PR #2826, opened 2015-03-10
Log output from cron jobs.

pmclanahan:


pmclanahan:
@jgmize r?

jgmize:
This looks like it should work as intended, and I know that's basically what we were doing before, but I wonder if perhaps it would be better to redirect a log file instead `/dev/null`?

pmclanahan:
I'd be in favor of that. I'll take a look at the logrotate config and see if we can just drop a file in /var/log and have it handled.

pmclanahan:
or, of course, get ops to configure logrotate for us for this.

pmclanahan:
@jgmize okay. now this should log every command's output to a log file in /var/log/bedrock per environment. I did not name the files based on date. I figured we could create a logrotate config to handle rotation and compression. I did add a job that will log the date hourly since many (most) of the commands don't output any date or time information. If you know of a better solution for this then we should do it.

bin/gen-crons.py
```diff
@@ -7,6 +7,7 @@
 
 HEADER = '!!AUTO-GENERATED!! Edit {template}.tmpl instead.'
 TEMPLATE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', 'etc', 'cron.d'))
+LOG_DIR = '/var/log/bedrock'
```
jgmize: I see that this dir already exists on the admin node, but perhaps we should still add a couple of lines to this script to create it if necessary?

pmclanahan: Not a bad plan. I created the folder manually myself just to make sure I was setting a real, usable, and sane default here. I'll add creation.

pmclanahan:
Updated.

jgmize:
Sorry @pmclanahan, but Travis doesn't like your latest update. :(

pmclanahan:
Ah yes, because we check the cron output and `gen-crons.py` can't create the `/var/log/bedrock` folder on Travis. Hmm...

pmclanahan:
Changes made @jgmize. Travis running again.

jgmize:
This is not a blocker, but one issue is that all the cron jobs in a given environment are now logging to the same file and in at least some cases I don't see a way of identifying which command the output in the file would be-- this is still a huge improvement over `/dev/null` though :)

pmclanahan:
It's true. The architecture could be improved. Is this worth going with you think, or should we do something better?

pmclanahan:
I think to get much better might be a decent investment of time. I'd be so happy if we had a good logging system for the running apps and all of this other stuff as well.

jgmize:
I think this is better than what we have now, and is fine to merge as-is, and we should follow up with a broader discussion about logging in the bedrock devops meeting today.
r+

superawesome:
Can you send us a bug on setting up logrotate (or something) for this?

PR #2828, opened 2015-03-11
WIP Bug 1121460, move all plugincheck client side code into bedrock

schalkneethling:
This is a WIP, do not merge. This will move all of the client side code for plugincheck into bedrock and also adds actual real life tests \o/ More to come.


cargoworker:

schalkneethling:
This is now ready for code review an testing. Note: This brings in a rather big change in that plugincheck will now only work in Fx - Once we have this new version out in the wild, we can reevaluate supporting other browsers as we go forward.

schalkneethling:
@alexgibson r?

alexgibson:
Do we still need to add a redirect for non-Firefox users? (I'm assuming to /firefox/new or similar)

If we are doing this redirect, we can probably remove the conditional content for non-Firefox users that [gets triggered here](https://github.com/mozilla/bedrock/pull/2828/files#diff-0232b38b609901500bb2c6db6389ebefR8)?

schalkneethling:
@alexgibson Thanks for the review! I guess it would be safest to push this to a demo server first. This would possible also be helpful in testing the redirect for nonFx users.

alexgibson:
Thanks, @schalkneethling - regarding the redirect for non-Firefox users. We discussed this briefly in our PR triage late last week. The bedrock back-end developers felt it might not be a good idea to do a server-side redirect on this page, as it is such high-traffic to do a vary on user agent. Doing a redirect on the client side would be better, but also not really a very good experience for the user.

It might be best to proceed with displaying a message / download button to non-Firefox users instead. This might have l10n implications though (unless there is a suitable string we can re-use somewhere). Happy to discuss this further if you need.

media/js/plugincheck/lib/plugincheck.js
```diff
@@ -0,0 +1,277 @@
+/*globals Utils */
+
+(function(exports) {
```
schalkneethling: hmmm, seeing it is already a global, much like Utils, perhaps just updating the JSHint bit above? or is there a benefit I am missing?

media/js/plugincheck/lib/plugincheck.js
```diff
@@ -0,0 +1,277 @@
...
+        } else if (Utils.isMatch(installedVersion, knownVersions.vulnerable)) {
+
+            var vulnerableInfo = Utils.isMatch(installedVersion, knownVersions.vulnerable);
+            plugin['status'] = 'vulnerable';
+            plugin['vulnerability_description'] = vulnerableInfo.vulnerability_description;
+            plugin['vulnerability_url'] = vulnerableInfo.vulnerability_url;
+
+            return plugin;
+
+        } else if (!(installedVersion > knownAbsLatest)) {
```
schalkneethling: should probably just use versionCompare here

media/js/plugincheck/lib/plugincheck.js
alexgibson: Either way is fine by me :+1:

media/js/plugincheck/lib/version-compare.js
```diff
@@ -0,0 +1,76 @@
...
+*    <li>0 if the versions are equal</li>
+*    <li>a negative integer iff v1 < v2</li>
+*    <li>a positive integer iff v1 > v2</li>
+*    <li>NaN if either version string is in the wrong format</li>
+* </ul>
+*
+* @copyright by Jon Papaioannou (["john", "papaioannou"].join(".") + "@gmail.com")
+* @license This function is in the public domain. Do what you want with it, no strings attached.
+*/
+function versionCompare(v1, v2, options) {
```
schalkneethling: Seeing I did not actually write that, I think it is better to leave it as a separate lib. Will add `'use strict';`

schalkneethling:
@alexgibson Yeah, that was one of the possibilities discussed with Jen but, in the end it they went with just doing the redirect. I am totally ok with displaying the Fx download button instead though.

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -239,11 +238,21 @@ <h1 class="large">{{_('Check Your Plugins')}}</h1>
...
+
+<!-- dislayed to nonFx users -->
+<div class="not-supported">
+  <header id="main-feature">
+      <h1 class="large">{{_('Check Your Plugins')}}</h1>
+      <p class="preamble">{{_('Keeping your plugins up to date helps Firefox run safely and smoothly.')}}</p>
+  </header>
+
+  <section>
+    <p>{{_('Currently the plugin check service is only available to Firefox users.')}}</p>
```
alexgibson: Is this one a new string? 

CC @flodolo - not sure what fallback options we have here on this one.

schalkneethling: Indeed, I do not think there is a current string we can use but, perhaps @flodolo knows. Also, I believe I need to wrap it in something more than {{_ so that it uses translations as soon as they become available?

flodolo: We don't have anything like that in strings across mozilla.org :-\

@schalkneethling Are you aware of other string changes? It's hard for me to check this PR given the amount of code involved.

flodolo: l10n_has_tag ([example](https://github.com/mozilla/bedrock/blob/e916f902ad708a31cff33092a29c40c9ada25ee2/bedrock/firefox/templates/firefox/new.html#L56))

schalkneethling: @flodolo nope, no other string changes. So, just no `else` for the l10n_has_tag? It did not exist before so, there is no else ;)

flodolo: Can the page live without any string there? Should we remove the entire section without the tag? Anyhow, the else is not needed if we don't have an alternative string.

schalkneethling: Would not be ideal without the string but also not a train smash I believe. I am also guessing languages such as German, French etc. will filter in fast so, it should be ok to go live without the string. Let me ask Jen.

flodolo: The alternative is to use the tag, but expose the string and wait for a decent coverage before merging.

jbertsch: I'm fine with either option:
a.  go live without the string
b.  expose the string and wait for decent coverage before merging
with a slight preference for option A

schalkneethling: I would prefer to go with option A - would really like to get this out the door ;)

schalkneethling: @flodolo so I am just going to wrap it with:

    `{% if l10n_has_tag('plugincheck_nosupport') %}
    ....
    {% endif %}

No else, and then the strings can filter in as they become available so, English will have the string and the button, and all others will just have the button initially.

schalkneethling:
@alexgibson feel free to give this one more pass when you have a moment. Thanks!

flodolo:
String wrapping looks good. It would be nice to have this on a demo server to check that all the strings are in there.

schalkneethling:
@flodolo it is on demo5

flodolo:
Strings look good, but shouldn't it display a warning with JS disabled?

schalkneethling:
@flodolo yeah, you mean if people visit the page in general with Js disabled? Do we have a string already I can use?

flodolo:
Weird, I thought we were already using it in plugincheck. We have ```Please turn on JavaScript to display this page correctly.```

schalkneethling:
@flodolo alrighty, taken care of. Thanks

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -239,11 +238,25 @@ <h1 class="large">{{_('Check Your Plugins')}}</h1>
...
+
+<!-- dislayed to nonFx users -->
+<div class="not-supported">
+  <header id="main-feature">
+      <h1 class="large">{{_('Check Your Plugins')}}</h1>
+      <p class="preamble">{{_('Keeping your plugins up to date helps Firefox run safely and smoothly.')}}</p>
+  </header>
+
+  <section>
+    <p>
```
alexgibson: Should this `<p>` be inside the l10n tag, else if the locale does not have the string translated, we would just put outputting an empty tag?

media/css/plugincheck/plugincheck.less
```diff
@@ -246,6 +245,48 @@ html[dir="rtl"] {
 .footer-newsletter-form {
     padding: 40px 20px;
 }
+.warning {
+    .span(8);
```
alexgibson: As this is a `<p>` element, you can probably just avoid setting a `width` on it entirely as it's block level and will fit the content of its parent. This will mean you don't need to set any additional media queries too :+1:

media/css/plugincheck/plugincheck.less
```diff
@@ -246,6 +245,48 @@ html[dir="rtl"] {
...
+        display: none;
+    }
+    .not-supported {
+        display: block;
+        #main-feature {
+            padding-bottom: 28px;
+        }
+        section {
+            background-color: #fff;
+            margin-bottom: 32px;
```
alexgibson: Setting this `margin-bottom` and `padding` on the line below to `@baseLine` will allow the download button to fit it's container a bit neater at smaller viewport sizes.

alexgibson:
This looks great :+1: Could of small nits to tidy up, but otherwise it's an r+wc from me code-wise.

Thanks @schalkneethling :tada:

bedrock/mozorg/templates/mozorg/plugincheck.html
schalkneethling: Ah yes, that makes more sense. I just followed the examples in the page blindly ;) difference of course is they have an else fallback. Nice catch.

schalkneethling:
Thanks @alexgibson and @flodolo - Just waiting on Travis and then merge time :woohoo:

PR #2831, opened 2015-03-12
Bug 1133010. Update Fx family nav.

jpetto:
- Add V2 for testing
- Update V1 copy & links
- Decrease vertical height


cargoworker:

jpetto:
Okay, this should be ready for review. A couple notes:

- I've tested in IE 8 & 9 and things seem functional
- The JS for V1 and V2 of the nav is similar, but kept separate as only one will live on after the test period.
- Speaking of V1 and V2, only V1 shows for mobile, as V2 has no top level clickable links.
- Still waiting on confirmation that GA is behaving as expected, but no need to delay review.

bedrock/base/templates/macros.html
```diff
@@ -452,3 +450,289 @@
...
+
+<div id="fxfamilynav-wrapper">
+
+{# Temporary holding place for optional CTA. Will be moved to proper place after nav version is decided. #}
+<div id="fxfamilynav-cta-stager">{% if caller %}{{ caller() }}{% endif %}</div>
+
+<header id="fxfamilynav-header" class="fxfamilynav-header{% if (navclass) %} {{ navclass }}{% endif %}">
+  <div class="container">
+    <i class="fxfamilynav-logo">{{ _('Firefox') }}</i>
+    <nav id="fxfamilynav" class="fxfamilynav"  role="navigation">
```
alexgibson: Nit: remove extra space before `role`

alexgibson:
Bit of a color contrast issue with the Tabzilla tab here :/

tabzilla-contrast

bedrock/firefox/templates/firefox/developer.html
```diff
@@ -211,4 +211,3 @@
   </div>
 </section>
 {% endblock %}
-
```
alexgibson: Is there a reason why this page doesn't have the new nav also?

bedrock/firefox/templates/firefox/interest-dashboard.html
```diff
@@ -88,7 +100,7 @@ <h3 data-accordion-role="tab">{{ _('How is my interest categorized?') }}</h3>
   <div data-accordion-role="tabpanel">
     <p>
       {% trans %}
-        The Interest Dashboard uses both a site-matched and keywords text analysis to categorize interest. The former is based on the categorization of certain URLs we have a high confidence of belonging to a particular interest (i.e. ESPN being Sports). The latter uses statistical analysis of keywords that appear within the title of a webpage mapped to keywords that appear on Wikipedia pages. Those pages are then summarized to IAB’s (Internet Advertising Bureau) content categories. 
+        The Interest Dashboard uses both a site-matched and keywords text analysis to categorize interest. The former is based on the categorization of certain URLs we have a high confidence of belonging to a particular interest (i.e. ESPN being Sports). The latter uses statistical analysis of keywords that appear within the title of a webpage mapped to keywords that appear on Wikipedia pages. Those pages are then summarized to IAB’s (Internet Advertising Bureau) content categories.
```
alexgibson: Not sure if this page is localized, but would this white space change get flagged as a string change @flodolo?

flodolo: Page not localized, no issue there.

alexgibson:
Some z-index scrolly fun happening on `firefox/partners`

partners-z-index

media/css/firefox/family-nav.less
```diff
@@ -9,30 +9,76 @@
...
-    font-size: inherit;
+// big left hand logo
+.fxfamilynav-logo {
+    float: left;
+    width: 140px;
+    height: 146px;
+    margin: 8px @baseLine 0 (@baseLine / 2);
+    .fxfamilynav-sprite(0);
+    .image-replaced();
+    .transition(transform 0.3s ease);
```
alexgibson: Think this might not work in Safari quite right, since `transform` will need to be `-webkit-transform`

media/js/firefox/family-nav.js
```diff
@@ -10,7 +10,40 @@ if (typeof window.Mozilla === 'undefined') {
...
+    // first look for forced nav from URL
+    if (window.location.href.indexOf('?v1=') !== -1) {
+        showV1 = 'yes';
+    } else if (window.location.href.indexOf('?v2=') !== -1) {
+        showV1 = 'no';
+    // then check sessionStorage
+    } else {
+        try {
+            showV1 = sessionStorage.getItem('fxfamilynavV1');
+        } catch (ex) {}
```
alexgibson: Need to set a value for `showV1` in the `catch` here?

alexgibson: Oh, I see below, never mind :smile:

alexgibson:
Nav links overlap on IE7

ie7

alexgibson:
This looks good to far :+1: I'll do another pass later once you've had some time to address comments so far :smile:

bedrock/firefox/templates/firefox/developer.html
jpetto: There sure is, I just don't happen to know it exactly. But yes, from above we were told not to put the nav on this page. ¯\_(ツ)_/¯

media/css/firefox/family-nav.less
jpetto: Ah, good catch. Added another transition mixin for `-webkit-transform`.

alexgibson:
Should also try and remember to remove Tabzilla assets from all these pages while we're at it? Think of all those KB's saved :D

jpetto:
@alexgibson Just pushed a commit that I believe addresses all issues. One more pass when you sign on tomorrow?

alexgibson:
Not necessarily a blocker, but I notice there's some hover glitchyness on the nav links in IE8?

alexgibson:
Worth preventing the default event action on the primary links, to avoid the `#` being added to the URL on click?

nav-hash-link

alexgibson:
Got some mobile styling issues with Tabzilla:

tabzilla-mobile

bedrock/base/templates/macros.html
```diff
@@ -452,3 +450,304 @@
...
+  </div>{#-- /.container --#}
+</header>{#-- /#fxfamilynav-header --#}
+
+<header id="fxfamilynavV2-header" class="fxfamilynav-header{% if (navclass) %} {{ navclass }}{% endif %}">
+  <div class="container">
+    <i class="fxfamilynav-logo">{{ _('Firefox') }}</i>
+    <nav id="fxfamilynavV2" class="fxfamilynav" role="navigation">
+      <ul id="fxfamilynavV2-primary" class="primarynav">
+        <li id="desktops">
+          <a href="#" class="primary-link" data-id="desktops"><div>{{ _('For desktops &amp; laptops') }}</div></a>
```
alexgibson: These are the only visible links on mobile, but don't go anywhere?

alexgibson:
Side note: overall UX observation on the new family nav - I sometimes find the hover state behavior a bit confusing, since it is detached from the actual active state shown in the UI? (i.e. the blue underline).

alexgibson:
A few things to tidy up, but overall this is looking good :+1: 

Main observations:

- Some general usability problems on older browsers (IE8 and below) with the hover states flickering.

- Keyboard accessibility is also still not quite there, but we can follow up once we settle on a final design for this new nav. It's something we should keep in mind though when designing/implementing the final version, as it does impact both accessibility and general usability.

bedrock/base/templates/macros.html
jpetto: Only V1 is visible on mobile (for this exact reason), regardless of what shows at tablet/desktop breakpoints.

alexgibson: Ah ok, so I should ignore if I resize the browser window and only look at what is there on page load, right? Just making sure I understand the logic.

jpetto: Resizing should always show V1 for mobile, regardless of what version you have for tablet/desktop. If you are seeing V2 ever for mobile, that's a bug.

alexgibson: Hmm, I could have sworn I hit one of these links on mobile - but now I can't reproduce :/

alexgibson: Perhaps I was confusing things, let's assume I'm incorrect for now :smile:

jpetto:
Okay, pretty close now I think:

- Mobile tabzilla issue was a regression - should be fixed now
- IE 8 flickering is better
- Preventing default when clicking on V2 primary nav items

In terms of the hover state being confusing, I totally agree. Result of coding the page and thinking through UX simultaneously (due to time constraints). We are planning an improvement phase as soon as a winner is chosen between V1 and V2, at which point we'll address usability/accessibility issues (among other things).

alexgibson:
r+ from me, latest changes look good and hover states in old browsers are working as expected :+1:

PR #2832, opened 2015-03-12
Bug 1142510 - Update credits FAQ to fix inaccuracies and bad URL

gerv:
index.html in SVN no longer gets updated per-checkin, so the URL to see recent checkins is wrong. Also a few textual updates.


pmclanahan:
Thanks!

PR #2835, opened 2015-03-13
Bug 1142583 Remove MWC from fx family nav

sgarrity:
and redirect /mwc/ to /firefox/partners/


alexgibson:
r+ from me. merge at will

PR #2837, opened 2015-03-13
Fix bug 1142850: Add cron job to update external files hourly.

pmclanahan:
Also alter the `gen-crons.py` script to have an option to disable creation of the log dir for local test runs. This still runs the check when testing while there is a value in the env var CRON_LOG_DIR, so Travis will still hit that branch of the code.

### DIFF
--- a/bin/gen-crons.py
+++ b/bin/gen-crons.py
@@ -7,7 +7,6 @@ from jinja2 import Template
 
 HEADER = '!!AUTO-GENERATED!! Edit {template}.tmpl instead.'
 TEMPLATE_DIR = os.path.abspath(os.path.join(os.path.dirname(__file__), '..', 'etc', 'cron.d'))
-LOG_DIR = os.getenv('CRON_LOG_DIR', '/var/log/bedrock')
 
 
 def main():
@@ -23,6 +22,8 @@ def main():
                             'Only define for cron.d style crontabs.'))
     parser.add_option('-p', '--python', default='python2.6',
                       help='Python interpreter to use.')
+    parser.add_option('-l', '--logdir',
+                      help='Directory in which to store logs from cron jobs.')
 
     (opts, args) = parser.parse_args()
 
@@ -33,11 +34,11 @@ def main():
         parser.error('-t must be defined')
 
     # ensure log path exists
-    if not os.path.isdir(LOG_DIR):
+    if opts.logdir and not os.path.isdir(opts.logdir):
         try:
-            os.mkdir(LOG_DIR)
+            os.mkdir(opts.logdir)
         except OSError:
-            parser.error('failed to create log directory: ' + LOG_DIR)
+            parser.error('failed to create log directory: ' + opts.logdir)
 
     log_file = 'cron-{0}.log'.format(opts.template.split('-')[1])
     django_manage = 'cd {{dir}} && {py} manage.py'.format(py=opts.python)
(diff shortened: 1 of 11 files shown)


pmclanahan:
@jgmize r?

bedrock/base/tests.py
```diff
@@ -21,8 +21,11 @@ class TestCrontabFiles(TestCase):
     @classmethod
     def setUpClass(cls):
         for cron_file in CRONTAB_FILE_NAMES:
-            call([path('bin', 'gen-crons.py'), '-t', cron_file, '-w', '/path/to/www',
-                  '-s', '/path/to/src'])
+            command = [path('bin', 'gen-crons.py'), '-t', cron_file, '-w', '/path/to/www',
+                       '-s', '/path/to/src']
+            if 'CRON_LOG_DIR' not in os.environ:
+                command.append('--nologdir')
```
jgmize: This feels like logic that belongs in `gen-crons.py` itself.

bin/gen-crons.py
```diff
@@ -23,6 +23,8 @@ def main():
                             'Only define for cron.d style crontabs.'))
     parser.add_option('-p', '--python', default='python2.6',
                       help='Python interpreter to use.')
+    parser.add_option('-n', '--nologdir', action='store_true',
```
jgmize: What if we just remove the default on line 10, and add the CRON_LOG_DIR to the real environment and leave it out in our testing environment?

jgmize:
I noticed that you've added the cron job to prod, but not dev and stage. Is this intentional, and if so why not?

bedrock/base/tests.py
pmclanahan: How would `gen-crons.py` know it's being run during a test?

bin/gen-crons.py
pmclanahan: Are you sure of how to inject it into the real environment? Should be able to do it in global.conf I think. That likely would be cleaner to only do the log stuff if a `LOG_DIR` is defined.

jgmize: We could change https://github.com/mozilla/bedrock/blob/master/bin/update/deploy_base.py#L213-L219 to something like the following:

```python
def generate_cron_file(ctx, tmpl_name):
    with ctx.lcd(settings.WWW_DIR):
        ctx.local("CRON_LOG_DIR={cron_log_dir} {python} bin/gen-crons.py "
                  "-p {python} -s {src_dir} -w {www_dir} -t {template}".format(
                    cron_log_dir=getattr(settings, 'CRON_LOG_DIR',
                                         '/var/log/bedrock'),
                    python=PYTHON,
                    src_dir=settings.SRC_DIR,
                    www_dir=settings.WWW_DIR,
                    template=tmpl_name))
```

pmclanahan: HA! global.conf. I'm clearly not thinking straight today.

So then we'd have to file a bug to get Ops to add the log dir to chief settings... meh. Doesn't seem that big a deal.  This is all because we want the app to ensure a directory we already know exists exists. Let's not file a bug for it.

pmclanahan: *HOWEVER*. We could in deployment pass in a `-l /var/log/bedrock` and that would do the check if provided. We could just not output the log stuff at all if it's not.

bin/gen-crons.py
```diff
@@ -23,6 +22,8 @@ def main():
                             'Only define for cron.d style crontabs.'))
     parser.add_option('-p', '--python', default='python2.6',
                       help='Python interpreter to use.')
+    parser.add_option('-l', '--logdir',
```
pmclanahan: @jgmize how about this?

jgmize: Yes, I like this even better :)

pmclanahan:
> I noticed that you've added the cron job to prod, but not dev and stage. Is this intentional, and if so why not?

A couple reasons:

1. I didn't want to adapt the shell script for the other two environments because lazy.
2. `update_externalfiles` is already called during deployment, which I figured is plenty good enough for non-prod environments.

jgmize:
I can definitely get behind not setting up the cron job on dev, but it would be really nice if we could validate the cron job works in the stage environment before deploying to prod...

pmclanahan:
Fair enough. I'll update shortly.

pmclanahan:
Updated.

jgmize:
Thanks @pmclanahan!

PR #2840, opened 2015-03-16
Bug 1133146: Add geolocation capabilities.

pmclanahan:
Add functions for country lookup and MaxMind DB handling.


pmclanahan:
@jgmize r?

pmclanahan:
@jgmize bump. Jake's got the DB file ready and needs this in the repo so the new compiled `maxminddb` lib can be installed in the venvs.

bedrock/base/geo.py
```diff
@@ -0,0 +1,50 @@
...
+
+from django.conf import settings
+
+try:
+    import maxminddb
+except ImportError:
+    maxminddb = None
+
+
+logger = logging.getLogger(__name__)
```
jgmize: FYI this causes a warning `No handlers could be found for logger "bedrock.base.geo"` for me locally

pmclanahan: Yeah. Our logging sucks. Just thought I'd add support for when it doesn't.
Can be removed for now if you'd prefer.

jgmize: Where do you expect the error logs to go in production?

bedrock/base/geo.py
```diff
@@ -0,0 +1,50 @@
...
+
+try:
+    if maxminddb:
```
jgmize: Why is the `if` inside the `try`, instead of the other way around?

pmclanahan: Could be. I think it's just how the thing grew. I can rearrange if it'd make more sense.

bedrock/base/geo.py
pmclanahan: No where at present. I just thought it better to be sending them than not.

jgmize: I'm a little concerned about the performance impact of the logging calls anyway, so if we don't have an immediate use for them I'd rather remove them.

bedrock/base/geo.py
jgmize: It's a tiny nit, but if you don't mind :)

pmclanahan:
@jgmize this is updated btw.

jgmize:
Thanks, this lgtm now :)

PR #2842, opened 2015-03-17
Fix cron log adding extra user.

pmclanahan:


pmclanahan:
@jgmize this should fix it. r?

jgmize:
Good catch, @pmclanahan

PR #2843, opened 2015-03-17
Remove dev cron jobs: fix bug 114384

jgmize:


jgmize:
These cron jobs should not be run because they are redundant to, and can interfere with, the dev deploy scripts: https://github.com/mozilla/bedrock/blob/master/bin/update/deploy_dev_base.py#L12-L20

jgmize:
@pmclanahan r?

PR #2849, opened 2015-03-18
Fix Bug 1144949 - Update "contribute to this page" link in footer

kyoshino:
Tested the redirect locally.


cargoworker:

craigcook:
Updating the link looks good to me, but I wonder if the text should change as well (which would be a string change, of course). Maybe "Contribute to this site"?

kyoshino:
Added the new string behind the l10n flag. This is a small but site-wide change. @flodolo Do you find any issue?

flodolo:
This won't work. The string is in a shared file, it doesn't really make sense to have tags on those (we spill them all over bedrock). 

Also, I'm looking at the bug and don't understand why we're changing this string.
> We can keep "Contribute to this page" string, so no l10n needed.

If you really need to do it, I'll have to add it to main.lang and then the PR will have to wait until we have a decent l10n coverage (like in old times).

kyoshino:
We have talked about the string in the weekly meeting and agreed it makes sense to change it. "Contribute to this page" is already in `main.lang` so can we put the `site_footer_201503` tag in that file?

flodolo:
As I said, I don't really want to start having tags in [shared files](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L65). Every single page in Bedrock will get that tag, and once we start it's way too easy to create confusion. I think we'd need to change our tools too, because shared files are not supposed to be activated, hence we don't have checks for tags.

I'm happy to add the string to main.lang, but it will have to wait, and I'm not sure the change of URL can. Put a conditional if lang starts with en-US?

craigcook:
We would like to change the string because "Contribute to this page" is confusing and misleading, but we didn't want to block merging this otherwise minor update until all locales have translated it. I didn't realize l10n flags in the footer would be problematic.

I'm not sure what the best solution here is. I suppose we can just block the PR after all and give locales a chance to update.

flodolo:
As said, if you want to push out the URL we can expose the new string only to en-* (using the old one for other locales), and then remove the conditional later when we have enough locales with that translation.

craigcook:
Yeah, an <code>if LANG.startswith 'en'</code> conditional would at least let us merge this now, at the small cost of leaving a future maintenance task to remove it later. @kyoshino, does that sound OK?

kyoshino:
Okay, I have just updated the code to use `if LANG.startswith('en')`.

craigcook:
r+! :dango:

PR #2850, opened 2015-03-19
New jenkins.sh WIP

jgmize:


jgmize:
progress so far: https://ci.masterfirefoxos.com/view/www.mozilla.org/job/mozilla_bedrock_github/129/console

jgmize:
@glogiotatidis FYI I'm temporarily using ci.masterfirefoxos.com for the purposes of developing some bedrock docker tests, but I'm hoping we have a more appropriate place for these to live shortly.

jgmize:
still more work to do, but I have a green run here: https://ci.masterfirefoxos.com/view/www.mozilla.org/job/mozilla_bedrock_github/130/console

glogiotatidis:
> @glogiotatidis FYI I'm temporarily using ci.masterfirefoxos.com for the purposes of developing some bedrock docker tests, but I'm hoping we have a more appropriate place for these to live shortly.

By all means. Let work towards making this a webprod wide CI server :)

bin/jenkins.sh
```diff
@@ -5,61 +5,44 @@
...
-. $VENV/bin/activate
-pip install -q -r requirements/compiled.txt
 pip install -q -r requirements/dev.txt
+rm -rf src  # clean up after pip fails to do so
+
+#TODO: get mysql working in new jenkins
+#echo "Creating database if we need it..."
+#echo "CREATE DATABASE IF NOT EXISTS \`${JOB_NAME}\`"|mysql -u $DB_USER -h $DB_HOST
+
+cp bedrock/settings/local.py-dist bedrock/settings/local.py
```
pmclanahan: Should just have a `settings/jenkins.py` and set or use a jenkins environment variable to auto load it like we did for Travis.

pmclanahan: Never mind. We already do that with the `JENKINS_HOME` variable. Shouldn't need to do anything with local.py.

pmclanahan: Main point is that we don't need this line though. Right?

jgmize: Without that line, I was getting an error about a missing local module. I believe that's because the `JENKINS_HOME` env var is not automatically passed in to the container running the tests. I could pass it in explicitly, but I'd like to eliminate settings/jenkins.py entirely and override specific settings with other env variables.

pmclanahan: Sure. But we're working with the codebase as it is now. Be a whole lot
easier for you to just set an env var, or change the env var it's looking
for to one that's always there for jeninks.

jgmize:
Closing this in favor of a different approach based on #3107

PR #2851, opened 2015-03-19
Fix bug 1145087 - Lightning version isn't being retrieved from AMO for en-US

kewisch:


cargoworker:

alexgibson:
Thanks @kewisch, r+

PR #2854, opened 2015-03-20
Move Clacks-Overhead handling to bedrock.

pmclanahan:
Also only include the header when response is a 200
and for a page.


pmclanahan:
@jgmize r?

pmclanahan:
@fwenzel fyi

PR #2855, opened 2015-03-20
[fix bug 1145558] Firefox family nav pages should not load Tabzilla assets

alexgibson:


cargoworker:

cargoworker:

PR #2857, opened 2015-03-21
Add rnasync cron job to dev

jgmize:


jgmize:
@pmclanahan r?

etc/cron.d/bedrock-dev.tmpl
```diff
@@ -7,6 +7,7 @@ MAILTO="webops-cron@mozilla.com,cron-bedrock@mozilla.com"
 # add date/time info to cron log
 0    * * * * {{ user }} date {{ log }}
 
+*/6  * * * * {{ django_manage }} rnasync {{ log }}
```
pmclanahan: Why do we need RNA data synced every 6 minutes on dev?

jgmize: The release managers use the dev server to preview release notes.

jgmize: https://github.com/mozilla/bedrock/blob/master/bedrock/releasenotes/views.py#L62

pmclanahan: Ah. Cool.

PR #2861, opened 2015-03-23
Bug 1108828 - Heartbeat pages

craigcook:
I totally kludged together the template view, though it seems to work for me locally. Would appreciate guidance from someone more knowledgeable... @jgmize?


craigcook:
Marking as WIP since I'm waiting for final signoff on some copy and design details (plus whatever fixes are likely needed in the backend bits).

bedrock/firefox/views.py
```diff
@@ -575,3 +575,14 @@ def hello(request):
...
+
+class FeedbackView(TemplateView):
+
+    def get_template_names(self):
+        rating = self.request.GET.get('rating', 0)
+        if rating > '3':
+            template = 'firefox/feedback/happy.html'
+        else:
+            template = 'firefox/feedback/unhappy.html'
+        return template
```
jgmize: Please return a list for this method:
```python
return [template]
```

jgmize:
@craigcook I sent https://github.com/craigcook/bedrock/pull/4 against your branch for this PR.

craigcook:
@jgmize thanks! Merged your PR and then squashed the commits so I think it lost your credit, but my gratitude will live on forever.

jgmize:
> @jgmize thanks! Merged your PR and then squashed the commits so I think it lost your credit, but my gratitude will live on forever.

No worries about squashing out the PR commit-- I only sent a PR b/c I thought it would be easier for you than copy-pasting from comments. I think we're good on the backend now, and the pages lgtm locally, but I'd appreciate it if @alexgibson or @jpetto would take a more detailed look at the frontend code.

jgmize:
@craigcook I sent https://github.com/craigcook/bedrock/pull/5 to address the flake8 issues.

bedrock/firefox/templates/firefox/feedback/happy.html
```diff
@@ -0,0 +1,75 @@
...
+    {% set share_urls = {
+      'twitter': 'http://mzl.la/1MZOYmZ',
+      'googleplus': 'https://www.mozilla.org/firefox/new/?utm_source=google&amp;utm_medium=social&amp;utm_content=happyheartbeat&amp;utm_campaign=desktop',
+      'facebook': 'http://mzl.la/1zmGNa6'
+    } %}
+    {{ share_cta(_('Share Firefox'), share_urls, _('Proud to use Firefox, the only non-profit, non-corporate, non-compromised browser. Join me:'), 'feedback-share', '') }}
+
+    <ul class="social-links">
+      <li><a href="https://twitter.com/firefox" class="twitter">{{ _('Follow us on Twitter') }}</a></li>
+      <li><a href="https://facebook.com/firefox" class="facebook">{{ _('Like us on Facebook') }}</a></li>
```
jpetto: `rel="external"`?

bedrock/firefox/templates/firefox/feedback/happy.html
```diff
@@ -0,0 +1,75 @@
...
+
+    <aside class="action-secondary cta-android">
+      <a data-name="CTA Firefox for Android" href="{{ url('firefox.sms') }}?utm_source=happyheartbeat&amp;utm_medium=heartbeat&amp;utm_content=AndroidCTA&amp;utm_campaign=FirefoxforAndroid">
+        <h2>{{ _('Get Firefox on your Android') }}</h2>
+        <p>{{ _('Send yourself a text to download Firefox for Android, a better mobile browser.') }}</p>
+      </a>
+    </aside>
+
+    <aside class="action-secondary cta-addons">
+      <a data-name="CTA Firefox Addons" href="https://addons.mozilla.org/firefox/?utm_source=happyheartbeat&amp;utm_medium=heartbeat&amp;utm_content=AddOnsCTA&amp;utm_campaign=AddOns">
```
jpetto: `rel="external"`?

(How concerned are we with marking as external?)

media/css/firefox/feedback.less
```diff
@@ -0,0 +1,271 @@
...
+    font-family: 'FA-Icons-Contact';
+    src: url('/media/fonts/icons-contact.eot?#iefix') format('embedded-opentype'),
+         url('/media/fonts/icons-contact.woff') format('woff'),
+         url('/media/fonts/icons-contact.ttf') format('truetype');
+    font-weight: normal;
+    font-style: normal;
+}
+
+main {
+    padding: 0 20px;
```
jpetto: `@baseLine`

media/css/firefox/feedback.less
```diff
@@ -0,0 +1,271 @@
...
+.mozilla-share-cta {
+    text-align: center;
+    margin: 0 auto;
+}
+
+.primary {
+    .clearfix;
+    .span-all;
+    margin-bottom: @baseLine;
+    padding: 20px;
```
jpetto: `@baseLine`

media/css/firefox/feedback.less
```diff
@@ -0,0 +1,271 @@
...
+    background: #fff;
+    border-bottom: 1px solid @borderColor;
+
+    .copy {
+        .font-size(16px);
```
jpetto: `@largeFontSize`

media/css/firefox/feedback.less
```diff
@@ -0,0 +1,271 @@
...
+    p {
+        margin: 0 auto 1.5em;
+        max-width: (@gridColumnWidth * 4) + (@gridGutterWidth * 3);
+    }
+}
+
+.action-secondary {
+    .border-box;
+    .span(4);
+    padding: 120px 20px 20px;
```
jpetto: `@baseLine` here and a few more places below.

media/js/firefox/feedback-ga-tracking.js
```diff
@@ -0,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+$(function () {
```
jpetto: Is there a reason to use the `document.ready` style here? If not, better to switch to the common IIFE style.

media/js/firefox/feedback-ga-tracking.js
```diff
@@ -0,0 +1,52 @@
...
+            callback = function() {
+                window.location = href;
+            };
+            trackGenericLink('CTA click', name, callback);
+        }
+    });
+
+    // track clicks in the SUMO links
+    $('.support-links a').on('click', function(e) {
+        var label = $(this).text();
```
jpetto: Looks like this should be called `name` (or `name` below should be changed to `label`).

media/js/firefox/feedback-ga-tracking.js
```diff
@@ -0,0 +1,52 @@
...
+    function trackGenericLink(action, href, callback) {
+        if (typeof callback === 'function') {
+            gaTrack(['_trackEvent', 'Feedback Interactions', action, href], callback);
+        } else {
+            gaTrack(['_trackEvent', 'Feedback Interactions', action, href]);
+        }
+    }
+
+    // track clicks in the secondary CTAs
+    $('.action-secondary a').on('click', function(e) {
```
jpetto: Small nit-pick on these event listeners - could be slightly more efficient to delegate, like:

`$('.action-secondary').on('click', 'a', function(e) {`

This creates a single event listener on the container, instead of an event listener on each `<a>` tag.

media/css/firefox/feedback.less
craigcook: (gonna go off on a tangent here so pardon the lengthy comment, maybe this should be a separate discussion elsewhere)

I'm not sure I like the idea of using <code>@baseLine</code> as a stand-in for <code>20px</code> every time we need a 20px value. The variable is named "baseline" because it's the common baseline for regular body text (14px with a 20px baseline). Keeping a standard baseline helps maintain a consistent vertical rhythm across the site. It really makes the most sense to use that variable for vertical spacing, especially with typography like the vertical space between paragraphs (which should really use em units proportional to the text anyway), or vertical space between two stacked boxes.

As with all variables, the point of defining <code>@baseLine</code> is that we can change it in one place and it will update everywhere it's used as a global control for vertical rhythm. I don't think changing that variable should change the padding on the left and right sides of this particular box.

BUT, if I'm going to use a variable here it would be <code>@gridGutterWidth</code>, which is more suited to horizontal spacing and also happens to be 20px so the end result is the same. So I think I'll do that. I could use <code>@baseLine</code> to pad the top and bottom of the box, except this box doesn't have top and bottom padding (but the next one does).

So as a general rule we should really only use <code>@baseLine</code> for vertical spacing where it won't cause width problems if its value ever changes. And sometimes 20px should just be 20px. We can use <code>@gridGutterWidth</code> for horizontal spacing, and if that one ever changes the <code>@gridColumnWidth</code> should also change accordingly. (Of course, so many pages have deviated from sandstone's grid system by now that I would dread the fallout of changing any of these variables at all... but that really is a separate discussion.)

media/css/firefox/feedback.less
craigcook: If we're using variables here it should probably be <code>padding: @baseLine @gridGutterWidth;</code> to keep vertical and horizontal spacing consistent with sandstone. They both happen to be 20px so the output is redundant, but the two variables might not always match.

jpetto:
r+ on the code, but looks like it needs a rebase?

craigcook:
rebased!

jpetto:
Hm, looks like the function `hello_screen_sharing` that's supposed to be in `firefox/views.py` got lost in the rebase.

craigcook:
Doh. The file conflicted, I botched the resolution. Fixing.

PR #2863, opened 2015-03-23
Fix Bug 1133780 - Update social sharing on firefox/sms

kyoshino:
Replacing the current Share widget with the new one. This page is en-US only, no l10n required.
https://www.mozilla.org/en-US/firefox/sms/sent/


alexgibson:
r+ :racehorse:

PR #2864, opened 2015-03-24
[fix bug 1137311] Updates to Hello FTU

alexgibson:
Pref to view this page locally: 

`loop.gettingStarted.url` = `http://127.0.0.1:8000/%LOCALE%/firefox/%VERSION%/hello/start/`

Must use a version >= 38.1 to see the new template


bedrock/firefox/templates/firefox/hello/start.html
```diff
@@ -61,6 +65,10 @@
     </section>
     <section class="end">
       <h2>{{ _('You’ve started your first conversation. Wasn’t that easy?') }}</h2>
+      {% if l10n_has_tag('hello_screen_sharing') %}
```
alexgibson: Note to self: still need to add a waffle flag here

alexgibson:
Marking this as blocked for now - waiting on confirmation on if this feature will land in 38.0 or 38.1 (which may require additional updates to this PR).

ckprice:
Confirming that Hello screen sharing is 38.1

-- 
Cory Price
/ckprice

alexgibson:
Ok this has been updated to reflect the new 38.1 launch target. 

@pmclanahan would appreciate some r on my the view logic changes, in particular [this function](https://github.com/mozilla/bedrock/pull/2864/files#diff-784e0d4de42af29595e70315a6be5337R575)?

bedrock/firefox/views.py
```diff
@@ -572,6 +572,27 @@ def hello(request):
         {'video_url': videos.get(request.locale, videos.get('en-US'))})
 
 
+def hello_screen_sharing(version):
+    try:
+        if version.endswith('a1') or version.endswith('a2'):
```
pmclanahan: Are we sure it'll only be `a1` or `a2`?

bedrock/firefox/templates/firefox/hello/start-38.1.html
```diff
@@ -0,0 +1,20 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "/firefox/hello/start.html" %}
```
pmclanahan: NIT: Technically it works, but paths in the `extends` tag shouldn't start with a `/`.

bedrock/firefox/views.py
alexgibson: This is the only thing I felt unsure on. Dev edition is `a2` and Nightly `a1`, so as long as those don't change this will be ok. I wasn't sure if there's a more robust way though...

pmclanahan: We could use a regex...

```python
if re.search('a\d$', version):
    ...
```

alexgibson:
Thanks @pmclanahan updated.

alexgibson:
@jpetto r?

jpetto:
LGTM. Screen sharing is nice! r+ :computer:

PR #2866, opened 2015-03-25
Bug 1133146 support multiple firefox os versions

alexgibson:
Still waiting on some additional countries being added, so **do not merge** for now. But this is good for code review!


pmclanahan:
@jgmize like Alex said, we're just waiting for some country clarification. Please r? the python bits when you get time.

bedrock/firefox/views.py
```diff
@@ -250,17 +249,15 @@ def all_downloads(request, channel):
...
-    lang_file = 'firefox/os/index-new'
-    old_home = 'firefox/os/index.html'
-    new_home = 'firefox/os/index-new.html'
+@never_cache
+def firefox_os_geo_redirect(request):
+    country = get_country_from_request(request)
+    try:
+        version = settings.FIREFOX_OS_COUNTRY_VERSIONS[country]
+    except KeyError:
+        version = settings.FIREFOX_OS_COUNTRY_VERSIONS['default']
```
jgmize: NIT: I personally prefer the `get` method with a default over exception handling. How about something like this instead?

```python
version = settings.FIREFOX_OS_COUNTRY_VERSIONS.get(
    country,
    settings.FIREFOX_OS_COUNTRY_VERSIONS['default'])
```

jgmize:
Other than one tiny nit which you can feel free to ignore, the python bits lgtm. :+1:

bedrock/mozorg/templates/mozorg/robots.txt
```diff
@@ -6,5 +6,9 @@ disallow: /*/download/
 disallow: /*/firstrun/
 disallow: /*/newsletter/existing/
 disallow: /*/whatsnew/
+disallow: /*/firefox/os/1.1/
```
alexgibson: @pmclanahan - added these to `robots.txt` as suggested. Just checking these look ok?

pmclanahan: I saw that. Looks good!

bedrock/firefox/views.py
pmclanahan: Yeah. I waffled on that myself. I'll try it the other way.

alexgibson:
@craigcook r?

alexgibson:
Also ping @flodolo - there are two new strings here to update the page title & description (again, sorry). 

And also [this string](https://github.com/mozilla/bedrock/pull/2866/files#diff-1df3cc28c78c6d328f3b94934753903fR19) is no longer hidden behind an l10n tag. Hopefully it's translated by now, not quite sure what else we could revert to in this case.

flodolo:
I will add the 2 new strings, it's ok to leave the other one not behind tag, it affects only 4 locales out of 21, and they should catch up.

alexgibson:
@pmclanahan - looks like some of the views/tests need a rebase. I can take a stab at it, but perhaps you might be best in case I miss something that got re-ordered?

pmclanahan:
@alexgibson I'll do the rebase now and force push the branch. You're not making further changes right?

alexgibson:
@pmclanahan - go for it, thanks! :+1:

pmclanahan:
Rebase done. Should be good to go! :8ball:

craigcook:
Looks good on the front-end. r+ :hamburger: 

Should we squash commits, or are they distinct enough to keep separate?

pmclanahan:
Commits are fine w/ me as they are.

PR #2867, opened 2015-03-25
Fix bug 1145763 - add Allison Banks to leadership page

craigcook:


alexgibson:
@flodolo says no need to wait on l10n, so r+ from me :pineapple:

PR #2868, opened 2015-03-25
Update Tabzilla document to recommend using https, not protocol relative URLs

kyoshino:
Part of [Bug 1147139](https://bugzilla.mozilla.org/show_bug.cgi?id=1147139). mozilla.org is no longer serving content via http.


pmclanahan:
Seems fine to me. Objections anyone? @jgmize ?

PR #2870, opened 2015-03-26
Update legal-docs submodule to 2175891

jgmize:


jgmize:
@pmclanahan r?

pmclanahan:
There's been another merge to master in legal-docs since. Should we go with that. Seems important for bedrock actually.

jgmize:
@pmclanahan updated

PR #2878, opened 2015-03-30
Fix Bug 1146474 - Update /en-US/firefox/sms/ With Legal Text & Link

kyoshino:
No l10n required. I would be nice to have it in today's push.

### DIFF
--- a/bedrock/firefox/templates/firefox/android/sms-send.html
+++ b/bedrock/firefox/templates/firefox/android/sms-send.html
@@ -21,7 +21,7 @@
       Send yourself a text message <br>to download it from Google Play.
     {% endtrans %}
     </h2>
-    <p class="note">{{ _('SMS service available in the U.S. only') }}</p>
+    <p class="note">{{ _('SMS service available in the U.S. only. SMS &amp; data rates may apply. The intended recipient of the SMS must have consented. <a href="%s">Learn more</a>')|format(url('privacy.notices.websites') + '#campaigns') }}</p>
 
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/android/sms-send.html
```diff
@@ -21,7 +21,7 @@ <h1 class="large">{{_('Get Firefox for Android')}}</h1>
       Send yourself a text message <br>to download it from Google Play.
     {% endtrans %}
     </h2>
-    <p class="note">{{ _('SMS service available in the U.S. only') }}</p>
+    <p class="note">{{ _('SMS service available in the U.S. only. SMS &amp; data rates may apply. The intended recipient of the SMS must have consented. <a href="%s">Learn more</a>')|format(url('privacy.notices.websites') + '#campaigns') }}</p>
```
alexgibson: Any reason to use `&amp;` here and not just `&`?

kyoshino: Looks like other strings are using ` &amp;` and `&` is not escaped somehow.

alexgibson: Hmm, seems to get escaped to me. But if other strings are using it then let's not hold up

kyoshino: I guess the escape issue is site-wide. See https://www.mozilla.org/en-US/firefox/37.0/whatsnew/

PR #2880, opened 2015-03-31
Fix Bug 1149620 - Firefox /all/ download page now has Windows stubs instead of full installers

kyoshino:


bedrock/firefox/firefox_details.py
```diff
@@ -120,7 +120,7 @@ def _get_filtered_builds(self, builds, channel, version=None, query=None):
                 build_info['platforms'][platform] = {
                     'download_url': self.get_download_url(channel, version,
                                                           platform, locale,
-                                                          True),
+                                                          True, True),
```
pmclanahan: Won't this mean that nothing gets stub?

kyoshino:
@pmclanahan r? 

I'll send a separate PR for mcom-tests.

bedrock/firefox/firefox_details.py
kyoshino: Yes, this is in `_get_filtered_builds` and the function is only for the /all/ pages. The download buttons are not affected.

pmclanahan: Ah yes! Of course your're right. Thanks.

pmclanahan:
If you have time I'd feel a lot better if there were a regression test with this. Is that what you plan to add to mcom-tests?

kyoshino:
Yes, I'll add something to mcom-tests.

PR #2884, opened 2015-04-01
Fix Bug 1150083 - Make Fx38 Win64 build of beta Available on moz.org

kyoshino:
Note: this PR changes the download button code as well for consistency, but the win64 buttons are being disabled by CSS for now. So the visible changes is /firefox/beta/all/ only as requested in the bug.


pmclanahan:
Nice work @kyoshino. Thanks! :+1:

PR #2885, opened 2015-04-01
Bug 1148127 - fix redirects for product page

craigcook:
Rules were in the wrong order

### DIFF
--- a/etc/httpd/global.conf
+++ b/etc/httpd/global.conf
@@ -64,7 +64,7 @@ RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?about/drivers(\.html|/)?$ https://wiki.mozil
 RewriteRule ^/community/(?:directory|wikis|blogs|websites)\.html$ https://wiki.mozilla.org/Websites/Directory [L,R=301]
 
 # bug 885856
-RewriteRule ^/projects/index\.(de|fr|hr|sq)\.html$ /$1/products/ [L,R=301]
+RewriteRule ^/projects/index\.(de|fr|hr|sq)\.html$ /$1/firefox/products/ [L,R=302]
 
 # bug 856075
 RewriteRule ^/projects/technologies\.html$ https://developer.mozilla.org/docs/Mozilla/Using_Mozilla_code_in_other_projects [L,R=301]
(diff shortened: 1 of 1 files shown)


stephendonner:
As discussed in IRC, might we get a redirect test or two (however many are needed) landed in https://github.com/mozilla/mcom-tests/blob/master/tests/test_redirect.py ?

jgmize:
@craigcook we need to delete [line 284](https://github.com/mozilla/bedrock/blob/master/etc/httpd/global.conf#L284) in global.conf as well.

etc/httpd/global.conf
```diff
@@ -64,7 +64,7 @@ RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?about/drivers(\.html|/)?$ https://wiki.mozil
 RewriteRule ^/community/(?:directory|wikis|blogs|websites)\.html$ https://wiki.mozilla.org/Websites/Directory [L,R=301]
 
 # bug 885856
-RewriteRule ^/projects/index\.(de|fr|hr|sq)\.html$ /$1/products/ [L,R=301]
+RewriteRule ^/projects/index\.(de|fr|hr|sq)\.html$ /$1firefox/products/ [L,R=302]
```
jgmize: missing `/` between `$1` and `firefox`

etc/httpd/global.conf
```diff
@@ -882,8 +881,9 @@ RewriteRule ^/(contributor-data/[a-z]+.json)$ /b/$1 [L,PT]
 # bug 1108828
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/feedback(/?.*)$ /b/$1firefox/feedback$2 [PT]
 
+# bug 1148127
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?products(/?)$ /$1firefox/products/ [L,R=302]
```
jgmize: FTR, this reordering wasn't actually necessary, just the [removal of line 284](https://github.com/mozilla/bedrock/pull/2885/files#diff-cc7713eb3a36ef9c90c2f1b99f8966e8L284) -- it doesn't hurt anything to reorder this though, so don't worry about undoing it :)

PR #2886, opened 2015-04-01
Fix Bug 1150131 - Make scene=2 check for auto-downloads a substring search on all URL parameters

kyoshino:


media/js/base/global.js
```diff
@@ -190,3 +190,45 @@ function gaTrack(eventArray, callback) {
         }
     }
 }
+
+
+// Retrieve search params as a object for easier access
+// This is a simple version of https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams
+var _SearchParams = function (search) {
```
alexgibson: I would like to try and get us out of the habit of adding more and more stuff into `global.js`. I don't think this will be used by that many pages, so I think it could probably be in just it's own file and included in the bundle for `/new`

media/js/base/global.js
```diff
@@ -190,3 +190,45 @@ function gaTrack(eventArray, callback) {
...
+
+_SearchParams.prototype.set = function (key, value) {
+    this.params[key] = !isNaN(value) ? parseInt(value) : value;
+};
+
+_SearchParams.prototype.has = function (key) {
+    return key in this.params;
+};
+
+_SearchParams.prototype.delete = function (key) {
```
alexgibson: `delete` is (IIRC) a reserved word and will likely make older browsers blow up when added as a method on the prototype chain.

media/js/base/global.js
```diff
@@ -190,3 +190,45 @@ function gaTrack(eventArray, callback) {
...
+    search = search.match(/^\??(.*)/)[1];
+    search = search ? search.split(/[&;]/m) : [];
+
+    $.each(search, function (index, param) {
+        param = param.split('=');
+
+        var key = param[0];
+        var value = param[1]
+
+        params[key] = !isNaN(value) ? parseInt(value) : value;
```
alexgibson: `parseInt` should specify a `radix`

media/js/base/global.js
```diff
@@ -190,3 +190,45 @@ function gaTrack(eventArray, callback) {
...
+    });
+};
+
+_SearchParams.prototype.get = function (key) {
+    return this.params[key];
+};
+
+_SearchParams.prototype.set = function (key, value) {
+    this.params[key] = !isNaN(value) ? parseInt(value) : value;
```
alexgibson: `parseInt` should specify a `radix`

media/js/firefox/new.js
```diff
@@ -235,13 +237,15 @@
                 gaTrack(['_trackPageview', virtual_url]);
             }
 
+            var params = new _SearchParams();
```
alexgibson: Any reason to call this multiple times in the one file? Could it just be called once at the beginning?

media/js/test/spec/global.js
```diff
@@ -377,4 +377,39 @@ describe('global.js', function() {
 
     });
 
+    describe('gaTrack', function () {
```
alexgibson: Nit: this should be `_SearchParams` or something similar

media/js/base/global.js
```diff
@@ -190,3 +190,45 @@ function gaTrack(eventArray, callback) {
...
+
+    search = search || location.search || '';
+    search = search.match(/^\??(.*)/)[1];
+    search = search ? search.split(/[&;]/m) : [];
+
+    $.each(search, function (index, param) {
+        param = param.split('=');
+
+        var key = param[0];
+        var value = param[1]
```
alexgibson: Missing semicolon

media/js/base/global.js
kyoshino: :+1: to move this to a separate file.

media/js/base/global.js
kyoshino: You are right. Maybe `remove` works.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Keywords

media/js/base/global.js
kyoshino: Ahh. I always forget it as modern browsers don't need it.

kyoshino: [`parseFloat`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseFloat) might be better..

alexgibson:
I've given this some testing locally and also pushed to a demo for good measure. Change looks good to me and working as expected, thanks @kyoshino r+ :four_leaf_clover:

kyoshino:
Thank you! :smiley_cat:

PR #2889, opened 2015-04-02
Fix Bug 1148557 - /en-US/credit page gives 404: Page Not Found

kyoshino:
A simple mistake. The cache is returning `3600` instead of names, leading to a TypeError.


jgmize:
Thanks @kyoshino!

PR #2890, opened 2015-04-03
Added a script for Bug#1117866

svatsan:
This script checks every icalendar file present in <bedrock>/media/caldata
and validates it. Currently, it would run every time a commit is done on master,
but that can be changed to just run on commits to .ics files.

There's also a failure in MexicanHolidays.ics, that I have left as-is for the author
to fix - this would be a good test to check/verify that all folks use the same file encoding.

If there are any more changes, please let me know


.travis.yml
```diff
@@ -19,6 +19,7 @@ before_script:
   - python manage.py collectstatic --noinput -v 0
   - python manage.py update_product_details
   - svn checkout https://svn.mozilla.org/projects/mozilla.com/trunk/locales/ locale
+  - python scripts/check_calendars.py
```
pmclanahan: I'd prefer this go above the `mysql` line just so it fails more quickly before all that other work is done.

scripts/check_calendars.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python
```
pmclanahan: Move this to the `bin` directory instead. The `scripts` dir is for scripts that are to be run via a management command.

scripts/check_calendars.py
```diff
@@ -0,0 +1,30 @@
...
+import os
+from icalendar import Calendar
+
+def get_ics(filename):
+    return filename.endswith('ics')
+
```
pmclanahan: 2 blank lines for pep8

scripts/check_calendars.py
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env python
+import os
+from icalendar import Calendar
+
```
pmclanahan: 2 blank lines

scripts/check_calendars.py
```diff
@@ -0,0 +1,30 @@
...
+        # parse. So we set multiple=True below
+        cal_entries = Calendar.from_ical(fh.read(), multiple=True)
+        if cal_entries is None:
+            raise ValueError
+    except ValueError as ve:
+        print ve.message
+        raise
+    finally:
+        fh.close()
+
```
pmclanahan: 2 blank lines.

also it would be best (though not required) to nest all of this uner a `if __name__ == '__main__':`

scripts/check_calendars.py
```diff
@@ -0,0 +1,30 @@
...
+calendars_dir = os.path.join('media','caldata')
+ics_files = map(lambda x: os.path.join(calendars_dir, x), filter(get_ics, os.listdir(calendars_dir)))
+for f in ics_files:
+    try:
+        check_if_correct_parse(f)
+    except ValueError:
+        print "failed to parse the icalendar file: ", f
+        exit(-1)
```
pmclanahan: Should be a positive integer right?

svatsan:
@pmclanahan All included in. About the `-1` in exit status, I was just following the old *nix convention in returning negative exit codes for errors that cannot be handled by program. Also, travis docs have this to say: 

> All commands must exit with code 0 on success. Anything else is considered failure.

That said, I don't have a problem in returning positive exit codes. I have now changed it to return 1 on failure. Let me know if everything's fine - I'll autosquash them later to a single commit.

svatsan:
@pmclanahan one more thing to mention: currently, this script bails out at the first sign of error. Will you be merging multiple calendar files at a time? Do you think it would be a good idea to bail out after reporting errors in all files?

pmclanahan:
Re: exit status: Cool. Didn't know that. Thanks.

I believe that we often do merge multiple files at once. So it would be nice to have a comprehensive list of errors to give to the author of the PR.

svatsan:
@pmclanahan Okay, I have added that facility to parse all files every time this script is kicked off and squashed all commits.

pmclanahan:
I thought Travis was failing because it was detecting an actual error in the calendar files, but this time it's an import error. This is because we do some path munging in `manage.py` so that imports of our git submodules work (we'll be making this better soon). For now the easiest fix is to move `check_calendars.py` back into the `scripts` dir and call it in the `.travis.yml` file as `python manage.py runscript check_calendars`. That will setup the correct environment before calling your script.

jgmize:
Thanks @svatsan, I rebased and merged this in 7d13ad85ed18bb6204f0218bfa36037dbcf3423e. I also tweaked the commit message to trigger our github/bugzilla integration automatically marking the bug as fixed. 
I made a recording of the steps I used here: https://asciinema.org/a/18726

PR #2891, opened 2015-04-03
Bug 1133146: Fix l10n on new versioned FxOS pages.

pmclanahan:


flodolo:
Looks good and works locally, thanks.

PR #2892, opened 2015-04-04
Fix /firefox/channel on android: fix bug 1151140

jgmize:


kyoshino:
Maybe regressed from my mass refactoring. Tested locally and r+. Thanks @jgmize!

jgmize:
thanks @kyoshino

PR #2893, opened 2015-04-06
Execute callback in tabzilla when GA isn't defined

freaktechnik:
Currently the language switcher will not work, if a site using tabzilla doesn't have Google Analytics, since the actual event action will never be executed.

### DIFF
--- a/bedrock/tabzilla/templates/tabzilla/tabzilla.js
+++ b/bedrock/tabzilla/templates/tabzilla/tabzilla.js
@@ -352,6 +352,10 @@ var Tabzilla = (function (Tabzilla) {
     Infobar.prototype.trackEvent = function (action, label, value,
                                              nonInteraction, callback) {
         if (typeof(_gaq) !== 'object') {
+            if (callback) {
+                callback();
+            }
+
             return;
         }
 


kyoshino:
@freaktechnik I guess this is from https://github.com/mozillacommunitych/mozilla.ch/commit/c5dd86645638bcef7183d5e5453579ddb07babd6 but mozilla.ch is really using the translation bar? It's an opt-in feature, so you can just disable it completely, though your PR seems to make sense.

http://bedrock.readthedocs.org/en/latest/tabzilla.html

jpetto:
Hi @freaktechnik! Is there a bugzilla bug associated with this change? If not, can you file one (just for tracking purposes)? Also, can you include the bug number in your branch name and commit message?

freaktechnik:
@kyoshino we are currently using it yes, even though there are issues due to the en-CH hreflangs.
@jpetto There currently isn't as far as I can see. Would I open it in the mozilla.org product, aka with the link from the readme?

jpetto:
@freaktechnik - Yep, product is `www.mozilla.org` and component should be `Pages & Content`. It would be great if you could link to an example of the bug (if one exists in the wild).

Thanks!

freaktechnik:
I'll close this pull for now, since the new commit will be from a different branch.

PR #2895, opened 2015-04-06
Fix Bug 1151538 - Execute callback in tabzilla when GA isn't defined

freaktechnik:
Currently the language switcher will not work, if a site using tabzilla doesn't have Google Analytics, since the actual event action will never be executed.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1151538

See also #2893


jpetto:
Looks good to me. Thanks for the excellent test case as well.

r+

PR #2897, opened 2015-04-07
Bug 1150567. Fix fx family nav text overlap.

jpetto:
- Fix overlapping when clicking V2 top level items
- Fix overlapping when hovering then tab focusing

### DIFF
--- a/media/css/firefox/family-nav.less
+++ b/media/css/firefox/family-nav.less
@@ -202,7 +202,8 @@
                 padding-right: 0;
             }
 
-            & > a {
+            & > a,
+            & > .primary-link {
                 display: block;
                 color: #fff;
                 padding: 0 5px;
@@ -237,7 +238,8 @@
     }
 
     .primarynav {
-        a {
+        .primary-link {
+            cursor: pointer; // to account for <span>'s used in V2
             opacity: 0.7;
             height: 67px;
             line-height: 67px;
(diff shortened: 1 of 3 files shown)


media/css/firefox/family-nav.less
```diff
@@ -237,7 +238,7 @@
     }
 
     .primarynav {
-        a {
+        .primary-link {
```
craigcook: Maybe add a <code>cursor: pointer;</code> for :hover? They're no longer links but the spans are still interactive and it feels weird to get the default insert cursor.

media/css/firefox/family-nav.less
```diff
@@ -785,4 +787,28 @@ html[lang^='en'] .fxfamilynav .subsubnav a {
...
+                }
+            }
+
+            .primary-link:focus + .subnav {
+                display: none;
+            }
+        }
+
+        .subnav.active {
+            display: block !important;
```
craigcook: Is this <code>!important</code> justified? I think it's needed to override inline styles applied by JS, correct?

media/css/firefox/family-nav.less
jpetto: Ah, yeah, good call. Will add.

media/css/firefox/family-nav.less
jpetto: It unfortunately is, as it needs to override the `display: none;` values directly above, which in turn are overriding the no JS `display` values set prior.

craigcook:
LGTM! r+wc! WTG FTW BBQ! :pig:

media/css/firefox/family-nav.less
craigcook: That's what I thought, just making sure it wasn't a leftover from testing or something.

PR #2898, opened 2015-04-08
Fix bug 1144669: Remove vendor submodules and use peep.

pmclanahan:
Use peep (a pip wrapper) for installation of dependencies
into the virtualenv already setup for deployment.


pmclanahan:
@jgmize it's here!!!!!!! Let's see what Travis does with it.

pmclanahan:
Notes: There are some notable upgrades. Jinja2 is now the latest version (2.5.5 -> 2.7.3). South is also latest, but the docs claim no backward compatible issues. 

I left many of the deps as github urls. This was for 2 reasons.

1. I was lazy and it was a massive pain to figure out what version corresponded to the particular hash we were using.
2. Some of the deps just didn't have a release at our version and upgrading presented problems out of the scope of this PR.

I also changed some things instead of adding dependencies. Like moving from the backport "ordereddict" lib to the one in the "collections" stdlib in 2.7. 

Another change is that MySQLpy is now only installed via the prod.txt requirements. This is because I've talked to a few people who had to install all of the mysql libs just to get that one to compile, and the weren't interested in running mysql anyway. I think this is the most common scenario since sqlite works great and is the default following the install instructions.

pmclanahan:
Wow. Huge success! Travis passed!

pmclanahan:
@jgmize I was thinking that funfactory removal would be too much for this PR and that I'd do that in a followup. What do you think?

jgmize:
> @jgmize I was thinking that funfactory removal would be too much for this PR and that I'd do that in a followup. What do you think?

Agreed, this is already pretty huge. This will take me a bit to go through in detail, but it's looking great so far. Nice work, @pmclanahan :+1:

.gitmodules
```diff
@@ -1,78 +1,3 @@
...
-[submodule "vendor-local/src/html5lib"]
-	path = vendor-local/src/html5lib
-	url = https://github.com/html5lib/html5lib-python
-[submodule "vendor-local/src/django-pipeline"]
-	path = vendor-local/src/django-pipeline
-	url = git://github.com/cyberdelia/django-pipeline.git
-[submodule "vendor-local/src/whitenoise"]
-	path = vendor-local/src/whitenoise
-	url = git://github.com/evansd/whitenoise.git
+	url = git://github.com/mozilla/legal-docs.git
```
jgmize: Shouldn't this remain an https url?

.travis.yml
```diff
@@ -23,7 +23,8 @@ script: coverage run manage.py test
 before_install:
   - git submodule update --init --recursive
 install:
-  - pip install -r requirements/dev.txt
+  - python bin/peep.py install -r requirements/dev.txt --no-use-wheel
+  - python bin/peep.py install -r requirements/prod.txt --no-use-wheel
```
jgmize: Why not just add `-r prod.txt` to `dev.txt`?

jgmize: Never mind, I realize now that most devs won't necessarily want or need those packages locally, and given the system requirements it makes sense to keep them separate.

bedrock/settings/base.py
```diff
@@ -2041,7 +2041,7 @@ def JINJA_CONFIG():
     'bedrock.mozorg.context_processors.facebook_locale',
     'bedrock.firefox.context_processors.latest_firefox_versions',
     'jingo_minify.helpers.build_ids',
-))
+), exclude=('session_csrf.context_processor',))
```
jgmize: What was the reason for this change?

bin/update/deploy_base.py
```diff
@@ -133,7 +141,10 @@ def update_info(ctx):
 def peep_install(ctx):
     """Install things using peep."""
     with ctx.lcd(settings.SRC_DIR):
-        ctx.local('../venv/bin/peep install -r requirements/compiled.txt')
+        ctx.local(peep_install_cmd('prod'))
+        # update any newly installed libs to be relocatable
+        # this call is idempotent
+        ctx.local('virtualenv-2.7 --relocatable ../venv')
```
jgmize: I understand this call is supposed to be idempotent, and I *hope* this works out for us, but deploying virtualenvs this way still makes me nervous. In the meantime, hopefully if we run into any issues with this they'll show up in dev or stage first...

jgmize:
r+wc -- I think the legal-docs submodule url change is the only thing I'd fix before merging at this point, assuming that it was unintentional.

FYI here's an export of the notes I took as I was reviewing this https://people.mozilla.org/~jmize/bedrock-peepify.html

Nice work!

bedrock/settings/base.py
pmclanahan: I did. I didn't add "session_csrf" as a dependency since we don't use it, but unless I exclude it here funfactory will helpfully add it to context processors which will try and fail to import it.

.gitmodules
pmclanahan: Sure. I removed it accidentally and I guess added it back differently. I'll update.

bin/update/deploy_base.py
pmclanahan: Yeah. We unfortunately don't have a choice at the moment unless we rsync the virutalenv separately to a path on the webheads that exactly matches the one on the admin node.

pmclanahan:
Some notes on your excellent notes doc:

1. Another good way to remove all of the extra dirs from vendor-local is to use `git clean -dff`. You can call `git clean -ndff` to see a dry-run of what it would do.

Otherwise I really like that doc. How'd you make it?

pmclanahan:
Fixed `s/git:/https:/` and rebased. Should be ready when you are.

jgmize:
>  Another good way to remove all of the extra dirs from vendor-local is to use git clean -dff. You can call git clean -ndff to see a dry-run of what it would do.

Thanks for the tip, but in my case I have [a few things in the dir that I didn't want to get rid of](https://people.mozilla.org/~jmize/bedrock-peepify.html#sec-2-3).

> Otherwise I really like that doc. How'd you make it?

I used a [slightly customized](https://github.com/jgmize/dotfiles/blob/master/.spacemacs#L120-L142) [org-mode](http://orgmode.org/). I linked to the [exported](http://orgmode.org/org.html#Exporting) version, but I also uploaded the `.org` file and linked to it at the bottom of the doc.

PR #2900, opened 2015-04-09
[fix bug 1152783] Missing 'else if' in conditional statement on hello fte

alexgibson:


jpetto:
Going to go out on a limb and give this an r+. :evergreen_tree:

PR #2901, opened 2015-04-09
Fix Bug 1123399 - Move Thunderbird all download page to Bedrock

kyoshino:


kyoshino:
Looks like tests are failing to override product details, so the latest production data is used instead of the test data. Still investigating why...

kyoshino:
I could eliminate some failures by replacing `@patch.dict` with `@patch.object` but I think that's not solving the underlying issue. @pmclanahan @jgmize Any ideas?

pmclanahan:
So confusing. Did something in this change cause this to stop working? I'm pretty sure these tests were passing before.

kyoshino:
Okay, tests passed! The Firefox tests require `ThunderbirdDesktop()` to prevent product data from being overridden unexpectedly.

pmclanahan:
That is really strange. I do not think those classes should share a data dict. Perhaps this is due to the way we're mocking something?

pmclanahan:
DOH! It's a [bug in django-mozilla-product-details](https://github.com/mozilla/django-product-details/blob/master/product_details/__init__.py#L42). Instantiates a mutable type at module load time and just reuses that object for everything. I don't think it's worth fixing for this. 

r+

kyoshino:
We probably need to reorganize django-product-details anyway once the Ship It dataset, including Thunderbird Beta and Earlybird, is available. :smiley_cat:

pmclanahan:
Most definitely.

pmclanahan:
I just need to take a look at the page locally then I'll merge.

bedrock/thunderbird/templates/thunderbird/all.html
```diff
@@ -0,0 +1,126 @@
+{% extends "thunderbird/base-resp.html" %}
```
pmclanahan: This template seems to share a lot with `firefox/all.html`. Is it possible that we could make a useful base template for all of these pages?

kyoshino: I should have moved those macros to the base macros template. Will do.

kyoshino: Moved shared stuff to a new macro template.

pmclanahan:
Great work @kyoshino ! :+1:

etc/httpd/global.conf
```diff
@@ -884,3 +884,6 @@ RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?products(/?)$ /$1firefox/products/ [L,R=302]
 # bug 1133004
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/products(/?)$ /b/$1firefox/products$2 [PT]
 
+# bug 1123399
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/all(\.html)?$ /$1thunderbird/all/ [L,R=301]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/all(/?)$ /b/$1thunderbird/all$3 [PT]
```
pmclanahan: One last thing. I think the `$3` on this line should be `$2`.

etc/httpd/global.conf
```diff
@@ -884,3 +884,6 @@ RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?products(/?)$ /$1firefox/products/ [L,R=302]
 # bug 1133004
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/products(/?)$ /b/$1firefox/products$2 [PT]
 
+# bug 1123399
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/all(\.html)?$ /$1thunderbird/all/ [L,R=301]
```
pmclanahan: I think you can make the `.html` non-optional in this one. `thunderbird/all` will be caught by the next line.

pmclanahan:
Just a couple things with the global.conf changes and we can merge.

jgmize:
After this merges, we should also add some coverage to mcom-tests

kyoshino:
@pmclanahan Thanks, fixed the rest of the issues in global.conf.
@jgmize I'll send a PR for mcom-tests.

stephendonner:
:+1: for adding mcom-tests!

PR #2904, opened 2015-04-09
Update some php_url links

kyoshino:
While looking at [Bug 1146683](https://bugzilla.mozilla.org/show_bug.cgi?id=1146683) I have noticed that some links using `php_url` can be updated. The rest of the links are /MPL/ and /thunderbird/, but as I mentioned in the bug, this function could be retired now.


ckprice:
Pulled this down and verified all links are working r+ thanks Kohei!

PR #2905, opened 2015-04-10
[fix bug 1153282] Add GA Event To What's New 37 SMS Submit Button

alexgibson:


bensternthal:
R+ tested locally and event fires with correct data.

alexgibson:
Thanks @bensternthal

PR #2908, opened 2015-04-13
Fix bug 1149335 - Add FxOS device for South Africa

craigcook:


flodolo:
Not adding l10n tag to the PR, string already added and imported from Firefox codebase.

schalkneethling:
good is good and tested locally r+

jgmize:
thanks @craigcook, @flodolo, and @schalkneethling

PR #2909, opened 2015-04-13
Fix bug 1152159 - Add Patriot Act Section 215 home promo

craigcook:
Uses the waffle switch "patriot-act-home-promo"


jpetto:
LGTM. r+ :corn:

PR #2910, opened 2015-04-13
Bug 1154033. Add temp survey to Fx family tertiary nav.

jpetto:


media/css/firefox/family-nav.less
```diff
@@ -817,3 +817,29 @@ html[lang^='en'] .fxfamilynav .subsubnav a {
...
+.survey {
+    padding-bottom: 4px;
+    margin: 0 -10px 6px;
+
+    a {
+        display: block;
+        padding: 4px 10px;
+        background: #eee;
+
+        i {
```
craigcook: Might this work just as well with <code>:before</code> instead of an empty element in the HTML?

jpetto: Ah, agreed. Updating.

craigcook:
r+ :hamburger:

PR #2911, opened 2015-04-15
Fix Bug 1112277, stop using jsonp with plugincheck

schalkneethling:
@alexgibson r?

### DIFF
--- a/media/js/plugincheck/check-plugins.js
+++ b/media/js/plugincheck/check-plugins.js
@@ -163,7 +163,7 @@ $(function() {
 
     // only execute the plugincheck code if this is Firefox
     if (isFirefox()) {
-        PluginCheck.getPluginsStatus('https://plugins.mozilla.org/en-us/plugins_list.json?callback=?',
+        PluginCheck.getPluginsStatus('https://plugins.mozilla.org/en-us/plugins_list.json',
             function(response) {
                 displayPlugins(response);
             }


schalkneethling:
This depends on the work that needs to be done here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1112277#c20

To test this locally, change the URL in:
https://github.com/mozilla/bedrock/pull/2911/files#diff-0232b38b609901500bb2c6db6389ebefR166

to point to plugins.allizom.org as that has a * ACAO header.

schalkneethling:
Since https://bugzilla.mozilla.org/show_bug.cgi?id=1112277#c28 both plugins stage and prod has an ACAO of '*' so, let the testing begin. @alexgibson

alexgibson:
Locally this works for me against plugins.allizom.org, but I get an XHR error when testing against prod?

schalkneethling:
@alexgibson What is the error? cross origin blocking?

alexgibson:
"xhr" Object { readyState: 0, getResponseHeader: .ajax/v.getResponseHeader(), getAllResponseHeaders: .ajax/v.getAllResponseHeaders(), setRequestHeader: .ajax/v.setRequestHeader(), overrideMimeType: .ajax/v.overrideMimeType(), statusCode: .ajax/v.statusCode(), abort: .ajax/v.abort(), state: .Deferred/d.state(), always: .Deferred/d.always(), then: .Deferred/d.then(), 11 more… } plugincheck.js:272:17
    "status" "error" plugincheck.js:273:17
    "error" ""

    Syntax Error: JSON.parse: unexpected end of line data at line 1 column 1 of the JSON data

schalkneethling:
@alexgibson ok, so this really works now on prod and stage ;)
see: https://bugzilla.mozilla.org/show_bug.cgi?id=1112277#c32

alexgibson:
@schalkneethling - this is working for me locally now against prod r+ 

Good to merge?

schalkneethling:
Indeed, let's merge!

PR #2915, opened 2015-04-16
Add Dockerfile and docker-compose.yml

jgmize:


jgmize:
Notes on progress so far: https://people.mozilla.org/~jmize/bedrock-docker-dev.html

jgmize:
@pmclanahan this is still WIP but what I have so far is working for me locally. If you (or anyone else) would like to try checking this out and testing it locally, just [install Docker](https://docs.docker.com/installation/#installation), [install Compose](https://docs.docker.com/compose/install/), check out this branch, run `docker-compose up`, and once it's finished you should be able to see bedrock running on http://localhost:8000

pmclanahan:
Awesome! Trying it.

jgmize:
Update for anyone following along: it works on @pmclanahan's machine, but not on localhost because he's using boo2docker on OS X. He connects to `$(boot2docker ip):8000` instead.

jgmize:
While this works great on Linux, there are significant performance issues using virtualbox shared folders with the boot2docker solution on other systems. @pmclanahan  and I discussed implementing a gulp or grunt task that rsyncs to the boot2docker VM for local development to alleviate some of the performance issues, but that will be a separate PR.

glogiotatidis:
> While this works great on Linux, there are significant performance issues using virtualbox shared folders with the boot2docker solution on other systems.

I believe that this is due to the `vendor(-local)` directories hosting thousand of files. But if we're pip installing everything on the docker image we don't need to have the vendor dirs. Did you try removing those and check if the performance issues go away?

jgmize:
>>    While this works great on Linux, there are significant performance issues using virtualbox shared folders with the boot2docker solution on other systems.

> I believe that this is due to the vendor(-local) directories hosting thousand of files. But if we're pip installing everything on the docker image we don't need to have the vendor dirs. Did you try removing those and check if the performance issues go away?

Yes, we already removed vendor, and there is only one submodule in vendor-local at this point, but we still have a lot of files:

```sh
$ for d in $(ls -d */); do count=$(find $d -type f | wc -l); echo $count $d; done | sort -rn
11512 locale/
4508 static/
3744 media/
1076 node_modules/
871 bedrock/
787 mofo_security_advisories/
477 vendor-local/
161 lib/
94 docs/
35 bin/
8 etc/
7 scripts/
7 puppet/
4 requirements/
3 root_files/
2 wsgi/
```

jgmize:
@pmclanahan (and/or anyone else who wants to test this) I've updated this with a new script, which has a dependency on [fswatch](https://github.com/emcrisostomo/fswatch)-- since you already installed docker and docker-compose, you should (hopefully) be able to `brew install fswatch` and run `bin/b2d-dev` and (eventually) have a running environment. Please test and let me know.

pmclanahan:
awesome. testing now.

jgmize:
@pmclanahan I updated this to attempt to automatically install fswatch (and brew) on os x if needed.

jgmize:
It should now automatically install docker-compose if needed, as well.

jgmize:
ok @pmclanahan I've done some additional refactoring and taken an initial stab at the docs. Please test it out when you can.

@bensternthal you mentioned being interested in testing this-- give it a shot and let me know how it goes

@glogiotatidis if you get a chance to test the linux version of the docs I'd appreciate it, too

docs/docker.rst
```diff
@@ -0,0 +1,71 @@
...
+    $ bin/b2d-grunt
+
+Linux
+-----
+
+Build the dev docker image::
+
+    $ docker-compose build dev
+
+Run the ``bin/sync_all`` script inside a dev docker container::
```
glogiotatidis: Before this step I had to rename `local.py-dist` to `local.py`

glogiotatidis:
Tried on linux and it works fine, nice work @jgmize ! :star: 
Notes:
 - The `local.py-dist` setup as I commented above
 - I noticed that locales other than `en-US` redirect to `en-US/<locale>`. So requesting `/fr/` redirects to `/en-US/fr`.

docs/docker.rst
jgmize: Thanks, fixing that now.

jgmize:
Thanks again @glogiotatidis, I've added steps to clone the locales repo and copy the local settings file to the docs.

bensternthal:
I was able to get this to work with the following issues:

* initial docker-compose up stopped with error in below screenshot
* ran again and it worked
* ran b2d-dev initially and did not work
* ran b2d-dev again and it worked
* from the docs it's unclear the relationship between b2d-dev and docker-compose up
* unclear how to shut down container started by b2d-dev

screen shot 2015-05-22 at 11 30 36 am

I will test more later.

jgmize:
@bensternthal thanks for the testing and feedback. I wrote `b2d-dev` and the other `b2d-*` scripts to simplify things for OS X users so that you never need to call `docker-compose` directly. You can if you want, of course, but the scripts set up the environment, ensure boot2docker is running, etc for you, and just running `docker-compose up` will not work in OS X unless you basically run the equivalent commands first. The section of the docs that have the `docker-compose` commands are targeted at Linux users, who don't really need wrapper scripts beyond what is provided by docker-compose already, since there isn't a VM layer. 

To shut down the b2d container, hit ctrl-C in the terminal window where the script is running. To stop the VM, run `boot2docker stop`

I'll work on updating the docs to clarify this after I get done reviewing https://github.com/mozilla/bedrock/pull/2919.

groovecoder:
I hit this:
```
+  bin/b2d-dev :   18 > docker-compose run b2d bash -c 'if [ ! -e .synced ]; then bin/sync_all && touch .synced; fi'
+ '[' '!' -e ./manage.py ']'
                            + ./manage.py migrate --noinput
                                                           Traceback (most recent call last):
  File "./manage.py", line 10, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 354, in execute
    django.setup()
  File "/usr/local/lib/python2.7/dist-packages/django/__init__.py", line 21, in setup
    apps.populate(settings.INSTALLED_APPS)
  File "/usr/local/lib/python2.7/dist-packages/django/apps/registry.py", line 85, in populate
    app_config = AppConfig.create(entry)
  File "/usr/local/lib/python2.7/dist-packages/django/apps/config.py", line 87, in create
    module = import_module(entry)
  File "/usr/lib/python2.7/importlib/__init__.py", line 37, in import_module
    __import__(name)
  File "/usr/local/lib/python2.7/dist-packages/django_statsd/__init__.py", line 1, in <module>
    from django_statsd import patches
ImportError: cannot import name patches
+  bin/b2d-dev :   18 > trap - SIGTERM
+  bin/b2d-dev :   18 > kill -- -38185
Terminated: 15
```

jgmize:
@groovecoder did you start from a fresh checkout of bedrock, or did you possibly have an old checkout? I'm asking because I haven't been able to replicate your issue, and I think you might have an old `bedrock/settings/local.py` -- would you mind deleting it and running the `bin/b2d-dev` script again?

groovecoder:
I had an old checkout. I'll re-clone and try again.

docs/docker.rst
```diff
@@ -0,0 +1,86 @@
...
+    $ git svn clone https://svn.mozilla.org/projects/mozilla.com/trunk/locales/ locale
+    # or
+    $ svn checkout https://svn.mozilla.org/projects/mozilla.com/trunk/locales/ locale
+
+OS X
+----
+
+Launch a dev server container in a boot2docker VM::
+
+    $ bin/b2d-dev
```
groovecoder: After this step, I sat at this `Attaching to bedrock_b2d_1` message for a long time. Maybe should add a bit to the docs that says something like, "If everything works, you'll see `Attaching to bedrock_b2d_1`. then you can ..."

groovecoder:
The only issues I hit were:

1. I had an old `bedrock/settings/local.py` causing my `ImportError: cannot import name patches` error

2. I had switched networks between `b2d-dev` attempts, causing `CommandError: Problem connecting to Nucleus: ('Connection aborted.', gaierror(-2, 'Name or service not known'))`

Maybe turn these into a "Troubleshooting" section in the docs? E.g., MDN has an ["Errors during vagrant up"](https://kuma.readthedocs.org/en/latest/installation-vagrant.html#errors) section, and a broader ["Troubleshooting"](https://kuma.readthedocs.org/en/latest/troubleshooting.html) section.

jgmize:
We've expanded the scope of the docker dev environment to also include the apache and php aspects of the site, so it makes more sense to follow the pattern of a separate "env" repo, similar to https://github.com/mozilla/marketplace-env/ and https://github.com/mozilla/payments-env, and I'm closing this PR in favor of that approach.

PR #2920, opened 2015-04-21
[fix bug 1155830] Enable Optimizely on Firefox 37 whatsnew page

alexgibson:
Added waffle switch `firefox-whatsnew-optimizely`


alexgibson:
This one would be good to review & push today so GarethC can do a quick test on some SMS messaging.

PR #2922, opened 2015-04-21
Add mozID .less file/style for modal widget.

jpetto:
- Standardize indentation in mozilla-modal.js

According to Ty, this will be the standard UI going forward.

Desktop styles should be solid. Have not tested responsive styles (currently in copy/pasta state from `mozilla-modal.less`) as my instance doesn't need to be responsive.

@alexgibson r?


alexgibson:
Close button is a little cropped off-screen for me

close-button

alexgibson:
There's a couple of IE-specific filter property styles being inherited from sandstone which need over riding on the close button.

ie-close-button

media/css/base/mozid-modal.less
```diff
@@ -0,0 +1,152 @@
...
+    background: #000;
+    background: rgba(0, 0, 0, 0.85);
+    width: 100%;
+    height: 101%;
+    position: fixed;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    z-index: 9999999;
```
alexgibson: I know this is just copy/pasta from the old one, but do you think we should add another digit on here, just to be extra safe? :trollface:

media/js/base/mozilla-modal.js
```diff
@@ -1,152 +1,152 @@
...
+        var title = (options && options.title) ? options.title : '';
+
+        var $modal = $(
+            '<div id="modal" role="dialog" aria-labelledby="' + origin.getAttribute('id') + '" tabindex="-1">' +
+            '  <div class="window">' +
+            '    <div class="inner">' +
+            '      <header>' + title + '</header>' +
+            '      <div id="modal-close">' +
+            '        <a href="#close-modal" class="modal-close-text"> ' + closeText + '</a>' +
+            '        <button type="button" class="button">×</button>' +
```
alexgibson: I know this is technical debt, but would be good to include some proper text here and use an icon instead of `x`?

alexgibson:
Other than a couple of nits/oddities with the close button, this looks good :+1: 

I do wonder if we should implement this on an existing page when it goes into master, to test the water as such? It would be good to get those responsive styles down in one fell swoop. What do you think?

alexgibson:
P.S - for reference I tried testing the new styles on /firefox/hello/ and /firefox/os/

media/css/base/mozid-modal.less
jpetto: I mean, a couple more 9's couldn't hurt, right?

jpetto:
Good call on implementing on an existing page. I've got it running on `/firefox/hello/`. I tested in IE 8 - 10, Fx, Chrome, and Safari and all is looking good.

media/css/base/mozid-modal.less
```diff
@@ -0,0 +1,158 @@
...
+
+                a {
+                    // TODO: remove <a> element from mozilla-modal.js with mozID rollout
+                    display: none;
+                }
+
+                .button {
+                    background-color: transparent;
+                    // png fallback for IE8
+                    background-image: url("data:image/png;base64, iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAQAAABKfvVzAAAAhElEQVR4AY3Rsa0DMRDE0N0CrtAPh1Yv1kHF8kebDQFhUr5oij9+NHWxZvEUG9gXpDnAt2he4KUv8kMXTkJOFU5CPiCQlA8IRPIBgUg+IJCQDxAS8wGBhHyAkJgbOGC/WH7sF8vbrrRcr/RciOdCPBfiuRDPhXgupFiSZ7KLhxXyTDaff714nz5yfKKNAAAAAElFTkSuQmCC");
```
alexgibson: Are we OK relying on base64 here? Think this might cause really old browsers (e.g. IE7 and below) to throw an insecure content warning, iirc.

media/js/base/mozilla-modal.js
alexgibson: Now we're using an image could we get rid of the `x`, or is it still needed?

jpetto: It's still needed for all the pages using the old modal UI. :unamused:

alexgibson: Ah I see, ok carry on :smile:

media/css/base/mozid-modal.less
jpetto: Ugh, yeah, you are right. How long does a browser released in 2006 dictate terms? 9 years and counting. Will update.

alexgibson: I know :(

If this was just one page I wouldn't even have brought it up. But seeing as this could appear on any page I thought we should probably consider it still here.

alexgibson:
r+ :banana:

PR #2925, opened 2015-04-22
Add strings for iOS download button, part of Bug 1139961

kyoshino:
We need to expose those strings to meet the l10n string freeze on April 23. The logic behind the button is coming later.

### DIFF
--- a/bedrock/firefox/templates/firefox/includes/download-button.html
+++ b/bedrock/firefox/templates/firefox/includes/download-button.html
@@ -64,6 +64,9 @@
               {% if channel != 'alpha' %}
                 <span class="download-subtitle">{{ _('Get it free on Google Play') }}</span>
               {% endif %}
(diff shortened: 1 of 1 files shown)


kyoshino:
Cc @flodolo

bedrock/firefox/templates/firefox/includes/download-button.html
```diff
@@ -64,6 +64,9 @@
               {% if channel != 'alpha' %}
                 <span class="download-subtitle">{{ _('Get it free on Google Play') }}</span>
               {% endif %}
+            {% elif plat.os == 'ios' %}
```
craigcook: Let's add <code>waffle.switch('ios-active')</code> (after the platform check) so there's no risk of this being exposed in production until iOS actually launches.

kyoshino:
Thanks @craigcook, added `waffle.switch` in case.

PR #2928, opened 2015-04-23
Marks outdated plugins as outdated not vulnerable

schalkneethling:
@jgmize @alexgibson r?

### DIFF
--- a/bedrock/mozorg/templates/mozorg/plugincheck.html
+++ b/bedrock/mozorg/templates/mozorg/plugincheck.html
@@ -58,6 +58,9 @@
   {% if l10n_has_tag('plugin_outdated') -%}
     {# L10n: This is the string used to indicate that the plugin is outdated #}
     data-outdated="{{ _('outdated') }}"
+  {% else %}
+    {# L10n: This is the string used to indicate that the plugin is outdated #}
+    data-outdated="{{ _('vulnerable') }}"
   {% endif %}
 
   {# L10n: This is the string used to indicate that the plugin is vulnerable #}
--- a/bedrock/mozorg/templates/mozorg/plugincheck.html
+++ b/bedrock/mozorg/templates/mozorg/plugincheck.html
@@ -55,7 +55,7 @@
   {# L10n: This is the string used to indicate that the plugin is unknown #}
   data-unknown="{{ _('unknown') }}"
 
-  {% if l10n_has_tag('share_cta') -%}
+  {% if l10n_has_tag('plugin_outdated') -%}
     {# L10n: This is the string used to indicate that the plugin is outdated #}
     data-outdated="{{ _('outdated') }}"
   {% endif %}
--- a/bedrock/mozorg/templates/mozorg/plugincheck.html
+++ b/bedrock/mozorg/templates/mozorg/plugincheck.html
@@ -55,6 +55,11 @@
   {# L10n: This is the string used to indicate that the plugin is unknown #}
   data-unknown="{{ _('unknown') }}"
 
+  {% if l10n_has_tag('share_cta') -%}
+    {# L10n: This is the string used to indicate that the plugin is outdated #}
+    data-outdated="{{ _('outdated') }}"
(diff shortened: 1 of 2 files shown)


bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -55,6 +55,11 @@
   {# L10n: This is the string used to indicate that the plugin is unknown #}
   data-unknown="{{ _('unknown') }}"
 
+  {% if l10n_has_tag('plugin_outdated') -%}
+    {# L10n: This is the string used to indicate that the plugin is outdated #}
+    data-outdated="{{ _('outdated') }}"
+  {% endif %}
```
alexgibson: Do you need an else here, or locales that don't have the translation will be undefined?

schalkneethling: yeah, I was thinking about that. There was nothing before though so not sure what I would put in the else, unless I can do something like:

    data-outdated="{{_('vulnerable')}}"

but I very much doubt that, that will work, besides it being confusing ;) And suggestions?

alexgibson: I'm not sure what the best approach is here, but the translation needs to fall back to something. Agree it seems confusing :/

schalkneethling: @flodolo any suggestions here?

flodolo: How often are we going to show this string? If it's something we risk to display often, I think we need to live with either the confusing "vulnerable", or maybe "unknown".

schalkneethling: This can definitely be shown frequently, so I would go with the confusing outdate = "vulnerable" as unknown means something completely different in this context. Thanks @flodolo

schalkneethling:
@alexgibson @flodolo updated the l10n tag with an else clause.

alexgibson:
Code changes look fine to me, I think this is probably the best option we have. Just hope it doesn't lead to confusing / inaccurate bug reports? :/

r+ from me, care to squarsh commits @schalkneethling?

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -55,6 +55,14 @@
   {# L10n: This is the string used to indicate that the plugin is unknown #}
   data-unknown="{{ _('unknown') }}"
 
+  {% if l10n_has_tag('plugin_outdated') -%}
+    {# L10n: This is the string used to indicate that the plugin is outdated #}
+    data-outdated="{{ _('outdated') }}"
+  {% else %}
+    {# L10n: This is the string used to indicate that the plugin is outdated #}
```
alexgibson: Actually I wonder if we could use a better l10n comment here, to explain a bit clearer? Throughs @flodolo?

flodolo: Yes, definitely confusing. "This string is used as fallback in case 'outdated' is not translated yet.".

I won't expose this to localizers anyway (files are manually managed and cleaned).

schalkneethling: I definitely is confusing but the only fallback we have for right now. Are we good to merge with this as is? I will squash and merge once I get the 'ok' - Thanks everyone!

alexgibson: I think if we just update this comment with @flodolo's suggestion we're good to squash & merge.

alexgibson: Also is there a bug number we can attach to this?

schalkneethling: Let me create one.

schalkneethling:
This is fixed and merged in https://github.com/mozilla/bedrock/pull/2934

PR #2929, opened 2015-04-24
[no bug] Add static tabzilla tab assets

alexgibson:
This is not intended as a fix for [bug 1151579](https://bugzilla.mozilla.org/show_bug.cgi?id=1151579), but it seems most, if not all, new pages being worked on for spring campaign feature the newer Tabzilla word mark. I think most of us are not planning on implementing the full Tabzilla on newly designed pages.

There are already several occurrences of this asset appearing on individual pages in production, e.g.

https://github.com/mozilla/bedrock/blob/master/media/img/firefox/family/nav-sprite-high-res5.png
https://github.com/mozilla/bedrock/blob/master/media/img/firefox/family/index/family-sprite-high-res.png

Instead of creating more versions of this tab throughout the site, I think it might be a good idea to start using a central style sheet & directory for these new assets. It will make updating things easier in the near future, and we can add to this style sheet with more styles & colors for different designs. Eventually (e.g. after spring campaign) we could apply this to all bedrock pages and then move what we do here to the Tabzilla repo.

Thoughts?


jpetto:
Great idea. Wondering how we could best include the [`:before` content](https://github.com/mozilla/bedrock/blob/master/media/css/firefox/family-nav.less#L76) that allows a background color to be placed below the transparent wordmark. Docs and Less comments?

Also wondering if we should have a mozID sprite that contains the new tabzilla image. It looks like we're going to be using a white mozilla wordmark in the footer of mozID-style pages, and we have the new close icons for mozID style modals.

Does it make sense to create a "mozID" sprite/Less file for these common elements?

We could talk this over during the Monday triage perhaps?

alexgibson:
I think the mozID sprite should probably remain separate to Tabzilla, since Tabzilla will ultimately get used on more than just mozorg. Good suggestion for the :before content, I'm thinking we can probably just add a base style for this here and then let page/theme styles in sandstone set the background colour as needed. Other sites could do this too pretty easy.

Let's discuss this further on Monday, sounds good :+1:

alexgibson:
Added a new block to the base template so we can now do:

    {% block tabzilla_tab %}
      <div id="tabzilla">
        <a href="{{ url('mozorg.home') }}">Mozilla</a>
      </div>
    {% endblock %}

Individual pages can change the wordmark background color using:

    #tabzilla:before {
        background-color: purple;
    }

Each page can then override the unused asset blocks:

    {% block tabzilla_css %}{% endblock %}
    {% block tabzilla_js %}{% endblock %}

alexgibson:
Rebased and updated, thanks @jpetto! Think we should be good to merge?

jpetto:
I'd say we're good to merge, yeah. Travis error looks unrelated...

jpetto:
Nuts to you, Travis! :v:

PR #2930, opened 2015-04-24
[fix bug 1158119] Update Hello FTU to show screen sharing for 38.0.5

alexgibson:
Updating to show Hello FTU screen sharing for 38.0.5 instead of 38.1


alexgibson:
Setup instructions to test this locally:

1.) make sure you are set up to run UITour locally:
http://bedrock.readthedocs.org/en/latest/uitour.html#local-development

2.) in about config, set the following pref:
`loop.gettingStarted.url` = `http://127.0.0.1:8000/%LOCALE%/firefox/%VERSION%/hello/start/`

You can then visit the page by clicking the Hello icon and selecting "Tour" from the gear menu.

To verify this bug, you should just need to confirm the start-38.0.5.html gets loaded if your version of Firefox is greater than than number (for example, try testing in Dev Edition).

craigcook:
r+ :monkey:

PR #2931, opened 2015-04-24
Bug 1136912. Dev edition updates.

jpetto:


jpetto:
Should now be ready for review.

bedrock/firefox/views.py
```diff
@@ -22,6 +22,7 @@
 from lib import l10n_utils
 from lib.l10n_utils.dotlang import _
 from product_details.version_compare import Version
+import waffle
```
alexgibson: Would it be a worthwhile optimization to do `from waffle import switch_is_active` here? Not sure how much difference it makes, but perhaps @pmclanahan or @jgmize might have a preference.

bedrock/firefox/views.py
```diff
@@ -334,6 +335,7 @@ def firefox_partners(request):
         request, 'firefox/partners/index.html', 'firefox.partners.index', template_vars, form_kwargs)
 
 
+# Note: this function is also used to determine display for dev-whatsnew
```
alexgibson: Would it be worthwhile renaming this function to something more generic, to avoid confusion in the whatsnew view itself?

media/css/firefox/dev-firstrun-spring-campaign.less
```diff
@@ -0,0 +1,251 @@
...
+    position: relative;
+    width: @widthDesktop - (@gridGutterWidth * 2);
+    margin: 0 auto;
+}
+
+// Main intro
+
+.intro {
+    padding-top: @baseLine;
+    padding-bottom: @baseLine*3;
```
alexgibson: Tiny nit: spaces either side of `*` here and below

alexgibson:
Any reason why `media/js/firefox/dev-firstrun-spring-campaign.js` is pretty much a copy of the code in `media/js/firefox/dev-firstrun.js`? Seems like a lot of duplication, could they both just use the same file?

Also, `media/js/firefox/dev-firstrun.js` seems to have GTM stuff in, which this newer file doesn't.

media/js/firefox/dev-whatsnew.js
```diff
@@ -0,0 +1,96 @@
...
+                videoId: $video.data('video-id'),
+                events: {
+                    'onReady': onPlayerReady,
+                    'onStateChange': onPlayerStateChange
+                }
+            });
+
+            function onPlayerReady(event) {
+                event.target.playVideo();
+                gaTrack(['_trackEvent', 'Developer /whatsnew/ Interactions', 'play', videoTitle + ' video']);
```
alexgibson: `gaTrack` needs replacing with GTM

media/js/firefox/dev-whatsnew.js
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            function onPlayerStateChange(event) {
+                if (event.data === YT.PlayerState.ENDED) {
+                    gaTrack(['_trackEvent', 'Developer /whatsnew/ Interactions', 'finish', videoTitle + ' video']);
```
alexgibson: `gaTrack` needs replacing with GTM

media/js/firefox/dev-whatsnew.js
```diff
@@ -0,0 +1,96 @@
...
+
+    var tag = document.createElement('script');
+    tag.src = 'https://www.youtube.com/iframe_api';
+    var firstScriptTag = document.getElementsByTagName('script')[0];
+    firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
+
+    var trackClick = function (gaArgs, href, event) {
+        if (event.metaKey || event.ctrlKey) {
+            // Open link in new tab
+            gaTrack(gaArgs);
```
alexgibson: Same here

media/js/firefox/dev-whatsnew.js
```diff
@@ -0,0 +1,96 @@
...
+        } else {
+            event.preventDefault();
+            gaTrack(gaArgs, function() { window.location = href; });
```
alexgibson: same here `gaTrack`

alexgibson:
This is looking great - just a few things to tidy up / update :+1:

alexgibson:
Also, should we add some tests for the view logic?

bedrock/firefox/views.py
jpetto: Sounds logical to me. :+1:

bedrock/firefox/views.py
jpetto: Good idea. I'll rename to `show_devbrowser_firstrun_or_whatsnew`.

jpetto:
Ugh, yeah, so the duplicate/near-duplicate `*-spring-campaign.[less|js]` files. As requirements have been less that static recently, and since we had to waffle the `firstrun` template, I preemptively added them in the even that the behavior/styles needed to be significantly different. Since we're now at req freeze and the files are almost identical, I will drop the new `*-spring-campaign.` versions in favor of sharing the existing assets across both `firstrun` templates.

bedrock/firefox/views.py
jpetto: Trying to move to `from waffle import switch_is_active` gave me problems when writing tests/mocks, so leaving as `import waffle`.

bedrock/firefox/templates/firefox/dev-whatsnew.html
```diff
@@ -0,0 +1,112 @@
...
+        <h2>{{_('Performance Tools')}}</h2>
+        <p>{{_('Examine your web content’s frame rate, memory consumption and JS profile in a consolidated timeline view to help easily identify and troubleshoot performance bottlenecks.')}}</p>
+        <a href="https://developer.mozilla.org/docs/Tools/Performance" rel="external" class="more">{{_('Learn more about Performance Tools')}}</a>
+        <div class="responsive-video-container">
+          <div class="video" data-video-id="eH0R10Ga4Hs"></div>
+        </div>
+      </li>
+      {% if LANG.startswith('en-') %}
+        <li class="feature">
+          <a href="https://hacks.mozilla.org/?p=28808" rel="external">
```
alexgibson: hacks blog post is a 404?

bedrock/settings/static_media.py
```diff
@@ -1165,6 +1165,14 @@
         ),
         'output_filename': 'js/firefox_developer_firstrun-bundle.js',
     },
+    # will be needed when videos are ready (shortly after 6/2?)
```
alexgibson: Not sure I quite follow why this is commented out - isn't this bundle needed for locales who don't have the `l10n_has_tag('spring_release')` tag?

alexgibson:
Video thumbnails on the /whatsnew page look a little smaller than other pages, plus the play icon is a little misaligned:

whatsnew-thumbs

bedrock/firefox/templates/firefox/dev-whatsnew.html
```diff
@@ -0,0 +1,112 @@
...
+            <img src="{{ static('img/firefox/firstrun/dev/feature-whatwefixed.png') }}" alt="{{ _('Screenshot') }}" class="screenshot">
+          </a>
+          <h2>{{_('What we fixed')}}</h2>
+          <p>
+            {{ _('When we released Firefox Developer Edition, we asked for your feedback.') }}
+            {{ _('We received tons of great ideas, many of which have been implemented just in the last month.') }}
+          </p>
+          <a href="https://hacks.mozilla.org/?p=28808" rel="external" class="more">{{_('See how you helped')}}</a>
```
alexgibson: Just a note this hacks URL needs updating also

alexgibson:
Usually, we also make /firstrun also available at the /firefox/tour URL, should anyone want to see it again (Help -> Firefox Tour). Do we want to do that here?

bedrock/settings/static_media.py
jpetto: Ah! Good catch. It will be needed until the `dev-edition-spring-campaign` waffle switch is activated.

jpetto: No, wait. This bundle is not needed at the moment.

Videos did not come in for the features promoted on `/firefox/40.0a2/whatsnew/`, so we don't need any page-specific JS running. The videos will likely come in at some point, and when they do, we'll have this bundle ready.

bedrock/firefox/templates/firefox/dev-whatsnew.html
jpetto: [According to the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1159932#c15), this URL will be live on 6/2.

bedrock/settings/static_media.py
alexgibson: Hmm, but there are videos already in the /whatsnew template, for locales that don't have the flag `l10n_has_tag('spring_release')`, or am I misunderstanding?

jpetto: Sheesh, no, you are correct. I got lost in the conditions. Enabling the bundle.

jpetto:
The video thumbnails on `/whatsnew` are a bit small for non spring campaign URLs, but new assets weren't easy to get produced, so we rolled with the images we had.

Fixed the play icon alignment though.

alexgibson:
r+ from me, care to squarsh?

jpetto:
Squished!

PR #2933, opened 2015-04-24
Bug 1136459. Add iOS to fx family nav.

jpetto:
- Add "Apps" to template for L10n (may replace "Marketplace" in nav)

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -583,6 +583,16 @@ set share_urls = {
             <li><a href="{{ url('firefox.sync') }}" data-id="android-sync"><div>{{ _('Sync') }}</div></a></li>
           </ul>
         </li>
+        {% if waffle.switch('ios-active') %}
+        <li>
+          <a href="{{ url('firefox') }}" class="primary-link" data-id="ios"><div>{{ _('<span>for</span> iOS') }}</div></a>
+
+          <ul class="subnav" id="ios-subnav">
+            <li><a href="{{ url('firefox') }}" data-id="ios-index"><div>{{ _('Overview') }}</div></a></li>
+            <li><a href="{{ url('firefox.sync') }}" data-id="ios-sync"><div>{{ _('Sync') }}</div></a></li>
+          </ul>
+        </li>
+        {% endif %}
         <li>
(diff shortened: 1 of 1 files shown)


bedrock/base/templates/macros.html
```diff
@@ -584,6 +584,14 @@
           </ul>
         </li>
         <li>
+          <a href="{{ url('firefox') }}" class="primary-link" data-id="ios"><div>{{ _('<span>for</span> iOS') }}</div></a>
```
craigcook: Should probably wrap this whole thing in an <code>{% if waffle.switch('ios-active') %}</code>

bedrock/base/templates/macros.html
```diff
@@ -591,6 +599,7 @@
             <li><a href="{{ url('firefox.os.devices') }}" data-id="os-devices"><div>{{ _('Devices') }}</div></a></li>
             <li><a href="{{ url('firefox.partners.index') }}" data-id="os-partners"><div>{{ _('For partners') }}</div></a></li>
             <li><a rel="external" href="https://marketplace.firefox.com/" data-id="os-marketplace"><div>{{ _('Marketplace') }}<i class="icon-external"></i></div></a></li>
+            <li><a rel="external" href="https://marketplace.firefox.com/" data-id="os-apps"><div>{{ _('Apps') }}<i class="icon-external"></i></div></a></li>
```
craigcook: Since this is a string change do we need to tag it? Do we already have "Apps" as a string somewhere?

flodolo: The navigation system is in a shared file (main.lang), so we simply need to hope that it will be localized in time (had the same issue with the other strings).

jpetto:
Closing until Fx iOS is back on the schedule.

flodolo:
Just a side note: since we don't know when iOS will come out, I suggest to keep this PR in mind and use as many strings as possible from it (if not all of them).

Unfortunately I can't remove these strings now, they've been exposed for weeks to localization.

jpetto:
Unless the entire nav changes, I'm almost positive we'll use "for iOS" in the nav when the product ships. I heard rumblings that we will *not* be replacing "Marketplace" with "Apps" (meaning it would remain "Marketplace"), but I don't know if that's definite yet.

PR #2934, opened 2015-04-25
Marks outdated plugins as outdated not vulnerable

schalkneethling:
@alexgibson Fixed the branch name to point to the bug so, merging here and will close the previous pull request.


flodolo:
I always thought the important part for the bug reference is the commit message, not the branch?

schalkneethling:
@flodolo yeah probably, but ah well, did both ;)

PR #2936, opened 2015-04-28
[no bug] Add blue rays background style to sandstone

alexgibson:
Adding a shared style I need to use on a couple of upcoming pages, so make sense for this to go into master.


alexgibson:
Closing this as @jpetto has the :moneybag:

PR #2937, opened 2015-04-28
Fix Bug 1153588, support SeaMonkey on plugin check

schalkneethling:
@alexgibson r?

### DIFF
--- a/media/js/base/global.js
+++ b/media/js/base/global.js
@@ -114,6 +114,12 @@ function isFirefoxUpToDate(latest) {
     return (latestFirefoxVersion <= fx_version);
 }
 
+// Used on the plugincheck page to also support SeaMonkey
+function isSeaMonkey(userAgent) {
+    var ua = userAgent || navigator.userAgent;
+    return /SeaMonkey/i.test(ua);
+}
+
 // used in bedrock for desktop specific checks like `isFirefox() && !isFirefoxMobile()`
 // reference https://developer.mozilla.org/en-US/docs/Gecko_user_agent_string_reference
 function isFirefoxMobile(userAgent) {
--- a/media/js/test/spec/global.js
+++ b/media/js/test/spec/global.js
@@ -320,6 +320,51 @@ describe('global.js', function() {
         });
     });
 
+    describe('isSeaMonkey', function() {
+
+        it('should consider SeaMonkey as SeaMonkey', function() {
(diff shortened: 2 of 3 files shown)


media/js/test/spec/global.js
```diff
@@ -320,6 +320,51 @@ describe('global.js', function() {
...
+
+        it('should consider SeaMoneky as SeaMonkey', function() {
+            var ua = 'Mozilla/5.0 (Windows NT 6.1; WOW64; rv:40.0) Gecko/20100101 Firefox/40.0 SeaMonkey/2.37a1';
+            var result = isSeaMonkey(ua);
+            expect(result).toBeTruthy();
+        });
+
+        it('should not consider Firefox to be SeaMonkey', function() {
+            var ua = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10.8; rv:23.0) Gecko/20100101 Firefox/23.0';
+            var result = isFirefox(ua);
```
alexgibson: This should be testing `isSeaMonkey()`, right? Same for occurrences below.

schalkneethling: Oooooh ouch, copy/paste #fail

media/js/base/global.js
```diff
@@ -98,6 +98,11 @@ function isFirefox(userAgent) {
     );
 }
 
+function isSeaMonkey(userAgent) {
+    var ua = userAgent || navigator.userAgent;
+    return ((/SeaMonkey/i).test(ua));
```
alexgibson: Nit: all the brackets might not be required here: could this just be `return /SeaMonkey/i.test(ua);`?

alexgibson:
This looks good to me, thanks @schalkneethling. I think the function `isSeaMonkey` is probably ok to leave in `global.js` for now. It is likely pretty uncommon.

In the near future I would like to factor out all these UA detection methods into their own lib for consumption in bedrock, instead of hanging them off global scope everywhere. But this is a larger change to the code base and probably outside the scope of this bug. 

I think this is fine for the time being. Thanks and r+wc

media/js/test/spec/global.js
```diff
@@ -320,6 +320,51 @@ describe('global.js', function() {
         });
     });
 
+    describe('isSeaMonkey', function() {
+
+        it('should consider SeaMoneky as SeaMonkey', function() {
```
stephendonner: Uber (unsolicited) nit: s/SeaMoneky/SeaMonkey

schalkneethling: bugger, nice catch @stephendonner ;)

PR #2940, opened 2015-04-29
Add aurora-bars mixin.

jpetto:
- Use mixin on firefox family landing page.

### DIFF
--- a/media/css/firefox/family/index.less
+++ b/media/css/firefox/family/index.less
@@ -4,19 +4,12 @@
 
 @import "../../sandstone/lib.less";
 
-// potential general styles for new mozilla id themed pages
-@mozIDBlue: #0c99d5;
-
 .mozid {
     color: #fff;
 
     #outer-wrapper {
-        background: @mozIDBlue url(/media/img/firefox/family/index/aurora.png) right top no-repeat;
+        .aurora-bars();
         border-top: 4px solid #fff;
-
-        @media only screen and (max-width: @breakTablet) {
-            background-position: -200px -100px;
-        }
     }
 
     #wrapper {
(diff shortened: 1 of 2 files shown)


jpetto:
IE <= 9 will just get a solid background color, which I think is acceptable. I'd rather not have each page include a fallback stylesheet just for an image file of these bars for IE <= 9.

Thoughts?

jpetto:
@alexgibson r?

alexgibson:
I'm totally fine with IE <= 9 just having a solid background. this is what progressive enhancement is for after all :smile: 

Everything about this looks great, except for Safari, which seems to render exactly the same standard syntax very differently from every other browser :cry: 

safari

jpetto:
After a quick test, it seems Safari has problems with `transparent` used as a color stop. Good job, Safari.

I had a little luck replacing `transparent` with `rgba(255, 255, 255, 0)`, though Safari decides to render a noticeable bit of white even with a 0% opacity.

I'll dig in deeper tomorrow.

:disappointed:

craigcook:
> I had a little luck replacing transparent with rgba(255, 255, 255, 0), though Safari decides to render a noticeable bit of white even with a 0% opacity.

I've run into this before with transparent gradients. <code>rgba(255, 255, 255, 0)</code> is transparent white, and I think the default value for the <code>transparent</code> keyword is <code>rgba(0, 0, 0, 0)</code>, or transparent black (since "transparent" isn't really a color, it's a color with a zero alpha channel). You need to blend the opaque blue into the transparent version of itself to get a correct opaque-to-transparent gradient. I think you can use a LESS function for this: <code>fade(@auroraStartColor, 0%)</code> and it should convert hex to rgba.

jpetto:
Excellent call @craigcook. Tried your technique and all seems well, even in Safari.

Odd that some browsers interpret `transparent` as you'd logically expect (revealing the color underneath), while Safari treats it as a 0% alpha black.

alexgibson:
This looks great! r+ :tangerine:

PR #2950, opened 2015-05-04
Fix Bug 1150328, implement wom landing page

schalkneethling:
@alexgibson @craigcook Initial version, opening as a WIP, feel free to review the code though.


media/css/mozorg/contribute/friends.less
```diff
@@ -0,0 +1,95 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../../sandstone/lib.less";
+
+* {
+    .border-box();
```
alexgibson: We don't really have a policy on the default box-model used in bedrock per-say. I know many dev's do use this as modern default today, but I do wonder if we should maintain some sort of consistency across pages on bedrock. We do need to support older browsers on certain pages, and it can be a bit of a mind-shift swapping from one box modal to another depending upon the page.

Any thoughts on this on @craigcook or @jpetto ?

schalkneethling: @alexgibson It seems that we should be good on this in general, as far back as IE8:
http://caniuse.com/#search=border-box

alexgibson: Yep, we generally don't support IE < 8 unless it's a specific page, but my comment was more about general consistency than browser support.

schalkneethling: Gotcha!

jpetto: In this instance, I value consistency. @alexgibson has a good point about the mental shift required when moving to updating pages with a different box model approach.

My thoughts are to apply `border-box` on specific elements when/where needed until the time we shift *all* bedrock pages to use it by default.

FWIW, we ran in to a problem with this approach when applying the Fx Family nav to `/firefox/os/` (the widget case mentioned above).

bedrock/mozorg/templates/mozorg/contribute/friends.html
```diff
@@ -0,0 +1,61 @@
...
+{% endblock %}
+
+{% block contrib_page_title %}{{ _('Join Firefox Friends today!') }}{% endblock %}
+
+{% block alt_header %}
+<img src="{{ static('img/contribute/friends/firefox-friends-logo.png') }}" width="186" class="ff-logo" alt="{{_('Firefox Friends logo')}}" />
+<div class="ff-header-content">
+  <h1>{{_('Join Firefox Friends today!')}}</h1>
+  <p>{{_('Step to the frontline to help us tell the world how <br />awesome Firefox is and about all the important work <br />Mozilla does every day.')}}</p>
+  <a href="https://friends.mozilla.org" id="ff-join" class="ff-join-now" rel="external">Join now</a>
```
flodolo: Missing l10n wrapping

media/css/mozorg/contribute/friends.less
schalkneethling: @jpetto @alexgibson @craigcook makes total sense to me. To consistency! :beer:

schalkneethling:
@alexgibson @craigcook This is ready for code review. Thanks!

bedrock/mozorg/templates/mozorg/contribute/friends.html
```diff
@@ -0,0 +1,64 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "mozorg/contribute/contribute-base.html" %}
+
+{% block page_title %}
+    {{ _('Join Firefox Friends today!') }}
```
alexgibson: Nit: fix indentation

bedrock/mozorg/templates/mozorg/contribute/friends.html
```diff
@@ -0,0 +1,64 @@
...
+
+{% block js %}
+{% endblock %}
+
+{% block contrib_page_title %}{{ _('Join Firefox Friends today!') }}{% endblock %}
+
+{% block alt_header %}
+<img src="{{ static('img/contribute/friends/firefox-friends-logo.png') }}" width="186" class="ff-logo" alt="{{_('Firefox Friends logo')}}" />
+<div class="ff-header-content">
+  <h1>{{_('Join Firefox Friends today!')}}</h1>
```
alexgibson: Would the `<h1>` section here be more suited inside `<main role="main">` perhaps?

alexgibson:
Are the line breaks here needed, it looks a little odd in places at different screen widths?

line-break

alexgibson:
Remove default `text-shadow` on headings?

shadow

media/css/mozorg/contribute/friends.less
```diff
@@ -0,0 +1,153 @@
...
+@import "../../sandstone/lib.less";
+
+.firefox-friends {
+    #wrapper {
+        padding-bottom: 0;
+        width: 100%;
+    }
+    #masthead {
+        background: #373737 url(/media/img/contribute/friends/header-img-wide.jpg) 50% 0 no-repeat;
+        padding: 0 180px;
```
alexgibson: Would it be easier to set a fixed width here on `#masthead` instead of adjust the padding at different screen sizes? This would probably make it more consistent with other pages.

alexgibson: I guess then you would need to apply the background image to a parent element though, instead of `#masthead`. Maybe it is easier to leave it as-is.

media/css/mozorg/contribute/friends.less
```diff
@@ -0,0 +1,153 @@
...
+        width: 100%;
+        .border-box();
+
+        @media only screen and (max-width: 1100px) {
+            padding: 0 100px;
```
alexgibson: Is it worth making some mobile-sized images for the background, to help avoid small screens needing to load in the full sized image?

media/css/mozorg/contribute/friends.less
```diff
@@ -0,0 +1,153 @@
...
+    .ff-join-now {
+        display: block;
+        background-color: #f26c24;
+        color: #fff;
+        margin: 70px auto 0 auto;
+        padding: 10px 30px;
+        border: 0;
+        width: 200px;
+        .font-size(1.2rem);
+        .border-radius(6px);
```
alexgibson: Nit: no need to use the mixin for border-radius as unprefixed support is good enough now :)

alexgibson:
Do we need to add a `/b/` rewrite rule for this new URL to global.conf?

http://bedrock.readthedocs.org/en/latest/contribute.html#getting-a-new-bedrock-page-online

bedrock/mozorg/templates/mozorg/contribute/friends.html
```diff
@@ -0,0 +1,64 @@
...
+{% endblock %}
+
+{% block contrib_head %}{% endblock contrib_head %}
+
+{% block contrib_content %}
+<section class="ff-body-content">
+  <h2>{{_('How it works')}}</h2>
+
+  <div class="how">
+    <img src="{{ static('img/contribute/friends/connected.png') }}" width="155" alt="" />
```
alexgibson: These three images could probably be done as background images in CSS, using a sprite maybe?

bedrock/mozorg/templates/mozorg/contribute/friends.html
```diff
@@ -0,0 +1,64 @@
...
+  <div class="how">
+    <img src="{{ static('img/contribute/friends/rewarded.png') }}" width="155" alt="" />
+    <h3>{{_('Get rewarded')}}</h3>
+    <p>{{_('Track your impact and share often<br /> for opportunities to be recognized<br /> and rewarded.')}}</p>
+  </div>
+
+  <hr />
+
+  <h2>{{_('Watch the video')}}</h2>
+  <!-- this is temp as the video might not be ready for the 5/11 launch, see:
```
alexgibson: Is this still needed if the video is not ready?

media/css/mozorg/contribute/friends.less
```diff
@@ -0,0 +1,153 @@
...
+
+        @media only screen and (max-width: @breakMobileLandscape) {
+                margin-top: 35px;
```
alexgibson: Nit: fix indentation

alexgibson:
Could these do with breaking onto their own lines a little sooner in responsiveness? Starts to feel a little squished maybe

mobile-squish

alexgibson:
"How it works" section needs some RTL layout tweaks:

rtl

bedrock/mozorg/templates/mozorg/contribute/friends.html
schalkneethling: The `main` section wraps the blue area below the `header` but, I do feel that the above line is the main headline of the page and thus, it is defined as the `h1` as apposed to the "How it works" heading.

bedrock/mozorg/templates/mozorg/contribute/friends.html
schalkneethling: Yeah, I was going back and forward on this one. They are decorative only so, makes sense.

bedrock/mozorg/templates/mozorg/contribute/friends.html
alexgibson: Yeah, I agree "How it works" shouldn't be the `<h1>`. I just wondered that "Join Firefox Friends today!" feels like it belongs to the main contents of the document (whereas `#masthead` and `#colophon` belong to the body). But this is all minor semantics, do what you feel is best here :)

bedrock/mozorg/templates/mozorg/contribute/friends.html
schalkneethling: It was a placeholder, and the general feeling was that it would possibly be ready today but, seems not so, removing.

media/css/mozorg/contribute/friends.less
schalkneethling: Yup, I was actually discussing that with Ben, do we have a way to dynamically load the images or, do we just use media queries and load a smaller background image?

alexgibson: As long as it's a background image in CSS, I think most mobile browsers should only download the one they need :+1:

schalkneethling:
@alexgibson the line breaks were dictated in the design yes. I considered not using the `<br />` tags but, in the past it seemed that we sometimes have to. The catch-22 now is, removing those `br` tags is a string change and we cannot make string changes anymore :-/

flodolo:
It's 2 locales, I can fix them by hand (again), as long as we don't go live with a mix of English and translated text. But we really need to find a better solution to stray line-breaks given that this kind of layout is more and more common.

One would be to explain once and for all to other stakeholders that a webpage is not a pixel-perfect picture, things need to move around.

alexgibson:
> @alexgibson the line breaks were dictated in the design yes. I considered not using the <br /> tags but, in the past it seemed that we sometimes have to. The catch-22 now is, removing those br tags is a string change and we cannot make string changes anymore :-/

Hmm, yeah that's tricky to change now. Given that the containers are fully fluid, I'm not sure how to go about preventing awkward breaks in the copy. I understand this is a common ask from creative (been there too).

A technique I often use is setting a `max-width` on the whichever text element it is, and then adjusting it until it more or less fits what they want to see in English. This can then be adjusted at different breakpoints, or even per-locale if needed.

schalkneethling:
@flodolo @alexgibson So, the initial idea was to launch this page yesterday but, then it was decided to push it out another day to see if the video is ready. Seems it is not going to be though so, I am guessing this will need to go live today.

So @flodolo I reckon it is your call if you are open to a string change or would prefer to defer this till after the initial launch. Also, I totally agree on the pixel perfect comment but, as Alex mentioned sometimes it just is what it is.

schalkneethling:
@alexgibson @flodolo ok, content and styling updated for the 'How it works' paragraphs.

schalkneethling:
@alexgibson in terms of the /b/ question, as this should go live today, I reckon we do not need that?

alexgibson:
> @alexgibson in terms of the /b/ question, as this should go live today, I reckon we do not need that?

@schalkneethling - it's needed for the page to be viewable on production, so best add it here in this PR

media/css/mozorg/contribute/friends.less
```diff
@@ -0,0 +1,170 @@
...
+            text-shadow: 1px 1px 2px #333;
+
+            @media only screen and (max-width: 790px) {
+                .font-size(3rem);
+            }
+        }
+        p {
+            .font-size(1.4rem);
+            text-shadow: 1px 1px 2px #333;
+
```
alexgibson: Should we set a max-width on this element too?

flodolo:
Pushed changes to both es-ES and pt-BR.

The third column on es-ES looks really long (long words), but the main issue is the sprite, it doesn't cope with multiple lines in 1st and 2nd column.

schermata 2015-05-12 alle 15 59 28

alexgibson:
We're a bit tight on vertical background space on mobile. I'd suggest trimming down the 160px bottom padding for small viewports to give some slack?

mobile-padding

schalkneethling:
@alexgibson yeah, it is a bit tight, I reckon halving that will work. Updating.

etc/httpd/global.conf
```diff
@@ -205,6 +205,8 @@ RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?mission(/?)$ /b/$1mission$2 [PT]
 # bug 793754, 1054437
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?contribute(/?|/.+)$ /b/$1contribute$2 [PT]
 
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?contribute/friends(/?)$ /b/$1contribute/friends$2 [PT]
```
alexgibson: Nit: good practice to include the bug number here also

media/css/mozorg/contribute/friends.less
schalkneethling: whoops, indeed. How did I miss that, hmmmmm.

schalkneethling:
@flodolo thanks, looking into that.

media/css/mozorg/contribute/friends.less
```diff
@@ -0,0 +1,185 @@
...
+                width: 30%;
+            }
+
+            @media only screen and (max-width: 640px) {
+                display: block;
+                padding-left: 0;
+                width: 100%;
+            }
+
+            h3::before {
```
alexgibson: Unless I'm missing something, I think you can avoid the need for the IE specific style sheet if you just use `h3:before` instead of `h3::before`?

alexgibson:
Some slight squishening on the icons as I resize:

squish

media/css/mozorg/contribute/friends.less
schalkneethling: :before does not seem to work in Fx from what I could see, nor IE9+

alexgibson:
Other than these last two nits, this is looking great! r+wc thanks @schalkneethling

media/css/mozorg/contribute/friends.less
alexgibson: Strange, seems to work for me locally, at least, we use `:before` quite regularly on a lot of pages, e.g. on /firefox/os/

schalkneethling: Say whaaat? no it works for me as well. I must have done something else wrong. :-/

PR #2951, opened 2015-05-04
Fix bug 1160395 - link family page download button to /new scene2

craigcook:


bedrock/firefox/templates/firefox/family/index.html
```diff
@@ -36,12 +36,12 @@
       <div class="message" id="dlbar-oldfx-desktop">
         {{ _('You’re using an older version of Firefox for desktop.') }}
         {{ _('Update today to stay fast and safe.') }}
-        <a class="button-flat btn-download" href="{{ url('firefox.new') }}" data-product="Firefox for Desktop, outdated Firefox">{{ _('Download') }}</a>
+        <a class="button-flat btn-download" href="{{ url('firefox.new') }}?scene=2#download-fx" data-product="Firefox for Desktop, outdated Firefox">{{ _('Download') }}</a>
```
alexgibson: If memory serves me correctly, this might fail in old-IE?

Maybe that's ok for this page, but if not we may have to use a real download button. But then the button text would be "Free Download" instead of just "Download"

PR #2954, opened 2015-05-05
Bug 1144949 - Use "Contribute to this site" for all locales

flodolo:
The new string is now localized in 57 locales ( ~90% of our l10n user base). Time to remove the switch.


alexgibson:
r+ thanks @flodolo

PR #2955, opened 2015-05-05
Fix Bug 1139961 - Update universal download button to add link to App Store for iOS

kyoshino:
WIP. Do not merge.

* We don't know the download URL yet
* We don't know the versioning scheme yet
* We don't know if Tim Cook would love Firefox yet
* The tests are still failing because the iOS release notes are not ready yet

### DIFF
--- a/bedrock/firefox/templates/firefox/new.html
+++ b/bedrock/firefox/templates/firefox/new.html
@@ -53,20 +53,17 @@
 
 <main role="main">
   <div class="version-message-container">
-    {% if l10n_has_tag('ios_announce') %}
-      <div id="version-message-ios" class="version-message">
-        {{_('Firefox is coming soon to iOS! <a href="{url}">Sign up to learn more</a>.').format(url=url('newsletter.ios'))|safe }}
-      </div>
-    {% else %}
-      <div id="version-message-ios" class="version-message">
-        {{_('Sorry, Firefox is currently unsupported by iOS devices.')}}
-      </div>
-    {% endif %}
     <div id="version-message-android-latest" class="version-message">
       {{_('Congrats! You’re using the latest version of Firefox.')}}
     </div>
     <div id="version-message-android-old" class="version-message">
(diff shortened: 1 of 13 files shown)


bedrock/firefox/templates/firefox/new.html
```diff
@@ -53,21 +53,18 @@
...
     <div id="version-message-android-latest" class="version-message">
       {{_('Congrats! You’re using the latest version of Firefox.')}}
     </div>
     <div id="version-message-android-old" class="version-message">
       {{_('Looks like you’re using an older version of Firefox. Update on Google Play.')}}
     </div>
+    <div id="version-message-ios-latest" class="version-message">
+      {{_('Congrats! You’re using the latest version of Firefox.')}}
+    </div>
+    <div id="version-message-ios-old" class="version-message">
```
alexgibson: Given that iOS apps are auto updating, is this really required?

kyoshino: Hmm, so there would be no message, or just say "Thank you for using Firefox!" or something?

alexgibson: I'm not too sure to be honest, best check with Cory in the bug on what the behaviour should be here? I think we are likely looking to avoid adding new strings if possible.

kyoshino: Actually there are a couple of new strings in this PR:

* Get Firefox on your iOS device
* Learn more about Firefox for Android *instead of saying Mobile*
* Learn more about Firefox for iOS

flodolo: Wait, why are we adding strings at this point? That wasn't the plan :-\

alexgibson: I would probably check in the bug what needs changing here before we commit to introducing new strings.

kyoshino: I'll ask in the bug.

kyoshino:
Closing the issue until we know more.

PR #2956, opened 2015-05-06
Fix bug 1156822 - update contact info in participation guidelines

craigcook:
Also removed some outdated content (doesn't impact the overall guidelines)

@flodolo: This page doesn't seem to be localized so these content changes shouldn't need an l10n flag. Can you verify?


alexgibson:
Looks good to me, r+ and good to merge, once @flodolo gives the thumbs up on changing strings.

flodolo:
We don't localize this page :+1:

PR #2957, opened 2015-05-06
Fix syntax and deprecation warnings in tests

jgmize:
Moved parens to eliminate "always true" assertions
Removed deprecated simplejson module

### DIFF
--- a/bedrock/base/tests/test_accepted_locales.py
+++ b/bedrock/base/tests/test_accepted_locales.py
@@ -65,7 +65,7 @@ class AcceptedLocalesTest(test_utils.TestCase):
         """
         settings.DEV = True
         langs = get_dev_languages()
-        assert (langs == ['en-US', 'fr'] or langs == ['fr', 'en-US'],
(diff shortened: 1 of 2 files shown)


jgmize:
@pmclanahan r?

bedrock/base/tests/test_accepted_locales.py
```diff
@@ -65,7 +65,7 @@ def test_build_dev_languages(self):
         """
         settings.DEV = True
         langs = get_dev_languages()
-        assert (langs == ['en-US', 'fr'] or langs == ['fr', 'en-US'],
+        assert langs == ['en-US', 'fr'] or langs == ['fr', 'en-US'], (
```
pmclanahan: DOH! I did that. flake8 was complaining about the indentation. might not like this either.

jgmize: yeah, I missed this in the review :(
flake8 isn't complaining about this version.

PR #2958, opened 2015-05-06
Bug 1162101. FxOS Orange Klif Madagascar.

jpetto:
@craigcook r?


bedrock/firefox/templates/firefox/os/devices.html
```diff
@@ -2237,6 +2238,187 @@ <h4 class="device-details-title">{{ _('Specifications') }}</h4>
...
+                    <li class="stat-camera">
+                      <strong>{{ _('2.0 MP') }}</strong>
+                      {{ _('Rear camera') }}
+                    </li>
+                    <li class="stat-storage">
+                      <strong>{{ _('512 MB') }}</strong>
+                      {{ _('Internal memory') }} *
+                    </li>
+                    <li class="stat-battery">
+                      <strong>{{ _('7.0 h') }}</strong>
```
flodolo: Any chance of using "7 h", since we already have this string?

jpetto: It shall be done.

PR #2959, opened 2015-05-06
Fix Bug 1161783 - [traceback] TypeError: download_firefox() got an unexpected keyword argument 'build' on /styleguide/all-download-buttons/

kyoshino:


jgmize:
wfm locally after setting DEV=True
thanks @kyoshino!
r+

PR #2960, opened 2015-05-07
Fix Bug 1162669 - Thunderbird latest sysreq/relnotes links are redirected to website-archive

kyoshino:
Tested locally on Apache.


etc/httpd/global.conf
```diff
@@ -576,6 +576,7 @@ RewriteRule ^/(\w{2,3}(?:-\w{2}(?:-mac)?)?/)?thunderbird/([3-9]\d\.\d(?:a2|beta|
 RewriteRule ^/(\w{2,3}(?:-\w{2}(?:-mac)?)?/)?thunderbird/([3-9]\d\.\d(?:a2|beta|\.\d)?)/system-requirements(/)?$ /b/$1thunderbird/$2/system-requirements$3 [L,PT]
 RewriteRule ^/(\w{2,3}(?:-\w{2}(?:-mac)?)?/)?thunderbird/releases(/)?$ /b/$1thunderbird/releases$2 [L,PT]
 RewriteRule ^/(\w{2,3}(?:-\w{2}(?:-mac)?)?/)?thunderbird/releases/(.*)$ http://website-archive.mozilla.org/www.mozilla.org/thunderbird_releasenotes/en-US/thunderbird/releases/$2 [L,R=301]
+RewriteRule ^/(\w{2,3}(?:-\w{2}(?:-mac)?)?/)?thunderbird/latest/(.*)(/)?$ /b/$1thunderbird/latest/$2$3 [L,PT]
```
jgmize: Rather than making the optional trailing slash a capture group with`(/)?`, you can omit the parens and unconditionally include the slash in the target:
```
RewriteRule ^/(\w{2,3}(?:-\w{2}(?:-mac)?)?/)?thunderbird/latest/(.*)/?$ /b/$1thunderbird/latest/$2/ [L,PT]
```
You can simplify this even further by eliminating the `(.*)` and explicitly matching only `releasenotes`, which is the only thing that won't result in a 404 anyway:

```
RewriteRule ^/(\w{2,3}(?:-\w{2}(?:-mac)?)?/)?thunderbird/latest/releasenotes/?$ /b/$1thunderbird/latest/releasenotes/ [L,PT]
```

kyoshino: There are 2 URLs incorrectly redirecting to archive, so `/latest/(.*)/?$` should work.

jgmize:
Thanks @kyoshino!

kyoshino:
Thanks @jgmize!

PR #2963, opened 2015-05-07
Fix bug 945845: Escape error messages sent via AJAX.

pmclanahan:
The partnership page form did not properly sanitize user input
before including it in the page via AJAX.


pmclanahan:
@jgmize r?

jgmize:
thanks @pmclanahan

PR #2964, opened 2015-05-08
Fix bug 1162151 error on bin/sync_all

jgmize:


jgmize:
@pmclanahan r?

bin/sync_all
```diff
@@ -1,5 +1,10 @@
 #!/bin/bash
-cd $(readlink --canonicalize --no-newline ${0%/*})/..
+set -ex
```
pmclanahan: :+1:

pmclanahan:
Works well for me, seems much simpler, and even improves output. :clap:

PR #2965, opened 2015-05-08
updating bug bounty hall of fame for Q1 2015 and Q4 2014

rforbes:
This update just added some data to the html page for

### DIFF
--- a/bedrock/security/templates/security/bug-bounty/hall-of-fame.html
+++ b/bedrock/security/templates/security/bug-bounty/hall-of-fame.html
@@ -32,9 +32,52 @@
 
       <div class="accordion accordion-auto-init zebra">
 
+      <section id="year-2015">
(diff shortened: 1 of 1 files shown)


bin/sync_all
```diff
@@ -1,5 +1,4 @@
 #!/bin/bash
-cd $(readlink --canonicalize --no-newline ${0%/*})/..
```
flodolo: You don't want this in your PR ;-)

rforbes: dang it, you are right. that was an error.  I am sorry, I am new to this git thing.  What should I do?

pmclanahan:
Looks good and seems to work well. r+ from me. maybe @craigcook should double check the HTML though.

craigcook:
Looks good to me!

PR #2969, opened 2015-05-09
Bug1151534 campaign landing page

schalkneethling:
@alexgibson @craigcook r? Still a WIP but, feel free to have a look over the code.


bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,55 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+ {% extends "base-resp.html" %}
+
+{% block page_title_prefix %}{{_('Firefox Web browser')}} - {% endblock %}
+{% block page_title %}{{_('When it\'s personal, choose Firefox')}}{% endblock %}
```
flodolo: Use curly apostrophes in English, not straight ones.

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,55 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+ {% extends "base-resp.html" %}
+
+{% block page_title_prefix %}{{_('Firefox Web browser')}} - {% endblock %}
```
flodolo: Is lowercase `browser` wanted? We had this string in the past, but it was ` Firefox Web Browser `

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,55 @@
...
+{% block page_title %}{{_('When it\'s personal, choose Firefox')}}{% endblock %}
+
+{% block page_desc %}Choose Firefox when you want to browse the Web without the Web browsing you.{% endblock %}
```
flodolo: l10n wrapping

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,55 @@
...
+  {% stylesheet 'personal' %}
+{% endblock %}
+
+{% block site_header_nav %}{% endblock %}
+{% block site_header_logo %}{% endblock %}
+
+{% block alt_header %}
+<div class="header-content">
+  <img src="{{ static('img/firefox/personal/logo.png') }}" width="147" alt="{{_('Firefox logo')}}" />
+  <h1>{{_('When it\'s personal,<span>choose Firefox</span>')}}</h1>
```
flodolo: Curly apostrophe

bedrock/firefox/templates/firefox/personal.html
schalkneethling: That is what is here: https://docs.google.com/document/d/1QfQKx0DSOmcXIVPkA1nQ1TEWS5JbhIyZs0ZgsiyKXC4/edit

schalkneethling: If that causes a need for a new string, as apposed to being able to reuse something, I would think it can just be 'Browser'

flodolo: I would mostly ask to confirm that's wanted. The other string is not in a shared file, so we'd still need it in the new lang file.

AssholeDotzler:
Usted ama el dolor? Y para que te haga un favor... voy a hacer eso!


Yo saque sus huevos de oro de fracaso tan duro hasta que están sonando y que están cantando todas las melodías de dolor hacia adelante y hacia atrás en un instante! Yo soy el cascanueces chico que las grietas de su nueces que nutheads!

Y ahora vete a la mierda ESTÚPIDA de MIERDA COÑO!

----

Usted puede lamer mi Pene y mis Huevos de oro y besar mi Culo hijo de puta con un Trastorno de la personalidad desde el más profundo de los abismos del infierno desde los Abismos de Mozilla!

Vete a la mierda y que tu Madre te autista hijo de puta, vas a lamer Putin Huevos, Puto Maricon Fascista!

La mierda de su Perro, su Gato y su sucio Cabra de montaña que Maricon Pendejo!

Chupar Mozilla Pollas y todos juntos ir y chupar de Google Culo lleno de Mierda!

A la mierda en su rodilla le cock suckers! Y que Carajo Australis Chrome clon de mierda! Follando Mozilla gay lesbiana homosexual hijos de puta del infierno!

bedrock/firefox/templates/firefox/personal.html
ckprice: Just a note that the campaign landing page will be en-us only.

On Sunday, May 10, 2015, Francesco Lodolo <notifications@github.com> wrote:

> In bedrock/firefox/templates/firefox/personal.html
> <https://github.com/mozilla/bedrock/pull/2969#discussion_r30004986>:
>
> > @@ -0,0 +1,55 @@
> > +{# This Source Code Form is subject to the terms of the Mozilla Public
> > + # License, v. 2.0. If a copy of the MPL was not distributed with this
> > + # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
> > +
> > + {% extends "base-resp.html" %}
> > +
> > +{% block page_title_prefix %}{{_('Firefox Web browser')}} - {% endblock %}
>
> I would mostly ask to confirm that's wanted. The other string is not in a
> shared file, so we'd still need it in the new lang file.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/bedrock/pull/2969/files#r30004986>.
>


-- 
Cory Price
/ckprice

ckprice:
Please remove l10n tag - this page is en-us only.

On Saturday, May 9, 2015, Schalk Neethling <notifications@github.com> wrote:

> @alexgibson <https://github.com/alexgibson> @craigcook
> <https://github.com/craigcook> r? Still a WIP but, feel free to have a
> look over the code.
> ------------------------------
> You can view, comment on, or merge this pull request online at:
>
>   https://github.com/mozilla/bedrock/pull/2969
> Commit Summary
>
>    - Fix Bug 1151534, implement spring campaign landing page
>    - page layout complete
>
> File Changes
>
>    - *A* bedrock/firefox/templates/firefox/personal.html
>    <https://github.com/mozilla/bedrock/pull/2969/files#diff-0> (55)
>    - *M* bedrock/firefox/urls.py
>    <https://github.com/mozilla/bedrock/pull/2969/files#diff-1> (1)
>    - *M* bedrock/settings/static_media.py
>    <https://github.com/mozilla/bedrock/pull/2969/files#diff-2> (7)
>    - *A* media/css/firefox/personal.less
>    <https://github.com/mozilla/bedrock/pull/2969/files#diff-3> (69)
>    - *A* media/img/firefox/personal/data_money.png
>    <https://github.com/mozilla/bedrock/pull/2969/files#diff-4> (0)
>    - *A* media/img/firefox/personal/header.png
>    <https://github.com/mozilla/bedrock/pull/2969/files#diff-5> (0)
>    - *A* media/img/firefox/personal/logo.png
>    <https://github.com/mozilla/bedrock/pull/2969/files#diff-6> (0)
>    - *A* media/img/firefox/personal/personal_info.png
>    <https://github.com/mozilla/bedrock/pull/2969/files#diff-7> (0)
>    - *A* media/img/firefox/personal/you_choose.png
>    <https://github.com/mozilla/bedrock/pull/2969/files#diff-8> (0)
>
> Patch Links:
>
>    - https://github.com/mozilla/bedrock/pull/2969.patch
>    - https://github.com/mozilla/bedrock/pull/2969.diff
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/bedrock/pull/2969>.
>


-- 
Cory Price
/ckprice

flodolo:
Removed (I was the one who added it), but I think I'm a bit lost at this point (I assumed we needed together with whatsnew/firstrun for the 4 locales).

ckprice:
Yeah getting lost is very easy sorry :( The Pocket landing page will
require l10n for the 4 locales. There's no PR for that yet. We're targeting
15th for exposed strings on that page (Craig is doing the dev there).

On Sunday, May 10, 2015, Francesco Lodolo <notifications@github.com> wrote:

> Removed (I was the one who added it), but I think I'm a bit lost at this
> point (I assumed we needed together with whatsnew/firstrun for the 4
> locales).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/bedrock/pull/2969#issuecomment-100697250>.
>


-- 
Cory Price
/ckprice

flodolo:
Ah, now it's clear: campaign landing page ≠ pocket landing page. Thanks :+1: 
(note that the comments about curly apostrophes for en-US are still valid)

media/css/firefox/personal.less
```diff
@@ -0,0 +1,104 @@
...
+.header-content {
+    margin: 0 auto;
+    padding: 40px 0 80px 0;
+    width: 100%;
+    max-width: 1400px;
+    text-align: center;
+    h1 {
+        color: #bee1f5;
+        margin: 0 0 55px 0;
+        font-family: 'Open Sans';
```
ckprice: @schalkneethling just a note that this whole page should be Fira Sans.

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,55 @@
...
+  {{ download_firefox(force_direct=true, simple=true) }}
+</div>
+{% endblock alt_header %}
+
+{% block extrahead %}{% endblock %}
+
+{% block js_pretrack %}{% endblock %}
+
+{% block content %}
+<main role="main">
```
ckprice: @schalkneethling destination for these links is noted in the [copy document](https://docs.google.com/document/d/1QfQKx0DSOmcXIVPkA1nQ1TEWS5JbhIyZs0ZgsiyKXC4/edit).

schalkneethling: Yup, got that, thanks!

media/css/firefox/personal.less
schalkneethling: Alrighty, will make the switch.

ckprice:
@schalkneethling a couple notes comparing the [designs](https://bugzilla.mozilla.org/show_bug.cgi?id=1149243#c8) to this PR.

**Firefox watermark**

image 2015-05-12 at 11 22 05 am

The designs have the watermark's tail pretty much centered with the headline, but the PR shows it skewed left quite a bit.

**Mozilla tab**

Seems to escape the bounds of the container. The designs (and other mozorg pages), Tabzilla is anchored to the main wrapper on the page.

**Responsive**

There is a responsive PSD in the design bug which has the proof points stacking. Could we incorporate that?

Thanks!

schalkneethling:
@ckprice I left aligned the watermark because we are filling the entire view port and the watermark is cut off on the left so, does not work any other way unless I can get a watermark that is not cut off on the left. As mentioned above the one I extracted from the PSD is.

Will look at the positioning of tabzilla

I do not believe I have pushed the latest responsive bits, will do so now.

Thanks for the feedback!

schalkneethling:
@ckprice can you give this a review again?

ckprice:
@schalkneethling this looks much better.

It still needs a technical code review (@craigcook?).

I'd also like to get this on the demo server, and have Ty/Matej take a look at it in action.

Thanks!

schalkneethling:
Thanks for the feedback @ckprice - also pinging @alexgibson for code review here. Will see if there is a demo server available.

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,67 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+ {% extends "base-resp.html" %}
+
+{% block page_title_prefix %}{{_('Firefox Web Browser')}} - {% endblock %}
+{% block page_title %}{{_('When it&rsquo;s personal, choose Firefox')}}{% endblock %}
```
alexgibson: Nit: no need to use `&rsquo;` we can just do `’`

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,67 @@
...
+{% block site_css %}
+  {% stylesheet 'personal' %}
+{% endblock %}
+
+{% block site_header_nav %}{% endblock %}
+{% block site_header_logo %}{% endblock %}
+
+{% block alt_header %}
+<div class="header-content">
+  <h1>{{_('When it&rsquo;s personal,<span>choose Firefox</span>')}}</h1>
```
alexgibson: Same here

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,67 @@
...
+  {{ download_firefox(simple=true) }}
+
+  <div class="mobile-download">
+    <a href="https://play.google.com/store/apps/details?id=org.mozilla.firefox&referrer=utm_source%3Dmozilla%26utm_medium%3DReferral%26utm_campaign%3Dmozilla-org">
```
alexgibson: Escape with `amp;`

alexgibson: Actually we have a canonical link in setting you can use here: `{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}`

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,67 @@
...
+      <img src="{{ static('img/firefox/personal/google_play_badge.png') }}" width="172" alt="{{_('Download from Google Play')}}" />
```
alexgibson: We have high-res badges you can use here already in the repo:

    {{ high_res_img('firefox/android/btn-google-play.png', {'alt': 'Get it on Google Play', 'width': '129', 'height': '45', 'l10n': True}) }}

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,67 @@
...
+
+  <div class="mission">
+    <p class="choice"><span>You choose</span> how your information<br /> is collected and<br /> shared online.</p>
+    <a href="{{ url('privacy.privacy-day') }}">Get tips about your privacy</a>
+  </div>
+</main>
+
+<section class="mobile-download">
+  <p>{{_('Add Firefox to your smartphone or tablet')}}</p>
+  <a href="https://play.google.com/store/apps/details?id=org.mozilla.firefox&referrer=utm_source%3Dmozilla%26utm_medium%3DReferral%26utm_campaign%3Dmozilla-org">
```
alexgibson: Again, this link can be replaced by the canonical link to the play store in settings.

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,67 @@
...
+    <img src="{{ static('img/firefox/personal/google_play_badge.png') }}" width="172" alt="{{_('Download from Google Play')}}" />
```
alexgibson: Use high-res badge that's already in the repo

bedrock/settings/static_media.py
```diff
@@ -687,6 +687,13 @@
         ),
         'output_filename': 'css/persona-bundle.css',
     },
+    'personal': {
```
alexgibson: This is a bit, but should the CSS bundle be called `firefox-personal` to match the JS bundle name?

media/css/firefox/personal.less
```diff
@@ -0,0 +1,194 @@
+@import "../sandstone/lib.less";
+
+@font-face {
+    font-family: 'FiraSans Regular';
+    src: url('/media/fonts/FiraSans-Regular.woff') format('woff');
```
alexgibson: Do we need to add other Fira Sans font formats for other browsers?

media/css/firefox/personal.less
```diff
@@ -0,0 +1,194 @@
...
+    background: none;
+}
+#wrapper {
+    width: 100%;
+}
+#masthead {
+    background: #0c99d5 url(/media/img/firefox/personal/header.png) 0 0 no-repeat;
+
+    @media only screen and (min-width: @breakTablet) and (max-width: @breakDesktop) {
+        width: 100%;
```
alexgibson: The Tabzilla tab is floating to the top right of the page, but we should try and make it consistent with other mozorg pages, and set `#masthead` to a fixed width that's centered in the page. You would need to apply the background image to a parent container in this case though.

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,67 @@
...
+
+{% block extrahead %}{% endblock %}
+
+{% block js_pretrack %}{% endblock %}
+
+{% block content %}
+<main role="main">
+  <div class="mission">
+    <p><span>We believe</span> that your personal<br /> information belongs<br /> only to you.</p>
+    <a href="//mozilla.org/firefox/desktop/trust">See how we put you first</a>
```
alexgibson: Change href to `{{ url('firefox.desktop.trust') }}`

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,67 @@
...
+  </div>
+
+  <div class="mission">
+    <p class="data"><span>We never</span> act in a way that<br /> would compromise<br /> that belief.</p>
+    <a href="//mozilla.org/about/manifesto">Learn about our mission</a>
```
alexgibson: Change href to `{{ url('mozorg.about.manifesto') }}`

bedrock/settings/static_media.py
```diff
@@ -1326,6 +1333,13 @@
         ),
         'output_filename': 'js/firefox_hello_ie9-bundle.js',
     },
+    'firefox_personal': {
+        'source_filenames': (
+            'js/libs/jquery-1.7.1.min.js',
+            'js/firefox/personal.js'
```
alexgibson: Missing a trailing comma here

alexgibson:
We should be showing the mobile download buttons based on platform, not just by mobile screen size.

If a user hits the page on an Android based device, bedrock automatically adds a class of 'android' to the main `<html>` element. You can hook into that in your CSS.

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,67 @@
...
+  </div>
+
+  <div class="mission">
+    <p class="choice"><span>You choose</span> how your information<br /> is collected and<br /> shared online.</p>
+    <a href="{{ url('privacy.privacy-day') }}">Get tips about your privacy</a>
```
alexgibson: If you just add a class of `more` to these links sandstone automatically adds the `»`

media/css/firefox/personal.less
```diff
@@ -0,0 +1,194 @@
...
+    p.data:before {
+        background-position: 50% -174px;
+    }
+    p.choice:before {
+        background-position: 50% -380px;
+    }
+    a {
+        .font-size(1.2rem);
+        &:after {
+            content: ' \00BB';
```
alexgibson: No need for this in your CSS if you use the `more` class that's in sandstone.

media/css/firefox/personal.less
```diff
@@ -0,0 +1,194 @@
...
+    max-width: 1400px;
+    text-align: center;
+
+    @media only screen and (max-width: @breakMobileLandscape) {
+        padding-bottom: 60px;
+    }
+}
+.mission {
+    display: inline-block;
+    margin-left: 50px;
```
alexgibson: Instead of adding a left margin value to horizontal lists like these, could you just add some padding to either side maybe? (e.g. `padding: 0 25px;`). Then you wouldn't need to remove the left margin using `first-child` which isn't supported in older browsers.

media/js/firefox/personal.js
```diff
@@ -0,0 +1,28 @@
+$(function() {
+    'use strict';
+
+    var dntStatus = navigator.doNotTrack || navigator.msDoNotTrack;
+    var $body = $('body');
+    var $flashTalkingPixel = $('<img />', {
+        width: '1',
+        height: '1',
+        src: 'https://servedby.flashtalking.com/spot/8/6247;40428;4669/?spotName=Mozilla_Download_Conversion',
```
alexgibson: This looks fine code wise, but I'm going to go ahead and say the obligatory boo to tracking like this! :-1:  :(

media/js/firefox/personal.js
```diff
@@ -0,0 +1,28 @@
...
+    var fxMatch = navigator.userAgent.match(/Firefox\/(\d+)/);
+    if (fxMatch && Number(fxMatch[1]) < 32) {
+        // Can't say for sure if it is 1 or 0, due to Fx bug 887703
+        dntStatus = (dntStatus === 'yes') ? 'Unknown' : 'Unspecified';
+    } else {
+        dntStatus = { '0': 'Disabled', '1': 'Enabled' }[dntStatus] || 'Unspecified';
+    }
+
+    // only add the pixel if the value of dntStatus is enabled.
+    if (dntStatus === 'Enabled') {
```
alexgibson: At least we are trying to respect some requests for privacy :+1:

media/css/firefox/personal.less
alexgibson: Alternatively, instead of using a 100% fluid container, you could use the common viewport widths in bedrock and just set each item to 33% or similar.

bedrock/firefox/templates/firefox/personal.html
schalkneethling: huh, could have sworn I changed these. Ah well, thanks

bedrock/settings/static_media.py
schalkneethling: makes sense to me

media/css/firefox/personal.less
schalkneethling: We only have the woff fonts in bedrock, but I will grab the other from the Fira Repo

schalkneethling: Then again, any reason we cannot just use this? <link rel="stylesheet" href="//code.cdn.mozilla.net/fonts/fira.css"> @craigcook @alexgibson

schalkneethling: It will load all of the variants and versions of the font whereas we can be more selective if we have the fonts inside bedrock.

craigcook: I believe browsers shouldn't ("shouldn't" being the key word here; we should test to verify) download font files if the given page doesn't actually use them, so in theory just declaring a bunch of variants in CSS won't impact performance unless you use a lot of them on one page.

But that matter aside, I'd still vote for adding the other formats (TTF and EOT) to bedrock anyway since we know we'll need them in the future.

media/css/firefox/personal.less
```diff
@@ -0,0 +1,194 @@
...
+    font-family: 'FiraSans Regular';
+    src: url('/media/fonts/FiraSans-Regular.woff') format('woff');
+    font-weight: normal;
+    font-style: normal;
+}
+
+@font-face {
+    font-family: 'FiraSans Bold';
+    src: url('/media/fonts/FiraSans-Bold.woff') format('woff');
+    font-weight: normal;
```
craigcook: Instead separate families for Regular and Bold, we can just have "Fira Sans" as the family name and use <code>font-weight: bold;</code> same as we've done for Open Sans (see https://github.com/mozilla/bedrock/blob/master/media/css/sandstone/fonts.less). 

Fira Sans Light does need to be its own family name, though, since browsers have poor support for incremental weights (200, 500, 700, etc).

media/css/firefox/personal.less
schalkneethling: Handled it slightly differently, but the end result is the same ;)

media/js/firefox/personal.js
schalkneethling: +1

media/css/firefox/personal.less
schalkneethling: Yup, agree. Added the required one's to Bedrock

schalkneethling:
@ckprice in https://github.com/mozilla/bedrock/pull/2969#issuecomment-102397364 @alexgibson raises a good point, in that we should only show the Play store buttons if we are on a Android device and not just because we are at a mobile view port size.

Does this match the requirement or, should we just base it on view port size as it is currently?

If it should be as Alex suggests, I assume I just keep the green 'Free Download' button for non-android mobile view ports.

alexgibson:
When I resize to smaller viewports, the Tabzilla tab gets cut off, and the second line of the heading looks like it floats a bit off center

header

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,70 @@
...
+    <a rel="external" href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}">
+      {{ high_res_img('firefox/android/btn-google-play.png', {'alt': 'Get it on Google Play', 'width': '172', 'l10n': True}) }}
+    </a>
+  </div>
+</div>
+{% endblock alt_header %}
+
+{% block extrahead %}{% endblock %}
+
+{% block js_pretrack %}{% endblock %}
```
alexgibson: Is this block needed?

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,70 @@
...
+
+  <div class="mobile-download">
+    <a rel="external" href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}">
+      {{ high_res_img('firefox/android/btn-google-play.png', {'alt': 'Get it on Google Play', 'width': '172', 'l10n': True}) }}
+    </a>
+  </div>
+</div>
+{% endblock alt_header %}
+
+{% block extrahead %}{% endblock %}
```
alexgibson: Not sure if this one is needed either?

alexgibson:
Do we need a `/b/` pass-through for this page in global.conf?

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,70 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+ {% extends "base-resp.html" %}
```
alexgibson: Teeny nit: remove blank space

alexgibson:
List centering on mobile needs some responsive updates

mobile

Lord-of-Darkness:
Desarrolladores y líderes de Mozilla, que son travestis culo!

Alguien debe colgar ustedes para arriba en sus propias bolas y dejar que su perro muerda su caricatura de una polla!

Posteriormente deben ponerlo tiempo que sea necesario por la garganta hasta que traga a sí mismos en el suicidio!

Ustedes Mozilla está ... Estúpido de mierda coños!
--------------------------------------------------------
Developers and leaders of Mozilla, which are shemales ass!

Someone should hang you up in your own balls and let your dog bite his caricature of a cock!

They should then put it long as needed by the throat until he swallows themselves in suicide!

You Mozilla is... Stupid fucking pussies!

alexgibson:
I haven't seen the design, but should we be using the flat button styling here? The default here is the old glossy button

download

bedrock/firefox/templates/firefox/personal.html
schalkneethling: Does one not need to include the empty block to prevent the contents of it from the parent template included/rendered?

media/js/firefox/personal.js
```diff
@@ -0,0 +1,28 @@
+$(function() {
+    'use strict';
+
+    var dntStatus = navigator.doNotTrack || navigator.msDoNotTrack;
+    var $body = $('body');
+    var $flashTalkingPixel = $('<img />', {
+        width: '1',
+        height: '1',
+        src: 'https://servedby.flashtalking.com/spot/8/6247;40428;4669/?spotName=Mozilla_Download_Conversion',
+        class: 'flash-talking'
```
alexgibson: Hmm, the use of `class` here causes IE8 to throw an error :/

schalkneethling:
"Do we need a /b/ pass-through for this page in global.conf?"

Well this should only go live June 2 I believe so, not sure if it will be behind a waffle switch or hidden some other way but, we will eventually need the entry in global.conf

media/js/firefox/personal.js
schalkneethling: Oh wow, that is interesting. To be honest, I do not even need it so, will just get rid of it.

alexgibson: http://bugs.jquery.com/ticket/9019

schalkneethling:
"I haven't seen the design, but should we be using the flat button styling here? The default is the old glossy button"

I saw @jpetto added the styles for the flat look the other day, is there a parameter I can pass to `{{ download_firefox(simple=true) }}` or, do I need to use something completely different for that?

schalkneethling:
"List centering on mobile needs some responsive updates"

Ah yes, I know why, it is because of my change from `margin-right: 50px;` to `padding: 0 25px;` and now I am not clearing the padding correctly when switching to `display:block;`

bedrock/firefox/templates/firefox/personal.html
alexgibson: I believe the contents of the block are already empty in the parent. Very few pages use this block currently, IIRC

schalkneethling: in that case it will be fine, will double check on that.

media/js/firefox/personal.js
alexgibson: Wait, isn't this condition the wrong way around? We should not be adding the pixel when dnt is enabled?

schalkneethling: Yes it is ;p I forgot to change it around. hehe

bedrock/settings/static_media.py
```diff
@@ -1326,6 +1333,14 @@
         ),
         'output_filename': 'js/firefox_hello_ie9-bundle.js',
     },
+    'firefox-personal': {
+        'source_filenames': (
+            'js/libs/jquery-1.7.1.min.js',
```
alexgibson: Instead of including jquery here, you should probably just include the common `firefox-resp` bundle in the page. Currently you're not loading other files, such as `global.js` and `global-init.js`, so you're not getting the download button JS etc.

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -0,0 +1,70 @@
...
+  <a rel="external" href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}">
+    {{ high_res_img('firefox/android/btn-google-play.png', {'alt': 'Get it on Google Play', 'width': '172', 'l10n': True}) }}
+  </a>
+</section>
+{% endblock %}
+
+{% block email_form %}{% endblock %}
+
+{% block site_js %}
+  {% javascript 'firefox-personal' %}
```
alexgibson: If you add `{{ super() }}` here you will get the other files the page needs

bedrock/settings/static_media.py
schalkneethling: Yeah, I was trying to save a few bytes but, turns out a need a few of those files so, makes sense to just include the bundle.

media/css/firefox/personal.less
```diff
@@ -0,0 +1,201 @@
+@import "../sandstone/buttons.less";
```
alexgibson: This should already be included in sandstone.less, so no need to add it here

media/js/firefox/personal.js
```diff
@@ -0,0 +1,27 @@
...
+    var fxMatch = navigator.userAgent.match(/Firefox\/(\d+)/);
+    if (fxMatch && Number(fxMatch[1]) < 32) {
+        // Can't say for sure if it is 1 or 0, due to Fx bug 887703
+        dntStatus = (dntStatus === 'yes') ? 'Unknown' : 'Unspecified';
+    } else {
+        dntStatus = { '0': 'Disabled', '1': 'Enabled' }[dntStatus] || 'Unspecified';
+    }
+
+    // only add the pixel if the value of dntStatus is enabled.
+    if (dntStatus !== 'Enabled') {
```
alexgibson: What about if this is 'Unknown' or 'Unspecified'?, I would guess we probably shouldn't add the pixel for those cases either? Would it be better to check for `dntStatus !== 'Disabled'`?

media/css/firefox/personal.less
schalkneethling: Thought so to, but got a LESS error when I did not explicitly include it.

media/js/firefox/personal.js
schalkneethling: Yeah, not sure, it is the code that was requested in *the* bug so...

schalkneethling: So after the IRC discussion where we managed to confuse each other while actually agreeing :) what we are saying is, that this line should be:

```
if (dntStatus === 'Disabled') { .....
```

alexgibson: :+1: :dizzy:

schalkneethling:
This is now all good so, we now need two things:

1) Demo server, which we sort out on Monday
2) A waffle flag so this does not go live as soon as we merge it to master, but only on June 2

craigcook:
The design mock has a minimal footer similar to the one on /firefox/products but the current implementation has the full standard footer. We should confirm with Ty which footer the page should have.

ckprice:
Thanks Craig; I'm passing this around the design team for feedback.

schalkneethling:
@ckprice Thanks for the feedback. Seeing this is currently being reviewed on demo4 by everyone, let's see what this brings up.

alexgibson:
Occurred to me that we are still downloading both Open Sans and Fira Sans on this page, because Tabzilla. We could save nearly 200kb of page weight by using [tabzilla-static.less](https://github.com/mozilla/bedrock/blob/master/media/css/tabzilla/tabzilla-static.less). Worthwhile, @schalkneethling?

jgmize:
> Occurred to me that we are still downloading both Open Sans and Fira Sans on this page, because Tabzilla. We could save nearly 200kb (uncompressed) of page weight by using tabzilla-static.less

:+1: to doing this here *and everywhere else on the site*

schalkneethling:
@alexgibson definitely worth doing. Will bake it into the other requested changes.

schalkneethling:
@alexgibson oh wait, but if I use tz-static, does that not mean that it will not expand and collapse on click? Is that fin for this page?

alexgibson:
> @alexgibson oh wait, but if I use tz-static, does that not mean that it will not expand and collapse on click? Is that fin for this page?

correct, Tabzilla in it's current form is being nuked from orbit :boom:

schalkneethling:
@alexgibson right, so I can comfortably use the static version here, @ckprice ?

ckprice:
@schalkneethling if we're getting rid of Tabzilla in its current form, I think we should just use the static version. I assume that the static form also involves removing the carrot from the mozilla wordmark.

schalkneethling:
@ckprice alrighty, switching to static.

ckprice:
@schalkneethling have the [font/footer comments](https://docs.google.com/document/d/1WojRgiXrzXZibKqe-cwGxZJsx6g3jF-D1hOrE_Tn_Qg/edit) been implemented? Can we see an update on demo4 today?

schalkneethling:
@ckprice it has been and should be on demo4, although I have not pushed. I believe myself and @craigcook share demo4. Craig, if you have a moment, I pushed all the latest to https://github.com/mozilla/bedrock/tree/bug1151534-campaign-landing-page Thanks all!

media/css/firefox/personal.less
```diff
@@ -0,0 +1,323 @@
...
+            padding: @baseLine 0;
+            width: 100%;
+        }
+    }
+
+    .mozlogo {
+        .image-replaced();
+        display: inline-block;
+        margin: 0 20px -2px 0;
+        background: url(/media/img/firefox/family/index/family-sprite.png) -345px -485px no-repeat;
```
alexgibson: This is quite a large image sprite to load in just for the wordmark?

alexgibson:
Some CSS properties need overriding for IE on the download button, and some footer alignment issues in Chrome.

layout

Otherwise, this is looking great!

media/css/firefox/personal.less
schalkneethling: Yup, saw that a little while ago, was not aware it contained all of those assets.

schalkneethling:
Thanks @alexgibson I completely neglected to test this in IE again after switching to the flat button. The footer issue in Chrome is weird though, but I will have a look.

schalkneethling:
@alexgibson seeing that all I literally have to do to fix the button in IE is:

```
filter: none;
```

I am just going to stick that into personal.less and not create a separate file containing only that rule. Make sense?

alexgibson:
> I am just going to stick that into personal.less and not create a separate file containing only that rule. Make sense?

I think that should be fine yeah :+1:

craigcook:
I think someone commented elsewhere about the color mismatch between the hero background-image and background-color, which I'm still seeing. We should make that image transparent.

schalkneethling:
@craigcook yeah, for the life of me I can not understand why there is the difference but, making it transparent will definitely solve the issue, good call.

media/css/firefox/personal.less
```diff
@@ -0,0 +1,333 @@
...
+    }
+    h1:before {
+        display: block;
+        background: transparent url(/media/img/firefox/personal/logo.png) 0 0 no-repeat;
+        background-size: 148px;
+        margin: 0 auto @baseLine;
+        width: 148px;
+        height: 148px;
+        content: '';
+        .at2x('/media/img/firefox/personal/logo.png', 295, 295);
```
alexgibson: Looks like you're not supplying a valid unit type here for background-size?

Something like .at2x('/media/img/firefox/personal/logo.png', 148px); should work. Then you shouldn't need the line 103 above.

media/css/firefox/personal.less
```diff
@@ -0,0 +1,333 @@
...
+.mission-icons-sprite(@posx, @posy) {
+    display: block;
+    background-image: url(/media/img/firefox/personal/mission-icons-sprite.png);
+    background-position: @posx @posy;
+    background-size: 175px;
+    background-repeat: no-repeat;
+    margin-top: @baseLine;
+    height: 190px;
+    content: '';
+    .at2x('/media/img/firefox/personal/mission-icons-sprite.png', '50%', 0);
```
alexgibson: Again here I think `.at2x('/media/img/firefox/personal/mission-icons-sprite.png', 175px);` should be what you need? Then line 170 can be removed.

media/css/firefox/personal.less
```diff
@@ -0,0 +1,333 @@
...
+        cursor: pointer;
+        max-width: 300px;
+        max-height: 68px;
+        font-family: 'FiraSans Light';
+        font-weight: bold;
+        .font-size(1.5rem);
+        text-align: center;
+        .inline-block();
+        .border-radius(10px);
+        .transition();
```
alexgibson: Nit: may well be a copy-pasta from some legacy button code, but is `.transition()` a bit heavy handed?

schalkneethling: Yeah, that was a straight copy from .button-flat in buttons.less

media/css/firefox/personal.less
schalkneethling: Ah so that is why, ok. I saw that it takes the path, width, height and also, whether I should specify the actual size of the high-res or the scaled down size. From your comments above it seems the latter though.

Thanks for the info @alexgibson

alexgibson:
@schalkneethling care to squarsh?

schalkneethling:
:hatched_chick:

alexgibson:
r+ :zap:

PR #2970, opened 2015-05-11
Fix Bug 1161938, support other Gecko based browsers

schalkneethling:
@alexgibson r?

### DIFF
--- a/media/js/base/global.js
+++ b/media/js/base/global.js
@@ -87,15 +87,17 @@ function getFirefoxMasterVersion(userAgent) {
     return version;
 }
 
+// Used on the plugincheck page to also support all browsers based on Gecko.
+function isLikeFirefox(userAgent) {
+    var ua = userAgent || navigator.userAgent;
+    return (/Iceweasel/i).test(ua) || (/IceCat/i).test(ua) ||
+        (/SeaMonkey/i).test(ua) || (/Camino/i).test(ua) ||
+        (/like Firefox/i).test(ua);
+}
+
 function isFirefox(userAgent) {
     var ua = userAgent || navigator.userAgent;
-    // camino UA string contains 'like Firefox'
-    return (
-        (/\sFirefox/).test(ua) &&
-        !(/like Firefox/i).test(ua) &&
-        !(/Iceweasel/i).test(ua) &&
-        !(/SeaMonkey/i).test(ua)
-    );
+    return (/\sFirefox/).test(ua) && !isLikeFirefox(ua);
 }
 
 // 2015-01-20: Gives no special consideration to ESR builds
@@ -112,12 +114,6 @@ function isFirefoxUpToDate(latest) {
     }
 
     return (latestFirefoxVersion <= fx_version);
-}
-
-// Used on the plugincheck page to also support SeaMonkey
-function isSeaMonkey(userAgent) {
-    var ua = userAgent || navigator.userAgent;
-    return /SeaMonkey/i.test(ua);
 }
 
 // used in bedrock for desktop specific checks like `isFirefox() && !isFirefoxMobile()`
--- a/media/js/plugincheck/check-plugins.js
+++ b/media/js/plugincheck/check-plugins.js
@@ -152,7 +152,7 @@ $(function() {
     }
 
     // show main download button to non Fx traffic
-    if(!isFirefox() && !isSeaMonkey() ) {
+    if(!isFirefox() && !isLikeFirefox() ) {
         wrapper.addClass('non-fx');
     }
 
@@ -162,7 +162,7 @@ $(function() {
     }
 
     // only execute the plugincheck code if this is Firefox
-    if (isFirefox() || isSeaMonkey()) {
+    if (isFirefox() || isLikeFirefox()) {
         PluginCheck.getPluginsStatus('https://plugins.mozilla.org/en-us/plugins_list.json',
             function(response) {
                 displayPlugins(response);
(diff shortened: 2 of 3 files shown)


media/js/base/global.js
```diff
@@ -114,10 +114,11 @@ function isFirefoxUpToDate(latest) {
...
 }
 
-// Used on the plugincheck page to also support SeaMonkey
-function isSeaMonkey(userAgent) {
+// Used on the plugincheck page to also support all browsers based on Gecko.
+// Not including Camino as it has reached EOL [http://caminobrowser.org/]
+function isLikeFirefox(userAgent) {
     var ua = userAgent || navigator.userAgent;
-    return /SeaMonkey/i.test(ua);
+    return (/Iceweasel/i).test(ua) || (/SeaMonkey/i).test(ua);
```
alexgibson: Think we should also look for `(/like Firefox/i).test(ua)` here too?

alexgibson: Oh, that's because we don't want Camino, right? Do you know if there are any other Gecko browsers that use `like Firefox` in their UA?

schalkneethling: Yes, Camino is EOL so, I do not reckon we need to support it. Not sure which are desktop browsers there are, let me check.

alexgibson: Just thinking about general reuse since this is in global.js. Part of me expects a function `isLikeFirefox` to check for `like Firefox`. Seeing as Camino is EOL I'm fine not including it here, though maybe something more specific would be good?

schalkneethling: Ok so, apparently IceWeasel is now IceCat and although it does have a IceCat string in the userAgent it also uses the like Firefox string. see here:
http://useragentstring.com/pages/Browserlist/

So, we can either check specifically for IceCat and avoid including Camino or just use 'like Firefox' which will then include IceCat and Camino.

I am thinking let's check for IceCat specifically. Thoughts?

alexgibson: I'm happy adding a check for IceCat if it's been renamed. I think we should also add a check for `like Firefox`. If we want to disallow Camino, I think we should check for that specifically.

schalkneethling: I am actually not to concerned to support it or not, so let's check for IceCat and like Firefox in here. Then our basis are covered and the function name more closely match it's result. Will update the PR. Thanks for the feedback.

schalkneethling: hmmmm, seems the last two version of Camino switched to having the string 'Firefox compatible) in the user agent string and not 'like Firefox' anymore:

http://useragentstring.com/Camino8.723_id_19875.php

alexgibson: It still contains the word `Camino` though, which seems more specific?

alexgibson: While we're here we could do a little refactor too to avoid repetition, and change `isFirefox()` to do a check for `!isLikeFirefox()`. Seems neater - thoughts?

schalkneethling:
@alexgibson updated including refactor.

media/js/base/global.js
```diff
@@ -87,15 +87,16 @@ function getFirefoxMasterVersion(userAgent) {
     return version;
 }
 
+// Used on the plugincheck page to also support all browsers based on Gecko.
+function isLikeFirefox(userAgent) {
+    var ua = userAgent || navigator.userAgent;
+    return /Iceweasel/i.test(ua) || /IceCat/i.test(ua)
+        || /SeaMonkey/i.test(ua) || /Camino/i.test(ua);
```
alexgibson: Nit: `||` should come before the line break to appease jshint

media/js/base/global.js
```diff
@@ -87,15 +87,16 @@ function getFirefoxMasterVersion(userAgent) {
...
 function isFirefox(userAgent) {
     var ua = userAgent || navigator.userAgent;
-    // camino UA string contains 'like Firefox'
-    return (
-        (/\sFirefox/).test(ua) &&
-        !(/like Firefox/i).test(ua) &&
-        !(/Iceweasel/i).test(ua) &&
-        !(/SeaMonkey/i).test(ua)
-    );
+    return (/\sFirefox/).test(ua) && !isLikeFirefox(ua);
```
alexgibson: Think you can remove the parenthesis around `/\sFirefox/` here?

alexgibson:
I think this is looking good, thanks @schalkneethling!

I'll give the tests a look over first thing in the morning before merging, but otherwise this is an r+wc from me :dango:

media/js/base/global.js
alexgibson: Scratch that, looks like the parens are needed for old Gecko versions: http://mdn.beonex.com/en/Core_JavaScript_1.5_Reference/Global_Objects/RegExp/test.html

We should stick to using them here.

media/js/base/global.js
```diff
@@ -87,15 +87,16 @@ function getFirefoxMasterVersion(userAgent) {
     return version;
 }
 
+// Used on the plugincheck page to also support all browsers based on Gecko.
+function isLikeFirefox(userAgent) {
+    var ua = userAgent || navigator.userAgent;
+    return (/Iceweasel/i).test(ua) || (/IceCat/i).test(ua) ||
+        (/SeaMonkey/i).test(ua) || (/Camino/i).test(ua);
```
alexgibson: Weren't we going to add a test for `like Firefox` too, or maybe I'm confused?

schalkneethling: Oh, I thought we were replacing like Firefox with specifically checking for Camino but hey, let check for the other as well ;)

alexgibson:
@schalkneethling this all looks good to me, care to squash?

PR #2971, opened 2015-05-11
Remove some leftover playdoh references

jgmize:


jgmize:
@pmclanahan r?

jgmize:
I left the playdoh.wsgi file (and references to it) alone since that is currently used in the apache configs.

pmclanahan:
Wonderful! Thanks so much :clap:

PR #2976, opened 2015-05-13
Fix Bug 1164216, add video to wom page and small other tweaks

schalkneethling:
@alexgibson @craigcook r?

### DIFF
--- a/bedrock/mozorg/templates/mozorg/contribute/friends.html
+++ b/bedrock/mozorg/templates/mozorg/contribute/friends.html
@@ -49,6 +49,16 @@
     <p>{{_('Track your impact and share often for opportunities to be recognized and rewarded.')}}</p>
   </div>
 </section>
+
+{% if LANG.startswith('en') %}
+<hr />
+
+<section class="ff-video">
+  <h2>{{_('Watch the video')}}</h2>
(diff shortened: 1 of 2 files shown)


craigcook:
Since the video is in English I presume we should only show it to en-* locales. You can wrap the whole section in <code>{% if LANG.startswith('en') %}</code>.

Apart from that all looks good to me. r+wc :tomato:

schalkneethling:
Thanks @craigcook !! Very good point, will inquire about whether or not we want to show the video only to en-*

schalkneethling:
Asked https://bugzilla.mozilla.org/show_bug.cgi?id=1164216#c6

bedrock/mozorg/templates/mozorg/contribute/friends.html
```diff
@@ -49,6 +49,16 @@ <h3 class="rewarded">{{_('Get rewarded')}}</h3>
...
 </section>
+
+{% if LANG.startswith('en') %}
+<hr />
+
+<section class="ff-video">
+  <h2>{{_('Watch the video')}}</h2>
+  <iframe src="https://www.youtube.com/embed/ZKcv8HHyyy8" frameborder="0" allowfullscreen></iframe>
+</section>
+{% endblock %}
```
craigcook: oops, should be endif, not endblock

schalkneethling: whoops, meh

PR #2978, opened 2015-05-13
Use travis_retry for all actions requiring network access.

pmclanahan:
Just found out about this. Should reduce our failures due to network glitches.


jgmize:
Nice.

PR #2981, opened 2015-05-13
Bug 1157265: Upgrade Django to 1.7.8.

pmclanahan:
Upgrades some supporting packages as well. Moves all migrations to new format. Has to be deployed in the right order, with the faked migrations one first.


pmclanahan:
@jgmize this is ready for r? The issues I was having with peep failing to upgrade things was a pip version issue. I've solved it by having travis and deployment install a specific known-good version of pip.

jgmize:
r+, nice work!

PR #2983, opened 2015-05-13
Fix Bug 1164572, add higher res hero image and SVG logo with fallback

schalkneethling:
@alexgibson @craigcook r?


bedrock/settings/static_media.py
```diff
@@ -970,6 +970,13 @@
         ),
         'output_filename': 'js/contribute-form-bundle.js',
     },
+    'contribute-friends': {
+        'source_filenames': (
+            'js/libs/jquery-1.7.1.min.js',
```
craigcook: Does it specifically need to be this older version of jQuery? We've updated to 1.11.0 elsewhere, and I think 1.7 has some deprecated stuff that might throw errors.

bedrock/settings/static_media.py
```diff
@@ -970,6 +970,13 @@
...
+            'js/mozorg/contribute/friends.js'
```
craigcook: I think this line needs a trailing comma

craigcook:
Maybe we should also make a cropped down version of the hero image for small viewports, so it's a smaller file to download on mobile.

bedrock/settings/static_media.py
schalkneethling: Oh crap no, I must have overlooked the other version. Will change.

schalkneethling:
@craigcook yeah, @alexgibson also suggested that. Will do.

alexgibson:
This looks good to me! Only nit is you can optimize the SVG to make it a bit smaller. I was able to shave off around 25% using https://jakearchibald.github.io/svgomg/

Other than that, r+wc from me! :cake:

schalkneethling:
Optimized the SVG, just waiting for Travis. Thanks @alexgibson and @craigcook

PR #2985, opened 2015-05-14
Bug 1156383. Change "Televisions" to "Television".

jpetto:


bedrock/firefox/templates/firefox/os/devices-panasonic-tv.html
```diff
@@ -49,7 +49,13 @@
           <a id="device-nav-smartphones" href="#smartphones">{{ _('Smartphones') }}</a>
         </li>
         <li>
-          <a id="device-nav-tvs" href="#tvs">{{ _('Televisions') }}</a>
+          <a id="device-nav-tvs" href="#tvs">
+          {% if l10n_has_tag('fxos_devices_panasonic_tv_update_1') %}
```
flodolo: I wonder if we can get away with `LANG.startswith('en-')` and avoid the fourth update to this page in a week.

jpetto: Would be fine by me, but probably best to ask in [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1156383).

alexgibson:
We're going to push back on this string change in the bug, as the team thinks it makes more sense to leave it as a plural in the longer term. It also sounds nicer along side "devices" which is already plural.

Closing this one, thanks @jpetto!

PR #2988, opened 2015-05-15
Fix Bug 1164424 - Firefox Hello: 'TRY HELLO NOW' button is displayed when using FF38.0esr

kyoshino:
Tested locally with the `browser.uitour.testingOrigins` pref enabled on 38 ESR .

Because this code is only run by Firefox 35+, it might be safe to use `Promise` here, rather than adding another nested callback.

L10n note: There is one new string: “Sorry, Hello is not yet available in Firefox ESR.” Another one, “Get Firefox” is in firefox/desktop/tips.lang so it’s not new. Given the number of Firefox 38 ESR users who will see this message is very small at this time, I’m not sure if we have to wait for l10n. @flodolo how do you think?

### DIFF
--- a/bedrock/firefox/templates/firefox/hello/index.html
+++ b/bedrock/firefox/templates/firefox/hello/index.html
@@ -48,6 +48,10 @@ set share_urls = {
   {% endcall %}
 {% endblock %}
 
+{% block string_data %}
+data-try-now="{{ _('Try it now') }}"
+{% endblock %}
+
 {% block tabzilla_css %}{% endblock %}
 {% block tabzilla_js %}{% endblock %}
 
@@ -218,6 +222,7 @@ set share_urls = {
     <div class="container">
       <div class="ctacopy-wrapper">
         <h2 class="dltry-copy heading-secondary" id="ctacopy-nonfx">{{ _('Get Firefox and start your first conversation') }}</h2>
+        <h2 class="dltry-copy heading-secondary" id="ctacopy-esrfx">{{ _('You need a different version of Firefox to use Hello.') }}</h2>
         <h2 class="dltry-copy heading-secondary" id="ctacopy-oldfx">{{ _('Update your Firefox to start using Hello') }}</h2>
         <h2 class="dltry-copy heading-secondary" id="ctacopy-hellofx">{{ _('Start your first conversation today') }}</h2>
       </div>
--- a/media/js/firefox/hello/index.js
+++ b/media/js/firefox/hello/index.js
@@ -118,13 +118,46 @@
             showFxFooterMessaging();
         }
         // Hello exists in desktop version 35 and up
-        else if (w.getFirefoxMasterVersion() >= 35) {
+        else if (w.getFirefoxMasterVersion() >= 35 && 'Promise' in window) {
             showFxFooterMessaging();
 
-            // see if Hello is an available target in toolbar/overflow/customize menu
-            Mozilla.UITour.getConfiguration('availableTargets', function(config) {
+            Promise.all([
+                // check for the browser channel info
+                new Promise(function(resolve) {
+                    Mozilla.UITour.getConfiguration('appinfo', function(config) {
+                        resolve(config.defaultUpdateChannel);
+                    });
+                }),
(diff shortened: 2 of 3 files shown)


flodolo:
`Get Firefox`: firefox/tips.lang is an opt-in page translated in 42 languages, the Hello page has 56 locales enabled, so it's not necessarily already translated.

If we're adding new strings, "Get Firefox" seems pretty confusing when displayed to a user that's already using ESR.

media/js/firefox/hello/index.js
```diff
@@ -119,12 +119,38 @@
         }
         // Hello exists in desktop version 35 and up
         else if (w.getFirefoxMasterVersion() >= 35) {
-            showFxFooterMessaging();
```
alexgibson: Since `showFxFooterMessaging();` is now inside the promise, if it never fulfills (e.g. uitour doesn't work as expected), the fallback content doesn't show. We should make sure it degrades gracefully still

bedrock/firefox/templates/firefox/hello/index.html
```diff
@@ -218,6 +222,7 @@ <h3 class="heading-tertiary">{{ _('Get Firefox to get started') }}</h3>
     <div class="container">
       <div class="ctacopy-wrapper">
         <h2 class="dltry-copy heading-secondary" id="ctacopy-nonfx">{{ _('Get Firefox and start your first conversation') }}</h2>
+        <h2 class="dltry-copy heading-secondary" id="ctacopy-esrfx">{{ _('Sorry, Hello is not yet available in Firefox ESR.') }}</h2>
```
alexgibson: Looks like Matej suggests `"You need a different version of Firefox to use Hello"` here: https://bugzilla.mozilla.org/show_bug.cgi?id=1164424#c3

bedrock/firefox/templates/firefox/hello/index.html
```diff
@@ -48,6 +48,10 @@
   {% endcall %}
 {% endblock %}
 
+{% block string_data %}
+data-get-firefox="{{ _('Get Firefox') }}"
```
alexgibson: Could this be either `"Try it now"` or `"Get it now"`? https://bugzilla.mozilla.org/show_bug.cgi?id=1164424#c3

alexgibson:
I've tested this change in ESR 38 and FF 35 general release, and it works as expected :+1: 

Code wise, this looks good. Just need to make sure the layout doesn't break if those promises don't resolve and get gobbled up. Other than that, the JS looks fine to me, thanks @kyoshino!

We do still need to iron out the string issues before merging this change though.

media/js/firefox/hello/index.js
```diff
@@ -119,12 +119,38 @@
...
+            Promise.all([
```
alexgibson: Even though this is behind a piece of UA detection, I guess we should still add a feature check too just to be on the safe side?

Lord-of-Darkness:
Desarrolladores y líderes de Mozilla, que son travestis culo!

Alguien debe colgar ustedes para arriba en sus propias bolas y dejar que su perro muerda su caricatura de una polla!

Posteriormente deben ponerlo tiempo que sea necesario por la garganta hasta que traga a sí mismos en el suicidio!

Ustedes Mozilla está ... Estúpido de mierda coños!
--------------------------------------------------------
Developers and leaders of Mozilla, which are shemales ass!

Someone should hang you up in your own balls and let your dog bite his caricature of a cock!

They should then put it long as needed by the throat until he swallows themselves in suicide!

You Mozilla is... Stupid fucking pussies!

kyoshino:
Thanks @alexgibson @flodolo, changed the copy and fixed the code.

kyoshino:
@flodolo the new strings are okay for you? I still think that we don't have to wait for l10n in 56 locales as only a few people will see the message anyway.

flodolo:
Yep, good for me. I'll import "Try it now" from other files, the other string will be pushed when ready.

flodolo:
Note: strings extracted and added to ```firefox/hello.lang```

alexgibson:
This looks great, thanks @kyoshino, r+! :tada:

PR #2989, opened 2015-05-15
Fix Bug 1164781 - Spanish locales are redirected to /en-US instead of /es when page is not localized

kyoshino:


stephendonner:
Should/can we get some tests for this?

kyoshino:
Added a test!

pmclanahan:
Very nice @kyoshino :+1:

PR #2995, opened 2015-05-19
Bug 1159231- Remove old firefox/os page template

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1159231
Remove old firefox ox template, rename accordingly, update references in template, remove images from old template.


flodolo:
I think it would be safe to push this on a demo server?

For sure it breaks localization: the new template expects to find a `/firefox/os/index-new.lang`, we should use the same .lang file to avoid issues with production.

flodolo:
Just to give more context to the previous comment: we currently have two files, `/firefox/os/index.lang` and `/firefox/os/index-new.lang`. 

It's safe to remove `/firefox/os/index.lang` now, since it's not actually used. But, it's not safe to rename `/firefox/os/index-new.lang` to `/firefox/os/index.lang`:
* We'd need to timely sync pushes to prod for Bedrock and SVN.
* We have a lot of references out there to the current template as index-new.lang.

So my suggestion is to keep using `/firefox/os/index-new` as source for strings with ```add_lang_file```, even if the template is renamed to ``/firefox/os/index.html`.

malena:
@flodolo I've added back reference to index-new.

bedrock/firefox/templates/firefox/os/index.html
```diff
@@ -2,13 +2,51 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% from "macros.html" import fxfamilynav with context %}
+
 {% extends "firefox/base-resp.html" %}
 
 {% add_lang_files "firefox/os/devices" %}
```
flodolo: Need to add `firefox/os/index-new` here too.

malena:
Waiting on demo server to be available in a few days.

alexgibson:
I seem to remember there was some shared code between the old template and the new one in `firefox-os-ie.less`. The following lines may also be removable in this PR: https://github.com/mozilla/bedrock/blob/master/media/css/firefox/os/firefox-os-ie.less#L5-L63

Admittedly I haven't tested this, but it looks like there are at least some references to images now being removed here.

malena:
@alexgibson Yes good catch! Will take a look at ie CSS.

alexgibson:
Looks like we can also remove the following JS files as part of this change:

* https://github.com/mozilla/bedrock/blob/master/media/js/firefox/os/desktop.js
* https://github.com/mozilla/bedrock/blob/master/media/js/firefox/os/have-it.js
* https://github.com/mozilla/bedrock/blob/master/media/js/firefox/os/mobile.js

Let's get rid of the following libs as well, as this page looks to be the only bundle using them:

* https://github.com/mozilla/bedrock/blob/master/media/js/libs/jquery.plusslider.js
* https://github.com/mozilla/bedrock/blob/master/media/js/libs/jquery.color.js

malena:
@alexgibson I've removed those files :)

alexgibson:
Great, code-wise this gets an r+ from me, nice work :+1: 

@malena - care to squarsh commits?
@flodolo - are we still waiting on a demo server for this to test l10n before we merge?

flodolo:
Let me test the patch locally, hopefully it won't be necessary.

flodolo:
OK, tested again locally with a few locales, I can't see anything wrong. It's good to go for me :+1:

alexgibson:
Note: we also have #3005 to merge today. Whichever one goes into master first, the PR that's left will likely need a rebase to resolve conflicts. Let's discuss which one to merge first in our meeting later today.

PR #2996, opened 2015-05-19
Fix bug 1152344: Use django-pylibmc for default cache backend.

pmclanahan:
This will not change anything immediately. We still need
to change the local settings in the environments after
deployment of the changes.


pmclanahan:
@jgmize r?

pmclanahan:
Only setting we should need to chage though is `CACHES['default']['BACKEND'] = 'django_pylibmc.memcached.PyLibMCCache'`.

jgmize:
r+, nice work.

PR #2997, opened 2015-05-20
Bug 1164924 - Remove old firefox/sync page template

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1164924


flodolo:
Side note: I've already cleaned up the lang file, if you want to test.
* Deactivate locales still on the old template.
* Removed strings and tags used only in the old template.

malena:
This is currently ready on demo server: https://www-demo3.allizom.org/en-US/ 
I checked it in French and Spanish and saw a few places where translations were not there. Under the Events section. "Community Meetup" "Hour of Code". Is that a concern @flodolo?

flodolo:
Is demo3 hosting all the "clean-up PRs" or just this one?

The actual events are not localizable: we localize the section title, dates, but not the event details.
https://www.allizom.org/fr/

malena:
@flodolo Just this PR.

flodolo:
OK, then it's safe to focus only on https://www-demo3.allizom.org/firefox/sync

One more thing that needs to be tested locally is appropriate redirection for disabled locales with Bedrock set as production (DEV=False). For example bn-BD should redirect to something else (I guess en-US for you), fr should work and look completely localized.

flodolo:
Side note: the patch seems to work fine for me locally.

malena:
@flodolo Thanks! Yeah sorry not sure why I was concerned with home. Still getting used to this whole process. :)

schalkneethling:
All works perfectly locally for me as well. So, all this needs is to marge in the latest from master and then it's a r+ :lemon:

sgarrity:
This has been rebased. @malena walked through the valley the shadow of git in the process. I've confirmed that it matches the original changes that have already been r+'ed and I'll merge as soon as Travis gives the nod.

flodolo:
Adding tag to remember to clean up the tags.

PR #2998, opened 2015-05-20
Bug 1159230 - Remove old mozilla.org home page template

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1159230


sgarrity:
I *think* the [failed test](https://travis-ci.org/mozilla/bedrock/builds/63377548) here is due to a test that has been rendered out-of-date (or obsolete) by this PR.

This test is referencing `mozorg/home.html`, which is no longer used (and is removed by this PR): https://github.com/mozilla/bedrock/blob/master/lib/l10n_utils/tests/test_commands.py#L128

@flodolo Do you know if this test should be updated to point to the new home page template, or if it is still relevant?

flodolo:
I'm not familiar at all with that test, I think we can just move it to a different "stable" template (maybe firefox/new?)

malena:
Test are passing, ready for review.

bedrock/mozorg/views.py
```diff
@@ -422,7 +422,7 @@ def new_home(request, locale=None):
 
 def home(request):
```
jpetto: Now that we only have one home page template, do we need these two different views? Do we still need the `HomeTestView` class above? It seems all code paths now lead to the same `home.html` template.

bedrock/settings/static_media.py
```diff
@@ -583,41 +583,28 @@
...
         'source_filenames': (
-            'css/mozorg/home/home.less',
-            'css/mozorg/home/home-promo.less',
+            'css/mozorg/home/home-ie8.less',
         ),
-        'output_filename': 'css/home-2015-bundle.css',
+        'output_filename': 'css/home-ie8-bundle.css',
     },
-    'home-2015-ie8': {
+    'home-ie9': {
```
jpetto: Is this bundle used anywhere?

bedrock/settings/static_media.py
```diff
@@ -1399,13 +1377,13 @@
             'js/mozorg/home/ga-tracking.js',
             'js/mozorg/home/scroll-prompt.js',
         ),
-        'output_filename': 'js/home-2015-bundle.js',
+        'output_filename': 'js/home-bundle.js',
     },
-    'home-2015-ie9': {
+    'home-ie9': {
```
jpetto: We have a `matchmedia_addlistener` bundle set up for these cases (where it's just the addListener shim needed). Better to use that bundle and remove this one.

bedrock/settings/static_media.py
```diff
@@ -583,41 +583,28 @@
     },
     'home': {
         'source_filenames': (
-            'css/mozorg/home.less',
```
jpetto: Should we get rid of the `*.less` and `*.js` files from the removed template?

bedrock/mozorg/views.py
malena: @jpetto Thanks! I wasn't too sure about this. I will remove HomeTestView and just reference one view.

jpetto: I'm not 100% sure either. :smile: But as far as I can tell, there's no longer a reason to have 2 views and 1 class view. Should just need a single view. @jgmize - Care to weigh in?

If we can/do remove `HomeTestView`, some tests will need updating.

jgmize: Yes, I agree that we should consolidate to one view and update the tests as necessary.

jpetto:
@malena - I think [this gist](https://gist.github.com/jpetto/5376bf49ff61c392449e) will help get the tests functional, though @jgmize may have a more elegant solution.

malena:
@jpetto Thanks!! I've also made the rest of the changes you noted above.

bedrock/settings/static_media.py
```diff
@@ -630,41 +630,22 @@
...
-        ),
-        'output_filename': 'css/home-ie9-bundle.css',
-    },
-    'home-ie8': {
-        'source_filenames': (
-            'css/mozorg/home-ie8.less',
-        ),
-        'output_filename': 'css/home-ie8-bundle.css',
-    },
     'home-ie': {
```
jpetto: Is this bundle still needed? I don't see it included in any templates. If not, let's drop it and remove `home-ie.less`.

bedrock/settings/static_media.py
```diff
@@ -1464,7 +1436,7 @@
         ),
         'output_filename': 'js/growth-firstrun-test2-bundle.js',
     },
-    'home-2015': {
+    'home': {
         'source_filenames': (
             'js/base/mozilla-share-cta.js',
             'js/base/firefox-anniversary-video.js',
```
jpetto: Doesn't look like this video is on the home page anymore. Can probably remove this file from the bundle.

bedrock/settings/static_media.py
```diff
@@ -1464,7 +1436,7 @@
         ),
         'output_filename': 'js/growth-firstrun-test2-bundle.js',
     },
-    'home-2015': {
+    'home': {
         'source_filenames': (
             'js/base/mozilla-share-cta.js',
```
jpetto: Hm, I don't see our `share_cta` macro on the home page either. Maybe remove this file also?

malena:
@jpetto. Good eye! I assumed the stuff under home was current.

bedrock/mozorg/views.py
```diff
@@ -493,20 +480,11 @@ def home_tweets(locale):
     return []
 
 
-def new_home(request, locale=None):
+def home(request, locale=None):
```
jpetto: We can now also drop the `locale` param from this function signature. Just set it to `locale = l10n_utils.get_locale(request)` below.

jpetto:
One last comment on the `home` view definition. Can you squash the commits as well? Should be good after that.

Thanks!

malena:
@jpetto All squashed. Thanks!

jpetto:
Looks great @malena. Thanks!

r+ :zap:

flodolo:
Adding tag to remember to clean up the old lang file.

PR #2999, opened 2015-05-20
Update mozid modal header sizing

alexgibson:
Minor tweak to moz-id modal header styling, as per https://bugzilla.mozilla.org/show_bug.cgi?id=1136937#c27

### DIFF
--- a/media/css/base/mozid-modal.less
+++ b/media/css/base/mozid-modal.less
@@ -31,7 +31,9 @@ html.noscroll body {
             &> header {
                 display: block;
                 height: auto;
-                .font-size(60px);
+                .font-size(48px);
+                line-height: 1.3;
+                padding: 0 (@gridGutterWidth * 3);
                 text-align: center;
                 color: #fff;
                 .open-sans-light;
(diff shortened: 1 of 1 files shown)


media/css/base/mozid-modal.less
```diff
@@ -31,7 +31,8 @@ html.noscroll body {
             &> header {
                 display: block;
                 height: auto;
-                .font-size(60px);
+                .font-size(48px);
+                padding: 0 (@gridGutterWidth * 3);
```
jpetto: One addition - would be nice to drop the line height down to 1.3 or so. 1.5 is a bit much when the text wraps.

jpetto:
Minor request to update `line-height` for header while we're modifying the file. Other than that, r+. Merge when ready!

jpetto:
Oh, and :snake:

media/css/base/mozid-modal.less
alexgibson: Good suggestion :+1:

alexgibson:
Added `line-height: 1.3;` as suggested, merging just as soon as travis is green. Thanks @jpetto!

PR #3002, opened 2015-05-21
Bug 1166324 fxos orange klif cameroon

jpetto:
Both countries do indeed have the same purchase URL.

This needs to go out today.

@alexgibson or @sgarrity r?


alexgibson:
r+ :v:

PR #3003, opened 2015-05-21
Bug 1160480 - update Commit Access Policy to add named vouchers for ...

gerv:
…l10n and remove obsolete info.

Gerv

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/governance/policies/commit/access-policy.html
+++ b/bedrock/mozorg/templates/mozorg/about/governance/policies/commit/access-policy.html
@@ -66,14 +66,11 @@ Therefore, a good policy balances the use of technical and social controls to mi
 
 <p>
 {% trans try='https://wiki.mozilla.org/Build:TryServer',
-         user='https://developer.mozilla.org/docs/Creating_Mercurial_User_Repositories',
-         incubator='https://www.mozilla.org/hacking/incubator-repository.html'
+         user='https://developer.mozilla.org/docs/Creating_Mercurial_User_Repositories'
 %}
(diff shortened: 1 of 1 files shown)


craigcook:
@flodolo I'm pretty sure this page isn't localized. Can you confirm?

flodolo:
Confirmed: no localization.

bedrock/mozorg/templates/mozorg/about/governance/policies/commit/access-policy.html
```diff
@@ -99,21 +96,13 @@ <h3 id="Level1a">{{ _('Level 1a - Named Voucher') }}</h3>
 <table class="table module-table">
   <tr>
     <td>mozilla.org/trunk</td>
-    <td>David Boswell or Reed Loden</td>
+    <td>Reed Loden</td>
   </tr>
   <tr>
     <td>mozilla.com/tags/production</td>
     <td>John Slater</td>
```
craigcook: Not really a code comment, I just happened to notice this: John Slater is no longer with Mozilla, and I'm not sure who would be the current voucher for this repository. But we're also trying to stop using it as we move everything to git/bedrock. Maybe just delete the whole entry?

craigcook:
All looks good code-wise, just the question of whether we should also remove John Slater's name or even remove the whole entry. r+wc.

bedrock/mozorg/templates/mozorg/about/governance/policies/commit/access-policy.html
gerv: I know John has gone but I didn't know what to replace it with. If you are trying to stop using these two repos (note there are two with his name on) then we can just remove the entries. Does that work for you guys?

jbertsch:
Let's remove the entries with John's name.  Thanks, Craig!

bedrock/mozorg/templates/mozorg/about/governance/policies/commit/access-policy.html
craigcook: I checked with Jen. Let's remove them both since we're discouraging any use of those repos. Thanks!

gerv:
Done.

craigcook:
I squashed these two commits and merged in https://github.com/mozilla/bedrock/commit/f2875f8b57583b8c015c9d05deab997f05445243

PR #3006, opened 2015-05-22
Bug 1142147 Update content services screenshot

sgarrity:


craigcook:
The new high res image is a bit smaller than the one it's replacing. Is that intentional? 

The new one isn't quite twice the size of the base image (it's like 170% instead of 200%), but that is still plenty big enough to look good on a retina screen. Just want to verify it's the size you intended.

sgarrity:
@craigcook Good eye, sir! Yeah, I didn't get a large enough image to do a 2x high-res image. I think this is sufficient though - it's not shown at the 1-to-1 pixel ratio anyhow.

PR #3009, opened 2015-05-25
Fix Bug 1168097, change color, label of outdated plugins button

schalkneethling:
@alexgibson r? blue makes more sense then gray ;)


flodolo:
Why are we introducing a new string? I don't see any explanation in the bug, and to be honest we're having too many of these small changes lately across mozilla.org

flodolo:
Side note: i never noticed, but the button doesn't look good with long text
screenshot 2015-05-25 21 46 53

alexgibson:
I wonder for the button icons, could we just use a single PNG with a transparent arrow, so it works with any button color style now or in the future? We can probably fix the layout issue @flodolo spotted with a little extra padding and centering the vertical position of the icon.

As for the string change, I will leave this up to @flodolo and yourself to discuss. But I agree unless the change is really warranted, it might be worth not asking localizers to re-translate.

schalkneethling:
@alexgibson yes, I will update these icons to be as such, good point.

Then @flodolo - The main thing is to create a greater distinction between outdated and vulnerable plugins, as it currently seems to trip up a lot of users and cause some confusion. Have the same wording, icon and button color obviously adds to this but, I get the issue with translating so, I am thinking perhaps a color change and icon change is better.

Will also fix up the padding/centering with regards to long words. Thanks!

alexgibson:
This change looks good to me, only question is the string change - do we think losing the word 'now' from 'Update now' is warranted enough to require re-translation? Thoughts @schalkneethling and @flodolo?

flodolo:
Personally I don't think we should add a new string.

Just to give you an idea: this page has already 6 tags, that's probably the page with more tags if we exclude home page. We had a lot of small updates.

schalkneethling:
@alexgibson @flodolo nope, oversight. I meant to change it back.

PR #3013, opened 2015-05-27
[fix bug 1168744] Remove jQuery animations from mozilla modal

alexgibson:
@jpetto r?

A small but worthwhile perf improvement I think. Thoughts? 

I noticed Chrome shows a fair amount of jank opening the modal on /firefox/hello/, and exhibits the same perf on other pages where the new mozid modal styling is going to be used.

### DIFF
--- a/media/css/base/mozid-modal.less
+++ b/media/css/base/mozid-modal.less
@@ -20,6 +20,7 @@ html.noscroll body {
     left: 0;
     z-index: 9999999;
     overflow: auto;
+    .animation(sand-fade-in 0.3s ease-in 0s 1 normal both);
 
     .window {
         padding: 20px;
--- a/media/css/base/mozilla-modal.less
+++ b/media/css/base/mozilla-modal.less
@@ -20,6 +20,7 @@ html.noscroll body {
     left: 0;
     z-index: 9999999;
     overflow: auto;
+    .animation(sand-fade-in 0.3s ease-in 0s 1 normal both);
 
     .window {
         padding: 20px;
--- a/media/js/base/mozilla-modal.js
+++ b/media/js/base/mozilla-modal.js
@@ -78,10 +78,7 @@ Mozilla.Modal = (function(w, $) {
             }
         });
 
-        $modal.hide();
-        $modal.fadeIn('fast', function() {
-            $modal.focus();
-        });
+        $modal.focus();
 
         // close with escape key
         $d.on('keyup.' + evtNamespace, function(e) {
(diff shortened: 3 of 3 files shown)


media/js/base/mozilla-modal.js
```diff
@@ -78,10 +78,8 @@ Mozilla.Modal = (function(w, $) {
             }
         });
 
-        $modal.hide();
-        $modal.fadeIn('fast', function() {
-            $modal.focus();
-        });
+        $modal.show();
```
jpetto: Is this line necessary?

alexgibson: I believe so yeah?

media/css/base/mozid-modal.less
```diff
@@ -20,6 +20,7 @@ html.noscroll body {
     left: 0;
     z-index: 9999999;
     overflow: auto;
+    .animation(sand-fade-in 0.1s ease-in 0s 1 normal both);
```
jpetto: This animation is a little too fast to notice going in, and doesn't provide a nice fade out when dismissing the modal.

I wonder if we could do a class with `opacity: 1` along with a transition, then toggle that class in the JS?

alexgibson: So I used an animation here, as transitions don't occur when an element changes from display:none to display:block. This method also allows older browsers to degrade gracefully.

I can adjust the animation timing :+1: 

As for the fade out, this would require a little timeout trickery before hiding the modal element again. Worth it?

media/js/base/mozilla-modal.js
jpetto: Hm, I removed it (seemingly) without consequence. I think the only place the `#modal` element is ever set to `display: none` is on the (removed) line 81.

alexgibson: Hmm, ok i will double-check and remove it if so :+1:

alexgibson:
Removed the redundant 'show()' and increased the fade time to 300ms.

We don't have a fade-out anim like before, think we can live with that?

media/css/base/mozilla-modal.less
```diff
@@ -20,6 +20,7 @@ html.noscroll body {
     left: 0;
     z-index: 9999999;
     overflow: auto;
+    .animation(sand-fade-in 0.1s ease-in 0s 1 normal both);
```
jpetto: Should we up this fade-in time as well?

media/js/base/mozilla-modal.js
```diff
@@ -115,16 +112,17 @@ Mozilla.Modal = (function(w, $) {
...
         }
 
-        $('#modal').fadeOut('fast', function() {
-            $contentParent.append($content);
-            $(this).remove();
+        var $md = $('#modal');
 
-            // allow page to scroll again
-            $html.removeClass('noscroll');
+        $md.hide();
```
jpetto: Could probably get rid of this `hide()` call, too, right?

media/css/base/mozilla-modal.less
alexgibson: doh, good catch

alexgibson:
Updated, thanks @jpetto

jpetto:
Looking good. I don't mind the lack of transition when dismissing the modal. I might actually prefer it.

If we get push back, could do some CSS class/transition/ontransitionend type thing.

Anyway, r+! :baby_chick:

jpetto:
Oh, also, I didn't notice any performance difference between this PR and production in Chrome (or Fx), but I'm on board with dropping jQuery animations wherever we can, so, yeah. Merging.

PR #3014, opened 2015-05-27
Bug 1166983. Update headline on /firefox/new

jpetto:


bedrock/firefox/templates/firefox/new.html
```diff
@@ -93,12 +93,15 @@
...
+      {% if LANG == 'en-US' or l10n_has_tag('spring_release') %}
+        {# L10n: Line break below for visual formatting only. #}
+        {{ _('When it’s personal,<br>choose Firefox.') }}
+      {% elif l10n_has_tag('fx10_independent') %}
         {{ _('Choose Independent.') }}
         <br>
         {{ _('Choose Firefox.') }}
       {% else %}
-        {%- if LANG in ['en-GB', 'en-US', 'en-ZA'] -%}
+        {%- if LANG in ['en-GB', 'en-ZA'] -%}
```
flodolo: I think we should drop completely this piece of conditional at this point and use ```Committed to you, your privacy and an open Web``` as fallback.

bedrock/firefox/templates/firefox/new.html
```diff
@@ -93,12 +93,15 @@
     <div class="row">
       <img src="{{ static('img/firefox/new/firefox-logo.png') }}" alt="Firefox" />
       <h1 class="large">
-      {% if l10n_has_tag('fx10_independent') %}
+      {% if LANG == 'en-US' or l10n_has_tag('spring_release') %}
```
flodolo: Reading the bug this should be en-US only, so no need for a tag, right?

jpetto: Cory mentioned in IRC that the headline should eventually be served to all locales. Do we need the `LANG == 'en-US'` check in this case?

bedrock/firefox/templates/firefox/new.html
jpetto: Sounds good. Will update.

bedrock/firefox/templates/firefox/new.html
ckprice: @jpetto @flodolo sorry Jon, I gave you some bad info. Let's keep this JUST en-US only. I think it may eventually go to other locales, but the message needs to be vetted for a global audience before it can be shown.

jpetto:
Should go out with spring campaign.

alexgibson:
r+ on the code change, good to merge when ready.

alexgibson:
We got the go-ahead to start merging

PR #2932, opened 2015-04-24
Bug 1136937. Fx 38.0.5 firstrun.

jpetto:

PR #3015, opened 2015-05-27
Bug 1161808 - Pocket landing page

craigcook:
I still need to test IE and may have some cleanup relating to that but this is ready to start code review.


media/css/sandstone/buttons.less
```diff
@@ -18,7 +18,7 @@
     padding: 12px (@baseLine * 2);
     margin-bottom: @baseLine;
     background: @buttonFlatBlue;
-    color: #fff;
+    color: #fff !important;
```
craigcook: I'm not proud of this.

media/js/firefox/pocket.js
```diff
@@ -0,0 +1,75 @@
...
+
+    // Check Firefox version
+    if (window.isFirefox()) {
+        if (window.isFirefoxUpToDate()) {
+            $html.addClass('firefox-latest');
+        } else {
+            $html.addClass('firefox-old');
+        }
+
+        hasTarget('pocket').then(function(result) {
```
alexgibson: Might want to move this up inside the UITour.ping callback on line 55?

craigcook: Cood call. I also don't actually need the firefox-old class, that was a leftover from testing.

bedrock/firefox/templates/firefox/pocket.html
```diff
@@ -0,0 +1,116 @@
...
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/base-resp.html" %}
+
+{% block page_title %}{{ _('Pocket') }}{% endblock %}
+{% block body_id %}firefox-pocket{% endblock %}
+{% block body_class %}
+  {{ super() }}
+  mozid
```
alexgibson: Hmm, this won't conflict - but we should tread carefully (cc @jpetto), as there's a `mozID` body class being added to sandstone in another PR here: https://github.com/mozilla/bedrock/pull/2932/files#diff-8a894a393259a3ba520a09de584b6d1dR174.

It looks like the [family page](https://github.com/mozilla/bedrock/blob/master/media/css/firefox/family/index.less#L7) also has a body class (lowercase) of mozid. I don't think anything should break here or conflict due to case sensitivity, but we should probably be standardize this in some way.

bedrock/firefox/templates/firefox/pocket.html
```diff
@@ -0,0 +1,116 @@
...
+      </h2>
+
+      {# Has toolbar button #}
+      <p class="main-cta" id="cta-try"><button type="button" class="button-flat try-pocket">{{ _('Try it now') }}</button></p>
+
+      {# No toolbar button #}
+      <p class="main-cta" id="cta-learn"><a href="https://support.mozilla.org/kb/where-pocket-button-firefox" class="button-flat">{{ _('Learn more') }}</a></p>
+
+      {# Old Firefox or Not Firefox #}
+      <p class="main-cta" id="cta-download"><a href="{{ url('firefox.new') }}?scene=2#download-fx" class="button-flat-green">{{ _('Download Firefox') }}</a></p>
```
alexgibson: Should we use a real download button here?

bedrock/firefox/templates/firefox/pocket.html
```diff
@@ -0,0 +1,116 @@
...
+
+    </div>
+  </section>
+
+  <section class="content-secondary">
+    <div class="content">
+      {# L10n: Pocket features below are emphasized with <strong></strong> #}
+      <ul>
+        <li>
+          <img src="{{ static('img/firefox/pocket/pocket-logo.svg') }}" width="122" height="122" alt="">
```
alexgibson: Are we bothered about older browser to add a png fallback?

bedrock/firefox/templates/firefox/pocket.html
```diff
@@ -0,0 +1,116 @@
...
+        <a href="{{ url('legal.index') }}">{{ _('Legal Notices') }}</a>
+      </li>
+    </ul>
+  </div>
+</footer>
+{% endblock %}
+
+{% block js %}
+  <!--[if IE 9]>
+    {% javascript 'matchmedia_addlistener' %}
```
alexgibson: Is this needed? I don't see `matchMedia` being used anywhere?

media/css/firefox/pocket.less
```diff
@@ -0,0 +1,282 @@
...
+
+    .content {
+        padding: (@baseLine * 2) 0 (@baseLine * 3);
+    }
+
+    .mozlogo {
+        .image-replaced();
+        display: inline-block;
+        margin: 0 auto @baseLine;
+        background: url(/media/img/firefox/family/index/family-sprite.png) -345px -485px no-repeat;
```
alexgibson: Large image sprite to load just for the wordmark?

media/css/sandstone/buttons.less
alexgibson: I'm not proud of you either :trollface:

bedrock/firefox/templates/firefox/pocket.html
```diff
@@ -0,0 +1,116 @@
...
+
+        {# Current Firefox #}
+        <span class="has-firefox">{{ _('All you need is your Firefox Account.') }}</span>
+
+        {# Old Firefox or Not Firefox #}
+        <span class="get-firefox">{{ _('Get Firefox to start using Pocket right in your browser.') }}</span>
+      </h2>
+
+      {# Has toolbar button #}
+      <p class="main-cta" id="cta-try"><button type="button" class="button-flat try-pocket">{{ _('Try it now') }}</button></p>
```
alexgibson: If I load the page with pocket in the toolbar, then remove the icon while still on the page, licking theis button tries to load `/firefox/pocket/undefined`, which is a 404. I would perhaps change this to a regular link that goes to the SUMO article as a fallback, should someone decide to remove it after reading the page.

bedrock/firefox/templates/firefox/pocket.html
craigcook: Nope, don't need it, just a leftover. Will remove.

media/css/firefox/pocket.less
craigcook: Oops, you're right. Copied this CSS from the family page but need a different image.

bedrock/firefox/templates/firefox/pocket.html
craigcook: Ah, good idea!

bedrock/firefox/templates/firefox/pocket.html
craigcook: It's really just IE8, and I was thinking it could live without these images but it shows the broken image placeholder, which does look pretty bad... I could fix it with CSS but then I need to do a separate style sheet just for IE8 and that's more overhead than I think it warrants. So I'll use your JavaScript trick to include PNG fallbacks.

Then again, I could just use PNG for all browsers with two resolutions, and that could be a CSS sprite and cut three requests down to one... but then it's not SVG and SVG is cool. They're presentational images, really, so maybe they belong in CSS anyway...

OK, I've talked myself into it. Switching CSS pseudo-elements and a PNG sprite.

alexgibson: Heh, you don't even need me to discuss the best solution (if only I had a rubber-duck emoji)

bedrock/firefox/templates/firefox/pocket.html
craigcook: After much discussion on IRC, we need the button to say "Download Firefox" to be clear on what the person is downloading (they're not downloading Pocket), so this needs to be a link to /new. And we'll link directly to scene 2 because we lose a lot of potential downloads when folks have to click an extra time, while scene 2 fires the download automatically. This doesn't work in older IEs but we're willing to accept that loss.

alexgibson:
r+ :ghost:

PR #3017, opened 2015-05-28
Fix 1169026: Update SMS message IDs for widget

pmclanahan:
Leaves old sms view to send the old message ID.


pmclanahan:
This won't work. Have to change basket.

PR #3018, opened 2015-05-28
Fix 1169026: Update SMS message IDs for widget

pmclanahan:
Leaves old sms view to send the old message ID.

Have to wait for the basket changes to go live to merge this.


pmclanahan:
Blocked by the changes to basket in mozilla/basket#136 going live.

pmclanahan:
Based on feedback from Ben Niolet we're not doing this right now.

PR #3020, opened 2015-05-28
Bug 1168235 optimizely templates

craigcook:


bedrock/firefox/templates/firefox/releases/beta-notes.html
```diff
@@ -5,3 +5,9 @@
 {% extends "firefox/releases/release-notes.html" %}
 
 {% set channel_name = 'Beta' %}
+
+{% block optimizely %}
+  {% if waffle.switch('firefox-releasenotes-optimizely') %}
+     {% include 'includes/optimizely.html' %}
+  {% endif %}
+{% endblock %}
```
jgmize: This isn't necessary because you've already added it to the parent template.

jgmize:
r+wc

bedrock/firefox/templates/firefox/releases/beta-notes.html
craigcook: Ah, you're right, of course. I was rushing and didn't think it through. Sorry bout that.

PR #3024, opened 2015-05-29
[fix bug 1164295] Add pixel to /new for downloads

ckprice:
Buelle(r?)

### DIFF
--- a/media/js/firefox/new.js
+++ b/media/js/firefox/new.js
@@ -49,6 +49,8 @@
     var virtual_url = ('/' + locale + '/products/download.html' +
                        query_str + 'referrer=' + referrer);
 
+    var pixelAdded = false;
+
     var $html = $(document.documentElement);
 
     if (isFirefox()) {
@@ -170,11 +172,44 @@
                     $scene1.css('visibility', 'hidden');
                     $thankYou.focus();
                 }, 500);
+                // Added measurements fire only when the download is fired
+                if (!pixelAdded) {
+                    pixelAdded = true;
+                    spring_campaign_measurement();
+                }
             }
         }
 
         function show_scene_anim(scene) {
             show_scene(scene, true);
+        }
+
+        // Pixel to be removed on July 31st, 2015 (Bug 1168440)
+        function spring_campaign_measurement () {
+            var _dntStatus = navigator.doNotTrack || navigator.msDoNotTrack;
+            var fxMatch = navigator.userAgent.match(/Firefox\/(\d+)/);
+            var ie10Match = navigator.userAgent.match(/MSIE 10/i);
+            var w8Match = navigator.appVersion.match(/Windows NT 6.2/);
+
+            if (fxMatch && Number(fxMatch[1]) < 32) {
+                // Can't say for sure if it is 1 or 0, due to Fx bug 887703
+                _dntStatus = 'Unspecified';
+            } else if (ie10Match && w8Match) {
+                // IE10 on Windows 8 does not Enable based on user intention
+                _dntStatus = 'Unspecified';
+            } else {
+                _dntStatus = { '0': 'Disabled', '1': 'Enabled' }[_dntStatus] || 'Unspecified';
+            }
+
+            if (_dntStatus !== 'Enabled'){
+                var $body = $('body');
+                var $pixel = $('<img />', {
+                    width: '1',
+                    height: '1',
+                    src: 'https://servedby.flashtalking.com/spot/8/6247;40428;4669/?spotName=Mozilla_Download_Conversion'
+                });
+                $body.append($pixel);
+            }
         }
 
         // Pull download link from the download button and add to the


media/js/firefox/new.js
```diff
@@ -171,12 +171,39 @@
...
+            var fxMatch = navigator.userAgent.match(/Firefox\/(\d+)/);
+
+            if (fxMatch && Number(fxMatch[1]) < 32) {
+                // Can't say for sure if it is 1 or 0, due to Fx bug 887703
+                _dntStatus = (_dntStatus === 'yes') ? 'Unknown' : 'Unspecified';
+            } else {
+                _dntStatus = { '0': 'Disabled', '1': 'Enabled' }[_dntStatus] || 'Unspecified';
+            }
+
+            if (_dntStatus !== 'Enabled'){
```
schalkneethling: Myself and @alexgibson had a long chat about this and we both believe that this should be:

```
if (_dntStatus === 'Disabled') {
  ...
```
Just throwing it out there ;)

ckprice: @schalkneethling thanks for this. Guidance from the privacy team is that DNT should be tied to user intention, rather than default browser settings.

media/js/firefox/new.js
```diff
@@ -171,12 +171,39 @@
...
         }
 
         function show_scene_anim(scene) {
             show_scene(scene, true);
         }
 
+        // Pixel to be removed on July 31st, 2015 (Bug 1168440)
+        function spring_campaign_measurement () {
+            var _dntStatus = navigator.doNotTrack || navigator.msDoNotTrack;
+            var fxMatch = navigator.userAgent.match(/Firefox\/(\d+)/);
```
pmclanahan: We could (should?) always track IE since we can't know the users' choice since it's "enabled" by default.

media/js/firefox/new.js
schalkneethling: :+1:

media/js/firefox/new.js
```diff
@@ -171,12 +171,39 @@
...
+                _dntStatus = { '0': 'Disabled', '1': 'Enabled' }[_dntStatus] || 'Unspecified';
+            }
+
+            if (_dntStatus !== 'Enabled'){
+                var $body = $('body');
+                var $pixel = $('<img />', {
+                    width: '1',
+                    height: '1',
+                    src: 'https://servedby.flashtalking.com/spot/8/6247;40428;4669/?spotName=Mozilla_Download_Conversion',
+                    class: 'flash-talking'
```
schalkneethling: as @alexgibson mentioned, this bit needs to be removed as class is a reserved word.

media/js/firefox/new.js
ckprice: Good catch. We need to exclude IE from this check. Looking for code now.

ckprice:
Added a match to exclude IE (notably IE10) from this check, as DNT was Enabled be default, thus not implying user intention.

@jpetto r?

pmclanahan:
Once we work out this logic, we should also wrap GA and GTM in it for bug 1136169.

media/js/firefox/new.js
```diff
@@ -171,12 +171,41 @@
                     $thankYou.focus();
                 }, 500);
             }
+
+            // Added measurements fire only when the download is fired
+            spring_campaign_measurement();
```
jpetto: This should probably be moved into the `if (scene === 2)` block above, right?

Also, should add logic to make sure pixel is only added once. Currently, if you hit `/firefox/new`, click the CTA, then click the back button, the `spring_campaign_measurement` function is called again.

media/js/firefox/new.js
```diff
@@ -171,12 +171,41 @@
...
+        // Pixel to be removed on July 31st, 2015 (Bug 1168440)
+        function spring_campaign_measurement () {
+            var _dntStatus = navigator.doNotTrack || navigator.msDoNotTrack;
+            var fxMatch = navigator.userAgent.match(/Firefox\/(\d+)/);
+            var ieMatch = navigator.userAgent.match(/MSIE/);
+
+            if (fxMatch && Number(fxMatch[1]) < 32) {
+                // Can't say for sure if it is 1 or 0, due to Fx bug 887703
+                _dntStatus = (_dntStatus === 'yes') ? 'Unknown' : 'Unspecified';
+            } else if (ieMatch) {
```
jpetto: Since it's only IE10 on Windows 8 that has DNT enabled without explicitly being set by the user, we should only disregard DNT for that specific combo. Users on IE9 or IE10 on Windows 10 should have their DNT setting respected.

ckprice:
@jpetto r?

media/js/firefox/new.js
```diff
@@ -170,6 +170,8 @@
                     $scene1.css('visibility', 'hidden');
                     $thankYou.focus();
                 }, 500);
+                // Added measurements fire only when the download is fired
+                spring_campaign_measurement();
```
jpetto: This line can still be called twice - click CTA, click back button, click CTA again. Not a very likely scenario, but would be nice to put a check in before adding the pixel a second time.

ckprice:
These are the two main URL's this needs to fire at

/en-US/firefox/new/#download-fx
/en-US/firefox/new/?scene=2#download-fx

Thanks

media/js/firefox/new.js
```diff
@@ -177,6 +184,34 @@
...
+        // Pixel to be removed on July 31st, 2015 (Bug 1168440)
+        function spring_campaign_measurement () {
+            var _dntStatus = navigator.doNotTrack || navigator.msDoNotTrack;
+            var fxMatch = navigator.userAgent.match(/Firefox\/(\d+)/);
+            var ie10Match = navigator.userAgent.match(/MSIE 10/i);
+            var w8Match = navigator.appVersion.match(/Windows NT 6.2/);
+
+            if (fxMatch && Number(fxMatch[1]) < 32) {
+                // Can't say for sure if it is 1 or 0, due to Fx bug 887703
+                _dntStatus = (_dntStatus === 'yes') ? 'Unknown' : 'Unspecified';
```
jpetto: Could we just set `_dntStatus = 'Unspecified'` here? Regardless of the `yes` check, it ends up adding the pixel, right?

jpetto:
r+. Just waiting for Travis.

PR #3027, opened 2015-06-01
Fix bug 1152533 - Firefox Friends home promo

craigcook:


craigcook:
Adding a Do Not Merge for the moment while we decide for certain whether the new promo should be available to non-English locales. If we truly want it to be English-only I can unwrap the strings and put it in <code>if LANG.startswith('en')</code>.

schalkneethling:
Tested locally in Fx, Chrome and IE8, all looks good. Code is good so, other than waiting for the l10n question, this is a r+ Thx @craigcook :cake:

craigcook:
After discussion on IRC, we'll go ahead and expose strings and display the promo to any locales that translate it, even though the friends.mozilla.org site may not be localized in the same language. So I don't think we need any code changes, the usual l10n tag is good to go.

flodolo:
LGTM :+1:

PR #3028, opened 2015-06-02
[fix bug 1170243] Add link to Optimization Game on /whatsnew 40.0a2

alexgibson:


alexgibson:
@jpetto r?

alexgibson:
Still waiting on confirmation in the bug if this should also be added to /firstrun: https://bugzilla.mozilla.org/show_bug.cgi?id=1170243#c3

bedrock/firefox/templates/firefox/dev-firstrun-spring-campaign.html
```diff
@@ -85,6 +85,9 @@
         <h2>{{_('Performance Tools')}}</h2>
         <p>{{_('Examine your web content’s frame rate, memory consumption and JS profile in a consolidated timeline view to help easily identify and troubleshoot performance bottlenecks.')}}</p>
         <a href="https://developer.mozilla.org/docs/Tools/Performance" rel="external" class="more">{{_('Learn more about Performance Tools')}}</a>
+        {% if LANG.startswith('en-') %}
+          <a href="http://codepen.io/mozhacks/pen/xGgevw?editors=001?utm_source=mozilla.org&amp;utm_medium=referral&amp;utm_campaign=firefox-developer-game&amp;utm_content=firstrun-dev-tour" rel="external" class="more">{{_('Play the game on Codepen')}}</a>
```
jpetto: I can't decide if it would be better to just put a `<br>` here instead of adding the CSS rule. Thoughts?

alexgibson: I toyed over this as well, maybe <br> would better suit l10n?

alexgibson: Updated to remove the block level links and use a `<br>` instead

ckprice:
> Still waiting on confirmation in the bug if this should also be added to /firstrun and product page (I'm assuming yes): https://bugzilla.mozilla.org/show_bug.cgi?id=1170243#c3

Confirming that we'll be adding this to everywhere the Performance Tools blurb is (firstrun, whatsnew, dev ed).

jpetto:
r+! :tent:

PR #3031, opened 2015-06-03
Fix bug 1168947: Upgrade django-mozilla-product-details to 0.7

pmclanahan:


pmclanahan:
@jgmize r?

pmclanahan:
@sylvestre I have the cache timeout for product-details set for 1 hour. Is that okay or should we try to make it shorter? This will mean that changes to p-d will show up in at most 1 hour.

jgmize:
@pmclanahan should this PR also move the `./mange.py update_product_details` call from the deployment scripts to a cron job?

pmclanahan:
good point. we'll need a cronjob that will update it and sync it out. I'll add that.

pmclanahan:
Added cron jobs and update scripts. I think we should leave the current chief p-d deployment alone in case of emergency update. I'm looking into adding a sanity check to the files before push.

jgmize:
@pmclanahan would you mind rebasing this?

pmclanahan:
@jgmize done.

bin/update-scripts/stage/update-stage-product-details.sh
```diff
@@ -0,0 +1,12 @@
...
+
+DEPLOY_SCRIPT=/data/bedrock-stage/deploy
+WORKING_DIR=/data/bedrock-stage/src/www.allizom.org-django/bedrock
+SITE_NAME=www.allizom.org
+PYTHON=../venv/bin/python
+PD_PATH=lib/product_details_json
+
+cd $WORKING_DIR
+$PYTHON manage.py update_product_details
+$DEPLOY_SCRIPT -q "${SITE_NAME}-django/bedrock/$PD_PATH"
```
jgmize: It would be nice if this were refactored into a single common script that all the environments used. We could set all the environment variables in the cron template instead of the scripts themselves.

pmclanahan: I agree, but I think a lot of these scripts could use some refactoring for such a thing, so I'd rather do that in a separate PR in order to get them all on the same page. That large a refactor doesn't belong in this PR IMHO.

etc/cron.d/bedrock-prod.tmpl
```diff
@@ -9,6 +9,7 @@ MAILTO="webops-cron@mozilla.com,cron-bedrock@mozilla.com"
 
 */10 * * * * {{ user }} {{ source }}/bin/update-scripts/prod/update-prod-php.sh {{ log }}
 */15 * * * * {{ user }} {{ source }}/bin/update-scripts/prod/update-prod-locale-cron.sh {{ log }}
+0    * * * * {{ user }} {{ source }}/bin/update-scripts/prod/update-prod-product-details.sh {{ log }}
```
jgmize: @sylvestre mentioned in irc that he would prefer that this update with the same frequency that we do the rnasync. Any objections to changing that first `0` to a `*/5` here and on the stage template?

pmclanahan: I could, but I'd also need to reduce the cache timeout to 5 min as well or it'd be pointless. Seems a little short, but it would probably be okay.

PR #3033, opened 2015-06-03
Reorder contribute.json IRC contacts.

pmclanahan:
contributebot mentions the first 3 in its message.


pmclanahan:
Thought this order would be better for the welcome message. If anyone else would rather be the one pinged just say so.

PR #3035, opened 2015-06-04
Fix bug 1171181: Respond correctly when no redirect found.

pmclanahan:
Remove spaces from old file name in case its valid. Return 410
(gone) if no redirect url can be found.

### DIFF
--- a/bedrock/security/views.py
+++ b/bedrock/security/views.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 import re
 
+from django.core.urlresolvers import NoReverseMatch
 from django.db.models import Q
 from django.utils.decorators import method_decorator
 from django.views.decorators.http import last_modified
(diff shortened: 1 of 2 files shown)


pmclanahan:
@jgmize r?

bedrock/security/views.py
```diff
@@ -205,19 +206,22 @@ class KVRedirectsView(CachedRedirectView):
...
+                version = '{0}.{1}'.format(*version)
+                return reverse('security.product-version-advisories', kwargs={'product': product,
+                                                                              'version': version})
+
+            if url_component.endswith('ESR'):
+                return reverse('security.product-advisories',
+                               kwargs={'slug': url_component[:-3] + '-esr'})
+
+            return reverse('security.product-advisories', kwargs={'slug': url_component})
+        except NoReverseMatch:
```
jgmize: NIT: there is a non-trivial amount of logic inside this try block that has nothing to do with the actual exception. It works, but what if instead of having three different `return reverse(...)` that could each raise a `NoReverseMatch` you just stored the `viewname` and `reverse_kwargs` for each of them in variables and had a single call to `return reverse(viewname, reverse_kwargs)` inside the try block?

jgmize:
Other than my refactoring suggestion, this lgtm: feel free to merge as-is, or I'd be happy to take another look if you decide to change it.

bedrock/security/views.py
pmclanahan: Yeah. I didn't like it much easier, but was lazy. Good call though. I'll fix it up.

pmclanahan:
refactor complete. much better IMHO. take another quick look if you can.

bedrock/security/views.py
```diff
@@ -205,19 +206,26 @@ class KVRedirectsView(CachedRedirectView):
...
+        url_component = kwargs['filename'].replace(' ', '')
+        try:
+            return reverse(**self.get_redirect_args(url_component))
+        except NoReverseMatch:
+            return None
+
+    def get_redirect_args(self, url_component):
         if url_component == 'suite17':
-            return reverse('security.product-advisories', kwargs={'slug': 'mozilla-suite'})
+            return dict(viewname='security.product-advisories',
```
jgmize: NIT: using the literal `{}` to construct the `dict` is more readable (IMO) and faster.

jgmize:
yes, much better, thanks.

r+, you can construct your dicts however you want :)

bedrock/security/views.py
pmclanahan: yeah, mostly did it to avoid a larger diff, also because nested literal dicts aren't very readable IMHO. But I'd be happy either way really.

PR #3038, opened 2015-06-05
Deploy to webheads in series.

pmclanahan:
This is intended to cause commander to run `deploy_app`
for each host in series instead of parallel like it does now.
It also adds a 5s wait before allowing the next code push to happen.

TODO: Add a check for a 200 response code from the host before proceeding to
the next host.

### DIFF
--- a/bin/update/deploy_base.py
+++ b/bin/update/deploy_base.py
@@ -6,13 +6,13 @@ the systems that need it.
 """
 import random
 import re
+import time
 import urllib
 import urllib2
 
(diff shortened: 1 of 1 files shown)


pmclanahan:
@jgmize r?

jgmize:
This lgtm, but I'd like to test it before merging. I think it would be worth putting this branch on the `mozilla` repo so we can test deploying it to the stage environment, since that's the only non-prod environment with multiple web heads.

pmclanahan:
Done. We can test when we've both got a min today.

PR #3040, opened 2015-06-05
[fix bug 1171216] Campaign landing page footer off on Chrome at 760px breakpoint

alexgibson:
TIL: Chrome still considers `content: '';` as valid content on pseudo elements, needs `content: none;` instead.


schalkneethling:
r+ :watermelon:

PR #3041, opened 2015-06-05
[fix bug 1171899] Campaign landing page - unsupported platforms link color

alexgibson:
* Fixes download button link color for unsupported platforms and iOS messaging.
* Also tweaks the dl button link target a little, to fill its parent container.

### DIFF
--- a/media/css/firefox/personal.less
+++ b/media/css/firefox/personal.less
@@ -107,50 +107,65 @@ body {
         .at2x('/media/img/firefox/personal/logo.png', 148px);
     }
     .download-button.download-button-simple {
-        /* copied from buttons.less :: .button-flat - with amendments */
-        background: #6fbe4a;
         color: #fff;
-        padding: 12px (@baseLine * 2);
         margin-bottom: 0;
         border: none;
-        cursor: pointer;
-        max-width: 300px;
-        max-height: 68px;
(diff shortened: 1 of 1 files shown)


schalkneethling:
@alexgibson How do I trigger the alternate button text? I reckon for iOS I would use the simulator or Chrome but unsupported?

alexgibson:
> @alexgibson How do I trigger the alternate button text? I reckon for iOS I would use the simulator or Chrome but unsupported?

The simplest way is just to use the inspector, and change the class name on `<html>` from `osx` to `ios`, or `oldwin`.

schalkneethling:
@alexgibson Ah yes, that is even simpler :) Should these just be plain text? I am wondering for the 'Sign up to learn more', could this possible be styled as a flat green button?

schalkneethling:
I am wondering, could we do something like:

```
.oldwin .download-button .unsupported-download {
    max-width: 350px;
}
```
screen shot 2015-06-05 at 16 05 13

That way the text will be displayed on only two lines and less scrunched up.

alexgibson:
Good suggestion :+1:

alexgibson:
@schalkneethling updated, thanks

schalkneethling:
r+ :melon:

PR #3045, opened 2015-06-08
Bug 1162057. Implement aurora-bars mixin.

jpetto:
- Remove older, image-based implementations.


alexgibson:
Tested these pages on various browsers and they look good to me, r+! :coffee:

PR #3046, opened 2015-06-08
Fix bug 1171183 in LocaleURLMiddleware

jgmize:
Strip characters from url params that cannot be encoded in utf-8


jgmize:
@pmclanahan r?

PR #3048, opened 2015-06-09
Adds functional tests for /plugincheck

schalkneethling:
@alexgibson r?

### DIFF
--- a/tests/functional/plugincheck.js
+++ b/tests/functional/plugincheck.js
@@ -1,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global casper */
+'use strict';
+
+var config = require('../lib/config');
+var helpers = require('../lib/helpers');
+var path = '/plugincheck/';
+var url = config.base() + path;
+
+casper.test.begin('PluginCheck, Elements: ' + url, 5, function suite(test) {
+
+    casper.start();
+
+    casper.userAgent(config.userAgent.fx);
+
+    casper.thenOpen(url, function() {
+        test.assertHttpStatus(200);
+        test.assertResourceExists(/media\/img\/plugincheck\/screenshots\/addons-screenshot/);
+        // The tests below ensure that the newsletter section is always visible irrespective of browser
+        test.assertVisible('#newsletter-form', 'The newletter signup is visible in Fx');
+
+        casper.userAgent(config.userAgent.chrome);
+    });
+
+    casper.thenOpen(url, function() {
+        test.assertVisible('#newsletter-form', 'The newletter signup is visible in webkit');
+        test.assertVisible('#download-button-desktop-release', 'Download button visible in webkit');
+    });
+
+    casper.run(function() {
+        test.done();
+        helpers.done();
+    });
+});
+
+casper.test.begin('PluginCheck, conditional content Firefox: ' + url, 2, function suite(test) {
+
+    casper.start();
+
+    casper.userAgent(config.userAgent.fx);
+
+    casper.thenOpen(url, function() {
+        test.assertVisible('.plugin-status-container h2', 'Plugin Status heading exists');
+        test.assertNotVisible('.not-supported', 'Unsupported browser message not visible');
+    });
+
+    casper.run(function() {
+        test.done();
+        helpers.done();
+    });
+});
+
+casper.test.begin('PluginCheck, conditional content Webkit: ' + url, 2, function suite(test) {
+
+    casper.start(url, function() {
+        test.assertVisible('.not-supported', 'Unsupported browser message visible');
+        test.assertNotVisible('.plugin-status-container h2', 'Plugin Status heading not visible');
+    });
+
+    casper.run(function() {
+        test.done();
+        helpers.done();
+    });
+});
+
+casper.test.begin('PluginCheck, outated Firefox: ' + url, 1, function suite(test) {
+
+    casper.start();
+
+    casper.userAgent(config.userAgent.fxOutdated);
+
+    casper.thenOpen(url, function() {
+        test.assertVisible('.version-message-container', 'Update Firefox button visible');
+    });
+
+    casper.run(function() {
+        test.done();
+        helpers.done();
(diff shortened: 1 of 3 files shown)


tests/functional/plugincheck.js
```diff
@@ -0,0 +1,103 @@
...
+    casper.userAgent('Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:38.0) Gecko/20100101 Firefox/38.0');
+
+    casper.thenOpen(url, function() {
+
+        test.assertHttpStatus(200);
+
+        test.assertVisible('.plugin-status-container h2', 'Plugin Status heading exists');
+        test.assertNotVisible('.not-supported', 'Unsupported browser message not visible');
+
+        casper.userAgent('Mozilla/5.0 (Macintosh; Intel Mac OS X 10_10_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/43.0.2357.81 Safari/537.36');
```
alexgibson: Is this resetting the US back to the default? If so, would it make sense to do this automatically after every test in the `done()` function in `/lib/helpers.js`?

tests/functional/plugincheck.js
```diff
@@ -0,0 +1,103 @@
...
+    });
+
+    casper.thenOpen(url, function() {
+
+        test.assertHttpStatus(200);
```
alexgibson: Probably don't need to check the response code in every test? In other examples, I've just done this once as the first test in every suite

tests/functional/plugincheck.js
alexgibson: Never mind, this gets reset after each test block :+1:

tests/functional/plugincheck.js
schalkneethling: I was wondering about that. I was thinking that each block will reload the page so, perhaps just ensuring it actually loaded before stepping into the real tests might be a good 'sanity' check but, no reason to overdo things.

You reckon I should remove?

alexgibson: I think once should be enough, since using it everywhere quickly runs up the number of total tests?

schalkneethling: I am wondering though, perhaps for the two tests where the userAgent is changed, perhaps here it makes sense to do a status check inside `.thenOpen`?

tests/functional/plugincheck.js
```diff
@@ -0,0 +1,91 @@
...
+casper.test.begin('PluginCheck, Elements' + url, 4, function suite(test) {
+
+    casper.start();
+
+    casper.userAgent(config.userAgent.fx);
+
+    casper.thenOpen(url, function() {
+        test.assertHttpStatus(200);
+        test.assertResourceExists(/media\/img\/plugincheck\/screenshots\/addons-screenshot(\.\w{1,}){0,}\.jpg/);
+        // The tests below ensure that the newsletter section is always visible irrespective of browser
```
alexgibson: Is there any logic in the page to hide the newsletter? If not, is the extra test below necessary?

tests/functional/plugincheck.js
```diff
@@ -0,0 +1,91 @@
...
+    casper.run(function() {
+        test.done();
+        helpers.done();
+    });
+});
+
+casper.test.begin('PluginCheck, Download button present: ' + url, 1, function suite(test) {
+
+    casper.start(url, function() {
+        test.assertVisible('#download-button-desktop-release', 'Download button visible');
```
alexgibson: Could this be moved up into the first test?

tests/functional/plugincheck.js
```diff
@@ -0,0 +1,91 @@
...
+
+var config = require('../lib/config');
+var helpers = require('../lib/helpers');
+var path = '/plugincheck/';
+var url = config.base() + path;
+
+casper.options.pageSettings.loadPlugins = true;
+
+
+casper.test.begin('PluginCheck, Elements' + url, 4, function suite(test) {
```
alexgibson: Nit: need a space after `Elements`

tests/functional/plugincheck.js
schalkneethling: No, I guess if shows in the one, the other will be fine as well.

tests/functional/plugincheck.js
```diff
@@ -0,0 +1,91 @@
...
+
+/* global casper */
+'use strict';
+
+var config = require('../lib/config');
+var helpers = require('../lib/helpers');
+var path = '/plugincheck/';
+var url = config.base() + path;
+
+casper.options.pageSettings.loadPlugins = true;
```
alexgibson: Is this needed?

tests/functional/plugincheck.js
schalkneethling: Then again, seeing I am moving https://github.com/mozilla/bedrock/pull/3048#discussion-diff-32004107R84 up into this test, no harm in keeping the additional assert is there?

tests/functional/plugincheck.js
schalkneethling: Yup, moving.

tests/functional/plugincheck.js
schalkneethling: Nah, it is going to have no effect, removing.

tests/functional/plugincheck.js
```diff
@@ -0,0 +1,91 @@
...
+
+casper.test.begin('PluginCheck, Elements' + url, 4, function suite(test) {
+
+    casper.start();
+
+    casper.userAgent(config.userAgent.fx);
+
+    casper.thenOpen(url, function() {
+        test.assertHttpStatus(200);
+        test.assertResourceExists(/media\/img\/plugincheck\/screenshots\/addons-screenshot(\.\w{1,}){0,}\.jpg/);
```
alexgibson: Could this be simplified to just `test.assertResourceExists(/media\/img\/plugincheck\/screenshots\/addons-screenshot/);` maybe - thoughts?

schalkneethling: probably, I just tested it directly in a Regexp tester thingy and it seemed good and stable, but I reckon it does not matter what the file type is, nor those numbers although, perhaps the function will care.

schalkneethling: Seems it is totally happy with that :cake:

alexgibson:
r+ :banana:

PR #2986, opened 2015-05-14
Bug 1162048 use mozid modal

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1162048

This PR replaces the old mozilla-modal styles with the new mozid-modal. This was prompted by new modal implemented on this page: https://www.mozilla.org/en-US/firefox/hello/

The following pages were updated to work with the new styles:
https://www.mozilla.org/en-US/firefox/35.0a2/firstrun/ 
http://www.mozilla.org/en-US/about/manifesto/ 
https://www.mozilla.org/en-US/firefox/geolocation/
https://www.mozilla.org/en-US/firefox/developer/ 
https://www.mozilla.org/en-US/foundation/annualreport/2012/ 
https://www.mozilla.org/en-US/foundation/annualreport/2013/
https://www.mozilla.org/en-US/firefox/new/ 
https://www.mozilla.org/en-US/firefox/os/devices/ 
https://www.mozilla.org/en-US/firefox/partners/
https://www.mozilla.org/en-US/contribute/

Modals were removed from pages no longer using them:
https://www.mozilla.org/en-US/firefox/installer-help/ - modal no longer used

This page was not updated because mozilla-modal.js was not used to implement modal
https://www.mozilla.org/en-US/foundation/annualreport/2011/

These pages are no longer accessible:
firefox_os -  template to be removed in separate bug
firefox_firstrun -  removed modal,  modal no longer used in template
firefox_os_mwc_2015_preview - updated to new modal in case this becomes active

PR #3050, opened 2015-06-10
Fix Bug 1146683 - The php_url helper function is broken

ishanroy:


bedrock/mozorg/helpers/misc.py
```diff
@@ -35,17 +35,6 @@ def convert_to_high_res(url):
...
-@jinja2.contextfunction
-def php_url(ctx, url):
-    """Process a URL on the PHP site and prefix the locale to it."""
-    locale = getattr(ctx['request'], 'locale', None)
-
-    # Do this only if we have a locale and the URL is absolute
-    if locale and url[0] == '/':
-        return path.join('/', locale, url.lstrip('/'))
-    return url
-
```
jpetto: There are 1 too many blank lines between functions here, which is why the tests aren't passing. There should be two blank lines in-between each function.

jpetto:
Looks like there's one more use of `php_url` in `bedrock/foundation/templates/foundation/trademarks/policy.html` on [line 82](https://github.com/mozilla/bedrock/blob/master/bedrock/foundation/templates/foundation/trademarks/policy.html#L82) that needs to be updated.

jpetto:
Thanks for taking this on @ishanroy!

In addition to the two tasks requested above, we need to complete one more thing to finalize this pull request. To keep the git history clean, [squash your commits](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) and update your commit message to reference the bug number and description. For this PR, I would recommend a commit message like:

`[fix bug 1146683] Remove php_url helper function.`

If you need any assistance squashing your commits, just comment here and someone will help you along.

One final note - it's generally best to create a separate branch when working on a bug. For example, if you look at PR #2986, you can see the developer created a branch named after the bug (`bug-1162048-use-mozid-modal`). You don't have to create a new branch for this PR, but keep this approach in mind for future pull requests. :smile:

ishanroy:
Thanks for your help @jpetto. Will keep in mind to work on branches in future.

jpetto:
This change is looking great @ishanroy. Thanks!

r+ :cake:

ishanroy:
Thanks @jpetto. Was my first ever bug. Happy to help!

PR #3051, opened 2015-06-10
[fix bug 1172375] Update Hello FTE redirect for non-Firefox users from /new to Hello product page

alexgibson:
The Hello FTE URL is only accessed via an in-product link. If someone should happen to come across the link in another browser, we currently redirect to `/firefox/new/` using the `LatestFxView` class (the same one that gets used on `/whatsnew` and `/firstrun`). 

This change creates `HelloFxView` just for Hello, which redirects to `/firefox/hello` instead (there may well be a simpler / nicer way to do this, but I thought it may be worth using a separate class for Hello, since `LatestFxView` logic may change in the future)

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -620,7 +620,55 @@ def hello_screen_sharing(version):
     return version >= Version('38.0.5')
 
 
-class HelloStartView(LatestFxView):
+class HelloFxView(TemplateView):
+
+    """
+    Base class to be extended by views that require visitor to be
+    using latest version of Firefox. Classes extending this class must
+    implement either `get_template_names` function or provide
+    `template_name` class attribute.
+    """
+
+    @vary_on_headers('User-Agent')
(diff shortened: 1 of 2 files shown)


alexgibson:
@pmclanahan r?

bedrock/firefox/views.py
```diff
@@ -620,7 +620,55 @@ def hello_screen_sharing(version):
     return version >= Version('38.0.5')
 
 
-class HelloStartView(LatestFxView):
+class HelloFxView(TemplateView):
```
pmclanahan: This is a lot of duplicated code from `LatestFxView`. We could just update `LatestFxView

pmclanahan: I see your note about a separate class for this to avoid changes, but we can handle that if/when it comes up I think.

alexgibson: Thanks @pmclanahan - happy to defer to yourself on what you think is best for now. Closing this :)

PR #3052, opened 2015-06-10
[fix bug 1172375] Update Hello FTE redirect for non-Firefox users from /new to Hello product page

pmclanahan:


pmclanahan:
@alexgibson this is my rework of your commit from #3051. Let me know what you think.

alexgibson:
Thanks @pmclanahan - I agree this is the simpler solution for now. Tested locally and works as expected, thanks for the assist! :cake:

bedrock/firefox/views.py
```diff
@@ -414,16 +417,18 @@ def redirect_to(self):
...
         query = self.request.META.get('QUERY_STRING')
         query = '?' + query if query else ''
 
         user_agent = self.request.META.get('HTTP_USER_AGENT', '')
         if 'Firefox' not in user_agent:
-            return reverse('firefox.new') + query
+            return reverse(self.non_fx_redirect) + query
             # TODO : Where to redirect bug 757206
 
+        return None
```
jgmize: NIT: I personally prefer to leave this as implicit, but to each their own.

pmclanahan: It was implicit, but agibson had made his explicit, then the Zen crept into my head, and this is much more obvious... to each their own I guess. :guardsman:

PR #3057, opened 2015-06-12
Bug 1171763 - Delete two pages in research and remove from nav

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1171763


alexgibson:
This change looks good, thanks @malena! 

I noticed there are still a couple of links to the deleted pages here that also need removing: https://github.com/mozilla/bedrock/blob/master/bedrock/research/templates/research/research.html#L37-L44

Also, is it worth adding a redirect for the two removed URL's back to `/research`, should they be linked from anywhere else?

alexgibson:
I think these four URL patterns can also be removed: https://github.com/mozilla/bedrock/blob/master/bedrock/research/urls.py#L10-L13

Looks like `research/collaborations` and `research/publications` are danglers that shouldn't be there, perhaps they can be tidied up in the redirect too?

malena:
@alexgibson Cleaned up.

alexgibson:
This looks great thanks @malena! r+ :tangerine:

PR #3058, opened 2015-06-12
Add functional tests for Firefox Developer page

alexgibson:


tests/functional/firefox-developer.js
```diff
@@ -0,0 +1,66 @@
...
+    casper.run(function() {
+        test.done();
+        helpers.done();
+    });
+});
+
+casper.test.begin('Firefox Developer, Click download button: ' + url, 1, function suite(test) {
+
+    casper.start(url, function() {
+        this.click('.intro .download-button .os_win .download-link');
```
schalkneethling: Nice that we do not have to care about the host OS here. I reckon we will test that the correct button/message is shown when we add tests to test the download button specifically.

schalkneethling:
General question, do we need to test whether the share widget is available on pages that should have them? I was wondering about that on the plugincheck page as well.

schalkneethling:
Other than the general question above. r+ :cake:

tests/functional/firefox-developer.js
alexgibson: Yep, I think this test is mostly useful to ensure that no JS errors in the page prevent the download from happening, so which platform link we click doesn't really matter - the logic is the same. 

We currently have existing & more extensive download link tests in the mcom repo. It may make sense to keep the more complete tests there, but something we can look at & discuss in more detail later. Doing complete visual tests for the buttons is something that would be good as a stand alone test suite, but tricky to do as it depends on more than just user agent.

alexgibson:
With regard to the share widget, sure I can add a check just to make sure the element is there. I wouldn't say it's critical to the page though.

alexgibson:
@schalkneethling - added a check for the share widget, thanks for th r!

schalkneethling:
@alexgibson yeah, the share widget is not critical, just a general observation.

alexgibson:
@schalkneethling - think we're good to merge this one?

PR #3059, opened 2015-06-13
A/B testing for several tasks under /contribute/signup/

comzeradd:
@craigcook it may need some minor UI fixes, but you can start looking the code.


flodolo:
General note: don't use `\'`, use proper curly quotes.

I also don't see a bug reference, so I'm missing all the background of this PR. But keep in mind that this page is heavily localized, so strings need to be wrapped in l10n calls like `_('string')` and we need a system to hide content until localized, unless the first phase is en-US only.

comzeradd:
This is about [bug 1168359](https://bugzilla.mozilla.org/show_bug.cgi?id=1168359). I opened each variation separately then, but it makes more sense to be in one commit.

flodolo:
My personal feeling is that we shouldn't involve localized versions of the page in this, and just wait for a winner.

LucyeoH:
Hey @craigcook @comzeradd just checking in on the status of this? Is it still on track for the 18th?

craigcook:
@LucyeoH I'm reviewing/testing today, still on track for a merge and production push on the 18th unless we hit any major issues.

flodolo:
@craigcook Can you clarify if this is going to be en-US only?

I see there's a `/contribute/tasks` associated to the PR. As it is, that URL will be enabled for all locales, since the bound lang file is `mozorg/contribute/index`. But there are about 20 new strings in this PR.

craigcook:
Yes, only English for the testing, no l10n needed. Would it be best to unhook the lang file on these new pages?

flodolo:
From a quick test, removing the `add_lang_file` directive let us redirect visitors to en-US when visiting `contribute/tasks` (with production settings). Might be worth a second test, also not sure if there are other URLs to verify.

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,182 @@
...
+            <li class="option">
+                <a data-task-id="2" class="ab-task category-stumbling" href="https://oneanddone.mozilla.org/en-US/tasks/92/" target="_blank"></a>
+                <span class="label-text">{{ _('Download our Stumbler App') }}</span>
+            </li>
+            <li class="option">
+              <a data-task-id="3" class="ab-task category-donate" href="https://sendto.mozilla.org/page/contribute/givenow-seq#page-1" target="_blank"></a>
+              <span class="label-text">{{ _('Donate to Mozilla') }}</span>
+            </li>
+            <li class="option">
+              &nbsp;
```
craigcook: The page was obviously designed for four options per row, so I can see why you added these empty items to make it work with the layout. But I think we could instead update the CSS to handle different cases where there are two or three items per row, adjusting widths to keep them centered.

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,182 @@
...
+    <p class="section-tagline">{{ _('There are a lots of ways to contribute to Mozilla. Pick one below that you think would be right for you.') }}</p>
+  </div>
+{% endblock %}
+
+{% block contrib_content %}
+  {% if variation == 'a1' %}
+    <section class="section inquiry">
+      <div class="container inquiry-choices">
+        <div class="option-group">
+          <h4>{{ _('Only have a little time?') }}</h4>
```
craigcook: Let's make these <code>h2</code>s and we can adjust the font size in CSS (<code>h2</code> is more semantically correct but a little too big while <code>h4</code> is a little too small)

craigcook:
Responsive/mobile layout is pretty badly broken.

tasks-mobile

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,182 @@
...
+            </li>
+          </ul>
+        </div>
+      </div>
+    </section>
+  {% endif %}
+
+  <section class="section inquiry">
+    <div class="container center">
+      <a href="{{ url('mozorg.contribute.tasksthankyou') }}" class="tasks-more">{{ _('Not ready to try a task?') }} »</a>
```
craigcook: You can use the <code>button-flat</code> class to get the standard button styles, but you may still need another style hook to increase the font size.

media/css/mozorg/contribute/contribute-2015.less
```diff
@@ -948,6 +948,31 @@ textarea:focus {
 
     .container {
         padding-top: 0;
+
+        .tasks-more {
```
craigcook: With the <code>button-flat</code> class you won't need all these other properties, you might just need one for the font size.

media/css/mozorg/contribute/contribute-2015.less
```diff
@@ -948,6 +948,31 @@ textarea:focus {
...
+           background-color: #0096DD;
+           color: #FFF;
+           display: inline-block;
+           border: 0px none;
+           border-radius: 0.5em;
+           text-transform: uppercase;
+           font-weight: bold;
+           cursor: pointer;
+           transition: all 0.15s ease-in-out 0s;
+           font-size: 1.125rem;
```
craigcook: We have a mixin that takes a pixel font size value and outputs rem with a px fallback. This could be <code>.font-size(18px);</code>

media/css/mozorg/contribute/contribute-2015.less
```diff
@@ -956,6 +981,77 @@ textarea:focus {
         margin: 0;
         padding: 0;
     }
+
+    .signup-option {
+        font-size: 1.5em;
```
craigcook: Use the <code>.font-size()</code> mixin here

media/css/mozorg/contribute/contribute-2015.less
```diff
@@ -956,6 +981,77 @@ textarea:focus {
...
+.inquiry-choices {
+    ul {
+        list-style: none;
+    }
+
+    a {
+        position: relative;
+        display: block;
+
+        &:before {
```
craigcook: Instead of duplicating all this CSS it'll be simpler to hang these icons from a class we can apply equally to links or labels, thought that will mean also adding the class to the current form as well.

craigcook:
The HTML is looking pretty good, just a few minor nitpicks. Most of the fixes it needs on the front-end are in the CSS and I'm happy to make those tweaks in a separate PR if you'd like the help.

I'll defer to @pmclanahan and @jgmize to review the python bits. A test failed in Travis but I'm not sure what the problem is.

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks-survey.html
```diff
@@ -0,0 +1,64 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "mozorg/contribute/contribute-base.html" %}
+
+{% add_lang_files "mozorg/contribute/index" %}
```
flodolo: Since this needs to be en-US only, remove this line.

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks-thankyou.html
```diff
@@ -0,0 +1,27 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "mozorg/contribute/contribute-base.html" %}
+
+{% add_lang_files "mozorg/contribute/index" %}
```
flodolo: Since this needs to be en-US only, remove this line.

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,182 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "mozorg/contribute/contribute-base.html" %}
+
+{% add_lang_files "mozorg/contribute/index" %}
```
flodolo: Since this needs to be en-US only, remove this line.

comzeradd:
@craigcook PR updated

bedrock/mozorg/views.py
```diff
@@ -162,6 +162,54 @@ def contribute_signup(request):
...
+    def get_context_data(self, **kwargs):
+        cxt = super(ContributeTasks, self).get_context_data(**kwargs)
+        variation = self.request.GET.get('variation', 'b2')
+        cxt['variation'] = variation
+        return cxt
+
+    template_name = 'mozorg/contribute/contribute-tasks.html'
+
+
+def contribute_tasks(request):
```
pmclanahan: Why not just use `ContributeTasks.as_view()` in `urls.py`?

bedrock/mozorg/views.py
```diff
@@ -162,6 +162,54 @@ def contribute_signup(request):
...
+    return ContributeTasks.as_view()(request)
+
+
+class ContributeTasksSurvey(l10n_utils.LangFilesMixin, FormView):
+    template_name = 'mozorg/contribute/contribute-tasks-survey.html'
+    form_class = ContributeTasksForm
+
+    def get_context_data(self, **kwargs):
+        cxt = super(ContributeTasksSurvey, self).get_context_data(**kwargs)
+        task = self.request.GET.get('task', '')
```
pmclanahan: possible XSS vector. We should make sure that the `task` value is indeed being escaped in the template. I'd be more comfortable if we were validating the kind of value we're expecting `task` to be.

bedrock/mozorg/views.py
```diff
@@ -162,6 +162,54 @@ def contribute_signup(request):
     return view_class.as_view()(request)
 
 
+class ContributeTasks(l10n_utils.LangFilesMixin, TemplateView):
+
+    def get_context_data(self, **kwargs):
+        cxt = super(ContributeTasks, self).get_context_data(**kwargs)
+        variation = self.request.GET.get('variation', 'b2')
```
pmclanahan: Same XSS concern as below.

media/css/mozorg/contribute/contribute-2015.less
```diff
@@ -958,6 +971,16 @@ textarea:focus {
     }
 }
 
+.inquiry-choices {
+    ul {
+        list-style: none;
+    }
+
+    .option-double {
+        width: 440px;
```
craigcook: This is too wide in small viewports (320px is the narrowest we really plan for). You can add another rule in the mobile media query setting this to <code>width: auto;</code>

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,159 @@
...
+        <div class="option-group">
+          <h2>{{ _('Only have a little time?') }}</h2>
+          <ul class="option-list">
+            <li class="option">
+              <a data-task-id="1" class="ab-task category category-nightly" href="https://nightly.mozilla.org/" target="_blank"></a>
+              <span class="label-text">{{ _('Help us test by downloading Firefox Nightly') }}</span>
+            </li>
+            <li class="option">
+                <a data-task-id="2" class="ab-task category category-stumbling" href="https://oneanddone.mozilla.org/en-US/tasks/92/" target="_blank"></a>
+                <span class="label-text">{{ _('Download our Stumbler App') }}</span>
```
craigcook: Sorry I didn't notice this before, but these spans should be inside the links so the text is linked as well as the icon. Just moving the <code>&lt;/a&gt;</code> seems to work fine with no other CSS tricks needed, except we might then want to change the link color to match the body text color instead of the regular blue.

media/css/mozorg/contribute/contribute-2015.less
comzeradd: I added media queries for both tablet/mobile views.

bedrock/mozorg/views.py
comzeradd: I added some simple regex (similar to the one for the categories in the current signup page) to limit the variables to only the accepted range of values.

comzeradd:
PR updated. @craigcook I could use some advise on where to insert the js snippet from optimizely at the signup template. Ideally it should be inside the `<head>` since it may lead to redirection.

craigcook:
We have a <a href="https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/base-resp.html#L75">block in the base template</a> especially for Optimizely, so just wrap it in that and it'll appear in the head. You can (and possibly should) also wrap it in a page-specific waffle switch so we can easily turn the snippet off or on without pushing code changes. See PR #3020 for a recent example.

craigcook:
The pages are looking really close to done but could still use some CSS tweaks to adjust the layout when there are less than four options (I'd like to get the option groups centered rather than shifted left). Unfortunately I'm traveling today and won't have a chance to work on it or do any further review until tonight (Pacific time) so I don't know if this will make it to production on the 18th after all. @alexgibson might be able to help with some front-end review while I'm unavailable,

We usually try to avoid production pushes on Fridays just because there's always the risk of something going wrong and ruining everyone's weekend, but this may be low-risk enough to push on Friday if it's important. Or else we could finish it up and do a push next week from Whistler, though that's less than ideal as well.

media/css/mozorg/contribute/contribute-2015.less
```diff
@@ -968,6 +1001,12 @@ textarea:focus {
         margin-left: 10px;
     }
 
+    .section-subtitle {
+        margin: 10px;
+        text-align: left;
+        font-size: 1rem;
```
alexgibson: This should use the .font-size() mixin

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,176 @@
...
+
+{% block contrib_content %}
+  {% if variation == '1' %}
+    <section class="section inquiry">
+      <div class="container inquiry-choices">
+        <div class="option-group">
+          <h2>{{ _('Only have a little time?') }}</h2>
+          <ul class="option-list">
+            <li class="option">
+              <a data-task-id="1" class="ab-task category category-nightly" href="https://nightly.mozilla.org/" target="_blank">
```
alexgibson: Just a nit, but it would be nice if these links warned the user they open a new window. Could maybe add `title="{{ _('Opens new window') }}"` to these?

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,176 @@
...
+                <span class="label-text">{{ _('Help us test by downloading Firefox Nightly') }}</span>
+              </a>
+            </li>
+            <li class="option">
+                <a data-task-id="2" class="ab-task category category-stumbling" href="https://oneanddone.mozilla.org/en-US/tasks/92/" target="_blank">
+                  <span class="label-text">{{ _('Download our Stumbler App') }}</span>
+                </a>
+            </li>
+            <li class="option">
+              <a data-task-id="3" class="ab-task category category-activism" href="https://sendto.mozilla.org/page/contribute/givenow-seq#page-1" target="_blank">
```
alexgibson: Remove `#page1` hash in the URL here, else you can't see the header when the donate page loads?

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks-survey.html
```diff
@@ -0,0 +1,62 @@
...
+  <div class="container">
+    <h1 class="page-title">{{ _('Thanks!') }}</h1>
+    <p class="section-tagline">
+      {{ _('We’re thrilled to see that you’re interested in contributing and can’t wait to get you started.') }}
+      {{ _('Keep an eye on your inbox for an important message from us.') }}
+    </p>
+
+    {% if task %}
+      <div class="confirm-body survey">
+        <iframe src="http://www.surveygizmo.com/s3/2184089/Participation-ratings-survey-questions-for-each-selection-7-questions?task={{ task }}" frameborder="0" scrolling="no"></iframe>
```
alexgibson: Can the survey be `https://` here?

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,176 @@
...
+      </div>
+    </section>
+  {% elif variation == '2' %}
+    <section class="section inquiry">
+      <div class="container inquiry-choices">
+        <div class="option-group">
+          <h2>{{ _('Only have a little time?') }}</h2>
+          <ul class="option-list option-double">
+            <li class="option">
+              <a data-task-id="3" class="ab-task category category-activism" href="https://sendto.mozilla.org/page/contribute/givenow-seq#page-1" target="_blank">
```
alexgibson: Again remove `#page1`?

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,176 @@
...
+        <div class="option-group">
+          <h2>{{ _('Try something simple.') }}</h2>
+          <ul class="option-list">
+            <li class="option">
+              <a data-task-id="6" class="ab-task category category-video" href="https://air.mozilla.org/channels/project-meeting/" target="_blank">
+                <span class="label-text">{{ _('Attend a Mozilla community meeting right now!') }}</span>
+              </a>
+            </li>
+            <li class="option">
+              <a data-task-id="3" class="ab-task category category-activism" href="https://sendto.mozilla.org/page/contribute/givenow-seq#page-1" target="_blank">
```
alexgibson: Again remove `#page1`?

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,176 @@
...
+        <div class="option-group">
+          <h2>{{ _('Only have a little time?') }}</h2>
+          <ul class="option-list">
+            <li class="option">
+              <a data-task-id="1" class="ab-task category category-nightly" href="https://nightly.mozilla.org/" target="_blank">
+                <span class="label-text">{{ _('Help us test by downloading Firefox Nightly') }}</span>
+              </a>
+            </li>
+            <li class="option">
+                <a data-task-id="2" class="ab-task category category-stumbling" href="https://oneanddone.mozilla.org/en-US/tasks/92/" target="_blank">
```
alexgibson: I know this page is en-US only, but we should probably still remove the hard-coded locales in the links on these pages. This could just be: `https://oneanddone.mozilla.org/tasks/92/`

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,176 @@
...
+                <span class="label-text">{{ _('Attend a Mozilla community meeting right now!') }}</span>
+              </a>
+            </li>
+            <li class="option">
+              <a data-task-id="3" class="ab-task category category-activism" href="https://sendto.mozilla.org/page/contribute/givenow-seq#page-1" target="_blank">
+                <span class="label-text">{{ _('Donate to Mozilla') }}</span>
+              </a>
+            </li>
+            <li class="option">
+              <a data-task-id="2" class="ab-task category category-stumbling" href="https://oneanddone.mozilla.org/en-US/tasks/92/" target="_blank"></a>
```
alexgibson: Same here remove hardcoded `en-US` locale in the link

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,176 @@
...
+                <span class="label-text">{{ _('Help us test by downloading Firefox Nightly') }}</span>
+              </a>
+            </li>
+            <li class="option">
+              <a data-task-id="7" class="ab-task category category-testing" href="https://oneanddone.mozilla.org/" target="_blank">
+                <span class="label-text">{{ _('Pick from a list of first-time tasks here') }}</span>
+              </a>
+            </li>
+            <li class="option">
+              <a data-task-id="5" class="ab-task category category-helping" href="https://support.mozilla.org/en-US/get-involved/questions" target="_blank">
```
alexgibson: Same here remove hard-coded `en-US` locale in the link

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,176 @@
...
+                <span class="label-text">{{ _('Attend a Mozilla community meeting right now!') }}</span>
+              </a>
+            </li>
+          </ul>
+        </div>
+        <div class="option-group">
+          <h2>{{ _('Try something more challenging.') }}</h2>
+          <ul class="option-list option-double">
+            <li class="option">
+              <a data-task-id="5" class="ab-task category category-helping" href="https://support.mozilla.org/en-US/get-involved/questions" target="_blank">
```
alexgibson: Again, remove locale here

alexgibson:
Submitting the email form on /tasks-survey/ doesn't show any form of "Thank you" message to the user, it just reloads the page and doesn't make it clear it was a success?

alexgibson:
This is looking pretty good template wise - I agree with Craig it would be good to tweak the layout so things are more centered when there are less than 4 tasks horizontally. Other than that, I think we need to make sure that iframe is loaded over `https://` and fix the email thank you message.

I'll defer to @pmclanahan on the Python bits

comzeradd:
Thanks @alexgibson for the comments. I made some css tweaks to make options centered when we have 3 options.

alexgibson:
Layout is a little busted when there are only 2 tasks:

tasks

bedrock/mozorg/templates/mozorg/contribute/signup.html
```diff
@@ -10,6 +10,12 @@
 
 {% block page_title %}{{_('Get Involved: Volunteer Opportunities at Mozilla')}}{% endblock %}
 
+{% block optimizely %}
+  {% if waffle.switch('contribute-signup-optimizely') %}
+    <script src="//cdn.optimizely.com/js/246059135.js"></script>
```
alexgibson: If we're using Optimizely to redirect traffic for this test, can we please make sure to do it for en-US visitors only, as these new pages are not translated? Thanks

bedrock/mozorg/views.py
pmclanahan: Very good. Thanks!

pmclanahan:
Backend fixes look good to me. r+ on that. I understand other questions are still open, so I'm not merging. But others should feel free to merge if the rest is worked out.

alexgibson:
@comzeradd - I think if we can just tidy up the layout for the tasks, we should still be good to merge/push this today? If not, it may have to wait as we want to try and avoid a Friday push right before Whistler.

comzeradd:
I fixed the css regression. I added a new commit changing email recipient for the 'Other' category as @LucyeoH suggested.

alexgibson:
Thanks @comzeradd!

The alignment could do with some more work, but is definitely better than it was. I think we probably need to run with what we have today and tidy up post launch if need be. Before we merge can you please attach relevant bug numbers to each commit?

r+wc on the front-end :dango:

comzeradd:
Before merge I'm also waiting for details on gathering these emails on a new Exact Target bucket.

comzeradd:
There is no bug opened for the email change. Do you want me to open one?

media/css/mozorg/contribute/contribute-2015.less
```diff
@@ -958,6 +962,53 @@ textarea:focus {
...
+            vertical-align: top;
+            float: none;
+            width: 210px;
+        }
+    }
+
+    .option-double {
+        width: 440px;
+        display: inline-block;
+        text-align: left;
```
alexgibson: If you would like to fix the alignment when there's two items, change this to `text-align: center;`

media/css/mozorg/contribute/contribute-2015.less
```diff
@@ -958,6 +962,53 @@ textarea:focus {
...
+        float: left;
+
+        ul.option-list {
+            text-align: left;
```
alexgibson: change this to `text-align: center;`

media/css/mozorg/contribute/contribute-2015.less
```diff
@@ -958,6 +962,53 @@ textarea:focus {
...
+
+            li.option {
+                float: left;
```
alexgibson: Then remove this selector - things should look a little more balanced now?

alexgibson:
> There is no bug opened for the email change. Do you want me to open one?

It would be good to be able to associate this change with a bug for reference, if possible :+1:

alexgibson:
@comzeradd - I have to go offline, but when you have updated and are ready to merge, may I suggest you pop into #www on IRC and someone should be able to help merge/push. Thanks

craigcook:
@alexgibson Thanks for jumping in on this. I didn't mean to pass the buck, but I had originally hoped it would be done and merged on Wednesday. Then I got a late start on the review and found enough little things that needed fixing and before I knew it the week got away from me. Thanks for covering for me. I owe you (another) one :beers:

alexgibson:
@craigcook np - happy to help :beer:

craigcook:
Marking this as Do Not Merge pending the last few code tweaks and finalizing the Basket/Exact Target stuff. We'll finish it up and merge/push after #mozwww

comzeradd:
@craigcook Since the decision was to add these users to the `dontknow` bucket, I updated the code accordingly. This probably needs some review, since I don't have previous experience with basket.

craigcook:
@pmclanahan can you take a look at the basket bits? Should we put this on a demo server for some testing?

bedrock/mozorg/forms.py
```diff
@@ -223,6 +223,19 @@ class ContributeForm(forms.Form):
...
 
+class ContributeTasksForm(forms.Form):
+    required_attr = {'required': 'required'}
+
+    email = forms.EmailField(widget=EmailInput(attrs=required_attr))
+    name = forms.CharField(widget=forms.TextInput(attrs=required_attr))
+    privacy = forms.BooleanField(widget=PrivacyWidget)
+    category = forms.CharField(widget=widgets.HiddenInput(attrs={'value': 'dontknow'}))
+
+    def clean(self):
```
pmclanahan: This doesn't seem to do anything. Best to just remove it.

bedrock/mozorg/forms.py
```diff
@@ -223,6 +223,19 @@ class ContributeForm(forms.Form):
     office_fax = forms.CharField(widget=HoneyPotWidget, required=False)
 
 
+class ContributeTasksForm(forms.Form):
+    required_attr = {'required': 'required'}
+
+    email = forms.EmailField(widget=EmailInput(attrs=required_attr))
+    name = forms.CharField(widget=forms.TextInput(attrs=required_attr))
+    privacy = forms.BooleanField(widget=PrivacyWidget)
+    category = forms.CharField(widget=widgets.HiddenInput(attrs={'value': 'dontknow'}))
```
pmclanahan: Is `dontknow` the only valid value for this field? It seems so, but I don't see any validation of this. We should probably remove this field and send `dontknow` as the hard coded value in the view. Relying on this to come back unchanged from the browser won't work.

pmclanahan:
I commented on a couple of issues with the python bits. But the biggest remaining thing to do is to write some tests for the new form. If you're comfortable doing this @comzeradd that's great, otherwise ping me and I'm happy to help.

comzeradd:
@pmclanahan Thanks for the comments. I updated the form/view code, since it seems that basket requires some extra fields (eg. country).

bedrock/mozorg/views.py
```diff
@@ -162,6 +162,75 @@ def contribute_signup(request):
...
+        return reverse('mozorg.contribute.tasksthankyou')
+
+    def get_basket_data(self, form):
+        data = form.cleaned_data
+        basket_data = {
+            'email': data['email'],
+            'name': data['name'],
+            'country': data['country'],
+            'interest_id': 'dontknow',
+            'lang': 'en-US'
```
pmclanahan: Should use `l10n_utils.get_locale(self.request)` for future proofing.

comzeradd:
@pmclanahan PR updated

bedrock/mozorg/views.py
```diff
@@ -162,6 +162,75 @@ def contribute_signup(request):
...
+
+    def get_context_data(self, **kwargs):
+        cxt = super(ContributeTasks, self).get_context_data(**kwargs)
+        variation = self.request.GET.get('variation', '4')
+        match = self.variation_re.match(variation)
+        if match:
+            cxt['variation'] = variation
+        return cxt
+
+    template_name = 'mozorg/contribute/contribute-tasks.html'
```
pmclanahan: NIT: Odd to have a class variable after a method definition. Typically all class variable definitions would be at the top of the class.

pmclanahan:
Looking good. Still no tests for the new views though.

craigcook:
@comzeradd We still need to include some basic tests for the new views. Do you need any help with that? @pmclanahan can give you some guidance, or we might be able to add tests in a separate PR. We also want to do some testing on a demo server and hopefully get this merged and pushed this week, at last.

comzeradd:
Yes, I'd prefer we do this on separate PR. This will also help us make further minor fixes if we see this live on stage. I can also ping one of the backend guys from our team to help with the tests once this is merged.

pmclanahan:
We don't typically ever allow untested python into the codebase if we can help it. I'm willing to allow it if the manual tests on a demo server work out due to the tight timeline, but this isn't a good practice. Untested code is broken code by definition.

craigcook:
This PR is on demo5, e.g. https://www-demo5.allizom.org/en-US/contribute/tasks/?variation=1

bedrock/mozorg/templates/mozorg/contribute/contribute-tasks.html
```diff
@@ -0,0 +1,176 @@
...
+                <span class="label-text">{{ _('Help us test by downloading Firefox Nightly') }}</span>
+              </a>
+            </li>
+            <li class="option">
+                <a data-task-id="2" class="ab-task category category-stumbling" title="{{ _('Opens new window') }}" href="https://oneanddone.mozilla.org/tasks/92/" target="_blank">
+                  <span class="label-text">{{ _('Download our Stumbler App') }}</span>
+                </a>
+            </li>
+            <li class="option">
+              <a data-task-id="3" class="ab-task category category-activism" title="{{ _('Opens new window') }}" href="https://sendto.mozilla.org/page/contribute/givenow-seq" target="_blank">
```
TheoChevalier: Could we use donate_url() function here please? This way it will open a localized page when it's available. 
Here's an example https://github.com/mozilla/bedrock/blob/master/bedrock/tabzilla/templates/tabzilla/tabzilla.js#L696

ckprice:
Quick note: if we want to measure actions on the tasks and tasks-survey pages

e.g. Did users click more on [v1](https://www-demo5.allizom.org/en-US/contribute/tasks/?variation=1) vs [v2](https://www-demo5.allizom.org/en-US/contribute/tasks/?variation=2)?

Then the snippet will also have to be added to `tasks` and `tasks-survey`.

craigcook:
Looks like we also need the same optimizely snippet on <code>contribute-tasks.html</code> and <code>contribute-tasks-survey.html</code> to track the experiment (still wrapped in the same waffle switch).

comzeradd:
@craigcook I updated the PR with these changes

LucyeoH:
@craigcook @ckprice @pmclanahan do you mind checking this out today? @comzeradd can you tell me when this is merged? 

p.s. why is there a "Do Not Merge" label? Just because it's not R+'d yet?

craigcook:
@LucyeoH We're verifying the optimizely stuff on the demo server and will add some back-end tests so if all goes well we should be able to merge and push to production Monday.

pmclanahan:
Opened a new PR for this with tests (#3135). Closing this one.

PR #3060, opened 2015-06-13
Fix Bug 1174423 - Disable font-variant: small-caps for ga-IE, evaluate exclusion of other locales

kyoshino:
We still have to figure out locales need to be updated. So this is WIP. DO NOT MERGE.

### DIFF
--- a/media/css/l10n/ga-IE/intl.css
+++ b/media/css/l10n/ga-IE/intl.css
@@ -1,5 +1,6 @@
-/* Bug 854756 */
+/* Bug 854756, 1174423 */
 
 * {
   text-transform: none !important;
+  font-variant: normal !important;
 }
--- a/media/css/base/mozilla-share-cta.less
+++ b/media/css/base/mozilla-share-cta.less
@@ -42,7 +42,7 @@
         text-align: center;
         visibility: visible;
         text-shadow: none;
-        font-family: 'Open Sans', sans-serif;
+        .open-sans;
         opacity: 1;
         .transition(opacity .2s ease-in-out);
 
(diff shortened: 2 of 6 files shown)


EmiXLabs:
Also need for 'az' locale

flodolo:
Emin, do you need to disable both? Please leave a comment in the bug

EmiXLabs:
@flodolo looks like this pull request will not be merged as its not used in mozilla.org and problem is only for Chrome. So it can be passed.

flodolo:
mozilla.org is not used only with Firefox. It will be merged as soon as it's ready (again, see https://bugzilla.mozilla.org/show_bug.cgi?id=1174423)

EmiXLabs:
@flodolo I'm following that bug as I'm in Cc and as you mentioned in Bugzilla, need to replicate what we did for Turkish locale also in Azerbaijani, just adding same CSS changes will be OK.

kyoshino:
Sorry I'm late here :bow: Updated the PR based on the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1174423)'s discussion. @flodolo r?

media/css/l10n/ga-IE/intl.css
```diff
@@ -1,5 +1,5 @@
-/* Bug 854756 */
+/* Bug 1174423 */
 
 * {
-  text-transform: none !important;
```
flodolo: Let's keep this, we didn't receive any explicit request to remove it, and I don't know how bad it fails across browsers and versions.

flodolo:
Besides one comment, can you add 'az' with the same rules as 'tr'? Then we're good to go.

kyoshino:
Thanks @flodolo! Updated.

flodolo:
r+ from me, thanks Kohei :+1:

PR #3061, opened 2015-06-14
Add functional tests for newsletter page

schalkneethling:
@alexgibson r? Not 100% sure the way I am testing form submission 'fail' when validation fails but hey, that is what pull requests are for right ;)

### DIFF
--- a/tests/functional/newsletter.js
+++ b/tests/functional/newsletter.js
@@ -1,0 +1,90 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global casper */
+'use strict';
+
+var config = require('../lib/config');
+var helpers = require('../lib/helpers');
+var path = '/newsletter/';
+var url = config.base() + path;
+
+casper.test.begin('Newsletter, Elements: ' + url, 7, function suite(test) {
+
+    casper.start(url, function() {
+        test.assertHttpStatus(200);
+
+        test.assertVisible('#newsletter-form', 'Form is visible');
+        test.assertElementCount('#newsletter-form input[type="email"]', 1, 'One email field exists');
+        test.assertElementCount('#newsletter-form select', 2, 'Two select drop downs exist');
+        test.assertElementCount('#newsletter-form input[type="radio"]', 2, 'Two radio buttons exist');
+        test.assertElementCount('#newsletter-form input[type="checkbox"]', 1, 'One checkbox exist');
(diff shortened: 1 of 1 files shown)


alexgibson:
Agree I'm not sure the current form failure test feels quite right. Perhaps it would be nice just to affirm that an error message gets displayed when the server returns an error. To bypass HTML5 form validation that browsers automatically apply, maybe we could use the CasperJS `evaluate` method to set a `novalidate` attribute in the form before submitting. Then we might just be able to test that the error list gets shown in response. Just an idea, haven't actually tried this yet - but something we could try and figure out.

schalkneethling:
@alexgibson yeah, your suggestion sounds much better. I will take a look at this.

tests/functional/newsletter.js
```diff
@@ -0,0 +1,98 @@
...
+        test.done();
+        helpers.done();
+    });
+
+});
+
+casper.test.begin('Newsletter, Validation failed, form submission prevented: ' + url, 1, function suite(test) {
+
+    casper.start(url).thenEvaluate(function() {
+        document.querySelector('#id_email').removeAttribute('required');
```
alexgibson: Could this be simplified to `document.getElementById('newsletter-form').setAttribute('novalidate', true);`?

tests/functional/newsletter.js
```diff
@@ -0,0 +1,98 @@
...
+        test.done();
+        helpers.done();
+    });
+
+});
+
+casper.test.begin('Newsletter, Click element labels: ' + url, 2, function suite(test) {
+
+    casper.start(url, function() {
+        this.fill('#newsletter-form', {
```
alexgibson: Is this needed for the test?

tests/functional/newsletter.js
```diff
@@ -0,0 +1,98 @@
...
+
+});
+
+casper.test.begin('Newsletter, Successful form submission: ' + url, 1, function suite(test) {
+
+    casper.start(url, function() {
+        this.fill('#newsletter-form', {
+            email: 'noreply@mozilla.com',
+            privacy: true
+        });
```
alexgibson: If you pass `true` here at the end of `this.fill`, there's no need to fire the `click` below

alexgibson:
Other than a couple of nits, this looks great! r+wc thanks @schalkneethling

tests/functional/newsletter.js
schalkneethling: Ha! I had no idea about that attribute :blush: . That is indeed much simpler.

tests/functional/newsletter.js
schalkneethling: As I am not actually submitting the form, nope, it is not.

schalkneethling:
@alexgibson updated and squashed

PR #3065, opened 2015-06-16
Update homepage functional test for newsletter submission

alexgibson:


alexgibson:
@schalkneethling r? I got a couple of intermittent failures for this test yesterday, due to either slow network connectivity or sendto.mozilla.org being slow. This should ease things a bit.

tests/lib/config.js
```diff
@@ -28,6 +28,9 @@ var userAgent = {
 // set default viewport size to desktop
 casper.options.viewportSize = viewport.desktop;
 
+// set default wait timeout
+casper.options.waitTimeout = 5000;
```
schalkneethling: Is resetting it in helpers#done not enough? The default timeout in Casperjs is 5000

alexgibson: Figured it couldn't hurt, I guess the default Casper uses could always change in future versions?

schalkneethling: Fair enough :+1:

schalkneethling:
r+

PR #3067, opened 2015-06-16
Bug 1172673 - update Maker Party promo

craigcook:
Changing the color scheme


alexgibson:
Tested locally and changes look good to me, r+ to merge when ready :v:

craigcook:
Haven't gotten final word in the bug so to play it safe I think we should merge this change, as it was the design that was originally approved anyway. We can always change it again later.

PR #3068, opened 2015-06-16
Fix "grunt less" with explicit dependencies

jgmize:


jgmize:
@craigcook r?

alexgibson:
I don't think the `grunt less` command is really used anymore since the media to static conversion. We originally added it because it used to require two page refreshes to see changes made to CSS. Now with the media to static conversion, this is no longer the case. 

CSS is now also moved/compiled into the `/static` directory on page load as far as I understand, so even if we fix the compilation here, running the command still doesn't do anything much of use anymore. It's been on my to-do list to remove `grunt-less`, which I will make an effort to do today in a separate PR :smile: (thanks for the prompt) 

Note: It may still be worthwhile doing this PR, since technically any .less file should be able to compile individually?

alexgibson:
Opened #3070 to remove the `grunt less` task

schalkneethling:
The `grunt less` task has been removed. This lgtm :cake:

alexgibson:
I totally forgot we didn't merge this one, sorry!

PR #3072, opened 2015-06-17
[fix bug 1154783] Hello FTE: Capture new event for new room sharing button

alexgibson:


alexgibson:
@ckprice - r?

ckprice:
r+  :rocket:

image 2015-06-17 at 1 10 59 pm

Looks like it doesn't fire the event when the user clicks `+ Add a Service`, only if they actually click on a service, which I think is fine.

Thanks!

alexgibson:
Thanks @ckprice!

PR #3073, opened 2015-06-17
Bug 907793: Fix Open Sans to be able to render Romanian

pascalchevrel:
The patched files are provided by Jonathan Kew, see https://bugzilla.mozilla.org/show_bug.cgi?id=907793#c19


craigcook:
@pmclanahan Will we need to cachebust references in the CSS to make sure the new files are loaded?

pmclanahan:
@craigcook I think unfortunately yes :( Should just be able to add a special comment to files that reference these files:

```css
/*! updated 2015-06-17 */
```

Something like that.

pascalchevrel:
All the css files referencing these fonts are updated with the comment you suggested

pmclanahan:
Looks great Pascal. r+

PR #3074, opened 2015-06-17
Bug 1172673 - Update Webmaker promo copy

craigcook:
There was a late copy change that I missed.

@flodolo had you already exposed strings from the previous PR? If so, I'm really sorry.


flodolo:
String replaced, please use tag `webmaker_promo_makerparty2015b` for this one since we already have 12 locales with the other strings translated.

alexgibson:
r+ :penguin:

PR #3075, opened 2015-06-17
WIP: Add simple dict cache for more speed.

pmclanahan:
This I think will help. Going ahead and submitting the PR for a sanity check.


pmclanahan:
@jgmize let me know what you think of this direction. Just copied methods from `LocMemCache` that used pickle and removed pickle. Have to be careful of mutating the values, but with p-d and translations that isn't an issue.

jgmize:
@pmclanahan this seems reasonable to me. Slap some unit tests on that and let's ship it :)

pmclanahan:
Added tests from Django. Overkill, but definitely thorough :)

jgmize:
thorough. yes. I think we're covered now.

pmclanahan:
`WIP` tag removed. I think we're good. I did wonder about adding the cleanup to this PR, but I kept them as separate commits and figured that would be good enough :smile_cat:

jgmize:
good enough for me :)

PR #3082, opened 2015-06-27
[fix bug 1177810] get started with sync button does nothing

alexgibson:
Some fallout from the huge GTM migration PR :(

### DIFF
--- a/media/js/firefox/sync.js
+++ b/media/js/firefox/sync.js
@@ -126,6 +126,12 @@
         'data-download-version': 'Firefox for Android'
     });
 
+    // Firefox Sync sign in flow button
+    $('#cta-sync').on('click', function(e) {
+        e.preventDefault();
+        Mozilla.UITour.showFirefoxAccounts();
+    });
+
 
     Mozilla.FxFamilyNav.init({ primaryId: 'desktop', subId: 'sync' });
 })(window.jQuery, window.Mozilla);


media/js/firefox/sync.js
```diff
@@ -126,6 +126,12 @@
         'data-download-version': 'Firefox for Android'
     });
 
+    // Firefox Sync sign in flow button
+    $('#cta-sync').on('click', function(e) {
+        e.preventDefault();
+        Mozilla.UITour.showFirefoxAccounts();
```
schalkneethling: I guess there is no way to really test this with Casper as the tests are run in Phantom? Then again, would it actually be possible even if it was running in Slimer?

flodolo:
OCD note: fix the typo in the commit message

schalkneethling:
How do I get the get started button to show up via localhost? ;) I tried in various Fx versions, using 'forget' and clearing all things history but, alas it just does not show up. @flodolo

alexgibson:
> OCD note: fix the typo in the commit message

Heh, thanks - this is what I get for opening a PR after drinks in the hotel bar

media/js/firefox/sync.js
alexgibson: Yep - this is not possible with Casper since it requires a real Firefox with custom preferences. Need to work out something for the future - perhaps a custom build with Selenium

alexgibson:
> How do I get the get started button to show up via localhost? ;)

Please see the documentation for configuring UI tour to work locally http://bedrock.readthedocs.org/en/latest/uitour.html#local-development

schalkneethling:
aha, thanks @alexgibson

alexgibson:
Updated commit message typo, thanks @flodolo

media/js/firefox/sync.js
schalkneethling: hmmm, I wonder perhaps we can test that it goes to about:accounts?action=signup&entrypoint=uitour - will try it out.

schalkneethling:
r+

media/js/firefox/sync.js
alexgibson: It won't work in CasperJS, because it is UITour and only works in Firefox.

PR #3090, opened 2015-06-30
Remove junit as reporter from karma.conf

schalkneethling:
@alexgibson r? A small oversight I also missed ;p


alexgibson:
thanks @schalkneethling :beer:

PR #3092, opened 2015-07-01
Add functional tests for contact pages

alexgibson:


tests/functional/contact.js
```diff
@@ -0,0 +1,133 @@
...
+        test.assertVisible('#nav-spaces-select', 'Mobile navigation is visible');
+    });
+
+    casper.thenEvaluate(function(space) {
+        var select = document.getElementById('nav-spaces-select');
+        select.value = space;
+
+        var evt = document.createEvent('Event');
+        evt.initEvent('change', true, true);
+        select.dispatchEvent(evt);
```
schalkneethling: We might need to do this kinda thing i.e. dispatch events, in more places as we move forward so, perhaps it is worth wrapping the above three lines in a triggerEvent function inside helpers.js ?

alexgibson: Ooh, good call :+1:

tests/functional/contact.js
```diff
@@ -0,0 +1,133 @@
...
+});
+
+casper.test.begin('Contact, Mobile interaction: ' + spacesUrl, 4, function suite(test) {
+
+    casper.options.viewportSize = config.viewport.mobile;
+    casper.options.waitTimeout = 10000;
+
+    casper.start(url, function() {
+        test.assertNotVisible('#map', 'Map is not visible on mobile');
+        test.assertVisible('#nav-spaces-select', 'Mobile navigation is visible');
```
schalkneethling: Seeing that you check that the mobile navigation is not visible on desktop:
https://github.com/mozilla/bedrock/pull/3092/files#diff-291e556dd45b5f79745989905a959658R24

Perhaps we should here ensure that the desktop navigation is not visible on mobile, for both spaces and communities? Or perhaps just ensuring that it is not visible for the one, will be enough.

alexgibson: Makes sense, will do

tests/functional/contact.js
alexgibson: Hmm, might be trickier than I first though - this code needs to execute inside the remote DOM environment, and calling a function that's in `helpers.js` does not work here. I'll see what I can figure out.

alexgibson:
@schalkneethling - figured it out by creating a `clientScripts` helper file than can be used for remote DOM things. Should be good I think

alexgibson:
Travis is failing to update product details for some reason - very likely unrelated to this change.

schalkneethling:
Nice one! :cake:

PR #3094, opened 2015-07-02
Fix bug 1179847: Support pontoon.m.o in dev mode.

pmclanahan:
Since "allow-from" [isn't well supported across browsers](https://developer.mozilla.org/en-US/docs/Web/HTTP/X-Frame-Options#Browser_compatibility) I decided to just turn it off on dev (including demos). If anyone knows why this might be a bigger issue, let's discuss it here.


pmclanahan:
@jgmize r?

PR #3053, opened 2015-06-10
Bug 1164278. Fx Family nav FxOS distinction.

jpetto:

PR #3095, opened 2015-07-03
[fix bug 1179790] Header on firefox/desktop page misaligned in IE

alexgibson:
Need to use `top` instead of `margin-top` for this page since the intro container is set to `overflow: hidden;` (which makes IE have a sad when it comes to applying negative margin values).

### DIFF
--- a/media/css/firefox/desktop/index.less
+++ b/media/css/firefox/desktop/index.less
@@ -6,7 +6,8 @@
 @import 'common.less';
 
 #wrapper {
-    margin-top: -162px; // family nav
+    margin-top: 0;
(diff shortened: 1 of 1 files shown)


schalkneethling:
For some reason the hover interactions of the family nav does not work for me locally. Not just on the desktop page, even if I click on for Andriod, the hover states are not triggered. Does it work for you @alexgibson ?

alexgibson:
@schalkneethling - those got removed in #3053

schalkneethling:
Gotcha! r+ :cocktail:

PR #3097, opened 2015-07-03
Add functional tests for history page

alexgibson:


schalkneethling:
@alexgibson I am wondering, do we want to add some tests here to test the animation interactions? For example clicking the next button changes the current animate-in list item to animate-out, and the next list item to animate-in, indicating the transition happens.

That then also brings up the question of whether we should then test that, when the animation reaches the last slide, does it cycle back to the first item in the list and vice versa, clicking on previous when on the first slide, moved back to the last slide in the list.

Thoughts?

alexgibson:
@schalkneethling  - I gave this some thought too. Because this page uses a 3rd party JQuery plugin, I kinda came to the conclusion that adding functional tests at this level are perhaps a bit overly specific? Personally I think the best places for tests like "The slideshow plugin should add the expected CSS class when the button is clicked" should probably be in the plugin repo itself, which the author is responsible for maintaining. I can't see us altering the source code of the plugin here, so I figured probably just ensuring the plugin has loaded ok on the page is enough. What do you think? Open to suggestions if there's a middle-ground you would feel more comfortable with.

Also, because the plugin is set to auto-rotate it makes testing interactions a bit more prone to unpredictability.

schalkneethling:
@alexgibson aha, well then I totally agree with you. r+ :cherries:

PR #3099, opened 2015-07-06
[RTL] Mission Page doesn't support the RTL direction

mermi:
This is ready, please review, I filled a bug here #1180553


schalkneethling:
Page after the change:

screen shot 2015-07-06 at 15 14 16

r+ :cat:

PR #3100, opened 2015-07-06
[RTL] About page doesn't support the RTL direction exactly. 

mermi:
This is ready, please review, I filled a Bug here #1180558

### DIFF
--- a/media/css/mozorg/about-base.less
+++ b/media/css/mozorg/about-base.less
@@ -131,6 +131,10 @@ html[dir="rtl"] {
         margin-right: 0;
         margin-left: 490px;
     }
+    #main-content .intro {
+        float: right;
+
+    }
 }
 
 // Leadership


media/css/mozorg/about-base.less
```diff
@@ -131,6 +131,10 @@ html[dir="rtl"] {
         margin-right: 0;
         margin-left: 490px;
     }
+    #main-content .intro {
+        float: right;
+
```
flodolo: Style consistency: no empty line in the middle of the rule, empty line before the rule.

schalkneethling: I believe the empty line should be `clear: right;` @mermi ?

schalkneethling: although it does not seem to be needed

schalkneethling:
Other than removing the one empty line, this is an r+ Thanks @mermi \o/

schalkneethling:
Also, just FYI, naming your branches something like bug-bugnumber-bug-title and then making your commit message something like 'Fix Bug bugnumber, some descriptive message' enables the bot to comment in the bug and auto resolved fix it for you.

So this one could be:

bug-1180558-about-page-rtl, and the commit message 'Fix Bug 1180558, aligning video correctly fixes RTL'

mermi:
ok, I will rename the branch @schalkneethling and @flodolo I am checking the empty line

media/css/mozorg/about-base.less
mermi: Yes it's for clear: right @schalkneethling

PR #3101, opened 2015-07-06
Up the Casperjs waitTimeout to avoid test timeouts

schalkneethling:
@alexgibson r?

### DIFF
--- a/tests/lib/helpers.js
+++ b/tests/lib/helpers.js
@@ -10,7 +10,7 @@ var config = require('../lib/config');
 function done() {
     casper.options.pageSettings.userAgent = config.userAgent.phantomjs;
     casper.options.viewportSize = config.viewport.desktop;
-    casper.options.waitTimeout = 5000;
+    casper.options.waitTimeout = 10000;
 }
 
 module.exports = {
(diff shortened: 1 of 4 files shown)


alexgibson:
Other than the typo in the commit message, this is an r+ from me :cake:

tests/lib/helpers.js
```diff
@@ -10,7 +10,6 @@ var config = require('../lib/config');
 function done() {
     casper.options.pageSettings.userAgent = config.userAgent.phantomjs;
     casper.options.viewportSize = config.viewport.desktop;
-    casper.options.waitTimeout = 5000;
```
alexgibson: Wait, this should probably remain and be changed to `casper.options.waitTimeout = 5000;`, for the case where we may set a test specific value?

schalkneethling: Indeed, good catch. For some reason my mind was thinking, meh, we'll never need a longer timeout ;p

alexgibson: I saw thinking somewhere we might want to set a shorter timeout :P

schalkneethling:
@alexgibson updated

schalkneethling:
hmmmm, I updated the commit message, not sure why it does not reflect over here. Perhaps when you click the big green button it will show.

tests/lib/helpers.js
schalkneethling: Well of course, one might want to do that as well :)

PR #3105, opened 2015-07-07
Update TwitterCache on save: fix bug 1168640

jgmize:


jgmize:
@pmclanahan r?

bedrock/mozorg/tests/test_models.py
```diff
@@ -55,3 +56,9 @@ def test_errors_fail_silently(self, get_mock):
 
         # and even errors should be cached
         get_mock.assert_called_once_with(account='walter')
+
+    @patch('bedrock.mozorg.models.cache_tweets')
+    def test_cache_post_save(self, mock_cache_tweets, get_mock):
+        post_save.send(TwitterCache,
+                       instance=TwitterCache(account='me', tweets=['I twote']))
+        mock_cache_tweets.assert_called_once_with('me', ['I twote'])
```
pmclanahan: lol @ twote! :joy:

pmclanahan:
So the point is to only cache an empty DB for 10s so that someone can manually fix the missing tweets and have it show up quickly?

jgmize:
> So the point is to only cache an empty DB for 10s so that someone can manually fix the missing tweets and have it show up quickly?

There is also the possibility that the data is there, but there happens to be a temporary network or DB issue.

pmclanahan:
Cool. Only comment I have is that setting the cache in the save signal handler likely won't do much since it happens in one DC from the admin node, but it's not hurting anything and will hopefully be of more use soon :)

jgmize:
> Cool. Only comment I have is that setting the cache in the save signal handler likely won't do much since it happens in one DC from the admin node, but it's not hurting anything and will hopefully be of more use soon :)

The main issue this affects is dev since it wipes the DB on every deploy, which is in the same DC as the admin node.

PR #3107, opened 2015-07-08
12-factorify settings: fix bug 1182394

jgmize:


pmclanahan:
My primary comment so far is that this is looking good. But, I don't think we need to make every single configuration variable settable via the environment. I would stick to those things that do change from env to env right now, and we can convert others as we go. You've done the work now, but I see some things in there that are very unlikely to change outside of a code update.

jgmize:
@pmclanahan after looking over the changes I found several cases where I agreed with your assessment that they would require code changes anyway, and I've reverted those changes. Please let me know if there are any remaining that you don't think should be environment specific. I've also updated the PR to no longer throw an error if there is no settings/local.py file, but as we've previously discussed it will still use it if present. I've removed the WIP label, as I think this is ready for your "real" review and local testing :)

bedrock/settings/__init__.py
```diff
@@ -18,9 +18,7 @@
     try:
         from .local import *  # noqa
     except ImportError as exc:
-        exc.args = tuple(['%s (did you rename bedrock/settings/local.py-dist?)' %
-                          exc.args[0]])
-        raise exc
+        'local.py is supported, but no longer necessary'
```
pmclanahan: Should we print a warning here?

bedrock/settings/base.py
```diff
@@ -24,29 +26,31 @@ def path(*args):
...
 # Production uses MySQL, but Sqlite should be sufficient for local development.
 # Our CI server tests against MySQL.
-DATABASES = {
-    'default': {
-        'ENGINE': 'django.db.backends.sqlite3',
-        'NAME': 'bedrock.db',
-    }
-}
-SLAVE_DATABASES = []
+DATABASES = config(
```
pmclanahan: This is a bit hard to parse as a human, but it seems like it'll work well as a human managing the environment.

pmclanahan: The other strategy I was envisioning was an optional env var as a CSV of other env var names. e.g.:

```python
DATABASES = {}
DATABASE_URL_VARS = config('DATABASE_URL_VARS', default=None, cast=Csv())
if DATABASE_URL_VARS:
    # example usage
    # DATABASE_URL_VARS='DATABASE_URL_DEFAULT,DATABASE_URL_RW_MASTER'
    # DATABASE_URL_DEFAULT='sqlite:///bedrock.db'
    # DATABASE_URL_RW_MASTER='sqlite:///rw_master.db'
    for db_var in DATABASE_URL_VARS:
        db_name = db_var[13:].lower()
        DATABASES[db_name] = dj_database_url.parse(config(db_var))
else:
    DATABASES['default'] = config('DATABASE_URL', default='sqlite:///bedrock.db')
```

I don't think the above is as elegant a solution as yours, but it sure would make the environment vars and `.env` file easier to read.

bedrock/settings/base.py
```diff
@@ -24,29 +26,31 @@ def path(*args):
...
 DATABASE_ROUTERS = ('multidb.PinningMasterSlaveRouter',)
 
-# Override in local.py for memcached.
-CACHES = {
-    'default': {
-        'BACKEND': 'django.core.cache.backends.locmem.LocMemCache',
-        'LOCATION': 'translations'
-    }
-}
+CACHES = config(
```
pmclanahan: Same comment here. I kinda balk at JSON in env vars, but perhaps that's the way. Just kinds smells to me of a need for something else...

bedrock/settings/base.py
```diff
@@ -233,10 +235,10 @@ def JINJA_CONFIG():
     }
 
 
-MEDIA_URL = '/user-media/'
-MEDIA_ROOT = path('media')
-STATIC_URL = '/media/'
-STATIC_ROOT = path('static')
+MEDIA_URL = config('MEDIA_URL', default='/user-media/')
```
pmclanahan: I'd say the media and static `ROOT` configs are among those unlikely to change, but this can't hurt I think.

bedrock/settings/base.py
```diff
@@ -260,13 +262,16 @@ def JINJA_CONFIG():
 PIPELINE_COMPILERS = (
     'pipeline.compilers.less.LessCompiler',
 )
-PIPELINE_LESS_BINARY = path('node_modules', 'less', 'bin', 'lessc')
```
pmclanahan: these bin paths are unlikely to change as well right? assuming we stick with installing via NPM.

Note: as soon as we're in a better way with deployment I want to use npm directly and delete the `node_modules` dir from the repo. We'd use [npm-lockdown](https://github.com/mozilla/npm-lockdown) of course.

pmclanahan:
This is looking good. Let's talk briefly about the more complex configs e.g. `DATABASES` and then go with it. I do want a command or tool or something that will take a persons `local.py` and give them a `.env` file, but that can come later.

bedrock/settings/__init__.py
jgmize: Ideally not having a local.py would become the norm rather than the exception, so no, I wouldn't see the point of printing a warning here. Once we're farther along and have refactored the local settings into environment variables on all the servers though, it probably makes sense to add a deprecation warning if you *have* a local.py though.

bedrock/settings/base.py
jgmize: In the normal case, the environment variable and .env file will only have a single DATABASE_URL, which should be fairly easy to read. I see the support for multiple DBs as a necessary interim hack that we can eventually get rid of entirely once the main servers are only talking to a read-only replica and we have separate instances entirely that can actually talk to the DB masters. However, even as an interim hack an .env file with both a RO and RW DB connection should be fairly readable:
```json
DATABASES='{"default": "mysql://rw_user:password@rw-prod-host/rw-db-name", "readonly": "mysql://ro_user:password@ro-prod-host/ro-db-name"}'
```

bedrock/settings/base.py
jgmize: As I mentioned over Vidyo the other day, I would have preferred to use https://github.com/ghickman/django-cache-url but as far as I could tell it doesn't support features we're using in our prod config right now, and I didn't want to block this on adding support for those features.

bedrock/settings/base.py
jgmize: Agreed it's unlikely, this is more future proofing than strictly necessary, but I came to the same conclusion that it couldn't hurt :)

bedrock/settings/base.py
jgmize: I actually ran into a situation where the default wasn't working and I had to set the path locally, which is why I left this in.

requirements/base.txt
```diff
@@ -180,3 +180,12 @@ django-nose==1.4
 
 # sha256: aUBxjfw-_0JYIDrVAhCQkz5cBHB9XKjMnnPJSniU6p8
 pathlib==1.0.1
+
+# TarGZ
```
glogiotatidis: nit: remove this line

requirements/base.txt
```diff
@@ -180,3 +180,12 @@ django-nose==1.4
...
+# sha256: 8uJz7TSsu1YJYtXPEpF5NtjfAil98JvTCJuFRtRYQTg
+# Wheel
```
glogiotatidis: nit: remove this line

jgmize:
Removed unnecessary comments from requirements/base.txt

bedrock/settings/base.py
pmclanahan: Fair points. Let's go with it.

bedrock/settings/__init__.py
pmclanahan: Yes. Good call. That's what I was thinking but commented in the wrong place. Could print a deprecation warning in an `else` block here.

pmclanahan:
Looking good to me. We should get this on a demo server. For now testing on a demo will likely mean manually moving the local.py out of the way and creating a .env file in the /data/src/... dir on the admin node. I know we don't want to use .env files in real environments, but my guess is that we'd have to involve IT to get any real env vars injected into the apache env for real.

pmclanahan:
@jgmize speaking of `.env` files though, please add `.env` to the `.gitignore` file.

jgmize:
> Looking good to me. We should get this on a demo server. For now testing on a demo will likely mean manually moving the local.py out of the way and creating a .env file in the /data/src/... dir on the admin node. 

Yes, I'll push this to a demo server and test with an .env file today.

>I know we don't want to use .env files in real environments, but my guess is that we'd have to involve IT to get any real env vars injected into the apache env for real.

In the long run we'll use the environment variable management provided by our platform of choice of course, but for our current deployments in PHX1 and SCL3, .env files seem like the simplest solution to me. I'd rather not have to go through IT + puppet + apache configs for environment variable changes. I'd like to implement version control on these, perhaps with symlinks to a private git repo on the admin node?

> @jgmize speaking of .env files though, please add .env to the .gitignore file.

Good catch, done.

pmclanahan:
> I'd like to implement version control on these, perhaps with symlinks to a private git repo on the admin node?

Sounds good. By "private git repo" do you mean local to only the admin node? That seems good to me. No need to involve github in this.

jgmize:
> Sounds good. By "private git repo" do you mean local to only the admin node? That seems good to me. No need to involve github in this.

Yes, exactly.

jgmize:
@pmclanahan I found some additional settings that were being overridden by the demo1 local.py, which I've added as env vars in base.py. This required me to override_settings in a couple of tests. I also found that we are using some custom OPTIONS in the DATABASES settings in the local.py so I needed to make a change there-- the python side is a little more readable, but in order for decouple to deserialize the json correctly it has to be all on one line :(

I verified that everything works as expected both with and without the local.py on demo1, after manually converting the settings in local.py to a .env file. The original local.py is now named local.py.bak in the src dir, and if we need to push a branch that isn't based on this one we will need to restore it.

pmclanahan:
Nice. I'll take a look.

jgmize:
rebased to fix merge conflict with https://github.com/mozilla/bedrock/commit/17208e0327bef65a787b22c1e9b44385c338ceda#diff-7194d3de4cc9718f18e49e996f1e50b0R266

pmclanahan:
I think we're ready to go with this. You agree?

jgmize:
> I think we're ready to go with this. You agree?

ready when you are, @pmclanahan :)

pmclanahan:
:+1:

PR #3108, opened 2015-07-08
Fix Bug 1181422 - Plugincheck says ESR 38.1.0 is out of date

kyoshino:
I’ll implement a new solution using the Tour API as part of Bug 939470, but this is still required.

### DIFF
--- a/media/js/base/global.js
+++ b/media/js/base/global.js
@@ -138,6 +138,17 @@ function isFirefox31ESR(userAgent, buildID) {
         buildID && buildID !== '20140716183446';
 }
 
+// Detect Firefox 38 ESR by the *non-ESR* build IDs.
+// 38.0: 20150508094354, 38.0.1: 20150513174244, 38.0.5: 20150525141253
+// https://wiki.mozilla.org/Releases/Firefox_38/Test_Plan
+function isFirefox38ESR(userAgent, buildID) {
+    userAgent = userAgent || navigator.userAgent;
+    buildID = buildID || navigator.buildID;
+
+    return !isFirefoxMobile(userAgent) &&
+        getFirefoxMasterVersion(userAgent) === 38 && buildID &&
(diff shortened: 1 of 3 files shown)


media/js/base/global.js
```diff
@@ -138,6 +138,17 @@ function isFirefox31ESR(userAgent, buildID) {
...
 
+// Detect Firefox 38 ESR by the *non-ESR* build IDs.
+// 38.0: 20150508094354, 38.0.1: 20150513174244, 38.0.5: 20150525141253
+// https://wiki.mozilla.org/Releases/Firefox_38/Test_Plan
+function isFirefox38ESR(userAgent, buildID) {
+    userAgent = userAgent || navigator.userAgent;
+    buildID = buildID || navigator.buildID;
+
+    return !isFirefoxMobile(userAgent) &&
+        getFirefoxMasterVersion(userAgent) === 31 && buildID &&
```
alexgibson: Shouldn't this be `38`?

kyoshino: Ahh, true! Thanks @alexgibson

alexgibson:
It feels like adding another function each time we have an ESR release is going to become a little cumbersome. Is it worth refactoring to use a single common function perhaps, which we can then update as new ESR releases come out, and older ones become non-supported? Open to suggestions on what you think might work best here @kyoshino

kyoshino:
I'll refactor the detection code as part of Bug 939470. 31 ESR goes end-of-life in a month so we can remove `isFirefox31ESR` anyway.

alexgibson:
Ok sounds good to me, thanks @kyoshino :+1:

alexgibson:
Note to self: I need to manually download and test this on ESR 38 before merging.

alexgibson:
Tested this manually on ESR 38 and works as expected, thanks @kyoshino! r+

PR #3109, opened 2015-07-08
Add karma coverage

schalkneethling:
@alexgibson Opening this for feedback and discussion. Even though the includeAllSources flag on karma-coverage seems to be a bug, it will not help us much as some JS files will not be instrumented without a test as it needs some stubbing, specifically Mozilla.UITour

With that said, this is a WIP and any and all feedback is welcomed. :closed_umbrella:


schalkneethling:
Another question is, do we want to include the coverage report? I guess we do but, perhaps not the HTML report?

schalkneethling:
@alexgibson Oh, run `grunt test` and it will generate the coverage report in a folder called `coverage`. Just open the `index.html` to see the result

alexgibson:
Thanks @schalkneethling - Some thoughts:

In a perfect world it would be great have unit/behavioral tests for everything here, but practically speaking I wonder if it may be more effective to aim for a mix of functional & units tests where most suited? While I can see us writing functional tests at a bare minimum for pretty much everything, I don't know if refactoring all this JS to make it more easily testable will necessarily prove achievable, given our often limited time & resources. It might be better to aim for more writing more easily testable code going forward, as many pages get redesigned and rebuilt?

I would most certainly include everything in `/media/js/base/` - these are common libraries we maintain and use on multiple pages, so good coverage here is the biggest win. Also things like plugincheck and UITour are areas that certainly warrant having tests written in Jasmine, since it requires a fair amount of mocking and does not lend as easily to functional tests.

alexgibson:
I would probably recommend we don't include the coverage report in the repo, and just add it to `.gitignore`. Is it possible to make it live under `tests/unit/` too perhaps?

schalkneethling:
"I would probably recommend we don't include the coverage report in the repo, and just add it to .gitignore. Is it possible to make it live under tests/unit/ too perhaps?"

Agreed however, do we perhaps want consider pushing the results to coveralls.io?

Also, I can definitely change the location of the output, and will add an entry for it in .gitignore

alexgibson:
> Agreed however, do we perhaps want consider pushing the results to coveralls.io?

Does it need to be in the repo for that, or is it just something travis/jenkins can generate?

schalkneethling:
@alexgibson To get it to coveralls.io I am sure we can have it part of what travis does, but will confirm as part of the work on that.

schalkneethling:
Yeah, I agree. I looked at some of the code and it is going to be hard to unit test and, as you mentioned, a lot of it is covered by the functional tests.

There is no way we will have 100% of all the code and, I do not believe it is needed. What is important is to have 100% coverage of the common libraries, as you mentioned, and other functional code, utils and the UI-Tours.

So, for this I am going to focus on adding everything in /base and then we can discuss which other files we need to add in https://github.com/mozilla/bedrock/issues/3110 ~ There I will add *everything* and then we can clean it up to only the one's that are required.

Sound like a plan?

alexgibson:
Sounds like a great plan to me :+1:

schalkneethling:
So, this JS coverage is being bundled up with the Python coverage stats and the JS is now dragging it down to an unacceptable low I guess :( Ideas?

alexgibson:
r+ :cake:

PR #3113, opened 2015-07-09
Bug 1171859. Build & use smooth scrolling polyfill.

jpetto:


jpetto:
@alexgibson Thoughts?

media/js/base/mozilla-smoothscroll.js
```diff
@@ -0,0 +1,58 @@
...
+                window.scrollTo(opts);
+            };
+        // use jQuery if it's available
+        } else if (window.jQuery) {
+            var $htmlBody = $('html, body');
+
+            _smoothScroll = function(opts) {
+                $htmlBody.animate({
+                    scrollTop: opts.top,
+                    scrollLeft: opts.left
```
alexgibson: When passing a single value for `opts.top` like in the example here, would it better to fall back to a `0` value for `opts.left` here instead of passing `undefined`, and vice versa?

alexgibson: Perhaps it would be good to start with a set of default properties, which could then be overridden when calling the function?

    var _options = {
        'behavior': 'smooth',
        'left': 0,
        'top': 0
    };

alexgibson:
I think this looks great - only thing other than the one suggestion above, would be to add some jasmine JS tests for this?

tests/unit/spec/mozilla-smoothscroll.js
```diff
@@ -0,0 +1,76 @@
...
+            ].join();
+
+            $(scrollMarkup).appendTo('body');
+        });
+
+        afterEach(function () {
+            $('#scrollMarkup').remove();
+        });
+
+        // TODO: This test works as long as Fx is specified as the testing browser,
```
alexgibson: Instead of relying on a specific browser to run the tests, how about passing a stub variable used only for testing? I've used a similar approach elsewhere for read-only properties like `document.hidden`. All the tests really need to make sure if when the native API is not supported, jQuery gets called instead with the expected values etc, if that makes sense?

jpetto:
@alexgibson - Updated tests. Take another look?

tests/unit/spec/mozilla-smoothscroll.js
```diff
@@ -0,0 +1,103 @@
...
+
+            $(scrollMarkup).appendTo('body');
+        });
+
+        afterEach(function () {
+            $('#scrollMarkup').remove();
+        });
+
+        it('should use window.scrollTo when native smooth scrolling is supported', function () {
+            $('#nav a').on('click', function(e) {
```
alexgibson: It's a nit, but is the on `click` stuff necessary? Tests could be a bit shorter if calling `Mozilla.smoothScroll()` directly?

media/js/base/mozilla-smoothscroll.js
```diff
@@ -0,0 +1,73 @@
...
+Mozilla.smoothScroll = (function() {
+    'use strict';
+
+    var _smoothScroll;
+    var $htmlBody;
+
+    var _init = function(unitTest) {
+        var hasSmoothScroll;
+
+        if (unitTest) {
```
alexgibson: Could this be simplified to something like `hasSmoothScroll = useNative || 'scrollBehavior' in document.documentElement.style;`?

tests/unit/spec/mozilla-smoothscroll.js
```diff
@@ -0,0 +1,103 @@
...
+                    top: $(target).offset().top,
+                    unitTest: 'jQuery'
+                });
+            });
+
+            spyOn(window, 'scrollTo');
+
+            $('#link1').trigger('click');
+
+            expect(window.scrollTo).not.toHaveBeenCalled();
```
alexgibson: Use `.toHaveBeenCalledWith()` here too?

tests/unit/spec/mozilla-smoothscroll.js
```diff
@@ -0,0 +1,103 @@
...
+
+        it('should use jQuery fallback when smooth scrolling is not supported', function () {
+            $('#nav a').on('click', function(e) {
+                e.preventDefault();
+
+                var target = $(this).attr('href');
+
+                Mozilla.smoothScroll({
+                    top: $(target).offset().top,
+                    unitTest: 'jQuery'
```
alexgibson: Is passing this variable needed for tests that aren't related to the native API call?

alexgibson: Oh, I see it's used when the plugins gets initialized? Fine to leave if so

tests/unit/spec/mozilla-smoothscroll.js
alexgibson: oh, maybe spy on `$.fn.animate` here?

tests/unit/spec/mozilla-smoothscroll.js
jpetto: Yeah, good call. I was trying too hard to mimic user behavior.

tests/unit/spec/mozilla-smoothscroll.js
jpetto: Ah, yeah, will spy on `$.fn.animate`.

tests/unit/spec/mozilla-smoothscroll.js
jpetto: Yeah, it's necessary to force the plugin to re-initialize for each test.

media/js/base/mozilla-smoothscroll.js
jpetto: I don't think so? If `useNative` were falsy, behavior would fall back to baked in browser support, which wouldn't work if I want to make sure `hasSmoothScroll` is false for some tests. Right?

alexgibson: Ah I see - I was assuming it would just be `undefined` except for in the instance of the one test that needs it. But then I see you're using the `unitTest` value elsewhere.

jpetto: Yeah, `unitTest` is dual purpose right now - forces re-initialization, and determines use of native scroll. Alternative could be passing in a different variable for testing to force re-init if you think that's better?

alexgibson: Happy to go with what you feel is best :+1:

tests/unit/spec/mozilla-smoothscroll.js
```diff
@@ -0,0 +1,84 @@
...
+            spyOn(window, 'scrollTo');
+
+            Mozilla.smoothScroll({
+                top: $('#anchor1').offset().top,
+                unitTest: 'native'
+            });
+
+            expect(window.scrollTo).toHaveBeenCalledWith({
+                behavior: 'smooth',
+                top: $('#anchor1').offset().top,
```
alexgibson: Any reason not to pass values directly here instead of need to add temp markup to the DOM?

jpetto: Ah, no, not anymore. I originally was going to test that the window scrolled to the correct position, but it doesn't appear the window scrolls at all when using Karma (even though the calls to scroll are made). Will simplify.

alexgibson:
Looks good to me, r+ :seedling: 

Care to squash?

jpetto:
Squarshed! :eggplant:

PR #3116, opened 2015-07-10
Fix bug 1167727: Ingest events from dev events calendar.

pmclanahan:


pmclanahan:
@jgmize r?

bedrock/events/tests/test_data/reps.ical
```diff
@@ -10,7 +10,7 @@ DTSTART;TZID=Europe/Skopje:20141007T180000
 DTEND;TZID=Europe/Skopje:20141007T190000
 LOCATION:KIKA\, Skopje\, Macedonia\, F.Y.R. of
 SUMMARY:Community meeting in Skopje
-DESCRIPTION:It&#39;s been a long summer and finally we have enough people back
+DESCRIPTION:It&#39;s been a long summer and finally we have enough people back
```
kewisch: I know this is not your change, but I just noticed it randomly: iCalendar doesn't use html escapes. This should be `'` and if html escaping is needed then it should probably be done elsewhere

bedrock/events/models.py
```diff
@@ -14,40 +14,58 @@
 from bedrock.events.countries import country_to_continent
 
 
+def calendar_id_from_google_url(feed_url):
+    return feed_url.split('/')[5]
+
+
+def calendar_url_for_event(event, calendar_id):
+    template = 'https://www.google.com/calendar/embed?showCalendars=0&mode=week&wkst=1&' \
+               'src={calendar_id}&ctz=America/Los_Angeles&' \
```
jgmize: Do the ampersands need to be escaped here, or are they escaped later?

requrements.txt
```diff
@@ -0,0 +1,212 @@
+-f /Users/pmclanahan/.pip/wheels
```
jgmize: Pretty sure this was accidentally included in your commit ;)

jgmize:
@pmclanahan re: https://bugzilla.mozilla.org/show_bug.cgi?id=1167727#c21 I didn't see any issues other than possibly needing to escape some ampersands and removing the accidentally committed file. If you can address those minor issues and push it to one of the demo envs when you get a chance that sounds great to me-- you might even temporarily enable the cron job in the demo env so we can test that as well.

bedrock/events/tests/test_data/reps.ical
pmclanahan: This is just test data. It's what I got from reps.mozilla.org when I originally wrote the tests. I think it's a good test if it's still representative of what reps.m.o is sending us, but I've not checked in a while.

bedrock/events/tests/test_data/reps.ical
```diff
@@ -22,17 +22,17 @@ BEGIN:VEVENT
...
 DTSTAMP:20141007T184543
-DTSTART;TZID=Africa/Johannesburg:20141008T060000
-DTEND;TZID=Africa/Johannesburg:20141009T130000
-LOCATION:Mary Fitzgerald Square\, Newtown\, Johannesburg\, Johannesburg\, 
+DTSTART;VALUE=DATE:20150719
+DTEND;VALUE=DATE:20150720
+LOCATION:Mary Fitzgerald Square\, Newtown\, Johannesburg\, Johannesburg\,
  South Africa
 SUMMARY:Tech 4 Africa
-DESCRIPTION:Tech4Africa is a place for new ideas and to encourage people t
```
pmclanahan: Oops. Didn't realize that my helpful editor had stripped a bunch of whitespace and made this diff so huge. Anyone care? I can revert if need be.

requrements.txt
pmclanahan: woah. strange. yeah. thanks.

flodolo: OCD attack, please fix the filename

bedrock/events/models.py
pmclanahan: Why would we need to escape ampersands? IIRC they only need to be escaped if you want to actually display an ampersand in the page.

jgmize: > Why would we need to escape ampersands? IIRC they only need to be escaped if you want to actually display an ampersand in the page.

You can usually get away without escaping ampersands in URLs, but it's safer to default to escaping them. See the top answer to http://stackoverflow.com/questions/19441750/do-ampersands-still-need-to-be-encoded-in-urls-in-html5 for a more detailed explanation.

pmclanahan: Very interesting indeed. Thanks.

pmclanahan: Autoescape in the template layer is escaping these. I think that's fine. Other option is to pre-escape them here and use `|safe` in the template.

pmclanahan:
This is on demo1: https://www-demo1.allizom.org/en-US/contribute/events/

requrements.txt
pmclanahan: @flodolo whole file was in the commit mistakenly. OCD fixed by deletion :)

bedrock/events/models.py
jgmize: > Autoescape in the template layer is escaping these. I think that's fine. Other option is to pre-escape them here and use |safe in the template. 

Autoescape in the template is great, thanks :)

bedrock/events/models.py
```diff
@@ -14,40 +14,58 @@
 from bedrock.events.countries import country_to_continent
 
 
+def calendar_id_from_google_url(feed_url):
+    return feed_url.split('/')[5]
+
+
+def calendar_url_for_event(event, calendar_id):
+    template = 'https://www.google.com/calendar/embed?showCalendars=0&mode=week&wkst=1&' \
```
jgmize: `mode=week` works, but given the sparseness of the data I think `mode=agenda` is more useful. Thoughts?

pmclanahan: That'd be fine with me. I just thought that the agenda view looked weird as a full page. We could popup a window, but that's pretty lame. I'm fine with whatever people want though really.

jgmize:
r+, I think this is good to go as-is code-wise. While I personally would prefer the tests not actually touch the DB, that's not worth blocking on, and neither is the mode of the calendar it links to, as it is trivially changed later.

pmclanahan:
@jgmize we have approval in the bug to go ahead with this. I think I should remove the crons from the dev config before merging. Thoughts?

jgmize:
> @jgmize we have approval in the bug to go ahead with this. I think I should remove the crons from the dev config before merging. Thoughts?

As discussed over vidyo, we're going to leave the dev cron job in to let it run on the dev server at least overnight for testing purposes, and follow up with a separate PR to disable the cron job on dev once testing is complete.

PR #3117, opened 2015-07-11
Bug 1183678 - Add SUMO links for Reading List and Reader View to /firefox/android

pollti:


flodolo:
Is there a bug reference for this change? On the technical side, it's adding new strings on a localized page, which means you need to use a ```l10n_has_tag``` conditional (try a search in the codebase for reference). But I'd like to know more on the reasons for adding new strings/links.

pollti:
There is no bmo reference for this. Someone asked in IRC if it would be possible to add those links here. If this isn't good for l10n or any other reason, we can of course close this PR and not add the links.

jpetto:
@pollti - I have [filed a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1183678) for this change. For historical reference, can you add a comment specifying who originally requested adding the links?

The code change looks good, though as @flodolo mentioned, you'll need to wrap the change in an `l10n_has_tag` block to ensure non-translated strings aren't displayed. You can view an example at the bottom of [the file you're modifying](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/android/index.html#L323). I'd suggest `android_reader_links` as the L10n tag name.

Lastly, when you update your end, please reference the bug in your commit message. Something like:

> Bug 1183678. Add SUMO links for reading list and reader view to /firefox/android/.

Let me know if you have any questions or need any assistance.

Thanks!

pollti:
Thanks for your comments and the bug report!
I was in IRC with my mobile device when I was asked to make this change, so I don't have logs and don't remember the original requester - I'm sorry for that.
I think we don't need an `else` part here, because the links aren't that important. Do you agree with that?

jpetto:
Yep, agreed we don't need an `else` condition. Just the `if` to make sure the strings are localized.

Do you have a bugzilla account? We should get the bug assigned to you.

pollti:
My bmo account is https://bugzilla.mozilla.org/user_profile?login=pollti%40gmx.net. (I assigned it to me when adding this pull request as an attachment.)

jpetto:
Looks great. r+

Thanks @pollti!

pollti:
Thanks for your assistance! @jpetto and @flodolo

PR #3124, opened 2015-07-14
Bug 1175544. Upgrade jQuery to 1.11.3.

jpetto:
Tested in Fx, Chrome, Safari, IE 7 - 10 and all seemed well. Tests are all passing.


alexgibson:
All unit & functional tests are passing :+1: 

I'll do some additional manual checking of high-impact pages on older browsers, but I think this change should be good.

alexgibson:
Tested the main download pages as low as IE6 and all seems well. Ran through most JS heavy pages in FF / Chrome / Safari and things generally seem well r+ :runner:

PR #3125, opened 2015-07-14
Bug 1171928 use tabzilla static across mozorg

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1171928

### DIFF
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -42,10 +42,6 @@
 
     {{ l10n_css() }}
 
-    {% block tabzilla_css %}
-      {% stylesheet 'tabzilla' %}
-    {% endblock %}
-
     <!--[if lte IE 8]>
       {# Only needed for IE before v9 #}
       <script src="{{ static('js/libs/html5shiv.js') }}"></script>
@@ -92,28 +88,31 @@
     {# for headers not to be confined by #wrapper (like fx family nav) #}
     {% block site_header_unwrapped %}{% endblock %}
 
-    <div id="wrapper">
+      <div id="wrapper">
 
-    {% block site_header %}
-      <header id="masthead">
+      {% block site_header %}
+        <header id="masthead">
+
         {% block tabzilla_tab %}
-        <a href="{{ url('mozorg.home') }}" id="tabzilla" data-infobar="{% block tabzilla_infobar %}{{ settings.TABZILLA_INFOBAR_OPTIONS }}{% endblock %}">Mozilla</a>
+          <div id="tabzilla">
+            <a href="{{ url('mozorg.home') }}">Mozilla</a>
+          </div>
         {% endblock %}
 
         {% block site_header_nav %}
-        <nav id="nav-main" role="navigation">
-          <span class="toggle" role="button" aria-controls="nav-main-menu" aria-expanded="false" tabindex="0">{{_('Menu')}}</span>
-          <ul id="nav-main-menu">
-            <li class="first mission-item"><a href="{{ url('mozorg.mission') }}">{{_('Mission')}}</a></li>
-            <li class="about-item"><a href="{{ url('mozorg.about') }}">{{_('About')}}</a></li>
-            <li class="products-item"><a href="{{ url('firefox.family.index') }}">{{_('Products')}}</a></li>
-            <li class="last contribute-item"><a href="{{ url('mozorg.contribute') }}">{{_('Get Involved')}}</a></li>
-          </ul>
-        </nav>
+          <nav id="nav-main" role="navigation">
+            <span class="toggle" role="button" aria-controls="nav-main-menu" aria-expanded="false" tabindex="0">{{_('Menu')}}</span>
+            <ul id="nav-main-menu">
+              <li class="first mission-item"><a href="{{ url('mozorg.mission') }}">{{_('Mission')}}</a></li>
+              <li class="about-item"><a href="{{ url('mozorg.about') }}">{{_('About')}}</a></li>
+              <li class="products-item"><a href="{{ url('firefox.family.index') }}">{{_('Products')}}</a></li>
+              <li class="last contribute-item"><a href="{{ url('mozorg.contribute') }}">{{_('Get Involved')}}</a></li>
+            </ul>
+          </nav>
         {% endblock %}
 
         {% block site_header_logo %}
-        <h2><a href="{{ url('mozorg.home') }}">{{ high_res_img('img/sandstone/header-mozilla-stone.png', {'alt': 'Mozilla', 'width': '170', 'height': '45'}) }}</a></h2>
+          <h2><a href="{{ url('mozorg.home') }}">{{ high_res_img('img/sandstone/header-mozilla-stone.png', {'alt': 'Mozilla', 'width': '170', 'height': '45'}) }}</a></h2>
         {% endblock %}
 
         {% block alt_header %}{% endblock %}
@@ -123,30 +122,31 @@
         {% block breadcrumbs %}{% endblock %}
 
         {% block site_header_share %}{% endblock %}
-      </header>
-    {% endblock %}
 
-    {% block mosaic %}{% endblock %}
+        </header>
+      {% endblock %}
 
-    {% block messages %}
-      {% if messages %}
-        <ul class="messagelist billboard">
-          {% for message in messages %}
-            <li{% if message.tags %} class="{{ message.tags }}"{% endif %}>{{ message }}</li>
-          {% endfor %}
-        </ul>
-      {% endif %}
-    {% endblock messages %}
+      {% block mosaic %}{% endblock %}
 
-    {% block content %}{% endblock %}
+      {% block messages %}
+        {% if messages %}
+          <ul class="messagelist billboard">
+            {% for message in messages %}
+              <li{% if message.tags %} class="{{ message.tags }}"{% endif %}>{{ message }}</li>
+            {% endfor %}
+          </ul>
+        {% endif %}
+      {% endblock messages %}
 
-    {% block email_form %}{{ email_newsletter_form() }}{% endblock %}
+      {% block content %}{% endblock %}
 
-    </div><!-- close #wrapper -->
+      {% block email_form %}{{ email_newsletter_form() }}{% endblock %}
 
-    {% block site_footer %}
-      {% include 'includes/site-footer.html' %}
-    {% endblock %}
+      </div><!-- close #wrapper -->
+
+      {% block site_footer %}
+        {% include 'includes/site-footer.html' %}
+      {% endblock %}
 
     </div><!-- close #outer-wrapper -->
 
@@ -156,9 +156,6 @@
 
     {% block site_js %}
       {% javascript 'common-resp' %}
-    {% endblock %}
-    {% block tabzilla_js %}
-      <script src="{{ settings.CDN_BASE_URL }}{{ url('tabzilla') }}"></script>
     {% endblock %}
     {% block js %}{% endblock %}
     {% if settings.DEV %}
--- a/bedrock/firefox/templates/firefox/all.html
+++ b/bedrock/firefox/templates/firefox/all.html
@@ -30,7 +30,7 @@
   sky firefox-all-{{ channel }} {% if channel == 'alpha' %}blueprint{% endif %}
 {% endblock %}
 
-{% block site_css %}
+{% block page_css %}
   {% stylesheet 'firefox_all' %}
 {% endblock %}
 
--- a/bedrock/firefox/templates/firefox/australis/fx38_0_5/firstrun.html
+++ b/bedrock/firefox/templates/firefox/australis/fx38_0_5/firstrun.html
@@ -23,10 +23,7 @@
 
 {% block body_class %}mozID{% endblock %}
 
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
-
-{% block site_css %}
+{% block page_css %}
   {% stylesheet 'firefox_fx38_0_5_firstrun' %}
 {% endblock %}
 
--- a/bedrock/firefox/templates/firefox/australis/tour-base.html
+++ b/bedrock/firefox/templates/firefox/australis/tour-base.html
@@ -23,8 +23,6 @@
 
 {% block body_class %}sky{% endblock %}
 
-{% block site_css %}{% endblock %}
-
 {% block string_data %}
 {# L10n: This is the label for a button that opens the tour modal dialog #}
 data-open="{{ _('Open tour') }}"
@@ -56,7 +54,9 @@ data-has-tour="True"
 
 {% block site_header %}
   <header id="masthead">
-    <a href="{{ url('mozorg.home') }}" id="tabzilla">Mozilla</a>
+    <div id="tabzilla">
+      <a href="{{ url('mozorg.home') }}">Mozilla</a>
+    </div>
     {% block site_header_nav %}{% endblock %}
     {% block site_header_logo %}{% endblock %}
     {% block breadcrumbs %}{% endblock %}
--- a/bedrock/firefox/templates/firefox/base-resp.html
+++ b/bedrock/firefox/templates/firefox/base-resp.html
@@ -18,7 +18,7 @@
 {% block twitter_id %}firefox{% endblock %}
 
 {% block site_css %}
-  {% stylesheet 'firefox-resp' %}
+    {% stylesheet 'firefox-resp' %}
 {% endblock %}
 
 {% block site_header_nav %}
--- a/bedrock/firefox/templates/firefox/dashboard-base.html
+++ b/bedrock/firefox/templates/firefox/dashboard-base.html
@@ -8,6 +8,7 @@
 {% block page_title %}{% endblock %}
 
 {% block site_css %}
+  {{ super() }}
   {% stylesheet 'firefox-dashboard' %}
 {% endblock %}
 
--- a/bedrock/firefox/templates/firefox/desktop/index.html
+++ b/bedrock/firefox/templates/firefox/desktop/index.html
@@ -21,7 +21,7 @@
   {% endif %}
 {% endblock %}
 
-{% block site_css %}
+{% block page_css %}
   {% stylesheet 'firefox_desktop' %}
 {% endblock %}
 
@@ -32,9 +32,6 @@
     {{ download_firefox(simple=True, small=True, dom_id="sticky-download-desktop") }}
   {% endcall %}
 {% endblock %}
-
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
 
 {% block content %}
 <main role="main" id="main-content">
--- a/bedrock/firefox/templates/firefox/desktop/tips.html
+++ b/bedrock/firefox/templates/firefox/desktop/tips.html
@@ -12,15 +12,16 @@
 {% block body_id %}firefox-desktop-tips{% endblock %}
 {% block body_class %}firefox-desktop-tips{% endblock %}
 
-{% block site_css %}
+{% block page_css %}
   {% stylesheet 'firefox_desktop_tips' %}
 {% endblock %}
 
 {% block site_header %}
   <header id="masthead">
     <div class="container">
-      <a href="{{ url('mozorg.home') }}" id="tabzilla" data-infobar="{{ settings.TABZILLA_INFOBAR_OPTIONS }}">Mozilla</a>
-
+      <div id="tabzilla">
+        <a href="{{ url('mozorg.home') }}">Mozilla</a>
+      </div>
       {% block site_header_nav %}
         {% include 'firefox/includes/top-menu.html' %}
       {% endblock %}
--- a/bedrock/firefox/templates/firefox/family/index.html
+++ b/bedrock/firefox/templates/firefox/family/index.html
@@ -25,9 +25,6 @@
   {% endif %}
 {% endblock %}
 
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
-
 {% block body_class %}mozid{% endblock %}
 
 {% block site_header %}
--- a/bedrock/firefox/templates/firefox/hello/index.html
+++ b/bedrock/firefox/templates/firefox/hello/index.html
@@ -52,9 +52,6 @@ set share_urls = {
 data-try-now="{{ _('Try it now') }}"
 {% endblock %}
 
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
-
 {% block content %}
 
 <main role="main">
--- a/bedrock/firefox/templates/firefox/interest-dashboard.html
+++ b/bedrock/firefox/templates/firefox/interest-dashboard.html
@@ -13,8 +13,7 @@
 #}
 {% block page_title %}{{ _('Firefox Interest Dashboard') }}{% endblock %}
 
-{% block site_css %}
-  {{ super() }}
+{% block page_css %}
   {% stylesheet 'firefox-interest-dashboard' %}
 {% endblock %}
 
@@ -23,9 +22,6 @@
 {% block site_header_unwrapped %}
   {{ fxfamilynav('desktop', '', 'dark') }}
 {% endblock %}
-
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
 
 {# The contents of the dashboard_intro block appears at the top of the page
    below the page title. Paragraphs should be wrapped like this:
--- a/bedrock/firefox/templates/firefox/tiles.html
+++ b/bedrock/firefox/templates/firefox/tiles.html
@@ -5,8 +5,8 @@
 {% from "macros.html" import fxfamilynav with context %}
 
 {% extends "/firefox/dashboard-base.html" %}
-{% block site_css %}
-  {{ super() }}
+
+{% block page_css %}
   {% stylesheet 'firefox-tiles' %}
 {% endblock %}
 
@@ -30,9 +30,6 @@ Our New Tab Experience
 {% block site_header_unwrapped %}
   {{ fxfamilynav('desktop', '', 'dark') }}
 {% endblock %}
-
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
 
 {# The contents of the dashboard_intro block appears at the top of the page
    below the page title. Paragraphs should be wrapped like this:
--- a/bedrock/foundation/templates/foundation/annualreport/2011.html
+++ b/bedrock/foundation/templates/foundation/annualreport/2011.html
@@ -25,7 +25,9 @@
 
 {% block site_header %}
   <header id="masthead">
-    <a href="{{ url('mozorg.home') }}" id="tabzilla" data-infobar="{{ settings.TABZILLA_INFOBAR_OPTIONS }}">{{ _('Mozilla') }}</a>
+    <div id="tabzilla">
+      <a href="{{ url('mozorg.home') }}">{{ _('Mozilla') }}</a>
+    </div>
 
     {% block site_header_nav %}{% endblock %}
 
--- a/bedrock/settings/static_media.py
+++ b/bedrock/settings/static_media.py
@@ -87,12 +87,14 @@ PIPELINE_CSS = {
     'common': {
         'source_filenames': (
             'css/sandstone/sandstone.less',
+            'css/tabzilla/tabzilla-static.less',
         ),
         'output_filename': 'css/common-bundle.css',
     },
     'responsive': {
         'source_filenames': (
             'css/sandstone/sandstone-resp.less',
+            'css/tabzilla/tabzilla-static.less',
         ),
         'output_filename': 'css/responsive-bundle.css',
     },
@@ -152,7 +154,6 @@ PIPELINE_CSS = {
     },
     'dnt': {
         'source_filenames': (
-            'css/sandstone/sandstone-resp.less',
             'css/firefox/family-nav.less',
             'css/base/mozilla-accordion.less',
             'css/firefox/dnt.less',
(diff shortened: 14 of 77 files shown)


bedrock/base/templates/base-resp.html
```diff
@@ -60,6 +56,7 @@
     {# Global styles, hidden from IE7 and lower #}
     {% block site_css %}
       {% stylesheet 'responsive' %}
+      {% stylesheet 'tabzilla-static' %}
```
alexgibson: I haven't dug into this PR yet, but given that the `tabzilla-static` CSS file is pretty tiny - is it actually a performance hit to include it like this and incur an extra http request, as opposed to bundling it in the `responsive` common CSS bundle?

malena: Yes the only css in the 'responsive' bundle is 'sandstone-resp.less', so I can definitely add it. Glad you suggested. I caught a few other places where tabzilla-static was part of page specific bundles I can remove.

bedrock/foundation/templates/foundation/annualreport/2011.html
```diff
@@ -11,6 +11,7 @@
 {% block body_class %}sand{% endblock %}
 
 {% block extrahead %}
+  {% stylesheet 'tabzilla-static' %}
```
jpetto: Could we add `tabzilla-static.less` to the `annual_2011` bundle instead of including two separate stylesheets?

malena: Ah yes! Will definitely change that.

media/css/firefox/desktop/trust.less
```diff
@@ -4,9 +4,10 @@
 
 @import 'common.less';
 
-// customize family nav
-.fxfamilynav-tabzilla:before {
-    background-color: @textColorSecondary;
+#fxfamilynav-tabzilla {
+    &:before {
```
sgarrity: Could these nested LESS selectors be simplified? To this:
```
#fxfamilynav-tabzilla:before {
```

bedrock/firefox/templates/firefox/all.html
```diff
@@ -30,7 +30,7 @@
   sky firefox-all-{{ channel }} {% if channel == 'alpha' %}blueprint{% endif %}
 {% endblock %}
 
-{% block site_css %}
+{% block page_css %}
```
jpetto: The `firefox_all` bundle in `static_media.py` still references `sandstone-resp.less`, which results in duplicate CSS.

bedrock/firefox/templates/firefox/desktop/index.html
```diff
@@ -33,9 +33,6 @@
   {% endcall %}
 {% endblock %}
 
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
```
jpetto: These empty blocks still exist in many templates. Should remove everywhere, no?

bedrock/firefox/templates/firefox/all.html
jpetto: On a related note, moving all these pages from `site_css` to `page_css` will result in 1 extra HTTP request per page, as we're now loading two separate bundles. Are we okay adding this overhead?

jpetto:
In review, I noticed at least a couple pages (`/firefox/pocket/` and `/firefox/independent/`) that load `sandstone-resp.less` twice - once in the `page_css` bundle and once in the `site_css` bundle. This PR did not introduce the code duplication, but is touching those CSS blocks, so I figured I'd mention it here. Should be cleaned up sooner rather than later, whether here or in a separate PR.

bedrock/firefox/templates/firefox/all.html
craigcook: > On a related note, moving all these pages from site_css to page_css will result in 1 extra HTTP request per page, as we're now loading two separate bundles. Are we okay adding this overhead?

I think we need to weigh the potential cost of the extra request against the potential savings of smaller files, especially if the global site styles are cached from a previous page. I'd love to do some performance testing to compare load times.

media/css/firefox/family-nav.less
```diff
@@ -4,6 +4,12 @@
 
 @import "../sandstone/lib.less";
 
+#tabzilla {
+    &:before {
+        background-color: @textColorSecondary;
+    }
+}
+
```
jpetto: This selector isn't necessary as the file stands. The family nav has its own tabzilla implementation (down on line 57). However, it probably is best to switch the family nav over to the standard implementation...which would require a little bit more work. Could be broken out to a separate PR - up to you.

bedrock/settings/static_media.py
```diff
@@ -826,6 +807,12 @@
         ),
         'output_filename': 'css/tabzilla-min.css',
     },
+    'tabzilla-static': {
```
jpetto: Doesn't look like this is used anymore after the change to the 2011 annual report?

bedrock/firefox/templates/firefox/all.html
jgmize: >> On a related note, moving all these pages from site_css to page_css will result in 1 extra HTTP request per page, as we're now loading two separate bundles. Are we okay adding this overhead?

> I think we need to weigh the potential cost of the extra request against the potential savings of smaller files, especially if the global site styles are cached from a previous page. I'd love to do some performance testing to compare load times.

+1, I've been guilty of [obsessing over HTTP request reduction](http://blog.getify.com/obsessions-http-request-reduction/) myself, but sometimes splitting the files up can yield better performance, despite the request overhead.

bedrock/firefox/templates/firefox/desktop/index.html
malena: Yes!

bedrock/firefox/templates/firefox/all.html
jpetto: Yeah, seems from a macro level to be an "it depends" situation. Does a cached resource count against the max parallel HTTP requests limit? The home page currently has 30 HTTP requests, and it seems the max parallel limit is around 8 in modern browsers, so maybe it doesn't matter? This is outside of my ken. :smile_cat:

Bring it up in tomorrow's meeting?

jpetto: FWIW, my gut feels the approach here (shared/cached bundle and page-specific bundle) is the right thing to do, but my gut cannot produce any supporting evidence.

jpetto:
Can remove a [little bit of the instructions](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/macros.html#L503) for the family nav macro.

alexgibson:
Friendly drive-by comment. If you rebase, there's a couple more empty tabzilla blocks that can be removed here when this lands:

https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/hello/start-40.0.html#L62-L64

:car: :runner:

bedrock/firefox/templates/firefox/all.html
malena: As discussed in meeting I'm going to leave the page_css block in. We can make another bug for testing better performance.

malena:
Thanks @alexgibson. Removing.

malena:
I've tackled the commented issues. Another review would be great!

malena:
@jpetto Whenever you have a chance this is ready to be reviewed.

jpetto:
Wondering if we should update the [tabzilla docs](http://bedrock.readthedocs.org/en/latest/tabzilla.html) as well. The docs have instructions for non-bedrock sites, so maybe not yet, but should probably file a bug to keep it on the radar?

bedrock/firefox/templates/firefox/australis/fx38_0_5/firstrun.html
```diff
@@ -23,10 +23,7 @@
 
 {% block body_class %}mozID{% endblock %}
 
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
-
-{% block site_css %}
+{% block page_css %}
   {% stylesheet 'firefox_fx38_0_5_firstrun' %}
```
jpetto: This bundle needs pruning - has its own reference to `sandstone.less` and `tabzilla-static.less`.

bedrock/firefox/templates/firefox/australis/tour-base.html
```diff
@@ -23,8 +23,6 @@
 
 {% block body_class %}sky{% endblock %}
 
-{% block site_css %}{% endblock %}
```
jpetto: Pages using this base template (e.g. [firefox/australis/firstrun-tour.html](https://github.com/malena/bedrock/blob/bug-1171928-use-tabzilla-static-across-mozorg/bedrock/firefox/templates/firefox/australis/firstrun-tour.html#L27)) end up with flip-flopped CSS order. The page specific stuff ends up in `site_css`, while the common global stuff ends up in `page_css` (due to inheriting from [firefox/base.html](https://github.com/malena/bedrock/blob/bug-1171928-use-tabzilla-static-across-mozorg/bedrock/firefox/templates/firefox/base.html#L18)).

It's kind of a goose chase to follow the inheritance chain. :confused: However, you can rather easily hack the site to load a specific firstrun template by adding a line like `template = 'firefox/australis/firstrun-tour.html'` to [line 493 in `firefox/views.py`](https://github.com/malena/bedrock/blob/bug-1171928-use-tabzilla-static-across-mozorg/bedrock/firefox/views.py#L493).

bedrock/firefox/templates/firefox/base-resp.html
```diff
@@ -18,7 +18,7 @@
 {% block twitter_id %}firefox{% endblock %}
 
 {% block site_css %}
-  {% stylesheet 'firefox-resp' %}
+    {% stylesheet 'firefox-resp' %}
```
jpetto: Nit - extra indentation (only because you're still making updates :smile:)

bedrock/firefox/templates/firefox/base.html
```diff
@@ -15,7 +15,7 @@
 {% block facebook_id %}14696440021{# facebook.com/Firefox #}{% endblock %}
 {% block twitter_id %}firefox{% endblock %}
 
-{% block site_css %}
+{% block page_css %}
```
jpetto: Hm, in `firefox/base-resp.html`, you leave this block as `site_css`. Any reason this template should be different?

bedrock/firefox/templates/firefox/dashboard-base.html
```diff
@@ -7,7 +7,7 @@
 {% block page_title_prefix %}{% endblock %}
 {% block page_title %}{% endblock %}
 
-{% block site_css %}
+{% block page_css %}
   {% stylesheet 'firefox-dashboard' %}
```
jpetto: Just noticed this bundle loads `menu-resp.less`, which isn't needed on the two templates that inherit from this template (`firefox/interest-dashboard.html` and `firefox/tiles.html`).

Also, those two inheriting pages have their CSS in `site_css`, which results in a weird CSS load order.

bedrock/firefox/templates/firefox/desktop/index.html
```diff
@@ -21,7 +21,7 @@
   {% endif %}
 {% endblock %}
 
-{% block site_css %}
+{% block page_css %}
```
jpetto: The other desktop pages (`fast.html`, `trust.html`, `customize.html`) still use the `site_css` block. Should those be switch to `page_css` as well?

bedrock/firefox/templates/firefox/family/index.html
```diff
@@ -25,9 +25,6 @@
   {% endif %}
 {% endblock %}
 
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
-
```
jpetto: This page is still using the `site_css` block. Should we switch to `page_css`?

bedrock/firefox/templates/firefox/hello/index.html
```diff
@@ -52,9 +52,6 @@
 data-try-now="{{ _('Try it now') }}"
 {% endblock %}
 
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
-
```
jpetto: This page is still using the `site_css` block as well.

bedrock/firefox/templates/firefox/interest-dashboard.html
```diff
@@ -24,9 +24,6 @@
   {{ fxfamilynav('desktop', '', 'dark') }}
 {% endblock %}
 
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
-
```
jpetto: Another one using `site_css`.

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -31,9 +31,6 @@
   {{ fxfamilynav('desktop', '', 'dark') }}
 {% endblock %}
 
-{% block tabzilla_css %}{% endblock %}
-{% block tabzilla_js %}{% endblock %}
-
```
jpetto: This page has confusing CSS inheritance. Might be better to switch this template from `site_css` to `page_css`?

media/css/firefox/desktop/trust.less
```diff
@@ -173,6 +172,11 @@ html[dir="rtl"] {
 }
 
 @media (max-width: @breakTablet) {
+
+    #fxfamilynav-tabzilla:before {
+        background-color: transparent; 
+    }
```
jpetto: Nice catch!

media/css/firefox/family-nav.less
```diff
@@ -4,6 +4,10 @@
 
 @import "../sandstone/lib.less";
 
+#tabzilla:before {
+    background-color: @textColorSecondary;
+}
```
jpetto: Is this selector necessary? Most (all?) pages using the family nav do not have an element with the id `tabzilla` - instead `fxfamilynav-tabzilla` is used.

media/css/styleguide/styleguide.less
```diff
@@ -36,6 +36,10 @@ a[href="#TODO"] { color: #f00 !important; }
   .offset(3);
 }
 
+#tabzilla:before {
+    background-color: #484848;
```
jpetto: Use `@textColorSecondary` instead?

jpetto:
A little bit more cleanup to do. Almost there! :persevere:

bedrock/firefox/templates/firefox/base.html
malena: Probably my mistake when updating site_css to page_css. This is a base template so it should stay site_css. Fixing this.

bedrock/firefox/templates/firefox/base-resp.html
malena: np :)

bedrock/firefox/templates/firefox/dashboard-base.html
malena: Thinking the dashboard-base.html shouldn't change to page_css instead leave as an overwriting site_css. Otherwise the CSS load order gets wacky since the child templates have their own site overwrites like below in tiles.html

{% block site_css %}
  {{ super() }}
  {% stylesheet 'firefox-tiles' %}
{% endblock %}

Perhaps leaving interest-dashboard.html with a site_css block and the two child templates can change to the page_css block.

What do you think @jpetto?

Not really sure what this {{ super() }} does but if it doesn't affect if I use page_css or site_css block, I would think using page_css would be better for the child templates.

bedrock/firefox/templates/firefox/desktop/index.html
malena: Reason they remain site_css is because they don't need tabzilla-static css because these templates all use the familynav. However index.html should also remain as site_css otherwise it has extra css it doesn't need. I would need to add back sandstone-less file into the  firefox-desktop bundle, since thats the only css index.html needed to inherit.

For me to update the site_css to page_css for these I would need to update the base template to use familynav. But since we don't want that (I am assuming familynav was specifically added to pages) for now leaving them as site_css and making a PR for maybe down the road removing this familynav implemenation of the tabzilla tab and using tabzilla-static instead. 

The only thing I'll be fixing here is adding back sandstone-less to firefox-desktop and updating the index to site_css block since firefox-desktop would have all the needed css.

Does that make sense @jpetto?

bedrock/firefox/templates/firefox/family/index.html
malena: Yes I can switch to page_css and there is no issue besides the fact that it inherits template-resp.css as well. Doesn't change the layout, but wondering if we even need template-resp.css at all? This may be another PR. 

Should we update or leave as is? @jpetto. I'm tempted to update, even if it adds this template-resp.css.

malena: Thinking I will create a new bug to tackle removing template-resp.css @alexgibson, @jpetto. For now I will update to page_css.

jbertsch:
Should we use the new tabzilla that espressive built here?  https://github.com/mozilla/tabzilla

bedrock/firefox/templates/firefox/hello/index.html
malena: This page uses the familynav not tabzilla which is why it uses site_css to overwrite. Will leave as is and we can tackle replacing familynav implementation with tabzilla in another PR.

bedrock/firefox/templates/firefox/interest-dashboard.html
malena: This page also uses familynav not tabzilla. Changing it to page_css breaks the menu. Thinking a few PR's need to come out of the bug, one of them being replacing the familynav with tabzilla-static.

malena:
@jbertsch Yep! I was pointed to this doc http://bedrock.readthedocs.org/en/latest/tabzilla.html#tabzilla-static, but I believe it's the same tabzilla-static. :)  

However there are a few sections with a different implementation of the tabzilla called familynav. I'm thinking it would be best to create another PR to replace those. It would be best to get this PR merged first. 

Thoughts @alexgibson @jpetto @craigcook?

media/css/styleguide/styleguide.less
malena: Yep, updating to var.

jpetto:
Agree that the Fx family nav tabzilla update should happen in a separate PR.

bedrock/settings/static_media.py
```diff
@@ -349,8 +343,6 @@
     },
     'firefox_fx38_0_5_firstrun': {
         'source_filenames': (
-            'css/sandstone/sandstone.less',
-            'css/tabzilla/tabzilla-static.less',
```
jpetto: Gah. So now this page no longer has `tabzilla-static.less`, because it's not included in the `firefox` bundle (which this page inherits from `firefox/base.html`). I'll leave this one up to you - should `tabzilla-static.less` be included in the `firefox` bundle, or added a la carte in each page?

bedrock/firefox/templates/firefox/australis/tour-base.html
jpetto: So the `firefox/australis/firstrun-tour.html` page has broken/missing CSS, but I think only because it's using the `site_css` block. Testing locally, switching `firefox/australis/firstrun-tour.html` to use `page_css` instead fixes the issue.

jpetto: Ah, also, `firefox/australis/firstrun-tour.html` doesn't load `tabzilla-static.less`. Maybe another reason to add it to the `firefox` LESS bundle?

bedrock/firefox/templates/firefox/dashboard-base.html
jpetto: Yep, I think leaving `site_css` in `dashboard-base.html` and using `page_css` in `interest-dashboard.html` and `tiles.html` is the way to go.

`{{ super() }}` basically means "insert the content of the parent template's block", meaning you can kind of extend a block instead of straight up overwriting.

So, I think `dashboard-base.html` would look like:

```
{% block site_css %}
  {# brings in the site_css bundle specified in the parent template (firefox/base-resp.html) #}
  {{ super() }}
  {% stylesheet 'firefox-dashboard' %}
{% endblock %}
```

Each child page would then have a basic `page_css` block, e.g.

```
{% block page_css %}
  {% stylesheet 'firefox-interest-dashboard' %}
{% endblock %}
```

bedrock/firefox/templates/firefox/family/index.html
jpetto: Agree that updating to `page_css` is the way to go for this PR.

I opened [bug 1186605](https://bugzilla.mozilla.org/show_bug.cgi?id=1186605) a couple days ago to audit LESS & JS bundles. Please make a note of this instance in that bug so we don't forget. :smile:

bedrock/firefox/templates/firefox/hello/index.html
jpetto: I'm okay leaving this page using `site_css`, as long as we make a note that it needs to be updated when we update tabzilla in the Fx family nav.

Do we have a bug for that yet? If not, can you file one and note this page will need its CSS block/bundle updated?

bedrock/firefox/templates/firefox/interest-dashboard.html
jpetto: As noted above, I think we can change this block to `page_css` as long as `dashboard-base.html` gets updated as well.

bedrock/firefox/templates/firefox/tiles.html
jpetto: As noted above, as long as we change `dashboard-base.html` to `site_css`, let's change this page to `page_css`.

bedrock/settings/static_media.py
jpetto: Feel free to do whatever is easier here (which may be just adding `tabzilla-static.less` to this page-specific bundle). I don't want to make us add any more high level decisions to this already large PR. We can re-think when we tackle [bug 1186605](https://bugzilla.mozilla.org/show_bug.cgi?id=1186605) if necessary.

jpetto:
I think we're just about there! This is hopefully the last round of feedback. :construction_worker:

bedrock/firefox/templates/firefox/hello/index.html
malena: I've opened a bug for this issue. https://bugzilla.mozilla.org/show_bug.cgi?id=1188852

bedrock/firefox/templates/firefox/family/index.html
malena: Great @jpetto! I have added another note in bug 1186605 regarding possible removal of template-resp.css.

bedrock/firefox/templates/firefox/dashboard-base.html
malena: @jpetto Thanks for explanation. I've also seen 

{% block extrahead %}
  {% stylesheet 'foo' %}
{% endblock %}

Does this do the same? Just asking to understand. Either way I've updated to the changes you have noted above. So now child templates extending from dashboard-base.html will have a page_css block instead.

bedrock/firefox/templates/firefox/tiles.html
malena: Yup! Changes made.

bedrock/firefox/templates/firefox/dashboard-base.html
jpetto: So `{% block extrahead %}` is basically a general purpose container for anything you might want to put in the `<head>` - CSS, JS, `<meta>` tags, etc. If you place CSS bundles in it, they will be shown to *all* browsers, whereas CSS bundles placed in `site_css` and `page_css` will be hidden from IE 7 and below. In short, it's a way to force CSS on IE 7 and lower.

bedrock/firefox/templates/firefox/interest-dashboard.html
malena: yep! updated dashboard-base.html

bedrock/settings/static_media.py
malena: I added tabzilla-static.less back to the firefox bundle because this is a shared bundle as referenced within firefox/base.html.

It wasn't inheriting styles correctly because the base was including the firefox bundle inside of the page_css block instead of the site_css block. 

So basically all base.html templates should not be using the page_css block otherwise that would most likely be overwritten by child pages (as it should be).

bedrock/firefox/templates/firefox/australis/fx38_0_5/firstrun.html
malena: removed sandston.less and tabzilla.less from firefox_fx38_0_5_firstrun bundle

bedrock/firefox/templates/firefox/australis/tour-base.html
malena: Added tabzilla and sandstone to firefox less bundle which should be inherited by all the australis templates.

media/css/firefox/family-nav.less
malena: Yeah I think this was my first edits when I missed some dependencies. I removed it.

malena:
@jpetto I've tackled the rest of the commented issues. Ready for another review.

bedrock/firefox/templates/firefox/desktop/index.html
jpetto: Yep, this makes sense. We'll clean it up in [bug 1186605](https://bugzilla.mozilla.org/show_bug.cgi?id=1186605).

bedrock/firefox/templates/firefox/desktop/tips.html
```diff
@@ -12,15 +12,16 @@
...
   {% stylesheet 'firefox_desktop_tips' %}
 {% endblock %}
 
 {% block site_header %}
   <header id="masthead">
     <div class="container">
-      <a href="{{ url('mozorg.home') }}" id="tabzilla" data-infobar="{{ settings.TABZILLA_INFOBAR_OPTIONS }}">Mozilla</a>
-
+      <div id="tabzilla">
+        <a href="{{ url('mozorg.home') }}" data-infobar="{{ settings.TABZILLA_INFOBAR_OPTIONS }}">Mozilla</a>
```
jpetto: We no longer need the `data-infobar` attribute here, as static tabzilla doesn't support any additional infobar messaging.

bedrock/foundation/templates/foundation/annualreport/2011.html
```diff
@@ -25,7 +25,9 @@
 
 {% block site_header %}
   <header id="masthead">
-    <a href="{{ url('mozorg.home') }}" id="tabzilla" data-infobar="{{ settings.TABZILLA_INFOBAR_OPTIONS }}">{{ _('Mozilla') }}</a>
+    <div id="tabzilla">
+      <a href="{{ url('mozorg.home') }}" data-infobar="{{ settings.TABZILLA_INFOBAR_OPTIONS }}">{{ _('Mozilla') }}</a>
```
jpetto: Don't need the `data-infobar` attribute here or below.

media/css/styleguide/styleguide.less
jpetto: Forgot to update to `@textColorSecondary`?

malena: ooops.. must have edited /static instead of /media. Added back in.

malena:
@jpetto Removed data-infobar in tabzilla and updated less. Good for final review hopefully!!

jpetto:
Okay, for the final check, I've been giving the site a visual once over. Found a few issues:

- `/firefox/releases/` is missing `tabzilla-static.less`. Looks like this template (`/firefox/releases/index.html`) needs updating to `page_css`, and also needs its `firefox_releases_index` LESS bundle updated.
- `/firefox/new/` shouldn't rely on the `site_css` block. This page should use the `extrahead` block to make sure *all* CSS is sent to *all* browsers (`site_css` and `page_css` are hidden from IE 7 and below). Currently, this page is a little off in terms of CSS in IE 7. Let's revert this page's LESS bundle, then add `tabzilla-static.less` to it. `/firefox/new/` is in dire need of improvements, which should be starting soon in [bug 1187057](https://bugzilla.mozilla.org/show_bug.cgi?id=1187057).
- Styleguide sub-pages have ended up with a duplicate next/prev nav kind of tucked away under tabzilla:
screen shot 2015-07-30 at 12 51 17 pm

So...close... :partly_sunny:

malena:
Hey @jpetto I've updated the first and third bullet point. 

For the second bullet, the firefox/new page is looking fine to me. The firefox/new.html page is currently using the block extrahead: https://github.com/malena/bedrock/blob/bug-1171928-use-tabzilla-static-across-mozorg/bedrock/firefox/templates/firefox/new.html#L37 and it's also inheriting the tabzilla-static. No need to add it back in to the firefox_new bundle?

Maybe I'm missing something from this second point?

jpetto:
The problem with `/firefox/new/` only shows up in IE 7 or below. Those browsers do not get any CSS loaded in `site_css` or `page_css` - meaning no `tabzilla-static.less` inheritance. Here's how it looks in my IE 7 VM:

screen shot 2015-07-30 at 3 46 46 pm

This is kind of the reason the `extrahead` block exists - for the outlier pages that can't follow our standard CSS degradation rules.

`/firefox/new/` is probably the most important page on the site, which is why we want it to look nice, even on *super* old versions of IE. In the future, this page will likely become even more of a special & unique :snowflake:, possibly not inheriting from any base templates at all.

The styleguide and releases pages look good. One more to go!

malena:
Sorry @jpetto. Totally end of day brain fart. K it's all updated.

bedrock/firefox/templates/firefox/new.html
```diff
@@ -33,7 +33,6 @@
   <h2>{{ high_res_img('img/firefox/new/header-firefox.png', {'alt': _('Firefox for desktop'), 'width': '223', 'height': '84'}) }}</h2>
 {% endblock %}
 
-{% block site_css %}{% endblock %}
```
jpetto: This emptying of the `site_css` block needs to remain, otherwise modern browsers get served duplicate CSS.

This is the last thing :exclamation:

malena: @jpetto done

jpetto:
r+!

dance

PR #3127, opened 2015-07-15
Fix bug 1179537 - remove "social media" option from get involved page

craigcook:


schalkneethling:
r+ assuming +1 on the bug.

PR #3129, opened 2015-07-15
Bug 1184281. Fix family nav on tertiary pages.

jpetto:


alexgibson:
Tested pages locally, works as expected r+ :dango:

PR #3132, opened 2015-07-16
Bug 1180325 - Grammar error on the sync feature page

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1180325


sgarrity:
I merged this without taking into account the l10n side-effects. This will now need to be updated to only use the updated string in en-US.

PR #3138, opened 2015-07-17
Fix Bug 1133266 - Move Thunderbird Privacy Policy to Bedrock

kyoshino:
The privacy policy doc is already in [legal-docs](https://github.com/mozilla/legal-docs/blob/master/thunderbird_privacy_policy/en-US.md) which has been merged to Bedrock as part of #3144.

The archive page is no longer updated, so the strings are not wrapped for l10n.


malena:
r+

PR #3141, opened 2015-07-20
Casper tests contribute stories

schalkneethling:
@alexgibson r? Here I combined the tests for https://www.mozilla.org/en-US/contribute/stories/ and individual community member pages such as /contribute/stories/shreyas/ into one test suite as they are essentially the same page.

### DIFF
--- a/tests/functional/contribute-stories.js
+++ b/tests/functional/contribute-stories.js
@@ -1,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public" + '/n' +
+ "* License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global casper */
+'use strict';
+
+var config = require('../lib/config');
+var helpers = require('../lib/helpers');
+var path = '/contribute/stories/shreyas/';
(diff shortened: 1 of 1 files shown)


tests/functional/contribute-stories.js
```diff
@@ -0,0 +1,65 @@
...
+        helpers.done();
+    });
+});
+
+casper.test.begin('Contribute Stories, Individual Story Page: ' + url, 5, function suite(test) {
+    casper.start(url, function() {
+        this.click('.stories-other .person:first-child a.url');
+    });
+
+    casper.waitForUrl(/contribute\/stories\/shreyas\/$/, function() {
```
alexgibson: Better just to load this page directly maybe? This seems like a separate test going on here , rather than anything directly related to the content on `/contribute/stories/`. I don't think verifying that a link from one page to the other is working really adds much, seeing as we have tests for both pages individually?

schalkneethling: Yeah I agree. When I realized that they are in fact exactly the same page, I was wondering whether I should not just load the individual member page and be done with it. This settles it then.

alexgibson:
Care to squash?

alexgibson:
r+ :gem:

PR #3145, opened 2015-07-21
Fix Bug 1185457, remove required attrbiute from field when selection changes and container collapse

schalkneethling:
@alexgibson r?

### DIFF
--- a/media/js/mozorg/contribute/contribute-2015.js
+++ b/media/js/mozorg/contribute/contribute-2015.js
@@ -155,8 +155,17 @@
                 show_area(categoryarea);
             }
         } else {
+            // get and store the previous area before collapsing and hiding
+            // it's container else, we will not be able to get att it using
+            // the :visible psuedo selector.
+            var previouslySelectedArea = $('.area:visible');
+            previouslySelectedArea.find('select')
+                .prop('selectedIndex', 0)
+                .attr('required', false);
             $areascontainer.slideUp('fast', function() {
-                $('.area:visible').hide().find('select').prop('selectedIndex', 0);
+                // instead of using .hide() which is going to change in the
+                // upcoming jQuery 3 release, just remove the style attrbiute.
+                previouslySelectedArea.removeAttr('style');
             });
         }
     };
--- a/tests/functional/contribute-signup.js
+++ b/tests/functional/contribute-signup.js
@@ -39,7 +39,7 @@ casper.test.begin('Contribute Signup, Elements: ' + url, 15, function suite(test
     });
 });
 
-casper.test.begin('Contribute Signup, Areas of interest interaction: ' + url, 8, function suite(test) {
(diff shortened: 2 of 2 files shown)


media/js/mozorg/contribute/contribute-2015.js
```diff
@@ -155,8 +155,17 @@
                 show_area(categoryarea);
             }
         } else {
+            // get and store the previous area before collapsing and hiding
+            // it's container else, we will not be able to get att it using
```
alexgibson: Nit: typo `at`?

media/js/mozorg/contribute/contribute-2015.js
```diff
@@ -155,8 +155,17 @@
...
+            // the :visible psuedo selector.
+            var previouslySelectedArea = $('.area:visible');
+            previouslySelectedArea.find('select')
+                .prop('selectedIndex', 0)
+                .attr('required', false);
             $areascontainer.slideUp('fast', function() {
-                $('.area:visible').hide().find('select').prop('selectedIndex', 0);
+                // instead of using .hide() which is going to change in the
+                // upcoming jQuery 3 release, just remove the style attrbiute.
+                previouslySelectedArea.removeAttr('style');
```
alexgibson: This works, but if you want to make it work fully in jQuery 3, I believe we would need to use a class such as `.hidden` on each area, and then remove/add it when needed. As far as I understand, even calling `slideDown` on an element which is set to `display: none;` in the page's CSS wouldn't work. 

Up to you if you want to go that far in this PR of course (much of this page would probably need attention to make it all work).

alexgibson:
r+wc :racehorse:

media/js/mozorg/contribute/contribute-2015.js
schalkneethling: hmmmm, so the problem here is with slideDown which will set `style="display:block"` inline and not add a class that I can remove so, that is why I opted for just removing the `style` attribute here. Perhaps utilizing the callback on slideDown one could override that behavior and instead and the class `.hidden`?

alexgibson: If I read the code correctly, it's the area container that is having `slideDown` applied. Here, you're removing the styles applied to `.area`, which is having `fadeIn` and `fadeOut` applied in the `show_area()` function.

But honestly, this is largely out of scope for this bug I think. I'm not sure if this will fix the issue for jQuery 3, but it still works here.

schalkneethling: Bah! but of course. We will have to do one big audit once jQuery 3.0 get's to beta I reckon.

alexgibson: Yeah, we can probably make a good tracking bug for things we could do to fix / make the transition easier prior to updating. It will take some time certainly.

PR #3146, opened 2015-07-21
Fix Bug 1185475, ensure all fields are implcitly or explicitly associated a11y

schalkneethling:
@craigcook @alexgibson r? See comments here for more:
https://bugzilla.mozilla.org/show_bug.cgi?id=1185475#c1

### DIFF
--- a/bedrock/mozorg/templates/mozorg/contribute/signup.html
+++ b/bedrock/mozorg/templates/mozorg/contribute/signup.html
@@ -122,17 +122,17 @@
         <h2 class="section-title">{{ _('Now tell us a little about yourself:') }}</h2>
         <ul>
           <li>
-            <label for="person-name">{{ _('Name') }}</label>
+            <label for="id_name">{{ _('Name') }}</label>
             {{ form.name }}
             {{ form.name.errors }}
           </li>
           <li>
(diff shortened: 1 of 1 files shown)


craigcook:
Ugh, I should have done this months ago. How embarrassing. Thanks for cleaning up my mistakes!

We can also add <code>for</code> attributes to the two email format radios. Even though the label wraps around the control, an explicit association can't hurt.

Otherwise r+wc! :monkey:

schalkneethling:
@craigcook I had it like that, and also added `for` attributes for the categories list but, after talking on the a11yslackers gitter channel (https://twitter.com/stevefaulkner/status/517996868740329472), it seems that, unless we are concerned with <= IE7, implicitly linking the label with the field is handled perfectly by assistive technologies.

With that said however, I am more than happy to add the explicit associations. Thoughts? Thanks!

craigcook:
Yeah, only really old IE requires a <code>for</code> attribute to make labels clickable. It's not really necessary these days but it also doesn't hurt anything either. I don't have a strong opinion on the matter, just thought I'd mention it. Happy to merge without them.

schalkneethling:
Thanks @craigcook :cake:

PR #3149, opened 2015-07-22
Use lazy EMAIL_BACKEND and add more env vars

jgmize:


jgmize:
@pmclanahan r?

@jpetto this should fix your local error handling issue.

bedrock/settings/base.py
```diff
@@ -453,10 +453,15 @@ def JINJA_CONFIG():
 # This can be changed to use session once we do add a database.
 MESSAGE_STORAGE = 'django.contrib.messages.storage.cookie.CookieStorage'
 
-EMAIL_BACKEND = config(
-    'EMAIL_BACKEND',
-    default='django.core.mail.backends.console.EmailBackend' if DEBUG else
+
+def lazy_email_backend():
```
pmclanahan: I think this is a fine solution, but the other way we solve these kinds of things is to have them in `settings/__init__.py`. I'm thinking we'll eventually want to get everything out of `__init__.py`, so my question is whether you think this PR is a good chance to do some of that?

jgmize: If you don't make it lazy, you either can't override it from local.py (if you put it after the `from .local import *`) or you can't default it based on the DEBUG setting in local.py, unless I'm mistaken.

jgmize: Once we're done eliminating local.py I don't really see a point in keeping settings as a dir anyway-- I'd rather just have settings.py pull everything from env vars.

pmclanahan: > If you don't make it lazy, you either can't override it from local.py (if you put it after the from .local import *) or you can't default it based on the DEBUG setting in local.py, unless I'm mistaken.

No, you're right. You'd have to do some nasty `if 'SOME_SETTING` not in globals():` trick, and using `lazy` seems less bad.

Also you're right about pulling settings back into a single file, well, plus the separate static_media file. My vote would be to keep that separate.

PR #3150, opened 2015-07-22
Fix /firefox/unsupported-systems/ menu.

jpetto:
Noticed this page's menu was broken when looking through Fx URLs the other day. Didn't think it warranted a bug.


alexgibson:
Nice spot, r+ :yen:

PR #3151, opened 2015-07-23
[fix bug 1186485] Welcome page for Firefox users on Windows 10

alexgibson:


bedrock/firefox/templates/firefox/win10-welcome.html
```diff
@@ -0,0 +1,144 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/base-resp.html" %}
+{% add_lang_files "firefox/whatsnew_40" %}
```
alexgibson: @flodolo - worth renaming/moving the lang file for this new template?

flodolo: Yes, let's do it. I also need to update the URL for reference.

alexgibson:
Ok marking this as ready - some notes for the reviewer:
* You'll need to have UITour set up locally for the page to work fully and test in Firefox 40 or above.
* This page is being launched by Firefox on Windows10 machines only. QA is happening to make sure it behaves as expected on that platform, but the functionality still works on other platforms (such as OSX) Only the page copy is Win10 specific.
* There are two page states, depending on if Firefox is your default browser when the page loads.
* This is a public URL, but it's not being linked to from anywhere. As such, we're not too concerned how it looks on other browsers, as long as it doesn't appear drastically broken.

Happy to talk through any other questions that may arise.

bedrock/firefox/templates/firefox/win10-welcome.html
```diff
@@ -0,0 +1,140 @@
...
+  <section class="firefox-default-cta hidden">
+    <h2>{{ _('Make Firefox your default in 3 easy steps:') }}</h2>
+    <ol>
+      <li>{{ _('Click the button below') }}</li>
+      <li>{{ _('Scroll down to “Web browser”') }}</li>
+      <li>{{ _('Choose Firefox') }}</li>
+    </ol>
+    <button id="set-default" type="button" class="button-flat-dark">{{ _('Let’s do it') }}</button>
+  </section>
+  <section class="firefox-learn-links">
```
schalkneethling: Should this not be using the hidden class as well, as apposed to `display: none;` in CSS?

schalkneethling: Ah, never mind. I see you are not using JS here but CSS to animate :+1:

media/css/sandstone/fira-fonts.less
```diff
@@ -0,0 +1,70 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@font-face {
+    font-family: 'Fira Sans';
+    src: url('/media/fonts/FiraSans-Regular.eot?#iefix') format('embedded-opentype'),
+         url('/media/fonts/FiraSans-Regular.woff2') format('woff2'),
```
schalkneethling: Recently watched a talk on font performance and usability and in it, the guy suggests putting woff2 first and then woff, otf etc. Perhaps worth considering here?

alexgibson: Would this not effect old IE downloading multiple file types?

media/js/firefox/win10-welcome.js
```diff
@@ -0,0 +1,150 @@
...
+    };
+
+    /**
+     * Track the first time Welcome page tab is made visible in GA
+     * @param {docHidden} boolean value for testing purposes only.
+     */
+    Win10Welcome.checkTabVisibility = function(docHidden) {
+        var hidden = docHidden !== undefined ? docHidden : document.hidden;
+        if (hidden) {
+            $document.one('visibilitychange.win10', Win10Welcome.trackTabVisibility);
```
schalkneethling: `document.on(....` ?

alexgibson: Only needs to happen once, so `$document.one` is correct here.

schalkneethling: After looking at the tests, it seems this should indeed be `.one`

schalkneethling: Aaaah yes, http://api.jquery.com/one/

media/css/sandstone/fira-fonts.less
schalkneethling: I have not personally tested it so, perhaps. Should be that the browser downloads the first format it supports though.

alexgibson: I think this warrants some further investigation. Most instances of bullet-proof `@font-face` syntax recommend putting `.eot` first. Not quite sure the reason, but I believe old IE's may have had issues with different ordering. I'm going to leave this as-is for now and we can always revisit when we're more confident.

schalkneethling:
This is looking real good and works great. I just need to test this on a Win10 machine but so far :cake: 

One thing that does concern me however is the color contrast if the yellow text links. The contrast/luminosity is way to low. I know this is as it was designed but, perhaps we can make a case to change it to white?

The SVG icons lighting up when hovered over is great but, the text links, not so much. See screenshot below:

screen shot 2015-08-04 at 11 04 34

alexgibson:
Good call on the link contrast - let me ping Lee and we'll find a suitable alternative :+1:

schalkneethling:
Other than the color contrast issue, this is a big r+ Nicely done @alexgibson :dragon:

alexgibson:
Thanks @schalkneethling - Lee agreed to change the link color to white, so I have made it so. I'm going to hold off merging this until I have approval. Thanks for the review!

schalkneethling:
Yeah! My pleasure @alexgibson

ckprice:
@alexgibson cleared for takeoff

alexgibson:
Thanks @ckprice, merging!

PR #3152, opened 2015-07-23
Remove coveralls ping from travis

schalkneethling:
@alexgibson r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,16 +25,13 @@ before_script:
   - sh -e /etc/init.d/xvfb start
 script:
   - grunt test
-  - coverage run manage.py test
+  - python manage.py test
 before_install:
   - travis_retry git submodule update --init --recursive
 install:
   - travis_retry python bin/peep.py install -r requirements/pip.txt --no-use-wheel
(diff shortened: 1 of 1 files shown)


alexgibson:
Does this just remove the travis ping, or stop coveralls altogether?

schalkneethling:
From what I recall speaking to @pmclanahan, this will stop coverage from running on the python code and also stop the ping to coveralls. I have been meaning to remove it so we can add the JS unit tests but, if coveralls is misbehaving, perhaps we should just leave it be.

.travis.yml
```diff
@@ -25,7 +25,7 @@ before_script:
   - sh -e /etc/init.d/xvfb start
 script:
   - grunt test
-  - coverage run manage.py test
+  - manage.py test
```
pmclanahan: has to be `python manage.py test`

schalkneethling:
You know what, I bet that needs to be `./manage.py test

.travis.yml
schalkneethling: Thanks! Just figured it needs to be something like that as it is not run by coverage anymore :blush:

schalkneethling:
Whoops, we need to remove the after-success block as well else it still pings coveralls.

PR #3154, opened 2015-07-24
Redirect windows-10 welcome: fix bug 1187343

jgmize:


jgmize:
@pmclanahan or @alexgibson r?

alexgibson:
Do we still need a pass-though redirect in global.conf for it to get to this redirect?

jgmize:
> Do we still need a pass-though redirect in global.conf for it to get to this redirect?

Good catch, yes we do. Updating.

jgmize:
Updated to include global.conf PT

alexgibson:
r+ :speedboat:

PR #3157, opened 2015-07-26
Adds functional tests for styleguide

schalkneethling:
@alexgibson r?

### DIFF
--- a/tests/functional/styleguide.js
+++ b/tests/functional/styleguide.js
@@ -1,0 +1,168 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public" + '/n' +
+ "* License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global casper */
+'use strict';
+
+var config = require('../lib/config');
+var helpers = require('../lib/helpers');
+var path = '/styleguide/';
+var url = config.base() + path;
+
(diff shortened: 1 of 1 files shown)


tests/functional/styleguide.js
```diff
@@ -0,0 +1,81 @@
...
+        test.assertVisible('#sidebar', 'The sidebar navigation is visible');
+    });
+
+    casper.run(function() {
+        test.done();
+        helpers.done();
+    });
+});
+
+casper.test.begin('Styleguide, Navigation flow: ' + url, 11, function suite(test) {
```
alexgibson: Feels like there's a lot going on in this one test? Might it be better to split this up a bit?

I don't think we gain much by clicking a link, waiting for another page to load and then checking the URL. It would be much faster just to check that the link's href is what we expect? If we want to test something at that URL, then make it it's own test? Doing this for a few key style guide pages would be enough I think.

schalkneethling: Makes sense, I was thinking to do an full navigation flow but, perhaps better to make the tests smaller and more succinct. It does make the entire suite run slower now, because there is a lot of `waiting` involved.

Did you have any particular pages in mind?

schalkneethling:
@alexgibson updated. Let me know if this covers enough of the bases or if you reckon I need to add more. Thanks!

alexgibson:
r+ care to squash?

PR #3159, opened 2015-07-27
Fix Bug 1172986, show message when no plugins are found

schalkneethling:
@alexgibson r?

### DIFF
--- a/bedrock/mozorg/templates/mozorg/plugincheck.html
+++ b/bedrock/mozorg/templates/mozorg/plugincheck.html
@@ -101,66 +101,79 @@
       <p class="enable-js warning">{{_('Please turn on JavaScript to display this page correctly.')}}</p>
 
       <h2>{{_('Plugin Status')}}</h2>
-      <ul>
-        <li>{{_('Step 1: Click Update to update a plugin.')}}</li>
-        <li>{{_('Step 2: Complete all recommended updates before restarting your browser.')}}</li>
-      </ul>
 
-      <div class="plugincheck-loader"></div>
+      <div class="plugincheck-loader hidden"></div>
 
-      <div id="sec-plugin-vulnerable">
-        <table class="table status">
-          <caption>{{_('Potentially vulnerable plugins')}}</caption>
-          <thead>
-            <tr>
-              <th class="plugin-status-detail" scope="col">{{_('Plugin')}}</th>
-              <th class="plugin-status-status" scope="col">{{_('Status')}}</th>
-              <th class="plugin-status-action action" scope="col">{{_('Action')}}</th>
-            </tr>
-          </thead>
-          <tbody id="plugin-vulnerable"></tbody>
-        </table>
+      <div id="no-plugins" class="hidden">
+      {% if l10n_has_tag('no_plugins') %}
+      <p>
+        {% trans bugzilla_url='https://bugzilla.mozilla.org/enter_bug.cgi?product=Plugin%20Check&component=UI' %}
+          No plugins were found. If you believe this is an error, <a href="{{ bugzilla_url }}">please file a bug</a> and let us know.
+        {% endtrans %}
+      </p>
+      {% endif %}
       </div>
 
-      <div id="sec-plugin-outdated" class="shadow-separator">
-        <table class="table status">
-          <caption>{{_('Outdated Plugins')}}</caption>
-          <thead>
-            <tr>
-              <th class="plugin-status-detail" scope="col">{{_('Plugin')}}</th>
-              <th class="plugin-status-status" scope="col">{{_('Status')}}</th>
-              <th class="plugin-status-action action" scope="col">{{_('Action')}}</th>
-            </tr>
-          </thead>
-          <tbody id="plugin-outdated"></tbody>
-        </table>
-      </div>
+      <div id="plugins" class="hidden">
+        <ul>
+          <li>{{_('Step 1: Click Update to update a plugin.')}}</li>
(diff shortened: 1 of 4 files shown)


bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -101,66 +101,77 @@ <h1 class="large">{{_('Check Your Plugins')}}</h1>
...
-          <thead>
-            <tr>
-              <th class="plugin-status-detail" scope="col">{{_('Plugin')}}</th>
-              <th class="plugin-status-status" scope="col">{{_('Status')}}</th>
-              <th class="plugin-status-action action" scope="col">{{_('Action')}}</th>
-            </tr>
-          </thead>
-          <tbody id="plugin-outdated"></tbody>
-        </table>
+      {% if l10n_has_tag('no_plugns') %}
```
flodolo: typo

schalkneethling: dang it :blush:

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -101,66 +101,77 @@ <h1 class="large">{{_('Check Your Plugins')}}</h1>
...
-          <thead>
-            <tr>
-              <th class="plugin-status-detail" scope="col">{{_('Plugin')}}</th>
-              <th class="plugin-status-status" scope="col">{{_('Status')}}</th>
-              <th class="plugin-status-action action" scope="col">{{_('Action')}}</th>
-            </tr>
-          </thead>
-          <tbody id="plugin-outdated"></tbody>
-        </table>
+      {% if l10n_has_tag('no_plugins') %}
```
alexgibson: There's no else conditional for this l10n tag. Might it be better just to show the message in English here if no translation is available? Seems like this state is an unlikely occurrence. But if someone does hit it, better to show some form of textual feedback than nothing at all? Thoughts @flodolo?

schalkneethling: I agree, the reason I left it with no else, was because currently there is no message if there are no plugins so, the current behavior does not change so, it felt like a safe fallback.

alexgibson: Yeah. We do have to be a little careful conditionally including html like this, for example when using `$('#no-plugins')` on a non-localized page. Here we're ok and no JS will break, but it does increase scope for error. If we don't remove the l10n tag, perhaps we should put the tag inside the element as an additional measure.

flodolo: In this specific case I prefer no message to an English message.

alexgibson: Ok, then I think we should just move the l10n tag to wrap the text only, and not the whole HTML element.

tests/functional/plugincheck.js
```diff
@@ -81,3 +81,19 @@ casper.test.begin('PluginCheck, outated Firefox: ' + url, 1, function suite(test
...
     });
 });
+
+casper.test.begin('PluginCheck, no plugins found message: ' + url, 2, function suite(test) {
+    casper.start();
+
+    casper.userAgent(config.userAgent.fx);
+
+    casper.thenOpen(url, function() {
+        test.assertVisible('#no-plugins', 'The no plugins container is visible');
```
alexgibson: Is this just assuming no plugins since it is launching with Casper? Should probably add a comment explaining this assumption if so. Also, can we be sure this will always be the case?

schalkneethling: Yes, that is the assumption so, will add a comment about that. Whether this will always be the case, well, I believe so. Casper does have a flag to tell it to load plugins but, having tried that previously it did not seem to work for me. With that said then, seeing that, that flag is false by default, this would be a safe assumption:

http://docs.casperjs.org/en/latest/modules/casper.html#pagesettings

bedrock/mozorg/templates/mozorg/plugincheck.html
schalkneethling: I was also thinking about the JS error probability and yes, as you stated, in this case it will just fail silently but, better to err on the side of caution ;)

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -101,66 +101,77 @@ <h1 class="large">{{_('Check Your Plugins')}}</h1>
...
-              <th class="plugin-status-status" scope="col">{{_('Status')}}</th>
-              <th class="plugin-status-action action" scope="col">{{_('Action')}}</th>
-            </tr>
-          </thead>
-          <tbody id="plugin-outdated"></tbody>
-        </table>
+      <div id="no-plugins" class="hidden">
+      {% if l10n_has_tag('no_plugins') %}
+        {% trans bugzilla_url='https://bugzilla.mozilla.org/enter_bug.cgi?product=Plugin%20Check&component=UI' %}
+        <p>No plugins were found. If you believe this is an error, <a href="{{ bugzilla_url }}">please file a bug</a> and let us know.</p>
```
alexgibson: Nit: move the `<p>` outside of the `trans` block?

tests/functional/plugincheck.js
alexgibson: Ok sounds good to me :+1:

alexgibson:
r+wc, care to squarsh?

PR #3164, opened 2015-07-29
[fix bug 1186221] FxOS Orange Klif Egypt availability.

jpetto:
- Add MTN to list of One Touch Fire E operators.


schalkneethling:
r+ :large_orange_diamond: @flodolo good to merge? Does not seem like there is something l10n wise we need based on https://bugzilla.mozilla.org/show_bug.cgi?id=1186221#c1 Thanks!

flodolo:
Yes, we don't block merges on this page.

PR #3167, opened 2015-07-30
Fix bug 1182720 - update Webmaker promos after Maker Party

craigcook:


flodolo:
Note: I don't plan to push these requests out before Aug 7, we have tons of tile with that deadline for the release.

media/css/mozorg/home/home-promo.less
```diff
@@ -839,31 +839,6 @@ html[lang|="en"] {
...
-
-/****************************************************************************/
-// Maker Party
-.promo-large-landscape.makerparty {
-    background: #e14164; // pinkish red
-
-    h2 {
-        padding: 110px @baseLine 0 @baseLine;
-        margin-top: @baseLine;
-        background-image: url(/media/img/home/voices/promos/makerparty/maker-party.png);
```
jpetto: Do we need to keep this image file?

jpetto:
Aside from a possible orphaned image, looks good. Not sure if we need to wait for image removal to merge?

r+wc

PR #3168, opened 2015-07-30
[fix bug 1150231] Fx40 firstrun experience.

jpetto:
Will need to merge prior to Fx40 launch on 8/11 so we can test FxA integration on stage. FxA should be ready to test on stage on Tuesday, 8/4.

### DIFF
--- a/bedrock/firefox/templates/firefox/australis/fx40/firstrun.html
+++ b/bedrock/firefox/templates/firefox/australis/fx40/firstrun.html
@@ -1,0 +1,131 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% add_lang_files "firefox/whatsnew_38" %}
+
+{% extends "firefox/base.html" %}
+
+{% block extra_meta %}<meta name="robots" content="noindex">{% endblock %}
+
+{#- This will appear as <meta property="og:title"> which can be used for social share -#}
+{% block page_og_title %}
+  {{_('Choose the independent browser')}}
+{% endblock %}
+
+{#- This will appear as <meta property="og:description"> which can be used for social share -#}
+{% block page_og_desc %}
+  {{_('Firefox is non-profit, non-corporate, non-compromised. Choosing Firefox isn’t just choosing a browser. It’s a vote for personal freedom.')}}
+{% endblock %}
+
+{#- Override <meta property="og:url"> for social share -#}
+{% block page_og_url %}{{ url('firefox') }}{% endblock %}
+
+{% block body_class %}mozID{% endblock %}
+
+{% block site_css %}
+  {% stylesheet 'firefox_fx40_firstrun' %}
+{% endblock %}
+
+{% block optimizely %}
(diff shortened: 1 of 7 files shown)


jpetto:
[testing locally (against stomlinson's demo server)](https://bugzilla.mozilla.org/show_bug.cgi?id=1150231#c4)
[testing on demo4 (against FxA stage server)](https://bugzilla.mozilla.org/show_bug.cgi?id=1150231#c28)

URL for Fx40 firstrun is: `/firefox/40.0/firstrun/`

alexgibson:
Working locally \o/ - that took quite some fiddling

bedrock/firefox/templates/firefox/australis/fx40/firstrun.html
```diff
@@ -0,0 +1,134 @@
...
+{% block page_og_desc %}
+  {{_('Firefox is non-profit, non-corporate, non-compromised. Choosing Firefox isn’t just choosing a browser. It’s a vote for personal freedom.')}}
+{% endblock %}
+
+{#- Override <meta property="og:url"> for social share -#}
+{% block page_og_url %}{{ url('firefox') }}{% endblock %}
+
+{% block body_class %}mozID{% endblock %}
+
+{% block tabzilla_css %}{% endblock %}
```
alexgibson: Nit: old `tabzilla_css` and `tabzilla_js` blocks no longer needed.

alexgibson:
RTL styles could do with a little tweak, mainly for the header logo:

rtl-header

alexgibson:
Signup seems to be working ok for me locally, and code looks good. I will trust that QA and growth team have verified all the postMessage & GA events are firing as expected :+1:

bedrock/settings/base.py
```diff
@@ -689,6 +689,20 @@ def facebook_tab_url_lazy():
 # Optimize.ly project code
 OPTIMIZELY_PROJECT_ID = None
 
+# Fx Accounts iframe source
+# Prod, stage, dev, & demos have this set in their local.py files.
+#
+# For local testing:
+#   - Point FXA_IFRAME_SRC to stomlinson's instance in your own local.py:
+#     FXA_IFRAME_SRC = 'https://stomlinson.dev.lcip.org/'
```
alexgibson: Are we going to get a nicer way to test locally in the long run? Seems like depending on stomlinson's own dev instance is probably not what we want long-term?

alexgibson:
r+wc from me

If FxA signup is going to be a permanent fixture for /firstrun, the only other thing I think would be good would be add some JS jasmine tests to verify all the correct things happen for the right messages received etc. Would require some refactoring of the JS, but worthwhile doing post-launch perhaps?

bedrock/settings/base.py
jpetto: A good question. If we end up keeping FxA in the firstrun experience, we'll definitely need a better way to test locally. There is an option to run the entire FxA stack locally, but it's arguably harder to get running that bedrock (and definitely more volatile), which makes it rather prohibitive for us bedrock folk.

jpetto:
Yes, agree that tests would be good if FxA remains a part of firstrun.

Added an RTL fix for the logo.

alexgibson:
r+ :roller_coaster:

PR #3170, opened 2015-07-31
Update home-promo.less

amjadm61:
It would be better to change the background color of the firefox developer edition tile into black. Because, most of the developers would like to see their working window with a black background.

### DIFF
--- a/media/css/mozorg/home/home-promo.less
+++ b/media/css/mozorg/home/home-promo.less
@@ -951,7 +951,7 @@ html[lang|="en"] {
 /****************************************************************************/
 // Firefox Developer Edition
 .promo-large-portrait.firefox-developer {
-    background: #005da5; // dark blue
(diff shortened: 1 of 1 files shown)


pmclanahan:
I believe the former blue color was set to go with the dev edition logo. Also I know a lot of devs who use the light-theme in dev edition :) Thanks for the suggestion though.

@craigcook opinions?

craigcook:
Pure black is a bit too strong for the home page, especially for one of the large promos. We try to use slightly muted colors that work well together while still being overall bold and colorful. Giving any one promo an extremely bright or extremely dark color would make it too dominant and would overpower the rest of the page. 

I should also mention that the dev edition promo was changed to gray recently because we wanted to use a similar dark blue for the MDN 10th anniversary promo. So even though pure black might be too much, maybe a "light black" is close enough.

PR #3171, opened 2015-07-31
Clean up/optimize Fx desktop page templates.

jpetto:
Noticed while reviewing PR #3125 that `site_header` in `desktop-base.html` isn't needed anymore (due to family nav).

Didn't think a bug was warranted.


schalkneethling:
r+ :clap:

PR #3172, opened 2015-07-31
Fixed white space and word overwrap via mobile

yglazko:
Two-line fix, release notes no longer overflow and force the page to create a white space as they once did.


media/css/firefox/releasenotes.less
```diff
@@ -447,6 +447,8 @@ article {
     .section-items {
         > li {
             padding-left: 0;
+            word-wrap:break-word;
+            width:50%;
```
jpetto: I think the `word-wrap` property is all we need here? Setting the `width` to 50% pushes text a bit close to the right edge, and results in some overly skinny text on `<li>` elements without an icon - like the "Reference" line on the Android release notes:

screen shot 2015-08-03 at 10 09 01 am

craigcook: I'd take it further and suggest removing the left padding and setting the tags to <code>display:block</code>, <code>position:static</code>, and <code>width:auto</code> so they just stack up. Seems like a simpler fix and will be much more readable on small screens without squeezing the text into columns.

releasenotes

yglazko:
Followed Craig's suggestion above, added a new commit.

craigcook:
r+! :watermelon: 

Can you squash these two commits and reference the bug number in the message? If your commit message begins with "Fix bug 1189856" our bot will automatically resolve the bug when it merges to master.

craigcook:
Oops, the rebase went a bit pear-shaped. Closing this PR in favor of #3185 from a clean branch.

PR #3173, opened 2015-08-03
Remove dead code from browser-tour.js and update tests

alexgibson:
Paying off some technical debt by removing dead code from old tours. Also tidying up the tests a little. Didn't think this warranted a bug (unless someone thinks otherwise!).


schalkneethling:
All good r+ :moneybag:

PR #3175, opened 2015-08-03
[bug 1171928] Fix static tabzilla on Firefox 36 tour pages

alexgibson:
* `/firefox/36.0/whatsnew/`
* `/firefox/36.0/firstrun/`
* `/firefox/36.0/tour/`


schalkneethling:
There is an opacity set on tabzilla on these pages:

```
#tabzilla {
    opacity: 0.4;
}
```

It is set in common.css Guessing this is not intentional?

alexgibson:
As far as I can recall, this was part of the design for these pages. They wanted to give more visual balance to other elements on the page.

schalkneethling:
In that case r+ :milky_way:

PR #3176, opened 2015-08-03
[fix bug 1190378] No CSS on thunderbird system-requirements pages

alexgibson:
Haven't fully tested this change locally yet, but should fix things.


alexgibson:
Confirmed locally this fix does what it says on the tin :+1:

PR #3177, opened 2015-08-03
Update mozilla-pager.js tests

alexgibson:
- Tests we're not stopping auto-rotate after execution due to spy function not calling through. This wasn't causing failures, but could potentially cause confusion/conflicts in other tests.


schalkneethling:
r+ :cake:

PR #3183, opened 2015-08-05
Bug 1150180 - update "learn more about tiles" page

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1150180


bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +46,99 @@
...
 {% if LANG in ['en-US'] %}
-When you open a new tab in Firefox, you see several rectangles that link to different websites.
-Others call them squares, windows or bookmarks; we call them Tiles.
+  <h2>{{ _('Personalize your New Tab experience') }}</h2>
+  <p>{{ _('You are in complete control of your New Tab page. Customize your experience by rearranging, pinning or removing sites, and set up New Tab just the way you want it.') }}</p>
+  <ol>
+    <li>{{ _('Drag a site to exactly where you want it.') }}
+      <img src="{{ static('img/firefox/tiles/drag-site.png') }}" alt="{{ _('Screenshot') }}">
+    </li>
+    <li>{{ _('Once you\'ve found the perfect spot for your site, pin it to make it stay.') }}</li>
```
schalkneethling: Just use a `’` then you do not need to escape.

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +46,99 @@
...
+    <li>{{ _('Make a site disapear by deleting it from your New Tab page.') }}</li>
```
schalkneethling: 'disapear' > 'disappear'?

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +46,99 @@
...
+  </ol>
+  <img src="{{ static('img/firefox/tiles/pin-site.png') }}" alt="{{ _('Screenshot') }}">
+  <p>
+    <a href="https://support.mozilla.org/en-US/kb/new-tab-page-show-hide-and-customize-top-sites">{{ _('Learn more about customizing New Tab') }}</a>
```
schalkneethling: I know this is `en-US` only, but I reckon we should not hard-code the `en-US` into the URL

flodolo:
I'm temporarily adding a "Do Not Merge" tag because it's far for clear what we need to do from a l10n point of view (discussion ongoing via email). If we need to localize this page, we should also remove all the conditionals in the content.

Given that we also need this page for Aug 10, we might just need to merge this anyway and fix it later. In that case feel free to remove the tag and merge away.

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +46,99 @@
...
+  </p>
 {% else %}
-{{ _('When you open a new tab in Firefox, you see several rectangles that link to different websites.') }}
-{{ _('Some people call them squares, windows or bookmarks; we call them Tiles.') }}
+  <h2>{{ _('What are Tiles?') }}</h2>
+  <p>{{ _('When you open a new tab in Firefox, you see several rectangles that link to different websites.') }}</p>
+  <p>{{ _('Some people call them squares, windows or bookmarks; we call them Tiles.') }}</p>
```
schalkneethling: I do not see this content in the new copy document. Should this remain? Perhaps targeted at a different locale? Perhaps it is just the diff that makes this seem out of place but, just wanted to double check.

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +46,99 @@
...
-
-<h3>Easy to control</h3>
-
-<p>
-Tiles are easily pinned, moved around or removed using simple drag-and-drop and close interfaces.
-If you do not want to see any Tiles, you can deactivate them completely in two clicks through the new tab gear control.
-</p>
-
+  <h2>{{ _('If you\'re getting started') }}</h2>
+  <p>{{ _('If you\'re using Firefox for the first time, the New Tab page is already populated with useful and interesting sites and content. This includes both referrals to sponsored sites from commercial partners, labeled as "Sponsored," and content we consider important to our mission of advancing the public benefit of the Internet.') }}</p>
```
schalkneethling: Here and above, you can just use `’`

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +46,99 @@
...
-</p>
-<p>
-Mozilla respects your privacy and any data collection by Mozilla via Tiles will be done in accordance with our <a href="{{ url('privacy.notices.firefox') }}">Privacy Notice</a>.
-The data we collect is to help us deliver the most interesting and relevant content to our users.
-For more on what data Mozilla collects via Tiles, see below.
+  <h2>{{ _('Personalization with your privacy in mind') }}</h2>
+  <p>{{ _('Mozilla promotes a transparent exchange between users and content providers to make the Web enjoyable and sustainable.') }}</p>
+  <p>{{ _('If suggestions are enabled, Firefox will make suggestions to you based on categories identified in your browsing history and will let you know every time why are you seeing specific suggestions.') }}</p>
+  <p>{{ _('In delivering Suggested sites, Firefox does not share your personal data, nor place cookies on your computer. Content providers will only receive aggregated information about interactions users have with their tiles, but individual information is never shared or sold.') }}</p>
+  <p>{{ _('You remain in control of your browsing history, and you can disable suggestions whenever your chose.') }}</p>
```
flodolo: choose?

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +46,99 @@
...
-{{ _('We just know the total number of views and clicks an individual Tile receives. This helps us deliver the most interesting and relevant content.') }}
-{{ _('You can deactivate click and impression tracking by clicking on the gear icon on the top right corner of the new tab page, and selecting Classic to show only unenhanced History Tiles or Blank mode.') }}
+  <h2>{{ _('Respects your privacy') }}</h2>
+  <p>
+    {{ _('As with all Mozilla products, Tiles provide users with extensive control over their personal data.') }}    
+    {{ _('Data collection is limited to the aggregate number of impressions, clicks and Tile placement.') }}   
+    {{ _('Nothing is ever personally identifiable. In other words, we don’t know who clicks on what.') }}    
+    {{ _('We just know the total number of views and clicks an individual Tile receives. This helps us deliver the most interesting and relevant content.') }}   
+    {{ _('You can deactivate click and impression tracking by clicking on the gear icon on the top right corner of the new tab page, and selecting Classic to show only unenhanced History Tiles or Blank mode.') }}
```
flodolo: `… the new tab page, and selecting Classic to show only unenhanced History Tiles or Blank mode.`
I think this should be "New Tab" for consistency with the rest of the document, also wondering if there should be a comma before 'or' for clarity.

bedrock/firefox/templates/firefox/tiles.html
schalkneethling: I know it is like this in the copy but, I am pretty sure "disable suggestions whenever your chose." should be "disable suggestions whenever *you choose*."

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +46,99 @@
...
-{{ _('We just know the total number of views and clicks an individual Tile receives. This helps us deliver the most interesting and relevant content.') }}
-{{ _('You can deactivate click and impression tracking by clicking on the gear icon on the top right corner of the new tab page, and selecting Classic to show only unenhanced History Tiles or Blank mode.') }}
+  <h2>{{ _('Respects your privacy') }}</h2>
+  <p>
+    {{ _('As with all Mozilla products, Tiles provide users with extensive control over their personal data.') }}    
+    {{ _('Data collection is limited to the aggregate number of impressions, clicks and Tile placement.') }}   
+    {{ _('Nothing is ever personally identifiable. In other words, we don’t know who clicks on what.') }}    
+    {{ _('We just know the total number of views and clicks an individual Tile receives. This helps us deliver the most interesting and relevant content.') }}   
+    {{ _('You can deactivate click and impression tracking by clicking on the gear icon on the top right corner of the new tab page, and selecting Classic to show only unenhanced History Tiles or Blank mode.') }}   
+  </p>
```
schalkneethling: There are other instances like the above, I wonder whether we cannot just wrap the content inside a `{% trans %}` block, instead of each line separately? @flodolo

flodolo: We tend to separate strings to avoid invalidating huge paragraphs when a single comma changes.

media/css/firefox/tiles.less
```diff
@@ -4,10 +4,136 @@
...
+        }
+    }
+}
+
+.main-column {
+    width: 580px;
+
+    .intro {
+        font-size: @baseLine;
+        font-size: 1.2rem;
```
schalkneethling: You should use the `.font-size()` mixin here, for example:

```
.font-size(@baseLine);
```

Then you do not need the two separate `font-size` entries. @see https://github.com/mozilla/bedrock/blob/master/media/css/sandstone/lib.less#L24

media/css/firefox/tiles.less
```diff
@@ -4,10 +4,136 @@
...
+        font-style: italic;
+        font-weight: 100;
```
schalkneethling: You can remove the `font-weight` and then do:

```
  font-family: "Open Sans Light";
  font-style: italic;
```

media/css/firefox/tiles.less
```diff
@@ -4,10 +4,136 @@
...
+    }
+
+
+    section {
+        padding-top: 2.5em;
+        border-top: solid 3px @borderColor;
+
+        h4 {
+            font-weight: bold;
+            font-style: italic;
```
schalkneethling: This is a bit of a tricky one. Because you are inheriting `Open Sans Light` from `sandstone.less` here, what you are actually getting in the browser is faux bold but, we are also not currently downloading `Open Sans Bold Italic` so, the italic will also be faux. The italic seems the lesser of the two evils ;) though so, I would suggest adding:

```
font-family: "Open Sans";
```
If you feel that the faux italic does not look great, you can add the required font to this page: https://github.com/mozilla/bedrock/blob/master/media/css/sandstone/fonts.less#L66

Perhaps @alexgibson has some thoughts on this as well?

media/css/firefox/tiles.less
```diff
@@ -4,10 +4,136 @@
...
+                float: left;
+                width: 46%;
+                margin-right: 20px;
+            }
+
+            li:nth-child(3) {
+                padding-left: .5em;
+                box-sizing: border-box;
+            }
+        }
```
schalkneethling: This is a tricky one but, I think we can clean this up a little so, couple of suggestions:

```
ol {
  font-family: "Open Sans";
  font-style: italic;
  font-weight: bold;
}
```
Let the `ol` just do the numbering and, seeing you are going to need `bold/italic` here again, it might very well be worth it to add the font to the page.

```
li:before {
    content: ") ";
}

With the `ol` doing the numbering, we only need to add the parenthesis. I am honestly not even sure if we should go that far, I know it is in the design but, I do not know whether it would be a problem to just have `1.` as apposed to '1.)'

I would also suggest using the images as a background image using the `:after' pseudo selector. Getting items two and three to line up is going to be a huge headache. You just need to set the container size so, that for items 2 and 3, you are cropping out the part of the image that is not relevant.

If you do that, you can then also get rid of the `li:nth-child(3)` rule.

File: bedrock/firefox/templates/firefox/tiles.html (hunk 8 above)
Comment (schalkneethling) [inline bedrock/firefox/templates/firefox/tiles.html]: Ah, good to know, thanks!

**Comment by schalkneethling** [comment]
I would also suggest resizing some of the images that are much larger than needed, except for images that are meant for hidpi. Also, did you run them through something like https://tinypng.com/ or http://pmt.sourceforge.net/pngcrush/ to crunch down the file size?

**Comment by schalkneethling** [comment]
The layout also does not quite match the design here https://bug1150180.bmoattachments.org/attachment.cgi?id=8640617&t=snKOkYdzLgjedt6QSLFPa8

screen shot 2015-08-06 at 13 23 21

The closest I could get was:

screen shot 2015-08-06 at 13 22 45

Not a 100% match either but, a bit closer. I did this by making the following changes:

```
.main-column {
    width: 800px;
}
.sidebar {
    float: right;
    margin: -200px -190px 0 10px;
    width: 280px;
}
```
Note however that I did this in dev tools so, did not do any cross browser testing meaning, your mileage may vary :) (cough IE8)

**Comment by schalkneethling** [comment]
That's all for now, thanks @malena

File: bedrock/firefox/templates/firefox/tiles.html (hunk 1 above)
Comment (malena) [inline bedrock/firefox/templates/firefox/tiles.html]: Thanks changing all instances!

File: bedrock/firefox/templates/firefox/tiles.html (hunk 2 above)
Comment (malena) [inline bedrock/firefox/templates/firefox/tiles.html]: Oooh I caught a few typos but I guess I missed a few. Good eye!

File: bedrock/firefox/templates/firefox/tiles.html (hunk 4 above)
Comment (malena) [inline bedrock/firefox/templates/firefox/tiles.html]: Yeah this is the old content that just seems out of place.

File: bedrock/firefox/templates/firefox/tiles.html (hunk 6 above)
Comment (malena) [inline bedrock/firefox/templates/firefox/tiles.html]: changed to "disable suggestions whenever you choose."

File: bedrock/firefox/templates/firefox/tiles.html (hunk 7 above)
Comment (malena) [inline bedrock/firefox/templates/firefox/tiles.html]: This is actually the old copy that is now being removed :)

File: bedrock/firefox/templates/firefox/tiles.html (hunk 8 above)
Comment (malena) [inline bedrock/firefox/templates/firefox/tiles.html]: Ahh nice to know @flodolo. I can adjust the new copy to do the same.

File: media/css/firefox/tiles.less (hunk 9 above)
Comment (malena) [inline media/css/firefox/tiles.less]: YES! Thanks.

**Comment by malena** [comment]
@schalkneethling for the comment regarding layout, I can see what you mean by the design not matching. I am also assuming you are viewing it in a larger screen. Wondering if changing the line breaks as you can see in your fixed screenshot (with the paragraph under the "Personalize your New Tab experience") is acceptable though? I would need to confirm with the design team if they would rather have the copy line breaks match the design or if they are okay with making it wider for larger screens so that the image on the right cuts off like in the design. Thanks for pointing this out!

File: media/css/firefox/tiles.less (hunk 9 above)
Comment (malena) [inline media/css/firefox/tiles.less]: Thanks! Still getting used to all the mixins we have. Updating.

**Comment by schalkneethling** [comment]
@malena I totally agree, best to loop in the design folks to get clarity.

File: media/css/firefox/tiles.less (hunk 12 above)
Comment (malena) [inline media/css/firefox/tiles.less]: As discussed splitting images into two so it remains meaningful. :)

File: media/css/firefox/tiles.less (hunk 11 above)
Comment (malena) [inline media/css/firefox/tiles.less]: @schalkneethling Great! I didn't even realize this! Will add font-family: "Open Sans"; Yes I agree italic is lesser of two evils. I don't think it's necessary to import required font as it looks great now that it's bold.

**Comment by malena** [comment]
@schalkneethling Confirmed with creative and the copy should maintain the line breaks so it's okay that the image doesn't bleed out on the right side. So no changes would be needed :)

**Comment by malena** [comment]
Keeping Do Not Merge Label for now since this needs to go live on August 10th.

File: bedrock/firefox/templates/firefox/tiles.html [hunk 13]
```diff
@@ -45,197 +41,74 @@
...
-        No data is collected when a user chooses to opt out of the enhanced Tiles experience.
-      {% endtrans %}
-      {% endif %}
-    </p>
-  </div>
+<section class="tls-started">
+  <h2>{{ _('If you’re getting started') }}</h2>
+  <p>
+    {{ _('If you’re using Firefox for the first time, the New Tab page is already populated with useful and interesting sites and content.') }} 
+    {{ _('This includes both referrals to sponsored sites from commercial partners, labeled as "Sponsored," and content we consider important to our mission of advancing the public benefit of the Internet.') }}
```
TheoChevalier: Could we have “smart quotes” here for consistency please? I'd prefer we make the change now rather than breaking the string freeze for that minor change.

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +41,74 @@
...
-      <li><a href="https://blog.mozilla.org/netpolicy/2014/08/21/trust-should-be-the-currency/">
-        {{ _('Trust should be the currency') }}
-      </a></li>
-      {% endif %}
-    </ul>
-  </div>
+<section class="tls-privacy">
+  <h2>{{ _('Personalization with your privacy in mind') }}</h2>
+  <p>{{ _('Mozilla promotes a transparent exchange between users and content providers to make the Web enjoyable and sustainable.') }}</p>
+  <p>{{ _('If suggestions are enabled, Firefox will make suggestions to you based on categories identified in your browsing history and will let you know every time why are you seeing specific suggestions.') }}</p>
```
TheoChevalier: I'm French, so I might be wrong, but isn't it supposed to be "every time why you are seeing"?

media/css/firefox/tiles.less
```diff
@@ -4,10 +4,135 @@
...
+            width: 100%;
+        }
+    }
 }
+
+.main-column {
+    width: 580px;
+
+    .intro {
+        .font-size(@baseLine);
```
alexgibson: Generally, `@baseLine` represents the baseline vertical rhythm in sandstone. Does no harm using it for a font size declaration, but using `.font-size(20px);` would be more appropriate.

media/css/firefox/tiles.less
```diff
@@ -4,10 +4,135 @@
...
+        font-family: "Open Sans Light";
```
alexgibson: If you use `.open-sans-light;` here instead of directly using `font-family`, you'll get all the default fallback fonts as well as locale specific ones.

media/css/firefox/tiles.less
```diff
@@ -4,10 +4,135 @@
...
+        padding-bottom: 2em;
+    }
+
+
+    section {
+        padding-top: 2.5em;
+        border-top: solid 3px @borderColor;
+
+        h4 {
+            font-family: "Open Sans";
```
alexgibson: Same here use `open-sans;`

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +41,74 @@
...
+    </li>
+    <li>{{ _('Once you’ve found the perfect spot for your site, pin it to make it stay.') }}
+      <img src="{{ static('img/firefox/tiles/pin-site.png') }}" alt="{{ _('Screenshot') }}">
+    </li>
+    <li>{{ _('Make a site disappear by deleting it from your New Tab page.') }}
+      <img src="{{ static('img/firefox/tiles/delete-site.png') }}" alt="{{ _('Screenshot') }}">
+    </li>
+  </ol>
+  <p>
+    {{ _('<a href="%(url)s">Learn more about customizing New Tab</a>')|format(url='https://support.mozilla.org/en-US/kb/new-tab-page-show-hide-and-customize-top-sites') }}
```
alexgibson: Any reason to wrap the whole link in an l10n tag here and not just the link text?

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +41,74 @@
...
+  </p>
+
+  <h4>{{ _('Top sites') }}</h4>
+  <p>{{ _('New Tab will show the “top sites” you visited most recently and frequently, so you can access them quickly.') }}</p>
+  <img src="{{ static('img/firefox/tiles/top-sites.png') }}" alt="{{ _('Screenshot') }}">
+
+  <h4>{{ _('Suggested sites') }}</h4>
+  <p>
+    {{ _('See sites and content suggested to you based on categories that Firefox identifies in your browsing history.') }} 
+    {{ _('They will be clearly labeled in your new tab as “Suggested,” or when the suggestion is from a commercial partner, they will be labeled as “Sponsored,” and will indicate the identified interest category.') }}
```
flodolo: new tab -> New Tab?

media/css/firefox/tiles.less
schalkneethling: But of course! I completely neglected to mention it, thanks @alexgibson

bedrock/firefox/templates/firefox/tiles.html
malena: @alexgibson  Well I took the example from the docs: http://bedrock.readthedocs.org/en/latest/l10n.html  Maybe we should update the docs as well?

malena:
Just got an email about new copy updates. Will be updating again.

bedrock/firefox/templates/firefox/tiles.html
flodolo: Yeah, that doc doesn't have great examples: while this approach is needed if you have partial links `_('<a href="%(url)s">Learn more</a> about customizing New Tab')`, when the entire string is a URL it's cleaner to avoid HTML tags in the string `<a href="https://support.mozilla.org/en-US/kb/new-tab-page-show-hide-and-customize-top-sites">_('Learn more about customizing New Tab')</a>`

alexgibson: Yes, that example is a bit confusing in this context. I assume it's purpose was to highlight that it's important not to hard-code URL's in the string. We're not doing that here anyhoo which is good, but seeing as the link is separated from any other localizable text, we should be fine not to need a substitution and just wrap the link text instead. Hope that makes sense?

alexgibson: Also, we should probably omit `/en-US/` from the link `href`, and let SUMO handle the locale negotiation.

malena: hmm I've tried a few variations of what you are mentioning and getting errors. 
`<a href="%(url)s">{{ _('Learn more about customizing New Tab')|format(url='https://support.mozilla.org/kb/new-tab-page-show-hide-and-customize-top-sites') }}</a>`
Not sure what the syntax for something like this would be? Above I get page not found.

flodolo: This? And yes, en-US in the link should be removed
```
<a href="https://support.mozilla.org/kb/new-tab-page-show-hide-and-customize-top-sites"> {{ _('Learn more about customizing New Tab') }}</a>
```

alexgibson: <a href="https://support.mozilla.org/kb/new-tab-page-show-hide-and-customize-top-sites">{{ _('Learn more about customizing New Tab') }}</a>

This should do it?

malena: oooh ic. I thought I had to separate the url. Thinking it's more complicated than it is. Thanks for clearing it up!

bedrock/firefox/templates/firefox/tiles.html
malena: Yes! New copy had same mistake and I pointed it out so it's being fixed :)

media/css/firefox/tiles.less
malena: Ahh okay good to know.

media/css/firefox/tiles.less
malena: Oooh learning lots. Thanks guys updating.

bedrock/firefox/templates/firefox/tiles.html
malena: @TheoChevalier sorry not sure what you mean by smart quotes? Maybe a dumb question, but do you mean I should be wrapping the copy in double quotes for this line?

malena: got it!

malena:
Okay copy has been updated and comments have been addressed. @schalkneethling @alexgibson Thanks!

TheoChevalier:
Thanks for the update @malena ;) As flod said, the page is now being localized. If the PR is r+'ed, can it be merged today? This way localizers can see the page on dev server and start doing QA.

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +41,80 @@
...
-        Aggregate data from the New Tab page includes impressions, clicks and tile placement.
-      {% endtrans %}
-      {% endif %}
-    </p>
-  </div>
+<section class="tls-discover">
+  <h2>{{ _('Discover new sites and content on the Web') }}</h2>
+  <p>
+    {{ _('If enabled, Firefox displays website links as thumbnails (miniaturized images) or logos on your New Tab page.') }}
+    {{ _('These are called Tiles.') }}
```
flodolo: Not sure if it was like this from the beginning, but we went a step too far in splitting sentences. Can we have one single sentence: `If enabled, Firefox displays website links as thumbnails (miniaturized images) or logos on your New Tab page. These are called Tiles.`?

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +41,80 @@
...
+  </p>
+  <p>
+    {{ _('Tiles save you time by giving you quick access to your frequently visited sites, as well as showing new sites and content that are similar to sites you already love.') }} 
+    {{ _('Here are the types of Tiles you may see:') }}
+  </p>
```
flodolo: Same issue here: `Tiles save you time by giving you quick access to your frequently visited sites, as well as showing new sites and content that are similar to sites you already love. Here are the types of Tiles you may see:`

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -45,197 +41,74 @@
...
-  </div>
+<section class="tls-personalize">
+  <h2>{{ _('Personalize your New Tab experience') }}</h2>
+  <p>
+    {{ _('You are in complete control of your New Tab page.') }}
+    {{_('Customize your experience by rearranging, pinning or removing sites, and set up New Tab just the way you want it.') }}
+  </p>
+  <ol>
+    <li>{{ _('Drag a site to exactly where you want it.') }}
+      <img src="{{ static('img/firefox/tiles/drag-site.png') }}" alt="{{ _('Screenshot') }}">
```
craigcook: If this alt text can't be descriptive of what the image shows, I think an empty alt would be better than just "screenshot." I think the captions alone are good enough.

craigcook:
The text that's in bold italics is really just Open Sans Italic with fake bolding so it looks a bit chunky. You can add a real Open Sans Bold Italic to the style sheet and it'll render nicer (though at the expense of downloading yet another font). 
https://github.com/mozilla/bedrock/blob/master/media/css/sandstone/fonts.less#L66-L73

bedrock/firefox/templates/firefox/tiles.html
```diff
@@ -245,15 +118,9 @@ <h3 data-accordion-role="tab">{{ _('Where can I learn more?') }}</h3>
 
 {% block dashboard_sidebar_contents %}
 
-<h4>{{ _('Tiles Preferences') }}</h4>
+<img class="tab" src="{{ static('img/firefox/tiles/new-tab-enhanced.png') }}" alt="{{ _('Screenshot') }}">
```
craigcook: These last two images should also have empty alts.

craigcook:
r+ :cactus:

PR #3184, opened 2015-08-05
Bug 1148187 Redirect /access pages to archive

sgarrity:


etc/httpd/global.conf
```diff
@@ -569,6 +569,13 @@ RewriteRule ^/access/windows/at-apis.html$ https://developer.mozilla.org/en-US/d
 RewriteRule ^/access/windows/msaa-server.html$ https://developer.mozilla.org/en-US/docs/Web/Accessibility/Implementing_MSAA_server [L,R=301]
 RewriteRule ^/access/windows/zoomtext.html$ https://developer.mozilla.org/en-US/docs/Mozilla/Accessibility/ZoomText [L,R=301]
 
+# bug 1148187
+RewriteRule ^/access/keyboard(.*)$ http://website-archive.mozilla.org/www.mozilla.org/access/access/keyboard$1 [L,R=301]
+RewriteRule ^/access/qa(.*)$ http://website-archive.mozilla.org/www.mozilla.org/access/access/qa$1 [L,R=301]
+RewriteRule ^/access/slideshow(.*)$ http://website-archive.mozilla.org/www.mozilla.org/access/access/slideshow$1 [L,R=301]
+RewriteRule ^/access/type-ahead(.*)$ http://website-archive.mozilla.org/www.mozilla.org/access/access/type-ahead$1 [L,R=301]
+RewriteRule ^/access/unix(.*)$ http://website-archive.mozilla.org/www.mozilla.org/access/access/unix$1 [L,R=301]
```
jgmize: Since these are the last of the /access urls, and they come below the slew of /access urls that were redirected to MDN, couldn't we just have a catch-all redirect like the one below?

`RewriteRule ^/access/(.*) http://website-archive.mozilla.org/www.mozilla.org/access/access/$1 [L,R=301]`

Ideally we'd actually make /access a /b PT here and do the redirects in python instead of adding one more thing that we need to port, but the `redirect` helper function in utils doesn't support this kind of pattern matching yet, and I don't want to block this PR on that.

sgarrity: Good idea - will do.

sgarrity: Updated. @jgmize The trailing `$` isn't need on the first argument, right?

jgmize:
> Updated. @jgmize The trailing $ isn't need on the first argument, right?

Right, the `(.*)` at the end will capture anything regardless. The reason you see $ at the end of so many regexes is to avoid partial matches: `^/foo` will match `/foobar` but `^/foo$` won't.

PR #3186, opened 2015-08-05
[fix bug 1168440] Remove Spring pixel

ckprice:
@craigcook r? Looks like discussions are still ongoing, but for now it's removed.


craigcook:
This is just a reversion of 24a349beb681957ea95397d305f90c0d73dc7963 so not too much to review. r+

PR #3188, opened 2015-08-06
[fix bug 1189821] Replace & remove old share widget.

jpetto:
- Improve firefox/desktop/tips/ RTL
- Move firefox/new/ JS to footer


schalkneethling:
This lgtm. Are we waiting for anything in terms of approval? Seems like a r+ to me @jpetto ?

jpetto:
As far as I know, we're good to go. Gareth confirmed moving the JS to the footer on `/firefox/new/` is okay in terms of GA, so as long as it's working (I tested back to IE 7), we should be merge ready.

schalkneethling:
r+ :camel:

PR #3190, opened 2015-08-06
Add a link to Firefox OS Security Advisories

kyoshino:
https://github.com/mozilla/foundation-security-advisories/pull/5 added Security Advisories for Firefox OS. We need a link to the index page. No bug. Tested locally.


jgmize:
Good catch, thanks @kyoshino

PR #3192, opened 2015-08-07
Bug 1191923 - remove waffle switch for Pocket page

craigcook:
The Firefox OS TV page, too. We no longer need to waffle these pages. Eliminating some debt.

https://bugzilla.mozilla.org/show_bug.cgi?id=1191923


alexgibson:
r+ :aerial_tramway:

PR #3193, opened 2015-08-07
[fix bug 1188852] Fx family nav tabzilla static update.

jpetto:
As static tabzilla only has a white tab, updated mobile style of family nav from white bg to blue bg. Tycho gave it a :+1:.


media/css/firefox/family-nav.less
```diff
@@ -49,45 +48,16 @@
...
-        background: transparent;
-        position: absolute;
-        left: 28px;
-        top: 0;
-        z-index: 9;
-    }
 }
 
-// change top border color to match inverted tabzilla image
+// change tabzilla color for tablet and below
```
schalkneethling: Tiny nit: "change Tabzilla wordmark color for tablet and below"?

schalkneethling: disregard, all is good here ;p

schalkneethling:
r+ :cake:

PR #3194, opened 2015-08-07
Fix Bug 1192354 - Remove Firefox 31 ESR detection

kyoshino:
Do not merge before August 11. @alexgibson r?

### DIFF
--- a/media/js/base/global.js
+++ b/media/js/base/global.js
@@ -123,31 +123,27 @@ function isFirefoxMobile(userAgent) {
     return /Mobile|Tablet|Fennec/.test(ua);
 }
 
-// Detect Firefox 31 ESR using navigator.buildID. 20140716183446 is the *non-ESR*
-// build ID that can be found at https://wiki.mozilla.org/Releases/Firefox_31/Test_Plan
-// Since this is a simple and only way to detect ESR, there would be some false
-// positives if the browser is a pre-release (Nightly, Aurora, Beta), semi-official
-// (e.g. Ubuntu build by Canonical) or unofficial version of Firefox 31, because
-// those builds have a different build ID from the official non-ESR build.
-function isFirefox31ESR(userAgent, buildID) {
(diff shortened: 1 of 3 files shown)


media/js/base/global.js
```diff
@@ -123,31 +123,24 @@ function isFirefoxMobile(userAgent) {
...
-// Detect Firefox 38 ESR by the *non-ESR* build IDs.
-// 38.0: 20150508094354, 38.0.1: 20150513174244, 38.0.5: 20150525141253
-// https://wiki.mozilla.org/Releases/Firefox_38/Test_Plan
-function isFirefox38ESR(userAgent, buildID) {
-    userAgent = userAgent || navigator.userAgent;
-    buildID = buildID || navigator.buildID;
+    var ESRs = {
+        // 38.0, 38.0.1, 38.0.5 and 38.0.6
+        // https://wiki.mozilla.org/Releases/Firefox_38/Test_Plan
+        38: ['20150508094354', '20150513174244', '20150525141253', '20150605094246'],
```
alexgibson: Remove trailing comma

media/js/base/global.js
```diff
@@ -123,31 +123,24 @@ function isFirefoxMobile(userAgent) {
...
+    };
 
-    return !isFirefoxMobile(userAgent) &&
-        getFirefoxMasterVersion(userAgent) === 38 && buildID &&
-        ['20150508094354', '20150513174244', '20150525141253'].indexOf(buildID) === -1;
+    return !isFirefoxMobile(userAgent) && buildID &&
+        version in ESRs && ESRs[version].indexOf(buildID) === -1;
```
alexgibson: While this works fine in the setting it's being used here, should this function be run in isolation it will return `undefined` in older browsers. Is it worth being a bit more diligent here?

kyoshino: Yeah, this function seems expected to be called after checking `isFirefox`. I'll add `!isFirefox(userAgent)` so it should always work.

alexgibson:
Looks great, thanks @kyoshino! r+ :boat:

PR #3115, opened 2015-07-10
[fix bug 1178217] Updates to Hello FTU for Firefox 40

alexgibson:

PR #3198, opened 2015-08-10
[fix bug 1190716] Update Hello FTU for Firefox 40 to interactive version

alexgibson:


alexgibson:
Ping @flodolo - as per comments in [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1190716), I believe we now have the final two new strings needed for this PR.

flodolo:
Looks good to me, even if it's getting harder to follow all the code changes :-)

We have 2 new strings behind `hello_ftu_firefox_40_interactive`, it's probably easier to check after merge and push to prod how many are actually unused.

alexgibson:
> We have 2 new strings behind hello_ftu_firefox_40_interactive, it's probably easier to check after merge and push to prod how many are actually unused.

Sounds good - I'll file a follow-up bug to this one after this goes out, and we'll see if we can clean up the tags in this page :+1:

alexgibson:
Ok I think this should be good for review.

@jpetto r? (I think you've done the Hello dance before, so you're probably the most familiar?)

Testing notes:
- Please test using Firefox 40
- Make sure to have [UITour working locally](http://bedrock.readthedocs.org/en/latest/uitour.html#local-development)
- In about:config, set "loop.gettingStarted.url" to e.g. "http://localhost:8000/%LOCALE%/firefox/%VERSION%/hello/start
- Access the tour by clicking the Hello button, then click the gear icon and select "Tour"

There is no new functionality here other than a few updated strings and bits in the template. It is basically the same as what we shipped prior to PR #3115, but with the code refactored for new tests.

media/js/firefox/hello/hello-ftu.js
```diff
@@ -0,0 +1,471 @@
...
+     * @param callback (function)
+     */
+    HelloFTU.showHelloPanel = function(callback) {
+        // Only open the info panel if tab is visible
+        if (HelloFTU.documentHidden || document.hidden) {
+            return;
+        }
+        // Make sure loop icon is available before opening Hello panel (bug 1111828)
+        Mozilla.UITour.getConfiguration('availableTargets', function (config) {
+            if (config.targets && $.inArray('loop', config.targets) !== -1) {
```
jpetto: Is this operation expensive? Is it feasible to cache the result of this check so it only has to be executed once? I suppose the user *could* move the icon between page load and initializing the tour...

jpetto:
Tour worked well in Fx 40. One question in `hello-ftu.js`, but not a blocker.

r+ from me, but looks like there's still [discussion going on in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1190716#c16).

media/js/firefox/hello/hello-ftu.js
alexgibson: Yeah, this gets called at multiple times in the tour depending on the state, so it's there in the event that the user could remove the Hello button at any time. It's only really costly in that it requires firing a custom event each time, but I think it's better to pay that cost here.

alexgibson:
Thanks @jpetto - holding off merging this until folks decide what to do in the bug. Should hopefully be resolved soon one way or another. Will ping here if it requires any code updates (@flodolo - really sorry if this ends up being another string change :cry:).

alexgibson:
Ok, we're now unblocked on merging this one. There's been one last string change (cc @flodolo or @pascalchevrel), but we're now good to expose these to localizers.

@jpetto - I added a new commit for the string change. If all looks good, I'll squash and we can finally get this one out the door.

jpetto:
String change looks good. Looks like we're ready to merge after a squash?

alexgibson:
Thanks @jpetto - squashed!

jpetto:
Waiting until all the iOS stuff is out the door and solid before pushing more changes. Plan to merge either late 9/3 or 9/4.

jpetto:
As iOS is delayed, and will be behind waffle switches, we should be safe to merge this now. :balloon:

PR #3199, opened 2015-08-10
Bug 1152501 - remove Patriot Act campaign promo

craigcook:
Also relates to bug 1192764. This promo used a waffle switch that we shouldn't need to maintain because the whole promo can be removed.


jgmize:
thanks @craigcook

PR #3200, opened 2015-08-10
[fix bug 1191522] Enable Optimizely on /firefox/desktop/ pages.

jpetto:
As [commented in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1191522#c4), one switch for all `/firefox/desktop/` pages should be sufficient.

Will need to add to [the wiki page](https://wiki.mozilla.org/Mozilla.org/Optimizely) after merge.


jgmize:
Don't we also need to remove the existing optimizely block from  `bedrock/firefox/templates/firefox/desktop/index.html` so that it doesn't override the base template?

jpetto:
Uhm, yes. Yes we do. I even checked for the existing one...but obviously not hard enough. :grimacing:

PR #3204, opened 2015-08-11
[fix bug 1150231] Update FxA iframe URL params.

jpetto:


jpetto:
New/updated params requested [in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1150231#c58).

craigcook:
r+ :alien:

PR #3212, opened 2015-08-17
Bug 1171369

SujaySKumar:
I would love it if anybody has any other idea of fixing this bug

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/history-details.html
+++ b/bedrock/mozorg/templates/mozorg/about/history-details.html
@@ -17,7 +17,7 @@
       {% trans coderush='http://clickmovement.org/coderush',
                sourcerelease='http://web.archive.org/web/20021001071727/wp.netscape.com/newsref/pr/newsrelease558.html'
       %}
-        The Mozilla project was <a href="{{ coderush }}">created in 1998</a>
+        The Mozilla project was created in 1998
         with the <a href="{{ sourcerelease }}">release of the Netscape browser
         suite source code</a>.
       {% endtrans %}
--- a/root_files/humans.txt
+++ b/root_files/humans.txt
@@ -87,6 +87,7 @@ Developers: lismanb
 Developers: kenrick95
 Developers: comzeradd
 Developers: yglazko
+Developers: SujaySKumar
 
 Localizers on SVN
 Localize: akilaa@wippies.fi
@@ -210,4 +211,3 @@ Localize: veeven@gmail.com
 Localize: w.benes@gmail.com
 Localize: witia@aviary.pl
 Localize: yoshino@mozilla-japan.org
-


bedrock/mozorg/templates/mozorg/about/history-details.html
```diff
@@ -17,7 +17,7 @@ <h1 class="title-shadow-box">{{ _('History of the Mozilla Project') }}</h1>
       {% trans coderush='http://clickmovement.org/coderush',
                sourcerelease='http://web.archive.org/web/20021001071727/wp.netscape.com/newsref/pr/newsrelease558.html'
       %}
-        The Mozilla project was <a href="{{ coderush }}">created in 1998</a>
+        The Mozilla project was created in 1998
```
TheoChevalier: This page is localized, this means you'll need to wrap this change around l10n tags so that the old translation is kept while the new string is being localized. See https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templates/mozorg/about/history.html#L103 as an example

jgmize: There's no need for a new string; this can be a simple change: code rush is available at https://air.mozilla.org/code-rush/

vendor-local/src/legal-docs
```diff
@@ -1 +1 @@
-Subproject commit 3c6ae9377f3e9ba59101dcd2d8f1cbfae2dcfc04
+Subproject commit 0e7351517172d17e56e9d044dc6e5dc320bf8811
```
jgmize: Please remove this submodule change from the PR.

schalkneethling:
Hey @SujaySKumar - Thanks for the PR, there are just a couple of things you need to do and we can merge this.

First we need to get rid of that subproject commit. To do that, checkout your branch and then type:

```
git reset HEAD~2
```
If you now run `git status`, you will see the two files you made changes to as well as the change in `vendor-local`. Now:

```
cd vendor-local/src/legal-docs/
git checkout master
cd /bedrock
```
If you now do a `git status` you will see only your changes \o/ Next, go ahead and make the additional change suggested by @jgmize above, with regards to simply pointing the link to the new home of coderush on air.mozilla

Now, you are ready to update this PR, just a few small things left. Let's rename your branch to more closely follow the standard we use for branch names:

```
git branch -m bug1171369-fix-non-working-coderush-link
```

Now:

```
git add --all
git commit -m 'Fix Bug 1171369, point coderush video link to air mozilla'
git push origin bug1171369-fix-non-working-coderush-link # change origin to be whatever you remote branch is called
```
The above does mean that you will have to open a new PR from the new branch but, that's is OK. When you have opened the new PR, just close this one and reference the new PR.

Let me know if you run into any problems.

SujaySKumar:
Hey @*schalkneethling <https://github.com/schalkneethling>* - I have
committed the changes. Thank you for your help.
Can you explain to me what the legal-docs are meant for? Why was the file
getting changed even though I wasn't changing it? And how did following
your instructions rectify it? (i.e checkout master inside legal-docs)
If it is too big to explain, you can point me to some articles and I will
read up on it.

schalkneethling:
@SujaySKumar It is a submodule that exists outside the project so, sometimes updates happen there and your repo was not synced with it.

When a submodule changes you need to run, from the root of your project:

```
git submodule update --init --recursive
```
For more on submodules see http://www.git-scm.com/book/en/v2/Git-Tools-Submodules

PR #3213, opened 2015-08-17
Links to teach.mozilla.org/tools/ instead of a 404 page

SujaySKumar:


schalkneethling:
@SujaySKumar is there a bug connected to this? Also, you have a submodule commit in your PR. Are you comfortable with Git to take care of it?

SujaySKumar:
The commit in the legal-docs? I can do that. But I have an issue. I have not changed the file in any way. But still, it is getting changed every time i try to commit something. Do you have any idea what the issue might be with the legal-docs?

SujaySKumar:
Oh Im so sorry...I forgot to add the bug ID to the commit.
It is this bug Bug 1195376 (https://bugzilla.mozilla.org/show_bug.cgi?id=1195376)

schalkneethling:
We have that submodule problem here again :) See my comments on your other PR but here, do:

```
git reset HEAD # as you only have 1 commit and not 2
```
as apposed to:

```
git reset HEAD~2
```

PR #3215, opened 2015-08-17
[fix bug 1195391] Remove video from Fx38+ firstrun.

jpetto:
- Enable display of FxA frame error (see [FxA issue #2934](https://github.com/mozilla/fxa-content-server/issues/2934))

### DIFF
--- a/media/js/firefox/australis/fx38_0_5/firstrun.js
+++ b/media/js/firefox/australis/fx38_0_5/firstrun.js
@@ -3,78 +3,6 @@
 
     var $document;
 
-    var hasVideo = $('#video').length > 0;
-    var $ctaSignup = $('#cta-signup');
-    var $ctaSignin = $('#cta-signin');
-    var $videoFrame = $('#video-frame');
-    var $videoTitle = $('#video-title');
-    var $video = $('#firstrun-video');
-
-    var videoOnLoad = false;
-
-    // if locale has video, do A/B test
-    if (hasVideo) {
-        videoOnLoad = (Math.random() >= 0.5);
-
-        // manual override to test videos
-        if (window.location.href.indexOf('v=') !== -1) {
-            var parts = window.location.href.split('v=');
-            var variation = parts[1];
-
-            if (variation === '1') {
-                videoOnLoad = false;
-            } else {
-                videoOnLoad = true;
-            }
-        }
-
-        window.dataLayer.push({
-            'event': 'dataLayer-initialized',
-            'page': {
-                'category': 'firstrun-38.0.5',
-                'variation': (videoOnLoad) ? '1' : '2'
-            }
-        });
-
-        $video.on('play', function() {
-            // GA track video play
-            window.dataLayer.push({
-                'event': 'firstrun-38.0.5-video',
-                'interaction': 'start',
-                'videoTitle': 'When its Personal Campaign Video'
-            });
-        }).on('ended', function() {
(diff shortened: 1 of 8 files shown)


jpetto:
We should have enough data on video interaction by 8/19, so this PR can merge either 8/19 or 8/20.

Video removal code (the bulk of this PR) can be reviewed at any time.

FxA frame error display code is being discussed in the [FxA issue](https://github.com/mozilla/fxa-content-server/issues/2934). Will remove the WIP label when FxA solution is agreed upon.

jpetto:
I believe all code in this PR is now ready for review.

If you haven't set up Fx for testing FxA on demo4, follow the [instructions posted in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1150231#c28). After this is complete, you should be able to sign up for a new account against the FxA staging server.

Then, to test the fix in this PR, follow the [instructions in the FxA issue](https://github.com/mozilla/fxa-content-server/issues/2934#issuecomment-131576572). To clarify, prior to this PR, if the FxA iframe errored for whatever reason, the user never gets any visual feedback, just an empty white box. After this PR (and with the AdBlock hack to force an FxA error mentioned above), the user should see the FxA iframe's 500 error message after 1.5 seconds.

jpetto:
@schalkneethling or @sgarrity r?

media/js/firefox/australis/fx38_0_5/firstrun.js
```diff
@@ -83,14 +11,14 @@
...
-            $ctaSignup.on('click', function(e) {
+            $('#cta-signup').on('click', function(e) {
                 e.preventDefault();
 
                 Mozilla.UITour.showFirefoxAccounts();
             });
 
             // signin CTA opens hamburger menu
-            $ctaSignin.on('click', function(e) {
+            $('#cta-signin').on('click', function(e) {
```
schalkneethling: Any reason we cannot keep this one and the one just above as cached variables as before?

schalkneethling:
This look good code wise, still need to do testing on demo4. One question though, any reason we are not removing the video and poster files?

media/js/firefox/australis/fx38_0_5/firstrun.js
jpetto: Since they're only being used once, I figured we didn't need to put them in variables. I'm happy to go back to `var`'s if you think that's better.

jpetto:
The video/poster are still in use on a couple Fx 38 whatsnew pages. Those pages probably aren't getting much traffic now, but I'd say the assets should remain as long as those pages do (just in case).

jpetto:
Just got [confirmation in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1195391#c3) that we're good to merge & push this pending review.

media/js/firefox/australis/fx38_0_5/firstrun.js
schalkneethling: Should be fine. Let's leave as is.

schalkneethling:
The good looks good to me so r+ unless @sgarrity has additional comments. Then I reckon when QA is happy, we can merge :cake:

jpetto:
@shane-tomlinson tested on demo4 and [gave it a :+1:](https://github.com/mozilla/fxa-content-server/issues/2934#issuecomment-132612508), so I think we're good to merge?

PR #3219, opened 2015-08-18
Bug 1136436 - Firefox for iOS product page

craigcook:
@flodolo, as usual, let me know if I've botched any strings. There's still a lot of front-end work to do but strings should be frozen.


TheoChevalier:
Took a quick look since flod is sleeping. Everything looks good to me. But I'll let flod double check in case I missed something.

Just making sure the following changes are what we want:
"Add Firefox to your iPhone, iPad and iPod touch." has been changed into "Add Firefox to your iPhone, iPad and iPod Touch." (touch vs. Touch)
Same goes for "Mobile Web browser for your iPhone, iPad and iPod Touch"

New strings:

* You’ll only receive updates and announcements about the iOS version of Firefox.
Thanks for your interest in Firefox for iOS.
Send Firefox to your iOS device
Don’t forget to check your inbox to confirm your request to get updates.

craigcook:
> "Add Firefox to your iPhone, iPad and iPod touch." has been changed into "Add Firefox to your iPhone, iPad and iPod Touch." (touch vs. Touch)

Correct, Apple brand guidelines say it's "iPod touch" with a lowercase t. We didn't know that last time so we're fixing it now. There are a few new strings and a few deleted strings, but I think those are the only *changed* strings.

TheoChevalier:
ok :) Missed a few strings in my comment, tough. One of them is "Beta release". Currently it's in uppercase in the page, is it supposed to be like this in the final design? If so, I guess we'll need to do what we usually do, putting the string in capitals in the template, so that locales can use lowercase when needed.

craigcook:
"Beta release" should be capitalized as it is in the template (uppercase B, lowercase r) and we'll use CSS to make it appear as all caps on the rendered page.

TheoChevalier:
Ok, so we will exclude asian locales with CSS rules?

flodolo:
@TheoChevalier  text-transform is disabled for all relevant locales with [site-wide rules](https://github.com/mozilla/bedrock/tree/master/media/css/l10n).

TheoChevalier:
Missed this change… Thanks flod

flodolo:
Looks like we're removing a few strings:
* Bookmarks
* Sync your Firefox browsing history on iOS
* First tap the <b>tab</b> button. Then tap the %(gear_icon)s icon to start using Sync.
* Not now
* Learn more in the App Store
* Add Firefox for iOS to your iPhone, iPad and iPod Touch.

Who's going to see the newsletter section? Only supported locales, or all locales?

flodolo:
Tested locally, can't see anything wrong with the strings. We need this on a demo server though.

craigcook:
Viewable on demo1 https://www-demo1.allizom.org/firefox/ios/ (still a lot to do but I'll be pushing regular updates)

craigcook:
As discussed on IRC and email, the strings that I believed were removed were not, in fact, intended to be removed (there are two competing copy documents with different text). I've restored the bits that were removed, and also changed "Beta release" to "Preview release."

flodolo:
Do we need "Learn more in the App Store."? I don't see it in the code

craigcook:
You're right, I missed "Learn more in the App Store" and also forgot the gear icon alt text. Updated. Sorry.

media/js/firefox/ios.js
```diff
@@ -0,0 +1,145 @@
...
+    // Get country code via geo-ip lookup
+    function getGeoLocation () {
+        try {
+            COUNTRY_CODE = geoip_country_code().toLowerCase();
+        } catch (e) {
+            COUNTRY_CODE = '';
+        }
+    }
+
+    $script('//geo.mozilla.org/country.js', function() {
```
alexgibson: Any reason not to always use `https://`?

media/js/firefox/ios.js
```diff
@@ -0,0 +1,145 @@
...
+
+    var $body = $('body');
+    var $html = $('html');
+    var COUNTRY_CODE = '';
+
+    // Countries in which Firefox for iOS is available.
+    // Add two-letter country codes (lowercase) to this array to
+    // to expand the market. We'll remove this logic for wide release.
+    // Note: This is the country code, NOT locale code; they often differ.
+    var market_countries = ['nz'];
```
alexgibson: Nit: use camelCase for variable names

media/js/firefox/ios.js
```diff
@@ -0,0 +1,145 @@
...
+            COUNTRY_CODE = geoip_country_code().toLowerCase();
+        } catch (e) {
+            COUNTRY_CODE = '';
+        }
+    }
+
+    $script('//geo.mozilla.org/country.js', function() {
+        getGeoLocation();
+
+        for (var i=0; i<market_countries.length; i++) {
```
alexgibson: Nit: spaces between operators

media/js/firefox/ios.js
```diff
@@ -0,0 +1,145 @@
...
+
+    $('#sync-instructions .btn-dismiss').on('click', function(e) {
+        e.preventDefault();
+        $html.removeClass('noscroll');
+        $body.append($instructions);
+        $fill.remove();
+    });
+
+    // Sync stuff
+    window.dataLayer = window.dataLayer || [];
```
alexgibson: Is this needed?

bedrock/firefox/templates/firefox/ios.html
```diff
@@ -0,0 +1,146 @@
...
+      </div>
+
+    </div>{#-- /.sync-cta --#}
+
+  </div>{#-- /.container --#}
+</section>{#-- /#sync --#}
+
+{% if has_widget %}
+  <div id="send-to-modal-container">
+    {{ send_to_device(platform='ios', title_text=_('Send Firefox to your iOS device')) }}
```
alexgibson: Not sure if this is required, but if you add `include_logo=True` you'll get the Firefox logo and rounded corners on the modal container

craigcook: We'll be trying an alternate style for iOS, I just haven't done it yet.
widget-head

media/js/firefox/ios.js
craigcook: Probably not, just an artifact of copy/paste.

media/js/firefox/ios.js
craigcook: Probably not. This was copied from the [Firefox OS page](https://github.com/mozilla/bedrock/blob/master/media/js/firefox/os/firefox-os.js#L164)

media/js/firefox/ios.js
```diff
@@ -0,0 +1,159 @@
...
+            COUNTRY_CODE = geoip_country_code().toLowerCase();
+        } catch (e) {
+            COUNTRY_CODE = '';
+        }
+    };
+
+    $.getScript('https://geo.mozilla.org/country.js', function() {
+        getGeoLocation();
+
+        for (var i = 0; i < marketCountries.length; i++) {
```
schalkneethling: This can be optimized a bit by caching the length of the array:

```
for (var i = 0, l = marketCountries.length; i < l; i++) {
```

media/js/firefox/ios.js
```diff
@@ -0,0 +1,159 @@
...
+    var state = 'Unknown';
+    var syncCapable = false;
+
+    var swapState = function(stateClass) {
+        $body.removeClass('state-default');
+        $body.addClass(stateClass);
+    };
+
+    // Sniff UA for Firefox for iOS
+    var isFirefoxiOS = function(userAgent) {
```
schalkneethling: I would suggest moving this function into `global.js` and adding a few unit tests for it in:
https://github.com/mozilla/bedrock/blob/master/tests/unit/spec/base/global.js

media/js/firefox/ios.js
```diff
@@ -0,0 +1,159 @@
...
+    // Send page state to GA if it hasn't already been sent
+    if (syncCapable === false) {
+        window.dataLayer.push({
+            event: 'ios-page-interactions',
+            interaction: 'page-load',
+            loadState: state
+        });
+    }
+
+})(window.jQuery, window.Mozilla);
```
schalkneethling: You can shorten this to just `jQuery, Mozilla` I reckon.

media/css/firefox/ios.less
```diff
@@ -0,0 +1,690 @@
...
+.fx-logo {
+    float: left;
+
+    a {
+        display: block;
+        width: 110px;
+        height: 115px;
+        margin-top: @baseLine;
+        .image-replaced();
+        background: url(/media/img/firefox/family/nav-sprite7.png) 0 0 no-repeat;
```
schalkneethling: I do not believe we need these `background` fallbacks. I have tested this with JS disabled and IE8 and all seems fine. Is there a specific reason you include these?

schalkneethling:
Misalignment of icons under the hero space in Safari 8.0.8 ;(

screen shot 2015-09-01 at 14 07 37

schalkneethling:
@craigcook Is there a way to have the send to device widget load? It seems to be based on `SEND_TO DEVICE_LOCALES` but, even with en-US I do not see it. Also, what needs to be set for the beta flag content to be shown? Thanks!

jpetto:
@craigcook I took a shot at refactoring `ios.js` and put [the commit](https://github.com/jpetto/bedrock/commit/a9d7c519ec0a9c9248e688ec2b158410681f5c60) up on [my fork of the branch](https://github.com/jpetto/bedrock/tree/bug-1136436-ios-product-page). I haven't noticed any geolocation script load issues after the refactor. Give it a look?

schalkneethling:
Ah ok, so adding the class `in-market` to the body tag shows the pre release header content.

media/js/firefox/ios.js
craigcook: Yeah, I'm sure we'll do that soon, but we decided to do it on just this page for now to reduce the scope (messing with the platform detection in global.js has more risk and needs thorough testing).

media/css/firefox/ios.less
craigcook: I'm not sure I follow. This is for the Firefox logo, and really it's only temporary because once FxiOS has a wide release we'll be adding the family nav to this page, which is also why I used the family sprite instead of adding a new logo image (though there are other logo images I could use instead). This logo is a bit of a hack until we can add the family nav.

craigcook:
> Is there a way to have the send to device widget load? It seems to be based on SEND_TO DEVICE_LOCALES but, even with en-US I do not see it. Also, what needs to be set for the beta flag content to be shown? Thanks!

We're explicitly hiding the widget for the initial release, but we intend to set up an A/B test of some sort to measure performance of the widget vs the direct app store button. So for now the widget is built in and never shown. You can show it by setting the page to the in-market state (either add the class via dev tools or add your country to the array in the JS) and unhiding the send-to-device button (`.get-fxios .send-to-device`).

craigcook:
> Misalignment of icons under the hero space in Safari 8.0.8 ;(

Yeah, I was using flexbox which I actually didn't need anyway. Converted to classic floats so we should be good now.

media/js/firefox/ios.js
```diff
@@ -0,0 +1,167 @@
...
+
+    // send-to-device form
+    var $widget = $('#send-to-modal-container');
+    var sendToDeviceForm = new Mozilla.SendToDevice();
+
+    // Countries in which Firefox for iOS is available.
+    // Add two-letter country codes (lowercase) to this array to
+    // to expand the market. We'll remove this logic for wide release.
+    // Note: This is the country code, NOT locale code; they often differ.
+    var marketCountries = ['nz', 'us'];
```
craigcook: Oops, note to self: Remove US. (didn't mean to commit this)

craigcook:
We've been approved so this is ready for final review and merge.

The App Store vs Send-to-Device test is nearly ready and may require a few tweaks to the GA stuff, but we shouldn't delay for that, we can add it separately.

media/css/firefox/ios.less
schalkneethling: Oh no, I meant when using the at2x mixin, we do not need a background fallback CSS rule so, for example here, you can remove the line `background: ....` and just use the at2x mixin by itself. Unless, as mentioned above, there is a specific reason you are using both.

schalkneethling:
In IE8 the logo is clipped, I am guessing because it does not support background size :-/ Changing the width and height on `.fx-logo a` to 145px respectively solves the problem though.

ie8 - win7_1

media/css/firefox/ios.less
craigcook: Ah! You're right, of course. The declarations are redundant. (I'm still working on a non-retina screen so I tend to do low resolution first and add the larger images later, and I didn't properly clean up after myself). I'll fix that.

craigcook:
> In IE8 the logo is clipped, I am guessing because it does not support background size :-/ Changing the width and height on .fx-logo a to 145px respectively solves the problem though.

So, as a purely lazy hack, I was using the family nav sprite instead of adding a new logo file (I was racing against time last week trying to get something up on demo before a meeting). But now I've added a proper single logo image at the correct size so IE8 (and others) should get the 110x115 logo without needing `background-resize`.

bedrock/firefox/templates/firefox/ios.html
```diff
@@ -0,0 +1,146 @@
...
+    <h1>Firefox<br> for iOS</h1>
+  {% else %}
+    <h1>{{ _('Firefox for iOS') }}</h1>
+  {% endif %}
+
+    <div class="get-fxios {% if has_widget %}show-widget{% endif %}">
+      <h2>{{ _('Add Firefox to your iPhone, iPad and iPod touch.') }}</h2>
+
+      <p class="appstore-badge">
+        <a href="{{ settings.APPLE_APPSTORE_FIREFOX_LINK }}"><img src="{{ l10n_img('firefox/ios/btn-app-store.svg') }}" alt="{{ _('Download on the App Store') }}"></a>
```
schalkneethling: IE8 does not support SVG and thus shows a broken image :( IE8 :angry:

schalkneethling:
@craigcook or @alexgibson Is there some about:config trickery one needs to employ to get these sections to show up based on your signed in status?

https://github.com/mozilla/bedrock/pull/3219/files#diff-bc0bba81284aed12c94d08220c3c4999R104 and below. Thanks!

craigcook:
> Is there some about:config trickery one needs to employ to get these sections to show up based on your signed in status?

There is indeed, Look for `browser.uitour.testingOrigins` (or create it if you don't have it) and add your localhost URL so it'll be whitelisted. You can also add demo*, dev, stage, etc in a comma-separated list, if you like.

You'll also need to change `browser.uitour.requireSecure` to `false` or else it'll require SSL.

schalkneethling:
Those are the one's, thanks @craigcook

schalkneethling:
@craigcook So I discovered two issues:

1) If the user has an add blocker installed, geodude might get blocked (it does for me) and thus, the JS will never run and the download button will not show up, even for people who *are* in-market.
2) People with JS disabled will also never see the download button.

Had a chat with @alexgibson and we came up with the following that solves both problems:
https://gist.github.com/schalkneethling/41e17be33d9af6d68138

Now, it is true that with this, some people might see the download button that are not in market, but rather that, then a bunch of in market people never seeing it at all, right? ;)

schalkneethling:
One thing I find a little confusing, perhaps it is just me ;) The page states:

"Sign in to your Firefox Account to access….", and if you are signed in, it states below this "You are already signed in!"

But in fact, the signed in status is, your sync sign in status not fx accounts so, perhaps the previous line should read: "Sign in to Firefox Sync to access…."

Probably a bit late for l10n but, thought I would add a comment.

schalkneethling:
I do not believe there is anything else so, once the above final points are addressed. I believe we are ready to :shipit: Thanks for the great work @craigcook!

craigcook:
Just got confirmation from jbertsch that the default state should be to assume the visitor is out-of-market unless we positively determine they are in a market country. So if JS is off/fails, or geodude is blocked/fails, the fallback state is to show the email signup form and not the app store button.

There's a possibility of changing this if we see any alarming trends in the near future (like a surprisingly high number of email signups from market countries that should have gotten the app store) but for now we'll leave it as-is.

Not sure what do do about IE8 and the SVG button at this point. We could make PNG versions for every locale, and maybe we'll do that for the wider release if we see enough traffic from IE8, but for this limited release I'm willing to let IE8 users take another one for the team.

So with that, I think we may be good to merge?

craigcook:
WAIT! We should get a final thumbs up from product before merging/pushing this page, and there may still be some question of whether the app store link is functional in New Zealand. So let's not merge just yet...

schalkneethling:
r+ :space_invader:

media/js/firefox/ios.js
```diff
@@ -0,0 +1,168 @@
...
+
+    // send-to-device form
+    var $widget = $('#send-to-modal-container');
+    var sendToDeviceForm = new Mozilla.SendToDevice();
+
+    // Countries in which Firefox for iOS is available.
+    // Add two-letter country codes (lowercase) to this array to
+    // to expand the market. We'll remove this logic for wide release.
+    // Note: This is the country code, NOT locale code; they often differ.
+    var marketCountries = ['nz'];
```
schalkneethling: @craigcook Just realized, should this have been removed or changed?

craigcook: Nope, this is as it should be. It's a bit hacky but this array seemed like the simplest way of listing countries for the soft launch. We'll just add more countries by hand as the release goes wider. Once we go global we'll remove all of the geo stuff entirely.

schalkneethling: Ooooh, so it is launching in NZ only. Interesting, did not know that. Ok, all good then. Just wanted to make sure I did not miss something.

PR #3224, opened 2015-08-19
[fix bug 1187829] Update waypoints JS lib.

jpetto:
- Remove orphaned 'products' bundle & JS file.


schalkneethling:
I saw the Alex mentioned that we should switch to using the waypoints version that is not dependent on jQuery but, also saw you mentioned the sticky.waypoints part does still require jQuery. Is this the reason why you stuck with the jQuery dependent version of the base waypoints library?

jpetto:
Yep, I stuck with the jQuery build of Waypoints to support the sticky plugin/shortcut.

schalkneethling:
All good r+ :spaghetti:

alexgibson:
Missed one here: https://github.com/mozilla/bedrock/blob/master/media/js/firefox/australis/common.js#L24

Doesn't break anything on the page, but I'll open a quick PR to tidy it up.

PR #3226, opened 2015-08-20
Fix Bug 1124038 - Move Thunderbird ESR page to Bedrock

kyoshino:


etc/httpd/global.conf
```diff
@@ -915,5 +915,9 @@ RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/products(/?)$ /b/$1firefox/products$
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/all\.html?$ /$1thunderbird/all/ [L,R=301]
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/all(/?)$ /b/$1thunderbird/all$2 [PT]
 
+# bug 1124038
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/organizations/(?:all-esr\.html|faq/)$ /$1thunderbird/organizations/ [L,R=301]
```
jpetto: Should the trailing slash after `faq` be optional?

kyoshino: :+1:

jpetto:
This, along with the other Thunderbird related PRs, is currently up on demo4: https://www-demo4.allizom.org/en-US/thunderbird/

kyoshino:
https://www-demo4.allizom.org/en-US/thunderbird/organizations/

bedrock/thunderbird/templates/thunderbird/organizations.html
```diff
@@ -0,0 +1,44 @@
...
+  {% stylesheet 'thunderbird-organizations' %}
+{% endblock %}
+
+{% block content %}
+  <header id="main-feature" class="stacked">
+    <h1>{{ _('Mozilla Thunderbird <span class="large">Extended Support Release</span>') }}</h1>
+    <h4>{{ _('A community-lead project that allows organizations to benefit from the speed, flexibility and security of Thunderbird while getting the support they need.') }}</h4>
+  </header>
+
+  <main id="main-content">
```
schalkneethling: hmmm, it seems like there is a bit of an inconsistency in our use of `role="main"` in the repo. A lot of pages have it, while there are also those that do not. For the sake of greater consistency, I would vote for adding the role attribute here. See the browser support section here for more: http://html5doctor.com/the-main-element/

bedrock/settings/static_media.py
```diff
@@ -859,6 +859,13 @@
         ),
         'output_filename': 'css/facebookapps_downloadtab-bundle.css',
     },
+    'thunderbird-organizations': {
+        'source_filenames': (
+            'css/base/menu.less',
+            'css/firefox/organizations.less',
+        ),
+        'output_filename': 'css/thunderbird-organizations-bundle.css',
```
schalkneethling: Nit: Naming convention would dictate this to be `thunderbird_organizations-bundle.css`

media/css/firefox/template-resp.less
```diff
@@ -24,6 +24,10 @@
     display: block;
 }
 
+.stacked h4 {
+    line-height: 1.3;
+}
```
schalkneethling: Curious, what does this and the rule added in `template.less` affect?

bedrock/settings/static_media.py
kyoshino: Some are using underscores and others are using hyphens. I prefer the latter for file names.

media/css/firefox/template-resp.less
kyoshino: Somehow `firefox/template-resp.less` used on this page, and adding this rule fixes the `line-height` on the Firefox ESR page as well so that's fine. The same rule was added `firefox/template.less` for consistency.

bedrock/thunderbird/templates/thunderbird/organizations.html
kyoshino: [ARIA in HTML](http://www.w3.org/TR/html-aria/) says `role="main"` is one of "Default Implicit ARIA semantics - SHOULD NOT be used" so I omit the role.

schalkneethling:
I have a tiny nit that probably only bother me ;) The nit is that the "Extended Support Release" text is not aligned with it's surrounding content:

screen shot 2015-08-27 at 10 13 38

Adding a `margin-left: -4px;` rule to the `.stacked h1 span` block solves this.

screen shot 2015-08-27 at 10 14 13

Again, this might only be bugging me so, I am going to leave it to you to make the call, unless someone else feels the same ;p

schalkneethling:
Other than the above, r+ :cake: Thanks @kyoshino!

bedrock/thunderbird/templates/thunderbird/organizations.html
```diff
@@ -0,0 +1,44 @@
...
+{% block body_id %}thunderbird-organizations{% endblock %}
+
+{% block page_css %}
+  {% stylesheet 'thunderbird-organizations' %}
+{% endblock %}
+
+{% block content %}
+  <header id="main-feature" class="stacked">
+    <h1>{{ _('Mozilla Thunderbird <span class="large">Extended Support Release</span>') }}</h1>
+    <h4>{{ _('A community-lead project that allows organizations to benefit from the speed, flexibility and security of Thunderbird while getting the support they need.') }}</h4>
```
schalkneethling: I wonder about this `h4` directly after the `h1`. Perhaps it would be a better call to use a paragraph element here? It reads more like a paragraph rather than a sub-heading.

schalkneethling:
@kyoshino only two small things left here, it needs a rebase and we should be good to merge. Thanks!

kyoshino:
Thanks @schalkneethling, fixed the remaining issues.

schalkneethling:
Thanks @kyoshino :cake:

PR #3228, opened 2015-08-20
Fix Bug 1124037 - Move Thunderbird Start and What's New pages to Bedrock

kyoshino:


jpetto:
This, along with the other Thunderbird related PRs, is currently up on demo4: https://www-demo4.allizom.org/en-US/thunderbird/

kyoshino:
https://www-demo4.allizom.org/en-US/thunderbird/beta/start/
https://www-demo4.allizom.org/en-US/thunderbird/earlybird/start/
https://www-demo4.allizom.org/en-US/thunderbird/earlybird/whatsnew/
https://www-demo4.allizom.org/en-US/thunderbird/nightly/start/
https://www-demo4.allizom.org/en-US/thunderbird/nightly/whatsnew/

malena:
Not sure if Chrome should matter for these pages but this is how the header looks like for all these pages in Chrome. Also if you add a clear:both to sections then the page would be one column which resembles the original page. Unless there is a reason these pages are in two columns?
screenshot 2015-08-26 11 14 22

kyoshino:
@malena As noted in `thunderbird/start.css`, I have used some CSS3 properties like [`margin-start`](https://developer.mozilla.org/en-US/docs/Web/CSS/margin-start) and those are not supported by Chrome yet. Since this page is viewed only in Gecko-based Thunderbird, it's totally fine.

kyoshino:
We could improve the page layout once the Thunderbird team prepares new content.

malena:
@kyoshino Okay great! I didn't see any other issues.

malena:
r+

malena:
Oh actually never mind. This shouldn't be necessary.

malena:
r+

PR #3232, opened 2015-08-21
[fix bug 1190614] Build alternate, optimized /firefox/new/.

jpetto:
After review/merge is complete, will test with Optimizely. See [bug 1197271](https://bugzilla.mozilla.org/show_bug.cgi?id=1197271).


jpetto:
Branch is up on demo3:

https://www-demo3.allizom.org/en-US/firefox/new/2/

@alexgibson r?

bedrock/firefox/templates/firefox/new2.html
```diff
@@ -0,0 +1,228 @@
...
+{% endblock %}
+
+{# all CSS must be in extrahead block for old IE #}
+{% block site_css %}{% endblock %}
+
+{% block extrahead %}
+  {% stylesheet 'firefox_new2' %}
+{% endblock %}
+
+{% block optimizely %}
```
alexgibson: Is Optimizely needed on both the control page and this one?

media/css/firefox/new2.less
```diff
@@ -0,0 +1,583 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../sandstone/lib.less";
+
+.sky #outer-wrapper {
+    background: #eaeff2;
+    background: -moz-radial-gradient(center, ellipse cover, rgba(234,239,242,1) 0%, rgba(212,221,228,.5) 60%),
```
alexgibson: We can save a few bytes by removing all this prefix cruft. Probably just need the -webkit prefix?

media/js/firefox/new2.js
```diff
@@ -0,0 +1,280 @@
...
+                    if (navigator.appVersion.indexOf('MSIE') !== -1) {
+                        return;
+                    }
+
+                    $directDownloadLink.trigger('click');
+                }
+            }
+
+            // prep both scenes for display
+            $downloadStarted.removeClass('invisible');
```
alexgibson: Is there a reason why we need to rely on opacity here as opposed to using something like `visually-hidden`?

bedrock/settings/static_media.py
```diff
@@ -1171,6 +1180,20 @@
...
     },
+    'firefox_new2': {
+        'source_filenames': (
+            'js/libs/jquery-1.11.3.min.js',
+            'js/libs/spin.min.js',
+            'js/base/global.js',
+            'js/base/global-init.js',
+            'js/base/search-params.js',
+            'js/newsletter/form.js',
+            'js/base/mozilla-input-placeholder.js',
```
alexgibson: Can we safely omit this file? I think only up-to-date Firefox users see the newsletter, in which case this polyfill shouldn't be needed.

media/js/firefox/new2.js
```diff
@@ -0,0 +1,280 @@
...
+
+            // An iframe can not be used here to trigger the download because
+            // it will be blocked by Chrome if the download link redirects
+            // to a HTTP URI and we are on HTTPS.
+            function trackAndRedirect(url, virtualUrl) {
+                // Delay to initiate download is required to allow animation
+                // to finish loading in IE. If delay is removed, the DOM will
+                // unload before the animation completes and the page will
+                // stop in a half-animated state.
+                window.setTimeout(function() {
```
alexgibson: Is this 500ms delay still needed now we removed the animation?

bedrock/firefox/templates/firefox/new2.html
jpetto: Yep. :unamused:

media/js/firefox/new2.js
jpetto: Using `opacity` for the transition. JS decides which "scene" to show when the page loads, so thought a short transition from `opacity: 0;` to `opacity: 1;` was more pleasing. Any reason `visually-hidden` mixin would be preferable?

bedrock/settings/static_media.py
jpetto: Ah, good catch! Removing.

media/js/firefox/new2.js
alexgibson: > Any reason visually-hidden mixin would be preferable?

Ah I see. I wasn't quite clear on why this bit was needed. I didn't really notice the transition effect when I tested the page locally, but I see it now. I had assumed you we're wanting to hide content until the page had loaded, which is why I suggested using `visibility` as this has broader support in old browsers. But if this is only for visual effect then we're probably all good?

media/css/firefox/new2.less
jpetto: I'm in with that. Should we keep both `-webkit` prefixes, or just the most recent?

media/js/firefox/new2.js
jpetto: Yeah, just a little visual nicety.

media/js/firefox/new2.js
jpetto: I'm going to remove it and put the branch on a demo to test in IE. Hopefully you're right. :+1:

media/css/firefox/new2.less
```diff
@@ -0,0 +1,579 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../sandstone/lib.less";
+
+.sky #outer-wrapper {
+    background: #eaeff2;
+    background: -webkit-gradient(radial, center center, 0px, center center, 60%, color-stop(0%,rgba(234,239,242,.9)), color-stop(60%,rgba(212,221,228,.5))),
```
alexgibson: I think we can get rid of this one too? Only the most recent -webkit prefixed version of the standard property should be needed still.

media/js/firefox/new2.js
alexgibson: One thought that just occurred to me - could this possibly effect download conversions negatively on slower connections? (e.g. waiting for the window load event to see the content?)

jpetto: Hm, that is a good thought. They would see the headline and logo and such, but not the main content. I could make scene 1 visible by default, then fade in scene 2 on `window.load` if instructed by URL params.

This catch is that it's possible the download button would be visible but then disappear as the user is reaching for it. However, the download would start automatically in this case (on IE 9+, at least). Probably worth it?

jpetto:
Tested removing the 500ms delay on the `trackAndRedirect` on [demo3](https://www-demo3.allizom.org/en-US/firefox/new/2/) in IE 7 & 8 and all seems well. :raised_hands:

jpetto:
Went with the heavier, but seemingly better in this case, jQuery `fade` rather than transitioning opacity. Flow is much closer to original (scene 1 always loads, then transitions to scene 2), rather than loading scene 2 directly (when URL dictates) as in previous commit. Result is faster display of content (including download button) on slower internet connections.

Left this change in a separate commit in case we want to back it out.

[demo3](https://www-demo3.allizom.org/en-US/firefox/new/2/) has been updated with the latest.

alexgibson:
Looks like Safari barfs a bit on the animation now that we removed the 500ms delay on the download :crying_cat_face:

media/js/firefox/new2.js
```diff
@@ -0,0 +1,272 @@
...
+        if (!noScene2 && $('.download-button-wrapper:visible').length > 0) {
+            var $downloadButtonLinks = $('.download-button-wrapper .download-other-desktop').detach();
+            $downloadButtonLinks.css('display', 'block').insertBefore('#firefox-screenshot');
+        }
+
+        if (site.platform === 'android') {
+            $('#download-button-android .download-subtitle').html(
+                $('.android.download-button-wrapper').data('upgradeSubtitle'));
+
+            $downloadStarted.removeClass('invisible');
```
alexgibson: Is this line and the one below still needed?

jpetto: D'oh! Nope, not needed. Deleting.

alexgibson:
Code looks good :+1: 

Tested demo3 on IE6-11, Edge, Chrome, Safari, Firefox, Opera and seems to all be working as expected.

r+ care to squarsh?

jpetto:
Waiting until Optimizely test is ready before merge.

jpetto:
As having the new URL live wont hurt anything, we're [good to merge before the Optimizely test is ready](https://bugzilla.mozilla.org/show_bug.cgi?id=1197271#c5).

PR #3236, opened 2015-08-24
Add functional tests for the Firefox DNT page

schalkneethling:
@alexgibson r?

### DIFF
--- a/tests/functional/firefox-dnt.js
+++ b/tests/functional/firefox-dnt.js
@@ -1,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public" + '/n' +
+ "* License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global casper */
+'use strict';
+
+var config = require('../lib/config');
+var helpers = require('../lib/helpers');
+var path = '/firefox/dnt/';
+var url = config.base() + path;
+
+casper.test.begin('Firefox DNT, Elements: ' + url, 6, function suite(test) {
+    casper.start(url, function() {
+        test.assertHttpStatus(200);
+
+        test.assertVisible('#dnt-status-wrapper', 'The DNT status section is visible');
+
+        var href = this.getElementAttribute('.sidebar-inset > p a', 'href');
+        test.assert(href.indexOf('support.mozilla.org/kb/how-do-i-stop-websites-tracking-me') > -1, 'How to enable DNT in Fx link exists');
+
+        test.assertVisible('.sidebar-box', 'IE9 DNT instructions visible');
+
+        test.assertVisible('#faq', 'The FAQ section is visible');
+        test.assertElementCount('#faq section', 8, 'There are eight sections in the FAQ');
+    });
+
+    casper.run(function() {
+        test.done();
+        helpers.done();
+    });
+});
+
+casper.test.begin('Firefox DNT, FAQ interaction: ' + url, 2, function suite(test) {
+    casper.start(url, function() {
+        this.click('#faq section:first-of-type h3');
+    });
+
+    casper.waitForSelector('#faq-1-panel[aria-hidden="false"]', function() {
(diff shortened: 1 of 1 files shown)


tests/functional/firefox-dnt.js
```diff
@@ -0,0 +1,53 @@
...
+        helpers.done();
+    });
+});
+
+casper.test.begin('Firefox DNT, FAQ interaction: ' + url, 2, function suite(test) {
+    casper.start(url, function() {
+        this.click('#faq section:first-of-type h3');
+
+        var hidden = this.getElementAttribute('#faq-1-panel', 'aria-hidden');
+        test.assert(hidden === "true", 'The first FAQ section expanded');
```
alexgibson: Nit: mixed quotes

tests/functional/firefox-dnt.js
```diff
@@ -0,0 +1,53 @@
...
+
+        this.click('#faq section:first-of-type h3');
+    });
+
+    casper.wait(1000, function() {
+        var hidden = this.getElementAttribute('#faq-1-panel', 'aria-hidden');
+        test.assert(hidden === "false", 'The first FAQ section collapsed');
```
alexgibson: Nit: mixed quotes

tests/functional/firefox-dnt.js
```diff
@@ -0,0 +1,53 @@
...
+    casper.start(url, function() {
+        this.click('#faq section:first-of-type h3');
+
+        var hidden = this.getElementAttribute('#faq-1-panel', 'aria-hidden');
+        test.assert(hidden === "true", 'The first FAQ section expanded');
+
+        this.click('#faq section:first-of-type h3');
+    });
+
+    casper.wait(1000, function() {
```
alexgibson: I think generally we should try and avoid arbitrary timeouts - would `waitUnitlVisible` be a more robust alternative here?

schalkneethling: I actually tried that before this solution but, it seems that for some reason the "animation" of opening and closing the containers trips up Casper. It can probably even be 500ms, it is just to wait for the animation to finish.

With all of that said. I will give it a try again, and let you know.

alexgibson: Could always see if `waitForSelector` works too when aria-hidden is changed?

schalkneethling: Yeah, just tried:

```
casper.waitWhileVisible('#faq-1-panel', function() {
    test.assert(true, 'The first FAQ section collapsed');
});
```
And I get the timeout error:

```
Timeout selector visibility change: #faq-1-panel
Exiting test run to avoid a race in which results may be unpredictable.
```

tests/functional/firefox-dnt.js
```diff
@@ -0,0 +1,53 @@
...
+        helpers.done();
+    });
+});
+
+casper.test.begin('Firefox DNT, FAQ interaction: ' + url, 2, function suite(test) {
+    casper.start(url, function() {
+        this.click('#faq section:first-of-type h3');
+
+        var hidden = this.getElementAttribute('#faq-1-panel', 'aria-hidden');
+        test.assert(hidden === 'true', 'The first FAQ section expanded');
```
alexgibson: Wait, isn't the logic for this backward? Maybe I'm confused? :confused: 

This seems to work? https://gist.github.com/alexgibson/02ca16bba328d2762a83

schalkneethling: It seems to be yes, but it totally works, or seems to anyhow :-/

schalkneethling: That Gist looks much better, I will switch. Thx

alexgibson: I think it's because the property is only applied after the animation completes, so you're racing here?

schalkneethling:
@alexgibson updated.

tests/functional/firefox-dnt.js
schalkneethling: Yup, that makes total sense to me.

alexgibson:
r+ :cake:

PR #3237, opened 2015-08-24
Fix Bug 1196471, remove id from download button to avoid duplicate ids

schalkneethling:
@craigcook r?


alexgibson:
I would vote for leaving this functionality in the download button helper, and either:

a). Specify two different ID's on the home page for each button (simplest change, this functionality already exists).
b). Modify the download button helper so that an ID is only included in the markup if explicitly defined using the macro.

Others may disagree (in which case feel free to chime in), but I think having the ability to define an ID for a given download button could still come in useful.

schalkneethling:
After hearing yesterday that the functionality actually does exist to pass a custom id, I totally agree that we should leave it as is and just pass a custom id to one or both of the buttons on the home page.

schalkneethling:
I have made the change to specify a unique id for one of the buttons instead of removing the id attribute.

alexgibson:
r+ :cake:

PR #3242, opened 2015-08-24
Bug 1166055 - Refactor RTL handling in CSS, part 1

kyoshino:
Let’s introduce RTL-aware mixins to make our coding easier. There will be 6 PRs (or more?) to get all the things fixed, and this is the first step. @mermi might give some help :-)

1. Add mixins, update `.span`-related stuff
2. Update shared LESSes under `base`, `sandstone` and `tabzilla`
3. Update `firefox` LESSes, part 1
4. Update `firefox` LESSes, part 2
5. Update `mozorg` LESSes
6. Update other LESSes

### DIFF
--- a/docs/coding.rst
+++ b/docs/coding.rst
@@ -131,6 +131,28 @@ sizes in either fixed units (``px``, ``pt``, etc) or element-relative units (``e
 See `this post by Jonathan Snook <http://snook.ca/archives/html_and_css/font-size-with-rem>`_
 for more info.
 
+Use the useful RTL-aware mixins for direction-related CSS properties including
+``right``, ``left``, ``float``, ``clear``, ``margin``, ``border``, ``padding``
+and ``text-align``, when applicable. Those can be found in ``sandstone/lib.less``.
+For example::
+
+    .box {
+        .float-left;
+        .margin-right(10px);
+    }
+
+will be converted by LESS as::
+
+    html[dir="ltr"] .box {
+        float: left;
+        margin-right: 10px;
+    }
+
+    html[dir="rtl"] .box {
+        float: right;
+        margin-left: 10px;
+    }
+
 Configuring your code editor
 ----------------------------
 
--- a/media/css/base/mozilla-modal.less
+++ b/media/css/base/mozilla-modal.less
@@ -125,7 +125,7 @@ html[dir="rtl"] {
 @media only screen and (min-width: @breakTablet) and (max-width: @breakDesktop) {
     #modal .inner {
         .span_narrow(12);
-        float: none;
+        .float-none;
         margin: @baseLine auto (@baseLine * 3);
     }
 }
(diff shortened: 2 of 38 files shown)


craigcook:
I'm a little worried about the overhead of a bunch of separate mixins for every property. What about a single mixin that takes a property and two values? Something like [what I did here for Master Firefox OS](https://github.com/mozilla/masterfirefoxos/blob/master/masterfirefoxos/base/static/stylus/lib.styl#L47-L64). That's Stylus though and I'm not sure how to accomplish the equivalent in Less.

The other thing giving me pause is hanging all styles from `html[dir]`. It means old browsers that don't support attribute selectors would get nothing, any document lacking a `dir` on the `html` element would get nothing (unlikely, but could happen), and it adds two points of specificity to every selector so things get harder to override.

It would be better to have a baseline rule with a simple selector, then modify it with a more specific rule for the other direction (most Earth languages are LTR so we tend to assume that as the baseline and treat RTL as the exception). That's what we currently do by hand but we could do the same with mixins.

kyoshino:
Treating RTL as an exception makes it a little bit difficult to override the style in media queries, based on my experience. Also, we don't have to think about older browsers; on Bedrock those stylesheets are not served to IE7 and below anyway. That's why I didn't put a baseline rule, rather applied the rule both to LTR and RTL. For readability, property-specific mixins might be better, I guess, than one single `bidi-style`.

craigcook:
I didn't mean IE specifically, I meant any/all browsers that may not support attribute selectors, whatever they may be. That's not a great concern for known browsers, but it's still the case that we'd be using a more complex CSS3 selector for basic CSS1 properties like `float` and `margin` without also offering some baseline with a simple selector for browsers we can't even imagine. 

The other points also still stand: it requires a `dir` attribute and increases specificity. We should favor the simplest, least specific selector whenever possible.

And I'm not sure I understand what you mean about overriding in media queries. Can you give an example? With something like: 
```
.foo { float: left; } 
[dir='rtl'] .foo { float: right; }
```
the more specific selector wins. Do you mean if you then need to override the more specific RTL style in a media query? Doesn't the query itself make it more specific?

jpetto:
Are users negatively affected by our current handling of RTL? Would user experience improve or decline with this change?

Are bedrock developers negatively affected by our current handling of RTL? Would dev experience improve or decline with this change?

I agree with @craigcook that we do not want to be overly specific for LTR CSS. I don't see the reason to add `html[dir='ltr']` before all these selectors. This would add weight to the compiled CSS files for minimal to no gain.

Would also like clarification on the "Treating RTL as an exception makes it a little bit difficult to override the style in media queries" comment.

If a change is warranted in terms of dealing with RTL CSS, I'd prefer something less intrusive/cumbersome, like @craigcook's example in Master Firefox OS (which does appear to be [supported in LESS](http://lesscss.org/features/#parent-selectors-feature-changing-selector-order)).

kyoshino:
As discusses in the meeting, I'm closing this for now and will work on a new approach on a different branch.

PR #3244, opened 2015-08-24
Fix Bug 1191136 - Download page says my Firefox is out of date when Firefox is not available for Firefox OS TV

kyoshino:


media/js/base/site.js
```diff
@@ -64,10 +64,7 @@
             if (pf.indexOf("Mac") !== -1) {
                 return 'oldmac';
             }
-            if (pf === '' &&
-                    /Firefox/.test(ua) &&
-                    /Mobile|Tablet/.test(ua) &&
-                    !(/Android/.test(ua))) {
+            if (pf === '' && /Firefox/.test(ua)) {
```
schalkneethling: So this then lumps together Firefox OS and Firefox OS TV but, do we not still need the `!(/Android/.test(ua)` part?

alexgibson: Do we also know for sure that `navigator.platform === ''` on Firefox OS TV? We only seem to have the UA string looking in the bug.

kyoshino: We have `/android/i.test(ua)` on line 41 so the Android UA check isn't necessary here. About the platform, actually I don't know. Maybe ` /Firefox/.test(ua)` is enough...

alexgibson: I don't think we should remove `pf === ''` as this was specific to Firefox OS. It would be good to get confirmation on what this value is, else we may be safer simply checking for `Viera`?

media/js/base/site.js
```diff
@@ -64,10 +64,7 @@
             if (pf.indexOf("Mac") !== -1) {
                 return 'oldmac';
             }
-            if (pf === '' &&
-                    /Firefox/.test(ua) &&
-                    /Mobile|Tablet/.test(ua) &&
-                    !(/Android/.test(ua))) {
+            if (/Firefox/.test(ua)) {
```
alexgibson: I don't think we should remove `pf === ''` as this was specific to Firefox OS. Only testing for `/firefox` in the UA and relying on the ordering of these conditionals feels a little brittle?

media/js/base/site.js
kyoshino: I'll ask the `navigator.platform` value in the bug.

jgmize: Based on the UA patterns in https://developer.mozilla.org/en-US/docs/Web/HTTP/Gecko_user_agent_string_reference#Firefox_OS and https://bugzilla.mozilla.org/show_bug.cgi?id=1191136#c0 the following regex should allow us to match the FxOS UA exclusively:

`/^Mozilla\/([0-9.])+ \(.*\) Gecko\/([0-9.])+ Firefox\/([0-9.])+$/`

alexgibson: `/^Mozilla\/[0-9.]+ \(.*\) Gecko\/[0-9.]+ Firefox\/[0-9.]+$/.test(navigator.userAgent)`

Gave this a quick whirl and it seems to return `true` for me for Firefox desktop and Android UA strings?

jgmize: > Gave this a quick whirl and it seems to return true for me for Firefox desktop and Android UA strings?

Oops, you're right @alexgibson, my mistake. Never mind, it seems the only way to differentiate is indeed with the navigator.platform value :/

kyoshino: Got a confirmation of `navigator.platform`. It's empty on the TV as well. So I guess this is now good to go.

alexgibson: Great - thanks @kyoshino!

alexgibson:
We're re-using existing strings here, so l10n is not effected by this change. Code looks fine, and as per the bug we look good as far as UA detection goes. r+ :dancer:

PR #3254, opened 2015-08-28
Fix Bug 1196468, fix header nesting for download button

schalkneethling:
@craigcook r? I navigated around the site with JS turned off and all looks good.

### DIFF
--- a/bedrock/firefox/templates/firefox/includes/download-button.html
+++ b/bedrock/firefox/templates/firefox/includes/download-button.html
@@ -4,7 +4,7 @@
 
 {% macro alt_buttons(builds) %}
   <div class="download download-dumb">
-    <h4>{{ _('Download Firefox') }} — {{ locale_name|safe }}</h4>
+    <p role="heading" class="download-heading">{{ _('Download Firefox') }} — {{ locale_name|safe }}</p>
     <ul>
       {% for plat in builds -%}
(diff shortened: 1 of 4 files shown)


bedrock/firefox/templates/firefox/includes/download-button.html
```diff
@@ -4,7 +4,7 @@
 
 {% macro alt_buttons(builds) %}
   <div class="download download-dumb">
-    <h4>{{ _('Download Firefox') }} — {{ locale_name|safe }}</h4>
+    <p role="heading" class="download-heading"><b>{{ _('Download Firefox') }} — {{ locale_name|safe }}</b></p>
```
craigcook: Should we go with `strong` for emphasis? Maybe the `heading` role alone implies enough importance, in which case we shouldn't need the `b` either.

schalkneethling: I guess this also brings up the question of whether we should add the attribute `aria-labeledby` to the `ul`. That means the `p` will need an id attribute as well. As the emphasis is not visual but more for auditory purposes, I would think the mapping for `role="heading"` would suffice and that you are correct, we probably do not need either `strong` nor `b`

schalkneethling:
@craigcook I tested this with VoiceOver and it sounds to me like the mapping to a heading is happening correctly. With that I removed the bold element wrapping the text.

craigcook:
r+! :palm_tree:  Wanna squarsh and merge?

PR #3256, opened 2015-08-28
Fix Bug 1199621, reduce bottom padding of heading on thunderbird email providers page

schalkneethling:
@jpetto r? I generally do not like to use id's in CSS but, in this case it is the only way to target the specific page, as it is used by the thunderbird/features page as well.

### DIFF
--- a/media/css/thunderbird/features.less
+++ b/media/css/thunderbird/features.less
@@ -8,6 +8,12 @@
     line-height: 1.3;
 }
 
+#email-providers {
+    main > section {
+        margin-top: 0;
+    }
+}
+
 [role="directory"] ul {
     display: block;
     overflow: hidden;


jpetto:
This may be too much thought for such a small change, but...

Since the standard is a `48px` gap between `#main-feature` and the following content, perhaps you could instead override the `40px` top margin on the `section` for this page?

Something like:

```
#email-providers main > section {
    margin-top: 0;
}
```

Would be a bit more consistent with the `thunderbird/features` page.

schalkneethling:
@jpetto no that is totally valid. Good thinking. I will update the PR.

schalkneethling:
@jpetto updated.

jpetto:
LGTM. r+ :chart:

PR #3257, opened 2015-08-28
[fix bug 1195369] Redirect iOS users on /firefox/new/ to iOS product page.

jpetto:
This PR must be merged after #3219, but can be code reviewed now.

Currently up on demo3, and can be tested with a (temporary) URL hack, e.g.

http://www-demo3.allizom.org/firefox/new/?referrer=http://www.google.com


jpetto:
@schalkneethling or @sgarrity r?

alexgibson:
Would be good to add a test for `getDomainFromURL` that correctly handles a non valid URL, just for good measure?

media/js/firefox/new-ios-redirect-helper.js
```diff
@@ -0,0 +1,37 @@
...
+if (typeof window.Mozilla === 'undefined') {
+    window.Mozilla = {};
+}
+
+(function(Mozilla) {
+    'use strict';
+
+    Mozilla.FirefoxNewIosRedirectHelper = {
+        getDomainFromURL: function(url) {
+            var matches = url.match(/^https?\:\/\/([^\/?#]+)(?:[\/?#]|$)/i);
```
alexgibson: Regex isn't one of my strong points, but out of curiosity what is the non-capturing group `(?:[\/?#]|$)` needed for? It doesn't seem to be an issue, but the tests seem to pass both with and without it.

alexgibson:
Code looks good, and testing on demo3 using an iOS device it seems to be working as expected. One question/suggestion but otherwise it's an r+ from me (optional wc).

Before this merges, do we need to do any sort of GA verification on demo3?

media/js/firefox/new-ios-redirect-helper.js
```diff
@@ -0,0 +1,37 @@
...
+                domain = domain.replace('www.', '');
+            }
+
+            return domain;
+        },
+        isSearchEngine: function(domain) {
+            var result = false;
+
+            // only concerned with these top 4 for now
+            if ((/bing\.|duckduckgo\.|google\.|yahoo\./i).test(domain)) {
```
alexgibson: Should we be checking specifically for `search.yahoo` here, as they have several web properties?

media/js/firefox/new-ios-redirect-helper.js
jpetto: Clearly not my strongest point, either. :grin:

You're correct - it isn't needed. Was trying to be too fancy by accounting for trailing querystring or hash, but there doesn't seem to be a need to care about anything beyond the domain itself. Removing.

media/js/firefox/new-ios-redirect-helper.js
jpetto: Good call. Confirmed with Gareth - will update to `search.yahoo.`.

jpetto:
@alexgibson Updated!

Gareth did test on demo3 to make sure the `utm_` params were coming through as expected. I don't *think* we need additional testing.

alexgibson:
Changes look good, tests pass r+ from me to merge when ready :+1:

jpetto:
Added a waffle switch to the JS block that performs the redirect.

jpetto:
Waffle switch name is `firefox-new-ios-redirect`.

craigcook:
Waffle working as expected. r++! :bamboo:

PR #3259, opened 2015-08-29
Add woff2 config to .htaccess

alexgibson:
Fwiw, we're only using `.woff2` fonts on [one page](https://www.mozilla.org/firefox/windows-10/welcome/) so far, which is only being shown a single time to Firefox users on Windows 10. But we should still add this going forward?

### DIFF
--- a/static/.htaccess
+++ b/static/.htaccess
@@ -2,6 +2,7 @@
 AddType image/svg+xml .svg
 AddType application/octet-stream .exe
 AddType application/font-woff .woff
+AddType application/font-woff2 .woff2
 AddType application/vnd.ms-fontobject .eot
 AddType application/octet-stream .ttf
 
@@ -33,7 +34,7 @@ FileETag MTime
 </FilesMatch>
 
 # fonts
-<FilesMatch "\.(eot|svg|ttf|woff)$">
+<FilesMatch "\.(eot|svg|ttf|woff|woff2)$">
     ExpiresByType application/octet-stream "access plus 1 year"
 </FilesMatch>
 


alexgibson:
@pmclanahan r?

jpetto:
r+! :checkered_flag:

pmclanahan:
I don't believe this file is in use any longer actually. Bedrock itself serves up these files now and ***should*** already be doing the right thing. I'll check.

pmclanahan:
Yes. Looks good (see below). I'll file a PR soon to delete this .htaccess file (I meant to do this a while ago. apologies).

```
$ curl -I https://mozorg.cdn.mozilla.net/media/fonts/FiraSans-Regular.186ecd9aa674.woff2
HTTP/1.1 200 OK
Accept-Ranges: bytes
Access-Control-Allow-Origin: *
Cache-Control: public, max-age=315360000
Content-Type: application/octet-stream
Date: Mon, 31 Aug 2015 17:19:18 GMT
Last-Modified: Mon, 10 Aug 2015 14:55:58 GMT
Server: ECAcc (fty/2F52)
X-Backend-Server: bedrock6.webapp.phx1.mozilla.com
X-Backend-Server: bedrock4.webapp.scl3.mozilla.com
X-Cache: HIT
X-Cache-Info: cached
X-Cache-Info: caching
Content-Length: 129452
```

alexgibson:
Ah, thanks @pmclanahan! This does make sense now, as I was curious why it worked ok without said changes - this explains things. I was just trying to cover our bases, but looks like it's already done. Thanks!

PR #3217, opened 2015-08-18
Fix Bug 836462 - Move Thunderbird landing page to Bedrock

kyoshino:
* A [screenshot](https://bugzilla.mozilla.org/attachment.cgi?id=8649109) can be found in the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=836462).
* Imported the localized strings from html to lang files in [r145109](https://viewvc.svn.mozilla.org/vc?revision=145109&view=revision) using a [script](https://gist.github.com/kyoshino/824667f6e279bd76c517).

PR #3262, opened 2015-08-29
Fix Bug 1199916 - Make image helpers always work without img prefix

kyoshino:
See the second commit in #3217 for why this should be fixed.

### DIFF
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -112,7 +112,7 @@
         {% endblock %}
 
         {% block site_header_logo %}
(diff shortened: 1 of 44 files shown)


schalkneethling:
This makes sense to me and is a couple less characters to type ;p I am going to leave this open for comment by other @jpetto @alexgibson etc. If this is going to get merged, we will have to update the relevant docs here:

http://bedrock.readthedocs.org/en/latest/coding.html#embedding-images

Thanks!

kyoshino:
`docs/coding.rst` in this commit will update the doc :smile_cat:

jpetto:
As other image helpers are implicitly aware of the top level `img` directory, and don't require `img/` in the path, I think this is a good update. Consistency ftw.

r+ from me. Pinging @pmclanahan to double-check the Python side of things.

PR #3263, opened 2015-08-31
Fix Bug 1199447, add new promo for thimble

schalkneethling:
Still waiting on some feedback from the bug so, marking as WIP and DNM with l10n


schalkneethling:
@alexgibson @flodolo r?

bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -185,7 +185,15 @@ <h2 class="primary go">{{ _('Teach the Web') }}</h2>
         <li id="promo-7" class="item promo-face">
           {{ promo_face(3) }}
         </li>
-      {% if l10n_has_tag('webmaker_promo_introducing') %}
+      {% if l10n_has_tag('webmaker_promo_thimble') %}
+        <li id="promo-8" class="item promo-small-landscape thimble" tabindex="0" data-name="Thimble">
+          <a class="panel-link" rel="external" href="https://thimble.mozilla.org/">
+            <div class="primary">
```
alexgibson: I don't think we need this `<div class="primary">`?

bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -185,7 +185,15 @@ <h2 class="primary go">{{ _('Teach the Web') }}</h2>
...
+              <h2 class="go">{{ _('Try the Thimble code editor') }}</h2>
```
alexgibson: Not sure we need `class="go"` her either?

schalkneethling:
Nope, they were not needed. \o/ for less HTML and CSS

media/css/mozorg/home/home-promo.less
```diff
@@ -890,6 +890,29 @@ html[lang|="en"] {
...
+/***************************************************************************/
+// Thimble
+.promo-small-landscape.thimble {
+
+    background-color: #2D9961;
+
+    h2 {
+        .at2x('/media/img/home/panel/thimble/mozilla-thimble.png', 69px, 75px);
+        background-repeat: no-repeat;
+        margin: 30px 20px;
```
alexgibson: Nit: slightly nicer centering with `25px @baseLine`?

media/css/mozorg/home/home-promo.less
```diff
@@ -890,6 +890,29 @@ html[lang|="en"] {
...
+        padding: 8px 0 0 85px;
+        height: 75px;
+    }
+}
+
+.html-rtl {
+    h2 {
```
alexgibson: Overly specific selector - should this be `.promo-small-landscape.thimble h2`?

alexgibson:
I would suggest we give non-english locales as much room for text as possible, similarly to how we handle the other small promos? https://github.com/mozilla/bedrock/blob/master/media/css/mozorg/home/home-promo.less#L756

media/css/mozorg/home/home-promo.less
schalkneethling: Hah, I had it like that and then changed it. Totally agree, targeting all h2 when it is rtl is not a good idea ;p

schalkneethling:
Ok, I did some tweaks and testing. Tested with German (below), Italian and French, and the spacing looks good.

screen shot 2015-09-01 at 16 20 49

Also tested RTL using Arabic.

screen shot 2015-09-01 at 16 24 20

schalkneethling:
Ok, the spacing looks better now after a small tweak for the on EN languages.

alexgibson:
r+ on the code changes - care to squash?

I think Flod might be on PTO, so pinging @pascalchevrel to see if we're ok to merge this. We're using an l10n tag so it should be fine, but we need to make sure folks are aware there's a new string.

PR #3264, opened 2015-08-31
Fix Bug 1150649 - Move Thunderbird all-beta and channel pages to Bedrock

kyoshino:
Do not merge yet. 2 things:

* The Beta logo requested in [Bug 1196942](https://bugzilla.mozilla.org/show_bug.cgi?id=1196942) is not yet provided.
* The Beta and Earlybird versions are hard-coded because those are not available via product-details. Since the new Ship It library is on the way, we could probably remove the setting soon.


bedrock/thunderbird/templates/thunderbird/base.html
```diff
@@ -0,0 +1,31 @@
...
+{% block site_css %}
+  {% stylesheet 'firefox' %}
+{% endblock %}
+
+{% block site_header_nav %}
+  {% include 'thunderbird/includes/top-menu.html' %}
+{% endblock %}
+
+{% block site_header_logo %}
+<h2><a href="{{ url('thunderbird.index') }}">{{ high_res_img('img/thunderbird/template/logo-wordmark.png', {'alt': 'Thunderbird', 'width': '260', 'height': '72'}) }}</a></h2>
```
schalkneethling: Seeing that we merged the PR to change `high_res_img` etc. you no longer need 'img/...'

bedrock/thunderbird/templates/thunderbird/channel.html
```diff
@@ -0,0 +1,69 @@
...
+    <li id="earlybird-link"><a href="#earlybird"><img src="{{ static('img/thunderbird/channel/toggler-earlybird.png') }}" alt="Thunderbird Earlybird" id="toggler-logo-earlybird" /></a></li>
+  </ul>
+
+  <a href="#" id="carousel-left" aria-hidden="true"></a>
+  <a href="#" id="carousel-right" aria-hidden="true"></a>
+
+  <div class="pager-content">
+
+    <div id="beta" class="pager-page">
+      <h2 class="channel-title channel-title-beta">{{ high_res_img('img/thunderbird/channel/title-beta.png', {'alt': 'Thunderbird Beta', 'width': '634', 'height': '74'}) }}</h2>
```
schalkneethling: same here as above

bedrock/thunderbird/templates/thunderbird/includes/top-menu.html
```diff
@@ -0,0 +1,25 @@
+<nav id="nav-main" role="navigation">
+  <span class="toggle" role="button" aria-controls="nav-main-menu" aria-expanded="false" tabindex="0">{{_('Menu')}}</span>
```
schalkneethling: This is visible in the UI but probably should not be. Also, I do not believe it should have the `role="button` as it does not act like a button and is static text.

bedrock/thunderbird/templates/thunderbird/includes/top-menu.html
```diff
@@ -0,0 +1,25 @@
...
+    <li id="nav-main-addons">
+      <a href="https://addons.mozilla.org/thunderbird/" tabindex="0" aria-owns="nav-main-addons-submenu" aria-haspopup="true">{{_('Add-ons')}}</a>
+      <ul aria-expanded="false" id="nav-main-addons-submenu" class="submenu">
+        <li class="first"><a href="https://addons.mozilla.org/thunderbird/extensions/" tabindex="-1">{{_('Add-ons')}}</a></li>
+        <li class="last"><a href="https://addons.mozilla.org/thunderbird/themes/" tabindex="-1">{{_('Themes')}}</a></li>
+      </ul>
+    </li>
+    <li class="last" id="nav-main-support">
+      <a href="https://support.mozilla.org/products/thunderbird" tabindex="0" aria-owns="nav-main-addons-submenu" aria-haspopup="true">{{_('Support')}}</a>
+      <ul aria-expanded="false" id="nav-main-addons-submenu" class="submenu">
```
schalkneethling: Is there a reason to have this drop down on support? As the main text and the drop down links to the same page, perhaps we just need the main menu entry?

schalkneethling:
There is a fair amount of Python in here, @pmclanahan mind taking a look at those?

pmclanahan:
*looks around* :eyes:

bedrock/settings/base.py
```diff
@@ -847,3 +847,10 @@ def facebook_tab_url_lazy():
 STATSD_HOST = config('STATSD_HOST', default='127.0.0.1')
 STATSD_PORT = config('STATSD_PORT', cast=int, default=8125)
 STATSD_PREFIX = config('STATSD_PREFIX', default='bedrock')
+
+# Thunderbird pre-release versions. Remove this once the new Ship-It library
+# provides the data. (Bug 1150649)
+THUNDERBIRD_VERSIONS = {
```
pmclanahan: This will almost certainly become out-of-date quickly. I see that there is a mention of `40.0b1` in product-details in the `thunderbird_history_development_releases.json` file, but it would be a little tricky to get it out of there reliably. If we could, we could then infer the alpha version. It's potentially error prone but possibly more reliable than having to remember to update this setting. What do you think?

etc/httpd/global.conf
```diff
@@ -936,9 +927,20 @@ RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/features(/?)$ /b/$1thunderbird/f
...
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/all-beta\.html$ /$1thunderbird/beta/all/ [L,R=301]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/((?:beta|earlybird)/)?(all|notes|system-requirements)(/?)$ /b/$1thunderbird/$2$3$4 [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/channel(/?)$ /b/$1thunderbird/channel$2 [L,PT]
+
+# bug 1081917, 1029829, 1029838
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/([3-9]\d\.\d(?:a2|beta|\.\d)?)/releasenotes(/)?$ /b/$1thunderbird/$2/releasenotes$3 [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/([3-9]\d\.\d(?:a2|beta|\.\d)?)/system-requirements(/)?$ /b/$1thunderbird/$2/system-requirements$3 [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/releases(/)?$ /b/$1thunderbird/releases$2 [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/releases/(.*)$ http://website-archive.mozilla.org/www.mozilla.org/thunderbird_releasenotes/en-US/thunderbird/releases/$2 [L,R=301]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?thunderbird/latest/(.*)/?$ /b/$1thunderbird/latest/$2/ [L,PT]
```
pmclanahan: I think you can leave off the `/?` at the end of the match and the `/` at the end of the destination. There could be some mismatch there that may make for some strange URLs.

pmclanahan:
Just a couple of small things, but looks good to me. If the answer to the unfortunate hard coding of alpha and beta versions is that we just have to deal with it for now, then we can go with it. I just wanted to make sure we'd exhausted all other options.

Perhaps @sylvestre is willing to add the current alpha and beta version numbers to the existing SVN-based product details?

sylvestre:
Beta versions are available on the next version of product details ( https://ship-it.mozilla.org/json/thunderbird_versions.json but vpn only for now )
Please submit a bug for aurora/alpha, I will manage that later. Thanks!

pmclanahan:
@sylvestre apparently I'm not authorized to see that (LDAP prompt that doesn't work for me). But for this at least we're only able to use what we currently have. Thanks for the info.

bedrock/thunderbird/templates/thunderbird/base.html
kyoshino: Yes! No more `img/`!

bedrock/thunderbird/templates/thunderbird/includes/top-menu.html
kyoshino: Will remove the drop down menu.

kyoshino:
If we can use Ship It real soon, that's much appreciated. Even though Thunderbird Beta is not updated frequently, hardcoding should be avoided as much as possible. I'll file a bug for Earlybird.

bedrock/thunderbird/templates/thunderbird/includes/top-menu.html
kyoshino: Hmm, it should not be visible on non-responsive views. Will hide it.

etc/httpd/global.conf
kyoshino: :+1:

kyoshino:
Current status:

* The Thunderbird logo with a Beta sash is not provided at lest for now, so let's proceed with a [2-column layout](https://bug1196942.bmoattachments.org/attachment.cgi?id=8660215) like the current page to avoid confusion.
* Ship It is not ready yet but coming soon. Is it okay to hard-code the version numbers before the API is available? Thunderbird Beta is not frequently updated like Firefox Beta; there are only two during the release cycle, I guess.

sylvestre:
I will try to update the product details tomorrow.

kyoshino:
@sylvestre That's much appreciated!

kyoshino:
The product-details lib has been updated by @sylvestre and I have removed the hard-coded pre-release versions in my PR. The two blockers now cleared, so we can now proceed!

schalkneethling:
@kyoshino What about the Thunderbird Beta logo? We proceeding without that?

schalkneethling:
Ah ok, so the bug seems to suggest we move ahead as is. Alrighty!

kyoshino:
Yep, we use the current Beta logo without any modification.

schalkneethling:
@kyoshino This looks good to me from a front-end perspective. One question though. The Thunderbird landing page does not have the navigation that is on the other pages, is this meant to be?

This means there is no direct link to releases from this page if a user wants to download the Beta so, they need to click on "Systems & Languages" and will then have access to the navigation.

Other than that, if @pmclanahan or someone can have a look over the Python side of things, we should be good to merge. From my general overview of the Python it looks all good. Thanks!

kyoshino:
The Thunderbird landing page is based on the Firefox equivalent, that's why the header navigation is missing so far. But personally I'd like to have it for a better UX. Will send a separate PR to do so.

schalkneethling:
@kyoshino sounds great! Thanks! Just waiting on the go ahead in terms of the Python bits, and we can merge this.

kyoshino:
According to the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1196942), the Creative team allows us to create the Thunderbird Beta logo with a sash, so we can switch back to the nice-y carousel design while not creating confusion. I'll give it a try.

pmclanahan:
Seems like we're good to go as soon as we get the new beta logo and rebase this branch. I'll take one last look when that's done. Please ping me here when ready @kyoshino. Thanks again!

schalkneethling:
@kyoshino Correct me if I am wrong but, I believe we can go ahead for now without the Thunderbird beta logo so, this just needs a rebase.

kyoshino:
Sorry I'm late here :bow: Just rebased the PR. Should be good to merge.

kyoshino:
I'll send another PR once the Thunderbird Beta logo is created so we can switch to the same design as the Firefox Channel page.

schalkneethling:
@pmclanahan You want to give this one last look over?

PR #3266, opened 2015-09-01
[bug1200464] Fixed the redirection bugs in ja locale

chikoski:
This pull request will change the site's redirection settings to fix the bug1200464, which navigate users who attempt to access ja-localzed relesenotes to wrong address.

### DIFF
--- a/etc/httpd/global.conf
+++ b/etc/httpd/global.conf
@@ -125,11 +125,10 @@ RewriteRule ^/ja/?$ http://www.mozilla.jp/ [L,R]
 # bug 1051686
 RewriteRule ^/ja/firefox/organizations/?$ http://www.mozilla.jp/business/downloads/ [L,R=301]
 
-# bug 1138280
-RewriteRule ^/ja/(firefox|thunderbird)/(beta/)?notes/ http://mozilla.jp/$1$2/notes/ [L,R=301]
-RewriteRule ^/ja/mobile/(beta/?)notes/ http://mozilla.jp/android/$1$2/notes/ [L,R=301]
-RewriteRule ^/ja/(firefox|thunderbird)/(\d+\.\d+(\.\d+)*)(beta)?/releasenotes(/?|/.+)$ http://mozilla.jp/firefox/$1/$2$3/releasenotes/ [L,R=301]
-RewriteRule ^/ja/mobile/(\d+\.\d+(\.\d+)*)(beta)?/releasenotes(/?|/.+)$ http://mozillla.jp/firefox/android/$1$2/releasenotes/ [L,R=301]
+# bug 1138280, bug 1200464
+RewriteRule ^/ja/(firefox|thunderbird)/(beta/)?notes/? http://www.mozilla.jp/$1/$2notes/ [L,R=301]
+RewriteRule ^/ja/(firefox|thunderbird)/((android/)?[0-9.]+(beta)?)/releasenotes/? http://www.mozilla.jp/$1/$2/releasenotes/ [L,R=301]
+
 
 ## Redirect things to django!
 


etc/httpd/global.conf
```diff
@@ -128,8 +128,8 @@ RewriteRule ^/ja/firefox/organizations/?$ http://www.mozilla.jp/business/downloa
 # bug 1138280
 RewriteRule ^/ja/(firefox|thunderbird)/(beta/)?notes/ http://mozilla.jp/$1$2/notes/ [L,R=301]
 RewriteRule ^/ja/mobile/(beta/?)notes/ http://mozilla.jp/android/$1$2/notes/ [L,R=301]
-RewriteRule ^/ja/(firefox|thunderbird)/(\d+\.\d+(\.\d+)*)(beta)?/releasenotes(/?|/.+)$ http://mozilla.jp/firefox/$1/$2$3/releasenotes/ [L,R=301]
-RewriteRule ^/ja/mobile/(\d+\.\d+(\.\d+)*)(beta)?/releasenotes(/?|/.+)$ http://mozillla.jp/firefox/android/$1$2/releasenotes/ [L,R=301]
+RewriteRule ^/ja/(firefox|thunderbird)/((android/)?\d+\.\d+(\.\d+)*(beta)?)/releasenotes(/?|/.+)$ http://mozilla.jp/$1/$2/releasenotes/ [L,R=301]
+RewriteRule ^/ja/mobile/(\d+\.\d+(\.\d+)*(beta)?)/releasenotes(/?|/.+)$ http://mozillla.jp/firefox/android/$1$2/releasenotes/ [L,R=301]
```
jgmize: mozilla.jp currently redirects to www.mozilla.jp, so we should probably save a redirect and change the targets of all of these redirects to www.mozilla.jp

jgmize: Actually, I think that we can safely remove line 132 entirely, as `/ja/mobile/` should get redirected to `/ja/firefox/android/` which would then get caught by the single rule on 131. Also, I believe we can simplify the instances of `\d+\.\d+(\.\d+)*` to a more readable `[0-9.]`, and `(/?|/.+)` can be safely simplified to `/?`. How about something like this?
```
RewriteRule ^/ja/(firefox|thunderbird)/((android/)?[0-9.](beta)?)/releasenotes/?$ http://mozilla.jp/$1/$2/releasenotes/ [L,R=301]
```

etc/httpd/global.conf
```diff
@@ -126,10 +126,9 @@ RewriteRule ^/ja/?$ http://www.mozilla.jp/ [L,R]
 RewriteRule ^/ja/firefox/organizations/?$ http://www.mozilla.jp/business/downloads/ [L,R=301]
 
 # bug 1138280
-RewriteRule ^/ja/(firefox|thunderbird)/(beta/)?notes/ http://mozilla.jp/$1$2/notes/ [L,R=301]
-RewriteRule ^/ja/mobile/(beta/?)notes/ http://mozilla.jp/android/$1$2/notes/ [L,R=301]
-RewriteRule ^/ja/(firefox|thunderbird)/(\d+\.\d+(\.\d+)*)(beta)?/releasenotes(/?|/.+)$ http://mozilla.jp/firefox/$1/$2$3/releasenotes/ [L,R=301]
-RewriteRule ^/ja/mobile/(\d+\.\d+(\.\d+)*)(beta)?/releasenotes(/?|/.+)$ http://mozillla.jp/firefox/android/$1$2/releasenotes/ [L,R=301]
+RewriteRule ^/ja/(firefox|thunderbird)/(beta/)?notes/ http://www.mozilla.jp/$1$2/notes/ [L,R=301]
+RewriteRule ^/ja/(firefox|thunderbird)/((android/)?[0-9]+\.[0-9]+(\.[0-9]+)*(beta)?)/releasenotes/? http://www.mozilla.jp/$1/$2/releasenotes/ [L,R=301]
```
chikoski: I have updated my redirection rules with review comment.

    [0-9]+\.[0-9]+(\.[0-9]+)*(beta)?

This part may give us some confusion, but it is need to redirect the access to the URLs such as /ja/firefox/38.0.5beta/relasenotes/

jgmize: My [original comment](https://github.com/mozilla/bedrock/pull/3266#discussion_r38385204) had a typo in it: my intended suggestion was to replace `\d+\.\d+(\.\d+)*` with `[0-9.]+` which is shorter, IMO more readable, and matches an arbitrary number of dotted numerals. I didn't intend to ask you to replace each `\d` with `[0-9]`-- I'm indifferent to that particular change, and I apologize for the unintended pedantry.

I see that you addressed the other changes as well, and the new version lgtm so far-- I'm going to do a bit of testing locally first and then merge if everything works as expected.

etc/httpd/global.conf
```diff
@@ -126,10 +126,9 @@ RewriteRule ^/ja/?$ http://www.mozilla.jp/ [L,R]
 RewriteRule ^/ja/firefox/organizations/?$ http://www.mozilla.jp/business/downloads/ [L,R=301]
 
 # bug 1138280
-RewriteRule ^/ja/(firefox|thunderbird)/(beta/)?notes/ http://mozilla.jp/$1$2/notes/ [L,R=301]
-RewriteRule ^/ja/mobile/(beta/?)notes/ http://mozilla.jp/android/$1$2/notes/ [L,R=301]
-RewriteRule ^/ja/(firefox|thunderbird)/(\d+\.\d+(\.\d+)*)(beta)?/releasenotes(/?|/.+)$ http://mozilla.jp/firefox/$1/$2$3/releasenotes/ [L,R=301]
-RewriteRule ^/ja/mobile/(\d+\.\d+(\.\d+)*)(beta)?/releasenotes(/?|/.+)$ http://mozillla.jp/firefox/android/$1$2/releasenotes/ [L,R=301]
+RewriteRule ^/ja/(firefox|thunderbird)/(beta/)?notes/ http://www.mozilla.jp/$1$2/notes/ [L,R=301]
```
pmclanahan: Looks to me like this would result in `http://www.mozilla.jp/firefoxbeta/notes/`. That is to say that I think there should be a `/` character between the `$1` and `$2` in the destination URL.

pmclanahan: Actually, if `beta/` is matched it'll also mean a double slash before `/notes`. How about this:

```apache
RewriteRule ^/ja/((?:firefox|thunderbird)/)(beta/)?notes/ http://www.mozilla.jp/$1$2notes/ [L,R=301]
```

etc/httpd/global.conf
chikoski: Ah, I misunderstood you intention. My original rule intended to describe rule strictly, but as you mentioned it is low readable. I took your suggestion and updated my pull request. Thanks!

etc/httpd/global.conf
chikoski: It is my mistake. I fixed this in my last pull-request, as well. Thanks again.

etc/httpd/global.conf
```diff
@@ -126,10 +126,9 @@ RewriteRule ^/ja/?$ http://www.mozilla.jp/ [L,R]
 RewriteRule ^/ja/firefox/organizations/?$ http://www.mozilla.jp/business/downloads/ [L,R=301]
 
 # bug 1138280
-RewriteRule ^/ja/(firefox|thunderbird)/(beta/)?notes/ http://mozilla.jp/$1$2/notes/ [L,R=301]
-RewriteRule ^/ja/mobile/(beta/?)notes/ http://mozilla.jp/android/$1$2/notes/ [L,R=301]
-RewriteRule ^/ja/(firefox|thunderbird)/(\d+\.\d+(\.\d+)*)(beta)?/releasenotes(/?|/.+)$ http://mozilla.jp/firefox/$1/$2$3/releasenotes/ [L,R=301]
-RewriteRule ^/ja/mobile/(\d+\.\d+(\.\d+)*)(beta)?/releasenotes(/?|/.+)$ http://mozillla.jp/firefox/android/$1$2/releasenotes/ [L,R=301]
+RewriteRule ^/ja/(firefox|thunderbird)/(beta/)?notes/ http://www.mozilla.jp/$1/$2/notes/ [L,R=301]
```
pmclanahan: Still broken because if `(beta/)` matches it'll result in a doubled slash before `notes`. Look at the line I suggested again and see if it'll work.

chikoski: Ah, I understood what you mean.
I changed that rule as
    RewriteRule ^/ja/(firefox|thunderbird)/(beta/)?notes/? http://www.mozilla.jp/$1/$2notes/ [L,R=301]

It will work as same as you mentioned before. (by the way I found another typo in your suggestions).

Thanks

pmclanahan:
Looks much better! Thanks!

PR #3267, opened 2015-09-01
[fix bug 1169038] Firefox Push FAQ page

alexgibson:


alexgibson:
@flodolo - I would not yet consider strings to be final. Expect some edits, but I will update here once ready.

alexgibson:
Ok I think we are good string wise according to folks in [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1169038) - removing DNM tag.

alexgibson:
For those that are curious, this page is accessed via a "Learn more" link in the new Push Notifications prompt in Firefox 42. Traffic to this page is likely quite small still, as there aren't many sites using the feature yet. But we're currently starting to hit 404's in production for this page, so it would be good to try and get this one reviewed and merged soon.

schalkneethling:
r+ :spades:

PR #3269, opened 2015-09-01
removed "the new"

marcoscaceres:
"the new" is confusing... I though it was talking about a "new" site, not the current one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 bedrock
 =======
 
-*bedrock* is the code name of the new mozilla.org. It is as shiny,
+*bedrock* is the code name of mozilla.org. It is as shiny,
 awesome, and open sourcy as always. Perhaps even a little more.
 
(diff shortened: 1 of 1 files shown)


pmclanahan:
When that was written, it was new :)

marcoscaceres:
sure, understood. But now, it's awesome but not so new :)

alexgibson:
Ironically, we're just only now finishing migrating from the "old" mozilla.org :wink:

README.md
```diff
@@ -1,7 +1,7 @@
 bedrock
 =======
 
-*bedrock* is the code name of the new mozilla.org. It is as shiny,
+*bedrock* is the code name of mozilla.org. It is as shiny,
```
pmclanahan: This is not exactly accurate yet. It's the code name for the django-based web site that has been gradually replacing the old site for years now. We're almost done, so maybe this will be accurate soon. But I think something more like "Bedrock is the name we use for the code behind www.mozilla.org" would be good.

craigcook:
I updated the commit message to clarify the change and merged this in e684963deb11e216879b00ac65dc85a0b183b32e

Thanks!

PR #3270, opened 2015-09-01
Bug 987852 Add redirects for archived MPL pages

sgarrity:
These redirects are for this set of pages moved to the website-archive server in this commit: https://viewvc.svn.mozilla.org/vc?revision=145674&view=revision


schalkneethling:
This works as expected. Checked the regexes against the folder structure in SVN and also ensured that the documents are live on http://website-archive.mozilla.org/ so r+ :cake:

PR #3271, opened 2015-09-02
Update partners page functional tests

alexgibson:


alexgibson:
@schalkneethling r? Been seeing some intermittent test failures for this page. Seems like we we're trying to run tests a bit early, rather than waiting for the page JS to finish initializing (this page does some lazy loading of JS).

schalkneethling:
r+ :monorail:

PR #3272, opened 2015-09-03
Add sslify for Deis deployments.

glogiotatidis:
When SSLify is enabled (disabled by default) will redirect http requests
to https.

@pmclanahan or @jgmize r?

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -294,6 +294,7 @@ DOMAIN_METHODS = {
 }
 
 MIDDLEWARE_CLASSES = (
+    'sslify.middleware.SSLifyMiddleware',
     'bedrock.mozorg.middleware.MozorgRequestTimingMiddleware',
     'django_statsd.middleware.GraphiteMiddleware',
     'bedrock.tabzilla.middleware.TabzillaLocaleURLMiddleware',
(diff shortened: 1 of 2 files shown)


requirements/base.txt
```diff
@@ -187,3 +187,6 @@ dj-database-url==0.3.0
 
 # sha256: ufv3R_qOcR0zDy9Da5suz17tnrZ6Y3yoGrzRHCq-3sg
 python-decouple==2.3
+
+# sha256: 9r_SEEjH38lbOWWaPad3ddbbHB96dFgFzMbGE494O20
+django-sslify==0.2.7
```
glogiotatidis: This is a python only dep so it should be fine to install on SCL

bedrock/settings/base.py
```diff
@@ -847,3 +848,7 @@ def facebook_tab_url_lazy():
 STATSD_HOST = config('STATSD_HOST', default='127.0.0.1')
 STATSD_PORT = config('STATSD_PORT', cast=int, default=8125)
 STATSD_PREFIX = config('STATSD_PREFIX', default='bedrock')
+
+SSLIFY_DISABLE = config('DISABLE_SSL', default=True, cast=bool)
+SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
+USE_X_FORWARDED_HOST = True
```
glogiotatidis: Not sure if the last two lines will affect the current SCL cluster. If yes we can only enable them when `not SSLIFY_DISABLE`

pmclanahan: That's probably safest.

glogiotatidis: Done

pmclanahan:
r+ will merge after the iOS page is out the door. Just don't want repo churn at this moment.

glogiotatidis:
> r+ will merge after the iOS page is out the door. Just don't want repo churn at this moment.

ok. Meanwhile I updated deis env variables `DISABLE_SSL=False` on both eu and us clusters.

pmclanahan:
Awesome. Thanks!

PR #3273, opened 2015-09-03
[fix bug 1201328] Share widget not working on mobile

alexgibson:


jpetto:
Looks good to me. I suppose we could have a larger discussion regarding implementing both hover and click handling, but this is a solid fix for a real problem, so good to merge now.

r+ :point_up_2:

PR #3275, opened 2015-09-03
Bug 1201198 Add back commiter agreement redirects

sgarrity:
These were unintentionally removed in 769f05dd60334f54182f6cfdb91f6cff13a851ac


schalkneethling:
r+ :cake:

PR #3274, opened 2015-09-03
[fix bug 1201467] MPL 2.0 FAQ template not found

alexgibson:
- rename template directories for /MPL/ to lowercase for consistency

PR #3276, opened 2015-09-03
Fix Bug 1201196, fix broken link to text file for MPL 1.1

schalkneethling:
@malena r?

### DIFF
--- a/bedrock/mozorg/templates/mozorg/mpl/historical.html
+++ b/bedrock/mozorg/templates/mozorg/mpl/historical.html
@@ -21,27 +21,27 @@
         <p>These files are made available for historical and reference purposes only, and should not be used to license software.</p>
         <dl>
             <dt><a href="{{ url('mozorg.mpl.1.1.index') }}">Mozilla Public License, version 1.1</a></dt>
-            <dd>This is the Mozilla Public License, version 1.1. It is also available in <a href="index.txt">plain text</a>. MPL 1.1 is obsoleted by MPL 2.0, and Mozilla strongly recommends against using this license for new code.</dd>
+            <dd>This is the Mozilla Public License, version 1.1. It is also available in <a href="{{ static('MPL/1.1/index.txt') }}">plain text</a>. MPL 1.1 is obsoleted by MPL 2.0, and Mozilla strongly recommends against using this license for new code.</dd>
 
             <dt><a href="{{ url('mozorg.mpl.1.1.faq') }}">MPL version 1.1 FAQ</a></dt>
             <dd>This is the FAQ for MPL 1.1. Because MPL 1.1 has been replaced by MPL 2.0, this FAQ should be used for historical reference purposes only.</dd>
 
-            <dt><a href="NPL/1.1/">Netscape Public License, version 1.1</a></dt>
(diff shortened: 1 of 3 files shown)


alexgibson:
Can we please merge this one after #3274 to avoid rebase issues?

malena:
There is actually broken links on two more pages that would need the same edits.
mpl/index.html (one instance)
mpl/historical.html (two instances)

schalkneethling:
@malena updated

bedrock/mozorg/templates/mozorg/mpl/historical.html
```diff
@@ -36,10 +36,10 @@
             <dd>Since legal documents are difficult for non-lawyers to read, Mozilla has in the past provided a version of the license which contains informal explanations of various sections of the license, in the form of hyperlinks.  Note that <em>these explanations are not the license.</em> The explanations are not a legal document, or legal advice. If you need to know <em>exactly</em> what the license requires, you need to read and understand the license itself; if you need legal advice, you need to talk to a lawyer. The annotations are merely a good faith effort to explain the license in simpler language.</dd>
 
             <dt><a href="NPL/1.0/">Netscape Public License, version 1.0</a></dt>
-            <dd>This is the license under which the Mozilla source code was initially released. It is also available in <a href="NPL/1.0/index.txt">plain text</a>. The NPL is obsolete and should not be used.</dd>
+            <dd>This is the license under which the Mozilla source code was initially released. It is also available in <a href="http://website-archive.mozilla.org/www.mozilla.org/mpl/MPL/1.0/index.txt">plain text</a>. The NPL is obsolete and should not be used.</dd>
```
malena: This is referencing NPL/1.0/index.txt not MPL/1.0/index.txt.
However since NPL is archived wondering if the url would be similar except point to NPL instead?

bedrock/mozorg/templates/mozorg/mpl/historical.html
```diff
@@ -36,10 +36,10 @@
             <dd>Since legal documents are difficult for non-lawyers to read, Mozilla has in the past provided a version of the license which contains informal explanations of various sections of the license, in the form of hyperlinks.  Note that <em>these explanations are not the license.</em> The explanations are not a legal document, or legal advice. If you need to know <em>exactly</em> what the license requires, you need to read and understand the license itself; if you need legal advice, you need to talk to a lawyer. The annotations are merely a good faith effort to explain the license in simpler language.</dd>
 
             <dt><a href="NPL/1.0/">Netscape Public License, version 1.0</a></dt>
```
malena: Not sure if these would need to be updated to archived url or if the redirect would work? Maybe no change is needed but I was a little confused about this when I updated these docs.

schalkneethling: I do not see a rewrite rule for NPL/1.0 here https://github.com/mozilla/bedrock/blob/master/etc/httpd/global.conf#L597 Does the link work on production?

schalkneethling: Nope, these do not seem to work on production. I reckon I will fix the few MPL links and once we have done a push to prod to get all those new redirect rules in, we an revisit these.

schalkneethling: Actually, I am going to fix those URLs based on the new rewrite rules because these changes will most likely be pushed to prod along with the rewrite rules.

malena:
r+

PR #3278, opened 2015-09-03
Bug 1136436 - add waffle switch for iOS product page

craigcook:
Switch name is 'ios-active'


jpetto:
Looks good! r+ :panda_face:

PR #3283, opened 2015-09-04
Bug 1137724: Move iOS newsletter to interests section.

pmclanahan:


pmclanahan:
@kyoshino r? Looks like that's all that's needed right?

kyoshino:
:+1: on my local server. The position, that should probably be under Firefox OS as requested, can be controlled by Basket.

screen shot 2015-09-10 at 13 57 46

pmclanahan:
Thanks Kohei!

PR #3290, opened 2015-09-08
Bug 1201914 Fix MPL URLs that include .html suffix

sgarrity:
Also fix the order of the rules, and add a missing archive redirect for missing.html (get it?).

### DIFF
--- a/etc/httpd/global.conf
+++ b/etc/httpd/global.conf
@@ -875,10 +875,10 @@ RewriteRule ^/media/flash/playerWithControls.swf - [F]
 # Bug 896474
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?about(/?|/.+)$ /b/$1about$2 [PT]
 
-# Bug 987852
-RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL(/?|/.+)$ /b/$1MPL$2 [PT]
-RewriteRule ^/MPL/1.1/index\.txt /media/MPL/1.1/index.txt [L,PT]
+# Bug 987852 & 1201914
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/(.+).html$ /$1MPL/$2/ [L,R=301]
 RewriteRule ^/MPL/2.0/index\.txt /media/MPL/2.0/index.txt [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL(/?|/.+)$ /b/$1MPL$2 [PT]
 
 # Bug 1040970
 RewriteRule ^/mozillacareers$ https://wiki.mozilla.org/People/mozillacareers?utm_medium=redirect&utm_source=mozillacareers-vanity [L,R=301]
--- a/etc/httpd/global.conf
+++ b/etc/httpd/global.conf
@@ -596,6 +596,7 @@ RewriteRule ^/MPL/1\.0/(.*) http://website-archive.mozilla.org/www.mozilla.org/m
 RewriteRule ^/MPL/2\.0/process/(.*) http://website-archive.mozilla.org/www.mozilla.org/mpl/MPL/2\.0/process/$1 [L,R=301]
 RewriteRule ^/MPL/NPL/(.*) http://website-archive.mozilla.org/www.mozilla.org/mpl/MPL/NPL/$1 [L,R=301]
 RewriteRule ^/MPL/boilerplate-1\.1/(.*) http://website-archive.mozilla.org/www.mozilla.org/mpl/MPL/boilerplate-1\.1/$1 [L,R=301]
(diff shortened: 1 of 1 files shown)


etc/httpd/global.conf
```diff
@@ -875,10 +876,18 @@ RewriteRule ^/media/flash/playerWithControls.swf - [F]
...
+# Bug 987852 & 1201914
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/1.1/FAQ.html /$1/MPL/1.1/FAQ/ [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/2.0/combining-mpl-and-gpl.html /$1/MPL/2.0/combining-mpl-and-gpl/ [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/2.0/differences.html /$1/MPL/2.0/differences/ [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/2.0/FAQ.html /$1/MPL/2.0/FAQ/ [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/2.0/permissive-code-into-mpl.html /$1/MPL/2.0/permissive-code-into-mpl/ [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/2.0/Revision-FAQ.html /$1/MPL/2.0/Revision-FAQ/ [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/historical.html /$1/MPL/historical/ [L,PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/license-policy.html /$1/MPL/license-policy/ [L,PT]
+RewriteRule ^/MPL/2.0/index\.txt /media/MPL/2.0/index.txt [L,PT]
```
pmclanahan: This line appears to be a dupe of the line below it.

sgarrity: > This line appears to be a dupe of the line below it.

Fixed.

etc/httpd/global.conf
```diff
@@ -875,10 +876,18 @@ RewriteRule ^/media/flash/playerWithControls.swf - [F]
 # Bug 896474
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?about(/?|/.+)$ /b/$1about$2 [PT]
 
-# Bug 987852
-RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL(/?|/.+)$ /b/$1MPL$2 [PT]
-RewriteRule ^/MPL/1.1/index\.txt /media/MPL/1.1/index.txt [L,PT]
+# Bug 987852 & 1201914
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/1.1/FAQ.html /$1/MPL/1.1/FAQ/ [L,PT]
```
pmclanahan: These ".html" urls should be redirects, not pass-throughs. Are there no longer any URLs with ".html" at all? Alternately are these the full list? If it is, it looks like we can pretty much just do a single rule for this:

```apache
RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/(.+).html$ /$1/MPL/$2/ [L,R=301]
```

sgarrity: > These ".html" urls should be redirects, not pass-throughs. Are there no longer any URLs with ".html" at all? Alternately are these the full list? If it is, it looks like we can pretty much just do a single rule for this:

Good idea. I've updated with this single rule. The only other `.html` urls are redirects to the website-archive that happen higher up in the file.

I left it as a separate commit, but will squash once we're all set.

etc/httpd/global.conf
```diff
@@ -875,10 +876,10 @@ RewriteRule ^/media/flash/playerWithControls.swf - [F]
 # Bug 896474
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?about(/?|/.+)$ /b/$1about$2 [PT]
 
-# Bug 987852
-RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL(/?|/.+)$ /b/$1MPL$2 [PT]
-RewriteRule ^/MPL/1.1/index\.txt /media/MPL/1.1/index.txt [L,PT]
+# Bug 987852 & 1201914
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?MPL/(.+).html$ /$1/MPL/$2/ [L,R=301]
```
pmclanahan: Oops. This is probably my fault. Shouldn't be a slash between `$1` and `MPL` in the destination: `/$1MPL/$2/`.

sgarrity: > Oops. This is probably my fault. Shouldn't be a slash between $1 and MPL in the destination: /$1MPL/$2/.

Fixed & squashed.

PR #3291, opened 2015-09-08
[fix bug 1198516] Implement send to device tests on /firefox/android

jpetto:


bedrock/firefox/views.py
```diff
@@ -655,3 +666,11 @@ def get_template_names(self):
             template = 'firefox/feedback/unhappy.html'
 
         return [template]
+
+
+def android(request):
+    # check for variant in querystring for send-to-device testing
+    variant = request.GET.get('v', '')
```
alexgibson: Is this a possible xss vector seeing as it gets embedded in the template? Would it be better to limit the possible values to either `a`, `b`, or `c`?

media/js/firefox/android.js
```diff
@@ -283,14 +283,85 @@
...
+            $('#send-to-device').removeClass('logo').appendTo($sendToDeviceEmbeddedWidget);
+
+            // show the widget
+            $sendToDeviceEmbeddedWidget.removeClass('send-to-device-variant');
+
+            // update nav & footer buttons to scroll to embedded widget
+            $sendToLinks.on('click', function(e) {
+                e.preventDefault();
+
+                Mozilla.smoothScroll({
```
alexgibson: Nice! :+1:

media/css/firefox/android.less
```diff
@@ -583,11 +583,130 @@ html[dir="rtl"] #phone-wrapper figure {
...
+        display: none;
+    }
+
+    #send-to-device-embedded-widget #send-to-device {
+        display: block;
+        background: #005189;
+        border-radius: 6px;
+        color: #fff;
+        width: auto;
+        margin-right: 20px;
```
alexgibson: Nit: `@gridGutterWidth`?

media/css/firefox/android.less
```diff
@@ -583,11 +583,130 @@ html[dir="rtl"] #phone-wrapper figure {
...
+
+        // largely appropriated from send-to-device.less
+        h2 {
+            max-width: none;
+            .font-size(24px);
+            padding-bottom: 20px;
```
alexgibson: Nit: `@baseLine`?

alexgibson:
Aside from one question about the query param being embedded in the template, and a couple of nits - this looks good to me. Tested locally and seems to be working well. r+wc on the front end :+1: 

Would appreciate if @pmclanahan could take a look at the python bits (in particular the basket ID passing?).

pmclanahan:
*looks* :eyes:

bedrock/firefox/views.py
pmclanahan: Yes. Was about to say the same. We need to sanitize this input to only accept the values we expect.

bedrock/firefox/views.py
```diff
@@ -215,9 +219,16 @@ def send_to_device_ajax(request):
     data_type = 'email' if '@' in phone_or_email else 'number'
     data[data_type] = phone_or_email
     form = SendToDeviceWidgetForm(data)
+
     if form.is_valid():
         phone_or_email = form.cleaned_data.get(data_type)
         platform = form.cleaned_data.get('platform')
+
+        # check for android and send to device test and update message if present
+        send_to_device_test = request.POST.get('android-send-to-device-test')
```
pmclanahan: This is unsaitized user-provided data. Should be validated to ensure it's one of the values we expect.

bedrock/firefox/views.py
jpetto: Good catch. Will add a check for 3 acceptable values.

jpetto:
@alexgibson @pmclanahan Updated. Take one more look?

alexgibson:
Looks good to me, but I'll defer to @pmclanahan for the final word :+1:

pmclanahan:
Nice work Jon! r+ :rocket:

pmclanahan:
Oops. You probably wanted to squarsh that. Oh well. Apologies.

PR #3293, opened 2015-09-08
Add mgmnt command to download Maxmind DB

pmclanahan:


pmclanahan:
@glogiotatidis r?

The only hiccup is that if we want to use this for the GeoIP-Lite DB directly from Maxmind, it's a gzip download, so we'd have to handle gunzipping it. I don't think we need to spend time on that right now, but it's something we could do in future. But this should work for now for grabbing the real DB. Now to get it updating on CI and uploading.

bedrock/base/management/commands/get_geoip_db.py
```diff
@@ -0,0 +1,25 @@
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+
+import requests
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        db_url = settings.MAXMIND_DB_URL
+        if db_url is None:
```
glogiotatidis: nit: I'd go with the more generic `if db_url` just in case db_url is set to empty string for example.

glogiotatidis:
> The only hiccup is that if we want to use this for the GeoIP-Lite DB directly from Maxmind, it's a gzip download, so we'd have to handle gunzipping it. I don't think we need to spend time on that right now, but it's something we could do in future. 

+1

> But this should work for now for grabbing the real DB. Now to get it updating on CI and uploading.

I don't remember whether we agreed that I do the jenkins job or you'll do it. Happy to take over just send me the maxmind scripts and url. 

r+ :art:

glogiotatidis:
I created the jenkins script to upload the file to S3, including the needed IAM roles and buckets. See #3298 

I just need to include the maxmind scripts to download the db.

glogiotatidis:
Maybe you want to call the script [here](https://github.com/mozilla/bedrock/blob/master/docker/run-prod.sh#L4) to complete this in one pull request?

pmclanahan:
> Maybe you want to call the script here to complete this in one pull request?

Yes. Good call. Adding.

pmclanahan:
This is the script to get the DB from Maxmind: https://github.com/maxmind/geoipupdate

We just need to install that on the jenkins box. I have the config file with our credentials.

pmclanahan:
@glogiotatidis I had an idea last night. We already get and update this file on the admin node in our DC. Can we just throw a script on there for now to upload this file to s3 and worry with Jenkins later?

pmclanahan:
Just leaving a note here: Giorgos and I discussed the above, and it's already working great in Jenkins. Case closed. All that's left is to merge this and add the env vars to Deis. 

@glogiotatidis sound right? Merge this if you think it's ready.

bedrock/base/management/commands/get_geoip_db.py
```diff
@@ -0,0 +1,25 @@
...
+            raise CommandError('Request timed out: ' + db_url)
+        except requests.RequestException as e:
+            raise CommandError(str(e))
+
+        if resp.status_code == 200:
+            with open(settings.MAXMIND_DB_PATH, 'wb') as geoip_file:
+                for chunk in resp:
+                    geoip_file.write(chunk)
+
+        self.stderr.write('GeoIP file downloaded successfully.')
```
glogiotatidis: Why not `stdout` or event `logging`? Since this is not an error it shouldn't go to stderr.

bedrock/base/management/commands/get_geoip_db.py
```diff
@@ -0,0 +1,25 @@
...
+from django.core.management.base import BaseCommand, CommandError
+
+import requests
+
+
+class Command(BaseCommand):
+    def handle(self, *args, **options):
+        db_url = settings.MAXMIND_DB_URL
+        if not db_url:
+            return
```
glogiotatidis: Maybe we want to add some `logging.debug` here to say that we skipped downloading

glogiotatidis:
Apart for the logging comments, I tested this locally and works as expected. I also updated `bedrock` deis app with the needed env vars so this will just work when merged.

Nice job! :dart:

pmclanahan:
Logging updated. Thanks!

bensternthal:
FYI, I noticed this morning bedrock does not appear to be working.

http://bedrock-dev.eu-west.moz.works/
Results in 502.

Unknown if related I see many of these in papertrail.
```IOError: [Errno 13] Permission denied: '/GeoIP2-Country.mmdb'```

pmclanahan:
Yeah. Stuff seems broken in both regions. Not sure what the deal is yet.

glogiotatidis:
Default maxmind save path is `../maxmind.db` in which the docker user did not have write permissions. I added `MAXMIND_DB_PATH` env variable to the eu cluster and things are fixed there. us cluster fails for another reason atm.

PR #3294, opened 2015-09-08
Fix bug 1202877 - Fx iOS available in Australia and Austria

craigcook:


craigcook:
@schalkneethling r?

There are a few Aussies and Bavarians we can ask to verify on prod after we push.

schalkneethling:
r+ :sheep:

PR #3295, opened 2015-09-09
Add a Procfile and clock process.

pmclanahan:
* Add jobs via APScheduler to mimic ones currently running via cron.


pmclanahan:
@glogiotatidis This works well. I don't yet like the output from the clock proc (or lack thereof). I think it's a problem with bedrock's logging config, but I've yet to figure out the problem. We've got some logging config to do anyway for papertrail I think, so I think we can fix that part later.

glogiotatidis:
Looks and works good r+. Re logging maybe it's useful to add a scheduler event listener class to log when commands run independently of the output of the actual commands. 

https://apscheduler.readthedocs.org/en/latest/userguide.html#scheduler-events

pmclanahan:
Added event listener.

pmclanahan:
Updated once more to avoid removing `git` from the image in the Dockerfile so that `update_security_advisories` can use it.

glogiotatidis:
Event listener looks OK. I still can't figure the logging issue. Maybe @jgmize has any idea why `logging.info()` in jobs does not work?

glogiotatidis:
Since we moved external files to db we probably want to add a clock entry?

See also #3325

pmclanahan:
Yes. Have other things to fix as well. Will do that.

scripts/clock.py
```diff
@@ -0,0 +1,92 @@
...
+            raise
+        else:
+            self.log('finished successfully')
+
+    def log(self, message):
+        print('Clock job {0}: {1}'.format(self.name, message))
+
+
+@scheduled_job('interval', seconds=5)
+def job_should_pass():
```
pmclanahan: This one and the next one are obviously not for prod. They should help with testing what things look like in papertrail and whether things work or not at all :)

pmclanahan:
@jgmize this should be good for testing. All of the changes to add the virtualenv are in a single commit, so it should be decently easy to push a non-venv branch for testing both ways.

jgmize:
Unfortunately this approach is currently blocked by https://github.com/deis/deis/issues/4605 but there is an open PR https://github.com/deis/deis/pull/4606 that may resolve it soon. In the meantime I've filed https://github.com/mozilla/bedrock/pull/3422 as a replacement for this approach.

pmclanahan:
So I've modified this so that we get the extra logging for each job that I added from before, but it just runs all of the jobs in sequence. I then just call the `./manage.py runscript clock` from the jenkins script added in #3422. We should then be able to add snitches to papertrail for e.g. `Clock job job_rnasync: finished successfully`.

pmclanahan:
NM. Can't reopen this apparently. Opening new PR.

PR #3297, opened 2015-09-09
[bug 1190614] Update /firefox/new/2/ pass through for all locales

alexgibson:


alexgibson:
@jpetto r?

jpetto:
This looks right to me. Seems the old `en-US` only rule was in use during the PHP to bedrock migration, and is obviated by the [more generalized fix for bug 1044352](https://github.com/mozilla/bedrock/blob/master/etc/httpd/global.conf#L252).

Apologies for missing this the first time around.

r+

alexgibson:
> Apologies for missing this the first time around.

I should have picked up on it during code review too :anguished:

PR #3299, opened 2015-09-09
Add newrelic.ini for docker.

glogiotatidis:
The idea is that we populate `NEW_RELIC_APP_NAME` and `NEW_RELIC_LICENSE_KEY` in the environment and the NR agent grabs the data from there. 

I believe that webops have a different approach with their own `newrelic.ini` file and we continue to respect that here as long as webops don't store their `newrelic.ini` in the root of the repo.

- I tested that on a demo app on deis and works fine
- Added NR environment variables on both clusters for `bedrock-dev` app. `NEW_RELIC_APP_NAME=bedrock-deis-dev`

@pmclanahan r?

### DIFF
--- a/wsgi/playdoh.wsgi
+++ b/wsgi/playdoh.wsgi
@@ -1,6 +1,8 @@
 # flake8: noqa
-
 import os
+
+from decouple import config
+
 
 try:
     import newrelic.agent
@@ -9,7 +11,7 @@ except ImportError:
 
 
 if newrelic:
-    newrelic_ini = os.getenv('NEWRELIC_PYTHON_INI_FILE', False)
+    newrelic_ini = config('NEWRELIC_PYTHON_INI_FILE', default=False)
     if newrelic_ini:
         newrelic.agent.initialize(newrelic_ini)
     else:
--- a/newrelic.ini
+++ b/newrelic.ini
@@ -1,0 +1,210 @@
+# ---------------------------------------------------------------------------
+
+#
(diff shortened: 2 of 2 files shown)


wsgi/playdoh.wsgi
```diff
@@ -9,7 +11,7 @@ except ImportError:
 
 
 if newrelic:
-    newrelic_ini = os.getenv('NEWRELIC_PYTHON_INI_FILE', False)
```
pmclanahan: Are you sure this env var isn't used anywhere?

glogiotatidis: good catch, updated back to the original value

wsgi/playdoh.wsgi
```diff
@@ -9,7 +11,7 @@ except ImportError:
...
+    newrelic_ini = config('NEWRELIC_PYTHON_INI_FILE', default='newrelic.ini')
```
pmclanahan: The default was `False` before so that not defining the Env var would effectively turn NR off w/o having to uninstall the lib. Do you think that's valuable? We can leave it like this, but I also think it's fine to require the env var for NR to work.

glogiotatidis: Agree that changing default can cause errors in this double deployment scenario we're in. I changed it to default to `False` again and updated the `bedrock-dev` apps with the needed env variables.

pmclanahan:
Great! Thanks!

PR #3300, opened 2015-09-10
Run tests against the built docker image.

glogiotatidis:
This runs only `./manage.py test`. I see in `travis.yml` that we run `grunt test` and pre-script linters like `flake8`, `runscript check_calendars`.

We probably want to run all these in docker too.  Thoughts @pmclanahan ?

### DIFF
--- a/bedrock/base/tests/test_accepted_locales.py
+++ b/bedrock/base/tests/test_accepted_locales.py
@@ -41,7 +41,7 @@ class AcceptedLocalesTest(test_utils.TestCase):
         cls.PROD_LANGUAGES = settings.PROD_LANGUAGES
         cls.DEV_LANGUAGES = settings.DEV_LANGUAGES
         settings.PROD_LANGUAGES = ('en-US',)
-        os.rename(cls.locale, cls.locale_bkp)
+        shutil.move(cls.locale, cls.locale_bkp)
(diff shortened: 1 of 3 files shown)


glogiotatidis:
@pmclanahan r?

pmclanahan:
I'm actually fine with only running the django tests in the docker container. Those other tests are purely for the dev process IMHO. The things that we're trying to catch in the test run in the docker image are build errors, so I don't think that running the JS tests again gets us much.

glogiotatidis:
As longs as we keep travis around this works. Removing the WIP tag, feel free to review and merge.

pmclanahan:
Definitely not planning on getting rid of CI for testing master branch and PRs. Whether it'll be Travis or something else I can't say, but we'll definitely have something.

pmclanahan:
I don't see any dependency changes, so does this mean that the docker image already gets the dev dependencies so that the tests will run?

glogiotatidis:
> I don't see any dependency changes, so does this mean that the docker image already gets the dev dependencies so that the tests will run?

Yes https://github.com/mozilla/bedrock/blob/master/Dockerfile#L20
You can also check the test job I run to verify that this is working https://ci.us-west.moz.works/job/mozilla_bedrock_giorgos/7/

PR #3302, opened 2015-09-10
Bug 686489 new download fennec apk download page

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=686489


bedrock/firefox/templates/firefox/partners/index.html
```diff
@@ -459,7 +459,12 @@ <h2 class="tween">
         <div class="intro tween">
           {% if l10n_has_tag('mwc_2015') %}
             <h4>{{ _('A superior mobile experience') }}</h4>
-            <a href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}"><img src="{{ static('img/firefox/partners/google-play.png') }}" alt="{{ _('Get it on Google Play') }}" class="play" /></a>
+            <div class="play-button"> 
+              <a href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}"><img src="{{ static('img/firefox/partners/google-play.png') }}" alt="{{ _('Get it on Google Play') }}" class="play" /></a>
+              <small class="download-other">
+                <a href="{{ url('firefox.android.all') }}">{{_('Systems &amp; Languages')}}</a>
+              </small>
```
schalkneethling: The color contrast here is not good and is generally hard to read. 

screen shot 2015-09-14 at 12 03 22

I would suggest we make it:

```
.partner-article.dark a {
    color: white;
    font-size: .75rem;
}
```
screen shot 2015-09-14 at 12 05 57

bedrock/firefox/templates/firefox/personal.html
```diff
@@ -37,6 +37,9 @@
     <a rel="external" href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}">
       {{ high_res_img('firefox/android/btn-google-play.png', {'alt': 'Get it on Google Play', 'width': '172', 'l10n': True}) }}
     </a>
+    <small class="download-other">
+      <a href="{{ url('firefox.android.all') }}">{{_('Systems &amp; Languages')}}</a>
+    </small>
```
schalkneethling: Here also, contrast and legibility can be improved:

screen shot 2015-09-14 at 12 07 13

Here I would suggest:

```
.download-other {
    a {
        color: #fff;
        .open-sans-light();
        font-size: .95rem;
    }
}
```
screen shot 2015-09-14 at 12 11 59

media/css/firefox/android/all.less
```diff
@@ -0,0 +1,100 @@
...
+            }
+        }
+        .download {
+            padding: 10px 0;
+            a {
+                display: block;
+                padding: 5px 5px 5px 35px;
+                background-position: left center;
+                background-repeat: no-repeat;
+                background-image: url(/media/img/firefox/android/android-download.svg);
```
schalkneethling: You could here do:

```
background: transparent url(/media/img/firefox/android/android-download.svg) left center no-repeat;
```
Are we concerned with IE8 support on this page? The reason I ask is because IE8 does not support SVG :-/

media/css/firefox/partners.less
```diff
@@ -493,9 +493,17 @@
...
+    #android-overview .intro .play-button {
         display: block;
         margin: 5px 0 5px 20px;
+        float: right;
+        .download-other {
+            text-align: center;
+            clear: both;    
+            display: block;
+            .open-sans();
+            .font-size(11px);
```
schalkneethling: Nit: I would rearrange the rules above as:

``` 
clear: both;    
display: block;
.open-sans();
.font-size(11px);
text-align: center;
```

bedrock/firefox/templates/firefox/android/all.html
```diff
@@ -0,0 +1,59 @@
...
+  <main id="main-content" role="main">
+    <div class="build-table-container">
+      <table class="table">
+        <caption>
+          <h3>{{ _('APK Versions') }}</h3>
+        </caption>
+        <thead>
+          <tr>
+            <th>{{ _('Version') }}</th>
+            <th>{{ _('Download') }}</th>
```
schalkneethling: For a11y I would add `scope="col"`to the above two headers.

schalkneethling:
Do we need an L10N flag here? @flodolo

flodolo:
No. We don't localize the /all pages, and all other templates use an existing `Systems &amp; Languages` string.

bedrock/firefox/templates/firefox/partners/index.html
malena: Wondering about consistency. The other link is yellow. Changing it to white would make it look like plain text no? I was sticking to the current link theme in that section. Should be changing that?

screenshot 2015-09-14 15 34 40

bedrock/firefox/templates/firefox/personal.html
malena: Yes the contrast for this one is pretty dark. But not sure white would be okay because that is the colour of text? How about the link colour used in https://www.mozilla.org/en-US/firefox/os/2.0/. It's slightly off white.

screenshot 2015-09-14 15 46 42

media/css/firefox/android/all.less
malena: There could be cases where a person is seeing this page on their desktop on ie8, but not sure if that would be a concern. I can double check with @jgmize? Changing graphic wouldn't be a problem either.

alexgibson: Could include a png fallback for the svg? https://css-tricks.com/a-complete-guide-to-svg-fallbacks/

schalkneethling: Alternatively, could we use the button that is used on the firefox/partners page perhaps?

bedrock/firefox/templates/firefox/personal.html
schalkneethling: So, the off white is fine, as long as we bump the font size as well. I have to say that the links under the normal download button personally, is to small as well and the line-height is much to tight but, this is up for an overhaul so, I will leave that as is for now :)

I would ideally here go for white, bump the size a little and, if there is concern that it will not seem like a link (although I believe we are safe), we can give it an underline by default. Thoughts?

bedrock/firefox/templates/firefox/partners/index.html
schalkneethling: Good catch, that will make things confusing. Also, after some additional testing, turns out this blue is just a terrible background color ;) The contrast here is definitely way to low and the legibility is not good at all but, let's leave that as is for now, and I will open a bug to explore how we can resolve this.

schalkneethling: I have found a winning combination for this that I will propose in a new bug for discussion.

malena: Okay will leave this as is for now.

bedrock/firefox/templates/firefox/personal.html
malena: Okay I've increased the font size as suggested in earlier thread above to .75rem and updated the colour as well. Should be better to read now. I agree the links under the download buttons are a little small ><.

media/css/firefox/android/all.less
malena: I've added a css fallback to png version.

media/css/firefox/android/all.less
```diff
@@ -0,0 +1,99 @@
...
+                padding: 10px 0;
+            }
+        }
+        .download {
+            padding: 10px 0;
+            a {
+                display: block;
+                padding: 5px 5px 5px 35px;
+                background: transparent url(/media/img/firefox/android/android-download.png) left center no-repeat;
+                background: transparent url(/media/img/firefox/android/android-download.svg) left center no-repeat;
```
alexgibson: Might some browsers download the SVG here and still not support it? I think there's a trick here to use https://css-tricks.com/a-complete-guide-to-svg-fallbacks/

malena: Okay I added the css fallback for svg.

schalkneethling:
Looks good @malena Care to squash?

schalkneethling:
r+ :snowman:

PR #3305, opened 2015-09-10
Bug 1203589 - 'Download a fresh copy' should link to firefox/all/#loc…

pascalchevrel:
…alecode


alexgibson:
At smaller viewport sizes the localized language text could do with some padding to match the non localized text:

lang-padding

alexgibson:
Other than one small nit, I think this is a great improvement for our visitors. Nice work @pascalchevrel!

pascalchevrel:
Thanks @alexgibson, I have updated the PR to add padding to localized language names on smaller viewports

alexgibson:
Ah, I almost forgot - we're currently running a performance A/B test on the /firefox/new/ page. @pascalchevrel could you please also update [this template](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/new2.html) so we don't accidentally lose this change once the test is over? Thanks again.

pascalchevrel:
new2.html updated

alexgibson:
r+ :cake:

PR #3310, opened 2015-09-11
[fix bug 1199738] Update FxOS device availability.

jpetto:


flodolo:
No need to wait for l10n, I'm tempted to leave the country names around in case they come back. Tagging in the meantime so I won't forget about it.

media/css/firefox/os/get_device.less
```diff
@@ -97,102 +102,95 @@
...
+    }
+    .intex {
+        background: transparent url('/media/img/firefox/os/logo/intex.png') top center no-repeat;
+    }
+    .jbhifi {
+        background: transparent url('/media/img/firefox/os/logo/jbhifi.png') top center no-repeat;
     }
     .mtn {
         background: transparent url('/media/img/firefox/os/logo/mtn.png') top center no-repeat;
     }
```
schalkneethling: Should we remove this as MTN is no longer a provider?

schalkneethling:
r+ /wc :snake:

alexgibson:
Now that we do server-side geo location for many of the countries here, does this list also need to be updated? https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L805-L842

Are the phones still available in those countries (just not online), or are they no longer available to buy period?

jpetto:
Hm, I don't believe this update should dictate changes to the list in `base.py`. Based on a [comment in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1199738#c7), the phones may still be available on shelves (even if not available online).

media/css/firefox/os/get_device.less
jpetto: Well, none of the providers in the commented out list are currently active, but I figured it best to keep them around for easy re-activation should they be needed. It would be a shame if we got an MTN link in a couple months and had to re-create the class.

Or are you saying MTN is no longer in the cell service/phone business?

PR #3313, opened 2015-09-14
[fix bug 1203746] Add Romania availabiliy to Orange Klif.

jpetto:


alexgibson:
r+ just waiting on travis :basketball:

PR #3314, opened 2015-09-15
firefox/menu.less was moved to base/menu.less and oldIE.css still ref…

pascalchevrel:
…erences old path


alexgibson:
r+ thanks @pascalchevrel!

PR #3316, opened 2015-09-15
[fix bug 1199258] Redirect fxa signin users on fx40 firstrun.

jpetto:
Need to wait until FxA prod is updated (possibly 9/16?) before merging, but should be reviewed and tested on demo4 now.

https://www-demo4.allizom.org/en-US/firefox/40.0/firstrun/

Instructions for testing FxA staging can be found [here](https://gist.github.com/jpetto/f46f1850737b7a838cad).

Summary of problem/fix:

Currently, users signing in to FxA at `/firefox/40.0/firstrun/` get a less than ideal experience, as the links displayed post-login are contained within the iframe. Our short term fix is to immediately redirect users that sign in to their FxA account settings.

This fix requires a change on the FxA side to force the sign in to complete immediately, as opposed to waiting for an XHR request. Without the FxA change, a redirect will invalidate the sign in.

This PR, tested against FxA stage, should redirect users signing in to FxA to the FxA settings screen immediately with no errors, keeping them signed in to FxA.

### DIFF
--- a/bedrock/firefox/templates/firefox/australis/fx40/firstrun.html
+++ b/bedrock/firefox/templates/firefox/australis/fx40/firstrun.html
@@ -69,7 +69,7 @@
     </div>
 
     <div class="rightcol">
-      <iframe id="fxa" scrolling="no" data-src="{{ settings.FXA_IFRAME_SRC }}?utm_campaign=fxa-embedded-form&amp;utm_medium=referral&amp;utm_source=firstrun&amp;utm_content=fx-{{ version }}&amp;entrypoint=firstrun&amp;service=sync&amp;context=iframe&amp;style=chromeless"></iframe>
+      <iframe id="fxa" scrolling="no" data-src="{{ settings.FXA_IFRAME_SRC }}?utm_campaign=fxa-embedded-form&amp;utm_medium=referral&amp;utm_source=firstrun&amp;utm_content=fx-{{ version }}&amp;entrypoint=firstrun&amp;service=sync&amp;context=iframe&amp;style=chromeless&amp;haltAfterSignIn=true"></iframe>
     </div>
   </div>
 </section>
--- a/media/js/firefox/australis/fx40/firstrun.js
+++ b/media/js/firefox/australis/fx40/firstrun.js
@@ -56,12 +56,14 @@
                 });
 
                 break;
-            // track user with FxA account logging in (very rare for firstrun)
+            // track & redirect user with FxA account logging in
             case 'login':
                 window.dataLayer.push({
                     'event': 'firstrun-fxa',
                     'interaction': 'fxa-login'
                 });
+
+                window.location.href = fxaIframeSrc + '/settings';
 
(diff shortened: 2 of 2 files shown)


jpetto:
@alexgibson r?

alexgibson:
Code-wise this all looks good. I signed up for a new Firefox Account on stage, and created a new profile described above using Firefox 40. Signing into the iframe on demo4 resulted in me being redirected as expected, so r+ there :+1: 

One thing I did encounter, which may be another related issue - is that once signed into the account, viewing /firstrun again displays the "Manage Account" section within the iframe again and no redirect happens. Not entirely sure if this is intended behavior or not? (see screenshot below). It also seems impossible to log back out again here (disconnecting Sync in Preferences isn't enough). The only option seems to be to delete the account? :/

signed-in

shane-tomlinson:
> One thing I did encounter, which may be another related issue - is that once signed into the account, viewing /firstrun again displays the "Manage Account" section within the iframe again and no redirect happens. Not entirely sure if this is intended behavior or not? (see screenshot below). 

We currently provide no facility to completely disable settings, the hope is since this is the firstrun flow, few users should see the page.

> It also seems impossible to log back out again here (disconnecting Sync in Preferences isn't enough). The only option seems to be to delete the account? :/

This is intended behavior. If a user signs into FxA for Sync, then signs out of FxA, they also kill the session token that Sync uses, effectively signing them out of Sync too.

We are planning a protocol for the browser to notify FxA if the user signs out from the browser's preferences panel; this outcome is a disconnect in the Sync preferences will cause FxA to believe the user is signed out too.

shane-tomlinson:
@jpetto - code looks good, and you are right, wait until FxA train 45 is out the door to push this, or else the Unexpected Error message will appear again.

shane-tomlinson:
> We are planning a protocol for the browser to notify FxA if the user signs out from the browser's preferences panel; this outcome is a disconnect in the Sync preferences will cause FxA to believe the user is signed out too.

I forgot a reference - https://github.com/mozilla/fxa-content-server/issues/2917#issuecomment-137590716 The bug mentions fennec, but the functionality will be available to all Fx variants.

jpetto:
Since it looks like we're good here, I'm going to take this branch off demo4 in favor of #3318.

jgmize:
> @jpetto - code looks good, and you are right, wait until FxA train 45 is out the door to push this, or else the Unexpected Error message will appear again.

@shane-tomlinson would you mind dropping a link here to the FxA train schedule?

shane-tomlinson:
> @shane-tomlinson would you mind dropping a link here to the FxA train schedule?

The train schedule follows a 2 week cadence, the release date is variable. Here is The FxA schedule as a whole - https://www.google.com/calendar/embed?src=mozilla.com_urbkla6jvphpk1t8adi5c12kic%40group.calendar.google.com&ctz=Europe/London

Train 45 was released to production last night, you are good to merge!

PR #3317, opened 2015-09-15
Fix bug 1204652 - Firefox for iOS available in Canada

craigcook:
We should hold off merging this until we're pretty confident we won't need to do any other pushes (though it's also easy to revert if we have to). This update should not push to prod until we get confirmation that FxiOS is available in the Canada App Store.


schalkneethling:
r+

schalkneethling:
:maple_leaf:

PR #3318, opened 2015-09-15
[fix bug 1190614] Implement optimized /firefox/new/.

jpetto:
This PR looks much bigger than it is. Large part of the diff is replacing current `/firefox/new/` files with already merged `/firefox/new/2/` files.

Also removes a few images in `img/firefox/new/` that I couldn't find in use anywhere.

[As requested in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1190614#c44), did another round of testing in Safari to `/firefox/new/`, `/firefox/new/?scene=2`, and `/firefox/new/#download-fx`. All seems well there.


bedrock/firefox/templates/firefox/new.html
```diff
@@ -104,118 +104,85 @@ <h1 class="large">
...
-          {{ share_cta(_('Share'), share_urls, share_text, 'share-firefox-new', 'sky mini') }}
+        <div class="desktop-latest-links-wrapper latest-links-wrapper">
+          <ul>
+            <li class="hide-for-refresh"><a href="{{ url('firefox.desktop.index') }}">{{_('Learn more about Firefox for desktop')}}</a></li>
+            <li><a href="https://support.mozilla.org/products/firefox?utm_source=mozilla.org&amp;utm_medium=referral&amp;utm_campaign=need-help-link">{{_('Need help?')}}</a></li>
+            {% if l10n_has_tag('mobile_links') %}
+            <li class="hide-for-refresh"><a href="{{settings.GOOGLE_PLAY_FIREFOX_LINK }}">{{_('Get Firefox on your Android device')}}</a></li>
+            <li class="hide-for-refresh"><a href="{{ url('firefox.os.index') }}">{{_('Learn about Firefox OS')}}</a></li>
+            {% endif %}
+            <li><a href="{{ url('firefox.all') }}">{{_('Download a fresh copy')}}</a></li>
```
alexgibson: We should add back in `#{{ LANG }}` to the `href` here which was added in #3305

alexgibson:
Given this a test locally and looks good (much of this has already been reviewed). Can't find any references to the removed images so I think we're good there also.

Think it's worth pushing this to a demo for some final testing just to make sure we look good?

media/js/firefox/new.js
```diff
@@ -148,167 +146,114 @@
...
+                if (noScene2) {
                     // if using an unsupported platform just try to drop the URL hash
                     if (window.history && window.history.replaceState) {
                         var uri = window.location.href.split('#')[0];
                         window.history.replaceState({}, '', uri);
                     }
                 } else {
-                    show_scene(2);
+                    showScene(2);
                     // For IE < 11 we supress the auto-download since this
```
alexgibson: We should totally make a bug to do this now that the PHP site is no more? Needn't be part of this PR, but we *should* be able to now limit the popup behavior to IE 6/7/8?

bedrock/firefox/templates/firefox/new.html
jpetto: Ah, missed that in my rebase. Will add.

media/js/firefox/new.js
jpetto: Oooh, yes, great idea. Make that bug sir. :guardsman:

alexgibson: I filed [bug 1205282](https://bugzilla.mozilla.org/show_bug.cgi?id=1205282)

jpetto:
Re-added `#{{ LANG }}` to the `firefox/all` link (originally from #3305) and pushed to demo4:

https://www-demo4.allizom.org/en-US/firefox/new/

alexgibson:
Tested on demo4 on IE 6-11, MS Edge, Firefox, Chrome, Safari & Opera and all seems well. r+ from me! :birthday:

PR #3319, opened 2015-09-15
[fix bug 1199738] Update FxOS devices features.

jpetto:


alexgibson:
For the ZTE Open the features tab says `Up to 9 h`, but the specs tab says `Up to 9 h 20 m`

alexgibson:
Other than the one inconsistency this looks ok to me

It does seem a little confusing to be using decimals for talk time in some places, and hour/minutes in other places. But we already have a mix of different specs in the template already, so not a blocker.

r+wc :iphone:

alexgibson:
We're holding off merging this until l10n catches up, so keeping the do-not-merge label for now.

jpetto:
Good catch on the ZTE Open @alexgibson. I have updated the specs tab to just read "Up to 9 h" to match the features. Will at least give L10n one less string to localize/change.

flodolo:
We have 21 locales complete by now, which covers about 75% of our population. Given the missing locales, I think we're good to merge here.

PR #3320, opened 2015-09-16
Bug 1008100: fix typo in 404 template

flodolo:
Re-using existing bug for reference.

The old PHP template has a slightly different text, “here” came after a
comma, so I guess blind copy and paste.
https://www.mozilla.org/en-US/404


schalkneethling:
r+ :cake:

PR #3321, opened 2015-09-16
[bug 1197897] Optimizely Test on Windows 10 /welcome/ page

alexgibson:


alexgibson:
Still discussing final copy/messaging - but code wise this should be good for review.

alexgibson:
To test, please make sure to have [UITour enabled for local development](http://bedrock.readthedocs.org/en/latest/uitour.html#local-development), and make sure Firefox is **not** set as your default browser.

alexgibson:
@jpetto r?

alexgibson:
We're good to go in terms of copy, so removing the do-not-merge label.

bedrock/firefox/views.py
```diff
@@ -647,5 +647,30 @@ def android(request):
...
+        if variant == '1':
+            template = 'firefox/win10_variants/variant-1.html'
+        elif variant == '2':
+            template = 'firefox/win10_variants/variant-2.html'
+        elif variant == '3':
+            template = 'firefox/win10_variants/variant-3.html'
+        elif variant == '4':
+            template = 'firefox/win10_variants/variant-4.html'
+        elif variant == '5':
+            template = 'firefox/win10_variants/variant-5.html'
```
jpetto: Could shorten this up a little? Something like:

```
if (variant in ['1', '2', '3', '4', '5'] and self.request.locale == 'en-US'):
    template = 'firefox/win10_variants/variant-' + variant + '.html'
```

jpetto:
Tested locally and on Windows 10 and all worked as expected. Aside from small suggestion, looking good. r+wc

By the way, :heart: the organization and structure of the templates. :smile:

bedrock/firefox/views.py
alexgibson: Ah, much more succinct :sunrise:

alexgibson:
Thanks for the feedback @jpetto - updated :+1:

jpetto:
r+ :fire:

PR #3323, opened 2015-09-16
Bug 1205265 - Use proper quotes for Romanian in privacy/tips

Jobava:
If you look at this page: https://www.mozilla.org/ro/privacy/tips/  the right side quote in green and italic has English-style quotemarks, and they should be Romanian: „”

The en-US page: https://www.mozilla.org/en-US/privacy/tips/

Probably some other locales also need their own specific quote marks in that less rule (German or Russian):

https://www.mozilla.org/ru/privacy/tips/ (should be « » like in French)

https://www.mozilla.org/de/privacy/tips/ (should be „ “ for Germany and Austria)

Honorable mention to [Cristi Silaghi](https://twitter.com/cristi_silaghi_) for bringing this matter to my attention!

### DIFF
--- a/media/css/privacy/privacy-day.less
+++ b/media/css/privacy/privacy-day.less
@@ -269,8 +269,19 @@ html[lang^='en'] #tips-nav-direct li a span:after {
         content: "« ";
         left: -1em;
     }
+    
     &:after {
         content: " »";
+    }
+}
+
+[lang="ro"] .tip aside .quote {
+    &:before{
+        content: "„";
+    }
+    
+    &:after {
+        content: "”";
     }
 }
 


Jobava:
This requires further testing, I will add an additional commit later today.

Jobava:
Bugzilla bug report: https://bugzilla.mozilla.org/show_bug.cgi?id=1205265

Jobava:
This looks alright now after the last changes.

Screenshot: https://i.imgur.com/HPvb8G8.png

craigcook:
This looks good. Should we go ahead and add 'de' and 'ru' to the respective rules so they can get their correct punctuation as well? Also any other languages that should use angle quotes or lower quotes (It seems like German uses both, but angle quotes are more formal or old-fashioned perhaps).

Then if you can squash your commits I think we can merge this. You can ask @flodolo for some guidance if you have any trouble squashing.

Jobava:
I squashed my two commits and also removed that "merge" and other commits which don't belong to me.

flodolo:
One final nit on the commit message: it needs to be clear without the entire discussion as background, e.g. "Bug 1205265 - Use proper quotes for Romanian in privacy/tips"

I would leave out other locales for now, we can always add them if they ask. For future pages, probably a good idea to avoid this structure all together, or create a site-wide object with the proper quotes for each locale.

Jobava:
OK, rebased again and will use that style in the future.

flodolo:
Great, thanks. And the nitpicker in me thank you for fixing also the rule above yours :+1:

flodolo:
Trying the green button, hoping to not break things. Great work Jobava!

PR #3324, opened 2015-09-16
Fix Bug 1188811, fix color contrast of links in fxos styleguide

schalkneethling:
@alexgibson r?

### DIFF
--- a/media/css/styleguide/styleguide.less
+++ b/media/css/styleguide/styleguide.less
@@ -254,6 +254,14 @@ body.dark {
         color: #eee;
         text-shadow: 0 1px #333;
     }
+    a {
+        &:link {
+            color: #00C7EF;
+        }
+        &:visited {
+            color: #F4F4F4;
+        }
+    }
     #sidebar {
         h2 a {
             color: @darkThemeText;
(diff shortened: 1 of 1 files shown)


alexgibson:
Code-wise this looks fine, although we do now lose the visited link styles for all blue links on these pages? 

I don't know if there is a really a better option for dark styles such as this. I guess it's better to have sufficient contrast than it is to lose track of what's been visited already? (arguably lack of contrast means losing track of things already).

schalkneethling:
Hmmm yes, I thought about that. I would suggest we use the Light Gray [#F4F4F4] from FxOS palette for the visited state. Thoughts?

alexgibson:
I would suggest we probably consult with our style guide overlords in the bug. We could just pick an active style here if need be, but would be good to check first?

schalkneethling:
Agree

alexgibson:
Contrast looks good to me :+1: 

visited

Although I guess it's not quite as obvious which links are visited?

alexgibson:
Happy to go with whatever is the decided option here between yourself and Ty. (I don't know if there is really an ideal solution without re-assessing our link styling on a broader basis)

schalkneethling:
@alexgibson So, Ty is not overly concerned with the visited link color so I went with #f4f4f4
I agree that it is not ideal but yes, we definitely need to think about this some more. For this one though, I reckon we are good to merge. Thanks!

alexgibson:
r+ :tractor:

PR #3325, opened 2015-09-17
Don't run update_externalfiles in Dockerfile.

glogiotatidis:
External files are stored in database since baf6555.


jgmize:
good catch, @glogiotatidis. Thanks!

PR #3331, opened 2015-09-18
Bug 1196002 - Improve tablet view of /firefox/tiles/

rishiloyola:


media/css/firefox/tiles.less
```diff
@@ -109,7 +109,17 @@
...
-
+ 
+    #main-feature h1{
+        .font-size(64px);
+        float: none;
+        width: auto;
+    }
+    
+    .main-column{
+        width: auto;
```
jpetto: Just adjusting the width alone doesn't do the trick here, as the element is still `float`ed to the left. This leaves content running off the right edge of the window.

screen shot 2015-09-18 at 5 58 01 pm

You'll need to remove the float on this element - `float: none;`

media/css/firefox/tiles.less
```diff
@@ -109,7 +109,17 @@
 }
 
 @media only screen and (min-width: @breakTablet) and (max-width: @breakDesktop) {
-
+ 
+    #main-feature h1{
```
jpetto: Coding style fix - please make sure there is a space after the selector and before the opening `{`. (Here and in `.main-column` below.)

jpetto:
@rishiloyola - I've made a couple comments on your code. Let me know if you have any questions there.

When you have addressed the comments, please squash your 5 commits into a single commit. See the [squashing commits](http://bedrock.readthedocs.org/en/latest/contribute.html#squashing-your-commits) section of the docs.

Post in this PR if you need any help.

Thanks!

rishiloyola:
@jpetto I squashed my 5 commits into a single commit. Its ready to test.

jpetto:
@rishiloyola The layout is looking good!

One last request is to clean up the commit message. We try to keep all commit messages short and beginning with a reference to the bug being addressed. For this PR, I recommend a commit message like:

`[fix bug 1196002] Improve tablet view on /firefox/tiles/.`

This lets us more easily read through the history of commits.

Once you make that change, we'll be ready to merge.

rishiloyola:
Ok. You can now merge it.

Thanks.

jpetto:
Looking good @rishiloyola! Thanks for the contribution!

r+

PR #3335, opened 2015-09-21
Fix bug 1206021: Upgrade django-pipeline

pmclanahan:
The latest version does error on a broken import in .less files.

I also found and fixed (because it errored) another broken import that
doesn't appear to have been affecting the site.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -226,7 +226,7 @@ def JINJA_CONFIG():
             'jinja2.ext.loopcontrols', 'lib.l10n_utils.template.l10n_blocks',
             'lib.l10n_utils.template.lang_blocks',
             'jingo_markdown.extensions.MarkdownExtension',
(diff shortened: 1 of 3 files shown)


pmclanahan:
@jpetto @alexgibson @schalkneethling this should fix it. If you would attempt to break things locally and see if it does indeed break `./manage.py collectstatic`. You will probably have to ensure that your local `static` dir is empty. Best way to do that is by running the following commands:

```bash
$ cd static
$ git clean -fdx
$ cd ..
```
Then run `collectstatic` again and see if it errors where you expect.

pmclanahan:
Another good  (possibly better) test would be to create a branch of your own based on this one, submit a PR, and see if it fails as expected. Travis does **not** cache the static directory, so it should do a full run of the assets every time.

schalkneethling:
It works! \o/ https://sneethling.pastebin.mozilla.org/8847094

schalkneethling:
r+ :fireworks:

PR #3339, opened 2015-09-22
Fix bug 1207187 - remove R. Davies from leadership page

craigcook:


schalkneethling:
This is a r+ but I guess we should only merge this tomorrow?

craigcook:
Yeah, should probably hold off merging until we're ready for a push on the 23rd.

schalkneethling:
23rd it is.

PR #3345, opened 2015-09-25
Bug1208358: update the redirection rules for the resources under /ja/about

chikoski:
This pull request will update the redirection rules to fix the issue that ja-locale users can not acquire some resources served under /about. The details are filed as bug1208358.


chikoski:
I failed to create appropriate pull request. I close it

PR #3347, opened 2015-09-25
[fix bug 1205632] Redirect /js/language/ to archive

alexgibson:


alexgibson:
@pmclanahan r? Trying out this fancy-pants new redirect goodness. Not sure if I did everything right, but test seems to have passed? :+1:

pmclanahan:
Perfect! Nice work!

My only note is that if this url isn't already going to bedrock via a rule in `global.conf` we'll also need to add that. We'll be done with `global.conf` soon, but for now it's still a thing.

So either verify such a `RewriteRule` exists or add one, then it's an r+. :rocket:

alexgibson:
Ah yes good catch, I think I'll need to add a `RewriteRule`.

alexgibson:
@pmclanahan - updated, think this should do it?

alexgibson:
Oh, hang on - I just noticed this URL redirects to MDN?

https://www.mozilla.org/js/language/

bedrock/mozorg/redirects.py
```diff
@@ -228,6 +228,10 @@ def decider(request, **kwargs):
     # Bug 1073269 /dnt/ -> /firefox/dnt/
     redirect(r'^dnt/?$', 'firefox.dnt'),
 
+    # bug 1205632
+    redirect('^js/language(?P<path>.*)$',
```
pmclanahan: I think we should go ahead and move the [other](http://viewvc.svn.mozilla.org/vc/projects/mozilla.com/trunk/.htaccess?view=markup#l753) [redirects](http://viewvc.svn.mozilla.org/vc/projects/mozilla.com/trunk/.htaccess?view=markup#l694) here as well. Those can go above this one and then this one will catch the rest. Redirects are tried in order, and before the other URLs.

etc/httpd/global.conf
```diff
@@ -951,3 +951,6 @@ RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/pocket(/?)$ /b/$1firefox/pocket$2 [P
 
 # bug 1136436
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/ios(/?)$ /b/$1firefox/ios$2 [PT]
+
+# bug 1205632
+RewriteRule ^/js/language/?$ /b/js/language/ [PT]
```
pmclanahan: This will only forward the specific `/js/language/` URL and not any URL with anything after that. Should be this I think:

```apache
RewriteRule ^(/js/language(/.+)?)$ /b$1 [PT]
```

test_redirects/map_globalconf.py
```diff
@@ -128,6 +128,9 @@
     # bug 1051686
     url_test('/ja/firefox/organizations/', 'http://www.mozilla.jp/business/downloads/'),
 
+    # bug 1205632
+    url_test('/js/language/E262-3-errata.html',
```
pmclanahan: TESTS!! \o/ :fireworks:

etc/httpd/global.conf
pmclanahan: That rule is a bit different, I know, but it's what I've been using when moving a redirect to bedrock.

bedrock/mozorg/redirects.py
alexgibson: Sounds good, will update :+1:

etc/httpd/global.conf
alexgibson: Bah - of course, thanks!

test_redirects/map_globalconf.py
alexgibson: :heart: that we can do these :)

alexgibson:
Thanks for the feedback @pmclanahan - I've added the additional redirects/tests as suggested :+1:

etc/httpd/global.conf
```diff
@@ -951,3 +951,6 @@ RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/pocket(/?)$ /b/$1firefox/pocket$2 [P
 
 # bug 1136436
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/ios(/?)$ /b/$1firefox/ios$2 [PT]
+
+# bug 1205632
+RewriteRule ^(/js/language(/.+)?)$ /b$1 [PT]
```
pmclanahan: One more thing. I think that this won't match `/js/language/' because the `.+` means it needs at least one character after the trailing slash. Just make it `.*` and we should be good.

pmclanahan:
Looks great. Just update that rewrite rule and we'll be done \o/ :rocket: 

r+wc

etc/httpd/global.conf
alexgibson: nice catch.

alexgibson:
Thanks @pmclanahan updated

PR #3351, opened 2015-09-28
Update karma to v0.13.10

alexgibson:


alexgibson:
Tests are green - @schalkneethling r?

Note: I had an optional dependency fail (engine.io) when updating Karma locally, but this was fixed by updating my version of XCode for OSX 10.10.5. Even if I didn't update XCode, Karma still runs ok as socket.io is not used to run the tests.

schalkneethling:
r+ :checkered_flag:

PR #3352, opened 2015-09-28
Bug 1197244 - Newsletters Control Center:  use a separate .lang file …

pascalchevrel:
…for all the newsletter stuff

* newsletter.lang is no longer loaded as a default langfile for all templates (because only 8 locales do the newsletters and default files are assigned to all locales)
* all newsletter templates and newsletter python files except the hacks blog get their strings extracted to mozorg/newsletters.lang
* newsletter/views.py was loading strings from the contribute page, now they are also loaded from mozorg/newsletters.lang
* mozorg/forms.py now also loads strings from mozorg/newsletter.lang
* One string was corrected in a template (2 spaces instead of 1)
* tests and documentation updated

The mozorg/newsletters.lang file already has all the string copied from newsletter.lang and the file is on production on svn for the 8 locales that translate newsletters.

Strings in newsletter.lang that were used outside of newsletters were moved to main.lang

We will remove the old newsletter.lang from svn only once bedrock will run the new code and will no longer include newsletter.lang by default for all pages.

### DIFF
--- a/bedrock/newsletter/forms.py
+++ b/bedrock/newsletter/forms.py
@@ -20,6 +20,8 @@ from lib.l10n_utils.dotlang import _, _lazy
 
 _newsletters_re = re.compile(r'^[\w,-]+$')
 
+LANG_FILES = ['mozorg/newsletters']
+
 
 def validate_newsletters(newsletters):
     if not newsletters:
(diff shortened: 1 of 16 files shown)


bedrock/newsletter/forms.py
```diff
@@ -20,6 +20,7 @@
 
 _newsletters_re = re.compile(r'^[\w,-]+$')
 
+LANG_FILES = ['mozorg/newsletters']
 
 def validate_newsletters(newsletters):
```
pmclanahan: Travis failed because pep8 requires 2 blank lines above a function.

pmclanahan:
LGTM. Just one pep8 issue to fix and then I think Travis will be happy and we can merge. Thanks Pascal!

PR #3354, opened 2015-09-29
Fix Bug 1205357, implement fall campaign landing page

schalkneethling:


bedrock/firefox/templates/firefox/control.html
```diff
@@ -0,0 +1,57 @@
...
+  <h1>{{ high_res_img('firefox/control/header-firefox.png', {'alt': _('Firefox for desktop'), 'width': '223', 'height': '84'}) }}</h1>
+  <h2>{{_('Keep your identity yours with<br /> the new Firefox.')}}</h2>
+    {{ download_firefox(simple=true) }}
+</div>
+{% endblock alt_header %}
+
+{% block content %}
+<main role="main">
+  <div class="blurb">
+    <h2>More Protection. The Most Privacy.</h2>
```
flodolo: All these strings need to be wrapped for l10n

schalkneethling: Meh, totally slipped my mind somehow.

flodolo:
Strings extracted locally but not exposed for localization yet. Please warn me if there's any change from now on.

One information that I need is on which demo server this page will live, since localizers need to be able to test their work.

schalkneethling:
Asked about headline string wrapping here: https://bugzilla.mozilla.org/show_bug.cgi?id=1205357#c3

schalkneethling:
@flodolo The url for this page will be /firefox/choose

bedrock/firefox/templates/firefox/choose.html
```diff
@@ -0,0 +1,58 @@
...
+
+{% block site_header_nav %}{% endblock %}
+{% block site_header_logo %}{% endblock %}
+
+{% block alt_header %}
+<div class="header-content">
+  <div id="tabzilla">
+    <a href="{{ url('mozorg.home') }}">Mozilla</a>
+  </div>
+  <h1>{{ high_res_img('firefox/choose/header-firefox.png', {'alt': _('Firefox for desktop'), 'width': '223', 'height': '84'}) }}</h1>
```
flodolo: Why "Firefox for desktop" and not just "Firefox" (unlocalizable)?

schalkneethling: I copied it from another page but, no reason to have it like that. Let me change it to just Firefox.

schalkneethling:
@alexgibson This one is ready for review r?

bedrock/firefox/templates/firefox/choose.html
```diff
@@ -0,0 +1,58 @@
...
+</main>
+{% endblock %}
+
+{% block email_form %}{% endblock %}
+
+{% block site_footer %}
+  {% include 'firefox/includes/simple_footer.html' %}
+{% endblock %}
+
+{% block site_js %}
```
alexgibson: Can just remove this block since it's not doing anything here in the inherited template?

bedrock/firefox/templates/firefox/choose.html
```diff
@@ -0,0 +1,58 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "base-resp.html" %}
```
alexgibson: Since this is in the `/firefox/` app should it be extending `firefox/base-resp.html`?

media/css/firefox/choose.less
```diff
@@ -0,0 +1,179 @@
...
+        }
+    }
+    .download-button.download-button-simple {
+        color: #fff;
+        margin-bottom: 0;
+        border: none;
+        font-weight: bold;
+        .font-size(1.5rem);
+        text-align: center;
+        .inline-block();
```
alexgibson: I believe this may be superfluous since `.inline-block` is already applied in `buttons.less`

media/css/firefox/choose.less
```diff
@@ -0,0 +1,179 @@
...
+            .download-heading {
+                color: #fff;
+            }
+        }
+    }
+    .download-content {
+        text-transform: uppercase;
+    }
+}
+.download-button {
```
alexgibson: Nit: Unnecessary indentation

media/css/firefox/choose.less
```diff
@@ -0,0 +1,179 @@
...
+    }
+}
+
+main {
+    margin: 0 auto;
+    padding: 90px 0 120px 0;
+    width: 100%;
+    max-width: 1400px;
+    text-align: center;
+}
```
alexgibson: Nit: line break after `}`

bedrock/firefox/templates/firefox/choose.html
```diff
@@ -0,0 +1,58 @@
...
+  <div class="blurb">
+    <h2>{{ _('The most private browsing') }}</h2>
+    <p>{{ _('Take control of your web by using Private Browsing with Tracking Protection, only from Firefox.') }}</p>
+    <a href="{{ url('firefox.desktop.trust') }}" class="more">{{ _('Learn more about this') }}</a>
+  </div>
+
+  <div class="blurb">
+    <h2 class="data">{{ _('Here for you, not selling your info') }}</h2>
+    <p>{{ _('Discover easy tips about how to take back control of your online experience, only from Mozilla.') }}</p>
+    <a href="{{ url('firefox.desktop.trust') }}" class="more">{{ _('Get smart on data tracking') }}</a>
```
alexgibson: This is the same URL as the previous link, guessing they still need to be updated?

bedrock/firefox/templates/firefox/choose.html
```diff
@@ -0,0 +1,58 @@
...
+  </div>
+  <h1>{{ high_res_img('firefox/choose/header-firefox.png', {'alt': _('Firefox'), 'width': '223', 'height': '84'}) }}</h1>
+
+  <h2 data-tracking="{{ _('Browse Freely.<br> Choose Firefox.') }}" data-selling="{{ _('Take Control.<br> Choose Firefox.') }}">{{_('Stand for choice.<br> Choose Firefox.')}}</h2>
+  {{ download_firefox(simple=true) }}
+</div>
+{% endblock alt_header %}
+
+{% block content %}
+<main role="main">
```
alexgibson: Would this be better as another element instead of `<main>`, since the `<h1>` and download button are not within it?

bedrock/firefox/templates/firefox/choose.html
schalkneethling: Yup, still waiting on what those should be.

media/css/firefox/choose.less
schalkneethling: So, you are saying rather something like:

```
.download-button .download-list .os_android .download-link .download-subtitle {
    margin: 0;
    .font-size(1.2rem);
    text-align: inherit;
}
```

alexgibson: Unless there's a reason to nest all the indentation here, yes.

ckprice:
@flodolo @schalkneethling sorry, but have to get title and desc. Will have today updated in the [copy doc](https://docs.google.com/document/d/1BN0brWqkOK5-HE_uP1EzR0amVq_uWwonDNo_46HKe6k/edit).

ckprice:
@flodolo @schalkneethling  updated copy document with final `<title>`. The 'B' wants to be lower case, and there is no period after `the  new Firefox`.

bedrock/firefox/templates/firefox/choose.html
```diff
@@ -0,0 +1,56 @@
...
+    <p>{{ _('Take control of your web by using Private Browsing with Tracking Protection, only from Firefox.') }}</p>
+    <!-- Still waiting on final URL for this link -->
+    <a href="{{ url('firefox.desktop.trust') }}" class="more">{{ _('Learn more about this') }}</a>
+  </div>
+
+  <div class="blurb">
+    <h2 class="data">{{ _('Here for you, not selling your info') }}</h2>
+    <p>{{ _('Discover easy tips about how to take back control of your online experience, only from Mozilla.') }}</p>
+    <!-- For python this will be url('teach.smarton.tracking') - Currently not 100% whether smart on will launch on or before the same date as this page so, using the URL below to avoid errors. -->
+    <a href="https://mozilla.org/teach/smarton/tracking/" class="more">{{ _('Get smart on data tracking') }}</a>
```
alexgibson: Perhaps it might be better to waffle this section until we know that Smarton is going to be live on time, else currently this link is 404?

alexgibson: If we don't want to use waffle, perhaps a good fallback URL could be `/privacy/tips/`?

alexgibson:
Other than what to do with the two URL's we're waiting on, this looks good to me! r+

bedrock/firefox/templates/firefox/choose.html
```diff
@@ -0,0 +1,56 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/base-resp.html" %}
+
+{% block page_title_prefix %}{{_('Firefox Web browser')}} — {% endblock %}
+{% block page_title %}Keep your identity yours with the new Firefox{% endblock %}
```
alexgibson: Wait, this title is not wrapped for localization?

bedrock/firefox/templates/firefox/choose.html
```diff
@@ -0,0 +1,56 @@
...
+
+{% block site_header_nav %}{% endblock %}
+{% block site_header_logo %}{% endblock %}
+
+{% block alt_header %}
+<div class="header-content">
+  <div id="tabzilla">
+    <a href="{{ url('mozorg.home') }}">Mozilla</a>
+  </div>
+  <h1>{{ high_res_img('firefox/choose/header-firefox.png', {'alt': _('Firefox'), 'width': '223', 'height': '84'}) }}</h1>
```
alexgibson: Nit: I don't believe we need to localize `Firefox`

alexgibson:
Other than the two remaining URLs and an unwrapped string, this page looks good to me, r+wc :wink: :racehorse:

bedrock/firefox/templates/firefox/choose.html
schalkneethling: Yeah, let use that for now.

bedrock/firefox/templates/firefox/choose.html
schalkneethling: I have no idea how that slipped through the cracks :-/

bedrock/firefox/templates/firefox/choose.html
```diff
@@ -0,0 +1,58 @@
...
+  {{ download_firefox(simple=true) }}
+</div>
+{% endblock alt_header %}
+
+{% block content %}
+<section class="page-content">
+  <div class="blurb">
+    <h2>{{ _('The most private browsing') }}</h2>
+    <p>{{ _('Take control of your web by using Private Browsing with Tracking Protection, only from Firefox.') }}</p>
+    {% if waffle.switch('tracking-protection') %}
```
alexgibson: This waffle switch seems a bit pointless, since it also uses the same switch name that turns the page itself on/off? Is there any reason why this needs waffle, seeing as it currently points to an existing page anyway, and not the tracking protection page?

schalkneethling: Hmmmm, very true. I was thinking both would most likely be flipped at the same time but yeah, seeing the previous link is not a 404 or something like that, it is safe to merge and can update the URL once the correct page is in master.

Reverting.

PR #3358, opened 2015-09-29
Fix Bug 1209455 - Error 404 for all Thunderbird release notes URLs on mozilla.org

kyoshino:


bedrock/thunderbird/urls.py
```diff
@@ -31,20 +34,26 @@
...
     url('^thunderbird/(?:%s/)?system-requirements/$' % channel_re,
         bedrock.releasenotes.views.latest_sysreq,
-        {'product': 'Thunderbird'}, name='thunderbird.sysreq'),
+        {'product': 'thunderbird'}, name='thunderbird.latest.sysreq'),
     url('^thunderbird/latest/system-requirements/$',
         bedrock.releasenotes.views.latest_sysreq,
         {'product': 'thunderbird', 'channel': 'release'}),
+    url(sysreq_re,
+        bedrock.releasenotes.views.system_requirements,
+        {'product': 'Thunderbird'}, name='thunderbird.sysreq'),
```
jgmize: Shouldn't 'Thunderbird' be lower case here?

kyoshino: No, this one is upper-case somehow. I think everything should be lower case, but it likely requires change to Nucleus.

kyoshino: I'll comment on the mixture of lower- and upper-cases.

kyoshino: Comment added. I'm happy to send a separate PR to fix the inconsistency, as `firefox/urls.py` has the same issue.

jgmize: > Comment added. I'm happy to send a separate PR to fix the inconsistency, as firefox/urls.py has the same issue.

Thanks, I agree that shouldn't block this.

PR #3360, opened 2015-09-29
Move redirects from global.conf to app.

pmclanahan:
This handles most of the redirects in global.conf.
Some remain, but this is now to the point where
the test_redirects suite and mcom-tests passes against a demo
server running Apache.


pmclanahan:
@jgmize it has arrived. r?

jgmize:
@pmclanahan I think this is good except for the flake8 errors travis is complaining about. r+wc

jgmize:
This is nice work, very thorough, great attention to detail. I really appreciate you slogging through all this.

pmclanahan:
Thanks. I think "slog" is a good word for it ;)

bedrock/firefox/views.py
```diff
@@ -290,7 +290,7 @@ def dnt(request):
 def all_downloads(request, channel):
     if channel is None:
         channel = 'release'
-    if channel == 'developer':
+    if channel == 'developer' or channel == 'aurora':
```
jgmize: Nice catch.

PR #3361, opened 2015-09-29
Fix  Bug 1209706 - [traceback] django.core.urlresolvers:NoReverseMatch: Reverse for 'thunderbird.all' with arguments '()' and keyword arguments '{}' not found. 0 pattern(s) tried: []

kyoshino:


jgmize:
Thanks @kyoshino

PR #3364, opened 2015-09-29
Fix Bug 1209708 - [traceback] exceptions:UnicodeEncodeError: 'ascii' codec can't encode character u'\u2211' in position 10: ordinal not in range(128)

kyoshino:


stephendonner:
@jgmize @pmclanahan r?

pmclanahan:
@kyoshino mind adding a test with the problem URL to the `test_redirects` suite?

jgmize:
> @kyoshino mind adding a test with the problem URL to the test_redirects suite?

@pmclanahan I don't think @kyoshino has that information, b/c it isn't in the bug. Looking at the info about the error in New Relic, I see the escaped version of the malformed URL:
`/it/thunderbird/2.0.0.23/start/NBVC%e2%88%91BNM` but a proper test would include the the original junk data instead of the escaped equivalent, and I'm not sure if that should necessarily be a blocker for this PR.

kyoshino:
I guess new tests are not necessary here, because it's caught by a recently added test :smiley_cat:

pmclanahan:
@jgmize when I paste the url you included in your comment into my browser on `www.mozilla.org` I get a 500. So I think it'd be a good test to add to the suite to make sure that's a 404 instead.

pmclanahan:
@kyoshino well that's why I asked. Our tests are currently passing but this URL is throwing a 500.

kyoshino:
Hmm, will take a look tomorrow.

jgmize:
> @jgmize when I paste the url you included in your comment into my browser on www.mozilla.org I get a 500. So I think it'd be a good test to add to the suite to make sure that's a 404 instead.

I should have realized that copying and pasting it into the URL bar would unescape it automatically and render as /it/thunderbird/2.0.0.23/start/NBVC∑BNM. Thanks for pointing that out, @pmclanahan. I can understand the desire for a specific test case for this junk URL, I just want to make sure we don't start "enumerating badness"-- there are many possible urls that may throw unicode errors similar to the one we saw in this bug that are out of scope of this PR.

pmclanahan:
Agreed. But it'd be nice to have one to verify this I think, and should cover the full range of errors of this type for this URL. It's not really a big deal, it just seemed like the kind of thing that'd be easy to reintroduce later.

kyoshino:
I cannot reproduce 500 locally. It just says 404.

pmclanahan:
Cool. We'll just get this in then.

PR #3367, opened 2015-09-30
Fix remaineder of mcom-test redirects.

pmclanahan:


pmclanahan:
@jgmize this should do it (I think) r?

test_redirects/map_globalconf.py
```diff
@@ -213,8 +213,8 @@
     url_test('/firefox/mobile/faq/?os=firefox-os', '/firefox/os/faq/'),
 
     # Bug 986174
-    url_test('/firefox/mobile/features/', '/firefox/android/'),
-    url_test('/firefox/mobile/faq/', '/firefox/android/faq/'),
+    url_test('/{m,{firefox/,}mobile}/features/', '/firefox/android/'),
+    url_test('/{m,{firefox/,}mobile}/faq/', '/firefox/android/faq/'),
```
jgmize: will you explain the nested grouping in the first arg on this line and the line above please? it seems like it would test the following:
`/m/faq/`
`/firefox//faq/`  <-- note the double slash
`/mobile/faq`

jgmize: nm, I think I've got it. If I'm reading this correctly now, 216 should expand to:
`/m/features/`
`/firefox/mobile/features/`
`/mobile/features/`
and 217 to:
`/m/faq/`
`/firefox/mobile/faq/`
`/mobile/faq/`
which lgtm.

pmclanahan: Sorry, just saw this. You're right in your 2nd comment. I know it's a bit confusing when nested like that, and I could write it out fully, but this does work and makes sense to me, and I think will to others if they know bash expansion and regex. Also [braceexpand](https://pypi.python.org/pypi/braceexpand) is a wonderful tool.

PR #3368, opened 2015-09-30
Redirect /mobile/platforms/ to sumo. bug 884933

jgmize:


jgmize:
@pmclanahan r?

pmclanahan:
LGTM. Need to update the `PT` rule in `global.conf` to allow `/mobile/platforms/` through to bedrock in legacy deployment.

PR #3370, opened 2015-09-30
[fix bug 1186605] Audit LESS & JS bundles.

jpetto:


alexgibson:
Making some notes on progress as I test this

- [x] Cleaned out `static` directory and ran `collectstatic` sucessfully

bedrock/settings/static_media.py
```diff
@@ -798,24 +746,22 @@
...
         'output_filename': 'css/styleguide-docs-mozilla-pager-bundle.css',
     },
     'styleguide-docs-send-to-device': {
         'source_filenames': (
-            'css/sandstone/sandstone-resp.less',
             'css/base/send-to-device.less',
         ),
         'output_filename': 'css/styleguide-docs-send-to-device-bundle.css',
     },
+    # no longer used on bedrock. possibly referenced on external sites? should investigate.
```
alexgibson: Yep we need to keep this around as long as external sites may still be using tabzilla. I think we really need to set a public date for deprecation and let people know, before we remove this from bedrock.

alexgibson:
All the pages I've tested are looking fine so far. I'll dig in a little further to the code changes themselves, but things are looking good so far :+1:

bedrock/firefox/templates/firefox/channel.html
```diff
@@ -24,7 +24,7 @@
 
 {% block body_id %}channel{% endblock %}
 
-{% block extrahead %}
+{% block page_css %}
```
alexgibson: Should this page remain using `extrahead`, since it's a download page and older browser support is more critical?

bedrock/thunderbird/templates/thunderbird/index.html
```diff
@@ -15,9 +15,7 @@
   <h2>{{ high_res_img('thunderbird/landing/wordmark.png', {'alt': _('Thunderbird'), 'width': '304', 'height': '84'}) }}</h2>
 {% endblock %}
 
-{% block site_css %}{% endblock %}
-
-{% block extrahead %}
+{% block page_css %}
```
alexgibson: Again, should a download page like this use `extrahead`?

alexgibson:
Pages that inherit the [base template in the legal app](https://github.com/mozilla/bedrock/blob/master/bedrock/legal/templates/legal/base-resp.html) aren't getting the `nav_main_resp` bundle, so the mobile nav is non-functional.

alexgibson:
Same for `/press/speakerrequest/`

alexgibson:
Actually, this is the same for any template outside of the /firefox app that is responsive (and does not use `mozorg_common`)

jpetto:
Ah, rats. I forgot `nav-main-resp.js` is needed to for the mobile menu and not just desktop dropdowns. Hm. I'm going to give the approach a little re-think. May make more sense to restore the Firefox common bundle and put `nav-main-resp.js` back in `common`. :confounded:

alexgibson:
Ok, will hang fire fox now - ping me when you're ready to start review again :+1:

bedrock/firefox/templates/firefox/channel.html
jpetto: Hm, good point. Going to pull all CSS into `extrahead` here for older browsers.

bedrock/thunderbird/templates/thunderbird/index.html
jpetto: I did think about this one. It's not a browser, and the download button is still visible and functional as-is on IE 7, so I figured it was okay to send fancy CSS only modern browsers. What do you think?

alexgibson: As long as the download button is still visible and works, then I agree it's fine to change this one :smile:

jpetto:
Okay, added a commit that essentially adds `nav-main-resp.js` back to the `common` bundle, and serves that `common` bundle to almost every page on the site. Thought is caching a single bundle will offset unnecessary weight on some pages.

bedrock/settings/static_media.py
```diff
@@ -1763,5 +1627,5 @@
             'js/libs/matchMedia.addListener.js',
         ),
         'output_filename': 'js/matchmedia_addlistener-bundle.js',
-    },
+    }
```
alexgibson: Teeny nit: isn't the trailing comma preferred?

alexgibson:
@jpetto - I think this is looking good. All pages should be covered now that `nav-main-resp.js` is back in the common bundle. Aside from one tiny nit, it's an r+wc from me. Care to squarsh?

alexgibson:
Good to fix here also - I noticed there are couple of JS bundles that are being output in the wrong directory. They're currently residing in `/css` once output, but should be in `/js`

https://github.com/mozilla/bedrock/blob/master/bedrock/settings/static_media.py#L1451
https://github.com/mozilla/bedrock/blob/master/bedrock/settings/static_media.py#L1460
https://github.com/mozilla/bedrock/blob/master/bedrock/settings/static_media.py#L1469

jpetto:
Addressed the latest comments and squashed. Thanks @alexgibson!

alexgibson:
r+ :bomb:

PR #3373, opened 2015-10-01
Fix softlinkstatic to actually create symlinks

jgmize:


jgmize:
as pointed out by @stephendonner in IRC:
5:16 PM <•stephend> one is that loading.gif issue - http://www.webpagetest.org/result/150930_6B_1R8G/
5:16 PM <•stephend> (seems to be there, just with a dynamic asset-hashed filename
5:17 PM <•stephend> other is http://www.webpagetest.org/result/150930_6P_21WH/ - https://www.allizom.org/en-US/contact/spaces/
5:18 PM <•stephend> missing https://www.allizom.org/media/img/contact/moz-map-pin.png asset

There are still references to original, unhashed image filenames so we need to continue serving them from the /static dir.

docker/softlinkstatic.py
```diff
@@ -7,10 +7,11 @@
...
-for path in static_files['paths']:
-    full_path = os.path.join('./static', path)
-
-    try:
-        os.unlink(full_path)
-    except OSError:
-        pass
+for orig, hashed in static_files['paths'].items():
+    if '?' in orig:
+        continue
```
pmclanahan: would be nice to catch these too, but I get ya. Could just split on `?`, but my other idea for this doesn't see those query params, so we can fix that way.

jgmize: Before I wrote it this way I validated that in every instance of an `orig` with a query param, there is already a corresponding `orig.split('?')[0]` in `static_files['paths']`, so I didn't see a point in processing those twice.

pmclanahan: Oh cool. Of course you didn't, because there isn't one. Didn't realize that. Nice catch.

PR #3374, opened 2015-10-01
Add Hostname Middleware.

glogiotatidis:
Adds `X-Backend-Server` header in responses.

Added the ENABLE setting so we don't touch the non-Docker installs but I'm happy to remove it if it doesn't make a difference.

### DIFF
--- a/bedrock/mozorg/middleware.py
+++ b/bedrock/mozorg/middleware.py
@@ -69,3 +69,9 @@ class ClacksOverheadMiddleware(object):
         if response.status_code == 200:
             response['X-Clacks-Overhead'] = 'GNU Terry Pratchett'
         return response
+
+
+class HostnameMiddleware(object):
+    def process_response(self, request, response):
+        response['X-Backend-Server'] = settings.HOSTNAME
+        return response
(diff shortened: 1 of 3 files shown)


glogiotatidis:
@jgmize r?

jgmize:
I agree with keeping it a setting that is off by default, since we're already generating the x-backend-server header in apache and there's no point in generating a duplicate. It would be nice if the env var determined whether the middlware was in settings.MIDDLEWARE_CLASSES at all instead of checking the setting on each request. It's a bit of a micro-optimization nit, but given that this is in middleware it *might* be worth it-- what do you think?

glogiotatidis:
Makes testing and code a bit more compliated but here it is if you prefer

pmclanahan:
My opinion is that this "micro-optimization" makes settings a good bit less readable. I personally prefer the way you had it, but it's not wrong this way.

jgmize:
@glogiotatidis I used your latest modifications to the jenkins pipeline to deploy this to bedrock-dev, and validated that no X-backend-server header was added until I ran `deis config:set ENABLE_HOSTNAME_MIDDLEWARE=1 -a bedrock-dev `, after which all requests started returning the header as expected. 

I like the pattern you've come up with on making the middleware optional, and may submit a future PR expanding its use-- there are several middlewares that I think would benefit from it, making the middleware as simple as possible at the cost of some readability in the settings. I personally prefer this tradeoff, and would have already merged this but I respect @pmclanahan's differing opinion on this and will hold off on merging pending further discussion.

pmclanahan:
I don't wanna block it. If you both like it then go for it.

jgmize:
thanks @glogiotatidis and @pmclanahan

PR #3375, opened 2015-10-01
Fix Bug 1196573, update private browsing blurb on trust page

schalkneethling:
@alexgibson r?

@flodolo There are still an open question re the URL and explicit line breaks in the sub head and copy.


alexgibson:
All these string changes will need l10n tags?

alexgibson:
Also should we waffle this so it turns on once 42 is released?

flodolo:
URL doesn't block us, since it's not part of the string. But yes, everything should be behind a tag, and we should display the DNT blurb for locales not ready.

flodolo:
Actually, correction. We shouldn't display the DNT bit, just display the old Private Browsing text.

bedrock/firefox/templates/firefox/desktop/trust.html
```diff
@@ -48,25 +48,28 @@
...
+              {{ _('Sometimes it’s nice to go undercover.') }}
+  -            {{ _('Turn this feature on to protect your browsing history from others.') }}
+            </p>
 
-          <a class="more" rel="external" data-interaction="outbound link" href="https://support.mozilla.org/kb/private-browsing-browse-web-without-saving-info">{{ _('Learn how to open a private window') }}</a>
-        </li>
+            <a class="more" rel="external" data-interaction="outbound link" href="https://support.mozilla.org/kb/private-browsing-browse-web-without-saving-info">{{ _('Learn how to open a private window') }}</a>
+          </li>
+        {% endif %}
         {% if l10n_has_tag('forget_button') %}
```
schalkneethling: Do we have enough locales translated for the forget button or, is there still some that fallback to the lightbeam blurb? I removed the CSS for lightbeam and will need to add it back if the fallback is still displayed often. @flodolo ?

flodolo: We have 9 locales that still don't have the strings translated, and we'd have to disable the page for them.

Having said that, I'm not against cleaning the template, but I don't think that's something we should do now, already too much going on.

bedrock/firefox/templates/firefox/desktop/trust.html
```diff
@@ -48,25 +48,28 @@
...
+          <li id="privacy-browsing">
+            <h3>{{ _('Private Browsing') }}</h3>
 
-          <p>
-            {{ _('Sometimes it’s nice to go undercover.') }}
-            {{ _('Turn this feature on to protect your browsing history from others.') }}
-          </p>
+            <p>
+              {{ _('Sometimes it’s nice to go undercover.') }}
+  -            {{ _('Turn this feature on to protect your browsing history from others.') }}
```
flodolo: Extra trailing minus

bedrock/firefox/templates/firefox/desktop/trust.html
```diff
@@ -48,25 +48,28 @@
...
+            </p>
 
-          <a class="more" rel="external" data-interaction="outbound link" href="https://support.mozilla.org/kb/private-browsing-browse-web-without-saving-info">{{ _('Learn how to open a private window') }}</a>
-        </li>
```
flodolo: This should be behind a tag as well.

flodolo:
Not sure how final this is in terms of code, but alignment looks pretty bad outside of en-US
schermata 2015-10-01 alle 13 21 24

schalkneethling:
Ah, I know what is wrong there.

bedrock/firefox/templates/firefox/desktop/trust.html
schalkneethling: Agreed

bedrock/firefox/templates/firefox/desktop/trust.html
```diff
@@ -48,25 +48,28 @@
       </p>
 
       <ul class="feature-list">
-        <li id="privacy-dnt">
-          <h3>{{ _('Do Not Track') }}</h3>
+        {% if l10n_has_tag('tracking_protection') %}
```
alexgibson: We probably want to add a waffle switch here, so the new copy only shows if the locale has the l10n flag **and** the waffle switch is `true`, e.g. something like:

    {% if l10n_has_tag('tracking_protection') and waffle.switch('tracking_protection') %}

alexgibson: Note: we seem to use a convention of using hyphens `-` in waffle switches looking at other pages, as opposed to underscores `_`.

bedrock/firefox/templates/firefox/desktop/trust.html
```diff
@@ -48,25 +48,28 @@
       </p>
 
       <ul class="feature-list">
-        <li id="privacy-dnt">
-          <h3>{{ _('Do Not Track') }}</h3>
+        {% if l10n_has_tag('tracking-protection') %}
```
flodolo: Can you switch to `tracking_protection` like the other page? We should not have issues anymore with hyphens, but still, underscore are safe and consistent ;-)

schalkneethling:
Reade for code review @alexgibson r?

schalkneethling:
@alexgibson waffle flag added.

alexgibson:
This one has some centering issues also:

trust-alignment

alexgibson:
Still alignment issues on this one at smaller viewports:

trust-alignment

schalkneethling:
Oh man, I do not know how I do not pick these up :puzzled:

media/css/firefox/desktop/trust.less
```diff
@@ -48,24 +48,38 @@
...
+            line-height: 1.3;
         }
-
-        &#privacy-forget {
-            background-position: -1310px top;
+        li {
+            float: none;
+            display: inline-block;
+            background: url(/media/img/firefox/desktop/trust/sprite-privacy.png) center top no-repeat;
+            margin-right: 30px;
```
alexgibson: Is it intentional that this has a right margin of `30px`, while the `last-child` below has a left margin of `20px`?

media/css/firefox/desktop/trust.less
```diff
@@ -48,24 +48,38 @@
...
+            padding-top: 180px;
+            vertical-align: top;
+
+            @media only screen and (max-width: @breakMobileLandscape) {
```
alexgibson: Should this be `@breakTablet` seeing as it goes into single column here?

media/css/firefox/desktop/trust.less
schalkneethling: No, they should both be 20px

alexgibson:
r+ looking good! :pineapple:

PR #3376, opened 2015-10-01
[fix bug 1209643] Links from foundation/moco result in 404

alexgibson:


alexgibson:
@pmclanahan - r? This fixes a couple of 404's on `/foundation/moco`. 

I took the opportunity to update and move the redirects to bedrock. I added tests, which seem to work ok (woo!). Only part I wasn't too sure on was adding the pass-through for each URL in `global.conf`. I tried to follow what's been done elsewhere (fingers crossed I've got it right).

bedrock/mozorg/redirects.py
```diff
@@ -222,6 +222,13 @@ def decider(request, **kwargs):
     redirect(r'^legal/eula/firefox-2/?$', '/about/legal/eula/firefox-2/'),
     redirect(r'^legal/eula/firefox-3/?$', '/about/legal/eula/firefox-3/'),
 
+    # Bug 1209643
+    redirect(r'^legal/bylaws_amendment_1(\.html|/)?', '/foundation/documents/bylaws-amendment-1/'),
```
pmclanahan: This is fine, but you can use the view name here as well for a bit more future-proofing. (e.g. `foundation.documents.bylaws`, etc.)

alexgibson: ah of course, thanks.

pmclanahan: could also combine these first two lines:

```python
redirect(r'^legal/bylaws_amendment_(?P<n>[12])(\.html|/)?', '/foundation/documents/bylaws-amendment-{n}/'),
```

etc/httpd/global.conf
```diff
@@ -498,12 +498,12 @@ RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?foundation/documents/(index|mozilla-200.-fin
...
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?foundation/trademarks/l10n-policy(/?)$ /$1foundation/trademarks/ [L,R=301]
 
-# bug 960689
-RewriteRule ^/en-US/legal/bylaws_amendment_1(?:\.html)?$ /foundation/documents/bylaws-amendment-1/ [L,R=301]
-RewriteRule ^/en-US/legal/bylaws_amendment_2(?:\.html)?$ /foundation/documents/bylaws-amendment-2/ [L,R=301]
-RewriteRule ^/en-US/legal/articles(?:\.html)?$ /foundation/documents/articles-of-incorporation/ [L,R=301]
-RewriteRule ^/en-US/legal/amendment(?:\.html)?$ /foundation/documents/articles-of-incorporation/amendment/ [L,R=301]
-RewriteRule ^/en-US/legal/bylaws(?:\.html)?$ /foundation/documents/bylaws/ [L,R=301]
+# bug 960689, 1209643
+RewriteRule ^(/(\w{2,3}(?:-\w{2})?/)?legal/bylaws_amendment_1(?:\.html)?)$ /b$1 [PT]
```
pmclanahan: Similar shortcut available here too.

pmclanahan:
LGTM. It'd be nice to use the view names in the calls to `redirect`, but other than that you don't really need to modify anything. Very nice work!

bedrock/mozorg/redirects.py
alexgibson: This is good because I learnings' things :+1:

alexgibson:
Thanks for the suggestions @pmclanahan,

I switched to using view names in the redirects, and added shortcuts for the `bylaws-amendment` redirects/rewrite rules. I left it as a separate commit for easier review, thanks!

bedrock/mozorg/redirects.py
```diff
@@ -222,6 +222,12 @@ def decider(request, **kwargs):
     redirect(r'^legal/eula/firefox-2/?$', '/about/legal/eula/firefox-2/'),
     redirect(r'^legal/eula/firefox-3/?$', '/about/legal/eula/firefox-3/'),
 
+    # Bug 1209643
+    redirect(r'^legal/bylaws_amendment_(?P<n>[12])(\.html|/)?', '/foundation/documents/bylaws-amendment-{n}/'),
```
pmclanahan: Hmm. Maybe it would be a nice feature to allow substitution from source url to view name as well... Not now, but that seems like it might come in handy for things like this.

pmclanahan:
r+ very nice work! Squarsh and merge at will.

bedrock/mozorg/redirects.py
alexgibson: I did wonder if there was a way to do this as well, +1 from me in that it could be handy :+1:

alexgibson:
Squashed and will merge as soon as Travis is green. Thanks for the review @pmclanahan!

PR #3377, opened 2015-10-01
Fix bug 1210368: Update dev to use locales from git.

pmclanahan:
This will require a change on the admin node, so please
coordinate merging this with me so this server change can
be made just prior to merging this.


pmclanahan:
@jgmize r?

This also affects all of the demos which just use a symlink to the dev locales. Should be a good test :)

jgmize:
r+, I'll leave the merging to you after you make the change to the admin node.

PR #3378, opened 2015-10-01
Fix Bug 1196564, update tracking protection blurb on android page

schalkneethling:
@alexgibson r? Same issue here as with the trust page. Awaiting final feedback regarding line breaks in copy.


bedrock/firefox/templates/firefox/android/index.html
```diff
@@ -311,36 +311,28 @@ <h2 class="heading">{{ _('Keep your private info private') }}</h2>
...
-
-        <a rel="external" href="https://www.youtube.com/watch?v=Ewk0YtFTBec&amp;list=PLA8ABF6B7A66A0C5E&amp;index=15">
-          {% if l10n_has_tag('android_spring_updates') %}
-            {{ _('Learn how to open a private window') }}
-          {% else %}
-            {{ _('See how Private Browsing works') }}
-          {% endif %}
-        </a>
-      </li>
+    {% if l10n_has_tag('tracking_protection') %}
```
alexgibson: Again we probably need to use a waffle switch here? Suggest using the same name for the switch as in #3375

bedrock/firefox/templates/firefox/android/index.html
```diff
@@ -311,36 +311,28 @@ <h2 class="heading">{{ _('Keep your private info private') }}</h2>
...
+
+          <a class="more" rel="external" data-interaction="outbound link" href="https://support.mozilla.org/kb/private-browsing-browse-web-without-saving-info">{{ _('Learn more about Private Browsing') }}</a>
+        </li>
+      {% else %}
+        <li id="private-browsing">
+          <h3>{{ _('Private Browsing') }}</h3>
+
+          <p>
+            {{ _('Sometimes it’s nice to go undercover.') }}
+-            {{ _('Turn this feature on to protect your browsing history from others.') }}
```
alexgibson: There's a random `-` character on the start of this line

bedrock/firefox/templates/firefox/android/index.html
```diff
@@ -311,36 +311,28 @@ <h2 class="heading">{{ _('Keep your private info private') }}</h2>
...
+          </p>
+
+          <a class="more" rel="external" data-interaction="outbound link" href="https://support.mozilla.org/kb/private-browsing-browse-web-without-saving-info">{{ _('Learn more about Private Browsing') }}</a>
+        </li>
+      {% else %}
+        <li id="private-browsing">
+          <h3>{{ _('Private Browsing') }}</h3>
+
+          <p>
+            {{ _('Sometimes it’s nice to go undercover.') }}
```
alexgibson: Would this line and the one below be better as a single string instead of split into two? Thoughts @flodolo?

bedrock/firefox/templates/firefox/android/index.html
flodolo: Rogue minus here too.

bedrock/firefox/templates/firefox/android/index.html
```diff
@@ -311,36 +311,28 @@ <h2 class="heading">{{ _('Keep your private info private') }}</h2>
...
+-            {{ _('Turn this feature on to protect your browsing history from others.') }}
+          </p>
+
+          <a class="more" rel="external" data-interaction="outbound link" href="https://support.mozilla.org/kb/private-browsing-browse-web-without-saving-info">{{ _('Learn how to open a private window') }}</a>
```
flodolo: Need to restore the existing condition on the string. Also, is the URL change expected?

bedrock/firefox/templates/firefox/android/index.html
flodolo: I'm not understanding this PR at all. I have the feeling you're copying code from the desktop/trust PR, but the current page is completely different. You're basically introducing more string as fall backs for new strings.

alexgibson:
Page is looking kinda broken for me?

bedrock/firefox/templates/firefox/android/index.html
alexgibson: Yep, we should fall back to the old content here, not introduce more new content as a fallback (as that won't be translated)

schalkneethling:
Ok, restoring the correct fallback and @alexgibson I will see if I can replicate what you see. Which browser/version are you see the above?

alexgibson:
> Which browser/version are you see the above?

Firefox 43 Dev Edition

bedrock/firefox/templates/firefox/android/index.html
```diff
@@ -311,18 +311,17 @@ <h2 class="heading">{{ _('Keep your private info private') }}</h2>
     </p>
 
     <ul>
-      <li id="dnt">
-        <h3>{{ _('Do not track') }}</h3>
+      {% if l10n_has_tag('tracking_protection') %}
```
alexgibson: Note: still need a waffle switch here

alexgibson: Also indentation needs fixing

alexgibson:
There's still some layout /responsive issues with the changes to the icons. At smaller viewport sizes things become misaligned a bit.

schalkneethling:
@alexgibson waffle flag added.

alexgibson:
Still some alignment issues with centering the icons

icon-alignment

schalkneethling:
Yup, found the problem there.

alexgibson:
Nearly there, but still some alignment issues it seems?

android-icon-alignment

alexgibson:
@schalkneethling - looking good! r+ care to squarsh?

alexgibson:
Just need functional tests updating for the link changes and then we can merge this :+1:

schalkneethling:
Functional test updated.

alexgibson:
r+ :bee:

PR #3380, opened 2015-10-02
Enable database connection reuse for mysql and postgres.

pmclanahan:
Let's try this.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -40,6 +40,10 @@ DATABASES = config(
                            cast=dj_database_url.parse,
                            default='sqlite:///bedrock.db')}))
 
+if DATABASES['default']['ENGINE'].endswith('psycopg2'):
+    # let the DB handle connection killing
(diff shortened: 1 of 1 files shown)


pmclanahan:
@jgmize this should work. worth a shot maybe.

bedrock/settings/base.py
```diff
@@ -40,6 +40,10 @@ def path(*args):
                            cast=dj_database_url.parse,
                            default='sqlite:///bedrock.db')}))
 
+if DATABASES['default']['ENGINE'].endswith(('mysql', 'psycopg2')):
+    # let the DB handle connection killing
+    DATABASES['default']['CONN_MAX_AGE'] = None
```
jgmize: In theory this should be better, but I'm a little concerned about making a major change like this to our current infra-- I'd want to test this out with a canary deployment, which would take time away from other, higher priority changes that need to be made. What do you think about only enabling this for postgres for now, since it will be easier to test the impact of the change in the new infra?

pmclanahan:
Sounds good. Updated.

PR #3382, opened 2015-10-02
Fix bug 1210396: Add newsletter strings to bedrock.

pmclanahan:
I did it in the view rather than the template, but it works
just as well.

### DIFF
--- a/bedrock/newsletter/views.py
+++ b/bedrock/newsletter/views.py
@@ -51,6 +51,149 @@ unknown_address_text = _lazy(
 invalid_email_address = _lazy(u'This is not a valid email address. '
                               u'Please check the spelling.')
 
+NEWSLETTER_STRINGS = {
+    u'about-mozilla': {
+        'description': _lazy(u'Bringing you the best new opportunities to help support the open Web.'),
+        'title': _lazy(u'Mozilla Communities')},
+    u'about-standards': {
+        'description': _lazy(u''),
+        'title': _lazy(u'About Standards')},
+    u'addon-dev': {
+        'description': _lazy(u''),
+        'title': _lazy(u'Addon Development')},
+    u'addons': {
+        'description': _lazy(u''),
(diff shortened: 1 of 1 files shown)


pmclanahan:
@pascalchevrel this seems to work well, but some strings aren't translated when I run this locally that are on production. In my searching it seems that there are still strings in `newsletter.lang` that are being used in prod and haven't been moved to `mozorg/newsletters.lang`. Not sure if that's the deal or not, but since I remember `newsletter.lang` being removed from the default files recently I figured it may be relevant. Let me know what you think.

pmclanahan:
@pascalchevrel bump

bedrock/newsletter/views.py
```diff
@@ -51,6 +51,149 @@
...
+        'title': _lazy(u'Addon Development')},
+    u'addons': {
+        'description': _lazy(u''),
+        'title': _lazy(u'About Addons')},
+    u'affiliates': {
+        'description': _lazy(u'A monthly newsletter to keep you up to date with the '
+                             u'Firefox Affiliates program.'),
+        'title': _lazy(u'Firefox Affiliates')},
+    u'ambassadors': {
+        'description': _lazy(u'A monthly newsletter on how to get involved with Mozilla on your campus. '),
```
pascalchevrel: There is an extra trailing space on this string

pascalchevrel:
I rechecked all the strings in your PR and I think the missing strings are actually those that are changed as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1199799

I don't know if you want to update your PR to reflect the changes asked in this bug or if you prefer a separate PR.

pmclanahan:
@pascalchevrel I've updated this PR to change the names of those newsletters. Is that all that you needed?

pascalchevrel:
yes  I think it's fine, sorry for the delay I was on PTO

PR #3384, opened 2015-10-02
[fix bug 1191009] Update events filters.

jpetto:
- Show only events that haven't started on home page.


pmclanahan:
Looks great! Thanks Jon. r+

PR #3387, opened 2015-10-02
Fix bug 1210980: Do not redirect Fx 28.0 release notes to archive.

pmclanahan:
From IRC:

```
<CamJN> Hey all, https://www.mozilla.org/en-US/firefox/28.0/releasenotes/ is 404ing
```

28.0beta is in the archive, but 28.0 should be served by bedrock.


pmclanahan:
@jgmize r?

for a test I'll be adding that URL to mcom-tests since we can't rely on a 200 response during a Travis run due to lack of nucleus data.

PR #3389, opened 2015-10-05
Add healthcheck URL.

glogiotatidis:
This URL will return 200 OK. To be used with Deis health checks.

### DIFF
--- a/bedrock/base/views.py
+++ b/bedrock/base/views.py
@@ -1,0 +1,5 @@
+from django.http import HttpResponse
+
+
+def health_check(request):
+    return HttpResponse('OK')
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -206,6 +206,7 @@ SUPPORTED_NONLOCALES = [
     'telemetry',
     'webmaker',
     'contributor-data',
+    'healthz',
 ]
 
 ALLOWED_HOSTS = config(
@@ -875,5 +876,8 @@ SSLIFY_DISABLE = config('DISABLE_SSL', default=True, cast=bool)
 if not SSLIFY_DISABLE:
     SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
     USE_X_FORWARDED_HOST = True
+SSLIFY_DISABLE_FOR_REQUEST = [
+    lambda request: request.get_full_path() == '/healthz/'
+]
 
 HOSTNAME = platform.node()
--- a/bedrock/urls.py
+++ b/bedrock/urls.py
@@ -48,6 +48,8 @@ urlpatterns = (
     url(r'^facebookapps/',
         include('bedrock.facebookapps.urls')),
 
+    url(r'^healthz/$', 'bedrock.base.views.health_check'),
+
     # Uncomment the admin/doc line below to enable admin documentation:
     # url(r'^admin/doc/', include('django.contrib.admindocs.urls')),
 )


pmclanahan:
Why do we need a special url for this? I thought it'd be better to have a real page checked which would ensure that the whole site is working. The home page even ensures that the DB is working. I'm probably misunderstanding the purpose of this thing.

vendor-local/src/legal-docs
```diff
@@ -1 +1 @@
-Subproject commit fdf03d97dff7a53e6331018649c357fa6eea9b9e
+Subproject commit 3e3426f74e477265f3231c9b332cfe169ae35264
```
pmclanahan: Probably didn't mean to include this.

glogiotatidis:
Valid question. I'm still not convinced that this is needed but I wanted to build a test image to test it, thus the pull request. 

Deis health checks against a relative url e.g. `/` instead of a full url e.g. `http://bedrock-dev.us-west.moz.works` which makes sense. The problem is that we redirect all requests to https and therefore the health check against `/en-US/` returns `301` instead of `200`. 

This PR attempts to solve this by providing a non-https URL to check against.

I'll investigate further and update here.

glogiotatidis:
Docs here: http://docs.deis.io/en/latest/using_deis/config-application/?highlight=health#custom-health-checks

pmclanahan:
Ah! I'd forgotten about the HTTPS 301 even though your PR clearly opts `/200/` out of that specifically. Seems good then.

pmclanahan:
r+ from me then as soon as the submodule change is removed.

jgmize:
There is a convention of using `/healthz` as the endpoint for health checks that we may want to adopt. It's the example endpoint in the [kubernetes docs](http://kubernetes.io/v1.0/docs/user-guide/liveness/README.html), deis will [adopt it internally](https://github.com/deis/deis/issues/4482) as the project switches to k8s in v2, and other large open source projects such as cloud foundry are [adopting it](https://github.com/openshift/origin/blob/b2e97700db09415052d60c02c9edf49bfaf506b3/pkg/cmd/server/origin/handlers.go#L35) as well.

glogiotatidis:
Rebased and tested http://giorgosbedrock.eu-west.moz.works/healthz/

glogiotatidis:
Merge at will

glogiotatidis:
please hold merging this, I want to test another thing.

glogiotatidis:
I tried to enabled checks on my app with

`deis_eu config:set HEALTHCHECK_URL=/healthz/  HEALTHCHECK_INITIAL_DELAY=50 -a giorgosbedrock`

and I get `detail: aborting, app containers failed to respond to health check`

jgmize:
@glogiotatidis I suspected that your HEALTHCHECK_INITIAL_DELAY was too small, due to the time it takes to download the maxmind db, so to test that theory I set HEALTHCHECK_INITIAL_DELAY=300 on bedrock-dev on both clusters and triggered [bedrock pipeline #144](https://ci.us-west.moz.works/view/Bedrock%20Pipeline/?fullscreen=true) by building [bedrock_base_image with parameters TAG=pr/3389/head](https://ci.us-west.moz.works/view/Bedrock/job/bedrock_base_image/144/console).

jgmize:
The deploy was successful and https://bedrock-dev.us-west.moz.works/healthz/ returns OK for me now. We can probably tune HEALTHCHECK_INITIAL_DELAY lower than 300, but I believe this should be fine to merge as-is. Do you agree, @glogiotatidis?

glogiotatidis:
Interesting. I'm wondering whether the healthcheck delay starts the moment the unit files get created *or* when the port opens. I'd expect the later thus the small delay in my setup but it's probably the first. 

My only concern is if Deis will need at least 300 seconds (i.e. 5 minutes) to make a deploy? Otherwise I'm good to merge.

jgmize:
> Interesting. I'm wondering whether the healthcheck delay starts the moment the unit files get created or when the port opens. I'd expect the later thus the small delay in my setup but it's probably the first.

The publisher [listens](https://github.com/deis/deis/blob/v1.10.1/publisher/server/publisher.go#L47) to docker API events on each host, calls [publishContainer](https://github.com/deis/deis/blob/v1.10.1/publisher/server/publisher.go#L64) when the container is created, which does a [health check](https://github.com/deis/deis/blob/v1.10.1/publisher/server/publisher.go#L146) which will [sleep for the configured delay](https://github.com/deis/deis/blob/v1.10.1/publisher/server/publisher.go#L206) before testing the configured url, and only publish to etcd on success. 

> My only concern is if Deis will need at least 300 seconds (i.e. 5 minutes) to make a deploy? Otherwise I'm good to merge.

Yes, I'd like to eliminate this delay as well, but as this is set outside of the code I don't think that's a blocker for merge. I only set it that high to test my theory-- we can experiment with different values and see what works.

PR #3391, opened 2015-10-05
Bug 1196569 - Private Browsing product page

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1196569


bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,136 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import fxfamilynav with context %}
+
+{% extends "/firefox/base-resp.html" %}
+
+{% block page_title %}{{ _('Keep your Firefox in sync') }}{% endblock %}
```
flodolo: Need to update the page title.

bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,136 @@
...
+
+  <section class="intro">
+    <div class="container">
+      <div class="header">
+        <h1>
+          {{ _('More protection.') }}
+          <br>
+          {{ _('The most privacy.') }}
+          <br>
+          {{ _('Only from Firefox.') }}
```
flodolo: We have already [exposed](More protection. <br>The most privacy. <br>Only from Firefox.) a similar string for the campaign landing page, I'd suggest to reuse the same
`More protection. <br>The most privacy. <br>Only from Firefox.`

sgarrity: @flodolo To be clear here, it should change in this case because we already have that string in the landing page, but usually this would be a better way to wrap the strings, right?

flodolo: Yes and no. This is not horrible for localization, but it kind of forces you to have 3 sentences on 3 lines, when maybe you want to try something different.

ckprice:
>Need to update the page title.

Sorry guys, I need to get this from Matej. Will have today.

ckprice:
@flodolo @malena copy document updated with title and desc.

[TITLE]

Firefox Web browser — Private Browsing with Tracking Protection — Mozilla


[DESCRIPTION]

Get more protection and the most privacy, only from Firefox. No other browser gives you so much control over your digital footprint.

bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,136 @@
...
+  </section>
+
+  <section class="control-center">
+    <div class="container">
+      <img src="{{ static('img/firefox/private-browsing/mask.png') }}">
+      <h2>{{ _('The Control Center has you covered') }}<sup aria-describedby="control-center-caveat">*</sup></h2>
+      <p>
+        {{ _('When you see the shield to the left of the address bar, that means Firefox is protecting you from potential trackers.') }}
+        {{ _('You can disable Tracking Protection for a single site, or turn it off entirely.') }}
+        {{ _('It’s up to you.') }}
```
flodolo: Please keep this two sentences together
```
{{ _('You can disable Tracking Protection for a single site, or turn it off entirely. It’s up to you.') }}
```

flodolo:
Besides a couple of notes on strings, I can't see anything wrong from a quick text. Waiting to double check on next push.

malena:
@flodolo Updated strings.

bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,130 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import fxfamilynav with context %}
+
+{% extends "/firefox/base-resp.html" %}
+
+{% block page_title %}{{ _('Firefox Web browser — Private Browsing with Tracking Protection — Mozilla ') }}{% endblock %}
```
flodolo: Please remove the trailing space.

flodolo:
Strings extracted locally, please warn if anything changes. Plan to expose them within 24 hours.

ckprice:
@flodolo we have no new strings aside from the Title and Desc were added to the [copy document](https://docs.google.com/document/d/15Egy67jk_eD1LqFoWGGzh-nMXH1t2iEpV3K0VkTMDQE/edit) yesterday.

I haven't stepped through this PR, but if all the strings in the copy doc are in, then we are good to go here.

flodolo:
@ckprice Thanks, luckily this PR had the correct title and desc from the start (just double checked).

malena:
Yep I added the correct title and desc as soon as it became available in the doc. I am still working on layout/design for now. @flodolo let me know when I should push this to demo server.

flodolo:
Note: the page is available on demo5 only with the /b redirect
https://www-demo5.allizom.org/b/firefox/private-browsing

etc/httpd/global.conf
```diff
@@ -285,6 +285,9 @@ RewriteRule ^/en-US/firefox/channel/android(/?)$ /b/en-US/firefox/channel/androi
 # bug 1059518
 RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/tiles(.*)$ /b/$1firefox/tiles$2 [PT]
 
+# bug 1196569 
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)/firefox/private-browsing(.*)$ /b/$1firefox/private-browsing$2 [PT]
```
schalkneethling: I reckon you need something like this:

```
# bug 1205357
RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/choose(/?)$ /b/$1firefox/choose$2 [PT]
```

schalkneethling: Reckon `(.*)` is causing not to?

jpetto:
@malena - Just a heads up that some additional GTM requests have come through in [bug 1217883](https://bugzilla.mozilla.org/show_bug.cgi?id=1217883).

malena:
Okay thanks! @jpetto

flodolo:
Nit found by a localizer: small#tracking-protection-caveat takes the entire width of the page, small#control-center-caveat doesn't (37%).

malena:
@flodolo Ahh okay yes the PS design has the caveat for small#control-center-caveat break at a smaller width hence my css width rule, but since this page is localized it would be better to maintain the same wider width. Thanks!

malena:
r?

flodolo:
"Try Private Browsing" is not hard-coded
https://bugzilla.mozilla.org/show_bug.cgi?id=1218543

I remember checking this page and not noticing this issue, we have this string [extracted](https://github.com/mozilla-l10n/www.mozilla.org/blob/master/en-US/firefox/private-browsing.lang#L20), when did it change?

bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,170 @@
...
+{% block content %}
+
+<main role="main">
+  <section class="intro">
+    <div class="container">
+      <header>
+        <h1>{{ _('More protection. <br>The most privacy. <br>Only from Firefox.') }}</h1>
+        <div class="download-buttons">
+          <div class="firefox-latest-cta">
+            <a href="https://www.mozilla.org" class="button-flat-dark" data-highlight="privateWindow" role="button"> Try Private Browsing </a>
```
craigcook: Unwrapped string, "Try private browsing"

bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,170 @@
...
+      </p>
+    </div>
+  </section>
+
+  <section class="cta" id="cta">
+    <div class="container">
+      <div class="download-buttons">
+        <div class="firefox-latest-cta">
+          <h2>{{ _('Try it now to stay in control') }}</h2>
+          <a href="https://www.mozilla.org" class="button-flat-dark" data-highlight="privateWindow" role="button"> Try Private Browsing </a>
```
craigcook: Unwrapped string, "Try private browsing"

bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,170 @@
...
+      </div>
+      <div class="blurb with-tracking">
+        <h3>{{ _('With Tracking Protection')}}</h3>
+        <p>{{ _('Minimize the ability of hidden third parties from tracking your browsing activity across many sites.') }}</p>
+      </div>
+      <p class="footer">
+        <strong>
+          <a class="android-link" href="https://support.mozilla.org/en-US/kb/private-browsing-firefox-android">{{ _('Learn how to use Private Browsing') }}</a>
+          <a class="ios-link" href="https://support.mozilla.org/kb/private-browsing-firefox-ios
+">{{ _('Learn how to use Private Browsing') }}</a>
```
craigcook: Nit: unnecessary line break here

bedrock/firefox/templates/firefox/private-browsing.html
malena: Oooh yes, there was some confusion with what text was originally in there for that line. The copy had a different sentence that appeared after l10n. So during that confusion I left it like that to remember to check if it. @flodolo Don't worry, the copy hasn't changed! I made sure of that!

flodolo:
Can you push an update to demo server? That's important for testing and check how the button looks.

bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,169 @@
...
+{% block content %}
+
+<main role="main">
+  <section class="intro">
+    <div class="container">
+      <header>
+        <h1>{{ _('More protection. <br>The most privacy. <br>Only from Firefox.') }}</h1>
+        <div class="download-buttons">
+          <div class="firefox-latest-cta">
+            <a href="https://www.mozilla.org" class="button-flat-dark" data-highlight="privateWindow" role="button">{{ _('Try Private Browsing') }}</a>
```
schalkneethling: You can use the url helper here:
`{{ url('mozorg.home') }}

bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,169 @@
...
+              </a>
+            </div>
+            <div class="ios-download-button">
+              <a href="{{ settings.APPLE_APPSTORE_FIREFOX_LINK }}&amp;ct=mozorg-ios_page-appstore-button">
+                <img src="{{ l10n_img('firefox/ios/btn-app-store.svg') }}" alt="{{ _('Download on the App Store') }}">
+              </a>
+            </div>
+          </div>
+          <div class="firefox-old-cta">
+            <p>{{ _('Your Firefox is not up to date. Download the latest version to use the new Private Browsing.') }}</p>
```
schalkneethling: Not sure if it matters, and probably to late for a string change but, seems a little strange that this string does include the "with Tracking Protection" part?

bedrock/firefox/templates/firefox/private-browsing.html
alexgibson: This URL should probably fall back to something more useful than the homepage. If someone doesn't have the `privateWindow` target in their pallet, then the button will fall back to a regular link instead of highlighting the icon. A SUMO article like https://support.mozilla.org/kb/private-browsing-use-firefox-without-history/may be more appropriate.

bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,169 @@
...
+  <section class="control-center" id="control-center">
+    <div class="container">
+      <h2>{{ _('The Control Center has you covered') }}</h2>
+      <p>
+        {{ _('When you see the shield to the left of the address bar, that means Firefox is protecting you from potential trackers.') }}
+        {{ _('You can disable Tracking Protection for a single site, or turn it off entirely. It’s up to you.') }}
+      </p>
+      <p class="footer">
+        <strong>
+          <a class="android-link" href="https://support.mozilla.org/kb/mixed-content-blocker-firefox-android">{{ _('Learn more about the Control Center') }}</a>
```
schalkneethling: This link says "Control Center" but the caveat states that Fx on Android does not have a control center?

bedrock/firefox/templates/firefox/private-browsing.html
```diff
@@ -0,0 +1,169 @@
...
+      </p>
+    </div>
+  </section>
+
+  <section class="cta" id="cta">
+    <div class="container">
+      <div class="download-buttons">
+        <div class="firefox-latest-cta">
+          <h2>{{ _('Try it now to stay in control') }}</h2>
+          <a href="https://www.mozilla.org" class="button-flat-dark" data-highlight="privateWindow" role="button">{{ _('Try Private Browsing') }}</a>
```
schalkneethling: You can use the url helper here as well.

media/css/firefox/private_browsing/private-browsing-conditionals.less
```diff
@@ -0,0 +1,170 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../../sandstone/lib.less";
+
+.download-buttons  {
+    > div {
```
schalkneethling: Nit: No need to nest, can just be `.download-buttons > div`

media/css/firefox/private_browsing/private-browsing-conditionals.less
```diff
@@ -0,0 +1,170 @@
...
+
+.android {
+    .footer .android-link {
+        display: block;
+    }
+    .footer .desktop-link {
+        display: none;
+    }
+    .download-buttons {
+    }
```
schalkneethling: Empty rule?

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../../sandstone/lib.less";
+
+/*-------------------------------------------------------------------------*/
+// Common
+
+#wrapper {
+    border-top: 2px solid #fff;
+    background-color: #0095dd;
```
schalkneethling: Should this be @mozIDBlue https://github.com/mozilla/bedrock/blob/master/media/css/sandstone/lib.less#L66

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+    .font-size(25px);
+    line-height: 1.2;
+    margin: @baseLine auto;
+}
+
+p {
+    color: @mozIDGrey;
+    .open-sans-light();
+    .font-size(20px);
+
```
schalkneethling: Nit: Superfluous empty line.

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+}
+
+.footer {
+    margin: 45px auto 0 auto ;
+    text-align: center;
+  
+    a {
+        display: none;
+    }
+    a::after {
```
schalkneethling: Use a single `:` which will that ensure IE8 can join the party :tada: :)

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+        content:" »";
```
schalkneethling: Nit: Use single quotes

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+
+.tracking-protection,
+.when-to-use,
+.control-center {
+    h2 {
+        margin-bottom: (@baseLine * 3);
+        padding: 0 30px;
+    }
+    p {
+        margin: @baseLine auto 0 auto;
```
schalkneethling: Nit: You can leave off the last `auto`

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+    }
+}
+
+.intro {
+    margin-top: 50px;
+
+    header {
+        width: 490px;
+        padding-top: 80px;
+        float: left;
```
schalkneethling: Nit: Would prefer if the order was `float`, `padding-top`, `width`

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+    background-color: #fff;
+  
+    h3 {
+        .open-sans-extrabold();
+        .font-size(30px);
+        margin: 10px auto; 
+    }
+    .blurb {
+        background: url('/media/img/firefox/private-browsing/sprite-tracking-protection.png') no-repeat;
+        background: linear-gradient(transparent, transparent), url('/media/img/firefox/private-browsing/sprite-tracking-protection.svg');
```
schalkneethling: Nit: Use gradient mixin? https://github.com/mozilla/bedrock/blob/master/media/css/sandstone/lib.less#L290

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+        width: 465px;
+    } 
+    .with-tracking {
+        background-position: -410px 0;
+    }
+    .private-browsing {
+        background-position: 135px 0;
+        padding-right: 24px;
+    }
+    .private-browsing::after {
```
schalkneethling: Single `:` here as well

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+        color: #6fbf4a;
+        content: "+";
```
schalkneethling: Single quotes

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+    }
+    li {
+        float: left;
+        margin: 0;
+        text-align: center;
+        width: 330px;
+       
+        p {
+            background: url('/media/img/firefox/private-browsing/sprite-features.png');
+            background: linear-gradient(transparent, transparent), url('/media/img/firefox/private-browsing/sprite-features.svg');
```
schalkneethling: Gradient mixin?

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+        background-position: 80px top;
+    }
+    li:first-child + li p {
+        background-position: -190px top;
+    }
+}
+
+.control-center {
+    background: url('/media/img/firefox/private-browsing/control-center.png');
+    background: linear-gradient(transparent, transparent), url('/media/img/firefox/private-browsing/control-center.svg');
```
schalkneethling: And here

media/css/firefox/private_browsing/private-browsing.less
```diff
@@ -0,0 +1,489 @@
...
+    background-repeat: no-repeat;
+    background-color: #fff;
+    padding-top: 320px;
+    text-align: center;
+
+    p {
+        .font-size(25px);
+    }
+    .footer a::after {
+        content:" »";
```
schalkneethling: Single quotes and single `:` above

schalkneethling: Nit: Space after `content:`

media/js/firefox/private-browsing.js
```diff
@@ -0,0 +1,47 @@
...
+                $shield.addClass('blocked');
+            });
+        } else {
+            $html.addClass('firefox-old');
+        }
+    } else {
+        $html.addClass('non-firefox');
+    }
+
+})(window.jQuery, window.Modernizr, window.site);
```
schalkneethling: Nit: You can shorten this to `jQuery, Modernizr, site`, `window` is implied. Although, do you need Modernizr?

media/js/firefox/private-browsing.js
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+;(function($, Modernizr, site) {
+    'use strict';
+
+    var $html = $(document.documentElement);
```
schalkneethling: Nit: You can shorten this to `$('html');`

schalkneethling:
Not sure if it is only on this page (I reckon there may be others) but, the color contrast of the secondary links under the download button is not good at all. In fact, the text is almost illegible.

screen shot 2015-10-27 at 12 29 20

schalkneethling:
The license text in the footer under the wordmark is effected by the paragraph rule here:
https://github.com/mozilla/bedrock/pull/3391/files#diff-17c14532df4ae71b74268ce01e6aba9aR49

causing the font size to be to big. Perhaps add a body (or a class on `main`) class such as `private-browsing` and then wrap the `p` and the header rules inside it.

screen shot 2015-10-27 at 12 40 41

bedrock/firefox/templates/firefox/private-browsing.html
alexgibson: This one should be a more appropriate fallback link too, like the other CTA above :+1:

media/css/firefox/private_browsing/private-browsing-conditionals.less
```diff
@@ -0,0 +1,170 @@
...
+       display: block; 
+    }
+    .cta {
+        .container {
+            padding: 60px 20px;
+        }
+        h2 {
+            .font-size(30px);
+            line-height: 1.4;
+            width: 440px;
```
schalkneethling: I do not believe you need to make the width smaller here. It can just use the `500px` from `.cta h2`. Making it smaller causes a disconnect between the message and the button I feel.

Currently:

screen shot 2015-10-27 at 12 52 25

As apposed to:

screen shot 2015-10-27 at 12 52 38

schalkneethling:
Alignment problems when showing the Google Play download button:

screen shot 2015-10-27 at 12 59 56

And just above the footer:

screen shot 2015-10-27 at 13 00 04

schalkneethling:
Possibly also on ios (although in the ios simulator it looked fine) - This is replacing for example android, with ios on the `html` element.

screen shot 2015-10-27 at 13 02 16

And above the footer:

screen shot 2015-10-27 at 13 02 25

media/js/firefox/private-browsing.js
```diff
@@ -0,0 +1,47 @@
...
+        window.dataLayer.push({
+            event: 'private-browsing-page-interactions',
+            interaction: 'scroll',
+            section: section        
+        });
+    }
+
+    Mozilla.HighlightTarget.init('.button-flat-dark');
+
+    if (window.isFirefox()) {
```
alexgibson: Just a side note: this will return true for Firefox for Android as well as Desktop. Perhaps this is intentional, but it's worth pointing out that the UITour functionality will only work on the desktop browser. I'm not sure what Firefox for Android users should see here? (sorry for the drive-by comment).

schalkneethling:
IE is being a ... on smaller viewports (IE9)

windows 7 - ie9

and

windows 7 - ie92

bedrock/firefox/templates/firefox/private-browsing.html
malena: The copy has been reviewed and approved. Too late to make any changes.

media/js/firefox/private-browsing.js
malena: Ohh okay. Firefox on Android should see what Chrome on Android sees. So I would need to add another check to see if mobile too.

media/js/firefox/private-browsing.js
malena: When I remove the window reference, window.isFirefox() won't work.  Also I thought we used Modernizr on all pages? Especially this one which needs to work in IE since we are using HTML5 elements.

media/js/firefox/private-browsing.js
malena: Ahh yes, but actually I've seen $(document.documentElement); on other pages. Wondering why that has been used over $('html')?

media/js/firefox/private-browsing.js
alexgibson: Fwiw, I think it's ok to keep these as `window.jQuery` etc, as it makes it obvious they are global. As for Modernizr, this should not be required unless you are using any of its features specifically. We include an [HTML5 Shiv](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/base-resp.html#L45-L48) is all pages by default for IE.

media/css/firefox/private_browsing/private-browsing-conditionals.less
malena: The design had the sentence break specifically in those spots for desktop. Not sure if I should be changing that? I can push current to demo and send link to design for confirmation on this.

media/js/firefox/private-browsing.js
malena: Ahh okay great @alexgibson. I just try to keep my page consistent with others and I actually just copied this from another page. Will remove it.

media/css/firefox/private_browsing/private-browsing.less
malena: @schalkneethling  Hmm that's only there for IE fallback, not really a gradient. Wondering if using a mixin would make it hard to understand what's happening here? I can make a comment perhaps? What do you think?

malena:
@schalkneethling I've tackled the ie9 layout issues. Also learned ie9 does not support top:inherit. :confused:

media/css/firefox/private_browsing/private-browsing.less
schalkneethling: depends on which IE version(s) this is for. It is only supported by IE10+ and the mixin adds the `-ms-filter` part for older versions of IE that does not support `linear-gradient`.

http://caniuse.com/#search=linear-gradient

schalkneethling: With that said IE looks good in general so, maybe it is not an issue at all.

schalkneethling:
IE :angry:

bedrock/firefox/templates/firefox/private-browsing.html
schalkneethling: :+1:

media/css/firefox/private_browsing/private-browsing-conditionals.less
schalkneethling: Perhaps have them take a look, not a blocker. Thx @malena

schalkneethling: This bug confirms that I was wrong https://bugzilla.mozilla.org/show_bug.cgi?id=1218848

malena:
@schalkneethling I've updated the font colour of that text that wasn't very contrasty. I also updated alignment issues on mobile. Will be doing another round of tests once demo5 is updated but for now this is ready for another pass when you have a chance. Thanks!

schalkneethling:
When the `html` element class is set to `ios js non-firefox loaded` you will see in the screen shot below, that the critters are causing horizontal scrolling at small viewports:

screen shot 2015-10-28 at 16 58 17

From what I can see, it is caused by the margins set by this selector:
`.android .intro .critter-animation, .ios .intro .critter-animation`

Removing that solves the problem.

schalkneethling:
screen shot 2015-10-28 at 16 45 28

These icons seems to be ever so slightly off center.

schalkneethling:
@malena Do we not have high-res versions of the images? From what I can see in Fx, when loading the page on high-def, it still loads the same PNG files as it does for normal resolutions.

malena:
@schalkneethling I got svg images but had issues with supporting ie > 10 and ie < 9 at the same time. The original linear-gradient fix I had doesn't work on ie11 since I have to use the prefix -ms-linear-gradient but then the ordering of the rules made it either work in ie >10 but not in ie11. Anyways, I replaced them with PNG's because I couldn't find a good working solution. I can ask for high res versions to swap them out.

schalkneethling:
@malena If you have access to the original files, the high-res will just be double the size of the standard PNG. Then you can use the `at2x` mixin to load the different versions.

With that said, I will not hold up this PR while we wait for those images or, create them ourselves. If all the other items are addressed, we can merge this one and add the high-res images in a separate PR.

Thanks @malena

malena:
@schalkneethling okay sounds good! I can address images other PR. Can you check the android fix I just committed?

schalkneethling:
@malena The changes for ios and android are all good. The icons are still a bit off center. I had a look and adjusting a few of the rules as follows, seems to do the trick:

```
.when-to-use li:first-child p {
    background-position: 90px top;
}

.when-to-use li:first-child + li p {
    background-position: -180px top;
}

.when-to-use li p {
    transparent url("/media/img/firefox/private-browsing/sprite-features.png") no-repeat scroll -490px top / 743px auto;
}
```

I am also wondering if it would not be better to give each of those list element a class. Then you can avoid selectors such as `.when-to-use li:first-child + li p` for example and just have:

```
.when-to-use li.classname p
```

Or even, place the class on the `p`, and then have:

```
.when-to-use p.classname
```
Let me know what you think. This is super close. In fact, I believe this is the last item that needs to be addressed. Thanks!

schalkneethling:
The shield here at around the desktopBreakpoint is not completely centered.

screen shot 2015-10-28 at 20 30 38

I find that adjusting the margin as below, resolves the problem.

```
#tracking-protection-animation .shield-container {
    margin: 0 0 0 -25px
}
```
Give it a try and let me know what you think.

malena:
@schalkneethling I have updated the alignment. I left the html how it was and didn't add any class names. Just wanted to deal with the css for now.

schalkneethling:
w00t r+ :de: Thanks @malena, looking great! - Just needs a quick squash.

malena:
@schalkneethling wahoo!! squashed :dancer:

PR #3393, opened 2015-10-06
Update tests to pass without l10n files.

pmclanahan:
Mostly involved updating some mocks and tests.


pmclanahan:
@jgmize looks like Travis passed \o/ r?

PR #3394, opened 2015-10-06
Fix bug 1196555 - iOS product page updates for 1.1

craigcook:


bedrock/firefox/templates/firefox/ios.html
```diff
@@ -122,6 +124,38 @@
...
+    <h2 class="heading">{{ _('Keep your private info private') }}</h2>
+
+    <p class="lead">
+      {{ _('We believe your personal information belongs to you.') }}
+      {{ _('Firefox for iOS is filled with features to help keep it that way. ') }}
+    </p>
+
+    <ul>
+      <li id="private-browsing">
+      {% if waffle.switch('tracking-protection') %}
```
alexgibson: Does Firefox for iOS support tracking protection at all? (It was my understanding it does not yet, but I could be wrong).

alexgibson: When I try in Aurora 1.1 for iOS, creating a new private tab says “Firefox won't remember any of your history or cookies, but bookmarks will be saved.” Doesn't seem to mention anything about trackers currently. If this isn't going to change, it would effect the line of copy below as well as this headline.

craigcook: I don't think it does but wanted to confirm, hence the waffle switch. I suspect I'll be removing this.

bedrock/firefox/templates/firefox/ios.html
```diff
@@ -122,6 +124,38 @@
...
+        <h3>{{ _('Private Browsing with<br> Tracking Protection') }}</h3>
+      {% else %}
+        <h3>{{ _('Private Browsing') }}</h3>
+      {% endif %}
+        <p>{{ _('Browse without having your history, searches or cookies saved, and block hidden third parties from following you across the Web.') }}</p>
```
alexgibson: This string mentions blocking third parties, which needs clarification too I guess.

bedrock/firefox/templates/firefox/ios.html
craigcook: Actually, since I'm about 99% sure FxiOS won't have tracking protection in 1.1, I'll go ahead and remove it here and below. We can add it later if we're wrong.

bedrock/firefox/templates/firefox/ios.html
```diff
@@ -93,6 +104,8 @@
       <li class="sync-history">{{ _('History') }}</li>
       <li class="sync-passwords">{{ _('Passwords') }}</li>
       <li class="sync-tabs">{{ _('Open tabs') }}</li>
+      <li class="sync-bookmarks">{{ _('Bookmarks') }}</li>
+      <li class="sync-reading-list">{{ _('Reading List') }}</li>
```
flodolo: Even if we never shipped, the page is localized and live for several language. We need new strings behind a flag.

flodolo:
@m-alexis I plan to push out the strings for this page in the next few hours. [Based on the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1196555#c2) everything seems correct to me, but would you mind double checking?

bedrock/firefox/templates/firefox/ios.html
```diff
@@ -122,6 +144,36 @@
...
 </section>{#-- /#sync --#}
 
+{% if l10n_has_tag('tracking_protection') and waffle.switch('firefox-ios-global') %}
+<section class="section section-privacy" id="privacy">
+  <div class="container">
+    <h2 class="heading">{{ _('Keep your private info private') }}</h2>
+
+    <p class="lead">
+      {{ _('We believe your personal information belongs to you.') }}
+      {{ _('Firefox for iOS is filled with features to help keep it that way. ') }}
```
schalkneethling: Nit: trailing space after `way. ` - not sure if it is changeable now.

bedrock/firefox/templates/firefox/ios.html
```diff
@@ -122,6 +144,36 @@
...
+    </p>
+
+    <ul>
+      <li id="private-browsing">
+        <h3>{{ _('Private Browsing') }}</h3>
+        <p>{{ _('When activated, Firefox for iOS won’t accept cookies, remember passwords or record any sites that you’ve visited. ') }}</p>
```
schalkneethling: Nit: There is a trailing space here as well so, perhaps those are intentional?

bedrock/firefox/templates/firefox/ios.html
```diff
@@ -122,6 +144,36 @@
...
+        <a rel="external" href="https://support.mozilla.org/kb/clear-your-browsing-history-and-other-personal-dat">{{ _('Learn how to clear your history') }}</a>
+      </li>
+    </ul>
+  </div>{# /.container #}
+</section>
+{% endif %}
+
 {% if has_widget %}
   <div id="send-to-modal-container">
     {{ send_to_device(platform='ios', title_text=_('Send Firefox to your iOS device'), ios_link=settings.APPLE_APPSTORE_FIREFOX_LINK + '&ct=mozorg-ios_page-appstore-button_sd') }}
```
schalkneethling: Nit: `&` to `&amp;`

media/css/firefox/ios.less
```diff
@@ -76,20 +79,21 @@ a.go {
 // Main hero space
 .section-intro {
     .container {
-        padding: 40px 500px 80px 10px;
+        padding: 60px 500px 80px 10px;
         min-height: 550px;
-        background-position: 570px 0;
+        background-position: 570px 20px;
```
schalkneethling: Nit: `@baseLine` for `20px` ?

bedrock/firefox/templates/firefox/ios.html
craigcook: Hm... I suspect that will break the string for l10n, but I'm not sure how strictly we deal with whitespace. @flodolo?

bedrock/firefox/templates/firefox/ios.html
craigcook: Not intentional, just failed to clean them up as I was adding copy.

media/css/firefox/ios.less
craigcook: If it was a padding or margin I'd agree, but for `background-position` I think it should be a more explicit value.

schalkneethling: Ah yes, very good point. :+1:

bedrock/firefox/templates/firefox/ios.html
flodolo: We strip them on our side of tools (both reference and translations). The [current reference](https://github.com/mozilla-l10n/www.mozilla.org/blob/master/en-US/firefox/ios.lang#L124-L126) string has no trailing whitespaces but the demo pages looks fully localized right now.
https://www-demo1.allizom.org/it/firefox/ios/

If doesn't look like we strip them when we [store source](https://github.com/mozilla/bedrock/blob/master/lib/l10n_utils/dotlang.py#L70) (unlike we do for translations), maybe we should.

flodolo: Tried the patch locally, removed the whitespace and the string remains translated. Not completely sure why it happens, but it seems to work.

flodolo: I shouldn't be allowed to read code at 7AM: it works because we [strip the line](https://github.com/mozilla/bedrock/blob/master/lib/l10n_utils/dotlang.py#L61) before we try to figure out what it is (which also makes the following two `.strip()` on comments and translations not really useful).

schalkneethling: Thanks @flodolo

bedrock/firefox/templates/firefox/ios.html
```diff
@@ -122,6 +144,36 @@
...
+    <p class="lead">
+      {{ _('We believe your personal information belongs to you.') }}
+      {{ _('Firefox for iOS is filled with features to help keep it that way. ') }}
+    </p>
+
+    <ul>
+      <li id="private-browsing">
+        <h3>{{ _('Private Browsing') }}</h3>
+        <p>{{ _('When activated, Firefox for iOS won’t accept cookies, remember passwords or record any sites that you’ve visited. ') }}</p>
+        <a rel="external" data-interaction="outbound link" href="https://support.mozilla.org/kb/private-browsing-browse-web-without-saving-info">{{ _('Learn more about Private Browsing') }}</a>
```
schalkneethling: We can update this now to `{{ url('firefox.private-browsing') }}`

schalkneethling:
A few nits here and there and a IE9-- issue but, looking good

craigcook:
Waiting for some feedback from Ty on the last round of layout changes, hoping to be able to merge this today.

craigcook:
I didn't hear from Ty and I'm on PTO next week, so I'm taking the Do Not Merge flag off this and considering it mergeable (assuming an r+ of course). If he asks for any design tweaks we can add them later.

schalkneethling:
So all is good now with `firefox-ios-global` set to `on` but, when the flag is set to off, the last of the icons in the sync section drops down in all browsers. Seeing that it is in the `off` state, I cannot merge this as the flag will be in the `off` state on production.

With that said, we will not have a production push until Monday PST some time so, I can merge this, and then open a PR that resolves the problem. Once that one is then also merge, we can safely push to production.

@craigcook If you are still around and have a spare moment to address this it would be great but, not a train smash. If by Monday morning, GMT+2, you have not been able to get to it, I will proceed as described above. Thanks!

craigcook:
@schalkneethling Oops, Bookmarks should have been waffled as well. I had removed it for local testing and didn't put it back. Fixed now!

schalkneethling:
r+ :apple:

PR #3397, opened 2015-10-06
Fix bug 1136459 - Add Firefox for iOS to family nav

craigcook:


flodolo:
Just confirming: we already have all 4 strings in main.lang

jpetto:
Layout maintains, though is about to burst at the seams for the longest locale I could find (xh):

screen shot 2015-10-09 at 9 38 35 am

We may want to shrink the Fx logo for tablet view at some point, or add a bit of padding under tabzilla to allow text to flow all the way to the right edge.

Anyway, r+! :iphone:

PR #3402, opened 2015-10-06
Upgrade to product-details 0.8.1

pmclanahan:
We can now turn on p-d storage in the database.

Unfortunately the test suite still requires p-d data so running the tests in DB mode doesn't pass because there's no data in the test db. We could create a test fixture and possibly then remove the need to pull p-d data at travis.


pmclanahan:
@jgmize r?

bedrock/settings/base.py
```diff
@@ -113,6 +113,8 @@ def path(*args):
 PROD_DETAILS_DIR = path('lib', 'product_details_json')
 PROD_DETAILS_CACHE_NAME = 'product-details'
 PROD_DETAILS_CACHE_TIMEOUT = 60 * 15  # 15 min
+PROD_DETAILS_STORAGE = config('PROD_DETAILS_STORAGE',
+                              default='product_details.storage.PDFileStorage')
```
jgmize: Why not default to `PDDatabaseStorage`?

pmclanahan:
Added ***DNM*** tag because I want to add a data migration to product-details lib.

pmclanahan:
This is now updated for p-d 0.8.1 and includes the data migration and recent data in the lib. 

@jgmize r?

bedrock/settings/base.py
```diff
@@ -28,6 +28,7 @@ def path(*args):
 
 # Is this a dev instance?
 DEV = config('DEV', cast=bool, default=False)
+PROD = config('PROD', cast=bool, default=False)
```
pmclanahan: Not sure if this is the way to go, but like we discussed the other day, we could set this in all the prod envs once, then use it for other things like this. I'm open to other suggestions though for sure.

bedrock/settings/base.py
```diff
@@ -110,9 +111,11 @@ def path(*args):
 
 # Tells the product_details module where to find our local JSON files.
 # This ultimately controls how LANGUAGES are constructed.
-PROD_DETAILS_DIR = path('lib', 'product_details_json')
 PROD_DETAILS_CACHE_NAME = 'product-details'
 PROD_DETAILS_CACHE_TIMEOUT = 60 * 15  # 15 min
+default_pdstorage = 'PDDatabaseStorage' if PROD else 'PDFileStorage'
```
jgmize: Remind me why we don't just always default to `PDDatabaseStorage`?

pmclanahan: Just no reason to create all of that confusion with all of the contributors with existing copies. They don't often run `./manage.py migrate`. This way their day is uninterrupted.

jgmize: I'm not sure if "PROD" is the right name for the setting, since I would want to default to the db for dev and demo instances as well, but in the interests of time I'm willing to merge this as-is until we can come up with a better name.

PR #3403, opened 2015-10-06
Update jenkins include_l10n script.

glogiotatidis:
This introduces PROD_ONLY env variable which will be used by jenkins
jobs after l10n to determine execution path: When PROD_ONLY is true
only push_to_stage jobs run and push_to_dev jobs are not run.

PROD_ONLY is true when the job is time triggered and therefore it's
updating only locales and not code. This image with the updated locales
should be pushed to staging and production and not to dev which maybe
already in a newer code version.

Secondly a image build from code merges would run `svn up` and update
the locales making the next time triggered job to believe that there are
no locale changes and therefore skip pushing the latest version to
production. To tackle this FORCE_TIME_TRIGGER_UPDATE gets introduced to
force pushing previously checkout locales to production on time triggers.

 - [ ] Update bedrock_l10n_image to use this script when merged.

### DIFF
--- a/docker/jenkins/include_l10n.sh
+++ b/docker/jenkins/include_l10n.sh
@@ -7,13 +7,16 @@
 set -xe
 
 # Used to trigger downstream Jenkins jobs
+PARAM_FILE=.parameters
 TRIGGER_FILE=.docker-updated
-rm -rf $TRIGGER_FILE
+FORCE_TIME_TRIGGER_UPDATE=.timetriggerupdate
+rm -rf $TRIGGER_FILE $PARAM_FILE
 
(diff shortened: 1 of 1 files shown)


glogiotatidis:
The pipeline runs this code and the last runs are successful tests of the changes

pmclanahan:
Just an FYI: We'll be switching to a github repo for l10n files later this week. See https://bugzilla.mozilla.org/show_bug.cgi?id=1210368#c5 for details.

pmclanahan:
Just as a tl;dr: the prod l10n files will be here:

https://github.com/mozilla-l10n/bedrock-l10n

And dev files will be here:

https://github.com/mozilla-l10n/www.mozilla.org

PR #3405, opened 2015-10-06
Fix bug 1209648 - Update /new page for iOS

craigcook:
I'm reusing the 'Add Firefox to your iPhone, iPad and iPod touch' tagline from the product page, but without a period so the string isn't an exact match.

Also, should I use the `firefox_ios` l10n tag or something else?


bedrock/firefox/templates/firefox/new.html
```diff
@@ -118,12 +120,18 @@ <h1 class="large" id="main-headline">
         <div class="android download-button-wrapper" data-upgrade-subtitle="{{_('Get it free on Google Play')}}">
           {{ download_firefox(platform='android', small=True, icon=False, dom_id='download-button-android') }}
         </div>
+        <div class="ios download-button-wrapper">
+          {{ download_firefox(platform='ios') }}
```
craigcook: Note: this button doesn't work. That'll be a separate PR that will effectively block this one.

flodolo:
The `firefox_ios` tag is fine. Even with the final period we couldn't reuse the existing string, since it's not in a shared file.

bedrock/firefox/templates/firefox/new.html
```diff
@@ -203,6 +218,7 @@ <h1 class="large" id="main-headline">
   <meta itemprop="operatingSystem" content="Mac">
   <meta itemprop="operatingSystem" content="Linux">
   <meta itemprop="operatingSystem" content="Android">
+  <meta itemprop="operatingSystem" content="iOS">
```
jpetto: Not a huge deal, but this should maybe be behind the waffle switch as well?

craigcook: Oh yeah, especially if this is merging almost a month ahead. I shall waffle it.

media/css/firefox/new.less
```diff
@@ -341,7 +341,8 @@ html[lang|="en"] #main-headline span:first-child {
     width: 100%;
 }
 
-.download-button-wrapper.android {
+.download-button-wrapper.android,
+.download-button-wrapper.ios {
```
jpetto: I think we'll need some CSS to display this `.download-button-wrapper.ios` for users *on* iOS. Something like [we do for Android](https://github.com/mozilla/bedrock/blob/master/media/css/firefox/new.less#L274) maybe?

jpetto: Along these lines, do we need to alter the display for users on Fx iOS, like we do for users on Fx Android?

craigcook: Doh, I knew I was forgetting something. With all the branch-juggling I lost track of what was left to do in which. I'll add the CSS to display the button (remembering that the button won't actually display unless you have the updated button from the other branch, but you can also just switch it to android for local testing).

As for versions, I'm not sure we can actually determine latest version of FxiOS yet. It's on a different version number than the other Firefoxes and I don't know what's involved in getting "1.1" to be treated as latest, but only for FxiOS... that may take some extra work on the product details backend and I'm not sure what's in there already at this stage. I'll ask, but I think for this release we may not be able to do any special "you're up to date" or "you need a refresh" logic, that may have to come later.

media/css/firefox/new.less
```diff
@@ -278,6 +278,21 @@ html.android {
     }
 }
 
+/* iOS */
+html.ios {
```
jpetto: This selector already exists way up on line 106.

jpetto:
I might have the `/new` use case fever, but I feel like this PR can simplify instead of complicate. Here are our use cases:

1. **Current** - Users on iOS are immediately redirected to `/firefox/ios/`, meaning it's impossible to see the `.ios-links-wrapper` and `#version-message-ios` content. Users with JS turned off just get the standard desktop platform download buttons.

2. **iOS launch** - Users on iOS will be shown the download button, which will link them to the App Store. Based on current Android behavior, there should be no other links visible (unless I'm unaware of a requirement).

If the above is true, we could avoid adding most of the new conditionals and even remove some old stuff: the `.ios-links-wrapper` and `#version-message-ios` elements, as well as all existing selectors nested in the `html.ios` selector. (Though your additions to this selector would still be required for the download button.)

Does that make sense, or am I [in too deep](http://ecx.images-amazon.com/images/I/51eNT6tfYKL._SY300_.jpg)?

jpetto:
Aside from one image we can likely delete, this is looking great. I don't even see any mcom tests that need updating. r+wc

I noticed the browser image for iOS is identical to that used for Android. Not a blocker for this PR - just wondering if there's been any discussion. Can always update the image later.

craigcook:
Deleted the icon and squashed commits. 

Also filed https://bugzilla.mozilla.org/show_bug.cgi?id=1215743 to get a browser image, which we can just drop in later. However, what with the existing image and our aggressive caching of static files, I assume we also have to rename it or otherwise cache-bust it in the HTML anyway.

jpetto:
All looks good to me. Gareth confirmed GTM change is okay to go live now. It shall be merged.

PR #3406, opened 2015-10-07
Fix Bug 1139961 - Add iOS to universal download button

craigcook:
Props to Kohei for doing most of the work in #2955

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -256,5 +256,11 @@ class FirefoxAndroid(ProductDetails):
         return self.store_url
 
 
+class FirefoxIOS:
+    def get_download_url(self):
+        return settings.APPLE_APPSTORE_FIREFOX_LINK
+
+
 firefox_desktop = FirefoxDesktop()
 firefox_android = FirefoxAndroid()
+firefox_ios = FirefoxIOS()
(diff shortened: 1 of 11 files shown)


flodolo:
Removing tag since this PR doesn't actually have any new strings.

schalkneethling:
Following on from IRC:

Setting the classes on the `html` element to: `ios x86 js non-firefox loaded` on the landing page, causes the button inside the grey bar (below the promos) to look as follows:

screen shot 2015-11-06 at 11 51 07

Note: This is how it is currently in production as well so, not introduced by this PR.

schalkneethling:
Ran into the below whilst testing the following scenario:

I have the `firefox-ios-global` switch set to `off` and opened the landing page in the iOS simulator (set to iPhone 6S[if that makes a difference somehow, shouldn't]). With these I get the following:

screen shot 2015-11-06 at 12 29 45

schalkneethling:
With the switch set to `on` I see the following:

screen shot 2015-11-06 at 12 35 58

craigcook:
With the waffle switch off you should see only the coming soon message and not the button, but I'm seeing both as well. I'll work on it.

craigcook:
Alright, I think I sorted out how to keep the button hidden when the switch is off, though it meant tacking on an extra class as a handy toggle.

jpetto:
Closing in favor of rebased #3542.

PR #3407, opened 2015-10-07
[fix bug 1211907] Remove /firefox/independent and /firefox/personal pages

alexgibson:


alexgibson:
CC @flodolo re tidying up any lang files

schalkneethling:
There was one reference but, you would not pick it up as it is in the new campaign landing page but, I fixed that by copying the assets into the new landing page asset folder. r+ :balloon:

PR #3408, opened 2015-10-07
Do not include product details in docker image.

glogiotatidis:
Merge after #3402


jgmize:
Thanks @glogiotatidis. Merged #3402 earlier.

PR #3409, opened 2015-10-07
Create initial functional tests using pytest

davehunt:
This is the first step in replacing the CasperJS tests with pytest tests. Once this has landed we'll work our way through the CasperJS tests and either migrate them to this stack or remove them. We'll then do the same with the mcom-tests.

Feedback would be appreciated from @bobsilverberg, @alexgibson, @pmclanahan and @jgmize. The plan is to run these in Web QA's Jenkins instance in addition to the mcom-tests, but I'd like to also get these running in the new Jenkins pipeline, which @glogiotatidis may be able to help with.


docs/testing.rst
```diff
@@ -8,148 +8,176 @@
...
-* ``config.json`` a configuration file used by PhantomJS when running tests.
 
 Installation
 ------------
 
-The specific versions of CasperJS, PhantomJS and Jasmine/Karma which are needed to run the
-tests, can be installed by running the following command from the root directory of the project::
+First follow the :ref:`installation instructions for bedrock<install>`, which
+will install the specific versions of Jasmine/Karma which are needed to run the
+unit tests, and guide you through installing pip and setting up a virtual
```
alexgibson: should this be `installing peep` instead of `installing pip`?

docs/testing.rst
```diff
@@ -8,148 +8,176 @@
...
-    npm test -- tests/functional/home.js --config=tests/config.json
+To run a single test file you must tell py.test to execute a specific file
+e.g. ``tests/functional/test_newsletter.py``:
 
-You can also easily run the tests against any bedrock environment by specifying the
-``--domain`` variable. For example, to run all functional tests against dev::
+.. code-block:: bash
 
-    npm test -- tests/functional --domain=https://www-dev.allizom.org --config=tests/config.json
+    $ py.test --driver Firefox --html results.html -n auto tests/functional/test_newsletter.py
```
alexgibson: When I try to run this line I get the following error:

    usage: py.test [options] [file_or_dir] [file_or_dir] [...]
    py.test: error: unrecognized arguments: -n

If I omit `-n auto` the tests seem to run fine

alexgibson:
Is it possible to specify the output directory for the HTML report? Perhaps it could live under `/tests/functional`, and we could add it to .gitignore?

alexgibson:
Looks like Travis is failing because `python manage.py test` is trying to run the Selenium tests as well as the unit tests?

alexgibson:
@davehunt - this looks great! Couple of niggles, but tests are running and passing for me locally. Great job on the docs :+1:

alexgibson:
I also got a `pytest` warning, but I assume this is ok for now:

    WI1 /Users/agibson/Sites/bedrock/venv/lib/python2.7/site-packages/pytest_selenium/pytest_selenium.py:89 'pytest_runtest_makereport' hook uses deprecated __multicall__ argument

docs/testing.rst
davehunt: peep doesn't need to be installed as it's in bedrock/bin, and peep depends on pip, so that does still need to be installed.

docs/testing.rst
davehunt: You probably don't have pytest-xdist installed. I added it to the requirements file. While it's not strictly required, it allows you to run the tests in multiple processes, and therefore in parallel.

alexgibson: If I type `pip freeze` it lists `pytest-xdist==1.13.1`. I'll try reinstalling and see if that helps

docs/testing.rst
alexgibson: ah ok, thanks :+1:

docs/testing.rst
alexgibson: I uninstalled `pytest-xdist` and then reinstalled requirements but still the same error?

davehunt:
Updated based on your comments @alexgibson and hopefully fixed the Travis issue.

davehunt:
The pytest warning should be safe to ignore. I suspect pytest-selenium will support the latest and previous versions of pytest, so will probably drop 2.6.x soon, and once the next version after 2.8.x is out we can remove the reference to the deprecated argument.

requirements/test_functional.txt
```diff
@@ -0,0 +1,8 @@
+# sha256: bwGLPxbC1cSQ-u-cot0Pk4LpMPi3gDp1cVGMTDQE78o
+pytest==2.8.1
```
alexgibson: `test_redirects.txt` uses `pytest 2.7.2` - should we be using the same version for consistency?

alexgibson:
Seems the only way I can get all the dependencies installed to run the functional tests was to use `pip` instead of `peep`. This was using a fresh venv.

    pip install -r requirements/test_functional.txt

As soon as I did this, everything runs fine. I wonder if peep isn't installing all the required dependencies?

pmclanahan:
> Seems like peep isn't installing everything that's needed?

peep does indeed not install anything you don't tell it to. you should pip install into a fresh venv then pip freeze to get the full requirements list.

tests/functional/test_newsletter.py
```diff
@@ -0,0 +1,38 @@
...
+    page = NewsletterPage(base_url, selenium).open()
+    assert '' == page.email
+    assert 'United States' == page.country
+    assert 'English' == page.language
+    assert page.html_format_selected
+    assert not page.text_format_selected
+    assert not page.privacy_policy_accepted
+
+
+@pytest.mark.nondestructive
```
bobsilverberg: Is this really nondestructive? I would have thought that signing up for a mailing list was destructive.

tests/pages/newsletter.py
```diff
@@ -0,0 +1,87 @@
...
+        el = self.selenium.find_element(*self._language_locator)
+        Select(el).select_by_visible_text(value)
+
+    @property
+    def html_format_selected(self):
+        el = self.selenium.find_element(*self._html_format_locator)
+        return el.is_selected()
+
+    def select_html_format(self):
+        self.selenium.find_element(*self._html_format_locator).click()
```
bobsilverberg: For `accept_privacy_policy`, you ensure that the privacy policy is not already selected, but for `select_html_format` (and `select_text_format`) you do not do the same. Is that intentional? Are these meant as toggles which will simply change the state regardless of the current state?

tests/pages/page.py
```diff
@@ -0,0 +1,42 @@
...
+
+    _root_locator = None
+
+    def __init__(self, selenium, root=None):
+        self.selenium = selenium
+        self.timeout = TIMEOUT
+        self.root = root
+
+        if self.root is None and self._root_locator is not None:
+            self.root = self.selenium.find_element(*self._root_locator)
```
bobsilverberg: So you could end up with `self.root` == `None`, correct? Is that okay? Is it up to the PageRegion developer to either ensure that isn't the case or to make sure they don't use `self.root`? In the past I think we defaulted `self.root` to `selenium` so that `self.root` was always available. I'm not saying that's what we should do, but I just wanted to check your thinking on this.

bobsilverberg:
@davehunt LGTM. I asked a few questions above.

requirements/test_functional.txt
davehunt: We could potentially even combine these and have a tests.txt, what do you think?

tests/functional/test_newsletter.py
davehunt: I agree, what do you think @alexgibson? This would mean the test wil be skipped for sensitive URLs. I'll also add a regex to setup.cfg to indicate sensitive URLs. Would `mozilla\.(com|org)` be suitable?

tests/pages/newsletter.py
davehunt: These are radio buttons, so additional clicks do not toggle the selected state.

tests/pages/page.py
davehunt: I'm not keen on `self.root` defaulting to `self.selenium`, though perhaps we could default it to the document root (if that's even possible?). I was thinking that root is totally optional, and for page regions that don't specify it, using `region.root` would not make sense. Instead those page regions should use `self.selenium` directly.

davehunt:
@alexgibson I've updated the requirements file for peep, however having all packages pinned means they'll get out of date regularly. I would recommend using something like https://requires.io/github/mozilla/bedrock/requirements/?branch=master to keep on top of them.

tests/functional/test_newsletter.py
alexgibson: Yes I think it makes sense to mark this as destructive :+1:

alexgibson: Also maybe worth adding a note to the guidelines about which type of test should generally not be marked as `nondestructive`?

alexgibson:
> @alexgibson I've updated the requirements file for peep, however having all packages pinned means they'll get out of date regularly. I would recommend using something like https://requires.io/github/mozilla/bedrock/requirements/?branch=master to keep on top of them.

I think there's still a dependency missing when installing with peep (I'm still seeing `py.test: error: unrecognized arguments: -n`).

As for the suggestion to use something like `requires.io`, this sounds good to me. Although I guess this scenario is true for any project that specifically locks down dependencies with something like peep.

As for the requirements file, my only concern is we currently have pytest v2.7.2 in `test_redirects.txt` and `v2.8.2 in `test_functional.txt`. Would it be better for everything to be on a single version? Combining the two requirements files may be a simple solution.

 Any thoughts or strong feeling on the above, @pmclanahan?

tests/functional/test_newsletter.py
davehunt: Good idea. I'm going to enhance pytest-selenium to accept a sensitive url regex in setup.cfg. At the moment it requires either either a command line argument or environment variable. Supporting setup.cfg means it can easily be configured for a project and stored in the repo.

alexgibson:
As of the latest commit this all installs and runs fine for my using peep in a fresh venv :+1:

tests/functional/test_newsletter.py
davehunt: Documentation updated.

schalkneethling:
Installation was quick and easy. Running the tests yields the following though:
https://sneethling.pastebin.mozilla.org/8848680

Is this expected? Did I miss a step? @davehunt

davehunt:
@schalkneethling I believe the test failed because you do not have the localisations installed. I found this too with a local instance, and @alexgibson pointed me towards some doc for installing the locales.

pmclanahan:
@alexgibson in an ideal world yes, we'd have these requirements synced. But for now it really depends on if they'll all be run by the same system. If these are to be run in travis e.g., then I'd say yes we should sync them and put them in the same file even. If not, then it's fine for now to leave them separate.

schalkneethling:
I cloned the locales, reran the tests and all is good.

pmclanahan:
We just fixed this in master. l10n files are no longer required for passing
tests (at least the django tests).

davehunt:
@pmclanahan do you mean you're not running tests that require l10n, or that the tests are running but don't need to pass? This particular test is checking the functionality of changing the language of the home page, so it will need at least one alternative locale to be present to pass.

pmclanahan:
Ah yes. My problem is that I was thinking of the other test suite. Please ignore me :)

alexgibson:
@davehunt - how about we combine the two testing requirements files into a combined `testing.txt`. Then we can at least ensure they are both using the same pytest version if someone wants to run both suites of tests? Other than that, I think we're all good here :+1:

davehunt:
Rebased and squashed. I also added a default locale (en-US) and the ability to override when creating a page object instance. This is because I noticed different locators for the footer newsletter sign up from en-GB. /cc @alexgibson @bobsilverberg @m8ttyB @schalkneethling @pmclanahan @jgmize @glogiotatidis.

alexgibson:
@davehunt - looks like travis is failing because some of the hashes don't match the versions in the updated requirements?

davehunt:
Yeah, I'm not sure why Travis is getting different hashes to me..?

bobsilverberg:
@davehunt If you are not already familiar with it, you could try `peepin` [1], a tool written by our own beloved @peterbe, which will generate the appropriate hashes for you. I'm not sure if this will resolve the differences between your local install and Travis, but it might be worth trying.

[1] https://pypi.python.org/pypi/peepin

pmclanahan:
`peepin` +1. It'll get hashes for all of the packages. The issue is that PyPI can have different files for the same package (e.g. .tar.gz, .zip, .wheel, etc.). These will all have different hashes, and different versions of pip prefer different ones.

davehunt:
Okay, the difference was the `--no-wheel` argument that Travis is passing. I've updated the requirements file to support when this argument is present. I had problems using `peepin` with an existing requirements file, which I'll raise separately.

alexgibson:
Travis is happy, everything installs fine for me using a fresh venv, tests all pass! r+ :beers:

PR #3410, opened 2015-10-07
Symlink static files.

glogiotatidis:
Uses `collectatic -l` and reduces the image size by [~140mb](https://imagelayers.io/?images=mozorg%2Fbedrock:7d1d934dd94d9946fc6ce20f461e5c207edb3ab8,mozorg%2Fbedrock:2725cbd0b0abacb2cac78123a3432c9b0ae48f75)

### DIFF
--- a/docker/bin/softlinkstatic.py
+++ b/docker/bin/softlinkstatic.py
@@ -10,8 +10,9 @@ with open(STATIC) as static_fp:
 for orig, hashed in static_files['paths'].items():
     if '?' in orig:
         continue
-    orig_path = os.path.join('./static', orig)
-    hashed_filename = os.path.split(hashed)[1]
+    hashed_path = os.path.join('./static', hashed)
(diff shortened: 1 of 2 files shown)


pmclanahan:
Dig it. Would be nice if pipeline respected the `-l` flag as well, but this works well in my tests.

We'll have to be aware of this when we want to keep old hashed files around for cache reasons on the new infra. Since the hashed files are just links their content can now be updated. We'll figure that out later when we need to do this though. Just wanted to implant that notion in your brain :)

Nice work! :cake:

glogiotatidis:
> We'll have to be aware of this when we want to keep old hashed files around for cache reasons on the new infra. Since the hashed files are just links their content can now be updated. We'll figure that out later when we need to do this though. Just wanted to implant that notion in your brain :)

good point, I'll keep in my mind.

Thanks!

PR #3411, opened 2015-10-07
Set PYTHONDONTWRITEBYTECODE in dockerfiles.

glogiotatidis:


docker/dockerfiles/bedrock_base
```diff
@@ -17,9 +17,11 @@ RUN pip install -r requirements/pip.txt
 
 # Install app
 COPY ./bin/peep.py /app/bin/peep.py
+ENV PYTHONDONTWRITEBYTECODE=1
 RUN ./bin/peep.py install --no-cache-dir -r requirements/dev.txt
 RUN ./bin/peep.py install --no-cache-dir -r requirements/prod.txt
 RUN ./bin/peep.py install --no-cache-dir -r requirements/docker.txt
+ENV PYTHONDONTWRITEBYTECODE=0
```
jgmize: Why are you setting it back to 0? Why not just set it to 1 and then leave it?

glogiotatidis: Don't we benefit from pyc files in production?

jgmize: since pyc files [only help at startup time](http://www.network-theory.co.uk/docs/pytut/CompiledPythonfiles.html), and our startup time delays are dominated by other factors, I would argue that no, we don't actually get any benefit from them so we might as well not have them in the docker images.

glogiotatidis: We're right @jgmize, I'd a nebulous understanding of pyc files. Thanks!

jgmize:
Thanks again, @glogiotatidis

PR #3412, opened 2015-10-08
Fix Bug 1205858, adds new default contact us tab to contact page

schalkneethling:
@jpetto r?

@flodolo I am pretty sure this is not localized but, just putting an l10n label on this in case. If indeed it is not needed, you can just remove the label. Thanks!

### DIFF
--- a/bedrock/mozorg/templates/mozorg/contact/contact-landing.html
+++ b/bedrock/mozorg/templates/mozorg/contact/contact-landing.html
@@ -1,0 +1,65 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% extends "mozorg/contact/contact-base.html" %}
+
+{% block page_title %}{{ _('Contacts, Spaces and Communities') }} — {{ _('Contact us') }}{% endblock %}
+
+{% block contact_entry %}
+  <div id="entry-container">
+    <section class="entry contact-us" id="contact" data-tab="contact">
+      <header>
+        <h2>{{ _('Contacting Mozilla')}}</h2>
+      </header>
+
+      <ul>
+        <li><a href="https://support.mozilla.org/">{{ _('I’m having problems with using Firefox')}}</a></li>
+        <li><a href="https://sendto.mozilla.org/page/contribute/givenow-seq?preset=2&ref=Donation_Form&utm_campaign=Donation_Form&utm_source=mozilla.org&utm_medium=referral&utm_content=contact_us_page">{{ _('I want to donate to Mozilla')}}</a></li>
+        <li><a href="{{ url('foundation.trademarks.policy') }}">{{ _('I have questions about copyright trademarks and Mozilla')}}</a></li>
+        <li><a href="mailto:research_requests@mozilla.com">{{ _('I need information for an academic paper')}}</a></li>
+        <li><a href="https://wiki.mozilla.org/People:MozSpaces_Guidelines">{{ _('I want to hold an event in a Mozilla space')}}</a></li>
+        <li><a href="https://bugzilla.mozilla.org/form.dev-engagement-event">{{ _('I want Mozilla to sponsor my event')}}</a></li>
+        <li><a href="{{ url('styleguide.identity.mozilla-branding') }}">{{ _('I have a question about using Mozilla logos')}}</a></li>
+      </ul>
+
+      <p>
+      {% trans link_sumo='https://support.mozilla.org' %}
+        To reach Mozilla anywhere in the world, check out the locations above. For help
+        with any of our products, including Firefox for desktop and Android, Firefox OS
+        and Thunderbird, please <a href="{{ link_sumo }}">visit our support site</a>.
+      {% endtrans %}
+      </p>
+
+      <div class="contact-online">
+        <h4>{{ _('Find us online') }}</h4>
+        <ul>
+          <li><a href="https://facebook.com/mozilla" hreflang="en-US" rel="external">{{ _('Facebook') }}</a></li>
+          <li><a href="https://twitter.com/mozilla" hreflang="en-US" rel="external">{{ _('Twitter') }}</a></li>
+          <li><a href="{{ url('mozorg.about.forums.forums') }}">{{ _('Forums &amp; e-mail lists') }}</a></li>
+          <li><a href="https://wiki.mozilla.org/IRC">{{ _('IRC channels') }}</a></li>
+        </ul>
+      </div>
+
+      <div class="contact-join">
+        <h4>{{ _('Join us') }}</h4>
+        <ul>
+          <li><a href="{{ url('mozorg.contribute') }}">{{ _('Get involved') }}</a></li>
+          <li><a href="https://careers.mozilla.org">{{ _('Career opportunities') }}</a></li>
+          <li><a href="https://mozillians.org">{{ _('Mozillians') }}</a></li>
+        </ul>
+      </div>
+
+      <div class="contact-help">
+        <h4>{{ _('Questions &amp; feedback') }}</h4>
+        <ul>
+          <li><a href="https://support.mozilla.org/">{{ _('Get support') }}</a></li>
+          <li><a href="{{ url('foundation.licensing') }}">{{ _('Browser distribution &amp; licensing') }}</a></li>
+          <li><a href="https://input.mozilla.org/feedback">{{ _('Firefox feedback') }}</a></li>
+          <li><a href="https://support.mozilla.org/kb/websites-look-wrong-or-appear-differently">{{ _('Report a broken site') }}</a></li>
+          <li><a href="https://developer.mozilla.org/docs/Mozilla/QA/Bug_writing_guidelines">{{ _('How to report software issues') }}</a></li>
+        </ul>
+      </div>
+    </section>
+  </div>
+{% endblock %}
--- a/media/css/mozorg/contact-spaces.less
+++ b/media/css/mozorg/contact-spaces.less
@@ -146,15 +146,23 @@
     position: relative;
 }
 
-// Tabs to toggle spaces or communities
+// Tabs to toggle contact us, spaces or communities
 .category-tabs {
     .open-sans;
     .font-size(@largeFontSize);
-    left: 22px;
+    left: 18px;
     margin: 0;
     position: absolute;
     text-align: center;
     top: -40px;
+
+    @media only screen and (max-width: @breakTablet) {
+        .font-size(@baseFontSize - 2);
+    }
+
+    @media only screen and (max-width: @breakMobileLandscape) {
+        left: 15px;
+    }
 
     li {
         .inline-block;
@@ -162,6 +170,23 @@
         position: relative;
         right: 8px;
         width: 218px;
+
+        @media only screen and (max-width: @breakTablet) {
+            width: 145px;
+
+            &.current a {
+                background-position: 100% -60px;
+            }
+        }
+
+        @media only screen and (max-width: @breakMobileLandscape) {
+            left: 0;
+            width: 95px;
+
+            &:first-child {
+                left: 0;
+            }
+        }
 
         a {
             background: url(/media/img/firefox/pager-tabs.png) 0 0 no-repeat;
@@ -171,16 +196,18 @@
             position: relative;
             z-index: 10;
 
+            @media only screen and (max-width: @breakTablet) {
+                background: #fff none;
+                border-top-left-radius: 8px;
+                border-top-right-radius: 8px;
+            }
+
         }
 
         &.current a {
             background-position: 100% 0;
             color: @textColorSecondary;
             z-index: 20;
-        }
-
-        &:first-child {
-            left: 8px;
         }
     }
 }
@@ -435,6 +462,14 @@
             margin: 0 auto;
         }
     }
+}
+
+/* override float rule for contact us tab */
+.entry.contact-us {
+    float: none;
+    border-top: none;
+    width: 100%;
+    .clearfix();
 }
 
 .entry-space .card {
@@ -740,23 +775,6 @@
 
     #page-content {
         margin-top: 40px;
-    }
-
-    .category-tabs {
-        .font-size(@baseFontSize);
-        left: 0;
-
-        li {
-            width: 145px;
-
-            a {
-                background: url(/media/img/firefox/pager-tabs.png) 0 -60px no-repeat;
-            }
-
-            &.current a {
-                background-position: 100% -60px;
-            }
-        }
     }
 
     .entry {
--- a/media/js/mozorg/contact-spaces.js
+++ b/media/js/mozorg/contact-spaces.js
@@ -114,13 +114,13 @@
 
             //set the current tab navigation item
             $('ul.category-tabs li[data-id="' + tab + '"]').addClass('current');
-
-            // set the current list menu navigation item
             $('.nav-category li[data-id="' + id + '"]').addClass('current');
 
-            // hide the community sub menu's
-            $('.accordion .submenu').hide();
-            mozMap.toggleCommunitySubMenu();
+            if (tab === 'communities') {
+                // hide the community sub menu's
+                $('#nav-communities .submenu').hide();
+                mozMap.toggleCommunitySubMenu();
+            }
         },
 
         /*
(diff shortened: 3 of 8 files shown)


flodolo:
Confirming, not localized.

bedrock/mozorg/templates/mozorg/contact/contact-landing.html
```diff
@@ -0,0 +1,66 @@
...
+        <li><a href="https://support.mozilla.org/">{{ _('I’m having problems with using Firefox')}}</a></li>
+        <li><a href="https://sendto.mozilla.org/page/contribute/givenow-seq?preset=2&ref=Donation_Form&utm_campaign=Donation_Form&utm_source=mozilla.org&utm_medium=referral&utm_content=contact_us_page">{{ _('I want to donate to Mozilla')}}</a></li>
+        <li><a href="https://www.mozilla.org/foundation/trademarks/policy/">{{ _('I have questions about copyright trademarks and Mozilla')}}</a></li>
+        <li><a href="mailto:research_requests@mozilla.com">{{ _('I need information for an academic paper')}}</a></li>
+        <li><a href="https://wiki.mozilla.org/People:MozSpaces_Guidelines">{{ _('I want to hold an event in a Mozilla space')}}</a></li>
+        <li><a href="https://bugzilla.mozilla.org/form.dev-engagement-event">{{ _('I want Mozilla to sponsor my event')}}</a></li>
+        <li><a href="https://www.mozilla.org/styleguide/identity/mozilla/branding/">{{ _('I have a question about using Mozilla logos')}}</a></li>
+        <li><a href="#press_enquiry" id="press_enquiry_toggle">{{ _('I’m a member of the Press and have a question for Mozilla') }}</a></li>
```
jpetto: Should this `<li>` be removed until the press contact form is completed ([bug 1210426](https://bugzilla.mozilla.org/show_bug.cgi?id=1210426))?

bedrock/mozorg/templates/mozorg/contact/contact-landing.html
```diff
@@ -0,0 +1,66 @@
...
+  <div id="entry-container">
+    <section class="entry contact-us" id="contact" data-tab="contact">
+      <header>
+        <h2>{{ _('Contacting Mozilla')}}</h2>
+      </header>
+
+      <ul>
+        <li><a href="https://support.mozilla.org/">{{ _('I’m having problems with using Firefox')}}</a></li>
+        <li><a href="https://sendto.mozilla.org/page/contribute/givenow-seq?preset=2&ref=Donation_Form&utm_campaign=Donation_Form&utm_source=mozilla.org&utm_medium=referral&utm_content=contact_us_page">{{ _('I want to donate to Mozilla')}}</a></li>
+        <li><a href="https://www.mozilla.org/foundation/trademarks/policy/">{{ _('I have questions about copyright trademarks and Mozilla')}}</a></li>
```
jpetto: Use the `url` helper here?

bedrock/mozorg/templates/mozorg/contact/contact-landing.html
```diff
@@ -0,0 +1,66 @@
...
+        <li><a href="mailto:research_requests@mozilla.com">{{ _('I need information for an academic paper')}}</a></li>
+        <li><a href="https://wiki.mozilla.org/People:MozSpaces_Guidelines">{{ _('I want to hold an event in a Mozilla space')}}</a></li>
+        <li><a href="https://bugzilla.mozilla.org/form.dev-engagement-event">{{ _('I want Mozilla to sponsor my event')}}</a></li>
+        <li><a href="https://www.mozilla.org/styleguide/identity/mozilla/branding/">{{ _('I have a question about using Mozilla logos')}}</a></li>
```
jpetto: And here?

bedrock/mozorg/templates/mozorg/contact/contact-landing.html
```diff
@@ -0,0 +1,66 @@
...
+        with any of our products, including Firefox for desktop and Android, Firefox OS
+        and Thunderbird, please <a href="{{ link_sumo }}">visit our support site</a>.
+      {% endtrans %}
+      </p>
+
+      <div class="contact-online">
+        <h4>{{ _('Find us online') }}</h4>
+        <ul>
+          <li><a href="https://facebook.com/mozilla" hreflang="en-US">{{ _('Facebook') }}</a></li>
+          <li><a href="https://twitter.com/mozilla" hreflang="en-US">{{ _('Twitter') }}</a></li>
```
jpetto: Put a `rel="external"` on these two social links?

bedrock/mozorg/templates/mozorg/contact/contact-landing.html
```diff
@@ -0,0 +1,66 @@
...
+          <li><a href="/about/forums/">{{ _('Forums &amp; e-mail lists') }}</a></li>
```
jpetto: Use `url` helper here?

bedrock/mozorg/templates/mozorg/contact/contact-landing.html
schalkneethling: Indeed, completely neglected to remove it.

bedrock/mozorg/templates/mozorg/contact/contact-landing.html
schalkneethling: Yeah, good point. This was all copied from preexisting code but, no reason not to improve it ;)

media/js/mozorg/contact-spaces.js
```diff
@@ -115,12 +115,16 @@
             //set the current tab navigation item
             $('ul.category-tabs li[data-id="' + tab + '"]').addClass('current');
 
-            // set the current list menu navigation item
-            $('.nav-category li[data-id="' + id + '"]').addClass('current');
+            if (tab === 'spaces') {
+                // set the current list menu navigation item for the active space
+                $('.nav-category li[data-id="' + id + '"]').addClass('current');
```
jpetto: I believe this line is still needed for both `spaces` and `communities`, as both have side navs.

media/js/mozorg/contact-spaces.js
```diff
@@ -115,12 +115,16 @@
...
+            }
 
-            // hide the community sub menu's
-            $('.accordion .submenu').hide();
-            mozMap.toggleCommunitySubMenu();
+            if (tab === 'communities') {
+                // hide the community sub menu's
+                $('.accordion .submenu').hide();
```
jpetto: I know the logic isn't changed in this PR, but I feel this line would be clearer if targeted at the `nav-communities` ID, as it appears the intent is to affect a specific element and not all `.accordion` elements. Thoughts?

Also, looks like a conditional around these two `communities` related lines should have been there before this PR, as they shouldn't need to fire for `spaces` or `contact`. :+1:

media/js/mozorg/contact-spaces.js
```diff
@@ -778,6 +805,30 @@
         },
 
         /*
+         * Show the contact us page content.
+         * Determined using data-id attribute and .current list item.
+         */
+        showContactUsContent: function(url, cacheId) {
+            var $current = $('.category-tabs li.current');
+            var id = $current.data('id');
```
jpetto: Doesn't look like this local variable is used.

media/js/mozorg/contact-spaces.js
```diff
@@ -778,6 +805,30 @@
...
+            var contentUrl = url || current.prop('href');
```
jpetto: Missing `$` in front of `current`.

media/css/mozorg/contact-spaces.less
```diff
@@ -437,6 +437,11 @@
     }
 }
 
+/* override float rule for contact us tab */
+.entry.contact-us {
+    float: none;
```
jpetto: Should include the `.clearfix()` mixin here, as contains floated content?

alexgibson:
Before merging this it might be a good idea to double-check mcom-tests to see if anything needs updating :+1:

jpetto:
Looks like there's a bug on mobile, as the select boxes for both Spaces and Communities are visible when toggling back to the "Contact" tab.

screen shot 2015-10-08 at 1 39 32 pm

Also, desktop & tablet views looks a *little* bit off due to the amount of white space on the right side of the content. Since there's no side nav, perhaps we should remove the width restriction for the Contact content?

media/css/mozorg/contact-spaces.less
schalkneethling: Did not run into any floating issues without the mixin so, perhaps it is not required?

schalkneethling:
@jpetto addressed all comments. @alexgibson so, this one will be blocked until the mcom tests are updated?

This, for example, will need to be updated:
https://github.com/mozilla/mcom-tests/blob/master/tests/test_contact.py#L29

alexgibson:
I think I might have spotted a history bug in this change, but will double-check soon to see if I can reproduce (currently mobile). Yes I think we'll likely need to update mcom-tests for this change too.

jpetto:
Hm, found another layout issue - tabs are unevenly spaced at mobile wide:

screen shot 2015-10-08 at 3 37 25 pm

Could probably get rid of that top border for Contact mobile view as well, as there's no select box up top to separate.

media/css/mozorg/contact-spaces.less
jpetto: The `.clearfix` on the parent `.container` is saving us from layout issues. If we had content underneath "Questions & feedback" however, there would be issues. I tend to err on the side of clearing floats. Doesn't hurt anything. :wink:

media/js/mozorg/contact-spaces.js
jpetto: Still need to remove the conditional around this line?

jpetto: Ah, I see it's repeated below. Would be easier/more DRY to just have the line once, not wrapped in a conditional though, yeah?

media/js/mozorg/contact-spaces.js
```diff
@@ -115,12 +115,17 @@
...
+                // set the current list menu navigation item for the active space
+                $('.nav-category li[data-id="' + id + '"]').addClass('current');
+            }
 
-            // hide the community sub menu's
-            $('.accordion .submenu').hide();
-            mozMap.toggleCommunitySubMenu();
+            if (tab === 'communities') {
+                // hide the community sub menu's
+                $('#nav-communities').hide();
```
jpetto: Ah, but this still has to target the child `.submenu`'s. Should have been more clear in my previous comment.

jpetto:
Jeez, just noticed the tabs are unevenly spaced at desktop size as well. So subtle (and noisy with map bg) that I didn't notice it at first.

screen shot 2015-10-08 at 4 13 55 pm

media/js/mozorg/contact-spaces.js
```diff
@@ -311,6 +322,8 @@
                 $('#nav-spaces, #meta-spaces').fadeOut(100, function () {
                     $('#nav-communities, #meta-communities').show();
                 });
+            } else if (tab === 'contact') {
+                $('#nav-spaces, #meta-spaces, #nav-communities, #meta-communities').fadeOut(100);
```
jpetto: For consistency, let's keep the logic uniform and make the condition here `(tab === 'contact' && wideMode)`. Then we can simplify/ the `toggleMobileNav` function below to be:

```
} else if (tab === 'contact' && !wideMode) {
    $('#nav-communities-select, #meta-communities, #nav-spaces-select, #meta-spaces').hide();
}
```

jpetto:
I'm seeing a history bug - possibly the one @alexgibson noticed. Load `/contact/`, click the "Spaces" tab, wait for content to load, click back button. The content does not switch back to "Contact". If you go from "Spaces" to "Communities" (or vice versa), the content updates correctly using the back/forward buttons.

alexgibson:
> I'm seeing a history bug - possibly the one @alexgibson noticed. Load /contact/, click the "Spaces" tab, wait for content to load, click back button. The content does not switch back to "Contact". If you go from "Spaces" to "Communities" (or vice versa), the content updates correctly using the back/forward buttons.

Yep this is what I'm seeing too. Specifically, I think the new code needs to handle [this condition](https://github.com/mozilla/bedrock/blob/master/media/js/mozorg/contact-spaces.js#L160-L161)

schalkneethling:
Ok, PR opened against mcom tests https://github.com/mozilla/mcom-tests/pull/443

media/js/mozorg/contact-spaces.js
schalkneethling: :+1:

alexgibson:
I did a bit of debugging locally on the history bug and I think these areas may still need updating:

https://github.com/mozilla/bedrock/blob/master/media/js/mozorg/contact-spaces.js#L534-L538
https://github.com/mozilla/bedrock/blob/master/media/js/mozorg/contact-spaces.js#L395

Also, we have a much [simplified version](https://github.com/mozilla/bedrock/blob/master/media/js/mozorg/contact-spaces-ie7.js) for older IE browsers. Does this need updating at all for this new tab?

media/js/mozorg/contact-spaces.js
jpetto: Just a reminder to switch this to `$('#nav-communities .submenu').hide();` in the next update.

schalkneethling:
@jpetto test all the things but, I need to update the IE bits that @alexgibson mentioned above. Was not even aware of that. Thanks Alex

jpetto:
The history bug seems to be fixed. :+1: 

However, during testing, I noticed something odd. On any initial page load (`/contact/`, `/contact/spaces/`, `/contact/communities/`), the JS wipes out the main content, then an XHR request is made requesting the exact same page, which re-populates the main content. Essentially, it looks like the mechanism for smoothly swapping content when clicking tabs is operating at page load as well, resulting in a totally unnecessary request/performance hit.

Check the network tab in dev tools and you'll see an XHR request to the same page initially loaded. :confounded: 

I believe these three functions are all culprits:

`showSpace`
`showCommunityContent`
`showContactUsContent`

This behavior doesn't look to be introduced with this PR, but it sure would be nice to fix it up soon. Could be in this PR, or file a new bug if other matters are pressing.

alexgibson:
@jpetto - nice spot. Not sure when this crept in, but you're right we should fix it. Probably just need to add a check to make sure the ajax doesn't get called on page load :+1:

schalkneethling:
Thanks @jpetto I will resolve that in this PR as well.

schalkneethling:
@jpetto updated. Let me know if there are other items you run into. Thanks!

jpetto:
The last thing I see is that the initial content is not cached, which can cause an extra XHR request. If I load `/contact/` and click the "Spaces" tab, an XHR request is made to get the "Spaces" content. All good there. If I then click the back button or the "Contact" tab, another XHR request is made to get the "Contact" content (even though that content was retrieved on initial page load).

Maybe we could factor out the [caching mechanism from the AJAX request](https://github.com/mozilla/bedrock/blob/master/media/js/mozorg/contact-spaces.js#L1078-L1084) and use it in the [`setInitialContentState`](https://github.com/mozilla/bedrock/blob/master/media/js/mozorg/contact-spaces.js#L395) function to cache initially loaded content?

schalkneethling:
@jpetto So, it seems the cacheid is not being set, let me look into it. Will update the PR and let you know. Thanks!

schalkneethling:
@jpetto yes, so as you mentioned and I just confirmed, the cacheId is not set and hence the ajax request happens. Now the problem is, on initial load there are no ajax requests as the appropriate page is loaded directly.

I guess the way to get this to work is to change the way the page works overall in that the content is always loaded via Ajax, even on initial load i.e. the same page is loaded for `/contact/`, `/contact/spaces/` and 'contact/communities/` but, based on the URL, the appropriate content is loaded via Ajax thus, setting the cacheId from initial load.

This should probably be part of the refactor that is on the cards for the very near future. Perhaps there is a stop gap I am overlooking that we can implement in this bug but else, it is probably out of scope for this bug. Thoughts? Thanks!

jpetto:
I *think* the fix could be in factoring out the caching mechanism from the AJAX request, then using it on page load *and* AJAX requests. From what I've seen of the code, AJAX isn't required to populate the cache, but they are tightly coupled at the moment.

However, I feel the page is at a good state in this PR. The redundant XHR request at page load has been removed, which is a big win (and was also outside the scope of this bug). In the interest of getting this new tab out the door, I'd be fine with opening another bug to finish up polishing the caching performance.

Just needs a squash and we'll be good to go. r+! :rocket:

jpetto:
Also, just noting that [this pre-existing rewrite rule](https://github.com/mozilla/bedrock/blob/master/etc/httpd/global.conf#L298) should handle this new page:

`RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?contact(.*)$ /b/$1contact$2 [PT]`

schalkneethling:
Awesome, thx for the review @jpetto Squashed. With regards to the rewrite rule yes, it will work just fine. There was a redirect that would redirect requests to `spaces/` but is removed here.

So, let's merge this first. I will then merge master onto the inquiry page, add the link and we can merge that one.

PR #3413, opened 2015-10-08
Move to using npm-lockdown for Node dependency management.

pmclanahan:
* Remove node modules that were checked into the repo.
* Add npm-lockdown to the repo (similar to how peep.py is in the repo).
* Add npm install to deployments (legacy and Docker).
* Update docs to mention npm installation and how to use lockdown to manage
  deps.


pmclanahan:
Blocked by webops getting NPM installed on our admin server for legacy infra deployments. Went ahead and filed this so that people could give it a look over and discuss the change.

@alexgibson I'm particularly interested in your feedback here.

pmclanahan:
[Webops bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1212909) has been filed to add npm BTW.

alexgibson:
@pmclanahan If the plan is to no longer have the production dependent node modules checked into the repo itself, then I think this makes great sense. Even minor version changes to modules are not immune to causing new bugs and regressions (I have no idea how people even use Bower like this, but thankfully at least npm has something like npm-lockdown). I guess this change does make us more dependent on npm reliability of service, but we'd be far from alone here. So, I'm all for this approach :+1: 

Out of interest what are the main benefits of this change for us when it comes to our new deployment world (I'm guessing a smaller repo helps?).

pmclanahan:
Smaller repo does help. It also makes it quicker and easier to make changes to dependencies. My guess is that we'll be doing more with Node and have more dependencies there over time, and checking them all in will get pretty unwieldy pretty quickly. Also we're already dependent on npm for test runs, so we might as well lock those down as well.

pmclanahan:
We're unblocked on this as IT has installed npm. I'll test on a demo and report back.

lockdown.json
```diff
@@ -0,0 +1,1152 @@
...
+    "1.2.1": "9bb5304d2e0b56698b2c758b08a3eaa9daa58a39"
+  },
+  "camelcase-keys": {
+    "1.0.0": "bd1a11bf9b31a1ce493493a930de1a0baf4ad7ec"
+  },
+  "caseless": {
+    "0.11.0": "715b96ea9841593cc33067923f5ec60ebda4f7d7",
+    "0.6.0": "8167c1ab8397fb5bb95f96d28e5a81c50f247ac4"
+  },
+  "casperjs": {
```
alexgibson: We just recently removed this from `package.json` (along with phantomJS), so it might be worth updating the lockdown file here, as they are no longer needed :)

pmclanahan:
I've successfully deployed demos of this to deis and demo3. I think we're about as ready as we'll ever be to do this. Final thoughts?

pmclanahan:
@alexgibson @jgmize ? ^^

jgmize:
lgtm from a deployment perspective, but I'm unclear on the tradeoffs for frontend devs. If @alexgibson is happy with this approach then let's go for it.

alexgibson:
I'm fine with this approach in general, and keeping the size of the repo down is good. If we're in a good state to merge/deploy this then sounds good to me!

PR #3414, opened 2015-10-08
Fix django tests in jenkins.

pmclanahan:
Also switches both Jenkins and Travis tests to run in-memory sqlite DB.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ cache:
 before_script:
   - flake8 bedrock lib
   - python manage.py runscript check_calendars
-  - mysql -e 'create database bedrock_test;'
   - python manage.py version
   - python manage.py migrate --noinput
   - python manage.py collectstatic --noinput -v 0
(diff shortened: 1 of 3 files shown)


pmclanahan:
@glogiotatidis this should fix the jenkins test runs. I was able to reproduce the failed tests locally and these changes fixed them when run with the `JENKINS_HOME` envvar set.

glogiotatidis:
Thanks @pmclanahan, nice work, r+! :beers:. I filed #3415 with an updated unit test run script for jenkins. 

fyi the jenkins docker pipeline does not make use of `bedrock/settings/jenkins.py` and instead uses environment variables. Since we're not using ci.m.o either we can remove this file.

pmclanahan:
I think it does make use of `jenkins.py` because that file is imported in `settings/__init__.py` if the `JENKINS_HOME` env var is set, which it is in the test runs I saw.

pmclanahan:
Ah! I see what you mean from #3415. The test run doesn't see those env vars. Good to know. We can delete that file then.

PR #3418, opened 2015-10-09
Bug 1210368: Use git-based locale data on dev/demo servers

pmclanahan:
This is unreverting #3385 and doing #3386 again but only for dev/demo.


pmclanahan:
@jgmize r?

PR #3419, opened 2015-10-09
Update l10n docs now that translations are on GitHub

pascalchevrel:
Note: I also updated a line about the shortcut for source, it's not specific to zsh, been using it in bash for years :)

### DIFF
--- a/docs/contribute.rst
+++ b/docs/contribute.rst
@@ -127,28 +127,28 @@ Server architecture
 - *URLs:* http://www-demo1.allizom.org/ , http://www-demo2.allizom.org/ and
   http://www-demo3.allizom.org/
 - *PHP SVN branch:* trunk, updated every 10 minutes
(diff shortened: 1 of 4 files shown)


docs/contribute.rst
```diff
@@ -127,28 +127,28 @@ Server architecture
 - *URLs:* http://www-demo1.allizom.org/ , http://www-demo2.allizom.org/ and
   http://www-demo3.allizom.org/
 - *PHP SVN branch:* trunk, updated every 10 minutes
-- *Bedrock locale SVN branch:* trunk, updated every 10 minutes
+- *Bedrock locales dev repo:* master, updated via a webhook on pushes
```
jgmize: Technically only the new infrastructure is using the webhook, and the current infrastructure is still using a cron job. Even with the webhook deploys are not instantaneous, so as long as that expectation is not being set I think this is fine.

PR #3420, opened 2015-10-12
[fix bug 1212033] Create tracking protection animation

alexgibson:
Usage:

1.) Make sure you have [UITour set up for local development](http://bedrock.readthedocs.org/en/latest/uitour.html#local-development).
2.) Include `js/firefox/australis/australis-uitour.js` and  `js/base/mozilla-highlight-target.js` in your page JS bundle.
3.) Include markup in the template `{% include 'firefox/includes/tracking-protection-animation.html' %}`
4.) Add less file to page bundle `media/css/firefox/tracking-protection-animation.less`
5.) If user is on up-to-date Firefox and eligible to see the second part of the animation (i.e. critters disappear when clicking the button), then add the class `firefox-up-to-date` to any parent element in order to trigger loading the relevant assets.
6.) When the user clicks the private browsing CTA, a custom event is fired which we can listen for in order to play the second stage of the animation:

    var $shield = $('#tracking-protection-animation');

    Mozilla.HighlightTarget.init('.button-flat-dark');

    $('.button-flat-dark').on('highlight-target', function() {
        $shield.addClass('blocked');
    });


alexgibson:
@malena r?

malena:
r+ :dancer:

PR #3416, opened 2015-10-09
Update l10n docker image to use git l10n repo.

glogiotatidis:
Use https://github.com/mozilla-l10n/bedrock-l10n for locales. This code already runs on https://ci.us-west.moz.works/job/bedrock_l10n_image/

- [ ] Update l10n job to use this script when merged.
- [x] Add webhook

PR #3421, opened 2015-10-12
Bug 1210368: Convert legacy infra from svn to git for locales.

pmclanahan:
This just changes the deployment and cron scripts. I'll need to manually
update the `locale` dir on the admin node for this to work, so please coordinate
the merging of this with me.

### DIFF
--- a/bin/update-scripts/prod/update-prod-locale.sh
+++ b/bin/update-scripts/prod/update-prod-locale.sh
@@ -1,15 +1,18 @@
 #!/bin/bash
 
-UPSTREAM=$(svn info https://svn.mozilla.org/projects/mozilla.com/trunk/locales | grep Revision | cut -f2 -d' ')
 cd /data/bedrock/src/www.mozilla.org-django/bedrock
 
 pushd locale > /dev/null
-RUNNING=$(svnversion ./ | tr -d '[A-Z]')
-if [ "$RUNNING" != "$UPSTREAM" ]; then
-    svn -q up
-    echo -e "finished at $(date)" > /data/bedrock/src/www.mozilla.org-django/bedrock/media/locale_finished.txt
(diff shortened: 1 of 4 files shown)


pmclanahan:
@jgmize r?

@glogiotatidis this is just for legacy infra. Do you think you could knock out the required changes for docker tomorrow? I can take a stab at it as well, but it's nearly the end of my day. Let me know if not and I can hack on it in the morning.

jgmize:
> This just changes the deployment and cron scripts. I'll need to manually
update the locale dir on the admin node for this to work, so please coordinate
the merging of this with me.

I think the updating of the locale dir could have been included in the update scripts by including something along the lines of the following before the `pushd locale`:

```sh
if [[ ! -e locale/.git ]]; then
    rm -rf locale
    git clone git clone https://github.com/mozilla-l10n/bedrock-l10n.git
fi
```

However, I have no objection to you doing this manually as it's a one-time operation, and this looks fine to merge as-is. r+

> @glogiotatidis this is just for legacy infra. Do you think you could knock out the required changes for docker tomorrow? [...]

I believe he already did this in #3416, which I will take a look at next.

pmclanahan:
> I believe he already did this in #3416, which I will take a look at next.

Yes. Just saw that. WOO! Nice work Giorgos :)

pmclanahan:
> I think the updating of the locale dir could have been included in the update scripts

I did consider that, but decided against it because, as you said, one time thing. Though now that you make it look so easy I might throw it in :)

pmclanahan:
I added the check for the `.git` dir in deployment instead of cron since the deployment will fail in the `update_info` task if the locale dir isn't a git repo.

jgmize:
r+, nice work.

PR #3422, opened 2015-10-13
Add periodic db update script for jenkins

jgmize:
I've set up an example of this approach at https://ci.us-west.moz.works/view/Bedrock/job/bedrock_periodic_db_update_dev_us_west/ -- to use this, we would set up a periodic job executing this new script for each app and cluster that needed it.

### DIFF
--- a/docker/jenkins/periodic_db_update.sh
+++ b/docker/jenkins/periodic_db_update.sh
@@ -1,0 +1,16 @@
+#!/bin/bash -ex
+# Needs DEIS_CONTROLLER URL, DEIS_USERNAME, DEIS_PASSWORD,
+# DEIS_PROFILE, and DEIS_APPLICATION environment variables.
+#
+# To set them go to Job -> Configure -> Build Environment -> Inject
+# passwords and Inject env variables
+#
+
+deis login $DEIS_CONTROLLER --username=$DEIS_USERNAME --password=$DEIS_PASSWORD
+deis run -a $DEIS_APPLICATION -- '\
+./manage.py rnasync; \
+./manage.py cron update_ical_feeds; \
+./manage.py update_product_details; \
+./manage.py update_externalfiles; \
+./manage.py update_security_advisories; \
+./manage.py cron update_tweets;'


jgmize:
@glogiotatidis r?

docker/jenkins/periodic_db_update.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+# Needs DEIS_CONTROLLER URL, DEIS_USERNAME, DEIS_PASSWORD,
+# DEIS_PROFILE, and DEIS_APPLICATION environment variables.
+#
+# To set them go to Job -> Configure -> Build Environment -> Inject
+# passwords and Inject env variables
+#
+
```
glogiotatidis: Add `set -xe`  to
 - (-x) Print commands just before execution - with all expansions and substitutions done, and words marked - useful for debugging.
 - (-e) and to exit immediately if a command fails.

docker/jenkins/periodic_db_update.sh
```diff
@@ -0,0 +1,16 @@
...
+deis whoami || deis login $DEIS_CONTROLLER --username=$DEIS_USERNAME --password=$DEIS_PASSWORD
```
glogiotatidis: I'd avoid this micro-optimization to limit the possibility of errors due to false or outdated logins and instead I'd always `deis login`

glogiotatidis:
Both this and the jenkins job look good. r+ Nice job @jgmize :sunny:

docker/jenkins/periodic_db_update.sh
jgmize: Good catch. I had included this in the shebang (`#!/bin/bash -xe`) originally, but then pasted over it with an adaptation of your excellent commentary from another script. Will update before merge.

docker/jenkins/periodic_db_update.sh
jgmize: Fair enough, will remove the optimization.

PR #3424, opened 2015-10-13
Fix Bug 1210426, adds press inquiry page

schalkneethling:
@jpetto @pmclanahan r?

### DIFF
--- a/bedrock/press/forms.py
+++ b/bedrock/press/forms.py
@@ -15,6 +15,70 @@ from bedrock.mozorg.forms import (DateInput, EmailInput, HoneyPotWidget,
 SPEAKER_REQUEST_FILE_SIZE_LIMIT = 5242880  # 5MB
 
 
+class PressInquiryForm(forms.Form):
+    jobtitle = forms.CharField(
+        required=True,
+        widget=forms.TextInput(
+            attrs={
+                'class': 'required',
+                'required': 'required',
+                'aria-required': 'true'
+            }
+        )
+    )
+    name = forms.CharField(
+        required=True,
+        widget=forms.TextInput(
+            attrs={
+                'class': 'required',
+                'required': 'required',
+                'aria-required': 'true'
+            }
+        )
+    )
+    media_org = forms.CharField(
+        required=True,
+        widget=forms.TextInput(
+            attrs={
+                'class': 'required',
+                'required': 'required',
+                'aria-required': 'true'
+            }
+        )
+    )
+    inquiry = forms.CharField(
+        required=True,
+        widget=forms.Textarea(
+            attrs={
+                'class': 'required',
+                'required': 'required',
+                'aria-required': 'true'
+            }
+        )
+    )
+    deadline = forms.CharField(
+        required=True,
+        widget=DateInput(
+            attrs={
+                'class': 'required',
+                'required': 'required',
+                'aria-required': 'true'
+            }
+        )
+    )
+
+    # honeypot
+    office_fax = forms.CharField(widget=HoneyPotWidget, required=False)
+
+    def clean_office_fax(self):
+        cleaned_data = super(PressInquiryForm, self).clean()
+        honeypot = cleaned_data.pop('office_fax', None)
+
+        if honeypot:
+            raise forms.ValidationError(
+                _('Your submission could not be processed'))
+
+
 class SpeakerRequestForm(forms.Form):
     # event fields
     sr_event_name = forms.CharField(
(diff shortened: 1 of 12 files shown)


bedrock/press/templates/press/press-inquiry.html
```diff
@@ -0,0 +1,103 @@
...
+
+      <form name="mozilla-contact-us" id="press_inquiry" class="press-inquiry" action="{{ url('press.press-inquiry') }}" method="post">
+
+        {{ csrf() }}
+
+        <fieldset>
+          <div class="field required">
+            <label for="id_jobtitle">{{ _('Job Title') }}</label>
+            {% if form.jobtitle.errors %}
+              <p class="error-msg"> - {{_('Please enter your job title.')}}</p>
```
jpetto: Looks like an errant ` - ` in this tag.

bedrock/press/templates/press/includes/resources.html
```diff
@@ -0,0 +1,16 @@
+<aside class="sidebar">
+  {#
+    This list should likely be factored out to an include when all press pages
+    have been ported to bedrock.
+  #}
```
jpetto: Can probably drop this comment now, eh? :grin:

bedrock/press/forms.py
```diff
@@ -15,6 +15,70 @@
...
+            attrs={
+                'class': 'required',
+                'required': 'required',
+                'aria-required': 'true'
+            }
+        )
+    )
+    deadline = forms.CharField(
+        required=True,
+        widget=forms.DateInput(
```
jpetto: Could get a `type="date"` rendered if you use our custom `DateInput` widget here. (Just `widget=DateInput` instead of `widget=forms.DateInput`.)

bedrock/press/templates/press/press-inquiry.html
```diff
@@ -0,0 +1,103 @@
...
+    <div class="main-column">
+    {% if not form_success %}
+
+      {% if form.errors %}
+        <div class="tooltip">
+          {{ _('An error has occurred with your submission. Please review your information below and try again.') }}
+        </div>
+      {% endif %}
+
+      <form name="mozilla-contact-us" id="press_inquiry" class="press-inquiry" action="{{ url('press.press-inquiry') }}" method="post">
```
jpetto: Just happened to notice the `<body>` id and this form's id are identical.

media/css/press/press.less
```diff
@@ -88,6 +88,16 @@ fieldset {
     }
 }
 
+.press-inquiry {
+    .error-msg {
+        background-color: #aa1401;
```
jpetto: Small nit - unnecessary nesting.

media/css/press/press-ie.less
```diff
@@ -0,0 +1,4 @@
+.tooltip {
+    margin-bottom: 20px;
+    max-width: 75%;
+}
```
jpetto: So the tooltip seems to overlap the following field label a bit in modern browsers as well. This is in dev edition:

screen shot 2015-10-13 at 4 13 40 pm

Looks like instead of making an IE specific CSS file & bundle, we could just move these rules into the main `press.less`? Or add a bit of top margin to `form.press-inquiry` (since that selector already exists)?

jpetto:
Reminder to add a link to this page from new `/contact/` page (coming in #3412). May need to add the link in that PR, depending on which lands first.

bedrock/press/templates/press/press-inquiry.html
schalkneethling: Yes, I was going to have it inline but then decided against it. So this is no longer relevant.

bedrock/press/templates/press/includes/resources.html
schalkneethling: Ha! indeed

bedrock/press/templates/press/press-inquiry.html
schalkneethling: Seeing I do not actually use it for anything, I am just removing it.  Nice catch

media/css/press/press-ie.less
schalkneethling: I completely forgot about `novalidate` attribute to have these browsers skip validation and thus, did not test it :-/ Me :disappointed:

jpetto:
Looking good to me! I feel confident in the Python bits, as they're well tested and based on the existing speaker request form. r+ :candy: 

Only thing left is to squash, and then add a link to this page from the new `/contact/` page (from #3412), which could be done here or there, depending on which we want to land first.

pmclanahan:
> I feel confident in the Python bits, as they're well tested and based on the existing speaker request form. r+ :candy:

Agreed with Jon. Nice work Schalk!

PR #3429, opened 2015-10-14
Fix Bug 1214022, additions to the mozilla trademark list

schalkneethling:
@alexgibson r?


schalkneethling:
Good for review, just waiting for confirmation on the bug. Thanks!

alexgibson:
The new images are twice the size of the others, I would suggest sizing them to be consistent?

trademarks

bedrock/foundation/templates/foundation/trademarks/list.html
```diff
@@ -19,9 +19,15 @@ <h1 class="page-title">{{ _('List of Mozilla Trademarks') }}</h1>
       <th>{{ _('Firefox logo') }}&reg;</th>
     </tr>
     <tr>
+      <td>
+        {{ high_res_img('trademarks/hello-logo-wordmark.png', { 'alt': 'Firefox Hello logo' }) }}
```
alexgibson: The `alt` text should be wrapped for l10n here and below.

schalkneethling: Yes, I was wondering whether this page is localized. I am guessing it is? If so, this probably needs an l10n label as a heads up to @flodolo

alexgibson: I don't believe it is localized (but @flodolo can confirm). Either way, existing strings are already wrapped on this page so we should still wrap new ones.

schalkneethling:
@alexgibson Yes, the size of the logos and whether or not it should be just the logo or, logo and wordmark, is what I am waiting confirmation on. With that said, I agree with your suggestion and will make the update. Thanks!

alexgibson:
@schalkneethling - if we are waiting on possible changes based on feedback in the bug, is it best to wait on review here if the code is likely to be changing again?

bedrock/foundation/templates/foundation/trademarks/list.html
flodolo: No, not localized (but good thing in general to have strings wrapped).

schalkneethling:
@alexgibson nope, the code won't change. The only possible change are the images and possibly their size.

alexgibson:
Please let me know when changes are made and this is good to go, and I will check again :+1:

alexgibson:
@schalkneethling - any update on this one yet, is it good for review / merge-able?

schalkneethling:
@alexgibson Looks like it might be, just waiting on final confirmation from Jen.

schalkneethling:
@alexgibson Ok, this is good for review and merge. Thank you

bedrock/foundation/templates/foundation/trademarks/list.html
```diff
@@ -15,127 +15,113 @@ <h1 class="page-title">{{ _('List of Mozilla Trademarks') }}</h1>
...
-      <th>{{ _('Camino') }}&reg;</th>
+      <td><img width="100" height="100" alt="{{ _('SeaMonkey logo') }}®" src="{{ static('img/trademarks/seamonkey-logo-r.png') }}" /><span>&reg;</span></td>
+      <th>{{ _('SeaMonkey logo') }}</th>
     </tr>
     <tr>
-      <td><img width="100" height="100" alt="{{ _('Mozilla Communities') }}™" src="{{ static('img/trademarks/community-wordmark.png') }}" /></td>
-
-      <th>{{ _('Mozilla Communities') }}&trade;</th>
+      <td><img width="100" height="100" alt="{{ _('Foxkeh logo') }}®" src="{{ static('img/trademarks/foxkeh-logo-r.png') }}" /><span>&reg;</span></td>
+      <th>{% trans %}Foxkeh logo<br /><small>This trademark is licensed under the Creative Commons Attribution-Non-Commercial-NoDerivs 2.1 Japan.</small>{% endtrans %}</th>
```
alexgibson: Might this be better as two separate strings rather than using a trans block?

alexgibson:
One minor nit, but otherwise looks good r+wc :turtle:

bedrock/foundation/templates/foundation/trademarks/list.html
schalkneethling: That would be better. I did not change it and it is as it was before I believe but, I like your suggestion and, this is not currently translated so, will not be breaking l10n

PR #3431, opened 2015-10-14
Add email settings for sending from AWS SES.

pmclanahan:
Set some env vars and it'll do the right thing.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -478,8 +478,11 @@ def lazy_email_backend():
 
 EMAIL_BACKEND = config('EMAIL_BACKEND', default=lazy(lazy_email_backend, str)())
 EMAIL_HOST = config('EMAIL_HOST', default='localhost')
+EMAIL_PORT = config('EMAIL_PORT', default=25, cast=int)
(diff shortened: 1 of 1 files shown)


pmclanahan:
In case we can use this, here it is. Shouldn't cause problems to merge. Have to set `EMAIL_BACKEND=django_ses.SESBackend` env var to enable, as well as the AWS settings I just added. Then it should "Just Work(tm)" as long as we have enabled the SES service in our account.

pmclanahan:
I've changed this from using the `django-ses` package to just using SMTP. We just needed more settings to be configurable via env vars.

bedrock/settings/base.py
```diff
@@ -478,8 +478,11 @@ def lazy_email_backend():
 
 EMAIL_BACKEND = config('EMAIL_BACKEND', default=lazy(lazy_email_backend, str)())
 EMAIL_HOST = config('EMAIL_HOST', default='localhost')
+EMAIL_PORT = config('EMAIL_PORT', default=25, cast=int)
+EMAIL_USE_TLS = config('EMAIL_USE_TLS', default=False, cast=bool)
```
jgmize: Why is this default False?

PR #3432, opened 2015-10-14
Switch to using DB prod-details backend when DEBUG is False

pmclanahan:


pmclanahan:
@jgmize r?

pmclanahan:
I guess we don't need this as much now that all of the envs have the `PROD` variable set, but we could do this so we don't grow reliance on that.

jgmize:
> I guess we don't need this as much now that all of the envs have the PROD variable set, but we could do this so we don't grow reliance on that.

I didn't use the PROD setting; I set `PROD_DETAILS_STORAGE = 'product_details.storage.PDDatabaseStorage'` in  the stage and prod envs, and didn't touch the dev or demos. I'm not sure that tying PROD_DETAILS_STORAGE to DEBUG won't cause confusion in someone's local dev env down the line: for example, imagine a month or two after this merges, if someone decides to test something with `DEBUG = False` locally, but normally runs with `DEBUG = True`, they will encounter an error because their product details DB is empty, but it won't be immediately obvious to them as to why. I'm wondering if we should just switch to DB by default everywhere, and just send out an email letting people know they will need to run `./manage.py migrate` in their local env after pulling the change.

pmclanahan:
> I'm wondering if we should just switch to DB by default everywhere, and just send out an email letting people know they will need to run ./manage.py migrate and ./manage.py update_product_details (or bin/sync_all, which will take care of both and a few other things as well) in their local env after pulling the change.

That's fine I think. Really they only ***need*** to run `migrate` since we have the fancy data migration. But sure. In reality I'm happy the way it is, I was just doing this to hit a middle ground between annoying us and annoying other devs running the site. Since we've already set the settings on the running instances, we can probably just close this and forget about it until there's a compelling reason to rock the boat again.

PR #3433, opened 2015-10-14
Fix Bug 1204579 - Consolidate Thunderbird rewrite rules and add some redirects

kyoshino:
I forgot to send this earlier… The most of the work mentioned in the bug has actually been done in #3360. This will remove the rest of the legacy Thunderbird redirect rules so the PHP server can be terminated soon.


jgmize:
thanks @kyoshino!

PR #3436, opened 2015-10-15
Fix Bug 1196255 - tabzilla invisible on mobile view of /privacy/tips/

kyoshino:
This works on both Firefox 41 and 42 (and later).


kyoshino:
Ah, no, this breaks the layout on Firefox 41. Will figure out a better way.

alexgibson:
Thanks @kyoshino r+ :banana:

PR #3438, opened 2015-10-15
Ping NewRelic on deploy.

glogiotatidis:
More stats, more logs, more fun!

(I already updated the env variables in jenkins)


jgmize:
Thanks @glogiotatidis!

PR #3439, opened 2015-10-15
Add functional tests for /firefox/dnt/

alexgibson:
* Adds tests for `/firefox/dnt/` page.
* I also added a new class in the `page/base.py` for the Firefox Family navigation (which appears on multiple pages) for easier code reuse when it might be needed elsewhere. Not totally sure if this is the right approach, but good to discuss.
* Removes redundant casper tests.


alexgibson:
@davehunt r?

davehunt:
Hey @alexgibson this looks great! I do have a few recommendations which I'll add individual comments for.

tests/functional/test_firefox_dnt.py
```diff
@@ -0,0 +1,25 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+from selenium.common.exceptions import TimeoutException
```
davehunt: I recommend adhering to PEP8 for Python code. For imports standard libraries, external libraries, and internal libraries should be separated by a blank line. See more about imports here: http://legacy.python.org/dev/peps/pep-0008/#imports

tests/functional/test_firefox_dnt.py
```diff
@@ -0,0 +1,25 @@
...
+
+from ..pages.firefox_dnt import FirefoxDNTPage
+
+@pytest.mark.nondestructive
```
davehunt: Another PEP8 tip: top level functions should be surrounded with two blank lines. See http://legacy.python.org/dev/peps/pep-0008/#blank-lines for details. I would recommend using [flake8](https://pypi.python.org/pypi/flake8) as a precommit hook to check your code for compliance. We could also add a check in Travis CI so we avoid landing code that does not adhere.

tests/pages/base.py
```diff
@@ -41,3 +46,18 @@ def languages(self):
...
+
+        def click_tertiary_menu(self):
+            self.selenium.find_element(*self._tertiary_nav_button).click()
+            Wait(self.selenium, self.timeout).until(
+                expected.visibility_of_element_located(self._tertiary_nav_menu))
+
+        @property
+        def tertiary_menu_visible(self):
+            el = self.selenium.find_element(*self._tertiary_nav_menu)
+            return el.is_displayed()
```
davehunt: I think I would have a function directly under the base page named `open_x_menu` you've used both 'family' and 'tertiary', so I'd pick one and use it consistently. This function could return your `PageRegion` object, with functions and properties related to the menu itself. Rather than returning whether the menu is visible, a better _functional_ test might click one of the menu items and check that you get the expected response. Otherwise, with the mentioned suggestions your test might look something like this:

```python
menu = page.open_tertiary_menu()
assert menu.is_displayed
```

davehunt: Also, where possible I would recommend providing a root element for a `PageRegion`. If you avoid having the button that opens the menu within this region as suggested above, you could use `_root_locator = (By.ID, 'fxfamilynav-tertiarynav')`. This means that any locators within this region are looked up from the scope of this element rather than the entire document. While the performance advantage is negligible, it can help when using locators that are not unique.

tests/pages/firefox_dnt.py
```diff
@@ -0,0 +1,32 @@
...
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as expected
+from selenium.webdriver.support.select import Select
+from selenium.webdriver.support.ui import WebDriverWait as Wait
+
+from .base import BasePage
+
+
+class FirefoxDNTPage(BasePage):
```
davehunt: If you're going to have a bunch of Firefox pages, you could put this in `pages/firefox/do_not_track.py`. Also, I generally prefer not to abbreviate, so I'd call this `DoNotTrackPage` under the `pages.firefox` module. In my mind this makes it more accessible to anyone unfamiliar with the DNT abbreviation.

tests/pages/firefox_dnt.py
```diff
@@ -0,0 +1,32 @@
...
+
+    _url = '{base_url}/{locale}/firefox/dnt'
+
+    _dnt_status = (By.CSS_SELECTOR, '#dnt-status > h5')
+    _faq_1_header = (By.ID, 'faq-1-header')
+    _faq_1_panel = (By.ID, 'faq-1-panel')
+
+    def dnt_status_is_visible(self):
+        el = self.selenium.find_element(*self._dnt_status)
+        return el.is_displayed()
```
davehunt: I'd rename this to `is_do_not_track_status_visible`. You can also add the `@property` decorator so that you can call this without needing the `()`.

davehunt: Also, if the line is short enough, you may not need to define an `el` variable. I've done this mainly to keep line lengths down.

tests/pages/firefox_dnt.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def click_faq_1_header(self):
+        self.selenium.find_element(*self._faq_1_header).click()
+        Wait(self.selenium, self.timeout).until(
+            expected.visibility_of_element_located(self._faq_1_panel))
+
+    def faq_1_panel_is_visible(self):
+        el = self.selenium.find_element(*self._faq_1_panel)
+        return el.is_displayed()
```
davehunt: These are fine, though I'd personally name them `show_answer_to_first_question` and `is_answer_to_first_question_displayed`. That said, I'm going to use this as an opportunity to share another practice we've used when multiple items are on a page.

davehunt: If you create a `PageRegion` for frequently asked questions, such as:

```python
class FrequentlyAskedQuestion(PageRegion):

    _question_locator = (By.TAG_NAME, 'h3')
    _answer_locator = (By.TAG_NAME, 'div')

    def show_answer(self):
        assert not self.is_answer_displayed, 'Answer is already displayed'
        self.root.find_element(*self._question_locator).click()
        Wait(self.selenium, self.timeout).until(
            expected.visibility_of_element_located(self._answer_locator))

    @property
    def is_answer_displayed(self):
        return self.root.find_element(*self._answer_locator).is_displayed()
```

Then you could create and return this page region for each question on the page. In the main page object you could have something like:

```python
_faqs_locator = (By.CSS_SELECTOR, '#faq > section')

@property
def frequently_asked_questions(self):
    return [FrequentlyAskedQuestion(self.selenium, root=el) for el in \
               self.selenium.find_elements(*self.faqs_locator)]
```

Then in the test you could check the first item like so:

```python
question = page.frequently_asked_questions[0]
question.show_answer()
assert question.is_answer_displayed
```

Or you could easily check them all:

```python
for question in page.frequently_asked_questions:
    question.show_answer()
    assert question.is_answer_displayed
```

I'd also suggest having a `hide_question` to test that functionality. Wherever you assert something is visible, I recommend using the same property to check the opposite - if at all possible. This reduced the chance of false positive.

Note that my code samples are untested. :smile:

tests/pages/firefox_dnt.py
```diff
@@ -0,0 +1,32 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as expected
+from selenium.webdriver.support.select import Select
```
davehunt: I think this is an unused import. If you use flake8 it should report these too.

tests/pages/base.py
davehunt: If we create a `pages.firefox` module, and this menu only applies to those pages, we could move it into `pages/firefox/base.py`, make that extend `pages/base.py` and make the appropriate pages extend the new `FirefoxBasePage`. That said, I don't feel strongly about this - it's a better organisational structure if appropriate but adds an additional layer.

alexgibson:
@davehunt - thanks for such a great review! I'll work through these and come back to you if I have any questions. As for the pep8 issues, we already use a hook in Travis for this, but it looks like it isn't checking these files. Will have to try and work out why.

tests/pages/firefox_dnt.py
alexgibson: I think this makes sense. I did consider this initially, but wasn't quite sure if this would be the right way to do it. I'll make it so!

tests/pages/firefox_dnt.py
alexgibson: I think using this pattern sounds certainly more flexible, good suggestions :+1:  I think it is perhaps overkill to repeat clicking on every item, as it is essentially testing the same code multiple times, but I think opening / closing a section should be a minimum requirement for a base test.

alexgibson:
@davehunt - think I've addresses most of your feedback hopefully, let me know if I misunderstood anything?

My only slight concern here is with the `test_open_and_close_faq_panel` test. As discussed, for this to pass the window needs to be large enough to open/close the panel without causing Selenium to scroll, else the sticky navigation header causes a miss-click when closing the menu. The test passes fine for me now we're maximizing the browser window for each test, but it still feels a little brittle. Is it worth writing a custom scrolling function for this use case?

tests/functional/conftest.py
```diff
@@ -0,0 +1,11 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+
+@pytest.fixture
+def selenium(selenium):
+    selenium.maximize_window()
```
davehunt: I think we should specify the window size so that tests are consistent regardless of the resolution of the machine running the browser. See [set_window_size](http://seleniumhq.github.io/selenium/docs/api/py/webdriver_remote/selenium.webdriver.remote.webdriver.html#selenium.webdriver.remote.webdriver.WebDriver.set_window_size). I note that the CasperJS tests defaulted to 1000x600. We might want to pick something a little more standard.

tests/pages/firefox/do_not_track.py
```diff
@@ -0,0 +1,49 @@
...
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as expected
+from selenium.webdriver.support.ui import WebDriverWait as Wait
+
+from .base import FirefoxBasePage
+from ..page import PageRegion
+
+
+class FrequentlyAskedQuestion(PageRegion):
```
davehunt: Nit: I would put this as a subclass beneath `DoNotTrackPage`.

tests/functional/firefox/do_not_track.py
```diff
@@ -0,0 +1,30 @@
...
+import pytest
+
+from ...pages.firefox.do_not_track import DoNotTrackPage
+
+
+@pytest.mark.nondestructive
+def test_family_navigation(base_url, selenium):
+    page = DoNotTrackPage(base_url, selenium).open()
+    page.family_navigation.open_menu()
+    assert page.family_navigation.is_menu_displayed
```
davehunt: Nit: I would have an `open_family_navigation_menu` on `FirefoxBasePage` that returned a `FamilyNavigationMenu` object. This menu would then only have an `is_displayed` property. In my opinion this would make for a slightly more readable test:

```python
page = DoNotTrackPage(base_url, selenium).open()
menu = page.open_family_navigation_menu()
assert menu.is_displayed
```

davehunt:
This looks great. All of my comments are just nits, so please feel free to ignore them. For the scrolling, I think it may be a problem at some point, and we can avoid it by scrolling particular elements into the bottom of the viewport instead of the top. I think having a sensible default window size will mean we don't have to worry about this until it comes up again though. Here's an example of [scrolling to avoid floating banners at the top of the page](https://github.com/mozilla/sumo-tests/blob/6efc9d70258d5928de07243b4d45ae071ff79d89/pages/desktop/questions_page.py#L48-L49).

alexgibson:
@davehunt updated, thanks!

davehunt:
r+

PR #3440, opened 2015-10-15
Modify clock script to just run the jobs instead of the schedule.

pmclanahan:
Basically the same as #3295, but just runs all the jobs with the enhanced logging insetad of scheduling them. This will make it very easy to enable the scheduler process when deis is ready.

### DIFF
--- a/scripts/clock.py
+++ b/scripts/clock.py
@@ -6,7 +6,7 @@ from apscheduler.schedulers.blocking import BlockingScheduler
 
 from bedrock.events.cron import cleanup_ical_events, update_ical_feeds
 from bedrock.mozorg.cron import update_tweets
-from scripts import update_firefox_os_feeds, update_tableau_data
+from scripts import update_firefox_os_feeds
 
 
 schedule = BlockingScheduler()
@@ -37,16 +37,6 @@ class scheduled_job(object):
 
     def log(self, message):
         print('Clock job {0}: {1}'.format(self.name, message))
-
-
-@scheduled_job('interval', seconds=5)
-def job_should_pass():
-    print('DO ALL THE THINGS!!!')
-
-
-@scheduled_job('interval', seconds=5)
-def job_raise_exception():
-    raise ValueError("Stuff's broke yo!")
 
 
 @scheduled_job('interval', minutes=30)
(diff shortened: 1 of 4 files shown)


pmclanahan:
@jgmize r?

scripts/clock.py
```diff
@@ -0,0 +1,91 @@
...
+
+
+@scheduled_job('interval', hours=1)
+def job_ical_feeds():
+    update_ical_feeds()
+    cleanup_ical_events()
+
+
+@scheduled_job('interval', hours=1)
+def job_update_firefox_os_feeds():
```
pmclanahan: Note: `update_firefox_os_feeds` and `update_tableau_data` were left out of the previous version. Not sure we even have access to the tableau DB from AWS (likely we do not).

pmclanahan:
Are we not happy with this approach? I know @bensternthal wanted to get this in and get some alerts setup to watch for these log messages. Are you not happy about adding the dependency before we get the ability to run a separate process @jgmize? We could always create a separate app that would run the clock proc for real and give it the same settings for DB and whatnot, it'd just be a pain.

jgmize:
> Are we not happy with this approach? I know @bensternthal wanted to get this in and get some alerts setup to watch for these log messages. 

Sorry, @pmclanahan, I'm not sure this will accomplish that goal-- I believe the outputs from the print statements will return to the deis client and won't show up in papertrail.

> Are you not happy about adding the dependency before we get the ability to run a separate process @jgmize? 

Now that you mention it, I'm not that excited about the fact that this PR adds a fairly large dependency and then ends up not actually using any of its functionality. However, based on my reading of https://github.com/deis/deis/pull/4606 the issue is entirely on the deis client side, so we could try deploying a bedrock app with a procfile using either the older python version of the deis client, or if we were feeling adventurous we could build a deis client based on that PR and try using it.

pmclanahan:
ADVENTURE TIME!

pmclanahan:
@jgmize looks like said deis PR was merged. Proceed how you will.

pmclanahan:
@jgmize Just following up on some PRs. Have you tried the new deis client that's supposed to fix the Procfile issue? Should I just close this one?

jgmize:
@pmclanahan I haven't had a chance to build a deis client binary based on master to test it, but that fix should be in the 1.12 version of the client, which should be released a week from today according to http://docs.deis.io/en/latest/roadmap/schedule/#minor-releases

pmclanahan:
Closing in favor of #3960

PR #3441, opened 2015-10-15
Fix Bug 1215229, add email field to inquiry form

schalkneethling:
@jpetto r?


jpetto:
Looks good! :ice_cream: r+

PR #3443, opened 2015-10-16
Bug 1215136 - SmartOn Tracking

craigcook:
* Adds a form-title wrapper and title and subtitle params to the global newsletter macro.
* ~~Adds a "micro" version of the social sharing widget.~~ Didn't end up using this.


bedrock/teach/templates/teach/smarton/tracking.html
```diff
@@ -0,0 +1,376 @@
...
+  </section>
+
+  <section id="know" class="section section-know">
+    <div class="content">
+
+      <h2 class="section-title">{{ _('What is tracking?') }}</h2>
+
+      <p class="section-intro">
+        {{ _('Every time we go online, we leave behind traces of our activity.') }}
+        {{ _('They’re often called our “digital footprints” and for good reason, because tracking is sort of like re-tracing our steps.') }}
```
flodolo: Please group the first 2 sentences into one string
```
{{ _('Every time we go online, we leave behind traces of our activity. They’re often called our “digital footprints” and for good reason, because tracking is sort of like re-tracing our steps.') }}
```

bedrock/teach/templates/teach/smarton/tracking.html
```diff
@@ -0,0 +1,376 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "teach/smarton/base.html" %}
+
+{% block page_title %}{{ _('Get smart on tracking') }}{% endblock %}
+{% block page_desc %}{{ _('What you do online is your business, and you can keep it that way. Get smart on tracking.') }}{% endblock %}
```
craigcook: String added 20 October: "What you do online is your business, and you can keep it that way. Get smart on tracking."

jpetto:
@craigcook - Just a heads up that some additional GTM requests have been made in [bug 1217883](https://bugzilla.mozilla.org/show_bug.cgi?id=1217883).

alexgibson:
Do we need to remember to update the newsletter form CSS in [oldIE.less](https://github.com/mozilla/bedrock/blob/master/media/css/sandstone/oldIE.less#L536) too?

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+        text-align: center;
+
+        a {
+            .trailing-arrow();
+            display: inline-block;
+            padding: .25em 20px;
+            border: 2px solid rgba(255, 255, 255, .5);
+            border-radius: 5px;
+            margin-top: .15em;
+            transition: all .2s ease;
```
alexgibson: Use `.transition()` mixin?

alexgibson: Also is using `all` a bit heavy handed?

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+            }
+
+            &:before {
+                content: '« ';
+                text-align: center;
+                width: 15px;
+                position: absolute;
+                left: 4px;
+                opacity: 0;
+                transition: all .2s ease;
```
alexgibson: Same as above, use `.transition()` mixin and `all` sparingly.

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+            padding-bottom: 20px;
+        }
+    }
+
+    a {
+        .border-box();
+        width: 100%;
+        display: block;
+        padding: 10px 10px 24px;
+        transition: border-width .1s ease, padding-bottom .1s ease;
```
alexgibson: Use `.transition()` mixin

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+        a:active {
+            text-decoration: none;
+        }
+    }
+
+    .series-intro {
+        .span(9);
+        .font-size(18px);
+        margin: 0 0 1em;
+        color: rgba(255, 255, 255, .8);
```
alexgibson: Does this need a hex fallback for `color`?

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+        p {
+            margin: 0;
+        }
+    }
+}
+
+// Pull quotes
+.pull-quote {
+    .font-size(18px);
+    .open-sans-light();
```
alexgibson: Nit: no need for parenthesis

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+    font-style: italic;
+    margin-bottom: .5em;
+    position: relative;
+}
+
+.attribute {
+    color: rgba(0, 0, 0, .5);
```
alexgibson: Hex fallback for `color`?

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+        vertical-align: middle;
+        padding: 0 10px;
+    }
+
+    .title {
+        width: 33%;
+    }
+
+    .tagline {
+        .open-sans-light();
```
alexgibson: Nit: no need for parenthesis

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+.columns-2 .col { width: 50%; }
+.columns-3 .col { width: 33.3%; }
+.columns-4 .col { width: 25%; }
+.columns-5 .col { width: 20%; }
+
+
+
+@media screen and (max-height: 650px) {
+
+    // .nav-steps {
```
alexgibson: Is this needed? If not safe to remove?

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+
+}
+
+
+/* -------------------------------------------------------------------------- */
+// Specific styles for the Tracking page
+#smarton-tracking {
+    .section-ask {
+        background: url(/media/img/teach/smarton/tracking/tracking-hero-bg.svg) center -100px no-repeat;
+        background-size: 100% auto;
```
alexgibson: Use `.background-size()` mixin.

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+
+        @media only screen and (max-width: @breakTablet) {
+            background-size: 150% auto;
```
alexgibson: Use `.background-size()` mixin.

media/css/teach/smarton.less
```diff
@@ -0,0 +1,1431 @@
...
+
+    .topic-surveillance {
+        background: @theme-surveillance;
+    }
+}
+
+.series-description {
+    .border-box;
+    padding: 0 ((@gridColumnWidth * 2) + (@gridGutterWidth));
+    margin: 0 auto @baseLine*2;
```
alexgibson: Nit space wither side of `*`

alexgibson:
Worth putting the images here through https://tinypng.com/ ?

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
...
+}
+
+;(function($) {
+    'use strict';
+
+    var $window = $(window);
+    var $document = $(document);
+    var $body = $('body');
+    var $header = $('.page-header');
+    var $nav = $('.nav-steps');
```
alexgibson: Looks like this variable isn't used.

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    // Check for an adjusted scrollbar every 100ms.
+    setInterval(adjustScrollbar, 100);
+
+    // Scroll smoothly to the linked section
+    $document.on('click', '.nav-steps a[href^="#"]', function(e) {
+        e.preventDefault();
+        // Extract the target element's ID from the link's href.
+        var elem = $(this).attr('href').replace( /.*?(#.*)/g, "$1" );
```
alexgibson: Nit: mixed double and single quotes

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
...
+
+    // Trigger the chart when it scrolls into view
+    $('.sidebar-chart').each(function(){
+        var stage = $(this).children('.chart-stage');
+        var circleId = stage.attr('id');
+        var circleVal = stage.data('chart-value');
+        var waypoint = new Waypoint({
+            element: stage,
+            handler: function(direction) {
+                if (direction == 'down') {
```
alexgibson: Nit use `===`

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
...
+        });
+    });
+
+    // Track user scrolling through each section down the page
+    $('.main .section').each(function() {
+        var sectionId = $(this).attr('id');
+        var waypoint = new Waypoint({
+            element: $(this),
+            handler: function(direction) {
+                if (direction == 'down') {
```
alexgibson: Nit use `===`

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// create namespace
+if (typeof window.Mozilla === 'undefined') {
```
alexgibson: Doesn't look like this is needed for this file?

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
...
+                        $firstSection.removeAttr('style');
+                        $header.find('.nav-steps li').removeClass();
+                    }
+                }
+            }
+        }
+    }
+
+    // Check for an adjusted scrollbar every 100ms.
+    setInterval(adjustScrollbar, 100);
```
alexgibson: Calling this function on `setInterval` isn't too great performance wise. Is there a reason why you've created a custom function for the sticky header, instead of using `Waypoint.Sticky` (which already seems included in the page bundle)?

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
...
+
+    // Scroll smoothly to the linked section
+    $document.on('click', '.nav-steps a[href^="#"]', function(e) {
+        e.preventDefault();
+        // Extract the target element's ID from the link's href.
+        var elem = $(this).attr('href').replace( /.*?(#.*)/g, "$1" );
+        $('html, body').animate({
```
alexgibson: We have a fancy new [native scrolling polyfill](https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-smoothscroll.js) which falls back to jQuery `animate` in older browsers, which would be nice here if suitable?

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
...
+    });
+
+    // Check window width
+    if (queryWidth.matches) {
+        wideMode = true;
+    } else {
+        wideMode = false;
+    }
+
+    queryWidth.addListener(function(mq) {
```
alexgibson: This is throwing an error in IE9 - need to include the polyfill in a conditional comment [like here](https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templates/mozorg/about/manifesto.html#L337-L339)

alexgibson:
IE has some nav overflow that could do with fixing:

ie-nav

the page layout is also pretty broken in IE8 (although text is still readbale)

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
...
+
+    // Play videos in a modal and track the clicks
+    $('a.video-open').attr('role', 'button').on('click', function(e) {
+        e.preventDefault();
+
+        var $this = $(this);
+        var embedSrc = $this.data('embed-src');
+        var embedWrap = $('#' + $this.data('element-id'));
+
+        Mozilla.Modal.createModal(this, embedWrap, {
```
alexgibson: I know this has been discussed as an issue already, but I still think opening this website in an iframe is less than ideal even with a modal being used. If I close the modal and try to bail on the 100mb download, it continues to load in the background. Re-opening the modal also starts the download over again from the start :(

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
...
+            scrollTop: $(elem).offset().top - 83
+        }, 1000, function() {
+            $(elem).attr('tabindex','100').focus().removeAttr('tabindex');
+        });
+    });
+
+    var $footer = $('#footer-cta');
+    var footHeight = $footer.height();
+    var tallMode = false;
+    var queryHeight = matchMedia('(min-height: 650px)');
```
alexgibson: This throws an error is all browsers that don't support `matchMedia`. Ideally we should move all matchMedia logic inside it's own function, and only call it after using a feature detect.

media/js/teach/smarton.js
```diff
@@ -0,0 +1,303 @@
...
+        });
+    }
+
+    function adjustScrollbar() {
+        if (didScroll) {
+            didScroll = false;
+            var scrollTop = $window.scrollTop();
+            if ($header.length > 0) {
+                if( scrollTop >= headTop.top ) {
+                    if(!fixed) {
```
alexgibson: Nit: conditionals in this file could do with some tidy up in places, e.g. spaces after `if`, spaces before opening brackets `{` etc. Obviously this is less important than other things, so feel free to skip it if time is crunched.

alexgibson:
Aspect ratio on the thumbnail is a little off at smaller viewport sizes

thumbnail

alexgibson:
Could make a bit more use of available space here at mobile landscape size?

mobile

alexgibson:
Close button on sticky footer overlaps a bit on resize

screen shot 2015-10-30 at 11 53 16 am

alexgibson:
Seems to be a bit of funkyness on the sticky email footer once it's been expanded (i.e. the form details are showing). Scrolling up & down the page makes it jump around a bit, as does resizing the browser (although the latter is something most users probably won't do).

alexgibson:
Overall things are looking good :+1: Still some tidy up to do, and some bugs / optimizations to work out in the sticky header/footer logic. I still think loading that huge non-optimized website in the iframe is a step too far given it's current state, but happy to defer to whatever decision has been made there.

media/js/teach/smarton.js
craigcook: Only because this is how I did it on another page three years ago and I just copied from there. Laziness + ignorance! I'll switch to Waypoint.Sticky.

media/js/teach/smarton.js
craigcook: Yeah, it's still not perfect but may be the best we can do. They don't want to send people away from our site and opening a new tab might be even worse. 

I'm writing in the iframe src when the modal opens to prevent the page from loading while it's hidden, hence why it starts over if you close and reopen the modal, but I do need to also destroy it when it closes so the loading can stop.

media/css/teach/smarton.less
craigcook: I was still tinkering with this while working on mobile nav... it's meant to shave a few more pixels off the header in short viewports, but now that I'm fairly happy with then mobile nave I think I can uncomment it.

media/js/teach/smarton.js
craigcook: Knew I forgot something... Added now.

media/js/teach/smarton.js
```diff
@@ -0,0 +1,300 @@
...
+    setInterval(adjustScrollbar, 100);
+
+    // Scroll smoothly to the linked section
+    $document.on('click', '.nav-steps a[href^="#"]', function(e) {
+        e.preventDefault();
+        // Extract the target element's ID from the link's href.
+        var elem = $(this).attr('href').replace( /.*?(#.*)/g, '$1' );
+        Mozilla.smoothScroll({
+            top: $(elem).offset().top - headHeight
+        }, function() {
```
alexgibson: Hmm, smoothScroll doesn't accept a callback function like this, so you may have to rework this logic a little?

craigcook: Turns out we don't need to fire anything after scrolling anyway. :sweat_smile:

craigcook:
So I remembered what this does for the scrolling nav:
`$(elem).attr('tabindex','100').focus().removeAttr('tabindex');`

It adds a tabindex to the target element so it can receive focus (normally only links and form controls can receive focus), focuses on it, then removes the tabindex. Otherwise clicking in the nav scrolls down the page but the link you clicked still has focus and hitting tab goes to the next link in the navbar, rather than the first link in the section you just scrolled to. If you tab through the rest of the nav, the next link after it will be the first link at the top of the page so the whole page jumps back up to the top. This hack means you can jump down the page and tab right into the section in view.

media/js/teach/smarton.js
```diff
@@ -0,0 +1,261 @@
...
+    // Scroll smoothly to the linked section
+    $document.on('click', '.nav-steps a[href^="#"]', function(e) {
+        e.preventDefault();
+        // Extract the target element's ID from the link's href.
+        var elem = $(this).attr('href').replace(/.*?(#.*)/g, '$1');
+        Mozilla.smoothScroll({
+            top: $(elem).offset().top - $pageHeader.height()
+        }, function() {
+            // Add a tabindex to the target element so it can recieve focus, focus it, then remove tabindex
+            $(elem).attr('tabindex','100').focus().removeAttr('tabindex');
```
alexgibson: Like I said, this line here won't work, as the smooth scroll doesn't accept a callback as the last parameter. You might need to rethink how you're doing this (or if the hack is really needed, should natural tab order of the nav not be respected? But maybe I'm misunderstanding what you mean).

alexgibson: Re-reading your explanation I think I see where you're coming from, but this will still need to be reworked somehow if you want to keep it, but there's no callback for when a native scroll event has ended. Maybe something we could look to add in the helper.

craigcook: Ah I see now, didn't realize there was no callback, it wasn't throwing an error and seemed to work as expected initially, so not sure what I was seeing. I just tested again and, while there was still no error, it wasn't actually working after all. So, removed again.

I have mixed feelings, honestly. I'd like to respect natural tab order since I imagine that's what serious keyboard users will expect, and yet, the sticky nav still lets you jump down the page and it's jarring to then be swept back to the top, defeating the purpose of the nav links. Frankly I'm not a fan of this kind of sticky nav in the first place because it's usually more trouble than it's worth and I don't know that it actually helps anyone. Maybe some day we can do user testing with real people.

media/css/teach/smarton.less
craigcook: It inherits a fallback color from the header (`#fff`)

craigcook:
To do:
- [ ] video thumbnail aspect scaling
- [ ] IE testing/hacking

Going to work on IE over the weekend. I'd love to avoid a separate stylesheet but IE8 will probably need a few special rules to compensate for its lack of flexbox and SVG support.

media/js/teach/smarton.js
```diff
@@ -0,0 +1,258 @@
...
+        });
+    }
+
+    // Scroll smoothly to the linked section
+    $document.on('click', '.nav-steps a[href^="#"]', function(e) {
+        e.preventDefault();
+        // Extract the target element's ID from the link's href.
+        var elem = $(this).attr('href').replace(/.*?(#.*)/g, '$1');
+        Mozilla.smoothScroll({
+            top: $(elem).offset().top - $pageHeader.height()
```
alexgibson: On mobile viewports the `$pageHeader` seems to be miscalculated, due to the nav height changing dynamically. This results in the scroll position being incorrect when you click the nav links.

craigcook:
After fighting with IE8 for a good chunk of time I got the layout pretty acceptable, but there were still weird issues with the sticky header and the circles script that draws the pie charts was throwing an error. I finally decided it likely wasn't worth the effort to make everything look good in IE8, so instead I'm just hiding the CSS and JS from IE 8 and lower, and giving them just a few extra style rules to improve readability. It's a cop-out, but I'm prepared to live with it. 

We can perhaps spend a bit more time on IE issues for the next two smarton pages (thus retroactively fixing them for this one) but I've been giving serious thought to dropping IE8 for the whole site when we eventually do the redesign. We should discuss that further, but for this page right now I think this is the simplest way forward.

alexgibson:
I have no qualms whatsoever in dropping support for IE8 on pages like this, and totally agree we should consider ending support for it in general (bar the odd download page) :+1:

media/js/teach/smarton.js
```diff
@@ -0,0 +1,258 @@
...
+                embedWrap.find('.modal-video-frame').attr('src', embedSrc);
+            },
+            onDestroy: function() {
+                embedWrap.find('.modal-video-frame').removeAttr('src');
+            }
+        });
+
+        // Track the click and record the video ID
+        window.dataLayer.push({
+            event: 'smarton-interactions',
```
alexgibson: Nit: GA event calls should have quotes, e.g.

    window.dataLayer.push({
        'event': 'smarton-interactions',
        'interaction': 'video modal open',
        'video': $this.data('element-id')
    });

Same for here and elsewhere in this file.

alexgibson:
One mobile nav positioning issue remains, and a minor nit with the GA event calls - but otherwise this gets an r+wc from me :ghost:. 

I think this page is in good shape, both scrolling and sticky header are noticeably smoother with the latest updates, nice work! :+1:

media/js/teach/smarton.js
craigcook: Hmm yeah, that's not good. Not sure how best to work around it... We could supply a fixed offset that's the estimated height of the bar after it sticks/collapses, but that may be off by a bit if the title wraps to two lines (which it likely will for longer titles, especially when localized). 

Perhaps better to forego the sticky nav entirely on mobile? Then it wouldn't collapse and the height would remain the same, but we lose the title and the indication of which section you're in. Maybe we can live with that. 

I'll be at a conference all day but will have laptop+wifi so I'll come up with something today.

alexgibson: I think we'd either need to calculate the height offset onclick, or like you say forego the sticky nav on mobile if that doesn't work out. Happy to go with whatever you think is best/have time for. Enjoy the conf!

media/js/teach/smarton.js
```diff
@@ -0,0 +1,258 @@
...
+        });
+    }
+
+    // Scroll smoothly to the linked section
+    $document.on('click', '.nav-steps a[href^="#"]', function(e) {
+        e.preventDefault();
+        // Extract the target element's ID from the link's href.
+        var elem = $(this).attr('href').replace(/.*?(#.*)/g, '$1');
+        Mozilla.smoothScroll({
+            top: $(elem).offset().top
```
alexgibson: This calculation is now broken on desktop if you remove the nav height entirely?

craigcook: Oops, still tinkering with a fix/workaround for this issue, should have put the height back before that last push. If you have any ideas I'm very open to suggestions.

alexgibson: I think the only way to do this and retain the nav for mobile would be to calculate the height of the header after the button is clicked, instead of once the page has loaded. Either that or use a fixed known height, one for desktop and one for mobile (don't know how that effects l10n though).

craigcook: Decided to just unstick the header in small viewports, which also means we get some screen real estate back. The scroll offset now depends on the viewport size, and with the unsticky header it's more predictable.

alexgibson:
Thanks @craigcook - I think this looks fine for now. I notice the nav remains sticky if I resize the browser to mobile after the page has loaded, but I think this is acceptable for now as most people won't resize their browser. We can always follow up later with some tweaks if needed :+1:

craigcook:
Hm yeah, resizing is awkward, though the matchMedia stuff should update. We'll put it on the "fix it later" list. Thanks!

m-alexis:
Craig, I still see the "see all smartons" link in the header. Has that been removed but not pushed to demo1? I know you're at a conference today and Ben volunteered to make that change if needed.

ckprice:
@m-alexis this was pushed to [stage](https://www.allizom.org/en-US/teach/smarton/tracking/), it's not in the header there.

m-alexis:
ah ok. false alarm!

PR #3444, opened 2015-10-16
Fix Bug 939470 - mozilla.org tells user it has up-to-date Firefox when it doesn't

kyoshino:
Now it works as expected. Please push this to any available test server so we can confirm the new behaviour. But I'm still wondering how to add tests for the `_getFirefoxDetails` method. Any ideas...?

### DIFF
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -3,7 +3,7 @@
 <!doctype html>
 {# Note the "windows" class, without javascript platform-specific
      assets default to windows #}
-<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions }}">
+<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|safe }}">
   <head>
     <meta charset="utf-8">{# Note: Must be within first 512 bytes of page #}
 
--- a/bedrock/base/templates/base.html
+++ b/bedrock/base/templates/base.html
@@ -3,7 +3,7 @@
 <!doctype html>
 {# Note the "windows" class, without javascript platform-specific
      assets default to windows #}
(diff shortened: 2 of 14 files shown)


alexgibson:
Thanks for taking a look at this big @kyoshino! I haven't really dug into the code yet, but is there a technical reason why all this code needs to move into (what was) `site.js`? JS in the head is render blocking, so we should try and be extra diligent about moving more code here if it may not be strictly required?

kyoshino:
Thanks @alexgibson! I forgot the fact `site.js` was in `<head>`. I think merging environment detection code into one single namespace makes sense, but yes, the Firefox part can/should be loaded separately only on requited pages. Will revise the PR tomorrow and fix tests as well.

alexgibson:
Thanks @kyoshino - I appreciate the work you put in so far! I think we should leave site.js intact, and only make additions to it that effect initial page rendering (e.g layout specific CSS classes). Functions like the Firefox ones should remain in global.js. Doesn't need to be part of this PR, but eventually I'd like to move the Firefox functions to their own module that we can load explicitly when needed.

kyoshino:
Okay, let's reorganize the Firefox functions in a separate PR. Will revise this shortly.

kyoshino:
Alright, the new `getFirefoxDetails` method just does the right thing.

kyoshino:
Okay, as we talked on Vidyo today, let's discuss this further with wider stakeholders on the bug or mailing list.

Technically this does the right thing, so how about testing it on non-`/firefox/new/` pages using GA events, so we can analyse how many users are hitting the fallback code?

alexgibson:
@kyoshino - I'd be fine with creating a new function for this, which we can use as-and-when we choose on specific pages? I still think we should probably limit the event firing to only users whom we know are on a non-up-to-date ESR release version. If this is a way to unblock this PR, then let's go with that for now, while we decide how to handle /firefox/new/ :+1:

kyoshino:
As I said, this bug is not limited to ESR, but exists with all minor updates. Currently we don't have any data so some GA testing might be good anyway. I'll file a separate bug and PR to add the new function with GA events except for /firefox/new/.

alexgibson:
Ok @kyoshino that sounds good. I understand about the minor point release, I keep forgetting as this bug started out with ESR in mind :+1:

kyoshino:
Will send a new PR once the decision for /firefox/new/ is made.

PR #3446, opened 2015-10-16
Fix Bug 1215643 - Adjust schema.org software applications markup on /firefox/new/ and also add to homepage

kyoshino:


craigcook:
Checked with http://linter.structured-data.org/ and all seems good. r+ :tropical_fish:

PR #3447, opened 2015-10-19
[fix bug 1215633] Temporarily redirect /firefox/firstrun/learnmore/ to SUMO

alexgibson:


alexgibson:
@pmclanahan r?

I can't quite work out what I need to do to fix this failing test. It passes fine if I omit `permanent=False`, but looks like it's not getting the expected status code when it's temporary?

alexgibson:
Ah, think I fixed it by adding `status_code=302` to the test.

PR #3449, opened 2015-10-19
Fix intermittent test failure on firefox dnt page

alexgibson:
* Waits for accordion to fully expand before clicking again to collapse.

### DIFF
--- a/tests/pages/firefox/do_not_track.py
+++ b/tests/pages/firefox/do_not_track.py
@@ -3,7 +3,6 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from selenium.webdriver.common.by import By
-from selenium.webdriver.support import expected_conditions as expected
 from selenium.webdriver.support.ui import WebDriverWait as Wait
 
 from .base import FirefoxBasePage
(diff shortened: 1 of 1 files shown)


alexgibson:
@davehunt r? This should hopefully eliminate both scenarios we discussed.

tests/pages/firefox/do_not_track.py
```diff
@@ -31,15 +31,23 @@ class FrequentlyAskedQuestion(PageRegion):
...
     def show_answer(self):
         assert not self.is_answer_displayed, 'Answer is already displayed'
-        el = self.root.find_element(*self._answer_locator)
-        self.root.find_element(*self._question_locator).click()
-        Wait(self.selenium, self.timeout).until(expected.visibility_of(el))
+        question = self.root.find_element(*self._question_locator)
+        answer = self.root.find_element(*self._answer_locator)
+        # Prevent Selenium automatically scrolling this element to the top of
+        # the page, which can lead to unintented clicks on the sticky header
+        self.selenium.execute_script('arguments[0].scrollIntoView(false);', question)
```
davehunt: I'm not convinced we need this force scroll into view. I think the issue is much more likely to be the element presenting as visible before it's finished transitioning.

tests/pages/firefox/do_not_track.py
```diff
@@ -31,15 +31,23 @@ class FrequentlyAskedQuestion(PageRegion):
...
+        question.click()
+        Wait(self.selenium, self.timeout).until(expected.visibility_of(answer))
 
     def hide_answer(self):
         assert self.is_answer_displayed, 'Answer is already hidden'
+        # Wait until the accordion is fully expanded before clicking it again
+        Wait(self.selenium, self.timeout).until(
+            expected.presence_of_element_located(self._expanded_locator))
```
davehunt: Rather than having another locator, we should reuse the answer locator and wait for it to have the appropriate attribute. Also, we should do this in `show_answer` after clicking the question, to prevent any further action until after the answer is fully displayed.

```python
answer = self.root.find_element(*self._answer_locator)
Wait(self.selenium, self.timeout).until(
    lambda m: answer.get_attribute('aria-expanded') == 'true')
```

alexgibson:
@davehunt updated, thanks

davehunt:
I wasn't able to replicate the intermittent locally, but I think this makes the interaction more robust. r+

PR #3451, opened 2015-10-19
Bug 1216225 - Add function to provide accurate Firefox channel/version info based on mozUITour API

kyoshino:
This is basically a copy of #3444 but I have removed the change to /firefox/new/, added GA event tracking for testing purposes, and shortened the timer to 200ms.


media/js/base/global.js
```diff
@@ -149,6 +137,174 @@ function isFirefoxESR(userAgent, buildID) {
...
+/**
+ * Use the async mozUITour API of Firefox to retrieve the user's browser info,
+ * including the update channel and accurate, patch-level version number. This
+ * API is available on Firefox 35 and later.
+ * http://bedrock.readthedocs.org/en/latest/uitour.html
+ *
+ * @param  {Function} callback - callback function to be executed
+ * @return {None}
+ */
+function getFirefoxDetails(callback) {
```
alexgibson: It would be good to get some sort of test coverage for this function. I don't think we need to go as far as mocking the custom events firing/receiving, but it would be nice to at least sanity check that the callbacks are formed as we expect and dealt with correctly. This may mean splitting the logic here up into several smaller functions which can be tested individually. If this needs to be split up, might it be better as it's own module rather than leaving everything in global?

media/js/base/global.js
```diff
@@ -34,15 +34,19 @@ function init_download_links() {
...
-    if (isFirefox() && !isFirefoxUpToDate()) {
-        // look at each button to see if it's set to check for old firefox
-        $('.download-button').each(function() {
-            var $button = $(this);
-
-            if ($button.hasClass('download-button-check-old-fx')) {
-                // replace subtitle copy
-                $button.find('.download-subtitle').text(window.trans('global-update-firefox'));
+    if (isFirefox()) {
+        window.getFirefoxDetails(function(data) {
```
alexgibson: This is firing a UITour event for every page on bedrock. Looking at the code however, it seems as though there is only a single occurrence of this particular button style on the [`/plugincheck` page](https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templates/mozorg/plugincheck.html#L100). This also means the /plugincheck page is firing this custom event twice to retrieve the same information. 

Instead of calling this function automatically for every page on this site, should we instead omit the up-to-date logic here and call it explicitly from `check-plugins.js`?

media/js/base/global.js
```diff
@@ -149,6 +137,174 @@ function isFirefoxESR(userAgent, buildID) {
...
+    // Trigger the API
+    document.addEventListener('mozUITourResponse', listener);
+    document.dispatchEvent(new CustomEvent('mozUITour', {
+        'bubbles': true,
+        'detail': {
+            'action': 'getConfiguration',
+            'data': { 'configuration': 'appinfo', 'callbackID': callbackID }
+        }
+    }));
+};
```
alexgibson: Unnecessary semicolon.

media/js/base/global.js
```diff
@@ -149,6 +137,174 @@ function isFirefoxESR(userAgent, buildID) {
...
+            }
+        }
+
+        if (isUpToDate) {
+            break;
+        }
+    }
+
+    return isUpToDate;
+};
```
alexgibson: Unnecessary semicolon.

media/js/base/global.js
```diff
@@ -149,6 +137,174 @@ function isFirefoxESR(userAgent, buildID) {
...
+ * mainly used for security notifications.
+ *
+ * @param  {Boolean} strict - whether the minor and patch-level version numbers
+ *                   should be compared. Default: true
+ * @param  {Boolean} isESR - whether the Firefox update channel is ESR
+ * @param  {String}  userVer - browser's version number
+ * @return {Boolean} result
+ */
+function isFirefoxReallyUpToDate(strict, isESR, userVer) {
+    strict = strict === undefined ? true : strict;
```
alexgibson: Let's add `'use strict';`

media/js/base/global.js
```diff
@@ -149,6 +137,174 @@ function isFirefoxESR(userAgent, buildID) {
...
+ * Use the async mozUITour API of Firefox to retrieve the user's browser info,
+ * including the update channel and accurate, patch-level version number. This
+ * API is available on Firefox 35 and later.
+ * http://bedrock.readthedocs.org/en/latest/uitour.html
+ *
+ * @param  {Function} callback - callback function to be executed
+ * @return {None}
+ */
+function getFirefoxDetails(callback) {
+    var callbackID = Math.random().toString(36).replace(/[^a-z]+/g, '');
```
alexgibson: Let's add 'use strict';

media/js/base/global.js
alexgibson: This is firing a UITour event for every page on bedrock. Looking at the code however, it seems as though there is only a single occurrence of this particular button style on the [`/plugincheck` page](https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templates/mozorg/plugincheck.html#L100). This also means the /plugincheck page is firing this custom event twice to retrieve the same information. 

Instead of calling this function automatically for every page on this site, should we instead omit the up-to-date logic here and call it explicitly from `check-plugins.js`?

kyoshino:
Thanks @alexgibson!! I'll revisit this tomorrow.

media/js/base/global.js
kyoshino: :+1: Let's add those new functions to a new js module and cache the API's query result on it.

kyoshino:
* Put the new functions on `Mozilla.Client`. With a separate PR, I'll move other related functions such as `isFirefox()` to the new module.
* Modified `update_download_text_for_old_fx()` so `getFirefoxDetails()` is now called only when needed.
* Modified `getFirefoxDetails()` so it now returns a cached result when available.
* Added a test for `getFirefoxDetails()`, though it's still failing somehow.

media/js/base/global.js
```diff
@@ -34,15 +34,22 @@ function init_download_links() {
...
-        $('.download-button').each(function() {
-            var $button = $(this);
-
-            if ($button.hasClass('download-button-check-old-fx')) {
-                // replace subtitle copy
-                $button.find('.download-subtitle').text(window.trans('global-update-firefox'));
+    var $buttons = $('.download-button');
+
+    // if using Firefox and the page has download buttons
+    if (isFirefox() && $buttons.length) {
```
alexgibson: This is still firing a UITour event for every bedrock page that has a download button, which is still way more than we need. We should be as conservative as possible.

media/js/firefox/desktop/tips.js
```diff
@@ -15,8 +15,12 @@
 
     // only show download button for users on desktop platforms, using either a non-Firefox browser
     // or an out of date version of Firefox
-    if ((window.isFirefox() && window.isFirefoxUpToDate()) || $html.hasClass('android') || $html.hasClass('ios') || $html.hasClass('fxos')) {
-        $('#footer').addClass('hide-download');
+    if (window.isFirefox() || $html.hasClass('android') || $html.hasClass('ios') || $html.hasClass('fxos')) {
```
alexgibson: We need to rethink conditional logic such as this. `getFirefoxDetails()` should only be called for Firefox Desktop users. Other browsers / platforms should not be penalized with delayed loading times, especially for mobile.

media/js/base/mozilla-client.js
```diff
@@ -0,0 +1,153 @@
...
+     * Detect whether the user's Firefox is up to date or outdated. This data is mainly used for security notifications.
+     *
+     * @param  {Boolean} strict - whether the minor and patch-level version numbers should be compared. Default: true
+     * @param  {Boolean} isESR - whether the Firefox update channel is ESR
+     * @param  {String}  userVer - browser's version number
+     * @return {Boolean} result
+     */
+    Client._isFirefoxUpToDate = function (strict, isESR, userVer) {
+        strict = strict === undefined ? true : strict;
+        isESR = isESR === undefined ? window.isFirefoxESR() : isESR;
```
alexgibson: If we're moving to this method for ESR detection, I don't think we should rely on the old `isFirefoxESR()` function as a fallback? The main advantage of this bug was to remove the need for having to manually maintain build ID's.

kyoshino: Yes, the legacy `isFirefoxESR` function should be gone... [`navigator.buildID` will be removed from Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=583181). There would be some false positives in case the API doesn't work, but `isFirefoxESR` also has false positives as noted, so it should be acceptable.

media/js/firefox/desktop/tips.js
kyoshino: Changed the if-else logic here.

media/js/base/mozilla-client.js
kyoshino: Removed the `isFirefoxESR` function.

media/js/base/global.js
kyoshino: I think calling the API is not so expensive, and download buttons are not on every page... For people browsing a couple of pages, caching `FirefoxDetails` with `sessionStorage` might help?

alexgibson: I don't think we should be storing this kind of information in session storage. Please, let's just call this where needed.

alexgibson: It may not be expensive in the grand scheme of things, but it is wasteful taking into account the high number of pages that don't need this user information.

kyoshino: Hmm, okay. Looks like the `download-button-check-old-fx` class is currently not used on any page. I'll update the jQuery selector to check the class name first.

alexgibson: Thanks, yeah like I said it's only used on one page /plugincheck.

kyoshino:
Modified the `update_download_text_for_old_fx` function and added caching with `sessionStorage` (not `localStorage`, of course) to minimize API calls. And the test is now passing!

media/js/base/mozilla-client.js
```diff
@@ -0,0 +1,171 @@
...
+                'channel': channel,
+                'isUpToDate': isUpToDate,
+                'isESR': isESR
+            };
+
+            callback(details);
+
+            // Cache the data with sessionStorage
+            try {
+                sessionStorage.setItem(ssKeyName, JSON.stringify(details));
```
alexgibson: Like I mentioned in the previous comment, I don't think we should be storing user data is session storage like this unless there is a really good reason. Session storage is also synchronous, so another reason I think we should avoid using it.

kyoshino: Okay, removed `sessionStorage`.

alexgibson:
@kyoshino - looks like this one needs a rebase?

One other observation looking at the latest code is we no longer seem to be returning the fallback straight away if the user is on FF < 35. Seems like this might be better than waiting for the timeout, since we know it will not work on older versions?

Will try and get this up on a demo server soon so we can do some more thorough testing. Thanks for your patience :+1:

kyoshino:
Thanks @alexgibson! Rebased and added back the immediate fallback for Firefox < 35.

media/js/base/global.js
```diff
@@ -34,15 +34,16 @@ function init_download_links() {
...
-        // look at each button to see if it's set to check for old firefox
-        $('.download-button').each(function() {
-            var $button = $(this);
-
-            if ($button.hasClass('download-button-check-old-fx')) {
+    // look at each button to see if it's set to check for old firefox
+    var $buttons = $('.download-button-check-old-fx');
+
+    // if using Firefox and the page has download buttons
+    if ($buttons.length && isFirefox()) {
```
alexgibson: Should we be checking for Firefox Desktop here explicitly? (`isFirefox()` will return true for Android as well, if I remember correctly).

media/js/firefox/desktop/tips.js
```diff
@@ -15,7 +15,13 @@
 
     // only show download button for users on desktop platforms, using either a non-Firefox browser
     // or an out of date version of Firefox
-    if ((window.isFirefox() && window.isFirefoxUpToDate()) || $html.hasClass('android') || $html.hasClass('ios') || $html.hasClass('fxos')) {
+    if (window.isFirefox()) {
```
alexgibson: Same here, would Firefox for Android also fall into this?

media/js/firefox/desktop/index.js
```diff
@@ -41,7 +41,11 @@
         });
     }
 
-    if (window.isFirefoxUpToDate()) {
-        $('#overview-intro-up-to-date').addClass('active');
+    if (window.isFirefox()) {
```
alexgibson: Same here, would Firefox for Android also fall into this?

media/js/firefox/family-index.js
```diff
@@ -24,11 +24,9 @@
 
     // Check Firefox version
     if (isFirefox()) {
```
alexgibson: Same here, would Firefox for Android also fall into this?

media/js/plugincheck/check-plugins.js
```diff
@@ -162,8 +162,12 @@ $(function() {
     }
 
     // show for outdated Fx versions
-    if (isFirefox() && !isFirefoxUpToDate() && !isFirefoxESR()) {
-        outdatedFx.show();
+    if (isFirefox()) {
```
alexgibson: And again here?

media/js/base/global.js
```diff
@@ -34,15 +34,16 @@ function init_download_links() {
...
 function update_download_text_for_old_fx() {
-    // if using an out of date firefox
-    if (isFirefox() && !isFirefoxUpToDate()) {
-        // look at each button to see if it's set to check for old firefox
-        $('.download-button').each(function() {
-            var $button = $(this);
-
-            if ($button.hasClass('download-button-check-old-fx')) {
+    // look at each button to see if it's set to check for old firefox
+    var $buttons = $('.download-button-check-old-fx');
```
alexgibson: We could also move this selector inside the `isFirefox` conditional to avoid querying the DOM for other browsers?

media/js/base/mozilla-client.js
```diff
@@ -0,0 +1,165 @@
...
+     */
+    Client.getFirefoxDetails = function (callback) {
+        // Fire the callback function immediately if cache exists
+        if (this.FirefoxDetails) {
+            callback(this.FirefoxDetails);
+
+            return;
+        }
+
+        var self = this;
```
alexgibson: Seeing as this is a static object, is there any need to use `this`, and not reference `Client` explicitly?

media/js/base/global.js
kyoshino: Not sure why the platform check is missing in the original code, but I can add it.

kyoshino: Hm, maybe it's okay to handle Firefox for Android as well here.

media/js/base/mozilla-client.js
kyoshino: :+1:

media/js/firefox/desktop/index.js
kyoshino: This one is also okay, I guess.

media/js/firefox/desktop/tips.js
kyoshino: The original condition has `$html.hasClass('android')` so it's fine.

media/js/firefox/family-index.js
kyoshino: It's fine as well IMO.

media/js/plugincheck/check-plugins.js
kyoshino: It's fine as well IMO.

kyoshino:
So basically Firefox for Android can be handled the same way as the desktop equivalent. Firefox for iOS is a different beast, but the current `isFirefox()` function returns `false` in that case. I'll refactor those functions in the next iteration anyway.

media/js/base/global.js
alexgibson: What I'm getting at is the new UITour function should not be called for desktop users only? It won't work on Android?

kyoshino:
Rebased against master. Tests passed!

alexgibson:
@kyoshino - the reason I queried about Android users was because the new UITour function is now desktop specific. It won't work on Android, so we need to make sure we take this into account somehow. Currently would Android users have to wait an extra 200ms to see content?

media/js/base/global.js
kyoshino: I haven't tested yet on Android, actually. If the API doesn't work on Android for Firefox, it can be another exception to immediately fallback, like Firefox < 35. Let me test on Android now...

kyoshino:
Confirmed the API is not working on Android. Okay, then I'll add a platform check in `Client.getFirefoxDetails()` so the 200ms delay will be gone.

media/js/base/global.js
alexgibson: The API doesn't work for Firefox Android, this I'm confident in fwiw.

alexgibson:
Sounds great, thanks @kyoshino! :+1:

kyoshino:
Added a platform check as well as a note on the UITour document.

alexgibson:
I've pushed this to demo2 so we can do some manual testing, thanks @kyoshino :+1:

alexgibson:
@kyoshino - when testing on demo2 it occurred to me it might be helpful to be able to call `Mozilla.Client.FirefoxDetails` in the console, and see if it was the uitour event or the fallback that was returned. Would it be worthwhile adding `accurate` to this object too, it might come in useful?

alexgibson:
I've given this some testing on demo2 in various Firefox release and ESR versions, and things are looking good code-wise, thanks @kyoshino!

The only scenario I've not been able to test is an outdated ESR that also supports the UITour appInfo function (since it was only introduced in Firefox 35). Looking at the code I think it should handle this fine, but as an extra measure could we please add a unit test for this combination?

The only other minor visual change introduced here is on the /firefox/products page. Now there is a visible jump when the download bar at the top of the page hides again. It's only a minor thing really, but I was wondering if we could perhaps change the logic here to be hidden by default for Firefox users, instead of visible?

I also still need to sync up with Gareth to make sure we're recording this data in GA.

alexgibson:
I would also appreciate if either @pmclanahan or @jgmize could take a peek at the Python code here. LGTM, but a second pair of eyes would be appreciated.

pmclanahan:
Python bits look good. r+

kyoshino:
As I said in the meeting, the API doesn't work in the tests anyway so we are already covering the old version scenario I guess.

Updated the style and script for the Firefox family page to avoid the download bar bumping.

alexgibson:
> As I said in the meeting, the API doesn't work in the tests anyway so we are already covering the old version scenario I guess.

What I was meaning was, don't the current tests stub out `isESR`? So something that is ESR but not in`data-esr-versions` (sorry if I didn't make sense).

kyoshino:
When the API doesn't work (in the tests), `isESR` will be `false` and `channel` will be `release` even on ESR, as the legacy `navigator.buildID`-based detection is now removed. Hm  :confused:

alexgibson:
I'm probably not being clear, I'm just talking about stubbing out a test for `_isFirefoxUpToDate`, where `isESR` is true but the version is not in `data-esr-versions`?

kyoshino:
Thanks @alexgibson for clarification! Added the 5th test for `_isFirefoxUpToDate` saying "should consider outdated if user version is much less than latest version, regardless of strict option".

alexgibson:
Perfect, thanks @kyoshino! :smile:

alexgibson:
Code-wise this gets a big r+ from me, thanks @kyoshino! :bow:

alexgibson:
@kyoshino - could you please squash these 2 commits, then I believe we're good to merge thanks :+1:

kyoshino:
Doh, removed that unintentional commit!

alexgibson:
Thanks @kyoshino!

PR #3453, opened 2015-10-20
Fix Bug 1206781, adds page for b2gdroid fte

schalkneethling:
Waiting on feedback regarding GA tracking for the contribute link.


bedrock/firefox/templates/firefox/os/b2gdroid.html
```diff
@@ -0,0 +1,24 @@
...
+      {% stylesheet 'firefox_os_b2gdroid' %}
+    {% endblock %}
+  </head>
+  <body>
+    <div class="wrapper">
+      <header>
+        <h1>{{ _('Welcome') }}</h1>
+      </header>
+      <main rol="main">
+        <p>{{ _('Welcome to B2GDroid, an experiment from Mozilla that let’s you experience Firefox OS. B2GDroid is an early preview and we’d love your help to shape the future of the Web.') }}</p>
```
flodolo: Argh. (lets)

bedrock/firefox/templates/firefox/os/b2gdroid.html
```diff
@@ -0,0 +1,24 @@
...
+        <p>{{ _('To find out how to get involved, visit:') }} <a href="{{ url('mozorg.contribute') }}" rel="external">https://www.mozilla.org/contribute</a></p>
+      </main>
+      <footer>
+        <button class="more">Next</button>
```
flodolo: Wrap string

schalkneethling:
Update regarding l10n: "Not for Nov. 2nd as the OS localization won't be ready for then, but it will be
needed thereafter." - from https://bugzilla.mozilla.org/show_bug.cgi?id=1206781#c19

alexgibson:
@schalkneethling - looks like this is not needed on bedrock according to the bug, shall we close this PR?

schalkneethling:
Yeah, nobody outright confirmed but, all indicators seems to point in that direction. I reckon I will do a quick follow up with @m-alexis to confirm before deleting.

schalkneethling:
looks like this is no longer required so, closing

PR #3456, opened 2015-10-20
Bug-1215145-fix-lang-selector

TheJJ100100:
Hi, I think I have fixed bug 1215145, fixing the language selector.

https://bugzilla.mozilla.org/show_bug.cgi?id=1215145

I only just realized it was assigned to someone after completing necessary changes. I wasn't sure whether to do a pull request or not, so I did anyway.

Thank You.

### DIFF
--- a/bedrock/base/templates/includes/lang_switcher.html
+++ b/bedrock/base/templates/includes/lang_switcher.html
@@ -4,8 +4,8 @@
 
 {% if translations|length > 1 %}
 <form id="lang_form" method="get" action="#">
-  <label for="language">{{ _('Other languages:') }}</label>
-  <select id="language" name="lang" dir="ltr">
+  <label for="page-language-select">{{ _('Other languages:') }}</label>
+  <select id="page-language-select" name="lang" dir="ltr">
     {% for code, label in translations|dictsort -%}
       {# Don't escape the label as it may include entity references
        # like "Português (do&nbsp;Brasil)" (Bug 861149) #}
--- a/bedrock/tabzilla/templates/tabzilla/tabzilla.js
+++ b/bedrock/tabzilla/templates/tabzilla/tabzilla.js
@@ -380,7 +380,7 @@ var Tabzilla = (function (Tabzilla) {
         var offeredLang;
         var availableLangs = {};
         var $links = $('link[hreflang]');
-        var $options = $('#language option');
+        var $options = $('#page-language-select option');
 
         // Make a dictionary from alternate URLs or a language selector. The key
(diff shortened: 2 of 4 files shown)


TheJJ100100:
I'm not sure whether to change language in:

"<label for="language">Other languages:</label>"

to  "page-language-select"

jpetto:
Hey @TheJJ100100 -

It doesn't make sense to change the id of both the footer language selector *and* the language section on the `/firefox/android/` page to the same value. That's exactly the problem we're trying to solve, as an id can only be used once per page.

As [stated in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1215145#c0), I recommend changing only the id of the footer language selector. Doing so will require finding and updating use of this selector in related LESS and JS files. A quick search shows current uses [here](https://github.com/mozilla/bedrock/blob/master/bedrock/tabzilla/templates/tabzilla/tabzilla.js#L383) and [here](https://github.com/mozilla/bedrock/blob/master/media/js/base/global.js#L64). Be sure to perform your own thorough search to make sure all our bases are covered.

Thanks.

PS - Check out this [Markdown Cheatsheet](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#code) for tips on formatting code in GitHub comments.

TheJJ100100:
@jpetto Is this any better or worse?

jpetto:
Hi @TheJJ100100 -

It looks like you've found and updated files associated with an id change to the footer language select, but that's not the id you've changed. You've changed the id of the `<section>` on the `/firefox/android/` page. I recommend reverting the id change on `/firefox/android/index.html` and, as originally requested, changing the id on `base/templates/includes/lang_switcher.html` [here](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/includes/lang_switcher.html#L7-L8).

Please be sure to test the fix locally before pushing up new commits. When changing the language select in the footer of a page, you should be automatically redirected to the same page translated in the newly selected language.

Thanks!

TheJJ100100:
Done.

jpetto:
The code is looking good now. Nice! Last step is to squash the 8 commits down into one, and adjust the commit message to reference the bug and a short description of the problem fixed. I'd recommend the following as a commit message:

`[fix bug 1215145] Fix lang selector on /firefox/android/`.

If you aren't comfortable with [squashing commits](http://bedrock.readthedocs.org/en/latest/contribute.html#squashing-your-commits), let me know and I'll provide assistance.

jpetto:
Looks great. Thanks @TheJJ100100.

PR #3457, opened 2015-10-21
Adds latest fx version to transbar jsonp

schalkneethling:
This is needed for the decoupled update bar @alexgibson r?


schalkneethling:
Closing this in favor of a new/better approach

PR #3458, opened 2015-10-21
Add Python functional tests for contribute page

alexgibson:
* Adds functional tests for the `/contribute` page.
* Also adds a new class to the base page object for the embedded newsletter form, which can be reused on other pages.


alexgibson:
@davehunt r?

tests/pages/base.py
```diff
@@ -44,3 +49,83 @@ def select_language(self, value):
...
+            # Click the submit button to expand the form when not visible.
+            # Ideally here we would focus on the email field, but Selenium has issues
+            # dealing with focus events when the window is not active.
+            # See bug https://bugzilla.mozilla.org/show_bug.cgi?id=704583
+            self.root.find_element(*self._submit_button_locator).click()
+            Wait(self.selenium, self.timeout).until(
+                expected.visibility_of_element_located(self._privacy_policy_checkbox_locator))
+
+        @property
+        def is_form_detail_visible(self):
```
davehunt: Note: We've used `_displayed` instead of `_visible` elsewhere.

tests/pages/base.py
```diff
@@ -44,3 +49,83 @@ def select_language(self, value):
...
+            assert el.is_selected(), 'Privacy policy has not been accepted'
+
+        def click_sign_me_up(self):
+            self.root.find_element(*self._submit_button_locator).click()
+            Wait(self.selenium, self.timeout).until(
+                expected.visibility_of_element_located(self._thank_you_locator))
+
+        @property
+        def sign_up_successful(self):
+            el = self.selenium.find_element(*self._thank_you_locator)
```
davehunt: Note: This locator isn't within the `PageRegion` root. I think that's fine though.

tests/pages/contribute.py
```diff
@@ -0,0 +1,39 @@
...
+
+    _url = '{base_url}/{locale}/contribute'
+
+    _video_play_locator = (By.CSS_SELECTOR, '.video-play')
+    _video_close_locator = (By.ID, 'modal-close')
+    _video_modal_locator = (By.ID, 'modal')
+
+    _event_link_locator = (By.CSS_SELECTOR, '.extra-event .event-link')
+
+    def open_video_modal(self):
```
davehunt: Suggestion: A more descriptive name might be `play_video`. The video could even be a `PageRegion` with an `is_displayed` property and `close` function.

tests/pages/base.py
alexgibson: Yep correct, the element is actually a sibling of the root here.

tests/functional/test_contribute.py
```diff
@@ -0,0 +1,51 @@
...
+    page.newsletter.accept_privacy_policy()
+    page.newsletter.click_sign_me_up()
+    assert page.newsletter.sign_up_successful
+
+
+@pytest.mark.nondestructive
+def test_sign_up_fails_when_missing_required_fields(base_url, selenium):
+    page = ContributePage(base_url, selenium).open()
+    with pytest.raises(TimeoutException):
+        page.newsletter.click_sign_me_up()
```
davehunt: You might want to tweak the test names, as the last three are specific to the newsletter form on the contribute page but they don't make that clear. Perhaps `test_newsletter_default_values`, `test_successful_newsletter_sign_up` and `test_newsletter_sign_up_fails_when_missing_required_fields`.

Alternatively, you could create a `TestNewsletter` class and move these tests beneath. This would then appear in reports as `test_contribute.py::TestNewsletter.test_default_values`, etc.

tests/functional/test_contribute.py
```diff
@@ -0,0 +1,51 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+from selenium.common.exceptions import TimeoutException
+
+from ..pages.contribute import ContributePage
+
+
+@pytest.mark.nondestructive
+def test_open_close_modal(base_url, selenium):
```
davehunt: I'd either rename this to `test_open_close_video_model` or perhaps simply `test_play_video`.

davehunt:
Excellent, only minor nits, which you can feel free to ignore. r+

tests/pages/contribute.py
alexgibson: The video itself has a different location in the DOM to where the modal itself is created, so I'm not sure a `PageRegion` is more appropriate here (e.g. only the close button would be a child of the root). Will update the naming convention still :+1:

PR #3459, opened 2015-10-21
Add support for ios release notes

sylvestre:


sylvestre:
This is more for a proof of concept. There is no tests, some dead code, some links to update.

sylvestre:
@craigbook If I want this to be merged, am I on the right track? Thanks

craigcook:
@sylvestre You're on the right track, but your previous comment indicated you were still working on it so I tagged the PR to make sure nobody merged prematurely. Will you be adding further changes or is this ready for a final review now?

sylvestre:
No, I still need to work on this :) Thanks!

sylvestre:
@craigcook it is ok for me now!

bedrock/settings/base.py
```diff
@@ -734,6 +734,9 @@ def facebook_tab_url_lazy():
                             urlquote('utm_source=mozilla&utm_medium=Referral&'
                                      'utm_campaign=mozilla-org'))
 
+# Link to Firefox for iOS on Itunes
+ITUNES_FIREFOX_LINK = ('https://itunes.apple.com/nz/app/firefox-web-browser/id989804926')
```
alexgibson: We already have a link in `base.py` to the appropriate app store link here: https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L743

sylvestre:
@alexgibson Fixed. I wonder how I didn't see that :|

bedrock/firefox/templates/firefox/releases/release-notes.html
```diff
@@ -127,8 +127,12 @@
           {% if release.product == 'Firefox for Android' %}
           <li class="try3"><a href="https://support.mozilla.org/products/mobile/download-and-install">{{ _('Install') }}</a></li>
           {% else %}
+          {% if release.product == 'Firefox for iOS' %}
+          <li class="try3"><a href="https://support.mozilla.org/products/ios/install-and-update-firefox-ios">{{ _('Install') }}</a></li>
+          {% else %}
           <li class="try3"><a href="https://support.mozilla.org/products/firefox/download-and-install">{{ _('Install') }}</a></li>
           {% endif %}
+          {% endif %}
           <li class="try4"><a href="https://addons.mozilla.org/firefox/">{{ _('Customize') }}</a></li>
```
jgmize: Standard addons will not work with Firefox for iOS, so we should remove this link.

jgmize:
I did some testing locally after running `./manage.py rnasync` and settings DEV=True to ignore the `is_public` field currently set to `False` for the Firefox for iOS 1.1 release notes draft
 
* /firefox/ios/1.1/releasenotes/ works as expected
* /firefox/ios/latest/releasenotes/ does not yet redirect, because there is not yet an entry product-details
* /firefox/ios/1.1/system-requirements/ is a 404, because the system requirements field is currently empty for that release
* the default "complete list of changes" link to the bugzilla search will not work,  it needs to be overridden with the `bug_search_url` field for that release

Nothing on the list above requires a change to this PR, so once you remove the "customize" element I mentioned earlier, I think this will be good to merge, and you can follow up on the other items in product-details and nucleus. r+wc, nice work!

sylvestre:
Thanks for this great testing. I removed the customize element. 
I will fix the rest in a future PR

jgmize:
Thanks @sylvestre.

Since /firefox/ios/1.1/releasenotes/ will continue to return a 404 in production until the release has been marked public in nucleus, it was safe for me to merge this now, and we can validate the changes in product-details and nucleus against www-dev where the url will work due to `settings.DEV = True`.

sylvestre:
Excellent, many thanks. Do you know when https://www-dev.allizom.org/en-US/firefox/ios/1.1/releasenotes/ is going to be available ?

jgmize:
@sylvestre I forgot that we will need a change to global.conf to pass the traffic through from apache to bedrock. https://www-dev.allizom.org/b/en-US/firefox/ios/1.1/releasenotes/ is working now (note the `/b/` in the url) and I will follow up with a PR for the apache config change.

PR #3462, opened 2015-10-22
Exclude redirected locales from homepage language selector functional tests

alexgibson:
This should avoid intermittent failures like this https://webqa-ci.mozilla.com/job/bedrock.dev.win10.ie/14/HTML_Report/

For reference, said redirects are here: https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/redirects.py#L13-L20

### DIFF
--- a/tests/functional/test_l10n.py
+++ b/tests/functional/test_l10n.py
@@ -13,9 +13,10 @@ from ..pages.home import HomePage
 def test_change_language(base_url, selenium):
     page = HomePage(base_url, selenium).open()
     initial = page.footer.language
-    available = page.footer.languages
-    available.remove(initial)  # avoid selecting the same language
-    new = random.choice(available)  # pick a random lanugage
+    # avoid selecting the same language or locales that have homepage redirects
+    excluded = [initial, 'ja', 'zh-TW', 'zh-CN']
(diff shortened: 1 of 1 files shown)


alexgibson:
@davehunt r?

tests/functional/test_l10n.py
```diff
@@ -15,6 +15,10 @@ def test_change_language(base_url, selenium):
     initial = page.footer.language
     available = page.footer.languages
     available.remove(initial)  # avoid selecting the same language
+    excluded_locales = ['ja', 'zh-TW', 'zh-CN']  # avoid selecting locales that get redirected
+    for locale in excluded_locales:
+        if locale in available:
+            available.remove(locale)
```
davehunt: We could simplify this using a list comprehension:

```python
excluded = [initial, 'ja', 'zh-TW', 'zh-CN']
available = [l for l in page.footer.languages if l not in excluded]
```

alexgibson: ah, much nicer! :+1:

alexgibson:
@davehunt updated, thanks

davehunt:
r+

PR #3463, opened 2015-10-22
Fix Bug 1214822, redirect ios users to ios product page from choose

schalkneethling:
@alexgibson r?

### DIFF
--- a/bedrock/firefox/templates/firefox/choose.html
+++ b/bedrock/firefox/templates/firefox/choose.html
@@ -12,6 +12,13 @@
   {% stylesheet 'firefox_choose' %}
 {% endblock %}
 
+{#- See https://bugzilla.mozilla.org/show_bug.cgi?id=1214822#c6 -#}
+{% block js_pretrack %}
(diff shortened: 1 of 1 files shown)


alexgibson:
Should this redirect be waffled, like on the [`/new` page](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/new.html#L40-L44)?

schalkneethling:
So, the /choose page is going to be redirected to /new until campaign launch so, we probably do not need this to be waffled? Unless the iOS redirect should only become active after the global iOS launch.

alexgibson:
The bug suggests iOS users should be redirected to `firefox/ios` until after the iOS global launch. As far as I understand this is after the campaign launch, hence why we're also using the waffle switch on `/new`. Perhaps best ask in the bug for clarification.

schalkneethling:
@alexgibson That makes sense to me but, let me triple check.

alexgibson:
r+

PR #3464, opened 2015-10-22
Fix Bug 1217353, adds new infobar app to bedrock

schalkneethling:
@pmclanahan r?

### DIFF
--- a/bedrock/infobar/templates/infobar/infobar.jsonp
+++ b/bedrock/infobar/templates/infobar/infobar.jsonp
@@ -1,0 +1,14 @@
+{% set_lang_files "tabzilla/tabzilla" %}
+_({
+    {# L10n: You can replace 'in your language' in this string by the name of your language if you want. -#}
+    "message": "{{ _('Would you like to see this page in your language?')|js_escape }}",
+    {# L10n: Your translation can just say Yes -#}
+    "accept": "{{ _('Yes, please!')|js_escape }}",
+    {# L10n: Your translation can just say No -#}
+    "cancel": "{{ _('No, thanks.')|js_escape }}",
+    {# L10n: Update bar messaging -#}
+    "update_message": "{{ _('Looks like you’re using an older version of Firefox.')|js_escape }}",
+    "update_accept": "{{ _('Update to stay fast and safe.')|js_escape }}",
+    "update_cancel": "{{ _('Later')|js_escape }}",
+    "latestfx": "{{ latest_firefox_version }}"
+})
--- a/bedrock/infobar/urls.py
+++ b/bedrock/infobar/urls.py
@@ -1,0 +1,12 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from django.conf.urls import url
+
+from . import views
+
+
+urlpatterns = (
+    url(r'^infobar\.jsonp$', views.infobar_jsonp, name='infobar'),
+)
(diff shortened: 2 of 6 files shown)


bedrock/infobar/templates/infobar/infobar.jsonp
```diff
@@ -0,0 +1,14 @@
...
+    "message": "{{ _('Would you like to see this page in your language?')|js_escape }}",
+    {# L10n: Your translation can just say Yes -#}
+    "accept": "{{ _('Yes, please!')|js_escape }}",
+    {# L10n: Your translation can just say No -#}
+    "cancel": "{{ _('No, thanks.')|js_escape }}",
+    {# L10n: Update bar messaging -#}
+    "update-message": "{{ _('Looks like you’re using an older version of Firefox.')|js_escape }}",
+    "update-accept": "{{ _('Update to stay fast and safe.')|js_escape }}",
+    "update-cancel": "{{ _('Later')|js_escape }}",
+    "latestfx": "{{ latest_firefox_version }}",
```
pmclanahan: Trailing comma. Bad for JS right?

bedrock/infobar/urls.py
```diff
@@ -0,0 +1,11 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from django.conf.urls import url
+
+from . import views
+
```
pmclanahan: NIT: 2 spaces after imports.

pmclanahan:
Looking good. Is there a separate bug/issue for adding the bar back to bedrock?

bedrock/infobar/templates/infobar/infobar.jsonp
schalkneethling: It used to have a trailing comma but, let me change that and see if all is still good.

schalkneethling:
@pmclanahan Not yet, no.

bedrock/infobar/views.py
```diff
@@ -0,0 +1,56 @@
...
+
+    return inner_last_modified
+
+
+def _resp(request, path, ctype, context=None):
+    resp = l10n_utils.render(request, path, context, content_type=ctype)
+
+    is_enabled = not settings.TEMPLATE_DEBUG and settings.CDN_BASE_URL
+    if is_enabled and isinstance(resp, HttpResponseRedirect):
+        # CDN URL should be protocol relative, but that won't work
```
pmclanahan: I don't think this CDN stuff is going to work. The CDN only works with certain URLs, and I don't remember "/infobar" being one of them. I believe "/tabzilla" was added specially. I think we're fine just letting this be a normal bedrock URL.

bedrock/infobar/templates/infobar/infobar.jsonp
pmclanahan: May be an old rule, but I think to be as broadly supportive of browsers as possible a trailing comma is still bad. I could be wrong though.

bedrock/infobar/views.py
schalkneethling: I copied this from the current code in tabzilla. Should I just remove the above function and have:

```
resp = l10n_utils.render(request, 'infobar/infobar.jsonp', 'application/javascript', content_type=ctype)
resp['Access-Control-Allow-Origin'] = '*'
return resp
```
inside

```
def infobar_jsonp(request):
```

pmclanahan: Basically yes. It's `l10n_utils.render(request, 'infobar/infobar.jsonp', content_type='application/javascript')`.

schalkneethling:
@pmclanahan we good to merge?

PR #3466, opened 2015-10-22
Port remaining redirects from global.conf

pmclanahan:
This is the last of 'em. For global.conf that is.


pmclanahan:
@jgmize r?

jgmize:
Nice work, r+

PR #3467, opened 2015-10-22
Remove fx40-firstrun waffle switch. Bug 1150231.

jpetto:


jgmize:
@jpetto this lgtm other than the flake8 errors travis is complaing about. r+wc.

jpetto:
Ah, unused `import`. Fixed!

PR #3469, opened 2015-10-23
Fix Bug 1217211, redirect choose to new until November 3rd

schalkneethling:
@pmclanahan r?

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -10,6 +10,7 @@ from cgi import escape
 
 from django.conf import settings
 from django.http import HttpResponsePermanentRedirect, HttpResponseRedirect
+from django.utils.encoding import force_text
 from django.views.decorators.cache import never_cache
 from django.views.decorators.csrf import csrf_exempt, csrf_protect
 from django.views.decorators.http import require_POST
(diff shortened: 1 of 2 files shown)


bedrock/firefox/views.py
```diff
@@ -281,6 +282,14 @@ def fx_home_redirect(request):
     return HttpResponseRedirect(reverse('firefox.new'))
 
 
+def choose(request):
+    if waffle.switch_is_active('tracking-protection-redirect'):
```
pmclanahan: We normally make the state we want in prod now be the "off" state so that we don't have to make sure the flag is created and "ON" in all of the prod instances before we push.

pmclanahan: need to import `waffle`.

schalkneethling: Ah, gotcha. `waffle` is important on line 26

pmclanahan: I still think that the default state should be to redirect. Do you disagree?

pmclanahan: Oh... nm. You switched it. I got confused b/c the name is now odd... hmm... `tracking-protection-redirect == on` means don't redirect. I'm fine w/ that, but perhaps confusing. Sorry for the churn. Are you cool w/ that or want to rename?

PR #3470, opened 2015-10-23
Create a Dockerfile for running the Python functional tests

davehunt:
To build the Docker image:

```
cp docker/dockerfiles/bedrock_functional_tests Dockerfile
docker build -t bedrock_functional_tests .
rm Dockerfile
```

To run the tests in Docker containers using Selenium Grid:

```
docker run -d --name selenium-hub selenium/hub:2.48.2
docker run -d --link selenium-hub:hub selenium/node-firefox:2.48.2
docker run -v `pwd`/results:/app/results \
--link selenium-hub:hub \
-e SELENIUM_HOST=hub \
bedrock_functional_tests
```

For additional nodes run several `node-firefox` containers by repeating the command to run it.

To run the tests using Sauce Labs:

```
docker run -v `pwd`/results:/app/results \
-e DRIVER=SauceLabs \
-e SAUCELABS_USERNAME=[user] \
-e SAUCELABS_API_KEY=[key] \
-e BROWSER_NAME=Firefox \
-e PLATFORM="Windows 10" \
-e SCREEN_RESOLUTION=1280x1024 \
bedrock_functional_tests
```

Replace `[user]` and `[key]` with the values provided by Sauce Labs.

You will find the HTML report and JUnit results in the `results` folder on your local machine.

@glogiotatidis could you take a look at this and provide feedback? I'd likely need some help hooking this up in Jenkins.

### DIFF
--- a/bin/run-functional-tests.sh
+++ b/bin/run-functional-tests.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+set -xe
+
+# Defaults
+: ${PYTEST_PROCESSES:="1"}
+: ${BASE_URL:="https://www-dev.allizom.org"}
+: ${DRIVER:="Remote"}
+: ${SELENIUM_HOST:="localhost"}
+: ${SELENIUM_PORT:="4444"}
+: ${BROWSER_NAME:="firefox"}
+: ${TESTS_PATH:="tests/functional"}
+: ${RESULTS_PATH:="${TESTS_PATH}/results"}
+
+# Common arguments
+CMD="py.test"
+CMD="${CMD} -n ${PYTEST_PROCESSES}"
+CMD="${CMD} --base-url ${BASE_URL}"
+CMD="${CMD} --driver ${DRIVER}"
+CMD="${CMD} --html ${RESULTS_PATH}/index.html"
+CMD="${CMD} --junitxml ${RESULTS_PATH}/junit.xml"
+
+# Remote arguments
+if [ "${DRIVER}" = "Remote" ]; then
+  CMD="${CMD} --host ${SELENIUM_HOST}"
+  CMD="${CMD} --port ${SELENIUM_PORT}"
+  CMD="${CMD} --capability browserName \"${BROWSER_NAME}\""
+  if [ -n "${BROWSER_VERSION}" ]; then CMD="${CMD} --capability version \"${BROWSER_VERSION}\""; fi
+  if [ -n "${PLATFORM}" ]; then CMD="${CMD} --capability platform \"${PLATFORM}\""; fi
+fi
+
+# Sauce Labs arguments
+if [ "${DRIVER}" = "SauceLabs" ]; then
+  CMD="${CMD} --capability browserName \"${BROWSER_NAME}\""
+  if [ -n "${BROWSER_VERSION}" ]; then CMD="${CMD} --capability version \"${BROWSER_VERSION}\""; fi
+  if [ -n "${PLATFORM}" ]; then CMD="${CMD} --capability platform \"${PLATFORM}\""; fi
+  if [ -n "${SELENIUM_VERSION}" ]; then CMD="${CMD} --capability selenium-version \"${SELENIUM_VERSION}\""; fi
+  if [ -n "${BUILD_TAG}" ]; then CMD="${CMD} --capability build \"${BUILD_TAG}\""; fi
+  if [ -n "${SCREEN_RESOLUTION}" ]; then CMD="${CMD} --capability screenResolution \"${SCREEN_RESOLUTION}\""; fi
+  if [ -n "${PRIVACY}" ]; then CMD="${CMD} --capability public \"${PRIVACY}\""; fi
+fi
+
+if [ -n "${MARK_EXPRESSION}" ]; then CMD="${CMD} -m \"${MARK_EXPRESSION}\""; fi
+
+CMD="${CMD} ${TESTS_PATH}"
+eval ${CMD}
--- a/docker/dockerfiles/bedrock_functional_tests
+++ b/docker/dockerfiles/bedrock_functional_tests
@@ -1,0 +1,25 @@
+FROM python:2-slim
+
+WORKDIR /app
+
+# Run the tests
+CMD ["/app/bin/run-functional-tests.sh"]
+
+# Defaults
+ENV PYTEST_PROCESSES 5
+ENV MARK_EXPRESSION sanity
+ENV PRIVACY "public restricted"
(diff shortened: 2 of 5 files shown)


requirements/test_redirects.txt
```diff
@@ -0,0 +1,9 @@
+-r dev.txt
+-r test.txt
+
+# sha256: -WfKOb25jhYpmmnEWpRMXUNFOTYV7WRwux5iyjUGv0E
```
glogiotatidis: Both `braceexpand` and `pytest-django` are in `test.txt`

requirements/test_redirects.txt
```diff
@@ -0,0 +1,9 @@
+-r dev.txt
```
glogiotatidis: Do we need many packages from `dev.txt`? Does it make sense to add them in `test.txt` and don't include `dev.txt` here?

docker/dockerfiles/bedrock_functional_tests
```diff
@@ -0,0 +1,28 @@
...
+RUN pip install -r requirements/pip.txt
+
+# Install requirements
+COPY ./bin/peep.py /app/bin/peep.py
+RUN ./bin/peep.py install --no-cache-dir -r requirements/test.txt
+
+COPY . /app
+
+# Run the tests
+CMD ["/app/bin/run-functional-tests.sh"]
```
glogiotatidis: This is fine here. For an extra optimization step it can be moved after `WORKDIR`. It only defines the command to run on `docker run` and it doesn't have to be on the end of the file.

bin/run-functional-tests.sh
```diff
@@ -0,0 +1,41 @@
...
+  if [ -n "${BROWSER_VERSION}" ]; then CMD="${CMD} --capability version \"${BROWSER_VERSION}\""; fi
+  if [ -n "${PLATFORM}" ]; then CMD="${CMD} --capability platform \"${PLATFORM}\""; fi
+  if [ -n "${SELENIUM_VERSION}" ]; then CMD="${CMD} --capability selenium-version \"${SELENIUM_VERSION}\""; fi
+  if [ -n "${BUILD_TAG}" ]; then CMD="${CMD} --capability build \"${BUILD_TAG}\""; fi
+  if [ -n "${SCREEN_RESOLUTION}" ]; then CMD="${CMD} --capability screenResolution \"${SCREEN_RESOLUTION}\""; fi
+  if [ -n "${PRIVACY}" ]; then CMD="${CMD} --capability public \"${PRIVACY}\""; fi
+fi
+
+CMD="${CMD} ${TESTS_PATH}"
+echo "Running: ${CMD}" && eval ${CMD}
```
glogiotatidis: I like to use `set -xe` at the start of my bash scripts. `-x` prints the commands before executing them, similarly to what you do here with `Running: ${CMD}` and `-e` exits immediatelly after a failed cmd.

docker/dockerfiles/bedrock_functional_tests
```diff
@@ -0,0 +1,28 @@
+FROM python:2
+
+WORKDIR /app
+
+# Defaults
```
glogiotatidis: What do you think about deduplicating the defaults and keeping in the dockerfile only the docker related defaults?

docker/dockerfiles/bedrock_functional_tests
```diff
@@ -0,0 +1,28 @@
+FROM python:2
```
glogiotatidis: nit: `python:2-slim` is a lighter img

glogiotatidis:
Amazing work @davehunt ! I'm super happy that selenium provides it's own docker images and we can easily build upon that. 

We could possibly automate the start and stop of selenium + node + tests using docker-compose but it's probably good enough as is.

Re CI:

This is the new functional tests suite. Should we run this after unit tests step? Do we need locales installed for these tests to pass?

requirements/test_redirects.txt
davehunt: The redirect tests do, as they use `pytest-django` that at least requires `Django` itself. I didn't really want to add `dev.txt` as a requirement for the functional tests as they only currently run against an independently running instance of Bedrock.

bin/run-functional-tests.sh
davehunt: +1 thanks

docker/dockerfiles/bedrock_functional_tests
davehunt: nice

docker/dockerfiles/bedrock_functional_tests
davehunt: Yeah, I considered that myself. :)

docker/dockerfiles/bedrock_functional_tests
davehunt: Ah, I was thinking of it more like a script. :)

davehunt:
Thanks @glogiotatidis. I've addressed your review comments. Let me know if it looks good now and if so I'll go ahead and squash in preparation for merge.

> This is the new functional tests suite. Should we run this after unit tests step? Do we need locales installed for these tests to pass?

I think we should run this after the unit tests. Most of the tests expect the en-US locale (and explicitly target it) however there is at least one test that switches the current language on the home page. We could either include this but require l10n, or exclude this test by indicating a 'sanity' suite for the build pipeline. We could also mark any tests requiring l10n and exclude those easily enough.

/cc @alexgibson @jgmize @pmclanahan

alexgibson:
I'm going to open a PR shortly that will mark specific tests as sanity tests (we can discuss what to mark and what not to mark in that PR). I can make sure we leave out any tests that touch l10n or specific locales if that helps here.

requirements/test_redirects.txt
glogiotatidis: OK. I realized that this is not installed in docker and only used in travis, correct?

glogiotatidis:
> I'm going to open a PR shortly that will mark specific tests as sanity tests (we can discuss what to mark and what not to mark in that PR). I can make sure we leave out any tests that touch l10n or specific locales if that helps here.

Yes that would be helpful thank you. 

The reason is that we include l10n as the very last building step of bedrock.

glogiotatidis:
Nice work @davehunt :cake: r+

alexgibson:
Opened #3476 to add markers for sanity tests

glogiotatidis:
I'm available to create the new pipeline step when this and #3476 get merged.

requirements/test_redirects.txt
davehunt: That's correct, although I suspect at some point we may want to include the redirect tests.

davehunt:
I've added support for a `MARK_EXPRESSION` environment variable. In the build pipeline we can set this to `sanity` to restrict test collection to just the sanity tests.

davehunt:
Squashed commits ready for merge.

requirements/test_redirects.txt
pmclanahan: Yes! :+1:

glogiotatidis:
Thanks @davehunt :dizzy:

PR #3471, opened 2015-10-23
Add Python functional tests for home page

alexgibson:


alexgibson:
@davehunt r?

tests/pages/base.py
```diff
@@ -129,3 +133,64 @@ def click_sign_me_up(self):
         def sign_up_successful(self):
             el = self.selenium.find_element(*self._thank_you_locator)
             return el.is_displayed()
+
+    class MozillaNewsletterEmbedForm(PageRegion):
```
davehunt: As mentioned on IRC, I think we should have a base `NewsletterEmbedForm` which is then extended for additional forms. The locators, properties, and functions are mostly identical between these two page region objects.

tests/pages/home.py
```diff
@@ -2,9 +2,54 @@
...
+    @property
+    def is_next_event_displayed(self):
+        return self.root.find_element(*self._next_event_locator).is_displayed()
+
+    @property
+    def is_events_list_displayed(self):
+        return self.root.find_element(*self._events_list_locator).is_displayed()
+
+
+class Promos(PageRegion):
```
davehunt: When I suggested this I was imagining a `Promo` page region that would apply to each promo. I think what you have is simple enough to not even need this page region. I think you could simply have a `total_displayed_promos` property that returns the number of promos displayed on the page. Note that this could be different from the total number in the DOM.

tests/functional/test_home.py
```diff
@@ -0,0 +1,62 @@
...
+import pytest
+from selenium.common.exceptions import TimeoutException
+
+from ..pages.home import HomePage
+
+
+@pytest.mark.nondestructive
+def test_promos_are_present(base_url, selenium):
+    page = HomePage(base_url, selenium).open()
+    assert page.promos.length == 16
```
davehunt: `length` is confusing from a UI testing perspective. As mentioned in my previous comment I think a property on the `HomePage` object for total displayed promos would make sense. This would need to find all the promos and return the number of them that are displayed.

tests/functional/test_home.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@pytest.mark.nondestructive
+def test_tweet_is_present(base_url, selenium):
+    page = HomePage(base_url, selenium).open()
+    assert page.promos.tweet_is_present
```
davehunt: I think we should be checking if this is displayed rather than just present. Are there known circumstances when this will not be displayed? If so, it would be great to get a negative test - it will give confidence that we'll pick up a regression if either state changes.

tests/functional/test_home.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+@pytest.mark.nondestructive
+def test_download_button_is_displayed(base_url, selenium):
+    page = HomePage(base_url, selenium).open()
+    assert page.download_button_is_displayed
```
davehunt: Nit: Sometimes we're using `is_x_displayed` and sometimes we're using `x_is_displayed`. Eventually these inconsistencies are bound to creep in, I just thought I'd flag it as I noticed it.

tests/functional/test_home.py
```diff
@@ -0,0 +1,62 @@
...
+
+def test_mozilla_newsletter_successful_sign_up(base_url, selenium):
+    page = HomePage(base_url, selenium).open()
+    success = page.mozilla_newsletter.success_url_slug
+    page.mozilla_newsletter.expand_form()
+    page.mozilla_newsletter.type_email('noreply@mozilla.com')
+    page.mozilla_newsletter.select_country('United Kingdom')
+    page.mozilla_newsletter.accept_privacy_policy()
+    page.mozilla_newsletter.click_sign_me_up()
+    assert success in selenium.current_url
```
davehunt: I think we should abstract the URL slug into the page object entirely. Perhaps a property such as `sign_up_successful` that returns a boolean based on this slug being in the URL.

tests/functional/test_home.py
```diff
@@ -0,0 +1,62 @@
...
+    page = HomePage(base_url, selenium).open()
+    assert page.download_button_is_displayed
+
+
+@pytest.mark.nondestructive
+def test_upcoming_events_are_displayed(base_url, selenium):
+    page = HomePage(base_url, selenium).open()
+    events = page.upcoming_events
+    assert events.is_next_event_displayed
+    assert events.is_events_list_displayed
```
davehunt: Are there cases where these should not be displayed? It would be good to get some negative tests too.

davehunt:
Looking good @alexgibson. I wonder if we need to start marking sanity/smoke tests yet. Are all of these currently tests that you would want to run on every commit?

tests/pages/base.py
alexgibson: I started down this route initially but then got swayed on if it created too much complexity, but will give this a revisit. Agree there is a bit too much repetition currently :+1:

tests/functional/test_home.py
alexgibson: I'm trying to keep things relatively simple here, as the staggered fade in effect on the promos makes this a bit tricky for timing. I think for the errors we want to catch here, checking for existence in the DOM should be enough.

tests/functional/test_home.py
alexgibson: Only cases where they shouldn't be displayed is if the cron fails, so I think this should probably be enough.

alexgibson:
> Are all of these currently tests that you would want to run on every commit?

I think some of these probably don't need to be run on every commit (e.g. the event, tweet tests etc)

tests/pages/home.py
alexgibson: Sounds good :+1:

tests/functional/test_home.py
alexgibson: Will do.

tests/functional/test_home.py
davehunt: I see, in that case this is fine, but our page object should return a boolean. If necessary we could add an `is_element_present` to `page.py` that catches the element not found exception and returns an appropriate boolean.

alexgibson: ah, nice suggestion :+1:

tests/functional/test_home.py
alexgibson: Actually, there are some locales where the twitter promo isn't displayed. Perhaps we could add a negative test for a certain locale.

tests/functional/test_home.py
davehunt: I try to keep these sorts of additional to a minimum, but this will add value. We could also have an is_element_displayed when needed. It was always implicit waits that made these messy, and we're not using those. :smile:

tests/functional/test_home.py
alexgibson: I'm trying to add this method to the `MozillaNewsletterEmbedForm` PageRegion like so:

    @property
    def sign_up_successful(self):
        return self._success_url_slug in self.selenium.current_url

However when I try to assert this method:

    assert page.mozilla_newsletter.sign_up_successful

Selenium then [throws an error](https://gist.github.com/alexgibson/459cfec0a2c630d6a8ba), since the URL has changed and it can no longer find the PageRegion (`#mozorg-newsletter-form`). Any idea how I can work around this to accomplish what you suggest?

davehunt: If you make `mozilla_newsletter` the root locator then it's still present after the page reloads.

alexgibson:
Thanks for the review @davehunt - updated!

tests/functional/test_home.py
```diff
@@ -0,0 +1,68 @@
...
+@pytest.mark.nondestructive
+def test_promos_are_present(base_url, selenium):
+    page = HomePage(base_url, selenium).open()
+    assert page.expected_promos == 16
+
+
+@pytest.mark.nondestructive
+def test_tweet_is_present(base_url, selenium):
+    page = HomePage(base_url, selenium).open()
+    assert page.is_tweet_promo_present
```
davehunt: Did you say there are cases where this is not displayed? I noticed it wasn't there for en-GB. If that's expected then we can instantiate the page object with a specific locale for a negative test.

tests/pages/newsletter_embed.py
```diff
@@ -0,0 +1,107 @@
...
+
+    @property
+    def sign_up_successful(self):
+        el = self.selenium.find_element(*self._thank_you_locator)
+        return el.is_displayed()
+
+
+class MozillaNewsletterEmbedForm(NewsletterEmbedForm):
+
+    _root_locator = (By.ID, 'mozorg-newsletter-form')
```
davehunt: Did you not need to modify this to the parent element, to allow verification of the submission being successful?

tests/pages/page.py
```diff
@@ -29,6 +29,12 @@ def url(self):
     def wait_for_page_to_load(self):
         return self
 
+    def is_element_present(self, locator):
+        try:
+            return self.selenium.find_element(*locator)
+        except:
```
davehunt: We should catch the specific exception `NoSuchElementException` from `selenium.common.exceptions`.

tests/pages/base.py
```diff
@@ -3,11 +3,11 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from selenium.webdriver.common.by import By
-from selenium.webdriver.support import expected_conditions as expected
 from selenium.webdriver.support.select import Select
 from selenium.webdriver.support.ui import WebDriverWait as Wait
 
 from .page import Page, PageRegion
+from .newsletter_embed import NewsletterEmbedForm, MozillaNewsletterEmbedForm
```
davehunt: In the past we've use a `regions` module for locating page regions. Otherwise these may appear to be full page objects.

tests/pages/home.py
```diff
@@ -2,9 +2,54 @@
...
+
+    @property
+    def is_promo_grid_displayed(self):
+        Wait(self.selenium, self.timeout).until(
+            expected.visibility_of_element_located(self._promo_grid_locator))
+        return self.selenium.find_element(*self._promo_grid_locator).is_displayed()
+
+    @property
+    def expected_promos(self):
+        return len(self.selenium.find_elements(*self._promo_tile_locator))
```
davehunt: Just a caution, but you're returning all promos here regardless of whether they're displayed. I know you've mentioned that you're concerned about checking for visibility of certain promos. I might suggest at least renaming the property to `number_of_promos_present` for a little bit of extra clarity.

tests/pages/newsletter_embed.py
alexgibson: Not sure I follow, when the page reloads it's a completely different site, so changing the parent element shouldn't make a difference? I managed to fix the test by using `assert newsletter.sign_up_successful` instead of `assert page.newsletter.sign_up_successful`, which is what I thought you we're meaning?

tests/functional/test_home.py
alexgibson: Forgot to add this one, good catch :+1:

tests/pages/newsletter_embed.py
davehunt: Ah, I misunderstood your original issue, and it didn't help that the behaviour appears to be different for en-GB. I'm not sure I like the page region being applied for a completely different site, but as it's just the URL I think it probably makes the most sense here.

tests/pages/newsletter_embed.py
```diff
@@ -0,0 +1,107 @@
...
+    _country_locator = (By.ID, 'country')
+    _success_url_slug = 'sign-up-for-mozilla'
+
+    def click_sign_me_up(self):
+        self.root.find_element(*self._submit_button_locator).click()
+        Wait(self.selenium, self.timeout).until(
+            lambda s: self._success_url_slug in s.current_url)
```
davehunt: We could now change this to: `Wait(self.selenium, self.timeout).until(lambda s: self.sign_up_successful)`

alexgibson:
@davehunt updated thanks

tests/functional/test_home.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+@pytest.mark.nondestructive
+def test_tweet_is_not_present(base_url, selenium):
+    page = HomePage(base_url, selenium, 'de').open()
+    assert not page.is_tweet_promo_present
```
davehunt: :+1:

tests/pages/regions.py
```diff
@@ -0,0 +1,106 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
davehunt: Certainly good enough for now, however as this increases we may need a `regions` directory with suitable file names for the various regions. This will be useful for discovery.

davehunt:
r+

PR #3472, opened 2015-10-25
Patch 1

Klofutar:


kewisch:
Can you squash these commits into one? Aside from that r+ from the calendar perspective.

Klofutar:
How do I do that? And what does this "r+" mean?

craigcook:
Hi Klofutar, welcome to bedrock!

Squashing commits in git merges multiple commits into a single, new commit, so the code history is a bit tidier. <a href="http://davidwalsh.name/squash-commits-git">Here's a pretty basic introduction</a> and <a href="http://www.git-scm.com/book/en/v2/Git-Tools-Rewriting-History">here's a more detailed guide</a>. It entails doing an "interactive rebase" to modify your local git history and then force-pushing your changes to overwrite the remote history. It can be a little scary at first so if you're not sure about trying to squash on your own we can do it for you when we merge these changes.

The "r" business is code review shorthand, where "r?" means "review requested", "r+" means it has passed code review and been accepted, and "r-" means it has failed review and been rejected. You may also sometimes see "r+wc" which means "r+ with comments" for when a code change has overall passed review but there were a few comments to be addressed. Once those comments are addressed the r+ is implied without further review.

Klofutar:
Filled a bug on Bugzilla instead.  Thanks for the help, though.

PR #3473, opened 2015-10-26
Fix bug 1218447 - Update Slovenian holiday calendar

craigcook:


craigcook:
@kewisch Mind giving this a look when you have a chance? I just took it directly from #3472 which you had previously r+ed, but would appreciate second look to make sure I haven't missed something.

PR #3474, opened 2015-10-26
Bug 1210864 - Add Private Browsing to Firefox Family subnav

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1210864

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -469,9 +469,11 @@
           <a data-id="desktop" href="{{ url('firefox.desktop.index') }}" class="primary-link"><div>{{ _('<span>for</span> Desktop') }}</div></a>
 
           <ul class="subnav" id="desktop-subnav">
-            <li{% if activesub in ('index', 'trust', 'customize', 'fast') %} class="hassubsubnav"{% endif %}>
+            {# note: Removing hassubsubnav for now because after localization nav items get pushed to two lines #}
+            <li {# {% if activesub in ('index', 'trust', 'customize', 'fast') %} class="hassubsubnav"{% endif %} #}>
               <a href="{{ url('firefox.desktop.index') }}"{% if activesub == 'index' %} class="selected"{% endif %}><div>{{ _('Overview') }}</div></a>
 
+            {# note: Removing items for now because after localization nav items get pushed to two lines
               {% if activesub in ('index', 'trust', 'customize', 'fast') %}
               <ul class="subsubnav" id="desktop-subsubnav">
                 <li><a href="{{ url('firefox.desktop.trust') }}"{% if activesub == 'trust' %} class="selected"{% endif %}>{{ _('Trusted') }}</a></li>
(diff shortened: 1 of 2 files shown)


craigcook:
Should add the link to the iOS subnav as well.

malena:
@craigcook Oh I thought that's what you would do? I am confusing what  you were going to add the menu.

craigcook:
Sorry, I added the iOS top link and subnav previously but couldn't add Private Browsing to the iOS subnav until the page was ready. I'm happy to add it separately once this merges but it doesn't have to wait, it can go into the same commit if you're up for it.

malena:
@craigcook Ahh okay no problemo!

malena:
r?

malena:
Includes a waffle switch. r?

bedrock/base/templates/macros.html
```diff
@@ -482,6 +482,9 @@
             </li>
             <li><a href="{{ url('firefox.sync') }}"{% if activesub == 'sync' %} class="selected"{% endif %}><div>{{ _('Sync') }}</div></a></li>
             <li><a href="{{ url('firefox.hello') }}"{% if activesub == 'hello' %} class="selected"{% endif %}><div>{{ _('Hello') }}</div></a></li>
+            {% if l10n_has_tag('tracking_protection') and waffle.switch('tracking-protection') %}
```
flodolo: Remove the l10n tag check, only waffle is needed. We can't use tags on a shared thing like navigation.

jpetto:
We knew it would eventually come to this - the subnav is now too long in some locales. Here's a shot of es-ES at tablet viewport:

screen shot 2015-10-29 at 3 50 16 pm

Unless there are other tenable ideas, I think our best option is to drop the sub-sub nav (Trusted/Flexible/Fast) for the desktop pages. Will need to get sign-off from jbertsch & habber.

jbertsch:
https://bugzilla.mozilla.org/show_bug.cgi?id=1210864#c4

media/css/firefox/family-nav.less
```diff
@@ -281,7 +281,7 @@
                 opacity: 0.8;
                 height: 77px;
                 line-height: 77px;
-                max-width: 125px; // uhhh, seems to work?
+                //max-width: 145px;
```
jpetto: I think we should keep this restriction in place, and should set it to `129px`. At tablet width, the `#desktop-subnav` is 556px wide, which, divided by 4 (the max elements now possible in the subnav) is 139. Each `<a>` has a left and right padding of 5px, hence 129.

If this restriction isn't in place, text will not wrap to two lines, which means an `<li>` could get pushed underneath and out of view for some locales.

bedrock/base/templates/macros.html
```diff
@@ -479,9 +481,13 @@
                 <li><a href="{{ url('firefox.desktop.fast') }}"{% if activesub == 'fast' %} class="selected"{% endif %}>{{ _('Fast') }}</a></li>
               </ul>
               {% endif %}
+            #} 
             </li>
             <li><a href="{{ url('firefox.sync') }}"{% if activesub == 'sync' %} class="selected"{% endif %}><div>{{ _('Sync') }}</div></a></li>
             <li><a href="{{ url('firefox.hello') }}"{% if activesub == 'hello' %} class="selected"{% endif %}><div>{{ _('Hello') }}</div></a></li>
+            {% if waffle.switch('tracking-protection') %}
+                <li><a href="{{ url('firefox.private-browsing') }}"{% if activesub == 'private-browsing' %} class="selected"{% endif %}><div>{{ _('Private Browsing') }}</div></a></li>
```
jpetto: Super small nit - each of these has one extra tab stop (should be 2 spaces, but is 4).

jpetto:
One small CSS issue, but otherwise looking good! r+wc

malena:
@jpetto updated and squashed commits.

jpetto:
Gah, I'm sorry to ask this late, but can we make a special CSS rule for `en-*` only to make sure "Private Browsing" fits on one line? I know this will be requested later if we don't do it now. (I thought I tested this during my last review, but apparently I was in tablet mode.)

I think we can just add a bit of CSS to line 428:

```
html[lang^='en'] .fxfamilynav {
    .subnav > li > a {
        max-width: 149px;
    }

    .subsubnav a {
        text-transform: uppercase;
    }
}
```

malena:
@jpetto no problemo updated!

jpetto:
Looks great! Just waiting for Travis... :hourglass:

PR #3475, opened 2015-10-26
Add directory for images relating to Open Graph

ednapiranha:
This is a directory for supporting Open Graph and related social sharing images that are specific to Mozilla-related projects. First image is the DevEdition logo.


pmclanahan:
A PR from @ednapiranha! Just like olde timey Flux :D :wave:

PR #3478, opened 2015-10-27
Fix Bug 1217896, wrap GA in DNT conditional

schalkneethling:
@alexgibson @jpetto r?


schalkneethling:
Oh, forgot to mention. To test this locally, you need to fill this[1] with a dummy id or else, the GA code does not load anyway locally.

1. https://github.com/mozilla/bedrock/blob/master/bedrock/settings/local.py-dist#L14

schalkneethling:
Initial tests by cmore has been successful. Next is for it to go through a round of QA.

media/js/base/dnt-helper.js
```diff
@@ -0,0 +1,46 @@
...
+
+    // only set platform if this is IE
+    if (isIE) {
+        platform = platformRegEx.exec(ua).toString();
+    }
+
+    if (fxMatch && parseInt(fxMatch[1], 10) < 32) {
+        // Can't say for sure if it is 1 or 0, due to Fx bug 887703
+        dntStatus = 'Unspecified';
+    } else if (isIE && $.inArray(platform, anomalousWinVersions) !== -1) {
```
alexgibson: We can't use jQuery here since this is in the `<head>`. This is currently throwing an undefined error for all IE versions. Note we can't just use `indexOf()` here either, since this only works for strings in older IE versions :cry:

media/js/base/dnt-helper.js
```diff
@@ -0,0 +1,46 @@
+/**
+ * Returns true or false based on whether doNotTack is enabled. It also takes into account the
+ * anomalies, such as !bugzilla 887703, which effect versions of Fx 31 and lower. It also handles
+ * IE versions on Windows 7, 8 and 8.1, where the DNT implementation does not honor the spec.
+ * @see https://bugzilla.mozilla.org/show_bug.cgi?id=1217896 for more details
+ * @params {string} [dnt] - An optional mock doNotTrack string to ease unit testing.
+ * @params {string} [ua] - An optional mock userAgent string to ease unit testing.
+ * @returns {boolean} true if enabled else false
+ */
+function dntEnabled(dnt, ua) {
```
alexgibson: Since this is in global scope, worth prefixing with an underscore `_dntEnabled`?

media/js/base/dnt-helper.js
schalkneethling: Oh wow, interesting. I did not see those errors. But yeah, will sort this out in vanilla JS

media/js/base/dnt-helper.js
schalkneethling: Sure thing, makes sense

media/js/base/dnt-helper.js
```diff
@@ -0,0 +1,56 @@
...
+    // http://www.regexr.com/3c2el
+    var platformRegEx = /Windows.+?(?=;)/g;
+    var ieRegEx = /MSIE|Trident/i;
+    var isIE = ieRegEx.test(ua);
+    var oldIE = false;
+    var platform = '';
+
+    if (isIE) {
+        // in IE 8 and below, indexOf on the Array object is undefined
+        oldIE = typeof Array.prototype.indexOf !== 'function';
```
alexgibson: Nit: instead of using a variable such as `oldIE` like this, could we just do this instead?

    // with old version of IE, DNT did not exist so we simply return false
    if (typeof Array.prototype.indexOf !== 'function') {
        return false;
    }

media/js/base/dnt-helper.js
```diff
@@ -0,0 +1,56 @@
...
+        // We are only concerned with the platform if this is IE
+        platform = platformRegEx.exec(ua).toString();
```
alexgibson: If there's no match here this will throw a type error?

media/js/base/dnt-helper.js
```diff
@@ -0,0 +1,53 @@
...
+    // Matches from Windows up to the first occurance of ; un-greedily
+    // http://www.regexr.com/3c2el
+    var platformRegEx = /Windows.+?(?=;)/g;
+    var ieRegEx = /MSIE|Trident/i;
+    var isIE = ieRegEx.test(ua);
+    var platform = '';
+
+    if (isIE) {
+        // We are only concerned with the platform if this is IE
+        platform = platformRegEx.exec(ua).toString();
```
alexgibson: This will still throw an exception if there's no match, as you would be calling `toString()` on `null`

media/js/base/dnt-helper.js
```diff
@@ -0,0 +1,53 @@
...
+    var ieRegEx = /MSIE|Trident/i;
+    var isIE = ieRegEx.test(ua);
+    var platform = '';
+
+    if (isIE) {
+        // We are only concerned with the platform if this is IE
+        platform = platformRegEx.exec(ua);
+        // With old versions of IE, DNT did not exist so we simply return false;
+        if (typeof Array.prototype.indexOf !== 'function') {
+            return false;
```
alexgibson: Would it be worthwhile moving this conditional return to the same block as the rest of the statements at the end of the function? Might make things slightly simpler to follow? e.g.

    if (isIE && typeof Array.prototype.indexOf !== 'function') {
        return false;
    else if {
        //etc...
    }

schalkneethling: The initial idea I had in mind was to return as quickly as possible if it is oldIE. With that said, as it stands now, if it is the first `if` of the conditional block, it will work just as well I reckon. Will update the code.

media/js/base/dnt-helper.js
```diff
@@ -0,0 +1,47 @@
...
+    // Matches from Windows up to the first occurance of ; un-greedily
+    // http://www.regexr.com/3c2el
+    var platformRegEx = /Windows.+?(?=;)/g;
+    var ieRegEx = /MSIE|Trident/i;
+    var isIE = ieRegEx.test(ua);
+    var platform = platformRegEx.exec(ua);
+
+    // With old versions of IE, DNT did not exist so we simply return false;
+    if (isIE && typeof Array.prototype.indexOf !== 'function') {
+        return false;
```
alexgibson: I guess for consistency we could now also do `dntStatus = 'Unspecified';` here?

media/js/base/dnt-helper.js
```diff
@@ -0,0 +1,47 @@
...
+    } else if (fxMatch && parseInt(fxMatch[1], 10) < 32) {
+        // Can't say for sure if it is 1 or 0, due to Fx bug 887703
+        dntStatus = 'Unspecified';
+    } else if (isIE && platform && anomalousWinVersions.indexOf(platform.toString()) !== -1) {
```
alexgibson: Is there a reason to turn the array into a string here? Sorry if I'm misunderstanding something

alexgibson: Oh, never mind sorry I misread

media/js/base/dnt-helper.js
schalkneethling: Well, then we need the other conditionals evaluated before we return. As we know we are going to return false, we might as well return here. Thoughts?

alexgibson: I'm not sure how much of an optimization this is really, seems very small? I was more thinking of consistency with the other conditionals, but will leave this up to you :+1:

media/js/base/dnt-helper.js
```diff
@@ -0,0 +1,47 @@
...
+    // List of Windows versions known to not implement DNT according to the standard.
+    var anomalousWinVersions = ['Windows NT 6.1', 'Windows NT 6.2', 'Windows NT 6.3'];
+
+    var fxMatch = ua.match(/Firefox\/(\d+)/);
+    // Matches from Windows up to the first occurance of ; un-greedily
+    // http://www.regexr.com/3c2el
+    var platformRegEx = /Windows.+?(?=;)/g;
+    var ieRegEx = /MSIE|Trident/i;
+    var isIE = ieRegEx.test(ua);
+    var platform = platformRegEx.exec(ua);
```
alexgibson: As we discussed on IRC, suggest changing `exec` to `match`?

alexgibson:
From the [latest comment in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1217896#c25) it sounds like there's no real QA happening for this change currently. @schalkneethling - could you liaise with rbillings on WebQA to make sure this gets coverage? It would be great to make sure the final code is also on the demo for when it gets reviewed.

alexgibson:
Code-wise this looks good to me, r+

Let's wait on QA doing some testing, but please squash these commits so we're ready to merge when needed.

PR #3480, opened 2015-10-27
[fix bug 1197395] Add mobile promo to /firefox/new/ scene 2.

jpetto:
I didn't re-add the [download trigger timeout](https://github.com/mozilla/bedrock/blob/c252326ef602a1c8675c1434830460fe3ab12032/media/js/firefox/new.js#L322-L330) for two reasons:

1. I couldn't reproduce the issue where the new scene 2 image would stop partially downloaded. I tried with and without throttling network to Edge speeds, on IE 7, 8, and modern browsers, clicking the download button on scene 1 and the download button on the home page (which goes directly to scene 2).
2. The new scene 2 image isn't functional (as the old install images were). *If* the image were to stop midway through download, the only cost to the user would be an incomplete promotional image. I'm thinking this would be acceptable to increase download responsiveness by 1-1.5 seconds.


jpetto:
Based on conversation in the [design bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1208148#c11), the image used here is likely going to change.

Also, to not put more strings in front of localizers during the fall campaign push, should not be merged until after 11/3.

The above aside, code is ready for review.

jpetto:
Image has been updated to be more multi-locale friendly. Ready for review and merge.

jpetto:
Note - I changed the `download-content` class to `scene-content` because the download buttons uses `download-content`, which is often used in Optimizely to target download links.

bedrock/firefox/templates/firefox/new.html
```diff
@@ -150,9 +158,36 @@ <h1 class="large" id="main-headline">
...
+
+      <div id="thankyou" tabindex="0">
+        {{ _('While it installs, add it to your Android and iOS devices.') }}
+      </div>
 
-    <div id="download-started" class="download-content hidden">
+      <div class="mobile-download-buttons">
+        <ul>
+          <li class="android">
+            <a rel="external" href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}">
```
schalkneethling: Not introduced by this PR but, just noticed that this link reads:
`https://play.google.com/store/apps/details?id=org.mozilla.firefox&referrer=utm_source%3Dmozilla%26utm_medium%3DReferral%26utm_campaign%3Dmozilla-org`

Note the `referrer=utm_source....`

Should it not be `referrer=&amp;utm_source...`

bedrock/firefox/templates/firefox/new.html
```diff
@@ -150,9 +158,36 @@ <h1 class="large" id="main-headline">
...
+              {{ high_res_img('firefox/android/btn-google-play.png', {'alt': _('Get it on Google Play'), 'width': '129', 'height': '45', 'l10n': True}) }}
+            </a>
+          </li>
+          <li class="ios">
+            <a href="{{ settings.APPLE_APPSTORE_FIREFOX_LINK }}&amp;ct=mozorg-scene_2-appstore-button">
+              <img src="{{ l10n_img('firefox/ios/btn-app-store.svg') }}" alt="{{ _('Download on the App Store') }}" width="152" height="45">
+            </a>
+          </li>
+        </ul>
+      </div>
```
schalkneethling: So here is a fun fact about the above:
The scene switches once download is clicked in all browsers but:

1) IE6 - There are no clickable areas nor badges/buttons
2) IE7/8 - The Android badge is shown but not the iOS one (SVG)

bedrock/firefox/templates/firefox/new.html
jpetto: Did a bit of research here, and it is formatted as [requested in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1099429#c0). Will add a comment in `base.py` to clarify for the next person who thinks that link looks funky. :grin:

jpetto:
Fixed the IE6 display issue.

As for the lack of SVG support in IE 6 - 8, I'm inclined to just let the `alt` text perform its graceful (and localized) degradation for this secondary CTA. Apple provides pre-localized App Store buttons, but only in SVG format (as far as I can tell). The way I see it, our options are:

1. Leave it alone and let the `alt` text do its thing for non-SVG browsers
2. Create localized App Store buttons in PNG format.
3. Show an HTML button with black background and white text in place of the SVG for IE 6 - 8.

I'd say option 2 is too much effort for ancient browsers. Option 3 wouldn't take a a *ton* of time, but would add complexity and dev time, again for ancient browsers.

If this weren't the `/new` page, I'd say option 1 for sure. As it is, I could be talked in to option 1 or 3.

Thoughts?

bedrock/firefox/templates/firefox/new.html
schalkneethling: :+1:

schalkneethling:
@jpetto Indeed, the only reason I even tested IE6/7 was because this is `/new`. I guess it is not entirely my/our decision but, I am leaning towards option 1. You want to fly this by someone or just go with option 1?

jpetto:
Agreed we should get sign-off on option 1. I'll needinfo jbertsch in the bug and report back here.

schalkneethling:
@jpetto signoff received here https://bugzilla.mozilla.org/show_bug.cgi?id=1197395#c39 - care to squash

jpetto:
Squashed!

Mostly for appearances, habber and cmore were leaning towards creating an HTML button fallback. I don't disagree with them, but I agree more with you and jbertsch - let's not add more code to the page for ancient browsers.

schalkneethling:
r+

PR #3481, opened 2015-10-28
Bug 1139654 - use sentence case for iOS home promo button

craigcook:
Apple's brand guidelines frown upon the uppercase I so we'll override the `text-transform` for just this promo.


schalkneethling:
r+ :hamburger:

PR #3482, opened 2015-10-28
Fix Bug 1218491, developer edition promo additions

schalkneethling:
@alexgibson r?


schalkneethling:
Waiting on a couple of clarifying questions but, those are only related to images and whether this is en-US only or en-* only so, good for general code review.

bedrock/firefox/templates/firefox/developer.html
```diff
@@ -93,6 +93,28 @@
 <section class="feature-overview">
   <div class="container">
     <ul class="features">
+    {% if LANG.startswith('en-') %}
```
flodolo: These should be behind a tag, not en-US only. We'll want to localize them at some point after Nov 3rd.

alexgibson:
This only updates the `/firefox/developer` feature page, yet the bug mentions /whatsnew and /firstrun also?

bedrock/firefox/templates/firefox/developer.html
alexgibson: Should this also be behind a waffle switch?

schalkneethling: @flodolo The bug mentions this is en-US only. I asked about it in the bug because I thought perhaps it could be en-* so, still waiting on that. If it will be translated later on, then I reckon using a tag would be better.

@alexgibson I reckon that would be a good idea yes.

flodolo: The page is fully localized, preventing our localizers to work on it after Nov 3rd seems wrong, unless the blurbs are disappearing after a few days.

ckprice:
I think we can do en-*.

Alex is correct - this bug entails adding the blurbs to both [/whatsnew](https://www.mozilla.org/en-US/firefox/38.0a2/whatsnew/) and [/firstrun](https://www.mozilla.org/en-US/firefox/38.0a2/tour/) as well as the product page.

Basically these two blurbs should take the top two places.

schalkneethling:
@ckprice @flodolo So, should this be:

```
{% if l10n_has_tag('firefox_dev_fall') and LANG.startswith('en-') and waffle.switch('firefox-dev-fall') %}
```
or will the current implementation be sufficient? As this page is localized, perhaps the current impl will be fine.

@ckprice So, I also updated the other pages but, whatsnew and firstrun only has 2 blurbs. I assume I just replace them with the new two as I have done here?

flodolo:
Checking for tag and en-* doesn't make sense: if it's en-US, you don't need tags, those are for localization.

I still don't understand why we're making this en-US only. The landing page is localized, we should have content parity after Nov 3rd, so it should be behind a tag.

schalkneethling:
@flodolo I though as much. @ckprice is the intention for these blurbs to never be translated?

ckprice:
They can eventually be translated. Sorry, I noted en-us only to indicate
that translation wasn't required for a Nov 3 launch

On Thursday, October 29, 2015, Schalk Neethling <notifications@github.com>
wrote:

> @flodolo <https://github.com/flodolo> I though as much. @ckprice
> <https://github.com/ckprice> is the intention for these blurbs to never
> be translated?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/mozilla/bedrock/pull/3482#issuecomment-152118432>.
>


-- 
Cory Price
/ckprice

bedrock/firefox/templates/firefox/dev-whatsnew.html
```diff
@@ -43,7 +39,24 @@
       <p>{{_('Get to know the features that make it the most complete browser for building the Web.')}}</p>
     </header>
     <ul class="features">
-    {% if l10n_has_tag('spring_release') %}
+    {% if l10n_has_tag('firefox_dev_fall') and waffle.switch('firefox-dev-fall') %}
+      <li class="feature">
+        <a href="https://hacks.mozilla.org/2015/11/developer-edition-44-creative-tools-and-more" rel="external">
```
alexgibson: This link is currently 404. That's probably intentional as the article is not ready, but just commenting for record.

bedrock/firefox/templates/firefox/dev-firstrun.html
```diff
@@ -70,14 +58,31 @@
 
 {% block content %}
 
-<main role="main" {% if l10n_has_tag('new_doorhanger') %}class="sync-reminder"{% endif %} >
```
alexgibson: If we're removing this l10n tag, I believe you can remove the class here as well. It was used as a conditional logic check in the page's JS here: https://github.com/mozilla/bedrock/blob/master/media/js/firefox/dev-firstrun.js#L279-L280

If we update this JS and always execute the code in the conditional there, then this class is no longer needed. Sorry if this seems confusing (it was a temporary hack to avoid having to wait on l10n).

bedrock/firefox/templates/firefox/dev-firstrun.html
```diff
@@ -70,14 +58,31 @@
...
+<main role="main" class="sync-reminder">
```
alexgibson: If we're removing this l10n tag, I believe you can remove the class here as well. It was used as a conditional logic check in the page's JS here: https://github.com/mozilla/bedrock/blob/master/media/js/firefox/dev-firstrun.js#L279-L280

If we update this JS and always execute the code in the conditional there, then this class is no longer needed. Sorry if this seems confusing (it was a temporary hack to avoid having to wait on l10n).

bedrock/firefox/templates/firefox/dev-firstrun.html
```diff
@@ -104,7 +108,7 @@
           <div class="video" data-video-id="eQqNfkqIJdw"></div>
         </div>
       </li>
-    {% if not l10n_has_tag('spring_release') %}
+    {% elif not l10n_has_tag('spring_release') %}
```
alexgibson: Could this just be an `{% else %}` now? These tags are confusing

bedrock/firefox/templates/firefox/dev-firstrun.html
```diff
@@ -116,17 +120,26 @@
           <div class="video" data-video-id="R_qDaLQ8ghg"></div>
         </div>
       </li>
+      <li class="feature">
```
alexgibson: We're repeating the same `Page Inspector` block twice in the template here. Could we avoid the repetition by only wrapping the l10n tag around the first item which changes, instead of repeating this here?

alexgibson: This is also within the old l10n block, not sure why this is all here?

ckprice:
Regarding [/firstrun](https://www-demo4.allizom.org/en-US/firefox/44.0a2/tour/)

Looks like `Performance Tools` and `Page Inspector` have dropped off (ref [production](https://www.mozilla.org/en-US/firefox/44.0a2/tour/)). Could we add these to the first two positions under `Features and Tools`?

alexgibson:
As discussed with @flodolo  on IRC we can simplify the firstrun template like so: https://irccloud.mozilla.com/pastebin/rd0lPoio

schalkneethling:
Ok so adding the page inspector the the `Features and tools` section means we will be duplicating it again in the HTML. One option is that we remove it from the `featured` section as it now be in the above mentioned section.

Regarding the `performance tools`, it will cause the same duplication so again, perhaps we can remove it from the `featured` section. With that said, the `featured` section will then only have one fallback, which is the `Web Audio Editor` which is behind a `spring_release l10n` tag so, perhaps the duplication here is unavoidable.

Thoughts @ckprice @flodolo ?

flodolo:
We have 4 locales that are enabled but missing the `spring_release` tag: gd, id, oc, uz.

For sake of simplifying things we could decide to disable the page for them (it means they haven't updated the page in 6 months), and get rid of that tag too. Otherwise, we need to duplicate things.

schalkneethling:
Thanks @flodolo, will also await Cory's reply.

bedrock/firefox/templates/firefox/dev-firstrun.html
```diff
@@ -147,7 +137,31 @@
             <div class="video" data-video-id="R_qDaLQ8ghg"></div>
           </div>
         </li>
-      {% endif %}
+        <li class="feature">
+          <a href="https://developer.mozilla.org/docs/Tools/Performance" rel="external">
+            <img src="{{ static('img/firefox/firstrun/dev/performance-tools.png') }}" alt="{{ _('Screenshot') }}" class="screenshot">
+          </a>
+          <h2>{{_('Performance Tools')}}</h2>
```
schalkneethling: This one is now duplicated but, I do not think we can get around it unless, it is ok to have only 1 featured item at the top of the page for locales that does not have `firefox_dev_fall` enabled

schalkneethling:
Demo4 updated with the latest changes.

bedrock/firefox/templates/firefox/developer.html
```diff
@@ -57,14 +52,12 @@
...
 
-  {% if l10n_has_tag('new_doorhanger') %}
   <p class="feedback-note">
     {{ _('Firefox Developer Edition automatically sends feedback to Mozilla.') }}
     <a href="{{ url('privacy.notices.firefox') }}#telemetry" class="more">{{ _('Learn more') }}</a>
   </p>
-  {% endif %}
 
-  <img src="{{ static('img/firefox/developer/screenshot.jpg') }}" alt="{{ _('Firefox Developer Edition') }}" class="screenshot" />
+  {{ high_res_img('firefox/developer/challenger-hero.png', {'alt': 'Firefox Developer Edition', 'width': '832', 'height': 'auto', 'class': 'screenshot'}) }}
```
alexgibson: Should this image also be behind the l10n / waffle flag as a conditional, otherwise some locales might see the screenshot image but not have the challenge listed in the links below?

alexgibson:
@flodolo - can you confirm if we should still be keeping `l10n_has_tag('spring_release')` in the /developer and /whatsnew templates. It looks removed from /firstrun but not the others.

alexgibson:
@schalkneethling - as confirmed with @flodolo on IRC. If we are removing the `l10n_has_tag('spring_release')` then it should be done for all templates in this change.

bedrock/firefox/templates/firefox/developer.html
schalkneethling: Doh! Of course it does. thx @alexgibson

bedrock/firefox/templates/firefox/developer.html
```diff
@@ -48,23 +39,19 @@
...
-  {% if l10n_has_tag('new_doorhanger') %}
   <p class="feedback-note">
     {{ _('Firefox Developer Edition automatically sends feedback to Mozilla.') }}
     <a href="{{ url('privacy.notices.firefox') }}#telemetry" class="more">{{ _('Learn more') }}</a>
   </p>
-  {% endif %}
 
-  <img src="{{ static('img/firefox/developer/screenshot.jpg') }}" alt="{{ _('Firefox Developer Edition') }}" class="screenshot" />
+  {% if l10n_has_tag('firefox_dev_fall') and waffle.switch('firefox-dev-fall') %}
+    {{ high_res_img('firefox/developer/challenger-hero.png', {'alt': 'Firefox Developer Edition', 'width': '832', 'height': 'auto', 'class': 'screenshot'}) }}
```
alexgibson: Not sure if this alt string should be localized, but it was previously so best wrap it here too?

schalkneethling:
The only possible problem I see at the moment, is that the `whatsnew` page' fallback could show either 3 or 4 featured items. This is different from the default of 2. Displaying 4 will probably be fine but, three might not look ideal.

bedrock/firefox/templates/firefox/developer.html
```diff
@@ -93,7 +80,28 @@
 <section class="feature-overview">
   <div class="container">
     <ul class="features">
-    {% if l10n_has_tag('spring_release') %}
+    {% if l10n_has_tag('firefox_dev_fall') and waffle.switch('firefox-dev-fall') %}
+      <li class="feature">
+        <a href="https://hacks.mozilla.org/2015/11/developer-edition-44-creative-tools-and-more" rel="external">
+          {{ high_res_img('firefox/developer/creative-tools.png', {'alt': 'Screenshot', 'width': '218', 'height': '118', 'class': 'screenshot'}) }}
```
alexgibson: Same here `Screenshot` should be wrapped for l10n.

bedrock/firefox/templates/firefox/developer.html
```diff
@@ -93,7 +80,28 @@
...
+        </a>
+        <h2>{{ _('Visual Editing Tools') }}</h2>
+        <p>
+          {{ _('Visually edit CSS and HTML animations playing on the page from the Inspector panel. Verify alignment, dimensions and padding of any element using the Ruler and Measurement tools.') }}
+        </p>
+        <a href="https://hacks.mozilla.org/2015/11/developer-edition-44-creative-tools-and-more" class="more" rel="external">{{ _('Learn more about Visual Editing Tools') }}</a>
+      </li>
+      <li class="feature">
+        <a href="http://devtoolschallenger.com/" rel="external">
+          {{ high_res_img('firefox/developer/devtools-challenger.png', {'alt': 'Screenshot', 'width': '218', 'height': '118', 'class': 'screenshot'}) }}
```
alexgibson: Same here `Screenshot`

bedrock/firefox/templates/firefox/developer.html
schalkneethling: Agreed. I did have it wrapped, not sure why I changed it :confused:

alexgibson:
Layout is broken on whatsnew when the new l10n tag is false

schalkneethling:
@alexgibson yup, that is what I was afraid off

schalkneethling:
@alexgibson How does one test that? I know how to enable and disable waffle switches but not sure about l10n tags.

alexgibson:
On firstrun we now have both "Web Audio Editor" and "Performance Tools" twice for locales that don't have the new tag enabled.

firstrun

bedrock/firefox/templates/firefox/dev-whatsnew.html
```diff
@@ -110,7 +98,5 @@
 {% block email_form %}{% endblock %}
 
 {% block js %}
-  {% if not l10n_has_tag('spring_release') %}
-    {% javascript 'firefox_developer_whatsnew' %}
-  {% endif %}
+  {% javascript 'firefox_developer_whatsnew' %}
```
alexgibson: There are no videos on this page any longer, so is this JS file needed?

alexgibson:
On firstrun we're now falling back to "Performance Tools & Web Audio Editor" for the top spots, instead of the current "Performance Tools & Page Inspector". Is this intentional?

schalkneethling:
No, I would think they need to be the same.

schalkneethling:
So, now on firstrun, if the locale does not have firefox_dev_fall the featured section will have performance and inspector but, under the 'features and tools' section performance will be duplicated. Perhaps then I should wrap it in a firefox_dev_fall tag.

In that case, if the locale does not have the tag it will have performance and inspector as featured but, performance will not be duplicated under the second section.

schalkneethling:
Ok so now:

1) If has firefox_dev_fall tag:

Features section shows Visual editing tools and devtools challenger
Under the Features and tools, it also shows performance and inspector

2) If has not firefox_dev_fall tag:

Features section shows performance and inspector
Under the Features and tools, performance and inspector is not shown

alexgibson:
The gist is if the features are in the top section, they do not need to be repeated underneath.

alexgibson:
I can check this PR again over the weekend, but if you need this merged today @schalkneethling best ask someone else to give it a check over?

schalkneethling:
@jpetto If you have a moment r?

bedrock/firefox/templates/firefox/dev-firstrun.html
```diff
@@ -19,11 +19,7 @@
 {% block page_og_url %}{{ url('firefox.developer') }}{% endblock %}
 
 {% block page_og_desc %}
-  {% if l10n_has_tag('meta_update_201503') %}
-    {{_('Built for those who build the Web. Introducing the only browser made for developers.')}}
-  {%- else -%}
-    {{ _('Introducing the only browser made for developers like you.') }}
-  {% endif %}
+  {{_('Built for those who build the Web. Introducing the only browser made for developers.')}}
```
jpetto: Just making sure we can lose this `l10n` tag.

bedrock/firefox/templates/firefox/dev-firstrun.html
```diff
@@ -56,10 +46,8 @@
 data-sync-icon="{{ static('img/firefox/dev-firstrun/sync-blue.png') }}"
 data-sync-icon-high-res="{{ static('img/firefox/dev-firstrun/sync-blue-high-res.png') }}"
 
-{% if l10n_has_tag('new_doorhanger') %}
-  data-sync-reminder-title="{{ _('Before you go&hellip;') }}"
-  data-sync-reminder-text="{{ _('If you continue without syncing your new Developer Edition profile, you could lose access to important browsing data from your existing Firefox profile.') }} {{ _('To sync now, choose the Sync option from this menu.') }}"
-{% endif %}
+data-sync-reminder-title="{{ _('Before you go&hellip;') }}"
+data-sync-reminder-text="{{ _('If you continue without syncing your new Developer Edition profile, you could lose access to important browsing data from your existing Firefox profile.') }} {{ _('To sync now, choose the Sync option from this menu.') }}"
```
jpetto: Same with these two above.

bedrock/firefox/templates/firefox/dev-firstrun.html
```diff
@@ -92,10 +97,9 @@
           <div class="video" data-video-id="eH0R10Ga4Hs"></div>
         </div>
       </li>
-    {% endif %}
       <li class="feature">
         <a href="https://www.youtube.com/watch?v=eQqNfkqIJdw" rel="external" class="video-play">
-          <img src="{{ static('img/firefox/firstrun/dev/feature-inspector-large.jpg') }}" alt="{{ _('Screenshot') }}" class="screenshot">
+          <img src="{{ static('img/firefox/firstrun/dev/feature-inspector.png') }}" alt="{{ _('Screenshot') }}" class="screenshot">
```
jpetto: I think we need to keep this as `feature-inspector-large.jpg`, as otherwise the image sizes are mismatched:

screen shot 2015-10-30 at 4 24 46 pm

bedrock/firefox/templates/firefox/dev-firstrun.html
jpetto: Just verifying that this looks good. If `firefox-dev-fall` is off, I see it in the top featured section. If `firefox-dev-fall` is on, I see it below in the "Features and Tools" section. :+1:

bedrock/firefox/templates/firefox/dev-firstrun.html
jpetto: @flodolo verified we can lose these tags in IRC. :+1:

bedrock/firefox/templates/firefox/developer.html
```diff
@@ -48,23 +39,19 @@
...
-  {% if l10n_has_tag('new_doorhanger') %}
   <p class="feedback-note">
     {{ _('Firefox Developer Edition automatically sends feedback to Mozilla.') }}
     <a href="{{ url('privacy.notices.firefox') }}#telemetry" class="more">{{ _('Learn more') }}</a>
   </p>
-  {% endif %}
 
-  <img src="{{ static('img/firefox/developer/screenshot.jpg') }}" alt="{{ _('Firefox Developer Edition') }}" class="screenshot" />
+  {% if l10n_has_tag('firefox_dev_fall') and waffle.switch('firefox-dev-fall') %}
+    {{ high_res_img('firefox/developer/challenger-hero.png', {'alt': _('Firefox Developer Edition'), 'width': '832', 'height': 'auto', 'class': 'screenshot'}) }}
```
jpetto: Hm, just noting we're sending a 156KB image down to mobile users. Not a blocker, but maybe a good place to try out the fancy new `srcset` and `sizes` attributes after fall campaign?

jpetto:
Only one image issue. Other than that, looking good! :hamster: r+wc

bedrock/firefox/templates/firefox/dev-firstrun.html
schalkneethling: Doh! Copy/Paste fail.

schalkneethling: Fixed up.

schalkneethling:
Ok, took care of all the things. I also managed to squish the high-res hero image a little more.

schalkneethling:
Thanks @alexgibson, @jpetto and @flodolo! This should be good to merge now.

bedrock/firefox/templates/firefox/dev-whatsnew.html
```diff
@@ -68,49 +81,18 @@
           <a href="https://hacks.mozilla.org/?p=28808" rel="external" class="more">{{_('See how you helped')}}</a>
```
flodolo: I checked the link and this doesn't seem right (it's about fx40).

@ckprice Are we exposing the whatsnew page at all to localizations? How can I know? I was preparing this page and realized we never translated this blurb (added in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1136912)).

ckprice: >@ckprice Are we exposing the whatsnew page at all to localizations? How can I know?

Yes, we are.

alexgibson: As discussed with @flodolo on IRC - we're still going to merge this as-is and push all strings out for l10n.

PR #3484, opened 2015-10-28
Add functional tests build step.

glogiotatidis:
This is already used in https://ci.us-west.moz.works/job/bedrock_functional_tests/

/cc @pmclanahan @jgmize @davehunt 

- [ ] Update Jenkins job to use these scripts after merging.

### DIFF
--- a/docker/jenkins/run_functional_tests.sh
+++ b/docker/jenkins/run_functional_tests.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+#
+# Starts Selenium Hub and NUMBER_OF_NODES (default 5) firefox nodes.
+# Waits until at least one node is ready and then runs sanity tests
+# only against a local bedrock instance started for this job.
+#
+set -xe
+
+SELENIUM_VERSION=${SELENIUM_VERSION:-2.48.2}
+
+cp docker/dockerfiles/bedrock_functional_tests Dockerfile
+docker build -t bedrock_functional_tests:$GIT_COMMIT --pull=true .
+
+docker pull selenium/hub:${SELENIUM_VERSION}
+docker pull selenium/node-firefox:${SELENIUM_VERSION}
+
+docker run -d --name bedrock-code-${BUILD_NUMBER} -e ALLOWED_HOSTS="*" -e SECRET_KEY=foo -e DEBUG=False -e DATABASE_URL=sqlite:////tmp/temp.db mozorg/bedrock_code:$GIT_COMMIT
+docker run -d --name bedrock-selenium-hub-${BUILD_NUMBER} selenium/hub:${SELENIUM_VERSION}
+
+for NODE_NUMBER in `seq ${NUMBER_OF_NODES:-5}`;
+do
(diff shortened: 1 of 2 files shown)


docker/jenkins/run_functional_tests.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+set -xe
+
+cp docker/dockerfiles/bedrock_functional_tests Dockerfile
+docker build -t bedrock_functional_tests:$GIT_COMMIT --pull=true .
+
+docker pull selenium/hub:2.48.2
+docker pull selenium/node-firefox:2.48.2
```
davehunt: We'll want to update the tag when new releases are made. Perhaps we should use an environment variable, as we use it in four places already?

docker/jenkins/run_functional_tests.sh
```diff
@@ -0,0 +1,15 @@
...
+
+docker run -d --name bedrock-code-${BUILD_NUMBER} -e ALLOWED_HOSTS="*" -e SECRET_KEY=foo -e DEBUG=False -e DATABASE_URL=sqlite:////tmp/temp.db mozorg/bedrock_code:$GIT_COMMIT
+docker run -d --name bedrock-selenium-hub-${BUILD_NUMBER} selenium/hub:2.48.2
+docker run -d --name bedrock-selenium-node-${BUILD_NUMBER} --link bedrock-selenium-hub-${BUILD_NUMBER}:hub --link bedrock-code-${BUILD_NUMBER}:bedrock selenium/node-firefox:2.48.2
```
davehunt: We can probably afford to spin up a few nodes (perhaps 5?) so tests can run in parallel. We could also do this via Docker compose to start the various containers and use [scale](https://docs.docker.com/compose/reference/scale/) for the nodes.

davehunt:
Awesome work @glogiotatidis! :smile_cat:

jgmize:
@glogiotatidis would you mind taking a look at https://ci.us-west.moz.works/job/bedrock_functional_tests/lastFailedBuild/console?

davehunt:
I just noticed a failure in the pipeline, and it may be caused by the tests running before the Selenium nodes are ready.

docker/jenkins/run_functional_tests.sh
glogiotatidis: I added an env variable, although I like to make changes in the repo when possible so we have history. We can do either way now.

docker/jenkins/run_functional_tests.sh
glogiotatidis: Added a loop controlled by an env variable that defaults to 5. ATM it seems that although we spin more node we still need about 75' to run the tests but I guess we'll see the benefits of more node as the tests increase

glogiotatidis:
Added a check that wait until at least one node of the started nodes is ready.

This is ready for another look

davehunt:
Nice. r+

jgmize:
Nice work @glogiotatidis and thanks @davehunt!

PR #3485, opened 2015-10-28
Add Python functional tests for /firefox/developer page

alexgibson:


alexgibson:
@davehunt r?

tests/functional/firefox/test_developer.py
```diff
@@ -0,0 +1,58 @@
...
+    page = DeveloperPage(base_url, selenium).open()
+    assert page.is_primary_download_button_displayed
+    assert page.is_secondary_download_button_displayed
+
+
+# Skipping this test until bug 1219251 is fixed.
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1219251
+# @pytest.mark.sanity
+# @pytest.mark.nondestructive
+# def test_play_video(base_url, selenium):
```
davehunt: Use the `skipif` mark to skip tests conditionally or unconditionally:

```python
@pytest.mark.skipif(reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1219251')
```

tests/functional/firefox/test_developer.py
```diff
@@ -0,0 +1,58 @@
...
+#     page = DeveloperPage(base_url, selenium).open()
+#     video = page.developer_videos[0]
+#     video.play()
+#     assert video.modal_is_displayed
+#     video.close()
+
+
+@pytest.mark.nondestructive
+def test_newsletter_default_values(base_url, selenium):
```
davehunt: Is this another unique newsletter? If not, how many pages are we going to test the newsletter from? I wonder if we should have some newsletter tests that are parametrised by page.

tests/pages/firefox/developer.py
```diff
@@ -0,0 +1,50 @@
...
+from selenium.webdriver.support.ui import WebDriverWait as Wait
+
+from .base import FirefoxBasePage
+from ..page import PageRegion
+
+
+class DeveloperPage(FirefoxBasePage):
+
+    _url = '{base_url}/{locale}/firefox/developer'
+    _primary_download_locator = (By.CSS_SELECTOR, '.intro .download-button')
```
davehunt: Nit: Keep an empty line before the locators.

tests/pages/firefox/developer.py
```diff
@@ -0,0 +1,50 @@
...
+    _secondary_download_locator = (By.CSS_SELECTOR, '.dev-footer-download .download-button')
+
+    @property
+    def is_primary_download_button_displayed(self):
+        return self.selenium.find_element(*self._primary_download_locator).is_displayed()
```
davehunt: If the element is not present we'll get an exception instead of `False`. I wonder if we should add a `is_element_displayed` to `page.py`, catch the `NoSuchElementException` there and return `False`. I think it would be cleaner in the result of a failure and perhaps makes the page objects cleaner too. We could do this in a separate patch though, and update other uses.

tests/pages/firefox/developer.py
```diff
@@ -0,0 +1,50 @@
...
+
+    @property
+    def is_secondary_download_button_displayed(self):
+        return self.selenium.find_element(*self._secondary_download_locator).is_displayed()
+
+    @property
+    def developer_videos(self):
+        return [DeveloperVideoModal(self.selenium)]
```
davehunt: Is this incomplete due to the bug? It's currently returning a list with only one item.

tests/pages/firefox/developer.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+class DeveloperVideoModal(PageRegion):
```
davehunt: Looking at the page, I think we should have multiple `Video` PageRegion objects with a `play` function that returns a `VideoModal` PageRegion object, which should have a `close` function and `is_displayed` property. If this is the same modal used in the other tests then we could move it to the common `regions` directory.

tests/functional/firefox/test_developer.py
alexgibson: Looking at the code it seems to be the same newsletter as we're already testing elsewhere. I'll omit these tests for this page. I like the suggestion of parametrised tests by page, but will save this for another PR.

tests/pages/firefox/developer.py
alexgibson: Just going by example of other tests here, but I think it's a good suggestion. I'll add a method for this to `page.py`

tests/functional/firefox/test_developer.py
alexgibson: nice :+1:

tests/pages/firefox/developer.py
alexgibson: I think this sounds like a good abstraction, makes sense to have a generic `Modal` PageRegion since this is the same JS plugin used throughout the site. They don't always play videos, but do use common elements in the DOM.

alexgibson:
@davehunt - thanks for the review, updated.

tests/pages/regions/modal.py
```diff
@@ -0,0 +1,26 @@
...
+
+    def close(self):
+        modal = self.selenium.find_element(*self._root_locator)
+        self.root.find_element(*self._close_locator).click()
+        Wait(self.selenium, self.timeout).until(expected.staleness_of(modal))
+
+    @property
+    def is_displayed(self):
+        Wait(self.selenium, self.timeout).until(
+            expected.visibility_of_element_located(self._root_locator))
```
davehunt: This wait will cause a timeout to elapse if we want to check that the modal is _not_ displayed. I would suggest having this property return faster, and using the return value as the condition of a wait whenever we expect the modal to become visible. For example:

```python
def play_video(self):
    self.selenium.find_element(*self._video_play_locator).click()
    Wait(self.selenium, self.timeout).until(lambda s: Modal(s).is_displayed)
```

davehunt:
r+ wc

alexgibson:
@davehunt updated - is this what you had in mind?

tests/pages/contribute.py
```diff
@@ -14,26 +14,13 @@ class ContributePage(BasePage):
...
-        close = self.selenium.find_element(*self._video_close_locator)
-        close.click()
-        Wait(self.selenium, self.timeout).until(
-            expected.staleness_of(close))
-
-    @property
-    def video_modal_is_displayed(self):
-        return self.selenium.find_element(*self._video_modal_locator).is_displayed()
+        Wait(self.selenium, self.timeout).until(lambda s: Modal(s).is_displayed)
+        return Modal(self.selenium)
```
davehunt: You could instantiate Modal just once, but otherwise this looks good.

davehunt:
r+

alexgibson:
@davehunt - I made one last change with regard to initiating the PageRegions, as briefly discussed on IRC. Would appreciate if you could take one last look when you get time (no rush).

davehunt:
r+ with or without comment addressed. I think this works find for now, but may evolve a little more in the future.

PR #3488, opened 2015-10-29
Fix Bug 1218015 - Remove recipient of emails from submissions via https://www.mozilla.org/en-US/about/legal/fraud-report/

kyoshino:


schalkneethling:
r+ :cake:

PR #3489, opened 2015-10-29
Fix Bug 1195783 - CCK2 is no longer available on AMO.

kyoshino:


schalkneethling:
r+ :cake: but holding of a merge until after the fall campaign pushes are done.

PR #3396, opened 2015-10-06
Fix bug 1136463 - add iOS to Firefox family page

craigcook:

PR #3490, opened 2015-10-30
Bug 1136463 - add Private Browsing to Fx Family page

craigcook:
This includes a bunch of the CSS for adding iOS as well, but none of those rules will actually be applied to the page until #3396 lands. This PR needs to merge before #3396 (which will then need rebasing, of course).


alexgibson:
Mozilla wordmark is cropped slightly

mozilla

alexgibson:
Private browsing icon is also cropped slightly at smaller viewports

private-browsing

alexgibson:
Not necessarily a blocker, but the product icons are no longer bottom aligned in IE:

ie-grrrrr

alexgibson:
A few cropping / alignment issues but otherwise looking good :+1:

craigcook:
I've fixed the sprite clipping problems (low-res and high-res sprites weren't matching) and the IE layout (I've been spoiled by flexbox; had to do some positioning for non-flex). Should be good to go.

alexgibson:
Changes look good! r+ :cactus:

PR #3491, opened 2015-10-30
Fix bug 1220104 - restore aurora bars to iOS product page

craigcook:


alexgibson:
Thanks @craigcook r+ - I would have got this when I finished reviews, but thanks (now get some rest!)

PR #3492, opened 2015-10-30
[fix Bug 1220138 - Adding Mozilla Sweden website link to community page]

swarnava:
Adding mozilla.ch on community page.


flodolo:
mozilla.ch is the Swiss community, not Swedish...

PR #3493, opened 2015-10-30
Fix intermittent failure in l10n language selector test

alexgibson:
https://webqa-ci.mozilla.com/job/bedrock.dev.win10.ie/34/
https://webqa-ci.mozilla.com/job/bedrock.dev.win10.ie/23/

I noticed these two failures both happen when redirected to the `or` locale. My hunch is that IE is matching the URL (www-dev.allizom.**or**g) before the page is redirected, resulting in the last assert failing when the selector becomes stale.

### DIFF
--- a/tests/functional/test_l10n.py
+++ b/tests/functional/test_l10n.py
@@ -18,5 +18,5 @@ def test_change_language(base_url, selenium):
     available = [l for l in page.footer.languages if l not in excluded]
     new = random.choice(available)
     page.footer.select_language(new)
-    assert new in selenium.current_url, 'Language is not in URL'
+    assert '/{0}/'.format(new) in selenium.current_url, 'Language is not in URL'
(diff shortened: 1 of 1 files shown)


alexgibson:
@davehunt r?

tests/functional/test_l10n.py
```diff
@@ -18,5 +18,5 @@ def test_change_language(base_url, selenium):
     available = [l for l in page.footer.languages if l not in excluded]
     new = random.choice(available)
     page.footer.select_language(new)
-    assert new in selenium.current_url, 'Language is not in URL'
+    assert '/' + new + '/' in selenium.current_url, 'Language is not in URL'
```
davehunt: Good catch! I'd prefer to use string formatting, such as:

```python
assert '/{0}/'.format(new) in selenium.current_url
```

davehunt:
r+ wc

PR #3500, opened 2015-11-02
Bug 1220844 - Remove the interactive embed from SmartOn Tracking

craigcook:
Fall back to a plain link, opens in a new tab.

### DIFF
--- a/media/js/teach/smarton.js
+++ b/media/js/teach/smarton.js
@@ -208,6 +208,13 @@
         });
     });
 
+    // Open embed in a new tab/window
+    $('a.video-open').on('click', function(e) {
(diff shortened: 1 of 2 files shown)


media/js/teach/smarton.js
```diff
@@ -208,6 +208,13 @@
         });
     });
 
+    // Open embed in a new tab/window
+    $('a.video-open').on('click', function(e) {
+        e.preventDefault();
+        window.open(this.href);
```
jgmize: Why is this js necessary?

craigcook: This opens the link in a new tab. They will ask for it. I know it.

PR #3501, opened 2015-11-03
Bug 1220883 - Update address and map marker of Beijing office

l-hedgehog:


alexgibson:
I believe the co-ordinates here may also need updating? https://github.com/mozilla/bedrock/blob/master/media/js/mozorg/contact-data.js#L67

l-hedgehog:
@alexgibson, yes, I updated it. It looks like GitHub somehow decides to not show it in the diff view by default.

alexgibson:
Thanks @l-hedgehog, I totally missed that! This change looks fine to me, r+

I'll merge this a bit later, once fall campaign has launched and things calm down a little.

l-hedgehog:
@alexgibson, thanks.

PR #3502, opened 2015-11-03
Start from debian:jessie.

glogiotatidis:
Using the python images from registry does not play well with our outdated docker version on the CI server. Debian images are OK for now.

Error https://ci.us-west.moz.works/job/bedrock_functional_tests/65/console

 @davehunt  r?


davehunt:
r+

glogiotatidis:
Thanks!

PR #3503, opened 2015-11-03
Fix firefox/choose url config.

pmclanahan:


bedrock/firefox/urls.py
```diff
@@ -34,7 +34,7 @@
         views.all_downloads, name='firefox.all'),
     page('firefox/channel', 'firefox/channel.html'),
     redirect('^firefox/channel/android/$', 'firefox.channel', locale_prefix=False),
-    url('firefox/choose', views.choose, name='firefox.choose'),
+    url(r'^firefox/choose/$', views.choose, name='firefox.choose'),
```
jgmize: should we make the trailing slash optional?

pmclanahan: trailing slashes aren't optional in urls.py files. it is optional in global.conf.

pmclanahan: django fixes them for us if the version w/o the slash isn't found.

pmclanahan:
@ckprice it fixes it. it's not working in some situations, and it's wrong in the code. if you'd like us to not fix it then we'll hold off.

PR #3504, opened 2015-11-03
Fix SmartOn short URL for social sharing

alexgibson:


craigcook:
Oh crap, is bitly not ssl? Sorry about that, I should have verified. Thanks for the quick fix!

PR #3505, opened 2015-11-03
Bug 1221167 - Make alt text on Google Play button localizable

flodolo:
Tested locally Android and Family page, not sure how to check the PB one.


alexgibson:
It appears as though the android badges are currently unused on the PB page (I see no logic to hide/show them). I have inspected the page though and can see the correct alt text is used there, so r+ :rocket:

PR #3507, opened 2015-11-03
[fix bug 1217229] Create stand-alone Fx mobile download page.

jpetto:
Will be en-US only for the time being, so English-only image is okay for now.


jpetto:
Updated image to match locale-friendly image used for updated `/firefox/new/` scene 2.

bedrock/firefox/templates/firefox/mobile-download.html
```diff
@@ -0,0 +1,78 @@
...
+          </li>
+          <li class="ios">
+            <a href="{{ settings.APPLE_APPSTORE_FIREFOX_LINK }}&amp;ct=mozorg-heartbeat_landing-button">
+              <img src="{{ l10n_img('firefox/ios/btn-app-store.svg') }}" alt="{{ _('Download on the App Store') }}" width="152" height="45">
+            </a>
+          </li>
+        </ul>
+      </div>
+
+      <div id="mobile-device-lineup"></div>
```
craigcook: Ordinarily I would complain about this empty element since we're not doing any fancy JS reveal so the image could just be a background on `#scene` or `#stage` but I suppose we don't need to overthink this page right now.

media/css/firefox/mobile-download.less
```diff
@@ -0,0 +1,222 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../sandstone/lib.less";
+
+.sky #outer-wrapper {
+    background: #eaeff2;
+    background: -webkit-radial-gradient(center, ellipse cover, rgba(234,239,242,1) 0%, rgba(212,221,228,.4) 60%),
```
craigcook: Should we use the radial gradient mixin?

craigcook:
I leave it to you to switch to the gradient mixin if you want to, but it's not a big deal. With the quick-and-dirty nature of the page and the limited audience I'm inclined not to nitpick as much as I might otherwise, so r+wc.

bedrock/firefox/templates/firefox/mobile-download.html
```diff
@@ -0,0 +1,78 @@
...
+  <section id="stage">
+    <div id="scene">
+      <div id="thankyou" tabindex="0">
+        {{ _('Add it to your Android and iOS devices.') }}
+      </div>
+
+      <div class="mobile-download-buttons">
+        <ul>
+          <li class="android">
+            <a rel="external" href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}">
```
craigcook: Does this need the same params as the app store button below?

media/css/firefox/mobile-download.less
```diff
@@ -0,0 +1,222 @@
...
+            float: none;
+            margin: 0 auto (@baseLine / 2);
+        }
+    }
+
+    #mobile-device-lineup {
+        display: none;
+    }
+
+    .footer-newsletter-form {
```
craigcook: No newsletter form? Probably don't need this.

bedrock/firefox/templates/firefox/mobile-download.html
jpetto: Yeah, I just went with what was easiest. If this page ever becomes "official", we'll clean it up.

media/css/firefox/mobile-download.less
jpetto: Ah, good call. Will remove.

bedrock/firefox/templates/firefox/mobile-download.html
jpetto: I don't think so - I believe we get decent GA from the Google Play store, but with the App Store we need additional params to determine the source.

media/css/firefox/mobile-download.less
jpetto: Don't think I can use the mixin, as we use multiple/layered gradients all in one `background` declaration.

jpetto:
Removed the superfluous `.footer-newsletter-form` class. Should be good to go?

media/css/firefox/mobile-download.less
craigcook: Ah yeah, I see now. Nevermind!

PR #3509, opened 2015-11-03
Change /firefox/releasenotes/ from 500 to 302

jgmize:


jgmize:
Fix errors like https://rpm.newrelic.com/accounts/263620/applications/9716170/traced_errors/9e7727-38ab87ae-8260-11e5-900f-b82a72d22a14

jgmize:
@pmclanahan r?

etc/httpd/global.conf
```diff
@@ -603,7 +603,7 @@ RewriteRule ^(/(?:\w{2,3}(?:-\w{2})?/)?mobile/android-download(-beta)?\.html)$ /
 RewriteRule ^(/(?:\w{2,3}(?:-\w{2})?/)?projects/calendar(/.*)?)$ /b$1 [PT]
 
 # bug 947890
-RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/latest/releasenotes(/?)$ /b/$1firefox/latest/releasenotes$2 [PT]
+RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?firefox/(?:latest/)?releasenotes(/?)$ /b/$1firefox/latest/releasenotes$2 [PT]
```
pmclanahan: I don't think `firefox/releasenotes/` should pass through to `firefox/latest/releasenotes/`. And since it was throwing 500s I assume some other rule is already passing it through to bedrock.

jgmize: It only returns a 500 if the request actually gets to bedrock, so we get 500s in AWS, but in SCL3 it returns a 404 because it doesn't pass through. Apologies, I meant to add a comment earlier explaining that.

Would you prefer a separate line that passes `/firefox/releasenotes/` through as `/firefox/releasenotes/` and doesn't explicitly link them together in `global.conf`?

pmclanahan: Ah. Gotcha. I just don't want a mismatch to make things strange later. My preference would be to make the `(latest/)?` bit capture and use it in the destination so that Django sees the real request.

jgmize:
> My preference would be to make the (latest/)? bit capture and use it in the destination so that Django sees the real request.

@pmclanahan I updated the PR to match your preference :)

PR #3512, opened 2015-11-04
Add functional tests for remaining contribute pages

alexgibson:


alexgibson:
@davehunt r?

tests/pages/contribute/stories.py
```diff
@@ -0,0 +1,36 @@
...
+
+from selenium.webdriver.common.by import By
+from selenium.webdriver.support.ui import WebDriverWait as Wait
+
+from .base import ContributeBasePage
+
+
+class ContributeStoriesPage(ContributeBasePage):
+
+    _url = '{base_url}/{locale}/contribute/stories/shreyas'
```
davehunt: Did you mean to hard-code the story? It seems there are a few, and not all have this story toggle locator. Perhaps this page object could override the `open` and `url` methods to allow specific stories to be opened, but default to 'shreyas'. Thinking about it, perhaps that's adding too much complexity for a simple test. Alternatively, perhaps add a comment to indicate that we're testing a specific story.

alexgibson: They do all have the toggle, but we would essentially be testing the same template logic three times, so I considered it overkill for just one small test. Will add a comment to make it clear the intention.

tests/functional/contribute/test_events.py
```diff
@@ -0,0 +1,19 @@
...
+@pytest.mark.nondestructive
+def test_events_table_is_displayed(base_url, selenium):
+    page = ContributeEventsPage(base_url, selenium).open()
+    assert page.events_table_is_displayed
+
+
+@pytest.mark.nondestructive
+def test_next_event_is_displayed(base_url, selenium):
+    page = ContributeEventsPage(base_url, selenium).open()
+    assert page.next_event_is_displayed
```
davehunt: Is it worth checking that there's at least one event listed, or that all events have date/event/location details? The total at the top of the page doesn't appear to match the number of events listed for me - should it?

alexgibson: The total at the top of the page is hard coded, not dynamic. I don't think we need to go as far as covering that each field from the DB is populated in the template. If there was an issue with the cron or elsewhere, this region would not get displayed.

tests/pages/contribute/stories.py
davehunt: https://www.mozilla.org/en-US/contribute/stories/ruben/ doesn't have the toggle - is that a bug?

alexgibson: Ah, you are quite right - nope it's not a bug I believe.

tests/functional/contribute/test_signup.py
```diff
@@ -0,0 +1,62 @@
...
+    category = page.contributor_categories[0]  # Select coding category
+    area = page.category_areas[0]  # Coding category areas
+    category.select_category
+    assert area.changes_to_required_field
+    assert area.is_displayed
+    category = page.contributor_categories[4]  # Select helping category
+    category.select_category
+    area.wait_to_collapse()
+    assert not area.is_required
+    assert not area.is_displayed
```
davehunt: I think these could either be split into separate tests or we could parameterise them. I also think we could make the test a little more readable by avoiding selecting categories by index. Something like the following could work as a parameterised test:

```python
@pytest.mark.sanity
@pytest.mark.nondestructive
@pytest.mark.parametrize('category, is_required', ['coding', True, 'helping', False])
def test_toggle_category_select(base_url, selenium, category, is_required):
    page = ContributeSignUpPage(base_url, selenium).open()
    page.select_category(category)
    assert page.is_area_displayed
```

This would involve a new method in the page object that clicked a category based on the label, which could be a property of the `ContributorCategory` region. For area, you could return True if any of the areas are displayed instead of matching category to area by index.

After writing out the example above, I think it might make more sense as two separate functional tests.

alexgibson: I did wonder if this was perhaps too long also, will take this on board and see what I can do :+1:

tests/functional/contribute/test_signup.py
```diff
@@ -0,0 +1,62 @@
...
+    page.type_email('noreply@mozilla.com')
+    page.select_country('United Kingdom')
+    page.select_text_format()
+    page.accept_privacy_policy()
+    page.click_start_contributing()
+    assert '/{0}'.format('?c=coding') in selenium.current_url, 'Category is not in URL'
+
+
+@pytest.mark.skipif(reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1221526')
+def test_successful_sign_up_helping(base_url, selenium):
```
davehunt: As above, these tests could be combined using parameterisation, but I'd at least make the category selection be by label instead of index.

tests/pages/contribute/signup.py
```diff
@@ -0,0 +1,115 @@
...
+    @property
+    def changes_to_required_field(self):
+        assert not self.is_required, 'Area field is already required'
+        Wait(self.selenium, self.timeout).until(lambda s: self.is_required)
+        return self.is_required
+
+    @property
+    def is_required(self):
+        el = self.selenium.find_element(*self._area_locator)
+        return el.get_attribute('required') == 'true'
```
davehunt: It appears that any area displayed is required. Do we need the additional check for required?

alexgibson: The idea here is that the areas are only set as required when they are made visible, hence the need to check for these cases specifically.

tests/pages/contribute/signup.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def wait_to_expand(self):
```
davehunt: This doesn't appear to be called from anywhere.

alexgibson:
@davehunt - updated thanks, good for a second look?

tests/functional/contribute/test_signup.py
```diff
@@ -0,0 +1,58 @@
...
+def test_toggle_category_without_areas(base_url, selenium):
+    page = ContributeSignUpPage(base_url, selenium).open()
+    page.select_helping_category()
+    assert not page.is_areas_region_displayed
+
+
+def test_successful_sign_up_coding(base_url, selenium):
+    page = ContributeSignUpPage(base_url, selenium).open()
+    page.select_coding_category()
+    assert page.is_coding_area_displayed
```
davehunt: We don't need this duplicate assert, `test_toggle_category_with_areas` will fail if this is False, and the next line will also fail anyway.

tests/functional/contribute/test_signup.py
```diff
@@ -0,0 +1,58 @@
...
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+def test_toggle_category_without_areas(base_url, selenium):
+    page = ContributeSignUpPage(base_url, selenium).open()
+    page.select_helping_category()
+    assert not page.is_areas_region_displayed
+
+
+def test_successful_sign_up_coding(base_url, selenium):
```
davehunt: Do we want to skip this one too?

tests/functional/contribute/test_signup.py
alexgibson: Good catch, forgot to remove this line :+1:

tests/functional/contribute/test_signup.py
alexgibson: It's only certain categories that don't work against dev currently. I think we should be ok to keep this one?

tests/pages/contribute/signup.py
```diff
@@ -0,0 +1,89 @@
...
+    def select_coding_category(self):
+        self.selenium.find_element(*self._coding_category_locator).click()
+        self.wait_for_coding_area_to_display()
+
+    def select_helping_category(self):
+        self.selenium.find_element(*self._helping_category_locator).click()
+
+    def wait_for_coding_area_to_display(self):
+        # required attribute is set once the area finishes animating
+        assert not self.is_coding_area_required, 'Coding area is already required'
```
davehunt: This looks like it could cause a race condition. I don't think we need `wait_for_coding_area_to_display` or the initial check for the coding area to be not required. The wait can go directly into `select_coding_category` as it's only called from there anyway.

tests/functional/contribute/test_signup.py
davehunt: Should we also add `assert page.is_coding_area_required`?

tests/pages/contribute/signup.py
alexgibson: Ok, good call :+1:

tests/functional/contribute/test_signup.py
davehunt: Oops, I meant this on line 16

alexgibson: Ok, understood :)

tests/functional/contribute/test_signup.py
```diff
@@ -0,0 +1,58 @@
...
+    page.select_coding_category()
+    assert page.is_coding_area_displayed
+
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+def test_toggle_category_without_areas(base_url, selenium):
+    page = ContributeSignUpPage(base_url, selenium).open()
+    page.select_helping_category()
+    assert not page.is_areas_region_displayed
```
davehunt: We should also check this is `True` in `test_toggle_category_with_areas`. I know it's implied, but it's good to check both states of a boolean property if possible.

alexgibson: Ok, makes sense

tests/functional/contribute/test_signup.py
davehunt: Ah, ok

alexgibson:
@davehunt updated, thanks!

davehunt:
r+ looks great!

alexgibson:
Thanks for the review! Will merge this one shortly

PR #3514, opened 2015-11-05
Fix bug 1221632: Disable auto-escape for newsletter titles and descriptions

pmclanahan:


pmclanahan:
@flodolo this works for me. can you give it a try as well?

flodolo:
Tested locally and works for me. Also compared the page content before/after the patch, the only change are those `&nbsp;`.
r+ on my side.

pmclanahan:
Thanks flod.

PR #3515, opened 2015-11-05
Fix Bug 1218583, swap gear and webmaker promos and update content

schalkneethling:
@alexgibson r?

### DIFF
--- a/bedrock/mozorg/templates/mozorg/home/home.html
+++ b/bedrock/mozorg/templates/mozorg/home/home.html
@@ -174,7 +174,17 @@
             <a class="fxos-link" href="{{ url('firefox.desktop.index') }}">{{ _('Learn more') }}</a>
           </div>
         </li>
-      {% if l10n_has_tag('webmaker_promo_introducing') %}
+      {% if l10n_has_tag('gear_promo_eoy') %}
+        <li id="promo-6" class="item promo-large-landscape gear" tabindex="0">
+          <h2 class="primary go">{{ _('Get the gear, <br>support the mission') }}</h2>
+          <a class="panel-link" rel="external" href="https://gear.mozilla.org?ref=holiday2015&amp;utm_campaign=holiday2015&amp;utm_source=mozilla.org&amp;utm_medium=referral" data-promotion-id="promo-6" data-promotion-name="Get the gear, support the mission" data-promotion-class="promo-large-landscape" data-promotion-type="tile">
+            <div class="secondary">
+              <h3>{{ _('Shop for t-shirts, hoodies, hats<br> and more at Mozilla Gear.') }}</h3>
+              <p class="more">{{ _('Shop now') }}</p>
+            </div>
+          </a>
+        </li>
+      {% elif l10n_has_tag('webmaker_promo_introducing') %}
         <li id="promo-6" class="item promo-large-landscape webmaker" tabindex="0">
           <h2 class="primary go">{{ _('Introducing<br> the new Webmaker') }}</h2>
           <a class="panel-link" rel="external" href="https://webmaker.org/" data-promotion-id="promo-6" data-promotion-name="Introducing the new Webmaker" data-promotion-class="promo-large-landscape" data-promotion-type="tile">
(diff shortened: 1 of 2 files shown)


alexgibson:
If a locale does not have the tag `gear_promo_eoy` enabled, but does have `webmaker_promo_introducing`, they will see two Web Maker promos. Is this intended?

schalkneethling:
@alexgibson good catch, updated.

bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -233,10 +243,17 @@ <h2 class="primary go">{{ _('Teach the Web') }}</h2>
           </a>
         </li>
       {% endif %}
-      {% if l10n_has_tag('gear_store') %}
+      {# only show webmaker promo below if new gear store promo is shown #}
+      {% if l10n_has_tag('gear_promo_eoy') %}
```
alexgibson: This doesn't guarantee that a locale which has the `gear_promo_eoy` tag also has the web maker strings translated. Perhaps we could do something like `if not l10n_has_tag('gear_promo_eoy') and l10n_has_tag('webmaker_promo_introducing')`?

Does this make sense @flodolo?

alexgibson: Wait, I'm not sure this this would work either? hmmm

bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -233,10 +243,17 @@ <h2 class="primary go">{{ _('Teach the Web') }}</h2>
...
+        <li id="promo-9" class="item promo-small-landscape webmaker">
+          <a class="panel-link" href="https://webmaker.org/" data-promotion-id="promo-9" data-promotion-name="Introducing the new Webmaker" data-promotion-class="promo-small-landscape" data-promotion-type="tile">
+            <h2>{{ _('Introducing<br> the new Webmaker') }}</h2>
+          </a>
+        </li>
+      {% elif l10n_has_tag('gear_store') %}
         <li id="promo-9" class="item promo-small-landscape gear">
           <a class="panel-link" href="https://gear.mozilla.org/?utm_source=gear.mozilla.org&amp;utm_medium=referral&amp;utm_content=mozillaorg_smallblock" data-promotion-id="promo-9" data-promotion-name="Get gear" data-promotion-class="promo-small-landscape" data-promotion-type="tile">
-            <h2>{{ _('Official Mozilla gear is here') }}</h2>
+              <h2>{{ _('Official Mozilla gear is here') }}</h2>
```
alexgibson: Nit: fix indentation

alexgibson: Could locales end up seeing two gear store promos here too?

alexgibson:
Briefly talked through this PR with @flodolo on IRC, and we think the simplest solution will be to accept that some locales might see either two Gear Store promos, or two Web Maker promos. There is just too much complexity with all the existing tags in this page.

media/css/mozorg/home/home-promo.less
```diff
@@ -866,10 +866,10 @@ html[lang|="en"] {
     background: #0ad; // bright blue
 
     h2 {
-        padding: 30px 20px 10px 140px;
+        padding: @baseLine 20px 10px 140px;
```
alexgibson: These changes are breaking an older web maker promo that's already on the page:

teach-the-web

I think we should probably use a new class name for this newer promo, instead of altering existing ones.

alexgibson:
`Introducing<br> the new Webmaker` is also a lot of text for a small promo. We should probably use a smaller base font size for all locales, and then use a [larger font size for English locales](https://github.com/mozilla/bedrock/blob/master/media/css/mozorg/home/home-promo.less#L780-L784).

alexgibson:
Is this still missing the regular default font size for English?

media/css/mozorg/home/home-promo.less
```diff
@@ -873,6 +873,18 @@ html[lang|="en"] {
...
 
+.promo-small-landscape.introducing-webmaker {
+    background: #0ad; // bright blue
+
+    h2 {
+        padding: 30px 20px 20px 140px;
+        min-height: 64px;
+        background: url(/media/img/home/voices/promos/webmaker/webmaker-logo.svg) 15px 15px no-repeat;
+        .background-size(100px, 100px);
+        .font-size(1.1rem);
```
alexgibson: Nit: I know it works the same, but use `px` value here for consistency? (it get's converted to `rem` units anyway, and keeps for consistency with other promos/pages).

schalkneethling:
@alexgibson I reckon the size works for English as well so, we can keep it as is. Unless you feel that we should bump the size?

alexgibson:
Yeah true, it does work for English as there's more space than before - but it's not consistent with the other small promo font sizes?

schalkneethling:
@alexgibson updated.

alexgibson:
r+

PR #3517, opened 2015-11-05
Fix Bug 1221826, update header and footer donate urls en-US only

schalkneethling:
@jpetto r?

### DIFF
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -113,7 +113,7 @@
                 <a data-link-type="firefox" data-element-location="nav click" href="{{ url('firefox.family.index') }}">{{_('Firefox')}}</a>
               </li>
               <li class="last donate-item">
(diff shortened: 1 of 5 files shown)


schalkneethling:
Need to wait for bug 1221736 to be resolved before merging this. Just to avoid the links ending up in production before.

jpetto:
Code looks good. r+

From the bug and local testing, it appears the `donate.mozilla.org` links seamlessly redirect to `sendto.mozilla.org`. We sure we need to wait to merge?

schalkneethling:
@jpetto if it redirects with no problem, then let's merge this. The reason I was holding off was because one of the comments said the links do not work. I assume the person actually meant, they are still being redirected.

jpetto:
Verified links are redirecting properly. When redirect is removed, link will correctly send user to `donate.mozilla.org`. Merge time!

PR #3518, opened 2015-11-05
Fix Bug 1222144, fixes source parameter on manifesto page

schalkneethling:
@jpetto r?


jpetto:
Yep, looks good to me. r+ :pager:

PR #3519, opened 2015-11-05
Fix bug 1214294: redirect en-US/firefox/os to 2.5

pmclanahan:


pmclanahan:
@malena these are the changes you need. I went ahead and filed a PR just to have somewhere to put them. You're welcome to just take the diff and apply the bits you need locally. I added a dummy template and urls.py change so that tests would pass, but you only really need the stuff in `views.py` and `test_views.py`.

malena:
okay thanks @pmclanahan!

PR #3522, opened 2015-11-06
Fix Bug 1222577, replace viewsource with mdn 10th promo

schalkneethling:
@jpetto r?


media/css/mozorg/home/home-promo.less
```diff
@@ -1042,24 +1042,6 @@ html[lang|="en"] {
 
 
 /****************************************************************************/
-// ViewSource Conf
-.promo-large-landscape.viewsource {
-    background-color: #00b085;
-
-    h2 {
-        padding-top: 120px;
-        .at2x('/media/img/home/voices/promos/viewsource/logo.png', 280px, 55px);
```
craigcook: We can also delete these images.

jpetto:
Yep, what @craigcook said. :grin:

Other than that, looks good. r+wc

media/css/mozorg/home/home-promo.less
schalkneethling: I thought about it but then, thought perhaps to keep it around for future use? There is gonna be a ViewSource again after all :)

craigcook: True, but that's probably a ways off and we'll almost certainly have a new home page by then. View Source might even have a new logo by then for all we know. Deleting the images needn't be a blocker for merge, but it's nice to keep things tidy.

schalkneethling: :+1: Agreed, will clean it up.

schalkneethling:
@jpetto updated. now with images removed ;p

PR #3523, opened 2015-11-06
[fix bug 1222181] Update PB CTA URL for Android/iOS.

jpetto:
- Fix bug on Windows where download button shows for
  up to date Fx users.

- Fix bug on Windows where two download buttons show
  for out of date Fx users.

### DIFF
--- a/media/js/base/global.js
+++ b/media/js/base/global.js
@@ -126,6 +126,13 @@ function isFirefoxMobile(userAgent) {
     return /Mobile|Tablet|Fennec/.test(ua);
 }
 
+// iOS does not follow same version numbers as desktop & Android, so may not be safe to
+// simply add this check to isFirefox. Will require further investigation/testing.
+function isFirefoxiOS(userAgent) {
+    var ua = userAgent || navigator.userAgent;
+    return /FxiOS/.test(ua);
+}
+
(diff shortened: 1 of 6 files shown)


media/js/base/global.js
```diff
@@ -126,6 +126,13 @@ function isFirefoxMobile(userAgent) {
     return /Mobile|Tablet|Fennec/.test(ua);
 }
 
+// iOS does not follow same version numbers as desktop & Android, so may not be safe to
+// simply add this check to isFirefox. Will require further investigation/testing.
+function isFirefoxiOS(userAgent) {
```
alexgibson: Looks like this function is used in `/media/firefox/ios.js` too. Should we update that to use the global version here to?

alexgibson:
Not introduced by this change, but I see a couple of linting errors when viewing the JS file locally. Worth tidying up here too?

`Line 5: site is defined but never used`
`Line 17: use ===`

alexgibson:
New logic seems good, r+wc :wine_glass:

media/js/base/global.js
jpetto: Yep - good call.

jpetto:
Put the fixes in a separate commit for easy checking.

alexgibson:
Changes look good, care to squash?

PR #3525, opened 2015-11-08
Bug 1209648 - Add iOS browser image

craigcook:
The platform images require a specific filename so I'm not sure what we can do to make the new images overwrite the cached ones. @pmclanahan, any ideas?

Also, both images are actually the same size, which is apparently the case for all the so-called high-res browser images except for Mac. Creative bug is https://bugzilla.mozilla.org/show_bug.cgi?id=1215743 if you need a proper 2x version from Ty.

This branch is in the moz repo so someone else can take over while I'm out.


pmclanahan:
@craigcook where are these images referenced in the code? I can only find mention of either of them in a JS test.

pmclanahan:
If they're just referenced in a page, then our cache busting will work. If it's in CSS and no other CSS changes are being made, then it might need a ***special*** comment in the less file.

alexgibson:
r+ :boat:

PR #3534, opened 2015-11-10
[fix bug 1221328] Update /firefox/android/ send to device widget.

jpetto:


media/js/firefox/android.js
```diff
@@ -283,85 +288,21 @@
...
-    // set up default behavior for send to device widget - opens in modal
-    // used for no variant (no test) and variant "a"
-    var defaultSendToDevice = function() {
-        // add class to display the widget
-        $('#send-to-device-default').removeClass('send-to-device-variant');
 
-        // reference modal wrapper for widget
-        $sendToDeviceModal = $('#send-to-modal-container');
+        // add hidden input to update basket id
+        $('#send-to-device-form').append('<input type="hidden" name="send-to-device-basket-id" value="android-embed">');
```
alexgibson: If `android-embed` is now the default basket ID for this form, would it be better in the macro itself as opposed to added here in JS? I'm assuming this is triggering a specific email too, in which case two questions come to mind:

1.) Is it translated into all the languages we need to make this default?
2.) If so, should we just make this email the default basket ID for android, instead of sticking with the name `android-embed`? Would this be too much of a pain to make renaming worthwhile?

alexgibson:
Can we also get rid of the variant logic in the android view? https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/views.py#L656

media/js/firefox/android.js
jpetto: As per [ADavis' comment in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1221328#c7), we want to keep the unique basket ID for this page.

However, your question regarding translation is still **very** valid. [Asked in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1221328#c8). Marking as do not merge until we have a resolution.

jpetto:
Hm, yeah, I think we can get rid of the variant logic for now. There is [another round of tests coming](https://bugzilla.mozilla.org/show_bug.cgi?id=1221338), but easy enough to go into the history to re-animate when they're ready.

jpetto:
[bniolet](https://bugzilla.mozilla.org/show_bug.cgi?id=1221328#c27) and [ADavis](https://bugzilla.mozilla.org/show_bug.cgi?id=1221328#c22) have given the green light in the bug, so we're good for final review and merge.

alexgibson:
Sorry for missing this first time around, but looks like the widget is off center on small viewports:

mobile

media/js/firefox/android.js
```diff
@@ -23,6 +23,11 @@
         timeout: 2000
     };
 
+    // send to device vars
+    var hasWidget = $('#intro').data('has-widget') !== 'False';
```
alexgibson: Nit: even though jQuery is relax on data-attributes, I believe the standards way is to use camelCase: `$('#intro').data('hasWidget')`

alexgibson:
r+wc from me :frog:

media/js/firefox/android.js
jpetto: Ah, yes indeed cameCase would be better. :+1:

jpetto:
Fixed tablet/mobile centering and updated jQuery data attribute format.

Agree that a unique basket id on this page complicates things. I believe it's been decided that the increased conversions are worth the dev cost, but I'd like to keep this on the radar as something to possibly improve in 2016.

PR #3537, opened 2015-11-11
[bug 1221303] Update pixel campaign id's on /firefox/os/2.5/

alexgibson:


alexgibson:
@malena r?

alexgibson:
Malena is on PTO today - @jpetto or @schalkneethling r?

alexgibson:
For the record, I'm [going by this for direction](https://bugzilla.mozilla.org/show_bug.cgi?id=1221303#c6).

media/js/firefox/os/firefox-os-2.5.js
```diff
@@ -8,22 +8,23 @@
     $('.cta-button').on('click', function(e) {
         if (!window._dntEnabled()) {
```
jpetto: Would it be better to put this `if` statement around the event listener? Or am I over-thinking it?

alexgibson: Could do yeah, why not :)

jpetto:
I think your inference on which tag belongs on which CTA is correct, based on the little detail we have. :+1:

alexgibson:
Thanks @jpetto updated. Let me know if looks good and I'll squash

jpetto:
Looking good. Squarsh time!

alexgibson:
Squished, merge at will :anchor:

PR #3538, opened 2015-11-11
[fix bug 1221643] Remove /whatsnew/b/ from bedrock

alexgibson:
This URL was for an Optimizely A/B test, which is no longer running. There should be zero traffic going to `/firefox/42.0/whatsnew/b/` so should be safe to remove.

This change should leave `/firefox/42.0/whatsnew/` and `/firefox/42.0/firstrun/learnmore/` both intact.


alexgibson:
@jpetto r?

jpetto:
LGTM! r+

PR #3539, opened 2015-11-11
[fix bug1223813] Add iOS id for Send to Device form SMS messaging

alexgibson:


pmclanahan:
r+ just waiting for Travis.

PR #3542, opened 2015-11-11
Fix Bug 1139961 - Update universal download button to add link to App…

jpetto:
… Store for iOS

Props to Kohei

### DIFF
--- a/bedrock/firefox/helpers.py
+++ b/bedrock/firefox/helpers.py
@@ -60,7 +60,7 @@ def download_firefox(ctx, channel='release', small=False, icon=True,
     :param channel: name of channel: 'release', 'beta' or 'alpha'.
     :param small: Display the small button if True.
     :param icon: Display the Fx icon on the button if True.
-    :param platform: Target platform: 'desktop', 'android' or 'all'.
+    :param platform: Target platform: 'desktop', 'android', 'ios', or 'all'.
     :param dom_id: Use this string as the id attr on the element.
     :param locale: The locale of the download. Default to locale of request.
     :param simple: Display button with text only if True. Will not display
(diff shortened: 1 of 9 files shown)


bedrock/firefox/helpers.py
```diff
@@ -145,14 +152,15 @@ def download_firefox(ctx, channel='release', small=False, icon=True,
     data = {
         'locale_name': locale_name,
         'version': version,
-        'product': 'firefox-android' if platform == 'android' else 'firefox',
+        'product': 'firefox-%s' % platform,
```
jpetto: Is this going to be okay? Old possible values were `firefox` and `firefox-android`. New possible values would be `firefox`, `firefox-android`, `firefox-ios`, and `firefox-all`.

Didn't run in to any issues while manual/automated testing, but want to double check.

jgmize: the possible values are controlled by product-details, so other values could be introduced (either intentionally or not) and we need to identify what would happen in that case.

jpetto: As far as I can tell, this entire `data` hash is sent to the `download-button.html` template and that's it. I don't see any spot in that template where the `product` value is used, so...maybe this change has zero impact?

jgmize: yeah, seems like it to me too, but I thought I might be missing something. unless @pmclanahan knows something I don't, the 'product' key doesn't seem to be used at all here?

jpetto: Could be a legacy value from an old version of `download-button.html`?

jpetto:
Possible lingering questions on the `download_firefox` helper, but testing shows all is working as expected.

r+ from me, but would like a confirming r+ from @jgmize and/or @pmclanahan before merging.

bedrock/firefox/helpers.py
jgmize: I believe you're right. I think we're safe here, and we can follow up with removing this and any other unused keys from the data dict in a separate PR.

pmclanahan: Must be. As far as I can tell this change seems fine and has no affect. Obviously this changes the value since it can no longer be just "firefox", but it doesn't seem to be used.

PR #3544, opened 2015-11-12
fix bug 1224060

chikoski:
This pull request will add a redirection rule to navigate Japanese users, who access to releasenotes of Firefox for iOS, to Japanese version of the releasenotes they want to read.

### DIFF
--- a/bedrock/firefox/redirects.py
+++ b/bedrock/firefox/redirects.py
@@ -241,4 +241,8 @@ redirectpatterns = (
 
     # bug 960543
     redirect(r'^firefox/(?P<vers>[23])\.0/eula', '/legal/eula/firefox-{vers}/'),
+
+    # bug 1224060
+    redirect(r'^ja/firefox/ios/(?P<vers>[0-9]+(\.[0-9]+)*)/(?P<file>releasenotes|system-requirements)',
+             'http://www.mozilla.jp/firefox/ios/{vers}/{file}/', locale_prefix=False),
 )


bedrock/firefox/redirects.py
```diff
@@ -231,4 +231,8 @@ def firefox_mobile_faq(request, *args, **kwargs):
 
     # bug 960543
     redirect(r'^firefox/(?P<vers>[23])\.0/eula', '/legal/eula/firefox-{vers}/'),
+
+    # bug 1224060
+    redirect(r'^ja/firefox/ios/(?P<vers>[0-9]+(\.[0-9]+)*)/(?P<file>releasenotes|system-requirements)',
+             'http://www.mozilla.jp/firefox/ios/{vers}/{file}/'),
```
pmclanahan: Need to add the argument `locale_prefix=False` to the call to `redirect()` so that your regex will match against the locale. Otherwise I think this is okay. I looked to see if this should be a redirect to `https://www.mozilla.jp` but that gives a cert error, which is very unfortunate. I'd love for that to be fixed and a PR filed here to update all redirects to use TLS.

I'd also like to see a test for this added to `test_redirects/map_globalconf.py`.

chikoski: Thanks. 
Added local_prefix flag. 

www.mozilla.jp does not services https, so I did not change the destination URL.

pmclanahan:
Looks good. Just waiting on a redirect test to be added to `test_redirects/map_globalconf.py` and we'll be ready to merge.

PR #3545, opened 2015-11-12
Fix Bug 1224075, reference correct image file for ios badge on family index page

schalkneethling:
@alexgibson r?


alexgibson:
Looks like this is a dupe of #3548, sorry didn't realise there was already a PR for this.

PR #3546, opened 2015-11-12
Fix Bug 1224082 - Update subject line of emails from submissions via https://www.mozilla.org/en-US/about/legal/fraud-report/

kyoshino:


jgmize:
thanks @kyoshino!

PR #3550, opened 2015-11-12
Add Python functional tests for /firefox/os/devices page

alexgibson:


alexgibson:
@davehunt r?

tests/functional/firefox/os/test_devices.py
```diff
@@ -0,0 +1,58 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from ....pages.firefox.os.devices import DevicesPage
```
davehunt: Once #3562 lands we can make this an absolute import.

tests/functional/firefox/os/test_devices.py
```diff
@@ -0,0 +1,58 @@
...
+    menu = page.open_family_navigation_menu()
+    assert menu.is_displayed
+
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+def test_open_close_modal(base_url, selenium):
+    page = DevicesPage(base_url, selenium).open()
+    page.select_location('Germany')
+    modal = page.open_modal()
```
davehunt: It's not clear what the `open_modal` method does. Could we rename it to `get_a_phone`?

tests/functional/firefox/os/test_devices.py
```diff
@@ -0,0 +1,58 @@
...
+    assert modal.is_displayed
+    modal.close()
+
```
davehunt: We could `assert not modal.is_displayed` here.

tests/functional/firefox/os/test_devices.py
```diff
@@ -0,0 +1,58 @@
...
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+def test_open_close_phone(base_url, selenium):
+    page = DevicesPage(base_url, selenium).open()
+    phone = page.open_phone_detail()
+    assert phone.is_displayed
+    assert phone.is_features_displayed
+    phone.close()
```
davehunt: We could also `assert not` for the two previous asserts to check that they're no longer displayed.

tests/functional/firefox/os/test_devices.py
alexgibson: Hmm, we don't do this elsewhere else, as the modal is removed from the DOM once it closes? This is already asserted in the `close()` method as well.

tests/functional/firefox/os/test_devices.py
```diff
@@ -0,0 +1,58 @@
...
+    assert tv.is_features_displayed
+    tv.close()
+
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+def test_toggle_detail_tabs(base_url, selenium):
+    page = DevicesPage(base_url, selenium).open()
+    phone = page.open_phone_detail()
+    assert phone.is_displayed
```
davehunt: Duplicate of previous assert.

tests/functional/firefox/os/test_devices.py
```diff
@@ -0,0 +1,58 @@
...
+    phone.click_specifications()
+    assert phone.is_specifications_displayed
```
davehunt: We should `assert not` for this before clicking the specifications tab.

tests/pages/firefox/os/devices.py
```diff
@@ -0,0 +1,98 @@
...
+
+    _location_select_locator = (By.ID, 'location')
+    _get_phone_button_locator = (By.CSS_SELECTOR, '.location-select-wrapper button')
+    _phone_thumbnail_locator = (By.CSS_SELECTOR, '#smartphones li:first-child > .device-thumbnail')
+    _phone_detail_locator = (By.CSS_SELECTOR, '.smartphone-detail-list .device-detail:first-child')
+    _tv_thumbnail_locator = (By.CSS_SELECTOR, '#tvs li:first-child > .device-thumbnail')
+    _tv_detail_locator = (By.CSS_SELECTOR, '.tvs-detail-list .device-detail:first-child')
+
+    def select_location(self, value):
+        assert self.is_button_disabled, 'Get phone button is already enabled'
```
davehunt: This would fail if a test needed to select a location twice. It's probably not an issue for the test cases you're plannign to write, but I don't think we should include this assert.

tests/pages/firefox/os/devices.py
```diff
@@ -0,0 +1,98 @@
...
+        el = self.selenium.find_element(*self._location_select_locator)
+        Select(el).select_by_visible_text(value)
+        Wait(self.selenium, self.timeout).until(lambda s: not self.is_button_disabled)
+
+    @property
+    def is_button_disabled(self):
+        el = self.selenium.find_element(*self._get_phone_button_locator)
+        return el.get_attribute('disabled') == 'true'
```
davehunt: I would suggest using the `is_enabled()` method provided by Selenium from the element, rather than checking the attribute ourselves. Also, the property name is not clear to be. Could we rename it to `is_get_a_phone_enabled`?

tests/pages/firefox/os/devices.py
```diff
@@ -0,0 +1,98 @@
...
+    def open_modal(self):
+        modal = Modal(self.selenium)
+        self.selenium.find_element(*self._get_phone_button_locator).click()
+        Wait(self.selenium, self.timeout).until(lambda s: modal.is_displayed)
+        return modal
+
+    def open_phone_detail(self):
+        el = self.selenium.find_element(*self._phone_detail_locator)
+        detail = DeviceDetail(self.selenium, root=el)
+        assert not detail.is_displayed, 'Phone detail is already displayed'
```
davehunt: I'm not sure we need this assert. It would prevent a test from opening another phone details when one is already open, which is a valid scenario.

tests/pages/firefox/os/devices.py
```diff
@@ -0,0 +1,98 @@
...
+        self.selenium.find_element(*self._phone_thumbnail_locator).click()
+        # wait until close button is displayed for detail pane to initialize
+        Wait(self.selenium, self.timeout).until(lambda s: detail.is_close_button_displayed)
+        return detail
+
+    def open_tv_detail(self):
+        el = self.selenium.find_element(*self._tv_detail_locator)
+        detail = DeviceDetail(self.selenium, root=el)
+        assert not detail.is_displayed, 'TV detail is already displayed'
```
davehunt: Ditto previous comment.

tests/pages/firefox/os/devices.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @property
+    def is_specifications_displayed(self):
+        return self.is_element_displayed(self._specifications_locator)
+
+    @property
+    def is_close_button_displayed(self):
+        return self.is_element_displayed(self._close_button_locator)
+
+    def click_specifications(self):
```
davehunt: I think `show_specifications` would be a better name here.

tests/pages/firefox/os/devices.py
```diff
@@ -0,0 +1,98 @@
...
+        assert not self.is_specifications_displayed, 'Specifications tab is already displayed'
+        self.root.find_element(*self._spacifications_link_locator).click()
+        Wait(self.selenium, self.timeout).until(lambda s: self.is_specifications_displayed)
+
+    def click_features(self):
```
davehunt: Ditto re: show vs click.

davehunt:
Nice, just a few nits. Let's see if we can land #3562 and rebase this. Otherwise r+ with comments addressed.

tests/functional/firefox/os/test_devices.py
davehunt: Fair enough then. :)

alexgibson:
Thanks @davehunt - updated

tests/functional/firefox/os/test_devices.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+def test_open_close_tv(base_url, selenium):
+    page = DevicesPage(base_url, selenium).open()
+    tv = page.open_tv_detail()
+    assert tv.is_displayed
+    assert tv.is_features_displayed
+    tv.close()
```
davehunt: Could we add `assert not` for these too?

alexgibson: Gah, missed this one thanks

tests/functional/firefox/os/test_devices.py
```diff
@@ -0,0 +1,60 @@
...
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+def test_toggle_detail_tabs(base_url, selenium):
+    page = DevicesPage(base_url, selenium).open()
+    phone = page.open_phone_detail()
+    assert not phone.is_specifications_displayed
+    phone.show_specifications()
+    assert phone.is_specifications_displayed
+    phone.show_features()
+    assert phone.is_features_displayed
```
davehunt: It might be excessive, but we could assert that the specifications is no longer displayed once features is displayed.

tests/pages/firefox/os/devices.py
```diff
@@ -0,0 +1,94 @@
...
+    @property
+    def is_features_displayed(self):
+        return self.is_element_displayed(self._features_locator)
+
+    @property
+    def is_specifications_displayed(self):
+        return self.is_element_displayed(self._specifications_locator)
+
+    @property
+    def is_close_button_displayed(self):
```
davehunt: Nit: Missed this last time, but I think `button` is superfluous in the property name.

davehunt:
Sorry, a couple of nits I missed first time. r+ wc.

alexgibson:
np, updated thanks

PR #3551, opened 2015-11-12
Bug 1221303 - Update pixel id

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1221303

Adding pixel id to both buttons.

### DIFF
--- a/bedrock/firefox/templates/firefox/os/ver/2.5.html
+++ b/bedrock/firefox/templates/firefox/os/ver/2.5.html
@@ -15,7 +15,7 @@
     <h1>{{ _('Experience Firefox OS on your Android device')}}</h1>
     <p>{{_('Introducing Firefox OS 2.5 Developer Preview, an experimental app that lets you use Firefox OS on your Android device.')}}</p>
     <div class="primary-cta">
-      <a href="https://d2yw7jilxa8093.cloudfront.net/B2GDroid-mozilla-central-nightly-latest.apk" class="cta-button" data-id="Mozilla_B2G_Activity_Tag_1">{{_('Get the Android App')}}</a>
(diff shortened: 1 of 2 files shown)


bedrock/firefox/templates/firefox/os/ver/2.5.html
```diff
@@ -15,7 +15,7 @@
     <h1>{{ _('Experience Firefox OS on your Android device')}}</h1>
     <p>{{_('Introducing Firefox OS 2.5 Developer Preview, an experimental app that lets you use Firefox OS on your Android device.')}}</p>
     <div class="primary-cta">
-      <a href="https://d2yw7jilxa8093.cloudfront.net/B2GDroid-mozilla-central-nightly-latest.apk" class="cta-button" data-id="Mozilla_B2G_Activity_Tag_1">{{_('Get the Android App')}}</a>
+      <a href="https://d2yw7jilxa8093.cloudfront.net/B2GDroid-mozilla-central-nightly-latest.apk" class="cta-button" id="52161" data-id="Mozilla_B2G_Activity_Tag_1">{{_('Get the Android App')}}</a>
```
alexgibson: Nit: HTML5 is a bit more relaxed I think, but traditionally id's can't start with a number. Even if this does work OK, I would perhaps still favor using a data attribute for this sort of use case. Something like `data-pixel` perhaps?

malena: sure I can keep it as a data attr to keep it consistent. let me update.

media/js/firefox/os/firefox-os-2.5.js
```diff
@@ -8,6 +8,7 @@
     function trackClick(e) {
         e.preventDefault();
         var id = $(e.target).data('id');
+        var pixelId = $(e.target).attr('id');
```
alexgibson: Seeing as we now need two attributes, we could query the dom once and cache the result e.g.:

    var $elm = $(e.target);
    var id = $elm.data('id');
    var pixel = $elm.data('pixel');

alexgibson:
r+ :dancers:

PR #3553, opened 2015-11-12
[fix bug 1224228] Show App Store badge for non-newsletter locales and iOS users

alexgibson:
We're currently showing the send-to-device widget to locales that don't have the newsletter, so they currently see a blank modal: https://www.mozilla.org/it/firefox/ios/

This PR also shows the App Store badge for iOS users, as it makes more sense to link to it directly.

### DIFF
--- a/media/css/firefox/ios.less
+++ b/media/css/firefox/ios.less
@@ -701,16 +701,32 @@ a.go {
     top: -30px;
 }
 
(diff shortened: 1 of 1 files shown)


alexgibson:
@jpetto r?

jpetto:
r+ :hotsprings:

PR #3557, opened 2015-11-12
Fix Bug 1223211 - https://www.mozilla.org/en-US/about/governance/policies/commit/ add information on how to create a public key

kyoshino:


schalkneethling:
r+ Thanks @kyoshino

PR #3559, opened 2015-11-13
Fix Bug 1223622, add new fundraising promo tile to landing page

schalkneethling:
@alexgibson r? Note @flodolo, not l10n required at this stage - https://bugzilla.mozilla.org/show_bug.cgi?id=1223622#c26


alexgibson:
We use some slightly different classes for touch devices, as we found older devices don't transition visibility. If we're writing some custom CSS for this promo and not using the existing styles, we should probably follow suit here.

media/css/mozorg/home/home-promo.less
```diff
@@ -1066,6 +1066,55 @@ html[lang|="en"] {
...
+    }
+
+    h3 {
+        padding-top: 90px;
+    }
+
+    &.show {
+        a.panel-link {
+            border: none;
+            transition: visibility 0s ease-in-out 0s
```
schalkneethling: @alexgibson I am guessing this is what you are referring to? I guess I am also missing the `-webkit-` prefix version here. Also, it used to have a border transition but, as I am setting border to `none` I removed that.

alexgibson: Take a look at the existing code for reference. Note we also have touch specific CSS classes here which differ slightly https://github.com/mozilla/bedrock/blob/master/media/css/mozorg/home/home-promo.less#L311

alexgibson:
Also could these images be a good candidate for SVG maybe? Not sure what the difference in file size would be but could be worth considering?

schalkneethling:
@alexgibson They definitely could. I will see what file size I can get them to.

media/css/mozorg/home/home-promo.less
schalkneethling: So `has-touch` will be added to this promo by JS right? As is the case for the others. With regards to the above, I could just have `border: none;` and not override the `transition`. Would that prevent any potential problems?

alexgibson: I think that should be fine yeah, the important bit is to make sure your cover the right classes (both `.show` and `.show-touch`, and avoid altering the transitions unless really necessary.

schalkneethling:
@alexgibson The SVG works great! And the file sizes are tiny \o/

media/css/mozorg/home/home-promo.less
```diff
@@ -1066,6 +1066,54 @@ html[lang|="en"] {
...
+        width: 134px;
+        height: 115px;
+    }
+
+    h3 {
+        padding-top: 90px;
+    }
+
+    &.show,
+    &.show-touch {
```
alexgibson: This `.show-touch` doesn't have enough specificity to apply the border override, since the original rule also has `.promo-grid.has-touch` as a parent.

alexgibson:
Looks great, r+ :yen: Care to squash?

PR #3561, opened 2015-11-13
[fix bug 1221739] Add a redirect for Hello's Feedback survey links

alexgibson:
Moved the existing redirect to `firefox/redirects.py` and added tests for both redirects.


pmclanahan:
Perfect! Thanks Alex! :palm_tree:

PR #3562, opened 2015-11-13
Use absolute imports for page objects

davehunt:
This removes the need for the rather ugly relative imports.


alexgibson:
This looks good to me, I think we can squash these two commits?

alexgibson:
r+

PR #3564, opened 2015-11-13
Fix intermittent failures caused by sticky navigation

alexgibson:
https://webqa-ci.mozilla.com/job/bedrock.dev.win10.firefox/87/


alexgibson:
@davehunt r? - not sure if this is the best approach, but seems to work. I was thinking a `ViewPort` class that both `Page` and `PageRegion` could inherit from might work ok, but you probably have a better suggestion :smile:

alexgibson:
Closing in favour of #3566

PR #3565, opened 2015-11-16
Add Python functional tests for /firefox/os/devices/tv page

alexgibson:


alexgibson:
@davehunt r?

tests/functional/firefox/os/test_tv.py
```diff
@@ -0,0 +1,34 @@
...
+    page = TVPage(base_url, selenium).open()
+    assert not page.is_previous_enabled
+    page.click_next()
+    screen = page.wait_for_screen_to_display('tv2')
+    assert page.is_previous_enabled
+    assert screen.is_displayed
+    page.click_previous()
+    screen = page.wait_for_screen_to_display('tv1')
+    assert not page.is_previous_enabled
+    assert screen.is_displayed
```
davehunt: Instead of maintaining a dictionary of the screens, I would propose having `screens` and `thumbnails` as properties. These would be regions for interacting with and determining if they're displayed. The test would then look something like:

```python
def test_next_previous_buttons(base_url, selenium):
    page = TVPage(base_url, selenium).open()
    assert page.screen[0].is_displayed
    assert not page.is_previous_enabled
    page.click_next()
    assert page.screen[1].is_displayed
    page.click_previous()
    assert not page.is_previous_enabled
    assert page.screen[0].is_displayed
```

We could then even add coverage by testing all screens:

```python
def test_next_previous_buttons(base_url, selenium):
    page = TVPage(base_url, selenium).open()
    assert not page.is_previous_enabled
    for screen in page.screens[:-1]:
        assert screen.is_displayed
        page.click_next()
    assert not page.is_next_enabled
    for screen in reversed(page.screens)[:-1]:
        screen.is_displayed
        page.click_previous()
    assert not page.is_previous_enabled
    assert page.screens[0].is_displayed
```

This could similarly apply to the thumbnails. If you used `for i in range(len(page.screens[:-1]))` then you could reference both screens and thumbnails by a common index, allowing you to check the correct thumbnail is highlighted when using the next/previous buttons, and vice versa.

alexgibson:
@davehunt - thanks for the useful suggestions, updated

alexgibson:
@davehunt - one thing I noticed I'm not doing quite right is reversing the array of page objects. The tests pass, but I think I'm iterating over one too many times still.

alexgibson:
Ok, think I got it.

tests/functional/firefox/os/test_tv.py
```diff
@@ -0,0 +1,46 @@
...
+    page = TVPage(base_url, selenium).open()
+    assert not page.is_previous_enabled
+    screens = page.screens
+    thumbnails = page.thumbnails
+    for i in range(len(screens))[:-1]:
+        assert screens[i].is_displayed
+        assert thumbnails[i].is_selected
+        page.click_next()
+    assert not page.is_next_enabled
+    for i in range(len(screens))[::-1][:-1]:
```
davehunt: Yeah, it seems `reversed` can be used but it's not as clean as I'd hoped. I think we can make this a little cleaner though by using the start, stop, step arguments for `range`. I've testing this locally:

```python
for i in range(len(screens) - 1):
    assert screens[i].is_displayed
    assert thumbnails[i].is_selected
    page.click_next()
assert not page.is_next_enabled
for i in range(len(screens) - 1, 0, -1):
    assert screens[i].is_displayed
    assert thumbnails[i].is_selected
    page.click_previous()
```

tests/functional/firefox/os/test_tv.py
```diff
@@ -0,0 +1,46 @@
...
+@pytest.mark.nondestructive
+def test_click_thumbnails(base_url, selenium):
+    page = TVPage(base_url, selenium).open()
+    screens = page.screens
+    thumbnails = page.thumbnails
+    for i in range(len(thumbnails))[:-1]:
+        assert screens[i].is_displayed
+        assert thumbnails[i].is_selected
+        thumbnails[i + 1].click()
+    for i in range(len(screens))[::-1][:-1]:
```
davehunt: I think this should be `thumbnails`, but I'd also suggest click and then check, so we can use the index unmodified. Also, using start, stop, range works well here too:

```python
assert screens[0].is_displayed
assert thumbnails[0].is_selected
for i in range(1, len(thumbnails)):
    thumbnails[i].click()
    assert screens[i].is_displayed
    assert thumbnails[i].is_selected
for i in range(len(thumbnails) - 1, -1, -1):
    thumbnails[i].click()
    assert screens[i].is_displayed
    assert thumbnails[i].is_selected
assert screens[0].is_displayed
assert thumbnails[0].is_selected
```

tests/pages/firefox/os/tv.py
```diff
@@ -0,0 +1,65 @@
...
+class Thumbnails(FirefoxBasePageRegion):
+
+    _link_locator = (By.TAG_NAME, 'a')
+
+    @property
+    def is_selected(self):
+        return self._root.get_attribute('class') == 'selected'
+
+    def click(self):
+        self._root.click()
```
davehunt: I think we should try to avoid calling `_root`, although I do think we may need to have a way to interact with the root element. In most cases the root element would be a container and not something we'd directly interact with. In this case, looking at the page source, I think the `Thumbnails` root locator would be `.pager-tabs li` and then the `_link_locator` above would be used in the `click` and `is_selected` methods.

alexgibson:
Thanks @davehunt updated

tests/pages/firefox/os/tv.py
```diff
@@ -0,0 +1,65 @@
...
+
+class Screens(FirefoxBasePageRegion):
+
+    _image_locator = (By.TAG_NAME, 'img')
+    _next_button_locator = (By.CSS_SELECTOR, '.pager-next')
+    _previous_button_locator = (By.CSS_SELECTOR, '.pager-prev')
+
+    @property
+    def is_displayed(self):
+        self.wait.until(lambda s: self.is_element_displayed(self._image_locator))
```
davehunt: We shouldn't wait for the element to be displayed as it would cause a timeout if we wanted to verify that the element is not displayed. Instead we should move this wait into the `click_next` and `click_previous` methods. The wait would probably need to be for the current screen to not be displayed after the click. The following seems to work well for me:

```python
screen = next(s for s in self.screens if s.is_displayed)
self.find_element(self._next_button_locator).click()
self.wait.until(lambda s: not screen.is_displayed)
```

tests/pages/firefox/os/tv.py
```diff
@@ -0,0 +1,65 @@
...
+    @property
+    def is_previous_enabled(self):
+        return self.find_element(self._previous_button_locator).is_enabled()
+
+
+class Screens(FirefoxBasePageRegion):
+
+    _image_locator = (By.TAG_NAME, 'img')
+    _next_button_locator = (By.CSS_SELECTOR, '.pager-next')
+    _previous_button_locator = (By.CSS_SELECTOR, '.pager-prev')
```
davehunt: We don't need these next/previous locators here.

tests/pages/firefox/os/tv.py
davehunt: It probably also makes sense for us to wait for `is_selected` after clicking.

tests/pages/firefox/os/tv.py
```diff
@@ -0,0 +1,65 @@
...
+    @property
+    def thumbnails(self):
+        return [Thumbnails(self.base_url, self.selenium, root=el) for el in
+                self.find_elements(self._thumbnails_locator)]
+
+    def click_next(self):
+        self.find_element(self._next_button_locator).click()
+
+    def click_previous(self):
+        self.find_element(self._previous_button_locator).click()
```
davehunt: I think these methods should be `show_next_screen` and `show_previous_screen` so we're describing the intent and not the specific interaction.

tests/functional/firefox/os/test_tv.py
```diff
@@ -0,0 +1,48 @@
...
+    page = TVPage(base_url, selenium).open()
+    screens = page.screens
+    thumbnails = page.thumbnails
+    assert screens[0].is_displayed
+    assert thumbnails[0].is_selected
+    for i in range(1, len(thumbnails)):
+        page.click_thumbnail(i)
+        assert screens[i].is_displayed
+        assert thumbnails[i].is_selected
+    for i in range(len(thumbnails) - 2, -1, -1):
```
davehunt: With this as `-2` wouldn't we skip a thumbnail?

tests/pages/firefox/os/tv.py
```diff
@@ -0,0 +1,72 @@
...
+        screen = next(s for s in self.screens if s.is_displayed)
+        self.find_element(self._next_button_locator).click()
+        self.wait.until(lambda s: not screen.is_displayed)
+
+    def show_previous_screen(self):
+        screen = next(s for s in self.screens if s.is_displayed)
+        self.find_element(self._previous_button_locator).click()
+        self.wait.until(lambda s: not screen.is_displayed)
+
+    def click_thumbnail(self, i):
```
davehunt: Nit: Could we name the argument `index` for clarity?

davehunt: I wonder if we should add a comment that this method is necessary because the thumbnail region does not have a reference to the page. Perhaps not, but it might help us to remember to clean this up if we add that ability.

tests/functional/firefox/os/test_tv.py
alexgibson: We we're clicking the last thumbnail twice before i think, but it went unnoticed as now we're waiting for the current screen not to be displayed.

davehunt:
I like it, thanks for your patience @alexgibson! :cake:

tests/pages/firefox/os/tv.py
alexgibson: will do :+1:

alexgibson:
Thanks for the excellent review @davehunt, updated and squashed. If all looks good I'll merge this shortly.

davehunt:
r+

PR #3566, opened 2015-11-16
Refactor page and region objects under a common base class

davehunt:
Also introduces a method for scrolling elements into view, which avoids the family header navigation on affected pages.

### DIFF
--- a/tests/pages/firefox/base.py
+++ b/tests/pages/firefox/base.py
@@ -4,18 +4,23 @@
 
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support import expected_conditions as expected
-from selenium.webdriver.support.ui import WebDriverWait as Wait
 
 from pages.base import BasePage
 from pages.page import PageRegion
 
+# this should be enough to scroll elements so they are not beneath the header
+HEADER_OFFSET = -100
+
 
 class FirefoxBasePage(BasePage):
 
-    def open_family_navigation_menu(self):
-        navigation = self.FamilyNavigation(self.selenium)
-        navigation.open_menu()
-        return navigation
+    @property
+    def family_navigation(self):
+        return self.FamilyNavigation(self.base_url, self.selenium)
+
+    def scroll_element_into_view(self, locator):
+        return super(FirefoxBasePage, self).scroll_element_into_view(
+            locator, y=HEADER_OFFSET)
 
     class FamilyNavigation(PageRegion):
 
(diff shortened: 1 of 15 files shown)


davehunt:
Would you mind taking a look over this @bobsilverberg? It's likely to heavily influence any common page object model package that we distribute in the future. There's a lack of documentation, but I think most of it is pretty straightforward. The part I'm least sure on it the expansion of any kwargs in the URL. While this is powerful, I think it's not particularly obvious what's happening.

alexgibson:
Note: I've temporarily disabled the /os/devices/ tests that are effected by the sticky header until this PR merges, as they are failing in the bedrock pipeline https://github.com/mozilla/bedrock/pull/3567

bobsilverberg:
@davehunt I am looking now. We should find some time to discuss it as there's a lot in there, most of which I like. I was surprised by at least one thing though, the `find_element` and `find_elements` methods, which I actually really like, but started removing from some projects as it felt like re-implementing the selenium api.

davehunt:
I ran the entire suite locally 10 times, and one test is failing 50% of the time. I'll take a look into this and see if I can stabilize it.

davehunt:
> We should find some time to discuss it as there's a lot in there, most of which I like. I was surprised by at least one thing though, the `find_element` and `find_elements` methods, which I actually really like, but started removing from some projects as it felt like re-implementing the selenium api.

Yep, agreed. I surprised myself by adding this, but it does make sense when we're sharing a common parent for the Page and PageRegion objects.

davehunt:
I've addressed the remaining intermittent. This seems stable for me locally now.

alexgibson:
@davehunt - I'm still seeing one test fail for me consistently: `FAIL tests/functional/firefox/test_do_not_track.py::test_open_and_close_faq_panel`

alexgibson:
Could you also please rebase your branch and remove the skipped tests for the device page? Thanks!

davehunt:
> @davehunt - I'm still seeing one test fail for me consistently: `FAIL tests/functional/firefox/test_do_not_track.py::test_open_and_close_faq_panel`

Could you provide details of the failure you're seeing?

davehunt:
> Could you also please rebase your branch and remove the skipped tests for the device page? Thanks!

Done.

alexgibson:
> Could you provide details of the failure you're seeing?

Here you go: https://gist.github.com/alexgibson/8290be6f5876cca70c85

davehunt:
Pushed a fix for the `test_open_and_close_faq_panel` failure.

alexgibson:
This looks good to me and tests pass locally, r+ :whale:

tests/pages/firefox/base.py
```diff
@@ -24,10 +29,16 @@ class FamilyNavigation(PageRegion):
...
         @property
-        def is_displayed(self):
+        def is_menu_displayed(self):
             return self.is_element_displayed(self._nav_menu_locator)
+
+
+class FirefoxBasePageRegion(PageRegion):
+
+    def scroll_element_into_view(self, locator):
+        return super(FirefoxBasePageRegion, self).scroll_element_into_view(
```
bobsilverberg: It's too bad we have to repeat this. I think I understand why it's required, but I wonder if there's another solution? Maybe something like a mixin?

tests/pages/firefox/developer.py
```diff
@@ -28,14 +27,14 @@ def is_secondary_download_button_displayed(self):
...
 
 
 class Video(PageRegion):
 
     def play(self):
-        modal = Modal(self.selenium)
-        self.root.click()
-        Wait(self.selenium, self.timeout).until(lambda s: modal.is_displayed)
+        modal = Modal(self.base_url, self.selenium)
+        self._root.click()
```
bobsilverberg: I'm wondering why this is just about the only place we need to refer to `self._root` in a page object. What's special about this and can we avoid it?

PR #3569, opened 2015-11-17
Bug 1217971 - Add lang file directive to firefox/choose

flodolo:
This lets us override strings coming from new.lang (loaded in base-resp.html). Thanks to @pascalchevrel  for the idea.

### DIFF
--- a/bedrock/firefox/templates/firefox/choose.html
+++ b/bedrock/firefox/templates/firefox/choose.html
@@ -4,6 +4,8 @@
 
 {% extends "firefox/base-resp.html" %}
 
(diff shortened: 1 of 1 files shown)


flodolo:
@alexgibson Can you double check? When loading firefox/choose in German, you should see the "Übernimm die Kontrolle. Hol dir Firefox." as headline instead of "Übenehmen Sie die Kontrolle. Entscheiden Sie sich für Firefox."

alexgibson:
I can confirm this works locally, r+ :spaghetti:

PR #3571, opened 2015-11-17
Removes period from after Dec on eoy homepage promo

schalkneethling:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1214038#c12 - @jpetto r?


craigcook:
This string hasn't yet been extracted for l10n so we're safe to change it. r+ :tropical_drink:

PR #3573, opened 2015-11-18
Fix Bug 1225737, shorten tweet text on cybersecurity and wrap in en-US conditional

schalkneethling:
@alexgibson r?

### DIFF
--- a/bedrock/teach/templates/teach/smarton/security.html
+++ b/bedrock/teach/templates/teach/smarton/security.html
@@ -14,8 +14,15 @@
 {% block smarton_content %}
 <article class="main">
 
+  {# see bug 1188591, comment 10 for details regarding the need for the en-* conditional #}
+  {% if LANG.startswith('en-') %}
+    {% set tweet_text = _('Stay updated, and make sure your passwords are as buff as you are. Get smart on cybersecurity.') %}
+  {% else %}
+    {% set tweet_text = _('Look both ways, stay updated, and make sure your passwords are as buff as you are. Get smart on cybersecurity.') %}
+  {% endif %}
+
   {# Navbar macro. Params: Page title, tweet text, Twitter URL, Facebook URL, Google+ URL #}
-  {{ smarton_navbar(_('Cybersecurity'), _('Look both ways, stay updated, and make sure your passwords are as buff as you are. Get smart on cybersecurity.') + hashtag, 'http://mzl.la/cybersecurity', 'https://www.mozilla.org/teach/smarton/security/', 'https://www.mozilla.org/teach/smarton/security/') }}
+  {{ smarton_navbar(_('Cybersecurity'), tweet_text + hashtag, 'http://mzl.la/cybersecurity', 'https://www.mozilla.org/teach/smarton/security/', 'https://www.mozilla.org/teach/smarton/security/') }}
 
   <section id="ask" class="section section-ask">
     <div class="content">


bedrock/teach/templates/teach/smarton/security.html
```diff
@@ -15,7 +15,12 @@
 <article class="main">
 
   {# Navbar macro. Params: Page title, tweet text, Twitter URL, Facebook URL, Google+ URL #}
-  {{ smarton_navbar(_('Cybersecurity'), _('Look both ways, stay updated, and make sure your passwords are as buff as you are. Get smart on cybersecurity.') + hashtag, 'http://mzl.la/cybersecurity', 'https://www.mozilla.org/teach/smarton/security/', 'https://www.mozilla.org/teach/smarton/security/') }}
+  {# see bug 1188591, comment 10 for details regarding the need for the en-US conditional #}
+  {% if LANG == 'en-US' %}
```
alexgibson: Should we not shorten the string for all en locales and not just en-US?

flodolo: Not really, given that the page is already live (I sent an email yesterday to warn localizers about it). Unless we want to add this as an alternative string behind a tag.

alexgibson: Note I said all `en` locales, not all locales. Will the text not be too long there too?

flodolo: Blame lack of coffee :coffee: 
Yes, extending to en-* makes sense

bedrock/teach/templates/teach/smarton/security.html
```diff
@@ -15,7 +15,12 @@
...
+    {{ smarton_navbar(_('Cybersecurity'), _('Stay updated, and make sure your passwords are as buff as you are. Get smart on cybersecurity.') + hashtag, 'http://mzl.la/cybersecurity', 'https://www.mozilla.org/teach/smarton/security/', 'https://www.mozilla.org/teach/smarton/security/') }}
```
alexgibson: Could we also just change the string based on conditional, instead of needing to duplicate the macro code?

schalkneethling: I was wondering about that as well, not entirely sure how to accomplish that, unless it is done the same way as the hashtag perhaps?

bedrock/teach/templates/teach/smarton/security.html
schalkneethling: As I was doing this I was wondering the same thing as we ran into a similar scenario in the not to distant past.

bedrock/teach/templates/teach/smarton/security.html
alexgibson: Yep using `set` should work

bedrock/teach/templates/teach/smarton/security.html
```diff
@@ -14,8 +14,11 @@
 {% block smarton_content %}
 <article class="main">
 
+  {# see bug 1188591, comment 10 for details regarding the need for the en-* conditional #}
+  {% set tweet_text = _('Stay updated, and make sure your passwords are as buff as you are. Get smart on cybersecurity.') if LANG.startswith('en-') else _('Look both ways, stay updated, and make sure your passwords are as buff as you are. Get smart on cybersecurity.') %}
```
alexgibson: Nit: would this be a little more readable on more than one line?

      {% set tweet_text = _('Stay updated, and make sure your passwords are as buff as you are. Get smart on cybersecurity.') %}
    {% else %}
      {% set tweet_text = _('Look both ways, stay updated, and make sure your passwords are as buff as you are. Get smart on cybersecurity.') %}
    {% endif %}

alexgibson:
One minor suggestion, which is up to you. r+wc :bear:

bedrock/teach/templates/teach/smarton/security.html
schalkneethling: Yeah, I prefer that as well.

PR #3575, opened 2015-11-18
Add main site navigation tests for desktop and mobile views

davehunt:
This replaces the following tests from https://github.com/mozilla/mcom-tests:
* [test_about > test_navbar_links_are_visible](https://github.com/mozilla/mcom-tests/blob/c9eacd8f62b11d4a2ab70a230db779932235c801/tests/test_about.py#L36)
* [test_about > test_navbar_link_destinations_are_correct](https://github.com/mozilla/mcom-tests/blob/c9eacd8f62b11d4a2ab70a230db779932235c801/tests/test_about.py#L45)
* [test_about > test_navbar_link_urls_are_valid](https://github.com/mozilla/mcom-tests/blob/c9eacd8f62b11d4a2ab70a230db779932235c801/tests/test_about.py#L56)
* [test_mozillabased > test_navbar_links_are_visible](https://github.com/mozilla/mcom-tests/blob/c9eacd8f62b11d4a2ab70a230db779932235c801/tests/test_mozillabased.py#L92)

The main differences are:
* These tests check the main navigation links from the __home__ page instead of the __about__ and __mozilla-based__ pages. We could easily parameterise these tests if we wanted to run them from multiple initial pages.
* The tests these replace were checking the links were visible, destination URLs were correct, and destination URLs responded with an OK response code. The new tests click the links (would throw if they were not visible), and check the destination URL is accurate for the expected page.
* The new tests also check the navigation for small screens by restricting the viewport.

### DIFF
--- a/tests/functional/test_navigation.py
+++ b/tests/functional/test_navigation.py
@@ -1,0 +1,40 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.home import HomePage
+
+
+@pytest.mark.nondestructive
+def test_navigation(base_url, selenium):
+    page = HomePage(base_url, selenium).open()
+    about_page = page.navigation.open_about()
+    assert about_page.url in selenium.current_url
+
+    page.open()
+    participate_page = page.navigation.open_participate()
+    assert participate_page.url in selenium.current_url
+
+    page.open()
+    firefox_page = page.navigation.open_firefox()
+    assert firefox_page.url in selenium.current_url
+
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+@pytest.mark.viewport('mobile')
+def test_mobile_navigation(base_url, selenium):
+    page = HomePage(base_url, selenium).open()
+    page.navigation.show()
+    about_page = page.navigation.open_about()
+    assert about_page.url in selenium.current_url
+
+    page.open().navigation.show()
+    participate_page = page.navigation.open_participate()
+    assert participate_page.url in selenium.current_url
+
+    page.open().navigation.show()
+    firefox_page = page.navigation.open_firefox()
+    assert firefox_page.url in selenium.current_url
--- a/tests/pages/base.py
+++ b/tests/pages/base.py
@@ -22,6 +22,10 @@ class BasePage(Page):
         return self
 
     @property
+    def navigation(self):
+        return self.Navigation(self.base_url, self.selenium)
+
+    @property
     def footer(self):
         return self.Footer(self.base_url, self.selenium)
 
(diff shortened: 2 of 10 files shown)


davehunt:
@alexgibson r?

tests/pages/base.py
```diff
@@ -33,6 +37,47 @@ def newsletter(self):
...
+        def show(self):
+            assert not self.is_displayed, 'Menu is already displayed'
+            self.find_element(self._toggle_locator).click()
+            self.wait.until(lambda s: self.is_displayed)
+            return self
+
+        @property
+        def is_displayed(self):
+            menu = self.find_element(self._menu_locator)
+            return menu.is_displayed() and menu.size['height'] == 177
```
alexgibson: Is there a reason we need to check the height of the menu here? Technically this could change depending on the string/translation.

tests/functional/test_navigation.py
```diff
@@ -0,0 +1,47 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.home import HomePage
+
+
+@pytest.mark.nondestructive
```
alexgibson: Should we add this as a santiy test?

tests/functional/test_navigation.py
```diff
@@ -0,0 +1,47 @@
...
+    page.open()
+    firefox_page = page.navigation.open_firefox()
+    assert firefox_page.url in selenium.current_url
+
+    page.open()
+    donate_page = page.navigation.open_donate()
+    assert donate_page.url in selenium.current_url
+
+
+@pytest.mark.nondestructive
```
alexgibson: Likewise, could this be a sanity test?

tests/functional/test_navigation.py
davehunt: Sounds like a good idea.

tests/pages/donate.py
```diff
@@ -0,0 +1,10 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from pages.page import Page
+
+
+class DonatePage(Page):
+
+    _url = 'https://donate.mozilla.org/'
```
alexgibson: While I understand why this is here, it seems a bit odd to have a page object for something that's not part of our site. Maybe this is the easiest solution though?

tests/pages/base.py
davehunt: Ah, yes! I'd love to have a better indication that the menu is showing. I can't use `is_displayed` because then it tries clicking while it's still animating. I _could_ wait for the height to stop changing, but ideally I'd wait for something more concrete. Do you have any suggestions?

tests/functional/test_navigation.py
davehunt: Just the desktop one, or both?

alexgibson: I'd say both?

tests/pages/donate.py
davehunt: We could have the URL in the tests if that's preferred? I don't have a preference, and your concern is valid.

tests/pages/base.py
alexgibson: Looking at the code, we add `aria-expanded="true"` to the `_toggle_locator` element once the menu opens. Perhaps we could use this?

tests/pages/donate.py
alexgibson: I think seeing as this is an external page let's put this URL in the test and avoid having this file here? Perhaps add a comment to the test for clarity?

tests/pages/base.py
davehunt: It looks like this attribute is added before the menu is open, as this is failing for me in 2/5 tests.

alexgibson: I wonder if we can adjust the code here to make testing easier, I'll take a look

alexgibson: Opened a PR which should hopefully make things easier: https://github.com/mozilla/bedrock/pull/3577

alexgibson:
@davehunt - re: sanity tests, given that these take some time to complete (and one link is also external), perhaps it would be better to have a single santity test that just opens/closes the mobile menu, but doesn't traverse all the links. Thoughts?

davehunt:
> re: sanity tests, given that these take some time to complete (and one link is also external), perhaps it would be better to have a single santity test that just opens/closes the mobile menu, but doesn't traverse all the links. Thoughts?

I agree that we should probably avoid depending on an external site for our sanity tests. Do we even need to check the donate menu item given that we're planning to introduce a much more thorough link checker in the near future? I'd suggest dropping the donate check from both tests, and marking just the mobile test as sanity.

alexgibson:
> I agree that we should probably avoid depending on an external site for our sanity tests. Do we even need to check the donate menu item given that we're planning to introduce a much more thorough link checker in the near future? I'd suggest dropping the donate check from both tests, and marking just the mobile test as sanity.

Ok, sounds reasonable to me :+1:

davehunt:
Rebased and ready for review again. Note that we need #3577 to land before we merge this, but after review I can squash and prepare for the merge.

alexgibson:
LGTM r+ pending #3577 

Let's squash & wait for the other PR to merge. Will give these tests a final run once that's in master :+1:

davehunt:
Squashed and rebased!

alexgibson:
r+ :fries:

PR #3576, opened 2015-11-18
Allow LinkChecker to crawl the site for broken links

davehunt:
This change allows the [LinkChecker](http://wummel.github.io/linkchecker/) tool to run against the site and report any broken links. This is much more thorough than the link checking performed in https://github.com/mozilla/mcom-tests/ and can be run as a separate scheduled job that would not necessarily block a deploy due to the unavailability of external sites.

With this change I'm able to run LinkChecker locally (still going strong after 20 minutes). If this is merged and deployed to any environment I'd be happy to set up a Jenkins job to run once a week (alternative suggestions welcome) and send the results to whomever may be interested (again, suggestions?). I'd then like to use this as an opportunity to remove all the link checking from mcom-tests. :link: :smile:

/cc @alexgibson @jgmize @pmclanahan

### DIFF
--- a/bedrock/mozorg/templates/mozorg/robots.txt
+++ b/bedrock/mozorg/templates/mozorg/robots.txt
@@ -1,3 +1,6 @@
+user-agent: LinkChecker
+disallow:
+crawl-delay: 1
 user-agent: *
 {% if disallow_all -%}
 disallow: /


davehunt:
I ran this locally with 4 threads (the default of 10 was too high and caused timeouts). I ignored links to humans.txt as this is not available locally, and I excluded checking external sites. Here's the HTML report: http://people.mozilla.org/~dhunt/linkchecker.html

> That's it. 105975 links checked. 10 warnings found. 41 errors found.
Stopped checking at 2015-11-18 22:56:17+001 (52 minutes, 59 seconds) 

I suspect a lot of the failures are related to redirects that wouldn't occur on production (I tested a couple and they seemed okay to me). The whitespace warnings are interesting, and probably relatively easy to fix.

I also looked into disabling the robots.txt checking in the tool. This has been requested as an enhancement to the tool, but the author is concerned with it leading to the tool being blacklisted. He suggests he may be interested in a patch that modifies the user agent when ignoring robots.txt so that any blacklisting would avoid affecting the default use of the tool. It's been over a year since any change landed, though there are recently active issues and pull requests.

I suspect the easiest thing is for us to allow the tool to crawl the site in robots.txt as done in this patch.

pmclanahan:
My first reaction is that I'd rather not alter the site for this tool. I appreciate that it's a good one, but it seems unmaintained and therefore likely that we'll need to find something else soon regardless. If it's not too tough to remove the robots.txt checking in the tool itself then I recommend we fork and modify.

pmclanahan:
We further discussed this in our weekly meeting and decided to go with this. We can always fork the tool later if this doesn't work out for some reason. Thanks again Dave.

davehunt:
@pmclanahan Do we know when this is likely to land in production? I've been watching https://bedrock-prod-deis.us-west.moz.works/robots.txt and don't see the changes yet.

pmclanahan:
doing a prod push this morning. should be there soon.

jgmize:
```sh
curl -k -H "Host: www.mozilla.org" https://$(dig +short bedrock-prod.us-west.moz.works | head -n 1)/robots.txt
```
now gives me the expected results:
```
user-agent: LinkChecker
disallow:
crawl-delay: 1
user-agent: *
disallow: /*/download/
disallow: /*/firstrun/
disallow: /*/newsletter/existing/
disallow: /*/whatsnew/
disallow: /*/firefox/os/1.1/
disallow: /*/firefox/os/1.3/
disallow: /*/firefox/os/1.3t/
disallow: /*/firefox/os/1.4/
disallow: /b/
```

Note that https://bedrock-prod-deis.us-west.moz.works/robots.txt gives different results:
```
user-agent: LinkChecker
disallow:
crawl-delay: 1
user-agent: *
disallow: /
```

This is also expected, because the view [checks the host header](https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/views.py#L533) to prevent robots from crawling sites other than www.mozilla.org.

PR #3577, opened 2015-11-18
[fix bug 1225896] Adjust mobile nav menu aria state timing

alexgibson:
This is a small change to make functional testing easier (see: https://github.com/mozilla/bedrock/pull/3575#discussion_r45218569)

### DIFF
--- a/media/js/base/nav-main-resp.js
+++ b/media/js/base/nav-main-resp.js
@@ -414,11 +414,9 @@ NavMain.openSmallMenu = function()
         return;
     }
 
-    $('#nav-main-menu')
-        .slideDown(150)
-        .removeAttr('aria-hidden');
-
-    NavMain.toggleButton.addClass('open').attr('aria-expanded', true);
+    $('#nav-main-menu').slideDown(150, function() {
+        NavMain.toggleButton.addClass('open').attr('aria-expanded', true);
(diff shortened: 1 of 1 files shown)


alexgibson:
@davehunt - does this change work for you ok?

davehunt:
@alexgibson yep this is now passing 100% of the time, thanks!

alexgibson:
Great, @jpetto r just to make sure I haven't fudged anything else here?

media/js/base/nav-main-resp.js
```diff
@@ -414,11 +414,9 @@ NavMain.openSmallMenu = function()
...
     }
 
-    $('#nav-main-menu')
-        .slideDown(150)
-        .removeAttr('aria-hidden');
-
-    NavMain.toggleButton.addClass('open').attr('aria-expanded', true);
+    $('#nav-main-menu').slideDown(150, function() {
+        NavMain.toggleButton.addClass('open').attr('aria-expanded', true);
+    }).removeAttr('aria-hidden');
```
alexgibson: Ah, yes good suggestion. We don't need it for this test directly, but I think it makes sense to make sure both functions match in terms of timing.

alexgibson:
Thanks @jpetto - updated

PR #3581, opened 2015-11-19
Bug 1182225 release notes redesign - m

malena:
Bug 1182225 - Change Front-End of Release Notes


sgarrity:
I've replaced this PR with https://github.com/mozilla/bedrock/pull/3581 based on the same commits but with some additional updates.

PR #3582, opened 2015-11-19
Fix bug 1226329: Improve x-backend-server header

pmclanahan:
This will turn on x-backend-server for every instance with a DEIS_APP
env var (which should be all of them running on Deis). It also adds the
value of said var to the header response. Also if DEIS_DOMAIN env var is set
that will be added to the end of that.


pmclanahan:
@jgmize r?

jgmize:
Nice work, @pmclanahan

PR #3584, opened 2015-11-20
Bug 1226536 - Private browsing page: add non-breaking space before external link decorator

flodolo:


alexgibson:
We could have just added a CSS class of `more` to the HTML link to to get the carat auto-magically instead of repeat the CSS. But this still works, plus cleans up some whitespace :smile: r+

PR #3585, opened 2015-11-20
Add Python functional tests for styleguide

alexgibson:
Note: I'm only testing the menu interaction on the styleguide, as the pages are all really just static content/links.


alexgibson:
@davehunt r?

tests/pages/styleguide.py
```diff
@@ -0,0 +1,46 @@
...
+    _link_locator = (By.CLASS_NAME, 'toggle')
+    _menu_locator = (By.CSS_SELECTOR, 'ul > li')
+    _sub_menu_locator = (By.CSS_SELECTOR, 'ul > .has-children')
+
+    @property
+    def sub_menu(self):
+        return [NavigationMenu(self.base_url, self.selenium, root=el) for el in
+                self.find_elements(self._sub_menu_locator)]
+
+    def show_menu(self):
```
davehunt: Could we rename this to `expand`? I think it makes the test a little more readable.

tests/pages/styleguide.py
```diff
@@ -0,0 +1,46 @@
...
+        assert not self.is_displayed, 'Menu is already displayed'
+        self.find_element(self._link_locator).click()
+        self.wait.until(lambda m: self._root.get_attribute('aria-expanded') == 'true')
+
+    def hide_menu(self):
```
davehunt: As above, but `collapse`?

davehunt:
r+ with comments for the test and page object

alexgibson:
@schalkneethling r on the front-end changes? I did a little tidy up and added an aria state to the styleguide menu (better for a11y, but also makes testing a little more robust).

schalkneethling:
r+ :radio:

PR #3592, opened 2015-11-22
Bug 1226263 - Firefox Student Ambassadors Form Validation

alexlovesprogramming:
Make sure user can't submit invalid first/last names.

### DIFF
--- a/tests/functional/test_navigation.py
+++ b/tests/functional/test_navigation.py
@@ -7,6 +7,7 @@ import pytest
 from pages.home import HomePage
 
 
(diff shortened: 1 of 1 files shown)


schalkneethling:
We cannot enforce this:

1) There are cases such as María del Pilar where the surname does not start with a capital
2) If validation fails based on the above, how will we notify users with a11y problems that the reason the form submission failed was because the name or surname did not start with a capital letter?

The current message simply states "Names cannot begin with a lowercase letter."

Perhaps this is an oversight on validating both the name and surname with the same rule?

PR #3593, opened 2015-11-23
Fix Bug 1225275, swap pb and ios, update color for ios and gear

schalkneethling:
@alexgibson r?


alexgibson:
Not directly part of this change, but there's also a waffle switch here we can remove: https://github.com/schalkneethling/bedrock/blob/bug1225275-homepage-promo-tile-cleanup-non-l10n/bedrock/mozorg/templates/mozorg/home/home.html#L221

r+wc :sweet_potato:

schalkneethling:
@alexgibson updated

PR #3594, opened 2015-11-23
[fix bug 1218451] Skip newsletter form submission tests on stage

alexgibson:
@davehunt r?


alexgibson:
After discussion with @davehunt we're going to look at explore repeating these tests before failing the run, as opposed to skipping them on stage entirely. Closing this for now

PR #3595, opened 2015-11-23
Bug 1224211. Hid the horizontal scroll bar.

alexlovesprogramming:


alexlovesprogramming:
r?alexgibson

alexgibson:
Thanks for the PR @alexlovesprogramming - I've taken a look at your PR and this does indeed hide the scroll bar :+1: 

Could you please just add `overflow-x: hidden;` to the [existing html selector on line 7](https://github.com/mozilla/bedrock/blob/master/media/css/firefox/nightly_firstrun.less#L7), instead of creating a duplicate? Then I think we're good to merge this change, thanks!

alexgibson:
r+ thanks @alexlovesprogramming! :cake:

PR #3596, opened 2015-11-23
Wait for the page URL to be in the current URL when loading pages

davehunt:
@alexgibson r?


alexgibson:
LGTM, let's see how this goes, r+

PR #3598, opened 2015-11-23
Bug 1227199 - Put EOY fundraising tile behind a tag

flodolo:


flodolo:
@schalkneethling r?

@m-alexis Are you fine with merging this? We still need to fix the URLs, but at least localizers can see what they're localizing

m-alexis:
@flodolo yeah ok to merge to get localization started.

PR #3600, opened 2015-11-24
Add Python functional tests for Firefox OS pages

alexgibson:


alexgibson:
@davehunt r?

tests/functional/firefox/os/version/test_base.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@pytest.mark.nondestructive
+@pytest.mark.parametrize('version', [('2.0'), ('1.4'), ('1.3t'), ('1.3'), ('1.1')])
+def test_news_is_displayed(version, base_url, selenium):
+    page = FirefoxOSBasePage(base_url, selenium, version=version).open()
+    assert page.is_news_displayed
+
+
+@pytest.mark.nondestructive
```
alexgibson: Note to self: this one should be a sanity test

tests/functional/firefox/os/version/test_2_0.py
```diff
@@ -0,0 +1,17 @@
...
+
+from pages.firefox.os.version.v2_0 import FirefoxOSPage
+
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+def test_app_category_selector(base_url, selenium):
+    page = FirefoxOSPage(base_url, selenium).open()
+    app_groups = page.app_groups
+    for group in app_groups:
```
davehunt: We can just do `for group in page.app_groups` without the need for the `app_groups` variable.

tests/functional/firefox/os/version/test_2_0.py
```diff
@@ -0,0 +1,17 @@
...
+        group.select()
+        assert group.is_active
```
davehunt: Do we need to check that a subset of the apps are faded out?

tests/functional/firefox/os/version/test_base.py
```diff
@@ -0,0 +1,33 @@
...
+    page = FirefoxOSBasePage(base_url, selenium, version=version).open()
+    assert page.is_news_displayed
+
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+@pytest.mark.parametrize('version', [('2.0'), ('1.4'), ('1.3t'), ('1.3'), ('1.1')])
+def test_cta_buttons_are_present(version, base_url, selenium):
+    page = FirefoxOSBasePage(base_url, selenium, version=version).open()
+    # test for presence of buttons only as display is dependent upon geo location
```
davehunt: In Firefox it should be possible to fake the geolocation data by setting **geo.prompt.testing** to **true**, **geo.prompt.testing.allow** to **true**, and **geo.wifi.uri** to a data URL with JSON content describing the fake location. I've not tested this recently, and it would require some work, but I wanted to let you know that it should be possible, if this was an important feature to test.

tests/functional/firefox/os/version/test_base.py
```diff
@@ -0,0 +1,33 @@
...
+@pytest.mark.parametrize('version', [('2.0'), ('1.4'), ('1.3t'), ('1.3'), ('1.1')])
+def test_news_is_displayed(version, base_url, selenium):
+    page = FirefoxOSBasePage(base_url, selenium, version=version).open()
+    assert page.is_news_displayed
+
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+@pytest.mark.parametrize('version', [('2.0'), ('1.4'), ('1.3t'), ('1.3'), ('1.1')])
+def test_cta_buttons_are_present(version, base_url, selenium):
```
davehunt: What does 'cta' mean? Could we expand it to make the test name more readable?

tests/functional/firefox/os/version/test_base.py
```diff
@@ -0,0 +1,33 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.firefox.os.version.base import FirefoxOSBasePage
+
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+@pytest.mark.parametrize('version', [('2.5'), ('2.0'), ('1.4'), ('1.3t'), ('1.3'), ('1.1')])
```
davehunt: Do we want all versions to run in the sanity suites? If not, you can mark individual parameters:

```python
@pytest.mark.parametrize('version', [
    pytest.mark.sanity(('2.5')),
    pytest.mark.sanity(('2.0')),
    ('1.4'),
    ('1.3t'),
    ('1.3'),
    ('1.1')])
```

tests/functional/firefox/os/version/test_base.py
```diff
@@ -0,0 +1,33 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
davehunt: The name of this file `test_base.py` suggests to me that this is a base test class that's extended. Perhaps rename it to something like `test_all.py` or `test_common.py`?

alexgibson: Sounds like a valid suggestion :+1:

tests/functional/firefox/os/version/test_base.py
alexgibson: Nice, did not know we could do this :smile:

tests/functional/firefox/os/version/test_base.py
alexgibson: It means 'call to action', will update to make it more obvious

tests/functional/firefox/os/version/test_base.py
alexgibson: Interesting suggestion, although in this case I'm not sure it will work for what we want, since we use an IP based lookup using `geo.mozilla.org`. This is something I would like to explore testing in the future perhaps, although I'm trying to push us to stop relying on it for our UI's, since the domain is now blacklisted by some anti-tracking addons. For now I'm going to add a TODO to explore a way to test this in the future.

alexgibson:
Thanks for the review @davehunt, updated

alexgibson:
@davehunt - when you have some time, please let me know if there are any other things to address. Thanks!

davehunt:
Sorry for the delay, GitHub notifications have been going to spam for a few days! :cry: This looks good, with my only question being whether we should check that selecting a category in `test_app_category_selector` fades a number of app icons out.

alexgibson:
> This looks good, with my only question being whether we should check that selecting a category in test_app_category_selector fades a number of app icons out.

I think we are probably ok just to assume that each category will fade out at least one icon, without making the test overly complicated.

PR #3602, opened 2015-11-24
Fix Bug 1214038, adds end of year fundraising takeover

schalkneethling:
Not ready for review, opening for l10n


bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -449,6 +449,10 @@ <h2 class="primary go">{{ _('Help Mozilla users') }}</h2>
   <button id="scroll-prompt">{{ _('Discover more') }}</button>
 </main>
 
+{% if waffle.switch('fundraiser-home-takeover') and l10n_has_tag('eoy_fundraising_2015') %}
```
flodolo: We need a different tag at this point, this one is associated to other strings.

schalkneethling: Ok `eoy_fundraising_takeover` ? Also, please note that there is different copy in the copy and design bug than that on the design provided in https://bugzilla.mozilla.org/show_bug.cgi?id=1214038#c15

Also, not sure, is the copy final? https://bugzilla.mozilla.org/show_bug.cgi?id=1214034#c3 @m-alexis

bedrock/mozorg/templates/mozorg/home/includes/fundraising/takeover.html
```diff
@@ -0,0 +1,48 @@
...
+   <section class="body">
+     <div class="content-container">
+       <header>
+         <h1>{{ high_res_img('home/takeover/wordmark.png', {'alt': 'Mozilla', 'width': '348', 'height': '92'}) }}</h1>
+         <h2>{{ _('Proudly non-profit. <br>Donate before Dec 31.') }}</h2>
+       </header>
+       <form id="fundraiser-form" method="get" action="{{ donate_url('mozillaorg_takeover') }}#amount-25">
+         <fieldset>
+           <label for="donate100">
+             <input type="radio" value="100" id="donate100" name="amount"> {{ _('$100') }}
```
flodolo: I'm not completely sure this is a good approach: are localizers supposed to know which currency we support? I think this should be determined server side.

Maybe talk with @alicoding to see how donate.mozilla.org works

bedrock/mozorg/templates/mozorg/home/includes/fundraising/takeover.html
```diff
@@ -0,0 +1,48 @@
...
+
+           <label for="donate-custom">{{ _('Enter your own amount') }}</label>
+           <input type="text" name="amount-custom" id="donate-custom" />
+
+           <label for="monthly">
+             <input type="radio" value="monthly" id="monthly" name="donation-type"> {{ _('Monthly') }}
+           </label>
+
+           <label for="onetime">
+             <input type="radio" value="onetime" id="onetime" name="donation-type"> {{ _('Onetime') }}
```
flodolo: One time?

bedrock/mozorg/templates/mozorg/home/includes/fundraising/takeover.html
flodolo: Alternatively
http://bedrock.readthedocs.org/en/latest/l10n.html#currency

bedrock/mozorg/templates/mozorg/home/home.html
flodolo: I'm talking about the strings in the tile (that's where we're currently using that tag).

The proposed tag works for me.

bedrock/mozorg/templates/mozorg/home/includes/fundraising/takeover.html
flodolo: I also wonder how you determine the currency to send to the form. Maybe create a list of supported locales+currency in settings, and fallback to en-US?

bedrock/mozorg/templates/mozorg/home/includes/fundraising/takeover.html
schalkneethling: I also thought so, but in the design it was one work :-/ Will fix.

flodolo: I know it's "One time" in https://donate.mozilla.org/en-US/ but "One-time" in the last design I see. Not confusing at all...

schalkneethling:
@flodolo Ok, the copy here is final so, feel free to expose. I am going to update the amounts to use the format you pointed to in http://bedrock.readthedocs.org/en/latest/l10n.html#currency

flodolo:
I need to see how the final string for the format looks like. Also, it looks like the design is actually using "one-time" https://drive.google.com/a/mozilla.com/file/d/0B2qu3ExKWIlvTGlhZExDTThWYUE/view

bedrock/mozorg/templates/mozorg/home/includes/fundraising/takeover.html
schalkneethling: The URL has two parameters `&amount=50` and `&currency=eur` so, I reckon we will have to have a list in settings as you suggested above.

schalkneethling:
Is using `One-time` ok for l10n?

flodolo:
Yes, it's just a word to localize, no difference between "One-time", "One time" and "Onetime" ;-)

alexgibson:
Note when this merges there are likely some home page functional tests that will skipping until the takeover has been removed again.

schalkneethling:
@alexgibson ok, this is ready for code review. The only thing is, whilst all the documentation suggests that from IE9 and up both SVG and background-size is supported, the background image behaves completely wacko in IE. Everywhere else it's fine.

Perhaps you have run into this before?

flodolo:
A few issues spotted while testing Italian: text overflows behind the image.

I also tried sv-SE, and it looks like the currency is not identified correctly on d.m.o (but that's a bug on their end, I think).

alexgibson:
It's not very obvious what this text field is for, is the label supposed to be hidden? At first I assumed I had to type one of the displayed amounts into the field, not realizing those are buttons.

text-field

alexgibson:
Donate button overflows the modal on mobile

mobile

media/js/mozorg/home/takeover.js
```diff
@@ -0,0 +1,88 @@
...
+    // once a user types into the custom amount field
+    // clear all other options.
+    $donateCustom.keyup(clearOptions);
+
+    // the name of the parameter that indicates the amount the user
+    // wish to donate has to be called amount. Seeing that we have
+    // radio buttons with that name as well as an input field, we need
+    // to disable the input field prior to form submission to avoid a
+    // scenario where we have ?amount=50&amount=
+    $form.on('submit', function(event) {
```
alexgibson: Nit: `event` is not used

media/css/mozorg/home/takeover.less
```diff
@@ -0,0 +1,216 @@
...
+.fundraising-takeover * {
+    .border-box;
+}
+
+.fundraising-takeover {
+    position: absolute;
+    top: 0;
+    left: 0;
+    background: @backgroundColor url('/media/img/home/takeover/hart.svg') center top no-repeat;
+    background-size: 100%;
```
alexgibson: Please use the `background-size()` mixin

media/css/mozorg/home/takeover.less
```diff
@@ -0,0 +1,216 @@
...
+    color: @foregroundColor;
+    width: 100%;
+    height: 720px;
+    .border-box;
+    z-index: 999;
+
+    @media only screen and (max-width: @breakDesktopWide) {
+        background-size: 1500px;
```
alexgibson: Use `.background-size()` mixin

media/css/mozorg/home/takeover.less
```diff
@@ -0,0 +1,216 @@
...
+        background-position: center 0;
+    }
+
+    @media only screen and (max-width: @breakTablet) {
+        background-size: 1000px;
```
alexgibson: Use `.background-size()` mixin

alexgibson:
If I load the following URL the donate form displays with "one time" selected instead of monthly. If the parameter correct?

https://donate.mozilla.org/en-US/?amount=50&recurring_acknowledge=monthly

schalkneethling:
I have posted some questions on the bug to get more clarity on some of the above:
https://bugzilla.mozilla.org/show_bug.cgi?id=1214038#c31

alexgibson:
As mentioned earlier, this change is going to make some of our functional tests fail. I'd suggest skipping the following tests by adding `@pytest.mark.skipif(reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1214038')` to each test below:

`tests/functional/test_home.py`

* `test_mozilla_newsletter_default_values`
* `test_mozilla_newsletter_successful_sign_up`
* `test_mozilla_newsletter_sign_up_fails_when_missing_required_fields`

`tests/functional/test_navigation.py`

* `test_navigation`
* `test_mobile_navigation`

`tests/functional/test_l10n.py`

* `test_change_language`

We then need to stop skipping these tests once the takeover is removed.

media/js/mozorg/home/takeover.js
```diff
@@ -0,0 +1,88 @@
...
+    });
+
+    // if a user clicks on 'continue to mozilla.org',
+    // close the modal and overlay
+    $closeModal.on('click', function(event) {
+        event.preventDefault();
+        toggleModal();
+    });
+
+    try {
```
alexgibson: Seeing as this promo is waffled, should we add a check here for `length` to make sure it exists in the DOM before continuing?

alexgibson:
It would be nice to be able to close the takeover by hitting `esc`, I think we've done this in the past on previous years?

media/js/mozorg/home/takeover.js
schalkneethling: Good point!

alexgibson:
I'm not sure if using a placeholder is a good solution as a label replacement, as it gets cut off even in English:

placeholder

alexgibson:
Long text gets cut off behind Curious Mr Fox:

text-overflow

media/css/mozorg/home/takeover.less
```diff
@@ -0,0 +1,220 @@
...
+
+.fundraising-takeover {
+    position: absolute;
+    top: 0;
+    left: 0;
+    background: @backgroundColor url('/media/img/home/takeover/hart.svg') center top no-repeat;
+    .background-size(100%);
+    color: @foregroundColor;
+    width: 100%;
+    height: 720px;
```
alexgibson: Is there a reason why this needs a fixed height? Given that string lengths can push content down, does it make life easier to remove this hard-coded value and allow `auto`?

schalkneethling:
Yeah, that is the reason I removed it :-/ - I have a specific break to avoid text going behind Mr. Fox, hmmmm

media/css/mozorg/home/takeover.less
schalkneethling: I guess that is fine. It was specified in the design as 720px so, I set it to that.

schalkneethling:
I wonder @flodolo, do we have the string `Other amount` translated in bedrock already?

flodolo:
No, we don't. 

You can always check on: https://transvision.mozfr.org (select mozilla.org as repository).

schalkneethling:
I wonder, what if we just use the word `Other` in that field? @flodolo @alexgibson Thoughts?

schalkneethling:
@alexgibson I cannot replicate what you show in the screen shot where Mr. Fox overlaps the text:
http://screencast.com/t/kn1TZPRaxz

flodolo:
Still don't have the string and too late for that: it's in a page that not all locales have (actually, very few), and probably used in a completely different context.

media/css/mozorg/home/takeover.less
```diff
@@ -0,0 +1,223 @@
...
+        @media only screen and (max-width: @breakTablet) {
+            margin-top: (@baseLine * 3);
+        }
+
+        @media only screen and (max-width: 360px) {
+            margin-top: @baseLine;
+        }
+
+        header {
+            max-width: 530px;
```
alexgibson: This style is applied for all breakpoints, breaking smaller layouts. We could just use `width` here instead of `max-width` and set an value for each breakpoint to make sure things reflow nicely.

media/css/mozorg/home/takeover.less
```diff
@@ -0,0 +1,241 @@
...
+            @media only screen and (max-width: @breakTablet) {
+                width: 70px;
+            }
+
+            @media only screen and (max-width: 360px) {
+                margin: 5px 2px;
+                width: 130px;
+            }
+
+            &:hover,
```
alexgibson: Nit: add `focus` state here also?

schalkneethling: Yes, good call.

media/css/mozorg/home/takeover.less
```diff
@@ -0,0 +1,241 @@
...
+    .recurring {
+        margin-top: 10px;
+
+        label {
+            margin-right: @baseLine;
+            .font-size(1rem);
+        }
+    }
+
+    input[type="submit"] {
```
alexgibson: Would be nice to have some sort of hover/focus state for the submit button too?

bedrock/mozorg/templates/mozorg/home/includes/fundraising/takeover.html
```diff
@@ -0,0 +1,57 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+ <article id="fundraising_takeover" class="fundraising-takeover hidden">
+   <div class="header-strip">
+     <p class="contained"><a href="{{ url('mozorg.home') }}" id="close_takeover" class="more">{{ _('Continue to mozilla.org') }}</a></p>
```
alexgibson: Add `role="button"` here?

media/js/mozorg/home/takeover.js
```diff
@@ -0,0 +1,91 @@
...
+    var $optionLabels = $('label[class!="own-amount"]', $donateOptionsSection);
+    var $options = $('input[type="radio"]', $donateOptionsSection);
+
+    // sets the mozorg.takeover state to true indicating that the user has
+    // seen the takeover modal during their current session.
+    function storeSession() {
+        try {
+            sessionStorage.setItem('mozorg.takeover.seen', true);
+        } catch (ex) {
+            console.log('Error thrown while setting takeover modal state in sessionStorage: ', ex);
```
alexgibson: Ironically, these console logs would end up throwing an error in older browsers (e.g. where sessionStorage is unsupported.) I think we should just remove these.

media/js/mozorg/home/takeover.js
```diff
@@ -0,0 +1,91 @@
...
+            // only show the takeover modal if the user has not already seen it
+            // during this session.
+            if (sessionStorage.getItem('mozorg.takeover.seen')) {
+                // show the takeover
+                toggleModal();
+                // set mozorg.takeover.seen to true
+                storeSession();
+            }
+        } catch (ex) {
+            console.log('Error thrown while getting takeover modal state from sessionStorage: ', ex);
```
alexgibson: Same here, remove `console.log`

media/js/mozorg/home/takeover.js
schalkneethling: So essentially, catch and swallow.

media/js/mozorg/home/takeover.js
```diff
@@ -0,0 +1,106 @@
...
+        event.preventDefault();
+        toggleModal();
+    });
+
+    // modal is waffled so, first ensure it exists
+    if ($takeoverModal.length > 0) {
+        try {
+            // only show the takeover modal if the user has not already seen it
+            // during this session.
+            if (sessionStorage.getItem('mozorg.takeover.seen')) {
```
alexgibson: This doesn't work, I don't see the takeover since the value is initially `null`. Suggest changing to `sessionStorage.getItem('mozorg.takeover.seen') !== 'true'` (make sure to set the value to `true` as a string above once visible).

media/js/mozorg/home/takeover.js
```diff
@@ -0,0 +1,106 @@
...
+                // show the takeover
+                toggleModal();
```
alexgibson: It feels like we're swallowing a bit too much logic in this try/catch, would it be better moving this to it's own function called `initTakeover()` or similar?

alexgibson:
When I hover over the dontate button, the fox/background does a little 1px jump?

media/js/mozorg/home/takeover.js
schalkneethling: Dang it! It is set like hat for debugging. Should be `!session....`

media/css/mozorg/home/takeover.less
```diff
@@ -0,0 +1,249 @@
...
+
+        @media only screen and (max-width: @breakTablet) {
+            margin: @baseLine auto;
+        }
+
+        &:hover,
+        &:focus {
+            background-color: #fff;
+            color: #333;
+            border: 1px solid #333;
```
alexgibson: This `1px` is causing the takeover to do a little jump when you hover over submit.

media/js/mozorg/home/takeover.js
schalkneethling: Ok, fixed up

media/css/mozorg/home/takeover.less
schalkneethling: Good catch

media/js/mozorg/home/takeover.js
schalkneethling: I completely agree. :+1:

media/js/mozorg/home/takeover.js
```diff
@@ -0,0 +1,106 @@
...
+        event.preventDefault();
+        toggleModal();
+    });
+
+    // modal is waffled so, first ensure it exists
+    if ($takeoverModal.length > 0) {
+        try {
+            // only show the takeover modal if the user has not already seen it
+            // during this session.
+            if (!sessionStorage.getItem('mozorg.takeover.seen')) {
```
alexgibson: Could i suggest we be more explicit than using `!`, seems like it is an easy thing to miss (as evident :P)

schalkneethling: You are correct, only saw the suggestion after changing to `!`

alexgibson:
Styling needs some RTL attention:

rtl

bedrock/mozorg/templates/mozorg/home/includes/fundraising/takeover.html
```diff
@@ -0,0 +1,43 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+ <article id="fundraising_takeover" class="fundraising-takeover hidden">
+   <div class="header-strip">
+     <p class="contained"><a href="{{ url('mozorg.home') }}" id="close_takeover" class="more" role="button">{{ _('Continue to mozilla.org') }}</a></p>
+   </div>
+   <section class="copy contained">
+     <header>
+       <h1>{{ high_res_img('home/takeover/wordmark.png', {'alt': 'Mozilla', 'class': 'moz-logo', 'width': '276', 'height': '93'}) }}</h1>
```
alexgibson: Nit: we probably shouldn't be using multiple `<h1>` on the page for a11y?

alexgibson:
Donate button breaks out the viewport on mobile with longer strings (this is `it`).

mobile-submit

bedrock/mozorg/templates/mozorg/home/includes/fundraising/takeover.html
schalkneethling: Yeah, I was thinking about that but, as the `article` starts a new context it is technically fine but, I know not all screen readers works very well with that document outline so, probably safer to just have two `h2` elements

alexgibson:
by no means a blocker, but I notice Safari adds some extra thick drop shadow to the fox to try and make him look shifty!

safari-fox

media/js/mozorg/home/takeover.js
```diff
@@ -0,0 +1,113 @@
...
+        if ($donateCustom.val() === '') {
+            $donateCustom.attr('disabled', 'disabled');
+        }
+    });
+
+    // if a user clicks on 'continue to mozilla.org',
+    // close the modal and overlay
+    $closeModal.on('click', function(event) {
+        event.preventDefault();
+        toggleModal();
```
alexgibson: Should we pass `true` here to unbind the listener?

media/js/mozorg/home/takeover.js
```diff
@@ -0,0 +1,113 @@
...
+    function initTakeOver() {
+        // show the takeover
+        toggleModal();
+        // set mozorg.takeover.seen to true
+        storeSession();
+
+        // trap key events, listening for escape.
+        // we only want to do this here so, we do not attach the event
+        // if we are not showing the modal.
+        $('body').on('keyup.takeover', function(event) {
```
alexgibson: We could also cache `$('body')` as a variable since we're also removing the event later on?

schalkneethling:
Yeah, seems like it does not support transparency properly. Or perhaps how it is generated in Illustrator.

media/js/mozorg/home/takeover.js
```diff
@@ -0,0 +1,119 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+$(function () {
+    'use strict';
+
+    // indexOf on Array is undefined in versions of IE lte8
+    var oldIE = typeof Array.prototype.indexOf !== 'function';
```
alexgibson: This might work but it just sounds wrong? If we're gonna use browser detection, let's just check the UA string :smile:

`var oldIE = (site.platform === 'windows' && /MSIE\s[1-8]\./.test(navigator.userAgent));`

alexgibson: Alternatively, we could rename it to `cutsTheMustard`? And feature detect `indexOf` and `sessionStorage`?

schalkneethling: hey what the heck, let's sniff away for now

schalkneethling: On second thought, let's test for indexOf ;)

alexgibson: works for me :+1:

alexgibson:
r+ from me :beers: 

Care to squash? We can merge whenever the takeover is approved my MoCo

PR #3603, opened 2015-11-25
Fix Bug 1219957 - Move browser detection code from global.js to mozilla-client.js

kyoshino:
Move the browser detecting functions to the new Mozilla Client module and offer each static properties for faster access. This PR also removes GA events added in #3451.

### DIFF
--- a/media/js/base/mozilla-client.js
+++ b/media/js/base/mozilla-client.js
@@ -10,11 +10,98 @@ if (typeof Mozilla === 'undefined') {
 (function() {
     'use strict';
 
+    /**
+     * Provide information on the user's browsing environment, including the platform and browser details.
+     *
+     * @namespace
+     * @see {@link https://developer.mozilla.org/en-US/docs/Gecko_user_agent_string_reference}
+     */
     var Client = {};
 
     /**
-     * Get the user's Firefox version number. '0' will be returned on non-Firefox browsers.
+     * Detect whether the user's browser is Firefox on any platform. This includes WebKit-based Firefox for iOS.
      *
+     * @private
+     * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
+     * @return {Boolean} result
+     */
+    Client._isFirefox = function (ua) {
+        ua = ua || navigator.userAgent;
+
+        return /\s(Firefox|FxiOS)/.test(ua) && !Client._isLikeFirefox(ua);
+    };
+
+    /**
+     * Detect whether the user's browser is Firefox for Windows, OS X or Linux.
+     *
+     * @private
+     * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
+     * @return {Boolean} result
+     */
+    Client._isFirefoxDesktop = function (ua) {
+        ua = ua || navigator.userAgent;
+
+        return /\sFirefox/.test(ua) && !/Mobile|Tablet|Fennec/.test(ua) && !Client._isLikeFirefox(ua);
+    };
+
+    /**
+     * Detect whether the user's browser is Firefox for Android.
+     *
+     * @private
+     * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
+     * @return {Boolean} result
+     */
+    Client._isFirefoxAndroid = function (ua) {
+        ua = ua || navigator.userAgent;
+
+        return /\sFirefox/.test(ua) && /Android/.test(ua);
+    };
+
+    /**
+     * Detect whether the user's browser is Firefox for iOS.
+     *
+     * @private
+     * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
+     * @return {Boolean} result
+     */
+    Client._isFirefoxiOS = function (ua) {
+        ua = ua || navigator.userAgent;
+
+        return /FxiOS/.test(ua);
+    };
+
+    /**
+     * Detect whether the user's browser is the Browser app on Firefox OS.
+     *
+     * @private
+     * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
+     * @param  {String} pf - browser's platform name, navigator.platform is used if not specified
+     * @return {Boolean} result
+     */
+    Client._isFirefoxFxOS = function (ua, pf) {
+        ua = ua || navigator.userAgent;
+        pf = (pf === '') ? '' : pf || navigator.platform;
+
+        return /Firefox/.test(ua) && pf === '';
+    };
+
+    /**
+     * Detect whether the user's browser is Gecko-based. Used on the plugincheck page to support all Gecko browsers.
+     *
+     * @private
+     * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
+     * @return {Boolean} result
+     */
+    Client._isLikeFirefox = function (ua) {
+        ua = ua || navigator.userAgent;
+
+        return /Iceweasel|IceCat|SeaMonkey|Camino|like\ Firefox/i.test(ua);
+    };
+
+    /**
+     * Get the user's Firefox version number. '0' will be returned on Firefox for iOS and non-Firefox browsers.
+     *
+     * @private
      * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
      * @return {String} version number
      */
@@ -23,12 +110,24 @@ if (typeof Mozilla === 'undefined') {
 
         var matches = /Firefox\/(\d+(?:\.\d+){1,2})/.exec(ua);
 
-        return (matches !== null && matches.length > 0) ? matches[1] : '0';
+        return (matches && !Client._isLikeFirefox(ua)) ? matches[1] : '0';
+    };
+
+    /**
+     * Get the user's Firefox major version number. 0 will be returned on Firefox for iOS and non-Firefox browsers.
+     *
+     * @private
+     * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
+     * @return {Number} major version number in integer
+     */
+    Client._getFirefoxMajorVersion = function (ua) {
+        return parseInt(Client._getFirefoxVersion(ua), 10);
     };
 
     /**
      * Detect whether the user's Firefox is up to date or outdated. This data is mainly used for security notifications.
      *
+     * @private
      * @param  {Boolean} strict - whether the minor and patch-level version numbers should be compared. Default: true
      * @param  {Boolean} isESR - whether the Firefox update channel is ESR. Default: false
      * @param  {String}  userVer - browser's version number
@@ -37,7 +136,7 @@ if (typeof Mozilla === 'undefined') {
     Client._isFirefoxUpToDate = function (strict, isESR, userVer) {
         strict = strict === undefined ? true : strict;
         isESR = isESR === undefined ? false : isESR;
-        userVer = userVer === undefined ? this._getFirefoxVersion() : userVer;
+        userVer = userVer === undefined ? Client._getFirefoxVersion() : userVer;
 
         var $html = $(document.documentElement);
 
@@ -92,8 +191,8 @@ if (typeof Mozilla === 'undefined') {
      */
     Client.getFirefoxDetails = function (callback) {
         // Fire the callback function immediately if cache exists
-        if (this.FirefoxDetails) {
-            callback(this.FirefoxDetails);
+        if (Client.FirefoxDetails) {
+            callback(Client.FirefoxDetails);
 
             return;
         }
@@ -137,11 +236,11 @@ if (typeof Mozilla === 'undefined') {
         };
 
         // Prepare fallback function in case the API doesn't work
-        var userVer = this._getFirefoxVersion();
+        var userVer = Client._getFirefoxVersion();
         var fallback = function () { onRetrieved(false, userVer, 'release'); };
 
         // If Firefox is old or for Android, call the fallback function immediately because the API is not implemented
-        if (parseFloat(userVer) < 35 || window.isFirefoxMobile()) {
+        if (parseFloat(userVer) < 35 || Client._isFirefoxAndroid()) {
             fallback();
 
             return;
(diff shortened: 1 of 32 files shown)


kyoshino:
@alexgibson r?

media/js/base/mozilla-client.js
```diff
@@ -123,17 +185,6 @@ if (typeof Mozilla === 'undefined') {
             };
 
             callback(details);
-
-            // Log the info with GA. Remove this once we have sufficient data
-            window.dataLayer = window.dataLayer || [];
-            window.dataLayer.push({
```
alexgibson: It may take a little while for Gareth to get us GA data from the last PR, so to avoid blocking the merge of this change, should we leave this in for now? We can always file a follow up bug once we have enough data.

media/js/base/mozilla-client.js
```diff
@@ -141,7 +192,7 @@ if (typeof Mozilla === 'undefined') {
         var fallback = function () { onRetrieved(false, userVer, 'release'); };
 
         // If Firefox is old or for Android, call the fallback function immediately because the API is not implemented
-        if (parseFloat(userVer) < 35 || window.isFirefoxMobile()) {
+        if (parseFloat(userVer) < 35 || this._isFirefoxAndroid()) {
```
alexgibson: Nit: just use `Client` instead of `this`?

alexgibson:
Thanks @kyoshino - I'll dig into testing the specific pages here, but this looks to be some excellent paying off of tech debt :+1:

media/js/base/mozilla-client.js
kyoshino: Okay, will remove the change from this PR.

media/js/base/mozilla-client.js
kyoshino: :+1:

media/js/base/mozilla-client.js
```diff
@@ -161,6 +212,20 @@ if (typeof Mozilla === 'undefined') {
         }));
     };
 
+    // Append static properties for faster access
+    Client.isFirefox = Client._isFirefox();
+    Client.isFirefoxAndroid = Client._isFirefoxAndroid();
+    Client.isFirefoxiOS = Client._isFirefoxiOS();
+    Client.isFirefoxDesktop = !Client.isFirefoxAndroid;
```
kyoshino: Um, this is wrong. Will fix it.

kyoshino:
Fixed the `isFirefoxDesktop` detection. Also, changed `isFirefox` to return `true` also for Firefox for iOS. That means some code now requires `!client.isFirefoxiOS` to exclude iOS, but I think it makes sense.

media/js/firefox/new.js
```diff
@@ -80,8 +58,8 @@
                 $html.addClass('firefox-latest');
 
                 // if user is on release channel and has latest version, offer refresh button
-                uiTourGetConfiguration('appinfo', function(config) {
-                    if (config.defaultUpdateChannel === 'release' && config.version === latestFirefoxVersionFull) {
+                client.getFirefoxDetails(function(data) {
```
alexgibson: Should we add an `if (client.isFirefoxDesktop)` conditional here?

media/js/firefox/sync.js
```diff
@@ -32,15 +33,15 @@
     };
 
     // Variations 1-5 are Firefox
-    if (window.isFirefox()) {
+    if (client.isFirefox) {
```
alexgibson: Looks like there's nothing for Firefox for iOS to fall into here below, should we add `!client.isFirefoxiOS`?

alexgibson: Actually no need, the new `if (client.isFirefoxDesktop)` below should be enough :+1:

media/js/firefox/desktop/intro-anim.js
```diff
@@ -5,9 +5,11 @@
 (function(Mozilla, $) {
     'use strict';
 
+    var client = Mozilla.Client;
+
     function cutsTheMustard() {
         // Bug (1083079) fixed but effects Firefox 35
-        if (window.isFirefox() && window.getFirefoxMasterVersion() === 35 && $('html').hasClass('osx')) {
+        if (client.isFirefox && client.FirefoxMajorVersion === 35 && client.platform === 'osx') {
```
alexgibson: Firefox 35 is quite old now, I think we're good to remove this check as the bug has been fixed for a while :+1:

alexgibson:
This is looking great - I've pushed this branch to demo2 for some testing: https://www-demo2.allizom.org/en-US/. 

Will try and take a closer look at this soon, thanks @kyoshino!

media/js/firefox/new.js
```diff
@@ -43,34 +45,10 @@
...
-        var latestFirefoxVersionFull = $html.attr('data-latest-firefox');
-
-        // get latest full version (no point release info) for initial check
-        var latestFirefoxVersion = parseInt(latestFirefoxVersionFull.split('.')[0], 10);
-
-        if (window.isFirefoxUpToDate(latestFirefoxVersion + '')) {
+    if (client.isFirefox && !client.isFirefoxiOS) {
+        // Detect whether the Firefox is up-to-date in a non-strict way. The minor version and channel are not
+        // considered. This can/should be strict, once the UX especially for ESR users is decided. (Bug 939470)
+        if (client._isFirefoxUpToDate(false)) {
```
alexgibson: I like they way this is setting us up for a simple change in the near future :+1:

alexgibson:
The function `_isFirefoxAndroid()` currently returns true for Firefox on iOS: 

`Mozilla.Client._isFirefoxAndroid('Mozilla/5.0 (iPhone; CPU iPhone OS 8_3 like Mac OS X) AppleWebKit/600.1.4 (KHTML, like Gecko) FxiOS/1.1 Mobile/12F69 Safari/600.1.4');`

This seems to be the only way that `_isFirefoxDesktop` correctly returns `false` for Firefox on iOS. Both will likely need to be updated.

media/js/base/mozilla-client.js
```diff
@@ -27,8 +100,20 @@ if (typeof Mozilla === 'undefined') {
...
 
     /**
+     * Get the user's Firefox major version number. 0 will be returned on non-Firefox browsers.
+     *
+     * @private
+     * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
+     * @return {Number} major version number
+     */
+    Client._getFirefoxMajorVersion = function (ua) {
+        return parseFloat(Client._getFirefoxVersion(ua));
```
alexgibson: Technically, won't this return more than just the major version number? e.g. `Won't this return more than the major version number alone? e.g. `Mozilla.Client._getFirefoxMajorVersion("Mozilla/5.0 (Macintosh; Intel Mac OS X 10.10; rv:44.1) Gecko/20100101 Firefox/44.1");`

media/js/plugincheck/check-plugins.js
```diff
@@ -171,7 +173,7 @@ $(function() {
     }
 
     // only execute the plugincheck code if this is Firefox
-    if (isFirefox() || isLikeFirefox()) {
+    if ((client.isFirefox && !client.isFirefoxiOS) || client.isLikeFirefox) {
```
alexgibson: Could this be simplified to `client.isFirefoxDesktop || client.isLikeFirefox`?

media/js/plugincheck/check-plugins.js
```diff
@@ -157,13 +159,13 @@ $(function() {
     }
 
     // show main download button to non Fx traffic
-    if(!isFirefox() && !isLikeFirefox() ) {
+    if((!client.isFirefox && !client.isLikeFirefox) || client.isFirefoxiOS) {
```
alexgibson: Why would we want to show the download button to users already on Firefox for iOS?

tests/unit/spec/base/mozilla-client.js
```diff
@@ -9,6 +9,172 @@ describe('mozilla-client.js', function() {
...
+        });
+
+        it('should return false for other browsers', function() {
+            expect(window.Mozilla.Client._isFirefoxDesktop('Mozilla/4.0 (compatible; MSIE 8.0; Windows NT 5.1; Trident/4.0; .NET CLR 2.0.50727; .NET CLR 3.0.04506.648; .NET CLR 3.5.21022; .NET CLR 3.0.4506.2152; .NET CLR 3.5.30729)')).toBeFalsy();
+            expect(window.Mozilla.Client._isFirefoxDesktop('Mozilla/5.0 (Macintosh; Intel Mac OS X 10_8_4) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/29.0.1547.57 Safari/537.36')).toBeFalsy();
+        });
+
+    });
+
+    describe('_isFirefoxAndroid', function () {
```
alexgibson: It would be good to add a test for the Firefox iOS user agent here to make sure it returns false.

media/js/base/mozilla-client.js
alexgibson: If we're relying on the expectation that the minor version number is never in the UA string for this to work, I'd feel more comfortable if we explicitly made this return only the major version, like the function name suggests.

alexgibson:
This is looking great, just need to make sure we firm up our UA tests a little for any unexpected false positives :+1:

media/js/base/mozilla-client.js
kyoshino: I was actually thinking old [releases](https://www.mozilla.org/en-US/firefox/releases/) like 3.6 but yes, 38.4.0 should be considered as `38`, not `38.4`... I'll use `parseInt` instead.

media/js/firefox/desktop/intro-anim.js
kyoshino: Okay, will remove this.

media/js/firefox/new.js
kyoshino: :+1:

media/js/plugincheck/check-plugins.js
kyoshino: So true.

media/js/plugincheck/check-plugins.js
kyoshino: And `client.isFirefoxAndroid` maybe.

tests/unit/spec/base/mozilla-client.js
kyoshino: Okay

kyoshino:
Fixed the remaining issues @alexgibson have pointed out. I have replaced `client.isFirefox && !client.isFirefoxiOS` with `client.isFirefoxDesktop || client.isFirefoxAndroid` for a better readability.

tests/unit/spec/base/mozilla-client.js
```diff
@@ -9,18 +9,209 @@ describe('mozilla-client.js', function() {
...
+
+        it('should return true for Firefox Android on Phone', function() {
+            expect(window.Mozilla.Client._isFirefoxAndroid('Mozilla/5.0 (Android; Mobile; rv:26.0) Gecko/26.0 Firefox/26.0')).toBeTruthy();
+        });
+
+        it('should return true for Firefox Android on Tablet', function() {
+            expect(window.Mozilla.Client._isFirefoxAndroid('Mozilla/5.0 (Android; Tablet; rv:26.0) Gecko/26.0 Firefox/26.0')).toBeTruthy();
+        });
+
+        it('should return true for Firefox OS on Phone', function() {
```
alexgibson: Wait, why are we returning true here for Firefox OS when the function is `_isFirefoxAndroid`?

media/js/base/mozilla-client.js
```diff
@@ -10,11 +10,84 @@ if (typeof Mozilla === 'undefined') {
...
+     * Detect whether the user's browser is Firefox for Android.
      *
+     * @private
+     * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
+     * @return {Boolean} result
+     */
+    Client._isFirefoxAndroid = function (ua) {
+        ua = ua || navigator.userAgent;
+
+        return /\sFirefox/.test(ua) && /Mobile|Tablet|Fennec/.test(ua) && !Client._isLikeFirefox(ua);
```
alexgibson: From the tests it looks like this returns true for Firefox OS? Would this be more robust if we we're checking for `android` specifically?

tests/unit/spec/base/mozilla-client.js
kyoshino: Aha, bad copy & paste! Will fix it.

media/js/base/mozilla-client.js
kyoshino: I just also noticed that Firefox for Maemo still returns `true` here. It's no longer `isFirefoxMobile`...

kyoshino:
Alright, the detecting functions are now more robust, and the tests are more readable and comprehensive. I have also added `Client.isFirefoxFxOS` for consistency.

media/js/base/mozilla-client.js
```diff
@@ -10,11 +10,96 @@ if (typeof Mozilla === 'undefined') {
...
+     * Detect whether the user's browser is the Browser app on Firefox OS.
+     *
+     * @private
+     * @param  {String} ua - browser's user agent string, navigator.userAgent is used if not specified
+     * @return {Boolean} result
+     */
+    Client._isFirefoxFxOS = function (ua) {
+        ua = ua || navigator.userAgent;
+
+        return /\sFirefox/.test(ua) && /Mobile|Tablet/.test(ua) && !/Android/.test(ua);
```
alexgibson: If we're adding Firefox OS to the mix, I think we should remain consistent with our current detection method in site.js: https://github.com/mozilla/bedrock/blob/master/media/js/base/site.js#L67-L69

kyoshino: Thanks, added `navigator.platform` check to be consistent with `site.js`.

alexgibson:
@kyoshino - since this Pr was opened a few other pages have been created which use the old `global.js` functions. Could you please rebase and update where needed? Then I'll get this back on a demo server. Thanks for your patience!

kyoshino:
@alexgibson Rebased and removed the legacy functions :smiley_cat:

alexgibson:
I've put this back up on demo2 and done some preliminary testing and so far things look good. I still need to do some further testing on Android & iOS on certain pages.

alexgibson:
I don't have a device with Firefox for Android to test with, so I'm not able to verify the browser specific messaging shown on the following pages:

https://www-demo2.allizom.org/en-US/firefox/android/
https://www-demo2.allizom.org/en-US/firefox/hello/
https://www-demo2.allizom.org/en-US/firefox/ios/
https://www-demo2.allizom.org/en-US/firefox/new/
https://www-demo2.allizom.org/en-US/plugincheck/ (does plugincheck even work on Firefox for Android?)
https://www-demo2.allizom.org/en-US/firefox/private-browsing/
https://www-demo2.allizom.org/en-US/firefox/sync/

@kyoshino - do you have a device to help test with here? If not, I can see if someone from WebQA can help to verify these pages.

kyoshino:
I'll test those pages with my Android phone and tablet.

kyoshino:
LGTM on Nexus 5X and 9. Confirmed the plugincheck page worked even on Android when [Flash Player was manually installed](https://support.mozilla.org/en-US/kb/how-do-i-watch-flash-videos-firefox-android).

alexgibson:
Thanks for helping to verify @kyoshino, and thanks again for this PR. Nice work! r+ :rabbit:

kyoshino:
Thanks @alexgibson!

PR #3604, opened 2015-11-25
Bug 1224565 - Fix: Add pdf url

malena:
https://bugzilla.mozilla.org/show_bug.cgi?id=1224565


malena:
@jpetto r?

jpetto:
Code looks good to me. r+

Waiting for Travis...but should we also wait for the PDF to be available on `static.mozilla.com` just to be safe?

jpetto:
PDF is available!

PR #3605, opened 2015-11-26
Fix Bug 1227250, replace all instances of sendto donate links with donate.

schalkneethling:
@alexgibson @flodolo r?

### DIFF
--- a/bedrock/foundation/templates/foundation/annualreport/2009/broadening-our-scope.html
+++ b/bedrock/foundation/templates/foundation/annualreport/2009/broadening-our-scope.html
@@ -187,7 +187,7 @@
           festival="http://www.drumbeat.org/",
           movies="http://www.drumbeat.org/",
           subtitles="http://www.drumbeat.org/",
-          fund="https://donate.mozilla.org/"
+          fund=donate_url('annualreport2009')
         %}
           <img src="{{ logo }}" width="100" height="95" align="left" alt="">
           Mozilla's work to build openness and empowerment into the Internet has historically
--- a/bedrock/mozorg/helpers/misc.py
+++ b/bedrock/mozorg/helpers/misc.py
@@ -349,11 +349,7 @@ def press_blog_url(ctx):
 @jingo.register.function
 @jinja2.contextfunction
 def donate_url(ctx, source=''):
-    """Output a link to the donation page taking locales into account.
-
-    Uses the locale from the current request. Checks to see if we have
-    a donation page that match this locale, returns the localized page
-    url or falls back to the US page url if not.
+    """Output a donation link to the donation page formatted using settings.DONATE_PARAMS
 
     Examples
     ========
@@ -365,33 +361,25 @@ def donate_url(ctx, source=''):
 
     For en-US this would output:
 
-        https://sendto.mozilla.org/page/contribute/EOYFR2013-tabzilla
+        https://donate.mozilla.org/en-US/?presets=100,50,25,15&amount=50&ref=EOYFR2015&utm_campaign=EOYFR2015&utm_source=mozilla.org&utm_medium=referral&utm_content=footer&currency=usd
 
     For de this would output:
 
-        https://sendto.mozilla.org/page/contribute/EOYFR2013-webDE
+        https://donate.mozilla.org/de/?presets=100,50,25,15&amount=50&ref=EOYFR2015&utm_campaign=EOYFR2015&utm_source=mozilla.org&utm_medium=referral&utm_content=footer&currency=eur
 
-    For fr this would output:
+    For a locale not defined in settings.DONATE this would output:
 
-        https://sendto.mozilla.org/page/contribute/EOYFR2013-webFR
-
-    For pt-BR this would output:
-
-        https://sendto.mozilla.org/page/contribute/EOYFR2013-webPTBR
+        https://donate.mozilla.org/ca/?presets=100,50,25,15&amount=50&ref=EOYFR2015&utm_campaign=EOYFR2015&utm_source=mozilla.org&utm_medium=referral&utm_content=footer&currency=usd
 
     """
     locale = getattr(ctx['request'], 'locale', 'en-US')
-    if locale not in settings.DONATE_LOCALE_LINK:
-        if 'default' in settings.DONATE_LOCALE_LINK:
-            locale = 'default'
-        else:
-            locale = 'en-US'
 
-        # The source of the new donation URL is different from the old one,
-        # so replace it if needed (Bug 1100548)
-        source = source.replace('mozillaorg', 'mozillaorg_default')
+    donate_url_params = settings.DONATE_PARAMS.get(
+        locale, settings.DONATE_PARAMS['en-US'])
 
-    return settings.DONATE_LOCALE_LINK[locale].format(source=source)
+    return settings.DONATE_LINK.format(locale=locale, presets=donate_url_params['presets'],
+        default=donate_url_params['default'], source=source,
+        currency=donate_url_params['currency'])
 
 
 @jingo.register.function
--- a/bedrock/mozorg/tests/test_helper_misc.py
+++ b/bedrock/mozorg/tests/test_helper_misc.py
@@ -22,17 +22,22 @@ TEST_FILES_ROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)),
                                'test_files')
 TEST_L10N_MEDIA_PATH = os.path.join(TEST_FILES_ROOT, 'media', '%s', 'l10n')
 
-TEST_DONATE_LOCALE_LINK = {
-    'default':
-        'https://sendto.mozilla.org/page/contribute/Give-Now?source={source}',
-    'en-US':
-        'https://sendto.mozilla.org/page/contribute/givenow-seq?'
-        'preset=2&source={source}&ref=EOYFR2014&utm_campaign=EOYFR2014'
-        '&utm_source=mozilla.org&utm_medium=referral&utm_content={source}',
-    'es-MX':
-        'https://sendto.mozilla.org/page/contribute/givenow-seq-es?'
-        'source={source}&ref=EOYFR2014&utm_campaign=EOYFR2014'
-        '&utm_source=mozilla.org&utm_medium=referral&utm_content=mozillaorg_ES',
+TEST_DONATE_LINK = ('https://donate.mozilla.org/{locale}/?presets={presets}'
+    '&amount={default}&ref=EOYFR2015&utm_campaign=EOYFR2015'
+    '&utm_source=mozilla.org&utm_medium=referral&utm_content={source}'
+    '&currency={currency}')
+
+TEST_DONATE_PARAMS = {
+    'en-US': {
+        'currency': 'usd',
+        'presets': '100,50,25,15',
+        'default': '50'
+    },
+    'es-MX': {
+        'currency': 'eur',
+        'presets': '100,50,25,15',
+        'default': '15'
+    },
 }
 
 TEST_FIREFOX_TWITTER_ACCOUNTS = {
@@ -370,7 +375,8 @@ class TestPressBlogUrl(TestCase):
         eq_(self._render('oc'), 'https://blog.mozilla.org/press/')
 
 
-@override_settings(DONATE_LOCALE_LINK=TEST_DONATE_LOCALE_LINK)
+@override_settings(DONATE_LINK=TEST_DONATE_LINK,
+    DONATE_PARAMS=TEST_DONATE_PARAMS)
 class TestDonateUrl(TestCase):
     rf = RequestFactory()
 
@@ -383,30 +389,34 @@ class TestDonateUrl(TestCase):
     def test_donate_url_no_locale(self):
         """No locale, fallback to default page"""
         eq_(self._render('', 'mozillaorg_footer'),
-            'https://sendto.mozilla.org/page/contribute/Give-Now?'
-            'source=mozillaorg_default_footer')
+            'https://donate.mozilla.org//?presets=100,50,25,15'
+            '&amp;amount=50&amp;ref=EOYFR2015&amp;utm_campaign=EOYFR2015'
+            '&amp;utm_source=mozilla.org&amp;utm_medium=referral'
+            '&amp;utm_content=mozillaorg_footer&amp;currency=usd')
 
     def test_donate_url_english(self):
         """en-US locale, default page"""
         eq_(self._render('en-US', 'mozillaorg_footer'),
-            'https://sendto.mozilla.org/page/contribute/givenow-seq?'
-            'preset=2&amp;source=mozillaorg_footer&amp;ref=EOYFR2014'
-            '&amp;utm_campaign=EOYFR2014&amp;utm_source=mozilla.org'
-            '&amp;utm_medium=referral&amp;utm_content=mozillaorg_footer')
+            'https://donate.mozilla.org/en-US/?presets=100,50,25,15'
+            '&amp;amount=50&amp;ref=EOYFR2015&amp;utm_campaign=EOYFR2015'
+            '&amp;utm_source=mozilla.org&amp;utm_medium=referral'
+            '&amp;utm_content=mozillaorg_footer&amp;currency=usd')
 
     def test_donate_url_spanish(self):
         """es-MX locale, a localized page"""
         eq_(self._render('es-MX', 'mozillaorg_footer'),
-            'https://sendto.mozilla.org/page/contribute/givenow-seq-es?'
-            'source=mozillaorg_footer&amp;ref=EOYFR2014'
-            '&amp;utm_campaign=EOYFR2014&amp;utm_source=mozilla.org&amp;'
-            'utm_medium=referral&amp;utm_content=mozillaorg_ES')
+            'https://donate.mozilla.org/es-MX/?presets=100,50,25,15'
+            '&amp;amount=15&amp;ref=EOYFR2015&amp;utm_campaign=EOYFR2015'
+            '&amp;utm_source=mozilla.org&amp;utm_medium=referral'
+            '&amp;utm_content=mozillaorg_footer&amp;currency=eur')
 
     def test_donate_url_other_locale(self):
         """No page for locale, fallback to default page"""
         eq_(self._render('pt-PT', 'mozillaorg_footer'),
-            'https://sendto.mozilla.org/page/contribute/Give-Now?'
-            'source=mozillaorg_default_footer')
+            'https://donate.mozilla.org/pt-PT/?presets=100,50,25,15'
+            '&amp;amount=50&amp;ref=EOYFR2015&amp;utm_campaign=EOYFR2015'
+            '&amp;utm_source=mozilla.org&amp;utm_medium=referral'
+            '&amp;utm_content=mozillaorg_footer&amp;currency=usd')
 
 
 @override_settings(FIREFOX_TWITTER_ACCOUNTS=TEST_FIREFOX_TWITTER_ACCOUNTS)
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -596,93 +596,307 @@ MOBILIZER_LOCALE_LINK = {
     'cs': 'https://wiki.mozilla.org/Mobilizer/MobilizerCzechRepublic/'
 }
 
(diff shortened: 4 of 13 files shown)


schalkneethling:
Just waiting for confirmation the leaving a default amount of 50 for all locales/currencies is ok:
https://bugzilla.mozilla.org/show_bug.cgi?id=1227250#c18

bedrock/settings/base.py
```diff
@@ -596,31 +596,31 @@ def facebook_tab_url_lazy():
...
 
-DONATE_SPANISH_LINK = ('https://donate.mozilla.org/es/?presets=100,50,25,15&amount=50'
+DONATE_SPANISH_LINK = ('https://donate.mozilla.org/es/?amount=50'
     '&ref=EOYFR2015&utm_campaign=EOYFR2015&utm_source=mozilla.org'
     '&utm_medium=referral&utm_content={source}&currency=eur')
 
 DONATE_LOCALE_LINK = {
     'default': (
-        'https://sendto.mozilla.org/page/contribute/Give-Now?source={source}'
+        'https://donate.mozilla.org/en-US/?source={source}'
```
flodolo: IMO the default should not have en-US

flodolo:
My personal feeling is that the DONATE_LOCALE_LINK thing is not the right approach (btw, it fails for Italian, since it doesn't link to https://donate.mozilla.org/it/ and €).

What's the difference between nn-NO and nb-no? Only the locale code.

So, it would make more sense to have a mapping like
```
DONATE_LOCALE_LINK = {
    'default': {
        'amount': 50,
        'currency': 'eur',
    },
   'en-US': {
        'amount': 50,
        'currency': 'usd',
    },
   'nb-NO': {
        'amount': 50,
        'currency': 'nok',
    },
...
```

And then use the `donate_url` helper function to build the link, including the current LOCALE in the URL.

bedrock/settings/base.py
schalkneethling: I actually agree and was thinking that, because when they do get the locale switching implemented we will not be forcing en-US with the default. Because there was a note added to always add a locale I thought it best to leave it in. With that said, it will currently fallback to en-US anyhow so.... I reckon dropping that off is fine.

schalkneethling:
The reason `it` fails to load for the correct locale is because we do not have an entry for it :-/ Which I found strange indeed. I do agree though but, perhaps we need to rethink/rework this mechanism in a separate bug?

Although I am open to discuss it here as well.

flodolo:
TBH I think this is the right bug to fix it, no point in replacing a broken link with another broken link.

flodolo:
Also note that random locales (or not available) will fallback to en-US
https://donate.mozilla.org/klingon/

schalkneethling:
@flodolo Fair enough, this change will be needed for the takeover work also I reckon. Let me have a look at this and make additional changes.

ScottDowne:
@flodolo 

Currently, random or not known locale fallback to the first browser set locale, if that's not known, it falls back to en-US.

We have an open bug to have this fallback to the next browser set locale, until we find a known one.

flodolo:
@ScottDowne That was my point. 

IMO right now the right thing to do is pass a locale code explicitly: if we simply use donate.mozilla.org we fail for too many locales (Italian, all Spanish variants). If we pass the wrong locale, we still fallback to en-US, so we're good.

When locale detection is fixed on the other side, we can think about removing the locale code and let d.m.o do its work.

bedrock/mozorg/tests/test_helper_misc.py
```diff
@@ -24,15 +24,15 @@
 
 TEST_DONATE_LOCALE_LINK = {
     'default':
-        'https://sendto.mozilla.org/page/contribute/Give-Now?source={source}',
+        'https://donate.mozilla.org/en-US/?source={source}',
     'en-US':
-        'https://sendto.mozilla.org/page/contribute/givenow-seq?'
-        'preset=2&source={source}&ref=EOYFR2014&utm_campaign=EOYFR2014'
-        '&utm_source=mozilla.org&utm_medium=referral&utm_content={source}',
+        'https://donate.mozilla.org/?amount=50'
```
ScottDowne: Should we be explicit with en-US here, and not rely on defaults where we can help it?

schalkneethling: We can be explicit. @flodolo also suggested that we rethink the way we build these URLs and perhaps use something more like https://github.com/mozilla/donate.mozilla.org/blob/master/data/currencies.js

Basically we should be able to inject more than just the `{source}`, ideally we should also do `{locale}` and `{currency}'. Perhaps we can also set thing like presets and defaults but, then we probably want to always set them unless, passing a parameter such as `presets=` would be fine on the donate.m.o side.

ScottDowne: Yeah, I think I agree. If you did something like that, we could get Andrea to give us a list of slightly different amounts and presets for mozilla.org, which might perform better.

I would also suggest to set the default amount to be the same as the second (middle button) in the preset.

schalkneethling:
@flodolo Ok, before I get to far into adding all the locales in `base.py` you wanna have a look at what I have there already? Also, I have some questions such as, which countries should all use Euro and which uses GBP?

Let me know your thoughts, thanks!

bedrock/mozorg/helpers/misc.py
```diff
@@ -349,11 +349,11 @@ def press_blog_url(ctx):
...
 def donate_url(ctx, source=''):
-    """Output a link to the donation page taking locales into account.
+    """Output a link to the donation page formatted using settings.DONATE
 
-    Uses the locale from the current request. Checks to see if we have
-    a donation page that match this locale, returns the localized page
-    url or falls back to the US page url if not.
+    Formats the donation url using custom values defined in settings.DONATE for
+    the curent locale. If a locale is not defined, it falls back to default, except
+    for the locale, which is always the locale from the request.
```
flodolo: Typo: current. Kind of hard to follow the second paragraph (do you really need the part after except)

bedrock/mozorg/helpers/misc.py
```diff
@@ -365,33 +365,31 @@ def donate_url(ctx, source=''):
...
-    if locale not in settings.DONATE_LOCALE_LINK:
-        if 'default' in settings.DONATE_LOCALE_LINK:
-            locale = 'default'
-        else:
-            locale = 'en-US'
+    index = locale
+
+    # if the current locale is not available in settings.DONATE,
+    # we want to fallback to the default params but, still pass
+    # the current locale in the URL
```
flodolo: Maybe add a note why we're doing it (waiting for donate.mozilla.org to fix the issue with locale detection).

bedrock/settings/base.py
```diff
@@ -1,3 +1,6 @@
+#!/usr/bin/env python
```
flodolo: Do you need the shebang here? It's not an executable file

bedrock/settings/base.py
flodolo: I guess you don't need this anymore, right?

flodolo:
Makes sense to me, even if I don't think we need the currency symbol?

Check https://github.com/mozilla/donate.mozilla.org/blob/master/data/locale-data.js for currency+locale (GBP is only en-GB)

schalkneethling:
Thx @flodolo - I was thinking the currency symbol could be useful for the front page takeover but, perhaps we do not really need it there? We could pass it to the template from the view based on locale.

flodolo:
You would know how to write the currency symbol, not how to write the string: before or after the number? With a space or without? Normal or non-breaking?

bedrock/mozorg/helpers/misc.py
schalkneethling: That is what happens when someone talks to you while you think and type ;p I will update it to make more sense.

bedrock/settings/base.py
schalkneethling: Nope, I was leaving the removal until the end.

bedrock/settings/base.py
schalkneethling: It is for the currency symbols. When running the tests it fails. Seeing that having the symbols here is not going to be of much use, I am going to remove them and then these lines as well.

schalkneethling:
@flodolo Indeed, it is a lot more complex than just knowing the symbol. The current implementation we are using for the takeover will probably work best then.

schalkneethling:
@flodolo @ScottDowne this is ready for review.

bedrock/mozorg/helpers/misc.py
```diff
@@ -365,33 +361,33 @@ def donate_url(ctx, source=''):
...
-            locale = 'default'
-        else:
-            locale = 'en-US'
+    index = locale
+
+    # if the current locale is not available in settings.DONATE,
+    # we want to fallback to the default params but, still pass
+    # the current locale in the URL. This is a workaround whilst
+    # donate.m.o does not handle automaic locale negotiation.
+    # see https://bugzilla.mozilla.org/show_bug.cgi?id=1227199#c5
```
flodolo:
```
# If the current locale is not available in settings.DONATE,
# fallback to the default params but still pass the current 
# locale in the URL. This is a temporary workaround since 
# donate.mozilla.org does not fully handle automatic
# locale negotiation.
```

bedrock/settings/base.py
```diff
@@ -596,93 +596,587 @@ def facebook_tab_url_lazy():
...
-        'https://sendto.mozilla.org/page/content/paypal-donate-thb/?'
-        'source={source}&ref=EOYFR2014&utm_campaign=EOYFR2014'
-        '&utm_source=mozilla.org&utm_medium=referral&utm_content=PPcurrency_THB'
-    ),
+DONATE_LINK = ('https://donate.mozilla.org/{locale}/?presets={presets}'
+    '&amount={default}&ref=EOYFR2015&utm_campaign=EOYFR2015'
+    '&utm_source=mozilla.org&utm_medium=referral&utm_content={source}'
+    '&currency={currency}')
+
+DONATE_PARAMS = {
```
flodolo: Please double check the list of locales with the locales actually supported on Bedrock, no point adding locales that will never be available on mozilla.org. Where did you get the current list?

From a very quick check (there might be others):
* af-ZA -> af
* au (not even a locale, it's a region, and we don't support en-AU anyway)
* ca-ES -> ca
* ch (again, not a locale)
* en-CA
* en-NZ
* en-PH
* et-EE -> et
* eu-ES
* fa-IR
* fl and fl-FL. Not even sure what this is… Did you mean Finnish (fi)?
* fr-FR, fr-CA
* fy
* hr-HR
* hu-HU
* id-ID
* ig-NG
* is-IS
* no
* xh-ZA

schalkneethling: Thanks @flodolo I am relying on your knowledge here. I have gone overboard on some I realize.

schalkneethling:
@flodolo ok, The locales are now only those in prod locales. How does this look?

flodolo:
To be completely honest I'm still not sure how you built this list.

The starting point should be these
https://github.com/mozilla/donate.mozilla.org/blob/master/data/currencies.js
https://github.com/mozilla/donate.mozilla.org/blob/master/data/locale-data.js

You should drop all the currencies that are not available on d.m.o

Drop: ach, af, az, be, bg, br, bs, ee, en-ZA, es-AR, es-CL, hr, hy-AM, ig, is, ka, kzt, km, ko, ln, mai, mk, my, ro, si, sq, sw, uk, ur, uz, vi, xh, yo, zh-CN, zh-TW, zu.

bn-BD and bn-IN should be 'in'

But please double check with the two files in d.m.o, given the size I probably missed some.

bedrock/settings/base.py
schalkneethling: How does the list look to you now @flodolo? Thanks!

bedrock/settings/base.py
```diff
@@ -596,93 +596,322 @@ def facebook_tab_url_lazy():
...
+        'currency': 'eur',
+        'presets': '100,50,25,15',
+        'default': '50'
+    },
+    'en-GB': {
+        'currency': 'gbp',
+        'presets': '150,100,50,25',
+        'default': '100'
+    },
+    'en-CA': {
```
flodolo: Remove this, not supported on Bedrock

bedrock/settings/base.py
```diff
@@ -596,93 +596,322 @@ def facebook_tab_url_lazy():
...
+        'currency': 'cad',
+        'presets': '150,100,50,25',
+        'default': '100'
+    },
+    'en-US': {
+        'currency': 'usd',
+        'presets': '100,50,25,15',
+        'default': '50'
+    },
+    'es': {
```
flodolo: No need for this, we redirect es to es-ES https://www.mozilla.org/es/

bedrock/settings/base.py
```diff
@@ -596,93 +596,322 @@ def facebook_tab_url_lazy():
...
+        'currency': 'inr',
+        'presets': '5000,3000,2000,1000',
+        'default': '3000'
+    },
+    'sk': {
+        'currency': 'eur',
+        'presets': '100,50,25,15',
+        'default': '50'
+    },
+    'si': {
```
flodolo: This is wrong. Did you mean 'sl' (Slovenian)? 'si' is Sinhala (India)

flodolo:
Minus a couple of nits, the list looks good to me (maybe ask someone from MoFo to double check if they're around?). Also, spot-checked it with de, it, sv-SE and everything works as expected.

bedrock/settings/base.py
schalkneethling: Indeed I did

bedrock/settings/base.py
schalkneethling: @ScottDowne want to give this a look over and see if there is anything you would like to change/remove

ScottDowne:
Any locales we don't currently support on d.m.o we can add.

They would initially be in English, but then we could at least add currency and presets.

schalkneethling:
@flodolo Thoughts?

flodolo:
I don't see any advantage in adding all locales, since we would just fall back to USD for the currency. 

We're passing the locale code nevertheless in the URL, even if a locale is not available in this array; only the params fallback to en-US.

schalkneethling:
@ScottDowne Does that work for you? If so, I reckon this is good to merge.

flodolo:
Note: mine doesn't really count as a proper code review, I'd suggest to have a second pair of eyes on it.

schalkneethling:
@pmclanahan r?

schalkneethling:
Thanks for all the help @flodolo and @ScottDowne

bedrock/foundation/templates/foundation/annualreport/2009/broadening-our-scope.html
```diff
@@ -183,11 +183,11 @@ <h4 class="section-title" id="drumbeat">
       <p>
         {% trans
           logo=static('img/foundation/annualreport/2009/drumbeat-logo.jpg'),
-          webcraft="http://www.drumbeat.org/",
-          festival="http://www.drumbeat.org/",
-          movies="http://www.drumbeat.org/",
-          subtitles="http://www.drumbeat.org/",
-          fund="https://donate.mozilla.org/"
+          webcraft="//www.drumbeat.org/webcraft",
```
pmclanahan: Why not specifically use "https://" links here instead of protocol relative? Means that it will be "http://" during local dev and could mask issues.

bedrock/mozorg/helpers/misc.py
```diff
@@ -365,33 +361,33 @@ def donate_url(ctx, source=''):
...
+    if locale not in settings.DONATE_PARAMS:
```
pmclanahan: NIT: Could just be:

```python
donate_url_params = settings.DONATE_PARAMS.get(locale, settings.DONATE_PARAMS['default'])
```

bedrock/settings/base.py
```diff
@@ -596,93 +596,312 @@ def facebook_tab_url_lazy():
...
-        'source={source}&ref=EOYFR2014&utm_campaign=EOYFR2014'
-        '&utm_source=mozilla.org&utm_medium=referral&utm_content=PPcurrency_THB'
-    ),
+DONATE_LINK = ('https://donate.mozilla.org/{locale}/?presets={presets}'
+    '&amount={default}&ref=EOYFR2015&utm_campaign=EOYFR2015'
+    '&utm_source=mozilla.org&utm_medium=referral&utm_content={source}'
+    '&currency={currency}')
+
+DONATE_PARAMS = {
+    'default': {
```
pmclanahan: This is mostly a nit, but I'd find it easier to understand and more explicit if this entry were `en-US` instead of `default`, then have the default for a locale that's not in the list be `en-US`. That way the code is clear that we're defaulting to en-US. Also it means that the majority of the traffic will go down the main code path (finding the locale in the dict) instead of the alternate branch (using the default).

pmclanahan:
Looks pretty good to me. Just a couple of comments I'm interested in your feedback on. 

r+wc

bedrock/foundation/templates/foundation/annualreport/2009/broadening-our-scope.html
schalkneethling: @alexgibson Has resolved these issues in a separate PR [https://github.com/mozilla/bedrock/pull/3608], the only change here is to use the `donate_url` helper for the `fund` variable

bedrock/mozorg/helpers/misc.py
schalkneethling: Thx @pmclanahan - If we remove `default` though, I guess we won't need the above at all

bedrock/settings/base.py
schalkneethling: I completely agree. The reason I had a `default` is two-fold. One, there was a default before (not a great reason) and two, I thought they might want to set the presets and default differently for people who fallback to `en-US`. I don't believe that is going to be the case so, I am going to make this change.

bedrock/foundation/templates/foundation/annualreport/2009/broadening-our-scope.html
alexgibson: It looks like you've accidentally reverted the changes when rebasing. This needs fixing again.

schalkneethling:
@pmclanahan updated, thanks!

bedrock/foundation/templates/foundation/annualreport/2009/broadening-our-scope.html
schalkneethling: Bah! Yes, there was a merge conflict and I wanted to keep my change to use the `donate_url` and did not notice I copied over that one as well. Fixing. Apologies for that @alexgibson

bedrock/foundation/templates/foundation/annualreport/2009/broadening-our-scope.html
```diff
@@ -183,11 +183,11 @@ <h4 class="section-title" id="drumbeat">
       <p>
         {% trans
           logo=static('img/foundation/annualreport/2009/drumbeat-logo.jpg'),
-          webcraft="http://www.drumbeat.org/",
-          festival="http://www.drumbeat.org/",
-          movies="http://www.drumbeat.org/",
-          subtitles="http://www.drumbeat.org/",
-          fund="https://donate.mozilla.org/"
+          webcraft="http://www.drumbeat.org/webcraft",
```
alexgibson: You're still reverting changes made in the other PR here. The only thing that needs to change is the donate link.

schalkneethling: :anguished: Fixing

alexgibson:
r+ from me (seeing as @pmclanahan already gave an r+wc, and we need this in master for the takeover PR), looks good. :hamburger:

bedrock/mozorg/helpers/misc.py
pmclanahan: We do need the above for locales not in the list.

pmclanahan: We'd just need to change it to: 

```python
donate_url_params = settings.DONATE_PARAMS.get(locale, settings.DONATE_PARAMS['en-US'])
```

schalkneethling: Ah, I thought it would just fallback to en-US. Will open a PR to fix that.

schalkneethling: Oh wait, I have this:
https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/helpers/misc.py#L377

pmclanahan: Oh yeah. That's perfect. All good :)

PR #3606, opened 2015-11-26
[fix bug 1228295] Set pager tabs on /firefox/android/ to disabled when inactive

alexgibson:
This is a minor code change to make functional testing easier, plus a (small) accessibility improvement.


schalkneethling:
r+ :pager:

PR #3607, opened 2015-11-26
Fix Bug 1228015, reduced font size on eoy home promo

schalkneethling:
@alexgibson @flodolo r? Main locales tested are:
en-gb, en-ca, de, pt-br, de, es-es, it and nl

### DIFF
--- a/media/css/mozorg/home/home-promo.less
+++ b/media/css/mozorg/home/home-promo.less
@@ -1073,7 +1073,8 @@ html[lang|="en"] {
 
     h2 {
         padding-top: 30px;
-        .font-size(32px);
+        .font-size(27px);
+        line-height: 1.2;
     }
 
     h2:before {
@@ -1088,6 +1089,7 @@ html[lang|="en"] {
 
     h3 {
         padding-top: 90px;
+        .font-size(24px);
     }
 
     .secondary {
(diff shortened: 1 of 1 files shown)


media/css/mozorg/home/home-promo.less
```diff
@@ -1105,6 +1107,27 @@ html[lang|="en"] {
     }
 }
 
+html[lang="de"],
+html[lang="it"] {
```
flodolo: I'm not seeing any problem with Italian and the current font-size. Are you?

media/css/mozorg/home/home-promo.less
```diff
@@ -1105,6 +1107,27 @@ html[lang|="en"] {
...
+    .promo-large-landscape.fundraiser {
+        h2 {
+            .font-size(24px);
+            line-height: 1.2;
```
flodolo: (not l10n) Do you need to repeat this?

flodolo:
I don't think we need Italian, but we definitely need to fix 'nl' when the tile is vertical

schermata 2015-11-26 alle 16 06 18

Probably a good idea to include es-ES. The text is currently on 3 lines, and the arrow sits right over the edge.

media/css/mozorg/home/home-promo.less
schalkneethling: It looked like the initial state had a really long second line so, thought it best to lessen the size. Perhaps not needed.

media/css/mozorg/home/home-promo.less
schalkneethling: Ha! Could have sworn I removed it.

alexgibson:
r+ :earth_asia:

PR #3609, opened 2015-11-26
[fix bug 1228355] Broken link on SmartOn Tracking page

alexgibson:
A tiny fix, but worthwhile given this page has a lot of traffic atm.


schalkneethling:
Good to know this is getting lots of traffic. r+ :traffic_light:

PR #3615, opened 2015-11-30
[fix bug 1208495] Update Mozilla news signup forms.

jpetto:


jpetto:
As per @jgmize's suggestion, will move removal of macro to separate PR to avoid race condition errors with rsync deployment.

alexgibson:
There may be a test on the homepage that needs updating here: https://github.com/mozilla/bedrock/blob/master/tests/functional/test_home.py#L51-L78

It may just be a case of updating the home page tests to use the existing newsletter macro that the contribute page test already uses: https://github.com/mozilla/bedrock/blob/master/tests/functional/test_home.py#L51-L78

alexgibson:
Looks like there are some occurrences of `mozorg_newsletter_form` that might not yet be covered in this PR? https://github.com/mozilla/bedrock/search?utf8=%E2%9C%93&q=mozorg_newsletter_form

jpetto:
Wow, 2 more instances of `mozorg_newsletter_form` were added since I first coded this change. Totally didn't expect/check for that when I rebased.

PR updated:
- Removed/replaced all instances of `mozorg_newsletter_form`
- Added `subtitle` param to `email_newsletter_form`
- Updated home page tests
- Added `mozorg_newsletter_form` macro back to account for possible deployment issues (will remove in follow-up PR)

alexgibson:
Code looks good, tests pass, forms all work and look nicely responsive locally, r+! :sake: 

The only reason I'm not merging this right now, is I'm not 100% certain if this will break any tests in the mcom-tests repo. I *think* we might be in the clear, but will need to take a look.

alexgibson:
It looks like we [removed the signup tests in mcom](https://github.com/mozilla/mcom-tests/pull/454) already, in favor of the bedrock tests. All that remains there is a link check, so I think we're good to merge this.

PR #3617, opened 2015-12-01
Fix Bug 1229244 - Users should be navigated to localized App Store

kyoshino:
@pmclanahan @jgmize r?

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -596,7 +596,7 @@
   {# If no App Store link is specified, use base with default "ct" param value #}
   {# See https://bugzilla.mozilla.org/show_bug.cgi?id=1196310#c23 #}
   {% if ios_link == '' %}
-    {% set ios_link = settings.APPLE_APPSTORE_FIREFOX_LINK + '&ct=mozorg' %}
+    {% set ios_link = firefox_ios_url('mozorg') %}
   {% endif %}
   {% set android_link = settings.GOOGLE_PLAY_FIREFOX_LINK %}
   <section id="send-to-device" class="{% if include_logo %}logo {% endif %}{% if include_title %}title{% else %}no-title{% endif %}">
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -277,8 +277,13 @@ class FirefoxIOS(ProductDetails):
         version = self.channel_map.get(channel, 'version')
         return self.mobile_details[version]
 
-    def get_download_url(self, channel='release', type=None):
-        return self.store_url
+    def get_download_url(self, channel='release', locale='en-US'):
+        countries = settings.APPLE_APPSTORE_COUNTRY_MAP
+
+        if locale in countries:
+            return self.store_url.format(country=countries[locale])
+
+        return self.store_url.replace('/{country}/', '/')
 
 
 firefox_desktop = FirefoxDesktop()
--- a/bedrock/firefox/templates/firefox/family/index.html
+++ b/bedrock/firefox/templates/firefox/family/index.html
@@ -51,7 +51,7 @@
       </div>
       <div class="message" id="dlbar-ios">
       {% if l10n_has_tag('tracking_protection') and waffle.switch('firefox-ios-global') %}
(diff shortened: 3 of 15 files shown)


craigcook:
Does Apple do a redirect on their end? Should we just use the general URL and let them sort it out?

I believe they do geolocation but I'm not sure if they also detect browser language, since they don't seem to distinguish language from country. Equating a language with a country is usually a bad idea. Just because someone is using a German browser doesn't mean they're in Germany, for instance. 

I'd worry about interfering with downloads if we assume a country based on a language and send someone to the wrong URL for their geolocation, depending on how these things are handled on Apple's end. This might be something to ask Mike Han about.

pmclanahan:
I'm inclined to agree with Craig, but I also can't get the appstore to detect the language I select, or automatically do anything based on country when I use e.g. a VPN in Germany. I also don't see much documentation around this issue, only recommendations from apple to just copy the URL out of the iTunes store :/ Maybe the URL component you're manipulating isn't country, but really is just language?

kyoshino:
Yeah, looks like Apple is not implementing redirects nor language negotiation at this moment. I know a language-country mapping is not the best solution, but it would be better to provide a link to localized page whenever possible from the UX perspective.

pmclanahan:
It depends on what that part of the appstore URL means. If it really means Country, then it could be bad UX because some apps aren't available in all countries. I know Fx is, but this could still be weird. If it's just language then this should be fine.

alexgibson:
I'm not entirely clear on how this works either, but I know that the version of the App Store you get is dependent on the country of the registered credit card you have associated with your account (so neither geolocation nor browser/phone language).

kyoshino:
Yes, so the URLs begin with `/jp/` (Japan) not `/ja/` (Japanese), `/us/` (United States) not `/en/` (English), etc.

craigcook:
Yeah, the URL indicates country but Apple seems to use that to mean language, so it could just be that Apple is "Doing It Wrong™" and the /jp/ URL really is just the Japanese translated page and has nothing to do with the user's country. As Alex said, since you have to have an AppleID before you can actually install anything, perhaps they do the geo/language stuff at that point. If that's the case, directing people to the `/jp/` URL if their language is `ja` should do no harm (other than perpetuating the "country = language" anti-pattern).

My worry is that, for example, someone in Switzerland using a French browser would be directed to the France App store and wouldn't be allowed to install because they're not in France. Hopefully Apple doesn't do that, so if we can confirm that people can install the app in their language from the App store for their country, regardless of what localized page they visit, we'll be in the clear.

alexgibson:
> Hopefully Apple doesn't do that, so if we can confirm that people can install the app in their language from the App store for their country, regardless of what localized page they visit, we'll be in the clear.

What I do know is that to install an app on a different store, you need to switch the country associated with your account (which the user can do). This does mean however, losing all the purchased apps you have installed via your current country/language, and also setting up a new credit card valid for that country in order to purchase again. It's not a user friendly system.

kyoshino:
As @craigcook said in the meeting today, apparently iTunes Preview page's country code is not related to the user's country selection on iTunes. So, even when I click the "View in iTunes" button on the Japanese page, iTunes opens the Canadian page because I have selected Canada as my country.

The `APPLE_APPSTORE_COUNTRY_MAP` lists 51 countries but only the following 11 pages provide localized content so far. I'll comment out others so they will fall back to the default (English) page.

https://itunes.apple.com/ar/app/apple-store/id989804926
https://itunes.apple.com/br/app/apple-store/id989804926
https://itunes.apple.com/cl/app/apple-store/id989804926
https://itunes.apple.com/de/app/apple-store/id989804926
https://itunes.apple.com/es/app/apple-store/id989804926
https://itunes.apple.com/fr/app/apple-store/id989804926
https://itunes.apple.com/it/app/apple-store/id989804926
https://itunes.apple.com/jp/app/apple-store/id989804926
https://itunes.apple.com/mx/app/apple-store/id989804926
https://itunes.apple.com/pt/app/apple-store/id989804926
https://itunes.apple.com/ru/app/apple-store/id989804926

One more thing: On Android devices, download links are opened with the Google Play app thanks to the [`market` pseudo-protocol](https://github.com/mozilla/bedrock/blob/master/media/js/base/global.js#L55-L64). On iOS devices, `itms-apps` can be used. I'll combine the change in this PR for easier testing.

PR #3618, opened 2015-12-01
Bug 1227250 - Update preset amounts for donation form

flodolo:
Based on https://github.com/mozilla/donate.mozilla.org/blob/master/data/currencies.js and considering the 2nd value as default

The last assumption needs to be confirmed


flodolo:
Also removed Persian, there's no support for irr on d.m.o.

PR #3619, opened 2015-12-01
Fix Bug 1229448, lowers donations for en-US and passes presets in URL

schalkneethling:
@flodolo r?


flodolo:
LGTM, waiting to see if we need to update the defaults for German

flodolo:
Based on @m-alexis comment you need to update the amounts for German too: 100,50,25,15
Not sure if it makes sense to update all 'eur' locales accordingly.

(also need to update the commit message and PR subject at this point)

flodolo:
Replaced by #3620

PR #3620, opened 2015-12-01
Bug 1229448 - Send presets in take-over donate form, update defaults

flodolo:


flodolo:
Noticed that before #3605 we had different values for de, es-*, fr, and pt-BR so updated all 'eur' and 'pt-BR' defaults.

craigcook:
Looks good to me. Verified locally that I'm seeing the updated values for different locales and that the presets and my selection are passed through to donate.mozilla.org. r+ :money_with_wings:

PR #3621, opened 2015-12-01
Fix Bug 1227311 - legal-docs Hello Privacy Notice commits for production

kyoshino:
Merging changes on legal-docs as per @mozcmcgill’s request.


jgmize:
thanks @kyoshino

PR #3623, opened 2015-12-02
Fix Bug 1214038, update to SVG and adds GA parameters

schalkneethling:
@alexgibson r?


alexgibson:
Could the updated SVG be optimized at all in https://jakearchibald.github.io/svgomg/ ?

Also seeing as it's referenced in CSS, you may need to add a cache buster comment to the Less file such as `/*! 2015-12-02 */`

Is it worth also rewording the commit message a little, to it make it a bit cleared on what this is updating? e.g. `Fix bug 1214038 - Add GA parameters and update SVG on homepage takeover`

alexgibson:
r+wc :banana:

PR #3625, opened 2015-12-02
Provide PageRegions with a reference to the page they are instantiated from

davehunt:
Addressing a limitation where page regions cannot locate and interact with elements outside of their scope. This allows us to avoid passing the base URL and selenium object for every page region (they're now extracted from the page object) and also solves the problem of referencing page methods/properties from a page region.

### DIFF
--- a/tests/pages/firefox/os/tv.py
+++ b/tests/pages/firefox/os/tv.py
@@ -18,12 +18,12 @@ class TVPage(FirefoxBasePage):
 
     @property
     def screens(self):
-        return [Screens(self.base_url, self.selenium, root=el) for el in
+        return [Screens(self, root=el) for el in
                 self.find_elements(self._screens_locator)]
 
     @property
     def thumbnails(self):
-        return [Thumbnails(self.base_url, self.selenium, root=el) for el in
+        return [Thumbnails(self, root=el) for el in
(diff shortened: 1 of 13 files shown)


davehunt:
Pinging @alexgibson for review and @bobsilverberg for feedback.

bobsilverberg:
Nice! I like it.

tests/pages/firefox/os/tv.py
```diff
@@ -71,5 +63,6 @@ def is_selected(self):
         return self.find_element(self._link_locator).get_attribute('class') == 'selected'
 
     def click(self):
+        screen = next(s for s in self.page.screens if s.is_displayed)
```
alexgibson: Nice :+1:

alexgibson:
I notice there's one test failing in `test_android.py`: 

`test_next_previous_buttons`
`TypeError: __init__() got multiple values for keyword argument 'root'`

Looks like the `PageRegion` needs updating there also. Otherwise, r+wc from me!

davehunt:
@alexgibson Ah yes, that one came in after a rebase and I didn't think to recheck or run the full suite again. It should be fixed now.

alexgibson:
Thanks @davehunt r+

PR #3628, opened 2015-12-03
Add functional tests for /firefox/new/ page

alexgibson:
These are tests for the page functionality when clicking the download button / landing on the page. As discussed in Orlando, we're going to leave the download link tests to the LinkChecker job for download pages.


alexgibson:
@davehunt r?

tests/functional/firefox/test_new.py
```diff
@@ -0,0 +1,27 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.firefox.new import FirefoxNewPage
+
+
+# We don't currently show the download button for up-to-date Firefox users.
+# TODO explore how could we run this test using a known out-of-date Firefox?
+@pytest.mark.skip_if_capabilities_include({'browserName': 'firefox'})
```
davehunt: Seeing this again, I wonder if it would be better to have marks like `skip_if_firefox`, which can then call common code with appropriate capabilities values to determine if we should skip. This would still give the same flexibility, at the cost of writing code to deal with these marks, but would make the tests more readable. We could then have something like `skip_if_latest_firefox` to assist with the TODO. These marks could then be combined on a single test, such as:

```python
@pytest.mark.skip_if_not_firefox
@pytest.mark.skip_if_latest_firefox
def test_click_download_button(base_url, selenium):
    page = FirefoxNewPage(base_url, selenium, params='').open()
    assert page.is_download_button_displayed
    page.click_download_button()
    assert page.is_thank_you_message_displayed
```

Of course, there are many ways we could slice this. Another option is to provide marks like `skip_if_browser_name` and take the browser name as an argument. That's more readable than the current code, but has its own costs. Ultimately, I'm happy to land something that works, and iterate on it.

Another thing I'm a little uneasy about is that we're currently using the actual capabilities, which is the safest way to know we're skipping for the right reason, but does require us to start a Selenium session (and browser). If we trusted our requested capabilities, we could even skip starting the unnecessary browser instance.

If he's around, I'd love to hear @bobsilverberg's thoughts on this too.

tests/functional/firefox/test_new.py
```diff
@@ -0,0 +1,27 @@
...
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+def test_click_download_button(base_url, selenium):
+    page = FirefoxNewPage(base_url, selenium, params='').open()
+    assert page.is_download_button_displayed
```
davehunt: If it wasn't displayed, the click would raise an exception, so I think we can remove this assertion.

tests/pages/firefox/new.py
```diff
@@ -0,0 +1,50 @@
...
+        return [DownloadButton(self, root=el) for el in
+                self.find_elements(self._download_button_locator)]
+
+    @property
+    def platform_download_button(self):
+        return next(s for s in self.download_buttons if s.is_displayed)
+
+    @property
+    def is_download_button_displayed(self):
+        return self.platform_download_button.is_displayed
```
davehunt: I think we can get away without a page region for the download button. As far as I can tell we only really need to click the download button (which implicitly checks that it's visible). We could therefore do something like:

```python
def download_firefox(self):
    els = [el for el in self.find_elements(self._download_button_locator) if el.is_displayed]
    assert len(els) == 1, 'Expected one download button to be displayed'
    els[0].click()
```

tests/functional/firefox/test_new.py
```diff
@@ -0,0 +1,27 @@
...
+    page.click_download_button()
+    assert page.is_thank_you_message_displayed
+
+
+@pytest.mark.sanity
+@pytest.mark.nondestructive
+def test_directly_load_thank_you(base_url, selenium):
+    page = FirefoxNewPage(base_url, selenium, params='?scene=2#download-fx').open()
```
davehunt: I'm not crazy about the `params` but don't have a better suggestion at this time. Perhaps we could subclass `FirefoxNewPage` and just alter the URL? I think my main issue is that this level of details shouldn't be in the tests if at all possible.

tests/pages/firefox/new.py
alexgibson: Nice suggestion :+1:

tests/functional/firefox/test_new.py
alexgibson: Ok, I wasn't sure if a subclass would be overkill for just changing the URL - but I would be open to using one if preferred. Agree it is more suited to live in the page object.

alexgibson:
@davehunt - updated thanks. Only open question that remains is how best to deal with the markers for these tests, which would be good to talk through in our meeting later today.

tests/functional/firefox/test_new.py
davehunt: After further discussion, we're going to go ahead with custom markers for the various conditions as we identify them. Here's a gist of a working patch based on this PR: https://gist.github.com/davehunt/566957e5cac7a40352bb

alexgibson:
@davehunt updated, thanks - seems to work well!

tests/pages/firefox/new.py
```diff
@@ -0,0 +1,33 @@
...
+
+    @property
+    def is_thank_you_message_displayed(self):
+        return self.is_element_displayed(self._thank_you_message_locator)
+
+    def wait_for_thank_you_message(self):
+        self.wait.until(lambda s: self.is_thank_you_message_displayed)
+
+
+class FirefoxNewScene2Page(FirefoxNewPage):
```
davehunt: `FirefoxNewScene2Page` is a little obscure to me, but then I'm not familiar with the code for the application. From my perspective something like `FirefoxNewThankYouPage` would make more sense, but if this will make sense to those working on the tests, leave it as it is.

tests/pages/firefox/new.py
```diff
@@ -0,0 +1,33 @@
...
+        assert len(els) == 1, 'Expected one download button to be displayed'
+        els[0].click()
+        self.wait_for_thank_you_message()
+
+    @property
+    def is_thank_you_message_displayed(self):
+        return self.is_element_displayed(self._thank_you_message_locator)
+
+    def wait_for_thank_you_message(self):
+        self.wait.until(lambda s: self.is_thank_you_message_displayed)
```
davehunt: You could move this into `FirefoxNewScene2Page` by overriding `wait_for_page_to_load`. Then the `download_firefox` method could instantiate and return this page object.

```python
    super(FirefoxNewPage, self).wait_for_page_to_load()
    self.wait.until(lambda s: self.is_thank_you_message_displayed)
    return self
```

tests/pages/firefox/new.py
```diff
@@ -0,0 +1,33 @@
...
+
+from pages.firefox.base import FirefoxBasePage
+
+
+class FirefoxNewPage(FirefoxBasePage):
+
+    _url = '{base_url}/{locale}/firefox/new/'
+
+    _download_button_locator = (By.CSS_SELECTOR, '#features-download .download-button .download-link')
+    _thank_you_message_locator = (By.ID, 'scene2-main')
```
davehunt: This, and the method that uses it can be moved to `FirefoxNewScene2Page`

tests/pages/firefox/new.py
alexgibson: > Then the download_firefox method could instantiate and return this page object.

Hmm, this part won't work, as the parameters are not strictly the same after clicking the button, as opposed to hitting the thankyou page directly.

alexgibson:
@davehunt updated - not exactly what you suggested but a bit cleaner I think.

davehunt:
Nice! r+

PR #3631, opened 2015-12-03
Fix bug 1218516: Add New Relic browser monitoring

pmclanahan:
Respects DNT. In this first pass it's only being added to /firefox/new/


pmclanahan:
@jpetto @alexgibson @jgmize thoughts? r?

alexgibson:
Please make sure this doesn't make it's way into the tracking protection tour template, that's the only page on the site we should never include this on (irrespective of DNT being enabled or not). Pretty sure this is safe, but just adding a note so we make sure to check.

pmclanahan:
@alexgibson the actual JS is only included on the firefox/new template so far, but just to prevent any future mistakes I can go ahead and add an empty `newrelic` block to the tour template. Sound good?

bedrock/firefox/templates/firefox/tracking-protection-tour.html
```diff
@@ -136,3 +136,6 @@
 {% block site_js %}
   {% javascript 'tracking-protection-tour' %}
 {% endblock %}
+
+{#- Ensure that this page never gets the New Relic JS -#}
+{% block newrelic %}{% endblock %}
```
pmclanahan: This should be good, right @alexgibson ?

bedrock/base/templates/base-resp.html
```diff
@@ -7,6 +7,9 @@
   <head>
     <meta charset="utf-8">{# Note: Must be within first 512 bytes of page #}
 
+    {% javascript 'site' %}
+    {% block newrelic %}{% endblock %}
```
jpetto: Out of curiosity, any reason to move the JS up in the `<head>`? Does New Relic need to come in this early?

bedrock/base/templates/includes/newrelic.html
```diff
@@ -0,0 +1,10 @@
+{% if settings.NEWRELIC_APP_ID %}
```
jgmize: How about we change this to
```
{% if settings.NEWRELIC_APP_ID and settings.NEWRELIC_BROWSER_LICENSE_KEY %}
```
since both are used in the snippet below?

bedrock/firefox/templates/firefox/tracking-protection-tour.html
alexgibson: This is the right template, thanks @pmclanahan :+1:

alexgibson: Could actually move this up to the top of the template next to where we already block out GA?

bedrock/base/templates/base-resp.html
pmclanahan: It was [recommended in the docs](https://docs.newrelic.com/docs/browser/new-relic-browser/installation-configuration/adding-apps-new-relic-browser#copy-paste-app). Not sure how important it is.

bedrock/base/templates/includes/newrelic.html
pmclanahan: done.

bedrock/firefox/templates/firefox/tracking-protection-tour.html
pmclanahan: Done.

pmclanahan:
If we'd rather go ahead and enable for the full site (or at least every page that inherits from the base templates) now that we've identified and fixed the tour page then I'm fine with that too. We need the firefox/new page particularly for our #mozlando experiment, but other pages would be wonderful to have data on as well.

bedrock/base/templates/base-resp.html
jpetto: I can't think of any reason moving the JS block here would cause any problems. Should be fine. Could even potentially improve rendering, as platform will be known prior to parsing CSS (for platform-specific rules and such).

jgmize:
> If we'd rather go ahead and enable for the full site (or at least every page that inherits from the base templates) now that we've identified and fixed the tour page then I'm fine with that too. We need the firefox/new page particularly for our #mozlando experiment, but other pages would be wonderful to have data on as well.

+1, MOAR DATA

pmclanahan:
@jgmize done.

pmclanahan:
So... What do we think? Can we land this before Mozlando? I think it'd be very useful to have these data for various of the sprinty things, also just because data. At the very least I'd like to try to get it in master before we leave so that it can go out with the mozlando pushes.

jgmize:
I think this could be very useful for the performance sprint, and we didn't find any issues when we did a canary deployment earlier. Now that we have DNT support and the ability to exclude it entirely from any pages we need to, I'm comfortable with a merge and push this morning.

PR #3632, opened 2015-12-04
Upgrade rna for "rnasync --force": fix bug 1230452

jgmize:
@pmclanahan r?


pmclanahan:
I think in order for it to pick up the change (in legacy infra at least) you need to set a version on the end: `#egg=rna==1.0.0`. Otherwise I think peep will skip it as it sees `rna` as already installed.

pmclanahan:
r+wc

jgmize:
Good catch, added version 1.0 to the end, and also [tagged](https://github.com/mozilla/rna/releases/tag/1.0) version in rna repo.

PR #3633, opened 2015-12-04
Bug 1230334 - Private Browsing: fix CTA button alignment for longer translations

flodolo:
See bottom of the page at
https://www.mozilla.org/et/firefox/private-browsing/


flodolo:
Natural width would be 230px, not 220px at breakDesktop, too bad we have per-locale rules that add 10px to some locales (e.g. German).

schalkneethling:
r+ :flags:

PR #3635, opened 2015-12-04
Add functional tests for Firefox partners page

davehunt:
This replaces the mcom-tests and CasperJS tests for the Firefox partners page.

@alexgibson r?


tests/pages/firefox/partners.py
```diff
@@ -0,0 +1,168 @@
...
+
+        @property
+        def items(self):
+            els = self.find_elements(self._items_locator)
+            return [self.MenuItem(self.page, root=el) for el in els]
+
+        class MenuItem(FirefoxBasePageRegion):
+
+            def click(self):
+                self._root.click()
```
davehunt: Ideally I'd have a wait here for the appropriate section to be displayed. Unfortunately the only way I can see to do this is to link the values of the data-section attributes of two elements. If possible, adding a class or attribute to the menu item to indicate that it is the current item (after any animations) would allow me to improve this.

tests/functional/firefox/test_partners.py
```diff
@@ -0,0 +1,70 @@
...
+def test_devices_menu(base_url, selenium):
+    page = PartnersPage(base_url, selenium).open()
+    devices_page = page.devices_menu.show_devices()
+    assert devices_page.url in selenium.current_url
+
+
+@pytest.mark.nondestructive
+def test_firefox_os_menu(base_url, selenium):
+    page = PartnersPage(base_url, selenium).open()
+    page.partner_menu.firefox_os.click()
```
alexgibson: Should this be `page.partner_menu.show_firefox_os()`?

tests/functional/firefox/test_partners.py
```diff
@@ -0,0 +1,70 @@
...
+    for item in reversed(page.firefox_os_menu.items):
+        item.click()
+
+
+@pytest.mark.nondestructive
+def test_marketplace_menu(base_url, selenium):
+    page = PartnersPage(base_url, selenium).open()
+    page.partner_menu.marketplace.click()
```
alexgibson: Should this be `page.partner_menu.show_marketplace()`?

tests/functional/firefox/test_partners.py
```diff
@@ -0,0 +1,70 @@
...
+    for item in reversed(page.marketplace_menu.items):
+        item.click()
+
+
+@pytest.mark.nondestructive
+def test_android_menu(base_url, selenium):
+    page = PartnersPage(base_url, selenium).open()
+    page.partner_menu.android.click()
```
alexgibson: Should this be `page.partner_menu.show_android()`?

tests/functional/firefox/test_partners.py
```diff
@@ -0,0 +1,70 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.firefox.partners import PartnersPage
+
+
+@pytest.mark.nondestructive
```
alexgibson: Make this one a sanity test?

tests/pages/firefox/partners.py
```diff
@@ -0,0 +1,168 @@
...
+class PartnersPage(FirefoxBasePage):
+
+    _url = '{base_url}/{locale}/firefox/partners'
+    _firefox_os_menu_locator = (By.CSS_SELECTOR, '#os nav')
+    _marketplace_menu_locator = (By.CSS_SELECTOR, '#marketplace nav')
+    _android_menu_locator = (By.CSS_SELECTOR, '#android nav')
+    _phone_overview_screen_locator = (By.ID, 'screen-overview')
+    _phone_os_screen_locator = (By.ID, 'screen-os')
+    _phone_marketplace_screen_locator = (By.ID, 'screen-marketplace')
+    _android_phone_locator = (By.CSS_SELECTOR, '.android-phone-visible')
```
alexgibson: More of a question than a suggestion, but which is better to use here: `CSS_SELECTOR` or `CLASS_NAME`?

tests/pages/firefox/partners.py
alexgibson: I've taken a look through the code for this and it looks like the horizontal scrolling sections just have their `left` position values changed and use a CSS transition to animate. We could add in a `transitionend` event to catch the end of the animation to add a class, but then this also means browsers must support this event for the test to pass. It looks like the code uses a `data-current` attribute to track which section is active, could we use this instead perhaps?

alexgibson: I know this means not waiting for the animation to finish, but I don't think that's necessarily a big issue here for these particular tests.

tests/functional/firefox/test_partners.py
davehunt: D'oh! I must have changed my mind at the last minute. :smile:

tests/pages/firefox/partners.py
davehunt: It's a good question, and I tend to use `CLASS_NAME` in cases like this (I'll update it). Amusingly, must of these variants just get converted to CSS selectors in the Selenium client.

tests/pages/firefox/partners.py
davehunt: Yeah, I don't think we need to wait for this transition at this point.

davehunt:
Ready for review again @alexgibson.

tests/functional/firefox/test_partners.py
```diff
@@ -0,0 +1,71 @@
...
+    assert devices_page.url in selenium.current_url
+
+
+@pytest.mark.nondestructive
+def test_firefox_os_menu(base_url, selenium):
+    page = PartnersPage(base_url, selenium).open()
+    page.partner_menu.show_firefox_os()
+    for item in reversed(page.firefox_os_menu.items):
+        item.click()
+
```
alexgibson: Should we be asserting something after each click for these tests? e.g. that the section is current once clicked

davehunt:
Review comments addressed. @alexgibson r?

alexgibson:
LGTM, r+

PR #3639, opened 2015-12-04
Fix Bug 1224614, adds eoy fundraising CTA to post heartbeat landing pages

schalkneethling:
@pmac r? Note: This should not be merged before December 15th, but is good for code review before then.

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -649,14 +649,31 @@ class HelloStartView(LatestFxView):
 
 class FeedbackView(TemplateView):
 
+    donate_url = ('https://donate.mozilla.org/'
+       '?ref=EOYFR2015&utm_campaign=EOYFR2015'
+       '&utm_source=Heartbeat_survey&utm_medium=referral'
+       '&utm_content=Heartbeat_{0}stars')
+
+    def get_score(self):
+        return self.request.GET.get('score', 0)
+
     def get_template_names(self):
-        score = self.request.GET.get('score', 0)
+        score = self.get_score()
         if score > '3':
             template = 'firefox/feedback/happy.html'
         else:
             template = 'firefox/feedback/unhappy.html'
 
         return [template]
+
+    def get_context_data(self, **kwargs):
+        context = super(FeedbackView, self).get_context_data(**kwargs)
+        score = self.get_score()
+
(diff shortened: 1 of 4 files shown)


schalkneethling:
@jpetto r?

bedrock/firefox/views.py
```diff
@@ -643,8 +643,18 @@ class HelloStartView(LatestFxView):
 
 class FeedbackView(TemplateView):
 
+    donate_url = ('https://donate.mozilla.org/'
+    '?ref=EOYFR2015&utm_campaign=EOYFR2015'
+    '&utm_source=Heartbeat_survey&utm_medium=referral'
+    '&utm_content=Heartbeat_{0}stars')
+
+    valid_stars = ('3', '4', '5')
```
jpetto: I think this would be better as a list `['3', '4', '5']` rather than a tuple.

jpetto: Actually, we probably don't need this variable. This list is only needed in one spot, so it could just be used in place in `get_content_data`.

bedrock/firefox/views.py
```diff
@@ -652,6 +662,15 @@ def get_template_names(self):
 
         return [template]
 
+    def get_context_data(self, **kwargs):
+        context = super(FeedbackView, self).get_context_data(**kwargs)
+        score = self.get_score()
+
+        if score in self.valid_stars:
```
jpetto: `if score in ['3', '4', '5']:`

bedrock/firefox/views.py
```diff
@@ -643,8 +643,18 @@ class HelloStartView(LatestFxView):
 
 class FeedbackView(TemplateView):
 
+    donate_url = ('https://donate.mozilla.org/'
+    '?ref=EOYFR2015&utm_campaign=EOYFR2015'
+    '&utm_source=Heartbeat_survey&utm_medium=referral'
+    '&utm_content=Heartbeat_{0}stars')
```
jpetto: Small nit - indent 3 lines below `donate_url`.

jpetto:
Couple small suggestions, but overall looks good. r+wc

schalkneethling:
@jpetto updated

jpetto:
r+ :stars: :mount_fuji:

PR #3643, opened 2015-12-08
Fix Bug 1226830, adds firstrun variants

schalkneethling:
@alexgibson mozlando r? :) thx

### DIFF
--- a/bedrock/firefox/templates/firefox/firstrun/firstrun-e.html
+++ b/bedrock/firefox/templates/firefox/firstrun/firstrun-e.html
@@ -1,0 +1,98 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% add_lang_files "firefox/whatsnew_38" %}
+
+{% extends "firefox/base.html" %}
+
+{% block extra_meta %}<meta name="robots" content="noindex">{% endblock %}
+
+{#- This will appear as <meta property="og:title"> which can be used for social share -#}
+{% block page_og_title %}
+  {{_('Choose the independent browser')}}
+{% endblock %}
+
+{#- This will appear as <meta property="og:description"> which can be used for social share -#}
+{% block page_og_desc %}
+  {{_('Firefox is non-profit, non-corporate, non-compromised. Choosing Firefox isn’t just choosing a browser. It’s a vote for personal freedom.')}}
+{% endblock %}
+
+{#- Override <meta property="og:url"> for social share -#}
+{% block page_og_url %}{{ url('firefox') }}{% endblock %}
+
+{% block body_class %}mozID variant-e{% endblock %}
+
+{% block site_css %}
+  {% stylesheet 'firefox_variants_firstrun' %}
+{% endblock %}
+
+{% block optimizely %}
+  {% if waffle.switch('firefox-firstrun-optimizely') %}
+    {% include 'includes/optimizely.html' %}
+  {% endif %}
+{% endblock %}
+
+{% block site_header %}
+  <header id="masthead">
+    {% block tabzilla_tab %}
+    <div id="tabzilla">
+      <a href="{{ url('mozorg.home') }}">Mozilla</a>
+    </div>
+    {% endblock %}
+    {% block site_header_nav %}{% endblock %}
+    {% block site_header_logo %}{% endblock %}
+    {% block breadcrumbs %}{% endblock %}
+  </header>
+{% endblock %}
+
+{% block content %}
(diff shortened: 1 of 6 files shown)


schalkneethling:
@craigcook If you have a moment, else Alex will grab it once he is free. Basically each template needs to look good at standard large screen sizes and then 800x600, Firefox only. Thanks!

schalkneethling:
To access the variations, access the following URLs:

en-US/firefox/firstrun/?v=e
en-US/firefox/firstrun/?v=f
en-US/firefox/firstrun/?v=g

chrismore:
@alexgibson @craigcook any chance we can get a review on this code before 10am on Wednesday? We need this code reviewed, any changes made, and pushed by 11am. Thanks!

craigcook:
I'm on it.

chrismore:
Thanks, Craig!

media/css/firefox/firstrun/variants.less
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    .leftcol {
+        background-color: #fff;
+        padding: @baseLine 35px 20px @baseLine;
+        border-radius: 6px;
+        width: 40%;
+    }
+
+    @media only screen and (max-width: 800px) {
```
craigcook: Suggest setting this breakpoint at `max-width: @widthDesktop` so it kicks in before we hit 800px. Otherwise we get horizontal scrolling and cutoff text anywhere between 1000px and 800px.

media/css/firefox/firstrun/variants.less
```diff
@@ -0,0 +1,162 @@
...
+
+        #outer-wrapper {
+            min-width: auto;
```
craigcook: I think `min-width: auto` may only work in combination with flexbox. Otherwise it could just be `min-width: 0`. But if this seems to work either way it may not matter.

media/css/firefox/firstrun/variants.less
```diff
@@ -0,0 +1,162 @@
...
+    }
+    .fxaccounts {
+        background-color: #fff;
+        margin: 0 auto;
+        padding: @baseLine;
+        border-radius: 8px;
+        width: 400px;
+    }
+
+    @media only screen and (max-width: 800px) {
```
craigcook: Same as above, try setting this at `@widthDesktop` (1000px)

bedrock/firefox/templates/firefox/firstrun/firstrun-e.html
```diff
@@ -0,0 +1,104 @@
...
+      {{ high_res_img('firefox/australis/fx38.0.5/firefox-logo.png', {'alt': 'Firefox', 'width': '142', 'height': '147'}) }}
+    </div>
+
+    {#  form left #}
+    <div class="leftcol">
+      <iframe id="fxa" scrolling="no" data-src="{{ settings.FXA_IFRAME_SRC }}?utm_campaign=fxa-embedded-form&amp;utm_medium=referral&amp;utm_source=firstrun&amp;utm_content=fx-{{ version }}&amp;entrypoint=firstrun&amp;service=sync&amp;context=iframe&amp;style=chromeless&amp;haltAfterSignIn=true"></iframe>
+    </div>
+
+    {#  copy right #}
+    <div class="rightcol">
```
craigcook: Just a thought: it would be nice if the copy came before the iframe in the source order, though I know that doesn't make floats easy. Since the iframe is a fixed width anyway, we could pad a container and position the iframe into the padded space, something like:

`.wrapper { position: relative; padding-left: 440px; }`
`.frame { width: 400px; position: absolute; left: 0; top: 0; }`

A bonus is that we could then let the content stack in small viewports (I know we're targeting 800px for now but this would let us accommodate even smaller).

craigcook:
Looking good overall, just needs those small responsive tweaks for variant e, which is the tricky one (f and g are good as-is).

chrismore:
Thanks @craigcook, @schalkneethling will get on the changes asap and we'll have a revised PR ready for review this morning.

chrismore:
@bensternthal, if all goes well here and we get an r+, we should be good to merge/push at 11am today.

media/css/firefox/firstrun/variants.less
schalkneethling: Good point. I noticed the horizontal scrollbar but though it was "ok". This is much better though.

schalkneethling:
@craigcook Updated. Thanks!

media/css/firefox/firstrun/variants.less
```diff
@@ -0,0 +1,163 @@
...
+    }
+
+    @media only screen and (max-width: @breakDesktop) {
+
+        #outer-wrapper {
+            min-width: auto;
+        }
+
+        #wrapper {
+            max-width: 800px;
```
craigcook: Add a `margin: 0 auto;` here to center it in the window.

craigcook:
One final tweak and we're good. r+wc

PR #3645, opened 2015-12-08
Font performance variant for /firefox/new/

jpetto:
Displays copy while font files are loading. Uses system font for devices less than 760px wide.

@alexgibson r?


jpetto:
Tested in Fx, Chrome, Safari, IE 10, 11, & Edge. IE 7, 8, 9 VMs all failed to connect to internet. :disappointed:

alexgibson:
@jpetto looks like we have some redirect failures ^ :(

alexgibson:
I managed to test in IE9 and fonts loaded async as expected :+1:

alexgibson:
fontfaceobserver.js seems to throw an `unexpected identifier` error in IE8. Should this perhaps need to be in a separate bundle? Would it be good to test this on a demo server perhaps?

media/js/firefox/new-v1.js
```diff
@@ -0,0 +1,43 @@
...
+        if (mq.matches) {
+            var fntOpenSans = new FontFaceObserver('Open Sans', {
+                weight: 'normal'
+            });
+            var fntOpenSansBold = new FontFaceObserver('Open Sans', {
+                weight: 'bold'
+            });
+            var fntOpenSansLight = new FontFaceObserver('Open Sans Light');
+
+            Promise.all([fntOpenSans.check(), fntOpenSansLight.check()]).then(function() {
```
alexgibson: Docs suggest we may need to include a `Promise` polyfill for browsers that don't have native support: https://github.com/bramstein/fontfaceobserver. Do we need to do this?

alexgibson:
Seems to work well nice - including the library seems to throw an error in IE < 9, which we probably want to avoid. Also one question about whether we need a Promises polyfill. Aside from that, things look good :+1: 

The test failures can hopefully be fixed too

bedrock/firefox/views.py
```diff
@@ -682,3 +683,18 @@ def get_template_names(self):
...
+@xframe_allow
+def new(request):
+    locale = request.locale
+    variant_locales = ['en-US', 'de']
+
+    # check for variant in querystring
+    variant = request.GET.get('v', '')
+
+    # ensure variant is one of 2 accepted values
+    if (variant not in ['1', '2'] or locale not in variant_locales):
```
alexgibson: Is variant 2 being used?

media/js/firefox/new-v1.js
jpetto: According to [the docs](https://github.com/bramstein/fontfaceobserver#installation), the version of `fontfaceobserver.js` used includes a small promise polyfill, so should be good there.

bedrock/firefox/views.py
jpetto: Variant 2 not being used yet, but likely will be early next week/at next push.

alexgibson:
Tested IE6-8 and page functions as expected. Tested IE9-10, Firefox, Chrome, Safari and font's load async as expected :+1:

jpetto:
Closing in favor of better named #3647.

PR #3651, opened 2015-12-10
[fix bug 1231201] /firefox/new/ performance test v2/4.

jpetto:
- v2: Custom/slim sandstone & buttons CSS
- v4: Async fonts, custom sandstone, no optimizely

### DIFF
--- a/bedrock/firefox/templates/firefox/new.html
+++ b/bedrock/firefox/templates/firefox/new.html
@@ -36,8 +36,15 @@
 {% block site_css %}{% endblock %}
 
 {% block extrahead %}
+  {# load fonts async #}
   {% if variant == '1' %}
     {% stylesheet 'firefox_new_v1' %}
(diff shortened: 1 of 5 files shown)


jpetto:
Shortcut for reviewer - `new-v2.less` is nearly identical to pre-existing `new.less`. Only change is removing three unused selectors.

schalkneethling:
Not introduced by this PR, but when the "Your system doesn't meet...." it is offset to the left. Seeing that everything else is centered, perhaps we should center this as well?

screen shot 2015-12-15 at 10 58 53

This uses the follow class values on the html element:
`oldmac armv7 js arm armv7up loaded`

Screen size is 320 x 533

schalkneethling:
r+wc :speedboat:

jpetto:
Opened #3660 to address unsupported platforms UX issue.

schalkneethling:
r+ :chocolate_bar:

PR #3652, opened 2015-12-10
[fix bug 1231693] Remove reference to private-browsing meta image

alexgibson:


jpetto:
r+, with consternation.

PR #3655, opened 2015-12-10
Turn off New Relic browser auto-instrumentation

pmclanahan:


pmclanahan:
@jgmize @glogiotatidis according to NR support if this is "false" then turning on "Pro" features in the UI won't auto-instrument and we can get rid of the confusing messaging.

PR #3657, opened 2015-12-14
Update to the latest stable test requirement package versions

davehunt:
Upgrades us to pytest-selenium with improved error handling. I also updated the other packages while I was there.


davehunt:
@alexgibson r?

alexgibson:
r+ :sparkler:

PR #3659, opened 2015-12-15
Fix Bug 1229614, set winner from experiment as default and make page responsive

schalkneethling:
@sgarrity r?


pmclanahan:
NIT: keep tense in commit message active. something like "set winner from experiment as default. make page responsive"

bedrock/firefox/tests/test_base.py
```diff
@@ -606,7 +606,7 @@ def test_fx_firstrun_40_0(self, render_mock):
         req = self.rf.get('/en-US/firefox/firstrun/')
         self.view(req, version='40.0')
         template = render_mock.call_args[0][1]
-        eq_(template, ['firefox/australis/fx40/firstrun.html'])
+        eq_(template, ['firefox/firstrun/firstrun.html'])
```
sgarrity: @schalkneethling Why is this part of the PR?

sgarrity:
@schalkneethling It looks like resulting `media/css/firefox/firstrun/firstrun.less` may be missing styles that were part of the winning variation that originally came from `/media/css/firefox/australis/fx40/firstrun.less`.

For example, the main headline font size appears incorrect due to this line being missing: 
https://github.com/mozilla/bedrock/blob/master/media/css/firefox/australis/fx40/firstrun.less#L30

I think the new firstrun.less should consist of the combined contents of `/media/css/firefox/australis/fx40/firstrun.less` *AND* the `.variant-g` portions of `variants.less`. Let me know if I'm off base.

In production, I see:
screen shot 2015-12-15 at 16 30 53

Testing locally with your PR, I see:
screen shot 2015-12-15 at 16 31 00

bedrock/settings/static_media.py
```diff
@@ -358,12 +358,11 @@
...
-    'firefox_variants_firstrun': {
+    'firefox_firstrun': {
         'source_filenames': (
             'css/sandstone/sandstone.less',
             'css/tabzilla/tabzilla-static.less',
-            'css/firefox/australis/fx40/firstrun.less',
-            'css/firefox/firstrun/variants.less',
+            'css/firefox/firstrun/firstrun.less',
         ),
         'output_filename': 'css/firefox_fx40_firstrun-bundle.css',
```
craigcook: I failed to catch this in my review of the variants last week, but this output bundle name is the same as line 346 above. If this new firstrun bundle is fully replacing the fx40 firstrun styles then we should delete the previous block in static_media, but if we need to keep the older fx40 styles and these are a new set for newer versions, the output bundle name should be different.

bedrock/firefox/tests/test_base.py
schalkneethling: So the default template has not changed to the one above so, the test fails if I do not change the template to the new one.

bedrock/settings/static_media.py
schalkneethling: Complete oversight on my part. Thanks @craigcook

schalkneethling:
@sgarrity thanks, I must have missed that. Will have a second look an make sure I have not missed other things.

schalkneethling:
Ok updated. At the moment I am not 100% sure whether we need to keep the previous version of firstrun around, although I am pretty confident we won't need to. With that though, I would like to get this into production as soon as we can so that we can take advantage of the increase in accounts signup we saw.

I can then confirm about the older versions and submit a PR to cleanup, if we are indeed removing it. Let me know your thoughts. Thanks!

chrismore:
@schalkneethling we don't need to keep the previous right-aligned form first run. This can be available as the default first run for anyone FIrefox 40 over newer.

schalkneethling:
Thanks @chrismore. @sgarrity cleaned up and ready for review.

media/css/firefox/firstrun/firstrun.less
```diff
@@ -0,0 +1,155 @@
...
+#intro {
+    color: #fff;
+    padding-bottom: (@baseLine * 3);
+    text-align: center;
+}
+
+header {
+    h1 {
+        margin-bottom: 0;
+        .font-size(4rem);
```
sgarrity: This seems to be `5rem` (from `.font-size(80px);`) in the variation in production.

schalkneethling: Ah yes, good catch.

schalkneethling:
@sgarrity updated

sgarrity:
This could merge as-is and be improved later, but it does appear that the page inside the iframe is responsive. This means we can likely make the entire page responsive and not need any min-widths.

https://accounts.firefox.com/?utm_campaign=fxa-embedded-form&utm_medium=referral&utm_source=firstrun&utm_content=fx-40.0&entrypoint=firstrun&service=sync&context=iframe&style=chromeless&haltAfterSignIn=true

schalkneethling:
@sgarrity updated and squashed.

media/css/firefox/firstrun/firstrun.less
```diff
@@ -0,0 +1,170 @@
...
+        .font-size(2.5rem);
+
+        @media only screen and (max-width: @breakTablet) {
+            .font-size(1.5rem);
+        }
+    }
+}
+
+main {
+    p {
```
sgarrity: Unnecessary nesting?

media/css/firefox/firstrun/firstrun.less
```diff
@@ -0,0 +1,170 @@
...
+    #outer-wrapper {
+        min-width: auto;
+    }
+
+    #masthead {
+        width: 100%;
+    }
+
+    main {
+        p {
```
sgarrity: Unnecessary nesting?

media/css/firefox/firstrun/firstrun.less
sgarrity: Unnecessary nesting?

sgarrity:
r+ :fireworks:

PR #3662, opened 2015-12-15
Fix bug 1232606: Fix newsletter titles from basket

pmclanahan:
Problem was that most titles and descriptions for newsletters are now in bedrock. The code path for ones coming from basket had an errant comma at the end, making the value a tuple instead of a string.


pmclanahan:
@jgmize r?

very quick one

PR #3665, opened 2015-12-16
Fix bug 1232477 - Add EKR to leadership page

craigcook:


kyoshino:
The bio is still 404 but otherwise LGTM!

PR #3668, opened 2015-12-17
Fix Bug 1233018, remove tiles copy from partnerships page and form

schalkneethling:
@jpetto r? This needs to be merged after https://github.com/mozilla/mcom-tests/pull/466 has been merged to avoid test breakage.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/partnerships.html
+++ b/bedrock/mozorg/templates/mozorg/partnerships.html
@@ -40,13 +40,10 @@
   <img src="{{ static('img/mozorg/partnerships/icon-firefox.png') }}" alt="{{ _('Firefox icon') }}" class="icon" />
   <h3>{{ _('Firefox for Desktop') }}</h3>
   <p>
-    {{ _('Opportunities include promoted content, distribution, customized versions and more of our PC, Mac &amp; Linux browser.') }}
+    {{ _('Opportunities include distribution, customized versions and more of our PC, Mac &amp; Linux browser.') }}
   </p>
 
   <ul>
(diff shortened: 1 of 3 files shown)


media/js/mozorg/partnerships.js
```diff
@@ -8,8 +8,6 @@
     var $sfForm = $('#sf-form');
     var $sfFormSubmit = $('#sf-form-submit');
     var $interest = $('#interest');
```
jpetto: Think we can drop this variable, too.

bedrock/mozorg/templates/mozorg/partnerships.html
```diff
@@ -40,14 +40,11 @@ <h1 class="large">{{ _('Partner with Mozilla') }}</h1>
...
   </p>
 
   <ul>
     <li>
-      <a href="{{ url('firefox.tiles') }}">{{ _('Learn more about promoted content in Mozilla Firefox') }}</a>
-    </li>
-    <li>
       <a href="{{ url('mozorg.partnerships-distribution') }}">{{ _('Learn more about Mozilla Firefox distribution policies') }}</a>
     </li>
   </ul>
```
jpetto: Can remove all form inputs with `tiles-field` class, as they will now never be displayed.

jpetto:
A note for later - if/when the form on `mozorg/contentservices/start.html` is removed, we can also remove the tiles related fields from `mozorg/forms.py/WebToLeadForm`.

jpetto:
Looks good! r+ :tropical_drink:

PR #3671, opened 2015-12-17
Bug 1150713 Remove straggler SMS page references

gauthierm:
Bug 1150713

 * Update link and content on happy feedback page
 * Remove SMS form from FX37 welcome page

Feedback page visible at `/firefox/feedback?score=4`

Welcome page visible at `/firefox/37.0/whatsnew/`

The current welcome page does geolocation to see if SMS is available. This can be triggered manually using `/firefox/37.0/whatsnew/?geo=us`. The updated page still does geolocation to populate the country flydown for the newsletter.

### DIFF
--- a/bedrock/firefox/templates/firefox/feedback/happy.html
+++ b/bedrock/firefox/templates/firefox/feedback/happy.html
@@ -4,7 +4,7 @@
 
 {% extends "firefox/base-resp.html" %}
 
-{% set_lang_files "firefox/feedback" %}
+{% set_lang_files "firefox/feedback" "firefox/sendto" %}
 
 {% block page_title_prefix %}{% endblock %}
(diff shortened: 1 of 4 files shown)


pmclanahan:
Looks okay to me. Mind taking a look @jpetto ?

jpetto:
Code looks good to me. Do we need an `L10n` tag around the string changes? @flodolo?

flodolo:
I'm a bit lost. Can someone confirm if we ever displayed ` bedrock/firefox/templates/firefox/feedback/happy.html ` to non en-US locales? I don't have the two strings that we're removing.

Anyhow, we don't need l10n tags since `Get Firefox for Android` is available from `download_button.lang` (shared), and `Send Firefox to your smartphone or tablet` is available in `firefox/sendto.lang`, which is already included here.

PR #3673, opened 2015-12-18
Fix Bug 1233620, do not display homepage takeover to Firefox for iOS users

schalkneethling:
@jpetto r?

### DIFF
--- a/media/js/mozorg/home/takeover.js
+++ b/media/js/mozorg/home/takeover.js
@@ -108,9 +108,10 @@ $(function () {
     // modal is waffled so, first ensure it exists
     if (arrayIndexOfSupported && $takeoverModal.length > 0) {
         try {
-            // only show the takeover modal if the user has not already seen it
-            // during this session.
-            if (sessionStorage.getItem('mozorg.takeover.seen') !== 'true') {
+            // only show the takeover modal if this is not Firefox for iOS and the
+            // user has not already seen it during this session.
+            var takeoverSeen = sessionStorage.getItem('mozorg.takeover.seen');
+            if (!window.isFirefoxiOS() && takeoverSeen !== 'true') {
                 initTakeOver();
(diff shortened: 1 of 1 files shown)


schalkneethling:
Just FYI, this one is urgent and should go out to production as soon as we can today. Thanks!

media/js/mozorg/home/takeover.js
```diff
@@ -108,9 +108,10 @@ $(function () {
     // modal is waffled so, first ensure it exists
     if (arrayIndexOfSupported && $takeoverModal.length > 0) {
         try {
-            // only show the takeover modal if the user has not already seen it
-            // during this session.
-            if (sessionStorage.getItem('mozorg.takeover.seen') !== 'true') {
+            // only show the takeover modal if this is not Firefox for iOS and the
+            // user has not already seen it during this session.
+            var takeoverSeen = sessionStorage.getItem('mozorg.takeover.seen');
+            if (isFirefoxiOS() === false && takeoverSeen !== 'true') {
```
alexgibson: Why not just do `!window.isFirefoxiOS()`?

alexgibson:
LGTM, one small nit but otherwise r+wc :banana:

media/js/mozorg/home/takeover.js
schalkneethling: Ah yes, I did not think to use window as it is implied but, at the same time I contradict myself and try to be explicit by checking against `false` and not `!` :p

schalkneethling:
@alexgibson updated. Thanks!

media/js/mozorg/home/takeover.js
alexgibson: Well, the other conditional is checking the value of a key, not for the existence of the key itself. This is just a simple boolean check :+1:

PR #3676, opened 2015-12-18
[fix bug 1231785] TP UI Tour does not allow linking to step 2

alexgibson:
To test this PR please first:

* [Configure UITour for local development](http://bedrock.readthedocs.org/en/latest/uitour.html#local-development).
* [Set the Tracking Protection tour URL](http://bedrock.readthedocs.org/en/latest/uitour.html#tracking-protection-uitour) to point to your local bedrock instance.


jpetto:
Code looks good and works as expected. r+! :tada:

PR #3677, opened 2015-12-18
[fix bug 1233729] Remove 3 images from manifesto page

alexgibson:
Also fixes a recent regression I introduced in #3650 (oops). Please test that the mosaic initializes on /about too.


jpetto:
LGTM. :palm_tree: r+!

PR #3638, opened 2015-12-04
Bug 1182225 release notes redesign

sgarrity:
This replaces PR https://github.com/mozilla/bedrock/pull/3581 from a shared branch as there are some more updates to push.

For some context, there was some initial work by @yglazko, then the bulk of the work was done by @malena. I have added a few (as yet unsquashed, for easier review) clean up commits at the end.

PR #3684, opened 2015-12-21
[fix bug 1234136] App Store button on release notes page is 404

alexgibson:
* Also added missing query param to a couple of buttons on other pages.

### DIFF
--- a/bedrock/firefox/templates/firefox/family/index.html
+++ b/bedrock/firefox/templates/firefox/family/index.html
@@ -134,7 +134,7 @@
           {% endif %}
           </a>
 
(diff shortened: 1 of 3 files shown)


kyoshino:
Yeah, my PR #3617 has been merged before #3638. Looks like there are 2 more no-param `firefox_ios_url()` at `firefox/family/index.html` L54 and `firefox/ios.html` L141. Do we need the param there also? Those links didn't have any param when I added the new method.

craigcook:
Not sure if we need the params on those other links but I think correcting this 404 is the more important fix right now, would be good to get it in before we break for the holidays. r+

PR #3688, opened 2015-12-21
[fix bug 1234263] Prevent New Relic JS for IE 6.

jpetto:


jpetto:
Switched from `userAgent` check to conditional comment.

jpetto:
@alexgibson r?

alexgibson:
Code-wise this gets an r+ from me. Should we merge this and test against stage, or push to a demo?

alexgibson:
Fwiw, I tried replicating the error locally on IE6 but couldn't seem to hit it.

PR #3693, opened 2015-12-22
Add functional tests for history page

alexgibson:


alexgibson:
@davehunt r? 

Note: I'm not testing clicking the next/previous buttons on this page, since the slideshow auto rotates. But it is checking that the slideshow initializes, and is not displayed on mobile viewport sizes. It's not ideal, but I think the auto-advance might make for temperamental tests when clicking. Thoughts?

tests/functional/test_history.py
```diff
@@ -0,0 +1,26 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.history import HistoryPage
+
+
+@pytest.mark.smoke
+@pytest.mark.nondestructive
+def test_is_slideshow_displayed(base_url, selenium):
```
davehunt: I'd drop the `is` from the test name.

tests/functional/test_history.py
```diff
@@ -0,0 +1,26 @@
...
+    page = HistoryPage(base_url, selenium).open()
+    assert page.is_slideshow_displayed
+    assert page.is_previous_button_displayed
+    assert page.is_next_button_displayed
+
+
+@pytest.mark.smoke
+@pytest.mark.nondestructive
+@pytest.mark.viewport('mobile')
+def test_is_list_displayed(base_url, selenium):
```
davehunt: As previous comment.

tests/functional/test_history.py
```diff
@@ -0,0 +1,26 @@
...
+    page = HistoryPage(base_url, selenium).open()
+    assert page.is_list_displayed
+    assert not page.is_previous_button_displayed
+    assert not page.is_next_button_displayed
```
davehunt: Do we need an `assert not page.is_slideshow_displayed`?

tests/functional/test_history.py
```diff
@@ -0,0 +1,26 @@
...
+from pages.history import HistoryPage
+
+
+@pytest.mark.smoke
+@pytest.mark.nondestructive
+def test_is_slideshow_displayed(base_url, selenium):
+    page = HistoryPage(base_url, selenium).open()
+    assert page.is_slideshow_displayed
+    assert page.is_previous_button_displayed
+    assert page.is_next_button_displayed
```
davehunt: Do we need an `assert not page.is_list_displayed`?

tests/pages/history.py
```diff
@@ -0,0 +1,35 @@
...
+    _url = '{base_url}/{locale}/about/history'
+
+    _slideshow_locator = (By.CSS_SELECTOR, '#slideshow-stage > .on')
+    _previous_button_locator = (By.CSS_SELECTOR, '.slide-control > .prev')
+    _next_button_locator = (By.CSS_SELECTOR, '.slide-control > .next')
+    _list_locator = (By.CSS_SELECTOR, '#slideshow > .slides')
+
+    @property
+    def is_slideshow_displayed(self):
+        self.wait.until(expected.presence_of_element_located(self._slideshow_locator))
```
davehunt: Is this a safe wait? Will this element always be present even when the slideshow is not displayed?

davehunt:
r+ with comments addressed, and I agree with not testing the next/previous buttons.

tests/pages/history.py
alexgibson: Hmm, no it's not. The `on` class is only added when the slideshow is displayed. I'll give this a rethink and refactor.

alexgibson:
@davehunt - updated. Hopefully this is a bit better.

davehunt:
There's a Travis failure due to an unused import of expected.

alexgibson:
@davehunt - fixed the Travis error. Please see my comment re: visibility of off screen transitioned content as to checking the number of slides visible (not sure we can test for this with the page in it's current state).

davehunt:
r+

PR #3695, opened 2015-12-22
[fix bug 1234555] Improve download buttons noJS fallback.

jpetto:


alexgibson:
A simple and worthwhile change. Looks good locally with JS disabled r+ :fireworks:

PR #3696, opened 2015-12-22
Fix Bug 1234595 - ESR release note select the iOS category

kyoshino:


alexgibson:
Nice work fixing the markup here too, r+ :cake: 

Care to squash the two commits, then I think we're good to merge?

kyoshino:
Squashed!

PR #3697, opened 2015-12-22
WIP [fix bug 1233779] Serve specific builds to WinXP users.

jpetto:
Still working through some CSS/display issues, but hoping for a review/sanity check on direction.

@alexgibson, @jgmize, @pmclanahan - Care to take a look?

### DIFF
--- a/media/js/base/site.js
+++ b/media/js/base/site.js
@@ -36,6 +36,7 @@
             }
             if (pf.indexOf("Win32") !== -1 ||
                     pf.indexOf("Win64") !== -1) {
+
                 return 'windows';
             }
             if (/android/i.test(ua)) {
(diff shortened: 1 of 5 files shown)


media/js/base/site.js
```diff
@@ -152,6 +153,11 @@
             if (version && parseFloat(version) >= 6.1) {
                 h.className += ' win7up';
             }
+
+            // Check for Windows XP users to alter download button behavior
+            if (/Windows NT 5\.1/i.test(navigator.userAgent)) {
```
pmclanahan: The [logic from the bouncer efforts](https://github.com/mozilla-services/go-bouncer/pull/16/files#diff-31a4a7824225629df071a1cba8b785d5R26) also looks for "XP". Not sure if this is a real thing, but might be worth checking for as well.

pmclanahan:
The direction seems good. Not sure we'll need it yet or not. It's looking like bouncer will roll out their changes soon, but it's great that we're ready to go if that effort fails.

jpetto:
Looks like the bouncer fix did the trick. Closing this PR, but will keep the branch around just in case.

pmclanahan:
Sounds good. Thanks for the great work @jpetto, we may yet need this or something like it.

PR #3699, opened 2015-12-23
Add details of emeritus status for owners and peers

gerv:
Hi,

At Mitchell's request (see mozilla.governance) we've just created a new status of "emeritus owner" or "emeritus peer" for ex module owners. This change documents that.

Gerv


alexgibson:
r+ thanks @gerv

PR #3700, opened 2015-12-23
Add functional tests for plugincheck page

alexgibson:


alexgibson:
@davehunt r?

Not necessarily something I expect to solve right now, but one open question I do have for this page with regard to testing is how we could potentially handle loading plugins via Saucelabs. It would be great to have two more base tests for this page - one test where Firefox loads with no plugins, and another test where it loads with at least one plugin. I'm going to add this to our issues list (unless there's a simple way we can do this).

davehunt:
r+

> Not necessarily something I expect to solve right now, but one open question I do have for this page with regard to testing is how we could potentially handle loading plugins via Saucelabs. It would be great to have two more base tests for this page - one test where Firefox loads with no plugins, and another test where it loads with at least one plugin. I'm going to add this to our issues list (unless there's a simple way we can do this).

I suspect the Firefox instances provided by Sauce Labs will have at least one plugin present. If it's possible to control plugins via the Firefox profile then we should be able to do what we need.

PR #3701, opened 2015-12-23
Add functional test for Windows 10 welcome page

alexgibson:


alexgibson:
@davehunt r?

Note: this page is only actively shown to Firefox users on Windows 10 (even though it's a publicly accessible URL). We already have good JS behavioral/unit test coverage for this page, which heavily uses UITour API. This PR just adds a basic functional test for non-Firefox browsers, to make sure we don't show anything unexpected should someone discover the page.

tests/pages/firefox/win10_welcome.py
```diff
@@ -0,0 +1,24 @@
...
+
+    _url = '{base_url}/{locale}/firefox/windows-10/welcome'
+
+    _main_contant_locator = (By.TAG_NAME, 'main')
+    _links_section_locator = (By.CLASS_NAME, 'firefox-learn-links')
+
+    @property
+    def is_firefox_default_messaging_displayed(self):
+        class_name = self.find_element(self._main_contant_locator).get_attribute('class')
+        return 'firefox-default' in class_name
```
davehunt: Nit: I would create a variable for the element rather than the element attribute, such as:

```python
el = self.find_element(self._main_contant_locator)
return 'firefox-default' in el.get_attribute('class')
```

tests/functional/firefox/test_win10_welcome.py
```diff
@@ -0,0 +1,15 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.firefox.win10_welcome import Win10WelcomePage
+
+
+@pytest.mark.skip_if_firefox
```
davehunt: It's not clear to me why we're skipping if Firefox.. perhaps an additional comment would help? How would the test fail if we did run it in Firefox?

davehunt:
r+ with comments

tests/functional/firefox/test_win10_welcome.py
alexgibson: The test would fail in Firefox depending on whether or not it was the default browser on the machine. It would also rely on UITour configs being set correctly. Will add comment as suggested!

PR #3702, opened 2015-12-23
Revert "[fix bug 1234263] Prevent New Relic JS for IE 6."

jgmize:
Reverts mozilla/bedrock#3688


jgmize:
Based on further testing, I believe we have confirmed the following:

* The NR js does *not* break IE6
* The NR js loads correctly on IE6 on XP SP3
* The NR js does not load on IE6 on XP SP2 due to a sha256 cert on https://js-agent.newrelic.com/
* This will prevent NR browser data from being collected from XP SP2 clients, but does not impair site functionality in any other way.

jgmize:
@jpetto or @alexgibson r?

jpetto:
Revert looks good, and agree we should be safe serving NR JS to XP SP2.

r+

PR #3704, opened 2015-12-24
Add functional tests for About page

alexgibson:


alexgibson:
@davehunt r?

tests/pages/about.py
```diff
@@ -2,9 +2,32 @@
...
+        return self.is_element_displayed(self._video_overlay_locator)
+
+    @property
+    def is_video_displayed(self):
+        return self.is_element_displayed(self._video_locator)
+
+    def play_video(self):
+        assert self.is_video_overlay_displayed, 'Video is already displayed'
+        self.find_element(self._video_overlay_locator).click()
+        self.wait.until(lambda m: not self.is_video_overlay_displayed)
```
davehunt: Nit: The argument passed to the condition is a Selenium object, iirc we conventionally name this `s` rather than `m`. We may already be inconsistent here though, and it has no effect on the code.

davehunt:
r+

PR #3705, opened 2015-12-24
Add functional test for Firefox Tiles page

alexgibson:
Also removes a redundant Casper test for the firefox/tiles page which I didn't think added much value.


alexgibson:
@davehunt r?

alexgibson:
Actually, I've decided to add a basic test here for the `/firefox/tiles/` page to test the family navigation menu. this is better than nothing :smile:

davehunt:
r+

PR #3706, opened 2015-12-25
Update GermanHolidays.ics

MozFux:
Bug 1224067 - Updated German Holidays until 2025


MozFux:
Do you know, why there is still the old file at https://www.mozilla.org/en-US/projects/calendar/holidays/#G ?

craigcook:
These static files are served from a CDN and cached for 30 days, but we've purged the cache manually so the updated file should be serving now. Some browsers/clients may have their own cache, of course, so there's not much we can do about that.

We may shorten the expiration time for calendar files to just a few days so updates become available quicker in the future.

MozFux:
Thanks, it's working now :)

kewisch:
I think it should be fine to stay at 30 days. The files are only updated max once a year (another PR is pending though), so if I can ask for a manual purge once a year that might be smarter than turning the caching time down to a few days. WDYT?

PR #3708, opened 2015-12-27
Bug 1235036 - Broken link on SmartOn Surveillance page

tieu:
Added en-us to url in order to fix the broken link.


craigcook:
Thanks!

Issue #3629, opened 2015-12-03
Add functional tests to improve coverage


davehunt:
The following functional tests/checks have been identified as desirable to improve test coverage:

- [ ] Viewing [Mozilla Spaces](https://www.mozilla.org/en-US/contact/spaces/) on a mobile viewport
- [ ] Viewing [Mozilla Communities](https://www.mozilla.org/en-US/contact/communities) on a mobile viewport
- [ ] Checking page footer is displayed on at least one page
- [x] Checking 404 page components are displayed
- [x] Thunderbird download pages
- [x] Firefox Desktop pages
- [x] Firefox for iOS page
- [x] Firefox channel page
- [x] Firefox installer help page
- [x] SmartOn pages
- [x] Android page sent-to-device form test (blocked by #3534)
- [ ] Firefox Products page conditional messaging for Android, iOS, out-of-date Firefox.
- [ ] Plugincheck page tests for both none and one (or more) plugins using Firefox.
- [ ] Sync page conditional messaging for Android, iOS, out-of-date Firefox.
- [ ] Firefox family navigation scroll tests, plus mobile viewport test.
- [x] Firefox OS 2.5 page news feed is displayed test (pending #3709)
- [x] Add test for partnerships form (see: https://github.com/mozilla/mcom-tests/blob/master/tests/test_partnerships.py#L33)

alexgibson:
I'm going to close this issue for now. There are still a few areas not covered in this initial list, but I think we have good enough coverage for most things now. Can always revisit in the future when we have some more bandwidth.

PR #3709, opened 2015-12-28
[fix bug 1233811] Add In The News to /firefox/os/2.5/.

jpetto:


jpetto:
Also adds `update_firefox_os_feeds` to `bin/sync_all`. @jgmize or @pmclanahan - Could you review that small bit? Change is simple enough, but want to make sure there are no unforeseen consequences.

pmclanahan:
change to `sync_all` seems fine to me.

media/css/firefox/os/firefox-os-2-5.less
```diff
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 @import "../../sandstone/lib.less";
+@import "in-the-news.less";
```
alexgibson: Any specific reason to use an `@import` here instead of including the file in the page bundle?

media/css/firefox/os/firefox-os.less
```diff
@@ -1,4 +1,5 @@
 @import "../../sandstone/lib.less";
+@import "in-the-news.less";
```
alexgibson: Same here?

alexgibson:
Change looks good to me :+1:

One question on the use of `@import` for the news feed styles, but aside from that it's an r+wc (change being optional). I wasn't sure if we we're now generally trying to avoid the use of `@import` and favoring bundles instead (with the exception of `lib.less`).

I've also made a [note here](https://github.com/mozilla/bedrock/issues/3629) to add a functional test for this change once it hits prod.

jpetto:
Good call on favoring bundles. Updated!

PR #3710, opened 2015-12-28
Fix link to home page on /credits: fix bug 1235426

jgmize:
Hard code link to eliminate caching localized link


jgmize:
@pmclanahan r?

PR #3711, opened 2015-12-29
Fix bug 1235480: Fix deletion of removed MFSA advisories

pmclanahan:
Also clean up any Products left without advisories after said deletions.


pmclanahan:
@jgmize r?

PR #3719, opened 2015-12-31
[fix bug 1235031] Update Telcel link (Mexico) for Firefox OS devices page

alexgibson:


tieu:
The new link only links to one product. 

Here's a link to list all devices running Firefox OS: http://www.telcel.com/bin/telcelcom/device/filter.9.telefonos-y-smartphones.22.order-popularidad.html?so=Firefox+OS

Should we use this one? Just in case the carrier obtains more Firefox OS devices.

alexgibson:
In this case we know there are not going to be any more Firefox OS consumer smartphones, so I think linking directly to the handset already on sale is the most appropriate choice. Thanks for the suggestion nonetheless!

PR #3720, opened 2015-12-31
[fix bug 1235539] 'Other releases' menu link on release notes page is not keyboard accessible

alexgibson:


media/js/firefox/releasenotes.js
```diff
@@ -11,16 +11,44 @@
...
+        $subMenuTitle.on('focus', function() {
+            $subMenu.attr('aria-expanded', true);
+        });
+
+        $subMenuTitle.on('click', function(e) {
+            e.preventDefault();
+        });
+
+        $subMenuLinks.on('focusout', function(e) {
+            if ($(e.target).parent().is('li:last-child')) {
```
jpetto: Small issue here - if you tab backwards/shift-tab from the last `<li>`, the menu disappears. Maybe needs a `$subMenuLinks.on('focus')` handler?

alexgibson: Hmm, good catch! A `focus` event handler doesn't seem to work here, so I think the solution will require a `keydown` event in order to pick up the shift key.

alexgibson:
Thanks @jpetto - updated.

jpetto:
Sweet, sweet keyboard navigation. :candy:

r+!

PR #3722, opened 2015-12-31
Bug 1226476 - Convert android/all/ to the same format as desktop uses

tieu:
Replaced old links with bouncer links.


jgmize:
The travis failures are due to 503s from pypi.python.org for test dependencies. These changes look good to me, and fix an important bug. Thanks @tieu!

PR #3724, opened 2016-01-01
Bug #1236432 - Miswording in Windows 10 welcome page

tieu:
As reported by @darchons . The button is located above the text "Click the button below"

Here's the fix for the typo.

### DIFF
--- a/bedrock/firefox/templates/firefox/win10-welcome.html
+++ b/bedrock/firefox/templates/firefox/win10-welcome.html
@@ -51,7 +51,7 @@
     <button id="set-default" type="button" class="button-flat-dark">{{ _('Let’s do it') }}</button>
     <h2>{{ _('Make Firefox your default in 3 easy steps:') }}</h2>
     <ol>
-      <li>{{ _('Click the button below') }}</li>
+      <li>{{ _('Click the button above') }}</li>
       <li>{{ _('Scroll down to “Web browser”') }}</li>
       <li>{{ _('Choose Firefox') }}</li>
     </ol>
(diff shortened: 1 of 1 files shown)


alexgibson:
@tieu - unfortunately we can't make a string change as simple as this, as it will break many existing translations. We can however use an l10n tag to provide some additional logic to the template, so the correct messaging/button position gets shown to locales which have the active translation. For example:

    {% if l10n_has_tag('win10_button_update') %}
      {{ _('Click the button above') }}
    {% else %}
      {{ _('Click the button below') }}
    {% endif %}

The same kind of logic could also be used to conditionally show the button either above or below the copy.

I also [filed a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1236432) for this issue, can you please make sure to include the bug number for reference in the commit message.

http://bedrock.readthedocs.org/en/latest/contribute.html#git-commit-messages

Thanks!

Pinging @flodolo for l10n awareness

alexgibson:
@tieu - please let me know if you have time to work on this fix some more. Not to worry if you don't, I can pick it up. Thanks

tieu:
I'll try and have it done by tonight.

Thanks.

tieu:
I've also opened a new PR for the win10-welcome.lang file: 

https://github.com/mozilla-l10n/bedrock-l10n/pull/3

Let me know if there should be any changes.

Thanks.

flodolo:
@tieu Thanks for the PR.

it wasn't actually needed but it's nice to see you almost nailed all passages

alexgibson:
Thanks for the update @tieu - I think we should display the button in the correct location, depending on the l10n tag, as well as change the appropriate string. We could do this using similar logic in the template e.g.

For the button above the copy:

      {% if l10n_has_tag('win10_button_update') %}
        <button id="set-default" type="button" class="button-flat-dark">{{ _('Let’s do it') }}</button>
      {% endif %}

For the button below the copy:

      {% if not l10n_has_tag('win10_button_update') %}
        <button id="set-default" type="button" class="button-flat-dark">{{ _('Let’s do it') }}</button>
      {% endif %}

alexgibson:
Closing in favor of #3759. I preserved your original commit @tieu, thanks for opening the PR!

PR #3726, opened 2016-01-04
[bug 1214038] Remove end of year fundraising takeover

alexgibson:


jpetto:
All appears to be in order. r+!

PR #3727, opened 2016-01-04
[fix bug 1236450] Remove end of year fundraising home page promo

alexgibson:


jpetto:
All looks good to me. r+!

PR #3728, opened 2016-01-04
Fix Bug 1226476 - Convert android/all/ to the same format as desktop uses

kyoshino:
Screenshot: https://bugzilla.mozilla.org/attachment.cgi?id=8703553


kyoshino:
@jgmize @pmclanahan r?

kyoshino:
These column labels are from [nightly.mozilla.org](https://nightly.mozilla.org/). Honeycomb (Android 3.x) is no longer supported so **Modern devices (Android 4.0+)** is correct.

kyoshino:
Oh wait, the support version is changing with Firefox 46. So we still have to say **Android 3.0+**. Will update the label and icon.

craigcook:
Page looks good to me. r+ on the front-end. :zap:

kyoshino:
Thanks @craigcook!

kyoshino:
@flodolo There are a couple of new strings here as well.

flodolo:
We don't localize the download pages, so good for us ;-)

kyoshino:
https://www.mozilla.org/en-US/firefox/all/ is localized, but only in Japanese.

flodolo:
Because they decided to localize it on their own. It's not tracked, not exposed to any other locale and not updated (there's also no en-US version).

jgmize:
@kyoshino would you mind rebasing this?

kyoshino:
Rebased!

jgmize:
Thanks again, @kyoshino!

kyoshino:
:smile_cat:

PR #3732, opened 2016-01-05
Fix Bug 1236791 - 301 redirect legacy query parameters on /firefox/new/

kyoshino:


pmclanahan:
This looks good to me. We should test to make sure CloudFlare is taking query params into account for caching responses, but I'm pretty sure it is.

PR #3738, opened 2016-01-05
Fix bug 1236910: Add redirects for SUMO

pmclanahan:


bedrock/mozorg/redirects.py
```diff
@@ -572,6 +572,9 @@ def decider(request, **kwargs):
     # bug 1003737
     redirect(r'^impressum/?$', '/about/legal/impressum/'),
 
+    # bug 1236910
+    redirect(r'^support(/.*)?$', 'https://support.mozilla.org/'),
```
alexgibson: The URL in question seems to be `https://bedrock-prod-deis.eu-west.moz.works/support/firefox/options`, do we need to be excluding the locale prefix here, or will this still catch and redirect?

alexgibson:
One question on requiring the locale prefix for the sumo redirect, but otherwise looks good to me r+wc (change being only required if needed).

bedrock/mozorg/redirects.py
pmclanahan: This should catch it with or without a locale prefix. I thought that would be best.

PR #3741, opened 2016-01-05
Allow security advisory to use single value for fixed_in

pmclanahan:


pmclanahan:
@cr r?

bedrock/security/tests/test_views.py
```diff
@@ -125,9 +125,9 @@ def test_latest_queryset_product(self):
 
     def test_latest_queryset_product_version(self):
         """Should advisories for a single product version."""
-        advisories_30 = [self.new_advisory(fixed_in=['Firefox 30.{0}'.format(i)])
+        advisories_30 = [self.new_advisory(fixed_in='Firefox 30.{0}'.format(i))
                          for i in range(5)]
-        advisories_29 = [self.new_advisory(fixed_in=['Firefox 29.0.{0}'.format(i)])
+        advisories_29 = [self.new_advisory(fixed_in='Firefox 29.0.{0}'.format(i))
```
cr: If I understand correctly, those changes to lines 117, 128 and 130 switch the tests from multiline format to oneline. Shouldn't both be tested?

bedrock/security/management/commands/update_security_advisories.py
```diff
@@ -129,7 +129,12 @@ def add_or_update_advisory(data, html):
         dateobj = parsedate(datestr).date()
         kwargs['announced'] = dateobj
     prodver_objs = []
-    for productname in data.pop('fixed_in'):
+
+    fixed_in = data.pop('fixed_in')
+    if isinstance(fixed_in, basestring):
+        fixed_in = [fixed_in]
+
+    for productname in fixed_in:
```
cr: This block looks good to me, now supporting both multiline and oneline. r+

bedrock/security/tests/test_views.py
pmclanahan: Yes. I was lazy and didn't add a specific test, but I changed these lines and left others in this test class as a list. Those are the two primary types used and I wanted to get this in before another advisory landed.

cr: Ah, at least there's coverage. r+ then from my side.

pmclanahan:
@jgmize I think we can merge this then. Look okay to you?

PR #3743, opened 2016-01-06
Remove smoke marker on Android l10n test

alexgibson:


alexgibson:
@davehunt - I missed this in the last PR. This one relies on locales being checked out, so should not have been a smoke test. Going to remove this marker.

PR #3744, opened 2016-01-06
Bug 1233015: Remove and redirect contentservices pages

pmclanahan:


alexgibson:
Looks like we can also remove the images in `media/img/contentservices`?

alexgibson:
Couple of tests that need removing too, but otherwise r+wc from me :beer:

pmclanahan:
> Looks like we can also remove the images in media/img/contentservices?

good catch. done.

pmclanahan:
Tests are green. Thanks Alex!

PR #3746, opened 2016-01-07
[fix bug 1235853] Remove facebookapps pages

alexgibson:


bedrock/facebookapps/templates/facebookapps/downloadtab.html
```diff
@@ -1,131 +0,0 @@
...
-                <button class="ir next">{{ _('Next') }}</button>
-            </div><!-- /#slider-container -->
-        </div><!-- /#scene3 -->
-    </div><!-- /#stage -->
-</div><!-- /#theater-firefox -->
-{% endblock %}
-
-{% block js %}
-    <div id="init-data"
-        data-app-id="{{ settings.FACEBOOK_APP_ID }}"
```
alexgibson: Note to self: we can probably remove this in settings. Same for line below

alexgibson: Removed from `settings`

bedrock/settings/base.py
```diff
@@ -569,15 +564,6 @@ def lazy_email_backend():
                          'tl_PH', 'tr_TR', 'uk_UA', 'vi_VN', 'zh_CN',
                          'zh_HK', 'zh_TW']
 
-
-# FACEBOOK_TAB_URL is lazily evaluated because it depends on the namespace
-# and app ID settings in local settings.
```
alexgibson: @pmclanahan - is there anything else we need to tidy up here? (do different envs havespecific app id settings?)

pmclanahan: I think this is sufficient. I'm sure there are `FACEBOOK_*` settings in the SCL3 environments, but I don't see anything in the AWS ones, so this is probably only working some of the time anyway :( I'll go in and delete these settings from the SCL3 envs after this is deployed.

alexgibson: Ok thanks @pmclanahan :+1:

schalkneethling:
I guess we can remove this as well? https://github.com/mozilla/bedrock/blob/master/etc/httpd/global.conf#L127

schalkneethling:
Not sure if it is because there are still some `.pyc` files inside the folder but, from what I can see, this folder while still exist after this is merged? https://github.com/mozilla/bedrock/tree/master/bedrock/facebookapps

schalkneethling:
Two open questions, other than that r+ :skull:

alexgibson:
> I guess we can remove this as well? https://github.com/mozilla/bedrock/blob/master/etc/httpd/global.conf#L127

I believe we still require the pass though for the redirect to work.

> Not sure if it is because there are still some .pyc files inside the folder but, from what I can see, this folder while still exist after this is merged? https://github.com/mozilla/bedrock/tree/master/bedrock/facebookapps

Not sure if I can do anything to change that? People can tidy up the folder on their local checkout once this merges.

schalkneethling:
Sounds good to me

PR #3748, opened 2016-01-07
Change caldata and font cache headers to one week

pmclanahan:
The previous timeout of one month was a bit much and caused a problem with a holiday calendar update.


pmclanahan:
@jgmize r?

@craigcook one week sound okay?

craigcook:
A week seems reasonable. These files change rarely but when they do change a month is a long time to wait.

PR #3749, opened 2016-01-08
Add download link checks

davehunt:
I'm concerned that if the LinkChecker download jobs failed to find the download links that they could give false confidence that these are not broken. We saw an issue recently where a conditional CSS block caused the LinkChecker to abort early, and the result was success. These tests are an improved version of those in mcom-tests, in that they now check the response code for all platforms and all locales (over 2,300 URLs).

I have not included the 'smoke' marker so these will not be run by the pipeline at this time. I have added a new 'headless' marker, which will allow us to isolate these tests, or to exclude them. It does not make sense, for example, to run these for multiple platforms as we would with the other functional tests.

I suggest replacing the LinkChecker download jobs in the pipeline with these new tests. Having these in the repository also makes it much easy to update the URLs of the download pages that we're testing. We can also add more asserts for these links if any come to mind.

I'm keen to hear your feedback, @alexgibson

### DIFF
--- a/tests/functional/conftest.py
+++ b/tests/functional/conftest.py
@@ -15,16 +15,23 @@ def capabilities(capabilities):
     return capabilities
 
 
+@pytest.fixture
+def firefox(selenium):
+    return selenium.capabilities.get('browserName').lower() == 'firefox'
+
+
+@pytest.fixture
+def internet_explorer(selenium):
+    return selenium.capabilities.get('browserName').lower() == 'internet explorer'
+
+
 @pytest.fixture(autouse=True)
-def filter_capabilities(request, selenium):
-    capabilities = selenium.capabilities
-    firefox = capabilities.get('browserName').lower() == 'firefox'
-    internet_explorer = capabilities.get('browserName').lower() == 'internet explorer'
-    if firefox and 'skip_if_firefox' in request.keywords:
+def filter_capabilities(request):
+    if request.node.get_marker('skip_if_firefox') and request.getfuncargvalue('firefox'):
         pytest.skip('Test must not be run on Firefox')
-    if not firefox and 'skip_if_not_firefox' in request.keywords:
+    if request.node.get_marker('skip_if_not_firefox') and not request.getfuncargvalue('firefox'):
         pytest.skip('Test must only be run on Firefox')
-    if internet_explorer and 'skip_if_internet_explorer' in request.keywords:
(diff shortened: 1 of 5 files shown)


tests/functional/conftest.py
```diff
@@ -15,16 +15,23 @@ def capabilities(capabilities):
...
 
 
+@pytest.fixture
+def firefox(selenium):
+    return selenium.capabilities.get('browserName').lower() == 'firefox'
+
+
+@pytest.fixture
+def internet_explorer(selenium):
+    return capabilities.get('browserName').lower() == 'internet explorer'
```
alexgibson: This should be `selenium.capabilities`

alexgibson:
This looks good to me @davehunt 

Given that there is no benefit to running these tests on multiple platforms, I agree isolating these tests in a separate job makes the most sense. They also take a while to run, so could slow down other jobs considerably. 

If we had these tests set up in their own Jenkins job, I would be happy to merge this even though the tests are currently failing due to [bug 1236045](https://bugzilla.mozilla.org/show_bug.cgi?id=1236045) and [bug 1234208](https://bugzilla.mozilla.org/show_bug.cgi?id=1234208). Before we do that however, I guess we should make sure to exclude these tests using the `headless` marker on other jobs we have set up in Jenkins? (outside of the pipeline). We could also retire the LinkChecker download jobs in favor of these tests (retaining only the full LinkChecker job).

Thoughts? Happy to discuss this further in our catch up later today.

alexgibson:
My other thought is perhaps these tests should live in a sibling directory, instead of under `/functional`? Separating them out might make it easier for running just the functional tests on their own, as currently the download tests would need to be manually excluded each time?

tests/functional/conftest.py
davehunt: Oops! Fixed :smile:

davehunt:
After speaking with @alexgibson we decided to keep these in the functional directory and add a note to the documentation on how to exclude these tests due to the long running time. I'll prepare Web QA's Jenkins to exclude headless tests from the existing jobs, and create a new job to run the headless suite.

alexgibson:
r+ from me, care to squash?

I'll hold off merging until we have Jenkins updated.

davehunt:
@alexgibson Rebased and squashed. Jenkins should be ready now.

PR #3752, opened 2016-01-11
Bug 1237282 - Broken links on governance pages due to relative protocol

alexlovescoding:


bedrock/mozorg/templates/mozorg/about/governance/organizations.html
```diff
@@ -20,7 +20,7 @@ <h1 class="title-shadow-box">{{_('Mozilla Organizations')}} </h1>
 
 <p>{{_('The <a href="//www.mozilla.org/foundation/">Mozilla Foundation</a> is a California non-profit corporation exempt from Federal income taxation under IRC 501(c)(3). The Foundation supports the existing Mozilla community and oversees Mozilla’s governance structure. It also actively seeks out new ways for people around the world to recognize and steward the Internet as a critical public resource.')}}</p>
```
alexgibson: While we're here, it would be good to fix the other protocol relative links. We can make all the mozilla.org links `https` explicitly. Even though they are not essentially "broken", using protocol relative links is now [considered an anti-pattern](http://www.paulirish.com/2010/the-protocol-relative-url/).

alexgibson: Actually, it would be even better to use the `url` helper here for the mozorg URLs. So for example here:

    <p>
        {% trans foundation_url=url('foundation.index') %}
	    The <a href="{{ foundation_url }}">Mozilla Foundation</a> is a California non-profit corporation exempt from Federal income taxation under IRC 501(c)(3). The Foundation supports the existing Mozilla community and oversees Mozilla’s governance structure. It also actively seeks out new ways for people around the world to recognize and steward the Internet as a critical public resource.
        {% endtrans %}
    </p>

This will then use the appropriate URL automatically and take protocol without needing to think about it.

alexgibson:
Many thanks for another PR @alexlovescoding! 

Changes look good. I've made a suggestion for a little more tidy up we could do here. Please let me know if you have any questions, or if something doesn't make sense. Thanks!

alexlovescoding:
Is there a mozorg.contact url?  I tried that, but it doesn't work.  Or should I enter the url manually for that one?

alexgibson:
> Is there a mozorg.contact url? I tried that, but it doesn't work. Or should I enter the url manually for that one?

Looks like you need `url('mozorg.contact.contact-landing')`

alexgibson:
Friendly ping @alexlovescoding - do you have bandwidth to finish this bug? Please let me know either way. I would be happy to finish it up if you're busy on other things (I can make sure to preserve your original commit for credit). Thanks!

alexlovescoding:
@alexgibson, there you go!  Sorry about the delay.

bedrock/mozorg/templates/mozorg/about/governance/organizations.html
```diff
@@ -14,18 +14,20 @@ <h1 class="title-shadow-box">{{_('Mozilla Organizations')}} </h1>
 
 <p>{{_('There are several organizations that support the Mozilla community and ')}}<a href="{{ url('mozorg.about.manifesto') }}">{{_('Mozilla’s principles.</a> They include the non-profit Mozilla Foundation as well as the wholly owned taxable subsidiary, the Mozilla Corporation. Mozilla considers itself a <a href="http://commonspace.wordpress.com/category/hybrid/">hybrid organization</a>, combining non-profit and market strategies to ensure the Internet remains a shared public resource.')}}</p>
 
-<p>{{_('This page provides information about the various organizations and their relationships to each other. If you have any questions about this, please feel free to <a href="//www.mozilla.org/contact">contact us</a>')}}</p>
+<p>{{_('This page provides information about the various organizations and their relationships to each other. If you have any questions about this, please feel free to ')}}<a href="{{ url('mozorg.contact.contact-landing') }}">{{_('contact us</a>.')}}</p>
```
alexgibson: I don't think this page is actively localized, but we should avoid splitting a single sentence into multiple strings like this. I would follow what you have done below for this paragraph as well.

alexgibson:
There are still a few more protocol relative links in `about/governance/roles.html` that need fixing.

Other than that, and the one comment above to address this is an r+wc :+1:

alexgibson:
@alexlovescoding - many thanks for your PR! I'm going to close this PR in favor of #3801. I made sure to preserve your original commit so you still receive credit when it merges. Thanks for contributing!

PR #3754, opened 2016-01-11
[fix bug 1237960] Remove old /whatsnew and /firstrun pages

alexgibson:
This PR removes some of our old `/firstrun` and `/whatsnew` pages, and renames/simplifies some of our templates now that these dependencies are gone. The diff looks a bit worse than this PR actually is, as there is no new code written - only deleting/moving/renaming templates, with some simplified inheritance. Here's a breakdown of what to expect in this change:

* All `/firstrun` and `/whatsnew` templates prior to FF 38.0.5 have been removed. The current ESR version is 38, so traffic to these pages should now be pretty low. These URLs now fall back to the generic 'Sync' promo pages (with no tour).
* The only place where we now have to maintain the full tour, is at `/firefox/tour/`. We're only keeping the most recent version of the tour with this change, which was is shown to users on 36.0 and above. All of the `/tour/` URLs below 36.0 will just be shown the same 'Sync' promo page as the legacy `/firstrun/` and `/whatsnew` URLs.
* Also removes the template `whatsnew-nightly-29.html`, which appears to have been orphaned for some time.


flodolo:
(marking to make sure I check out if I can remove some files)

alexgibson:
@jpetto r?

jpetto:
It looks like `media/img/firefox/australis/hello-2.png` and `media/img/firefox/australis/hello-2-high-res.png` are still referenced in the [`hello-animation.html` include](https://github.com/alexgibson/bedrock/blob/bug-1237960-remove-old-whatsnew-firstrun-pages/bedrock/firefox/templates/firefox/includes/hello-animation.html#L2).

bedrock/firefox/views.py
```diff
@@ -513,7 +499,7 @@ def get_template_names(self):
 
         if show_devbrowser_firstrun_or_whatsnew(version):
             template = 'firefox/dev-whatsnew.html'
-        elif version.startswith('42.') or version.startswith('43.'):
+        elif show_42_whatsnew(version):
             template = 'firefox/whatsnew_42/variant-a.html'
```
jpetto: While we're at it, should this template (and associated bundles/files) be renamed? It's not really a variant anymore.

jpetto:
Tested existing/remaining templates and all loaded without error. Spot checked many product-specific URLs and all loaded expected templates.

Aside from one image needing resurrection and a small renaming suggestion, all is looking good. r+wc

Thanks for the tough cleanup!

bedrock/firefox/views.py
alexgibson: Makes sense :+1:

alexgibson:
> It looks like media/img/firefox/australis/hello-2.png and media/img/firefox/australis/hello-2-high-res.png are still referenced in the hello-animation.html include.

Nice spot - I'll move this asset to a new, more permanent location

alexgibson:
On closer inspection it looks like the two Firefox Hello images used as fallback in the animation aren't actually being used any longer. Older browsers seem to be showing the CSS background image as fallback instead, so perhaps this is some legacy markup that can be removed? (it would be good to sanity check this, as I'm not too familiar with the animation).

Updated the /whatsnew 42 template name as suggested also :+1:

alexgibson:
Ok, I figured out where the Hello images we're used. I didn't realize the `hello-animation.html` include was used on the tour page, but not on the Hello product page. I've reinstated the images so they are no longer being removed :+1:

bedrock/firefox/templates/firefox/australis/fx36/tour.html
```diff
@@ -2,26 +2,37 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
 
-{% extends "firefox/australis/fx36/tour-base.html" %}
+{% extends "firefox/base.html" %}
 
 {% set_lang_files "firefox/australis/fx36_tour" %}
```
flodolo: Should this be changed to `firefox/australis/firefox_tour` at this point? I confess that I'm seriously lost at this point, and this looks like a scary change: based on the first file, I'd need to coordinate with the push to production, get rid of fx36_tour.lang and make sure all string are available in firefox_tour.lang

flodolo:
Since the diff is useless, I tried applied the patch and checking files. Besides an obsolete tag (`new_doorhanger`) that I've removed from lang files, I couldn't see anything wrong.

alexgibson:
> Since the diff is useless, I tried applied the patch and checking files. Besides an obsolete tag (new_doorhanger) that I've removed from lang files, I couldn't see anything wrong.

Thanks @flodolo for taking a look (sorry the diff looks worse than it probably is)

jpetto:
Looks good!

PR #3755, opened 2016-01-11
[fix bug 1238248] Decommission /firefox/push page

alexgibson:


alexgibson:
Going to add a do-not-merge tag to this just for now, will confirm when this is good to be removed.

alexgibson:
I think we are safe to review/merge this, as the new SUMO article is actively localized now, and the in-product link is already bypassing mozorg anyway (see bug for discussion).

jpetto:
Code all looks good, but can/should we remove the [entry from `firefox/urls.py`](https://github.com/alexgibson/bedrock/blob/bug-1238248-remove-firefox-push-page/bedrock/firefox/urls.py#L108)?

alexgibson:
> Code all looks good, but can/should we remove the entry from firefox/urls.py?

Good catch, removed the URL thanks.

jpetto:
:briefcase: r+!

PR #3757, opened 2016-01-11
Add a link to the bedrock instance in Jenkins when running the functional tests

davehunt:
@alexgibson this should fix the pipeline


alexgibson:
LGTM, also @jgmize gave a thumbs up on IRC. r+

PR #3758, opened 2016-01-11
Fix Bug 1237928 - [Firefox OS] Panasonic TV static page is outdated

kyoshino:


kyoshino:
Cc @flodolo for the updated strings

bedrock/firefox/templates/firefox/partners/index.html
```diff
@@ -254,7 +254,11 @@ <h3 id="os-operators-headline" class="tween">
               <div>
                 {% if l10n_has_tag('mwc_2015') %}
                   <h4>{{ _('Firefox OS at home') }}</h4>
-                  <a href="{{ url('firefox.os.tv') }}">{{ _('The power of the Web is coming to Panasonic TVs later this year') }}</a>
+                  {% if l10n_has_tag('fxostv_update_201601') %}
+                    <a href="{{ url('firefox.os.tv') }}">{{ _('The power of the Web is coming to Panasonic TVs again in 2016') }}</a>
```
flodolo: Is the bug asking to use 'Firefox OS will power new Panasonic UHD TVs unveiled at CES' here? It's quite unclear to me.

bedrock/firefox/templates/firefox/partners/index.html
```diff
@@ -254,7 +254,11 @@ <h3 id="os-operators-headline" class="tween">
               <div>
                 {% if l10n_has_tag('mwc_2015') %}
```
flodolo: Since you're touching this file, we should also take the occasion to get rid of this tag, all active locales have the strings translated.

bedrock/firefox/templates/firefox/partners/index.html
kyoshino: @flodolo You are right. Fixed the copy.

bedrock/firefox/templates/firefox/partners/index.html
kyoshino: Okay, will remove the tag. Actually there are still a couple of MWC 2015 traces on this page  :stuck_out_tongue:

kyoshino:
I have removed the MWC menu items as well as the MWC logo in the Recent News section.

alexgibson:
Updates to the /partners page will require a couple of tests removing I think…

kyoshino:
Hmm, I should probably remove all the WMC stuff with a separate PR.

kyoshino:
Okay, I just removed the closest `mwc_2015` tag around the new `fxostv_update_201601` tag. We have to fix the copy ASAP.

alexgibson:
Code changes look good to me, r+

I guess we should wait for [confirmation in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1237928#c11) from John Bernard that all looks ok before merging?

kyoshino:
John says we don't need the link to the blog now as MWC is approaching. I just removed it.

PR #3761, opened 2016-01-12
Fix bug 1238686 - Add Katharina Borchert to Leadership page

craigcook:
This is ready for initial review, but shouldn't merge until her Mozillians profile is public and her press bio is live and we can verify the links.


alexgibson:
r+ on the code changes. Good to merge whenever the bio URL is ready :steam_locomotive:

alexgibson:
Looks like the bio is now live: https://blog.mozilla.org/press/bios/katharina-borchert/

The [Mozillians profile](https://mozillians.org/en-US/u/Katharina/) does not seem to be public yet, but perhaps we can merge anyway? I can confirm the URL is correct after logging into Mozillians.

craigcook:
As of right now Katharina's profile is still non-public but rather than keep waiting we'll go ahead and merge. The URL is correct, it just needs to be set to public on the other end.

PR #3763, opened 2016-01-12
[bug 1237960] Remove orphaned firstrun JS files

alexgibson:
@jpetto - I just noticed 2 more JS files that I believe are now orphaned and can be removed also.


jpetto:
Verified these JS files are not referenced by any bundles. r+ :memo:

PR #3764, opened 2016-01-12
Remove unused button URL

kyoshino:
`mozorg/button.html` was removed in d8921e105bb6a77ebc2654dbcb2bf876b7244e43 but `mozorg/urls.py` still has the URL which is 404. This PR will simply remove it. No bug.


alexgibson:
Thanks @kyoshino r+ :chart_with_upwards_trend:

PR #3766, opened 2016-01-13
Add tests for partnerships page

davehunt:
Simply happy path test for requesting a partnership and a negative test when required fields are missing. This replaces the test in [mcom-tests](https://github.com/mozilla/mcom-tests/blob/56382844fe28c27d285c3aecb7614e31bdf4dd39/tests/test_partnerships.py), which only checked visibility of elements. It may be worth later adding interactions with more form fields, or adding tests for any field validation such as e-mail and website.

@alexgibson r?


alexgibson:
LGTM for the time being, thanks @davehunt r+

PR #3767, opened 2016-01-13
Rerun partnerships request test

alexgibson:
@davehunt - I should have picked this up in review. I think it's worth adding here.


alexgibson:
Closing this for now - after discussing on IRC we're going to see if we actually get any failures for this form first, before we add this.

PR #3768, opened 2016-01-13
Add functional test for 404 page

alexgibson:
@davehunt r?

### DIFF
--- a/tests/functional/test_not_found.py
+++ b/tests/functional/test_not_found.py
@@ -1,0 +1,14 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.not_found import NotFoundPage
+
+
+@pytest.mark.smoke
+@pytest.mark.nondestructive
+def test_is_download_button_displayed(base_url, selenium):
+    page = NotFoundPage(base_url, selenium).open()
+    assert page.download_button.is_displayed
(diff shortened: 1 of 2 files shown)


alexgibson:
Hmm... seems this gives the old legacy 404 page on prod and so the test fails... will investigate why.

tests/pages/page_not_found.py
```diff
@@ -0,0 +1,18 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
davehunt: Nit: No need for `page` in the file name as it's in the `pages` directory.

tests/functional/test_page_not_found.py
```diff
@@ -0,0 +1,13 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.page_not_found import NotFoundPage
+
+
+@pytest.mark.nondestructive
+def test_is_download_button_displayed(base_url, selenium):
```
davehunt: Should this be a smoke test?

alexgibson:
I believe we may be blocked on adding a test for this page by [Bug 1008100](https://bugzilla.mozilla.org/show_bug.cgi?id=1008100). I could add a test that triggers a 404 page using a known URL that passes through to bedrock, such as `/foundation/foobar/`, but this won't pass locally however when in dev mode.

davehunt:
If this is valuable coverage we could potentially make the path conditional based on environment, or mark the test as expected to fail based on environment. The best solution is getting the dependent bug fixed though. :smile:

alexgibson:
Now that #3942 is merged and deployed, I believe this is now unblocked from merging. 

Quick r @davehunt?

davehunt:
r+

PR #3772, opened 2016-01-14
Update Selenium to version 2.49.0

alexgibson:
@davehunt r?

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -34,9 +34,9 @@ pytest-variables==1.2
 # sha256: hP6NW_Tc3MSQAkRsR6FG0XrBD6zwDZCGZZBkrEO2wls
 requests==2.8.1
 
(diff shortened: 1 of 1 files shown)


davehunt:
r+ tests pass locally. Note that this is just updating the Python client, and will not update the Selenium server versions used by Sauce Labs or Docker in the build pipeline.

To update the server versions used in the pipeline, we must change the value of SELENIUM_VERSION in the global configuration in Jenkins. Note that this will affect both Sauce Labs and Docker jobs, and it appears that 2.49 is not yet available in [docker-selenium](https://github.com/SeleniumHQ/docker-selenium). To keep Docker pinned, we can set the SELENIUM_VERSION environment variable within the job configuration that uses Docker.

davehunt:
For what it's worth, the 2.49 client should be fully compatible with the 2.48.2 server, so there's no need to block merging this. If for some reason there is an incompatibility, we can simply revert this commit.

PR #3773, opened 2016-01-14
The current version of ios is 1.4, not 1.2

sylvestre:


alexgibson:
These looks like just test fixtures - I think you may be wanting to update this setting? https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L1111-L1112 (See #3770)

sylvestre:
Of course, sorry :/

alexgibson:
Thanks @sylvestre - this was only a very recent change :+1:

alexgibson:
Updated product details locally and I'm seeing `/firefox/ios/1.4/releasenotes/`display correctly. 

Thanks @sylvestre r+

PR #3760, opened 2016-01-11
Default GIT_COMMIT to HEAD running tests

jgmize:
This is mainly to make things more convenient when running locally.

PR #3774, opened 2016-01-14
Consolidate functional test scripts and enable verbose output

davehunt:
This should allow us to have a single bedrock_functional_tests job in Jenkins that accepts parameters. The defaults will be set to use Selenium Grid in Docker (as in the Build phase), and these will then be overridden when calling this job from the various test jobs using the properties files in the repo. This will allow control of the platforms, etc from the repo instead of storing all of that in Jenkins.

Pinging @glogiotatidis or @jgmize for review.


alexgibson:
@davehunt looks like this one needs a rebase

docker/jenkins/run_functional_tests.sh
```diff
@@ -1,45 +1,67 @@
-#!/bin/bash
-#
-# Starts Selenium Hub and NUMBER_OF_NODES (default 5) firefox nodes.
-# Waits until at least one node is ready and then runs smoke tests
-# only against a local bedrock instance started for this job.
-#
-set -xe
+#!/bin/bash -xe
+cp docker/dockerfiles/bedrock_functional_tests Dockerfile
+docker build -t bedrock_functional_tests:${GIT_COMMIT} --pull=true .
```
jgmize: I think https://github.com/mozilla/bedrock/pull/3760/files#diff-4db431ac309eff3c872644e2c0ee9b9bR10 is likely the source of the merge conflict, and if you include that change when you rebase it will hopefully make things a little more convenient when running locally.

jgmize:
Nice work @davehunt, this lgtm so far and I'll take a second pass after the rebase.

davehunt:
@alexgibson @jgmize rebased and incorporated neat enhancement from https://github.com/mozilla/bedrock/commit/4280807b5ac688a589bf2f4538c8e88bd94d1cdd

jgmize:
thanks @davehunt!

PR #3775, opened 2016-01-14
[fix bug 1239656] Clean up l10n tags on /firefox/partners page

alexgibson:


flodolo:
Thanks, it's an r+ on my side :balloon: 

Verified locally by cleaning up the file, and also confronting with the .lang file generated by Bedrock with this patch applied.

alexgibson:
Closing this in favor of #3781

PR #3776, opened 2016-01-14
Update sanity functional test markers

alexgibson:


alexgibson:
@davehunt r?

Given that we have to skip tests in IE where the info bar is triggered, I think it's worth having a sanity test specifically for the download button displaying correctly. Thoughts?

tests/functional/firefox/test_new.py
```diff
@@ -10,10 +10,17 @@
...
+@pytest.mark.nondestructive
+def test_download_button_displayed(base_url, selenium):
+    page = FirefoxNewPage(base_url, selenium).open()
+    assert page.is_download_button_displayed
+
+
 # skip tests that trigger download bar in IE due to https://github.com/SeleniumHQ/selenium/issues/448
 @pytest.mark.skip_if_firefox
 @pytest.mark.skip_if_internet_explorer
-@pytest.mark.smoke
```
alexgibson: No point in this one being a smoke test, since we currently skip it in Firefox

davehunt: I believe the smoke tests will also run in Google Chrome, is that enough to keep the marker on this test?

tests/pages/firefox/new.py
```diff
@@ -14,10 +14,15 @@ class FirefoxNewPage(FirefoxBasePage):
     _download_button_locator = (By.CSS_SELECTOR, '#features-download .download-button .download-link')
     _thank_you_message_locator = (By.ID, 'scene2-main')
 
-    def download_firefox(self):
+    @property
+    def is_download_button_displayed(self):
         els = [el for el in self.find_elements(self._download_button_locator) if el.is_displayed()]
         assert len(els) == 1, 'Expected one download button to be displayed'
-        els[0].click()
+        return els[0]
```
davehunt: A property prefixed `is_` should really return a boolean. We could create a private `_download_button` property that can be used from both `is_download_button_displayed` and `download_firefox`.

davehunt:
r+ wc

tests/functional/firefox/test_new.py
alexgibson: Good point! Will revert this deletion

alexgibson:
Thanks @davehunt - does this look ok?

tests/pages/firefox/new.py
```diff
@@ -14,10 +14,19 @@ class FirefoxNewPage(FirefoxBasePage):
...
+
+    @property
+    def is_download_button_displayed(self):
+        return self._download_button.is_displayed()
+
+    def download_firefox(self):
+        button = self._download_button
+        button.click()
```
davehunt: These can be one line. :)

davehunt:
r+ wc

PR #3777, opened 2016-01-14
Fix bug 1239801 - home promo missing App Store button on iOS

craigcook:
Fix bug 1239814 - home promo shows desktop links for Android devices

PR #3765, opened 2016-01-13
[fix bug 1237003] Update Tracking Protection tour strings for normal browsing mode

alexgibson:

PR #3779, opened 2016-01-14
[fix bug 1238061] Add iOS promos to /firefox/sync.

jpetto:


bedrock/firefox/templates/firefox/sync.html
```diff
@@ -212,9 +214,12 @@
   <div class="inner-wrapper">
     <div class="container">
       <header>
-        <h3>{{ _('Have an <b>Android smartphone</b> or <b>tablet</b>?') }}</h3>
-        <a href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}" class="sync-button green" data-interaction="button click" data-element-action="Sync CTA" id="cta-android-footer">
-          {{ _('DOWNLOAD FIREFOX FOR ANDROID') }}
+        <h3>{{ _('Get Firefox everywhere') }}</h3>
```
jpetto: @flodolo - This string exists on `/firefox/new`, though it is behind an `l10n` tag. However, according to the .lang files checker(http://l10n.mozilla-community.org/~pascalc/langchecker/?website=0&file=firefox/new.lang&action=translate&show), that string is translated for 92% of our user base.

I'm guessing the answers below are both "yes", but just to be explicit:

1. Is that a high enough percentage to merge as-is?

2. Should I `add_lang_files "firefox/new"` for this page?

flodolo: 2 Is not a good idea, we would import the state of that file (e.g. if new.lang is active, we activate sync even if it's not complete).

Given that we already have almost 60 locales, I'll copy that string into a shared file and it will be picked up automatically. So, merge at it is.

flodolo: Actually, we don't even need that, because [base-resp already includes firefox/new](https://github.com/jpetto/bedrock/blob/bug-1238061-add-ios-promos-to-sync-page/bedrock/firefox/templates/firefox/base-resp.html#L5)

In fact the title is already translated at the bottom
https://www-demo5.allizom.org/it/firefox/sync/

jpetto: Excellent! Thanks @flodolo.

alexgibson:
It would be good to [update this test](https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/test_sync.py#L19) to check that the new iOS App Store button is displayed, as well as the Android one.

media/css/sandstone/buttons.less
```diff
@@ -637,6 +637,7 @@ html[lang^='en'] {
 .android .download-button-desktop .download-list,
 .android .download-button-desktop small.os_win,
 .download-button-ios .os_ios,
+.ios .download-button .os_ios,
```
alexgibson: Is this still needed after a rebase, now that #3777 merged?

media/css/firefox/sync.less
```diff
@@ -133,6 +137,18 @@
     }
 }
 
+.primary-buttons {
+    display: flex;
+    justify-content: space-around;
```
alexgibson: Mixins aren't strictly needed here as this is shown to Firefox users only, but in #3765 I'm adding some just for general usefulness. Up to you if you would like use them here (happy to consolidate strategies, so to speak :smile:).

alexgibson:
A much needed update to this page, looks great :+1: 

Couple of questions / things that could be updated, but otherwise it's an r+wc from me :arrows_counterclockwise:

media/css/firefox/sync.less
jpetto: Happy to synchronize Swatches (did you have that show in the UK?). Should we get #3765 merged first, then I can rebase and convert?

media/css/sandstone/buttons.less
jpetto: Nope, #3777 fixed the issue. Will rebase.

media/css/firefox/sync.less
alexgibson: If you have time to review #3765 that would be great :smile: 

If this is a priority I'm happy to rebase my branch however if this merges first

jpetto:
@alexgibson - Rebased, moved to flexbox mixins, and updated the test.

tests/functional/firefox/test_sync.py
```diff
@@ -23,6 +23,13 @@ def test_android_download_button_displayed(base_url, selenium):
 
 @pytest.mark.smoke
 @pytest.mark.nondestructive
+def test_ios_download_button_displayed(base_url, selenium):
```
alexgibson: Nit: I think we can probably combine the two app store button tests into a single test, something like `test_app_store_buttons_displayed`?

tests/pages/firefox/sync.py
```diff
@@ -13,6 +13,7 @@ class FirefoxSyncPage(FirefoxBasePage):
 
     _primary_download_button_locator = (By.ID, 'download-button-desktop-release')
     _android_download_button_locator = (By.ID, 'cta-android-footer')
+    _ios_download_button_locator = (By.ID, 'cta-ios-footer')
```
alexgibson: We could also rename these locators to something a bit more semantic, now that we're using store badges instead of download buttons. Something like `_play_store_button_locator` and `_app_store_button_locator`?

tests/functional/firefox/test_sync.py
jpetto: Oh, certainly. Wasn't sure if it was better to have independent or combined tests. Will update.

alexgibson: Usually it's better to have separate tests for things have that separate interactions, but I think here it's fine to combine them as we're just checking the display of two buttons in the same region.

jpetto:
Improved naming of test vars/methods. :name_badge:

alexgibson:
LGTM, care to squash?

jpetto:
Squarshed!

PR #3780, opened 2016-01-15
Fix Bug 1226476 - Convert android/all/ to the same format as desktop uses

kyoshino:
#3728 was reverted in #3778 due the /thunderbird/all/ bustage. I have added the fix and basic test, taken from firefox.tests.test_base, as another commit for faster review. Once reviewed, I'll squash them. @jgmize r?


jgmize:
@kyoshino that last commit introduced a flake8 error: `bedrock/firefox/views.py:23:1: F401 '_' imported but unused`

kyoshino:
Yup, fixed the error.

kyoshino:
And squashed.

PR #3781, opened 2016-01-15
[fix bug 1239960] Decommission /firefox/partners page

alexgibson:
Still waiting to confirm what the redirects should be, so marking as do-not-merge for now.


alexgibson:
Redirects [confirmed in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1239960#c5), removing do-not-merge flag.

alexgibson:
Note: I may have to update an mcom redirect test before this merges.

alexgibson:
@jpetto - updated the `/mobile` and `/fennec` redirects to point to the product page. Thanks

bedrock/base/templates/macros.html
```diff
@@ -522,7 +522,6 @@
           <ul class="subnav" id="os-subnav">
             <li><a data-id="os-index" href="{{ url('firefox.os.index') }}"{% if activesub == 'index' %} class="selected"{% endif %}><div>{{ _('Overview') }}</div></a></li>
             <li><a data-id="os-devices" href="{{ url('firefox.os.devices') }}"{% if activesub == 'devices' %} class="selected"{% endif %}><div>{{ _('Devices') }}</div></a></li>
-            <li><a data-id="os-partners" href="{{ url('firefox.partners.index') }}"{% if activesub == 'partners' %} class="selected"{% endif %}><div>{{ _('For partners') }}</div></a></li>
```
jpetto: Super duper small nit - update instructions on the macro to not include `partners` as a valid sub nav ID for `os`.

bedrock/settings/static_media.py
```diff
@@ -1547,43 +1530,6 @@
         ),
         'output_filename': 'js/logo-prototype-bundle.js',
     },
-    'partners': {
-        'source_filenames': (
-            'js/libs/modernizr.custom.shiv-load.js',
```
jpetto: I don't see any more references to this Modernizr build. Can probably delete?

media/css/firefox/partners/desktop.less
```diff
@@ -1,861 +0,0 @@
...
-@import '../../sandstone/lib.less';
-
-/* {{{ Journal font */
-
-@font-face {
-    font-family: 'Journal';
-    src: url('/media/fonts/journal-webfont.eot');
-    src: url('/media/fonts/journal-webfont.eot?#iefix') format('embedded-opentype'),
-         url('/media/fonts/journal-webfont.woff') format('woff'),
-         url('/media/fonts/journal-webfont.ttf') format('truetype');
```
jpetto: Doesn't look like this font is used anywhere anymore. Should we remove it, or keep the files around for possible later use?

bedrock/firefox/views.py
```diff
@@ -308,28 +227,6 @@ def firefox_os_geo_redirect(request):
...
-        'locale_headline_url': LOCALE_FXOS_HEADLINES[press_locale]['url'],
-        'locale_headline_title': LOCALE_FXOS_HEADLINES[press_locale]['title'],
-        'js_common': JS_COMMON,
-        'js_mobile': JS_MOBILE,
-        'js_desktop': JS_DESKTOP,
-    }
-
-    form_kwargs = {
-        'interest_set': 'fx',
-        'lead_source': 'www.mozilla.org/firefox/partners/'}
```
jpetto: With the removal of this page, the [custom interest set](https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/forms.py#L253) is no longer in use. I feel it's a good thing to retain the ability to easily change the interests displayed, but I'm not exactly sure what we should do (if anything?) with the `WebToLeadForm` code at this point. [Here's one take on what we could do](https://gist.github.com/jpetto/7d4dde2790c9dbf50d1a).

Thoughts? Maybe @pmclanahan wants to weigh in?

jpetto:
Is this going to reduce our JS codebase by like 25% or what? Not sorry to see some of those files go.

Overall looks good (and really thorough). Couple questions regarding related code/files.

bedrock/settings/static_media.py
alexgibson: Will remove :+1:

media/css/firefox/partners/desktop.less
alexgibson: Good spot. I seem to remember this particular font is not very good for l10n, so I say let's remove :smile:

bedrock/firefox/views.py
alexgibson: Agree on retaining the use of this feature, will defer to yourself and @pmclanahan on what should be done (if anything) :+1:

bedrock/base/templates/macros.html
alexgibson: Will do!

alexgibson:
Addressed review feedback. Only thing I think remains is the question on the custom interest set. Thanks @jpetto!

jpetto:
Looks good! r+ :sweet_potato:

PR #3782, opened 2016-01-15
[fix bug 1239732] Developer page shows both Desktop and Android downoad buttons on IE6

alexgibson:
Note: Dev Edition downloads are currently 404 on XP, but hopefully that will be fixed very shortly in [bug 1239685](https://bugzilla.mozilla.org/show_bug.cgi?id=1239685)

### DIFF
--- a/bedrock/firefox/templates/firefox/developer.html
+++ b/bedrock/firefox/templates/firefox/developer.html
@@ -40,7 +40,7 @@
 
   <h1>{{ _('Built for those who build the Web') }}</h1>
   <h2>{{ _('The only browser made for developers like you.') }}</h2>
-  {{ download_firefox('alpha', icon=False, small=True, simple=True) }}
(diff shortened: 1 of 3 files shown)


jpetto:
This fix totally works. r+

I did a quick search to see if any other pages suffered from the same problem, and the only one I could find was [firefox/installer-help](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/installer-help.html#L70), which might be rather important for IE 6 users. Want to add the same fix to that template? I'll preemptively r+ the change. :smile_cat:

alexgibson:
Good spot - I've applied the same fix to the Dev Edition download button on the `/firefox/installer-help/` page and added a note to the commit message. Thanks @jpetto :+1:

alexgibson:
@jpetto - I had to update a couple of failing tests for the installer help page. Straight forward enough, but I'd appreciate a quick second before merging.

jpetto:
Test updates look good (and pass). r+!

PR #3783, opened 2016-01-15
[bug 1237960] Remove redundant /whatsnew FF 37 files

alexgibson:


bedrock/firefox/forms.py
```diff
@@ -10,8 +10,6 @@
 
 from lib.l10n_utils.dotlang import _lazy as _
 
-LANG_FILES = ['firefox/whatsnew-fx37']
-
```
jpetto: Not specific to this PR (as `firefox/whatsnew-fx37.lang` is already gone), but just checking that the string below ("Sorry. This number isn't...") is still able to be translated.

After a bit of testing, it looks like the `invalid` error message below is never used. I see a similar error message [in the `send_to_device` macro](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/macros.html#L606). Does that error message render the one below moot? Not really clear if we should remove the one below, or maybe just add a comment stating that it's never displayed?

Or I could just be confused.

jpetto:
Code is an r+, but would like a little more clarification on the `invalid` error string's status.

bedrock/firefox/forms.py
flodolo: I'm not completely sure that message was ever translated (we have about U.S. phones in sendto.lang).

jpetto: Looks like it was [translated for at least some locales](https://github.com/mozilla-l10n/www.mozilla.org/commit/06c998d2d46dcaf42afcae61502dd9593cc41b68#diff-399337a0fd4833ee8c0b111db134fe73L73).

alexgibson: I believe this string as actually remnants of the even older /firefox/sms page. It may have only been en-US originally in this case, and just serves as a default message. We could remove it, but then perhaps it is better to leave something there as a default?

flodolo: That page was done only for a [very small set of locales](https://github.com/mozilla-l10n/langchecker/commit/b732eca001f5a2405e2c93f1f106315bc84801f5#diff-913ec76ed4663fa42b51ed974ca60902L922) (that's one of the few I never dealt with).

Is this message actively used in any page?

alexgibson: > Is this message actively used in any page?

I don't believe so, no. The send-to-device widget used different messaging for this error, which is in the macro template itself.

jpetto: I'd suggest leaving the `invalid` string as-is, but adding a comment stating that it's not actually in use or localized, but is left in as a placeholder for potential future use. How's that sound?

alexgibson: Sounds good to me, thanks @jpetto

alexgibson: Added a comment re: the default error message string status as suggested.

jpetto:
Thanks @alexgibson!

PR #3784, opened 2016-01-15
Use Firefox latest ESR to run JS tests in Travis

alexgibson:
Puts us onto a rolling FF version for running the JS tests in PR's, while still ensuring some level of legacy coverage.


jgmize:
Good idea, @alexgibson.

PR #3785, opened 2016-01-15
Move the redirect tests into the tests directory and consolidate with mcom-tests

davehunt:
This is an initial attempt to move the more coverage from mcom-tests into bedrock. I've also moved the redirect tests into the `tests` directory and added documentation on how to run the tests. This also allows the functional UI tests to run with a Django live server, however that's currently not working for me. Note that this means it's necessary to specify a `--base-url` if you want to run the UI tests.

Pinging @pmclanahan and @jgmize for feedback.
Pinging @alexgibson for visibility.


tests/redirects/base.py
```diff
@@ -103,7 +106,7 @@ def assert_valid_url(url, location=None, status_code=301, req_headers=None, req_
             # strip off query for further comparison
             resp_location = resp_location.split('?')[0]
 
-        assert resp_location == get_abs_url(location, base_url)
+        assert re.match(get_abs_url(location, base_url), resp_location) is not None
```
pmclanahan: I'm not sure we can trust this change w/o more changes to the tests. Every `.` needs escaping for example. Perhaps we could instead make using a regex an option, either by adding a boolean param (e.g. `regex=True`), or passing in a compiled regex as the test:

```python
url_test('/stuff/whatnot/', re.compile(r'/stfu/donnie'))
```

tests/redirects/test_urls.py
```diff
@@ -0,0 +1,72 @@
...
+def test_locale_url(url, base_url):
+    url['base_url'] = base_url
+    assert_valid_url(**url)
+
+
+@pytest.mark.headless
+@pytest.mark.nondestructive
+@pytest.mark.parametrize('url', EXTERNAL_URLS, ids=itemgetter('url'))
+def test_external_url(url):
+    assert_valid_url(**url)
```
pmclanahan: Just so I'm clear, the rest of the redirect tests can be run locally by using the `smoke` mark, right?

davehunt: You can use marker expressions on the command line to filter tests. By default all tests will run. I've used the `smoke` marker to indicate tests that would be safe to run at all stages on our pipeline, leaving this test because it hits external URLs. To run just the smoke tests, add `-m smoke` to your command line. To exclude smoke tests, add `-m 'not smoke'`.

tests/redirects/base.py
davehunt: You're right, that would be a better approach.

davehunt:
@pmclanahan I've updated this to support a compiled regex pattern.

jgmize:
This [failed](https://travis-ci.org/mozilla/bedrock/builds/103171929) due to http->https redirects; I think adding an `export DISABLE_SSL=True` to the `before_script` section in `.travis.yml` should fix that. Please give that a shot and feel free to ping me if it doesn't. Other than that it's looking pretty good so far. Once we have a green run from travis and a rebase I'll be happy to take a final pass and merge if we don't find any other issues.

tests/redirects/map_globalconf.py
```diff
@@ -938,5 +942,28 @@
...
+    # from mcom-tests
+    url_test('/firefox/', '/en-US/firefox/'),
+    url_test('/en-US/firefox/', '/firefox/new/'),
+    url_test('/firefox/new/', '/en-US/firefox/new/'),
+    url_test('/fennec/', '/firefox/partners/'),
+    url_test('/firefox/mobile/', '/firefox/android/'),
+    url_test('/mobile/37.0{,beta,a2}/releasenotes', '/firefox/android/37.0{,beta,a2}/releasenotes/'),
+    url_test('/projects/firefox/3.6.13/whatsnew/', '/firefox/3.6.13/whatsnew/'),
+    url_test('/en-US/firefox/3.6.13/{firstrun,whatsnew}/', '/en-US/firefox/new/'),
+    url_test('/b2g/', '/firefox/partners/'),
```
alexgibson: Note: this test will fail once #3781 merges (and also adds an equivalent redirect test). Will just need to rebase depending on whichever PR merges first.

davehunt:
@jgmize I can't seem to get this to work with `DISABLE_SSL=True`. Is there something else I can try? I'm not sure why this would redirect to https on Travis but not for me locally... https://travis-ci.org/mozilla/bedrock/builds/104624564#L126

tests/redirects/map_locales.py
```diff
@@ -0,0 +1,34 @@
...
+# obsolete ab-CD-style names, which could be included in the visitors'
+# Accept-Language HTTP header and should be redirected to the respective
+# canonical locales
+LOCALE_VARIANTS = {
+    'en-US': ['en', 'en-CA'],
+    'es-ES': ['es', 'es-419'],
+    'fr': ['fr-FR'],
+    'ja': ['ja-JP-mac'],
+    'pt-BR': ['pt'],
+    'ta': ['ta-LK']}
```
davehunt: Are these locales/variants a decent representation for testing?

pmclanahan:
I think the failures are actually due to https://github.com/mozilla/bedrock/blob/98d6654326dfd8ae25b126d96c924cc68433f9f3/bedrock/firefox/views.py#L431-L438 and `DEV = False`.

tests/redirects/map_locales.py
pmclanahan: Looks representative of the vast majority of cases. :+1:

alexgibson:
> I think the failures are actually due to https://github.com/mozilla/bedrock/blob/98d6654326dfd8ae25b126d96c924cc68433f9f3/bedrock/firefox/views.py#L431-L438 and DEV = False.

Discussed this in our PR triage and we can safely remove these lines of code and their associated tests. I'd be happy to try help on this if you're not sure @davehunt.

davehunt:
Thanks for the guidance @alexgibson.

@pmclanahan, @jgmize - This is now ready for review again.

bedrock/firefox/tests/test_base.py
```diff
@@ -275,15 +275,6 @@ def test_rv_prefix(self, render_mock):
...
-    def test_fx_australis_secure_redirect(self, render_mock):
-        """Should redirect to https: for 29.0."""
-        url = '/en-US/firefox/whatsnew/'
-        req = self.rf.get(url)
-        with patch.object(req, 'is_secure', return_value=False):
-            resp = self.view(req, version='29.0')
-        eq_(resp['location'], 'https://testserver' + url)
-
     @override_settings(DEV=True)
     def test_fx_australis_secure_redirect_not_dev(self, render_mock):
```
alexgibson: Is this test now redundant also?

bedrock/firefox/tests/test_base.py
```diff
@@ -419,15 +410,6 @@ def test_fx_firstrun_40_0(self, render_mock):
...
-    def test_fx_australis_secure_redirect(self, render_mock):
-        """Should redirect to https:"""
-        url = '/en-US/firefox/firstrun/'
-        req = self.rf.get(url)
-        with patch.object(req, 'is_secure', return_value=False):
-            resp = self.view(req, version='29.0')
-        eq_(resp['location'], 'https://testserver' + url)
-
     @override_settings(DEV=True)
     def test_fx_australis_secure_redirect_not_dev(self, render_mock):
```
alexgibson: Same here? (also `test_fx_australis_secure_redirect_secure` below might be safe to remove)

bedrock/firefox/tests/test_base.py
alexgibson: `test_fx_australis_secure_redirect_secure` below might also be ok to remove now.

jgmize:
Agreed, the page-specific secure redirect tests are redundant and safe to remove now.

davehunt:
Now with more code removal goodness.

jgmize:
Thanks @davehunt!

PR #3788, opened 2016-01-18
[fix bug 1233773] TP UITour should close tab when not invoked from Private Window

alexgibson:
setup
-------

* This needs testing using the latest Nightly, at least `46.0a1 (2016-01-17)`.
* Please make sure set Tracking Protection to `Always` in `about:preferences#privacy` (not just in Private Browsing mode).
* Make sure you have [UITour working locally](http://bedrock.readthedocs.org/en/latest/uitour.html#local-development).

URL to test locally: `/firefox/46.01a/tracking-protection/start/?newtab=true`

To test behavior 1:

1.) Open a new tab (make sure you have more than one tab open).
2.) When you get to step 3 of the tour, click "Got it".
Expected result: The tab should close.

To test behavior 2:

1.) Open a single tab (the tour URL being the only tab in the active window).
2.) When you get to step 3 of the tour, click "Got it".
Expected result: Step 4 should be shown

Extra test for good measure:

Load the URL without `?newtab=true`.
Expected result: Step 4 should always be shown.

### DIFF
--- a/docs/uitour.rst
+++ b/docs/uitour.rst
@@ -638,6 +638,21 @@ Accepts one of the following options to be passed as an `id`:
 
     Only available in Firefox 42 onward.
 
+closeTab();
+^^^^^^^^^^^
+
+Closes the current tab.
+
+.. code-block:: javascript
+
+    Mozilla.UITour.closeTab();
+
+.. Important::
+
+    This function will do nothing when called from the last browser window when it contains
+    only one tab. You may need to provide a work around for this edge case in your code.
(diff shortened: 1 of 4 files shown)


alexgibson:
Note to self: I still need to update the UITour docs here.

alexgibson:
Added a note to the docs for the new `closeTab()` function.

schalkneethling:
Works as advertised and Karma tests all pass r+ :lock:

PR #3793, opened 2016-01-20
Add functional test for Firefox Installer Help page

alexgibson:
Also makes our download button display tests a bit more robust, by ensuring only one download button is displayed (useful for catching visual regressions like in #3782).

### DIFF
--- a/tests/pages/base.py
+++ b/tests/pages/base.py
@@ -34,6 +34,15 @@ class BasePage(Page):
     def newsletter(self):
         return NewsletterEmbedForm(self)
 
+    def download_button(self, locator):
+        """Return the displayed download button link determined by platform.
+
+        Only for use by page objects and not directly from tests.
+        """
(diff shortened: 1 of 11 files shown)


alexgibson:
@davehunt r?

tests/pages/base.py
```diff
@@ -34,6 +34,11 @@ def footer(self):
     def newsletter(self):
         return NewsletterEmbedForm(self)
 
+    def download_button(self, locator):
```
davehunt: I think we should add a docstring for this, or potentially mark is as private. It could be confusing for a test writer as this is only meant to be used from the page objects and not directly from a test.

davehunt:
r+ wc

tests/pages/base.py
alexgibson: Good suggestion, added a docstring.

PR #3795, opened 2016-01-20
[fix bug 1208495] Remove unused newsletter macro.

jpetto:
Follow-up to #3615.


craigcook:
`#mozorg-newsletter-form` is also referenced in form.js at https://github.com/mozilla/bedrock/blob/master/media/js/newsletter/form.js#L73 I think we can remove that function.

craigcook:
LGTM r+wc :newspaper:

jpetto:
Ah, good catch on the JS reference. Updated!

PR #3796, opened 2016-01-21
Fix failing Firefox Hello test

alexgibson:
@davehunt - I should have thought of this in the #3793. We can no longer test for a button not being displayed quite as easy. I'd like to address this properly, but this is a quick fix for now.


davehunt:
r+ perhaps a DownloadButton region can abstract some of this

PR #3797, opened 2016-01-21
[fix bug 1222986] Simplify conditional home page promos

alexgibson:
Still waiting on confirmation that the promos/locations/colours are all correct.

This PR removes much of the spaghetti logic and moves to a simpler solution for l10n. At maximum there will only ever be two possible promos in any given spot. The fallback promos are considered "evergreen" and will never change. When a new promo is added to a spot, the previous "non evergreen" promo is to be removed. This way we avoid the continual build up of promos, which will make maintenance simpler and localization easier to manage.


bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -301,17 +226,7 @@ <h2 class="primary go">{{ _('Mozilla Developer Network') }}</h2>
...
-          <h2 class="primary go">{{ _('Introducing <br>Firefox Friends') }}</h2>
-          <a class="panel-link" href="https://friends.mozilla.org" data-promotion-id="promo-11" data-promotion-name="Introducing Firefox Friends" data-promotion-class="promo-large-landscape" data-promotion-type="tile">
-            <div class="secondary">
-              <h3>{{ _('Join. Share. <br>Get rewarded.') }}</h3>
-              <p class="more">{{ _('Join Firefox Friends') }}</p>
-            </div>
-          </a>
-        </li>
-      {% else %}
+        {# Since MDN promo is evergreen we can't use a l10n tag so comment out the fallback.
```
alexgibson: @flodolo - just checking that strings within this comment are still extracted ok? (I wasn't 100% sure if this is ok)

flodolo: Yes, they are.

alexgibson:
Jen has confirmed in the the bug that the promo's and their respective fallback are all correct. @flodolo - any idea of the fallout here? I will defer to you on who/when to move forward.

flodolo:
Some notes:
* We should get rid of the tag `meta_update_201503` too.
* I'm not sure I understand why the MDN tile is potentially there twice. The base layer should not have duplicated items, in fact the base layer should give us a good starting point.
* We will have to disable quite a few locales. And what scares me is that the .lang file is the same before and after, which means we need to coordinate updates to l10n and push to prod.

alexgibson:
> We should get rid of the tag meta_update_201503 too.

Will do

> I'm not sure I understand why the MDN tile is potentially there twice. The base layer should not have duplicated items, in fact the base layer should give us a good starting point.

I've contested this in the bug already as well. Seems this is what they want to move forward with. I've made it clear in terms of implications and Jen says to move forward still.

> We will have to disable quite a few locales. And what scares me is that the .lang file is the same before and after, which means we need to coordinate updates to l10n and push to prod.

Whatever we need to do we can. Your call on how & when thanks.

flodolo:
> Whatever we need to do we can. Your call on how & when thanks.
I think the easiest way is to disable all the incomplete locales, then cleanup the file after this code has moved to production.

alexgibson:
@flodolo - updated as to what we agreed in the bug. This should be good for another look. Once you're happy I'll open it up for review. Thanks

flodolo:
Still need to cleanup the [meta tags](https://github.com/alexgibson/bedrock/blob/simplify-homepage-promos/bedrock/mozorg/templates/mozorg/home/home.html#L11-L23). Besides that, is good for review on my side.

flodolo:
Also: [mail sent to dev-l10n-web](https://groups.google.com/d/msg/mozilla.dev.l10n.web/qGr1CSdhkU4/x3Uj0tL4DgAJ).

alexgibson:
Thanks @flodolo - cleaned up the meta tags. Marking as ready for review.

media/css/mozorg/home/home-promo.less
```diff
@@ -791,33 +769,11 @@ html[lang|="en"] {
...
-            padding: 34px @baseLine @baseLine 120px;
-        }
-    }
-}
 
 /****************************************************************************/
 // Volunteer
 .promo-small-landscape.volunteer {
-    background-color: #56565a;
+    background: #0ad; // bright blue
```
craigcook: Any reason not to use shorthand for the whole thing?
`background: #0ad url(/media/img/home/voices/promos/volunteer/volunteer.png) 15px 25px no-repeat;`

craigcook:
Just the one non-blocker nitpick, and probably a few other spots we could use shorthand for backgrounds if you want to. Otherwise r+wc. :fried_shrimp:

alexgibson:
Thanks @craigcook updated - I've used shorthand on the promo I'm updating here, but I didn't go down the route of touching the others in this PR.

PR #3798, opened 2016-01-22
Fix Bug 1237070, implement lean data practices page

schalkneethling:
Waiting on one link, but the rest is open for code review. @craigcook r? Also note, this page is not localized and support is IE9 and above. As long as the content is readable in IE8 and lower, we are all good.


schalkneethling:
Just noticed some of the SVG files can do with some optimization. Will update the PR with those.

schalkneethling:
Ok, added the Github link and optimized the SVG files.

media/css/mozorg/about-lean-data.less
```diff
@@ -0,0 +1,303 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../sandstone/lib.less";
+
+.circular-image(@color: #333, @width: 180px, @height: 180px) {
```
craigcook: Doesn't look like this mixin us used anywhere.

schalkneethling: Ah yes, we no longer have circular images, good catch. removing.

craigcook:
r+ :rocket:

schalkneethling:
Ok @craigcook, all the ducks are in a row, merge at will. Thx

PR #3799, opened 2016-01-22
Use DownloadButton page region in functional tests

alexgibson:


alexgibson:
@davehunt r? Interested in your thoughts on this approach

tests/pages/regions/download_button.py
```diff
@@ -0,0 +1,30 @@
...
+        assert len(els) == 1, 'Expected one platform link to be displayed'
+        return els[0]
+
+    @property
+    def is_platform_link_displayed(self):
+        return self.platform_link.is_displayed()
+
+    @property
+    def is_displayed(self):
+        return self._root.is_displayed()
```
davehunt: How about making `platform_link` private, and in `is_displayed` doing something like:

```python
@property
def is_displayed(self):
    if self._root.is_displayed():
        return self._platform_link.is_displayed()
    else:
        return False
```

This could even be shortened to:

```python
@property
def is_displayed(self):
    return self._root.is_displayed() and self._platform_link.is_displayed() or False
```

davehunt:
r+ wc

tests/pages/regions/download_button.py
alexgibson: Nice suggestion, I like this approach :+1:

PR #3803, opened 2016-01-25
Add functional tests for Firefox iOS page

alexgibson:
@davehunt r?

### DIFF
--- a/media/css/firefox/ios.less
+++ b/media/css/firefox/ios.less
@@ -624,7 +624,8 @@ a.go {
     top: -30px;
 }
 
-.send-to-device {
+.send-to-device,
+.send-to {
     display: none;
 }
 
@@ -791,4 +792,3 @@ html[lang|="en"] {
       margin-top: 10px;
   }
 }
-
--- a/tests/functional/firefox/test_ios.py
+++ b/tests/functional/firefox/test_ios.py
@@ -1,0 +1,34 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+from selenium.common.exceptions import TimeoutException
+
+from pages.firefox.ios import IOSPage
+
+
+@pytest.mark.flaky(reruns=1, reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1218451')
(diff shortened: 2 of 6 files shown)


media/css/firefox/ios.less
```diff
@@ -624,7 +624,8 @@ a.go {
     top: -30px;
 }
 
-.send-to-device {
+.send-to-device,
+.send-to {
```
alexgibson: Note: this doesn't have any dependency on the test passing/failing here. Just fixing a visual discrepancy with a cropped button:

button-cropped

I supposed I *could* add a test for this also, although I wouldn't consider it the main page CTA.

alexgibson:
This is blocked by #3808. Will rebase after that merges and update.

tests/functional/firefox/test_ios.py
```diff
@@ -0,0 +1,34 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+from selenium.common.exceptions import TimeoutException
+
+from pages.firefox.ios import IOSPage
+
+
+@pytest.mark.flaky(reruns=1)
```
davehunt: It occurs to me that we should include a reason when marking tests as flaky. It should be safe to use `reason="My reason"` here, and the pytest-rerunfailures plugin could be enhanced to report this reason in the console summary.

tests/pages/firefox/ios.py
```diff
@@ -0,0 +1,30 @@
...
+    def is_app_store_button_displayed(self):
+        return self.is_element_displayed(self._app_store_button_locator)
+
+    @property
+    def is_send_to_device_button_displayed(self):
+        return self.is_element_displayed(self._send_to_device_button_locator)
+
+    def click_get_it_now(self):
+        modal = Modal(self)
+        self.find_element(self._send_to_device_button_locator).click()
```
davehunt: This is a bit confusing as there's a 'Get it now' button and then a 'Send' button. It might simply be that we need to rename this locator, or perhaps it will become more clear when you've brought in the changes from #3808.

tests/functional/firefox/test_ios.py
alexgibson: Makes sense, will do :+1:

alexgibson:
@davehunt - updated, thanks

davehunt:
r+

PR #3805, opened 2016-01-25
Add RewriteRule for /firefox/android/beta/all/

kyoshino:
Fix the intermittent 404. Follow up #3780. Part of Bug 1226476.


kyoshino:
@pmclanahan r?

PR #3806, opened 2016-01-25
Fix Bug 1239531 - Google logo on Play Store button is outdated

kyoshino:
* The new badge is available only in 11 locales so far.
* Due to the original images’ inconsistency, the dimension and colour tone vary slightly by locale.
* The images are optimized with ImageAlpha+Optim.


craigcook:
Sorry for the delay. r+ 
Let's hope Google adds more locales soon.

PR #3808, opened 2016-01-26
Fix intermittent failure in Android form submission test

alexgibson:
@davehunt r?

Looking at the sauce job here it seems IE sometimes triggers a scroll when filling in the form, preventing the submit button from being clicked due to the sticky header. Seen this intermittent a couple of times now.

http://saucelabs.com/jobs/34351f809d864d92acb92918ac205786

This change moves `SendToDevice` into it's own file in `pages.regions`, which I think makes sense as we don't include it in all our Firefox base pages. It also now inherits `FirefoxBasePageRegion` and calls `scroll_element_into_view` before clicking submit, which should hopefully fix the intermittent.

### DIFF
--- a/tests/pages/firefox/android.py
+++ b/tests/pages/firefox/android.py
@@ -5,6 +5,7 @@
 from selenium.webdriver.common.by import By
 
 from pages.firefox.base import FirefoxBasePage, FirefoxBasePageRegion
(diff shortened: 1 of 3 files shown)


davehunt:
r+

PR #3815, opened 2016-01-27
Fix Bug 1225170, implement new signup and task pages for contribute

schalkneethling:
@jpetto r? So, I do not foresee many, if any, code changes here. The biggest thing still to-do is some cross browser testing and Optimizely. The sooner we can start code review though, the better. Thx

- [x] Implement base HTML structure
- [x] Ensure all strings are wrapped even if page will not be localised currently
- [x] Add new or additional styling referencing margins and padding from design files
- [x] Add any functionality that needs to be implemented with JS
- [x] Make responsive and test vaious scenarios using a combination of Fx responsive view and Chrome's device emulation.
- [x] Optimize images and use SVG where there is a performance/maintenance benefit
- [x] Test in Fx, Chrome and Safari
- [x] Test to ensure readibility in IE8 and below
- [x] Test IE9 and above
- [x] Test High-DPI
- [ ] Ensure no accessibility issues was introduced using WAVE or similar
- [x] Ensure URLs etc. is set-up for Optimizely testing


schalkneethling:
Ah yes so, those test can be expected to fail until we have sorted out how to integrate all of this with Optimizely.

bedrock/settings/static_media.py
```diff
@@ -984,6 +997,21 @@
         ),
         'output_filename': 'js/contribute-studentambassadors-join-bundle.js',
     },
+    'contribute-taskview': {
+        'source_filenames': (
+            'js/base/global.js',
```
alexgibson: Drive by comment - no need to include files again that are already in the main site JS bundle (both `global.js` and `mozilla-client.js`).

schalkneethling: Oh awesome! thx @alexgibson

bedrock/mozorg/templates/mozorg/contribute/signup-alternate.html
```diff
@@ -0,0 +1,117 @@
...
+        <div class="tooltip hidden">
+          <p>{{ _('Stay connected with us for news, announcements and opportunities from the Firefox team feed.') }}</p>
+        </div>
+      </li>
+    </ul>
+  </section>
+
+  <section class="tasks-challenging">
+    <header>
+      <h2>{{ _('Try something more challenging...') }}</h2>
```
flodolo: You probably want … here (one unicode character)

schalkneethling: Ah yes, thx @flodolo

bedrock/mozorg/templates/mozorg/contribute/signup-alternate.html
```diff
@@ -0,0 +1,123 @@
...
+{% block body_id %}inquiry{% endblock %}
+
+{% block page_title %}{{_('Get Involved: Volunteer Opportunities at Mozilla')}}{% endblock %}
+
+{% block page_css %}
+  {% stylesheet 'contribute-signup' %}
+{% endblock %}
+
+{% block optimizely %}
+  {% if LANG == 'en-US' and waffle.switch(‘contribute-signup-optimizely') %}
```
jpetto: There's a fancy single quote at the start of the waffle switch name that's throwing errors.

bedrock/mozorg/templates/mozorg/contribute/signup-alternate.html
```diff
@@ -0,0 +1,123 @@
...
+  <section>
+    <header>
+      <h2>{{ _('Try something simple') }}</h2>
+    </header>
+
+    <ul class="simple">
+      <li class="task">
+        <h3><a href="{{ url('mozorg.contribute.taskview') }}whimsy/?v=2" class="do-task outline-button more">{{ _('Customize Your Desktop Browser') }}</a></h3>
+
+        <a href="javascript:void(0);" class="tooltip-trigger" title="{{ _('more info') }}">
```
jpetto: Inline JavaScript! :fearful: 

Would these be better as `<button>` elements?

bedrock/mozorg/templates/mozorg/contribute/signup-alternate.html
```diff
@@ -0,0 +1,123 @@
...
+<article>
+
+  <section>
+    <header>
+      <h2>{{ _('Try something simple') }}</h2>
+    </header>
+
+    <ul class="simple">
+      <li class="task">
+        <h3><a href="{{ url('mozorg.contribute.taskview') }}whimsy/?v=2" class="do-task outline-button more">{{ _('Customize Your Desktop Browser') }}</a></h3>
```
jpetto: When browsing the page, I instinctively clicked on the image for each task. Though they have a hover state, they aren't active links. What do you think about making the image clickable as well?

bedrock/mozorg/templates/mozorg/contribute/signup-new.html
```diff
@@ -0,0 +1,122 @@
...
+{% block body_id %}inquiry{% endblock %}
+
+{% block page_title %}{{_('Get Involved: Volunteer Opportunities at Mozilla')}}{% endblock %}
+
+{% block page_css %}
+  {% stylesheet 'contribute-signup' %}
+{% endblock %}
+
+{% block optimizely %}
+  {% if LANG == 'en-US' and waffle.switch(‘contribute-signup-optimizely') %}
```
jpetto: Same fancy single quote issue here.

bedrock/mozorg/templates/mozorg/contribute/tasks/v1/devtools-challenger.html
```diff
@@ -0,0 +1,33 @@
...
+  </section>
+
+  <ol class="task-steps">
+    <li id="step_one" class="one">
+      <header>
+        <h3>{{ _('Go to Dev Tools Challenger') }}</h3>
+      </header>
+      <div class="step-content">
+        <img src="{{ static('img/contribute/signup/devtools.png') }}" width="620" class="feature-img" alt="" />
+        <a href="http://devtoolschallenger.com/" rel="external" class="button-blue action" class="taskview-button" data-action="challenger" data-task="devtools" data-step="one" data-complete="true" data-variation="1" target="_blank">{{ _('Learn More') }}</a>
```
jpetto: Two `class` attributes here.

bedrock/mozorg/templates/mozorg/contribute/tasks/v1/follow-mozilla.html
```diff
@@ -0,0 +1,28 @@
...
+      <div class="step-content">
+        <img src="{{ static('img/contribute/signup/mozparticipate.png') }}" width="620" class="feature-img" alt="" />
+        <a href="https://twitter.com/intent/follow/?screen_name=mozparticipate" rel="external" class="follow-mozilla button-blue twitter" data-action="follow" data-task="follow-mozilla" data-step="one" data-complete="true" data-variation="1">{{ _('Follow @MozParticipate') }}</a>
+      </div>
+    </li>
+  </ol>
+{% endblock %}
+
+{% block thank_you_msg %}
+  {{ _('Good to have you join!') }}
```
jpetto: This thank you message doesn't quite seem to fit the task.

bedrock/mozorg/templates/mozorg/contribute/tasks/v1/joy-of-coding.html
```diff
@@ -0,0 +1,31 @@
...
+{% block task_content %}
+  <ol class="task-steps">
+    <li id="step_one" class="one">
+      <header>
+        <h3>{{ _('Watch The Intro!') }}</h3>
+        <p>{{ _('<a href="https://twitter.com/mike_conley" rel="external">@mike_conley</a> livehacks on real Firefox bugs while thinking out loud.') }}</p>
+      </header>
+      <div class="step-content">
+        <video id="joc" preload="none" controls poster="{{ static('img/contribute/signup/joc_poster.png') }}" data-action="play" data-task="joyofcoding" data-step="one" data-complete="true" data-variation="1">
+          <source src="{{ static('img/contribute/signup/joc-video.mp4') }}" type="video/mp4" />
```
jpetto: Should we put this directly on `videos.cdn.mozilla.net` like we've done with past videos?

bedrock/mozorg/templates/mozorg/contribute/tasks/v1/whimsy.html
```diff
@@ -0,0 +1,36 @@
...
+{% block task_content %}
+  <ol class="task-steps">
+    <li id="step_one" class="one">
+      <header>
+        <h3>{{ _('Install Whimsy') }}</h3>
+        <p>{{ _('Whimsy is an add-on to help make your Firefox more enjoyable, like replacing the tiles on your new tab page with animated gifs!') }}</p>
+      </header>
+      <div class="step-content">
+        <img src="{{ static('img/contribute/signup/whimsy.png') }}" width="620" class="feature-img" alt="" />
+        <iframe src="https://addons.cdn.mozilla.net/static/html/mozorg/whimsy.html" class="action" frameborder="0" class="whimsy-install-container" data-step="one" data-complete="false" data-variation="2"></iframe>
```
jpetto: Two `class` attributes here.

bedrock/mozorg/templates/mozorg/contribute/tasks/v2/devtools-challenger.html
```diff
@@ -0,0 +1,51 @@
...
+
+{% block more_tasks %}{{ url('mozorg.contribute.signup') }}?v=2{% endblock %}
+
+{% block taskview_title %}{{ _('Complete a CSS Task using our all-new Developer Tools Challenger') }}{% endblock %}
+{% block taskview_lead %}{{ _('The more you know about the web, the easier it is to make it your own! Participate in the web by testing your web literacy skills by completing a CSS task in Firefox Developer Edition.') }}{% endblock %}
+
+{% block task_content %}
+  <section class="get-firefox">
+    <p>{{ _('Dev Tools requires Firefox Developer Edition desktop browser. To try this task, download Firefox Developer Edition first.') }}</p>
+    <a href="{{ url('firefox.developer') }}" class="button-blue dev-edition">Download Firefox Dev Edition</a>
```
jpetto: Not that we're localizing for this test, but this string isn't wrapped.

bedrock/mozorg/templates/mozorg/contribute/tasks/v2/devtools-challenger.html
```diff
@@ -0,0 +1,51 @@
...
+  </section>
+
+  <ol class="task-steps">
+    <li id="step_one" class="one">
+      <header>
+        <h3>{{ _('Go to Dev Tools Challenger') }}</h3>
+      </header>
+      <div class="step-content">
+        <img src="{{ static('img/contribute/signup/devtools.png') }}" width="620" class="feature-img" alt="" />
+        <a href="http://devtoolschallenger.com/#d1600" rel="external" class="button-blue action" class="taskview-button" data-action="challenger" data-task="devtools" data-step="one" data-complete="false" data-variation="2" target="_blank">{{ _('Take the Challenge!') }}</a>
```
jpetto: Two `class` attributes.

bedrock/mozorg/templates/mozorg/contribute/tasks/v2/devtools-challenger.html
```diff
@@ -0,0 +1,51 @@
...
+      </div>
+    </li>
+    <li id="step_two" class="two">
+      <header>
+        <h3>{{ _('Tell your friends on Twitter to try Developer Tools, too!') }}</h3>
+      </header>
+      <div class="step-content">
+        <form name="tweet" id="signup-tweet" class="tweet-form" action="https://twitter.com/intent/tweet" method="get">
+          <fieldset>
+            <label for="tweet_txt" class="visually-hidden">Enter your tweet</label>
```
jpetto: Unwrapped string.

bedrock/mozorg/templates/mozorg/contribute/tasks/taskview-base.html
```diff
@@ -0,0 +1,49 @@
...
+{% block page_css %}
+  {% stylesheet 'contribute-taskview' %}
+{% endblock %}
+
+{% block site_header_nav %}{% endblock %}
+{% block site_header_logo %}{% endblock %}
+
+{% block alt_header %}
+  <h1>{{ _('Mozilla Participation Tasks') }}</h1>
+  <a href="{% block more_tasks %}{{ url('mozorg.contribute.signup') }}{% endblock %}" class="back">{{ _('I want to pick a different task.' )}}</a>
```
jpetto: Should this `more_tasks` URL have `?v=1` appended? Currently on all v1 tasks, clicking "I want to pick a different task" takes you to the current/existing signup page.

bedrock/mozorg/templates/mozorg/contribute/tasks/v2/firefox-mobile.html
```diff
@@ -0,0 +1,53 @@
...
+      </div>
+    </li>
+    <li id="step_two" class="two">
+      <header>
+        <h3>{{ _('Tell your friends about Firefox on Twitter') }}</h3>
+      </header>
+      <div class="step-content">
+        <form name="tweet" id="signup-tweet" class="tweet-form" action="https://twitter.com/intent/tweet" method="get">
+          <fieldset>
+            <label for="tweet_txt" class="visually-hidden">Enter your tweet</label>
```
jpetto: Unwrapped string.

bedrock/mozorg/templates/mozorg/contribute/tasks/v2/firefox-mobile.html
```diff
@@ -0,0 +1,53 @@
...
+            <textarea name="text" id="tweet_txt" maxlength="140">{{ _('Firefox is available everywhere! Now you can get it for desktop, iOS or Android!') }} https://www.mozilla.org/firefox/products</textarea>
+            <p class="char-counter">Number of characters left: <span class="char-count"></span></p>
```
jpetto: Unwrapped string.

bedrock/mozorg/templates/mozorg/contribute/tasks/v2/whimsy.html
```diff
@@ -0,0 +1,46 @@
...
+{% block task_content %}
+  <ol class="task-steps">
+    <li id="step_one" class="one">
+      <header>
+        <h3>{{ _('Install Whimsy') }}</h3>
+        <p>{{ _('Whimsy is an add-on to help make your Firefox more enjoyable, like replacing the tiles on your new tab page with animated gifs!') }}</p>
+      </header>
+      <div class="step-content">
+        <img src="{{ static('img/contribute/signup/whimsy.png') }}" width="620" class="feature-img" alt="" />
+        <iframe src="https://addons.cdn.mozilla.net/static/html/mozorg/whimsy.html" class="action" frameborder="0" class="whimsy-install-container" data-step="one" data-complete="false" data-variation="2"></iframe>
```
jpetto: Two `class` attributes.

bedrock/mozorg/views.py
```diff
@@ -234,7 +234,40 @@ class ContributeIndex(l10n_utils.LangFilesMixin, TemplateView):
...
+        template = 'mozorg/contribute/tasks/v{0}/{1}.html'.format(variant, task)
+        return [template]
+
+
 def contribute_signup(request):
+    # for the new signup landing pages, a variant will be specified
+    # using the URL parameter v
+    variant = request.GET.get('v', '')
+
+    if variant:
```
jpetto: For safety's sake, suggest changing this to:

```
if variant in ['1', '2']:
```

Also suggest wrapping the rest in an `else` to clarify logic flow.

bedrock/mozorg/views.py
```diff
@@ -234,7 +234,40 @@ class ContributeIndex(l10n_utils.LangFilesMixin, TemplateView):
     template_name = 'mozorg/contribute/index.html'
 
 
+class ContributeSignupNew(TemplateView):
+
+    def get_template_names(self):
+        variant = self.request.GET.get('v', '')
```
jpetto: Could avoid duplicating this logic by adding `variant` as an instance variable of the `ContributeSignupNew` class and passing the value from `contribute_signup`.

```python
class ContributeSignupNew(TemplateView):
    variant = None

    def get_template_names(self):
        template = 'mozorg/contribute/signup-alternate.html'
        
        if self.variant == '2':
        # no changes past this point
```

```python
# in contribute_signup
return ContributeSignupNew.as_view()(request, variant)
```

bedrock/mozorg/views.py
```diff
@@ -234,7 +234,40 @@ class ContributeIndex(l10n_utils.LangFilesMixin, TemplateView):
...
+        if variant == '2':
+            template = 'mozorg/contribute/signup-alternate.html'
+
+        return [template]
+
+
+class ContributeTaskView(TemplateView):
+
+    def get_template_names(self):
+        variant = self.request.GET.get('v', '')
```
jpetto: Can default `variant` to `1` here: `variant = self.request.GET.get('v', '1')`

bedrock/mozorg/views.py
```diff
@@ -234,7 +234,40 @@ class ContributeIndex(l10n_utils.LangFilesMixin, TemplateView):
...
+        task = filter(None, self.request.path.split('/'))[-1]
+
+        # if the variant was not set, this is a v1 variant
+        if not variant:
+            variant = '1'
+
+        template = 'mozorg/contribute/tasks/v{0}/{1}.html'.format(variant, task)
```
jpetto: For safety's sake, should make sure both `variant` and `task` contain valid/expected values. Something like:

```python
class ContributeTaskView(TemplateView):
    tasks = [
        'build-firefox',
        'devtools-challenger',
        'firefox-mobile',
        'follow-mozilla',
        'joy-of-coding',
        'whimsy',
    ]

    def get_template_names(self):
        variant = self.request.GET.get('v', '1')
        task = filter(None, self.request.path.split('/'))[-1]

        if variant in ['1', '2'] and task in self.tasks:
            template = 'mozorg/contribute/tasks/v{0}/{1}.html'.format(variant, task)
        else:
            raise Http404

        return [template]
```

bedrock/mozorg/views.py
```diff
@@ -242,6 +275,11 @@ def contribute_signup(request):
     return view_class.as_view()(request)
 
 
+def contribute_signup_alternate(request):
```
jpetto: Is this used/needed?

media/css/mozorg/contribute/contribute-common.less
```diff
@@ -0,0 +1,225 @@
...
+            overflow: hidden;
+            cursor: pointer;
+            margin: 0;
+
+            &:after {
+                position: absolute;
+                top: 4px;
+                left: 0;
+                content: '';
+                background: transparent url('/media/img/contribute/signup/burger.svg') 0 0 no-repeat;
```
jpetto: :hamburger:

media/css/mozorg/contribute/contribute-common.less
```diff
@@ -0,0 +1,225 @@
...
+        vertical-align: bottom;
+
+        @media only screen and (min-width: @breakTablet) and (max-width: @breakDesktop) {
+            .span_narrow(2);
+        }
+    }
+
+    a {
+        border-radius: 3px;
+        box-sizing: border-box;
```
jpetto: Use the `.border-box`mixin?

media/css/mozorg/contribute/contribute-common.less
```diff
@@ -0,0 +1,225 @@
...
+        position: absolute;
+        bottom: 0;
+        left: 0;
+        .transition(all .15s ease-in-out);
+
+        &:hover,
+        &:focus,
+        &:active {
+            background-color: lighten(@mozillaRed, 10%);
+            background-color: rgba(255, 255, 255, .1);
```
jpetto: Remove one of these `background-color`'s?

media/css/mozorg/contribute/taskview.less
```diff
@@ -0,0 +1,444 @@
...
+            margin-left: 45px;
+        }
+    }
+
+    .one:before {
+        content: '1';
+    }
+
+    .two:before {
+        content: '2';
```
jpetto: Just a note that if/when we do localize, will have to figure out a different way to display these numbers.

jpetto:
Looks like `media/img/contribute/signup/firefox-mobile-high-res.png` could use compressing.

Also, the task view pages could use a little UI improvement. The header copy is centered, but the task details and image are left aligned (at least at some viewport widths). Should probably center for all viewports.

media/js/mozorg/contribute/signup.js
```diff
@@ -0,0 +1,21 @@
+$(function() {
```
jpetto: Does this JS need to be in a document ready wrapper?

media/js/mozorg/contribute/signup.js
```diff
@@ -0,0 +1,21 @@
...
+    var $tooltipTrigger = $('.tooltip-trigger');
+    var $currentItem = {};
+    var $tooltip = {};
```
jpetto: Just curious why these are defaulted to empty objects?

media/js/mozorg/contribute/taskview.js
```diff
@@ -0,0 +1,273 @@
...
+            }
+        });
+    }
+
+    /**
+     * Handles blur and focus events on the main window, and completes
+     * the task step once the main window receives focus.
+     */
+    function handleFocusChange($step) {
+        var taskCompleted = false;
```
jpetto: Wait, wont this make the below `if` statement pass every time?

media/js/mozorg/contribute/taskview.js
```diff
@@ -0,0 +1,273 @@
...
+    'use strict';
+
+    var $taskSteps = $('.task-steps');
+    var $stepOne = $('#step_one');
+    var $thankYou = $('#thankyou');
+    var $getFirefox = $('.get-firefox');
+    var $actionButton = $('.action');
+    var $downloadButton = $('.download-link');
+
+    var $followButton = $('.follow-mozilla')
```
jpetto: Nit - missing semicolon.

media/js/mozorg/contribute/taskview.js
```diff
@@ -0,0 +1,273 @@
...
+        var maxLength = 140;
+        var tweetLength = $tweetField.val().length;
+
+        // get and show the initial number of characters
+        $charCount.text(maxLength - tweetLength);
+
+        $tweetField.on('keyup', function() {
+            tweetLength = $tweetField.val().length;
+            $charCount.text(maxLength - tweetLength);
+        })
```
jpetto: Nit - missing semicolon.

media/js/mozorg/contribute/taskview.js
```diff
@@ -0,0 +1,273 @@
+$(function() {
```
jpetto: Same as above - any reason this JS needs to be in a document ready wrapper?

media/js/mozorg/contribute/taskview.js
```diff
@@ -0,0 +1,273 @@
...
+        $thankYou[0].scrollIntoView();
+        $thankYou.focus();
+    }
+
+    /**
+     * Waits for the initial tab/window to become visible, and then
+     * proceeds to complete the relevant task step.
+     */
+    function handleVisibilityChange($step) {
+        $(document).on('visibilitychange.taskview', function() {
```
alexgibson: For browsers that don't support the `visibilitychange` event, this listener will never get unbound again since it won't fire. To avoid a build up of duplicate event listeners would it be better to call `off` first before calling `on`?

bedrock/mozorg/templates/mozorg/contribute/signup-alternate.html
schalkneethling: Yup, it was bothering me as well and Alex also mentioned it so, I am going to change this so the entire thing is clickable.

bedrock/mozorg/templates/mozorg/contribute/signup-alternate.html
schalkneethling: Yeah :blush: but, I read that this is apparently the preferred when an anchor links nowhere. I use to use just `#` or perhaps `#tooltip` but, I was going back and forward on this, and I reckon in the end, just using a button element will be best.

bedrock/mozorg/templates/mozorg/contribute/tasks/taskview-base.html
schalkneethling: Yup, that was a pending update. Good catch

bedrock/mozorg/templates/mozorg/contribute/tasks/v1/follow-mozilla.html
schalkneethling: Holly, your thoughts here?

bedrock/mozorg/templates/mozorg/contribute/tasks/v1/joy-of-coding.html
schalkneethling: Sounds like a good idea to me. I was not aware of that, no do I know how to get something uploaded there so, your assistance here will be much appreciated. Thanks

bedrock/mozorg/views.py
schalkneethling: Nice! Thank.

bedrock/mozorg/views.py
schalkneethling: Yup, that did concern me. Nice solutions, thanks @jpetto

bedrock/mozorg/views.py
schalkneethling: Nope, not anymore

media/css/mozorg/contribute/contribute-common.less
schalkneethling: Hmmm, this came from the old contribute2015 less file, wonder if that was done on purpose for some reason. Maybe the first is a fallback for IE?

media/css/mozorg/contribute/taskview.less
schalkneethling: Indeed

bedrock/mozorg/templates/mozorg/contribute/tasks/v1/follow-mozilla.html
habber: The rest of the thank you messages are pretty brief  How about "Thanks for following!"

Also, in the copy below the headline it mentions Firefox and in the second sentence which is no longer accurate for this task. Can you change the second sentence of that block of copy to "Follow us and learn how you can pitch in!"

media/js/mozorg/contribute/signup.js
schalkneethling: I guess not reallllllly but, perhaps safer to leave as is, else line two below might get an empty array if those elements does not exist in the DOM yet, right?

media/js/mozorg/contribute/signup.js
schalkneethling: Show intent of use, nothing else.

alexgibson:
Just a reminder that when this experiment goes live we should skip the functional tests in `tests/functional/contribute/test_signup.py` (I'm assuming this is the only contribute page this PR touches?).

You can skip the required tests using the following marker:

`@pytest.mark.skipif(reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1225170')`

media/js/mozorg/contribute/taskview.js
schalkneethling: Hmmm yes, so IE9 and lower does not support this so yes agreed. The thing is then, without having tested, that means that if someone in IE(unlikely but possible) completes any task that relies on this, the will not see the completion step so, perhaps there needs to be a fallback here for those browsers anyhow.

Will think about this some. Thx @alexgibson

media/js/mozorg/contribute/taskview.js
schalkneethling: That is very true. Wonder how this never got picked up. Anyhow, fixing.

bedrock/mozorg/templates/mozorg/contribute/tasks/v1/follow-mozilla.html
schalkneethling: Will do, thanks Holly

media/js/mozorg/contribute/taskview.js
alexgibson: Fwiw, I believe many browsers also used to require a vendor-prefix for this event (`webkitVisibilitychange` could still be out there). I wouldn't worry too much about a fallback perhaps (it's not like the page appears broken if someone doesn't see the message), but adding the Webkit prefix could still be worth doing?

media/js/mozorg/contribute/signup.js
jpetto: As long as the JS is below/after all markup (just above the `</body>` tag), you shouldn't have to worry about it. The DOM will be ready. It does feel uneasy at first, but I've not once seen a problem. #trustfalls

schalkneethling: :)

bedrock/mozorg/templates/mozorg/contribute/tasks/v1/whimsy.html
```diff
@@ -0,0 +1,36 @@
...
+{% block task_content %}
+  <ol class="task-steps">
+    <li id="step_one" class="one">
+      <header>
+        <h3>{{ _('Install Whimsy') }}</h3>
+        <p>{{ _('Whimsy is an add-on to help make your Firefox more enjoyable, like replacing the tiles on your new tab page with animated gifs!') }}</p>
+      </header>
+      <div class="step-content">
+        <img src="{{ static('img/contribute/signup/whimsy.png') }}" width="620" class="feature-img" alt="" />
+        <iframe src="https://addons.cdn.mozilla.net/static/html/mozorg/whimsy.html" frameborder="0" class="action whimsy-install-container" data-step="one" data-complete="false" data-variation="2"></iframe>
```
schalkneethling: Hey @jpetto so, I have seen locally, and have confirmed on demo4, that this install is getting blocked. Is it hard wired to only allow install from www.mozilla.org or, is there something else that needs to be done? Thanks!

jpetto: I'm *pretty* sure that the install is blocked on all domains other than mozilla.org. I'll dive into the old bugs where we did this previously to confirm.

schalkneethling: Thanks!

jpetto: Hm, after seeing the error for myself, I'm now worried it's due to the recently changed amo CDN url. This might be a real problem. Asking in #amo for assistance.

media/css/mozorg/contribute/taskview.less
craigcook: Aren't Arabic numerals universal? In any other case text in CSS is a problem but for just numerals it may be fine.

jpetto: I think I remember @flodolo mentioning there are some locales where Arabic numerals don't work...

schalkneethling:
@alexgibson re issuecomment-176677401 - When accessing the /contribute/signup/ page without any parameters, it will still load the current signup page, do we then still need to disable the tests?

alexgibson:
If the plan is to use Optimizely to start switching the URL on prod, then the tests still need to be skipped yes.

schalkneethling:
Thanks @alexgibson just wanted to make sure before skipping them. So, am I correct that we should be skipping all the tests in the referenced file? Seems like it to me. The reason I ask is, is there a way to say skip all tests in this file or, do I need to mark each one?

Also, there is one that will already be skipped due to another bug[https://github.com/mozilla/bedrock/blob/master/tests/functional/contribute/test_signup.py#L42], should I have both or, can I just replace the current one with the new one? I am thinking the latter, just want to double check.

alexgibson:
Just skip the individual tests using the marker I posted above should do the trick. No need to skip again the one that's already being skipped (just leave that marker as-is, as it references a different bug) :+1:

media/css/mozorg/contribute/contribute-common.less
craigcook: The first one will output a hex value so that'll be the fallback for old browsers that don't support rgba.

jpetto:
Not strictly a part of this PR, but would be good to update the [existing `signup.html` page](https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templates/mozorg/contribute/signup.html#L15) to use our Optimizely include.

media/css/mozorg/contribute/contribute-common.less
schalkneethling: Yup, that is pretty much what I believed to be the case.

schalkneethling:
@jpetto will do that.

bedrock/mozorg/templates/mozorg/contribute/tasks/v2/devtools-challenger.html
```diff
@@ -0,0 +1,63 @@
...
+    <li id="step_two" class="two">
+      <header>
+        <h3>{{ _('Tell your friends on Twitter to try Developer Tools, too!') }}</h3>
+      </header>
+      <div class="step-content">
+        <form name="tweet" id="signup-tweet" class="tweet-form" action="https://twitter.com/intent/tweet" method="get">
+          <fieldset>
+            <label for="tweet_txt" class="visually-hidden">{{ _('Enter your tweet') }}</label>
+            <textarea name="text" id="tweet_txt" maxlength="140">{{ _('Have you tried the latest animation tools in Firefox? Try Dev Tools Challenger http://devtoolschallenger.com/') }}</textarea>
+            <p class="char-counter">Number of characters left: <span class="char-count"></span></p>
```
jpetto: Might as well wrap this string?

bedrock/mozorg/templates/mozorg/contribute/tasks/v2/follow-mozilla.html
```diff
@@ -0,0 +1,58 @@
...
+      <header>
+        <h3>{{ _('Tweet This') }}</h3>
+      </header>
+      <div class="step-content">
+        <img src="{{ static('img/contribute/signup/startmozilla.png') }}" width="620" class="feature-img" alt="" />
+        <form name="tweet" id="signup-tweet" class="tweet-form" action="https://twitter.com/intent/tweet" method="get">
+          <fieldset>
+            <label for="tweet_txt" class="visually-hidden">Enter your tweet</label>
+            <textarea name="text" id="tweet_txt" maxlength="140">Hi @startmozilla! I’m a {enter your main skill here: developer, designer, writer, etc.} who would like to contribute to Mozilla!</textarea>
+            <p class="char-counter">Number of characters left: <span class="char-count"></span></p>
```
jpetto: Wrap strings here as well?

media/js/mozorg/contribute/taskview.js
jpetto: Can we replace this document ready wrapper with our standard IIFE wrapper?

media/js/mozorg/contribute/signup.js
```diff
@@ -0,0 +1,36 @@
+var $tooltipTrigger = $('.tooltip-trigger');
```
jpetto: To keep from polluting the global namespace, can we wrap this code in an IIFE like we do elsewhere?

media/js/mozorg/contribute/taskview.js
```diff
@@ -0,0 +1,316 @@
...
+        }
+
+        // for Chrome 13 and lower
+        if (typeof document.webkitHidden !== 'undefined') {
+            return 'webkitvisibilitychange'
+        }
+
+        // for older versions of IE
+        if (typeof document.msHidden !== 'undefined') {
+            return 'msvisibilitychange'
```
jpetto: Nit - add semicolons at the end of the `return` statements above.

media/js/mozorg/contribute/taskview.js
```diff
@@ -0,0 +1,316 @@
...
+
+        if (!taskCompleted) {
+            // once our original window receives focus again, complete the task.
+            window.onfocus = function() {
+                completeStep($step);
+            };
+            // ensure this only happens once.
+            taskCompleted = true;
+        }
+    }
```
jpetto: The logic in this function still isn't quite right. As it is function scoped, `taskCompleted` will be falsy every time the function runs, which results in the `onfocus` handler being called each time the original window regains focus (e.g. when switching back and forth from the Twitter popup).

A better approach may be to use jQuery's `one` function to make sure the handler only fires once. Something like:

```
$(window).one('focus', function() {
    completeStep($step);
});
```

media/js/mozorg/contribute/taskview.js
```diff
@@ -0,0 +1,316 @@
...
+
+    /**
+     * Waits for the initial tab/window to become visible, and then
+     * proceeds to complete the relevant task step.
+     */
+    function handleVisibilityChange($step) {
+
+        var $document = $(document);
+        var visibilityChange = getVisibilityStateEventKeyword();
```
jpetto: Since this function *can* be called twice on one page (Whimsy V2), I'd suggest moving `$document` and `visibilityChange` up a level in scope so they doesn't get re-declared/initialized on each call to `handleVisibilityChange`.

jpetto:
A few more small suggestions, but looking really good.

schalkneethling:
@jpetto ok, I reckon we are good to go with this one. Thanks for the review!

bedrock/mozorg/templates/mozorg/contribute/tasks/v2/devtools-challenger.html
```diff
@@ -0,0 +1,63 @@
...
+    <li id="step_two" class="two">
+      <header>
+        <h3>{{ _('Tell your friends on Twitter to try Developer Tools, too!') }}</h3>
+      </header>
+      <div class="step-content">
+        <form name="tweet" id="signup-tweet" class="tweet-form" action="https://twitter.com/intent/tweet" method="get">
+          <fieldset>
+            <label for="tweet_txt" class="visually-hidden">{{ _('Enter your tweet') }}</label>
+            <textarea name="text" id="tweet_txt" maxlength="140">{{ _('Have you tried the latest animation tools in Firefox? Try Dev Tools Challenger http://devtoolschallenger.com/') }}</textarea>
+            <p class="char-counter">{{ _('Number of characters left:') }}' <span class="char-count"></span></p>
```
jpetto: An errant single quote here - guessing from text editor auto-complete.

media/js/mozorg/contribute/taskview.js
```diff
@@ -0,0 +1,312 @@
...
+        }
+    }
+
+    /**
+     * Handles completion of Follow Mozilla interaction steps.
+     */
+    function followMozilla(event) {
+        var $this = $(event.target);
+        var intentURL = $this.attr('href');
+        var screenName = intentURL.substr(intentURL.indexOf('=') + 1);
```
jpetto: Getting an error on this line when working step 1 of [V2 on this task](https://www-demo4.allizom.org/en-US/contribute/task/follow-mozilla/?v=2). Maybe a conditional needed around `$this.data('action')`?

jpetto:
Found 2 (hopefully) final little bugs. Also want to verify with cmore that filtering the Optimizely snippet on `LANG == 'en-US'` wont conflict with language filtering in the experiment itself.

PR #3816, opened 2016-01-27
Remove duplicated requirements

davehunt:
This should fix the issues in Web QA's Jenkins caused by certain requirements appearing in both `base.txt` and `test.txt`.

### DIFF
--- a/requirements/test.txt
+++ b/requirements/test.txt
@@ -4,11 +4,6 @@
 # sha256: ZdKqaLKOfTEjO7K6jrMc2kDkZx-KwtayQeNYyWUqdLk
 apipkg==1.4
 
-# sha256: gqJTJFrtmHXxQd2Yp0YlcrypPQZM8xF70z3kth7hjmg
-# sha256: LCZCVPbPzmTDvZ1IiFIICTpzuQlWRfYA3gonfOAeoOU
-# sha256: h9QBPQYl1HiaT1a415oE1c5tsRUrtl8dOXRPdwmjZrQ
-beautifulsoup4==4.4.1
-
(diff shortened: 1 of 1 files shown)


requirements/test.txt
```diff
@@ -4,11 +4,6 @@
 # sha256: ZdKqaLKOfTEjO7K6jrMc2kDkZx-KwtayQeNYyWUqdLk
 apipkg==1.4
 
-# sha256: gqJTJFrtmHXxQd2Yp0YlcrypPQZM8xF70z3kth7hjmg
-# sha256: LCZCVPbPzmTDvZ1IiFIICTpzuQlWRfYA3gonfOAeoOU
-# sha256: h9QBPQYl1HiaT1a415oE1c5tsRUrtl8dOXRPdwmjZrQ
-beautifulsoup4==4.4.1
```
jgmize: This is newer than the version in requirements/base.txt, and I'm uncertain why we need it in requirements/base.txt at all. Perhaps we should remove the version in requirements/base.txt instead?

jgmize: scratch that, we're using BeautifulSoup in bedrock/privacy/views.py.

jgmize: Given that the tests pass with the older version of BeautifulSoup in requirements/base.txt, and that's the version being used in production now, the safest thing to do is to merge this as-is and update the version in requirements/base.txt in a later PR.

jgmize:
Thanks @davehunt

requirements/test.txt
davehunt: I agree, I thought it was safer to make the tests adhere to the packages pinned for the application.

PR #3819, opened 2016-01-28
[fix bug 1242610] Update send to device widget to use Mozilla Location Services

alexgibson:
Note: because the new MSL API requires CORS, the geo lookup will no longer work in IE9 and below. In the case of this widget this should still be fine I think, as the form degrades gracefully to email only (the geo request is only responsible for updating messaging related to SMS).

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -307,7 +307,7 @@
     {% set ios_link = firefox_ios_url('mozorg') %}
   {% endif %}
   {% set android_link = settings.GOOGLE_PLAY_FIREFOX_LINK %}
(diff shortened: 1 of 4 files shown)


hannosch:
Looks good to me, feedback+ :)

schalkneethling:
Code looks good to me, just not sure how to test the US part of the form usage.

alexgibson:
@schalkneethling perhaps it might be best if you let someone in the U.S confirm this PR works?

schalkneethling:
@alexgibson makes sense.

jpetto:
Confirmed that SMS is showing when in the :us:.

alexgibson:
@schalkneethling @jpetto rebased, thanks

PR #3822, opened 2016-01-28
Add URL tests that follow redirects and allow these to be run from the pipeline

davehunt:
This patch introduces the tests from [test_urls.py](https://github.com/mozilla/mcom-tests/blob/master/tests/test_urls.py) in mcom-tests. It also opens up the possibility of following all redirect tests to verify the final response code, however when I tested this I got a lot of 404s, most of these are probably expected due to random URLs used to verify redirects.

I've also renamed 'functional' to 'integration' in the Docker and Jenkins files and made it so that these can also run the 'redirect' tests. Note that `bedrock_functional_tests` and `bedrock_functional_tests_sauce` will need to be modified (and likely renamed) as soon as this is merged. Let's coordinate when this is ready to land.

I have noticed that due to a unicode character in a test name there's a problem generating the HTML report. This needs to be addressed before this can land, possibly by releasing a new version of pytest-html, but it shouldn't block getting a review.

@alexgibson, @pmclanahan, @jgmize: r?


davehunt:
The issue mentioned appears to be one in pytest core. I've raised https://github.com/pytest-dev/pytest/issues/1351 but for the time being we can simply turn off the HTML report generation in Jenkins by removing the command line option.

alexgibson:
@davehunt looks like this needs a rebase

tests/redirects/base.py
```diff
@@ -46,6 +47,8 @@ def url_test(url, location=None, status_code=requests.codes.moved_permanently,
     :param req_kwargs: Extra arguments to pass to requests.get()
     :param resp_headers: Dict of headers expected in the response.
     :param query: Dict of expected query params in `location` URL.
+    :param allow_redirects: Boolean indicating whether redirects should be followed.
```
pmclanahan: I would suggest `follow_redirects` instead. `allow_redirects` just seems confusing to me in a test suite mainly for redirects.

tests/redirects/base.py
```diff
@@ -101,8 +107,10 @@ def assert_valid_url(url, location=None, status_code=requests.codes.moved_perman
     :param resp_headers: Dict of headers expected in the response.
     :param base_url: Base URL for the site to test.
     :param query: Dict of expected query params in `location` URL.
+    :param allow_redirects: Boolean indicating whether redirects should be followed.
+    :param final_status_code: Expected status code after following any redirects.
     """
-    kwargs = {'allow_redirects': False}
+    kwargs = {'allow_redirects': allow_redirects}
```
pmclanahan: Ah. `allow_redirects` is from the `requests` lib. Still, in our test suite context I like `follow_redirects` better. Not a hard requirement though.

tests/redirects/map_globalconf.py
```diff
@@ -194,7 +194,7 @@
     # bug 860865, 1101220
     url_test('/firefox/all-{beta,rc}{/,.html}', '/firefox/beta/all/'),
     url_test('/firefox/all-aurora{/,.html}', '/firefox/developer/all/'),
-    url_test('/firefox/aurora/{all,notes,system-requirements}/'
+    url_test('/firefox/aurora/{all,notes,system-requirements}/',
```
pmclanahan: good catch

tests/redirects/test_urls.py
```diff
@@ -70,3 +42,46 @@ def test_404_url(base_url):
...
+    '/firefox/os/notes/1.1/',
+    '/firefox/partners/',
+    '/firefox/releases/',
+    '/firefox/speed/',
+    '/firefox/sync/',
+    '/firefox/tiles/',
+    '/firefox/unsupported-systems/',
+    '/firefox/unsupported/EOL/',
+    # Legacy URLs (Bug 1110927)
+    'firefox/panorama/',
```
pmclanahan: The above start with a slash, and the below do not. Only one of these is probably correct.

davehunt:
@alexgibson, @pmclanahan: Rebased and addressed review comments.

alexgibson:
Adding back a do-not-merge label as we need to sync some Jenkins config changes when this merges.

davehunt:
@alexgibson @pmclanahan @jgmize Travis is happy now. Feel free to review, and we can sync up before this is merged next week to make sure I can do the necessary Jenkins updates.

alexgibson:
This looks good to me Travis is now passing :smile:, although I'll defer to either @pmclanahan or @jgmize in terms of technical review. Thanks @davehunt!

jgmize:
r+ on the code, thanks @davehunt!

Will merge once Jenkins changes have been made.

docker/jenkins/run_integration_tests.sh
```diff
@@ -1,7 +1,21 @@
...
+docker build -t bedrock_integration_tests:${GIT_COMMIT} --pull=true .
+
+if [ -z "${BASE_URL}" ]; then
+  # start bedrock
+  docker run -d \
+    --name bedrock-code-${BUILD_NUMBER} \
+    -e ALLOWED_HOSTS="*" \
+    -e SECRET_KEY=foo \
+    -e DEBUG=False \
+    -e DATABASE_URL=sqlite:////tmp/temp.db \
```
pmclanahan: Just a note: if you don't need a DB that will stick around you can just set `DATABASE_URL=sqlite://` which will cause it to use a purely in-memory database and never create a file.

pmclanahan:
r+ from me as well

PR #3829, opened 2016-02-02
[fix bug 1240980] Update Firefox Hello product page

alexgibson:


alexgibson:
CC @flodolo - strings are final but design is still in progress. I will push this to a demo server and link in the bug.

flodolo:
The old page is going to be completely dismissed then? Or should we fallback while the new one gets translated?

alexgibson:
> The old page is going to be completely dismissed then? Or should we fallback while the new one gets translated?

The product itself it totally changing, so it makes no sense to keep the old one around. we'll be using a waffle flag and switching the new template on for when the new version is released.

bedrock/firefox/templates/firefox/hello/index-2016.html
```diff
@@ -0,0 +1,134 @@
...
+  {% call fxfamilynav('desktop', 'hello') %}
+    <a class="try-hello-button" href="https://support.mozilla.org">{{ _('Try Hello now') }}</a>
+    <a href="{{ url('firefox.new') }}">{{ _('Download now') }}</a>
+  {% endcall %}
+{% endblock %}
+
+{% block content %}
+  <header id="try-hello">
+    <div class="content">
+      <h1>{{ _('Plan. Shop. Decide. Together.') }}</h1>
```
alexgibson: Ping @flodolo - looks like we may want the word `Together` here to be visually formatted. I still need to confirm as the design for this section is still not quite final, but I should have predicted this in hindsight :/

How are we for locales having already translated this string? If the number is a lot, we could either change the string for en-US only, or use an l10n tag? Sorry for the trouble :/

flodolo: Can you define "visually formatted"? We already have [13 locales](https://l10n.mozilla-community.org/~pascalc/langchecker/?locale=all&website=0&file=firefox/hello-2016.lang), and I expect a few more after the weekend.

I honestly don't want to start a brand new page with a l10n tag already in it, but it's also extremely confusing to change a string only 2 days after we started localizing it :-\

alexgibson: > Can you define "visually formatted"?

They might want to make it bolder than the other words

> I honestly don't want to start a brand new page with a l10n tag already in it, but it's also extremely confusing to change a string only 2 days after we started localizing it :-\

If this is an issue, we can do this for en-US only if that works best. Thanks

flodolo: If it's not a problem I would keep this en-US only, especially if we don't know when the design will be clarified (the current one I've seen doesn't hint anything like that).

alexgibson: Ok will do, thanks @flodolo :+1:

alexgibson:
Ok, going to open this one up for review.

Please note: The "Try Hello" button is currently broken in Nightly due to a recent regression in the Hello add-on. This is due to be fixed in [bug 1247255](https://bugzilla.mozilla.org/show_bug.cgi?id=1247255) and is beyond control of the web page (this also effects the current hello page that is already in production).

bedrock/firefox/views.py
```diff
@@ -443,7 +443,10 @@ def hello(request):
         'zh-TW': 'https://videos.cdn.mozilla.net/uploads/FirefoxHello/firefoxhello_intro_chinese'
     }
 
-    return l10n_utils.render(request, 'firefox/hello/index.html',
+    if (waffle.switch_is_active('firefox-hello-2016')):
```
alexgibson: Ping @jgmize or @pmclanahan - new waffle switch here `firefox-hello-2016`. We don't need to enable it until March 8, but it would be great if we could have it set up & ready before then (else we could likely forget to do it).

pmclanahan: Not sure what you mean. We could set a calendar appointment. We can't (or more correctly I think shouldn't) schedule waffle switches right now just due to how complex it now is with 3 separate deployments. We can enable the switch whenever we want though.

alexgibson: I didn't mean schedule it, simply create the switch. Thought this was a thing that needed doing, but if I'm incorrect then carry on :smile:

pmclanahan: Ah! Not really necessary, yeah. If a switch is not in the database that is the same as `OFF`. So when we're ready to turn it on we'll create it as `ON` and it'll be good. We technically can create one and have it be `OFF`, but there's not much point.

media/css/firefox/hello/index-2016.less
```diff
@@ -0,0 +1,821 @@
...
+
+.js {
+    .fx-up-to-date,
+    .fx-out-of-date,
+    .non-fx,
+    .mobile,
+    .download-button,
+    .hello-wordmark,
+    .nav-download-button,
+    .try-hello-button, {
```
schalkneethling: Nit: I'm guessing the preprocessor is fine with this but, the last item should not have a `,`

media/css/firefox/hello/index-2016.less
```diff
@@ -0,0 +1,821 @@
...
+
+    .download-button,
+    .hello-wordmark, {
+        display: inline-block;
+    }
+
+    .fx-up-to-date,
+    .non-fx,
+    .mobile,
+    .try-hello-button, {
```
schalkneethling: Nit: Here as well

schalkneethling:
Seems like the "Need Help..." text should be centered here? It is when viewing at @breakTablet and smaller but not at larger viewports.

screen shot 2016-02-12 at 13 26 58

alexgibson:
> Seems like the "Need Help..." text should be centered here? It is when viewing at @breakTablet and smaller but not at larger viewports.

This is as per the design for it to be left aligned at desktop viewport.

schalkneethling:
Ok, I guess it is lining up with the right edge of the button. For systems that does not meet the requirements to run Fx, the message is almost unreadable(black on dark blue). Perhaps this can be changed to white?

screen shot 2016-02-12 at 13 30 46

schalkneethling:
This looks great! r+wc

alexgibson:
Thanks @schalkneethling - updated and squashed!

schalkneethling:
:fries:

PR #3830, opened 2016-02-03
[fix bug 1244888] Add FxA / firstrun setup info to the bedrock docs

alexgibson:


docs/firefox-accounts.rst
```diff
@@ -0,0 +1,86 @@
...
+.. _firefox-accounts:
+
+============================
+Firefox Accounts Signup Form
+============================
+
+Introduction
+------------
+
+Certain bedrock pages such as ``/firstrun`` feature a Firefox Accounts signup
```
jpetto: Only small nit would be qualifying this URL for new contributors - `/firefox/firstrun/`. Or possibly even changing it to `/firefox/accounts/`, as that one is even easier to test.

jpetto:
Looks great. One small suggestion, otherwise put it in the pizza. :+1:

alexgibson:
Thanks @jpetto updated

PR #3834, opened 2016-02-04
Add functional tests for firefox/all page

alexgibson:


alexgibson:
@davehunt r?

tests/pages/firefox/all.py
```diff
@@ -0,0 +1,37 @@
...
+                   self._build_rows_locator) if el.is_displayed()])
+
+    @property
+    def english_builds(self):
+        return len([el for el in self.find_elements(self._build_rows_locator)
+                   if 'english' in el.get_attribute('data-search')])
+
+    @property
+    def only_english_builds_are_displayed(self):
+        return self.english_builds == self.displayed_builds
```
alexgibson: I'm a bit wary that this is doing a lot of DOM querying. It might be a nicer idea to only get the `english_builds` length once, instead of needing a repeated query. Not sure about the best approach here to do this. Thoughts?

tests/functional/firefox/test_all.py
```diff
@@ -0,0 +1,26 @@
...
+
+from pages.firefox.all import FirefoxAllPage
+
+
+@pytest.mark.smoke
+@pytest.mark.nondestructive
+def test_search_language(base_url, selenium):
+    page = FirefoxAllPage(base_url, selenium).open()
+    page.type_search('english')
+    page.click_search()
```
davehunt: Does the click actually do anything? As far as I can tell simply typing in the term performs the search. Either way, I would create a `search_for` method in the page object that abstracts how it's actually done. The test shouldn't care.

alexgibson: The click could still be needed in some browsers (which don't support the `input` event). Good suggestion on the abstraction though :+1:

tests/functional/firefox/test_all.py
```diff
@@ -0,0 +1,26 @@
...
+    assert page.only_english_builds_are_displayed
```
davehunt: I'd rather have something like:

```python
for build in page.displayed_builds:
    assert 'english' in build.language.lower()
```

This would require a `build` region, which I think makes sense anyway.

davehunt: I'd also put the search term into a variable.

alexgibson:
@davehunt - thanks for the suggestions, updated

tests/pages/firefox/all.py
```diff
@@ -0,0 +1,42 @@
...
+class FirefoxAllPage(FirefoxBasePage):
+
+    _url = '{base_url}/{locale}/firefox/all'
+
+    _search_input_locator = (By.ID, 'language-search-q')
+    _submit_button_locator = (By.CSS_SELECTOR, '#language-search button')
+    _build_rows_locator = (By.CSS_SELECTOR, '#builds table > tbody > tr')
+
+    @property
+    def builds(self):
```
davehunt: I would make this private, as we don't expect tests to directly access builds that may not be displayed.

tests/pages/firefox/all.py
```diff
@@ -0,0 +1,42 @@
...
+        return [Build(self, root=el) for el in
+                self.find_elements(self._build_rows_locator)]
+
+    @property
+    def displayed_builds(self):
+        return [Build(self, root=el) for el in
+                self.find_elements(self._build_rows_locator) if el.is_displayed()]
```
davehunt: Add an `is_displayed` to `Build` and then you can simplify this to:

```python
    return [b for b in self._builds if b.is_displayed]
```

tests/pages/firefox/all.py
```diff
@@ -0,0 +1,42 @@
...
+
+    def builds_filtered_by_language(self, value):
+        languages = [el for el in self.builds if value in el.language]
+        return len(languages) == len(self.displayed_builds)
+
+    def search_for(self, value):
+        self.find_element(self._search_input_locator).send_keys(value)
+        self.find_element(self._submit_button_locator).click()
+        self.wait.until(lambda s: self.builds_filtered_by_language(value))
```
davehunt: This is essentially doing our assert before the test gets a chance to do it. Is there any other way to indicate that the search is complete? Have you seen that this wait is necessary?

If there's no other way and this is needed, then I would avoid the additional method that's only used here by moving the condition inline.

```python
expected_builds = [b for b in self._builds if value in b.language.lower()]
self.wait.until(lambda s: len(expected_builds) == len(self.displayed_builds))
```

alexgibson: Yes I think this wait is likely necessary, since updating the DOM is not instantaneous (and the operation also gets slower the more rows in the table exist). I'll move the condition inline.

tests/pages/firefox/all.py
alexgibson: ah, much nicer :+1:

alexgibson:
@davehunt updated, thanks!

alexgibson:
@davehunt - friendly ping reminder :smile:

davehunt:
Sorry, been sick. r+

alexgibson:
No problem, thanks again @davehunt

PR #3835, opened 2016-02-05
Fix token server uri for testing

vladikoff:


alexgibson:
Thanks @vladikoff! I tested this on our demo environment, but this still does not seem to work :(

alexgibson:
Not sure if this is expected or not, but if I search for `services.sync.tokenServerURI` in `about:config` the pref does not seem to be listed?

vladikoff:
Works for me with our fxa-local-dev configuration:

alexgibson:
Hmm... will give this another try

vladikoff:
> Not sure if this is expected or not, but if I search for services.sync.tokenServerURI in about:config the pref does not seem to be listed?

Here's what I got:

vladikoff:
@alexgibson try adding the identity one as well and make sure it is modified like the screenshot above https://github.com/mozilla/fxa-local-dev/blob/master/_scripts/profile.js#L105

alexgibson:
This still doesn't seem to be working quite right, even with a new profile. There must be another difference between our config and the one you link to. I'll try and see if I can figure it out.

alexgibson:
Together with #3835 we now have the issue fixed. Thanks @vladikoff :beer:

PR #3837, opened 2016-02-08
[fix bug 1245148] Split /firefox/new/ into 2 templates.

jpetto:


jpetto:
@alexgibson - Still a bit of a WIP, but would love a high-ish level sanity check. Any scenarios I've missed?

One concern I haven't yet fully tested is if the connection to the installer file takes longer than the redirect timeout. If this is indeed an issue, I may need to lose the direct download on scene 1 and fall back to relying on scene 2 to begin the download.

bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -0,0 +1,144 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/base-resp.html" %}
```
alexgibson: Could we make the scene1 and scene 2 templates more dry if they extended their own base template perhaps? Seems like a lot of the page meta stuff could probably be shared?

bedrock/firefox/views.py
```diff
@@ -506,4 +506,19 @@ def new(request):
...
+    template = 'firefox/new.html'
+    scene = request.GET.get('scene', None)
+    version = request.GET.get('v', None)
+    dlstarted = request.GET.get('dlstarted', None)
+
+    # separate scenes version of /new
+    if version == '2':
+        if scene == '2':
+            template = 'firefox/new/scene2.html'
+        # REVIEW: Should this explicitly check for a None value instead
```
alexgibson: I don't have a strong preference here, although I guess I would perhaps favor always showing the download button (scene1) in the event of an incorrectly typed link?

bedrock/settings/static_media.py
```diff
@@ -394,6 +394,26 @@
         ),
         'output_filename': 'css/firefox_new-bundle.css',
     },
+    'firefox_new_scene1': {
+        'source_filenames': (
+            'css/sandstone/sandstone-resp.less',
```
alexgibson: Would it be worth splitting these out into common CSS bundles? Could we potentially get faster page loads on scene 2 if the common bundle was already cached from scene1?

media/js/firefox/new/scene1.js
```diff
@@ -0,0 +1,119 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// set up namespace
+if (typeof Mozilla == 'undefined') {
+    var Mozilla = {};
+}
+
+Mozilla.Firefox = window.Mozilla.Firefox || {};
+
+;(function($, dataLaye, MFNew) {
```
alexgibson: Typo `dataLaye`

media/js/firefox/new/scene2.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+Mozilla.Firefox = window.Mozilla.Firefox || {};
+
+;(function($, dataLayer, MFNew) {
+    'use strict';
+
+    var $directDownloadLink = $('#direct-download-link');
+    var $stage = $('#stage');
+    var downloadStarted = Number(MFNew.params.get('dlstarted')) === 1;
```
alexgibson: Curious any reason you favor `Number` here over `parseInt` (which i guess we use more commonly)?

media/js/firefox/new/scene2.js
```diff
@@ -0,0 +1,72 @@
...
+
+    //$(document).ready(function() {
```
alexgibson: Is this no longer needed? :grimacing:

alexgibson:
> One concern I haven't yet fully tested is if the connection to the installer file takes longer than the redirect timeout. If this is indeed an issue, I may need to lose the direct download on scene 1 and fall back to relying on scene 2 to begin the download.

Testing in Chrome (esp on a fresh browser cache) I sometimes don't get a download when clicking the link before the page goes to scene 2 :cry:

It seems like triggering the download on scene 2 might be necessary here (although this does makes scene 2 even more important to be fast and robust)

bedrock/firefox/templates/firefox/new/scene1.html
jpetto: Only reason I haven't done this yet is I'm not sure all the meta stuff *should* be shared. I want to run the new pages by both garethc and habber to see if the scenes should have different titles, descriptions, etc.

Based on sprint timing, envisioning this as a task for next sprint - either customize meta info per scene/page, or create shared base template.

bedrock/firefox/views.py
jpetto: That was my thought as well. Will leave as a simple `else`.

bedrock/settings/static_media.py
jpetto: Hm, I think I agree that 1 additional HTTP request on scene 1 (or when going directly to scene 2) is worth the cache savings on a scene 1 > scene 2 user flow.

Will create a `firefox_new_common` bundle.

media/js/firefox/new/scene1.js
jpetto: d'oh

media/js/firefox/new/scene2.js
jpetto: Uh, `Number` doesn't require a radix? Not really a reason. Will switch to `parseInt` for consistency.

media/js/firefox/new/scene2.js
jpetto: Maybe not. Still experimenting with it removed.

jpetto:
Since scene 2 needs the logic to auto-trigger the download anyway, I may just fall back there. Since scene 1 (or any other page displaying **the** download button) handles the logic for displaying a proper link (for now), I don't think it's so bad to offload downloading to scene 2.

That being said, I'm still going to see if I can make scene 1 start the download (maybe a popup?). Doing it there is slightly faster for the user, and would avoid the half-downloaded image problem.

bedrock/firefox/templates/firefox/new/scene1.html
alexgibson: Ok sounds good to me :+1:

media/js/firefox/new/scene2.js
alexgibson: Up to you entirely, was just curious if you preferred one over the other.

jpetto:
Updates include:

- Common Less/JS bundles to improve cache performance going from scene 1 > scene 2
- Move download handling to scene 2
- Prune Less/JS files

Tested in IE 6 - 9 and modern browsers and all looked/functioned as expected.

jpetto:
@alexgibson - Should now be ready for real review.

media/js/firefox/new/scene1.js
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    //GA Custom Dimension in Pageview
+    window.dataLayer.push({
+        'event': 'set-state',
+        'state': state
+    });
+
+    // point download links to new version of scene 2
+    $('.download-link').attr('href', scene2URL);
```
alexgibson: This is over-writing all download button URL's, including iOS and Android buttons that should link directly to the app stores. Do we need to re-introduce the concept of platforms that are eligible for scene 2?

bedrock/firefox/templates/firefox/new/scene2.html
```diff
@@ -0,0 +1,105 @@
...
+    <div class="download-button-wrapper desktop" id="download-button-wrapper-desktop">
+      {{ download_firefox(force_direct=true, simple=true, dom_id='primary-download-button') }}
+    </div>
+
+    <p class="help">
+      {{_('Your download should begin automatically. If not, <a id="%s" href="">click here</a>. It could take a few minutes, but it’s worth the wait.')|format('direct-download-link')}}
+      <a rel="external" href="https://support.mozilla.org/products/firefox?utm_source=mozilla.org&amp;utm_medium=referral&amp;utm_campaign=need-help-link">{{_('Need help?')}}</a>
+    </p>
+
+    <div id="thankyou" tabindex="0">
```
alexgibson: Nit: do we need this `tabindex` still?

bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -0,0 +1,146 @@
...
+  </div>
+  <div class="android-latest-links-wrapper latest-links-wrapper">
+    <ul>
+      <li><a href="{{ url('firefox.android.index') }}">{{_('Learn more about Firefox for Mobile')}}</a></li>
+    </ul>
+  </div>
+  <div class="fxos-latest-links-wrapper latest-links-wrapper">
+    <ul>
+      <li><a href="{{ url('firefox.os.index') }}">{{_('Learn about Firefox OS')}}</a></li>
+      <li><a href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}">{{_('Discover all the ways to take Firefox on the go')}}</a></li>
```
alexgibson: I wonder if there is a more relevant link destination for this now? Not necessarily need be part of this PR, but the Firefox products page may be a better bet perhaps.

media/js/firefox/new/scene2.js
```diff
@@ -0,0 +1,70 @@
...
+    // hidden, but could be clicked by screen reader?)
+    $stage.on('click', '#direct-download-link, .download-link', function(e) {
+        e.preventDefault();
+
+        var url = $(e.currentTarget).attr('href');
+
+        // An iframe cannot be used here to trigger the download because
+        // it will be blocked by Chrome if the download link redirects
+        // to a HTTP URI and we are on HTTPS.
+        function trackAndRedirect(url, virtualUrl) {
```
alexgibson: Does keeping this as a function still serve a purpose now that it's only called once within this block?

media/js/firefox/new/scene2.js
```diff
@@ -0,0 +1,70 @@
...
+
+;(function($, dataLayer, MFNew) {
+    'use strict';
+
+    var $directDownloadLink = $('#direct-download-link');
+    var $stage = $('#stage');
+    var downloadStarted = parseInt(MFNew.params.get('dlstarted'), 10) === 1;
+
+    // pixel remains in perpetuity (https://bugzilla.mozilla.org/show_bug.cgi?id=1222945#c2)
+    if (!window._dntEnabled()){
```
alexgibson: As this is probably the least important bit of code on the page, is it worth moving it further down to just before the window `'load` conditional? Could probably be contained/called as a function seeing as it creates some variables that nothing else needs reference to?

alexgibson:
Looking good - I think the biggest remaining thing is how to handle button clicks that don't need scene2, and should just download directly.

Do you think this is worth putting up on a demo server for some more real-world testing? (I never really trust running localhost for changes to this page especially, although with this change that could get easier in the future :smile:).

bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -0,0 +1,146 @@
...
+    </ul>
+  </div>
+
+  <section id="download-buttons">
+    {# handles desktop users (windows, os x, linux) #}
+    <div class="download-button-wrapper desktop" id="download-button-wrapper-desktop">
+      {{ download_firefox(simple=true, dom_id='primary-download-button') }}
```
alexgibson: Note: changing the Id of the primary download button will cause one of our sanity tests to fail here: https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/test_new.py#L17

Unless there is a reason to change it, suggest just using the default ID that's already in place? This would then mean we don't have to skip that particular test.

alexgibson:
While this test is running you'll need to skip the following two tests:

FAIL tests/functional/firefox/test_new.py::test_click_download_button
FAIL tests/functional/firefox/test_new.py::test_directly_load_thank_you

bedrock/firefox/templates/firefox/new/scene1.html
jpetto: At one point I was using the newly specified ID, but no longer. Will remove so we can run the test.

media/js/firefox/new/scene2.js
jpetto: Uhm, nope. No need for the function - will remove.

bedrock/firefox/templates/firefox/new/scene2.html
jpetto: Yeah, probably not. Will remove.

media/js/firefox/new/scene1.js
jpetto: Tunnel vision! I have a fix lined up to only replace the URL if it already includes `scene=2`. Don't think we need any logic beyond what's already in the download button.

alexgibson: :+1: :tram:

bedrock/firefox/templates/firefox/new/scene1.html
jpetto: Hmm, this is a good point. Since only a very small audience will see this link, will just go ahead and change to `firefox.products`.

media/js/firefox/new/scene2.js
jpetto: I endorse both of these suggestions. Will make it so.

jpetto:
@alexgibson - Think I've addressed all comments. Another look through when you have time?

jpetto:
Also, [branch is up on demo5](https://www-demo5.allizom.org/en-US/firefox/new/?v=2).

bedrock/settings/static_media.py
```diff
@@ -1130,6 +1152,25 @@
         ),
         'output_filename': 'js/firefox_new-bundle.js',
     },
+    'firefox_new_common': {
+        'source_filenames': (
+            'js/base/search-params.js',
```
alexgibson: Is this still needed in the bundle?

jpetto: Hm, no it is not. And now that `common.js` is so small, we may not need that file, either...which would mean getting rid of the `common` bundle altogether. Working on a refactor.

alexgibson:
Tested on demo5:

IE 6-11, MS Edge, Chrome, Safari, Opera, Firefox, iOS - all working as expected :+1: 

Aside from the bundle comment above, this gets an r+wc from me. Nice work!

jpetto:
Removed `common.js` & `common` bundle. Refactored `scene1.js` & `scene2.js`. Pushed latest to demo5. Tested in IE 6 - 11, MSEdge, Chrome, Fx, and Safari.

Might be good to go after a squash?

alexgibson:
> Might be good to go after a squash?

Looks good to me, squash away :+1:

media/js/firefox/new/scene1.js
```diff
@@ -0,0 +1,113 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// set up namespace
+if (typeof Mozilla == 'undefined') {
+    var Mozilla = {};
+}
+
+Mozilla.Firefox = window.Mozilla.Firefox || {};
```
alexgibson: Wait, is this needed still? Same with the other JS file

jpetto: Ah, good catch. No namespace stuff needed anymore. Will remove and squash.

PR #3838, opened 2016-02-08
Fix Bug 1245208, add plugincheck layout variations for newsletter signup testing

schalkneethling:
@craigcook r?


schalkneethling:
Should have added this is the description of the PR. The pages are:

## Current
/plugincheck/

## Variants

/plugincheck/?v=b
/plugincheck/?v=c
/plugincheck/?v=d

media/css/plugincheck/plugincheck.less
```diff
@@ -381,10 +387,215 @@ html[dir="rtl"] {
...
+
+        .firefox {
+            font-weight: bold;
+        }
+    }
+
+    h4 {
+        .font-size(2rem);
+        line-height: 1.2;
+        letter-spacing: 0;
```
craigcook: I think this should be `letter-spacing: normal;` though the effect looks to be the same (probably depends on fonts).

craigcook:
Looking good, though the positioning of the logo in the header feels wrong at desktop and tablet sizes: 

plugincheck-header

I think it should be more aligned with the headline. We could put the logo inside the `main-feature` block to make it a bit easier, or else keep it where it is and use some padding and negative margin trickery to move things around like this (hacked in devtools):

plugincheck-header-2

media/css/plugincheck/plugincheck.less
schalkneethling: `normal` sounds more appropriate, will change.

schalkneethling:
Hmmm, yeah. I followed the design but, you are correct so, not sure. Should I just go ahead and change it or, first fly this by Michael?

craigcook:
Yeah, let's run it by Michael and see what he thinks.

designerham:
@schalkneethling and @craigcook I prefer the second version where the logo is in line with the header an sub head – much nicer. Thanks!

schalkneethling:
@craigcook updated, and ready to go. Thx

craigcook:
r+ :lollipop:

PR #3841, opened 2016-02-09
Fix Bug 1246806, ensure video element interactions work, remove underline from button hover state

schalkneethling:
@jpetto r?

### DIFF
--- a/media/js/mozorg/contribute/taskview.js
+++ b/media/js/mozorg/contribute/taskview.js
@@ -203,40 +203,69 @@
     }
 
     /**
+     * Called by joyOfCoding, monitors video playback and marks the video step
+     * as watched once we reach the 40sec mark.
+     * @param {object} $video - The video element as a jQuery object.
+     */
+    function markAsWatched($video) {
+        var videoElement = $video[0];
+
+        // a user can click play again after having watched the video the
+        // first time. Clicking on pause, for example, will also trigger the
+        // click event so, we need to ensure we only run the below on the
+        // first interaction.
+        if ($video.data('watched') !== true) {
+            $video.on('timeupdate.taskview', function(event) {
+                // user needs to watch at least 40 seconds before we mark
+                // this step as complete.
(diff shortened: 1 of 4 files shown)


media/js/mozorg/contribute/taskview.js
```diff
@@ -203,40 +203,72 @@
...
-                        // once the step has been completed,
-                        // remove the event listener.
-                        $jocVideo.off('timeupdate.taskview');
-                        $jocVideo.data('watched', true);
-                        trackInteraction('completed joy of coding', $this.data('variation'));
-                    }
-                });
-            }
+        // ensure the video is paused before playing and tracking the interaction.
+        if ($this.data('action') === 'play' && videoElement.paused) {
```
jpetto: Is it possible for this `if` block and the `if` block above to both execute? If not, I'd suggest wrapping this `if/else` block in an `else` for clarity.

media/js/mozorg/contribute/taskview.js
```diff
@@ -203,40 +203,72 @@
...
+            videoElement.play();
+
+            $this.attr('data-ga', true);
+            $jocVideo.attr('data-ga', true);
+
+            trackInteraction('play joy of coding', $this.data('variation'));
+            markAsWatched($this);
```
jpetto: I think this should be `markAsWatched($jocVideo)`?

media/js/mozorg/contribute/taskview.js
schalkneethling: Yeah, I was in two minds about that, probably safer to if/else

media/js/mozorg/contribute/taskview.js
schalkneethling: Oooooh, indeed.

jpetto:
All appears to be working properly now. :+1:

Care to squash?

jpetto:
Looks good! r+

PR #3842, opened 2016-02-10
[bug 1245208] Temporarily skip plugincheck newsletter test on non-Firefox browsers

alexgibson:
@schalkneethling r? This test fails in non-firefox browsers as the newsletter is no longer visible on the testing variations if the user does not use Firefox.

### DIFF
--- a/tests/functional/test_plugincheck.py
+++ b/tests/functional/test_plugincheck.py
@@ -14,6 +14,7 @@ def test_not_supported_message(base_url, selenium):
     assert page.is_not_supported_message_displayed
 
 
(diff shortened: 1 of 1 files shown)


schalkneethling:
@alexgibson darn it, indeed. So, is there an additional dependency one needs to install to run tests again Chrome? I tried specifying the driver as Chrome got a huge stack trace of errors. Thanks for the PR.

r+ :+1:

schalkneethling:
Ah ok, found the answer http://pytest-selenium.readthedocs.org/en/latest/user_guide.html#specifying-a-browser

alexgibson:
Yep, you just need to install Chrome driver. Thanks @schalkneethling

PR #3845, opened 2016-02-10
Fix bug 1246795 - Firefox Accounts home promo

craigcook:
Also fixes bug 1247103 - replace Webmaker promo with Teach promo

The Accounts promo is English only, no l10n (for now) so I haven't wrapped the strings. Other locales fall back to the evergreen private browsing promo.

The Teach promo we had previously and I believe the strings are still in place behind the `webmaker_promo_introducing` flag. @flodolo please correct me if I'm wrong and I can give it a new flag for the new string.


flodolo:
All the old strings have been removed, [this](https://l10n.mozilla-community.org/~pascalc/langchecker/?website=0&file=mozorg/home/index.lang&action=translate) is all we got.

This is basically a new string, so we need a different tag. Also, I'd suggest to remove the previous webmaker tag and string to avoid starting to pile up again.

Side question: if we plan to localize the FxA tile at some point, it would be better to expose the strings now (not sure what's up with all the en-US only tiles recently).

craigcook:
Ah, looks like I had outdated lang files, I still have the old strings locally. I'll give it a new tag.

The FxA promo links to the accounts page which isn't currently localized. If that page gets localized we could localize the promo as well, but I wouldn't want to link a localized promo to an English page. Can we expose the strings without actually activating the promo?

craigcook:
New flag for the Teach promo: `teach_the_web_promo`

And I've wrapped strings for the FxA promo and put it behind the flag `firefox_accounts_promo`, but it's also explicitly behind an "if English" check. If we do want to release it to more locales down the line we can remove the English check (assuming the accounts page is also localized).

flodolo:
OK, that makes sense. Leaving the strings wrapped is OK, at this point I'd remove the tag (we'd still need to remove the "if English" conditional is we decide to localize it).

We could drop the conditional and leave the tag, and I won't expose the strings to localizers, but the promo would show up on the dev server and it would be more confusing.

flodolo:
Added string and imported from the old file where possible
https://github.com/mozilla-l10n/www.mozilla.org/commit/1271824fe37fa7e5824af64d2310d9362c902923

I'll clean up the old string when this is merged to production.

media/img/home/voices/promos/webmaker/webmaker-teacher.svg
```diff
@@ -0,0 +1,3 @@
+<!DOCTYPE svg PUBLIC "-//W3C//DTD SVG 1.1//EN" "http://www.w3.org/Graphics/SVG/1.1/DTD/svg11.dtd" [
+	<!ENTITY ns_flows "http://ns.adobe.com/Flows/1.0/">
```
schalkneethling: Do we need the `DOCTYPE` etc. I reckon we can remove everything to just before the open `<svg....`

schalkneethling:
Just one comment about the webmaker SVG. Other than that, and the suggestion from @flodolo  to remove the tag and just keep the `en` conditional, the code looks good.

media/img/home/voices/promos/webmaker/webmaker-teacher.svg
craigcook: Huh, surprised the optimizer left this in... it certainly seems unnecessary. I'll update.

craigcook:
Removed the `firefox_accounts_promo` l10n flag and SVG cruft. I think we're all set.

schalkneethling:
r+ :sunny:

Issue #3844, opened 2016-02-10
New root certificate to add to /certs/


floatingatoll:
https://wiki.mozilla.org/MozillaRootCertificate references the file https://www.mozilla.org/certs/mozilla-root.crt for the current internal SHA-1 CA.

I would like to add two new files alongside it for the new internal SHA-2 CA, titled:

```
mozilla-root-sha2.crt
mozilla-root-sha2.crt.sha256sum
```

I'll email the files to @jgmize as it's not clear where I would put them in a pull request.

floatingatoll:
Emailed.

```
746eb47c02524cf15dea7e947be4b47cf6fb96ce1a41fe87d8ffc2ce64ba134d  ca_sha2.crt
```

jgmize:
Added in #3847, which I plan to push to production tomorrow morning.

jgmize:
> I would like to add two new files alongside it for the new internal SHA-2 CA, titled:
> mozilla-root-sha2.crt
> mozilla-root-sha2.crt.sha256sum

@floatingatoll I just noticed that the filenames you used when you reported this issue are different than those in your signed email attachment, which are what I used in #3847. Would like them renamed to match those above, or do you prefer them as they are?

floatingatoll:
Please rename them to match this request - thanks for checking!

PR #3847, opened 2016-02-10
Add certs to bedrock: bug 1149019

jgmize:


jgmize:
Fixes #3844

jgmize:
@pmclanahan r?

pmclanahan:
Looks good, but the files should be in `media/` not `static/`.

bedrock/base/static.py
```diff
@@ -12,11 +12,13 @@ class BedrockWhiteNoise(DjangoWhiteNoise):
     def __init__(self, application):
         super(BedrockWhiteNoise, self).__init__(application)
 
-        # TODO add tests for these in mcom-tests
+        # TODO add functional tests for these
```
pmclanahan: flake8 doesn't like the trailing whitespace on the comment :( seems a bit picky, but okay.

pmclanahan:
Move the files and fix flake8 and we're good. Thanks!

jgmize:
> Move the files and fix flake8 and we're good. Thanks!

Thanks @pmclanahan, good catch on the dir. I've fixed that and removed the trailing space in 6edb8a8

PR #3848, opened 2016-02-11
Tag docker image built w/ tagged commit as latest

jgmize:
This will make it easier to run the latest production docker image locally without having to look up the commit hash.

### DIFF
--- a/docker/jenkins/push2dockerhub.sh
+++ b/docker/jenkins/push2dockerhub.sh
@@ -16,4 +16,12 @@ COMMIT="${ghprbActualCommit:=$GIT_COMMIT}"
 docker tag -f $FROM_DOCKER_REPOSITORY:$COMMIT $DOCKER_REPOSITORY:$COMMIT
 
 # Push to docker hub
-docker push $DOCKER_REPOSITORY:$COMMIT
+docker push -f $DOCKER_REPOSITORY:$COMMIT
+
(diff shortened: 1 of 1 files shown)


jgmize:
@pmclanahan or @glogiotatidis r?

pmclanahan:
LGTM. I'll let Giorgos have a look as well before merging though.

glogiotatidis:
Maybe you want to tag the docker image with the git tag as well?

Thanks @jgmize r+

pmclanahan:
@glogiotatidis it looks like that's being done just above this change. Am I interpreting that code incorrectly?

glogiotatidis:
> @glogiotatidis it looks like that's being done just above this change. Am I interpreting that code incorrectly?

I didn't explain myself correctly. The current code will tag the docker image as `latest`. I'm suggesting that we tag it using the git tag as well. For example as `2016-02-10`

jgmize:
> I'm suggesting that we tag it using the git tag as well. For example as 2016-02-10

Not a bad idea, and it should only add a few seconds to a job that doesn't block anything else. I've updated this PR with the additional tagging in 529adea69c6d5e35d13b9901ca536931c50a246a.

pmclanahan:
Yup. I misunderstood. Thought you meant tag it with the git hash, but the "git tag" is a good idea.

PR #3850, opened 2016-02-11
Remove unsupported "-f" flag from docker push cmds

jgmize:
This is not implemented: https://github.com/docker/docker/issues/2619


jgmize:
@pmclanahan r?

Sorry, my mistake.

PR #3851, opened 2016-02-11
[fix bug 1243890] Update FxA iframe param for Fx46.

jpetto:
Have not yet updated `firefox/accounts.js` as I wanted to first get a sanity check on the approach.

Code is up on demo5:

[https://www-demo5.allizom.org/en-US/firefox/46.0/firstrun/](https://www-demo5.allizom.org/en-US/firefox/46.0/firstrun/)

Change isn't immediately obvious, so I'm logging the FxA iframe `src` to the console. If Fx 46+, `src` string should contain `context=fx_firstrun_v2`. Otherwise, `context=iframe`.

Actual change - after verifying email, Fx 46+ users see a blue "Sync Preferences" button in the iframe. Fx < 46 users do not get this button.

@alexgibson - Pre-review on approach? What do you think about the UITour fallback and timing?

### DIFF
--- a/media/js/firefox/firstrun/firstrun.js
+++ b/media/js/firefox/firstrun/firstrun.js
@@ -2,25 +2,31 @@
     'use strict';
 
     var $fxaFrame = $('#fxa');
-    var fxaIframeSrc = $('#intro').data('fxa-iframe-src');
+    var fxaIframeHost = $('#intro').data('fxa-iframe-host');
     var fxaFrameTarget = ($fxaFrame.length) ? $('#fxa')[0].contentWindow : null;
+    var fxaIframeSrc = $fxaFrame.data('src');
     var resizeTimer;
     var fxaHandshake = false;
 
     // remove trailing slash from iframe src (if present)
-    fxaIframeSrc = (fxaIframeSrc[fxaIframeSrc.length - 1] === '/') ? fxaIframeSrc.substr(0, fxaIframeSrc.length - 1) : fxaIframeSrc;
+    fxaIframeHost = (fxaIframeHost[fxaIframeHost.length - 1] === '/') ? fxaIframeHost.substr(0, fxaIframeHost.length - 1) : fxaIframeHost;
+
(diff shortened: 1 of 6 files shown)


media/js/firefox/firstrun/firstrun.js
```diff
@@ -63,15 +66,42 @@
...
+        console.log('uitour appinfo complete');
+        if (!fxaIframeSrcSet) {
+            // UITour succeeded - no need for fallback above
+            window.clearTimeout(fxaIframeTimer);
+
+            var version = parseInt(config.version, 10);
+
+            console.log('fx version = ' + version);
+
+            if (version >= 46) {
```
alexgibson: I may well be missing something, but if you're only after the major version number then using UITour for this could be overkill perhaps? If you needed the minor version or channel name then UITour is the only way, but if the major version number is the only thing of importance, then I would probably use `Mozilla.Client.FirefoxMajorVersion` instead. This is both synchronous and does not require any special config.

jpetto: I went right for the big guns, didn't I? Refactoring...

jpetto:
Okay, I think I've stopped trying too hard. :grin:

Ready for a real review.

alexgibson:
When I updated the bedrock docs for local testing, I totally forgot to include the bit about using the right port (`127.0.0.1:8111`) - could you add it back here? :smile:

jpetto:
> When I updated the bedrock docs for local testing, I totally forgot to include the bit about using the right port (127.0.0.1:8111) - could you add it back here? :smile:

I think I can handle that. :+1:

alexgibson:
Tested locally and both pages are working as expected, with the correct iframe query param and I see blue button after verifying email in FF 46+ :+1: 

r+wc from me once the docs are updated for my faux pas

Also noted that we should probably look to consolidate this code for the iframe sometime in the future, as there's a fair bit of duplication now?

PR #3853, opened 2016-02-12
Tag and push latest base image from tagged commits

jgmize:
@glogiotatidis r?

### DIFF
--- a/docker/jenkins/push2dockerhub.sh
+++ b/docker/jenkins/push2dockerhub.sh
@@ -20,8 +20,17 @@ docker push $DOCKER_REPOSITORY:$COMMIT
 
 GIT_TAG="$(git describe --tags --exact-match $GIT_COMMIT 2> /dev/null)"
 if [[ ! -z $GIT_TAG ]]; then
+    # tag and push docker image with git tag
     docker tag -f $FROM_DOCKER_REPOSITORY:$COMMIT $DOCKER_REPOSITORY:$GIT_TAG
     docker push $DOCKER_REPOSITORY:$GIT_TAG
+
+    # tag and push docker image with "latest"
     docker tag -f $FROM_DOCKER_REPOSITORY:$COMMIT $DOCKER_REPOSITORY:latest
     docker push $DOCKER_REPOSITORY:latest
+
+    # tag and push base image with latest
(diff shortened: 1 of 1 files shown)


docker/jenkins/push2dockerhub.sh
```diff
@@ -20,8 +20,17 @@ docker push $DOCKER_REPOSITORY:$COMMIT
...
+    # tag and push docker image with git tag
     docker tag -f $FROM_DOCKER_REPOSITORY:$COMMIT $DOCKER_REPOSITORY:$GIT_TAG
     docker push $DOCKER_REPOSITORY:$GIT_TAG
+
+    # tag and push docker image with "latest"
     docker tag -f $FROM_DOCKER_REPOSITORY:$COMMIT $DOCKER_REPOSITORY:latest
     docker push $DOCKER_REPOSITORY:latest
+
+    # tag and push base image with latest
+    BASE_IMAGE="$(docker run mozorg/bedrock bash -c "head -n 1 Dockerfile" | sed "s/FROM //")"
```
glogiotatidis: So there are two jenkins jobs that run the `push2dockerhub` script

* https://ci.us-west.moz.works/view/Bedrock/job/bedrock_push_dockerhub/configure
* https://ci.us-west.moz.works/view/Bedrock/job/bedrock_code_push_dockerhub/configure

Here you get the `FROM` of `mozorg/bedrock` which should be `mozorg/bedrock_code` but it's `bedrock_base` b/c the image squashing is not working the way it's supposed to work, but that's another bug.

I'm wondering whether it's better to implement another jenkins job to `bedrock_base_push_dockerhub`?

jgmize:
Good points, I missed the other job and this approach won't work with both. I think your suggestion of a separate bedrock_base_push_dockerhub job makes more sense. Thanks @glogiotatidis!

jgmize:
> [...] image squashing is not working the way it's supposed to work, but that's another bug.

@glogiotatidis would you mind following up with a bug describing the image squashing issue in more detail, or linking to the bug if you have already? I'd like to do some research on it but I don't have a clear understanding of the problem.

jgmize:
Created the bedrock_base_push job and configured the bedrock_base_image job to trigger it as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1248065

PR #3856, opened 2016-02-13
[bug 1248040] Update GTM data attributes on /firefox/new/

JamesDLorence:
Added attributes to tabzilla, simple footer, and 'click here' link.

### DIFF
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -102,7 +102,7 @@
 
         {% block tabzilla_tab %}
           <div id="tabzilla">
-            <a href="{{ url('mozorg.home') }}">Mozilla</a>
+            <a href="{{ url('mozorg.home') }}" data-link-type="nav" data-link-name="tabzilla">Mozilla</a>
           </div>
         {% endblock %}
 
--- a/bedrock/base/templates/base.html
+++ b/bedrock/base/templates/base.html
@@ -98,7 +98,7 @@
 
         {% block tabzilla_tab %}
         <div id="tabzilla">
-          <a href="{{ url('mozorg.home') }}">Mozilla</a>
+          <a href="{{ url('mozorg.home') }}" data-link-type="nav" data-link-name="tabzilla">Mozilla</a>
         </div>
         {% endblock %}
 
--- a/bedrock/firefox/templates/firefox/includes/simple_footer.html
+++ b/bedrock/firefox/templates/firefox/includes/simple_footer.html
@@ -12,9 +12,9 @@
       <li>{% include 'includes/lang_switcher.html' %}</li>
     </ul>
     <ul class="secondary">
-      <li><a href="{{ url('privacy') }}">{{_('Privacy')}}</a></li>
-      <li><a href="{{ url('privacy.notices.websites') }}#cookies">{{ _('Cookies') }}</a></li>
-      <li><a href="{{ url('legal.index') }}">{{_('Legal')}}</a></li>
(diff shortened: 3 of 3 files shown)


bedrock/firefox/templates/firefox/new.html
```diff
@@ -155,7 +155,7 @@ <h1 class="large" id="main-headline">
     <div id="scene2-main" class="scene-content scene2 hidden">
     {% if l10n_has_tag('firefox_new_mobile_promo') %}
       <p class="help">
-        {{_('Your download should begin automatically. If not, <a id="%s" href="">click here</a>. It could take a few minutes, but it’s worth the wait.')|format('direct-download-link')}}
+        {{_('Your download should begin automatically. If not, <a id="%s" href="" data-link-type="download" data-download-os="Desktop">click here</a>. It could take a few minutes, but it’s worth the wait.')|format('direct-download-link')}}
```
alexgibson: Because this link occurs in the middle of a string, adding in data attributes here breaks the translation for over over 100 locales. I suggest we avoid making changes like this in general, so we don't make extra work for our translators. Is there some other way you can track this link click?

alexgibson:
We're also currently testing a new split template for `/firefox/new`, which can be found here:

https://github.com/mozilla/bedrock/tree/master/bedrock/firefox/templates/firefox/new

Any changes made to the current template should probably also be reflected here, else changes will likely be lost once we move to the new split template model.

alexgibson:
There is also an occurrence of the tabzilla link in the non-responsive template to update: https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/base.html#L101

JamesDLorence:
@alexgibson - I remember you mentioned about the localization issue when adding data attributes so I will make sure to test that with every change I make going forward.

On your last comment, I was trying to keep scope to just changes that affected the /firefox/new/ page. This doesn't seem like a template that affects the /firefox/new/ page. Is that incorrect, or are you wanting me to just update all tabzilla buttons now?

Lastly, should I pull down new commits when I push my corrected commit, or should I just push the corrected commit?

alexgibson:
Any changes you make to the base templates will effect all pages across Mozilla.org, not just /firefox/new.

Just add your follow up commits here to address the code review changes. GitHub will tell you if the branch needs rebasing or not, so until that happens just carry on with your PR :)

JamesDLorence:
I realize that changes to templates can affect multiple pages, but I what I wanted to do to keep things organized was to focus on one page at a time where I made the necessary changes to it's templates only. I know that might mean other pages are changed, which is fine as nothing will affect the web analytics data until I make updates in your Google Tag Manager, but this is how we broke up the requirements to it will help me if can do it that way.

So for this specific example, I will eventually update all tabzilla links with data attributes, but I just wanted to focus on the template for the /firefox/new/ page and make sure that one was working fine.

Are you comfortable with doing it that way?

alexgibson:
If we are making a change to tabzilla in this PR we should do it to both base templates, it's a single line change and easy to forget in future.

alexgibson:
To be clear, this already changes 95% of pages on bedrock just updating the responsive template. It makes no sense not to do the rest.

JamesDLorence:
Sounds good, I'll make the change to the responsive template.

I just wanted to make sure you were good with me moving forward as I described.

JamesDLorence:
@alexgibson - I pushed up an amended commit which removed the translation issue and added in the addition of the data attributes to the non-responsive template.

alexgibson:
r+

PR #3858, opened 2016-02-14
Fix Bug 1246216 - Update Ari Jaaksi's Mozillians profile URL

kyoshino:


craigcook:
Looks like both profiles exist but the old one is "pending approval" and not publicly visible. The new one is approved and vouched, but also not publicly visible. I think he'll need to log in himself and set it to public, unless a Mozillians admin can do that. 

So r+ on this change and we should follow up separately to get his profile made public.

PR #3859, opened 2016-02-14
Fix Bug 1247291 - Updates needed to https://www.mozilla.org/en-US/about/legal/impressum/

kyoshino:
Screenshot: https://bugzilla.mozilla.org/show_bug.cgi?id=1247291#c2


schalkneethling:
Just an interesting observation. When accessing this page via `en-US` some copy on the page is actually in English. Is this page not meant just for German speaking users? Was just wondering if it should not always just redirect to `/de/`

As stated, just an observation. r+ Thanks @kyoshino :de:

kyoshino:
Yeah, will send a separate PR to set redirects from other languages.

PR #3860, opened 2016-02-14
Fix Bug 1237010 - [traceback] TypeError: unsupported operand type(s) for +: 'NoneType' and 'str' on /firefox/ios/releasenotes/

kyoshino:


jgmize:
Thanks @kyoshino

PR #3862, opened 2016-02-15
[bug 1239863] Decommission legacy Firefox OS pages

alexgibson:


flodolo:
Will look into the template to see how much we're missing and if it's worth localizing it for a couple of months.

flodolo:
From a quick check we basically don't have any of the strings used in the 2.5 template, so you can get rid of all the `l10n_has_tag` and clean up the strings in the template.

flodolo:
We can also remove the [extra lang files](https://github.com/mozilla/bedrock/pull/3862/files#diff-098c57b8ccbfc803cb1f167c55f019daL9). That would also avoid enabling the page on production because the existing index-new-lang is activated.

alexgibson:
Removed l10n tags and extra lang files from the template, thanks @flodolo

media/js/firefox/os/devices.js
```diff
@@ -197,31 +197,6 @@ if (typeof window.Mozilla === 'undefined') {
     }
     */
 
-    // only if coming from /firefox/os/, detect country
```
alexgibson: Note: removing this as part of [bug 1233052](https://bugzilla.mozilla.org/show_bug.cgi?id=1233052) now that the legacy pages are gone.

alexgibson:
Also added a second commit to remove the Firefox OS newsletter signup form. See bug https://bugzilla.mozilla.org/show_bug.cgi?id=1244200

flodolo:
Side note: it would be really helpful to have this merged, or at least on a demo server, since people started localizing but they can check their work yet.

alexgibson:
@flodolo - this is now up on demo2, thanks https://www-demo2.allizom.org/firefox/os/

bedrock/firefox/templates/firefox/os/index.html
```diff
@@ -77,220 +55,81 @@
...
-            <div class="primary-cta-signup signup hidden">
-              <a href="#email-form-wrapper" class="cta-button get-newsletter newsletter-signup-toggle" role="button" aria-haspopup="true">{{_('Sign me up')}}</a>
-              <p class="signup-txt">{{_('Sign up for our monthly newsletter and get the latest on Firefox OS')}}</p>
-            </div>
-            <a href="#get-device" class="primary-cta-phone cta-button get-phone hidden" role="button" aria-haspopup="true">{{_('Get a phone')}}</a>
+        <div class="hero-content">
+          <h1>{{ _('Experience Firefox OS on your Android device')}}</h1>
+          <p>{{_('Introducing Firefox OS 2.5 Developer Preview, an experimental app that lets you use Firefox OS on your Android device.')}}</p>
+          <div class="primary-cta">
+            <a href="https://d2yw7jilxa8093.cloudfront.net/B2GDroid-mozilla-central-nightly-latest.apk" class="cta-button" data-pixel="52161" data-id="Mozilla_B2G_Activity_Tag_1">{{_('Get the Android App')}}</a>
```
jpetto: Not sure how long-term this link is (the cloudfront URL seems random), but if it's going to be changing with some frequency, should consider moving to settings.

bedrock/firefox/redirects.py
```diff
@@ -291,7 +291,11 @@ def firefox_mobile_faq(request, *args, **kwargs):
              'https://support.mozilla.org/kb/about-tiles-new-tab'),
 
     # Bug 1239863
+    redirect(r'^firefox/os/1\.1/?$', 'firefox.os.index'),
+    redirect(r'^firefox/os/1\.3/?$', 'firefox.os.index'),
+    redirect(r'^firefox/os/1\.3t/?$', 'firefox.os.index'),
```
jpetto: Do we need a redirect for 1.4 as well?

tests/redirects/map_globalconf.py
```diff
@@ -986,6 +986,11 @@
              'https://support.mozilla.org/kb/about-tiles-new-tab'),
 
     # Bug 1239863
+    url_test('/firefox/os/2.5/', '/firefox/os/'),
+    url_test('/firefox/os/2.0/', '/firefox/os/'),
```
jpetto: Add a test for 1.4?

jpetto:
Aside from questions regarding 1.4 redirects/tests, looks great. r+wc

bedrock/firefox/templates/firefox/os/index.html
alexgibson: I don't think this is likely to change in the future, but it is already repeated more than once in the template it seems. I'll move it to settings :+1:

bedrock/firefox/redirects.py
alexgibson: This is why code-review is awesome. Totally forgot there was a 1.4 template to account for :wink:

tests/redirects/map_globalconf.py
alexgibson: Will do!

bedrock/firefox/templates/firefox/os/index.html
alexgibson: Also yeah, no idea where this cloudfront link came from? *shrugs*

alexgibson:
Thanks @jpetto updated

jpetto:
:candy:

PR #3863, opened 2016-02-15
[fix bug 1245521] TP Tour should always try and close tab when info panels are dismissed

alexgibson:


schalkneethling:
r+ :mask:

PR #3866, opened 2016-02-15
[bug 1240980] Mozilla Support link on Hello page should use title case

alexgibson:
Updating this for en locales only as it's not worth re-translating.


schalkneethling:
r+ :anchor:

PR #3869, opened 2016-02-16
[fix bug 1246461] Update Tracking Protection tour string for step 3 in non-private windows

alexgibson:


schalkneethling:
@alexgibson So, I only have one tab open in Fx Nightly. I then go to:
http://localhost:8000/en-US/firefox/47.0a1/tracking-protection/start/

When I get to step 3, there is a mismatch

screen shot 2016-02-16 at 11 52 38

However, if I append `&newtab=true`, all is good

screen shot 2016-02-16 at 11 53 05

Is there a flaw in my testing steps or, does this indicate a possible problem?

alexgibson:
@schalkneethling - the string in control center is dependent on whether or not you use a private window. You need to launch the tour in a private window to see "Disable protection for this session"

schalkneethling:
Aha, ok. Let me test that. Thx.

schalkneethling:
Following a discussion on IRC with @alexgibson I learned that in non-PBM `&newtab=true` is always present so, the "bug" in https://github.com/mozilla/bedrock/pull/3869#issuecomment-184602801 is not relevant.

r+ :honey_pot:

PR #3870, opened 2016-02-16
Fix Bug 1221087, set map state correctly for sub page landing state

schalkneethling:
@alexgibson r?

### DIFF
--- a/media/js/mozorg/contact-spaces.js
+++ b/media/js/mozorg/contact-spaces.js
@@ -406,8 +406,16 @@
         setInitialContentState: function () {
             // store initial content data id on page load
             var state = mozMap.getMapState();
+
             if (state === 'spaces') {
                 initContentId = $('#nav-spaces li.current').data('id');
+
+                // if we land directly on a specific space, like Mountain View,
+                // pan the marker to that space.
+                if (initContentId) {
+                    mozMap.panToMarker(initContentId);
+                }
+
                 // Show spaces nav+meta
                 if (wideMode) {
                     $('#nav-spaces, #meta-spaces').show();
@@ -416,6 +424,10 @@
                 }
             } else if (state === 'communities') {
                 initContentId = $('#nav-communities li.current').data('id');
+
+                var region = initContentId ? initContentId : 'communities';
+                mozMap.showCommunityRegion(region);
+
                 // Show community nav+meta
                 if (wideMode) {
                     $('#nav-communities, #meta-communities').show();
@@ -472,14 +484,14 @@
                 mozMap.clearCommunityLayers();
                 // unbind click events on community nav
                 mozMap.unbindCommunityNav();
+                // unbind click events on spaces nav
+                mozMap.unbindSpacesNav();
                 // add spaces marker layer.
                 mozMap.addSpacesMarkers();
                 // hide community legend
                 mozMap.hideMapLegend();
                 // reposition markers above the labels
                 mozMap.setLabelLayerIndex(1);
-
-                mozMap.updateTabState('contact');
             }
         },
 
@@ -554,7 +566,7 @@
          */
         panToMarker: function (id) {
             // if we're on the landing page zoom out to show all markers.
-            if (id === 'spaces') {
+            if (id === 'spaces' || id === 'contact') {
                 map.setView([28, 0], 2);
                 return;
             }
@@ -818,7 +830,7 @@
                 // update the page title
                 mozMap.setPageTitle(cacheId);
                 // this reuses the spaces view
-                mozMap.panToMarker('spaces');
+                mozMap.panToMarker('contact');
             } else {
                 $entryContainer.empty();
                 // request content via ajax


media/js/mozorg/contact-spaces.js
```diff
@@ -408,6 +409,7 @@
             var state = mozMap.getMapState();
             if (state === 'spaces') {
                 initContentId = $('#nav-spaces li.current').data('id');
+                mozMap.showSpace(state.url, null);
```
alexgibson: Looks like this causes a redundant ajax request, would it be better just to show the correct marker/map position here instead?

media/js/mozorg/contact-spaces.js
```diff
@@ -416,6 +418,7 @@
                 }
             } else if (state === 'communities') {
                 initContentId = $('#nav-communities li.current').data('id');
+                mozMap.setMapState();
```
alexgibson: Calling this here doesn't fix the bug for displaying the community map overlay?

media/js/mozorg/contact-spaces.js
```diff
@@ -160,6 +160,7 @@
                     // Hide spaces and community nav+meta
                     mozMap.toggleNav('contact');
                     mozMap.toggleMobileNav('contact');
+                    mozMap.setMapState();
```
alexgibson: Does this need to be called here? Looking at the code, it seems like this should only need calling if the active tab changes (see line 137)

schalkneethling: Yeah, it seemed like when you switch from a specific space, for example, back to contact us, the map does not change. Perhaps it is just loading really slowly for me?

media/js/mozorg/contact-spaces.js
schalkneethling: Hmmm, it worked for me, although it does take a looong time to load. Will test some more.

media/js/mozorg/contact-spaces.js
schalkneethling: Of course, I have it at the line right above this one :blush:

media/js/mozorg/contact-spaces.js
alexgibson: Hmm, I wonder if this is just masking over another bug perhaps - something that was added with the contact tab?

media/js/mozorg/contact-spaces.js
alexgibson: I think here you might want to use something like:

    var region = initContentId ? initContentId : 'communities';
mozMap.showCommunityRegion

schalkneethling:
@alexgibson Ok, I have made some updates and, for me, it seems like all the kinks are sorted now.

media/js/mozorg/contact-spaces.js
```diff
@@ -408,6 +409,7 @@
             var state = mozMap.getMapState();
             if (state === 'spaces') {
                 initContentId = $('#nav-spaces li.current').data('id');
+                mozMap.showSpace(state.url, initContentId);
```
alexgibson: This is still causing a redundant ajax request on page load. `mozMap.showSpace()` does a lot more than setting the map looking at the code. I seem to remember addressing this ajax issue was how the regression snuck in to begin with.

media/js/mozorg/contact-spaces.js
alexgibson: This still doesn't fix this issue you describe when I test locally (switch from a specific space, for example, back to contact us).

schalkneethling: Hmmmm, so I only see requests being made for spaces when it has not already been loaded. Are you seeing it always make a request to load the content or, are you referring to a different request?

media/js/mozorg/contact-spaces.js
schalkneethling: Hmmmm, so I only see requests being made for spaces when it has not already been loaded. Are you seeing it always make a request to load the content or, are you referring to a different request?

alexgibson: Hard-refresh on any Space URL and I see the same content load twice

media/js/mozorg/contact-spaces.js
schalkneethling: Hmmmmm, this is wacky. It seems like it is very unstable. Let me continue looking at it. Do we have functional tests for this page?

alexgibson: Yes we have functional tests - I don't think it is unstable per-say, just something isn't being set quite right. The contact tab should need to follow the same logic as the other two tabs.

schalkneethling:
@alexgibson ok, I have updated the PR. From what I can see, this works well. If you are seeing problems though or, redundant Ajax calls, please let me know.

media/js/mozorg/contact-spaces.js
```diff
@@ -478,8 +490,6 @@
                 mozMap.hideMapLegend();
                 // reposition markers above the labels
                 mozMap.setLabelLayerIndex(1);
-
-                mozMap.updateTabState('contact');
```
alexgibson: Not introduced in this change, but here if `state === 'contact'` is true, we're calling `mozMap.unbindCommunityNav();` to unbind the communities navigation, but we're seemingly not not also calling `mozMap.unbindSpacesNav();`

alexgibson:
r+wc

alexgibson:
Looks good, thanks @schalkneethling!

PR #3871, opened 2016-02-16
[fix bug 1245573] Remove padding on firstrun FxA iframe.

jpetto:
Removing padding improves iframe UI after account confirmation. With padding, grey bottom half of iframe is "floating". Without padding, grey bottom half is butted up against border. (See [the GitHub issue](https://github.com/mozilla/fxa-content-server/issues/3471) for screenshot.)

@alexgibson r?


alexgibson:
r+ :package:

PR #3872, opened 2016-02-16
Fix bug 1244346 - add encryption home promo

craigcook:
Replaces gear promo. This is English-only because the page we're linking to isn't localized, and I don't know if there are any plans to localize it.

Note: I didn't delete the gear images because they're still used by old Tabzilla.


bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -160,13 +160,13 @@ <h2 class="go">{{ _('Download Firefox') }}</h2>
             <a class="fxos-link" href="{{ url('firefox.desktop.index') }}">{{ _('Learn more') }}</a>
           </div>
         </li>
-      {% if l10n_has_tag('gear_promo_eoy') %}
-        <li id="promo-6" class="item promo-large-landscape gear" tabindex="0">
-          <h2 class="primary go">{{ _('Get the gear, <br>support the mission') }}</h2>
-          <a class="panel-link" rel="external" href="https://gear.mozilla.org?ref=holiday2015&amp;utm_campaign=holiday2015&amp;utm_source=mozilla.org&amp;utm_medium=referral" data-promotion-id="promo-6" data-promotion-name="Get the gear, support the mission" data-promotion-class="promo-large-landscape" data-promotion-type="tile">
+      {% if LANG.startswith('en') %}
```
jpetto: Update to `en-`? Not sure how critical it is, but would be nice for consistency.

bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -160,13 +160,13 @@ <h2 class="go">{{ _('Download Firefox') }}</h2>
...
+        <li id="promo-6" class="item promo-large-landscape encryption" tabindex="0">
+          <h2 class="primary go">{{ _('Encryption helps <br>protect your privacy') }}</h2>
```
jpetto: Do you think we should `display: none;` this `<br>` for small viewports? Text wraps a little funny for tablet and smaller.

jpetto:
Couple minor suggestions, but otherwise looks good. r+wc

bedrock/mozorg/templates/mozorg/home/home.html
craigcook: Hm yeah, it's not horrible, but it does look better without the break. Will update.

PR #3874, opened 2016-02-17
Fix Bug 1206178, add android/all link to android send to device widget

schalkneethling:
@jpetto r?

@flodolo We have this string already but, not on this page perhaps?

### DIFF
--- a/media/css/firefox/android.less
+++ b/media/css/firefox/android.less
@@ -645,13 +645,23 @@ html[dir="rtl"] #phone-wrapper figure {
             border-radius: 0 0 6px 6px;
             border-top: none;
 
-            ul {
+            ul.android {
+                text-align: center;
+
                 li {
-                    float: none;
-                    width: 100%;
+
+                    &.android-all,
+                    &.google-play {
+                        display: inline-block;
+                        width: auto;
+                    }
 
                     &.google-play {
-                        text-align: center;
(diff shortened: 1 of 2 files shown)


flodolo:
It's in a shared file (`download_button.lang`), so it will be picked up automatically.

schalkneethling:
Thx @flodolo

media/css/firefox/android.less
```diff
@@ -645,13 +645,24 @@ html[dir="rtl"] #phone-wrapper figure {
...
+            ul.android {
+                text-align: center;
+
                 li {
-                    float: none;
-                    width: 100%;
+
+                    &.android-all,
+                    &.google-play {
+                        float: none;
```
jpetto: Small nit (and existed before this PR) - looks like `float: none;` is already specified in [send-to-device.less](https://github.com/mozilla/bedrock/blob/master/media/css/base/send-to-device.less#L84).

jpetto:
Super small nit, but otherwise looks good. r+wc :white_flower:

schalkneethling:
Thanks @jpetto also, have signoff here https://bugzilla.mozilla.org/show_bug.cgi?id=1206178#c18

schalkneethling:
@jpetto updated, thx!

PR #3879, opened 2016-02-18
Update settings.FIREFOX_IOS_RELEASE_VERSION to 2.0

jgmize:


jgmize:
https://www.mozilla.org/en-US/firefox/ios/2.0/releasenotes/ is live, this PR will update the links to point to the new version.

jgmize:
@sylvestre until we have Firefox for iOS integrated into product details, we'll need to update this setting and do a production push before https://www.mozilla.org/en-US/firefox/ios/notes/ will redirect to the correct version, so if 2.1 is released before we have iOS versions in product details you may want to submit a PR similar to this one along with a time frame you'd like to see it published.

alexgibson:
r+ thanks @jgmize

PR #3880, opened 2016-02-18
[fix bug 1222137] Enable bedrock locale-based templates.

jpetto:
@pmclanahan r?

### DIFF
--- a/lib/l10n_utils/__init__.py
+++ b/lib/l10n_utils/__init__.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import re
+from os.path import splitext
 
 from django.conf import settings
 from django.http import HttpResponseRedirect
@@ -91,13 +92,21 @@ def render(request, template, context=None, **kwargs):
 
             return response
 
-        localized_tmpl = '%s/templates/%s' % (request.locale, template)
+        # Render try #1: Look for l10n template in locale/{{ LANG }}/templates/
+        l10n_tmpl = '%s/templates/%s' % (request.locale, template)
         try:
-            return django_render(request, localized_tmpl, context, **kwargs)
+            return django_render(request, l10n_tmpl, context, **kwargs)
         except TemplateDoesNotExist:
-            # If not found, just go on and try rendering the parent template.
             pass
 
+        # Render try #2: Look for locale-specific template in app/templates/
+        locale_tmpl = '.{}'.format(request.locale).join(splitext(template))
+        try:
+            return django_render(request, locale_tmpl, context, **kwargs)
+        except TemplateDoesNotExist:
+            pass
+
+    # Render try #3: Render originally requested/default template
     return django_render(request, template, context, **kwargs)
 
 
--- a/lib/l10n_utils/tests/test_template.py
+++ b/lib/l10n_utils/tests/test_template.py
@@ -4,7 +4,8 @@
 
 import os
 
-from django.test import override_settings
+from django.template import TemplateDoesNotExist
+from django.test import RequestFactory, override_settings
 
 from jingo import env
 from jinja2 import FileSystemLoader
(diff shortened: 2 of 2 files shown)


jpetto:
Hm, tests are all passing locally...

lib/l10n_utils/__init__.py
```diff
@@ -91,13 +91,21 @@ def render(request, template, context=None, **kwargs):
...
+        l10n_tmpl = '%s/templates/%s' % (request.locale, template)
         try:
-            return django_render(request, localized_tmpl, context, **kwargs)
+            return django_render(request, l10n_tmpl, context, **kwargs)
         except TemplateDoesNotExist:
-            # If not found, just go on and try rendering the parent template.
             pass
 
+        # Render try #2: Look for locale-specific template in app/templates/
+        locale_tmpl = template.replace('.html', '.' + request.locale + '.html')
```
pmclanahan: I think it'd be more robust to use some stdlib tools here:

```python
>>> from os.path import splitext
>>> template = 'los-angles/laziest/dude.html'
>>> '.{}'.format('pt-BR').join(splitext(template))
'los-angles/laziest/dude.pt-BR.html'
```

lib/l10n_utils/tests/test_template.py
```diff
@@ -125,3 +125,37 @@ def test_render_no_locale(self, django_render, get_lang_path):
...
+
+
+@patch.object(env, 'loader', FileSystemLoader(TEMPLATE_DIRS))
+@override_settings(ROOT=ROOT)
+class TestLocaleTemplates(TestCase):
+    def setUp(self):
+        self.client = Client()
+
+    def test_default_template(self):
+        '''
```
pmclanahan: NIT: These should be three double quotes per pep8 `"""`.

lib/l10n_utils/tests/test_template.py
```diff
@@ -125,3 +125,37 @@ def test_render_no_locale(self, django_render, get_lang_path):
         # Note: no .locale on request
         # Should not cause an exception
         render(request, '500.html')
+
+
+@patch.object(env, 'loader', FileSystemLoader(TEMPLATE_DIRS))
+@override_settings(ROOT=ROOT)
+class TestLocaleTemplates(TestCase):
+    def setUp(self):
+        self.client = Client()
```
pmclanahan: Importing `Client` and doing this is unnecessary. Django `TestCase` class already has a `self.client` available.

lib/l10n_utils/tests/test_template.py
```diff
@@ -125,3 +125,37 @@ def test_render_no_locale(self, django_render, get_lang_path):
...
+        No custom templates exist for en-US or de. Should load
+        test_files/templates/firefox/new.html.
+        '''
+        response = self.client.get('/en-US/firefox/new/')
+        self.assertTemplateUsed(response, 'firefox/new.html')
+
+        response = self.client.get('/de/firefox/new/')
+        self.assertTemplateUsed(response, 'firefox/new.html')
+
+    def test_bedrock_locale_template(self):
```
pmclanahan: Another way to do these tests if this isn't working (and I don't see any other use of `assertTemplateUsed` in bedrock) is to use mock:

```python
from django.template import TemplateDoesNotExist
from django.test import RequestFactory, override_settings

from mock import ANY, patch

from lib.l10n_utils import render


@patch.object(env, 'loader', FileSystemLoader(TEMPLATE_DIRS))
@override_settings(ROOT=ROOT)
class TestLocaleTemplates(TestCase):
    def setUp(self):
        self.rf = RequestFactory()

    @patch('lib.l10n_utils.django_render')
    def test_l10n_render(self, mock_dj_render):
        # using a list here means that the first call will raise the exception, and 2nd call
        # will return True. You could add another exception class to the list and it would mean
        # the 2nd call would also raise if you wanted to test it falling all the way through.
        mock_dj_render.side_effect = [TemplateDoesNotExist, True]
        request = self.rf.get('/')
        request.locale = 'de'
        render(request, 'firefox/new.html')
        mock_dj_render.assert_called_with(request, 'firefox/new.html', ANY)
```
I realize this seems a bit crazier, but it is more of a unit test and should be faster and you don't really need those test files if this works.

lib/l10n_utils/__init__.py
jpetto: :+1:

lib/l10n_utils/__init__.py
```diff
@@ -4,6 +4,8 @@
 
 import re
```
pmclanahan: NIT: Unnecessary blank line here. All stdlib imports can be together.

pmclanahan:
Hmm. This looks good, but tests failing. Looks like an unapplied mock. Likely due to `lib.l10n_utils` having been imported already. Might try...

```python
from lib import l10n_utils

@patch.object(l10n_utils, 'django_render')
class ...
```

You'd have to change calls to `render` to `l10n_utils.render`, but that'd be okay.

lib/l10n_utils/tests/test_template.py
```diff
@@ -125,3 +127,53 @@ def test_render_no_locale(self, django_render, get_lang_path):
         # Note: no .locale on request
         # Should not cause an exception
         render(request, '500.html')
+
+
+@patch.object(l10n_utils, 'django_render')
```
pmclanahan: We also have to patch `template_is_active`.

```python
@patch('lib.l10n_utils.template_is_active', Mock(return_value=True))
@patch('lib.l10n_utils.django_render')
```
That set of patches works for me.

lib/l10n_utils/tests/test_template.py
```diff
@@ -125,3 +127,53 @@ def test_render_no_locale(self, django_render, get_lang_path):
...
+        request.locale = 'es-ES'
+        render(request, 'firefox/new.html')
+        django_render.assert_called_with(request, 'firefox/new.es-ES.html', ANY)
+
+    def test_l10n_render(self, django_render):
+        """
+        Non en-US requests with an l10n template should render the l10n
+        template.
+        """
+        django_render.side_effect = [True]
```
pmclanahan: Don't need this line.

lib/l10n_utils/tests/test_template.py
```diff
@@ -14,6 +15,7 @@
 from nose.tools import eq_, ok_
 from pyquery import PyQuery as pq
 
+from lib import l10n_utils
```
pmclanahan: No longer need this line if you switch to how I have it below.

pmclanahan:
Yay! I think we're good here. Let me check one more time.

PR #3882, opened 2016-02-19
Bug 1244200 - update privacy link in newsletter forms

craigcook:
Changes "privacy policy" to "privacy notice" and updates the link.

@flodolo and @pmclanahan - This changes some strings but I'm unsure how to use an l10n tag in a .py file, or if that's even possible. I was unable to find any examples; I suspect that helper is just for templates.

### DIFF
--- a/bedrock/mozorg/forms.py
+++ b/bedrock/mozorg/forms.py
@@ -58,13 +58,13 @@ class PrivacyWidget(widgets.CheckboxInput):
         input_txt = super(PrivacyWidget, self).render(name, value, attrs)
 
         policy_txt = _(u'I’m okay with Mozilla handling my info as explained '
-                       u'in <a href="%s">this Privacy Policy</a>')
+                       u'in <a href="%s">this Privacy Notice</a>')
         return mark_safe(
             '<label for="%s" class="privacy-check-label">'
             '%s '
             '<span class="title">%s</span></label>'
             % (attrs['id'], input_txt,
-               policy_txt % reverse('privacy'))
+               policy_txt % reverse('privacy.notices.websites'))
         )
 
 
--- a/bedrock/newsletter/views.py
+++ b/bedrock/newsletter/views.py
@@ -610,7 +610,7 @@ def newsletter_subscribe(request):
             if 'email' in form.errors:
                 errors.append(_('Please enter a valid email address'))
             if 'privacy' in form.errors:
-                errors.append(_('You must agree to the privacy policy'))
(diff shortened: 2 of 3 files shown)


flodolo:
We can't, so this needs to wait until we get a decent coverage.

The alternative is to expose this only for English within a conditional, then flip the string for l10n when it's translated in enough locales. FYI, I'm not convinced this would affect the existing localizations.

bedrock/newsletter/views.py
```diff
@@ -610,7 +610,7 @@ def newsletter_subscribe(request):
             if 'email' in form.errors:
                 errors.append(_('Please enter a valid email address'))
             if 'privacy' in form.errors:
-                errors.append(_('You must agree to the privacy policy'))
+                errors.append(_('You must agree to the privacy notice'))
```
pmclanahan: We could do an l10n tag here since we have access to the `request` object and thus can know what locale we're rendering.

bedrock/mozorg/forms.py
```diff
@@ -58,13 +58,13 @@ def render(self, name, value, attrs=None):
         input_txt = super(PrivacyWidget, self).render(name, value, attrs)
 
         policy_txt = _(u'I’m okay with Mozilla handling my info as explained '
-                       u'in <a href="%s">this Privacy Policy</a>')
+                       u'in <a href="%s">this Privacy Notice</a>')
```
pmclanahan: This one would be pretty tough to do, but I think it's possible. There are some global variables in Django that tell us some l10n info that we could use. The other issue here is that we also don't know what template is being used, and therefore don't know which lang file to check for being active.

craigcook:
I've asked Jessilyn in the bug if this can wait a few weeks for l10n or if we need to get it to prod ASAP. Will update here once I get an answer but I'm hoping it'll be OK to wait.

craigcook:
Got confirmation in the bug that this can wait to merge until strings are localized, so no need for any tagging hackery. @flodolo if the code changes meet your approval you can extract/expose strings.

flodolo:
Didn't get any update, but I just saw the reply in the bug.

I'll expose the strings since I need to push out other urgent requests today.

craigcook:
Ahoy @flodolo! Any status update on this? Do we have enough coverage to merge?

flodolo:
Weird, I didn't get this notification in my mailbox, luckily I was checking l10n tagged PRs.

Yes, we're good to merge here.

schalkneethling:
r+ :smile:

PR #3884, opened 2016-02-19
Fix Bug 1249419 - Update links to Firefox Friends

kyoshino:
These pages are not localized.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/contribute/studentambassadors/landing.html
+++ b/bedrock/mozorg/templates/mozorg/contribute/studentambassadors/landing.html
@@ -51,7 +51,7 @@
         <section id="sec-what">
           <h2>{{ _('What We Do') }}</h2>
           <ul>
-            <li>{{ _('<a href="%s">Promote Firefox</a> for Desktop and Android')|format(url('mozorg.contribute.friends')) }}</li>
+            <li>{{ _('<a href="%s">Support Firefox</a> for Desktop and Android')|format(url('mozorg.contribute.friends')) }}</li>
             <li>{{ _('Support the successful launch of <a href="%s">Firefox&nbsp;OS</a> in communities (and campuses!) around the world')|format(url('firefox.os.index')) }}</li>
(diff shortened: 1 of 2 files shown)


bedrock/mozorg/templates/mozorg/contribute/studentambassadors/landing.html
```diff
@@ -51,7 +51,7 @@
         <section id="sec-what">
           <h2>{{ _('What We Do') }}</h2>
           <ul>
-            <li>{{ _('<a href="%s">Promote Firefox</a> for Desktop and Android')|format(url('mozorg.contribute.friends')) }}</li>
+            <li>{{ _('<a href="%s">Support Firefox</a> for Desktop and Android')|format(url('mozorg.contribute.friends')) }}</li>
```
craigcook: Since we're editing this line anyway, should we add iOS? "Support Firefox for desktop, Android and iOS" ?

bedrock/mozorg/templates/mozorg/contribute/studentambassadors/landing.html
```diff
@@ -51,7 +51,7 @@
...
             <li>{{ _('Support the successful launch of <a href="%s">Firefox&nbsp;OS</a> in communities (and campuses!) around the world')|format(url('firefox.os.index')) }}</li>
```
craigcook: Should we remove this?

craigcook:
Not specifically related to the Friends links bug but I see a few other little content edits we could make if we want to. But that's perhaps out of scope (especially since it might snowball into a lot of other changes if we go asking for input from stakeholders) so maybe best to leave it alone. I'm happy to merge it with no further edits. r+wc :fried_shrimp: 

Also, these are string changes but I don't think either of these pages is translated so it should have no impact. @flodolo can you confirm?

craigcook:
Oh and of course it says right at the top they're not localized, that's enough confirmation for me :)

flodolo:
Confirmed, we're not translating any of these.

kyoshino:
Yeah, I'll file a new bug for copy updates.

PR #3885, opened 2016-02-19
[bug 1239863] Remove Firefox OS from products page

alexgibson:
## Description
- Removes Firefox OS from the `/firefox/products` page.
- Firefox for iOS has moved into center spot, to make the layout more visually balanced.
- Also cleans up some l10n conditional tags, which means we can also remove Marketplace from the template. I checked with @flodolo and this is ok to do given locales that have active translations.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1239863

## Testing
- Page should now display three products horizontally on wide viewports, and stack vertically on small viewports.
- Page should still look ok in RTL.

## Checklist
- [x] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/family/index.html
+++ b/bedrock/firefox/templates/firefox/family/index.html
@@ -85,7 +85,7 @@
   </div>{#-- /.content --#}
 </header>{#-- /#header --#}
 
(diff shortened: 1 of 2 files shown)


alexgibson:
@craigcook r?

craigcook:
r+ :grapes:

PR #3886, opened 2016-02-19
[bug 1239863] Hide Firefox OS on Family navigation

alexgibson:
## Description
- Hides Firefox OS on Family Navigation menu
- Only shows Firefox OS in the navigation should the user already be on a Firefox OS page.

Note: this is a temporary hack, until we no longer have the /firefox/os and /firefox/os/ devices pages. I still need to get clarification if this is the correct solution, so marking as do-not-merge for now.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1239863

## Testing
Firefox OS primary link should not be displayed on any pages aside from the Firefox OS pages themselves. This should apply to both desktop and mobile viewports.

## Checklist
- [ ] Requires l10n changes.


craigcook:
r+! :tiger:

PR #3887, opened 2016-02-19
[fix bug 1249752] Test FxA email field on /firefox/sync/.

jpetto:
## Description
Add an email field to /firefox/sync/ for Fx users *not* signed in to Sync. This field will propagate to the FxA email field on /firefox/accounts/.

## Bugzilla
[bug 1249752](https://bugzilla.mozilla.org/show_bug.cgi?id=1249752)

## Testing
Must have a [profile configured for local or demo testing](http://bedrock.readthedocs.org/en/latest/firefox-accounts.html) and be signed out of Sync. Hit /firefox/sync/, enter an email, click the button, and notice the email field in the FxA iframe is populated.

Variations are currently on demo5 - [v2](https://www-demo5.allizom.org/en-US/firefox/sync/?v=2) and [v3](https://www-demo5.allizom.org/en-US/firefox/sync/?v=3).

Variations are only running for en-US, so no L10n impact.

Waiting on GA confirmation, but code should be ready for review.

### DIFF
--- a/bedrock/firefox/templates/firefox/sync.html
+++ b/bedrock/firefox/templates/firefox/sync.html
@@ -74,15 +74,20 @@
 
       <div class="buttons">
         <div class="show-fx-31-signed-in">
+        {% if l10n_has_tag('firefox_sync_mobile_buttons_update') %}
+          <p>
+            {{ _('Now sync Firefox with your iOS and Android devices.') }}
+          </p>
+        {% endif %}
           <ul class="primary-buttons">
             <li>
-              <a href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}" class="sync-button green" data-interaction="button click" data-element-action="Sync CTA" id="cta-android">
-                {{ _('Get Firefox for Android') }}
+              <a href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}" data-interaction="button click" data-element-action="Sync CTA" id="cta-android">
+                {{ high_res_img('firefox/android/btn-google-play.png', {'alt': _('Get it on Google Play'), 'width': '152', 'height': '45', 'l10n': True}) }}
               </a>
             </li>
             <li>
-              <a href="{{ firefox_ios_url('mozorg-sync_page-appstore-button') }}" class="sync-button green" data-interaction="button click" data-element-action="Sync CTA" id="cta-ios">
-                {{ _('Get Firefox for iOS') }}
+              <a href="{{ firefox_ios_url('mozorg-sync_page-appstore-button') }}" data-interaction="button click" data-element-action="Sync CTA" id="cta-ios">
+                <img src="{{ l10n_img('firefox/ios/btn-app-store.svg') }}" alt="{{ _('Download on the App Store') }}" width="152" height="45">
               </a>
             </li>
           </ul>
@@ -94,9 +99,30 @@
         </div>
 
         <div class="show-fx-31-signed-out">
+        {% if not version %}
           <button class="sync-button" data-interaction="button click" data-element-action="Sync CTA"  id="cta-sync">
             {{ _('Get started with Sync') }}
           </button>
+        {% else %}
+          <form id="fxa-email-form" action="{{ url('firefox.accounts') }}" method="get">
+            <input type="hidden" id="variation" name="v" value="{{ version }}">
+            <label for="fxa-email" class="fxa-email">
+            {% if version == '2' %}
+              {{ _('Enter your email to begin.') }}
+            {% elif version == '3' %}
+              {{ _('Create a Firefox account to begin.') }}
+            {% endif %}
+            </label>
+            <input type="email" id="fxa-email" class="fxa-email" placeholder="{{ _('YOUR EMAIL HERE') }}">
+            <button type="submit" class="sync-button" data-interaction="button click" data-element-action="Sync CTA" id="cta-sync-variation">
+            {% if version == '2' %}
+              {{ _('Get started with Sync') }}
+            {% elif version == '3' %}
+              {{ _('Set up Sync') }}
+            {% endif %}
+            </button>
+          </form>
+        {% endif %}
         </div>
 
         <div class="show-fx-29-30">
--- a/media/css/firefox/sync.less
+++ b/media/css/firefox/sync.less
@@ -48,6 +48,8 @@
     .font-size(@largeFontSize);
     text-transform: uppercase;
     .transition();
+    width: 60%;
+    margin: 0 auto;
     &:focus,
     &:active,
     &:hover {
@@ -67,7 +69,6 @@
 }
 
 .sky button.sync-button {
-    width: 100%;
     border: none;
     cursor: pointer;
 }
@@ -95,16 +96,24 @@
         padding-bottom: @baseLine * 2;
     }
     .buttons {
-        .span(8);
-        .offset(2);
+        .span(10);
+        .offset(1);
+        float: none;
         clear: both;
 
         > div {
             padding: 0 (@baseLine * 5) @baseLine;
         }
 
-        > div.show-fx-31-signed-in {
+        > div.show-fx-31-signed-in,
+        > div.show-fx-31-signed-out {
             padding: 0 @baseLine @baseLine;
+        }
+
+        > .show-fx-31-signed-in p {
+            .font-size(18px);
+            .open-sans-light();
+            text-align: center;
         }
     }
     .warning {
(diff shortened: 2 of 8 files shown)


media/js/firefox/accounts.js
```diff
@@ -20,6 +22,11 @@
         fxaIframeSrc = fxaIframeSrc.replace('context=iframe', 'context=fx_firstrun_v2');
     }
 
+    // if email address provided in querystring, send to FxA frame
+    if (params.get('fxa-email')) {
+        fxaIframeSrc += '&email=' + params.get('fxa-email');
```
jpetto: @shane-tomlinson - I just took a guess at this, but it appears to work locally and on [demo5](https://www-demo5.allizom.org/en-US/firefox/sync/). Is this official, or as yet unfinished/in development?

shane-tomlinson: @jpetto - [email is an officially supported parameter](https://github.com/mozilla/fxa-content-server/blob/master/docs/query-params.md#email? Before I say "go ahead!" what's feature/goal?

jpetto: @shane-tomlinson - We're planning to add an email address field to [/firefox/sync/](http://mozilla.org/firefox/sync) for Fx users that are *not* signed in to Sync. If they fill out the field and click the button, they'll be taken to [/firefox/accounts/](https://www.mozilla.org/firefox/accounts/) and have their email address populated in the FxA iframe.

shane-tomlinson: >  If they fill out the field and click the button, they'll be taken to /firefox/accounts/ and have their email address populated in the FxA iframe.

:shipit:

bedrock/firefox/templates/firefox/sync.html
```diff
@@ -94,9 +94,30 @@
         </div>
 
         <div class="show-fx-31-signed-out">
+        {% if not version %}
           <button class="sync-button" data-interaction="button click" data-element-action="Sync CTA"  id="cta-sync">
             {{ _('Get started with Sync') }}
           </button>
+        {% else %}
+          <form id="fxa-email-form" action="{{ url('firefox.accounts') }}" method="get">
+            <input type="hidden" id="variation" name="v" value="{{ version }}">
```
alexgibson: I guess as long as we don't have any other tests running on `firefox/accounts` this will be fine, but we would need to be careful not to cross pollute tests while this test is running?

media/js/firefox/accounts.js
```diff
@@ -20,6 +24,18 @@
         fxaIframeSrc = fxaIframeSrc.replace('context=iframe', 'context=fx_firstrun_v2');
     }
 
+    // check for fxaEmail in sessionStorage or URL params
+    try {
+        fxaEmail = sessionStorage.getItem('fxa-email');
+    } catch(ex) {
+        fxaEmail = params.get('fxa-email');
```
alexgibson: It makes me a little nervous that we're potentially passing user email addresses in the URL here. Under what circumstances does this happen, only if `sessionStorage` fails? Given that this page is a public URL which is also soon to be indexed by search engines, even if someone shares the URL on twitter without realizing their email is in the URL could be bad?

Do we also need to think about validating the email address, to try and prevent any potential XSS vectors?

media/css/firefox/sync.less
```diff
@@ -139,13 +142,39 @@
...
+        flex-basis: 100%;
+        margin-bottom: @baseLine;
         text-align: center;
+        .font-size(18px);
+        .open-sans-light();
+    }
+
+    button,
+    input {
+        flex: 1;
```
alexgibson: Use `.flex(1)` here for consistency?

alexgibson:
This is more of a UX / copy thing, but with the new button placement it seems slightly less clear to me what the page is telling me to download from the app store? ("Firefox", "Sync"?).

sync-buttons

alexgibson:
I don't believe this is any way related to this PR, but I currently can't seem to test that the email gets populated correctly, due to stomlinson's local dev instance currently returning a 500 error?

alexgibson:
Tested on demo5 and can confirm that the email address is populated correctly :+1:

bedrock/firefox/templates/firefox/sync.html
jpetto: Yeah, coordinating tests is still a manual process. At least it's pretty clear in Optimizely which tests are currently running. :confused:

media/js/firefox/accounts.js
jpetto: Hm, agree that we should do more to ensure the email address doesn't make it into the URL. I can check for `sessionStorage` and only show the email field if it's available.

Will also update to add a bit of validation on the email address when pulling from `sessionStorage`.

media/css/firefox/sync.less
jpetto: Ah, will do.

media/js/firefox/accounts.js
```diff
@@ -20,6 +24,18 @@
         fxaIframeSrc = fxaIframeSrc.replace('context=iframe', 'context=fx_firstrun_v2');
     }
 
+    // check for fxaEmail in sessionStorage or URL params
+    try {
+        fxaEmail = sessionStorage.getItem('fxa-email');
```
alexgibson: Should we also remove the item in `sessionStorage` at this point?

jpetto: Most definitely.

media/js/firefox/accounts.js
alexgibson: Sounds good :+1: I would favor relying on session storage as a hard-dependency, rather than falling back to reading via query param.

jpetto: Yep, agreed.

jpetto:
@habber - Care to chime in on the [UX comment above](#issuecomment-190689161)?

habber:
I agree that for a Firefox desktop user that is already signed into sync, another line of copy to support the app store buttons is ideal. I don't think it should block the PR and related testing we had planned. I'm looking for an existing string that we can use, but if I can not find one, will discuss with Jen about requesting copy from Troy.

habber:
Existing strings that I have found are more generic. They don't address mobile, or imply/recognize that the user is currently using Firefox desktop. (ie: "Download Firefox to start using Sync.")

Ideally we would have a line of copy similar to the following to support the buttons: 
"Get Firefox for mobile to get the most out of Sync" or "Now sync Firefox Desktop with Firefox for iOS or Android."

habber:
Update: Jen approves of adding the second suggestion above for this state of the page. 

Below "You're signed up and ready to access your Firefox anywhere, anytime." and above buttons, place the following line of text:

"Now sync Firefox with your iOS and Android devices."

Is there a font style that we typically use for this type of 3rd level messaging that would work here?

jpetto:
@flodolo - Adding a new string here behind tag `firefox_sync_mobile_buttons_update`. Any issues from your end?

flodolo:
No issue, but it will have to wait for a while before being exposed since I've pushed out new strings 12 hours ago.

media/js/firefox/accounts.js
```diff
@@ -20,6 +24,17 @@
...
+        sessionStorage.removeItem('fxa-email');
+    } catch(ex) {}
+
+    // if email address provided, send to FxA frame
+    if (fxaEmail) {
```
alexgibson: Should we also add the same basic email validation here on this side, just for good measure?

media/css/firefox/sync.less
```diff
@@ -95,17 +96,25 @@
...
         }
 
-        > div.show-fx-31-signed-in {
+        > div.show-fx-31-signed-in,
+        > div.show-fx-31-signed-out {
             padding: 0 @baseLine @baseLine;
         }
+
+        > .show-fx-31-signed-in p {
+            .font-size(1.125rem);
```
alexgibson: This is a nit, but our code is starting to jump around from using `px` to `rem` everywhere. I don't mind what unit we use, but I think we should remain consistent for our own sanity?

alexgibson:
Code changes look good, as does the new line of copy r+wc from me :+1:

media/js/firefox/accounts.js
jpetto: Yeah, couldn't hurt.

media/css/firefox/sync.less
jpetto: Ah, yeah, my bad. Took the value from edits in dev tools. Will convert to `px`.

jpetto:
Fixed and squarshed!

PR #3889, opened 2016-02-22
Fix Bug 1249892, filter out undefined entries in the response array

schalkneethling:
@alexgibson r?


alexgibson:
Code looks good, but is there anything else I should be looking for here? Not sure how to actually test this?

schalkneethling:
@alexgibson not really, the real problem exists in the database so, basically if the plugincheck page works for you, this effectively solves the problem of bad data breaking the page.

alexgibson:
I didn't realize the page was broken for everyone :/

Sorry I would have merged this sooner if I understood r+ :+1:

PR #3891, opened 2016-02-22
Add functional tests for /firefox/channel/ page

alexgibson:
## Description
Adds functional tests for the `/firefox/channel/` page.

## Testing
One thing I'm unsure on is if it's safe to use the normal `wait_for_page_to_load()` here when testing URL's that contain hashes (e.g. `#developer`), or whether it would be safer to use a custom method which can check for the correct HTML class to be added to the content before doing an assert. Tests seem to pass OK without anything extra, but I'm unsure if it's adequate enough.

## Checklist
- [ ] Requires l10n changes.


alexgibson:
@davehunt r?

davehunt:
r+

PR #3893, opened 2016-02-22
fixes bug1250519 - Remove BSD newsletters subscribe

alicoding:
This patch basically replace the BSD newsletters and replace it with Basket and it is a part of this sec-bug on bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1172325

### DIFF
--- a/bedrock/mozorg/tests/test_views.py
+++ b/bedrock/mozorg/tests/test_views.py
@@ -17,7 +17,6 @@ from django.test.utils import override_settings
 from captcha.fields import ReCaptchaField
 from bedrock.base.urlresolvers import reverse
 from jinja2.exceptions import TemplateNotFound
-from requests.exceptions import Timeout
 from mock import ANY, Mock, patch
 from nose.tools import assert_false, eq_, ok_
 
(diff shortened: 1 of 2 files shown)


alicoding:
I'm not sure why the test is failing here?

jgmize:
> I'm not sure why the test is failing here?

In [the travis build](https://travis-ci.org/mozilla/bedrock/builds/111057467) there is a flake8 failure:

`bedrock/mozorg/email_contribute.py:11:1: F401 'requests' imported but unused`

You removed the only use of requests in the file, and now just need to remove the `import requests` on [line 11](https://github.com/mozilla/bedrock/pull/3893/files#diff-101afe46fec28b6d3d4e493d5cc2b368L11).

alicoding:
@jgmize thanks! I just removed that line, and will see if travis will be happy this time :)

alicoding:
Also filed a bug on bugzilla https://bugzilla.mozilla.org/show_bug.cgi?id=1250519

bedrock/mozorg/tests/test_views.py
```diff
@@ -432,10 +432,8 @@ def test_webmaker_mentor_signup(self, mock_post, mock_subscribe):
         self.data.update(interest='education', newsletter=True)
         self.client.post(self.url_en, self.data)
 
-        assert_false(mock_subscribe.called)
-        payload = {'email': self.contact, 'custom-1788': '1'}
-        mock_post.assert_called_with('https://sendto.mozilla.org/page/s/mentor-signup',
-                                     data=payload, timeout=2)
+        mock_subscribe.assert_called_with(self.contact, 'mozilla-learning-network', source_url=ANY)
+        assert_false(mock_post.called)
```
jpetto: IMO, there's no longer a need for these `assert_false(mock_post.called)` assertions (or associated patched objects). Before this change, `email_contribute.py` could perform either a `post` or a `subscribe`, and part of these tests was to make sure the correct action was taken.

With this change, all actions are `subscribe`. The variation to test now seems to be the newsletter name/ID, which is already tested in the `mock_subscribe.assert_called_with` assertion.

Unless @jgmize or @pmac disagree, I'd say you can just remove the `assert_false(mock_post.called)` assertions and the associated patches.

alicoding:
I've updated my patch without `requests` and it seems to be passing now, but not sure if what I did was correct.

alicoding:
Tests are now finally passing O_o

bedrock/mozorg/tests/test_views.py
```diff
@@ -583,49 +571,38 @@ def test_with_autoresponse(self):
...
         self.client.post(self.url_en, self.data)
 
         mock_subscribe.assert_called_with(self.contact, 'about-mozilla', source_url=ANY)
-        assert_false(mock_post.called)
 
     @patch.object(ReCaptchaField, 'clean', Mock())
     @patch('bedrock.mozorg.email_contribute.basket.subscribe')
-    @patch('bedrock.mozorg.email_contribute.requests.post')
-    def test_webmaker_mentor_signup_timeout_fail(self, mock_post, mock_subscribe):
+    def test_webmaker_mentor_signup_timeout_fail(self, mock_subscribe):
```
jpetto: This test could now use some updating. As-is, it's no longer testing a timeout. It would be great if we could test a basket timeout, though I'm not sure exactly how to go about it. @pmclanahan may be able to offer a suggestion.

alicoding:
It will be very helpful if any one of you can take over this patch and help get this merge since I don't have the capacity to work on it.

pmclanahan:
Closing in favor of #3932. Thanks @alicoding!

PR #3895, opened 2016-02-23
Fix Bug 1249610, add new infobar to bedrock

schalkneethling:
Couple of notes:

The code for the new infobar lives at:
https://github.com/mozilla/mozilla-infobar/

I am mentioning that, because the code has thus already been reviewed. With that said though, I have made some changes based on bugs and edge cases I ran into whilst integrating it into Bedrock so, looking over the code during this pull request works for me.

Next thing, I have spot checked a bunch of pages (I will be sharing a Google Sheet with the details).

I tested the translation bar using en-US, de and fr. I switched locales using about:config in Fx, as I cannot find an addon that actually works.

I did not test this in any other browsers. The only bar that should appear in these will be the Translation Bar. I did test that the update bar does not show.

As with Fx, I cannot find a language switcher that works as advertised. I did attempt to adjust the language setting using Language and input settings in Chrome but, this does not have the desired effect. Any suggestions here are welcome.

@alexgibson @craigcook @jpetto @kyoshino

### DIFF
--- a/media/js/infobar/infobar.js
+++ b/media/js/infobar/infobar.js
@@ -187,25 +187,17 @@
      */
     InfoBar.prototype.getAcceptLangs = function(userLangs) {
         var userAcceptLangs;
-        userLangs = userLangs || navigator.languages || navigator.browserLanguage;
-
         // Note that navigator.language doesn't always work because it's just
         // the application's locale on some browsers. navigator.languages has
         // not been widely implemented yet, but the new property provides an
         // array of the user's accept languages that we'd like to see.
         // navigator.languages are only supported from Firefox/Chrome 32 and
-        // not at all by IE, Safari, Opera. navigator.language is not supported
-        // by IE so, we need to fallback to navigator.browserLanguage
-        if (navigator.languages) {
-            // Normalize all language strings for easier comparison.
-            userAcceptLangs = $.map(userLangs, function (lang) {
-                return InfoBar.prototype.normalize(lang);
-            });
-        } else if (navigator.browserLanguage) {
-            // the above properties only return one item but we need
-            // this as an array in the getOfferedLang function.
-            userAcceptLangs = [userLangs];
-        }
+        // not at all by IE, Safari, Opera.
+        userLangs = userLangs || navigator.languages;
+
+        userAcceptLangs = $.map(userLangs, function (lang) {
+            return InfoBar.prototype.normalize(lang);
+        });
 
         return userAcceptLangs;
     };
@@ -330,10 +322,10 @@
         }
 
         if (config.opened) {
-            config.bar.toggleClass('hidden');
+            config.bar.toggleClass('infobar-hidden');
             config.bar.attr('aria-hidden', true);
         } else {
-            config.bar.toggleClass('hidden');
+            config.bar.toggleClass('infobar-hidden');
             config.bar.attr('aria-hidden', false);
         }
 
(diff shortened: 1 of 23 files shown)


media/js/infobar/infobar.js
```diff
@@ -0,0 +1,509 @@
...
+        $acceptButton: {},
+        $cancelButton: {},
+        // info bar(s) the user wish to use
+        bars: [],
+        // The current instance of an InfoBar
+        bar: {},
+        opened: false,
+        json: {},
+        pageLang: '',
+        // list of available languags from either a list of hreflang links
```
flodolo: typo (languages)

media/js/infobar/infobar.js
```diff
@@ -0,0 +1,509 @@
...
+
+    /**
+     * Returns selected language's BiDi layout.
+     * False = left-to-right layout
+     * True = right-to-left layout
+     * @param {sting} lang - A language string such as en-US
+     */
+    InfoBar.prototype.getLanguageBidi = function(lang) {
+        // Languages using BiDi (right-to-left) layout
+        var LANGUAGES_BIDI = ['he', 'ar', 'fa', 'ur'];
```
flodolo: I would order this...

media/js/infobar/infobar.js
```diff
@@ -0,0 +1,509 @@
...
+     * @param {string} lang - The language string to be normalized
+     */
+    InfoBar.prototype.normalize = function (lang) {
+        return lang.replace(/^(\w+)(?:-(\w+))?$/, function (m, p1, p2) {
+            return p1.toLowerCase() + ((p2) ? '-' + p2.toUpperCase() : '');
+        });
+    };
+
+    /**
+     * Gets and sets the list of available languags from either a list of hreflang links
```
flodolo: Same typo

media/js/infobar/infobar.js
```diff
@@ -0,0 +1,509 @@
...
+            });
+            return availableLangs;
+        } else {
+            // If those lists cannot be found, there is nothing to do
+            return false;
+        }
+    };
+
+    /**
+     * Determines whether the users accept language(s) match the current
```
flodolo: user's accept language list matches

media/js/infobar/infobar.js
```diff
@@ -0,0 +1,509 @@
...
+                offeredLang = userLang;
+            }
+        });
+
+        // If there is no offered language return false, else true
+        return !offeredLang ? false : offeredLang;
+    };
+
+    /**
+     * Get's the list of the user's acceptLanguages as appropriate for the
```
flodolo: Get

media/js/infobar/infobar.js
```diff
@@ -0,0 +1,509 @@
...
+     */
+    InfoBar.prototype.isLikeFirefox = function(userAgent) {
+        var ua = userAgent || navigator.userAgent;
+        return (/Iceweasel/i).test(ua) || (/IceCat/i).test(ua) ||
+            (/SeaMonkey/i).test(ua) || (/Camino/i).test(ua) ||
+            (/like Firefox/i).test(ua);
+    };
+
+    /**
+     * Rturns true if this is a version of Firefox mobile excluding Firefox for iOS
```
flodolo: Returns

schalkneethling:
I have made an update to the code. It now determined the pageLang match as well as the offered language using only the first entry in the acceptLanguages array.

This means that, in Chrome, going to about:settings -> Language and input settings and adjusting your preferred language there, now works as expected.

schalkneethling:
I am working on adding more, porting, unit tests, and have found the following test to determine whether the user's and the page language match:

```
it('should return false if the user\'s language is the same as the page\'s language',
...
expect(setup(['ja', 'pt-PT', 'el', 'fr', 'en'], 'el')).toBeFalsy();
```

The above suggests that, if the page language is `el`, we have a match, because it is one of the entries in the accept languages array. This means, we should indeed be looping over all of the accept languages in the `userLangPageLangMatch` function for example.

Agree, disagree ;) Let me know, thx

jpetto:
@schalkneethling - I'm going to bring this up in our next sprint planning meeting. Is the work here primarily finished? Looks like we're waiting on UX approval in the bug. If the code is essentially complete (aside from a rebase), I'll suggest we get final approval on moving forward. If not, I'll get it properly prioritized (so it can stop staring us in the face at every PR triage).

Thanks!

schalkneethling:
@jpetto I believe the code is there. The main things that was still needed was more/better test coverage, manual testing of the integration into the site, and then there was some UX work that needed to be done.

jpetto:
Hmm, ok. Aside from the UX work, could you estimate about how much dev time remains? Feel free to be broad - estimating in half day increments would be just fine.

schalkneethling:
@jpetto I would say 3 days should be sufficient.

craigcook:
This pull request is almost a year old at this point and the future of the infobar is still uncertain. There's some current strategy forming around a revamped cross-site navbar that may take its place, as well as the new branding and design system coming soon that could render this even further out-of-date, so I don't think this PR can ever be merged. We're going to close it now. 

If you listen closely you'll hear the sound of a single bugle softly playing Taps in the distance 🎺

PR #3900, opened 2016-02-25
Bug 1249042 - testing copy/design of Firefox newsletter form

craigcook:
## Description
Creates a set of A/B tests on various Firefox pages to measure engagement with newsletter subscription forms with different designs and messaging. Testing will entail a redirect to alternate page templates via Optimizely with Google Analytics to track form submissions. [This test plan document](https://docs.google.com/document/d/1Q9QuRca6dDX1SdDYvVwPQQQ21I9wnLJIVZSp4JUNMRw/edit#heading=h.dda57fn2njtv) has some further detail on which variations appear on which page.

We're conducting this test targeting English only, so although these are new strings we should not expose them for l10n yet. Once we have some test data and see what messaging performs best we may settle on final strings for translation. Non-English locales viewing these pages should see either nothing at all (for pages that currently have no newsletter form) or the the standard "Get Firefox news" headline already translated.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1249042

## Testing
There are a total of six pages included in this round of testing. For each page, the base URL acts as the "A" variant and appending the query parameter `?v=b` to the URL renders the "B" variant. See the test plan doc for which headline appears on which page.

The desktop page is a bit special since the test form should only appear for people using the latest Firefox. Other browsers and older Firefoxes should see the old form plus a download button. No other pages do any special detection and show the same content to all browsers and versions.

## Checklist
- [ ] Requires l10n changes.

I haven't added any functional or integration tests because, honestly, I don't know how. Should probably try to learn that some day.


alexgibson:
@craigcook we don't need to add functional tests for things we're only running for a short time, such as an A/B test. We do however need to try and make sure the new variations don't cause any existing tests running in production to fail. I'll take a look for you and run the relevant tests.

craigcook:
@alexgibson I suspected as much, I was just following the template :octocat:

bedrock/firefox/templates/firefox/desktop/index-b.html
```diff
@@ -0,0 +1,31 @@
...
+        {% else %}
+          <h3>{{ _('Get Firefox news') }}</h3>
+        {% endif %}
+        </header>
+        {{ email_newsletter_form(title=' ') }}
+      </section>
+
+      <section class="original">
+        <div id="subscribe-wrapper">
+          {{ email_newsletter_form(title=_('Get Firefox news')) }}
```
alexgibson: Embedding two newsletters in the same template here is breaking the JS form submission, because it creates duplicate ID's on the same page. This is also causing [this functional test](https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/desktop/test_all.py#L24) to fail as a result of the duplicate ID's.

bedrock/firefox/templates/firefox/hello/index-2016-b.html
```diff
@@ -0,0 +1,22 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% extends "firefox/hello/index-2016.html" %}
```
alexgibson: This new Hello page is currently behind waffle, and won't be switched on until March 8. Should we be testing the old template instead? If we're switching templates in production mid-test, could this also effect results?

bedrock/firefox/templates/firefox/android/index-b.html
```diff
@@ -0,0 +1,43 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% extends "firefox/android/index.html" %}
+
+{% block email_form %}
+  <div id="subscribe-download-wrapper" class="ga-section" data-ga-label="Get Firefox news">
```
alexgibson: Do we need to repeat all of this template markup? As far as I can tell, only the newsletter title changes on this page. If so (I could be wrong), could we just create a new block in `index.html` that wraps only the newsletter marco, and overwrite that here instead?

alexgibson:
Thank you title message on the `/firefox/hello/` page needs contrast fixing.

hello-thankyou

alexgibson:
Thank you title on private browsing page still has text shadow and some rogue padding:

private-browsing-thank-you

alexgibson:
Just a few nits to tidy up, looks good :+1: 

Only problematic area is the `/desktop` page, which is having issues with the same form embedded twice on the page. Also an open question about the Hello page and waffle switches. Not sure what the plan is, but thought I should raise it in case folks weren't aware of the go-live date.

alexgibson:
I should add, except for the desktop page, all existing functional tests seem to be passing ok still :+1:

bedrock/firefox/templates/firefox/desktop/index-b.html
craigcook: Hmm you're right, of course. I knew repeating it was sloppy but hadn't thought about how it would break the ajax. I'll rework it to avoid the repetition.

bedrock/firefox/templates/firefox/hello/index-2016-b.html
craigcook: I think we're OK with this, but I'll confirm with Jen. Otherwise I can do the same thing for the current page as well, we'll just want to consider how redesigning the page in the middle of the test would impact data.

bedrock/firefox/templates/firefox/android/index-b.html
craigcook: Good idea, I'll make it so.

bedrock/firefox/templates/firefox/hello/index-2016-b.html
craigcook: Confirmed that we won't worry about testing on the Hello page until March 8, so we'll activate the optimizely snippet's waffle switch at the same time we activate the page's waffle switch.

craigcook:
@jpetto r?

jpetto:
Code looks good. :+1: 

Only nit I can find is the "this Privacy Policy" link is indistinguishable from the other text. I only noticed because I tried to click the label to check the box and got sent to a new page. I wouldn't call it a deal-breaker, but would be nice to give that text an underline if there's time.

craigcook:
Good call. Added `text-decoration: underline;`

craigcook:
Tests are set up in Optimizely and tagged with the bug number. We'll need to set a bunch of waffle switches to activate the snippets on each page:

`firefox-desktop-optimizely`
`firefox-android-optimizely`
`firefox-ios-optimizely`
`firefox-products-optimizely`
`firefox-private-browsing-optimizely`

jpetto:
The test for private browsing's Variation #1 didn't have the `?v=b` at the end of the redirect URL, so I added it. The rest of the tests look good.

bedrock/firefox/views.py
```diff
@@ -443,7 +443,13 @@ def hello(request):
     }
 
     if (waffle.switch_is_active('firefox-hello-2016')):
-        return l10n_utils.render(request, 'firefox/hello/index-2016.html')
+        variant = request.GET.get('v', '')
+        if variant in ['b']:
+            template = 'firefox/hello/index-2016-{0}.html'.format(variant)
```
jpetto: Not a blocker, but for future reference, could do this a bit simpler when there's only one available variant:

```
if variant == 'b':
    template = 'firefox/hello/index-2016-b.html'
```

craigcook: hmm, you're right of course... I started with the same code running the current plugincheck test, which has four variants, and for a while I believed there were going to be two for the desktop page so I kinda locked onto that notion and never thought to simplify it for a single variant, even for the others. 

Happy to update for cleanliness if you think it's worthwhile, but then maybe best not to go messing with it and risk further delay.

jpetto: Yeah, since the code is temporary and functional, I wouldn't worry about it.

jpetto:
Ready for merge as far as I'm concerned. We good to go?

PR #3906, opened 2016-02-25
Add Dockerfile for "git push deis" demos

jgmize:


pmclanahan:
Found an issue. When trying to use this for my django-csp experiments I couldn't get the app to run on deis. Finally figured out that it was because my new dep (django-csp) wasn't being installed because dep install doesn't happen in this build. We'll need to make that happen. Just repeating those commands should be pretty quick because they should mostly already be installed right? I'm experimenting with it now.

jgmize:
I've added a commit that installs the python requirements as well, which will add very little time if there are no new requirements and should work for new requirements automatically. However, for upgrades to existing packages in the base image we'll likely need to add a tweak to the Dockerfile to work around https://github.com/docker/docker/issues/12327 and if there are any new system requirements we'll need to explicitly add the appropriate "RUN apt-get" commands in the demo branch as well, and move it to the appropriate place in the base dockerfile for the actual PR.

pmclanahan:
This LGTM Josh. I think the issue of new debian packages will be far less frequent than new python or node ones. If my PR about using npm install happens we'll just have to add that step here as well, but it should be similarly quick.

r+

pmclanahan:
PS: I've not exactly tried this, but I did try a very similar change on my branch for the demo I pushed recently. So I'm pretty confident this'll work well.

jgmize:
Thanks @pmclanahan. I squashed my commits and deployed https://bedrock-demo-pr-3906.us-west.moz.works/en-US/ using this, so I can confirm this works. I'll follow up with docs and creds for the team as soon as I can.

PR #3910, opened 2016-02-26
Disable admin, auth, and sessions: fix bug 1251599

jgmize:


jgmize:
@pmclanahan r?

pmclanahan:
Looks like waffle requires the auth app to be in `INSTALLED_APPS`, but we can remove the rest.

pmclanahan:
Apart from that it looks good. r+wc.

jgmize:
Restored django.contrib.auth to INSTALLED_APPS

pmclanahan:
Tests passing. Thanks!

PR #3911, opened 2016-02-26
Replace obsolete tests for vary: user-agent

jgmize:
@pmclanahan r?

### DIFF
--- a/bedrock/firefox/redirects.py
+++ b/bedrock/firefox/redirects.py
@@ -152,7 +152,7 @@ redirectpatterns = (
     # Bug 979527
     redirect(r'^firefox/central/?$', is_firefox_redirector(
         'https://support.mozilla.org/kb/get-started-firefox-overview-main-features',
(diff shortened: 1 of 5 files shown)


pmclanahan:
To get those tests passing you have to have the custom vary header management middleware enabled. The issue is that we're now using these tests in multiple contexts. Maybe we should just enable said middleware at all times now? At the very least we need to change travis to have that middleware enabled so that these tests pass.

jgmize:
Thanks @pmclanahan, I added a commit to enable VaryNoCacheMiddleware by default.

jgmize:
> Thanks @pmclanahan, I added a commit to enable VaryNoCacheMiddleware by default.

... which naturally broke a couple of other tests that are expecting the previous default behavior. Working on a fix for that now.

pmclanahan:
LGTM. r+

Feel free to merge after tests pass and commits are squarshed.

:beers:

jgmize:
Tests finally pass; squashing.

PR #3916, opened 2016-03-01
[fix bug 1250225] Use srcset for high res images.

jpetto:
## Description
As `srcset` is now widely supported, we'd prefer to let browsers handle when to load high resolution images rather than using our own custom JS.

## Bugzilla
[bug 1250225](https://bugzilla.mozilla.org/show_bug.cgi?id=1250225)

## Testing
If possible, test on both standard and high-resolution screens. Also, check/run Jasmine tests.

## Checklist

### DIFF
--- a/media/js/base/mozilla-image-helper.js
+++ b/media/js/base/mozilla-image-helper.js
@@ -11,7 +11,6 @@
 
 $(document).ready(function() {
     Mozilla.ImageHelper.initPlatformImages();
-    Mozilla.ImageHelper.initHighResImages();
 });
 
 // create namespace
@@ -24,10 +23,7 @@ if (typeof Mozilla == 'undefined') {
 /**
  * ImageHelper object
  */
-Mozilla.ImageHelper = function() {
-};
-
-Mozilla.ImageHelper.is_high_dpi = null;
+Mozilla.ImageHelper = function() {};
 
 // }}}
 
@@ -38,8 +34,7 @@ Mozilla.ImageHelper.initPlatformImages = function() {
     $('.platform-img').each(function() {
         var $img = $(this);
         var data_attribute = 'src-';
-        var is_high_res = $img.data('high-res') && Mozilla.ImageHelper.isHighDpi();
-        var suffix;
+        var is_high_res = $img.data('high-res');
         var new_src;
 
         if (site.platform == 'oldwin') {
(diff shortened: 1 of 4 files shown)


jpetto:
@schalkneethling r?

media/js/base/mozilla-image-helper.js
```diff
@@ -38,8 +34,7 @@ Mozilla.ImageHelper.initPlatformImages = function() {
     $('.platform-img').each(function() {
         var $img = $(this);
         var data_attribute = 'src-';
-        var is_high_res = $img.data('high-res') && Mozilla.ImageHelper.isHighDpi();
-        var suffix;
+        var is_high_res = $img.data('high-res');
```
schalkneethling: Looking at the HTML that is now generated for images using the `high_res_img` helper, there seems to no longer be a `high-res` attribute set on the element. So, I guess we do not need this?

media/js/base/mozilla-image-helper.js
```diff
@@ -52,57 +47,24 @@ Mozilla.ImageHelper.initPlatformImages = function() {
...
-// {{{ initHighResImages()
-
-Mozilla.ImageHelper.initHighResImages = function() {
-    $('img[data-src][data-high-res="true"][data-processed="false"]').each(function() {
-        var $img = $(this);
-        var src = $img.data('src');
-        if (Mozilla.ImageHelper.isHighDpi()) {
-            src = $img.data('high-res-src');
+        // if high res requested and path provided, set srcset attribute
+        if (is_high_res && $img.data(data_attribute + '-high-res')) {
```
schalkneethling: I also do not see either of these in the HTML and, seeing that the `src` and `srcset` attributes are populated in the Jinja helper, do we need to set it in JS. Tested with and without JS enabled and it seems to work fine.

media/js/base/mozilla-image-helper.js
```diff
@@ -52,57 +47,24 @@ Mozilla.ImageHelper.initPlatformImages = function() {
...
+            $img.attr('srcset', $img.data(data_attribute + '-high-res') + ' 1.5x');
         }
-        this.src = src;
+
         $img.attr('data-processed', 'true');
```
schalkneethling: Perhaps we also do not need this line anymore?

schalkneethling:
Interestingly Firefox seems to always load the `high-res` image.

media/js/base/mozilla-image-helper.js
jpetto: We still need it when processing high res platform images. Platform images begin with an empty `src` attribute and no `srcset` attribute. We use JS to set `src` based on the current user's platform. If the platform image is marked with `data-high-res`, then the JS also adds/sets the `srcset` attribute.

media/js/base/mozilla-image-helper.js
jpetto: High res platform images still have the `data-src-{platform}-high-res` attributes, which are needed to correctly set/populate the `srcset` attribute.

media/js/base/mozilla-image-helper.js
jpetto: Ah, good catch. Since we no longer process "regular" (non-platform) high res images with JS, this is no longer needed.

media/js/base/mozilla-image-helper.js
schalkneethling: Right you are :+1:

jpetto:
Hm, I just did a spot test on `/firefox/choose/` and `/thunderbird/` in Dev Edition, Beta, and Release on both my high-res laptop and regular res external monitor and all load the expected images.

I did find a bug with platform images (setting `src` before `srcset`) that was causing Fx to download both images when on a high-res screen. Fixed in second commit (just issued).

schalkneethling:
@jpetto There is a test failing now after the last commit related to `initHighResImages`:
https://travis-ci.org/mozilla/bedrock/builds/112892417#L925

schalkneethling:
This is looking gooood :fireworks: r+ Just needs a squash.

jpetto:
Squarshed!

PR #3918, opened 2016-03-01
[fix bug 1251283] Update Hello FTU for Firefox 45

alexgibson:
## Description
Updates the Hello FTU page for Firefox 45 (Due March 8)

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1251283

## Testing
* Please test using **Firefox 45** and follow the setup instructions in the docs http://bedrock.readthedocs.org/en/latest/uitour.html#hello-ftu-uitour.
* When opening the Hello FTU, the Hello menu should automatically open if available in the toolbar.
* If the FTU page is loaded with the URL query param `?noopenpanel=1`, the Hello menu should not open (see [Bug 1250495](https://bugzilla.mozilla.org/show_bug.cgi?id=1250495) for context).
* Also up on demo1 using the instructions here: https://bugzilla.mozilla.org/show_bug.cgi?id=1251283#c9

## Checklist
- [x] Requires l10n changes.
- [ ] Related functional & integration tests passing (none for this page)


alexgibson:
Opening this now so we can get l10n started cc @flodolo.

flodolo:
@alexgibson Do you have a graphic mockup that I can share with localizers, since I doubt we'll put this on a demo server?

alexgibson:
@flodolo - added a screenshot in the bug https://bugzilla.mozilla.org/show_bug.cgi?id=1251283#c11

bedrock/firefox/templates/firefox/hello/start-45.html
```diff
@@ -0,0 +1,39 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/base-resp.html" %}
+
```
flodolo: Can we use a different lang file? I don't want to create a folder for such a short-lived page
```
{% add_lang_files "firefox/hello-start-45" %}
```

Double checked locally to verify that production would redirect correctly.

alexgibson: Lang file added, thanks

alexgibson:
Ok, opening this up for review - @schalkneethling r?

media/js/firefox/hello/start-45.js
```diff
@@ -0,0 +1,113 @@
...
+    if (client.isFirefoxDesktop) {
+        getAvailableTargets().then(function(targets) {
+            // if URL query param noopenpanel=1 exists then don't open the Hello menu.
+            var shouldOpenPanel = params.get('noopenpanel') === 1 ? false : true;
+            // Only open the Hello menu if target is in the user pallet.
+            if (shouldOpenPanel && targets && targets.indexOf('loop') > -1) {
+                openHelloMenu();
+                bindHelloEvents();
+
+                window.dataLayer.push({
```
schalkneethling: Could we use `trackInteraction('GetStarted');` here?

media/js/firefox/hello/start-45.js
```diff
@@ -0,0 +1,113 @@
...
+    function handleVisibilityChange() {
+        if (document.hidden) {
+            closeHelloMenu();
+            unbindHelloEvents();
+        } else {
+            bindHelloEvents();
+        }
+    }
+
+    if (client.isFirefoxDesktop) {
```
schalkneethling: Could we do:

```
// if URL query param noopenpanel=1 exists then don't open the Hello menu.
var shouldOpenPanel = params.get('noopenpanel') === 1 ? false : true;

if (client.isFirefoxDesktop && shouldOpenPanel) {
...
```

media/js/firefox/hello/start-45.js
alexgibson: Good suggestion :+1:

media/js/firefox/hello/start-45.js
alexgibson: Hmm.. I think I'd like to keep `shouldOpenPanel` in the function scope where the variable is used, since we only need to use it if Hello is in the menu.

schalkneethling: Ah ok. I was thinking that if we are either not on desktop or, the url param was passed to not open the Hello panel, we can basically bypass all the code inside the `if` statement, or would we still need to do call `getAvailableTargets`?

alexgibson: On second thought, I think this does make sense. Creates a variable in wider scope, but less code to execute :+1:

schalkneethling:
@alexgibson Functionality tested in Fx 45

[x] When opening the page at /firefox/hello it shows the static content and the door hanger is opened
[x] When opening the page at /firefox/hello/?noopenpanel=1 only static content is shown, no door hanger

Two open code items, other than that we're good. Thx r+wc

schalkneethling:
Also ran the Python tests and unit tests and all pass.

alexgibson:
@schalkneethling thanks for the review, updated. If all looks good, let me know and I'll squash. Thanks!

PR #3919, opened 2016-03-01
[fix bug 1252332] Redirect /sync to /firefox/sync

alexgibson:
## Description
Redirects `/sync/` to `/firefox/sync/`

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1252332

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/etc/httpd/global.conf
+++ b/etc/httpd/global.conf
@@ -760,3 +760,6 @@ RewriteRule ^(/(?:\w{2,3}(?:-\w{2})?/)?firefox/beta/feedbackprivacypolicy/?)$ /b
 
 # bug 1149019
 RewriteRule ^(/certs/.*)$ /b$1 [PT]
+
+# bug 1252332
+RewriteRule ^(/(?:\w{2,3}(?:-\w{2})?/)?sync/?)$ /b$1 [PT]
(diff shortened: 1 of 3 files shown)


jpetto:
Test passes & verified manually on local server. Pushed to test on demo5, and still get a 404. I think we need an entry in `global.conf` like `RewriteRule ^/(\w{2,3}(?:-\w{2})?/)?sync(/?)$ /b/$1sync/ [PT]`?

jpetto:
Added the rule mentioned above to `global.conf` and updated demo5 and no longer getting the 404.

alexgibson:
Bah, good catch - added a pass through in `global.conf`

etc/httpd/global.conf
```diff
@@ -760,3 +760,6 @@ RewriteRule ^(/(?:\w{2,3}(?:-\w{2})?/)?firefox/beta/feedbackprivacypolicy/?)$ /b
 
 # bug 1149019
 RewriteRule ^(/certs/.*)$ /b$1 [PT]
+
+# bug 1252332
+RewriteRule ^(/(?:\w{2,3}(?:-\w{2})?/)?sync/?)$ /b$1 [PT]
```
alexgibson: This is slightly different to the pass-through that @jpetto suggested (which also works), but it looks like a lot of re-write rules are now following this format.

@pmclanahan is this now the preferred format for adding a pass through in `global.conf`?

pmclanahan: This is what I was doing when converting old redirects to passthroughs just because it was easier :) Either way should be fine. This file is no longer in use in stage and prod, but will continue to affect demos and dev. I'll try soon to reduce this file to a single passthrough for the whole site and we can forget about it forever :fireworks:

jpetto:
It is time. r+ :shipit:

PR #3920, opened 2016-03-01
Docker Updates

glogiotatidis:
The idea is that we're re-using `DOCKER_IMAGE_TAG` in pre-squashed and squashed images and we end up using an unsquashed bedrock image.  This affects image size.

Took the opportunity to update other parts of the script too. I can git squash if that's preferred.

I tested this manually outside the pipeline and seems to be working.

### DIFF
--- a/docker/jenkins/build_image.sh
+++ b/docker/jenkins/build_image.sh
@@ -2,21 +2,18 @@
 #
 set -ex
 DOCKER_IMAGE_TAG=${DOCKER_REPOSITORY}:${GIT_COMMIT}
+TMP_DOCKER_TAG=${JOB_NAME}${BUILD_NUMBER}
 
 # If docker image exists and no force rebuild do nothing
 FORCE_REBUILD=`echo "$FORCE_REBUILD" | tr '[:upper:]' '[:lower:]'`
 if [[ $FORCE_REBUILD != "true" ]];
 then
-    if docker history -q $DOCKER_IMAGE_TAG 2> /dev/null;
+    if docker history -q $DOCKER_IMAGE_TAG > /dev/null;
     then
         echo "Docker image already exists, do nothing"
         exit 0;
     fi
 fi
-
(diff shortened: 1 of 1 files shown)


glogiotatidis:
@jgmize r?

jgmize:
Thanks @glogiotatidis, this lgtm. I'll take you up on your offer to squash the commits, and then we're good to merge :)

r+wc

glogiotatidis:
Squashed and rebased. Travis pass and I'm only changing jenkins build scripts so I'm ignoring CircleCI.

Thanks Josh.

glogiotatidis:
FYI Build [1064](https://ci.us-west.moz.works/job/bedrock_base_image/1064/) failed because the needed docker tags did not exists. I FORCE_BUILD-ed the job in [1066](https://ci.us-west.moz.works/job/bedrock_base_image/1066/] and fixed it

glogiotatidis:
Also updated bedrock_code_image and bedrock_l10n_image with `UPDATE_DOCKER_IMAGES=False` to compensate for the [default change](https://github.com/mozilla/bedrock/pull/3920/files#diff-ca6d150ef101826855c0929fc92b620fR25). This fixed failed build [946](https://ci.us-west.moz.works/job/bedrock_code_image/946/)

glogiotatidis:
So this looks to be working right. In the graph provided by [ImageLayers](https://imagelayers.io/?images=mozorg%2Fbedrock:6ef8ff5e6a106bca27413706c3fe649561e4ef7a,mozorg%2Fbedrock:21c970532fe7000ece1529f78fbe644a6062bd08) on the left column is the image built with the new script and on the right an older image. The size remains the same (expected) but the new image has one more layer which is the squashed `bedrock_base` layer. This will allow us to reuse this layer in image built from now on, which will reduce the build times (less squashing) and push/pull times to/from registries. Obviously less disk space too.

jgmize:
Nice, thanks @glogiotatidis!

PR #3923, opened 2016-03-02
Remove functional tests properties files

davehunt:
These have been moved to docker/jenkins/properties/functional_tests/ and have been split into per-push and per-tag configurations. Jenkins has now been updated so we can remove the old properties files.

@alexgibson @jgmize r?


alexgibson:
r+

PR #3926, opened 2016-03-02
Add CircleCI support

pmclanahan:
Also move from peep to pip8


pmclanahan:
@jgmize @alexgibson r?

pmclanahan:
Still need to cleanup the need for those CI specific settings files, but wanted to just get this working first.

bedrock/settings/__init__.py
```diff
@@ -6,8 +6,8 @@
 
 from .base import *  # noqa
 
-if os.getenv('TRAVIS', False):
-    from .travis import *  # noqa
+if os.getenv('CIRCLECI', False):
+    from .circleci import *  # noqa
```
jgmize: I'd prefer to eliminate circleci.py and use [environment variables](https://circleci.com/docs/environment-variables).

jgmize:
r+wc, nice work!

bedrock/settings/__init__.py
pmclanahan: I did. But I can probably go ahead and knock that out actually. I already did with the DB settings. I'll give it a go.

pmclanahan:
Went ahead and removed `circleci.py` and went a little crazy and converted from peep to pip8. I needed to do the pip8 thing anyway and wanted to base it on this work so I didn't have to also change it for travis.

mind giving that bit a look as well @jgmize ?

jgmize:
> Went ahead and removed circleci.py and went a little crazy and converted from peep to pip8. I needed to do the pip8 thing anyway and wanted to base it on this work so I didn't have to also change it for travis.

I understand wanting to base the pip8 work on the circleci, but wouldn't it have made more sense to go ahead and merge this PR after removing circleci.py and opened up a separate PR for the pip8 conversion? 

I've looked over the pip8 conversion and I don't see any issues with the changes themselves, and obviously it passes in circleci, but the commit has some changes to the legacy update scripts that can only be tested in the legacy environment. While we're *almost* done with the legacy deployments, we need to make sure not to break it until we have automatic l10n deploys to dev and everyone on the team can create and push demos in the new infra, so ideally I'd like to test the pip8 conversion in a legacy demo environment before merging. It looks like demo3 should be available; @pmclanahan would you mind testing the pip8 changes there when you get a chance?

pmclanahan:
That's a fair point. I'll split it out.

PR #3932, opened 2016-03-03
Fix bug 1250519: Remove BSD newsletter subscribe

pmclanahan:
This replaces PR #3893

### DIFF
--- a/bedrock/mozorg/tests/test_views.py
+++ b/bedrock/mozorg/tests/test_views.py
@@ -454,6 +454,7 @@ class TestContributeOldPage(TestCase):
     @patch('bedrock.mozorg.email_contribute.basket.subscribe')
     def test_webmaker_mentor_signup_timeout_fail(self, mock_subscribe):
         """Test Webmaker Mentor signup form when request times out"""
+        mock_subscribe.side_effect = basket.BasketException
         self.data.update(interest='education', newsletter=True)
         res = self.client.post(self.url_en, self.data)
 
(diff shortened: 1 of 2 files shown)


pmclanahan:
@jpetto r?

bedrock/mozorg/tests/test_views.py
```diff
@@ -426,49 +425,38 @@ def test_with_autoresponse(self):
...
         mock_subscribe.assert_called_with(self.contact, 'about-mozilla', source_url=ANY)
-        assert_false(mock_post.called)
 
     @patch.object(ReCaptchaField, 'clean', Mock())
     @patch('bedrock.mozorg.email_contribute.basket.subscribe')
-    @patch('bedrock.mozorg.email_contribute.requests.post')
-    def test_webmaker_mentor_signup_timeout_fail(self, mock_post, mock_subscribe):
+    def test_webmaker_mentor_signup_timeout_fail(self, mock_subscribe):
         """Test Webmaker Mentor signup form when request times out"""
-        mock_post.side_effect = Timeout('Timeout')
```
jpetto: Should this test also set a `side_effect` on `mock_subscribe` (as you did below)? It's not really testing any failures/timeouts as-is, is it?

jpetto:
Aside from one comment, looks good to me. r+wc

Just to double check, is the new `mozilla-learning-network` newsletter set up and ready for production?

bedrock/mozorg/tests/test_views.py
pmclanahan: Ah yes. Good catch.

pmclanahan: Updated.

pmclanahan:
@jpetto looking at the [JSON from basket](https://basket.mozilla.org/news/newsletters/) I see it in there. So that means that at least it does exist in basket and ExactTarget. Whether it's ready for prod or not I can't say, but the people in the bug seem ready so I think we're good to go.

jpetto:
Agreed. :shipit:

PR #3934, opened 2016-03-03
[fix bug 1250758] Add utm_ params to about:accounts.

jpetto:
## Description
To improve analytics on pages that open `about:accounts` using the `UITour.showFirefoxAccounts` method, we want to pass `utm_` params in to the privileged `about:accounts` tab.

## Bugzilla
[bug 1250758](https://bugzilla.mozilla.org/show_bug.cgi?id=1250758)

## Testing
~~**As of 2016-03-03, testing is broken in Nightly. This will be fixed when [bug 1253120](https://bugzilla.mozilla.org/show_bug.cgi?id=1253120) lands. (2016-03-04?)**~~

In Nightly and signed out of Sync, visit [`/firefox/sync/` on demo5](http://www-demo5.allizom.org/firefox/sync/) and click the "Get started with Sync" CTA. This should open `about:accounts` with two default `utm_` parameters added to the URL - `utm_campaign=page referral - not part of a campaign` (URL encoded) and `utm_content=/firefox/sync/`. Visiting [`/firefox/sync/?utm_foo=bar`](http://www-demo5.allizom.org/firefox/sync/?utm_foo=bar) and clicking the CTA should persist the `utm_foo=bar` query param to `about:accounts`.

## Checklist

### DIFF
--- a/docs/uitour.rst
+++ b/docs/uitour.rst
@@ -428,18 +428,43 @@ Passing ``defaultBrowser`` will set Firefox as the default web browser.
 
     ``setConfiguration('defaultBrowser')`` is only available in Firefox 40 onward.
 
-showFirefoxAccounts();
+showFirefoxAccounts(extraURLCampaignParams);
 ^^^^^^^^^^^^^^^^^^^^^^
 
-Allows a web page to navigate directly to ``about:accounts?action=signup``
+Allows a web page to navigate directly to
+``about:accounts?action=signup&entrypoint=uitour``. In Firefox 47 and beyond,
+optionally accepts an object of ``utm_*`` key/values, which will be encoded and
+appended to the ``about:accounts`` querystring.
+
+.. Important::
+
+    All keys in ``extraURLCampaignParams`` must begin with ``utm_``. If an
+    invalid key is present, the call to ``showFirefoxAccounts`` will fail.
 
 .. code-block:: javascript
 
+    // no extra utm_ campaign params. will open
+    // about:accounts?action=signup&entrypoint=uitour
     Mozilla.UITour.showFirefoxAccounts();
+
+    // with extra utm_ campaign params. will open
+    // about:accounts?action=signup&entrypoint=uitour&utm_foo=bar&utm_bar=baz
+    Mozilla.UITour.showFirefoxAccounts({
+        'utm_foo': 'bar',
+        'utm_bar': 'baz'
+    });
 
 .. Important::
 
     ``showFirefoxAccounts()`` is only available in Firefox 31 onward.
+    ``extraURLCampaignParams`` parameter only functional in Firefox 47 onward.
+
+.. note::
+
+    A convenience method named ``utmParamsFxA`` exists in
+    ``js/base/search-params.js`` that pulls all ``utm_`` params from the current
+    page's URL and places them in an object (along with pre-defined defaults)
+    ready to pass to ``showFirefoxAccounts``.
 
 resetFirefox();
 ^^^^^^^^^^^^^^^
--- a/media/js/base/search-params.js
+++ b/media/js/base/search-params.js
@@ -42,3 +42,40 @@ _SearchParams.prototype.toString = function () {
         return [encodeURIComponent(key), encodeURIComponent(value)].join('=');
     }).join('&');
 };
+
+_SearchParams.prototype.utmParams = function() {
+    var utms = {};
+
+    $.each(this.params, function (key, val) {
+        if (key.indexOf('utm_') === 0) {
+            utms[key] = val;
+        }
+    });
+
+    return utms;
+};
+
+_SearchParams.prototype.utmParamsFxA = function(pathname) {
+    pathname = pathname || window.location.pathname || '';
+
+    var utms = this.utmParams();
+
+    // set to default value if not specified in URL
+    if (!utms.utm_campaign) {
+        // utm_* values will be encoded on the product side, so no need to
+        // pre-emptively encode here
+        utms.utm_campaign = 'page referral - not part of a campaign';
+    }
+
+    // remove locale from pathname and store result in utm_content
+    // e.g. https://www.mozilla.org/it/firefox/sync/?foo=bar should
+    // have utm_content value of /firefox/sync/.
+    var matches = pathname.match(/\/[\w-]+(\/.*)$/);
+
+    if (matches && matches.length > 1) {
+        // no need to encode - will be done on product side
+        utms.utm_content = matches[1];
+    }
+
+    return utms;
+};
--- a/media/js/firefox/australis/australis-uitour.js
+++ b/media/js/firefox/australis/australis-uitour.js
@@ -177,8 +177,21 @@ if (typeof Mozilla == 'undefined') {
         });
     };
 
-    Mozilla.UITour.showFirefoxAccounts = function() {
-        _sendEvent('showFirefoxAccounts');
+    /**
+    * Request the browser open the Firefox Accounts page.
+    *
+    * @param {Object} extraURLCampaignParams - An object containing additional
+    * paramaters for the URL opened by the browser for reasons of promotional
+    * campaign tracking. Each attribute of the object must have a name that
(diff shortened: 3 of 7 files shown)


jpetto:
@alexgibson - Would love a second opinion on approach. Can you think of a better place for `utm_` param filtering/processing that `search-params.js`? Aside from `australis-uitour.js` (which I feel should continue to mirror Mozilla Central), `search-params.js` seemed like the best choice.

It's feels bad (in a non-DRY way) to repeat `new _SearchParams`/`params.utmParamsFxA()` for each instance of `showFirefoxAccounts`, but I'm not sure making a wrapper function is better.

media/js/firefox/australis/australis-uitour.js
```diff
@@ -56,14 +56,71 @@ if (typeof Mozilla == 'undefined') {
...
+        }
+    };
+
     Mozilla.UITour.registerPageID = function(pageID) {
         _sendEvent('registerPageID', {
             pageID: pageID
         });
     };
 
+    Mozilla.UITour.showHeartbeat = function(message, thankyouMessage, flowId, engagementURL,
```
alexgibson: Instead of pasting the whole contents of this file from mozilla central, I would only add what we need. So far I've only been including stuff that has actively been used on bedrock (and we have documentation for). Things such as `showHeartBeat()` we're never intended to be used in the context of mozorg, but from the browser itself (it is called from an invisible tab I believe).

media/js/firefox/australis/australis-uitour.js
```diff
@@ -177,15 +234,28 @@ if (typeof Mozilla == 'undefined') {
         });
     };
 
-    Mozilla.UITour.showFirefoxAccounts = function() {
-        _sendEvent('showFirefoxAccounts');
+    /**
+     * Request the browser open the Firefox Accounts page.
+     *
+     * @param {Object} extraURLCampaignParams - An object containing additional
```
alexgibson: Until we have a single documented place for changes, I would suggest adding this to the bedrock docs, so we don't need to look in two places. I believe the long-term plan is to have docs generated directly from these comments in mozilla central. But until that's in place, I would keep our current source of documentation up to date.

media/js/base/search-params.js
```diff
@@ -42,3 +42,37 @@ _SearchParams.prototype.toString = function () {
...
+};
+
+_SearchParams.prototype.utmParamsFxA = function(pathname) {
+    pathname = pathname || window.location.pathname || '';
+
+    var utms = this.utmParams();
+
+    // set to default value if not specified in URL
+    if (!utms.utm_campaign) {
+        utms.utm_campaign = 'page referral - not part of a campaign';
```
alexgibson: Do we need to encode both `utm_campaign` and `utm_content` using `encodeURIComponent`?

alexgibson:
I agree on keeping bedrock page specific changes out of `australis-uitour.js`. Approach looks good to me, I think `search-params.js` is the most appropriate place to house this :+1:

tests/unit/spec/base/search-params.js
```diff
@@ -42,6 +42,29 @@ describe('search-params.js', function() {
...
+            var utms = sp.utmParams();
+            var keys = Object.keys(utms);
+            expect(keys).toEqual(['utm_dude', 'utm_sport']);
+            expect(utms.utm_dude).toEqual('lebowski');
+            expect(utms.utm_sport).toEqual('bowling');
+        });
+
+        it('should return an object of utm_ values with defaults for FxA', function () {
+            var sp = new _SearchParams('utm_dude=lebowski&utm_sport=bowling&source=getfirefox');
+            var utms = sp.utmParamsFxA('/es-ES/firefox/sync/');
```
alexgibson: Suggest adding some extra parameters to this test for robustness? e.g. `/es-ES/firefox/sync/foo/?foo=bar&fizz=buzz`?

media/js/firefox/australis/australis-uitour.js
jpetto: Will do for sure. I think one more iteration of this file is coming down today. Will add docs after next update.

media/js/firefox/australis/australis-uitour.js
jpetto: Ah, good advice. I was a little concerned at how much was added here. Didn't know we were already picking and choosing. :smiley:

tests/unit/spec/base/search-params.js
jpetto: In non-testing use, no param will be sent to `utmParamsFxA`. The value will be set in the function to `window.location.pathname`, which doesn't contain any of the querystring. The testing value here is meant to mimic `pathname`, not a full URL.

alexgibson: Ah, makes sense :+1:

media/js/base/search-params.js
jpetto: According to [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1253120#c0), the values are encoded on the product side, so I don't think we want to preemptively encode here.

alexgibson: I see interesting - suggest adding that to the docs here for clarity?

jpetto: Will do. :+1:

jpetto:
Updated docs, cleaned up `australis-uitour.js` changes, and updated `showFirefoxAccounts` call on `/firefox/ios/`. (Other pre-existing calls to `showFirefoxAccounts` are on pages served to older versions of Fx, and since this update only affects Fx 47+, didn't see a reason to update those.)

Updated demo5 as well.

**Currently Nightly works to test these changes.**

docs/uitour.rst
```diff
@@ -428,19 +428,43 @@ Passing ``defaultBrowser`` will set Firefox as the default web browser.
...
-Allows a web page to navigate directly to ``about:accounts?action=signup``
+Allows a web page to navigate directly to
+``about:accounts?action=signup&entrypoint=uitour``. Optionally accepts an object
+of ``utm_*`` key/values, which will be encoded and appended to the
+``about:accounts`` querystring.
+
+.. Important::
+
+    All keys in ``extraURLCampaignParams`` must begin with ``utm_``. If an
+    invalid key is present, the call to ``showFirefoxAccounts`` will fail.
```
alexgibson: Does the call fail entirely? :fearful:  Or do the params just fail silently?

jpetto: Entire call fails, unfortunately. :disappointed:

docs/uitour.rst
```diff
@@ -428,19 +428,43 @@ Passing ``defaultBrowser`` will set Firefox as the default web browser.
...
 .. Important::
 
     ``showFirefoxAccounts()`` is only available in Firefox 31 onward.
 
+.. note::
+
+    A convenience method named ``utmParamsFxA`` exists in
+    ``js/base/search-params.js`` that pulls all ``utm_`` params from the current
+    page's URL and places them in an object (along with pre-defined defaults)
+    ready to pass to ``showFirefoxAccounts``.
```
alexgibson: Suggest adding a note here as to the version of Firefox which supports the additional parameter?

docs/uitour.rst
```diff
@@ -428,19 +428,43 @@ Passing ``defaultBrowser`` will set Firefox as the default web browser.
...
+    // with extra utm_ campaign params. will open
+    // about:accounts?action=signup&entrypoint=uitour&utm_foo=bar&utm_bar=baz
+    Mozilla.UITour.showFirefoxAccounts({
+        'utm_foo': 'bar',
+        'utm_bar': 'baz'
+    });
+
 .. Important::
 
     ``showFirefoxAccounts()`` is only available in Firefox 31 onward.
```
alexgibson: Suggest adding a note here as to the version of Firefox which supports the additional parameter?

jpetto: Ah, good call. Updated.

alexgibson:
Tested the new functionality works as advertised in Nightly, and still degrades gracefully in Release channel. r+ nice work! :fries:

PR #3939, opened 2016-03-04
Fix Bug 1253629 - Adjust timeout for getFirefoxDetails()

kyoshino:
Increasing the timeout to 300ms as @alexgibson suggested in the bug.


alexgibson:
Thanks again @kyoshino, r+! :bee:

PR #3940, opened 2016-03-07
Parameterize newsletter form functional tests

alexgibson:
## Description
Makes newsletter functional tests parameterized in order to reduce code duplication and allow for easier testing across multiple pages.

## Checklist
- [ ] Requires l10n changes.

Thoughts @davehunt?

### DIFF
--- a/tests/functional/test_newsletter_embed.py
+++ b/tests/functional/test_newsletter_embed.py
@@ -1,0 +1,72 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+from selenium.common.exceptions import TimeoutException
+
+from pages.home import HomePage
+from pages.about import AboutPage
+from pages.contribute.contribute import ContributePage
+from pages.mission import MissionPage
+from pages.firefox.all import FirefoxAllPage
+from pages.firefox.desktop.desktop import DesktopPage
+from pages.firefox.desktop.customize import CustomizePage
+from pages.firefox.desktop.all import FirefoxDesktopBasePage
+from pages.firefox.sync import FirefoxSyncPage
+from pages.plugincheck import PluginCheckPage
+from pages.smarton.landing import SmartOnLandingPage
+from pages.smarton.base import SmartOnBasePage
+
+
+@pytest.mark.nondestructive
+@pytest.mark.parametrize(('page_class', 'url_kwargs'), [
+    pytest.mark.smoke((HomePage, None)),
+    (AboutPage, None),
+    pytest.mark.smoke((ContributePage, None)),
+    (MissionPage, None),
+    (FirefoxAllPage, None),
+    pytest.mark.smoke((DesktopPage, None)),
+    (CustomizePage, None),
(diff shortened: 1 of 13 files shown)


tests/functional/test_newsletter_embed.py
```diff
@@ -0,0 +1,61 @@
...
+        reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1245208')
+    ('teach/smarton'),
+    pytest.mark.skip_if_not_firefox('teach/smarton/tracking',
+        reason='Newsletter is only shown to Firefox browsers.'),
+    pytest.mark.skip_if_not_firefox('teach/smarton/security',
+        reason='Newsletter is only shown to Firefox browsers.'),
+    pytest.mark.skip_if_not_firefox('teach/smarton/surveillance',
+        reason='Newsletter is only shown to Firefox browsers.')])
+def test_newsletter_default_values(slug, base_url, selenium):
+    page = MozillaBasePage(base_url, selenium, slug=slug).open()
```
davehunt: Rather than introduce `MozillaBasePage` I would make the parameters something like `page_object` and `url_kwargs`. Then you can pass `HomePage`, `AboutPage`, etc. and not have to duplicate the URL paths. The `url_kwargs` would be something like `{'slug': 'teach/smarton'}` and expanded when instantiating the page object.

davehunt:
+1 to reducing this duplication, but I'd like to see if we can do it without then duplicating the page URLs.

alexgibson:
@davehunt updated - hopefully this is along the right lines

tests/functional/test_newsletter_embed.py
```diff
@@ -0,0 +1,75 @@
...
+    page.newsletter.type_email('noreply@mozilla.com')
+    page.newsletter.select_country('United Kingdom')
+    page.newsletter.select_text_format()
+    page.newsletter.accept_privacy_policy()
+    page.newsletter.click_sign_me_up()
+    assert page.newsletter.sign_up_successful
+
+
+@pytest.mark.nondestructive
+@pytest.mark.parametrize('page_object', [HomePage, ContributePage])
```
davehunt: I know I suggested `page_object` but perhaps `page_class` would be a more accurate name?

tests/functional/test_newsletter_embed.py
```diff
@@ -0,0 +1,75 @@
...
+        reason='Newsletter is only shown to Firefox browsers.'),
+    pytest.mark.skip_if_not_firefox((SmartOnBasePage, {'slug': 'security'}),
+        reason='Newsletter is only shown to Firefox browsers.'),
+    pytest.mark.skip_if_not_firefox((SmartOnBasePage, {'slug': 'surveillance'}),
+        reason='Newsletter is only shown to Firefox browsers.')])
+def test_newsletter_default_values(page_object, kwargs, base_url, selenium):
+    if kwargs:
+        page = page_object(base_url, selenium, **kwargs).open()
+    else:
+        page = page_object(base_url, selenium).open()
```
davehunt: I wonder if this would work:

```python
url_kwargs = url_kwargs or {}
page = page_class(base_url, selenium, **url_kwargs).open()
```

Slightly less duplication. Also note `url_` prefix, which I think makes it a little clearer what these keywords are used for.

davehunt:
r+ wc

alexgibson:
Thanks @davehunt

PR #3941, opened 2016-03-07
[fix bug 1249726] Add iOS link to Firefox top menu navigation

alexgibson:
## Description
Adda missing link to Firefox iOS page on Firefox top menu navigation.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1249726

## Testing
This older style of navigation is still displayed on pages such as `/firefox/all/`

## Checklist
- [ ] Requires l10n changes (don't think so, but need to double-check)


flodolo:
It doesn't require any l10n (I don't plan to expose this string at all, I'm pretty sure Apple doesn't want iOS translated or transliterated)  ;-)

alexgibson:
> It doesn't require any l10n (I don't plan to expose this string at all, I'm pretty sure Apple doesn't want iOS translated or transliterated) ;-)

Would you prefer this string not to be wrapped at all?

flodolo:
Let's keep it wrapped just in case someone complains, thanks :+1:

alexgibson:
Marking as no-not merge for now, as I have one other open question that came in via the bug report.

alexgibson:
Removing do not merge flag, this is should be good to go as-is.

schalkneethling:
r+ :cake:

PR #3942, opened 2016-03-07
Pass everything through to bedrock in global.conf

jgmize:


pmclanahan:
Still says there are conflicts, but r+ whenever it's mergeable.

PR #3943, opened 2016-03-08
Add documentation for Bedrock deployment pipeline configuration

alexgibson:
This is mostly re-purposed from @davehunt's Google Doc, but reformatted here for easier access.

@jgmize @davehunt r?


davehunt:
nice! r+

PR #3945, opened 2016-03-08
[bug 1240980] Remove legacy Firefox Hello product page

alexgibson:
## Description
Removes the legacy Firefox Hello product page. Marking as do-not-merge for now until we have enabled the new page in production.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1240980

## Testing
As this PR both removes and moves files, clearing out your static assets before loading the page locally may be worthwhile, else a broken reference to a file may go un-noticed.

## Checklist
- [ ] Requires l10n changes.


alexgibson:
Ok I think this is now good for review.

bedrock/firefox/templates/firefox/hello/index.html
```diff
@@ -19,254 +19,131 @@
...
         </li>
       </ul>
-    </div>{# /.container #}
+    </div>
   </section>
-
-  <section id="download-try">
+  <div id="features">
+    <div class="split-container quick">
+      <section>
```
schalkneethling: Curious, is there any reason we cannot simply use the `section` elements without wrapping them inside the `div` elements?

alexgibson: This is just for positioning on the grid. How would you propose to do it using a single element?

alexgibson: I guess I would do it be writing specific CSS rules for each of the containing elements within the current `<section>` elements, but I think this may be more verbose than just letting a containing element handling things (especially when it comes to RTL). Happy to rewrite if you think this is worth it though.

schalkneethling: Ah, I see. Just curious really ;)

schalkneethling:
Not introduced here I believe but, small thing. The `line-height` feels tight to me in the newsletter section. Thoughts?

screen shot 2016-03-10 at 12 48 19

alexgibson:
> The line-height feels tight to me in the newsletter section.

Good spot :+1:

schalkneethling:
One small nit in the newsletter section, other than that r+wc :icecream:

alexgibson:
@schalkneethling updated thanks for the review!

alexgibson:
Wait, just found some more image references that need fixing

bedrock/firefox/templates/firefox/hello/start-45.html
```diff
@@ -27,8 +27,8 @@
   <section>
     <h2>{{ _('Made for sharing the Web.') }}</h2>
     <p>{{ _('Welcome to the all new Firefox Hello. Now when you invite a friend to a session, Firefox Hello will automatically share any Web page you’re viewing.') }} {{ _('Plan. Shop. Decide. Together.') }}</p>
-    {{ high_res_img('firefox/hello/2016/quick.png', {'alt': '', 'width': '45%', 'height': 'auto', 'class': 'screenshot'}) }}
```
alexgibson: Ok updated - @schalkneethling for reference I forgot to update the references on these 2 images. Should be good now.

schalkneethling: Hmm, interesting. I did clear out the static directory and those images did load. Good thing you caught it though.

alexgibson: If you cleared out static then the page *should* throw a 500 since the images wouldn't be there. Odd ¯\_(ツ)_/¯

bedrock/firefox/templates/firefox/hello/index-b.html
```diff
@@ -2,7 +2,7 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
 
-{% extends "firefox/hello/index-2016.html" %}
+{% extends "firefox/hello/index.html" %}
 
 {% set_lang_files "firefox/hello/index-2016" %}
```
schalkneethling: I guess referencing 2016 here is appropriate as it contains the new copy, correct?

alexgibson: Changing the name of the lang file is more trouble than it's worth. The directory is still the same, so this should be uneffected.

bedrock/firefox/templates/firefox/hello/index.html
```diff
@@ -6,11 +6,11 @@
 
 {% extends "firefox/base-resp.html" %}
 
-{% set_lang_files "firefox/hello" %}
+{% set_lang_files "firefox/hello-2016" %}
```
schalkneethling: Just to be sure, I note that `index-b` uses `{% set_lang_files "firefox/hello/index-2016" %}` as apposed to `{% set_lang_files "firefox/hello-2016" %}`

alexgibson: Ah, good spot - they should be the same, let me work out what's up there

bedrock/firefox/templates/firefox/hello/index-b.html
alexgibson: @flodolo - should this actually be `{% set_lang_files "firefox/hello-2016" %}`?

flodolo: Yes, `firefox/hello-2016.lang` is the [one we have](https://l10n.mozilla-community.org/~pascalc/langchecker/?locale=all&website=0&file=firefox/hello-2016.lang).

alexgibson:
Thanks @schalkneethling - I fixed the reference to the lang file on the testing template :+1:

schalkneethling:
r+ :v:

PR #3950, opened 2016-03-09
Mark bedrock-prod as sensitive URL for functional tests

alexgibson:
## Description
I noticed we're currently running form submission destructive tests against `bedrock-prod-deis` URLs, which we probably should not be doing.

@davehunt r?


davehunt:
well spotted, r+

PR #3953, opened 2016-03-09
Fix Bug 1254938 - Google Play button on /firefox/new/ scene 2 has wrong width

kyoshino:
Fix a regression introduced since #3806 has landed.


schalkneethling:
r+ Thanks @kyoshino

PR #3955, opened 2016-03-09
Switch to quay.io for demo docker image hosting

pmclanahan:


pmclanahan:
A little cleanup left (don't need the docker hub username and email any more), but other than that this works fine. More to do obviously if we want to switch the whole works over, but maybe just handling the demos like this is a good trial of the service?

jgmize:
As discussed in irc, I'm fine with switching the whole works over, and if we ever need to revert we'll have this in the git history. Once you clean up the unused docker hub username and email this looks good to merge.

r+wc, nice work :)

pmclanahan:
All cleaned up and ready to go.

PR #3956, opened 2016-03-09
Fix bug 1253413 - Foundation newsletter design/copy testing

craigcook:
## Description
This adds B variants of three pages to test design/copy of the Foundation newsletter signup form: home page, main about page, and the main legal page. We'll use Optimizely to redirect the pages to their B variants and use Google Analytics to track engagement.

This requires the following waffle switches:
`mozorg-home-optimizely`
`mozorg-about-optimizely`
`legal-index-optimizely`

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1253413

## Testing
Visiting one of these pages with the query param `?v=b` in the URL should show the B variant of that page. The A version of each page uses the currently standard form with the standard "Get Mozilla updates" headline while the B version shows the new design and copy. The home page and /about page already have forms so the A versions there are the current (control) versions and B is the new design. 

This test is for English only and other locales should show no form on /about/legal and show the current Firefox form on the home page and /about (Foundation newsletter is only in English so other locales get Firefox). These are new strings but the test is being conducted only in English so they should not be extracted/exposed for l10n at this time.

## Checklist
- [ ] Requires l10n changes.


craigcook:
@schalkneethling r?

bedrock/mozorg/views.py
```diff
@@ -502,11 +515,16 @@ def home_tweets(locale):
...
 
 
-def home(request):
+def home(request, template='mozorg/home/home.html'):
     locale = l10n_utils.get_locale(request)
 
+    variant = request.GET.get('v', '')
+
+    if variant == 'b':
+        template = 'mozorg/home/home-b.html'.format(variant)
```
schalkneethling: You do not need to call the `format` function on this string.

bedrock/legal/views.py
```diff
@@ -91,3 +91,16 @@ def impressum(request):
...
+
+def legal_newsletter_test(request, template='legal/index.html'):
+    """
+    Renders the legal index page template.
+    """
+
+    variant = request.GET.get('v', '')
+
+    if variant == 'b':
+        template = 'legal/index-b.html'.format(variant)
```
schalkneethling: `format` no needed here.

bedrock/mozorg/views.py
```diff
@@ -420,6 +420,19 @@ def plugincheck(request, template='mozorg/plugincheck.html'):
...
 
+def about_newsletter_test(request, template='mozorg/about.html'):
+    """
+    Renders the private browsing page template.
+    """
+
+    variant = request.GET.get('v', '')
+
+    if variant == 'b':
+        template = 'mozorg/about-b.html'.format(variant)
```
schalkneethling: `format` no needed here.

schalkneethling:
@alexgibson Are there some functional tests that should be disabled here?

alexgibson:
@schalkneethling - you can run the tests locally to find out.

I would look here:

https://github.com/mozilla/bedrock/blob/master/tests/functional/test_newsletter_embed.py
https://github.com/mozilla/bedrock/blob/master/tests/functional/test_about.py
https://github.com/mozilla/bedrock/blob/master/tests/functional/test_home.py

schalkneethling:
Maybe it is just me, but the line-height of the newsletter copy looks a little tight:

screen shot 2016-03-10 at 13 53 59

IE9 and 10 is not sizing the SVG properly, I believe it is because it wants the width and height attributes on the SVG element:
   
virtualbox_windows 7 - ie9_10_03_2016_16_31_23

And now, Edge (og noes!) seems to be chopping of the bottom of the SVG (perhaps it will also be happy when the width and height is added.
 
virtualbox_ie11 - win10_10_03_2016_16_39_29

As you can see from the last screen shot, IE11 is fine though....go figure.

craigcook:
Made a few font tweaks and fixed up the SVG. IE10 seems happy now, but I don't have an Edge VM to test. I guess I should download one...

bedrock/legal/urls.py
```diff
@@ -10,7 +10,10 @@
 from bedrock.legal_docs.views import LegalDocView
 
 urlpatterns = (
-    page('', 'legal/index.html'),
+    # page('', 'legal/index.html'),
+    url(r'^$',
+        views.legal_newsletter_test,
+        name='legal.index'),
```
jpetto: Even if it's coming back later, should remove the commented out line. Could also put the `url` call/args on a single line for conformity's sake.

media/css/mozorg/newsletter-test.less
```diff
@@ -0,0 +1,202 @@
...
+
+        ::-moz-placeholder {
+           color: #ccc;
+           font-weight: normal;
+        }
+
+        :-ms-input-placeholder {
+           color: #ccc;
+           font-weight: normal;
+        }
```
jpetto: Could combine the above?

craigcook: Most clients, when encountering a rule with an unknown selector, will ignore the entire rule. So for vendor prefixed selectors like this you have to repeat the rule for each vendor. Kinda sucks.

jpetto: Ah, I was wondering about that. Womp womp.

PR #3957, opened 2016-03-09
Replace vary user-agent w/ cache-control max-age=0

jgmize:


jgmize:
@pmclanahan r?

bedrock/firefox/views.py
```diff
@@ -284,7 +284,7 @@ class LatestFxView(TemplateView):
     """
     non_fx_redirect = 'firefox.new'
 
-    @vary_on_headers('User-Agent')
+    @cache_control(max_age=0)
```
pmclanahan: NIT: why not use the `never_cache` decorator instead. IIRC they should have the same result, but it's a bit more explicit.

pmclanahan:
Apart from that one nit this looks good. r+

bedrock/firefox/views.py
jgmize: I considered using the never_cache decorator, but then out of curiosity I looked at the source and saw that it calls [patch_cache_control](https://github.com/django/django/blob/1f29164ced01bde92c72d20d67bc453484a4f078/django/utils/cache.py#L41-L87) which was not necessary for this case, so I used this method as an optimization. Given the traffic levels of this particular view, I felt the optimization was justified over the very slight loss of readability.

pmclanahan: calling `patch_cache_control` is what the `cache_control` decorator does too. They seem identical really, just that the `cache_control` one is more about setting it for some > 0 time.

PR #3960, opened 2016-03-10
Add a clock process

pmclanahan:
Will allow us to remove jenkins periodic task


pmclanahan:
Refreshed and mostly copy-pasta'd from @glogiotatidis version in snippets (mozilla/snippets-service#175). Only pings DMS on one of the jobs for now, mostly to ensure that the process hasn't crashed. We can either add a way to add separate DMS urls for each task, or add an alert in DMS looking for the "CRASHED" messages, or both.

@glogiotatidis @jgmize r?

Procfile
```diff
@@ -0,0 +1,2 @@
+web: ./docker/run-prod.sh
```
jgmize: This will work on its own for a new app, but we've seen issues switching from "cmd" process type to the "web" proc type when deploying running apps. I wonder if it would be safer to `s/web/cmd/`?

pmclanahan: Have we tested adding a `cmd` proc type to a Procfile? That `cmd` type is added automatically by deis for a dockerfile deploy.

pmclanahan: Also these types of systems typically have special handling for the `web` type with ports and whatnot. We need to test all of this for sure. Perhaps I can push this to my demo to see what it does with an existing app?

glogiotatidis:
So here's something I don't get about new pip8 and hashes. When I installed apscheduler in snippets, I got an error that `tzlocal` and `futures` packages have no hashes.  

Here you only add `tzlocal` and you get no error for `futures`.

[`setup.py`](https://bitbucket.org/agronholm/apscheduler/src/ec775f9d2ce34f689fcbae28e2ab9200ddd6fa71/setup.py?fileviewer=file-view-default#setup.py-42) from `apscheduler` requires `tzlocal`, `futures` *and* `funcsigs` when in `python2.7`. Can anyone explain what's going on?

pmclanahan:
Yeah. futures is in base.txt which is always installed first. I'm not sure about `funcsigs`, but it doesn't seem to be required.

pmclanahan:
When I install it locally it doesn't install `funcsigs`:

```bash
$ pip install apscheduler
Collecting apscheduler
  Using cached APScheduler-3.0.5-py2.py3-none-any.whl
Collecting six (from apscheduler)
  Using cached six-1.10.0-py2.py3-none-any.whl
Collecting futures (from apscheduler)
  Downloading futures-3.0.5-py2-none-any.whl
Collecting pytz (from apscheduler)
  Using cached pytz-2015.7-py2.py3-none-any.whl
Collecting tzlocal (from apscheduler)
Installing collected packages: six, futures, pytz, tzlocal, apscheduler
Successfully installed apscheduler-3.0.5 futures-3.0.5 pytz-2015.7 six-1.10.0 tzlocal-1.2.2
```

pmclanahan:
@jgmize just did another test of deployment and process types:

1. destroyed, recreated and pushed latest master to my demo app (bedrock-demo-pmac)
2. pushed this branch with the `Procfile` changed to use `cmd` instead of `web` as the process type to my demo

It's still up and running w/ no issues. So maybe that'll work fine? No idea if changing that line in the Procfile will affect what is run, or if it's just continuing to use the `CMD` set in the dockerfile, but it is working, and we're not likely to change that.

jgmize:
@pmclanahan thanks for testing the `cmd` process type in the Procfile; I'm glad it works but you make some good points about potential for confusion. I believe you may have already tested switching from `cmd` to `web` with a single process, but I wanted to do some further testing with multiple processes and also observe the change in the etcd namespace to validate that the traffic was route as I would expect. Here are the steps I used:

0. `deis scale cmd=3 -a bedrock-demo-jgmize` to test multiple containers 
1. `git pr 3960`  -- https://github.com/jgmize/dotfiles/blob/master/.gitconfig#L15
2. `git co pr/3960`
3. `git co -b demo__jgmize`
4. `git push -f origin demo__jgmize`  -- origin = mozilla for me
5.  wait for circleci to deploy
6. `deis ps -a bedrock-demo-jgmize`
7. verify that 3 `cmd` process types are still running
8. `etcdctl exec-watch  --recursive /deis/services/bedrock-prod -- /bin/sh -c "env | grep ETCD | paste -s -d ' ' | systemd-cat -t upstream-watcher"`
9. watch results in papertrail while doing the following
10. `deis ps:scale clock=1`
11. `deis ps:scale web=3`
12. validate that traffic is being routed to both `cmd` and `web` process types, but not `clock`
13. `deis ps:scale cmd=0`
14. validate that only `web` and `clock` process types remain, and traffic flow was never interrupted

Based on the testing above I think it's safe to use the standard `web` process type in the Procfile, as long as we follow the same procedure for switching to the new process type upon deployment. I don't think there are any more blockers to merge and deploy this, as our existing processes will continue to run until we make the manual changes, but there are some follow-ups:

1. set the `DEAD_MANS_SNITCH_URL` environment variable in each deployment
2. deactivate the Jenkins jobs that the clock processes are replacing
3. documentation
4. consider setting up independent dead man's snitch urls for each recurring task

jgmize:
Unfortunately when this was deployed by Jenkins to the dev and stage environment, the Procfile was not automatically detected, and so the `web` and `clock` processes were not available initially. When I tried to run `deis scale web=5 -a bedrock-dev` I got a `Error: 400 BAD REQUEST detail: Container type web does not exist in application.` The reason for this is because the Procfile detection occurs client side, and the Procfile must be [in the current working directory (or specified by the --procfile argument)](http://docs.deis.io/en/latest/using_deis/process-types/#deploying-to-deis) when the `deis pull` command is executed, which was the case during the circleci deployment, but not the jenkins deployments. I fixed this manually for the dev and stage environments, and can do the same for prod deploy when this is ready to be activated. Unfortunately, further testing has revealed some additional issues that I didn't see in my earlier testing, so we aren't quite ready to switch this on in prod yet. Looking through the logs of the running clock processes, I saw the following errors repeated:
```
Clock job job_rnasync@bedrock-dev: CRASHED: 'RNAModelClient' object has no attribute 'objects'
Clock job job_update_product_details@bedrock-dev: CRASHED: permission denied for relation product_details_productdetailsfile
Clock job job_update_externalfiles@bedrock-dev: CRASHED: permission denied for relation externalfiles_externalfile
Clock job job_ical_feeds@bedrock-dev: CRASHED: permission denied for relation events_event
```

The other jobs, namely `update_firefox_os_feeds` and `update_security_advisories`, appear to be completing successfully.

jgmize:
The jenkins client update resolved the Procfile detection issue, but there was some breakage in the dev and stage environments due to the `cmd` process type having been scaled to 0 and the `web` process type being unavailable. After the jenkins deploys with the updated client I needed to scale web=5 and clock=1 on the dev and stage environments again.

pmclanahan:
Deis handles Procfiles in a very unfortunate way. It should be extracting it from the image at the root of `WORKDIR` IMHO. Having to run `deis pull` from a certain directory is at best unintuitive. But oh well.

Also glad you figured out the issues and that hopefully we're okay with things as they are.

jgmize:
> Deis handles Procfiles in a very unfortunate way. It should be extracting it from the image at the root of WORKDIR IMHO. Having to run deis pull from a certain directory is at best unintuitive. But oh well.

Keep in mind that my original assumption that it was a working directory problem was actually invalid: this was a client version issue, not a working directory issue. However, I agree that detecting and parsing the Procfile from the image would be a nice usability improvement.

jgmize:
Status update: this code is in production, but I have not enabled the clock process in prod and we are still dependent on the Jenkins scheduled jobs. The reason for this is due to the recurrence of the `'RNAModelClient' object has no attribute 'objects'` error. I thought this was resolved earlier, because I saw the job run successfully after fixing the DB permissions error, but subsequent runs revealed a failure pattern where the first run of the job would succeed, but all subsequent jobs fail with this error. I was unable to reproduce this error from the command line by running `./manage.py rnasync` in the deployed container, but after reading through the code, I realized the source of the problem: when [constructing the model_params](https://github.com/mozilla/rna/blob/833c5ad432925913c54be1c8954b2620e82b386a/rna/management/commands/rnasync.py#L36) we iterate through `RNAModelClient.model_map`, which is a [class attribute](https://github.com/mozilla/rna/blob/833c5ad432925913c54be1c8954b2620e82b386a/rna/clients.py#L73) that is [mutated as a side effect of the model_client method](https://github.com/mozilla/rna/blob/833c5ad432925913c54be1c8954b2620e82b386a/rna/clients.py#L111-L113). This does not cause issues in a short lived management command process, but in the long lived clock process the mutation of the class attribute causes this problem. One relatively simple fix for this would be to move the attribute from the class to an instance attribute, but this code has already been deprecated in the master branch of rna. OTOH, the master branch of RNA has other dependencies which block the immediate deployment, so it may be worth creating a branch off the currently deployed version of with just the patch to fix this particular problem. What are your thoughts, @pmclanahan?

jgmize:
I went ahead and created a branch on rna with the change described above: https://github.com/mozilla/rna/commit/fdcee9e4ba4f860483eaf9476be02d05b7a81f6e

jgmize:
This is now in prod in both clusters and the jenkins scheduled jobs are disabled.

PR #3962, opened 2016-03-10
[bug 1248040] Update GTM data attributes on family navigation and sid…

JamesDLorence:
…e menu, download modal, and /firefox/ios/ CTA buttons

Bug 1248040 - https://bugzilla.mozilla.org/show_bug.cgi?id=1248040

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -229,9 +229,9 @@
           <a data-id="ios" href="{{ url('firefox.ios') }}" class="primary-link" data-link-type="nav" data-link-name="for iOS"><div>{{ _('<span>for</span> iOS') }}</div></a>
 
           <ul class="subnav" id="ios-subnav">
-            <li><a href="{{ url('firefox.ios') }}"{% if activesub == 'index' %} class="selected"{% endif %}><div>{{ _('Overview') }}</div></a></li>
-            <li><a href="{{ url('firefox.sync') }}"{% if activesub == 'sync' %} class="selected"{% endif %}><div>{{ _('Sync') }}</div></a></li>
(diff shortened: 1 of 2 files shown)


bedrock/base/templates/macros.html
```diff
@@ -349,7 +349,9 @@ <h2 class="thank-you hidden">{{ _('Your download link was sent.') }}</h2>
               <input id="id-input" name="phone-or-email" type="text" required>
             </div>
             <div class="form-submit">
-              <button type="submit" class="button-flat" data-download-os="Android">{{ _('Send') }}</button>
+              <button type="submit" class="button-flat"
+              {% if platform == 'android' %} data-download-os="Android"
+              {% elif platform == 'ios' %} data-download-os="iOS" {% endif %}>{{ _('Send') }}</button>
```
alexgibson: While we're not actively using it on any current pages, there is another possible option here: `platform == 'select'`. In this instance the user gets to select either Android or iOS as their chosen platform before submitting the form. Is this something you would like to add attributes for now proactively? Happy to defer to what you decide here.

alexgibson:
Aside from one open question, code-wise this looks good. r+wc (change being optional)

bedrock/base/templates/macros.html
JamesDLorence: I don't see any easy way to pass which option is selected and store that in the data-attribute or I'd update it now. Unless this "select" feature is going live sometime soon, I'd like to just leave it as is.

alexgibson: To do this we would likely need to use JS to populate the data attribute value when someone makes a selection in the UI, however if you're OK not tracking this for now then we can just add this later if/when we need.

PR #3964, opened 2016-03-11
Update ESLint config

alexgibson:
## Description
Updates out ESLint config with a few changes:

- [x] Always require curly braces.
- [x] Variable and function names should use camelCase.
- [x] Strict comparison for equality when needed.
- [x] One variable declaration per line.
- [x] Require constructors to use initial caps.


schalkneethling:
r+ :+1:

PR #3961, opened 2016-03-10
Add cheet sheet for prod push to pipeline docs.

jpetto:
@jgmize r?

PR #3965, opened 2016-03-11
Add a script for properly tagging a release

pmclanahan:


jgmize:
Should this script include a `git fetch --tags`?

pmclanahan:
I assumed that the instructions from #3961 had been followed, but that's not a bad idea. I'll add it. Can't hurt.

pmclanahan:
@jgmize I added it, but it may not help since we can't know that the user has the master branch configured to fetch from the mozilla repo, and we can't know the name of the remote that points to mozilla :/ So I think this is as good as we get.

jgmize:
Thanks @pmclanahan

PR #3970, opened 2016-03-12
Fix Bug 1255892 - the link to the japanese build of dev edition is a 404

kyoshino:


pmclanahan:
Nice catch @kyoshino. It'd be nice to have a test for this, but we can add later. Let's get this fixed.

PR #3931, opened 2016-03-03
Move from peep to pip8

pmclanahan:

PR #3974, opened 2016-03-15
Remove redundant pipstrap

jgmize:
Followup from https://github.com/mozilla/bedrock/pull/3931/files#r55686643
@pmclanahan r?


pmclanahan:
Thanks!

PR #3978, opened 2016-03-16
Upgrade rna to fix clock process error in rnasync

jgmize:
@pmclanahan r?

See https://github.com/mozilla/bedrock/pull/3960#issuecomment-197457318 for details.


jgmize:
Verified that since this was deployed to dev and stage all rnasync jobs are completing successfully. This will be deployed with https://github.com/mozilla/bedrock/releases/tag/2016-03-16.2

PR #3979, opened 2016-03-16
Fix Bug 1255729 - Webmaker terms of service link is 404

kyoshino:


alexgibson:
Thanks @kyoshino r+ :fries:

PR #3981, opened 2016-03-16
[bug 1248040] Update GTM data attributes on firstrun and australis/fi…

JamesDLorence:
…rstrun tabzilla and footer, and base template navigation

Bug 1248040 - https://bugzilla.mozilla.org/show_bug.cgi?id=1248040

### DIFF
--- a/bedrock/base/templates/base.html
+++ b/bedrock/base/templates/base.html
@@ -106,16 +106,16 @@
         <nav id="nav-main" role="navigation">
           <ul>
             <li class="first about-item">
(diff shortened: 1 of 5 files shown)


alexgibson:
If you're adding attributes to the australis tour pages, you probably want to also update the base template here, which the firstrun template extends: https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/australis/fx36/tour.html

alexgibson:
r+wc

JamesDLorence:
I went through and added attributes to the tour.html page you pointed out. Good catch.

alexgibson:
r+

PR #3983, opened 2016-03-17
Fix Bug 1255785, remove logo prototype assets

schalkneethling:
## Description

Removes all assets and pages related to the logo prototype that was proposed a while back but, which will no longer be needed.

## Bugzilla

https://bugzilla.mozilla.org/show_bug.cgi?id=1255785

## Testing

I ran:

```
./manage.py test
grunt test
```
All tests pass. As this is not live I doubt there are any functional tests. I also did a search through the project for `logo-prototype` and removed everything it found.

## Checklist
- [ ] Requires l10n changes.


schalkneethling:
@alexgibson r?

alexgibson:
It looks like this page is live: https://www.mozilla.org/en-US/styleguide/identity/mozilla/logo-prototype/

I suggest we add a redirect back to /styleguide/ just to avoid any 404's from external links.

r+wc

alexgibson:
@schalkneethling - We should add a test for the redirect too?

tests/redirects/map_globalconf.py
```diff
@@ -985,6 +985,7 @@
              re.compile(r'/en-US/firefox/[\d\.]+/releasenotes/'),
              status_code=requests.codes.found),
     url_test('/firefox/brand/', '/styleguide/'),
+    url_test('/styleguide/identity/mozilla/logo-prototype/', '/styleguide/'),
```
alexgibson: Nit: instead of bundling this up in redirects that we're ported over from mcom-tests, it would be nice to keep this separate along with a reference to the bug number?

alexgibson:
r+ :confetti_ball:

PR #3984, opened 2016-03-17
Demos built by CircleCI should update product-details

pmclanahan:
It was failing to do so due to some environment settings.


pmclanahan:
@jgmize @alexgibson r?

alexgibson:
r+

PR #3985, opened 2016-03-17
Trigger Jenkins deployment pipeline from CircleCI

pmclanahan:
Instead of from Github directly since we have a bit more control.
This allows us to be sure that only pushes to master, and no other
branch, will trigger the full build pipeline.


pmclanahan:
@jgmize r?

PR #3986, opened 2016-03-17
Fix CircleCI demo deploy data

pmclanahan:
* Have it run the fxos feed update
* Have it use the .bedrock-demo-env settings

### DIFF
--- a/bin/circleci-demo-sync.sh
+++ b/bin/circleci-demo-sync.sh
@@ -3,12 +3,7 @@ set -ex
 
 [[ "$CIRCLE_BRANCH" != demo__* ]] && exit 0
 
-DEBUG=False
-DEV=True
-DATABASE_URL=sqlite:///bedrock.db
-DOCKER_CACHE_PATH=~/docker
-MOFO_SECURITY_ADVISORIES_PATH=$DOCKER_CACHE_PATH/security_advisories
-PROD_DETAILS_STORAGE=product_details.storage.PDDatabaseStorage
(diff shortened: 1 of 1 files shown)


pmclanahan:
This fixes the product-details issues as well as the fxos feed one. I've not tried to fix the twitter feeds yet because my understanding is that those require some secrets and we don't want those in the repo. I can work around this if necessary, but I didn't think they were super important.

@alexgibson r?

pmclanahan:
@jgmize you might be interested as well

alexgibson:
Thanks for this @pmclanahan - I guess the Twitter feed things are not super important to have on ad-hoc demo instances. We do have functional tests for such things, but as long as we expect failures here when running the tests against demo I think that's ok. It would be a nice-to-have so we can confidently get green runs, but then if it's a more complicated thing to do then we have to weigh up the level of effort required. I would be happy to go with what we have here for now, and we could always follow up in the future if we decided to :+1:

pmclanahan:
Let me give it a go then before we merge. I should maybe be able to add the twitter setting in Circle as a secret, and it'll use it while populating the DB. Might Just Work:tm:.

pmclanahan:
Okay. I've added the tweets updater and added the secret to circle. So maybe it'll work. I've also pushed this branch as `demo__pmac` to origin, so we'll see in a few min whether it worked.

alexgibson:
Seems all good on demo, so r+ from me!

https://bedrock-demo-pmac.us-west.moz.works/en-US/

PR #3989, opened 2016-03-18
Fix Bug 1244465 - Firefox /all/ download pages intermittently show localized strings in the title

kyoshino:
@pmclanahan r?


pmclanahan:
No good way to test this other than to see if we stop getting bugs filed I think :) Thanks @kyoshino!

PR #3990, opened 2016-03-18
Revert rna to 1.0

jgmize:


jgmize:
This reverts rna to the version we had before https://github.com/mozilla/bedrock/pull/3978

PR #3994, opened 2016-03-18
Add tagged release deployment to circle.yml

jgmize:
@pmclanahan r?

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -65,3 +65,8 @@ deployment:
     owner: mozilla
     commands:
       - bin/circleci-trigger-jenkins.sh
+  release:
+    tag: /\d{4}-\d{2}-\d{2}(\.\d)?/
+    owner: mozilla
+    commands:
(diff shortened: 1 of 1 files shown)


circle.yml
```diff
@@ -65,3 +65,8 @@ deployment:
     owner: mozilla
     commands:
       - bin/circleci-trigger-jenkins.sh
+  release:
+    tag: /\d+-\d+-\d+(\.\d+)?/
```
pmclanahan: I was actually curious if we could just add this line to the `jenkins` section. But this should definitely work.

We could also get more specific `tag: /\d{4}-\d{2}-\d{2}(\.\d)?/`, but that's not really necessary either (that's what I have in basket).

pmclanahan:
Some comments if you wanna try things, but no blockers. r+

jgmize:
Updated to use the more specific regex

PR #3995, opened 2016-03-18
[bug 1244200] Implement new Firefox newsletter form design

craigcook:
## Description
This standardizes a new design of the Firefox newsletter subscription form and adds it to most Firefox product pages (we'll add it to a few more soon, such as release notes, maybe dev edition). This also cleans up the extra templates and views from the previous round of A/B testing.

I've put the form's CSS into its own file that can be included in bundles as needed or even imported into another Less file. This hopefully makes the design fairly modular but since the form itself is rendered by a separate helper and is baked into a lot of base templates, other pages should still get the old standard form with the plain design unless we explicitly update them to the new design.

The headline is a new string which I've put behind the flag `fxnewsletter_subscribe_update`. I believe this should go in main.lang where the old string ("Get Firefox news") resides rather than repeat it in the lang files for each page. @flodolo please correct me if I'm wrong.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1244200

## Checklist
- [x] Requires l10n changes.


craigcook:
Also note: we're still running an A/B test on the /hello page so that one still has the two templates and view.

bedrock/firefox/templates/firefox/all.html
```diff
@@ -115,6 +115,19 @@
   </main>
 {% endblock %}
 
+{% block email_form %}
+  <aside id="newsletter-subscribe" class="ga-section" data-ga-label="Keep up with all things Firefox.">
+    <div class="content container">
+      {% if l10n_has_tag('fxnewsletter_subscribe_update') %}
```
flodolo: We don't translate this page. Only Japanese does, but it's untracked by our systems.

flodolo:
The string needs to go in main.lang indeed, which means we can't use a tag and need to wait until it's translated. 

And sadly I pushed strings just a few hours ago. This entire sprint revolution  is extremely hard to follow on my side, requests pop up without any notice and I can't basically plan anything.

craigcook:
I've removed the l10n tags since they won't work with main.lang and replaced them with `LANG.startswith('en-')` which we'll just need to remove later once we have good locale coverage for this new string.

media/css/firefox/desktop/common.less
```diff
@@ -593,11 +568,17 @@ html[dir="rtl"] {
         background-repeat: no-repeat;
         min-height: 120px;
         padding: 30px 0 20px 180px;
-        .font-size(40px);
         margin-left: -((@gridColumnWidth * 7) + (@gridGutterWidth * 6));
         margin-top: -30px;
         text-shadow: none;
 
+        h3 {
+            .font-size(40px);
```
schalkneethling: `(@baseLine * 2)` ?

schalkneethling:
@craigcook So, this is on the product pages correct? Anywhere else?

schalkneethling:
In IE9 the background gradient of the button is still in effect

ie9

schalkneethling:
Other than the previous comment, this is looking great.

craigcook:
The new form should appear on:
* /firefox/products
* /firefox/all
* /firefox/desktop (all pages)
* /firefox/ios
* /firefox/hello (two variants)
* /firefox/sync
* /firefox/private-browsing
* /firefox/pocket

/firefox/android is a slightly special case because the form shares space with the download button in the design so that page doesn't get the new form layout, but does get the new headline.

Turns out IE9 is showing a gradient because the mixin used by the `.button` class includes `-ms-filter`. I'm overrriding the background image but not the filter. So... I can just explicitly override the filter as well, OR maybe we should take the plunge of removing that filter from the gradient mixin? Then old IEs wouldn't show gradients anywhere that uses the same mixin. Maybe that's beyond the scope of this PR? At least something to consider, especially as we continue on the march toward dropping IE8.

schalkneethling:
r+ :dart:

PR #3998, opened 2016-03-21
Bug 1255882: Fix more 404s from GA report

pmclanahan:


pmclanahan:
@jgmize r?

bedrock/firefox/redirects.py
```diff
@@ -312,4 +314,58 @@ def firefox_mobile_faq(request, *args, **kwargs):
...
+    redirect(r'^projects/javaconnect(/.*)?$',
+             'http://developer.mozilla.org/en/JavaXPCOM'),
+    redirect(r'^projects/marketing(/.*)?$',
+             'http://contribute.mozilla.org/Marketing'),
+    redirect(r'^projects/minefield(/.*)?$',
+             'http://www.mozilla.org/projects/firefox/prerelease.html'),
+    redirect(r'^projects/minimo(/.*)?$', 'https://wiki.mozilla.org/Mobile'),
+    redirect(r'^projects/namoroka(/.*)?$', '/firefox/channel/'),
+    redirect(r'^projects/nspr(?:/.*)?$', 'https://developer.mozilla.org/docs/NSPR'),
+    redirect(r'^projects/netlib(/.*)?$', 'https://developer.mozilla.org/en/Necko'),
```
jgmize: This works, but is redirected to https://developer.mozilla.org/en-US/docs/Mozilla/Projects/Necko

bedrock/firefox/redirects.py
```diff
@@ -312,4 +314,58 @@ def firefox_mobile_faq(request, *args, **kwargs):
...
+             'https://developer.mozilla.org/En/DOM_Inspector'),
+    redirect(r'^projects/javaconnect(/.*)?$',
+             'http://developer.mozilla.org/en/JavaXPCOM'),
+    redirect(r'^projects/marketing(/.*)?$',
+             'http://contribute.mozilla.org/Marketing'),
+    redirect(r'^projects/minefield(/.*)?$',
+             'http://www.mozilla.org/projects/firefox/prerelease.html'),
+    redirect(r'^projects/minimo(/.*)?$', 'https://wiki.mozilla.org/Mobile'),
+    redirect(r'^projects/namoroka(/.*)?$', '/firefox/channel/'),
+    redirect(r'^projects/nspr(?:/.*)?$', 'https://developer.mozilla.org/docs/NSPR'),
```
jgmize: This works but is redirected to https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSPR

jgmize:
I started commenting on individual urls that worked but got redirected, then realized that since they are fixing things that are non-working right now, the best way to address them would be a followup to reduce the number of redirect hops, and should be considered outside of the scope of this PR, which does appear to resolve all the intended 404s, and as such I'm going to merge this as-is.

PR #3999, opened 2016-03-22
Fix Bug 1257703 - Update MOSPL with v1.1 revisions

kyoshino:
https://www.mozilla.org/en-US/about/patents/license/ is not localized.


alexgibson:
Thanks @kyoshino, this looks good to me :+1: r+

PR #4003, opened 2016-03-22
Add Optimizely project ID to demo deployments

alexgibson:


alexgibson:
@pmclanahan r?

I'm guessing we'll likely need to add this too for demo deployments (seems to be the same ID for all environments from what I can tell).

PR #4005, opened 2016-03-22
Remove old files associated with legacy infra

pmclanahan:


jgmize:
We haven't completely shut down the dev or demo environments in the legacy infra yet; let's hold off on merging this until we do.

pmclanahan:
Good point. Nearly forgot :) This should be fine to wait until then.

PR #4006, opened 2016-03-23
Add mozilla-banner.gif

pmclanahan:
Also move certs to root_files to avoid having to copy and gzip them on collectstatic.


pmclanahan:
@craigcook @jgmize r?

An argument was made in #www that this particular file has historic value and is well spread around the interwebs. I'm happy to put it back (it is quite cool after all). We can take these on a case by case basis as far as I'm concerned. I wanted 404 numbers but GA doesn't catch it because it's a 301 from bedrock and the img tag doesn't do the redirect or load the 404 w/ js. So we'll need to do some better log analysis. Also I added `images` to the locale exemptions to make those requests do a 404 immediately so we can better analyze.

PR #4007, opened 2016-03-23
Replace waffle with environment variables

pmclanahan:


pmclanahan:
@jgmize here's my first shot at the idea. Obviously we'll need to replace all current waffle use with this, but wanted feedback before going that far.

bedrock/mozorg/helpers/misc.py
```diff
@@ -37,6 +39,25 @@ def convert_to_high_res(url):
...
+
+    For example:
+
+        {% if switch('dude-and-walter') %}
+
+    would check for an environment variable called `SWITCH_DUDE_AND_WALTER`. The string from the
+    `switch()` call is converted to uppercase and dashes replaced with underscores.
+    """
+    env_name = 'SWITCH_' + name.upper().replace('-', '_')
+    return config(env_name, default=settings.DEV, cast=bool)
```
pmclanahan: I did look at the implementation in decouple and it loads all it's stuff just once at startup, so since we already use it in settings and this instances is a singleton it should be equally efficient to use this as `os.environ` directly, plus we get the benefit of the boolean casting from strings, the easy default, and managing locally via a `.env` file if devs want/need to.

jgmize:
thanks @pmclanahan, this lgtm so far.

pmclanahan:
@jgmize so. what do you think? I'm pretty happy with this actually. Maybe instead of switching every waffle switch at once we should merge this, then do the switches as they come up? Or at least as a separate PR so that new switches can be done this way while we're working on porting the others. There will be some coordination required for the actual port of switches and removal or waffle.

jgmize:
> [...] Maybe instead of switching every waffle switch at once we should merge this, then do the switches as they come up? Or at least as a separate PR so that new switches can be done this way while we're working on porting the others. There will be some coordination required for the actual port of switches and removal or waffle.

Seems reasonable to me. Go ahead and remove the "Do Not Merge" and WIP labels and I'll do some testing as soon as I get a chance.

pmclanahan:
done

PR #4010, opened 2016-03-24
Add supervisor to docker for dev/demo

pmclanahan:
Also add locale update to cron process in DEV mode.


pmclanahan:
@jgmize r?

I'd like to add more robust handling of locales as well (e.g. handling cloning of the proper repo based on `DEV` variable), but we can do that when/if we expand this technique beyond dev/demo.

docker/run-supervisor.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+
+set -xe
+
+enable_proc() {
+  cp "etc/supervisor_available/$1.conf" "etc/supervisor_enabled/$1.conf"
```
pmclanahan: I tried using a symlink instead and supervisor didn't seem to like it :disappointed:

jgmize:
Thanks @pmclanahan, nice work.

PR #4011, opened 2016-03-25
[fix bug 1258491] Update Fx Family Nav.

jpetto:
## Description
Update Fx Family nav:

- Remove subnav
- Add "Features" item and related page (/firefox/features/)

## Bugzilla
[bug 1258491](https://bugzilla.mozilla.org/show_bug.cgi?id=1258491)

## Testing
Check nav functionality and new /firefox/features/ page on all affected pages across browsers.

I tested on all browsers locally - Safari, Chrome, Fx, IE 6 - Edge.

## Notes
- A new nav, based on these changes, will be coming for FirefoxOS pages in [bug 1259851](https://bugzilla.mozilla.org/show_bug.cgi?id=1259851).
- Copy changes are likely on /firefox/features/ prior to merge, but I think layout/design will stand for initial launch.

## Checklist
- [x] Requires l10n changes.


flodolo:
What is the expectation about this PR (i.e. when do you plan to land)? 

As we talked on IRC, touching the main menu means adding strings to main.lang, and we can't hide those, we either need to wait or put them behind a `LANG.startswith('en-')` conditional.

P.S. I guess the checkbox "Requires l10n changes" should be checked

bedrock/firefox/templates/firefox/features.html
```diff
@@ -0,0 +1,72 @@
...
+
+  <ul id="fxfeature-list">
+    <li class="private-browsing">
+      <h2>{{ _('Private Browsing') }}</h2>
+
+      <p>
+        {{ _('No other browser gives you so much control over your digital footprint.') }}
+      </p>
+
+      <a href="{{ url('firefox.private-browsing') }}">{{ _('Learn more about Private Browsing') }}</a>
```
craigcook: Not sure how much design guidance we got for this page, but I might suggest embiggening these links or even wrapping the link around the entire block (which would mean some other wrapper with a class for the cta text).

jpetto:
> What is the expectation about this PR (i.e. when do you plan to land)?

As this update has major changes to the markup, CSS, and JS, I would really rather not use conditionals/L10n tags. @flodolo - How soon do you think we could have ample L10n coverage?

flodolo:
It would probably take from 1 to 2 weeks. Also note that we would need a demo server or, if that's not available, some image to give localizers as reference.

flodolo:
Correction: we already have "Features" and "Desktop", hopefully they're used in similar context, and they have both almost perfect coverage.

We should only give some locales a change to translate the new /features page, and the demo/issue image remains for that. Is the content stable and ready to be exposed for localization?

flodolo:
OK, I have a branch ready on www.mozilla.org l10n repository with the new strings, added a screenshot to the bug. Just  need a confirmation that we're good to expose these strings, and that we don't have other urgent requests incoming (will ask Mike about that).

jpetto:
String confirmation should be coming today in [bug 1260466](https://bugzilla.mozilla.org/show_bug.cgi?id=1260466). Demo URL is [https://bedrock-demo-jpetto-fxfamnav.us-west.moz.works/en-US/firefox/features/](https://bedrock-demo-jpetto-fxfamnav.us-west.moz.works/en-US/firefox/features/).

jpetto:
@craigcook - Made some UI improvements to the `/firefox/features/` page. Give it one more look over?

media/css/firefox/features.less
```diff
@@ -0,0 +1,205 @@
...
+        text-align: center;
+    }
+}
+
+#fxfeature-list {
+    .clearfix();
+    list-style-type: none;
+    margin: 0 auto;
+    padding: 0;
+    width: @breakDesktop - (@baseLine * 2);
```
craigcook: For grid-related widths we should probably use `@gridGutterWidth` rather than `@baseLine` (which is type-related so better for vertical spacing). They both happen to be 20px so the result is the same, but the point of making these variables is that they could change. If we someday decided to change `@baseLine` to 24px the grid might break.

media/css/firefox/features.less
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        @media (max-width: @breakTablet) {
+            float: none;
+            margin: 0 0 (@baseLine * 2);
+            min-height: 0;
+            width: auto;
+
+            p {
+                min-height: auto;
```
craigcook: `min-height: auto;` can cause trouble in older browsers (it wasn't valid in CSS 2, only recently added in CSS3). `min-height: 0;` is a safer way to override a min-height.

media/css/firefox/features.less
```diff
@@ -0,0 +1,205 @@
...
+        display: block;
+        padding: @baseLine;
+
+        &:hover {
+            text-decoration: none;
+        }
+    }
+}
+
+@supports (display: flex) {
```
craigcook: Yay for feature queries! I need to start doing more of this.

craigcook:
Looking good, just some nitpicks. r+wc :turtle: 
I guess this is blocked by l10n so shouldn't be merged for a while?

media/css/firefox/features.less
craigcook: Even better!

jpetto:
Yep, need to wait for L10n before merging. I think I've addressed all review comments. If we're good, I'll squash and start on the related FxOS nav.

flodolo:
@jpetto 
We have over 85% coverage on the Features page, I think we're good to merge here.

jpetto:
@craigcook - I don't think there's a reason to wait on the FxOS nav changes prior to merging this PR. I should have the FxOS PR ready to go a couple hours after this merges.

If you've given another once over after yesterday's rebase, feel free to click the green button. :grin:

PR #4012, opened 2016-03-28
Add management command for cloning/updating l10n files

pmclanahan:
This will be good for use locally as well as in CI so we don't have to worry with which repo to clone and docs can be more simple.


pmclanahan:
@jgmize r?

bedrock/settings/base.py
```diff
@@ -952,7 +955,7 @@ def lazy_email_backend():
 }
 
 MOFO_SECURITY_ADVISORIES_PATH = config('MOFO_SECURITY_ADVISORIES_PATH',
-                                       default=path('..', 'mofo_security_advisories'))
+                                       default=path('mofo_security_advisories'))
```
jgmize: I'm not opposed to this change or to the removal in `.bedrock_demo_env`, but why is it part of this commit and PR?

pmclanahan: I got a little cache happy. This helps me speed up demo deploys by making it easier to cache the MFSA repo, then include it in the docker build, then running the command in the cron process in the demos doesn't have to clone the whole repo every deploy. I can split it out though if you'd prefer.

jgmize: Ideally it would make more sense as a separate commit, but I'm comfortable merging as-is and will avoid further pedantry :)

jgmize:
Thanks @pmclanahan

PR #4017, opened 2016-03-30
[fix bug 1260370] Update Hello copy on /firefox/products/ page

alexgibson:
## Description
Updates the line of Hello copy on the /firefox/products page to better describe what the product now delivers.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1260370

## Checklist
- [x] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/family/index.html
+++ b/bedrock/firefox/templates/firefox/family/index.html
@@ -130,7 +130,11 @@
       <li class="product" id="product-hello">
         <a href="{{ url('firefox.hello') }}" data-link-type="button" data-link-name="Firefox Hello">
           <h2>{{ _('Firefox Hello') }}</h2>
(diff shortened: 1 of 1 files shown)


alexgibson:
CC @flodolo - do we already have this line of copy translated elsewhere?

flodolo:
No, we don't :-\

alexgibson:
Would it make things better if we suggested using this? `"Instantly browse any Web page with a friend."`?

flodolo:
We would still the tag, but at least I could import the string from existing files (assuming the space available is roughly the same)

alexgibson:
Discussed on IRC with @flodolo and we're going to keep the new string. Won't be pushed out for translation until later this week, but we're good to review/merge.

schalkneethling:
r+ :monkey:

PR #4020, opened 2016-03-31
[fix bug 1260990] Remove mozilla-video-tools.js from bedrock

alexgibson:
## Description
- Removes the usage of `mozilla-video-tools.js` from bedrock (I was going to lint this file, but removing seems like the better solution altogether :smile:).
- Video helper no longer includes Flash fallback, and uses simple fallback links to video files instead.
- Videos on `/about` and `/mission` now use the lighter `mozilla-video-poster.js` to retain styled poster images & play buttons.
- Videos on `/about` and `/mission` now work correctly in Safari again (no idea when they started needing two clicks? :grimacing:)

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1260990

## Testing
Pushed to demo and ran functional tests for `/about` and `/mission` in Firefox, Chrome and IE and all pass :smile:  

https://bedrock-demo-agibson.us-west.moz.works/en-US/about/
https://bedrock-demo-agibson.us-west.moz.works/en-US/mission/

## Checklist
- [ ] Requires l10n changes.


alexgibson:
Ok, going to mark this one as ready for review.

@pmclanahan - would appreciate some eyes on the Python code I'm removing here, just to make sure I'm not being overzealous in my cleanup?

pmclanahan:
Looks to me like the perfect amount of zealousness was applied :smiling_imp: Nice work!

r+ for the python bits, but I've not put it fully through it's paces, so I'll let someone else do the final r? and merge.

alexgibson:
Sounds good thanks @pmclanahan - this could do with some front-end review from someone else before we merge :+1:

schalkneethling:
Looking good. Tested in IE8 and @alexgibson I am wondering, should we hide the bottom shadow and perhaps add a headline when HTML5 video is not supported?

virtualbox_windows 7 - ie8_04_04_2016_14_18_32

schalkneethling:
r+ /wc :video_camera:

alexgibson:
Seeing as we're talking about dropping IE8 support very soon, I'm going to push back on applying any special styles for it here. Adding a headline is a good suggestion, but it does have l10n implications so I'm not sure what's best here? Both pages do have a video sub-head to help give context already, but adding a site-wide string is more complicated. Thoughts?

alexgibson:
I wonder if we already have a string we could use perhaps, let me check...

schalkneethling:
Agreed. If we do not already have a string, then let's not worry about that.

alexgibson:
@schalkneethling updated using the string "Watch the video", thanks.

schalkneethling:
r+

PR #4023, opened 2016-04-01
Fix Bug 1256668, update newsletter button styles site wide

schalkneethling:
## Description
Updates the button style for all newsletter buttons across the mozilla.org site.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1256668

## Testing
We do not have a public style reference yet, but you can use https://jsfiddle.net/71cewubk/15/ as a general resource here.

Navigate around the site and ensure that all newsletter buttons use a similar style. Also not that the copy for the buttons has changed from "Sign Me Up" to "Sign Up Now"

## Checklist
- [x] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/os/index.html
+++ b/bedrock/firefox/templates/firefox/os/index.html
@@ -45,7 +45,7 @@
 {% block site_header_unwrapped %}
   {% call fxfamilynav('os', 'index') %}
     <div id="cta-sticky">
-      <a href="#email-form-wrapper" class="primary-cta-signup cta-button-small newsletter-signup-toggle hidden" role="button" aria-haspopup="true">{{_('Sign me up')}}</a>
+      <a href="#email-form-wrapper" class="primary-cta-signup cta-button-small newsletter-signup-toggle hidden" role="button" aria-haspopup="true">{% if LANG.startswith('en-') %}{{ _('Sign Up Now') }}{% else %}{{ _('Sign me up') }}{% endif %}</a>
       <a href="#get-device" class="primary-cta-phone cta-button-small hidden" role="button" aria-haspopup="true">{{_('Get a phone')}}</a>
     </div>
   {% endcall %}
--- a/bedrock/mozorg/templates/mozorg/contribute/signup.html
+++ b/bedrock/mozorg/templates/mozorg/contribute/signup.html
@@ -164,7 +164,7 @@
           <label class="inline" for="id_newsletter">
             {{ form.newsletter }}
             {{ form.newsletter.errors }}
-            {{ _('Sign me up for community updates') }}
+            {{ _('Sign up now for community updates') }}
             {% if not LANG.startswith('en') %}{{ _('(English)') }}{% endif %}
           </label>
         </p>
--- a/bedrock/mozorg/templates/mozorg/contribute/studentambassadors/join.html
+++ b/bedrock/mozorg/templates/mozorg/contribute/studentambassadors/join.html
@@ -143,7 +143,7 @@
           </div>
         </section>
         <div class="form-submit">
-          <input type="submit" class="button" id="form-submit" value="{{_('Sign me up!')}}">
+          <input type="submit" class="button-newsletter" id="form-submit" value="{{_('Sign Up Now')}}">
         </div>
       </div>
   </form>
--- a/bedrock/newsletter/templates/newsletter/includes/form.html
+++ b/bedrock/newsletter/templates/newsletter/includes/form.html
@@ -72,7 +72,7 @@
 
     <div class="form-submit">
       <input type="submit" id="footer_email_submit"
-             value="{% if submit_text %}{{ submit_text }}{% else %}{{ _('Sign me up') }} »{% endif %}" class="button">
+             value="{% if submit_text %}{{ submit_text }}{% else %}{% if LANG.startswith('en-') %}{{ _('Sign Up Now') }}{% else %}{{ _('Sign me up') }}{% endif %}{% endif %}" class="button-newsletter {{ button_class }}">
 
       {% if details %}
         <p class="form-details">
--- a/media/css/sandstone/buttons-new.less
+++ b/media/css/sandstone/buttons-new.less
@@ -1,0 +1,62 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../sandstone/lib.less";
+
+@default: #0095dd;
+@defaultHover: #00afe5;
+@dark: #00539F;
+@darkHover: #0095DD;
+@white: #fff;
+@whiteHover: #f4f4f4;
+
+input[type="submit"].button-newsletter,
+.button-newsletter {
+
+    background: none;
+    color: @default;
+    border: 2px solid @default;
+    border-radius: 4px;
+    width: 100%;
+    min-height: 45px;
+    .font-size(18px);
+    font-weight: normal;
+    text-shadow: none;
+    box-shadow: none;
+    -webkit-appearance: none;
+    cursor: pointer;
+
+    &:hover,
+    &:focus {
+        color: @defaultHover;
+        border-color: @defaultHover;
+    }
+
+    &:active {
+        position: relative;
+        top: 1px;
+    }
+
+    &.dark {
+        color: @dark;
+        border-color: @dark;
+
+        &:hover,
+        &:focus {
+            color: @darkHover;
+            border-color: @darkHover;
+        }
+    }
+
+    &.white {
+        color: @white;
+        border-color: @white;
+
+        &:hover,
+        &:focus {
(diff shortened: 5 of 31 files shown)


bedrock/mozorg/templates/mozorg/contribute/signup.html
```diff
@@ -164,7 +164,7 @@ <h2 class="section-title">{{ _('Now tell us a little about yourself:') }}</h2>
           <label class="inline" for="id_newsletter">
             {{ form.newsletter }}
             {{ form.newsletter.errors }}
-            {{ _('Sign me up for community updates') }}
+            {{ _('Sign up now for community updates') }}
```
flodolo: This doesn't affect l10n, since the new sign-up form was never used outside of en-US, see https://www.mozilla.org/en-GB/contribute/signup/ vs https://www.mozilla.org/en-US/contribute/signup/

bedrock/mozorg/templates/mozorg/contribute/studentambassadors/join.html
```diff
@@ -143,7 +143,7 @@
           </div>
         </section>
         <div class="form-submit">
-          <input type="submit" class="button" id="form-submit" value="{{_('Sign me up!')}}">
+          <input type="submit" class="button-newsletter" id="form-submit" value="{{_('Sign Up Now')}}">
```
flodolo: Same here, page is not localized.

bedrock/newsletter/templates/newsletter/includes/form.html
```diff
@@ -72,7 +72,7 @@
 
     <div class="form-submit">
       <input type="submit" id="footer_email_submit"
-             value="{% if submit_text %}{{ submit_text }}{% else %}{{ _('Sign me up') }} »{% endif %}" class="button">
+             value="{% if submit_text %}{{ submit_text }}{% else %}{{ _('Sign Up Now') }}{% endif %}" class="button-newsletter {{ button_class }}">
```
flodolo: This one affects l10n, and you need to put it behind a en-* conditional like the `Keep up with<br> all things Firefox.Keep up with<br> all things Firefox.`string, since we can't have tags. Once we have coverage, we need to clean up the conditionals.

I suggest to add the code to clean up to this bug 
https://bugzilla.mozilla.org/show_bug.cgi?id=1258998

flodolo:
Unless I missed some string changes, there's only one affecting l10n, and it needs a en-* conditional

media/css/sandstone/sandstone-resp.less
```diff
@@ -2,6 +2,8 @@
 @import "reset.less";
 @import "fonts.less";
 @import "buttons.less";
+@import "buttons.less";
```
alexgibson: Remove duplicate `buttons.less` import?

media/css/sandstone/sandstone-resp.less
```diff
@@ -2,6 +2,8 @@
...
+@import "buttons-new.less";
```
alexgibson: Nit: Is this a new file which will eventually replace `buttons.less`? It makes sense now, but naming things `new` often only makes sense in the short term. Is there something else we could call it that might stand the test of time maybe?

media/css/sandstone/buttons-new.less
```diff
@@ -0,0 +1,55 @@
...
+        }
+    }
+
+    &.white {
+        color: #fff;
+        border-color: #fff;
+
+        &:hover,
+        &:focus {
+            color: #f4f4f4;
```
alexgibson: More of a design/UX comment, but I find it pretty hard to tell the difference when the hover style is applied to the white button class?

bedrock/firefox/templates/firefox/os/index.html
```diff
@@ -45,7 +45,7 @@
 {% block site_header_unwrapped %}
   {% call fxfamilynav('os', 'index') %}
     <div id="cta-sticky">
-      <a href="#email-form-wrapper" class="primary-cta-signup cta-button-small newsletter-signup-toggle hidden" role="button" aria-haspopup="true">{{_('Sign me up')}}</a>
+      <a href="#email-form-wrapper" class="primary-cta-signup cta-button-small newsletter-signup-toggle hidden" role="button" aria-haspopup="true">{{_('Sign Up Now')}}</a>
```
alexgibson: Does this one need an l10n flag or conditional?

flodolo: Conditional (at this point).

alexgibson:
I ran `test_newsletter_embed.py` and `test_newsletter.py` against your deployed instance on demo4 and tests pass :+1:

alexgibson:
Signup button is maybe a bit too larger here? https://www-demo4.allizom.org/en-US/newsletter/

media/css/sandstone/sandstone-resp.less
schalkneethling: whoops

media/css/sandstone/sandstone-resp.less
schalkneethling: it will replace `buttons.less`

alexgibson:
Signup button has no styling at all here: https://www-demo4.allizom.org/en-US/firefox/whatsnew/

media/css/sandstone/buttons-new.less
schalkneethling: Yup, we have discussed this but, no decisions has been made so, we are going with this for now.

schalkneethling:
Dang it, I missed the `whatsnew` pages entirely.

alexgibson:
Looks like we still have a newsletter variation test on the homepage that needs updating https://www-demo4.allizom.org/en-US/?v=b

alexgibson:
Do we need to update any of the styles in the newsletter center too, or are they out of scope for this PR? https://www-demo4.allizom.org/en-US/newsletter/recovery/

schalkneethling:
@alexgibson ok so, the what's new pages does not include sandstone and that is why the button has no style. Good catch! I did not have any of those what's new one's in the spreadsheet :-/

The newsletter recovery is out of scope for this one but, good to know about that page. Thanks!

alexgibson:
> @alexgibson ok so, the what's new pages does not include sandstone and that is why the button has no style. Good catch! I did not have any of those what's new one's in the spreadsheet :-/

All pages should have sandstone already, whatsnew pages included?

schalkneethling:
Looking at this, it seems like there are some that does not:

```
'firefox_tour_none': {
        'source_filenames': (
            'css/firefox/australis/australis-page-common.less',
            'css/firefox/sync-animation.less',
            'css/firefox/australis/australis-page-stacked.less',
        ),
        'output_filename': 'css/firefox_tour_none-bundle.css',
    },
```
If I change it to:

```
'firefox_tour_none': {
        'source_filenames': (
            'css/firefox/australis/australis-page-common.less',
            'css/firefox/sync-animation.less',
            'css/sandstone/buttons-new.less',
            'css/firefox/australis/australis-page-stacked.less',
        ),
        'output_filename': 'css/firefox_tour_none-bundle.css',
    },
```
The button is styles correctly.

alexgibson:
It looks like the firefox friends signup form is inheriting the height of the new email input, but that makes it disproportionate to the other fields. Perhaps we need to override the height change for this form?

friends-signup

media/css/sandstone/buttons-new.less
```diff
@@ -0,0 +1,55 @@
...
+    .font-size(18px);
+    font-weight: normal;
+    text-shadow: none;
+    box-shadow: none;
+    -webkit-appearance: none;
+    cursor: pointer;
+
+    &:hover,
+    &:focus {
+        color: #00afe5;
```
alexgibson: It's a small thing, but if these are to be common colors and repeated may they be worth assigning to variables?

alexgibson:
A few stragglers aside, but overall looking really good. I like the new larger font size for the buttons especially :+1:

media/css/sandstone/buttons-new.less
schalkneethling: I agree but, because there will be another round of revisions once folks have talked to the Firefox and identity designers, we might want to hold of just a tad.

alexgibson: If the colors could change, using a variable makes even more sense to me. But it's no big deal.

bedrock/firefox/templates/firefox/os/index.html
alexgibson: This still need a conditional

media/css/sandstone/buttons-new.less
schalkneethling: Agree, I was thinking more of naming the variables. Currently we have things like Firefox Red, Mozilla Blue etc. with that said it may, should not?,  be tied to tightly to the identity language as that might change as well. I am open to either though

alexgibson: I was thinking just similar to the class names you've already used may be enough?

alexgibson:
@schalkneethling LGTM, squish time?

alexgibson:
r+

PR #4027, opened 2016-04-01
[fix bug 1261078] Add tests to new /contribute/friends/ page

jpetto:
## Description
Adding functional tests to recently updated `/contribute/friends/` page. **Heavily** influenced by [existing newsletter tests](https://github.com/mozilla/bedrock/blob/master/tests/functional/test_newsletter.py).

## Bugzilla
[bug 1261078](https://bugzilla.mozilla.org/show_bug.cgi?id=1261078)

## Testing
Just run the tests:

`py.test --base-url http://localhost:8000 --driver Firefox --html tests/functional/results.html tests/functional/contribute/test_friends.py`

## Checklist

### DIFF
--- a/tests/functional/contribute/test_friends.py
+++ b/tests/functional/contribute/test_friends.py
@@ -4,12 +4,39 @@
 
 import pytest
 
+from selenium.common.exceptions import TimeoutException
 from pages.contribute.friends import ContributeFriendsPage
 
 
 @pytest.mark.smoke
 @pytest.mark.nondestructive
-def test_display_signup_form(base_url, selenium):
+def test_signup_default_values(base_url, selenium):
     page = ContributeFriendsPage(base_url, selenium).open()
     page.click_show_signup_form()
     assert page.is_signup_form_displayed
+    assert '' == page.email
+    assert 'United States' == page.country
+    assert page.html_format_selected
+    assert not page.text_format_selected
+    assert not page.privacy_policy_accepted
+    assert page.is_privacy_policy_link_displayed
+
+
+@pytest.mark.nondestructive
+def test_successful_sign_up(base_url, selenium):
+    page = ContributeFriendsPage(base_url, selenium).open()
+    page.click_show_signup_form()
+    page.type_email('success@example.com')
+    page.select_country('Germany')
+    page.select_text_format()
+    page.accept_privacy_policy()
+    page.click_sign_me_up()
+    assert page.sign_up_successful
+
+
+@pytest.mark.nondestructive
+def test_sign_up_fails_when_missing_required_fields(base_url, selenium):
+    page = ContributeFriendsPage(base_url, selenium).open()
+    page.click_show_signup_form()
+    with pytest.raises(TimeoutException):
+        page.click_sign_me_up()
--- a/tests/pages/contribute/friends.py
+++ b/tests/pages/contribute/friends.py
@@ -3,6 +3,8 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as expected
+from selenium.webdriver.support.select import Select
 
 from pages.contribute.base import ContributeBasePage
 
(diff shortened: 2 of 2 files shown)


tests/functional/contribute/test_friends.py
```diff
@@ -13,3 +14,29 @@ def test_display_signup_form(base_url, selenium):
     page = ContributeFriendsPage(base_url, selenium).open()
     page.click_show_signup_form()
     assert page.is_signup_form_displayed
+    assert '' == page.email
+    assert 'United States' == page.country
+    assert page.html_format_selected
```
alexgibson: Nit: as we're testing more than opening the form now, suggest renaming this test to `test_signup_default_values`?

alexgibson:
The form submission test currently fails against www-dev.allizom.org, likely a basket related issue?

tests/functional/contribute/test_friends.py
```diff
@@ -13,3 +14,29 @@ def test_display_signup_form(base_url, selenium):
...
+    assert not page.text_format_selected
+    assert not page.privacy_policy_accepted
+    assert page.is_privacy_policy_link_displayed
+
+
+@pytest.mark.flaky(reruns=1)
+def test_successful_sign_up(base_url, selenium):
+    page = ContributeFriendsPage(base_url, selenium).open()
+    page.click_show_signup_form()
+    page.type_email('thedude@urbanachievers.org')
```
alexgibson: Nit: suggest using `noreply@mozilla.com` for consistency with other tests

alexgibson:
A couple of nits, but otherwise looks good :+1: 

Only blocker is making sure the form works on our dev environment, which is likely a basket related issue.

jpetto:
@alexgibson - Updated based on review. Thanks!

@pmclanahan - Based on conversations last week, I'm guessing the dev failures are indeed a basket issue. Care to comment?

alexgibson:
If the tests fail on dev then we can't merge this as-is. Either we need to point bedrock dev instances to basket stage (if that's the new thing), or we fix basket dev so we can run tests against it.

jpetto:
Hm, yeah I believe basket dev is now used exclusively for new platform testing (see [https://bugzilla.mozilla.org/show_bug.cgi?id=1254685#c5](https://bugzilla.mozilla.org/show_bug.cgi?id=1254685#c5)).

pmclanahan:
The issue is that the dev db and the dev ExactTarget account are not kept up to date. But the issue with using stage or prod is that it's using the prod ExactTarget account and thus sending emails or signing up users for real. Not sure how to handle this. My preference would be for these tests to not rely on a working basket, but I'm not sure how to accomplish that.

alexgibson:
So is the answer that we should not be running form submission tests at all, given our current infrastructure? This seems less than ideal to me, but we seem to come round to this convo periodically.

alexgibson:
Adding a blocked label for now until we can get our basket dev woes mitigated.

alexgibson:
I believe now that #4034 has merged, if we rebase this branch and then update requirements, we should be able to try using the new test-specific email address for form submission.

jpetto:
Rebased, updated test req's, and changed test email to `success@example.com`. Ran tests against `localhost` and `https://www-dev.allizom.org` and all passed. Anything else we need to check prior to merge?

If we're good here, I'll file a bug to update existing email addresses in other functional tests.

tests/functional/contribute/test_friends.py
```diff
@@ -4,12 +4,39 @@
...
     assert page.is_signup_form_displayed
+    assert '' == page.email
+    assert 'United States' == page.country
+    assert page.html_format_selected
+    assert not page.text_format_selected
+    assert not page.privacy_policy_accepted
+    assert page.is_privacy_policy_link_displayed
+
+
+@pytest.mark.flaky(reruns=1)
```
alexgibson: We should be able to remove the `flaky` marker now, and seeing as this won't touch ET replace it with the `nondestructive` marker.

alexgibson:
Passes for me against dev :balloon: r+

PR #4030, opened 2016-04-04
Lint contribute page JS files

alexgibson:
## Description
Lint `/contribute` pages JS files.

## Bugzilla
N/A

## Testing
I must admit, I'm a little lost on all these different versions of the contribute pages. As far as I can tell, some are not used - but I don't know what the long term plan is for these pages so linted them anyway.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/js/mozorg/contribute/contribute-2015.js
+++ b/media/js/mozorg/contribute/contribute-2015.js
@@ -2,15 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-;(function($) {
+(function($) {
     'use strict';
 
     var $window = $(window);
     var $document = $(document);
     var $body = $('body');
     var $navList = $('#contribute-nav-menu');
-
-    var wideMode = false;
     var hasMediaQueries = (typeof matchMedia !== 'undefined');
 
     // If the browser supports media queries, check the width onload and onresize.
@@ -22,17 +20,14 @@
             this.resizeTimeout = setTimeout(checkWidth, 200);
         });
     } else {
-        wideMode = true;
         $body.removeClass('thin').addClass('wide');
     }
 
     function checkWidth() {
         if (window.matchMedia('screen and (min-width: 761px)').matches) {
-            wideMode = true;
             $body.removeClass('thin').addClass('wide');
             $navList.removeAttr('aria-hidden').show();
         } else {
-            wideMode = false;
             $body.removeClass('wide').addClass('thin');
             $navList.attr('aria-hidden', 'true').hide();
         }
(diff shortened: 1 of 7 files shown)


alexgibson:
Ok, I ran the current functional tests that we have for the most recent, `/contribute` page and they pass ok. I also manually viewed the older "faces video" page and that seems ok too.

The one thing I'm not sure about if this page: https://www.mozilla.org/en-US/contribute/tasks/. It doesn't seem to be linked to from anywhere afaict. Seems ok too, but it would be good to get someone who is familiar with the history of these pages to take a second look.

@schalkneethling or @craigcook r?

schalkneethling:
Hmmm, I have not seen that page before, and it is not part of the tests we did for the rework of the /signup page so, might have been an older initial stab at the more recent work? I did not work on it so, I reckon @craigcook will know more there.

media/js/mozorg/contribute/contribute-2015.js
```diff
@@ -149,10 +144,10 @@
         if ($.inArray(categoryarea, areas) !== -1) {
             if ($areascontainer.is(':hidden')) {
```
schalkneethling: Following the standard naming convention we do else where, should these not perhaps be `categoryArea` and `$areasContainer`?

media/js/mozorg/contribute/contribute-2015.js
```diff
@@ -193,10 +188,10 @@
     });
 
     // Show the specific area for the selected category
-    var show_area = function(categoryarea) {
+    var showArea = function(categoryarea) {
         // Get the ID of the previously selected area
-        var oldarea_id = $('.area:visible').attr('id');
-        var $oldarea = $('#' + oldarea_id);
+        var oldareaId = $('.area:visible').attr('id');
```
schalkneethling: maybe `oldAreaId`?

alexgibson:
@schalkneethling updated thanks

craigcook:
@schalkneethling as far as I can tell you built those task pages back in November (https://github.com/mozilla/bedrock/commit/d8f95ee37e70b2b3a4f27e88591698d92607b6e4). It was using an optimizely redirect that may have since been switched off so it's possible we're not sending any traffic to those page variants currently. I don't know the conclusion of that experiment. Maybe Jen knows?

craigcook:
Oh I see now, that particular page and a few others are indeed from an older stab, https://github.com/mozilla/bedrock/commit/f36e60b081602fffd3356f344ac1a6a1014a1eb5. At this point I think it's nothing but debt that needs to be cleaned up.

schalkneethling:
Alrighty, thanks @craigcook so, then we should just create a bug to remove that as some point but, that means this is safe to merge. \o/

schalkneethling:
@alexgibson care to squash?

alexgibson:
@schalkneethling squished, thanks!

PR #4034, opened 2016-04-04
Fix bug 1261886: Upgrade basket-client

pmclanahan:
v0.3.12 brings special testing email addresses:

* `success@example.com` will not hit basket and will return success.
* `failure@example.com` will not hit basket and will raise an exception.

These should be used to test live email forms for UX regresions.


pmclanahan:
@jgmize r?

@alexgibson @jpetto FYI

PR #4035, opened 2016-04-05
Lint common JS files and test specs

alexgibson:
## Description
- Lints our common JS files that are shared in many pages.
- Lints our JS test specs and adds them to the ESLint grunt task.

## Bugzilla
N/A

## Testing
I've given this some manual testing, and changes seem pretty safe. I guess the main thing is to make sure any renamed functions aren't also called from anywhere else that I may have missed. I've done a check and think all is ok, but a second sweep would be good.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/tests/unit/spec/base/mozilla-accordion.js
+++ b/tests/unit/spec/base/mozilla-accordion.js
@@ -3,7 +3,8 @@
  * Sinon docs: http://sinonjs.org/docs/
  */
 
-/* global describe, beforeEach, afterEach, it, expect, sinon, spyOn */
+/* global describe, beforeEach, afterEach, it, expect */
+/* eslint no-unused-vars: [2, { "varsIgnorePattern": "accordion" }] */
 
 describe('mozilla-accordion.js', function () {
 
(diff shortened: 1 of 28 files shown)


tests/unit/spec/base/mozilla-accordion.js
```diff
@@ -14,33 +15,33 @@ describe('mozilla-accordion.js', function () {
...
+                    '<div data-accordion-role="tabpanel"><p>Section 1 contents.</p></div>' +
+                '</section>' +
+                '<section>' +
+                    '<h2 data-accordion-role="tab">Section 2</h2>' +
+                    '<div data-accordion-role="tabpanel"><p>Section 2 contents.</p></div>' +
+                '</section>' +
+                '<section>' +
+                    '<h2 data-accordion-role="tab">Section 3</h2>' +
+                    '<div data-accordion-role="tabpanel"><p>Section 3 contents.</p></div>' +
+                '</section>' +
```
schalkneethling: Wait so, maybe I am not reading this correctly but, seeing you are now creating a string by concatenation, does this still need the array and the call to `join`?

alexgibson: I just forgot to remove the join most likely

schalkneethling: Aha, so I am not going crazy ;) There is another instance of it in a file further down.

tests/unit/spec/firefox/australis/browser-tour.js
```diff
@@ -25,71 +26,71 @@ describe('browser-tour.js', function() {
...
+            '<div id="ui-tour-mask" tabindex="-1">' +
+                '<div class="mask-inner">' +
+                    '<div class="logo"></div>' +
+                    '<div class="stage">' +
+                        '<h1>Mask title</h1>' +
+                        '<p>Mask sub heading</p>' +
+                    '</div>' +
+                '</div>' +
+            '</div>'
         ].join();
```
schalkneethling: over here.

schalkneethling:
Browsed around the site in Fx, Chrome and IE8, all works as expected. Ran Grunt tests, all pass. So, other than the two items mentioned r+ :100:

alexgibson:
@schalkneethling updated thanks

PR #4036, opened 2016-04-05
Lint plugincheck JS files

alexgibson:
## Description
- Lint plugincheck JS files
- Moves `mustache.js` to third-party `/media/js/libs` directory, so we don't have to explicitly ignore it

## Bugzilla
N/A

## Testing
Pretty minor changes, other than moving one file.

## Checklist
- [ ] Requires l10n changes.


media/js/plugincheck/lib/plugincheck.js
```diff
@@ -273,9 +272,11 @@
                 callback(pluginList);
 
             }).fail(function(xhr, status, error) {
+                /* eslint-disable no-console */
```
schalkneethling: oh that is neat! Love that one can do that.

schalkneethling:
r+ :zap:

PR #4037, opened 2016-04-05
Bug 1260370 - Follow-up: use underscores in l10n tag for consistency

flodolo:


flodolo:
@alexgibson Quick r? 

I realized while preparing the file that the tag uses dashes. It shouldn't be an issue anymore, but I prefer to keep using underscores for consistency.

alexgibson:
r+ just waiting on CircleCI

flodolo:
Thanks :cake:

PR #4038, opened 2016-04-05
Lint remaining mozorg app JS files

alexgibson:
## Description
- Lints remaining JS files in general mozorg app (only Firefox app remain after this PR!)
- Removes `existing.js` since it has been broken for some time, and serves only for GA purposes.

## Bugzilla
N/A

## Testing
Fairly minor changes to a few pages, mostly cosmetic and renaming of variables. But some manual testing of a few key pages wouldn't hurt.

## Checklist
- [ ] Requires l10n changes.


bedrock/settings/static_media.py
```diff
@@ -1006,12 +1006,6 @@
         ),
         'output_filename': 'css/contribute-signup-bundle.js',
     },
-    'existing': {
-        'source_filenames': (
-            'js/newsletter/existing.js',
```
alexgibson: Note: I'm removing this file as all it seems to be responsible for is GA tracking, and has been broken since before we migrated to GTM. I'm not sure how to best re-implement it for GTM without sending crap loads of events, so it would likely need re-working anyhow. No body seems to have missed the data in well over a year.

alexgibson:
Ok, I think this one is ready for review.

alexgibson:
This is also up on a demo for convenience: https://bedrock-demo-agibson.us-west.moz.works/en-US/

schalkneethling:
r+ :package:

PR #4040, opened 2016-04-05
[fix bug 1259851] Create custom FxOS nav.

jpetto:
## Description
A custom navigation for FxOS pages. Based off the Fx Family nav, yet completely decoupled as the Fx Family nav will likely continue to evolve.

## Bugzilla
[bug 1259851](https://bugzilla.mozilla.org/show_bug.cgi?id=1259851)

## Testing
Test all pages with new nav on all browsers. Code is a simplified version of the Fx Family nav, but we all know how risky it is to serve copy pasta without a taste test.

## Checklist
- [ ] Requires l10n changes.

**Note** - will add functional tests


bedrock/base/templates/macros.html
```diff
@@ -247,6 +247,136 @@
...
+  }
+#}
+
+<header id="fxosnav-header" class="fxosnav-header{% if (navclass) %} {{ navclass }}{% endif %}">
+  <div class="container">
+    <a href="{{ url('firefox.os.index') }}" class="fxosnav-logo">{{ _('Firefox') }}</a>
+    <nav id="fxosnav" class="fxosnav" role="navigation">
+      <ul id="fxosnav-primary" class="primary-nav">
+        <li{% if activenav == 'fxos' %} class="active"{% endif %}>
+          <a data-id="desktop" href="{{ url('firefox.os.index') }}" class="primary-link" data-link-type="nav" data-link-name="FxOS"><div>{{ _('FirefoxOS') }}</div></a>
```
flodolo: Never seen "FirefoxOS" written without a space.

jpetto:
Not sure how I missed the space between `Firefox` and `OS`...

flodolo:
Just to clarify: why are we updating the nav is we're trying to [get rid of all the OS references](https://bugzilla.mozilla.org/show_bug.cgi?id=1260807) at the same time?

jpetto:
Because the FxOS pages are slated to be around until 2018 (last I heard), and the Fx Family nav doesn't really apply to these pages anymore. The FxOS pages still need to be accessible (though I couldn't say to whom...).

flodolo:
If I'm correct the only string we're missing is `TV`.

flodolo:
I have the file ready in local with only `TV` added, please ping me if any string changes in this PR.

jpetto:
> I have the file ready in local with only `TV` added, please ping me if any string changes in this PR.

@flodolo - Just clarifying that we are no longer blocked by L10n?

flodolo:
Yes, I think you should go ahead and don't block given the visibility of this "TV".

alexgibson:
As this PR adds some new JS, you may want to rebase off current master to see if ESLint picks anything out :smile:

jpetto:
@craigcook - Added tests and rebased on master. Should now be good for a review. Thanks!

media/css/firefox/fxos-nav.less
```diff
@@ -0,0 +1,576 @@
...
+    }
+}
+
+// big left hand logo
+.fxosnav-logo {
+    .fxosnav-sprite(0);
+    .image-replaced();
+    .transform(translateY(30px));
+    .transition(-webkit-transform 0.3s ease);
+    .transition(transform 0.3s ease);
```
craigcook: I'm not sure repeating this mixin actually works. The output is:
```
-webkit-transition: -webkit-transform 0.3s ease;
transition: -webkit-transform 0.3s ease;
-webkit-transition: transform 0.3s ease;
transition: transform 0.3s ease;
```
I'm not sure of the support overlap with `-webkit-transition` and `-webkit-transform` (if there's any version of webkit that supports one but not the other). As it is, browsers that only support `-webkit-transition` will try to apply it to `transform` which they may not support.

For proper full coverage you'd really need:
```
-webkit-transition: -webkit-transform 0.3s ease, transform 0.3s ease;
transition: -webkit-transform 0.3s ease, transform 0.3s ease;
```
But at this point I wonder if we need to worry much about old webkits for this kind of minor enhancement.

craigcook:
Looks good apart from that repeated mixin, which isn't necessarily a blocker. r+wc

media/css/firefox/fxos-nav.less
jpetto: Ah, good catch. The only browser I'd really be concerned with is [Safari 8](http://caniuse.com/#search=transform), which may still have a decent user base. I think I'll forgo the mixin and just hand-code both the browser-prefixed and standard versions.

Any argument to that approach?

jpetto: On second thought, I agree with you that worrying about a transition for an old webkit browser isn't really worth changing up the code. The `transform` will still work. Will simply remove `.transition(-webkit-transform 0.3s ease);`

jpetto:
Note - also removed `.transition(-webkit-transform 0.3s ease);` from `family-nav.less`.

media/css/firefox/fxos-nav.less
craigcook: Yeah, if it was an important interaction I'd say go for maximum cross-browser robustity, but this is just moving the logo. If it jumps suddenly in older browsers and slides smoothly in newer ones, that's totally in keeping with progressive enhancement.

PR #4058, opened 2016-04-10
Fix Bug 1263445 - default ios version should be 3.0 instead of 2.0 in release notes

kyoshino:


kyoshino:
```
src/_pylibmcmodule.h:42:36: fatal error: libmemcached/memcached.h: No such file or directory
```
:confused:

jgmize:
The Circleci failure should be fixed by #4059

jpetto:
Looks good to me! :lollipop: 

Waiting on current build to finish before merging...

pollti:
Thank you very much!

PR #4060, opened 2016-04-11
Fix Bug 1263732 - Update Open Sans to eliminate kerning error, serve WOFF2

kyoshino:


flodolo:
Can we make sure this doesn't reintroduce issues like https://bugzilla.mozilla.org/show_bug.cgi?id=907793 or https://bugzilla.mozilla.org/show_bug.cgi?id=1115727 ?

jfkthame:
I checked a random one of the files in the PR (OpenSans-Light-webfont.woff, as it happens), and it looks like the incorrect Romanian glyph substitution (as described in https://bugzilla.mozilla.org/show_bug.cgi?id=907793#c10) is still present. Which presumably means this has not been fixed by the upstream Open Sans maintainers for any of the faces, and deploying these files will cause bug 907793 to reappear on the mozilla.org properties.

kyoshino:
Forgot about Bug 907793 :confused: Will figure out what should I do here. It would be perhaps better to use the upstream/official Open Sans along with a modified subset for Romanian. I have implemented the same hack for Fulah before: https://github.com/mozilla/bedrock/tree/master/media/fonts/l10n

kyoshino:
We need these 4 characters, right?
screen shot 2016-04-12 at 12 57 35 pm

kyoshino:
So I have added a modified subset for Romanian, in addition to the official Open Sans font family.

craigcook:
Since this update renames all the existing fonts I believe the old ones will still be cached on the CDN, so any external sites loading Open Sans by their old camelcase names will continue to get the old fonts, including the error. Then eventually the CDN will update and delete those files and external sites would start to get 404s for their fonts until they update their CSS.

It seems best to keep the original filenames and either force the CDN to refresh now or just let it refresh in time and external sites would continue get the kerning error for a while longer (we'll still get the immediate benefit of woff2 on mozorg).

craigcook:
Disregard my previous comment. According to @pmclanahan it should be safe to rename these files. External sites still hotlinking fonts from `mozorg.cdn.mozilla.net` will continue to get the old fonts (including the kerning error, sadly) regardless of what we do here. I'm generally in favor of standardizing all-lowercase filenames anyway so let's go for it.

In that case I think this is good to merge so a tentative r+ from me, but I wouldn't mind a second opinion (@schalkneethling, @alexgibson, @jpetto ?)

The Romanian subset does the trick for those characters.

jpetto:
This change looks good to me.

PR #4062, opened 2016-04-13
Upgrade product-details to latest backported for Django 1.7

pmclanahan:


pmclanahan:
@jgmize r?

pmclanahan:
Circle failed because the tarball doesn't contain a copy of the JSON like the package does. I'll fix that. 

Also this is based on [this product details branch](https://github.com/mozilla/django-product-details/tree/django-1.7-backport).

jgmize:
Thanks @pmclanahan

PR #4066, opened 2016-04-15
Bug 1264843 - embed FxA server in China within Fx China repack

l-hedgehog:
### Guidelines
Before you submit a pull request, please make sure you have done the following:

- [x] Review the guidelines for contributing to this repository (link above in GitHub ^).
- [x] All commit messages should be well written, and reference a Bugzilla bug number where applicable.

Note: If you're unsure about anything here or below, don't worry! Just fill out as much
information as you can.

## Description
Please explain the changes you're making, complete with a description of the issue and
a summary of the proposed solution.

Please refer to the bug description.

## Bugzilla
Include a direct link to the associated bug on Bugzilla.

https://bugzil.la/1264843

## Testing
Please leave some directions for the reviewer on how to test your pull request. List which
pages need attention, and what the reviewer should pay close attention to. If the change
requires testing in any specific browsers or platforms, please provide a list.

Just visit an modified page and append ?dist=mozillaonline to its url.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -950,6 +950,10 @@ OPTIMIZELY_PROJECT_ID = config('OPTIMIZELY_PROJECT_ID', default='')
 FXA_IFRAME_SRC = config('FXA_IFRAME_SRC',
                         default='https://accounts.firefox.com/')
 
+# Bug 1264843: embed FxA server in China within Fx China repack
+FXA_IFRAME_SRC_MOZILLAONLINE = config('FXA_IFRAME_SRC_MOZILLAONLINE',
+                                      default='https://accounts.firefox.com.cn/')
+
 # Google Play and Apple App Store settings
 from .appstores import (GOOGLE_PLAY_FIREFOX_LINK,  # noqa
                         APPLE_APPSTORE_FIREFOX_LINK, APPLE_APPSTORE_COUNTRY_MAP)
--- a/media/js/base/mozilla-client.js
+++ b/media/js/base/mozilla-client.js
@@ -205,10 +205,10 @@ if (typeof Mozilla === 'undefined') {
             }
 
             window.clearTimeout(timer);
-            onRetrieved(true, event.detail.data.version, event.detail.data.defaultUpdateChannel);
+            onRetrieved(true, event.detail.data.version, event.detail.data.defaultUpdateChannel, event.detail.data.distribution);
         };
 
-        var onRetrieved = function (accurate, version, channel) {
+        var onRetrieved = function (accurate, version, channel, distribution) {
             document.removeEventListener('mozUITourResponse', listener);
 
             var isESR = channel === 'esr';
@@ -217,6 +217,7 @@ if (typeof Mozilla === 'undefined') {
                 'accurate': accurate,
                 'version': version,
                 'channel': channel,
+                'distribution': distribution,
                 'isUpToDate': isUpToDate,
                 'isESR': isESR
             };
@@ -226,7 +227,7 @@ if (typeof Mozilla === 'undefined') {
 
         // Prepare fallback function in case the API doesn't work
         var userVer = Client._getFirefoxVersion();
-        var fallback = function () { onRetrieved(false, userVer, 'release'); };
+        var fallback = function () { onRetrieved(false, userVer, 'release', undefined); };
 
         // If Firefox is old or for Android, call the fallback function immediately because the API is not implemented
         if (parseFloat(userVer) < 35 || Client._isFirefoxAndroid()) {
--- a/media/js/base/mozilla-fxa-iframe.js
+++ b/media/js/base/mozilla-fxa-iframe.js
@@ -22,11 +22,18 @@ Mozilla.FxaIframe = (function() {
     // host domain for FxA
     var _host;
 
+    // alternative host domain for FxA within partner distribution
+    var _distHost;
+
     // reference to <iframe> element containing FxA
     var _$iframe;
 
     // stores full URL to FxA to be used for <iframe> src attribute
     var _src;
+
+    var _removeTrailingSlash = function(host) {
+        return (host[host.length - 1] === '/') ? host.substr(0, host.length - 1) : host;
+    };
 
     var _init = function(userConfig) {
         // store user supplied config
@@ -45,7 +52,7 @@ Mozilla.FxaIframe = (function() {
         }
 
         // make sure _host URL doesn't have a trailing slash
-        _host = (_host[_host.length - 1] === '/') ? _host.substr(0, _host.length - 1) : _host;
+        _host = _removeTrailingSlash(_host);
 
         // check user's Fx version to determine FxA iframe experience
         if (Mozilla.Client.FirefoxMajorVersion >= 46) {
@@ -54,6 +61,15 @@ Mozilla.FxaIframe = (function() {
 
         // initialize GA event name
         _gaEventName = _config.gaEventName || 'fxa';
+
+        if (_config.distribution && _config.distribution !== 'default') {
+            _distHost = $('#fxa-iframe-config').data(_config.distribution.toLowerCase() + 'Host');
+            if (_distHost) {
+                _distHost = _removeTrailingSlash(_distHost);
+                _src = _src.replace(_host, _distHost);
+                _host = _distHost;
+            }
+        }
 
         // call iframe to life by setting src attribute
         _$iframe.attr('src', _src);
--- a/bedrock/firefox/templates/firefox/accounts.html
+++ b/bedrock/firefox/templates/firefox/accounts.html
@@ -34,7 +34,7 @@
     </ol>
     <p class="signed-in">{{_('You’re signed in and ready to start using Sync.')}}</p>
   </div>
-  <aside id="fxa-iframe-config" data-host="{{ settings.FXA_IFRAME_SRC }}">
+  <aside id="fxa-iframe-config" data-host="{{ settings.FXA_IFRAME_SRC }}" data-mozillaonline-host="{{ settings.FXA_IFRAME_SRC_MOZILLAONLINE }}">
     <iframe id="fxa" scrolling="no" data-src="{{ settings.FXA_IFRAME_SRC }}?utm_campaign=fxa-embedded-form&amp;utm_medium=referral&amp;utm_source=accounts-page&amp;utm_content=fx-{{ version }}&amp;entrypoint=accounts-page&amp;service=sync&amp;context=iframe&amp;style=chromeless&amp;haltAfterSignIn=true"></iframe>
     <div class="mobile">
       <a class="button dark firefox-ios more" rel="external" href="https://support.mozilla.org/kb/sync-firefox-bookmarks-and-browsing-history-iOS">{{_('Set up Sync for iOS')}}</a>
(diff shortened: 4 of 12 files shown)


pmclanahan:
same deal as #4067

bedrock/mozorg/tests/test_helper_misc.py
```diff
@@ -593,3 +598,24 @@ def test_firefox_ios_url_param(self):
...
+class TestFxAIFrameSrc(TestCase):
+    rf = RequestFactory()
+
+    def _render(self, path='/'):
+        req = self.rf.get(path)
+
+        return render("{{ fxa_iframe_src() }}", {'request': req})
+
+    def test_fxa_iframe_src_no_dist(self):
+        eq_(self._render('/'), 'https://accounts.firefox.com/')
```
jpetto: Since different environments need different values for the FxA iframe src, could we use `settings.FXA_IFRAME_SRC` in place of the hard-coded FxA production URL?

bedrock/settings/base.py
```diff
@@ -937,6 +937,10 @@ def lazy_email_backend():
 FXA_IFRAME_SRC = config('FXA_IFRAME_SRC',
                         default='https://accounts.firefox.com/')
 
+FXA_IFRAME_SRC_DIST_MAP = {
```
jpetto: For future updates/reference, could we add a short description of the use of this dict along with a bug number?

jpetto:
In terms of FxA user experience, do we need to be concerned with `?dist=mozillaonline` links being shared/accessed outside of China?

jpetto:
Pinging @shane-tomlinson to see if there are any concerns/thoughts from the FxA side.

jpetto:
Also cc @rfk

rfk:
Thanks for looping me in @jpetto!

> do we need to be concerned with ?dist=mozillaonline links being shared/accessed outside of China

How likely is this to happen in practice, e.g. would it only happen if the user copy-pastes from the URL bar during the first-run flow?

I believe that accounts.firefox.com.cn is quite visually distinct from accounts.firefox.com, so this would hopefully reduce the potential for confusion.

bedrock/mozorg/tests/test_helper_misc.py
l-hedgehog: @jpetto I see. Or maybe I should just override `FXA_IFRAME_SRC` for this test, just like #4067 ?

bedrock/settings/base.py
l-hedgehog: I'll add comments to both PR later

bedrock/mozorg/tests/test_helper_misc.py
jpetto: Yep, overriding `FXA_IFRAME_SRC` for the test should also work.

jpetto:
> How likely is this to happen in practice, e.g. would it only happen if the user copy-pastes from the URL bar during the first-run flow?

Good question. @l-hedgehog care to comment here?

l-hedgehog:
@jpetto, @rfk, I'm afraid I don't really have an idea about how often this would happen. What I do know, in the future, I hope the "?dist=mozillaonline" querystring could be persisted when an user navigates between pages in the future.

So maybe I should try to patch the UITour API to provide partner distribution information to bedrock, so that the page can detect the mismatch and redirect accordingly?

Or instead of relying on querystring, I should add something like a custom HTTP header to https://www.mozilla.org requests with our extension? Will it work with cloudflare's CDN?

jgmize:
> Or instead of relying on querystring, I should add something like a custom HTTP header to https://www.mozilla.org requests with our extension? Will it work with cloudflare's CDN?

A custom header cannot be used for this purpose with our current configuration[0]. I believe it would be better to stick with the querystring approach.

[0] It may be technically possible on our enterprise plan, but it would involve opening a support ticket and having Cloudflare engineers add a custom cache key based on that header, and I would only pursue this strategy as a last resort.

alexgibson:
> So maybe I should try to patch the UITour API to provide partner distribution information to bedrock, so that the page can detect the mismatch and redirect accordingly?

If this is possible, exposing the information via the existing `getConfiguration('appinfo') API call would probably make the most sense. This would require a change to Firefox of course, but the patch might be uplifted fairly easily as UITour changes are generally pretty low risk.

If we had an API call like this, then potentially we could avoid needing query parameters entirely, and also mitigate the need for awkward redirects, since each page would be privy to the partner distribution information directly?

l-hedgehog:
@jgmize, I understand, that's what I thought when I asked "Will it work with cloudflare's CDN?"

@alexgibson, guess I really should go and file that UITour bug, maybe later today.

l-hedgehog:
Filed UITour bug as https://bugzil.la/1268369

alexgibson:
Hi @l-hedgehog - we we're just wondering what the status is of this PR, any updates?

l-hedgehog:
Hello @alexgibson, my patch for the UITour bug landed ~10 days ago. Recently I've been working on bugs from my other responsibilities. Hopefully I'll get back to this next week. Thanks for the patience.

l-hedgehog:
Hello @alexgibson, to remove the querystring, I moved the logic to client-side. Do you think the changes in my updated commit is the right way to do it? Thanks.

alexgibson:
Hi @l-hedgehog - just a heads up that in #4136 we're about to land a common shared library for the FxA iframe. This branch will likely need rebasing after it merges, but it should make adding this functionality a bit easier with less duplication across different pages.

alexgibson:
@l-hedgehog - looks like this now needs a rebase after #4136 merged.

l-hedgehog:
Hi @alexgibson, earlier today I rebased my patch, but it's not ready yet, the FxA frame tests are failing as a result of the new async `getConfiguration` call, and the recent updates to django-pipeline broke my local dev environment. (`DEBUG=True` in bedrock/settings/local.py no longer sets `PIPELINE_ENABLED` to `False` etc.)

I'll update this PR once I get them fixed.

alexgibson:
> the recent updates to django-pipeline broke my local dev environment. (DEBUG=True in bedrock/settings/local.py no longer sets PIPELINE_ENABLED to False etc.)

The updated docs in #4160 may be of some help here.

l-hedgehog:
> The updated docs in #4160 may be of some help here.

Thanks! It's helpful.

media/js/base/mozilla-client.js
```diff
@@ -217,6 +217,7 @@ if (typeof Mozilla === 'undefined') {
                 'accurate': accurate,
                 'version': version,
                 'channel': channel,
+                'distribution': distribution,
```
alexgibson: We should also add this value to the `window.dataLayer.push()` property on line 229 below.

media/js/base/mozilla-fxa-iframe.js
```diff
@@ -55,6 +62,15 @@ Mozilla.FxaIframe = (function() {
         // initialize GA event name
         _gaEventName = _config.gaEventName || 'fxa';
 
+        if (_config.distribution && _config.distribution !== 'default') {
+            _distHost = $('#fxa-iframe-config').data(_config.distribution.toLowerCase() + '-host');
```
alexgibson: Nit: the correct way to query data attributes is to use camelCase, so here it would be `.data(mozillaonlineHost)`. That said, this could get a little clunky with mixed casing depending on the `distribution` value. Would it perhaps be cleaner to drop the `-host` part?

alexgibson:
I think we should up this value to 400ms, to give a little more time before falling back to default values. The current 300ms seems a little tight. https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-client.js#L250

bedrock/settings/base.py
```diff
@@ -950,6 +950,11 @@ def lazy_email_backend():
 FXA_IFRAME_SRC = config('FXA_IFRAME_SRC',
                         default='https://accounts.firefox.com/')
 
+# Bug 1264843: embed alternative FxA servers within partner distributions
+FXA_IFRAME_SRC_DIST_MAP = config('FXA_IFRAME_SRC_DIST_MAP',
```
alexgibson: How likely do we think it is we'll get multiple partner distributions? This seems robust, but then I wonder if it's maybe overkill. Happy to defer to what you think here.

alexgibson:
@l-hedgehog - thanks for updating this. I'd like to get this pushed to a demo server so we can verify it works correctly with the China repack. I'll comment here / in the bug once that's done and we can test things out.

media/js/base/mozilla-fxa-iframe.js
l-hedgehog: I think the case of the `distribution` value doesn't matter here, Firefox interprets ` data-MozillaOnline-host="... ` as ` data-mozillaonline-host="... ` and it always should be `.data(mozillaonlineHost)` with `Host` as suffix.

I'm making it more explicit by using ` data-{{ dist|lower }}-host="... ` in templates.

bedrock/settings/base.py
l-hedgehog: I guess no other partner repacks will need this.

alexgibson: If it's unlikely that other partner repacks will need this, then I'd likely favor simplifying the list to a variable in settings? All we'd need is just one data attribute added to the FxA iframe.

l-hedgehog: Sure, PR updated to hard-code just one data attribute for MozillaOnline.

alexgibson:
@l-hedgehog - i pushed this to a demo server for testing. I'm entirely clear on what you'll need to do in order to ensure the signup flow works. I'm guessing there aren't any staging servers available for the FxA server in China?

https://www-demo1.allizom.org/en-US/firefox/accounts/
https://www-demo1.allizom.org/en-US/firefox/46.0/firstrun/

l-hedgehog:
@alexgibson, thank you! No, we don't have a public facing staging server, I'll make sure our production server is configured to work with demo1.

alexgibson:
> @alexgibson, thank you! No, we don't have a public facing staging server, I'll make sure our production server is configured to work with demo1.

Great, thanks @l-hedgehog - for reference we have demo1 through to demo5. Our dev server is www-dev.allizom.org, and staging is www.allizom.org.

alexgibson:
Hi @l-hedgehog - in #4186 we just added a new /firstrun template that uses the FxA iframe. Could you please rebase this PR and incorporate the changes here too?

I'd also like to check in on the status of this PR. Looking at the most recent comment in the bug it looks like your testing on demo server was a success? Is there anything left blocking on your end for merging this PR once we update it? Thanks

l-hedgehog:
Thanks @alexgibson, I've rebased the PR.

I'll have to finish #4067 before switching back to these pages in Fx China repack, but I think this PR can be merged first.

l-hedgehog:
Rebased to resolve the conflict with #4182.

alexgibson:
@l-hedgehog - thanks and sorry for the rebases. I'll try and get to this in the morning for one last look.

alexgibson:
Code wise this looks good, and testing on a demo server seems to have been successful judging by the bug. So this gets an r+ from me. Thanks again @l-hedgehog 👍

l-hedgehog:
Thanks @alexgibson, I'll try to finish #4067 soon.

PR #4067, opened 2016-04-15
 Bug 1264843: link to China build of Fx4A, for display within Fx China repack

l-hedgehog:
### Guidelines
Before you submit a pull request, please make sure you have done the following:

- [x] Review the guidelines for contributing to this repository (link above in GitHub ^).
- [x] All commit messages should be well written, and reference a Bugzilla bug number where applicable.

Note: If you're unsure about anything here or below, don't worry! Just fill out as much
information as you can.

## Description
Please explain the changes you're making, complete with a description of the issue and
a summary of the proposed solution.

Please refer to the bug description.

## Bugzilla
Include a direct link to the associated bug on Bugzilla.

https://bugzil.la/1264843

## Testing
Please leave some directions for the reviewer on how to test your pull request. List which
pages need attention, and what the reviewer should pay close attention to. If the change
requires testing in any specific browsers or platforms, please provide a list.

Just visit an modified page and append ?dist=mozillaonline to its url.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -377,6 +377,14 @@
 </header>{#-- /#fxosnav-header --#}
 {%- endmacro %}
 
+{% macro google_play_button(alt_href='', anchor_attributes={}, class_name='', extra_img_attributes={}, id='') -%}
+{% set optional_img_attributes = {'alt': _('Get it on Google Play'), 'width': '152', 'height': '45', 'l10n': True} %}
+{% do optional_img_attributes.update(extra_img_attributes) %}
+<a{% if class_name %} class="{{ class_name }}"{% endif %}{% if id %} id="{{ id }}"{% endif %} rel="external" href="{{ alt_href|default(settings.GOOGLE_PLAY_FIREFOX_LINK, True) }}"{% if not alt_href %} data-link-type="download" data-download-os="Android"{% endif %}{% for attr, val in anchor_attributes.items() %} {{ attr }}="{{ val }}"{% endfor %}>
+  {{ high_res_img('firefox/android/btn-google-play.png', optional_img_attributes) }}
+</a>
+{%- endmacro %}
+
 {% macro send_to_device(platform, include_title=True, title_text='', include_logo=False, ios_link='', spinner_color='') %}
   {# If no App Store link is specified, use base with default "ct" param value #}
   {# See https://bugzilla.mozilla.org/show_bug.cgi?id=1196310#c23 #}
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -380,7 +380,7 @@
 {% macro google_play_button(alt_href='', anchor_attributes={}, class_name='', extra_img_attributes={}, id='') -%}
 {% set optional_img_attributes = {'alt': _('Get it on Google Play'), 'width': '152', 'height': '45', 'l10n': True} %}
 {% do optional_img_attributes.update(extra_img_attributes) %}
-<a{% if class_name %} class="{{ class_name }}"{% endif %}{% if id %} id="{{ id }}"{% endif %} rel="external" href="{{ alt_href|default(settings.GOOGLE_PLAY_FIREFOX_LINK, True) }}"{% if not alt_href %} data-link-type="download" data-download-os="Android"{% endif %}{% for attr, val in anchor_attributes.items() %} {{ attr }}="{{ val }}"{% endfor %}>
+<a{% if class_name %} class="{{ class_name }}"{% endif %}{% if id %} id="{{ id }}"{% endif %} rel="external" href="{{ alt_href|default(settings.GOOGLE_PLAY_FIREFOX_LINK, True) }}"{% if not alt_href %} data-link-type="download" data-download-os="Android" data-mozillaonline-link="{{ settings.GOOGLE_PLAY_FIREFOX_LINK_MOZILLAONLINE }}"{% endif %}{% for attr, val in anchor_attributes.items() %} {{ attr }}="{{ val }}"{% endfor %}>
   {{ high_res_img('firefox/android/btn-google-play.png', optional_img_attributes) }}
 </a>
 {%- endmacro %}
--- a/bedrock/firefox/templates/firefox/android/index.html
+++ b/bedrock/firefox/templates/firefox/android/index.html
@@ -2,7 +2,7 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
-{% from "macros.html" import fxfamilynav, send_to_device with context %}
+{% from "macros.html" import fxfamilynav, google_play_button, send_to_device with context %}
 
 {% extends "firefox/base-resp.html" %}
 
@@ -34,9 +34,7 @@
 {% block site_header_unwrapped %}
   {% call fxfamilynav('android') %}
     <div class="{% if has_widget %}show-widget{% endif %}">
-      <a class="dl-button" rel="external" href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}" data-link-type="download" data-download-os="Android">
-        {{ high_res_img('firefox/android/btn-google-play.png', {'alt': _('Get it on Google Play'), 'width': '152', 'height': '45', 'l10n': True}) }}
-      </a>
+      {{ google_play_button(class_name='dl-button') }}
       <button id="send-to-button-header" class="send-to button dark" type="button" data-button-name="Get Firefox for Android">{{ _('Get Firefox for Android') }}</button>
     </div>
   {% endcall %}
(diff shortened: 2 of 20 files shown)


bedrock/mozorg/tests/test_helper_misc.py
```diff
@@ -551,6 +557,33 @@ def test_desktop_releasenotes_url(self, mock_reverse):
             'firefox.desktop.releasenotes', args=('42.0', 'release'))
 
 
+@override_settings(GOOGLE_PLAY_FIREFOX_LINK=TEST_GOOGLE_PLAY_FIREFOX_LINK,
+    GOOGLE_PLAY_DIST_MAP=TEST_GOOGLE_PLAY_DIST_MAP)
```
pmclanahan: Should be indented to match `GOOGLE_PLAY_FIREFOX_LINK`.

pmclanahan:
This looks okay to me, but I have some questions about the implementation. I've commented in the bug. Let's figure things out there and come back to this after. Thanks!

l-hedgehog:
Hello @alexgibson & @pmclanahan,

In this bug, I'm only interested in those Firefox for Android download links visible in desktop Firefox, and I didn't touch any other Fx4A links in this updated PR. But this feels like it'll complicate future maintenance, any suggestion on how to improve this? Thanks!

media/js/base/global-init.js
```diff
@@ -11,5 +12,6 @@ $(document).ready(function() {
     initLangSwitcher();
     $(window).on('load', function () {
         $('html').addClass('loaded');
+        Mozilla.Client.getFirefoxDetails(maybeSwitchToDistDownloadLinks);
```
alexgibson: We should add a conditional here to only run this on Firefox for Desktop e.g.

```
if (Mozilla.Client.isFirefoxDesktop) { 
    Mozilla.Client.getFirefoxDetails(maybeSwitchToDistDownloadLinks);
}
```

bedrock/base/templates/macros.html
```diff
@@ -377,6 +377,16 @@
 </header>{#-- /#fxosnav-header --#}
 {%- endmacro %}
 
+{% macro google_play_button(application='firefox', class_name='', extra_attributes=None, id='', include_rel=True, target_blank=False) -%}
```
alexgibson: I think it's safe that we can assume `rel=external` can always be attributed on the Play Store badge link, so I don't think it needs to be a parameter here?

bedrock/base/templates/macros.html
```diff
@@ -377,6 +377,16 @@
...
+{% set optional_attributes={'alt': _('Get it on Google Play'), 'width': '152', 'height': '45', 'l10n': True} %}
+{% if extra_attributes %}
+  {% do optional_attributes.update(extra_attributes) %}
+{% endif %}
+<a{% if class_name %} class="{{ class_name }}"{% endif %}{% if id %} id="{{ id }}"{% endif %}{% if include_rel %} rel="external"{% endif %} href="{% if application == 'stumbler' %}https://play.google.com/store/apps/details?id=org.mozilla.mozstumbler{% else %}{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}{% endif %}"{% if target_blank %} target="_blank"{% endif %}{% if application == 'firefox' %} data-link-type="download" data-download-os="Android" data-mozillaonline-link="{{ settings.GOOGLE_PLAY_FIREFOX_LINK_MOZILLAONLINE }}"{% endif %}>
```
alexgibson: There's a lot of logic here that's a bit difficult to read. Could we maybe pass the `href` as a parameter to the macro, that defaults to Firefox for Android instead, avoiding the long `application` conditional? Thoughts?

tests/unit/spec/base/global.js
```diff
@@ -90,4 +91,38 @@ describe('global.js', function() {
 
     });
 
+    describe('maybeSwitchToDistDownloadLinks', function() {
+
+        var $link,
+            defaultHref = 'https://test.example.com/?id=org.mozilla.firefox',
```
alexgibson: Nit: please use a separate `var` declaration per-line.

media/js/base/global-init.js
```diff
@@ -11,5 +12,8 @@ $(document).ready(function() {
     initLangSwitcher();
     $(window).on('load', function () {
         $('html').addClass('loaded');
+        if (Mozilla.Client.isFirefoxDesktop) {
```
alexgibson: Nit: please add a line of blank space above the conditional. A brief comment to give context as to what this is doing would be nice also.

alexgibson:
A few more suggestions / comments, but otherwise things look good from a code perspective. Once these are made I'll put this up on a demo for some testing before we merge. Thanks @l-hedgehog!

bedrock/base/templates/macros.html
l-hedgehog: Yes, it is a lot shorter. Thanks for the suggestion!

bedrock/base/templates/macros.html
```diff
@@ -377,6 +377,16 @@
 </header>{#-- /#fxosnav-header --#}
 {%- endmacro %}
 
+{% macro google_play_button(alt_href='', class_name='', extra_attributes=None, id='', target_blank=False) -%}
```
alexgibson: Another possible optimization: `target_blank` seems to be used very little, could it just be passed as an `extra_attribute`?

bedrock/firefox/templates/firefox/android/index.html
```diff
@@ -387,9 +383,7 @@ <h2 class="heading">{{ _('Don’t leave home without being in Sync') }}</h2>
         <div id="download-content">
           <h3>{{ _('Choose Firefox') }}</h3>
           <div id="send-to-device-footer">
-            <a class="dl-button" rel="external" href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}" data-link-type="download" data-download-os="Android" data-alt="{{ _('Get Firefox for Android') }}">
-              {{ high_res_img('firefox/android/btn-google-play.png', {'alt': _('Get it on Google Play'), 'width': '152', 'height': '45', 'l10n': True}) }}
-            </a>
+            {{ google_play_button(class_name='dl-button') }}{# data-alt="{{ _('Get Firefox for Android') }}" #}
```
alexgibson: Any reason why `data-alt` isn't being included here?

media/js/base/global-init.js
```diff
@@ -11,5 +12,11 @@ $(document).ready(function() {
     initLangSwitcher();
     $(window).on('load', function () {
         $('html').addClass('loaded');
+
+        /* Bug 1264843: In partner distribution of desktop Firefox, switch the
+           downloads to corresponding partner build of Firefox for Android. */
+        if (Mozilla.Client.isFirefoxDesktop) {
```
alexgibson: It just occurred to me that this might also be fine to call outside of the `load` event, inside the `ready` handler above.

bedrock/base/templates/macros.html
l-hedgehog: `target_blank` is for `a`, while `extra_attributes` is for `img` inside it. Is it okay to mix them up?

alexgibson: Ah good, point - ok let's leave this is is 👍

bedrock/firefox/templates/firefox/android/index.html
l-hedgehog: I wanted to get back to this after finishing the whole PR and figure out what it's for and whether it's still necessary, forgot to do that.

alexgibson: I believe it was for an A/B that may show an alternate string. Seeing as the string exists and is being localized, I think we should probably still retain it here.

bedrock/base/templates/macros.html
```diff
@@ -377,6 +377,16 @@
 </header>{#-- /#fxosnav-header --#}
 {%- endmacro %}
 
+{% macro google_play_button(alt_href='', class_name='', data_alt='', extra_attributes=None, id='', target_blank=False) -%}
```
alexgibson: While this works, `data_alt` is probably something that's a one off name, chosen randomly for a specific page. Instead of explicitly defining it here, I think it may be more robust to allow passing optional attributes on the link element (similar to what we floated previously as an idea). I'm happy to use a different parameter for these link attributes so as to not mix them up with the ones used for `img`. We could handle `target_blank` here in the same way then too, instead of needing to define it here.

l-hedgehog: PR updated, I kept the `class_name` and `id` as separate parameters.

alexgibson:
Thanks @l-hedgehog - testing locally this looks good to me. I've pushed the latest code to demo1 for a little further testing, but I think we're pretty much there. Sorry this took a while: https://www-demo1.allizom.org/en-US/

alexgibson:
r+ ⚡

l-hedgehog:
@alexgibson, thank you for the quick and thorough review!

@pmac, I'll remove all the hacks to in-product urls in China repack, so please expect an ~1% increase in firstrun/whatsnew traffic with the release of Fx 48. :smile:

PR #4068, opened 2016-04-15
fix the rtl css on the sync page

djug:
Currently the Arabic Sync page looks like this:
before

but it should look like this:
fixed


schalkneethling:
Thanks @djug r+ :flags:

PR #4069, opened 2016-04-15
fix the position of the plus sign in the private browsing page

djug:
the plus sign in the RTL pages is not in the right place:
before

and should appear like this:
fixed


schalkneethling:
r+ Thanks @djug :moneybag:

PR #4071, opened 2016-04-18
Fix bug 1258998 - Remove en-* conditional on newsletter title and button

flodolo:
Remove conditionals introduced in #3995 and #4023


schalkneethling:
r+ :clapper:

PR #4073, opened 2016-04-18
Follow up Bug 1261866 - Update links and redirects to mozilla.jp

kyoshino:
Forgot to add Flickr :camera: as part of #4032.


jpetto:
👍

PR #4076, opened 2016-04-20
[fix bug 1261482] Add onboarding page variations.

jpetto:
## Description
3 new onboarding pages to test in various combinations of first and second run.

Should hold off on merging until Optimizely test is built and all stakeholders have given green light.

## Bugzilla
[bug 1261482](https://bugzilla.mozilla.org/show_bug.cgi?id=1261482)

## Testing
Make sure each variation performs/renders as expected, and is displayed for the correct URLs. Testing will only be for en-US, so need for localization as of yet.

Testing URLs:

### Control
https://www-demo5.allizom.org/en-US/firefox/46.0/firstrun/?v=1 - simple/existing FxA template

### Variation 2
https://www-demo5.allizom.org/en-US/firefox/46.0/firstrun/?v=2 - simple/existing FxA template
https://www-demo5.allizom.org/en-US/firefox/46.0/secondrun/?v=2 - mobile download/private browsing template

### Variation 3
https://www-demo5.allizom.org/en-US/firefox/46.0/firstrun/?v=3 - simple/existing FxA template
https://www-demo5.allizom.org/en-US/firefox/46.0/secondrun/?v=3 - pin it/try search template

### Variation 4
https://www-demo5.allizom.org/en-US/firefox/46.0/firstrun/?v=4 - mobile download/private browsing template
https://www-demo5.allizom.org/en-US/firefox/46.0/secondrun/?v=4 - simple/existing FxA template

### Variation 5
https://www-demo5.allizom.org/en-US/firefox/46.0/firstrun/?v=5 - pin it/try search template
https://www-demo5.allizom.org/en-US/firefox/46.0/secondrun/?v=5 - complex/3-step FxA

### Variation 6
https://www-demo5.allizom.org/en-US/firefox/46.0/firstrun/?v=6 - complex/3-step FxA
https://www-demo5.allizom.org/en-US/firefox/46.0/secondrun/?v=6 - pin it/try search template

## Checklist
- [ ] Requires l10n changes.


jpetto:
As we need to do funnelcake testing, we are good to merge this PR after review. If last minute changes come in, we'll address in follow-up PRs.

bedrock/firefox/templates/firefox/onboarding/user-actions-mobileprivacy.html
```diff
@@ -0,0 +1,51 @@
...
+      </ul>
+    </li>
+    <li class="user-action privacy">
+      <h3>{{ _('Browse like no one’s watching') }}</h3>
+
+      <p>
+        {{ _('Use Private Browsing with Tracking Protection to cover your tracks.') }}
+      </p>
+
+      <a href="https://support.mozilla.org/kb/private-browsing-use-firefox-without-history/" class="trailing-arrow" id="try-pb" data-highlight="privateWindow">{{ _('Try it now') }}</a>
```
alexgibson: Could a better destination for this link be the `/firefox/private-browsing` page?

bedrock/firefox/templates/firefox/onboarding/user-actions-pinsearch.html
```diff
@@ -0,0 +1,48 @@
...
+
+      {# Default instructions - non-OS X #}
+      <ul class="instructions pinit-other">
+        <li>{{ _('Right-click the Firefox icon below') }}</li>
+        <li>{{ _('Click the pin to keep it there') }}</li>
+      </ul>
+
+      {# OS X specific instructions #}
+      <ul class="instructions pinit-mac">
+        <li>{{ _('Click and hold the Firefox icon below') }}</li>
```
alexgibson: Hmm... I place my dock to the left of the screen, not below. Could a better string be "Click and hold the Firefox icon in your dock"?

media/css/firefox/onboarding/firstrun-fxa.less
```diff
@@ -3,18 +3,7 @@
...
-    background: #fff;
-    border-radius: 50%;
-    height: @size;
-    left: @left;
-    opacity: 0;
-    position: absolute;
-    top: @top;
-    width: @size;
-}
+@import "onboarding-lib.less";
```
alexgibson: Nit: favor bundling less files over `@import`?

alexgibson:
Give this is likely a short lived page for the majority of users, I don't think it's anything to really worry much about - but I noticed keeping the page open with the little star animations make my laptop fans spin 😢

media/css/firefox/onboarding/onboarding-common.less
```diff
@@ -0,0 +1,318 @@
...
+        }
+    }
+
+    .mozlogo {
+        .at2x('/media/img/sandstone/footer-mozilla-white.png', 93px, 24px);
+        .image-replaced();
+        flex: 0 0 93px;
+        display: inline-block;
+        height: 24px;
+        margin-right: @baseLine;
```
alexgibson: Nit: `@gridGutterWidth`?

media/js/firefox/onboarding/fxa-complex.js
```diff
@@ -0,0 +1,102 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function($) {
```
alexgibson: By no means a blocker for this test, but it feels like we're starting to repeat most of the code in this file for each new page that uses FxA. 

Think it's worth filing a future bug to try and consolidate this in a standard base library that we can re-use?

media/js/firefox/onboarding/user-actions.js
```diff
@@ -0,0 +1,70 @@
...
+            }, function(reason) {
+                reject(reason);
+            });
+        });
+    }
+
+    function openSearchUI(e) {
+        e.preventDefault();
+        Mozilla.UITour.openSearchPanel(function() {
+            Mozilla.UITour.setSearchTerm('What is the weather?');
```
alexgibson: This may well be intentional, but the graphic shows "Today's weather" in the field, as opposed to what gets displayed here?

media/js/firefox/onboarding/user-actions.js
```diff
@@ -0,0 +1,70 @@
...
+            $trySearch.addClass('hidden');
```
alexgibson: I tried to push back against this in the Yahoo Search page a bit (and lost out), but I think this is a bit of a UX anti-pattern. First time I clicked the button I accidentally closed the search menu again. I do know where the menu is, but my first reaction was to try and click the button again.

media/js/firefox/onboarding/user-actions.js
```diff
@@ -0,0 +1,70 @@
...
+                if (config && config.targets) {
+                    resolve(config.targets);
+                } else {
+                    reject('UITour: targets property not found.');
+                }
+            });
+        });
+
+        return new Promise(function(resolve, reject) {
+            Promise.all([availableTargets]).then(function(results) {
```
alexgibson: Seeing as you're only waiting on a single `Promise` to resolve (`availableTargets`), is there any reason for the `Promise.all` here?

media/js/firefox/onboarding/user-actions.js
```diff
@@ -0,0 +1,70 @@
...
+
+    function initSearchUI() {
+        shouldPromoteSearch().then(function() {
+            $trySearch.removeClass('hidden');
+            $trySearch.on('click', openSearchUI);
+        });
+    }
+
+    // only 45+ should hit this page, but safer to double check
+    if (client.isFirefoxDesktop && client.FirefoxMajorVersion >= 43) {
```
alexgibson: I'm not sure if this check for `43` is necessary? We needed it on the Yahoo page since we also had `getConfiguration('search')`, but that's not being used here it seems. `getConfiguration('availableTargets')` should work for any version of Firefox that supports UITour. If you need some kind of simple check that UITour is working, a [ping](https://github.com/mozilla/bedrock/blob/master/media/js/base/uitour-lib.js#L245) callback might be enough, although I think the check for `isFirefoxDesktop` should be kept.

alexgibson:
Is the double headline expected?

double-headline

alexgibson:
iframe breaks out the parent container a little at small screen sizes (not a blocker, since this is a desktop test really).

mobile-iframe

alexgibson:
Looking really good, a couple of things to tidy up bot nothing major 👍

bedrock/firefox/templates/firefox/onboarding/user-actions-mobileprivacy.html
jpetto: I'd say so, yeah. Will update.

bedrock/firefox/templates/firefox/onboarding/user-actions-pinsearch.html
jpetto: Hmm, definitely a valid point. I will ping folks in the copy bug to see what they say.

media/css/firefox/onboarding/firstrun-fxa.less
jpetto: Well, this mini lib acts the same as our sandstone lib (only containing variables and mixins), which is why I followed the `@import` pattern. If you feel strongly, I'll move it to the bundle.

bedrock/firefox/templates/firefox/onboarding/user-actions-pinsearch.html
alexgibson: Hmm, maybe I didn't check closely enough, but would it not create duplicate `twinkle` animation rules?

media/css/firefox/onboarding/firstrun-fxa.less
alexgibson: Hmm, maybe I didn't check closely enough, but would it not create duplicate twinkle animation rules?

media/css/firefox/onboarding/onboarding-common.less
jpetto: Ah, yep.

media/js/firefox/onboarding/fxa-complex.js
jpetto: Yes indeed. I put such a task in our last sprint planning spreadsheet (and brought it up during the planning session), but did not yet file a bug. Will do that now.

media/css/firefox/onboarding/firstrun-fxa.less
alexgibson: Ah it's ok, on close inspection it does not seem to so perhaps this is indeed fine. Just seemed like a deviation from convention since it does generate some actual CSS.

media/js/firefox/onboarding/user-actions.js
jpetto: AFAIK, this is intentional to provide readable text in the graphic (as "What is the weather?" would need a really small font).

media/js/firefox/onboarding/user-actions.js
jpetto: Yep, I brought up this issue during dev and the decision was to keep the same UX as on the Yahoo page. 😕

media/js/firefox/onboarding/user-actions.js
jpetto: Was I being lazy? Yes. Will refactor.

media/js/firefox/onboarding/user-actions.js
jpetto: I did noodle over this during dev. Will give it a try with a ping instead.

jpetto:
> Is the double headline expected?

Unfortunately, it is. We can't suppress the headline in the iframe, and creative wanted consistent headlines across all three steps, so, yeah.

jpetto:
> Give this is likely a short lived page for the majority of users, I don't think it's anything to really worry much about - but I noticed keeping the page open with the little star animations makes my laptop fans spin (not a lot, but enough to notice)

Hmm, yeah, but I could see a user leaving the tab open. I'm going to limit the number of twinkles to like 14 (at ~3 seconds per), which should provide ample wonderment. 😀

alexgibson:
> Hmm, yeah, but I could see a user leaving the tab open. I'm going to limit the number of twinkles to like 14 (at ~3 seconds per), which should provide ample wonderment.

If it's any use, while experimenting on the stars for /firefox/new/, I found `scale()` to be a CPU hog. I think both opacity and translate should go to the GPU. Not sure if this would make a difference here or not.

alexgibson:
Just noticed the equals sign is also a little off center compared to the `+`

equals

jpetto:
> If it's any use, while experimenting on the stars for /firefox/new/, I found scale() to be a CPU hog. I think both opacity and translate should go to the GPU. Not sure if this would make a difference here or not.

Hmm, that is interesting. I thought all `transform`s were sent to the GPU. Anyway, I'd rather keep the scaling for a limited number of twinkles.

jpetto:
> Just noticed the equals sign is also a little off center compared to the +

D'oh! Forgot to carry the 20px padding. Fixing.

alexgibson:
> Hmm, that is interesting. I thought all transforms were sent to the GPU. Anyway, I'd rather keep the scaling for a limited number of twinkles.

AFAIK, only a limited number of properties get acceleration for free, but it depends largely on the browser too.

media/css/firefox/onboarding/firstrun-fxa.less
jpetto: You are totally right, `@keyframes` does generate CSS. My brain pattern matched it as a variable due to the `@`. I'm going to move `@keyframes twinkle` into `onboarding-common.less` and `firstrun-fxa.less` to keep with convention. This does result in some non-DRYness, but I think that's okay for now. When a winner/winners are chosen from this test, we'll refactor.

alexgibson:
Changes look good to me - the only thing I would say would be good to address still is the line of copy relating to the OSX dock location, but otherwise this is an r+wc from me 🍨

jpetto:
Woo! I've NI'd Matej regarding the copy. I'll squash by EOD.

Thanks!

jpetto:
Just received word/confirmation that the initial tests will only be run for Windows users, so that line of OS X specific copy shouldn't block us from merging.

Will leave the NI open on the copy bug for if/when we roll out that "pin it" page to a wider audience.

PR #4077, opened 2016-04-20
Bug 1266185: Add the ability to enable basic auth

pmclanahan:
Setting BASIC_AUTH_CREDS environment variable will now protect the entire site with that username and password. Value of the variable must be colon separated username and password (e.g. "theuser:thepass").

### DIFF
--- a/bedrock/base/middleware.py
+++ b/bedrock/base/middleware.py
@@ -4,12 +4,12 @@ Taken from zamboni.amo.middleware.
 This is django-localeurl, but with mozilla style capital letters in
 the locale codes.
 """
-
+import base64
 import urllib
 from warnings import warn
 
 from django.conf import settings
-from django.http import HttpResponsePermanentRedirect
+from django.http import HttpResponsePermanentRedirect, HttpResponse
 from django.utils.encoding import smart_str, force_text
 
 import tower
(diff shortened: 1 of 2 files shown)


bedrock/base/middleware.py
```diff
@@ -75,3 +75,28 @@ def process_request(self, request):
...
+            if 'HTTP_AUTHORIZATION' in request.META:
+                auth = request.META['HTTP_AUTHORIZATION'].split()
+                if len(auth) == 2:
+                    if auth[0].lower() == "basic":
+                        provided_auth = base64.b64decode(auth[1])
+                        if provided_auth == required_auth:
+                            # we're good. continue on.
+                            return None
+
+        response = HttpResponse(status=401,
```
jgmize: Shouldn't this be in an `else` block?

pmclanahan: No, but that's a good catch. It should be indented to be inside the topmost if block. It would be fine because of how we're handling the middleware enablement, but it's better to be correct here for sure.

pmclanahan:
@jgmize fixed and amended.

jgmize:
@pmclanahan I created a demo__jgmize branch off of this PR and then rebased off master, ran `deis config:set BASIC_AUTH_CREDS=wat:wat -a bedrock-demo-jgmize`, then pushed to trigger a deploy to https://bedrock-demo-jgmize.us-west.moz.works and everything appears to be working as expected. If you just rebase this off master that should fix the broken circleci test run, and then once that's passing feel free to merge. r+wc, nice work.

PR #4084, opened 2016-04-22
[fix bug 1261017] Issues with Mozilla.Modal in Firefox 47 at small viewports

alexgibson:
## Description
- Removes the `noscroll` option that is currently used on smaller viewports from `Mozilla.Modal`, as it causes issues in newer versions of Firefox (see bug link below for STR).

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1261017

## Testing
If memory serves me correctly, I think the `noscroll` option was added as a hack for early versions of Firefox OS due to some viewport bug. But honestly, I'm not quite sure. I think most browsers we care about today should be up to handling `position: fixed;`.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/css/base/mozilla-modal.less
+++ b/media/css/base/mozilla-modal.less
@@ -4,10 +4,6 @@
 
 @import "../sandstone/lib.less";
 
-html.noscroll body {
-    overflow: hidden;
-}
-
 #modal {
     background: #000;
     background: rgba(0, 0, 0, 0.85);
@@ -131,18 +127,6 @@ html[dir="rtl"] {
 }
 
 @media only screen and (max-width: @breakTablet) {
-    html.noscroll {
-        overflow: hidden;
-        height: 100%;
-        body {
-            height: 100%;
-            overflow: hidden;
-        }
-        #modal {
-            position: absolute;
-        }
-    }
-
     #modal .window {
         margin: 0 auto;
 
--- a/media/css/firefox/ios.less
+++ b/media/css/firefox/ios.less
@@ -746,6 +746,19 @@ a.go {
     display: block;
 }
 
+// used for Firefox on iOS Sync instructions overlay
+html.noscroll {
+    overflow: hidden;
+    height: 100%;
+    body {
+        height: 100%;
+        overflow: hidden;
+    }
+    #modal {
+        position: absolute;
+    }
+}
+
 
 
 // Footer
--- a/media/js/base/mozilla-modal.js
+++ b/media/js/base/mozilla-modal.js
@@ -8,7 +8,6 @@ Mozilla.Modal = (function(w, $) {
 
     var open = false;
     var $body = $('body');
-    var $html = $('html');
     var options = {};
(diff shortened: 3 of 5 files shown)


alexgibson:
@jpetto - r? 

Some kind of sanity check here would be appreciated to make sure I'm not removing something we still ultimately need :/

jpetto:
Hmm, I think one reason for `noscroll` was to avoid competing scrolls on mobile. When a modal is taller than a device's screen, scrolling can get funky if both the background page and the modal are scrollable. This isn't a show stopping blocker, but it's not a great user experience, either. 

In dev edition, if you open the modal and scroll up or down at all, the modal displays correctly. The behavior feels a little bit like a bug...

alexgibson:
> When a modal is taller than a device's screen, scrolling can get funky if both the background page and the modal are scrollable. This isn't a show stopping blocker, but it's not a great user experience, either. 

Can you point to any pages where this is an issue?

> In dev edition, if you open the modal and scroll up or down at all, the modal displays correctly. The behavior feels a little bit like a bug...

Not quite sure I understand, are you saying this is working correctly, or not quite?

jpetto:
I specifically tested on `/firefox/developer/` and `/contribute` using an iPhone 6, but I'd wager most modals will scroll at least a little with a phone in landscape mode. 

In terms of dev edition's current behavior, it does feel like a rendering bug, yeah. Just wondering if the fix should be in product rather than with the modal script.

I wouldn't exactly say the likely rare experience of competing scrollbars should stop the fix, but I'm also not sure this isn't a bug in dev edition.

alexgibson:
> I specifically tested on /firefox/developer/ and /contribute using an iPhone 6, but I'd wager most modals will scroll at least a little with a phone in landscape mode. 

The behavior doesn't seem too bad to me, everything is still accessible and works ok, but like you say the background can still scroll. To me this seems like an acceptable consequence to the content being inaccessible to people using Firefox on smaller screens. 

What I'm conscious of is this could also impact Firefox for Android. We used to have a bug for just this issue but I can't find it (filed by sgarity?). Perhaps it is a new bug in Firefox for desktop, or perhaps it is now intended behavior?

I'll file a bug for this and add a do-not-merge label for now, but given the bug makes content inaccessible I think we should still try and fix it.

jpetto:
Agree with that assessment. Should definitely be fixed one way or another.

In the event we do merge this PR, looks like there's still an `allowScroll` to be cleaned up in [`/firefox/os/devices.js`](https://github.com/mozilla/bedrock/blob/master/media/js/firefox/os/devices.js#L124).

alexgibson:
Filed a Firefox bug to see what they say: https://bugzilla.mozilla.org/show_bug.cgi?id=1266833

alexgibson:
I'm going to tentatively close this - looking in the bug it seems like it may actually be an e10s related issue, so will hopefully get prioritized to be fixed. I'll keep this branch around if/when we may need it.

PR #4086, opened 2016-04-25
Switch to using Gulp as a task runner and update dev dependencies

alexgibson:
- Switches to using [Gulp](http://gulpjs.com/) for a task runner instead of Grunt.
- Removes some dependency cruft we don't really use.
- Updates dev dependencies to latest versions of Karma/Jasmine/ESLint.


node_modules/.gitignore
```diff
@@ -5,7 +5,6 @@
 !.gitignore
 
 # less for less -> css
-!.bin/
```
alexgibson: @pmclanahan - this seems to create un-tracked changes in `node_modules/.bin/` when I do a fresh `npm install`. The line below should be enough to cover the `node_modules/.bin/lockdown` files that are already checked into the repo, unless I'm missing something?

alexgibson:
Well, CircleCI seems happy and this installs ok for me locally 👍

pmclanahan:
This looks good to me. Grunt isn't currently integrated to the build and deploy process, so as long as the prod dependency installation continues to work then I'm happy :)

:+1:

schalkneethling:
Just a question, any reason we cannot switch the npm scripts?

alexgibson:
> Just a question, any reason we cannot switch the npm scripts?

I thought about this, but we're currently on node v0.12.0 by the look if it, so the version of npm is less flexible for this type of thing. We may also soon be doing more front-end asset type things so gulp may come in useful.

PR #4089, opened 2016-04-25
Add hashes for zip requirements

jgmize:
We are seeing [failures](https://ci.us-west.moz.works/job/bedrock_base_image/1298/console) installing python requirements when building the base image due to hash mismatches. The reason for this is that for some requirements we have the tarball hashes but not the zipfiles.


jgmize:
@pmclanahan r?

PR #4090, opened 2016-04-25
Fix Bug 1245567 - Remove Firefox for Android 2.3 Gingerbread from www.mozilla.org

kyoshino:
This is due this Friday, April 29.

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -309,9 +309,16 @@ class FirefoxAndroid(_ProductDetails):
 
     def platforms(self, channel='release'):
         platforms = self.platform_labels.copy()
+        major_version = int(self.latest_version(channel).split('.', 1)[0])
+
+        # Android Gingerbread (2.3) is no longer supported as of Firefox 48
+        if major_version >= 48:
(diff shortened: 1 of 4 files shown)


kyoshino:
This has to be merged today, or there will be 404 links tomorrow. @pmclanahan @jgmize r?

bedrock/firefox/firefox_details.py
```diff
@@ -310,8 +310,14 @@ class FirefoxAndroid(_ProductDetails):
     def platforms(self, channel='release'):
         platforms = self.platform_labels.copy()
 
+        # Android Gingerbread (2.3) is no longer supported as of Firefox 48
+        if int(self.latest_version(channel).split('.', 1)[0]) >= 48:
```
pmclanahan: Looks like you're doing this split twice. Would be better to set a variable like:

```python
major_version = int(self.latest_version(channel).split('.', 1)[0])
```

pmclanahan:
Apart from a minor thing I commented on this looks good. r+wc.

bedrock/firefox/firefox_details.py
kyoshino: Ah thanks! Fixed.

PR #4091, opened 2016-04-25
Revert "Add supervisor to Procfile for bug 1266665"

jgmize:
Reverts mozilla/bedrock#4088


pmclanahan:
Doh! I knew there was a reason I'd not done this, but couldn't think of it. Thanks!

Guess we have to modify `run_prod.sh` to optionally run supervisor if a certain env var is set?

PR #4093, opened 2016-04-26
[fix bug 1261482] Update pin it copy for OS X.

jpetto:
## Description
Simple copy change proposed by @alexgibson in #4076 and approved by Matej in [the copy bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1259543#c3).

## Bugzilla
[bug 1261482](https://bugzilla.mozilla.org/show_bug.cgi?id=1259543#c3)


alexgibson:
r+

PR #4094, opened 2016-04-27
[fix bug 1267561] Roll out winning version of redesigned download page

alexgibson:
## Description
- This is still WIP, will update here when ready for code review once al page states have been approved.
- Rolls out the winner of the download page redesign tests from [Bug 1256241](https://bugzilla.mozilla.org/show_bug.cgi?id=1256241)
- New page will be en-US only for now, until we do a round of copy tests in the next sprint. After that is finished, we will update final strings and begin l10n to move locales over to the new page.
- This change also adds back the newsletter form, which is shown for up-to-date Firefox users on scene 1.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1267561

## Testing
- Branch is up on a demo here: https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/
- This has already been tested a fair bit when we did the initial test variations, but some additional checks (particularly with Firefox up-to-date and out-of-date page states) would still be worthwhile here since some new code has been added since.
- I ran the functional tests for this page in both Firefox and Chrome and they pass.

## Checklist
- [ ] Requires l10n changes (none for the time being)


alexgibson:
Ok, going to open this up for review as i believe everything is now approved.

media/css/firefox/new/redesign/scene1.less
```diff
@@ -377,3 +385,303 @@ html.fxos {
...
+                width: auto;
+                margin-bottom: @baseLine * 2;
+            }
+        }
+    }
+
+    h2 {
+        position: relative;
+        padding-top: 73px;
+        font-weight: bold;
```
jpetto: If we swapped in `.open-sans()` here, it saves us 112KB in an extra font load. The UI difference is barely noticeable (and looks better, IMO).

jpetto:
For out-of-date Firefox, the "Update to stay fast and safe" text isn't really recognizable as a link...which might be for the best, as the user probably just wants to download Firefox and not go to

jpetto:
One performance issue we should probably address, but looking great otherwise. r+wc

🌞

media/css/firefox/new/redesign/scene1.less
alexgibson: Nice catch 👍

alexgibson:
> For out-of-date Firefox, the "Update to stay fast and safe" text isn't really recognizable as a link...which might be for the best, as the user probably just wants to download Firefox and not go to SUMO.

I think there is probably some benefit in the user going to SUMO, since the article encourages the user to update manually instead of download a fresh copy entirely. I'll style the link using `@textColorSpaceYellow` to give it some visual highlight.

alexgibson:
@jpetto - thanks for the review, updated.

jpetto:
Looks like a winner to me! r+ 🔊

PR #4098, opened 2016-04-28
Remove stray files from old Vagrant and Cloud9 config

alexgibson:
@pmclanahan r? 

I noticed these files and figured they are probably no longer needed in the repo?


pmclanahan:
I filed a similar thing to remove all the files for the legacy deployment (#4005). The problem is that we're still using the old demos. So I'm pretty sure those puppet files can go, but not 100%.

alexgibson:
Ah, no worries. Let me close this and you can deal wit it when we need in #4005. Thanks

PR #4101, opened 2016-04-29
Bug1264966 finalize new contribute signup and tasks

schalkneethling:
## Description
This implements the final version of the new contribute/signup page and related tasks. NOTE: This does not include cleaning up all the pages and related files that will no become obsolete. It is probably best to do this as a separate PR.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1264966

## Testing
These pages all exist in some form on production already so there has been review done but, there are some significant changes to the JavaScript that warrants some review. 

## Checklist
- [x] Requires l10n changes.


flodolo:
See email I've just sent, I think we need to block this on l10n, otherwise we'll end up with a bunch of pages in mixed localized text and English (especially considering the deployment issue we have on prod).

Are strings final?

schalkneethling:
@flodolo Agreed on blocking here. The strings are final, yes.

flodolo:
One possible solution to avoid blocking anything: we need to use a different .lang file for the /signup page (`mozorg/contribute/signup`). We also need to update all the tasks templates to explicitly use that lang file with a add_lang_file directive.

This way the existing get involved main page remains online, and we can enable /signup as soon as it's ready. Everything needs to be tested thoroughly with production settings.

bedrock/mozorg/templates/mozorg/contribute/tasks/joy-of-coding.html
```diff
@@ -6,26 +6,28 @@
...
+        <h3>{{ _('Watch the intro.') }}</h3>
         <p>{{ _('<a href="https://twitter.com/mike_conley" rel="external">@mike_conley</a> livehacks on real Firefox bugs while thinking out loud.') }}</p>
       </header>
       <div class="step-content">
-        <video id="joc" preload="none" controls poster="{{ static('img/contribute/signup/joc_poster.png') }}" data-task="joyofcoding" data-step="one" data-complete="true" data-variation="1">
+        <video id="joc" preload="none" controls poster="{{ static('img/contribute/signup/joc_poster.png') }}" data-action="play" data-task="joyofcoding" data-step="one" data-complete="true">
           <source src="https://videos.cdn.mozilla.net/uploads/Videos%20creative/joc-video.mp4" type="video/mp4" />
         </video>
-        <button type="button" class="button-blue watch" data-action="play" data-task="joyofcoding" data-step="one" data-complete="true" data-variation="1">Watch Video</button>
+        <button type="button" class="button-blue watch" data-action="play" data-task="joyofcoding" data-step="one" data-complete="true">Watch Video</button>
```
flodolo: String is not wrapped. If we could use `Watch the video` it would be great, since we already have this string.

bedrock/mozorg/templates/mozorg/contribute/tasks/devtools-challenger.html
```diff
@@ -7,27 +7,29 @@
...
 
 {% block taskview_title %}{{ _('Try out our new animation editing tools in Developer Tools Challenger') }}{% endblock %}
-{% block taskview_lead %}{{ _('The more you know about the web, the easier it is to make it your own! Participate in the web by testing your web literacy skills using the new tools in Firefox Developer Edition.') }}{% endblock %}
+{% block taskview_lead %}{{ _('The more you know about the web, the easier it is to make it your own. Participate in the web by testing your web literacy skills using the new tools in Firefox Developer Edition.') }}{% endblock %}
 
 {% block task_content %}
   <section class="get-firefox">
     <p>{{ _('Dev Tools requires Firefox Developer Edition desktop browser. To try this task, download Firefox Developer Edition first.') }}</p>
-    <a href="{{ url('firefox.developer') }}" class="button-blue dev-edition" data-variation="1">Download Firefox Dev Edition</a>
+    <a href="{{ url('firefox.developer') }}" class="button-blue dev-edition">Download Firefox Dev Edition</a>
```
flodolo: String is not wrapped.

bedrock/mozorg/templates/mozorg/contribute/tasks/stumbler.html
```diff
@@ -0,0 +1,31 @@
...
+  <ol class="task-steps">
+    <li>
+      <header id="step_one" class="one">
+        <h3>{{ _('Install Mozilla Stumbler') }}</h3>
+        <p>{{ _('<a href="%s" target="_blank">Learn more about Mozilla Location Services</a> and discover more ways to get involved.')|format('https://location.services.mozilla.com/') }}</p>
+      </header>
+      <div class="step-content">
+        {{ high_res_img('contribute/signup/stumbler.png', {'alt': '', 'width': '620', 'class': 'feature-img'}) }}
+        <a rel="external" class="stumbler-button" href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}" target="_blank" data-action="install" data-task="stumbler" data-step="one" data-complete="true">
+          {{ high_res_img('contribute/signup/google-play.png', {'alt': _('Get it on Google Play'), 'width': '200', 'height': '59', 'data-action': 'install', 'data-task': 'stumbler', 'data-step': 'one', 'data-complete': 'true'}) }}
```
flodolo: Image in this page is not localized, while it's correctly localized in task/firefox-mobile/

flodolo:
Let's remove [this line](https://github.com/mozilla/bedrock/pull/4101.patch), so that /signup will use its own langfile.

Then you need to add a directive to the all the templates in /tasks
```
{% add_lang_files "mozorg/contribute/signup" %}
```

I did a quick test in my local VM, but it would be great if you could confirm: set `DEV = False`, /signup redirects to en-US if the locale's file in mozorg/contribute/signup.lang is not activated. 

I don't know why but individual task pages don't redirect and open the locale's page. Not sure if @pmclanahan has any ideas.

I also don't see the `Find your local Mozilla community.Find your local Mozilla community.`, I always see the newsletter block. Example: https://bedrock-demo-schalk-contribute.us-west.moz.works/it/contribute/task/follow-mozilla/

alexgibson:
We we're also skipping a bunch of functional tests while this test was running. If the old pages are being removed, then can the tests be removed too?

https://github.com/mozilla/bedrock/blob/master/tests/functional/contribute/test_signup.py

We don't need to add tests for the new page straight away, but we should file a follow up bug to create some post-launch.

schalkneethling:
@flodolo was unclear from the link in your previous comment but, I assume this is the line to remove from /signup correct?

```
{% add_lang_files "mozorg/contribute/index" %}
```

schalkneethling:
@flodolo The conditional l10n copy is below the newsletter block.

schalkneethling:
@alexgibson It definitely sounds to me like this is going to replace the current page at /signup so, I would say it is a good idea to remove that file you mentioned and file a bug to add tests for the new /signup page.

flodolo:
Changes look good from a l10n point of view, did you manage to test them with a prod setting to verify my findings?

As for the conditional content after newsletter: I see the code, but how do I see it in the page?

schalkneethling:
@flodolo You mean you do not see this?

screen shot 2016-05-03 at 20 12 29

flodolo:
Oh wow, I was expecting something else :-) Now I see it.

bedrock/mozorg/templates/mozorg/contribute/tasks/follow-mozilla.html
```diff
@@ -0,0 +1,45 @@
...
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "mozorg/contribute/tasks/taskview-base.html" %}
+
+{% add_lang_files "mozorg/contribute/signup" %}
+
+{% block page_title %}{{ _('Tweet to @StartMozilla and see how your technical skills can contribute') }}{% endblock %}
+
+{% block more_tasks %}{{ url('mozorg.contribute.signup') }}?v=2{% endblock %}
```
jpetto: Don't think we need this customized URL any longer.

bedrock/mozorg/templates/mozorg/contribute/tasks/follow-mozilla.html
```diff
@@ -0,0 +1,45 @@
...
+      </div>
+    </li>
+  </ol>
+{% endblock %}
+
+{% block thank_you_msg %}
+  {{ _('Thanks for following') }}
+{% endblock %}
+
+{% block try_another %}
```
jpetto: Don't think we need this block here anymore either.

bedrock/mozorg/templates/mozorg/contribute/tasks/joy-of-coding.html
```diff
@@ -4,28 +4,32 @@
...
       </header>
       <div class="step-content">
-        <video id="joc" preload="none" controls poster="{{ static('img/contribute/signup/joc_poster.png') }}" data-task="joyofcoding" data-step="one" data-complete="true" data-variation="1">
+        <video id="joc" preload="none" controls poster="{{ static('img/contribute/signup/joc_poster.png') }}" data-action="play" data-task="joyofcoding" data-step="one" data-complete="true">
           <source src="https://videos.cdn.mozilla.net/uploads/Videos%20creative/joc-video.mp4" type="video/mp4" />
         </video>
-        <button type="button" class="button-blue watch" data-action="play" data-task="joyofcoding" data-step="one" data-complete="true" data-variation="1">Watch Video</button>
+        <button type="button" class="button-blue watch" data-action="play" data-task="joyofcoding" data-step="one" data-complete="true">{{ _('Watch the video') }}</button>
+        <h4>{{ _('Get involved') }}</h4>
+        {{ _('<a href="%s" id="get-involved" target="_blank">Join the Coding Community</a>  and start collaborating with other Mozillians around the world.')|format('https://wiki.mozilla.org/Contribute/Coding') }}
```
jpetto: Nit - extra space after `</a>`

bedrock/mozorg/templates/mozorg/contribute/tasks/stumbler.html
```diff
@@ -0,0 +1,33 @@
...
+  <ol class="task-steps">
+    <li>
+      <header id="step_one" class="one">
+        <h3>{{ _('Install Mozilla Stumbler') }}</h3>
+        <p>{{ _('<a href="%s" target="_blank">Learn more about Mozilla Location Services</a> and discover more ways to get involved.')|format('https://location.services.mozilla.com/') }}</p>
+      </header>
+      <div class="step-content">
+        {{ high_res_img('contribute/signup/stumbler.png', {'alt': '', 'width': '620', 'class': 'feature-img'}) }}
+        <a rel="external" class="stumbler-button" href="{{ settings.GOOGLE_PLAY_FIREFOX_LINK }}" target="_blank" data-action="install" data-task="stumbler" data-step="one" data-complete="true">
+          {{ high_res_img('firefox/android/btn-google-play.png', {'alt': _('Get it on Google Play'), 'width': '152', 'height': '45', 'l10n': True, 'data-action': 'install', 'data-task': 'stumbler', 'data-step': 'one', 'data-complete': 'true'}) }}
```
jpetto: Do we need all the attributes on both the `<a>` and the ``?

bedrock/mozorg/templates/mozorg/contribute/tasks/taskview-base.html
```diff
@@ -15,7 +15,7 @@
 
 {% block alt_header %}
   <h1>{{ _('Mozilla Participation Tasks') }}</h1>
-  <a href="{% block more_tasks %}{{ url('mozorg.contribute.signup') }}?v=1{% endblock %}" class="try-another back" {% block pick_task_variation %}data-variation="1"{% endblock %}>{{ _('I want to pick a different task.' )}}</a>
+  <a href="{% block more_tasks %}{{ url('mozorg.contribute.signup') }}{% endblock %}" class="try-another back">{{ _('I want to pick a different task.' )}}</a>
```
jpetto: Don't think we need this `more_tasks` block any longer.

bedrock/mozorg/templates/mozorg/contribute/tasks/taskview-base.html
```diff
@@ -34,14 +34,28 @@ <h2 class="page-title">{% block taskview_title %}{% endblock %}</h2>
       <header>
         <h3>{% block thank_you_msg %}{% endblock %}</h3>
       </header>
-      <p><a href="{% block try_another %}{{ url('mozorg.contribute.signup') }}?v=1{% endblock %}" class="try-another" {% block task_variation %}data-variation="1"{% endblock %}>{{ _('Try another task?') }}</a></p>
+      <p><a href="{% block try_another %}{{ url('mozorg.contribute.signup') }}{% endblock %}" class="try-another">{{ _('Try another task?') }}</a></p>
```
jpetto: Looks like we can drop this `try_another` block as well.

jpetto:
Looks like `img/contribute/signup/whimsy.png` is no longer used & should be deleted.

media/css/mozorg/contribute/newsletter.less
```diff
@@ -0,0 +1,48 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../../sandstone/lib.less";
+
+.newsletter-module {
+    background: #fff;
+    color: #333;
```
jpetto: Use `@textColorPrimary` instead.

media/css/mozorg/contribute/taskview.less
```diff
@@ -204,36 +196,49 @@ main {
...
-    }
-
     .one.completed:before,
     .two.completed:before {
-        background: transparent url('/media/img/contribute/signup/complete-icon.svg') -30px -27px no-repeat;
-        background-size: 110px 110px;
+        position: absolute;
+        left: -65px;
+        background: transparent url('/media/img/contribute/signup/complete-icon.svg') -30px -33px no-repeat;
+        background-size: 90px 90px;
```
jpetto: Use `.background-size` mixin.

media/css/mozorg/contribute/taskview.less
```diff
@@ -204,36 +196,49 @@ main {
...
-        height: 63px;
-        .font-size(48px);
-        line-height: 1;
-        vertical-align: bottom;
+        border-radius: 0;
+        width: 90px;
+        height: 90px;
+
+        @media only screen and (max-width: @breakTablet) {
+            background-size: 70px 70px;
```
jpetto: Use `.background-size` mixin here and below.

media/css/mozorg/contribute/taskview.less
```diff
@@ -435,3 +403,64 @@ a[role="button"] {
 .hidden {
     display: none;
 }
+
+/* newsletter module styles */
+.newsletter-module {
```
jpetto: Could include `contribute/newsletter.less` in the bundle instead of repeating code here?

jpetto:
Looks like images can be run through ImageAlpha (again?) to save between 1 and 5%.

media/js/mozorg/contribute/taskview.js
```diff
@@ -108,92 +100,58 @@
...
+
+        if (taskAction === 'tweet') {
+            $tweetTxt = $('#tweet_txt').text();
+            intentURL += 'tweet?text=' + $tweetTxt + '&hashtags=QA1';
+            completeTwitterAction(intentURL, $this, 'tweeted to @startmozilla');
+        }
+
+        if (taskAction === 'follow') {
+            intentURL += 'follow/?screen_name=' + screenName;
+            completeTwitterAction(intentURL, $this, 'follow ' + screenName);
```
jpetto: Could refactor to be a bit more DRY.

media/css/mozorg/contribute/signup.less
```diff
@@ -102,87 +102,77 @@ main {
             height: 185px;
 
             @media only screen and (max-width: @breakDesktop) {
-                background-size: 400px 400px;
+                background-position: -15px -10px;
+                background-size: 446px 300px;
```
jpetto: Use `.background-size` mixin here and below.

bedrock/mozorg/templates/mozorg/contribute/tasks/stumbler.html
schalkneethling: Technically, no. However, I wanted to make doubly sure that should the event target turn out to be the `img` the attributes will still be available. I would think that it is safe so, I am going to remove it from the `img` tag.

media/css/mozorg/contribute/taskview.less
schalkneethling: I did try that initially, also tried using `newsletter-test.less` but, here we need a combination of the two in a way so, I found it simplest to code it up here. I reckon when we have moved to a single new newsletter style, this will become redundant.

bedrock/mozorg/templates/mozorg/contribute/tasks/follow-mozilla.html
```diff
@@ -0,0 +1,41 @@
...
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "mozorg/contribute/tasks/taskview-base.html" %}
+
+{% add_lang_files "mozorg/contribute/signup" %}
+
+{% block page_title %}{{ _('Tweet to @StartMozilla and see how your technical skills can contribute') }}{% endblock %}
+
+{% block more_tasks %}{{ url('mozorg.contribute.signup') }}{% endblock %}
```
jpetto: Do we still need this block? Isn't this the same URL specified in `taskview-base.html`?

bedrock/mozorg/templates/mozorg/contribute/tasks/taskview-base.html
jpetto: Just checking again here - do we still need this `more_tasks` block?

bedrock/mozorg/templates/mozorg/contribute/tasks/taskview-base.html
```diff
@@ -34,14 +34,28 @@ <h2 class="page-title">{% block taskview_title %}{% endblock %}</h2>
...
+
+{% block email_form_unwrapped %}
+  <section class="newsletter-module">
+    <div class="container">
+      {{ email_newsletter_form(newsletters='mozilla-foundation', title=_('Keep in touch with Mozilla. Sign up for our newsletter.'), button_class='dark') }}
+    </div>
+  </section>
+
+  <ul class="notices">
+    <li><a href="https://www.mozilla.org/contact/communities/" id="communities" target="_blank">{{ _('Find your local Mozilla community.') }}</a></li>
```
jpetto: Use the URL helper here instead?

media/css/mozorg/contribute/taskview.less
jpetto: Hmm, I'm a little confused. I don't see `contribute/newsletter.less` referenced anywhere, but it's being created in this PR. Is it for potential future use? Should we be adding it now if it isn't currently serving any purpose?

media/css/mozorg/contribute/taskview.less
```diff
@@ -435,3 +426,64 @@ a[role="button"] {
...
+    background: #fff;
+    color: #333;
```
jpetto: Use `@textColorPrimary` instead?

jpetto:
In `mozorg/views.py`, do we still need `class ContributeSignup` or `class ContributeSignupOldForm`? I no longer see those forms referenced, as it appears with this PR we will be serving only this new version of `/contribute/signup`.

jpetto:
Two interaction issues:

1. On the Joy of Coding task, when the video reaches halfway/completion, the page jumps down to the thank you message while the video is still playing, which is pretty poor UI if someone is actually interested in/engaged with the video.

2. Any link within a task that doesn't have a `data-task` attribute is inert. For example, "Learn more about Mozilla Location Services" and "Learn more about encryption" do nothing when clicked (in Fx Beta & Dev Edition).

bedrock/mozorg/templates/mozorg/contribute/tasks/taskview-base.html
schalkneethling: Dang it, I thought I removed it....hmmmm

schalkneethling: Ah! I removed one of them.

media/css/mozorg/contribute/taskview.less
schalkneethling: Ah yes, I moved the code into the main `taskview.less` and forgot to delete that one. Nice catch, thanks!

schalkneethling:
"On the Joy of Coding task, when the video reaches halfway/completion, the page jumps down to the thank you message while the video is still playing, which is pretty poor UI if someone is actually interested in/engaged with the video."

Agree. I actually asked this a while back. I reckon we need to make a special case for this one i.e. the message does not scroll into view as the others does.

schalkneethling:
@jpetto ok updated. I have not done a great deal of cleanup in this pull request. Partly so it does not become more of a pain to review, and because I am not sure how the signup page will work/behave once merged.

There may be a older form of the `/signup` page that needs to be kept while localization happens(which I believe is the case) or, whether this will be a cut over once merged. I do not believe there is a anything in place currently to prevent this page from being shown to everyone.

Perhaps this is something that still needs to be added? Also, @flodolo mentioned there was a redirect that pmac added to redirect some locales to an old version of `/signup`

Would be good to get clarification on this with Jen before we merge this.

bedrock/mozorg/templates/mozorg/contribute/tasks/encryption.html
```diff
@@ -0,0 +1,31 @@
...
+{% block task_content %}
+<ol class="task-steps">
+  <li>
+    <header id="step_one" class="one">
+      <h3>{{ _('Take the pledge') }}</h3>
+    </header>
+    <div class="step-content">
+      {{ high_res_img('contribute/signup/encryption.png', {'alt': '', 'width': '620', 'class': 'feature-img'}) }}
+      <a href="https://advocacy.mozilla.org/encrypt/signup" class="button-blue" data-action="sign" data-task="encryption" data-step="one" data-complete="true" target="_blank">{{ _('Take the pledge') }}</a>
+      <p>{{ _('<a href="%s" rel="external" target="_blank">Learn more about encryption</a>, why it’s essential and why we need to protect it.')|format('https://advocacy.mozilla.org/encrypt/social/1') }}</p>
```
flodolo: @schalkneethling 
Did you change this string after I extracted them? Because otherwise I don't understand why I don't have the link in mine :-\
https://github.com/mozilla-l10n/www.mozilla.org/blob/master/de/mozorg/contribute/signup.lang#L164

schalkneethling: It had to be moved to a different spot on the page and is now wrapped in a `p` as apposed to a `h3` but, other than that, there should be no change. If it effects l10n I am truly sorry.

schalkneethling:
Also, I always found it weird that Mike mentions links being on the page below the video for joy of coding but, there are no links. With the latest changes, there is also no link to airmo to watch his video series. Something to bring up with Jen etc. perhaps.

schalkneethling:
@jpetto Also, the thank you message currently states "Thanks for getting involved" - perhaps this is not appropriate any longer. Maybe something more like "Thanks for watching". And perhaps, based on my previous comment "Thanks for watching. Keep watching on Air Mozilla"

bedrock/mozorg/templates/mozorg/contribute/tasks/encryption.html
flodolo: This was definitely changed between the time I extracted strings and now.

It would be nice to know what happened, sadly it's not possible without history. Given that you can squash commit while merging PRs, maybe it would be nice to stop amending history on PRs.

flodolo:
Given that localization has started, I would not add new strings. In case, there's a `Thanks!` available.

schalkneethling:
@jpetto How do you feel about this one? Ready to merge? Thx

jpetto:
Okay, I think we're ready to merge this one, though some follow-up bugs need to be filed:

1. Clean up unused forms/templates/views/assets from the old `/contribute/signup/` pages. jbertsch confirmed that this new signup page will be the one and only (regardless of up-front L10n coverage), so there's no need to retain previous versions.

2. Improve copy/UX where necessary (as mentioned in above comments).

I think it makes sense to have these as separate bugs, as one will require L10n changes.

That all being said, r+! Nice work @schalkneethling! :hamburger: :fries: :beer:

jpetto:
Bugs filed:

1. [bug 1271424 - copy/UX improvements](https://bugzilla.mozilla.org/show_bug.cgi?id=1271424)
2. [bug 1271422 - clean up old code](https://bugzilla.mozilla.org/show_bug.cgi?id=1271422)

schalkneethling:
o/\o Thanks for the review and filing the bugs @jpetto

PR #4103, opened 2016-04-29
Fix Bug 1264972, roll out winning version of plugin check page

schalkneethling:
## Description
Rolls out the winning version of the plugin check page tests.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1264972

## Testing
This page has been in production for a while as one of the variations that was tested so, a check off the page in Fx and a look over the code changes should suffice.

## Checklist
- [ ] Requires l10n changes.


schalkneethling:
@flodolo maybe take a look over this final version and see whether it needs an l10n flag. Thanks!

flodolo:
I don't see any string change in the .html template. If that's correct, there's no l10n impact

alexgibson:
I seem to remember we conditionally skipped a functional test when this test was on. Does it still need to be skipped, or can we run it for all browsers again?

https://github.com/mozilla/bedrock/blob/master/tests/functional/test_plugincheck.py#L10

schalkneethling:
@alexgibson I would think it is safe to run that test again. Let me run it locally and see.

alexgibson:
Sorry I meant this test (related to the newsletter), it moved since we started skipping it since I refactored the newsletter tests: https://github.com/mozilla/bedrock/blob/master/tests/functional/test_newsletter_embed.py#L36

schalkneethling:
@alexgibson Ok, ran the newsletter_embed test agains Fx and the plugincheck test agains Chrome and all pass. In newsletter_embed, the test is only skipped in non Fx browsers:

```
pytest.mark.skip_if_not_firefox((PluginCheckPage, None),
```

schalkneethling:
this is ready for code review and there is a card in the backlog under story number 46

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -33,12 +33,6 @@
 
 {% block body_class %}sky{% endblock %}
 
-{% block optimizely %}
```
alexgibson: Any reason why this block is being removed (as opposed to just turning off the waffle switch on prod?)

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -89,10 +85,23 @@
...
 <main role="main">
   <header id="main-feature">
-      <h1 class="large">{{_('Check Your Plugins')}}</h1>
+      <a href="{{ url('firefox') }}">{{ high_res_img('plugincheck/firefox-logo.png', {'alt': 'Mozilla Firefox', 'width': '145', 'class': 'firefox-logo'}) }}</a>
+      <h2 class="large">{{_('Check Your Plugins')}}</h2>
       <p class="preamble">{{_('Keeping your plugins up to date helps Firefox run safely and smoothly.')}}</p>
   </header>
 
+  <section class="newsletter-container">
+    {% if LANG == 'en-US' %}
```
alexgibson: Is there a reason why this is `en-US` only, and other locales see a large block of empty space next to the newsletter?

alexgibson:
On longer strings, the subheading wraps underneath the Firefox logo in the main header

plugincheck-de

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -89,10 +85,23 @@
...
+      <header>
+        <h3>{{ _('<span class="firefox">Firefox</span> + You') }}</h3>
+        <h4>{{ _('Get Firefox tips, tricks, news and more') }}</h4>
+      </header>
+      {{ email_newsletter_form(button_class='white') }}
+    {% else %}
```
alexgibson: Is there a need for the `{% else %}` here? Could the `{{ email_newsletter_form(button_class='white') }}` include just come after the conditional header?

alexgibson:
Nit: form is a little off center at smaller viewports

form-small-viewport

alexgibson:
Is it intentional that the thank you message appears below where the main form was displayed?

thankyou

alexgibson:
If a server error message is displayed, the form jumps down the page a bit

alexgibson:
This is more of a question for my own curiosity, but what was the consensus over concerns that moving the form here interferes with the primary purpose of the page? Updating vulnerable plugins is pretty important - are lots of people going to sign up for this newsletter expecting help with their vulnerability? (only to get none).

alexgibson:
A few UI related bits that need some small tweaks, but otherwise looking good.

I am sort of concerned that the placement on the page is going to be distracting for users and a bit of a confusing UX, but I guess this is a wider discussion past the scope of this PR.

bedrock/mozorg/templates/mozorg/plugincheck.html
alexgibson: Hmm, this also means non en-Us visitors get no context for what the form is used for? Are they going to think it has something to do with checking their plugins?

bedrock/mozorg/templates/mozorg/plugincheck.html
schalkneethling: Nope, I thought we generally removed these blocks once no optimizely tests are run against the page. If we generally leave them in place, then I will do so here as well.

bedrock/mozorg/templates/mozorg/plugincheck.html
schalkneethling: I don't think we need these conditionals any longer. @flodolo?

bedrock/mozorg/templates/mozorg/plugincheck.html
alexgibson: When Optimizely tests are finished we generally just turn the waffle switch off again. Removing the block means we have to add it back again next time we want to run another test, which requires another code change and prod push.

bedrock/mozorg/templates/mozorg/plugincheck.html
schalkneethling: Also here @flodolo, do we still need the conditionals or do we have enough coverage to remove them? Thx

flodolo: As far as I can tell l10n has nothing to do with these conditionals, this is en-US content because the newsletter is available only in English?

bedrock/mozorg/templates/mozorg/plugincheck.html
flodolo: You created this one because of this, I don't know the answer
https://bugzilla.mozilla.org/show_bug.cgi?id=1180891

bedrock/mozorg/templates/mozorg/plugincheck.html
flodolo: For reference: we don't have those strings, they were never translated.

alexgibson: Fwiw, here you're showing the exact same newsletter for both en-US and other locales (hence my comment about the conditional being superfluous).

bedrock/mozorg/templates/mozorg/plugincheck.html
alexgibson: I would suggest we find a heading for the newsletter that's already translated for an easy fallback here. Without the heading, it's not immediately obvious what the form is for.

bedrock/mozorg/templates/mozorg/plugincheck.html
alexgibson: So it looks like we're currently showing using a different newsletter for en-US in master here: https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templates/mozorg/plugincheck.html#L286

Should this be what we are doing here also?

schalkneethling: @alexgibson Nope, this should now always use the Firefox newsletter. I reckon here we just need a fallback heading for langs that do not have the new heading text not translated. The form helper can then be called outside the `if else` as you mentioned.

schalkneethling:
"This is more of a question for my own curiosity, but what was the consensus over concerns that moving the form here interferes with the primary purpose of the page? Updating vulnerable plugins is pretty important - are lots of people going to sign up for this newsletter expecting help with their vulnerability? (only to get none)."

I agree, I was very surprised when I saw that this was the winning variation that should be rolled out. While it does have the potential of increasing newsletter signups, it has shown to effect interaction with the plugin status sections.

With this said, you make a good point about what people will believe the content of those newsletters will be. I should also mentioned however that this page is scheduled for a bug UX and design overhaul in the next sprint so, these problems might be resolved as part of that work.

alexgibson:
> I agree, I was very surprised when I saw that this was the winning variation that should be rolled out. While it does have the potential of increasing newsletter signups, it has shown to effect interaction with the plugin status sections.

> With this said, you make a good point about what people will believe the content of those newsletters will be. I should also mentioned however that this page is scheduled for a bug UX and design overhaul in the next sprint so, these problems might be resolved as part of that work.

I think what we're doing here could potentially be considered a [dark pattern](http://darkpatterns.org/). Given that we've noted ourselves a drop in interaction with the plugin section of the page, then we're pretty much putting our own internal KPI's above the security of our users (I'm being brash here, but security shouldn't be taken lightly). It's good to hear that there will be a design/UX overhaul soon, as this is something I think we should give more thought to.

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -89,10 +91,23 @@
...
+      <a href="{{ url('firefox') }}">{{ high_res_img('plugincheck/firefox-logo.png', {'alt': 'Mozilla Firefox', 'width': '145', 'class': 'firefox-logo'}) }}</a>
+      <h2 class="large">{{_('Check Your Plugins')}}</h2>
       <p class="preamble">{{_('Keeping your plugins up to date helps Firefox run safely and smoothly.')}}</p>
   </header>
 
+  <section class="newsletter-container">
+    <aside id="newsletter-subscribe">
+      <div class="content container">
+        {% if l10n_has_tag('plugincheck_newsletter') %}
+          {{ email_newsletter_form(title=_('<span class="firefox">Firefox</span> + You'), subtitle=_('Get Firefox tips, tricks, news and more'), button_class='white') }}
```
alexgibson: Could we style the `span` directly here instead of needed to include `class="firefox"` in the string for l10n?

media/css/newsletter/fxnewsletter-subscribe.less
```diff
@@ -128,6 +128,7 @@
 
     #footer-email-errors {
         .span-all;
+        padding: (@baseLine / 2) 0;
```
alexgibson: Suggest changing this to just `padding-bottom: @baseLine;` to avoid extra padding at the top too?

alexgibson:
Thank you message heading has some rogue `text-shadow`

text-shadow

media/css/plugincheck/plugincheck.less
```diff
@@ -15,22 +23,64 @@
...
+        letter-spacing: -0.25px;
         text-shadow: 0px 1px 0px rgba(255, 255, 255, 0.75);
         color: rgb(72, 72, 72);
+
+        @media only screen and (max-width: @breakDesktop) {
+            max-width: 500px;
+            .font-size(1.2rem);
+        }
+
+        @media only screen and (max-width: @breakDesktop) {
```
alexgibson: Duplicate media query to above?

bedrock/mozorg/templates/mozorg/plugincheck.html
schalkneethling: We can surely do that, and it seems safe as strings have not been extracted or exposed.

media/css/plugincheck/plugincheck.less
```diff
@@ -15,22 +23,64 @@
...
+            .font-size(3.5rem);
+        }
+    }
+
+    @media only screen and (max-width: @breakTablet) {
+        .large {
+            .font-size(3rem);
+        }
+
+        img[data-high-res] {
```
alexgibson: What is this selector for, is it perhaps overly-specific and could apply to more than what we intend?

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -89,10 +91,23 @@
...
-      <h1 class="large">{{_('Check Your Plugins')}}</h1>
+      <a href="{{ url('firefox') }}">{{ high_res_img('plugincheck/firefox-logo.png', {'alt': 'Mozilla Firefox', 'width': '145', 'class': 'firefox-logo'}) }}</a>
+      <h2 class="large">{{_('Check Your Plugins')}}</h2>
       <p class="preamble">{{_('Keeping your plugins up to date helps Firefox run safely and smoothly.')}}</p>
   </header>
 
+  <section class="newsletter-container">
+    <aside id="newsletter-subscribe">
+      <div class="content container">
+        {% if l10n_has_tag('plugincheck_newsletter') %}
```
alexgibson: Ping @flodolo - new l10n tag here for your reference

alexgibson:
Few small things/questions otherwise looking good, r+wc 🌵

media/css/plugincheck/plugincheck.less
schalkneethling: This overrides: https://github.com/mozilla/bedrock/blob/master/media/css/sandstone/sandstone.less#L313

I could do:
```
.js img[data-high-res].firefox-logo {
    display: block;
}
```

schalkneethling: hmmm, seems this may not be needed any longer.

schalkneethling: Not needed anymore indeed.

PR #4105, opened 2016-04-30
[bug 1266553] Add core dataLayer object and page ID data attributes

JamesDLorence:
bug 1266553 - https://bugzilla.mozilla.org/show_bug.cgi?id=1266553

-Updated those pages discussed with the data-gtm-page-id attribute.
-Added the core-datalayer.js file to house the library that will be built out for the core dataLayer object.
-Added Jasmine test for core-datalayer.js.

### DIFF
--- a/bedrock/thunderbird/templates/thunderbird/releases/index.html
+++ b/bedrock/thunderbird/templates/thunderbird/releases/index.html
@@ -4,6 +4,8 @@
 
 {% extends "thunderbird/base-resp.html" %}
 
+{% block gtm_page_id %}data-gtm-page-id="/thunderbird/releases/"{% endblock %}
+
 {% block page_title %}{{ _('Mozilla Thunderbird Release Notes') }}{% endblock %}
 {% block page_desc %}{{ _('Release notes for each version of Thunderbird.') }}{% endblock %}
 {% block body_id %}thunderbird-releases-index{% endblock %}
--- a/media/js/base/core-datalayer-init.js
+++ b/media/js/base/core-datalayer-init.js
@@ -1,0 +1,15 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// init core dataLayer object and push into dataLayer
+$(function() {
+    if (Mozilla && Mozilla.Analytics) {
+        var dataLayer = window.dataLayer = window.dataLayer || [];
+
+        dataLayer.push({
+            'event': 'core-datalayer-loaded',
+            'pageId': Mozilla.Analytics.getPageId()
+        });
+    }
+});
--- a/media/js/base/core-datalayer.js
+++ b/media/js/base/core-datalayer.js
@@ -1,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// create namespace
+if (typeof Mozilla == 'undefined') {
+    var Mozilla = {};
+}
+
+/**
+* Utility class for core dataLayer object to track contextual user and page data
+*/
+
+(function() {
+    var Analytics = {};
+
+    /** Returns page ID used in Event Category for GA events tracked on page.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} GTM page ID.
+    */
+    Analytics.getPageId = function(path) {
+        var pageId = $('html').attr('data-gtm-page-id');
+        var pathName = path ? path : document.location.pathname;
+
+        return pageId ? pageId : pathName.replace(/\/[^\/]*/, '');
+    };
+
+    Mozilla.Analytics = Analytics;
+
+})();
+
--- a/tests/unit/spec/base/core-datalayer.js
+++ b/tests/unit/spec/base/core-datalayer.js
@@ -1,0 +1,30 @@
+/* For reference read the Jasmine and Sinon docs
+ * Jasmine docs: http://pivotal.github.io/jasmine/
+ * Sinon docs: http://sinonjs.org/docs/
+ */
+
+/* global describe, beforeEach, afterEach, it, expect */
+
+describe('core-datalayer.js', function() {
+
+    describe('PageId', function(){
+        var html = document.documentElement;
+
+        afterEach(function() {
+            html.removeAttribute('data-gtm-page-id');
+        });
+
+        it('will grab data-gtm-page-id value if present on <html> element', function(){
+            html.setAttribute('data-gtm-page-id', 'test');
+
+            expect(Mozilla.Analytics.getPageId('/en-US/firefox/new/')).toBe('test');
+        });
+
+        it('will grab the pathname minus the first directory if no data-gtm-page-id value is present on <html> element', function(){
+            expect(Mozilla.Analytics.getPageId('/en-US/firefox/new/')).toBe('/firefox/new/');
+        });
+    });
+
+});
+
+
--- a/bedrock/base/templates/404.html
+++ b/bedrock/base/templates/404.html
@@ -5,6 +5,8 @@
 
 {% extends "base-resp.html" %}
 
+{% block gtm_page_id %}data-gtm-page-id="404"{% endblock %}
+
 {% block page_title %}{{ _('404: Page Not Found') }}{% endblock %}
 
 {% block extrahead %}
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -3,7 +3,7 @@
 <!doctype html>
 {# Note the "windows" class, without javascript platform-specific
      assets default to windows #}
-<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}">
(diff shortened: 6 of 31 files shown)


media/js/base/core-datalayer.js
```diff
@@ -0,0 +1,34 @@
...
+
+/**
+* Utility class for core dataLayer object to track contextual user and page data
+*/
+
+(function() {
+    var dataLayer = window.dataLayer = window.dataLayer || [];
+    var Analytics = {};
+
+    // Reurns page ID used in Event Category for GA events tracked on page
```
alexgibson: Typo `returns`

media/js/base/core-datalayer.js
```diff
@@ -0,0 +1,34 @@
...
+    Analytics.getPageId = function() {
+        var pageId = $('html').attr('data-gtm-page-id');
+
+        return pageId ? pageId : document.location.pathname.replace(/\/[^\/]*/, '');
+    };
+
+    // Push core datalayer object
+    dataLayer.push({
+        'event': 'core-dataLayer-loaded',
+        'pageId': Analytics.getPageId()
```
alexgibson: As we're writing tests for this file, and will be expanding on the functions defined here in the future, I think it makes sense to separate out when the functions are being run. Currently this is executing before the tests are run, which while isn't breaking anything, probably isn't ideal - and could lead to unexpected results when we expand on things in the future. 

I would suggest creating a file called `core-datalayer-init.js` which is not included in the test runner, similar to what we do with global.js here: https://github.com/mozilla/bedrock/blob/master/media/js/base/global-init.js

tests/unit/spec/base/core-datalayer.js
```diff
@@ -0,0 +1,27 @@
...
+        it('will grab data-gtm-page-id value if present on <html> element', function(){
+            $('html').attr('data-gtm-page-id', 'test');
+
+            expect(Mozilla.Analytics.getPageId()).toBe('test');
+        });
+
+        it('will grab the pathname minus the first directory if no data-gtm-page-id value is present on <html> element', function(){
+            $('html').removeAttr('data-gtm-page-id');
+
+            expect(Mozilla.Analytics.getPageId()).toBe('');
```
alexgibson: Hmm, this isn't really testing that the function works as expected. It could also start failing if we decided to change the location/context of where the tests run. In order to make testing a bit easier, I would suggest changing the function slightly so it accepts passing of parameters. Something like this should work:

```javascript
/**
 * Returns page ID used in Event Category for GA events tracked on page
 * @param  {String} id - GTM page ID
 * @return {String} path - URL path name used as a fallback if id
 */
Analytics.getPageId = function(id, path) {
    var pageId = id ? id : $('html').attr('data-gtm-page-id');
    var pathName = path ? path : document.location.pathname;

    return pageId ? pageId : pathName.replace(/\/[^\/]*/, '');
};
```

We could then test it more thoroughly, asserting multiple different paths return as expected.

tests/unit/spec/base/core-datalayer.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+/* global describe, beforeEach, afterEach, it, expect, sinon, spyOn, jasmine */
+
+describe('core-datalayer.js', function() {
+
+    describe('PageId', function(){
+
+        it('will grab data-gtm-page-id value if present on <html> element', function(){
+            $('html').attr('data-gtm-page-id', 'test');
```
alexgibson: If we're modifying this element in a test, we should clean up after each test so as not to avoid unexpected results

```javascript
afterEach(function () {
    html.removeAttribute('data-gtm-page-id');
});
```

We could also store our reference to this DOM node in a variable for reuse in the test suite:

```javascript
var html = document.documentElement;
```

alexgibson:
Do we also need to add page ID's to recently added pages like:

- `/firstrun/learnmore/`
- `/secondrun/`

alexgibson:
@JamesDLorence - this is looking good, thanks. Just a few recommendations for easier testing, let me know if anything doesn't make sense.

media/js/base/core-datalayer.js
JamesDLorence: I don't see a typo?

alexgibson: 'Reurns'

media/js/base/core-datalayer.js
JamesDLorence: Your request here makes sense, but I don't fully understand the benefit of having the test run before the functions are called? Am curious for my own benefit and so I make better decisions next time.

tests/unit/spec/base/core-datalayer.js
JamesDLorence: I like it. This makes a lot of sense.

media/js/base/core-datalayer.js
alexgibson: What I'm saying is, running the functions as they are loaded into the test runner, but before the tests run, can lead to unexpected behavior in said tests, or false positives etc. This function is harmless enough, but if we add others here it could lead to confusion depending on their behavior or order of execution.

tests/unit/spec/base/core-datalayer.js
JamesDLorence: Curious, does it really matter if I setup the `removeAttribute` call in an `afterEach` call or could I just do it in the spec since I only add it in the first spec?

media/js/base/core-datalayer.js
JamesDLorence: Gotchya. Thanks.

JamesDLorence:
@alexgibson - I am having a little trouble identifying the `/secondrun/` template, is it `onboarding-base.html`? And is the URL for the Developer edition pages the same as the normal '/firstrun` and `/whatsnew` URLs, or is there an added `/developer/` directory added or something?

alexgibson:
@JamesDLorence - looks like `/secondrun` has several possible templates: https://github.com/mozilla/bedrock/tree/master/bedrock/firefox/templates/firefox/onboarding

All but one seem to extend `onboarding-base.html`, but `fxa-simple.html` seems to extend `firefox/firstrun/firstrun.html`, so you may want to override `fxa-simple.html` with it's own specific page ID.

The dev edition pages are as follows:

- https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/dev-whatsnew.html
- https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/dev-firstrun.html

They can be accessed using the same URL convention as regular firstrun and whatnew pages, but they key difference is the version number in the URL . Dev Edition uses `a2` in the version, so for example `/firefox/48.0a2/whatsnew/`, `/firefox/48.0a2/firstrun/`.

JamesDLorence:
I had to add an `if/else` statement for the firstrun and secondrun data attributes becasue the templates were being used for different versions interchangeably.

media/js/base/core-datalayer-init.js
```diff
@@ -0,0 +1,18 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/* global initDownloadLinks, updateDownloadTextForOldFx, initMobileDownloadLinks,
```
alexgibson: You can remove these global eslint definitions as they are not part of this file.

tests/unit/spec/base/core-datalayer.js
```diff
@@ -0,0 +1,30 @@
+/* For reference read the Jasmine and Sinon docs
+ * Jasmine docs: http://pivotal.github.io/jasmine/
+ * Sinon docs: http://sinonjs.org/docs/
+ */
+
+/* global describe, beforeEach, afterEach, it, expect, sinon, spyOn, jasmine */
```
alexgibson: You can also remove `sinon, spyOn, jasmine` from here too.

alexgibson:
I notice we don't have page ID's for `firstrun/learnmore/` do we need to capture those too?

media/js/base/core-datalayer-init.js
```diff
@@ -0,0 +1,18 @@
...
+   initLangSwitcher */
+
+// init core dataLayer object and push into dataLayer
+$(document).ready(function() {
```
alexgibson: Let's change this ready handler to the equivalent but slightly smaller version:

```javascript
$(function() {
  // code
});
```

bedrock/firefox/templates/firefox/releases/index.html
```diff
@@ -1,5 +1,15 @@
 {% extends "firefox/base-resp.html" %}
 
+{% block gtm_page_id %}
+    {% if release.product == 'Firefox for Android' %}
```
alexgibson: `release` is undefined here and the page throws a 500

tests/unit/spec/base/core-datalayer.js
```diff
@@ -0,0 +1,30 @@
...
+        var html = document.documentElement;
+
+        afterEach(function() {
+            html.removeAttribute('data-gtm-page-id');
+        });
+
+        it('will grab data-gtm-page-id value if present on <html> element', function(){
+            html.setAttribute('data-gtm-page-id', 'test');
+
+            expect(Mozilla.Analytics.getPageId()).toBe('test');
```
alexgibson: For good measure let's pass in a `path` value here as well, just to make it clear that the data attitude is being used instead.

JamesDLorence:
I don't think I need the data attribute in `/release/index.html` so I just removed the whole block.

bedrock/thunderbird/templates/thunderbird/releases/index.html
```diff
@@ -4,6 +4,8 @@
 
 {% extends "thunderbird/base-resp.html" %}
 
+{% block gtm_page_id %}data-gtm-page-id="/thunderbird/releasenotes/"{% endblock %}
```
alexgibson: This should be `/thunderbird/releases/`

alexgibson:
Looking good, just one last amend and then I think we're good to merge this. r+wc, nice work 🍰

alexgibson:
@JamesDLorence - ping (in case you didn't notice). One last amend and then we can merge this?

JamesDLorence:
@alexgibson - sorry for the delay, I wast out OOO all last week.

alexgibson:
r+ 🐝

PR #4106, opened 2016-05-02
[fix bug 1267736] Move /patents content to /policy/patents

craigcook:
## Description
Moving the templates and updating/redirecting the URLs. I suspect a single pass-through style redirect would be better than redirecting each page individually, but I don't really grok regex enough to make that work. I'm very much open to suggestion.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1267736

## Testing
Verify that the redirects work (e.g. `/about/patents/guide/` should redirect to `/about/policy/patents/guide/`).
This shouldn't have any l10n impact as I don't think the pages are localized. If they are I think the only impact would be relocating lang files; there aren't any new or changed strings.

## Checklist
- [ ] Requires l10n changes.


jpetto:
LGTM. r+ 🏀

PR #4109, opened 2016-05-03
Fix Bug 1269700 - text overlapping in German Version on the Smarton site

kyoshino:


kyoshino:
@craigcook r?

craigcook:
Ah how embarrassing! Of course the buttons should be hidden in the "off" state, I just never noticed they weren't because it was white on white. It's obvious in German because the text is so much longer. r+ :shaved_ice:

PR #4114, opened 2016-05-05
[fix bug 1266481]  Add a maintenance message when email pref center is unavailable

craigcook:
## Description
Hides the form and adds a short maintenance message behind the waffle switch `newsletter-maintenance-mode`

@flodolo These are new strings so it may be better to put them in an l10n tag, but we don't have any useful fallback content to show if they aren't localized so maybe showing English is better than nothing? But I believe the plan is to get good locale coverage before we merge this. Let me know if a tag is preferable.

## Bugzilla
https://bugzilla.mozilla.org/show_bug.cgi?id=1266481

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/newsletter/templates/newsletter/existing.html
+++ b/bedrock/newsletter/templates/newsletter/existing.html
@@ -14,14 +14,23 @@
   {% block main_feature %}
     <div id="main-feature">
       <h1>{{ _('Manage your Email Preferences') }}</h1>
+
+    {% if waffle.switch('newsletter-maintenance-mode') %}
+      <p class="billboard">
+        {{ _('This page is in maintenance mode and is temporarily unavailable.') }}<br>
+        {{ _('To update your email preferences, please check back in a little while. Thanks!') }}
+      </p>
+    {% else %}
       <p>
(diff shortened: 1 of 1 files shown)


flodolo:
I guess English in this case it's better than nothing. Also, I hope people won't see this that frequently...

pmac:
@flodolo they shouldn't. this is being put in place now specifically for a brief period (12 to 24 hours) when we'll have to disable email preferences while the newsletter backend systems are moved from one provider to another. I will potentially be nice to have if our provider were to experience prolonged downtime in future, but we don't foresee that. We just thought it'd be preferable to have the message well translated since this page is so well translated and it'd be jarring to users and annoying to translators to have the maintenance message in English. Please let us know if there's another procedure you'd like to follow for this sort of short-lived but important content though.

flodolo:
I think it's fine to do it like this. The comment above was meant as "We don't need to wait for l10n, it's good to have the strings but they won't be extremely visible".

pmac:
Gotcha. Agreed.

craigcook:
Since this is behind a waffle switch and we won't actually need it for a while I think this is safe to merge (pending an r+) even while l10n is in progress.

PR #4117, opened 2016-05-09
[fix bug 126396] Redesigned Mozilla newsletter form

craigcook:
## Description
Removes the test code and makes the new design permanent on home, /about, and /about/legal.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1269396

## Testing
English locales should see the Mozilla form while other locales get the Firefox form, but it's the same design for both. This shouldn't have any l10n impact.

## Checklist
- [ ] Requires l10n changes.


alexgibson:
Would it worth making the "thank you" message centered when the form gets hidden, so there's not as much blank space?

thank-you

bedrock/legal/templates/legal/index.html
```diff
@@ -81,8 +81,16 @@ <h1 id="main-feature" class="center large">{{ self.page_title() }}</h1>
 
 {% block email_form %}
   {% if LANG.startswith('en-') %}
```
alexgibson: Suggest moving the `{% if LANG.startswith('en-') %}` conditional inside the `<aside id="newsletter-subscribe">` container to avoid repeating the markup?

bedrock/mozorg/templates/mozorg/about.html
```diff
@@ -126,3 +126,19 @@ <h1 class="title-shadow-box">{{ _('Get to know Mozilla') }}</h1>
 </article>
 
 {% endblock %}
+
+{% block email_form %}
+  {% if LANG.startswith('en-') %}
```
alexgibson: Same here, conditional could be inside the containing markup?

bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -347,17 +347,19 @@ <h2 class="primary go">{{ _('Mozilla Developer Network') }}</h2>
     </div>
   </section>
 
-{% block newsletter_test %}
-  <section class="module" id="newsletter-signup" data-ga-label="Get Mozilla updates">
+{% if LANG.startswith('en-') %}
```
alexgibson: Same again here?

media/css/mozorg/about-base.less
```diff
@@ -3,7 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 @import "../sandstone/lib.less";
-@import "newsletter-test.less";
+@import "../newsletter/moznewsletter-subscribe.less";
```
alexgibson: Suggest moving these to page bundles instead of using `@import`?

media/css/mozorg/home/home.less
```diff
@@ -4,7 +4,7 @@
 
 @import "../../sandstone/lib.less";
 @import "home-grid.less";
-@import "../newsletter-test.less";
+@import "../../newsletter/moznewsletter-subscribe.less";
```
alexgibson: Suggest moving these to page bundles instead of using `@import`?

alexgibson:
Looks good to me, just a couple of things to tidy up / suggestions. r+wc 📧

alexgibson:
One last thing - could the server error message styles be tweaked a little seeing as they are red on a red background?

server-error

craigcook:
@alexgibson All points spot on, thanks! Ive moved the imports into the bundles, made the thank you message span-all, and inverted the error message colors.

alexgibson:
r+ 🍰 !

PR #4121, opened 2016-05-11
Respect WEB_CONCURRENCY environment variable

jgmize:
@pmclanahan r?

### DIFF
--- a/docker/run-prod.sh
+++ b/docker/run-prod.sh
@@ -1,3 +1,3 @@
 #!/bin/bash -xe
 ./docker/run-common.sh
-gunicorn wsgi.app:application -b 0.0.0.0:${PORT:-8000} -w 2 --error-logfile - --access-logfile - --log-level ${LOGLEVEL:-info}
+gunicorn wsgi.app:application -b 0.0.0.0:${PORT:-8000} -w ${WEB_CONCURRENCY:-2} --error-logfile - --access-logfile - --log-level ${LOGLEVEL:-info}


docker/run-prod.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash -xe
 ./docker/run-common.sh
-gunicorn wsgi.app:application -b 0.0.0.0:${PORT:-8000} -w 2 --error-logfile - --access-logfile - --log-level ${LOGLEVEL:-info}
+gunicorn wsgi.app:application -b 0.0.0.0:${PORT:-8000} -w ${GUNICORN_WORKERS:-2} --error-logfile - --access-logfile - --log-level ${LOGLEVEL:-info}
```
pmac: Only note is that it might be better to make the variable non-server-specific. Maybe `WSGI_WORKERS`? On basket it's `WSGI_NUM_WORKERS`, but that seems verbose now. It's fine if you like `GUNICORN_WORKERS` too, just a suggestion.

r+

pmac:
Made a suggestion but you are free to take or leave it. I'm mostly surprised this wasn't already done :(

r+

pmac:
Also note that gunicorn itself looks for `WEB_CONCURRENCY` to set the number of workers by default. I like being explicit, but [it's a thing](http://docs.gunicorn.org/en/latest/settings.html#workers).

jgmize:
Good points, @pmclanahan, I should have thought to check the docs to see if there was an existing environment variable. I've updated it to use the documented `WEB_CONCURRENCY` you pointed out, but keeping our default at the current 2, rather than omitting the `-w` arg entirely and explicitly setting `WEB_CONCURRENCY` in every deployment.

PR #4123, opened 2016-05-11
[fix bug 1272030] Update /contribute/signup task event handling.

jpetto:
## Description

Improve stability of delegated click handling on various contribute tasks.

## Bugzilla link

[bug 1272030](https://bugzilla.mozilla.org/show_bug.cgi?id=1272030)

## Testing

Make sure task completion functions as expected in A-grade browsers and IE 10+. Make sure non-completion events (e.g. clicking "Learn more about encryption" link) work normally and do not initiate a task completion event.

IE 9 and below do not support `visibilityChange`, and thus do not visually indicate task completion. I could find no support for `document.msHidden` when testing IE 8 and 9, so I removed that conditional from the code.

## Checklist

### DIFF
--- a/media/js/mozorg/contribute/taskview.js
+++ b/media/js/mozorg/contribute/taskview.js
@@ -22,20 +22,17 @@
      * https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API
      */
     function getVisibilityStateEventKeyword() {
-        // for browser that support the event unprefixed
+        var evtName = undefined;
+
         if (typeof document.hidden !== 'undefined') {
-            return 'visibilitychange';
+            // for browser that support the event unprefixed
+            evtName = 'visibilitychange';
+        } else if (typeof document.webkitHidden !== 'undefined') {
+            // for Chrome 13 and lower
+            evtName = 'webkitvisibilitychange';
         }
 
-        // for Chrome 13 and lower
-        if (typeof document.webkitHidden !== 'undefined') {
-            return 'webkitvisibilitychange';
-        }
-
-        // for older versions of IE
-        if (typeof document.msHidden !== 'undefined') {
-            return 'msvisibilitychange';
-        }
+        return evtName;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jpetto:
@alexgibson or @schalkneethling r?

media/js/mozorg/contribute/taskview.js
```diff
@@ -22,20 +22,17 @@
...
+        if (typeof document.hidden !== 'undefined') {
+            // for browser that support the event unprefixed
+            evtName = 'visibilitychange';
+        } else if (typeof document.webkitHidden !== 'undefined') {
+            // for Chrome 13 and lower
+            evtName = 'webkitvisibilitychange';
         }
 
-        // for older versions of IE
-        if (typeof document.msHidden !== 'undefined') {
```
schalkneethling: So, would this not be needed for IE10? Although it is not mentioned here[1], it is mentioned in this article[2] that IE used a ms prefixed property when implementing page visibility in IE10

[1] https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API
[2] https://www.nczonline.net/blog/2011/08/09/introduction-to-the-page-visibility-api/

schalkneethling:
r+ w/c 🎐

media/js/mozorg/contribute/taskview.js
jpetto: I tested in my Windows 8 VM and IE 10 reported support for/worked with the standard `visibilitychange` event name. Can I use also shows [full support in IE 10](http://caniuse.com/#feat=pagevisibility).

I see a few blogs (including MDN) reference `document.msHidden`, but I couldn't find any official/MS documentation on it, nor did it work in IE 8 or 9. Maybe it was included in a preview release of IE 10 or something?

I suppose it's not a big deal to keep the check in the code, but as far as I can tell it's not doing anything. What do you think?

schalkneethling: I completely agree. Seeing it works without the check, let's not include it.

PR #4124, opened 2016-05-11
Update CORS handling to work with preflight

pmac:
Also enable CORS support for newsletter view. This should help static sites like for Viewsource to fully support newsletter signups with decent fallback for non-ajax posts.

### DIFF
--- a/bedrock/mozorg/middleware.py
+++ b/bedrock/mozorg/middleware.py
@@ -4,7 +4,6 @@
 
 import datetime
 from email.utils import formatdate
-import re
 import time
 
 from django.conf import settings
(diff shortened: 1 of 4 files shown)


pmac:
@stephaniehobson this should help you out with viewsource. You should also be able to steal the newsletters JS lib in bedrock and use it with slight modifications for the submission if you want.

pmac:
@jgmize r?

jgmize:
The tests are failing due to `ImportError: cannot import name CrossOriginResourceSharingMiddleware` in `bedrock/mozorg/tests/test_middleware.py` -- looks like you need to remove the tests for the middleware you removed.

pmac:
D'oh! Fixed. Thanks.

PR #4126, opened 2016-05-12
Use latest tagged commit instead of fetching url

jgmize:
@glogiotatidis r?


glogiotatidis:
Ok let's remove COMMIT_URL from the jenkins jobs as well.

glogiotatidis:
r+ !

PR #4129, opened 2016-05-16
Remove running selenium containers after local tests only.

glogiotatidis:
@davehunt when the $DRIVER=Remote it seems that we run the tests [locally on the CI node](https://github.com/mozilla/bedrock/blob/master/docker/jenkins/run_integration_tests.sh#L20-49). Is `Remote` the correct characterization or do we mean `Locally`?

### DIFF
--- a/docker/jenkins/cleanup_after_functional_tests.sh
+++ b/docker/jenkins/cleanup_after_functional_tests.sh
@@ -1,13 +1,15 @@
 #!/bin/bash
 
-docker stop bedrock-code-${BUILD_NUMBER}
-docker rm bedrock-code-${BUILD_NUMBER}
+if [ "${DRIVER}" = "Remote" ]; then
+    docker stop bedrock-code-${BUILD_NUMBER}
+    docker rm bedrock-code-${BUILD_NUMBER}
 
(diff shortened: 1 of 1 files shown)


davehunt:
`Remote` is a little misleading, but it means that the tests will run against a 'remote WebDriver' instance (an instance of a Selenium server). This server could be running locally, however in Bedrock's CI the instance is running in a docker container. A truly local instance would be $DRIVER=Firefox (for example) which would launch Firefox on the host instead of via a Selenium server. In Bedrock's CI we use Selenium Grid and docker so that we can run tests in parallel and do not need to rely on packages being available on the host.

glogiotatidis:
> Remote is a little misleading, but it means that the tests will run against a 'remote WebDriver' instance (an instance of a Selenium server). This server could be running locally, however in Bedrock's CI the instance is running in a docker container. A truly local instance would be $DRIVER=Firefox (for example) which would launch Firefox on the host instead of via a Selenium server. In Bedrock's CI we use Selenium Grid and docker so that we can run tests in parallel and do not need to rely on packages being available on the host.

understood, thanks!

glogiotatidis:
@jgmize r?

jgmize:
Thanks @glogiotatidis and @davehunt for the explanation

PR #4130, opened 2016-05-16
[fix bug 1268292] 'Refresh Firefox' button doesn't do anything when clicked

alexgibson:
## Description
- Only show the 'Refresh Firefox' button for up-to-date users if UITour is working.
- Now that the download button overhaul has landed, this PR also removes a ton of duplication for scene1.js, since the only two differences between the old and new pages are now obsolete 😄 

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1268292

## Testing
- Using up-to-date Firefox on release channel, visit the page with `browser.uitour.enabled` set to `false`. The button should no longer be seen (test both old and new templates).
- Would be good to vet that I've also not been too heavy-handed in removal of anything else that might not be obvious.

## Checklist
- [ ] Requires l10n changes.


jpetto:
This looks solid to me. Dropping an entire `.js.` file is a great win. 😀 

r+!

PR #4133, opened 2016-05-16
[Bug 1241969] don't ignore the requested language when opening /ab-CD/firefox/

rimas-kudelis:
Same applies to /ab-CD/firefox/channel/android/, so fixed it as well.

https://bugzilla.mozilla.org/show_bug.cgi?id=1241969


jgmize:
thanks @rimas-kudelis!

PR #4134, opened 2016-05-17
[fix bug 1271424] Improve copy and UX on some /contribute/signup tasks

alexgibson:
## Description
1.) Hides the "Want to get involved? Here’s how." steps on /contribute until we have email signups again. Shows a simple "Get Involved" CTA reusing existing strings.
2.) Adds a link underneath the Joy of Coding video that says "View more Joy of Coding videos on Air Mozilla" (since the video suggests there are links below on the page)
3.) Changes the Joy of Coding thank you message from "Thanks for getting involved" to "Thanks for watching".
4.) Changes the Dev Tools Challenger CTA button from "Learn More" to "Visit Dev Tools Challenger".
5.) Fixes the broken App Store button link on the firefox mobile task page.
6.) Changes the encryption task title from "Take the pledge" to "Become an encryption champion". See [Bug 1272023](https://bugzilla.mozilla.org/show_bug.cgi?id=1272023) for context.
7.) Moves the Stumbler "Learn more" copy from above the CTA to below.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1271424

## Checklist
- [x] Requires l10n changes.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/contribute/index.html
+++ b/bedrock/mozorg/templates/mozorg/contribute/index.html
@@ -108,6 +108,7 @@
 
   <section id="landing-howto" class="section landing-howto">
     <div class="container">
+      {# Hiding steps until we re-instate email signups
       <h2 class="section-title">{{ _('Want to get involved? Here’s how.') }}</h2>
       <ol class="steps">
         <li class="step1">{{ _('Choose what interests you and tell us how you’d like to help') }}</li>
@@ -115,6 +116,9 @@
         <li class="step3">{{ _('Create your Mozillians account and share your contributions with the world') }}</li>
       </ol>
       <p class="cta"><a href="{{ url('mozorg.contribute.signup') }}" class="button" data-link-type="button" data-link-name="Get started">{{ _('Get started') }}</a></p>
+      #}
+      <h2 class="section-title">{{ _('Join a global community of game-changers.') }}</h2>
+      <p class="cta"><a href="{{ url('mozorg.contribute.signup') }}" class="button" data-link-type="button" data-link-name="Get Involved">{{ _('Get Involved') }}</a></p>
     </div>
   </section>
 
--- a/bedrock/mozorg/views.py
+++ b/bedrock/mozorg/views.py
@@ -247,7 +247,7 @@ class ContributeSignupNew(TemplateView):
         return [template]
 
 
-class ContributeTaskView(TemplateView):
+class ContributeTaskView(l10n_utils.LangFilesMixin, TemplateView):
 
     tasks = [
         'devtools-challenger',
--- a/bedrock/mozorg/templates/mozorg/contribute/tasks/devtools-challenger.html
+++ b/bedrock/mozorg/templates/mozorg/contribute/tasks/devtools-challenger.html
@@ -24,7 +24,13 @@
       </header>
       <div class="step-content">
         <img src="{{ static('img/contribute/signup/devtools.png') }}" width="620" class="feature-img" alt="" />
-        <a href="http://devtoolschallenger.com/" rel="external" class="button-blue action" data-action="challenger" data-task="devtools" data-step="one" data-complete="true" target="_blank">{{ _('Learn More') }}</a>
+        <a href="http://devtoolschallenger.com/" rel="external" class="button-blue action" data-action="challenger" data-task="devtools" data-step="one" data-complete="true" target="_blank">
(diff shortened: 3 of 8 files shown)


alexgibson:
String changes are not yet confirmed or final - will update here when things are confirmed.

alexgibson:
OK, going to mark this one as ready for review. 

@flodolo - I've run by the string changes with Holly, so we should be good here. All new strings are behind an l10n tag.

bedrock/mozorg/views.py
```diff
@@ -247,7 +247,7 @@ def get_template_names(self):
         return [template]
 
 
-class ContributeTaskView(TemplateView):
+class ContributeTaskView(l10n_utils.LangFilesMixin, TemplateView):
```
jpetto: Out of curiosity, what's the purpose of this new param?

jpetto:
What do you think about left-aligning the "Watch the Video" button on the Joy of Coding task? All other blue buttons beneath the media (image/video) are left-aligned.

On that same note, the "Get it on Google Play" button on the Stumbler task is also centered. However, it's not a blue button, so...maybe that one's okay as is?

jpetto:
Code as-is looks good. One small UI suggestion, but otherwise solid. r+wc

bedrock/mozorg/views.py
alexgibson: I discovered this is required to use conditional l10n tags in the task templates. Without this, the templates throw an error.

alexgibson:
> What do you think about left-aligning the "Watch the Video" button on the Joy of Coding task? All other blue buttons beneath the media (image/video) are left-aligned.

I mulled this over too - seeing as you picked up on it also, I say let's make it all consistent 👍 

> On that same note, the "Get it on Google Play" button on the Stumbler task is also centered. However, it's not a blue button, so...maybe that one's okay as is?

The app store buttons on the mobile firefox task are left aligned I notice too, so perhaps this is reason to left align on the stumbler task too.

alexgibson:
@jpetto - thanks for the review, updated.

bedrock/mozorg/views.py
jpetto: Ah, good to know!

jpetto:
📋

bedrock/mozorg/templates/mozorg/contribute/index.html
```diff
@@ -108,13 +108,17 @@ <h3 class="fn given-name">Michael</h3>
...
+      {# Hiding steps until we re-instate email signups
       <h2 class="section-title">{{ _('Want to get involved? Here’s how.') }}</h2>
       <ol class="steps">
         <li class="step1">{{ _('Choose what interests you and tell us how you’d like to help') }}</li>
         <li class="step2">{{ _('Stand by for an email that will connect you to the Mozilla community') }}</li>
         <li class="step3">{{ _('Create your Mozillians account and share your contributions with the world') }}</li>
       </ol>
       <p class="cta"><a href="{{ url('mozorg.contribute.signup') }}" class="button" data-link-type="button" data-link-name="Get started">{{ _('Get started') }}</a></p>
+      #}
+      <h2 class="section-title">{{ _('Join a global community of game-changers.') }}</h2>
```
flodolo: @alexgibson 
Just realized that this introduced a new string without a tag, since we have only `Join a global community of game-changersJoin a global community of game-changers` in the file (no final dot). Can we use the existing one? Otherwise we need a tag :-\

alexgibson: Bah, sure I'll follow up with a fix. Sorry for the slip!

alexgibson: @flodolo are you sure? I copied the string directly from above in the same file (where it also has a period already) https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templates/mozorg/contribute/index.html#L27

alexgibson: Looks ok on dev: https://www-dev.allizom.org/de/contribute/

flodolo: Argh, sorry, my fault 😭 

I'm training @peiying2 to take over l10n for mozilla.org and didn't realize that the error was in the new update, adding a string without the period that is not part of the code, and not in the code itself.

I clearly shouldn't look at code in the morning.

PR #4139, opened 2016-05-19
[fix bug 1257796] Make scene2 of /firefox/new more resilient to unknown platforms

alexgibson:
## Description
- Adds a fallback `href` for the direct download link on scene2 that points to `/firefox/all/`, should the JS fail to update the direct download link.
- Only trigger the auto-download click if a supported visible platform link is detected.
- If a user lands directly on scene 2 with JS disabled, show all the download buttons and hide the (JS dependent) direct download link.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1257796

## Testing
- Test both the old and new scene2 templates with JS disabled.
- Test landing directly on scene2 with an unknown user agent (set `site.platform` to `other`). The page should no longer go into an infinite redirect loop by clicking the empty direct download link.

**Note:** I've only updated the `href` fallback link on the new template design, since this is en-US only and means we don't need to incur a string change. Hopefully soon we can start localization and migrate locales to a single design.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/new/redesign/scene2.html
+++ b/bedrock/firefox/templates/firefox/new/redesign/scene2.html
@@ -32,7 +32,11 @@
           </div>
 
           <p class="help">
-            {{_('Your Firefox should begin downloading automatically. If not, <a id="%s" href="">click here</a>.')|format('direct-download-link')}}
+            {# fallback_url is replaced by the platform download link via JS, but if
+               something fails the user should still get a link to a working download path. #}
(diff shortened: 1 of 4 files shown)


alexgibson:
@jpetto - not urgent, but would appreciate your review / input on these changes. They are edge cases, but i think worthwhile?

jpetto:
Tested old and new scene 2 templates with JS disabled and platform set to `other` and all is working as expected.

Nice structural improvements! 💪

PR #4143, opened 2016-05-20
[fix bug 1264749] Deprecate CSS support for IE 8.

jpetto:
## Description

Remove almost all uses of `extrahead` and serve `oldIE.less` to IE 8 and below.

## Bugzilla link

[bug 1264749](https://bugzilla.mozilla.org/show_bug.cgi?id=1264749)

## Testing

Check important pages first in IE 7 & 8. Consider spot-checking minor pages as well. I tested *most* pages changed in IE 8 and content was at least readable.

The following URLs have not been updated with this change:

- `/`
- `/new/`
- `/contact/spaces/` (and child pages)

The home and `/new/` pages have been left alone due to already working with IE 8 and being the most important. `/contact/spaces/` pages should require a specific PR, as the work to deprecate first-class CSS support is a bit more involved/tied to JS.

The following URLs rely on first-class CSS to hide/show content based on browser, which results in confusing/inappropriate content (usually extra buttons and/or copy) displayed for old IEs:

- `/firefox/sync/`
- `/firefox/desktop/`
- `/firefox/products/`
- `/firefox/hello/`
- `/firefox/ios/`
- `/firefox/private-browsing/`
- `/about/partnerships/`
- `/plugincheck/`
- `/contact/spaces/`

We should discuss (here or in the bug) if it's best to put no more effort into old IEs for this page, or to do the minimum required to hide the inappropriate content.

This branch is available at [https://bedrock-demo-jpetto.us-west.moz.works/](https://bedrock-demo-jpetto.us-west.moz.works/).

## Checklist


jpetto:
@stephendonner - Would you be able to help test these changes against IE 8?

alexgibson:
As part of this change, should we also relegate IE8 to get the (non-interactive) basic version of the contact pages?

https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templates/mozorg/contact/contact-base.html#L11-L25

alexgibson:
Hmm, on second thought maybe we should leave the above for another bug, since this change is just mainly about CSS and not JS?

alexgibson:
Tested home, /firefox/new/, /about, /mission, /about/manifesto on https://bedrock-demo-jpetto.us-west.moz.works/ in my IE8 VM and all look good 👍

alexgibson:
On the manifesto page I'm seeing some cropping on the faces (all browsers)

manifesto

alexgibson:
Open question:

Is there any reason why we still need to serve the full CSS using `{% block extrahead %}` to IE8 users and below on pages such as `/about` and `/about/mission`. Both pages look pretty poor on those browsers anyway, and I might even argue they equally as readable using `oldIE.less`?

Are there any other pages where we could re-assess our full-CSS coverage perhaps? https://github.com/mozilla/bedrock/search?utf8=%E2%9C%93&q=%7B%25+block+extrahead+%25%7D

alexgibson:
One minor regression on the manifesto page, but otherwise this looks good and fairly safe I think.

I do wonder if we could cull support for some pages as part of this change though - there still seem to be quite a few pages using `{% block extrahead %}`, some of which we could probably change to `{% block page_css %}`. Thoughts?

jpetto:
> Is there any reason why we still need to serve the full CSS using {% block extrahead %} to IE8 users on pages such as /about and /about/mission. Both pages look pretty poor on those browsers anyway, and I might even argue they equally as readable using oldIE.less?

This is a good question, and one that I thought about while working on this PR. I'm not sure it's best to actively force only the oldIE CSS to pages that currently render well in older IEs. (Both `/about/` and `/about/mission/` look pretty solid in IE 8.)

In my last commit, I did force `/about/manifesto` to use oldIE, as there was a visual issue with the CSS loaded in `extrahead` (specifically the faces sitting on top of header text). No extra energy spent/code written trying to support old IEs.

screen shot 2016-05-26 at 11 38 53 am

What do you think? Should we actively apply only `oldIE.less` to pages that currently look pretty good in older IEs? Or should we only deprecate them when visual/CSS changes are required?

craigcook:
There's an argument to be made for consistency while browsing around the site. If someone is using IE8 and finds that some pages are really plain and others are all fancy, the overall experience might feel more "broken" than if the whole site was consistently crappy.

alexgibson:
I think with the exception of a couple of pages (home page, download page?), we should probably try to be strict and say deprecation **means** deprecation? Some pages may currently look ok, but if we leave those there's a tendency to say we go on supporting any minor changes that those pages may incur. Deprecating them now may also make changing our JS support for crappy browsers easier in the near future?

I'd be happy to be convinced to keep supporting the pages if others think making them look worse is not a good idea though.

jpetto:
After mulling it over, I'm on board with the "deprecation means deprecation" argument, while leaving the home and download pages looking as-is/pretty good in IE 8. Will update the PR.

bedrock/firefox/templates/firefox/channel.html
```diff
@@ -24,8 +24,7 @@
 
 {% block body_id %}channel{% endblock %}
 
-{% block site_css %}{% endblock %}
-{% block extrahead %}
+{% block page_css %}
```
alexgibson: Hmm, I wonder if we should leave this particular page as-is? Because it still runs the page JS, it makes accessing the download buttons a bit more tricky. It's still possible, but you have to know to click on the respective logo etc. Perhaps we need to class this one similar to the `/contact/` pages, as it is closely tied to the JS still?

alexgibson:
There's a notification at the top of the /firefox/products page that gets shown by default when the page has no specific CSS applied. I wonder if instead of adding in code to hide this in older browsers, we could alter it slightly so it only ever gets shown in supported browsers? (maybe using `.hidden` by default?)

products-notification

alexgibson:
Something a little funky going on with the defaut `em` sizing for headings on the research pages:

research

alexgibson:
What we should do with pages that rely on first-class CSS to hide/show content is a good question. Many of these have been pretty broken looking in IE 6 / 7 for some time, so I'm open to what we do with them here for IE8 (if anything). A few options spring to mind:

1. We do nothing and don't worry about them (as we have been doing for IE 6 & 7)
2. We file separate bugs for pages we think are worth paying attention to, and re-architect things so conditional content only gets shown on browsers that need said content.
3. We maybe try and change our approach going forward for new pages. If a page needs conditional content, it should always have a class of `.hidden` in the markup by default. This can then be hidden by oldIE.less, and shown conditionally by page JS.

I think doing 3 definitely makes sense (?)

I'm open to doing either 1, or 2. depending on what the overall concusses is.

alexgibson:
Overall I think this change is looking good - I can't see any major issues with the code itself, just some open questions for discussion I think we need to decide on before moving forward.

jpetto:
Yeah, between the pages I listed above, `/firefox/channel/`, and some other pages (like `/research/` with the huge font (and it's not the only place that happens)), the rubber is really hitting the road in regards to deciding how much energy we put in to old browsers.

`/firefox/channel/` works, it's just a bad UI. Same goes for the huge font. (I'm not saying the bad UI should definitely be shipped, but you could make the argument.) The pages listed above with misleading copy/buttons displayed are probably the worst offenders, but even those display the primary content and allow a user to download Firefox.

It seems we have two decisions to make:

1. How should we handle the pages mentioned above right now? Leave them be until we get enough complaints? Put in hours of dev time to get them good enough in IE 8? Revert the `extrahead` removal on them for now and file bugs to fix them for IE 8- when there's time?

2. How should we build pages in the future that have CSS & JS tightly coupled (like `/firefox/products/`)? Do we need a standard way of building these that keeps IE 8- in mind? Do we continue to test in IE 8-? (Does that defeat the purpose of "deprecating IE 8 support"?)

alexgibson:
> How should we handle the pages mentioned above right now? Leave them be until we get enough complaints? Put in hours of dev time to get them good enough in IE 8? Revert the extrahead removal on them for now and file bugs to fix them for IE 8- when there's time?

Do you think we'll get complaints? Many of the pages have looked bad in IE6 & 7 for some time and we don't really seem to get complaints.. This is just upping the counter a little?

> How should we build pages in the future that have CSS & JS tightly coupled (like /firefox/products/)? Do we need a standard way of building these that keeps IE 8- in mind? Do we continue to test in IE 8-? (Does that defeat the purpose of "deprecating IE 8 support"?)

I made some suggestions above as to how we might do this. thoughts?

As a side note, If you feel strongly enough that we're not making the right choice here, and would rather go on supporting these pages in IE8 because it's not worth the effort of deprecating them (and making them look crappy), then I'm happy to do that too?

jpetto:
Weird, I didn't see your comment with the 1, 2, 3 list until just now. Way to go, Github!

Anyway, I totally agree with using `.hidden` going forward. That should make things easy enough.

I can take maybe another 15 minutes to try and fix the huge font issue. I bet I can figure it out in a short amount of time.

As for the other pages (those I listed in the initial comment and `/firefox/channel/`), I'm in favor of leaving them as they are in this PR. Our main concern with users of old, old browsers is that they can download Firefox. I don't think it's worth the time to re-work how content is displayed/hidden on existing pages (at least not while there's more important work to be done).

I'm also in favor of leaving `/contact/spaces/*` alone. Those pages look pretty good in old IE currently, and I don't see the effort needed to deprecate them to be worth it. The next time those pages are in for a serious refresh, we can drop old IE support.

> Do you think we'll get complaints? Many of the pages have looked bad in IE6 & 7 for some time and we don't really seem to get complaints.. This is just upping the counter a little?

We might? The `extrahead` removal does impact IE 6 and 7, meaning some pages will look worse in those browsers after this PR lands. However, the numbers on those browsers are even smaller than IE 8, and they can still download Firefox and read the primary content, so *hopefully* it's not a big deal.

> As a side note, If you feel strongly enough that we're not making the right choice here, and would rather go on supporting these pages in IE8 because it's not worth the effort of deprecating them (and making them look crappy), then I'm happy to do that too?

It's really tough to say. Due to `/`, `/firefox/new/`, and `/contact/spaces/`, we're forced into some level of compromise. I think the reason to push for *more* deprecation is to set expectations going forward, even though it does mean a little more work up front (which is already mostly done). I'm for doing the bigger change at this time.

alexgibson:
r+ from me, care to squash?

alexgibson:
Before we merge this, I thought perhaps @craigcook may have some thoughts on the discussion here when he has time? (else we'll merge as-is).

jpetto:
Squashed, but let's see if @craigcook has anything to add prior to merge.

craigcook:
I've finally spent some time on this and indeed some pages are pretty well broken in IE8, but they're just as broken as they've been in IE6 and 7 for a long time so it's hardly newsworthy. Overall the content is still readable and accessible and if anything the problems we see are because oldIE.less is doing *too much*. Rather than invest time into making tweaks to support dead browsers I think we can let it stand as-is. r+ 🎱 

Once we really kick off the redesign work that will replace sandstone we can build in a better strategy for oldIE. We should consider not serving any JavaScript to IE 6/7/8, for example.

PR #4144, opened 2016-05-23
Add functional tests for /contribute tasks

alexgibson:
## Description
- Adds functional tests for the new `/contribute` task pages.
- Adds task pages to parameterized newsletter tests.

Note: I decided against clicking the actual task CTA buttons that open new tabs, since it would introduce a dependency on loading third-party sites in our tests. I think asserting that our own pages look/function as expected is probably enough here. 

It would have been nice to test the "thank you" message after switching tabs, but I think this would likely prove to be more trouble in the long run that it is beneficial. The "thank you" message is only a piece of text, and not something I would class as critical to the page (i.e. even if it broke, people could still complete the tasks as normal).

## Bugzilla link
- N/A

## Testing
- I had to add a new selector for one of the firefox-mobile task page as a test hook. As such, this particular tests needs to be run locally to pass.

## Checklist
- [ ] Requires l10n changes.


alexgibson:
I need to rebase this after #4154 merged - will update shortly.

alexgibson:
Ok updated based on latest master and should be ready for review again.

jpetto:
Code looks good & tests are passing locally. 🔔 r+

PR #4145, opened 2016-05-23
Disable middleware using the Django way

pmac:
https://docs.djangoproject.com/en/1.7/topics/http/middleware/#marking-middleware-as-unused


pmac:
Just noticed this works in 1.7+. We should do this going forward.

@jgmize r?

jgmize:
thanks @pmac

PR #4149, opened 2016-05-23
[bug 1266576] Update dataLayer push to add newClickHref on link click…

JamesDLorence:
## Description
Adds updateDataLayerPush function that is called during the core dataLayer init so when a link click object is pushed to the dataLayer a new newClickHref property with the link's href value is added to the object. If the link's host matches the page's, then the host and locale are stripped before setting the newClickHref property.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1266576

### DIFF
--- a/media/js/base/core-datalayer.js
+++ b/media/js/base/core-datalayer.js
@@ -25,6 +25,37 @@ if (typeof Mozilla == 'undefined') {
         return pageId ? pageId : pathName.replace(/\/[^\/]*/, '');
     };
 
+    /** Monkey patch for dataLayer.push
+    *   Adds href stripped of locale to link click objects when pushed to the dataLayer,
+    *   also removes protocol and host if same as parent page from href.
+    */
+    Analytics.updateDataLayerPush = function() {
+        var dataLayer = window.dataLayer = window.dataLayer || [];
+
+        dataLayer.defaultPush = dataLayer.push;
+        dataLayer.push = function() {
+            for (var i = 0; i < arguments.length; i++) {
+                if (arguments[i].event === 'gtm.linkClick') {
+                    var element = arguments[i]['gtm.element'];
+                    var href = element.href;
+
+                    if (element.hostname === document.location.hostname) {
+                        var path = href.split(element.host)[1];
+                        var locale = path.split('/')[1];
+
+                        arguments[i].newClickHref = path.replace('/' + locale, '');
+                    } else {
+                        arguments[i].newClickHref = href;
+                    }
+
+                    dataLayer.defaultPush(arguments[i]);
+                } else {
+                    dataLayer.defaultPush(arguments[i]);
+                }
+            }
+        };
+    };
+
     Mozilla.Analytics = Analytics;
 
 })();
--- a/tests/unit/spec/base/core-datalayer.js
+++ b/tests/unit/spec/base/core-datalayer.js
@@ -7,7 +7,7 @@
 
 describe('core-datalayer.js', function() {
 
-    describe('PageId', function(){
+    describe('getPageId', function(){
         var html = document.documentElement;
 
         afterEach(function() {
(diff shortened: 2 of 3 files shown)


alexgibson:
@JamesDLorence - looks like you have some ESLint errors that need fixing here (see the CircleCI build details).

I'd also like to arrange a quick vidyo session to run through this change to better understand how it is working. Will send you an email, thanks again!

tests/unit/spec/base/core-datalayer.js
```diff
@@ -25,6 +25,60 @@ describe('core-datalayer.js', function() {
...
+    describe('updateDataLayerPush', function() {
+
+        beforeAll(function() {
+            Mozilla.Analytics.updateDataLayerPush();
+        });
+
+        beforeEach(function() {
+           var link = '<a id="link" href="https://www.mozilla.org/en-US/firefox/new/">';
+            $(link).appendTo('body');
+            window.linkElement = $('#link')[0];
```
alexgibson: Rather than use a global variable here it could be scoped inside the `updateDataLayerPush` block above;

tests/unit/spec/base/core-datalayer.js
```diff
@@ -25,6 +25,60 @@ describe('core-datalayer.js', function() {
...
+        });
+
+        afterEach(function() {
+            $('body').remove('#link');
```
alexgibson: Should we also be clearing `window.dataLayer` here after adding things to it in each test?

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,37 @@ if (typeof Mozilla == 'undefined') {
...
+                    var element = arguments[i]['gtm.element'];
+                    var href = element.href;
+
+                    if (element.hostname === document.location.hostname) {
+                        var path = href.split(element.host)[1];
+                        var locale = path.split('/')[1];
+
+                        arguments[i].newClickHref = path.replace('/' + locale, '');
+                    }
+                    else {
```
alexgibson: Nit: `else` should be on the same line as the closing if `}` bracket above.

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,37 @@ if (typeof Mozilla == 'undefined') {
...
+                        arguments[i].newClickHref = href;
+                    }
+
+                    dataLayer.defaultPush(arguments[i]);
+                }
+                else {
```
alexgibson: Nit: same here, `else` on the line above.

tests/unit/spec/base/core-datalayer.js
JamesDLorence: Are you talking about the window.linkElement declaration? I want linkElement to be deleted and re-instantiated after every test though as I alter it during one of the tests and I want each new test to start with the same link.

tests/unit/spec/base/core-datalayer.js
JamesDLorence: Since multiple pushes happen to the dataLayer on the page, I went with not deleting and re-instantiating it every time. Would you prefer it if I did?

tests/unit/spec/base/core-datalayer.js
alexgibson: Yes I'm talking about `linkElement`. You can do all you want above without the variable being in global scope.

tests/unit/spec/base/core-datalayer.js
alexgibson: We should clean up after each test/suite. Say we have a test later that expects `dataLayer.length` to be a certain value. These (unrelated) tests would cause unexpected results.

tests/unit/spec/base/core-datalayer.js
JamesDLorence: I was getting an error that `linkElement` was not defined when trying to run the tests with it not in the global scope and had `var linkElement = $('#link')[0];`

tests/unit/spec/base/core-datalayer.js
JamesDLorence: That does make sense. I'll make the change.

tests/unit/spec/base/core-datalayer.js
alexgibson: What I mean is put the variable inside the function scope of `updateDataLayerPush`:

```javascript
describe('updateDataLayerPush', function() {

    var linkElement;

    beforeEach(function() {
        var link = '<a id="link" href="https://www.mozilla.org/en-US/firefox/new/">';
        $(link).appendTo('body');
        linkElement = $('#link')[0];
    });
```

alexgibson:
Thanks @JamesDLorence r+

PR #4151, opened 2016-05-24
Upgrade django pipeline

alexgibson:
## Description
- Upgrades django-pipeline so we can disable automatic collection of static assets in local dev mode.
- Adds a new default gulp task, which copies assets from `/media` to `/static` as and when they change (this task copies everything in `/media` over when first run, and incrementally from then on).

This should hopefully give local bedrock development a nice speed boost 👍. Currently django-pipeline copies everything from `/media` to `/static` on each page load in dev mode (which is well over 100mb of files). This PR stops that from happening, and Gulp now handles copying over `/static` assets in a more efficient way. We could (and probably should) take this even further in the longer term, but this is hopefully a quick win.

## Bugzilla link
N/A

## Testing
- Once you pull the branch, update requirements for both bedrock and npm.
- Start bedrock locally, then just run `gulp`.


gulpfile.js
```diff
@@ -15,16 +16,29 @@ var lintPaths = [
     'gulpfile.js'
 ];
 
-gulp.task('test', function(done) {
+gulp.task('media:watch', function () {
+    return gulp.src('./media/**/*')
+        .pipe(watch('./media/**/*', {
```
pmac: Does this auto-compile .less files?

alexgibson: Not currently. The .less files are still being compiled by django-pipeline. We've only disabled the asset collecting. We can do the less compilation here too, if you can find a way to turn it off? either way, this is still a considerable time-saver as-is.

alexgibson:
Just a thought - will this change break integration tests that run in the pipeline via the docker container? Not sure if we would need to make sure assets are copied over there as they will no longer be automatically?

jgmize:
> Just a thought - will this change break integration tests that run in the pipeline via the docker container? Not sure if we would need to make sure assets are copied over there as they will no longer be automatically?

The integration tests are run against the bedrock_code image, which [runs collectstatic as part of the image build process](https://github.com/mozilla/bedrock/blob/5af8b70cf61a19473b216f4d98ae4ae402cf981d/docker/dockerfiles/bedrock_code#L7) so I don't think that should be an issue.

gulpfile.js
```diff
@@ -15,16 +17,38 @@ var lintPaths = [
     'gulpfile.js'
 ];
 
-gulp.task('test', function(done) {
+gulp.task('serve:backend', function () {
+    process.env.PYTHONUNBUFFERED = 1;
+    process.env.PYTHONDONTWRITEBITECODE = 1;
+    spawn('python', ['manage.py', 'runserver'], {
```
jgmize: By default runserver binds to localhost, so this doesn't work in a docker container. I got this working by adding a `0.0.0.0:8000` to the args list, but ideally it that could be a default overridden by an environment variable

jgmize:
r+wc, nice work @alexgibson and @pmac!

gulpfile.js
pmac: Ah yes! Good catch. I'll fix that.

pmac: @jgmize can you think of a reason to **not** have the default be `"0.0.0.0:$PORT"`, with a default value for the `PORT` env var of `8000`? That's what I'm leaning toward, but if we need to default to 127.0.0.1 then I can work that out as well.

jgmize: That's what I was thinking as well, thanks.

pmac:
runserver command updated to use `0.0.0.0` and a configurable port, defaulting to `8000`. I punted on the question of also running `collectstatic`. We can deal with that later when/if it becomes an issue.

PR #4153, opened 2016-05-24
[fix bug 1265912] 2015 transparency report

craigcook:
## Description
Adds a transparency section at /about/policy/transparency/ with an evergreen landing page and an initial report for 2015. In the future we can expect to publish a new report every six months following the same template with only minimal changes to the landing page (just adding/updating links).

Opening this PR in the private repo for initial review. Once this is r+ed and we're given the go-ahead to publish we can open an equivalent PR in the public repo to merge and push.

I've wrapped all the strings so this technically *could* be localized, but because it's legal/policy content I don't think we should actually do so.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1265912

## Testing
I haven't included any functional tests but we should probably do that in a follow-up in the near future (which might be a good project for me to learn how to write tests).

## Checklist
- [ ] Requires l10n changes.


bedrock/mozorg/templates/mozorg/about/policy/transparency/index.html
```diff
@@ -0,0 +1,252 @@
...
+        <p>
+          {{ _('Transparency is a key part of how Mozilla approaches user trust.') }}
+          {{ _('As an open source project that relies on open development, we build transparency into the way we write our code.') }}
+          {{ _('Additionally, our product documentation and notices describe how our products work and how we handle user data.') }}
+        </p>
+
+        <p>
+          {{ _('With this transparency in mind, we intend to publish bi­-annual transparency reports that help provide additional transparency to government disclosures and takedown requests.') }}
+        </p>
+        </p>
```
jpetto: Extra closing `</p>` tag.

bedrock/mozorg/templates/mozorg/about/policy/transparency/index.html
```diff
@@ -0,0 +1,252 @@
...
+          </dd>
+
+          <dt id="dfn-counter-notice">{{ _('Counter Notice') }}</dt>
+          <dd>
+            <p>{{ _('Documentation that meets the counter notification requirements <a href="%s">set forth here</a> in response to a <a href="%s">Takedown Notice</a>.')|format(url('legal.report-infringement'), url('mozorg.about.policy.transparency.index') + '#dfn-takedown-notice') }}</p>
+          </dd>
+
+          <dt id="dfn-threat-indicator">{{ _('Cybersecurity Threat Indicator') }}</dt>
+          <dd>
+            <p>{{ _('Pieces of information about a threat to a computer network or system, such as a vulnerability, piece of malicious code, or the IP address of an attacker. This definition is based on the Cybersecurity Information Sharing Act of 2015 (CISA); the full definition is at <a href="%s">6 U.S.C. § 1501(6)</a>.')|format('http://uscode.house.gov/view.xhtml?req=%28title:6%20section:1501%20edition:prelim%29%20OR%20%28granuleid:USC-prelim-title6-section1501%29&amp;f=treesort&amp;edition=prelim&amp;num=0&amp;jumpTo=true') }}</p>
```
jpetto: The external URL here is a 404 for me...

bedrock/mozorg/templates/mozorg/about/policy/transparency/jan-dec-2015.html
```diff
@@ -0,0 +1,247 @@
...
+            <th scope="col">{{ _('Data Produced') }}</th>
+          </tr>
+        </thead>
+        <tbody>
+          <tr>
+            <th scope="row">
+              <a href="{{ url('mozorg.about.policy.transparency.index') + '#dfn-search-warrant' }}">{{ _('Search Warrants') }}</a>
+            </th>
+            <td>0</td>
+            <td>N/A</td>
```
jpetto: I know this page is (likely) not being localized, but might be nice to wrap these `0` and `N/A` strings just in case?

bedrock/mozorg/templates/mozorg/about/policy/transparency/index.html
```diff
@@ -0,0 +1,252 @@
...
+{# Hide previous reports nav until we have a previous report to link to. Maybe move this to an include...
+      <nav class="content nav-previous">
+        <h3>{{ _('Previous Reports') }}</h3>
+        <ul>
+          <li><a href="#">January-December 2015</a></li>
+          <li><a href="#">January-June 2016</a></li>
+          <li><a href="#">August-December 2016</a></li>
+        </ul>
+      </nav>
+#}
```
jpetto: Let's move this to an include and add a comment here with instructions for future use.

Also, `nav` in a `nav`, eh? I've always heard that wasn't really proper (though I'll defer to you).

media/css/mozorg/about-transparency.less
```diff
@@ -0,0 +1,267 @@
...
+}
+
+.accordion-initialized .section-title {
+    margin-bottom: 0;
+}
+
+#report-sections.accordion-initialized [data-accordion-role='tab'] {
+    .border-box;
+    overflow: visible;
+    padding: 0 20px 0 40px;
```
jpetto: Use `@gridGutterWidth`? Not sure how much sticking to the sandstone grid system really matters anymore...

media/css/mozorg/about-transparency.less
```diff
@@ -0,0 +1,267 @@
...
+
+.section-extra .section-title {
+    .font-size(32px);
+    margin-bottom: 1em;
+}
+
+.section-defs dt {
+    position: relative;
+}
+.section-defs dt:target:before {
```
jpetto: Yay `:target`!

media/js/mozorg/about-transparency.js
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function($, Mozilla) {
+    'use strict';
+
+    var reportSections = $('#report-sections');
+
+    if (reportSections) {
```
jpetto: I think you want `if (reportSections.length)` here. jQuery will always return an object, even if the selector isn't found on the page.

media/js/mozorg/about-transparency.js
```diff
@@ -0,0 +1,31 @@
...
+        if (typeof matchMedia !== 'undefined') {
+
+            var queryWide = matchMedia('(min-width: 760px)');
+
+            if (queryWide.matches) {
+                Mozilla.Accordion.destroyAccordionById('report-sections');
+            } else {
+                reportSections = new Mozilla.Accordion(reportSections);
```
jpetto: This line is throwing errors and causing odd accordion behavior when squishing/expanding the browser, but it's only because you're assigning the accordion to the same variable you're using to create the accordion. `Mozilla.Accordion` expects an element, which it's getting the first time, but if you expand the contract again, `reportSections` has been converted to an instance of `Mozilla.Accordion`, which causes problems.

Easy fix though - just call `new Mozilla.Accordion(reportSections)` (here and below). No need to assign the result to a variable.

jpetto:
A few small cleanups/fixes, but really close. r+wc 🍌

media/js/mozorg/about-transparency.js
craigcook: Now see, jshint complained about that ("Do not use 'new' for side effects.") but eslint is fine with it. I never know who to trust...

media/js/mozorg/about-transparency.js
craigcook: doh, good catch

media/css/mozorg/about-transparency.less
craigcook: eh, I've lost faith in that variable (and our grid system in general; we don't consistently stick to it). The point of variables is that they can vary, but more often than not we just use this one as an alias for `20px` in places where the value probably shouldn't change anyway. If by some chance we decided to globally change our entire site's gutter width to 30px, I think I'd still want this particular element to be `padding: 0 20px 0 40px;`. The `@gridGutterWidth` variable is so entrenched and overused at this point changing its value would be disastrous. I'd like to phase it out with the redesign.

bedrock/mozorg/templates/mozorg/about/policy/transparency/index.html
craigcook: Yeah, it felt dirty and yet also semantically valid... the whole block is navigation, it just has distinct bits of navigation inside it. I guess it can be a `div` (or maybe an `aside`) with two `nav`s inside.

bedrock/mozorg/templates/mozorg/about/policy/transparency/jan-dec-2015.html
craigcook: I wasn't sure about N/A... it's an abbreviation of English words, but is also a sort of programmatic symbol. Maybe best to wrap it just in case. And 0 is just 0, I don't think we'd translate any other numerals either.

bedrock/mozorg/templates/mozorg/about/policy/transparency/index.html
craigcook: Looks like the `&amp;`s are at fault. It's a clunky mix of encoding. I'll clean it up.

PR #4154, opened 2016-05-24
Migrate page objects to PyPOM

davehunt:
## Description
The bedrock functional tests were a chance to create a new page object model in Python that could become a library shared across our web automation. This became [PyPOM](http://pypom.readthedocs.io/en/latest/), and has now been released as a standalone package. This patch switches to using the new library.

## Testing
All functional tests should continue to pass - I have successfully run them all locally in Firefox.

## Checklist
- [ ] Requires l10n changes.
- [X] Related functional & integration tests passing.


davehunt:
@alexgibson r?

tests/pages/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+def scroll_element_into_view(self, strategy, locator, x=0, y=0):
+    el = self.find_element(strategy, locator)
+    self.selenium.execute_script(
+        'arguments[0].scrollIntoView();'
+        'window.scrollBy(arguments[1], arguments[2]);', el, x, y)
+    return el
+
+WebView.scroll_element_into_view = scroll_element_into_view
```
alexgibson: Just so I understand this, we now override functions in PyPom this way if needed?

tests/pages/base.py
```diff
@@ -5,21 +5,21 @@
...
-    def __init__(self, base_url, selenium, locale='en-US', **kwargs):
-        super(BasePage, self).__init__(base_url, selenium, locale=locale, **kwargs)
+    def __init__(self, selenium, base_url, locale='en-US', **url_kwargs):
+        super(BasePage, self).__init__(selenium, base_url, locale=locale, **url_kwargs)
 
     def wait_for_page_to_load(self):
         super(BasePage, self).wait_for_page_to_load()
-        el = self.find_element((By.TAG_NAME, 'html'))
-        self.wait.until(lambda s: 'loaded' in el.get_attribute('class'))
+        self.wait.until(lambda s: self.is_element_present(
```
alexgibson: Just curious - Is this better performance wise, or is it just to neater? 😄

tests/pages/firefox/base.py
```diff
@@ -2,17 +2,25 @@
...
 
+from pypom import Region
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support import expected_conditions as expected
 
 from pages.base import BasePage
-from pages.page import PageRegion
 
-# this should be enough to scroll elements so they are not beneath the header
-HEADER_OFFSET = -100
```
alexgibson: I quite liked this being in a clearly defined variable with the comment, any reason to lose it be hard-coding the offset in the function itself?

tests/pages/newsletter.py
```diff
@@ -25,62 +25,62 @@ class NewsletterPage(BasePage):
...
 
     @property
     def is_privacy_policy_link_displayed(self):
-        return self.is_element_displayed(self._privacy_policy_link_locator)
+        return self.is_element_displayed(*self._privacy_policy_link_locator)
 
     def click_sign_me_up(self):
-        self.find_element(self._submit_button_locator).click()
+        self.find_element(*self._submit_button_locator).click()
         self.wait.until(expected.visibility_of_element_located(self._thank_you_locator))
```
alexgibson: Should this be `*self._thank_you_locator`?

alexgibson: Ah never mind, I see this is not needed.

alexgibson:
A couple of questions but code-wise this looks good, thanks @davehunt!

It would be great to run this against Chrome as well as Firefox before merging, as some tests are conditionally skipped in Firefox. Do you have time to run this? If not, I can see if I can find some time later.

tests/pages/__init__.py
davehunt: Essentially, yes. This isn't so much an override as it is adding a function to the WebView class, which is inherited by pages and regions. Ultimately we might patch PyPOM to include this functionality.

tests/pages/base.py
davehunt: I noticed a stale element exception occurring when loading subsequent pages via a click. This is most likely because the HTML element is found before the new page load starts, but is then stale by the time we check the attribute. This change resolved my stale element exceptions, but I now realise that it could mean the wait condition is met prematurely, though that's also possible with the current wait.

tests/pages/firefox/base.py
davehunt: We can restore it, though it's now only used in one place which is why I moved it.

davehunt:
> It would be great to run this against Chrome as well as Firefox before merging, as some tests are conditionally skipped in Firefox. Do you have time to run this?

Running them now.

tests/pages/base.py
alexgibson: Ok understood - I guess we'll see if this change has any side effects, but hopefully it will be more robust than before.

tests/pages/__init__.py
alexgibson: Got it, thanks 👍

tests/pages/firefox/base.py
alexgibson: I can imaging this value might need to change in future if/when we update the nav. For this reason I think making it more obvious what the dev needs to change is worthwhile. I would vote to either restore it, or make it more obvious via a comment in the function itself. Either would be fine.

davehunt:
All tests passed in Chrome.

alexgibson:
Thanks @davehunt - just one minor amend and then I think we're good to merge this. r+wc 🐎

tests/pages/firefox/base.py
davehunt: We can actually be smarter about this now and offset by the height of the family navigation element.. I'll update with this change.

davehunt:
@alexgibson let me know if you're happy with this approach - it removes the hard-coded offset entirely.

tests/pages/firefox/base.py
```diff
@@ -12,8 +12,11 @@
 class ScrollElementIntoView(object):
 
     def scroll_element_into_view(self, strategy, locator):
+        # scroll elements so they are not beneath the header
+        page = self.page if isinstance(self, Region) else self
+        offset = {'x': 0, 'y': -page.family_navigation.root.size['height']}
```
alexgibson: Hmm... are we sure this will be consistent across browsers?

alexgibson: The size of the header here also changes on scroll when it becomes sticky :/

davehunt: This allows it to be inconsistent - it will scroll based on the height of the header. If the header is taller in some browsers then we'll scroll further to accommodate it.

alexgibson: Ok, let's see how this goes 😉

alexgibson:
Let's see how the sticky header calculation goes - care to squash?

tests/pages/firefox/base.py
alexgibson: Hmm... one drawback now is scroll_element_into_view assumes the family navigation element is on the page. Could we maybe pass a locator to it instead?

alexgibson:
Actually I just made one more question/comment above re: hard-coding the family navigation locator ^

tests/pages/firefox/base.py
davehunt: Wouldn't it always be on pages that inherit from FirefoxBasePage?

alexgibson: Not every Firefox page has the family navigation (most do, but not all). I'm more thinking in future where we might need this function for some other element (even outside the firefox pages).

davehunt:
@alexgibson I've restored the hard-coded offset for now, with a comment

alexgibson:
r+

PR #4158, opened 2016-05-26
Update documentation for running local server using Gulp

alexgibson:
@pmac r?


pmac:
Thanks @alexgibson. We could also document how to disable these changes and keep using `./manage.py runserver` too, though I really don't know why anyone would. What do you think?

alexgibson:
> We could also document how to disable these changes and keep using ./manage.py runserver too, though I really don't know why anyone would. What do you think?

I think this could come in useful if someone has trouble with Gulp 👍

PR #4160, opened 2016-05-26
Add no gulp runserver docs

pmac:


pmac:
@alexgibson thoughts? I also realized that we've yet to mention `.env` file in the docs :( So maybe I should do that first, or not mention it here...

alexgibson:
> I also realized that we've yet to mention .env file in the docs :( So maybe I should do that first, or not mention it here...

Could the example here use `local.py` for now instead?

jgmize:
> Could the example here use local.py for now instead?

I recommend we update the docs to recommend environment variables (via `.env` files or otherwise) over `local.py`, and that we eventually remove `local.py` support altogether.

alexgibson:
> I recommend we update the docs to recommend environment variables (via .env files or otherwise) over local.py, and that we eventually remove local.py support altogether.

Agreed, if everyone is going to need a `.env` now to run the site locally, I think we might as well encourage the creation of one from the outset when installing the site.

pmac:
I'll take a stab at converting the `local.py` stuff to `.env`.

pmac:
I've added a commit to update the docs to replace `.env` file instructions in place of all mentions of `local.py`.

docs/install.rst
```diff
@@ -81,18 +81,12 @@ activated, so running the tests is as simple as::
 
 .. note::
 
-    If your local tests run fine, but when you submit a pull-request our Jenkins
+    If your local tests run fine, but when you submit a pull-request our CircleCI
     (continuous integration service) instance tells you the tests failed, it could
```
jgmize: We don't have a circleci instance. suggest rewording: "[...] pull request the tests fail in [CircleCI](https://circleci.com/gh/mozilla/bedrock), it could [...]"

docs/install.rst
```diff
@@ -81,18 +81,12 @@ activated, so running the tests is as simple as::
...
-    as possible by doing the following::
-
-        $ JENKINS_HOME=1 ./manage.py test
-
-    This tells Bedrock to use the jenkins settings. This will require you to have a local
-    MySQL database server running and configured correctly, but may help you debug. Alternately
-    you can move your ``settings/local.py`` to a backup, copy ``settings/jenkins.py`` to
-    ``settings/local.py`` and tweak the DB settings yourself to make it work.
+    be due to the difference in settings between what you have in ``.env``
+    and what Jenkins uses: ``.bedrock_demo_env``. You can run tests as close to Circle
```
jgmize: s/Jenkins/CircleCI/

jgmize:
r+wc

PR #4162, opened 2016-05-26
Restore waiting for the seed URL to be loaded in functional tests

davehunt:
The recent switch to PyPOM inadvertently dropped the wait for the seed URL to be loaded. This patch restores it. @alexgibson r?


alexgibson:
r+

PR #4163, opened 2016-05-26
Skip thank you page test due to flakyness in the pipeline

alexgibson:
@davehunt r?

https://ci.us-west.moz.works/job/bedrock_integration_tests_runner/6087/

I think this test is failing due to some kind of slight change in timing. It passes for me locally, and the page still works as expected when manual testing. But for some reason it is now failing in the pipeline. I need some time to try and figure it out and debug the issue, but for now I think we might need to skip this test :/


davehunt:
r+

PR #4166, opened 2016-05-30
Bug 1275626 home concept a

craigcook:
## Description
A new home page concept for testing (Concept B forthcoming). 

This will be English-only during the test and if this becomes the new home page we'll begin l10n at that time. Copy may change before that (and this entire concept may be discarded) so these strings shouldn't be extracted.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1275626

## Testing
Load the page by adding the query string `?v=a` to the URL.

The page needs a bit of cross-browser testing, especially for degradation in older, less capable browsers. IE 6/7/8 should just get basic styling. We're using a wee touch of flexbox so that might need some fallback for IE9.

## Checklist
- [ ] Requires l10n changes.


craigcook:
Lots of commits but we'll do a big squash at the end. Marking WIP and Do Not Merge until then.

bedrock/base/templates/base-new.html
```diff
@@ -0,0 +1,149 @@
+{% from "macros.html" import share_cta with context %}
```
alexgibson: Not a blocker for this PR, but with this change we will now have three base templates. There seems to be a lot of duplication here (particularly all the stuff in the `<head>`). Is it worth filing a follow up bug to consolidate / share some of this between templates?

bedrock/mozorg/templates/mozorg/home/home-a.html
```diff
@@ -0,0 +1,280 @@
...
+{% block content %}
+<main role="main" class="main">
+  <section class="section section-intro" id="intro">
+    <div id="home-slideshow">
+      <div class="slide" id="slide-one"></div>
+      <div class="slide" id="slide-two"></div>
+      <div class="slide" id="slide-three"></div>
+    </div>
+    <div class="content">
+      <h1 class="intro-tagline">
```
alexgibson: For SEO purposes (bleugh), should our logo be the `<h1>` on this page, as opposed to this tagline?

bedrock/mozorg/templates/mozorg/home/home-a.html
```diff
@@ -0,0 +1,280 @@
...
+    <div class="section-lead">
+      <div class="content">
+        <h2>{{ _('A network, hard at work') }}</h2>
+        <p>{{ _('At Mozilla, we bring together educators, builders, advocates and experts who work fearlessly to protect our individual freedoms on the internet.') }}</p>
+      </div>
+    </div>
+
+    <div id="encryption-video-container" class="moz-video-container">
+      <button class="moz-video-button" type="button" aria-controls="encryption-video">{{ _('Play video') }}</button>
+      <video controls="" id="encryption-video" poster="{{ static('img/home/new/encryption/encryption-video-poster.jpg') }}" preload="metadata">
```
alexgibson: If we use out video helper here for the markup, old crappy browsers will get fallback links to watch the video for free?

bedrock/mozorg/templates/mozorg/home/home-a.html
```diff
@@ -0,0 +1,280 @@
...
+        <source id="webm" src="https://d24kjznqej0s8a.cloudfront.net/2016/encryption_campaign/Mozilla_Encrypt_Beat2_landingpage.webm" type="video/webm">
```
alexgibson: `https://d24kjznqej0s8a.cloudfront.net` seems like a bit of a weird URL. Any reason this isn't on our regular video CDN?

bedrock/mozorg/templates/mozorg/home/home-a.html
```diff
@@ -0,0 +1,280 @@
...
+      </div>
+    </div>
+
+    <div id="aframe-wrapper"></div>
+
+    <div class="section-body">
+      <div class="content">
+        <h3>{{ _('Introducing A-Frame') }}</h3>
+        <p>{{ _('Check out this ground-breaking, open-source framework for easily creating web-based virtual reality experiences.') }}</p>
+        <p class="section-cta"><a href="#">{{ _('Learn more') }}</a></p>
```
alexgibson: Missing `href`?

bedrock/mozorg/templates/mozorg/home/home-a.html
```diff
@@ -0,0 +1,280 @@
...
+    </div>
+
+    <div class="footer-column">
+      <ul class="links-join">
+        <li><a href="{{ url('mozorg.contact.contact-landing') }}" data-link-type="footer" data-link-name="Contact Us">{{ _('Contact Us') }}</a></li>
+        <li class="wrap"><a href="{{ donate_url('footer') }}" class="donate" data-link-type="footer" data-link-name="Donate">{{ _('Donate') }}</a></li>
+        <li class="clear"><a href="https://wiki.mozilla.org/Webdev/GetInvolved/mozilla.org" data-link-type="footer" data-link-name="Contribute to this site">{{ _('Contribute to this site') }}</a></li>
+      </ul>
+
+      {% if fx_footer_links %}
```
alexgibson: This seems redundant here since we're not including any platform footer links. Can it be removed?

bedrock/mozorg/templates/mozorg/home/home-a.html
```diff
@@ -0,0 +1,280 @@
...
+      {{ email_newsletter_form(button_class='light') }}
+    {% endif %}
+    </div>
+  </aside>
+
+</main>
+{% endblock %}
+
+{% block site_footer %}
+<footer class="section section-footer" id="footer">
```
alexgibson: Would it actually make sense for this to be in the new base template? 

A little part of me dies a little each time we add another variation of our page footer (in a similar way to each time we add yet another Firefox logo to the repo 😉)

bedrock/base/templates/base-new.html
craigcook: New pages in the upcoming redesign should be totally divorced from sandstone and for now creating a new base template was a start down that path (and I wanted to get rid of the outer-wrapper and wrapper divs). We need to formulate a better plan for the redesign. This base template should be considered temporary, as hacky as it is, and the "new" in the filename is totally a joke :)

bedrock/mozorg/templates/mozorg/home/includes/survey.html
```diff
@@ -0,0 +1,50 @@
...
+
+  /* hide survey for smaller screens */
+  @media only screen and (max-width: 760px) {
+    #survey-message {
+      display: none;
+    }
+  }
+</style>
+
+<script>
```
alexgibson: I know this totally works, but we currently have some CSP hopes and dreams for bedrock. We should probably try and avoid using inline `<style>` and `<script>` tags as policy? (even though it might be a pain for something so simple).

media/css/mozorg/home/home-a.less
```diff
@@ -0,0 +1,1097 @@
...
+
+        img {
+            height: 30px;
+            margin-left: 60px;
+        }
+    }
+}
+
+// Global main nav
+.masthead-nav-main {
```
alexgibson: If this new nav could become used on more than one page in the future (even in concept b?), is it worth being in it's own less file?

media/css/mozorg/home/home-a.less
```diff
@@ -0,0 +1,1097 @@
...
+
+    .intro-tagline {
+        .border-box;
+        .font-size(32px);
+        font-style: italic;
+        line-height: 1.25;
+        margin: 0 auto;
+        padding: 100px 20px;
+        text-align: center;
+        text-shadow: 0 0 10px rgba(0, 0, 0, .01), 0 0 5px rgba(0, 0, 0, .025), 0 1px 1px rgba(0, 0, 0, .5);
```
alexgibson: Nit: worth splitting these declarations onto multiple lines for readability?

media/css/mozorg/home/home-a.less
```diff
@@ -0,0 +1,1097 @@
...
+            padding-top: 135px;
+            width: 250px;
+        }
+    }
+}
+
+.event-horizon {
+    background-color: #0150ab;
+    background: url('/media/img/firefox/new/redesign/stars.png') center 80px no-repeat,
+        linear-gradient(to bottom, rgb(30, 68, 136) 0%, rgb(19, 71, 146) 22%, rgb(2, 76, 161) 44%, rgb(0, 85, 184) 73%, rgb(4, 89, 190) 78%, rgb(17, 94, 199) 86%, rgb(51, 105, 219) 100%);
```
alexgibson: Nit: fix indentation so it lines up underneath `url()`?

bedrock/base/templates/base-new.html
alexgibson: Yeah, I'm not adverse to a new base template at all - I was more thinking of using shared includes for the bits that don't really change, it avoid having to update everything 3x.

alexgibson: As I said though, not a blocker for this PR - just something that might be worth considering, even if only temporary.

media/css/mozorg/home/home-a.less
```diff
@@ -0,0 +1,1097 @@
...
+            background-size: 120px 125px;
+            padding-top: 135px;
+            width: 250px;
+        }
+    }
+}
+
+.event-horizon {
+    background-color: #0150ab;
+    background: url('/media/img/firefox/new/redesign/stars.png') center 80px no-repeat,
```
alexgibson: This might be fine to leave in for now, but the design on /new is changing possibly as soon as the next sprint. We may just need to remember to move/rename the files for this page.

media/css/mozorg/home/home-a.less
```diff
@@ -0,0 +1,1097 @@
...
+    ul {
+        margin-bottom: 15px;
+        line-height: 1.5;
+    }
+
+    ul {
+        overflow: hidden;
+
+        &.fx-footer-links {
+            li {
```
alexgibson: Nit: unnecessary nesting?

media/css/mozorg/home/home-a.less
```diff
@@ -0,0 +1,1097 @@
...
+// Variables and Mixins
+
+@color-mozred-bright: saturate(lighten(@color-mozred, 5%), 10%);
+
+.invert-text() {
+    color: #fff;
+
+    a:link,
+    a:visited,
+    a:hover,
```
alexgibson: Does this need some form of distinct hover/focus styling by default?

media/js/mozorg/home/home-a.js
```diff
@@ -0,0 +1,174 @@
...
+    function isCanvasSupported() {
+        var elem = document.createElement('canvas');
+        return !!(elem.getContext && elem.getContext('2d'));
+    }
+
+    if (isCanvasSupported()) {
+        $aframeIframe = $('<iframe>').attr({
+            frameborder: 0,
+            id: 'aframe-iframe',
+            src: 'https://aframe.io/aframe/examples/animation-aframe-logo/?ui=false'
```
alexgibson: This is probably fine but worth asking - is `aframe.io` set up to handle the amount of traffic embedding it on the mozorg homepage might generate?

alexgibson:
By no means a blocker, but I seem to get a 1px high white line at the bottom of the page, after the main footer:

footer

alexgibson:
Looks like the a-frame animation adds over 1MB of JS to the page weight. It's good that it's not blocking per say, but it still adds to people's data bills 💸 . It does look nice, but I'm not sure if it does anything overly useful to explain what a-frame is to the user? Could some form of static image with some CSS animation be a nice (future) alternative?

alexgibson:
Just another general observation on a smaller screen - the side bar feels pretty huge, while the main content feels a little squished.

small-screen

Could the break point for the thinner side nav come a bit sooner maybe?

alexgibson:
Newsletter thank you message could run full width of the container?

newsletter

Could also make the spinner white using `data-spinner-color="#fff"`?

alexgibson:
With JS disabled some of the platform download links are off screen an inaccessible. I imagine this might also be true if the platform is unknown?

no-js

bedrock/base/templates/base-new.html
```diff
@@ -0,0 +1,149 @@
...
+
+    {{ l10n_css() }}
+
+    <!--[if lte IE 8]>
+      {# Only needed for IE before v9 #}
+      <script src="{{ static('js/libs/html5shiv.js') }}"></script>
+    <![endif]-->
+
+    <!--[if lte IE 8]>
+      {# Basic styles, only for IE7 and lower #}
```
alexgibson: Are we missing the oldIE.less stylesheet here?

bedrock/mozorg/templates/mozorg/home/home-a.html
```diff
@@ -0,0 +1,280 @@
...
+{% block body_class %}home-variant home-variant-a{% endblock %}
+
+{% block extra_meta %}
+<!-- validates bing webmaster tools -->
+<meta name="msvalidate.01" content="B7B177115A634927D608514DA17B2574">
+<!-- YouTube Verification -->
+<meta name="google-site-verification" content="U9a6gH32vLIykvntaDToj-ytYhlZ1AfAgVEKstixQIE">
+{% endblock %}
+
+{% block extrahead %}
```
alexgibson: If we're planning on serving basic styling to IE 6/7/8, should this be `{% block page_css %}`? Currently all browsers seem to be getting the full page CSS. Not quite sure if that's expected?

media/js/mozorg/home/home-a.js
```diff
@@ -0,0 +1,174 @@
...
+            handler: navState('firefox', 'innovation'),
+            offset: 10
+        });
+    }
+
+    // Change the navbar current item to match the section waypoint
+    function navState(current, previous) {
+        return function(direction) {
+            if (direction === 'down') {
+                $navPage.find('#nav-page-primary li').removeClass();
```
alexgibson: Nit: no need to repeat this line of JS in both conditionals?

bedrock/mozorg/templates/mozorg/home/home-a.html
alexgibson: This took me a while to work out - but currently IE 6/7/8 are getting served the page specific CSS, but are not getting any of `pebbles.less`. This means that things like download buttons are unstyled, so the page is showing links for every platform (including things like no-JS buttons, unsupported messaging etc.)

alexgibson:
Oh the joy of web fonts (IE9) D:

screen shot 2016-05-31 at 10 19 19 am

media/css/mozorg/home/home-a.less
```diff
@@ -0,0 +1,1097 @@
...
+            &:focus {
+                background-color: @color-mozred;
+            }
+        }
+    }
+
+    .secondary {
+        .flexbox;
+        margin: 0;
+        text-transform: uppercase;
```
alexgibson: Any reason why these are uppercase? During the recent download button overhaul to make things consistent across mozorg, using uppercase was ruled a big no-go (particularly in relation to download buttons). Maybe this is a UX question for Holly?

alexgibson:
Looks like IE9 needs the matchMedia event listener polyfill included in the page:

    SCRIPT438: Object doesn't support property or method 'addListener' home-a.js, line 156 character 9

It seems quite easy to forget adding this. Should we make it a part of the new base template?

media/js/mozorg/home/home-a.js
```diff
@@ -0,0 +1,174 @@
...
+        });
+    });
+
+    // Innovation section - aframe
+    function isCanvasSupported() {
+        var elem = document.createElement('canvas');
+        return !!(elem.getContext && elem.getContext('2d'));
+    }
+
+    if (isCanvasSupported()) {
```
alexgibson: We may not really care about this, but it looks like IE9 can't do the aframe stuff even though it does support Canvas:

    SCRIPT5009: 'Float32Array' is undefined aframe.js, line 1 character 360

alexgibson: Fwiw, looks like IE10 has some trouble too:

    SCRIPT445: Object doesn't support this action aframe.js, line 59346 character 3
    SCRIPT445: Object doesn't support this action aframe.js, line 59346 character 3
    SCRIPT445: Object doesn't support this action aframe.js, line 59346 character 3
    SCRIPT445: Object doesn't support this action aframe.js, line 59346 character 3

alexgibson: Seems to work fine in Edge \o/

alexgibson:
Nav button height doesn't seems to stretch on my phone:

img_2371

alexgibson:
While this test is running we need to skip the current home page tests to avoid failures in production:

- tests/functional/test_home.py
- tests/functional/test_navigation.py
- tests/functional/test_l10n.py

Note: to skip just add the following line to the top of each individual test:

    @pytest.mark.skipif(reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1275626')

alexgibson:
Looking good for a first pass 👍

bedrock/mozorg/templates/mozorg/home/home-a.html
jpetto: We appropriated the video straight from [https://advocacy.mozilla.org/encrypt/](https://advocacy.mozilla.org/encrypt/), which I think is fine for the test. (Not sure why they didn't use the regular video CDN...)

If we end up using this prototype as *the* homepage, I think moving the video to the conventional location is a good idea.

bedrock/mozorg/templates/mozorg/home/home-a.html
jpetto: D'oh! Will fix.

bedrock/mozorg/templates/mozorg/home/home-a.html
jpetto: We have another prototype coming up next sprint (that will ostensibly use the new base template created in this PR), so I wouldn't mind waiting to see what that footer looks like before refactoring this footer.

Will leave this up to @craigcook though.

(And every time we re-style the newsletter form, and...)

bedrock/mozorg/templates/mozorg/home/includes/survey.html
jpetto: Yeah, security is probably important. 😝 Will refactor into bundles.

media/css/mozorg/home/home-a.less
jpetto: I'd say we can refactor if/when the nav is needed elsewhere. We really don't know if this new concept will fly, so probably better to wait on componetizing.

media/css/mozorg/home/home-a.less
jpetto: Nice catch. At this point, I didn't see a reason to duplicate the files (and the user will get cache benefit if they download), but yeah, if `/new` is changed while this prototype lives, we'll need to move the assets under `/home`.

Will add a comment here to that effect.

media/js/mozorg/home/home-a.js
jpetto: I checked with the aframe team and they told me this page is hosted on GitHub, so we *should* be fine. Also, if the frame `src` fails to load, the user will just continue to see the fallback image.

jpetto:
> Looks like the a-frame animation adds over 1MB of JS to the page weight. It's good that it's not blocking per say, but it still adds to people's data bills 💸 . It does look nice, but I'm not sure if it does anything overly useful to explain what a-frame is to the user? Could some form of static image with some CSS animation be a nice (future) alternative?

Yeah, we discussed this. Given time constraints, embedding was really our only option for meaningful animation. If aframe ends up on the final homepage, the plan is to create a (much lighter) CSS animation.

media/js/mozorg/home/home-a.js
jpetto: Looks like IE 11 can't handle it either. Will add some code to prevent IE < Edge from attempting the aframe.

bedrock/base/templates/base-new.html
craigcook: Hmm... yeah... my original intent was to make a new one unrelated to sandstone, but we could probably use sandstone's oldIE and it should be just fine. I just haven't fired up IE8 to try it yet.

media/css/mozorg/home/home-a.less
craigcook: They were uppercase in the design mockup. I'll check with mham.

bedrock/mozorg/templates/mozorg/home/home-a.html
jpetto: Good call - will update to use the helper.

media/js/mozorg/home/home-a.js
jpetto: Good catch - will fix.

jpetto:
> Looks like IE9 needs the matchMedia event listener polyfill included in the page:
> 
> SCRIPT438: Object doesn't support property or method 'addListener' home-a.js, line 156 character 9
> 
> It seems quite easy to forget adding this. Should we make it a part of the new base template?

I think this is a good suggestion. Will add the `addListener` polyfill to the base template.

bedrock/mozorg/templates/mozorg/home/home-a.html
craigcook: Yeah, it's possible Concept B will have a different footer so let's leave it for now. I also copied the whole thing instead of using the include we already have simply because I wanted a slightly different structure. We'll rethink all this when we work on consolidating a new base template.

media/css/mozorg/home/home-a.less
craigcook: Done.

media/css/mozorg/home/home-a.less
craigcook: if we make this a more global mixin we probably should, maybe at least an underline on hover and focus, but for how it's used on this page at the moment it's really just shorthand for "make this stuff white." A default hover/focus style would likely have to be overridden.

craigcook:
@alexgibson I think we've addressed all the comments and I've squashed the commits. If you're satisfied please merge and tag it to kick off a prod push (minor stuff we could always handle in a follow-up PR). We're supposed to be showing the page in the sprint review at 9:30 Pacific and I'll try to be online early.

craigcook:
Oh, I guess we still need to add the skip tests, but I'm not entirely sure what to do there.

alexgibson:
> Oh, I guess we still need to add the skip tests, but I'm not entirely sure what to do there.

I'll add a commit for this before I merge.

alexgibson:
If the test run is green and things look good on stage, I'll tag a commit a bit later once I get some other work out the way. r+

PR #4167, opened 2016-05-31
[fix bug 1253001] Update partner submission page to use an email instead of salesforce

alexgibson:
## Description
- Update partner submission page to use an email instead of salesforce

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1253001

## Testing
https://bedrock-demo-agibson.us-west.moz.works/en-US/about/partnerships/

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/mozorg/views.py
+++ b/bedrock/mozorg/views.py
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+import jingo
 import json
 import re
 from cgi import escape
@@ -9,6 +10,7 @@ from cgi import escape
 from django.conf import settings
 from django.contrib.staticfiles.finders import find as find_static
 from django.core.context_processors import csrf
+from django.core.mail import EmailMessage
 from django.http import HttpResponseRedirect, Http404
 from django.views.decorators.cache import cache_page
 from django.views.decorators.csrf import csrf_exempt, csrf_protect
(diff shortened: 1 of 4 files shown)


alexgibson:
@pmac r?

Still waiting on confirmation that emails are coming through OK as expected, but I think this should be good for code review. This was a little out of my normal comfort zone (but I didn't think it looked too tricky), so some r would be appreciated :)

bedrock/mozorg/views.py
```diff
@@ -335,26 +340,18 @@ def process_partnership_form(request, template, success_url_name, template_vars=
...
-
-                r = requests.post('https://www.salesforce.com/servlet/'
-                                  'servlet.WebToLead?encoding=UTF-8', data)
-                msg = requests.status_codes._codes.get(r.status_code, ['error'])[0]
-                stat = r.status_code
 
+                subject = PARTNERSHIPS_EMAIL_SUBJECT
+                sender = PARTNERSHIPS_EMAIL_FROM
+                to = PARTNERSHIPS_EMAIL_TO
+                msg = jingo.render_to_string(request, 'mozorg/emails/partnerships.txt', data)
```
pmac: Looks like you should change this variable name. `msg` is used down on line 361 in the JSON value sent back to the client. I think you should make it 'ok' when everything is fine. But probably something like:

```python
try:
    email = EmailMessage(subject, msg, sender, to)
    email.send()
    stat = 200
    msg = 'ok'
except Exception:
    stat = 400
    msg = 'email send error'
```

pmac:
Looks good apart from my one note. r+wc 🎸

alexgibson:
Thanks @pmac! I've updated as suggested. Going to hold off merging until we get confirmation in the bug when the receiving email account is set up.

alexgibson:
Got [confirmation in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1253001#c16) that emails are being received ok, so I think we're good to merge here if all looks good.

PR #4170, opened 2016-06-01
[fix bug 1277252] Add scroll tracking to home prototype A

jpetto:
## Description

Add `dataLayer` call when scrolling down to primary sections on the page.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1277252

## Testing

Make sure `dataLayer` call is being made with appropriate values when scrolling down only.

Make sure no regressions caused by removing `'intro', 'intro'` Waypoint. (This Waypoint was in place to reset the nav highlight, but the `'action', 'intro'` Waypoint takes care of this when scrolling up.)

## Checklist

### DIFF
--- a/media/js/mozorg/home/home-a.js
+++ b/media/js/mozorg/home/home-a.js
@@ -8,10 +8,11 @@
     var $navPage = $('#nav-page');
 
     // waypoints
-    var introWaypoint;
     var actionWaypoint;
     var innovationWaypoint;
     var firefoxWaypoint;
+    var suppressScrollTracking = false;
+    var clickTargetSection;
 
     // intro slideshow
     var $slideshow = $('#home-slideshow');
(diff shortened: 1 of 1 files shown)


media/js/mozorg/home/home-a.js
```diff
@@ -97,38 +97,40 @@
 
     function enableWaypoints() {
         // fire the waypoints for each section, passing classes for the current and previous sections
-        introWaypoint = new Waypoint({
```
alexgibson: If this is no longer being used, can `var introWaypoint` and `introWaypoint.destroy();` be removed too?

media/js/mozorg/home/home-a.js
```diff
@@ -97,38 +97,40 @@
...
     // Change the navbar current item to match the section waypoint
     function navState(current, previous) {
         return function(direction) {
+            var $navPrimary;
+
             $navPage.find('#nav-page-primary li').removeClass();
 
             if (direction === 'down') {
-                $('#nav-primary-' + current).addClass('current');
+                $navPrimary = $('#nav-primary-' + current);
```
alexgibson: Nice optimization 👍

alexgibson:
If I'm at the top of the page and i click on "Firefox" in the nav, are we ok firing three scroll events? The user doesn't really see the middle sections, although the data would suggest that they did given the current code.

media/js/mozorg/home/home-a.js
jpetto: Oh, most definitely! Nice catch.

jpetto:
> If I'm at the top of the page and i click on "Firefox" in the nav, are we ok firing three scroll events? The user doesn't really see the middle sections, although the data would suggest that they did given the current code.

I'm guessing it's okay, as it's not unlike a user fast-scrolling down the page. Will defer to pg though.

jpetto:
pg has requested that tracking should happen on both up and down scrolling, and that clicking on the nav should suppress scroll tracking for any sections in between the current and requested section.

Will update the PR.

jpetto:
Okay, this should now be ready for review part II.

alexgibson:
Seems to work well locally, r+ 🚜

PR #4171, opened 2016-06-02
[fix bug 1277399] update Jim Cook's headshot on leadership page

craigcook:
## Description
Update Jim Cook's headshot on leadership page. It's just replacing an image, but does this need to be cachebusted?

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1277399

## Testing
Look at the picture. Notice that it is different. Get on with your life.


alexgibson:
r+ 🔎

PR #4172, opened 2016-06-03
[fix bug 1273837] Decommission Hello FTU and redirect to Hello product page

alexgibson:
## Description
- Removes the template pages found at `/firefox/%VERSION%/hello/start` and redirects to `firefox/hello`.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1273837

## Testing
I added tests for the redirect, although an extra pair of eyes on the redirect regexp would be appreciated.

## Checklist
- [x] Requires l10n changes.


alexgibson:
Ping @flodolo - there may be some lang files we can remove here too.

flodolo:
.lang files no, strings definitely.

bedrock/firefox/redirects.py
```diff
@@ -498,4 +498,7 @@ def firefox_mobile_faq(request, *args, **kwargs):
 
     # bug 1260423
     redirect(r'^firefox/choose/?$', 'firefox.new'),
+
+    # bug 1273837
+    redirect(r'^firefox(?:\/\d+\.\d+\.?(?:\d+)?\.?(?:\d+)?(?:a?)(?:\d*))?/hello/start/?$', 'firefox.hello'),
```
jpetto: I think there are one too many digit capturing groups here - the above would match `46.0.1.2` and `46.0.1.2a2`, which I don't think we need (uh, right?).

Assuming an `a` must always be followed by a number, I think you simplify slightly to:

`^firefox(?:\/\d+\.\d+(?:\.\d+)?(?:a\d+)?)?/hello/start/?$`

So, "firefox" followed by at least `#.#`, then optionally another `.#` and optionally `a#` at the end.

jpetto:
Can also get rid of the bundles in `static_media.py`?

jpetto:
Small suggestion on the regex and note to clean up `static_media.py`, but otherwise looks good. 🚎 r+wc

bedrock/firefox/redirects.py
alexgibson: Good catch on the extra digit, will update 👍

PR #4174, opened 2016-06-03
Bug 1255518 - Firefox download buttons: drop support for OS X 10.6-10.8 in August 2016

kyoshino:
This is required before Firefox 49 hits the Aurora channel next week on Thursday or Friday. I know the download button code is a total mess but we have to handle it anyway…

How to test this manually:

1. Make sure product-details is up-to-date
2. Change the condition in `helpers.py` from `>= 49` to `>= 48`
3. Spoof the browser’s UA string so that it looks like Snow Leopard, Lion or Mountain Lion, e.g. `Mozilla/5.0 (Macintosh; Intel Mac OS X 10.8; rv:49.0) Gecko/20100101 Firefox/49.0`
4. Load /firefox/channel/
5. Check if the desktop DevEdition button is disabled while other buttons are enabled

### DIFF
--- a/bedrock/firefox/helpers.py
+++ b/bedrock/firefox/helpers.py
@@ -182,6 +182,9 @@ def download_firefox(ctx, channel='release', platform='all',
     langs = firefox_desktop.languages
     locale_name = langs[locale]['native'] if locale in langs else locale
 
+    # Firefox 49+ requires OS X 10.9 Mavericks and later
+    mavericks_required = show_desktop and int(version.split('.', 1)[0]) >= 49
+
     data = {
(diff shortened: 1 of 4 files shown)


jpetto:
Code looks good and functions as expected. Do you have time to write tests around the new logic? Since we're impacting downloads, it would be really good to have tests.

Also, is a bug already filed to update this logic when 49 hits the release channel? Would be nice to reference that bug in some of the comments made in this PR.

jpetto:
We'll file separate bugs for adding tests & updating this logic before the release channel hits 49. Merging this before the upcoming DevEdition release.

PR #4177, opened 2016-06-07
[fix bug 1278426] IE Download events broken in IE9 and below

alexgibson:
## Description
- Fixes a bug where internal links we're being reported incorrectly in IE9. I had trouble reproducing this locally, but James seems confident this should fix the issue.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1278426

## Testing
Visit `/firefox/new` and click the download button. The `href` that gets passed in `Analytics.updateDataLayerPush()` should be `/firefox/new/?scene=2` (omitting the host and locale). The same applies to any internal link.

### DIFF
--- a/media/js/base/core-datalayer.js
+++ b/media/js/base/core-datalayer.js
@@ -29,8 +29,9 @@ if (typeof Mozilla == 'undefined') {
     *   Adds href stripped of locale to link click objects when pushed to the dataLayer,
     *   also removes protocol and host if same as parent page from href.
     */
-    Analytics.updateDataLayerPush = function() {
+    Analytics.updateDataLayerPush = function(host) {
         var dataLayer = window.dataLayer = window.dataLayer || [];
+        var hostname = host || document.location.hostname;
 
         dataLayer.defaultPush = dataLayer.push;
         dataLayer.push = function() {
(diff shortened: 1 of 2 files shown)


alexgibson:
@JamesDLorence - does this help with your issue?

alexgibson:
I pushed this to demo1 - @JamesDLorence can you please test?

https://www-demo1.allizom.org/en-US/firefox/new/

alexgibson:
Updated the existing test to hopefully stop any regression in the future.

JamesDLorence:
That looks good to me.

alexgibson:
> That looks good to me.

Does this mean you've tested with older IE on the demo server and verified the data now looks ok?

alexgibson:
Ok, got confirmation in the bug that data looks ok: https://bugzilla.mozilla.org/show_bug.cgi?id=1278426#c5

JamesDLorence:
Yes, it's been tested in IE9, 8 and 7 and it looked good in all of them.

alexgibson:
Ok, @jpetto r?

media/js/base/core-datalayer.js
```diff
@@ -39,11 +40,13 @@ if (typeof Mozilla == 'undefined') {
                     var element = arguments[i]['gtm.element'];
                     var href = element.href;
 
-                    if (element.hostname === document.location.hostname) {
-                        var path = href.split(element.host)[1];
+                    if (element.hostname === hostname) {
+                        // remove host and locale from internal links
+                        var path = href.replace(/^(?:https?\:\/\/)(?:[^\/])*/, '');
                         var locale = path.split('/')[1];
```
jpetto: Unsure of how problematic it may be on the site (if at all), but this code assumes a locale is present if the hostnames match. (If somehow a locale were *not* present, the first part of the URL would be cut off, e.g. `https://www.mozilla.org/firefox/new/` would be reported as just `/new/`.

Since our locales follow a predictable format, would it be worth it to be explicit in the code (with another regex) when looking for a locale? Something like:

`var locale = path.match(/^\/(\w{2}\-\w{2}|\w{2})\//);`

jpetto: If we agree to be more explicit in checking for the presence of a locale, would also be good to add a test.

jpetto:
One small-ish code specificity question/suggestion, but otherwise looks good. r+wc

media/js/base/core-datalayer.js
alexgibson: I think this is a worthwhile suggestion as an extra measure. I'll make it so :+1:

alexgibson:
@jpetto - thanks for the review. I updated and added an extra test.

jpetto:
LGTM! 🇬🇧

PR #4180, opened 2016-06-08
Fix Bug 1255518 - Firefox download buttons: drop support for OS X 10.6-10.8, take 2

kyoshino:
#4174 was reverted in #4179. This fixes the CSS rule as discussed on IRC.

### DIFF
--- a/media/css/sandstone/buttons.less
+++ b/media/css/sandstone/buttons.less
@@ -291,6 +291,7 @@ ul.download-list {
 .download-button .linux-arm-download,
 .download-button .unrecognized-download,
 .download-button .unsupported-download,
+.download-button .unsupported-download-osx,
 .download-button .nojs-download {
     display: none;
 }
@@ -319,7 +320,6 @@ ul.download-list {
 .linux .download-button .os_linux,
 .linux.x86.x64 .download-button .os_linux64,
 .windows .download-button .os_win,
-.osx .download-button .os_osx,
 .android .download-button .os_android,
 .download-button-android .os_android,
 .android .download-button-desktop .download-list,
@@ -333,6 +333,15 @@ ul.download-list {
     display: block !important;
 }
 
+// Add another rule for old browsers where :not isn't supported. If this is
+// mixed with the rule above, the entire rule will be invalid on those legacy
+// environments and some tests will fail.
+.osx:not(.pre-mavericks) .download-button .os_osx,
+.osx.pre-mavericks .download-button:not(.mavericks-required) .os_osx {
+    display: block !important;
+}
+
+.osx.pre-mavericks .download-button.mavericks-required .unsupported-download-osx,
 .windows.arm .download-button .unsupported-download,
 .linux.arm .download-button .linux-arm-download,
 .oldwin .download-button .unsupported-download,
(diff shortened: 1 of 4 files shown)


media/css/sandstone/buttons.less
```diff
@@ -333,6 +333,15 @@ ul.download-list {
     display: block !important;
 }
 
+// Add another rule for old browsers where :not isn't supported. If this is
+// mixed with the rule above, the entire rule will be invalid on those legacy
+// environments and some tests will fail.
+.osx:not(.pre-mavericks) .download-button .os_osx,
+.osx.pre-mavericks .download-button:not(.mavericks-required) .os_osx {
+    display: block !important;
+}
```
jpetto: This looks to be fine, as the only OS/browser combination to have trouble would be Firefox 3 on OS X 10.6 - 10.8, which is probably an *extremely* rare edge case. However, I believe those users would not see a download button at all. Is that right?

(I asked in #growth and #www about the number of Fx 3 users still out there, but haven't heard back.)

What would you think about displaying `.osx .download-button .os_osx` by default, then using the `:not` selector to hide them (where appropriate) from modern browsers?

jpetto:
Confirmed download buttons are visible in IE 7 and 8.

jpetto:
In regards to Fx 3 users, pg in #www said:

> There were 3.5 million sessions accounting for 2.6% of all Moz.org traffic in May 2016
> 95% of the Fx3.x users are on windows

If my math is right, that means 175,000 Fx 3 users visited mozilla.org *not* on Windows. It would be best if we can avoid completely hiding the download button for the Mac portion of those users, but given that this is a temporary change, I'm not sure it's a blocker.

media/css/sandstone/buttons.less
kyoshino: Where's Firefox 3 from? As per MDN, [`:not`](https://developer.mozilla.org/en-US/docs/Web/CSS/:not) is supported since Firefox 1.

jpetto: [Can i use says Fx 3 doesn't support CSS3 selectors](http://caniuse.com/#feat=css-sel3). I would assume MDN is more accurate though?

jpetto: Just tested in Fx 3.6 on Windows Vista and the `:not` selector does work. Woo!

media/css/sandstone/buttons.less
```diff
@@ -319,7 +320,6 @@ ul.download-list {
 .linux .download-button .os_linux,
 .linux.x86.x64 .download-button .os_linux64,
 .windows .download-button .os_win,
-.osx .download-button .os_osx,
```
jpetto: Since this line will need to be resurrected when release hits 49, maybe instead of deleting we comment it out with a short explanation?

media/css/sandstone/buttons.less
kyoshino: [Bug 71647](https://bugzilla.mozilla.org/show_bug.cgi?id=71647) fixed in 2001 😸

jpetto:
As Fx 3 does indeed support `:not` selectors, and this is a time-sensitive update, I think we're good to merge here. r+

media/css/sandstone/buttons.less
kyoshino: MDN sometimes provides inaccurate information so I don't trust the site neither.

PR #4182, opened 2016-06-09
[bug 1270316] Core DataLayer Additions

JamesDLorence:
## Description
Additions to the Core DataLayer Object, including making sure that the Firefox and plugin details have been pushed to the dataLayer before the Core DataLayer Object is pushed to ensure that data is available when a pageview is triggered.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1270316

### DIFF
--- a/media/js/base/core-datalayer-init.js
+++ b/media/js/base/core-datalayer-init.js
@@ -4,14 +4,32 @@
 
 // init core dataLayer object and push into dataLayer
 $(function() {
-    if (Mozilla && Mozilla.Analytics) {
-        var dataLayer = window.dataLayer = window.dataLayer || [];
+    var analytics = Mozilla.Analytics;
+    var client = Mozilla.Client;
+    var dataLayer = window.dataLayer;
 
-        dataLayer.push({
+    function sendCoreDataLayer() {
+        var dataLayerCore = {
             'event': 'core-datalayer-loaded',
-            'pageId': Mozilla.Analytics.getPageId()
-        });
+            'pageId': analytics.getPageId(),
+            'pageHasDownload': analytics.pageHasDownload(),
+            'pageHasVideo': analytics.pageHasVideo(),
+            'pageVersion': analytics.getPageVersion(),
+            // white listed for www.mozill.org, will always return false on other domains
+            'testPilotUser': 'testpilotAddon' in navigator ? 'true' : 'false'
+        };
+
+        dataLayer.push(dataLayerCore);
     }
 
-    Mozilla.Analytics.updateDataLayerPush();
+    if (client.isFirefoxDesktop || client.isFirefoxAndroid) {
+        client.getFirefoxDetails(function(details) {
+            dataLayer.push(details);
+            sendCoreDataLayer();
+        });
+    } else {
+        sendCoreDataLayer();
+    }
+
+    analytics.updateDataLayerPush();
 });
--- a/media/js/base/core-datalayer.js
+++ b/media/js/base/core-datalayer.js
@@ -25,6 +25,33 @@ if (typeof Mozilla == 'undefined') {
         return pageId ? pageId : pathName.replace(/\/[^\/]*/, '');
     };
 
+    /** Returns whether page has download button.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} string.
+    */
+    Analytics.pageHasDownload = function() {
+        return $('[data-download-os]').length ? 'true' : 'false';
+    };
+
+    /** Returns whether page has video.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} string.
+    */
+    Analytics.pageHasVideo = function() {
+        return ($('video').length || $('iframe[src^="https://www.youtube"]').length) ? 'true' : 'false';
+    };
+
+    /** Returns page version.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} version number from URL.
+    */
+    Analytics.getPageVersion = function(path) {
+        var pathName = path ? path : document.location.pathname;
+        var versionResults = /firefox\/(\d+(?:\.\d+)?\.\da?\d?)/.exec(pathName);
+
+        return versionResults ? versionResults[1] : null;
+    };
+
     /** Monkey patch for dataLayer.push
     *   Adds href stripped of locale to link click objects when pushed to the dataLayer,
     *   also removes protocol and host if same as parent page from href.
--- a/media/js/base/mozilla-fxa-iframe.js
+++ b/media/js/base/mozilla-fxa-iframe.js
@@ -72,6 +72,7 @@ Mozilla.FxaIframe = (function() {
     };
 
     var _sendGAEvent = function(type, extra) {
+        var dataLayer = window.dataLayer || [];
         var data = {
             'event': _gaEventName, // user-configurable, defaults to 'fxa'
             'interaction': type
@@ -82,7 +83,7 @@ Mozilla.FxaIframe = (function() {
             $.extend(data, extra);
         }
 
-        window.dataLayer.push(data);
+        dataLayer.push(data);
     };
 
     var _userCallback = function(callbackName, data) {
--- a/tests/unit/spec/base/core-datalayer.js
+++ b/tests/unit/spec/base/core-datalayer.js
@@ -25,6 +25,66 @@ describe('core-datalayer.js', function() {
         });
     });
 
+    describe('pageHasDownload', function() {
+
+        afterEach(function() {
+            $('.download').remove();
+        });
+
+        it('will return "true" when download button is present on page.', function() {
+            var downloadMarkup = '<a class="download" href="#" data-link-type="download" data-download-os="Desktop">';
+
+            $(downloadMarkup).appendTo('body');
+            expect(Mozilla.Analytics.pageHasDownload()).toBe('true');
+        });
+
+        it('will return "false" when download button is not present on page.', function() {
+            expect(Mozilla.Analytics.pageHasDownload()).toBe('false');
+        });
+    });
+
+    describe('pageHasVideo', function() {
+
+        afterEach(function() {
+            $('#htmlPlayer').remove();
+            $('#iframePlayer').remove();
+        });
+
+        it('will return "true" when HTML5 video is present on page.', function() {
+            var videoMarkup = '<video id="htmlPlayer"></video>';
+
+            $(videoMarkup).appendTo('body');
+            expect(Mozilla.Analytics.pageHasVideo()).toBe('true');
+        });
+
+        it('will return "true" when YouTube iframe video is present on page.', function() {
+            var videoMarkup = '<iframe id="iframePlayer" src="https://www.youtube-nocookie.com/embed/NqxUdc0P6YE?rel=0"></iframe>';
+
+            $(videoMarkup).appendTo('body');
+            expect(Mozilla.Analytics.pageHasVideo()).toBe('true');
+        });
+
+        it('will return "false" when download button is not present on page.', function() {
+            expect(Mozilla.Analytics.pageHasVideo()).toBe('false');
+        });
+    });
+
+    describe('getPageVersion', function() {
+        it('will return the Firefox version number form the URL if present', function() {
+            expect(Mozilla.Analytics.getPageVersion('https://www.mozilla.org/en-US/firefox/107.0.11/firstrun/')).toBe('107.0.11');
+            expect(Mozilla.Analytics.getPageVersion('https://www.mozilla.org/en-US/firefox/46.0.1/firstrun/learnmore/')).toBe('46.0.1');
+            expect(Mozilla.Analytics.getPageVersion('https://www.mozilla.org/en-US/firefox/11.0/whatsnew/')).toBe('11.0');
+            expect(Mozilla.Analytics.getPageVersion('https://www.mozilla.org/en-US/firefox/46.0.1a2/firstrun/')).toBe('46.0.1a2');
+            expect(Mozilla.Analytics.getPageVersion('https://www.mozilla.org/en-US/firefox/46.0a1/whatsnew/')).toBe('46.0a1');
+        });
+
+        it('will return null if no version number present in URL', function() {
+            expect(Mozilla.Analytics.getPageVersion('https://www.mozilla.org/en-US/')).toBeNull();
+            expect(Mozilla.Analytics.getPageVersion('https://www.mozilla.org/en-US/firefox/new')).toBeNull();
+            expect(Mozilla.Analytics.getPageVersion('https://www.mozilla.org/en-US/firefox/whatsnew')).toBeNull();
+        });
+    });
+
     describe('updateDataLayerPush', function() {
         var linkElement;
 
--- a/media/js/base/mozilla-client.js
+++ b/media/js/base/mozilla-client.js
@@ -222,17 +222,6 @@ if (typeof Mozilla === 'undefined') {
             };
 
             callback(details);
-
-            // Log the info with GA. Remove this once we have sufficient data
-            window.dataLayer = window.dataLayer || [];
-            window.dataLayer.push({
-                'event': 'firefox-details-retrieved',
-                'accurate': accurate,
-                'version': version,
-                'channel': channel,
-                'isUpToDate': isUpToDate,
-                'isESR': isESR
-            });
         };
 
         // Prepare fallback function in case the API doesn't work
@@ -246,8 +235,8 @@ if (typeof Mozilla === 'undefined') {
             return;
         }
 
-        // Fire the fallback function in .3 seconds
-        var timer = window.setTimeout(fallback, 300);
+        // Fire the fallback function in .4 seconds
+        var timer = window.setTimeout(fallback, 400);
 
         // Trigger the API
         document.addEventListener('mozUITourResponse', listener);


JamesDLorence:
Alex, if you'd like to talk through anything I'd be more than happy to hop on a call.

JamesDLorence:
I'll submit new PR to fix failed test after tomorrow's meeting with Alex and Gareth.

media/js/base/core-datalayer-init.js
```diff
@@ -5,13 +5,47 @@
...
+                    resolve(details);
+                });
+            });
+            promises.push(firefoxDetails);
+
+            if (/plugincheck/.test(document.location.pathname)) {
+                var pluginDetails = new Promise(function(resolve, reject) {
+                    document.addEventListener('pluginCheckDone', resolve);
+
+                    window.setTimeout(resolve, 400);
```
alexgibson: Some other things I'd like to discuss in the meeting tomorrow include the robustness of doing this, over using the current GA data provided by the plugincheck page itself.

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,55 @@ if (typeof Mozilla == 'undefined') {
...
+    /** Returns page type.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} GTM page type.
+    */
+    Analytics.getPageType = function(path) {
+        var pathName = path ? path : document.location.pathname;
+
+        if (/firefox\/(desktop|android|ios)\/$/.test(pathName)) {
+            return 'product page';
+        } else if (/firefox\/(hello|sync|accounts|private-browsing)\/$/.test(pathName)) {
```
alexgibson: This also seems like something we should discuss, as it will likely be out of date before long unless it's manually maintained, which isn't ideal.

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,55 @@ if (typeof Mozilla == 'undefined') {
...
+    }
+
+    /** Returns whether page is long scrolling.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} boolean.
+    */
+    Analytics.isLongScrollingPage = function(path) {
+        var pathName = path ? path : document.location.pathname;
+
+        if (/firefox\/(desktop|android|ios|hello|sync|private-browsing)\/$/.test(pathName)) {
```
alexgibson: This also seems like something that won't stand the test of time. I'd like to discuss something more dynamic / robust for reporting page length more accurately.

vendor-local/src/legal-docs
```diff
@@ -1 +1 @@
-Subproject commit 0b9d56b979421eb9606949ca3dfbd4140950e107
+Subproject commit a41df77591b5f4aba6b736410ecc94a81add29ee
```
alexgibson: This sub module commit seems to have been added by accident and should be removed.

media/js/base/core-datalayer-init.js
```diff
@@ -4,14 +4,31 @@
 
 // init core dataLayer object and push into dataLayer
 $(function() {
-    if (Mozilla && Mozilla.Analytics) {
-        var dataLayer = window.dataLayer = window.dataLayer || [];
+    var dataLayer = window.dataLayer = window.dataLayer || [];
```
alexgibson: Do we really need to repeat this seeing as it's already defined as a global in the [GTM snippet](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/includes/google-analytics.html#L7)? If we do need it for some reason, could it be shortened to `var dataLayer = window.dataLayer || [];`?

media/js/base/core-datalayer-init.js
```diff
@@ -4,14 +4,31 @@
...
+    function sendCoreDataLayer() {
+        var dataLayerCore = {
             'event': 'core-datalayer-loaded',
-            'pageId': Mozilla.Analytics.getPageId()
-        });
+            'pageId': Mozilla.Analytics.getPageId(),
+            'pageHasDownload': Mozilla.Analytics.pageHasDownload(),
+            'pageHasVideo': Mozilla.Analytics.pageHasVideo(),
+            'pageVersion': Mozilla.Analytics.getPageVersion(),
+            'testPilotUser': window.navigator.testpilotAddon ? window.navigator.testpilotAddon : 'false'
```
alexgibson: Could we maybe use `'testpilotAddon' in navigator ? window.navigator.testpilotAddon : false` instead? Also any reason we you're using a string for `'false'` at the end instead of a Boolean? (the property seems to return the latter from what I can see).

media/js/base/core-datalayer-init.js
```diff
@@ -4,14 +4,31 @@
...
+        };
+
+        dataLayer.push(dataLayerCore);
     }
 
-    Mozilla.Analytics.updateDataLayerPush();
+    if (Mozilla && Mozilla.Analytics) {
+        if (Mozilla.Client.isFirefox) {
```
alexgibson: Here we should use `client.isFirefoxDesktop || client.isFirefoxAndroid` instead, as Firefox for iOS doesn't have version  information available in the same way other platforms do. You could also use a local reference to `Mozilla.Client` using `var client = window.Mozilla.Client;`

media/js/base/mozilla-fxa-iframe.js
```diff
@@ -72,6 +72,7 @@ Mozilla.FxaIframe = (function() {
     };
 
     var _sendGAEvent = function(type, extra) {
+        var dataLayer = window.dataLayer = window.dataLayer || [];
```
alexgibson: Again is this needed?

media/js/plugincheck/check-plugins.js
```diff
@@ -257,6 +257,8 @@ $(function() {
                         'plugin-unknown-count': pluginTotals.unknownCount
                     });
 
+                    document.dispatchEvent(new CustomEvent('pluginCheckDone'));
```
alexgibson: I think we can remove this now after our last meeting too?

tests/unit/spec/base/core-datalayer.js
```diff
@@ -25,6 +25,56 @@ describe('core-datalayer.js', function() {
...
     });
 
+    describe('pageHasDownload', function() {
+
+        afterEach(function() {
+            $('.download').remove();
+        });
+
+        it('will return "true" when download button is present on page.', function() {
+            var downloadMarkup =   '<a class="download" href="#" data-link-type="download" data-download-os="Desktop">';
```
alexgibson: Nit: remove extra spaces before `=`

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,33 @@ if (typeof Mozilla == 'undefined') {
...
+        return ( $('video').length > 0 ) ? 'true' : 'false';
+    };
+
+    /** Returns page version.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} version number from URL.
+    */
+    Analytics.getPageVersion = function(path) {
+        var pathName = path ? path : document.location.pathname;
+        var versionResults = /firefox\/(\d{1,2}\.\d{1,2}\.?\d{0,2})/.exec(pathName);
```
alexgibson: This omits the channel numbers from the version URLs in Dev Edition (e.g. `49.0a2`) & Nightly (`50.0a1`). Is this intentional, or should we be capturing this info too?

media/js/base/core-datalayer-init.js
```diff
@@ -4,14 +4,31 @@
...
-        var dataLayer = window.dataLayer = window.dataLayer || [];
+    var dataLayer = window.dataLayer = window.dataLayer || [];
 
-        dataLayer.push({
+    function sendCoreDataLayer() {
+        var dataLayerCore = {
             'event': 'core-datalayer-loaded',
-            'pageId': Mozilla.Analytics.getPageId()
-        });
+            'pageId': Mozilla.Analytics.getPageId(),
```
alexgibson: Instead of repeating `Mozilla.Analytics` every time you could also use a local reference `var analytics = Mozilla.Analytics`;

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,33 @@ if (typeof Mozilla == 'undefined') {
         return pageId ? pageId : pathName.replace(/\/[^\/]*/, '');
     };
 
+    /** Returns whether page has download button.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} boolean.
+    */
+    Analytics.pageHasDownload = function() {
+        return ( $('[data-download-os]').length > 0 ) ? 'true' : 'false';
```
alexgibson: Could you shorten this to `return $('[data-download-os]').length ? true : false;`?

alexgibson: Also is there any reason you're looking for a data attribute specifically? Technically, looking for a class like `.download-link` would be faster?

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,33 @@ if (typeof Mozilla == 'undefined') {
...
+    };
+
+    /** Returns whether page has video.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} boolean.
+    */
+    Analytics.pageHasVideo = function() {
+        return ( $('video').length > 0 ) ? 'true' : 'false';
```
alexgibson: Could this also be shortened to `return $('video').length ? true : false;`

media/js/base/core-datalayer.js
alexgibson: Note: you're returning a String here and not a Boolean as the function description suggests?

media/js/base/core-datalayer.js
alexgibson: Note: you're returning a String here and not a Boolean as the function description suggests?

media/js/base/core-datalayer-init.js
JamesDLorence: The reason I'm using a string and not a Boolean is that Google converts Boolean to 0 or 1 which doesn't make for a great user experience in the GA reports. So instead I always will pass the strings 'true' or 'false' and not Boolean values.

alexgibson: Ah I see, that makes sense. In the case of this property here though, would that result in inconsistent data? (a mixture of '1' and 'false'?

media/js/base/core-datalayer-init.js
JamesDLorence: If you're comfortable with it being removed, that is fine with me. I just wasn't sure if you were concerned about there being a dependency on another file that is obscure. I will remove it though.

alexgibson: window.dataLayer is used all over the site already. If it was removed from the GTM snippet much breakage would already ensue, so I'm not sure adding this line everywhere is really worth it at this stage.

JamesDLorence: Sounds good then. I'll remove it.

media/js/base/core-datalayer-init.js
JamesDLorence: That is a problem so I'll update to pass the string 'true' if `testPilotAddon` is there.

media/js/base/core-datalayer-init.js
JamesDLorence: Will do.

media/js/base/mozilla-fxa-iframe.js
JamesDLorence: Will remove.

media/js/plugincheck/check-plugins.js
JamesDLorence: Will remove.

tests/unit/spec/base/core-datalayer.js
JamesDLorence: Will remove.

media/js/base/core-datalayer.js
JamesDLorence: This was not intentional. Will make the update to capture the Deve Edition and Nightly version numbers.

media/js/base/core-datalayer.js
JamesDLorence: Good call on the shortening.  And I went with using the data attribute over the class `.download-link` becasue I was not sure if all download link had that class and but I know that all download links do have `data-download-os`, or they should at least.

JamesDLorence: I've updated the function description from Boolean to String.

media/js/base/core-datalayer.js
JamesDLorence: Good call and I've updated the function description from Boolean to String.

media/js/base/core-datalayer-init.js
```diff
@@ -4,14 +4,31 @@
...
+        };
+
+        dataLayer.push(dataLayerCore);
     }
 
-    Mozilla.Analytics.updateDataLayerPush();
+    if (Mozilla && Mozilla.Analytics && Mozilla.Client) {
+        var client = Mozilla.Client;
+
+        if (client.isFirefox) {
```
alexgibson: This should be `client.isFirefoxDesktop || client.isFirefoxAndroid`

media/js/base/mozilla-fxa-iframe.js
```diff
@@ -72,6 +72,7 @@ Mozilla.FxaIframe = (function() {
     };
 
     var _sendGAEvent = function(type, extra) {
+        var dataLayer = window.dataLayer || [];
```
alexgibson: Is this needed?

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,33 @@ if (typeof Mozilla == 'undefined') {
...
+        return $('video').length ? 'true' : 'false';
+    };
+
+    /** Returns page version.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} version number from URL.
+    */
+    Analytics.getPageVersion = function(path) {
+        var pathName = path ? path : document.location.pathname;
+        var versionResults = /firefox\/(\d{1,2}\.(\d{1,2}|\da\d)\.?\d{0,2})/.exec(pathName);
```
alexgibson: This regexp is still failing for version numbers like `46.0.1a2` and `46.0a1`. It would be great to have test coverage for this 👍

alexgibson:
Looks like you have a few ESLint errors introduced in the last round of amends. Rather than wait to find out about them here, you can always run the check locally using the `gulp js:lint` command. Even better still, you can also install an ESLint plugin for most editors that will lint your code as you type 👍

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,33 @@ if (typeof Mozilla == 'undefined') {
...
+        return $('video').length ? 'true' : 'false';
+    };
+
+    /** Returns page version.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} version number from URL.
+    */
+    Analytics.getPageVersion = function(path) {
+        var pathName = path ? path : document.location.pathname;
+        var versionResults = /firefox\/(\d{1,2}(?:\.\d{1,2})?\.\da?\d?)/.exec(pathName);
```
alexgibson: If we restrict the regex to a maximum on 2 digits in the version number then this will no longer match when we reach `100`?

media/js/base/mozilla-fxa-iframe.js
alexgibson: OK, it looks like the test that fails due to `TypeError: window.dataLayer is undefined` was relying on where the declaration is being removed from `mozilla-client.js` above  :/

I think this is fine to leave in for now, although I'd like to find a more robust way to mock `window.dataLayer` for all tests in the future.

alexgibson:
I've pushed this branch to demo4 for testing: https://www-demo4.allizom.org/

media/js/base/core-datalayer.js
JamesDLorence: Yes, it will not work once you hit `100.0`. Would you like me to update it so that it will capture `100.0`?

JamesDLorence:
While testing, I noticed 2 things:
1. We don't capture that /teach/smarton/security and /teach/smarton/surveillance have videos since they are in `<iframe>` elements and not `<video>` elements. I'd like to update the logic to identify `<iframe>` elements that include 'youtube.com' in the src. Sound good?
2. None of the MozTourAPI calls are returning accurate data. Is that becasue we're on a demo server? Is there a way to test to ensure this it working properly?

alexgibson:
1. This sounds OK to me.
2. Sorry I should have mentioned this in the bug. You'll need to whitelist the demo server for UITour to work: http://bedrock.readthedocs.io/en/latest/uitour.html#local-development

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,33 @@ if (typeof Mozilla == 'undefined') {
...
+        return ($('video').length || $('iframe[src*="youtube"]').length) ? 'true' : 'false';
+    };
+
+    /** Returns page version.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} version number from URL.
+    */
+    Analytics.getPageVersion = function(path) {
+        var pathName = path ? path : document.location.pathname;
+        var versionResults = /firefox\/(\d{1,3}(?:\.\d{1,2})?\.\da?\d?)/.exec(pathName);
```
alexgibson: Sorry if I wasn't clear enough before, but I meant why restrict the major version number to a certain number of decimals at all?

media/js/base/core-datalayer-init.js
```diff
@@ -4,14 +4,31 @@
...
-        dataLayer.push({
+    function sendCoreDataLayer() {
+        var dataLayerCore = {
             'event': 'core-datalayer-loaded',
-            'pageId': Mozilla.Analytics.getPageId()
+            'pageId': analytics.getPageId(),
+            'pageHasDownload': analytics.pageHasDownload(),
+            'pageHasVideo': analytics.pageHasVideo(),
+            'pageVersion': analytics.getPageVersion(),
+            'testPilotUser': 'testpilotAddon' in navigator ? 'true' : 'false'
```
alexgibson: I think it would be useful to add a comment here that this property is white listed for www.mozilla.org only, and we can expect it to always return false for other domains (to avoid confusion in the future).

media/js/base/core-datalayer.js
```diff
@@ -25,6 +25,33 @@ if (typeof Mozilla == 'undefined') {
...
+    Analytics.pageHasDownload = function() {
+        return $('[data-download-os]').length ? 'true' : 'false';
+    };
+
+    /** Returns whether page has video.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} string.
+    */
+    Analytics.pageHasVideo = function() {
+        return ($('video').length || $('iframe[src*="youtube"]').length) ? 'true' : 'false';
```
alexgibson: Nit: let's make this a little bit more explicit and check for `iframe[src^="https://www.youtube"]`. This should cover both regular and nocookie domains, but also ensures it appears at the start of the `src` value.

alexgibson:
Aside from a couple of nits this is looking good, although I will defer to the results of your testing on demo before giving this an r+wc. It would also be nice if you could give this a rebase against the latest master before merging. Thanks!

media/js/base/core-datalayer.js
JamesDLorence: Good call.

JamesDLorence:
If you throw this version up on the demo server I will test it again and confirm everything is working as expected.

alexgibson:
> If you throw this version up on the demo server I will test it again and confirm everything is working as expected.

Latest code is now up on demo 4

JamesDLorence:
I've tested and everything looks good except I wanted to confirm that you were going to test and make sure that `navigator.testPilotAddon` is in fact that for test pilot users? If that checks out then everything looks good and we can go ahead and merge.

alexgibson:
> I've tested and everything looks good except I wanted to confirm that you were going to test and make sure that navigator.testPilotAddon is in fact that for test pilot users? If that checks out then everything looks good and we can go ahead and merge.

I've been taking a look at ways to confirm this, and the only way currently seems to be using manual verification.

1. Install and enable the testpilot addon by going to https://testpilot.firefox.com/
2. Visit https://www.mozilla.org/
3. Open the web console and pasting `'testpilotAddon' in window.navigator` returns 'true'.
4. Go to `about:addons` and either disable or remove the testpilot addon.
5. Visit https://www.mozilla.org/
6. Open the web console and pasting `'testpilotAddon' in window.navigator` returns 'false'.

JamesDLorence:
Ok, I've confirmed that `navigator.testpilotAddon` does return true for users that have installed the test pilot addon.

Unless you have any other concerns, please go ahead and merge this.

alexgibson:
Thanks, I'll take one final look before merging but I think we're good.

alexgibson:
r+ 🏁

PR #4185, opened 2016-06-15
Fix bug 1280217: FSA Major free text should pass to basket

pmac:
The field name in the clean method was misspelled.


pmac:
@jgmize r?

jgmize:
Test coverage for both cases would be nice, but I won't consider it a blocker for this simple fix.
r+

PR #4186, opened 2016-06-20
[fix bug 1278190] Implement improved /firstrun redesigned page

alexgibson:
## Description
- Adds new "horizon" template design to the existing /firstrun page.

**Note:** This is still blocked on final approval from creative, but we should be OK to being code review.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1278190

## Testing
This PR adds new template design only, so there should be no new logic in terms of the existing JS / signup flow. You can still test the form on the demo server below [using these instructions](http://bedrock.readthedocs.io/en/latest/firefox-accounts.html#demo-server-testing).
 
Demo server URL: https://www-demo3.allizom.org/en-US/firefox/47.0/firstrun/

## Checklist
- [ ] Requires l10n changes (should be no new strings for this PR).


craigcook:
Not really relevant to this PR, but I just noticed that we could/should remove the `title` elements from some of the SVGs. They're not especially meaningful and I imagine they're just output from Illustrator or Sketch (they look like layer names to me). It doesn't matter here because the images are used as CSS backgrounds, but titles in inline SVG can be read by screen readers so should be omitted for presentational images.

craigcook:
r+, will wait for the pipes to clear before I merge.

PR #4189, opened 2016-06-21
[bug 1279606] Move Page ID dataLayer push before GTM loads

JamesDLorence:
## Description
Move page ID dataLayer push to before the GTM container instantiates.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1279606


JamesDLorence:
Alex, have you had a chance to review this yet? I have a Call with Pmac and Gareth about where we're at with making GTM CSP compliant and this is blocking me.

alexgibson:
> Alex, have you had a chance to review this yet? I have a Call with Pmac and Gareth about where we're at with making GTM CSP compliant and this is blocking me.

Nope sorry I have other priorities I need to take care of first. Gareth understands this work is not in our sprint planning, so I'm afraid this will be blocked until I have time to take a look.

JamesDLorence:
Completely understand. I just wanted to be able to communicate with them where we're at. Thanks for the update.

alexgibson:
No worries, really I'd like to get #4182 merged and into production first before this one (this will also likely need rebasing come then too).

alexgibson:
@JamesDLorence - now that #4182 has merged, can you please rebase this branch against master? Thanks

JamesDLorence:
Rebased and ready! 👍

media/js/base/core-datalayer.js
```diff
@@ -2,50 +2,34 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-// create namespace
-if (typeof Mozilla == 'undefined') {
```
alexgibson: Because this is relying & adding to an existing namespace that's now in a different bundle altogether, I think we should probably keep this check, and also add a fallback for if `Mozilla.Analytics` is undefined also?

media/js/base/datalayer-page-id.js
```diff
@@ -0,0 +1,32 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+if (typeof Mozilla == 'undefined') {
+    var Mozilla = {};
+}
+
+(function() {
+    var dataLayer = window.dataLayer = window.dataLayer || [];
```
alexgibson: As we're now moving this out of the GTM snippet and into an external script, it would be worthwhile adding a comment here that states this is where the core `window.dataLayer` object is now defined?

alexgibson:
Now that `getPageId` is moving to it's own file and has it's own test suite, should we remove the [existing duplicate test](https://github.com/mozilla/bedrock/blob/master/tests/unit/spec/base/core-datalayer.js#L10-L26)?

media/js/base/datalayer-page-id.js
```diff
@@ -0,0 +1,32 @@
...
+
+    /** Returns page ID used in Event Category for GA events tracked on page.
+    * @param {String} path - URL path name fallback if page ID does not exist.
+    * @return {String} GTM page ID.
+    */
+    Analytics.getPageId = function(path) {
+        var pageId = document.getElementsByTagName('html')[0].getAttribute('data-gtm-page-id');
+        var pathName = path ? path : document.location.pathname;
+
+        return pageId ? pageId : pathName.replace(/\/[^\/]*/, '');
```
alexgibson: It would be nice to match the locale a little more explicitly here, like we did in `updateDataLayerPush`

alexgibson: Once you do this, adding an extra test to make sure a URL path that does not contain a locale functions as expected would be nice too.

alexgibson:
Small nit: for consistency sake I think we should rename `datalayer-page-id.js` to `core-datalayer-page-id.js`

alexgibson:
@JamesDLorence - this is now up on demo4 for testing: https://www-demo4.allizom.org/en-US/

JamesDLorence:
There was an issue with the first forward slash being removed from the `pageId` when removing the locale which I fixed, and I updated the name from `datalayer-page-id.js` to `core-datalayer-page-id.js`.

alexgibson:
Updated demo4 with the latest code mentioned above.

JamesDLorence:
Everything looks good to me. Thumbs up to merge from me.

alexgibson:
Sorry for the delay on getting back to this, I've been occupied by other priorities this week and I need to try and resolve an issue with one of our demo servers before I take a look at manually testing this.

alexgibson:
I went through some testing of a few edge case scenarios and there's one recommendation I'd like to make:

If for whatever reason a page exists where the `core-datalayer-page-id.js` is not present in the head, then any call to `window.dataLayer` in the [common base bundle](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/static_media.py#L1000-L1014) will now throw an error. I don't think this is a problem in reality, but I'd like to err on the side of caution and make sure a definition of `window.dataLayer` also exists in `core-datalayer-init.js` [here]https://github.com/mozilla/bedrock/blob/master/media/js/base/core-datalayer-init.js#L9). (e.g. `var dataLayer = window.dataLayer || [];`).

Aside from this one nit, this looks good to me r+wc

JamesDLorence:
Defined `window.dataLayer` in `core-datalayer-init.js` and rebased against master. It should be ready. 👍

alexgibson:
Thanks @JamesDLorence r+

PR #4190, opened 2016-06-22
[fix bug 1278184] Implement improved /new redesigned download page

alexgibson:
## Description
- New redesigned /new download page for `en-US` locale. This will be A/B tested against what's currently in production before being rolled out as the "final final new design".
- Updates meta description for both current and new en-US pages, as per [bug 1262236](https://bugzilla.mozilla.org/show_bug.cgi?id=1262236).

**Note:** This is still blocked on final approval from creative, but we should be OK to being code review.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1278184
https://bugzilla.mozilla.org/show_bug.cgi?id=1262236

## Testing

Demo server:
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/?v=1

URLs:
- `/en-US/firefox/new/?v=1`
- `/en-US/firefox/new/scene=2&v=1`

Page states:
- Firefox up-to-date (current release version).
- Firefox up-to-date (newer than current release version).
- Firefox out-of-date.
- Non-Firefox.
- iOS.
- Android.
- Unsupported.
- No-JS.

## Checklist
- [ ] Requires l10n changes.


alexgibson:
OK, I'm removing the do-not-merge label from this as we have the go ahead to ship the static (non-animated) version of the page, until we can get feedback from the creative team.

craigcook:
The mailbox graphic gets distorted at some breakpoints. There's a `background-size` defined in fxnewsletter-subscribe.less that needs an override.

mailbox-squish

craigcook:
The star background is clipped at smaller breakpoints. Not a huge issue but it does look awkward with the line just ending in mid-air. We could do `background-size: contain` but then the image is tiny. The other option would be applying that image to an element that fills the full width, but that might mean adding some other container inside it to center the content.

stars-clipped

bedrock/firefox/templates/firefox/new/horizon/scene1.html
```diff
@@ -0,0 +1,120 @@
...
+  <div class="main-container">
+    <div class="horizon">
+      <div class="inner-container">
+        <header id="masthead">
+          <div id="tabzilla">
+            <a href="{{ url('mozorg.home') }}" data-link-type="nav" data-link-name="tabzilla">Mozilla</a>
+          </div>
+          <h2>{{ high_res_img('firefox/template/header-logo-inverse.png', {'alt': 'Firefox', 'width': '185', 'height': '70'}) }}</h2>
+          <div class="version-message-container">
+            <div id="version-message-android-latest" class="version-message">
```
craigcook: Semantic nitpick: maybe these should be headings or paragraphs instead of generic divs. I think this is carried over from the older page (pre-rocket, even) but they probably should have been more semantically valuable elements there as well.

craigcook:
Looks good, just a few little graphic tweaks. r+wc 🌄

alexgibson:
Thanks for the review @craigcook - I added a followup commit for easier review 👍

PR #4191, opened 2016-06-22
Fix Bug 1281648 - default ios version should be 4.0 instead of 3.0 in release notes

kyoshino:


alexgibson:
r+ thanks @kyoshino 🐢

PR #4196, opened 2016-06-28
[fix bug 1282903] Swap system font for Open Sans on buttons.

jpetto:
## Description

Without a specific `font-family` set on `<input>` and `<button>` elements, the font falls back to that dictated by the UA stylesheet. It appears simply adding `font-family: inherit;` to the `.base-button` class fixes the issue.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1282903

## Testing

Check instances of `class="button"`. Specific examples are:

- `/about/`
- `/firefox/ios/`
- `/firefox/ios/testflight/`
- `/firefox/new/?v=1` (with Fx)
- `/firefox/new/` (with up-to-date Fx)

## Checklist

### DIFF
--- a/media/css/sandstone/sandstone-resp.less
+++ b/media/css/sandstone/sandstone-resp.less
@@ -366,12 +366,18 @@ input[type=time] {
     border-width: 1px;
     @shadow: 0 1px rgba(255,255,255,0.5);
     .font-size(@smallFontSize);
-    font-family: inherit;
     padding: (@baseLine / 5)  (@baseLine / 2);
     .border-radius(3px);
     .box-shadow(@shadow);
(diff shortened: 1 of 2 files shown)


jpetto:
@craigcook r?

This seems like the simplest fix, but maybe there's a better way?

craigcook:
This does the trick, but we might want to put it in sandstone for all inputs and buttons, not just this class? 

On other projects I've even done this in a CSS reset to make it a baseline style:
<pre><code>input, select, textarea, button { 
    font-family: inherit; 
    font-size: inherit; 
}</code></pre>

craigcook:
r+ 👪

PR #4198, opened 2016-06-30
[fix bug 1263268] Update Foundation board members

craigcook:
## Description
Removes Joi Ito from the active board and adds a new category for emeritus members, comprising Joi Ito and Mitch Kapor. Includes some copy changes but these pages aren't localized.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1263268


alexgibson:
LGTM, r+ 📷

bedrock/mozorg/templates/mozorg/about/leadership.html
```diff
@@ -13,7 +13,7 @@
   <h1 class="title-shadow-box">{{ _('Leadership') }}</h1>
 
   <p>
-  {{ _('Our leadership team is made up of both paid staff and volunteers who represent a breadth of experience and expertise in building products that people love and empowering communities to make a significant impact on the project and in the world.') }}
+  {{ _('Our leadership team is made up of both paid staff and volunteers with expertise building products and empowering communities.') }}
```
flodolo: CC @peiying2
This change landed without ever being flagged for l10n 👎 

Result: we have a half translated paragraph in all localized pages.

flodolo:
Need to fix both the paragraph and `Emeritus Board Members`

alexgibson:
Sorry @flodolo - I wasn't aware we localized this page, since we don't normally use l10n tags when making updates to the tempate. Will make sure to block merging next time before checking.

flodolo:
We do localize it, but clearly can't hide new members, so we never used tags. 
For the first paragraph it would have definitely made sense to use one.

alexgibson:
Would it help if I did a quick PR to add a tag on the paragraph?

craigcook:
I also didn't realize this page was localized so I didn't flag it. Sorry :/

flodolo:
To clarify: the one localized is `mozorg/about/leadership.html`, which is not mentioned in the title, and that's also another reason I missed it completely.

PR #4199, opened 2016-06-30
Do student ambassadors basket update with one call

pmac:
Basket now supports adding all of the data necessary for the FSA form in a single subscribe call. Let's do that as the old way is error prone.

### DIFF
--- a/bedrock/mozorg/forms.py
+++ b/bedrock/mozorg/forms.py
@@ -17,7 +17,6 @@ from django.utils.safestring import mark_safe
 from localflavor.us.us_states import STATE_CHOICES
 
 import basket
-from basket.base import request
 
 from lib.l10n_utils.dotlang import _
 from lib.l10n_utils.dotlang import _lazy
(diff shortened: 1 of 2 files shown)


pmac:
@jgmize r?

jgmize:
code changes and unit test coverage lgtm, but do we have integration test coverage for this or do we need to do manual e2e testing?

pmac:
I've validated the call in my local instance to my satisfaction against basket prod. I'm happy to push a demo but this seems to work and isn't a high priority thing. We don't have any automated tests that actually call to basket so that we don't gum up SFDC with a bunch of test contacts. We can and probably should look into that and basket is in a better place now to mark testing as such. Since it looks okay and works for me I'd be happy if we did a spot check on stage after merge.

jgmize:
> 

I've validated the call in my local instance to my satisfaction against basket prod. I'm happy to push a demo but this seems to work and isn't a high priority thing. We don't have any automated tests that actually call to basket so that we don't gum up SFDC with a bunch of test contacts. We can and probably should look into that and basket is in a better place now to mark testing as such. Since it looks okay and works for me I'd be happy if we did a spot check on stage after merge.

fair enough. r+, feel free to merge and spot check on stage when you have the time.

pmac:
Thanks! 🎆

pmac:
Tested and working on stage. Thanks again @jgmize. It's ready for prod any time.

PR #4200, opened 2016-07-01
Don't generate headless tests when running IE SauceLabs jobs

alexgibson:
This should avoid generating headless tests before collection when they're not needed. e.g:

https://ci.us-west.moz.works/job/bedrock_integration_tests_runner/6822/


alexgibson:
@jgmize - r? 

As far as I can tell this should work (locally I ran `-m "sanity and not headless"  --collect-only`)

PR #4201, opened 2016-07-01
[fix bug 1283208] Updates to home page concept A

craigcook:
## Description
A few updates to home page concept A:
* update nav to "Powered by People"
* update copy in Firefox section
* add image to intro slider
* move/animate survey message

The survey message pops up from the bottom when you scroll down to the Firefox section, and slides down again if you scroll back up.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1283208


craigcook:
@jpetto r? when you have a moment. And please let me know if any of the survey changes here are going to really screw with concept B and we can find a happy medium.

media/js/mozorg/home/home-a.js
```diff
@@ -115,6 +116,25 @@
...
         });
+
+        surveyWaypoint = new Waypoint({
+            element: '#firefox',
+            handler: function(direction) {
+                var $surveyMsg = $('#survey-message');
+                if ($surveyMsg.length) {
+                    if (direction === 'down') {
+                        // slide up when scrolling down
+                        $surveyMsg.addClass('stuck').css({ bottom: '-' + $surveyMsg.height() + 'px' }).animate({ bottom: '0' }, 500);
```
jpetto: Instead of asking the browser to calculate the height of the element (which is expensive), I'd recommend just picking a large enough static value and sticking to that. Looks like the element is about 38px tall, so if you just say the offset is like 90px, you'd be pretty safe at both one and two lines of text.

media/js/mozorg/home/home-a.js
```diff
@@ -115,6 +116,25 @@
...
+                    } else if (direction === 'up') {
+                        // slide down when scrolling up, then unstick
+                        $surveyMsg.animate({ bottom: '-' + $surveyMsg.height() + 'px' }, 500, function() {
+                            $surveyMsg.removeClass('stuck').removeAttr('style');
```
jpetto: As removing the `stuck` class reverts to `display: none;`, I don't think you need to worry about removing the `style` attribute.

jpetto:
A couple small performance suggestions, but otherwise looking good. r+wc

craigcook:
I think we're going to do a final quick review on a demo server before merging this though I don't expect any significant changes. Marking Do Not Merge for the moment.

jpetto:
LGTM! 📟

PR #4202, opened 2016-07-04
Update Selenium client to v2.53.6

alexgibson:
## Description
- Updates Selenium client to 2.53.6 (fixes an [issue running tests in Firefox 47.0](https://github.com/SeleniumHQ/selenium/issues/2110)).

## Testing
- Should work again in Firefox 47.0.1.


alexgibson:
@davehunt r?

davehunt:
r+ though you'll also need to update the server component to run tests against Firefox 47 in the pipeline, and at present this is not available via the docker images.

alexgibson:
> r+ though you'll also need to update the server component to run tests against Firefox 47 in the pipeline, and at present this is not available via the docker images.

We're currently [pinned to Firefox 45](https://github.com/mozilla/bedrock/blob/master/docker/jenkins/properties/integration_tests/per_push/firefox.properties#L3) in our automation and I don't expect this to change before we switch to Marionette. I believe this is the same for the docker images.

This is mainly just to get local testing working again, but with the above in mind, are there any other reasons why updating the client here could be problematic?

davehunt:
Nope, updating the client is a good idea, and Firefox 45 is ESR so it still supported by Selenium. I'm hoping to get the docker images updated once I can include Marionette support, which should make switching relatively easy.

PR #4208, opened 2016-07-07
Skip partnerships form submission test

alexgibson:
## Description
- Skips the `/about/partnerships` form submission test since this is now routed to a Google Group and can't be filtered. When time allows I'd like to find a solution (similar to our basket functional tests perhaps) that allow us to non-destructively test this form so we don't lose coverage for this page.

### DIFF
--- a/tests/functional/test_partnerships.py
+++ b/tests/functional/test_partnerships.py
@@ -8,6 +8,7 @@ from selenium.common.exceptions import TimeoutException
 from pages.partnerships import PartnershipsPage
 
 
+@pytest.mark.skipif(reason='Need to find a non-destructive solution for testing submission of form')
 @pytest.mark.nondestructive
 def test_request_partnership(base_url, selenium):
     page = PartnershipsPage(selenium, base_url).open()


alexgibson:
@pmac r?

Would also appreciate your thoughts on what we might be able to do longer term here to re-enable the test.

pmac:
What if instead of skipping the test we put a short circuit in the view code where if the email address submitted is our standard `success@example.com` then it does everything but send the email? That'd be pretty easy and we could just do it instead.

alexgibson:
> What if instead of skipping the test we put a short circuit in the view code where if the email address submitted is our standard success@example.com then it does everything but send the email? That'd be pretty easy and we could just do it instead.

I would like to do this very much, but it might have to wait until I have more time (probably next week)

pmac:
I can knock it out now. I'll have a PR shortly.

alexgibson:
> I can knock it out now. I'll have a PR shortly. 

Thanks @pmac - I owe you 🍺

alexgibson:
Closing in favor of #4210

PR #4210, opened 2016-07-07
Send no email for @example.com addresses

pmac:
Partnership form should not send the email if the submitted address is at example.com. This is a reserved testing domain and will never have valid email accounts. This is specifically so that the selenium tests for this page will not send mail to the list for these tests.


pmac:
@alexgibson look okay? I think this should work.

pmac:
potential replacement for #4208.

alexgibson:
Looks good to me, thanks for the quick assist @pmac! r+ (just waiting on Circle)

PR #4212, opened 2016-07-08
[fix bug 1282732] Redesign /plugincheck page

alexgibson:
## Description
- Redesigned /plugincheck page to focus on updating Flash / Silverlight and encouraging users to update their Firefox if out of date.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1282732

## Testing
https://bedrock-demo-agibson.us-west.moz.works/en-US/plugincheck/
- Test with & without Flash / Silverlight installed.
- Test out of date Firefox.
- Test non-Firefox.
- Test with out of date / vulnerable plugin status (I used mock data rather than spend time finding old plugins. Can assist with this if needed).
- Test no plugins.
- Test JS disabled.
- Test RTL and non English locales.

## Checklist
- [x] Requires l10n changes.


alexgibson:
Still need to confirm that strings are final, but will hopefully do that today.

alexgibson:
@peiying2 - Got confirmation in the bug that strings here are final.

alexgibson:
Adding a do-not-merge just until we get confirmation that strings all look OK for l10n, but this is still considered "final" and good for code review.

bedrock/mozorg/templates/mozorg/plugincheck-redesign.html
```diff
@@ -0,0 +1,226 @@
...
+{% endblock %}
+
+{% block site_header_logo %}{% endblock %}
+
+{% block content %}
+<main role="main">
+  <header class="container">
+      <h1 class="large">{{_('Check Your Plugins')}}</h1>
+      <p class="tagline">{{_('Plugins power the videos, animations and games in your Firefox. Keep an eye on their status and keep them up to date.')}}</p>
+      <p>{{_('Be sure to restart your Firefox after making all recommended updates.')}}</p>
```
jpetto: Small nit - indentation.

bedrock/mozorg/templates/mozorg/plugincheck-redesign.html
```diff
@@ -0,0 +1,226 @@
...
+    {{ download_firefox(dom_id="out-of-date-download-button") }}
+  </section>
+
+  <section id="not-supported-container" class="message-box container hidden">
+    <h2>{{_('Currently the plugin check service is only available to Firefox users on desktop and Android.')}}</h2>
+    {{ download_firefox(dom_id="not-supported-download-button") }}
+  </section>
+
+  <section id="enable-js-container" class="message-box container">
+      <h2>{{_('Please turn on JavaScript to display this page correctly.')}}</h2>
```
jpetto: ➡️

media/css/plugincheck/plugincheck-redesign.less
```diff
@@ -0,0 +1,602 @@
...
+        margin-bottom: -120px;
+    }
+
+    #plugin-status-container {
+        padding-top: 120px;
+    }
+}
+
+#enable-js-container {
+    h2 {
```
jpetto: Nit - unnecessary nesting.

media/css/plugincheck/plugincheck-redesign.less
```diff
@@ -0,0 +1,602 @@
...
+    }
+
+    .status-unknown {
+        .plugin-status-detail {
+            width: 80%;
+        }
+    }
+
+    .status-up-to-date {
+        .plugin-status-status {
```
jpetto: Nits - unnecessary nesting here and above.

media/css/plugincheck/plugincheck-redesign.less
```diff
@@ -0,0 +1,602 @@
...
+        display: inline-block;
+
+        &:first-child {
+            margin-right: 1em;
+        }
+    }
+}
+
+.html-rtl #sec-plugin-featured {
+    .plugin-meta span:first-child {
```
jpetto: Nit - nesting.

media/css/plugincheck/plugincheck-redesign.less
```diff
@@ -0,0 +1,602 @@
...
+
+            @media only screen and (max-width: @breakDesktop) {
+                .font-size(24px);
+            }
+        }
+
+        // overrides styles from fxnewsletter-subscribe.less
+        #newsletter-form-thankyou {
+            float: none;
+            margin-left: auto;
```
jpetto: Superfluous `margin-left` here?

media/css/plugincheck/plugincheck-redesign.less
```diff
@@ -0,0 +1,602 @@
...
+        content:'\2630 ';
+    }
+
+    span.menurev:after {
+        content:' \2630';
+    }
+}
+
+html[lang|="en"] {
+    .plugin-faq-container h2 {
```
jpetto: Nesting?

alexgibson: Will fix the nesting, although sometimes I think readability becomes harder one one line (we can chew over that another time 😄 ).

jpetto:
Would the Fx logos in `/img/firefox/common/` work (as opposed to those added to `/img/firefox/template/`)?

No problem if not - just trying to mitigate moar logos.

alexgibson:
> Would the Fx logos in /img/firefox/common/ work (as opposed to those added to /img/firefox/template/)?

Ooh I didn't look in `/common`. Very much favor not adding more 👍

media/css/plugincheck/plugincheck-redesign.less
jpetto: Totally agree in some situations. That's why I put a question mark on this one - I began doubting my nit-picking. 😉 

I will not argue if you think it's more readable as-is. Mostly pointing out for consistency.

alexgibson: In some of the cases you called out here it is most certainly unnecessary still, so thanks for pointing out.

media/js/plugincheck/check-plugins-redesign.js
```diff
@@ -0,0 +1,327 @@
...
+        var outdatedPluginsBody = $('#plugin-outdated');
+        var outdatedPluginsHtml = '';
+        var unknownPluginsSection = $('#sec-plugin-unknown');
+        var unknownPluginsBody = $('#plugin-unknown');
+        var unknownPluginsHtml = '';
+        var upToDatePluginsSection = $('#sec-plugin-uptodate');
+        var upToDatePluginsBody = $('#plugin-uptodate');
+        var upToDatePluginsHtml = '';
+
+        // If the latest response from the service was a vulnerable plugin,
```
jpetto: Should be "was a featured plugin"?

jpetto:
screen shot 2016-07-08 at 12 07 36 pm

I know it follows the wireframes, and I don't mean to step on toes, but a couple things seem off about the layout here:

1. All text is centered until you get to the featured plugins. Not a huge deal, but looks a little sloppy.
2. The "All plugins" text/plus sign are really far apart (on desktop) and aren't obviously interactable (did I just make up a word?). Maybe the plus should be next to the "All plugins" text? The bright blue of the newsletter signup doesn't help things here, as the eye is drawn to the heavy color (and away from the itty little plus sign off to the right).

alexgibson:
Yeah we've had minimal design input here. I'll relay your comments to Holly (some but not all we've discussed already).

alexgibson:
The other concern I have currently is that we could be hiding a potentially vulnerable plugin behind an expandable section. I would like to not collapse it at all, but it might take some persuasion. Will see what I can make happen.

media/js/plugincheck/check-plugins-redesign.js
```diff
@@ -0,0 +1,327 @@
...
+        var featuredPluginsHtml = '';
+        var outdatedPluginsSection = $('#sec-plugin-outdated');
+        var outdatedPluginsBody = $('#plugin-outdated');
+        var outdatedPluginsHtml = '';
+        var unknownPluginsSection = $('#sec-plugin-unknown');
+        var unknownPluginsBody = $('#plugin-unknown');
+        var unknownPluginsHtml = '';
+        var upToDatePluginsSection = $('#sec-plugin-uptodate');
+        var upToDatePluginsBody = $('#plugin-uptodate');
+        var upToDatePluginsHtml = '';
```
jpetto: As this function is called multiple times, would it be better to declare the section & body vars outside (so they don't get re-declared on each call)?

Also, since there's only ever one chunk of HTML created per plugin, could we just use a single, general HTML container var instead of declaring one for each status?

alexgibson: Good suggestion :+1:

media/js/plugincheck/tmpl/plugincheck-redesign.ui.tmpl.js
```diff
@@ -0,0 +1,45 @@
+/* exported featuredPluginsTmpl, otherPluginsTmpl, unknownPluginsTmpl */
+
```
jpetto: Just out of curiosity, are these template strings generated by something (I see that `exported` comment above...)? Or do we code by hand?

jpetto:
Tested in all scenarios listed and didn't notice any visual/functional issues.

A few small cleanup suggestions, but nothing major. Hopefully we can fix up the UI just a little prior to merge.

Happy to give another quick/final(?) look on Monday.

media/js/plugincheck/tmpl/plugincheck-redesign.ui.tmpl.js
alexgibson: These are exceptions for ESLint parsing, which need to be generated manually. Really we shouldn't be passing round global variables like this, which is why it complains. I'm not going to the effort of fixing this here since it's only related to this one page (plugincheck is also to become deprecated in the future anyway).

alexgibson:
Cleanups made, although I still need to address the UI feedback once Holly is available a little later today.

alexgibson:
Pushed another commit that addresses some of the UI cleanup issues (after some discussion with Holly):

- Featured plugins heading is now left aligned and styled to be consistent with other plugin table headings
- Accordion expand icons (plus + signs) now use default positioning and styling.
- If a plugin in the collapsed "All Plugins" section has a class of 'vulnerable', show a status/plugin count in the accordion header to alert the user.

jpetto:
Code changes are 👌. UI is definitely improved. r+

jpetto:
We ready to merge after a squash?

alexgibson:
I think this has been reviewed enough by those involved already so I think we're pretty safe to merge. Squashed! 👍

PR #4214, opened 2016-07-08
Improve handling of basket error getting newsletter data

pmac:
* Save a copy of data returned by https://basket.mozilla.org/news/newsletters/
* No longer cache newsletter languages, rely solely on newsletter data cache
* Upgrade Mock to 2.0 to get new methods like assert_called_once


pmac:
@jgmize r?

cc @alexgibson

bedrock/firefox/tests/test_views.py
```diff
@@ -135,6 +136,7 @@ def test_variant_android_invalid_test_value(self):
                                                lang='en-US')
 
 
+@override_settings(DEV=False)
```
pmac: This is here because this test failed for me locally when I had `DEV=True` in my `.env` file.

bedrock/newsletter/utils.py
```diff
@@ -76,3 +72,22 @@ def custom_unsub_reason(token, reason):
...
     return basket.request('post', 'custom_unsub_reason', data=data)
+
+
+def get_local_basket_newsletters_data():
+    """
+    Load newsletter data from a file previously saved from basket
+
+    To update the data run the following:
+
+        $ curl https://basket.mozilla.org/news/newsletters/ > bedrock/newsletter/basket_data.json
```
jgmize: Not a blocker for this PR, but I think we're going to need to either automate the updates of the json file in version control or have a documented manual update schedule so that we don't get too far out of sync.

jgmize:
code lgtm other than my non-blocking comment about keeping `basket_data.json` up to date, and I believe we have sufficient unit test coverage, but as with other basket-related PRs where we don't have sufficient integration test coverage I think we should wait to merge this until we can do manual testing right after this lands on stage.

r+, nice work @pmac

PR #4215, opened 2016-07-11
[fix bug 1282718] Create animated version of /new forest download page

alexgibson:
## Description
- Enables animation for the /new download page.
- Animation is being A/B tested against the static version to make sure it does not distract users from clicking the green button.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1282718

## Testing
Demo: https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/?v=1
- Animated version is accessible via `?v=1` on scene1.
- Animation uses progressive enhancement via `@supports`, so it only runs on modern browsers. I've tested in latest Firefox, Chrome, Safari, Edge and Opera and seems good.

## Checklist
- [ ] Requires l10n changes.


media/css/firefox/new/horizon/scene1.less
```diff
@@ -8,94 +8,106 @@
...
+ * Specific CSS applied when unprefixed animations, unprefixed transformations
+ * and CSS filter effects are supported.
+ */
+@supports (animation-name: sunrise) and (transform:translateY(120px)) and (filter:brightness(60%)) {
+    .horizonAnimation;
+}
+
+/**
+ * We need to repeat this block as YUI Compressor does not handle 'or' conditionals
+ * correctly when minifying :(
```
jpetto: Bummer!

jpetto:
First of all, this animation looks spectacular. 👏 👏 👏 

Code all looks good, too. Pretty simple way to achieve a really, really nice result.

My only (personal, non-scientific) suggestion would be to speed up the first half of the animation. I think the timing on the highlights is perfect, but before that could use almost a 2x speed increase.

Just a thought. Nothing blocking merging as far as I'm concerned. Happy to push the green button if we're ready. 😄

alexgibson:
Thanks @jpetto!

> My only (personal, non-scientific) suggestion would be to speed up the first half of the animation. I think the timing on the highlights is perfect, but before that could use almost a 2x speed increase.

After what it took to get this animation approved from both Ty and people higher up in creative, I'm not going to touch it any further (with a barge-pole). But you comments are welcome nonetheless 😆 

I'm good for this to merge if all else looks ok on your end 👍

PR #4216, opened 2016-07-11
Fix Bug 1285654 - Updates to policies and notices

kyoshino:


alexgibson:
Thanks again @kyoshino! r+ 🍃

PR #4218, opened 2016-07-12
[fix bug 1275629] Add survey to home prototype B.

jpetto:
## Description

Adds a survey to homepage prototype B. 

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1275629#c15

## Testing

Survey should appear after scrolling to the Firefox section. Survey should not be shown below 760px wide. Make sure no regressions on prototype A survey.

## Checklist

### DIFF
--- a/bedrock/mozorg/templates/mozorg/home/home-b.html
+++ b/bedrock/mozorg/templates/mozorg/home/home-b.html
@@ -262,4 +262,7 @@
 
 {% block js %}
   {% javascript 'home-b' %}
+
+  {# survey is en-US only - enforced via optimizely audience segmentation #}
+  {% include 'mozorg/home/includes/survey-b.html' %}
 {% endblock %}
--- a/bedrock/mozorg/templates/mozorg/home/includes/survey-b.html
+++ b/bedrock/mozorg/templates/mozorg/home/includes/survey-b.html
@@ -1,0 +1,6 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% stylesheet 'home-survey' %}
+{% javascript 'home-survey-b' %}
--- a/media/js/mozorg/home/home-b.js
+++ b/media/js/mozorg/home/home-b.js
@@ -2,10 +2,13 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-(function($) {
+(function($, Waypoint) {
     'use strict';
 
     var mozClient = window.Mozilla.Client;
+
+    var surveyWaypoint;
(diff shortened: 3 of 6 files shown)


bedrock/mozorg/templates/mozorg/home/home-b.html
```diff
@@ -262,4 +262,8 @@ <h1 class="logo"><a href="{{ url('mozorg.home') }}" data-link-type="footer" data
 
 {% block js %}
   {% javascript 'home-b' %}
+
+  {% if LANG == 'en-US' %}
```
alexgibson: Is this conditional needed seeing as the prototype is shown in English only?

media/js/mozorg/home/home-b.js
```diff
@@ -32,6 +54,11 @@
         $('#fx-download-link').addClass('hidden');
     }
 
+    // test for matchMedia
+    if ('matchMedia' in window) {
```
alexgibson: Could this block be inside the document ready handler below? `mqIsTablet` doesn't seem to be used anywhere else (the var could be defined there too?).

alexgibson:
A couple of small suggestions / questions but looks good to me. r+wc 📄

bedrock/mozorg/templates/mozorg/home/includes/survey-b.html
```diff
@@ -0,0 +1,6 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% stylesheet 'home-survey' %}
```
alexgibson: I now realize the other homepage survey is already doing this too. Up to you if you want to address this or not (it's more of a query as I don't see any real issues).

bedrock/mozorg/templates/mozorg/home/home-b.html
jpetto: Hm, there is a locale restriction, but it's in Optimizely (not in the view). I suppose this could be removed for now.

bedrock/mozorg/templates/mozorg/home/includes/survey-b.html
jpetto: I wondered this as well. I don't *think* adding a stylesheet here will trigger a reflow, as it doesn't contain rules impacting existing elements. Once the JS is executed, however, a reflow is likely due to a new element being injected into the DOM.

In other words, I don't think the location of the `<link>` tag has an effect, but the JS certainly does.

media/js/mozorg/home/home-b.js
jpetto: Good call.

bedrock/mozorg/templates/mozorg/home/includes/survey-b.html
alexgibson: Ok, let's leave it be 👍

PR #4222, opened 2016-07-15
WIP [fix bug 1287200] Begin restricting JS served to old IE.

jpetto:
## Description

Opening for discussion purposes only at this point.
## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1287200
## Testing

For now, only pages updated are:
- / (IE <= 8 gets full JS)
- /firefox/channel/ (IE <= 8 gets minimal JS)
- /firefox/ios/ (IE <= 8 gets minimal JS)
## Checklist

### DIFF
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -168,7 +168,7 @@
       The logic mess below will be simplified when all pages are refactored
       to behave well in old IE without page-specific JS.
     #}
-    {% if OLDIEJS != True %}
+    {% if force_modern_js != True %}
       {#
         Unless specifically requested by current page, hide first-class
         & page-specific JS from IE 8 and below.
@@ -179,7 +179,7 @@
       {% javascript 'common' %}
     {% endblock %}
     {% block js %}{% endblock %}
-    {% if OLDIEJS != True %}
+    {% if force_modern_js != True %}
       <!--<![endif]-->
 
       {#
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -161,13 +161,36 @@
     </div><!-- close #outer-wrapper -->
 
     <!--[if IE 9]>
-      <script src="{{ static('js/libs/matchMedia.js') }}"></script>
+      {% javascript 'matchmedia' %}
     <![endif]-->
 
+    {#
+      The logic mess below will be simplified when all pages are refactored
+      to behave well in old IE without page-specific JS.
+    #}
+    {% if OLDIEJS != True %}
+      {#
+        Unless specifically requested by current page, hide first-class
+        & page-specific JS from IE 8 and below.
+      #}
+      <!--[if !lte IE 8]><!-->
+    {% endif %}
     {% block site_js %}
       {% javascript 'common' %}
     {% endblock %}
     {% block js %}{% endblock %}
+    {% if OLDIEJS != True %}
+      <!--<![endif]-->
+
+      {#
+        Give old IE a slimmed down bundle with only jQuery, global, and
+        GTM scripts.
+      #}
+      <!--[if lte IE 8]>
+        {% javascript 'common-oldie' %}
+      <![endif]-->
+    {% endif %}
+
     {% if settings.DEV %}
       <script src="https://pontoon.mozilla.org/pontoon.js"></script>
     {% endif %}
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -164,32 +164,18 @@
       {% javascript 'matchmedia' %}
     <![endif]-->
 
-    {#
-      The logic mess below will be simplified when all pages are refactored
-      to behave well in old IE without page-specific JS.
-    #}
-    {% if force_modern_js != True %}
-      {#
-        Unless specifically requested by current page, hide first-class
-        & page-specific JS from IE 8 and below.
-      #}
-      <!--[if !lte IE 8]><!-->
-    {% endif %}
-    {% block site_js %}
-      {% javascript 'common' %}
-    {% endblock %}
-    {% block js %}{% endblock %}
-    {% if force_modern_js != True %}
-      <!--<![endif]-->
+    {# Serve first-class & page-specified JS to IE 9 and up #}
+    <!--[if !lte IE 8]><!-->
+      {% block site_js %}
+        {% javascript 'common' %}
+      {% endblock %}
+      {% block js %}{% endblock %}
+    <!--<![endif]-->
 
-      {#
-        Give old IE a slimmed down bundle with only jQuery, global, and
-        GTM scripts.
-      #}
-      <!--[if lte IE 8]>
-        {% javascript 'common-oldie' %}
-      <![endif]-->
-    {% endif %}
+    {# Serve minimal JS to IE 8 and below #}
+    <!--[if lte IE 8]>
+      {% javascript 'common-oldie' %}
+    <![endif]-->
 
     {% if settings.DEV %}
       <script src="https://pontoon.mozilla.org/pontoon.js"></script>
--- a/bedrock/base/templates/base.html
+++ b/bedrock/base/templates/base.html
@@ -150,7 +150,7 @@
       The logic mess below will be simplified when all pages are refactored
       to behave well in old IE without page-specific JS.
     #}
-    {% if OLDIEJS != True %}
+    {% if force_modern_js != True %}
       {#
         Unless specifically requested by current page, hide first-class
         & page-specific JS from IE 8 and below.
@@ -161,7 +161,7 @@
       {% javascript 'common' %}
     {% endblock %}
     {% block js %}{% endblock %}
-    {% if OLDIEJS != True %}
+    {% if force_modern_js != True %}
       <!--<![endif]-->
 
       {#
--- a/bedrock/base/templates/base.html
+++ b/bedrock/base/templates/base.html
@@ -146,10 +146,33 @@
 
     </div><!-- close #outer-wrapper -->
 
+    {#
+      The logic mess below will be simplified when all pages are refactored
+      to behave well in old IE without page-specific JS.
+    #}
+    {% if OLDIEJS != True %}
+      {#
+        Unless specifically requested by current page, hide first-class
+        & page-specific JS from IE 8 and below.
+      #}
+      <!--[if !lte IE 8]><!-->
+    {% endif %}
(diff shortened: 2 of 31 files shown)


bedrock/base/templates/base-resp.html
```diff
@@ -167,13 +167,36 @@
     </div><!-- close #outer-wrapper -->
 
     <!--[if IE 9]>
-      <script src="{{ static('js/libs/matchMedia.js') }}"></script>
+      {% javascript 'matchmedia' %}
```
jpetto: Nested conditional comments don't seem to work, meaning pages that have a conditional addition of `matchmedia.addListener` break the new logic in `base-resp.html`. `base-new.html` has both `matchMedia` and `matchMedia.addListener` bundled, which I'm totally in favor of and implementing here.

bedrock/firefox/urls.py
```diff
@@ -32,7 +32,7 @@
     url(r'^firefox/(?:%s/)?(?:%s/)?all/$' % (platform_re, channel_re),
         views.all_downloads, name='firefox.all'),
     page('firefox/accounts', 'firefox/accounts.html'),
-    page('firefox/channel', 'firefox/channel.html'),
+    page('firefox/channel', 'firefox/channel.html', OLDIEJS=True),
```
jpetto: One pain point of this approach will be adding this context var for *all* pages that haven't yet been tested without JS in old IE. I went this way so we could remove code as we went along, as opposed to adding code to opt pages in to the new, minimal JS way. Happy to hear some opposing thoughts here.

jpetto:
Would love some feedback on approach from @alexgibson, @craigcook, and @pmac. 😁

alexgibson:
I'm likely just missing a piece of the conversation that goes with this change, but what's the rationale behind going this route as opposed to simply no longer serving these pages JS? From our recent discussions I thought we we're planning on deprecating JS support, as opposed to some form of degraded support which is what this seems to be. I'm not opposed to the latter, but I think I need to understand the reason why we can't drop JS support entirely?

The reason I'm asking is because we often go to good lengths to make sure pages look "OK" with JS turned off. The /channel page for example seems more accessible with no JS than it does with just a common bundle for oldIE. Of course, this may well not be the case with all pages as I've not delved too far. But I'm curious what we're gaining by degraded support (this would likely be a good discussion).

jpetto:
My assumption is that we'll need to serve at least some JS to old IE for two reasons:

1) analytics
2) `triggerIEDownload` in `global.js`

I doubt we can get rid of #1, but am unsure on #2. I'll post a question in the bug and NI jbertsch.

I just pushed a commit that removes page JS on `/firefox/channel`. It's not quite as accessible as with no JS (mostly the menu and download buttons), but these could be remedied fairly easily, depending on decisions about minimal scripts.

alexgibson:
> 1) analytics

We would still load the GTM snippet for older browsers though, and since we rely on data attributes in HTML for most link clicks, this should continue to work as expected (I'm guessing though?) The only thing we'd lose for sure would be custom JS calls to `dataLayer.push()`, but given these are mostly in the page JS anyway losing those is probably unavoidable.

2) triggerIEDownload in global.js

Yeah, if we served the noJs view they would get the list of platform buttons, and have to click the one that says "Windows". This might be ok though? I guess we can't make the call here. Scene 2 of /firefox/new should degrade gracefully for no-JS now, and show the buttons when needed.

alexgibson:
I guess the other thing we'd lose re: analytics is the recently added core data layer stuff (which I forgot about in my previous comment) 😢 . I don't know how important this is though, in terms of the numbers our stats would be losing.

jpetto:
Yeah, I just noticed those new core scripts when starting this branch. I'll add pg to the bug to get his opinion on keeping/losing those for IE 8 and below.

jpetto:
So it has been requested that we keep the GA/GTM scripts, but we do have the go-ahead to test dumb download links.

I've updated this PR to reflect the above. The home page should still get full JS/CSS in IE 8, while `/firefox/channel/` and `/firefox/ios/` should get minimal JS/CSS (with dumb download links).

If/when we deem this approach move-forward worthy, I'll open a separate PR to add a variation on a TBD page to test smart vs dumb download buttons.

jpetto:
Okay, if the current direction is approved, the next step is to run a test on IE 8 pitting smart download buttons against the "dumb"/no-`js` class on `<html>` links.

@alexgibson @craigcook - Do you think we're ready to move forward here, or should we discuss the approach further?

To summarize: 

- We need to keep GA/GTM JS, which requires keeping various other high-level scripts (`site.js`, `mozilla-client.js`, `global.js`, `global-init.js`).
- We'll need to specifically flag any page (using a `context` variable) that needs to (temporarily) retain full JS support.
- For old IE, we will serve a custom `common` JS bundle that does re-use some scripts from the A-grade `common` bundle - meaning future changes/support on these shared scripts will need to keep IE 8 in mind.
- We'll be able to upgrade jQuery for A-grade browsers, which will result in a decent performance increase.

bedrock/base/templates/base-resp.html
alexgibson: This sounds good to me 👍

bedrock/settings/static_media.py
```diff
@@ -1000,7 +1000,7 @@
     # bundle should offset the extra weight.
     'common': {
         'source_filenames': (
-            'js/libs/jquery-1.11.3.min.js',
+            'js/libs/jquery-1.11.3.min.js',  # can now upgrade to jQuery 3?
```
alexgibson: We'll need to do a lot more testing with jQuery 3 to make sure there are no regressions as quite a lot has changed. But in theory this does indeed make it possible 👍

bedrock/settings/static_media.py
```diff
@@ -1013,6 +1013,18 @@
         ),
         'output_filename': 'js/common-bundle.js',
     },
+    'common-oldie': {
+        'source_filenames': (
+            'js/libs/jquery-1.11.3.min.js',  # upgrade to jQuery 1.12?
+            # will need to make sure scripts below are compatible with jQuery 1.12 & 3?
```
alexgibson: This is the only thing I think we'll need to be careful about. I do wonder if we can cut this down even further maybe. If we're showing dumb download buttons to old IE, could we theoretically cut out `global.js, `global.init.js` and `mozilla-client.js`?

media/js/base/site.js
```diff
@@ -189,7 +189,9 @@
             h.className += ' x64';
         }
 
-        // Add class to reflect javascript availability for CSS
-        h.className = h.className.replace(/\bno-js\b/, 'js');
+        // Add class to reflect javascript availability for CSS (for IE 9 and above)
+        if (!(platform === 'windows' && /MSIE\s[1-8]\./.test(navigator.userAgent))) {
```
alexgibson: Big thumbs up 👍

bedrock/base/templates/base.html
```diff
@@ -150,10 +150,33 @@
 
     </div><!-- close #outer-wrapper -->
 
+    {#
+      The logic mess below will be simplified when all pages are refactored
+      to behave well in old IE without page-specific JS.
+    #}
+    {% if force_modern_js != True %}
```
alexgibson: Does this conditional become somewhat redundant, seeing as we're keeping the `no-js` class set for IE < 9 in `site.js`? For example, the homepage has `'force_modern_js': True` in the view, yet it still gets the no-JS styling. Could this potentially lead to breakages introduced by the page JS expecting the `js` class to have been applied?

I think as long as pages look OK with `no-js` (and most do?), then would we even need this? This could simplify a fair bit of cruft in the templates/views, and also puts us further toward dropping support for realz.

Thoughts?

bedrock/base/templates/base-resp.html
```diff
@@ -167,13 +167,36 @@
...
     <![endif]-->
 
+    {#
+      The logic mess below will be simplified when all pages are refactored
+      to behave well in old IE without page-specific JS.
+    #}
+    {% if force_modern_js != True %}
```
alexgibson: Does this conditional become somewhat redundant, seeing as we're keeping the `no-js` class set for IE < 9 in `site.js`? For example, the homepage has `'force_modern_js': True` in the view, yet it still gets the no-JS styling. Could this potentially lead to breakages introduced by the page JS expecting the `js` class to have been applied?

I think as long as pages look OK with `no-js` (and most do?), then would we even need this? This could simplify a fair bit of cruft in the templates/views, and also puts us further toward dropping support for realz.

Thoughts?

alexgibson:
One final note: I'm not sure if this PR is intended to merge or not, but if we do go forward with this approach we'll also need to move our sanity tests up to IE9, as they're currently set to run in IE6 & 7 here: https://github.com/mozilla/bedrock/tree/master/docker/jenkins/properties/integration_tests/per_tag

Once we do this, we may want to consider some base "no-JS" tests, which we could run in older IE's?

bedrock/settings/static_media.py
jpetto: Well...maybe. If we want to hide the footer lang chooser from IE 8 and below (or just accept that it wont work), we can get rid of `global.js` and `global-init.js`. I'd be fine hiding the lang chooser from old browsers. Thoughts?

`mozilla-client.js` is required by `core-datalayer-init.js`, though I could pretty easily update `core-datalayer-init.js` to function without error if `Mozilla.Client` is unavailable.

I'll push a commit with these changes shortly.

alexgibson: > Well...maybe. If we want to hide the footer lang chooser from IE 8 and below (or just accept that it wont work), we can get rid of global.js and global-init.js. I'd be fine hiding the lang chooser from old browsers. Thoughts?

The footer lang switcher works without JS, but looks like it relies on `<noscript>`. Could we change this to use the `no-js` class instead?

jpetto: Oh, good call. Even better. 😀

bedrock/base/templates/base-resp.html
jpetto: The only major instance where I see this needed is on the `/contact/` pages, which probably means refactoring those pages to work well without JS (in IE <= 8) should become a required part of this PR.

I'll remove these conditionals and make a note that the `/contact/` pages need some work.

alexgibson: If this means we can avoid the conditional then I'm all for making the `/contact/` pages look ok without JS. We could retire the oldIE JS bundle for those pages 👍

jpetto:
Closing in favor of #5324.

PR #4225, opened 2016-07-18
Bug 1156344: Switch from Tower to Puente for i18n

pmac:
First step on the way to the Django 1.8.x upgrade.


pmac:
@jgmize r?

I've also pushed this to [my demo](https://bedrock-demo-pmac.us-west.moz.works). Main thing is to be sure translation still works :)

bedrock/newsletter/views.py
```diff
@@ -58,13 +58,13 @@
         'description': _lazy(u'Bringing you the best new opportunities to help support the open Web.'),
         'title': _lazy(u'Mozilla Communities')},
     u'about-standards': {
-        'description': _lazy(u''),
+        'description': u'',
```
jgmize: I'm going to be lazy and just ask: why does this no longer need to be lazy?

lib/l10n_utils/dotlang.py
```diff
@@ -108,7 +105,7 @@ def translate(text, files):
     if lang == settings.LANGUAGE_CODE:
         return Markup(text)
 
-    tweaked_text = tweak_message(text)
+    tweaked_text = strip_whitespace(text)
```
jgmize: Why isn't this using `tweak_message` from `lib.l10n_utils.extract`?

lib/l10n_utils/extract.py
```diff
@@ -0,0 +1,61 @@
+# mostly borrowed from tower
```
jgmize: Any chance we can borrow tests for these functions from tower as well, or would you be willing to add some?

jgmize:
running https://ci.us-west.moz.works/view/Bedrock/job/bedrock_integration_tests_runner/7208/ and https://ci.us-west.moz.works/view/Bedrock/job/bedrock_integration_tests_runner/7209/ against your demo now

lib/l10n_utils/dotlang.py
pmac: Good catch. This was a change from an earlier iteration of my attempt. I'll update.

pmac: Oh, nm, this is on purpose. `tweak_message` is specifically for the extract stuff because it has handles for the pluralization machinery. If you pass just a string to `tweak_message` it just calls `strip_whitespace` and returns, so these are functionally identical and this way is one fewer function call and a bit more straightforward code path.

lib/l10n_utils/extract.py
pmac: I'll check.

jgmize:
One [failure in the headless tests](https://ci.us-west.moz.works/view/Bedrock/job/bedrock_integration_tests_runner/7209/testReport/junit/tests.redirects/test_redirects/test_global_conf_url_693_firefox_pocket_/) will likely be fixed by a rebase, and we got some timeouts in the selenium tests that likely have more to do with the demo only being scaled to one web container than code changes; I'm rerunning it in https://ci.us-west.moz.works/view/Bedrock/job/bedrock_integration_tests_runner/7218/

jgmize:
the rerun failed with the same error; turns out it wasn't just a random timout. @alexgibson pointed out in #www that this looks like the available languages bug that was fixed in master, so a rebase would likely fix that as well.

pmac:
I added tests, rebased on master, and pushed to `demo__pmac`. 🎉

jgmize:
rerunning the [selenium](https://ci.us-west.moz.works/view/Bedrock/job/bedrock_integration_tests_runner/7219/) and [headless](https://ci.us-west.moz.works/view/Bedrock/job/bedrock_integration_tests_runner/7220/) tests against the freshly deployed demo.

jgmize:
All tests passed, code lgtm, and spot checking translated pages against the deployed demo everything looks consistent with prod.

pmac:
Okay. I'm pretty confident that this is ready then. I've rebased and checked through the code again. If you agree I think we're ready.

jgmize:
Nice work, @pmac 👍

PR #4226, opened 2016-07-18
[no bug] Refactor pebbles framework

craigcook:
## Description
This restructures and reorganizes the pebbles framework to make it more modular and extensible. The idea is to break it down into the smallest practical pieces so any given page can just `@import` the bits it needs. We'll still need some global baseline for the whole site but we don't have to load a monolithic framework on pages that don't need it. This PR establishes that baseline and sets us up to add new components as we create them.

Marking this as WIP to start some discussion with @jpetto, @alexgibson, @schalkneethling, and @stephaniehobson (this is heavily influenced by [stolen from] @stephaniehobson's work on MDN and View Source).

Some topics for discussion:

1. Should base elements like tables, forms, and buttons be part of the global styling or just imported as needed? E.g. if a page doesn't have any tables on it, do we just omit those styles? But having basic styling for all the general elements (not designed components) might be a good idea anyway if it doesn't add too much bloat. Similarly, I expect pretty much every page will have a button or two so maybe globally including button styles is better in the end. What do you think?

2. Should Fira Sans be global? If we don't include it in global.less we need to make sure it gets `@import`ed for any page that uses it (which will be almost every page). However, just on the off chance we have a page that doesn't use Fira Sans, leaving it out of the global style means we can avoid loading a bunch of fonts on a page that doesn't need them. For example, if we were to port a sandstone page to pebble but wanted to keep it Open Sans. It would mean reworking how we declare fonts but that's not impossible, I just wonder if it's worth the effort if we don't expect to need it. 

3. Should we break lib.less into separate files for variables and mixins? We did that in sandstone way back but found that we were always importing both, so merged them into one file. But with this more fragmented approach we could easily separate variables and mixins again and have a single lib.less that imports the other two. We could even break them further apart and have different variables and mixins in different files (colors.less and media-queries.less, for instance). Thoughts?

A few things still to do but not blocking this from merging:
* button styling (importing buttons.less directly from sandstone brings some unwelcome baggage with it but I don't want to redundant styles either).
* mobile menu styling (probably needs some design direction, and probably refactoring the JS a bit).
* start documenting this stuff in the style guide.


alexgibson:
I haven't really dug into this PR yet (past reading the above), but if I had one high-level wish above everything it would be to make `buttons.less` more modular? Currently, with Sandstone it gets included on every page by default which isn't ideal. But even then, sometimes on a page all you want is a simple button and not necessarily a full-fat download button. It would be great to have control over this on a per-page basis. Something like `buttons.less` and `download-buttons.less` perhaps? Just an idea though

craigcook:
Another topic for bikeshedding: there has been some previous discussion of switching to Stylus or even Sass instead of Less.

A few years ago Less seemed to be stagnating and we weren't sure it would last. Since then it's become much more active and there have been more frequent updates, so it's caught up to the competition. There's no longer a compelling reason to jump ship unless another preprocessor is clearly better.

I really like Stylus and would probably choose it for a new project if I was starting from zero, but it's also still a bit on the fringe so I might be a little gun-shy about using it on mozorg. It does seem to be growing and active so it's still worth considering.

Sass was previously rejected because it required Ruby, but now there's node-sass so it may be a viable option for us. Sass has the benefit of being the clear leader in preprocessors with a really active community and lots of resources and plugins. It's also something of a de facto standard so contributors and new hires are more likely to come to us with preexisting Sass knowledge over Less or Stylus.

All three are fairly equal, feature-wise, so the choice really comes down to other factors beyond just how they work. I'm inclined to stick with Less from sheer inertia. Switching would mean some back-end work to bake Sass/Stylus into our build process, and would mean continuing to maintain Less for a while (possibly indefinitely, but we can expect pieces of Sandstone to stick around for at least a few years). Or we could put some effort into porting all of our existing CSS to something new, but that may be harder than adding a second preprocessor (though there are tools and conversion scripts that can help). 

All that said, if there's a clear benefit to switching in the face of all the costs, now would be the time to do it. Converting Pebbles to another preprocessor wouldn't be too hard at this early stage but would become harder over time as we keep building on top of it.

Discuss! 🚲 🏡

jgmize:
I haven't used it myself yet, but [PostCSS](http://postcss.org/) seems to have some [distinct advantages](https://ai.github.io/about-postcss/en/) and can be used [together with less](http://webdesign.tutsplus.com/tutorials/using-postcss-together-with-sass-stylus-or-less--cms-24591) which could enable a smooth transition.

jpetto:
Like we do with Sandstone, I think it makes sense to shoot for a cache-friendly bundle of common CSS. This *probably* means base rules for tables, forms, links, etc. A little experimentation may be required to find our sweet spot - e.g. how many pages actually need `tables.less` versus file size increase.

As @alexgibson suggested, I think we can do a better job in regards to buttons - there's no reason to include the (comparably heavy) download button CSS on all pages.

I'm well in favor of the approach here of breaking related rules into separate `.less` files. Knowing when to split a single `.less` file into multiple is kind of a smell test thing. Is it hard to find stuff in the file? Sandstone's `lib.less` is ~550 lines, which I would say could use a refactor.

In terms of fonts, I don't think it hurts to include the Fira Sans CSS on every page, as the actual font files are only downloaded if required by an element on the page.

I'm also on board with trying out/switching to PostCSS. It's a little scary giving up control, but letting a robot make decisions on vendor prefixes and features support is probably a smart idea.

media/css/pebbles/includes/lib.less
```diff
@@ -243,12 +311,8 @@
...
 
 
-
-/****************************************************************************/
-// Classes used to hide content from users. Choose wisely!
-
-// Hide visually, but not to screen readers
-.visually-hidden() {
+// Hide an element visually, but not from screen readers
+.a11y-hidden() {
```
alexgibson: Could the name `a11y-hidden` sound a bit confusing, in that it could be interpreted as an element being hidden from screen readers? (whereas it is actually the reverse).

alexgibson:
In terms of pre-processors, my personal view is that they are all pretty much the same means to a common end. For 90% of what we write on mozorg they are all just as capable, so I don't care too much about what we use. Choosing one based on factors like tech stack / community / active maintenance is a very good suggestion. Personally, I'd favor Sass as it has so much activity in the community (plus tooling). That said, I think PostCSS is also well worth exploring as it means writing more potentially future proof source markup.

schalkneethling:
I agree that in terms of the pre-processor, we should choose the one with most activity/community support. Of course, we need to be aware of how the decisions effects our stack as well, and whether it will complicate local development and contribution in any way.

jpetto:
👍 to Sass.

media/css/pebbles/includes/lib.less
craigcook: Hm, fair point. I've just never liked "visually-hidden" because it sounds redundant, I wanted it to be clear this is hidden in an accessible way. But I suppose the comments make that clear enough. I'll change the name back.

craigcook:
So after a rousing bit of bikeshedding the consensus is to start using Sass for new work going forward. Rather than merge this refactor now only to have to immediately convert it to Sass, I'm going to close this PR and we'll combine the refactor with the Sass conversion in another PR soon.

Separately we'll also look into porting all of our existing Less to Sass (there are conversion tools) but in the near term they can coexist. We also need to do more research into PostCSS and see how it might fit into our workflow. It would also coexist with Less/Sass in the near term but there's a possibility of migrating to PostCSS alone down the road if it meets our needs.

jgmize:
> So after a rousing bit of bikeshedding the consensus is to start using Sass for new work going forward. [...] We also need to do more research into PostCSS and see how it might fit into our workflow. It would also coexist with Less/Sass in the near term but there's a possibility of migrating to PostCSS alone down the road if it meets our needs.

The potential switching costs are a major factor in this decision, so any switch away from the current language needs to be justified. The switching cost estimates need to include the intermediate stage where we would have parallel static asset build pipelines and dependencies on both node and ruby for an indefinite amount of time. I for one would find this very difficult to justify for just less+sass, and having less, sass, and postcss in the same project, even if "only temporarily" does not sound like a good idea to me at all, and I strongly suggest picking only one successor, if any.

PR #4229, opened 2016-07-19
Upgrade Jingo and Jinja2

pmac:
Prep for upgrade of Django to 1.8.x.

### DIFF
--- a/bedrock/base/tests/test_helpers.py
+++ b/bedrock/base/tests/test_helpers.py
@@ -4,7 +4,7 @@ import jingo
 
 
 def render(s, context={}):
-    t = jingo.env.from_string(s)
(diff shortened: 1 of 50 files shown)


pmac:
@jgmize here's the next step on our way to Django 1.8. This upgrades Jingo to a version that supports Djangos 1.7 and 1.8 as well as some supporting libs. It also removes waffle and converts all of that to the new `switch()` helper method that is backed by env vars.

pmac:
I've deployed this to [my demo](https://bedrock-demo-pmac.us-west.moz.works) and it seems to be functioning well.

jgmize:
Nice work as usual 👍

PR #4230, opened 2016-07-20
Fix languages not recognized by Django

pmac:
Using Django's i18n infra for activation of translation means that it looks for .mo files to decide whether to activate it. This means that only those languages that ship with Django will activate. Since we have our own i18n system anyway we can copy their mechnics with a much more simple system.

### DIFF
--- a/bedrock/base/context_processors.py
+++ b/bedrock/base/context_processors.py
@@ -1,7 +1,8 @@
 # from bedrock.base
 
 from django.conf import settings
-from django.utils import translation
+
+from lib.l10n_utils import translation
 
 
 def i18n(request):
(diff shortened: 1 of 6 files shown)


pmac:
@jgmize I got an email tonight saying that 3 letter languages like "dsp" weren't working. Turns out the real deal is what I explained above. So I think we have 2 options:

1. Merge this or something like it
2. Revert the Tower removal

I think this solves it and is actually better and more efficient for us. But we should discuss the risks and run the full suite on a demo, which I'm deploying now.

jgmize:
This approach seems workable to me. I'd like some unit tests and a headless test that validates that at least one of the affected languages is active, but manual testing on the deployed instance is probably sufficient for now and we can add those in a follow-up PR if you like.

pmac:
Recorded the issue in [bug 1287982](https://bugzilla.mozilla.org/show_bug.cgi?id=1287982)

PR #4231, opened 2016-07-20
Add unit-tests for translation utilities

pmac:
These are pretty simple tests, but so are the utilities. I've not yet come up with a way to test the thread-safety of these (though for our use that's not vital). Also we need some headless real tests but that would rely on installed translations so I'm not yet sure how to accomplish it. Let's discuss.


pmac:
@jgmize as I said above this is a pretty basic start, but to really test this right we'll probably need some further discussion. I'm leaving this here now so that we can get the conversation going tomorrow.

PR #4232, opened 2016-07-20
[fix bug 1287876] Unhide panels on home prototype B.

jpetto:
## Description

Removes snazzy slider to expose information at page load/without extra effort. Also updates the survey link for this prototype, and adds a bit of scroll tracking for wider viewports.

(I know the markup is now not the most semantic, but seeing as how it's a prototype that we need to update rather quickly, I didn't think an HTML refactor was warranted.)

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1287876

## Testing

Make sure no UI regressions at different viewport sizes/on older browsers.

## Checklist


craigcook:
r+ 🐼

PR #4233, opened 2016-07-20
Add ability to use SASS files as well as LESS

pmac:


pmac:
@alexgibson @jpetto @craigcook I got frustrated at django so I went ahead and did this to distract myself. Works for me locally. Give it a shot if you get a min.

jpetto:
I pulled this down, added a `test.scss` file to the `home` bundle, and received the following error:

`An output directory must be specified when compiling a directory`

I tried putting the `.scss` file in its own bundle, but that didn't help, either. I then tried compiling on the command line (`./node_modules/node-sass/bin/node-sass media/css/mozorg/home/test.scss > static/css/mozorg/home/test.css`), which worked without issue.

I then went in and cleared my `/static` directory (`git clean -fdx`), made a change to `test.scss`, and the pipeline compiled just fine. I put the `.scss` file back in the `home` bundle, and again no problems.

Finally, I added a new `.scss` file to the `firefox_channel` bundle, and again no issues.

Could be just a glitch on my local env, but wanted to make a note in case anyone else sees the same.

craigcook:
I'm doing a little experimenting with PostCSS, though right now I think Sass still makes the most sense. But we shouldn't commit to Sass until we're sure. Marking this as do-not-merge for the moment.

craigcook:
So after some tinkering and reading and evualting of PostCSS as a replacement for a preprocessor, I've come to the conclusion that, while it is technically possible, it isn't practical for bedrock. We're going with Sass. I can elaborate on the reasoning if anyone wants to know, but as it pertains to this PR, Sass is back on the plate.

pmac:
> Sass is back on the plate.

👍

alexgibson:
> I pulled this down, added a test.scss file to the home bundle, and received the following error:

> An output directory must be specified when compiling a directory

> I tried putting the .scss file in its own bundle, but that didn't help, either. I then tried compiling on the command line (./node_modules/node-sass/bin/node-sass media/css/mozorg/home/test.scss > static/css/mozorg/home/test.css), which worked without issue.

> I then went in and cleared my /static directory (git clean -fdx), made a change to test.scss, and the pipeline compiled just fine. I put the .scss file back in the home bundle, and again no problems.

> Finally, I added a new .scss file to the firefox_channel bundle, and again no issues.

> Could be just a glitch on my local env, but wanted to make a note in case anyone else sees the same.

Fwiw, I just tried replicating this but could not. Sass compilation worked as expected for me 👍

craigcook:
I finally tested this and I also didn't hit any snags once I npm installed and gulped. Sass files compile in the appropriate location, either as part of a bundle with Less files or in their own bundle. So I guess that's an r+ 🍶

PR #4236, opened 2016-07-22
Upgrade Django to 1.8.14

pmac:
This upgrades Django from 1.7 to 1.8 along with several other dependencies. Due to the major difference between 1.7 and 1.8 regarding template engine handling we've also had to simultaneously switch from Jingo to Django-Jinja.

I also forked jingo-markdown, updated it for use with django-jinja, and released it as [django-jinja-markdown](https://pypi.python.org/pypi/django-jinja-markdown). This has replaced jingo-markdown in the project.


pmac:
I've removed the blocking tags. It obviously needs a rebase and squash, but this is ready for review. @jgmize r?

pmac:
Blocked because it has to be deployed along with the new version of [Nucleus](https://nucleus.mozilla.org).

PR #4237, opened 2016-07-22
Switch from django-nose to py.test

pmac:
Remove most of nose libs. Nose itself is still required for nose.tools functions.


pmac:
@jgmize r?

jgmize:
Thanks @pmac

PR #4239, opened 2016-07-23
Bug 1167570 - Add "Source Code" download link to https://www.mozilla.org/firefox/all/

gerv:
## Description
LGPL requires us to provide access to source code where we provide access to binaries, and Firefox contains some LGPLed code. It doesn't have to be high profile; here's one suggested way.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1167570

## Testing
Loaded on my machine :-)

## Checklist
- [X] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/all.html
+++ b/bedrock/firefox/templates/firefox/all.html
@@ -62,7 +62,8 @@
     <h1>{{ self.page_title()  }}</h1>
     <h2>{{ self.page_desc() }}</h2>
     <p><a class="sysreq" href="{{ firefox_url(platform, 'sysreq', channel) }}">{{ _('Check the system requirements') }}</a><br />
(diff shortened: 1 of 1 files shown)


flodolo:
I'm confused: the bug talks about /firefox/new (localized, and we don't have such a string, so it should go behind a `l10n_has_tag` conditional) but you're adding the link to /firefox/all (not localized, so it's OK to add a string directly)?

I also don't remember ever seing a link to source in download buttons.

gerv:
Yeah, I thought /firefox/all was actually a better place to add the link than /firefox/new, because it's always the same and that's where we have all the "alternative" downloads.

See the bug for the LGPL text we are trying to obey.

flodolo:
Makes sense, l10n is not a blocker then 👍

alexgibson:
Thanks @gerv lgtm r+ 👍

PR #4244, opened 2016-07-25
Remove and clone locale dir if not a repo in l10n_update

pmac:


pmac:
@jgmize r?

pmac:
Maybe we don't need this now? Still wouldn't hurt I guess to have this be the behavior of this command?

jgmize:
I don't think we need this, but you're right that it probably wouldn't hurt to have it for local dev.

pmac:
Agreed. Not necessary. Let's close. Thanks.

pmac:
Reopening as we need this in order to make l10n updates work on www-dev.

pmac:
Hmm... Can't reopen. Will open a new one I guess.

pmac:
Opened #4385

PR #4245, opened 2016-07-25
Prevent warning about test runner change in Django 1.6

pmac:


pmac:
@jgmize r?

/cc @alexgibson thanks for the heads-up. This should fix it.

pmac:
IRC conversation context:

```
<agibson> pmac: since a day or so ago I get this warning when running bedrock locally. I’m assuming this is expected but just wanted to make sure: https://gist.github.com/alexgibson/398ff90a074eb33bfde2fd4a6aefdf51
```

pmac:
This works, and is pretty much a no-op. Merging. Thanks!

PR #4246, opened 2016-07-25
[fix bug 1288552] Stop traffic to onboarding variations.

jpetto:
## Description

As the onboarding test is complete, we should stop rendering the variations. There are still some funnelcake users trickling in - those hitting a `/firstrun/` variant URL will get the standard firstrun page, while those hitting any `/secondrun/` variation will be redirected to `/firefox/mobile-download/`.

After data is analyzed, we will file a follow-up bug/PR to appropriately deal with the variant templates & assets.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1288552

## Testing

Ensure all `/firstrun/` variant URLs (`?v=[1-6]`) render the modern/canonical firstrun template. Ensure all `/secondrun/` variant URLs (`/firefox/{version}/secondrun/?v=[1-6]`) redirect to `/firefox/mobile-download/`.

## Checklist


bedrock/firefox/redirects.py
```diff
@@ -504,4 +504,8 @@ def firefox_mobile_faq(request, *args, **kwargs):
 
     # bug 1283397
     redirect(r'^firefox/pocket/?$', 'https://getpocket.com/firefox/'),
+
+    # bug 1288552 - redirect /secondrun/ traffic from funnelcake test
+    redirect(r'^firefox(?:\/\d+\.\d+(?:\.\d+)?(?:a\d+)?)?/secondrun(?:/.*)?',
```
alexgibson: It would be good to add some tests for this redirect. Also, is this permanent, or should we use a temporary redirect?

alexgibson:
Aside from the missing test / redirect question this looks good, r+wc

My only other point for consideration would be that unless we're pretty certain that `/firstrun` is going to change as a result of this experiment (which I don't think it is, but could be wrong), would it be worth removing the orphaned templates now in this PR? Everything is still in version control, so resurrecting something in a few weeks time shouldn't be too difficult.

bedrock/firefox/redirects.py
jpetto: Uhm, permanent for now? It seems all data is pointing to showing less (and maybe no) first/second run content, so I think a permanent redirect is appropriate.

jpetto:
Added a test for the new redirect.

> would it be worth removing the orphaned templates now in this PR

I asked cmore about this when starting the work, and he recommended leaving the templates around until he finished analyzing the test results. firstrun *might* change - we'll see.

alexgibson:
> I asked cmore about this when starting the work, and he recommended leaving the templates around until he finished analyzing the test results. firstrun might change - we'll see.

Whatever happens I imagine it would need reworking anyway, since the Firefox team considers the sunrise / forest theme background the new brand direction for these pages. But if leaving the templates around will help, that's ok by me 👍

PR #4247, opened 2016-07-25
Revert "Remove .git from locale/.dockerignore"

jgmize:
Reverts mozilla/bedrock#4242


pmac:
We could do this, or we could just add `locale/.git` to the `.dockerignore` in the root of the repo. Thoughts?

jgmize:
> We could do this, or we could just add locale/.git to the .dockerignore in the root of the repo. Thoughts?

Yes, good idea, I like that better.

PR #4248, opened 2016-07-26
Remove Hello FTU instructions from UITour docs

alexgibson:
Just tidying up after myself (I forgot to remove this when we decommissioned the Hello FTU some time ago).


jpetto:
r+ 🚿

PR #4249, opened 2016-07-27
Fix Bug 1289796 - default ios version should be 5.0 instead of 4.0 in release notes

kyoshino:
Hope this would be the last time to hardcode the version on Bedrock ;-)


jpetto:
> Hope this would be the last time to hardcode the version on Bedrock ;-)

Me too. 😀 r+

PR #4250, opened 2016-07-27
[fix bug 1288517] Test dumb download links in IE <= 8.

jpetto:
## Description

Prep pages to serve updated/minimal JS to IE <= 8. ~~Traffic will be directed by Optimizely experiment (to be created).~~ Traffic will be directed using our Traffic Cop.

Also removes some of the code (bundles, template & view logic) from the /all test run in early October. (Assets from that test still exist as they will likely be re-purposed for a future test.)

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1288517

## Testing

~~Test `/firefox/new/?v=2` and `/firefox/new/` in modern browsers and in IE 7 & 8.~~
Make sure `/firefox/new/?v=1` functions correctly in IE 7 (regular download button), IE 8 (no JS download buttons), and modern browsers (regular download button). Make sure Traffic Cop experiment functions as expected.

## Checklist

### DIFF
--- a/bedrock/firefox/templates/firefox/new/scene1-ie8.html
+++ b/bedrock/firefox/templates/firefox/new/scene1-ie8.html
@@ -1,0 +1,15 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% add_lang_files "firefox/new/horizon" %}
+
+{% extends "firefox/new/scene1.html" %}
+
+{% block experiments %}
+  {{ super() }}
+  {% javascript 'oldie-nojs' %}
+{% endblock %}
+
+{% block js %}{% endblock %}
+
--- a/bedrock/firefox/templates/firefox/new/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/scene1.html
@@ -9,8 +9,8 @@
 {% extends "firefox/new/base.html" %}
 
 {% block experiments %}
-  {% if switch('experiment-firefox-new-all-links') and LANG == 'en-US' %}
-    {% javascript 'experiment-firefox-new-scene1-all-links' %}
+  {% if switch('experiment-firefox-new-oldie-nojs') %}
+    {% javascript 'experiment_firefox_new_oldie_nojs' %}
   {% endif %}
 {% endblock %}
 
@@ -18,9 +18,6 @@
 {% block extrahead %}
   {% stylesheet 'firefox_new_common' %}
   {% stylesheet 'firefox_new_scene1' %}
-  {% if version in ['1', '2'] %}
-    {% stylesheet 'firefox_new_scene1_variations' %}
-  {% endif %}
 {% endblock %}
 
 {% block body_id %}firefox-new-scene1{% endblock %}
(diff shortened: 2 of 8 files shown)


bedrock/firefox/templates/firefox/new/horizon/scene1-ie8.html
```diff
@@ -0,0 +1,15 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% add_lang_files "firefox/new/horizon" %}
```
jpetto: Do I need to include this line if it's in the `extends`-ed template?

jpetto:
Pages to test are not yet finalized, but this is the approach I'm looking at. We can get ahead of the game by reviewing this approach & code now. (Though I'm pretty sure we'll want to test `/firefox/new/`.)

What says you @alexgibson?

bedrock/firefox/templates/firefox/new/horizon/scene1-ie8.html
alexgibson: Yes I believe so

alexgibson:
This seems to work ok for me locally.

Unrelated to this PR, but I did notice the "dumb" download buttons stack vertically instead of horizontally in IE 7. Changing (this selector)[https://github.com/mozilla/bedrock/blob/master/media/css/sandstone/buttons.less#L255] to use the `.inline-block;` mixing seems to fix it.

alexgibson:
Think it's worth pushing this to a demo server? Clicking the dumb download links works fine for me locally, but I have a feeling (bug 1282386)[https://bugzilla.mozilla.org/show_bug.cgi?id=1282386] may effect this test once GTM is enabled :/

alexgibson:
Overall I think this approach for testing seems to work ok. My only slight concern is that we're not replicating the exact same JS oldIE bundle that we're proposing to use, but for a small test this seems ok 👍

alexgibson:
Last question - which browser(s) is this test going to run on? 

We have a [sanity test here](https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/new/test_download.py#L14) that runs in IE 6 & 7 to make sure exactly one download button is shown on the page. If this test is run on either of those browsers it would fail, so we would need to temporarily skip the test for IE if that may be an issue.

jpetto:
> Last question - which browser(s) is this test going to run on?

I believe we are only going to target IE 8 with the test, as it has a relatively high download rate. If it turns out we're going to test IE 8 and below, I'll be sure to `skipif` the impacted tests.

> Think it's worth pushing this to a demo server? Clicking the dumb download links works fine for me locally, but I have a feeling bug 1282386 may effect this test once GTM is enabled :/

Sigh, yes. It's on its way.

Also, good catch on the stacked download links. Replacing `display: inline-block` with the `.inline-block()` mixin does seem to do the trick. Will fix in this PR.

jpetto:
This is now up on my demo:

https://bedrock-demo-jpetto.us-west.moz.works/en-US/firefox/new/?v=2

As you feared, [bug 1282386](https://bugzilla.mozilla.org/show_bug.cgi?id=1282386) does indeed trigger the popup blocker. I guess we're blocked on this until that bug is resolved. 😞

alexgibson:
> As you feared, bug 1282386 does indeed trigger the popup blocker. I guess we're blocked on this until that bug is resolved. 😞

Well, this bug *is* already in production, and it's the kind of thing GA stats do not pick up on in terms of a test. I do however think it's a real pain for the user to have to accept the warning. Hopefully James & Gareth will come to some kind of conclusion soon. If it was up to me, I would be deprecating GTM support for IE < 9 along with this change. 😞

jpetto:
I don't think we can run the test as-is. It's one (bad) thing to have a popup warning on `/firefox/all/`. It's quite another to have it on `/firefox/new/`, which, even if the pre-popup clicks are counted, could well impact actual downloads.

alexgibson:
> I don't think we can run the test as-is. It's one (bad) thing to have a popup warning on /firefox/all/. It's quite another to have it on /firefox/new/, which, even if the pre-popup clicks are counted, could well impact actual downloads.

Totally agree 👍

jpetto:
As we are not targeting IE 6 or 7 (and thus shouldn't need to update any tests), I'm going to remove the WIP label here. Unfortunately, we're still blocked on the GTM issue. 😢

jpetto:
Rebased and updated 3 months after original review, so should probably be reviewed again. Still Do No Merge as we wait for [Peter to confirm audience segmentation in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1288517#c8).

craigcook:
LGTM, I get the expected results in IE7 and IE8. r+ pending the audience segment question.

PR #4251, opened 2016-07-27
[fix bug 1278380] Update newsletter thank you message

craigcook:
## Description
Updates the standard "thank you" message shown when subscribing to a newsletter. This uses the l10n tag "newsletter_thanks_072016" and I think the same string resides in both main.lang and in newsletter.lang. 

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1278380

## Checklist
- [x] Requires l10n changes.

### DIFF
--- a/bedrock/newsletter/templates/newsletter/mozilla-and-you.html
+++ b/bedrock/newsletter/templates/newsletter/mozilla-and-you.html
@@ -18,6 +18,6 @@
 {% endblock %}
 
 {% block success_confirm %}
-  <p>{% trans %}Thanks! Please check your inbox to confirm your subscription.{% endtrans %}</p>
-  <p>{% trans %}You'll receive an email from mozilla@e.mozilla.org to confirm your subscription. If you don't see it, check your spam filter. You must confirm your subscription to receive our newsletter.{% endtrans %}</p>
+  <h3>{{ _('Thanks!') }}</h3>
+  {{ super() }}
 {% endblock success_confirm %}
--- a/bedrock/mozorg/templates/mozorg/contribute/friends.html
+++ b/bedrock/mozorg/templates/mozorg/contribute/friends.html
@@ -94,13 +94,14 @@
   </div>{# /-- .ff-signup-wrapper #}
 
   <div id="newsletter-form-thankyou" class="thank newsletter-form newsletter-form-extra">
-    <h3>{{ _('Thanks! Please check your inbox to confirm your subscription to the Firefox Friends newsletter.') }}</h3>
+    <h3>{{ _('Thanks!') }}</h3>
 
     {# below matches text found in newsletter/includes/macros/email_form_thankyou() #}
     <p>
     {% trans %}
-      You'll receive an email from mozilla@e.mozilla.org to confirm your subscription.
-      If you don't see it, check your spam filter. You must confirm your subscription to receive our newsletter.
+      If you haven’t previously confirmed a subscription to a Mozilla-related
+      newsletter you may have to do so. Please check your inbox or your spam filter
+      for an email from us.
     {% endtrans %}
     </p>
   </div>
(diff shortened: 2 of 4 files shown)


craigcook:
cc @flodolo

flodolo:
String is in a shared file (main.lang), we can't use tags. If you don't want to block on l10n, you need a lang.startsWith('en-') for now.

craigcook:
Oh, I thought we could use tags with main.lang... ok, I'll revise the code and I'll find out if we can block on l10n since this appears on so many pages and I'm not sure it's urgently needed in prod.

alexgibson:
Does it maybe read a little odd now that we have the title and paragraph both starting with "Thanks!"?

craigcook:
> Does it maybe read a little odd now that we have the title and paragraph both starting with "Thanks!"?

Perhaps so. I'll ask Troy if he wants to change it, and I'll comment here if/when the string is truly frozen so we can start l10n.

craigcook:
Oh, good call. Seeing it in proper context I've also realized the headline in most places says "check your inbox to confirm" which is sometimes no longer needed and thus the whole reason for this update. So I've used the general "Thanks!" as the headline and removed it from the body. I can triple-check that Troy is happy with this and will comment here when we can consider this string frozen to start l10n.

craigcook:
And I just realized I made the same comment twice but totally rephrased because in between them I went ahead and changed the strings but forgot I had already commented. But I'm not gonna be a jerk and try to erase my mistakes. I fail at github comments and will leave it for all to see. Move along!

craigcook:
@flodolo @peiying2: Strings are frozen. Go!

bedrock/newsletter/templates/newsletter/mozilla-and-you.html
```diff
@@ -18,6 +18,12 @@ <h1 class="large">{{ _('Read all about it in our <span>newsletter</span>') }}</h
 {% endblock %}
 
 {% block success_confirm %}
-  <p>{% trans %}Thanks! Please check your inbox to confirm your subscription.{% endtrans %}</p>
-  <p>{% trans %}You'll receive an email from mozilla@e.mozilla.org to confirm your subscription. If you don't see it, check your spam filter. You must confirm your subscription to receive our newsletter.{% endtrans %}</p>
+  <h3>{{ _('Thanks! ') }}</h3>
```
jpetto: Looks like an extra space made it into this string.

bedrock/newsletter/templates/newsletter/mozilla-and-you.html
```diff
@@ -18,6 +18,12 @@ <h1 class="large">{{ _('Read all about it in our <span>newsletter</span>') }}</h
...
+  <p>
+  {% trans %}
+    If you haven’t previously confirmed a subscription to a Mozilla-related
+    newsletter you may have to do so. Please check your inbox or your spam filter
+    for an email from us.
+  {% endtrans %}
+  </p>
```
jpetto: Since this template extends `one_newsletter_signup.html` (which has the updated copy), instead of repeating the thank you message, you can import the parent template's content for this block alongside additions. Something like:

```
{% block success_confirm %}
  <h3>{{ _('Thanks!') }}</h3>
  {{ block.super }}
{% endblock %}
```

Test that out, but I believe it'll work.

craigcook: Good call, that works.

craigcook: Looks like it's just `{{ super() }}` rather than `{{ block.super() }}`

jpetto:
LGTM. r+ 📰

flodolo:
@craigcook We already have 31 locales done, all those who have newsletter. Feel free to merge when you want.

craigcook:
Thanks @flodolo!

PR #4253, opened 2016-07-28
Bug 1278981: Add service to sign Stub Attribution URL params

pmac:
Uses HMAC with SHA256

### DIFF
--- a/bedrock/firefox/tests/test_views.py
+++ b/bedrock/firefox/tests/test_views.py
@@ -3,16 +3,166 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 import json
+from urllib import unquote_plus
+from urlparse import parse_qs
 
 from django.test import override_settings
 from django.test.client import RequestFactory
 
 from bedrock.base.urlresolvers import reverse
-from mock import patch
+from mock import patch, Mock
 from nose.tools import eq_, ok_
 
 from bedrock.firefox import views
 from bedrock.mozorg.tests import TestCase
+
+
+@override_settings(STUB_ATTRIBUTION_HMAC_KEY='achievers',
+                   STUB_ATTRIBUTION_RATE=1)
+@patch.object(views, 'time', Mock(return_value=12345.678))
+class TestStubAttributionCode(TestCase):
+    def _get_request(self, params):
+        rf = RequestFactory()
+        return rf.get('/', params,
+                      HTTP_X_REQUESTED_WITH='XMLHttpRequest',
+                      HTTP_ACCEPT='application/json')
+
+    def test_not_ajax_request(self):
+        req = RequestFactory().get('/', {'source': 'malibu'})
+        resp = views.stub_attribution_code(req)
+        self.assertEqual(resp.status_code, 400)
+        assert 'cache-control' not in resp
+        data = json.loads(resp.content)
+        self.assertEqual(data['error'], 'Resource only available via XHR')
+
+    def test_no_valid_param_names(self):
+        req = self._get_request({'dude': 'abides'})
+        resp = views.stub_attribution_code(req)
+        self.assertEqual(resp.status_code, 400)
+        assert resp['cache-control'] == 'max-age=31536000'
+        data = json.loads(resp.content)
+        self.assertEqual(data['error'], 'no params')
+
+    def test_no_valid_param_data(self):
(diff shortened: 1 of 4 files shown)


pmac:
@jgmize this is ready for review. There are still some open questions, but this is the correct implementation based on the current situation.

/cc @ckprice

alexgibson:
I gave this a test locally by making an ajax GET request along with some params and it seems to work well 👍 

I'll defer to Josh on the code review side of things, but functionally I think this looks good.

stephendonner:
Hey @pmac @jgmize - I'd like to start being able to test this, coupled with https://github.com/mozilla-services/go-bouncer/pull/36 and, I think, the as-yet-to-be-coded/merged/stood-up Stub Attribution-download service.  Any chance of/when can we get this + the other PRs in, so I can test more fully?  Thanks!  /cc @oremj

jgmize:
I've only taken a cursory look at this code so far, given the current status (currently "Do Not Merge" + WIP), but I haven't seen anything obviously wrong yet. This is in need of a rebase, but that's not a blocker to deploying to a demo instance if needed.

alexgibson:
@stephendonner this is only one piece of the work that's required in order to test stub attribution. We're working on this along with the other parts this sprint. When things are ready for testing, we'll let you know. Thanks

stephendonner:
Thx @alexgibson - as you can tell/as usual, I'm anxious to test ;-)

pmac:
@jgmize it's ready for r? for real now.

@alexgibson I've made some changes. I changed it to expect exactly the params that should be in the URL of the page, so if you wanted to you could just tack it onto the AJAX request and let the service work it out. I also changed the returned JSON to use the parameter names that bouncer expects (`attribution_code` and `attribution_sig`).

bedrock/firefox/views.py
```diff
@@ -68,6 +77,19 @@
...
+    ('utm_source', '(not set)'),
+    ('utm_medium', '(direct)'),
+    ('utm_campaign', '(not set)'),
+    ('utm_content', '(not set)'),
+]
+STUB_VALUE_RE = re.compile(r'^[a-z0-9-.%()_]+$', flags=re.IGNORECASE)
+
+# available variations for onboarding first & second run tests - Q2 2016
+# https://bugzilla.mozilla.org/show_bug.cgi?id=1259608
+ONBOARDING_VARIATIONS = ['1', '2', '3', '4', '5', '6']
```
alexgibson: Is this needed here? rebase gremlins maybe? 👹

pmac: Bah! Indeed. I'll remove.

bedrock/firefox/views.py
```diff
@@ -86,6 +103,57 @@ def installer_help(request):
...
+    """Return a JSON response containing the HMAC signed stub attribution value"""
+    if not request.is_ajax():
+        return HttpResponseJSON({'error': 'Resource only available via XHR'}, status=400)
+
+    key = settings.STUB_ATTRIBUTION_HMAC_KEY
+    if key is None:
+        return HttpResponseJSON({'error': 'HMAC key not set'}, status=400)
+
+    rate = settings.STUB_ATTRIBUTION_RATE
+    if rate < 1 and random() > rate:
```
pmac: Hmm... I think you're right. Good catch.

bedrock/firefox/tests/test_views.py
```diff
@@ -3,18 +3,160 @@
...
+    def test_no_valid_param_data(self):
+        params = {'utm_source': 'br@ndt', 'utm_medium': 'ae#her'}
+        req = self._get_request(params)
+        resp = views.stub_attribution_code(req)
+        self.assertEqual(resp.status_code, 400)
+        data = json.loads(resp.content)
+        self.assertEqual(data['error'], 'no params')
+
+    def test_some_valid_param_data(self):
+        params = {'utm_source': 'brandt', 'utm_content': 'ae#her'}
```
pmac: Ah yes, good call. A better test would be a more likely invalid character. I'll fix.

pmac:
Updated to remove dynamic rate limiting and improve cache header usage.

pmac:
I think we're good here. Plus, this will not be used in prod until the frontend work is done, and it won't be turned on until we're ready and have an HMAC key. I'm merging this based on the previous feedback so that @alexgibson can more easily file his PR for the frontend bits.

Thanks all!

PR #4254, opened 2016-07-29
Bug 1282494: Use http://product-details.mozilla.org/ instead of the s…

pascalchevrel:
## Description
This patch override the default PROD_DETAILS_URL variable to use https://product-details.mozilla.org/1.0/ as a source for json files

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1282494

## Testing
Manual testing of download page

## Checklist

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -101,6 +101,10 @@ TEST_RUNNER = 'django.test.runner.DiscoverRunner'
 # http://www.i18nguy.com/unicode/language-identifiers.html
 LANGUAGE_CODE = 'en-US'
 
+# Use Ship It as the source for product_details
+PROD_DETAILS_URL = config('PROD_DETAILS_URL',
+                          default='https://product-details.mozilla.org/1.0/')
+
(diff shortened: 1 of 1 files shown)


bedrock/settings/base.py
```diff
@@ -101,6 +101,9 @@ def path(*args):
 # http://www.i18nguy.com/unicode/language-identifiers.html
 LANGUAGE_CODE = 'en-US'
 
+# Use Ship It as the source for product_details
+PROD_DETAILS_URL = 'https://product-details.mozilla.org/1.0/'
```
pmac: Could you make this configurable? It'd be nice to be able to switch back to SVN if we have to w/o having to deploy.

```python
PROD_DETAILS_URL = config('PROD_DETAILS_URL', default='https://product-details.mozilla.org/1.0/')
```

pmac:
Just one little change then I think we're ready. Thanks Pascal!

pascalchevrel:
The change with a default value is in the patch

PR #4256, opened 2016-07-31
Fix Bug 1290691 - Dehardcode iOS for Firefox version in Bedrock settings

kyoshino:


alexgibson:
Adding a blocker to this until we switch to the new product details as default.

kyoshino:
This is no longer blocked as #4267 has been merged.

alexgibson:
I pulled this locally but looks like I'm getting a `KeyError` for `'ios_version'` on the release notes page?

kyoshino:
I've got the same error but it's gone after running `bin/sync_all` 🔄

alexgibson:
I ran `bin/sync_all` again and it is indeed working, thanks @kyoshino! r+

PR #4257, opened 2016-08-01
[bug 1263268] Use l10n tag for leadership intro copy changes

alexgibson:
@flodolo r?

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/leadership.html
+++ b/bedrock/mozorg/templates/mozorg/about/leadership.html
@@ -13,7 +13,11 @@
   <h1 class="title-shadow-box">{{ _('Leadership') }}</h1>
 
   <p>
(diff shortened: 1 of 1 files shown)


bedrock/mozorg/templates/mozorg/about/leadership.html
```diff
@@ -13,7 +13,11 @@
   <h1 class="title-shadow-box">{{ _('Leadership') }}</h1>
 
   <p>
-  {{ _('Our leadership team is made up of both paid staff and volunteers with expertise building products and empowering communities.') }}
+  {% if l10n_has_tag('leadership-copy-update') %}
```
flodolo: Can we use underscores for consistency?

flodolo:
If would be great if you could change the tag, but it's an r+ from me (and thanks for quick fix)

alexgibson:
Changed to underscores as suggested, thanks

PR #4264, opened 2016-08-03
[fix bug 1262892] Add origin param to FxA iframe src.

jpetto:
## Description

The FxA team is working on an update to the mechanism that authorizes embedding of their iframe. This change is to allow that team to test their new approach. Functionally, nothing should change on mozilla.org due to this PR.

After the new auth mechanism is fully in place, we will file a follow-up PR to remove the current `ping` postMessage auth.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1262892

## Testing

Make sure nothing has changed/no regressions have appeared due to this change. Test locally, on [demo5](https://www-demo5.allizom.org/en-US/firefox/accounts/) (which points to @shane-tomlinson's FxA server for testing the new auth), and on [demo 4](https://www-demo4.allizom.org/en-US/firefox/accounts/) (which points to FxA stage).

(**Reminder:** The iframe only loads on Fx 39+, so shouldn't have to fire up a VM.)

## Checklist

### DIFF
--- a/docs/firefox-accounts.rst
+++ b/docs/firefox-accounts.rst
@@ -22,7 +22,7 @@ Local Development
 
     ``FXA_IFRAME_SRC=https://stomlinson.dev.lcip.org/``
 
-#. Set your local development server to run on ``127.0.0.1:8111``.
+#. Set your local development server to run on ``127.0.0.1:8111``, e.g. ``PORT=8111 gulp``.
 #. Quit Firefox.
 #. Create a new profile for testing called ``FxA Test Local`` by following the
    `instructions here`_.
--- a/media/js/base/mozilla-fxa-iframe.js
+++ b/media/js/base/mozilla-fxa-iframe.js
@@ -59,6 +59,9 @@ Mozilla.FxaIframe = (function() {
             _src = _src.replace('context=iframe', 'context=fx_firstrun_v2');
         }
 
+        // add origin (protocol + hostname + (optional) port) for embed authorization
+        _src += '&origin=' + encodeURIComponent(window.location.origin);
+
         // initialize GA event name
         _gaEventName = _config.gaEventName || 'fxa';
 
@@ -160,6 +163,9 @@ Mozilla.FxaIframe = (function() {
     };
 
     var _onLoaded = function(data) {
+        // remember iframe has loaded (new auth flow doesn't fire 'ping')
+        _handshake = true;
+
         _sendGAEvent('fxa-loaded');
         _userCallback('onLoaded', data);
     };


alexgibson:
I tested this locally and on both demo's and code-wise this looks fine, and can't see and regressions in terms of signing up and verifying.

I did notice one (probably unrelated) regression on @shane-tomlinson's FxA server (demo5). On some page loads (but not all), the iframe doesn't resize correctly. I also noticed it sometimes appears to resize when switching to another tab and then back again.

iframe

I couldn't reproduce this against the FxA staging server.

This PR is probably fine to merge, but I'll defer until you have a chance to look at the above issue.

jpetto:
@shane-tomlinson: I too see some strange resizing behavior on demo5 (which is pointing to your FxA instance). Every time I load [the `/firstrun` page](https://www-demo5.allizom.org/en-US/firefox/48.0/firstrun/), the iframe resizes itself too short for the content after about a second. This behavior is not present on [demo 4](https://www-demo4.allizom.org/en-US/firefox/48.0/firstrun/).

> I did notice one (probably unrelated) regression

Agree this issue is likely unrelated to these changes, but wanted you to take a look just in case.

shane-tomlinson:
Thanks @jpetto and @alexgibson - I'll have a look. There's a chance that when I removed code, I removed the call to the code that sends the "resize" message. :/

jpetto:
I think we're good to merge here, as the resizing issue is relative to the demo FxA server.

shane-tomlinson:
Thanks @jpetto - can you hold off on merging until I get to the bottom of the resizing problem? I don't want to see this merge only to have to ask you to back it out. :(

jpetto:
Code here is likely final, but want to hold off on pushing to master while @shane-tomlinson investigates the resizing issue.

shane-tomlinson:
@jpetto - I see the resizing problem now - it's due to [this block that resizes automatically if the `ping` message is not received](https://github.com/jpetto/bedrock/blob/abccdc76cba90321cbda7ff3b4b36fa39422d306/media/js/base/mozilla-fxa-iframe.js#L81:L85). The [`_handshake` variable is set to true](https://github.com/jpetto/bedrock/blob/abccdc76cba90321cbda7ff3b4b36fa39422d306/media/js/base/mozilla-fxa-iframe.js#L160) in response to a `ping` message, whereas with the new flow the first two messages to be received will be [`loaded`](https://github.com/jpetto/bedrock/blob/abccdc76cba90321cbda7ff3b4b36fa39422d306/media/js/base/mozilla-fxa-iframe.js#L129) or [`resize`](https://github.com/jpetto/bedrock/blob/abccdc76cba90321cbda7ff3b4b36fa39422d306/media/js/base/mozilla-fxa-iframe.js#L133). If either of those two events arrive within the specified timespan, can the resize be stopped?

jpetto:
D'oh! I should have caught that - been too long since I worked on that script. I've added checking for the `_handshake` in the `loaded` event, which appears to have fixed the issue. I've updated both [demo4](https://www-demo4.allizom.org/firefox/accounts/) and [demo5](https://www-demo5.allizom.org/firefox/accounts/), and both seem in good shape.

shane-tomlinson:
@jpetto - Whenever this merges, will it affect all browser versions where firstrun is available, or only the newest? And to double check, firstrun is available for Fx 40 and above?

shane-tomlinson:
Ref https://github.com/mozilla/fxa-content-server/pull/4008

shane-tomlinson:
@jpetto - this PR looks good to merge to me!

jpetto:
> Whenever this merges, will it affect all browser versions where firstrun is available, or only the newest? And to double check, firstrun is available for Fx 40 and above?

@shane-tomlinson - This PR will affect all instances of the FxA iframe, which currently means `/firstrun/` (for Fx 40+) and `/firefox/accounts/`.

pmac:
@jpetto based on my limited understanding of the `_handshake` var, it seems fine. r+

PR #4266, opened 2016-08-04
[fix bug 1274207] Download page should only show 'Refresh Firefox' button if reset is supported

alexgibson:
## Description
- A new UITour method was added in Firefox 48 that tells us if the user profile supports the "Reset Firefox" dialog that we show to up-to-date users on /firefox/new/. This prevents edge-case weirdness, like in [bug 1126559](https://bugzilla.mozilla.org/show_bug.cgi?id=1126559).

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1274207

## Testing
- Load `/firefox/new/` in Firefox 48 (make sure you update from product details first), with [UITour enabled](http://bedrock.readthedocs.io/en/latest/uitour.html#local-development). You should still see the "Refresh Firefox" button, and clicking it should still open up the dialog.
- Test this is working for both the old and new template designs.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/docs/uitour.rst
+++ b/docs/uitour.rst
@@ -285,6 +285,7 @@ Available ``type`` values:
 * ``'selectedSearchEngine'``
 * ``'search'``
 * ``'loop'``
+* ``'canReset'``
 
 Other parameters:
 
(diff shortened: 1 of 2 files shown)


jpetto:
Tested with both old and new templates and all looks good to me. r+ 🔘

PR #4275, opened 2016-08-15
Fix legal-docs and privacy 500

pmac:


kyoshino:
Yeah, replacing `None` with an empty string is the right answer here, as [explained by the extension author](https://github.com/aleray/mdx_outline/blob/master/mdx_outline.py#L176) 👍

pmac:
Nice. Thanks Kohei!

PR #4281, opened 2016-08-16
Fix Bug 1295670 - Update to the Firefox Cloud Services Privacy Notice for today Aug 16, 2016

kyoshino:
This has to be pushed to production today if possible.


jgmize:
thanks @kyoshino

PR #4283, opened 2016-08-16
Fix 1295691: Bedrock process starts without database

pmac:
* Alter config to start wsgi process even without database server
* Change some templates to avoid relying on db data
* No longer run migrations on container start

### DIFF
--- a/bedrock/events/utils.py
+++ b/bedrock/events/utils.py
@@ -2,6 +2,7 @@ from functools import wraps
 
 from django.conf import settings
 from django.core.cache import cache
+from django.db import DatabaseError
 
 from bedrock.events.models import Event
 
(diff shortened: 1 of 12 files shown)


pmac:
@jgmize r?

bedrock/events/utils.py
```diff
@@ -33,27 +34,42 @@ def wrapped(*args, **kwargs):
 
 @cache_memoized
 def current_and_future_event_count():
-    return Event.objects.current_and_future().count()
+    try:
+        return Event.objects.current_and_future().count()
+    except DatabaseError:
+        return None
```
jgmize: Returning `None` may expose us to a thundering herd problem here. It may be better to explicitly cache 0 here instead for a short period of time.

bedrock/settings/base.py
```diff
@@ -31,19 +31,40 @@ def path(*args):
...
-    cast=json.loads,
-    default=json.dumps(
-        {'default': config('DATABASE_URL',
-                           cast=dj_database_url.parse,
-                           default='sqlite:///bedrock.db')}))
 
-if DATABASES['default']['ENGINE'].endswith('psycopg2'):
-    # let the DB handle connection killing
-    DATABASES['default']['CONN_MAX_AGE'] = None
+class NonDefaultRouter(object):
```
jgmize: We should probably import this class from a separate file, say `routers.py`.

bedrock/settings/base.py
```diff
@@ -31,19 +31,40 @@ def path(*args):
...
+    def _return_dbname(self, *args, **kwargs):
+        return self.dbname
+
+    def _return_true(self, *args, **kwargs):
+        return True
+
+    db_for_read = _return_dbname
+    db_for_write = _return_dbname
+    allow_relation = _return_true
+    allow_migrate = _return_true
```
jgmize: A shorter suggestion:

```python
class NonDefaultRouter(object):
  def __init__(self, dbname):
    self.dbname = dbname

  db_for_read = db_for_write = lambda self, *a, **kw: self.dbname
  allow_relation = allow_migrate = lambda *a, **kw: True 
```

If you prefer the more verbose version I have no objections.

pmac: I thought about lambdas. It's a toss-up really. One is succinct and one is readable. Hmmm.

pmac:
Updated, tests pass now as well.

docker/run-common.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-./manage.py syncdb --noinput
+./manage.py migrate --noinput --database bedrock
```
jgmize: I believe this file can be removed altogether, along with `docker/run-dev.sh` which is no longer used.

pmac: don't need it for any local dev docker stuff?

jgmize:
nice work, @pmac

PR #4284, opened 2016-08-17
Remove remaining legacy infrastructure files

pmac:


pmac:
@jgmize r?

Didn't realize these were still here.

PR #4285, opened 2016-08-17
Remove special jenkins and c9 settings files

pmac:


bedrock/settings/__init__.py
```diff
@@ -1,20 +1,15 @@
...
-    if os.getenv('C9_USER'):
-        from .c9 import *  # noqa
-    try:
-        from .local import *  # noqa
-    except ImportError as exc:
-        'local.py is supported, but no longer necessary'
+try:
+    from .local import *  # noqa
+except ImportError as exc:
+    'local.py is supported, but no longer necessary'
```
jgmize: how much longer do you think we should continue to support local.py?

PR #4286, opened 2016-08-17
[fix bug 1295170] add transparency report for Jan-Jun 2016

craigcook:
## Description
Adds a transparency report for the first half of 2016, and adds a link to the old 2015 report. I don't believe any of this content is localized.

We're awaiting final sign-off from Denelle before it goes live so marking do-not-merge for now but we can still begin code review.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1295170

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/reports-nav.html
+++ b/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/reports-nav.html
@@ -1,0 +1,8 @@
+      <nav class="content nav-report three-up">
+        <ul>
+          <li><a href="{{ url('mozorg.about.policy.transparency.index') + '#faq' }}">{{ _('FAQ') }}</a></li>
+          <li><a href="{{ url('mozorg.about.policy.transparency.index') + '#definitions' }}">{{ _('Definitions') }}</a></li>
+          {# This link should point to the latest report. Past reports are added to 'includes/past-reports.html' #}
+          <li><a href="{{ url('mozorg.about.policy.transparency.jan-jun-2016') }}">{{ _('January–June 2016 Report') }}</a></li>
+        </ul>
+      </nav>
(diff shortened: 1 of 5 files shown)


alexgibson:
These changes all look good to me, r+ to merge once you get final sign off 👍

craigcook:
We've got the thumbs up to push this. Taking the DNM off but there have been a few little changes since it was originally r+ed so maybe @jpetto can take a quick look before we merge, just to be safe?

jpetto:
LGTM. r+

One unrelated thing I noticed is that in Fx (Release, Beta, & Dev Edition), the anchor links into the definitions do not jump to the correct scroll position. This is happening in production on the current report as well. Tested in Chrome & Safari and they work as expected.

Try hitting the current 2015 report and clicking on the "Court Order" link. It positions the red arrow correctly, but the page is scrolled all the way to the bottom. If you see this behavior too, I'll file a bug.

craigcook:
I do see that behavior, and it was even reported before but I didn't think it was worth fixing at the time. What's happening is that the browser jumps to the anchor, but then the accordion script kicks in and collapses the FAQ above, moving the anchor higher on the page than it was initially. When we hit the bottom of the page we can't scroll any lower so the focused anchor gets "pulled up" out of view because the content above it changes height. I suspect webkit doesn't try to jump to an anchor until after DOM events fire but Firefox jumps earlier.

A few possible solutions:

1. Move definitions above the FAQ so when the FAQ collapses it has no effect on the content above it.
2. Lose the accordion on the FAQ; if they don't collapse the scroll position doesn't change.
3. Use JS to readjust the scroll position after the FAQ collapses and bring the anchor back into view.

Of these, number 3 is the most complex code-wise, but I think provides the best UX and makes the most sense for the content (FAQ really should come before definitions, and without the accordion that's a really big wall of text to look at). I'll file a bug and see if we can squeeze it into the sprint as a quality enhancement.

PR #4291, opened 2016-08-18
Upgrade prod-details lib. Fix issue with database name.

pmac:


bin/sync_all
```diff
@@ -8,7 +8,7 @@ fi
 ./manage.py migrate --noinput --database bedrock
 ./manage.py rnasync
 ./manage.py cron update_ical_feeds
```
jgmize: outside the scope of this PR, but I just realized we probably need to add "--database bedrock" to the other mgmt cmds, including adding support where needed

pmac: We might, or we might not. We'll have to test. The router should make most things do the right thing. The p-d one was only a problem because of the manual transaction management it does.

PR #4292, opened 2016-08-18
Bug 1249774: Add SecurityMiddleware and settings for HSTS

pmac:
Also add setting for XSS filter header. All are disabled by default so that we can turn them on and test. Also remove sslify and replace with SecurityMiddleware

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -305,7 +305,6 @@ BASIC_AUTH_CREDS = config('BASIC_AUTH_CREDS', default=None)
 
 MIDDLEWARE_CLASSES = [
     'django.middleware.security.SecurityMiddleware',
-    'sslify.middleware.SSLifyMiddleware',
     'bedrock.mozorg.middleware.MozorgRequestTimingMiddleware',
     'django_statsd.middleware.GraphiteMiddleware',
     'corsheaders.middleware.CorsMiddleware',
@@ -407,10 +406,18 @@ SESSION_COOKIE_HTTPONLY = not DEBUG
 SESSION_COOKIE_SECURE = not DEBUG
 SESSION_ENGINE = 'django.contrib.sessions.backends.cache'
 
+# legacy setting. backward compat.
+DISABLE_SSL = config('DISABLE_SSL', default=True, cast=bool)
 # SecurityMiddleware settings
 SECURE_HSTS_SECONDS = config('SECURE_HSTS_SECONDS', default='0', cast=int)
 SECURE_HSTS_INCLUDE_SUBDOMAINS = False
 SECURE_BROWSER_XSS_FILTER = config('SECURE_BROWSER_XSS_FILTER', default=False, cast=bool)
+SECURE_SSL_REDIRECT = config('SECURE_SSL_REDIRECT', default=not DISABLE_SSL, cast=bool)
+SECURE_REDIRECT_EXEMPT = [
+    r'^healthz/$',
+]
+if SECURE_SSL_REDIRECT:
+    SECURE_PROXY_SSL_HEADER = ('HTTP_X_FORWARDED_PROTO', 'https')
 
 LOCALE_PATHS = (
     str(LOCALES_PATH),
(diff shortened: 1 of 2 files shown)


pmac:
@jgmize r?

bedrock/settings/base.py
```diff
@@ -406,6 +406,19 @@ def lazy_langs():
 SESSION_COOKIE_SECURE = not DEBUG
 SESSION_ENGINE = 'django.contrib.sessions.backends.cache'
 
+# legacy setting. backward compat.
+DISABLE_SSL = config('DISABLE_SSL', default=True, cast=bool)
```
jgmize: /me looks forward to getting rid of this poorly named backward compat env var

pmac: We could do that now. I could just go ahead and add the new config where appropriate if you'd like, and remove this line. I just did this so that we wouldn't have to immediately need to change prod config.

pmac: @jgmize so what do you think. Is it worth me adding the new config to every bedrock deis app now and removing this line before we merge, or should we do that later? I'm on the fence. It seems a bit disruptive, but it would be nice for this to be cleaner.

bedrock/settings/base.py
```diff
@@ -406,6 +406,19 @@ def lazy_langs():
...
+# SecurityMiddleware settings
+SECURE_HSTS_SECONDS = config('SECURE_HSTS_SECONDS', default='0', cast=int)
+SECURE_HSTS_INCLUDE_SUBDOMAINS = False
+SECURE_BROWSER_XSS_FILTER = config('SECURE_BROWSER_XSS_FILTER', default=False, cast=bool)
+SECURE_SSL_REDIRECT = config('SECURE_SSL_REDIRECT', default=not DISABLE_SSL, cast=bool)
```
jgmize: As long as we're discussing variable names, `SECURE_SSL_REDIRECT` is both redundant and inaccurate. How about shortening it to just `SECURE_REDIRECT`? If we wanted to get rid of `DISABLE_SSL` at the same time, we could change this to 
```python
SECURE_REDIRECT = config('SECURE_REDIRECT', default=True, cast=bool)
```
This wouldn't require any changes to deployed Deis apps, but it would mean changes to local dev .env files, and updating a couple of other places where we currently refer to that variable. I'm fine with leaving `DISABLE_SSL` in place for now for backwards compat and addressing that in a separate PR though.

bedrock/settings/base.py
jgmize: See comment below about this and `SECURE_SSL_REDIRECT`

pmac:
The thing is that `SECURE_SSL_REDIRECT` is a the variable that the [django middleware needs](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-SECURE_SSL_REDIRECT). So we need to use it. 

As for making the secure redirect the default this is once again our discussion about whether to make config defaults for dev or server. I suggest that it'll take a lot more of our time to make all devs change their config than it would to leave it default to non-ssl and adjust our server configs.

bedrock/settings/base.py
pmac: The thing is that `SECURE_SSL_REDIRECT` is a the variable that the [django middleware needs](https://docs.djangoproject.com/en/1.8/ref/settings/#std:setting-SECURE_SSL_REDIRECT). So we need to use it. 

As for making the secure redirect the default this is once again our discussion about whether to make config defaults for dev or server. I suggest that it'll take a lot more of our time to make all devs change their config than it would to leave it default to non-ssl and adjust our server configs.

jgmize:
Gotcha, suggestions withdrawn, let's leave everything as-is. For testing purposes, I pushed this to demo__hsts to trigger an automatic demo app creation and deployment, then I ran `deis config:set SECURE_HSTS_SECONDS=31536000 -a bedrock-demo-hsts` and validated the expected `strict-transport-security` header with curl:
```
$ curl -I https://bedrock-demo-hsts.us-west.moz.works
HTTP/1.1 301 MOVED PERMANENTLY
Cache-Control: max-age=600
Content-Type: text/html; charset=utf-8
Date: Mon, 22 Aug 2016 16:30:25 GMT
Expires: Mon, 22 Aug 2016 16:40:25 GMT
Location: https://bedrock-demo-hsts.us-west.moz.works/en-US/
Server: nginx/1.9.6
strict-transport-security: max-age=31536000
Vary: Accept-Language
X-Backend-Server: 37774b68542e.bedrock-demo-hsts
X-Robots-Tag: noodp
Connection: keep-alive
```

PR #4294, opened 2016-08-19
[fix bug 1281230] Update TestFlight page title & description

jpetto:
## Description

Update page title & description to help users find Firefox iOS Beta.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1281230

## Testing

Make sure I didn't cook any copy pasta.

## Checklist
- [x] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/testflight.html
+++ b/bedrock/firefox/templates/firefox/testflight.html
@@ -4,9 +4,23 @@
 
 {% extends "firefox/base-resp.html" %}
 
-{% block page_title_prefix %}{{ _('Firefox for iOS') }} — {% endblock %}
+{% block page_title_prefix %}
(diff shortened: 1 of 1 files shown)


alexgibson:
Changes look good to me, r+ 🍖 

/cc @peiying2 for l10n changes.

flodolo:
This pages is en-US only per bug 1266801, so you don't actually need the l10n_has_tag

alexgibson:
Bah, I did not know this :( The page being en-US only probably doesn't help raise our SEO much either, but then I'm guessing this is probably for good reason.

Jon, would you like to leave this as-is, or should we follow up and remove the superfluous tags?

alexgibson:
Opened #4296 to remove the l10n tags.

PR #4295, opened 2016-08-22
[fix bug 1297012] Remove New Relic JS snippet from base templates

alexgibson:
## Description
Removes the New Relic JS snippet from base templates to avoid accidentally enabling it.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1297012


alexgibson:
@jgmize @pmac r?

bedrock/settings/base.py
```diff
@@ -1046,9 +1046,6 @@ def lazy_email_backend():
     lambda request: request.get_full_path() == '/healthz/'
 ]
 
-NEWRELIC_BROWSER_LICENSE_KEY = config('NEWRELIC_BROWSER_LICENSE_KEY', default='')
```
alexgibson: Only thing I wasn't sure about here is if these settings are safe to remove. I can't see them being used anywhere outside of the JS snippet.

pmac:
Nice catch. Thanks Alex!

PR #4297, opened 2016-08-23
[fix bug 1295693] Add headless http status integration tests for pages that rely on external data

alexgibson:
@pmac @jgmize r?

I ended up deciding to test only the main pages that list generated content, as opposed to exhaustively parsing the links in each page as it ended up generating thousands of tests. I think this should generally give us enough confidence to alert us when something is broken, although I'd be happy to expand it a bit further if we think it's worth it.

### DIFF
--- a/tests/functional/test_generated_pages.py
+++ b/tests/functional/test_generated_pages.py
@@ -1,0 +1,67 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+import requests
+
+
+def pytest_generate_tests(metafunc):
+    if 'not headless' in metafunc.config.option.markexpr:
+        return  # test deslected by mark expression
+    base_url = metafunc.config.option.base_url
+    if not base_url:
+        pytest.skip(
+            'This test requires a base URL to be specified on the command '
+            'line or in a configuration file.')
+    paths = (
+        '/about/forums/',
+        '/about/legal/',
+        '/about/legal/terms/mozilla/',
+        '/about/legal/terms/services/',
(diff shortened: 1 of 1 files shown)


tests/functional/test_generated_content_pages.py
```diff
@@ -0,0 +1,71 @@
...
+        '/security/known-vulnerabilities/thunderbird-3.0/',
+        '/security/known-vulnerabilities/thunderbird-2.0/',
+        '/security/known-vulnerabilities/thunderbird-1.5/',
+        '/security/known-vulnerabilities/thunderbird-1.0/',
+        '/security/known-vulnerabilities/seamonkey-2.0/',
+        '/security/known-vulnerabilities/seamonkey-1.1/',
+        '/security/known-vulnerabilities/seamonkey-1.0/',
+        '/security/known-vulnerabilities/mozilla-suite/',
+        '/security/known-vulnerabilities/older-vulnerabilities/'
+        )
```
pmac: Super NIT: Deindent one level.

tests/functional/test_generated_content_pages.py
```diff
@@ -0,0 +1,71 @@
...
+    argvalues = []
+    for path in paths:
```
pmac: Could be done as a list comprehension:

```python
metafunc.parametrize('url', [base_url + path for path in paths])
```

pmac:
LGTM! Thanks Alex! I agree that this list should give us a good deal more confidence.

pmac:
Just a couple nits. You can change them or not. r+

tests/functional/test_generated_content_pages.py
alexgibson: Nice 👍

alexgibson:
Thanks @pmac updated

pmac:
👍 thanks again Alex!

PR #4299, opened 2016-08-24
[fix bug 1288779] redesign leadership page

craigcook:
## Description
Redesigns the leadership page with new headshots and adds executive bios (we'll be removing bios from the press blog). This PR looks bigger than it really is, but it's also bigger than you'd think it would be. 

Bios open in a modal so there's a lot of hidden content and slightly clunky fallback stuff for non-JS. Rather than start with the non-JS baseline and hang the JS-enabled styling from a `.js` class as I normally would, it seemed simpler here to assume JS and add fallback styling for the `.no-js` class. I can be persuaded otherwise and it may not be too hard to switch it around.

There's also some redundancy in the CSS because this is built mobile-first from pebbles, but also brings in `mozilla-modal.less` which is desktop-first from sandstone (a modal component in pebbles is coming soon so this can be tidied up).

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1288779

## Localization
As discussed in the bug, the bios add a LOT of new text that will change regularly so it's going to be an English-first page but we wanted to make it localizable for any locales that might want to take it on. The issue then will be maintaining future updates and avoiding zillions of l10n tags (the page seems to change at least monthly). We expect to be adding new bios soon so I suggest NOT exposing the new strings for l10n just yet, and removing the lang files for the old page to avoid mixed languages (a few old strings haven't changed, just added a ton of new stuff).

## Checklist
- [x] Requires l10n changes.


craigcook:
We are awaiting final headshots for Chris Beard, Chris Lawrence, and Jim Cook, as well as bios for Chris Lawrence and Angela Plohman. Hopefully we'll have those items in the next day or so. Marking WIP/DNM pending that content but we can begin initial code review.

alexgibson:
Modal seem to get cropped at certain screen widths:

modal-breakpoints

bedrock/mozorg/templates/mozorg/about/leadership.html
```diff
@@ -2,489 +2,1011 @@
...
-        <span class="title" itemprop="jobTitle">{{ _('Interim CTO and VP, Platform Engineering') }}</span>
-        <span class="note">{{ _('Steering Committee') }}</span>
-
-        <ul class="elsewhere">
-          <li><a class="url twitter" itemprop="url" href="https://twitter.com/david_bryant">@david_bryant</a></li>
-          <li><a class="url book" itemprop="url" href="https://mozillians.org/u/dbryant/">{{ _('Mozillians profile') }}</a></li>
+{% block site_header_logo %}{% endblock %}
+
+{% block content %}
+<main>
```
alexgibson: Nit: add `role="main"` here too?

media/css/mozorg/leadership.less
```diff
@@ -0,0 +1,901 @@
...
+        width: 230px;
+    }
+}
+
+.foreground {
+    display: none;
+
+    .snap img {
+        opacity: .1;
+        -webkit-animation: develop 2500ms ease-in forwards;
```
alexgibson: Nit: use `.animation()` mixin?

media/css/mozorg/leadership.less
```diff
@@ -0,0 +1,901 @@
...
+            position: absolute;
+        }
+
+        .shot-one { // MSurman
+            left: -12px;
+            top: 6px;
+            transform: rotate(-3deg);
+
+            img {
+                animation-delay: 2600ms;
```
alexgibson: Do we also need the `-webkit` vendor prefixes here?

media/css/mozorg/leadership.less
```diff
@@ -0,0 +1,901 @@
...
+
+        .shot-three {
+            left: 835px;
+            top: 90px;
+        }
+    }
+}
+
+@-webkit-keyframes develop {
+    0%   { -webkit-filter: invert(40%) sepia(100%); opacity: .1; }
```
alexgibson: Do we also need to include the un-prefixed `filter` rules here, or is it safe to assume that needing a vendor prefix for `keyframes` will also mean needing a prefix for the filter styles?

media/css/mozorg/leadership.less
```diff
@@ -0,0 +1,901 @@
...
+        z-index: 10;
+
+        .snap {
+            position: absolute;
+        }
+
+        .shot-one { // MSurman
+            left: -12px;
+            top: 6px;
+            transform: rotate(-3deg);
```
alexgibson: Do we need to use the `.transform()` mixin here too, or are we not worried about slightly older webkit browsers?

media/css/mozorg/leadership.less
```diff
@@ -0,0 +1,901 @@
...
+            text-decoration: underline;
+            color: @color-link;
+        }
+    }
+}
+
+
+// Headings
+.page-title {
+    .font-size(2.369rem);
```
alexgibson: Use `px` values for consitency, here and below?

media/css/pebbles/components/masthead.less
```diff
@@ -0,0 +1,121 @@
...
+
+@import (reference) '../lib.less';
+
+
+#masthead .content {
+    padding-top: 0;
+}
+
+.masthead-logo {
+    .font-size(1rem);
```
alexgibson: Again, we should stick to one standard unit for font-size declarations? It's `rem` here but `px` below.

media/js/mozorg/about-leadership.js
```diff
@@ -0,0 +1,108 @@
...
+    if (window.location.hash) {
+        var $target = $(window.location.hash);
+
+        if ($target.length && $target.hasClass('vcard has-bio')) {
+            $target.trigger('click');
+        }
+    }
+
+    // open external links in new windows
+    $('a[rel="external"]').on('click', function(e) {
```
alexgibson: This doesn't seem to be working, I'm guessing because of content is relocated within the modal. It should be noted however than opening links in new tabs is now considered to be both a security and performance issue. If we really need to do it, we should make sure to use `rel="external noopener noreferrer"`

alexgibson:
Not a blocker, but I notice the hover animation on the bio thumbnails feels very slow/laggy in Firefox. Almost like there's a delay on the hover animation. I don't seem to see the same issue in Chrome or Safari, so maybe it's something that Firefox struggles with :/

alexgibson:
This is a total new one to me, but it looks like IE9 supports `invert` and `sepia` through it's old proprietry `filter` implementation. Trouble is it can't animate them so the pics don't change. Should we put the animation behind a nice modern feature query using `@supports` perhaps?

ie9

alexgibson:
This is looking good 👍  - a few responsiveness tweaks on the modal needed, plus some transition sluggishness that may or may not be worth looking into.

My only other comment would be that it feels a little troubling to start using things from both pebbles and sandstone together in the same pages. I think this is fine for now, but it could be a slippery slope for technical debt if we're not careful. We seem to be using pebbles on more and more new pages now, so it would be good to get the refactor done soon (if that's still happening).

media/js/mozorg/about-leadership.js
craigcook: Oops, actually meant to remove this. Thanks for the reminder!

craigcook:
Good thinking on the feature queries, I totally need to do that and I'm not sure why I didn't think of it. (I know exactly why I didn't: because I hadn't tested IE yet).

I'm not proud about mixing sandstone and pebbles, and I had started to put together a modal style for pebbles... but I'm going to be refactoring the whole framework and converting it to Sass in the next sprint, and I kinda wanted to avoid adding too much to pebbles just yet. But you're right, it will only mean having to come back and do even more cleanup work on leadership later. I'll add a modal component to pebbles for this and it'll be easier to convert it to Sass all in one go. Future me will thank you.

media/css/mozorg/leadership.less
```diff
@@ -0,0 +1,932 @@
...
+
+    .prev {
+        left: -40px;
+        background: transparent url('/media/img/mozid/modal-left.svg') right center no-repeat;
+    }
+
+}
+
+/*
+#modal {
```
alexgibson: Can this be removed if it's no longer needed?

media/css/pebbles/includes/animations.less
```diff
@@ -0,0 +1,91 @@
...
+    }
+    90% {
+        transform: translate3d(0, 5px, 0);
+    }
+    100% {
+        transform: none;
+    }
+}
+
+@-webkit-keyframes sand-fade-in {
```
alexgibson: I remember prefixing these with `sand` (after sandstone) hastily to try and introduce some sort of namespace for the animation name as `fade-in` was quite generic. Should we rename to `pebbles-fade-in`, or something better?

media/css/mozorg/leadership.less
```diff
@@ -0,0 +1,932 @@
...
+        }
+
+        .shot-three {
+            left: 835px;
+            top: 90px;
+        }
+    }
+}
+
+@supports ((animation-name: develop) or (-webkit-animation-name: develop)) and ((filter: invert(40%)) or (-webkit-filter: invert(40%))) {
```
alexgibson: I remember using `or` in `@supports` broke when minifying when I tried it last. Have you pushed this to a demo server to make sure it still works? For the /new animation I had to use two separate `@supports` rules, one for prefixed and one for standards.

alexgibson: Yeap the animation does not seem to work on demo currently, so I suspect this might be the culprit: https://bedrock-demo-craigcook.us-west.moz.works/en-US/about/leadership/

media/css/pebbles/includes/animations.less
craigcook: Yeah this was a hasty copy-paste, I'll rename.

media/css/mozorg/leadership.less
craigcook: Yup, just a leftover I need to clean up.

media/css/mozorg/leadership.less
craigcook: Ah! That explains it! It works locally but not on demo and I was baffled as to why. I'll split it up, or maybe we can just drop the prefixes entirely and let old webkits be none the wiser? The developing effect is a totally superficial gimmick, I'd be fine with making it just a treat for the latest browsers.

craigcook:
Unblocked at last!!! It's been so long and there have been enough changes since the initial round of review that it might be best to review this again as if it's your first time seeing it.

PR #4300, opened 2016-08-24
Bug 1291359: Use referrer if no source_url for newsletters

pmac:


pmac:
@jgmize r?

pmac:
Restarted build and it's now fixed. Was "canceled" for some reason 😠 

r?

jgmize:
This is fine to merge as-is without my suggested simplification if we need (or if you just prefer) the conditional instead of `setdefault`. r+

bedrock/newsletter/views.py
```diff
@@ -581,6 +581,12 @@ def newsletter_subscribe(request):
                                                       'first_name',
                                                       'last_name', ]
                                if data[k]))
+
+            # NOTE this is not a typo; Referrer is misspelled in the HTTP spec
+            # https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.36
+            if not kwargs.get('source_url') and request.META.get('HTTP_REFERER'):
+                kwargs['source_url'] = request.META['HTTP_REFERER']
```
pmac: It'd be better if we didn't send `source_url` at all if we have no value.

pmac: And I also did it this way to make sure that we overwrite even an empty `source_url` value instead of just doing a test for the existence of the `source_url` param.

pmac:
Thanks Josh! 💯

bedrock/newsletter/views.py
jgmize: makes sense, thanks.

PR #4302, opened 2016-08-24
Add command to download, verify, and update product-details

pmac:
Downloads the files locally, checks them, then loads the modified files into the database.

### DIFF
--- a/bedrock/mozorg/management/commands/update_product_details_files.py
+++ b/bedrock/mozorg/management/commands/update_product_details_files.py
@@ -1,0 +1,93 @@
+from __future__ import print_function
+
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+from django.db import transaction
+
+from product_details.management.commands.update_product_details import Command as PDCommand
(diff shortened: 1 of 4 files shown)


pmac:
@jgmize r?

bedrock/mozorg/management/commands/update_product_details_files.py
```diff
@@ -0,0 +1,93 @@
...
+
+            self.db_storage.update('/', '', self.file_storage.last_modified('/'))
+            self.db_storage.update('regions/', '', self.file_storage.last_modified('regions/'))
+
+    def update_file_data(self, options):
+        # json dir set in settings
+        command = PDCommand()
+        command._storage = self.file_storage
+        command.is_db_storage = False
+        command.handle(**options)
```
jgmize: I know this whole thing is supposed to be a quick hack that we plan on replacing soon, and you would normally take a cleaner, less brittle approach, but I feel obligated to point out that we really shouldn't do this under normal circumstances.

pmac: Obligation complete. ✅

PR #4303, opened 2016-08-25
Fix Bug 1297571 - Updates needed to links on https://www.mozilla.org/en-US/foundation/trademarks/policy/

kyoshino:
This page is localized into Japanese only. I'll update the translation later.


alexgibson:
LGTM thanks @kyoshino r+

PR #4307, opened 2016-08-25
Data attribute additions

JamesDLorence:
## Description
-Added data attributes for download version and language on downloads on /firefox/all page. 
-Added data attributes for cta positions on cta buttons.
-Added Release Window Version to core datalayer object which pulls from the data-latest-firefox attribute on the `<html>` element on the page.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1286903
https://bugzilla.mozilla.org/show_bug.cgi?id=1286933

### DIFF
--- a/bedrock/base/templates/product-all-macros.html
+++ b/bedrock/base/templates/product-all-macros.html
@@ -84,10 +84,14 @@
 #}
 {% macro build_link(build, platform, tooltip) %}
   {% if build.platforms[platform] %}
-    <td class="download {{ platform }}"><a href="{{ build.platforms[platform].download_url }}" title="{{ tooltip }}"
-    {% if platform == 'android' %}data-link-type="download" data-download-os="Android"
-    {% elif platform == 'ios' %}data-link-type="download" data-download-os="iOS"
-    {% else %}data-link-type="download" data-download-os="Desktop"{% endif %}>{{ _('Download') }}</a></td>
+    <td class="download {{ platform }}">
+        <a href="{{ build.platforms[platform].download_url }}" title="{{ tooltip }}"
+           data-download-version="{{ platform }}" data-download-language="{{ build.name_en|lower}}" data-link-type="download"
+           {% if platform == 'android' %} data-download-os="Android"
+           {% elif platform == 'ios' %} data-download-os="iOS"
+           {% else %} data-download-os="Desktop"{% endif %}>{{ _('Download') }}
+        </a>
+    </td>
   {% else %}
     <td class="unavailable">{{ _('Not Yet Available') }}</td>
   {% endif %}
--- a/media/js/base/core-datalayer.js
+++ b/media/js/base/core-datalayer.js
@@ -40,6 +40,13 @@ if (typeof Mozilla.Analytics == 'undefined') {
         return versionResults ? versionResults[1] : null;
     };
 
+    /** Returns latest Fx version.
+    * @return {String} latest Fx version.
+    */
+    analytics.getLatestFxVersion = function() {
+        return $('html').data('latest-firefox');
+    };
+
     /** Monkey patch for dataLayer.push
     *   Adds href stripped of locale to link click objects when pushed to the dataLayer,
     *   also removes protocol and host if same as parent page from href.
--- a/tests/unit/spec/base/core-datalayer.js
+++ b/tests/unit/spec/base/core-datalayer.js
@@ -67,6 +67,23 @@ describe('core-datalayer.js', function() {
         });
     });
 
+    describe('getLatestFxVersion', function() {
+
+        afterEach(function() {
+            $('html').removeData('latest-firefox');
+        });
+
+        it('will return the Firefox version from the data-latest-firefox attribute from the html element if present', function() {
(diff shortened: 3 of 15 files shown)


JamesDLorence:
Hopefully it was ok that I squashed the two commits rather than making two different PRs.

alexgibson:
Looks like something has gone awry with your branch here, there's a second commit which is already in master (hence the catastrophic failures).

JamesDLorence:
I just rebased against mozilla:master and that seemed to have fixed it. Any idea how that happened?

bedrock/firefox/templatetags/helpers.py
```diff
@@ -11,7 +11,7 @@
 from bedrock.firefox.models import FirefoxOSFeedLink
 from bedrock.firefox.firefox_details import firefox_desktop, firefox_android, firefox_ios
 from bedrock.base.urlresolvers import reverse
-from lib.l10n_utils import get_locale
+from lib.l10n_utils import get_locale3
```
pmac: @JamesDLorence looks like this is a typo. Tests are complaining that it can't import this.

bedrock/base/templates/product-all-macros.html
```diff
@@ -84,10 +84,14 @@ <h4 class="not-found{% if builds %} hide{% endif %}">{{ _('No matching languages
...
 {% macro build_link(build, platform, tooltip) %}
   {% if build.platforms[platform] %}
-    <td class="download {{ platform }}"><a href="{{ build.platforms[platform].download_url }}" title="{{ tooltip }}"
-    {% if platform == 'android' %}data-link-type="download" data-download-os="Android"
-    {% elif platform == 'ios' %}data-link-type="download" data-download-os="iOS"
-    {% else %}data-link-type="download" data-download-os="Desktop"{% endif %}>{{ _('Download') }}</a></td>
+    <td class="download {{ platform }}">
+        <a href="{{ build.platforms[platform].download_url }}" title="{{ tooltip }}"
+           data-download-version="{{ platform }}" data-download-language="{{ build.name_en|lower}}"
+           {% if platform == 'android' %}data-link-type="download" data-download-os="Android"
```
alexgibson: We can assume that `data-link-type="download"` will always be true here, so we can move it outside of the conditional, like you have for `data-download-version` and `data-download-language`.

bedrock/firefox/templates/firefox/hello/index.html
```diff
@@ -40,15 +40,15 @@
       <header role="banner">
         <p class="notification">
           {{ _('Heads up! Firefox Hello will be removed from Firefox soon.') }}
-          <a rel="external" class="more" href="https://support.mozilla.org/kb/hello-status">{{ _('Learn more') }}</a>
+          <a rel="external" class="more" href="https://support.mozilla.org/kb/hello-status" data-cta-position="Nav">{{ _('Learn more') }}</a>
```
alexgibson: This page is being removed in #4315, so I don't think it's worth updating here in this pull request.

media/js/base/core-datalayer.js
```diff
@@ -40,6 +40,13 @@ if (typeof Mozilla.Analytics == 'undefined') {
         return versionResults ? versionResults[1] : null;
     };
 
+    /** Returns latest Fx version.
+    * @return {String} latest Fx version.
+    */
+    analytics.getLatestFxVersion = function() {
+        return document.getElementsByTagName('html')[0].getAttribute('data-latest-firefox');
```
alexgibson: Nit: we should probably retrieve the data attribute in the proper way where supported. Seeing as we can use jQuery here `$('html').data('latestFirefox')` should do it, which is shorter too.

alexgibson:
LGTM, r+wc 👍

media/js/base/core-datalayer.js
```diff
@@ -40,6 +40,13 @@ if (typeof Mozilla.Analytics == 'undefined') {
         return versionResults ? versionResults[1] : null;
     };
 
+    /** Returns latest Fx version.
+    * @return {String} latest Fx version.
+    */
+    analytics.getLatestFxVersion = function() {
+        return $('html').attr('data-latest-firefox');
```
alexgibson: Please use `.data('latestFirefox')` https://api.jquery.com/data/#data-html5

tests/unit/spec/base/core-datalayer.js
```diff
@@ -67,6 +67,23 @@ describe('core-datalayer.js', function() {
...
+            $('html').removeAttr('data-latest-firefox');
+        });
+
+        it('will return the Firefox version from the data-latest-firefox attribute from the html element if present', function() {
+            $('html').attr('data-latest-firefox', '48.0');
+
+            expect(Mozilla.Analytics.getLatestFxVersion()).toBe('48.0');
+        });
+
+        it('will return null if no data-latest-firefox attribute is present on the html element', function() {
```
alexgibson: Test says it should return `null` here but it's actually returning `undefined`? Not sure which one you're wanting, but though I should flag it.

alexgibson:
r+ thanks @JamesDLorence

PR #4311, opened 2016-08-26
Run all python tests in docker in circleci

pmac:


pmac:
@jgmize I'm sure much about how I'm doing things should change, but this is working. If you have time I'd appreciate some help getting this to a place where we could consider merging.

pmac:
@jgmize I think this is ready again. Might need to get it in sooner than later w/ the strange circle problems we've been having.

pmac:
@glogiotatidis got any time for an r? on this? I think it's pretty good now, but @jgmize is swamped.

alexgibson:
I checked this out and ran `make test` and all tests passed 👍 

I'm not qualified to review all the code here, but as far as I can tell everything works. So it gets an r+ from me at a high level.

docs/_build/.gitignore
```diff
@@ -1 +0,0 @@
-*
```
alexgibson: Did you mean to delete this? (I noticed the files we're left un-tracked after running the tests)

pmac: I didn't. Will restore. Good catch.

pmac: As it turns out these aren't required. I added an ignore for the `docs/_build` dir in the main `.gitignore` file. So I think we're okay here.

pmac:
@alexgibson thanks for the look. I think it's ready. I'd really like for @glogiotatidis to have a look before we merge just to make sure we don't mess anyone up w/ PR or demo build failures.

bin/circleci-docker-build.sh
```diff
@@ -0,0 +1,49 @@
...
+  rm -f "$MFSA_TAR_FILE"
+fi
+
+PD_TAR_FILE="$DOCKER_CACHE_PATH/pd_files.tgz"
+if [[ -f "$PD_TAR_FILE" ]]; then
+  tar xzf "$PD_TAR_FILE"
+  rm -f "$PD_TAR_FILE"
+fi
+
+if [[ "$CIRCLE_BRANCH" == demo__* ]]; then
```
glogiotatidis: Maybe out of the scope of this PR but it seems that all PRs get advantage of cached locales, database, mfsa and pd but only the ones with branch name matching `demo__*` replace the files.

Maybe the files should get deleted after extracting, only when branch name matches `demo__*`?

Makefile
```diff
@@ -0,0 +1,75 @@
...
+
+build-final: .docker-build
+	docker build -f docker/dockerfiles/bedrock_dev_final -t bedrock_dev_final .
+	touch .docker-build-final
+
+run: .docker-build
+	docker run --env-file docker/dev.env -p 8000:8000 -v "$$PWD:/app" ${DEV_IMG_NAME}
+
+django-shell: .docker-build
+	docker run -it --env-file docker/dev.env -v "$$PWD:/app" ${DEV_IMG_NAME} "./manage.py shell"
```
glogiotatidis: Remove quotes around `./manage.py shell` or this won't work

Makefile
```diff
@@ -0,0 +1,75 @@
...
+
+default:
+	@echo "You need to specify a subcommand."
+	@exit 1
+
+help:
+	@echo "build           - build docker containers for dev"
+	@echo "run             - docker-compose up the entire system for dev"
+	@echo ""
+	@echo "clean           - remove all build, test, coverage and Python artifacts"
```
glogiotatidis: Since this removes all `build` artifacts maybe it should also remove docker images?

pmac:
Great stuff! Thanks Giorgos. I'll fix it up.

bin/circleci-docker-build.sh
pmac: The idea behind caching that stuff at all is that it'll be faster to update for demo builds. PR builds don't really need the files, but they can use the cache if it's there. It's possible we could skip the extraction for PR builds and it would work fine too.

docker/run-tests.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash -xe
+
+flake8 bedrock lib tests
```
pmac: could do gulp lint here too. Have to gulp test in circle.yml because the docker container doesn't have the necessary browsers installed for those tests (and I couldn't get it working). So if I have to have the gulp stuff in circle anyway might as well do it there. I did have the flake8 thing in a separate `make lint` command, but this seemed more efficient, and it's not a bad idea to have people who run `make test` locally run it through flake8.

Makefile
```diff
@@ -0,0 +1,75 @@
...
+	-rm .docker-build
+
+test: .docker-build
+	docker run --env-file docker/test.env -v "$$PWD:/app" ${DEV_IMG_NAME} docker/run-tests.sh
+
+test-image: .docker-build-final
+	docker run --env-file docker/test.env ${FINAL_IMG_NAME} docker/run-tests.sh
+
+docs:
+	docker run --env-file docker/dev.env -v "$$PWD:/app" ${DEV_IMG_NAME} make -C docs/ clean && make -C docs/ html
```
pmac: huh... seemed to work. I'll check it out.

Makefile
glogiotatidis: If it makes sense to your workflow not to remove the images, go ahead keep it that way. What I was looking for with my comment was to question whether you actually want to remove all artifacts -which should include docker images since it's a build artifact- /or/ remove all but docker images.

Makefile
glogiotatidis: maybe you have sphinx installed on the host system as well, thus it didn't fail but it did't run in the docker container either

docker/run-tests.sh
glogiotatidis: lgtm!

pmac:
I made several improvements. It's still not a great solution for local dev, but it should work well for circle test runs and demo deployments until we move all of that to jenkins.

PR #4314, opened 2016-08-29
correct minor grammar nit ("when if")

mykmelez:
## Description

Under the Diversity and Inclusion section, in the sublist for "Some Mozillians may identify with activities or organizations that do not support the same inclusion and diversity standards as Mozilla. When this is the case:", there's an occurrence of "when if" in the sentence "when if (a) and (b) are met, other Mozillians should treat this as a private matter, not a Mozilla issue."

I suspect this was meant to be either "when" or "if" but not both, since either makes sense on its own, while both are ungrammatical.

## Bugzilla link

I haven't filed a bug but would be happy to do so if such a change rises to that level.

## Testing

I haven't tested, as I used GitHub's built-in support for editing files to fix this grammar nit.

## Checklist
- [ ] Requires l10n changes.


mykmelez:
I'm unsure if this requires l10n changes. I suppose it depends on how literally the text has been translated. Just in case, I've checked the "requires l10n changes" box.

I've also checked the "related functional & integration tests passing" box, since Circle CI tests succeeded.

flodolo:
Page is not localized, so no l10n changes.

PR #4315, opened 2016-08-31
[fix bug 1295332] Decommission and redirect Hello product page

craigcook:
## Description
Hello will be removed in Firefox 49, to be released on 13 September. This update removes the Hello product page, redirecting the /firefox/hello URL to a SUMO article at https://support.mozilla.org/kb/hello-status

This change needs go to production on 13 September once Firefox 49 is available and not before. We shouldn't merge until we're ready for the production push on the appointed day.

The only l10n impact should be cleaning up the lang files for the content we're removing, no new strings.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1295332

## Testing
Verify that the page redirects to SUMO. Also double-check that any other pages that link to the Hello page have been updated, that related static assets have been removed, tests updated/deleted, etc and that I haven't missed anything or accidentally removed anything we still need.

## Checklist
- [x] Requires l10n changes.


bedrock/firefox/redirects.py
```diff
@@ -119,9 +119,6 @@ def firefox_mobile_faq(request, *args, **kwargs):
     redirect(r'^seamonkey-transition\.html$',
              'http://www-archive.mozilla.org/seamonkey-transition.html'),
 
-    # bug 1121082
-    redirect(r'^hello/?$', 'firefox.hello'),
```
alexgibson: Any reason why these existing redirects are being removed? We should probably update their destination as opposed to remove them, to avoid serving 404's.

bedrock/firefox/redirects.py
```diff
@@ -499,9 +496,6 @@ def firefox_mobile_faq(request, *args, **kwargs):
     # bug 1260423
     redirect(r'^firefox/choose/?$', 'firefox.new'),
 
-    # bug 1273837
-    redirect(r'^firefox(?:\/\d+\.\d+(?:\.\d+)?(?:a\d+)?)?/hello/start/?$', 'firefox.hello'),
```
alexgibson: Same as above

bedrock/firefox/redirects.py
alexgibson: Ah never mind, I see this one has been re-added below 👍

bedrock/firefox/redirects.py
alexgibson: This one should definitely be retained though.

tests/redirects/map_globalconf.py
```diff
@@ -280,9 +280,6 @@
     url_test('/thunderbird/about/{board,press,staff}/',
              'http://website-archive.mozilla.org/www.mozilla.org/thunderbird/thunderbird/about/{board,press,staff}/'),
 
-    # bug 1121082
-    url_test('/hello/', '/firefox/hello/'),
```
alexgibson: This test should be kept and updated to match the redirect

tests/redirects/map_globalconf.py
```diff
@@ -1065,10 +1062,6 @@
     # Bug 1260423
     url_test('/firefox/choose', '/firefox/new/'),
 
-    # Bug 1273837
-    url_test('/firefox/{46.0,46.0.1,46.0a1,46.0a2}/hello/start', '/firefox/hello/'),
```
alexgibson: These tests should be updated too as opposed to removed.

alexgibson:
Aside from the `/hello/start/` redirect that shouldn't be removed, this all looks good to me 👍 

r+wc

alexgibson:
Oh, just noticed you can delete `'js/firefox/hello/index.js'` too.

bedrock/firefox/redirects.py
craigcook: Oops, it seemed like this was caught by the `firefox/hello` redirect and was redundant, but I must have done something wrong when I tested. I've restored this redirect and grouped it with the others.

tests/redirects/map_globalconf.py
craigcook: I wasn't sure of standard practice on keeping URL tests when effectively killing a page, but it makes sense to keep tests for the redirect. I've restored these.

tests/redirects/map_globalconf.py
```diff
@@ -1065,10 +1066,6 @@
...
-    url_test('/firefox/hello/start', '/firefox/hello/'),
```
alexgibson: Still need to restore this test too.

tests/redirects/map_globalconf.py
alexgibson: There still might be people using older versions of Firefox. The number might be small, but sending them to the SUMO page to explain why Hello no longer works is probably better than a 404 page.

alexgibson:
Looks good 👍 

r+ from me to merge when ready

PR #4321, opened 2016-09-01
[fix bug 1293779] Update VR image and stop variations on en-US home page

jpetto:
## Description

Stop serving test variations and update VR image on en-US specific home page.

en-US variant templates & assets should remain while we analyze experiment results.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1293779#c12

## Testing

Ensure no regressions?

## Checklist


bedrock/mozorg/views.py
```diff
@@ -254,15 +254,12 @@ def home_tweets(locale):
 def home(request, template='mozorg/home/home.html'):
     locale = l10n_utils.get_locale(request)
 
-    # en-US gets new template (or variation)
+    # en-US gets new template
+    # not using locale-specific templates naming scheme yet as page/test results
```
alexgibson: Incidentally, I'm not sure if it's currently possible to use a locale-specific template for en-US only, since it defaults to en-US as a fallback? At least, last time I tried to do this it didn't seem to work, but perhaps I was doing something wrong.

alexgibson:
r+ 👓

PR #4271, opened 2016-08-12
Bug 1276208: Upgrade nodejs in dockerfiles

pmac:
- Change the base image to quay.io/mozmar/base
- Upgrade prod node dependencies
- Switch to yuglify for js and css minification

(cherry picked from commit 0ab71c8ddf867c0e4727a8924c216daca3b717ba)

PR #4323, opened 2016-09-05
Convert gulpfile.js to use ES6

alexgibson:
Now that #4271 merged, we can start to do some nice things like use ES6 in our gulp file 😉 

Note: this also updates CircleCI to use node 4

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -4,7 +4,7 @@ machine:
   python:
     version: 2.7.10
   node:
-    version: 0.12.0
+    version: 4.2.6
   environment:
     # Django
     PIP_DISABLE_PIP_VERSION_CHECK: true
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -4,29 +4,29 @@
 
 /* global __dirname, require, process */
 
-var gulp = require('gulp');
-var karma = require('karma');
-var eslint = require('gulp-eslint');
-var watch = require('gulp-watch');
-var spawn = require('child_process').spawn;
+const gulp = require('gulp');
+const karma = require('karma');
+const eslint = require('gulp-eslint');
+const watch = require('gulp-watch');
+const spawn = require('child_process').spawn;
 
-var lintPaths = [
+const lintPaths = [
     'media/js/**/*.js',
     '!media/js/libs/*.js',
     'tests/unit/spec/**/*.js',
     'gulpfile.js'
 ];
 
-gulp.task('serve:backend', function () {
-    var devServerPort = process.env.PORT || 8000;
+gulp.task('serve:backend', () => {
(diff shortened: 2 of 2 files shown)


circle.yml
```diff
@@ -4,7 +4,7 @@ machine:
   python:
     version: 2.7.10
   node:
-    version: 0.12.0
+    version: 4.5.0
```
alexgibson: @pmac is this ok to set as the Node version CircleCI uses? I'm not entirely sure what version we're using in dockerfiles, other than it being 4.x.

pmac: It's the one from Ubuntu 16.04 (xenial), which appears to be 4.2.6. Looks like that one is [available on Circle](https://circleci.com/docs/build-image-trusty/#nodejs), so let's use 4.2.6.

alexgibson:
Thanks @pmac updated

gulpfile.js
```diff
@@ -4,51 +4,51 @@
...
 ];
 
-gulp.task('serve:backend', function () {
-    var devServerPort = process.env.PORT || 8000;
+gulp.task('serve:backend', () => {
+    const devServerPort = process.env.PORT || 8000;
     process.env.PYTHONUNBUFFERED = 1;
     process.env.PYTHONDONTWRITEBITECODE = 1;
-    spawn('python', ['manage.py', 'runserver', '0.0.0.0:' + devServerPort], {
+    spawn('python', ['manage.py', 'runserver', `0.0.0.0:${devServerPort}`], {
```
jpetto: 💯

gulpfile.js
```diff
@@ -4,51 +4,51 @@
...
+gulp.task('media:watch', () => {
     return gulp.src('./media/**/*')
         .pipe(watch('./media/**/*', {
             'verbose': true
         }))
         .pipe(gulp.dest('./static'));
 });
 
-gulp.task('js:test', function(done) {
+gulp.task('js:test', (done) => {
```
jpetto: Just mentioning that parens aren't necessary with a single argument. I haven't used ES6 enough to have a strong preference, but to this point I've been omitting the parens - probably just because I can. I'm good either way here.

jpetto:
Big time r+ on this one. I'll let you decide on parens/no parens for single argument arrow functions. Merge when ready!

gulpfile.js
alexgibson: I'm a bit torn on this one too. On one hand it saves a couple of bytes, but on the other hand it feels a bit inconsistent as parens are required at all other times, including when there are no arguments. I'm going to omit them for now as you pointed it out, but I've still not quite decided what my preference is here either.

PR #4325, opened 2016-09-06
[fix bug 1272635] replace manifesto mosaic with static images

craigcook:
## Description
The mosaic loads 68 separate images, which is really bad for performance. This replaces them with two static background images. Although the two image files are a bit large it's still fewer total bytes, but moreover it's down to only two HTTP requests, and the fact they're CSS backgrounds means they shouldn't be downloaded at all on small viewports (I believe inline images are still loaded with the HTML even if the CSS hides them). We sacrifice the randomizing and animation but the performance gain is worth it.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1272635


bedrock/settings/static_media.py
```diff
@@ -1558,14 +1558,6 @@
         ),
         'output_filename': 'js/tracking-protection-tour-bundle.js',
     },
-    'mosaic': {
-        'source_filenames': (
-            'js/libs/modernizr.custom.26887.js',
```
alexgibson: Lighter JS with this change too 👍

alexgibson:
LGTM, r+ 🚤

PR #4326, opened 2016-09-06
Fix Bug 1288544 - Implement mozilla.org/firefox/nightly/all to list all localized desktop Nightly builds

kyoshino:
* [Mockup](https://l10n.mozilla-community.org/~pascalc/screenshots/nightly/nightly_all_stars.png)
* [Demo server](https://bedrock-demo-nightly-all.us-west.moz.works/en-US/firefox/nightly/all/)
* No l10n required.


kyoshino:
Updated the warning copy as per request in the bug. So I think this is ready.

alexgibson:
Not related to this change, but the responsive sizing of the search panel is pretty messed up (probably from our site-wide button overhaul). I'll file a separate bug to fix it, unless it's something simple and you want to fix it here?

screen shot 2016-09-08 at 11 08 43 am

media/css/firefox/all.less
```diff
@@ -1,5 +1,42 @@
 @import "../sandstone/lib.less";
 
+body.nightly {
+    color: #FFF;
+    #outer-wrapper {
+        background: url('/media/img/firefox/horizon/stars.svg') center 60px no-repeat,
```
alexgibson: It might be worth adding a fallback here for browsers that don't support multiple backgrounds, else the text contrast becomes unreadable.

alexgibson:
Do we need to add some tests for the changes to `firefox_details.py`?

alexgibson:
This looks good to me 👍

Aside from the comments above, the page looks great. Feel free not to fix the search bar responsive styling if it turns out to be a can of inheritance worms, we can deal with it in a separate PR.

This gets an r+wc from me, although it would be good if @pmac could take a quick look at the Python changes before we merge.

kyoshino:
Nicholas also commented in the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1288544#c7) about the search button. As I replied, it can probably be hidden when JavaScript is enabled. I'll file a separate bug to deal with the button's style and behaviour.

Adding tests for the updated `firefox_details.py`...

alexgibson:
> As I replied, it can probably be hidden when JavaScript is enabled

I don't think every browser accepts the enter key as submit, there are always some edge cases. I think we should just fix the styling in a separate bug.

alexgibson:
I [filed a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1301363) for the search UI.

kyoshino:
The tests are failing now 😕 

> Action failed: set node.js version to 0.12.0

pmac:
No objections here. I've not tried it locally, but if you're satisfied @alexgibson then I am.

r+ 👍

PR #4330, opened 2016-09-07
Fix bug 1285837: Remove humans.txt

pmac:


craigcook:
A slightly disappointed r+ 😞

PR #4332, opened 2016-09-08
Fix Bug 1275483 - Create What's New page for Firefox Nightly

kyoshino:
[Demo server](https://bedrock-demo-nightly-whatsnew.us-west.moz.works/en-US/firefox/52.0a1/whatsnew/) is ready.

Note: The en-US locale is hardcoded in the MDN link, otherwise the GA query params are omitted on redirect. Until the issue is fixed at MDN side, let's keep the locale.

### DIFF
--- a/bedrock/firefox/redirects.py
+++ b/bedrock/firefox/redirects.py
@@ -24,8 +24,11 @@ redirectpatterns = (
     }),
 
     # bug 748503
-    redirect(r'^projects/firefox/[^/]+a[0-9]+/(?:firstrun|whatsnew)(?P<p>.*)$',
+    redirect(r'^projects/firefox/[^/]+a[0-9]+/firstrun(?P<p>.*)$',
              '/firefox/nightly/firstrun{p}'),
+
+    # bug 1275483
+    redirect(r'^firefox/nightly/whatsnew/?', 'firefox.nightly_firstrun'),
 
     # bug 840814
     redirect(r'^projects/firefox'
--- a/bedrock/firefox/templates/firefox/nightly_whatsnew.html
+++ b/bedrock/firefox/templates/firefox/nightly_whatsnew.html
@@ -1,0 +1,104 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{# Firefox Nightly What’s New page based on the Horizon theme for /firefox/new/ #}
+
+{% extends "/firefox/base-resp.html" %}
+{% add_lang_files "firstrun" "mobile" %}
+
+{% block extra_meta %}<meta name="robots" content="noindex">{% endblock %}
+
+{% block extrahead %}
+  {% stylesheet 'firefox_new_horizon_common' %}
+  {% stylesheet 'nightly_whatsnew' %}
+{% endblock %}
+
+{% block page_title_prefix %}{% endblock %}
+{% block page_favicon %}{{ static('img/firefox/favicon-nightly.png') }}{% endblock %}
+{% block page_title %}{{ _('You’ve just been upgraded to Firefox Nightly %(version)s!')|format(version=num_version) }}{% endblock %}
+{% block page_title_suffix %}{% endblock %}
+
+{% block body_id %}nightly-whatsnew{% endblock %}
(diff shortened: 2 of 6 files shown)


kyoshino:
The copy is being edited now. @pmac r? for the Python bits?

bedrock/firefox/redirects.py
```diff
@@ -24,9 +24,12 @@ def firefox_mobile_faq(request, *args, **kwargs):
     }),
 
     # bug 748503
-    redirect(r'^projects/firefox/[^/]+a[0-9]+/(?:firstrun|whatsnew)(?P<p>.*)$',
+    redirect(r'^projects/firefox/[^/]+a[0-9]+/firstrun(?P<p>.*)$',
```
pmac: Do we no longer need to handle "whatsnew" urls with the version number?

pmac: NM. I think I see what's up. The real view will just handle this now? But I think we shouldn't show the new page for every versioned URL right? We should redirect to the current URL?

craigcook:
Some of us discussed this page this morning and we had a few thoughts about improving the design.

* The subheading is competing a bit with the main heading, and since it's such a long piece of text it's really more of a paragraph than a title. I'd recommend making it white instead of orange and reducing the size a little bit so it's more of an introduction than a subheading.

* There is a lot of body text and each paragraph is quite short so it visually becomes a stack of thin lines. We could improve that by splitting it into two columns, possibly turning the second column into a sort of sidebar with a bulleted list. That would feel more balanced and easier on the eye, but it might also make each block of text too dense. Might need to play around with column widths and spacing to get it right.

* When the light text starts to run over the lighter background at the bottom it's harder to read. A nice way to solve this would be to use a "night" version of the background image, without the sun, or maybe with the sun just barely starting to show. That would give the whole page a darker background for contrast with the light text and also plays on the "nightly" name and the pre-release idea (before the sun rises).

These are all just presentational nitpicks and it doesn't have to be a blocker for merging this PR if we're short on time. If it's better to merge this first (once copy is final) I'd be happy to do a follow-up pull request tomorrow to touch up the design a bit.

bedrock/firefox/redirects.py
kyoshino: Chatted with @pascalchevrel and agreed to use versioned URLs like /firefox/52.0a1/whatsnew/ for flexibility. The page might literally show What's New in the future :wink:

kyoshino: This part is now handing the redirect:

```conf
    # bug 840814
    redirect(r'^projects/firefox'
             r'(?P<version>/(?:\d+\.\d+\.?(?:\d+)?\.?(?:\d+)?(?:[a|b]?)(?:\d*)(?:pre)?(?:\d)?))'
             r'(?P<page>/(?:firstrun|whatsnew))'
             r'(?P<rest>/.*)?$', '/firefox{version}{page}{rest}'),
```

pmac: Got it. Thanks 👍

pmac:
r+ on my end. I'll leave you all to work out the design nits.

kyoshino:
I'm quickly fixing the subheading and converting the body text in to 2 columns. The background could be revisited later.

kyoshino:
screen shot 2016-09-08 at 15 50 03

pascalchevrel:
I think it looks great :+1:

bedrock/firefox/templates/firefox/nightly_whatsnew.html
```diff
@@ -0,0 +1,108 @@
...
+            </p>
+            <p>
+              {% trans irc='ircs://irc.mozilla.org/nightly' %}
+                You can also interact with the Nightly community on IRC in the <a href="%(irc)s">#nightly</a> channel,
+                feel free to ask questions and talk about features, regressions or crashes you may be experiencing.
+              {% endtrans %}
+            </p>
+            <p>
+              {% trans %}
+                <strong>Go community!</strong>
```
craigcook: Should probably put the `strong` outside the `trans` so the HTML isn't part of the string.

craigcook:
I like the two columns, and that bit of text-shadow really helps too. Agreed that we can revisit the background and any other visual touches later.

Just one little nit about HTML in that string and then I think we can merge this. 🌃

kyoshino:
Solved @craigcook's comment and Marcia's feedback. We are good to go!

PR #4335, opened 2016-09-09
Fix bug 1260746: Add CSP

pmac:
CSP = Content Security Policy


craigcook:
I think we also need to whitelist youtube as a frame source. We have a few embedded videos.

pmac:
@craigcook you have example URLs with which I can test?

bedrock/settings/base.py
```diff
@@ -1069,3 +1075,51 @@ def lazy_email_backend():
...
+)
+CSP_CONNECT_SRC = (
+    '*.optimizely.com',
+    'www.googletagmanager.com',
+    'www.google-analytics.com',
+    '*.tiles.mapbox.com',
+    'api.mapbox.com',
+)
+CSP_REPORT_ONLY = config('CSP_REPORT_ONLY', default=False, cast=bool)
+if config('CSP_REPORT_ENABLE', default=True, cast=bool):
```
pmac: @glogiotatidis I think this is a good idea for lumbergh as well. It could be quite a lot of requests to the report URL if we make a mistake on a popular page.

pmac: Hmmm... might also use the setting to NOOP the view for even quicker mitigation.

craigcook:
/firefox/developer/ and the dev edition firstrun page have youtube videos on them, and there's one on the 2011 annual report. Those might be the only pages.

pmac:
@marumari this is csp work in progress for bedrock. just fyi.

alexgibson:
Could we also please add https://trackertest.org/ to the CSP whitelist? We load this as a hidden iframe on the Tracking Protection tour page in order to trigger the feature. The domain does not contain any real trackers, and is [owned by us](https://github.com/mozilla/trackertest/) (as in Mozilla).

alexgibson:
The other thing I'm not sure on is which domains to need for the FxA iframe embed. Here are the [prod values](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L960-L966), but maybe @jpetto can help fill confirm for stage / local dev.

alexgibson:
Ugh, there is also this nasty-ass tracking pixel :(

https://github.com/mozilla/bedrock/blob/master/media/js/firefox/new/scene2.js#L29

(personally I would not object to disallowing this entirely)

pmac:
> (personally I would not object to disallowing this entirely)

Agreed. Unfortunately this one is explicitly in our privacy policy so I can't really just remove it due to violation :( So much for that "we'll remove it after the australis campaign" promise 😠

pmac:
I've added more values and pushed my demo again.

* *.mozilla.com
* flashtalking.com
* creativecommons.org
* trackertest.org
* surveygizmo.com
* accounts.firefox.com(.cn)

jpetto:
On demo servers, we use `accounts.stage.mozaws.net` for FxA. We use a dev server (`stomlinson.dev.lcip.org`) for local development, but I don't think we need to be concerned there (uh, right?).

pmac:
Unless you turn on CSP locally it won't affect you. There's a setting to enable it and it is `False` by default for this reason. We should enable it for demos though. I'll add some settings so that we can alter these things per environment.

craigcook:
After a little more poking around, the dev edition pages are making use of the [youtube iframe API](https://developers.google.com/youtube/iframe_api_reference) to dynamically render the video embed, and that doesn't seem to work with youtube-nocookie.com. So I think we need to allow frames from both youtube.com and youtube-nocookie.com. In a few other places (smarton pages) we embed an iframe directly into the template and can use nocookie there just fine. For the iframe API I think we also need to accept JavaScript from youtube.com and possibly s.ytimg.com.

If we feel strongly enough about it we can rework the JS on the developer pages to avoid that third party API and roll our own iframe embed that would use the nocookie domain, but that's a less trivial update.

pmac:
Good catch @craigcook. I've updated the policy to allow those embeds to work as is. I think it's important that we get this change merged and then we can start to ratchet it back with some changes. The update is pushing to my demo now.

pmac:
So I've had to add "unsafe-eval" to the "script-src" directive because of snap.svg. I've [filed an issue with them](https://github.com/adobe-webplatform/Snap.svg/issues/475) in case it can be fixed or worked around. Barring that I wonder if we can find some other way to do what snap is doing for us?

/cc @craigcook @alexgibson

marumari:
Does the SVG contain script inside it?  You should be aware that (IIRC) Chrome will be blocking the execution of script inside SVG files unless it is delivered with a very specific MIME type.

On my personal site:
https://pokeinthe.io/2016/04/09/black-icons-with-svg-and-csp/

I have a CSP policy that is delivered exclusively for SVG files.  If the file is coming _from_ them, then their SVG policy applies, not ours.  I think.  Maybe.  I think.  Browser behavior around CSP and SVG is super messed up.

alexgibson:
> So I've had to add "unsafe-eval" to the "script-src" directive because of snap.svg. I've filed an issue with them in case it can be fixed or worked around. Barring that I wonder if we can find some other way to do what snap is doing for us?

I've not looked into what Snap is doing but I believe we're only using it on one set of pages (smarton), for a non-critical animation. If so, we could consider dropping it entirely and finding an alternative after this merges?

craigcook:
Yeah, we're only using Snap on the smarton landing page for the bouncy windowshade animation. If Snap is incompatible with our security policy then we should find a way to live without it, rather than lower our security across the board for this one script to do this one effect on this one page.

pmac:
@marumari no, it's a JS file in our repo. http://snapsvg.io/. It's for creating and manipulating SVG. But like @alexgibson suggested, perhaps we can just remove it.

marumari:
I took a brief look in their repo, but I don't see any calls to eval().  Perhaps I'm missing something, but it seems like maybe it's fixed in a newer version?

marumari:
Oh wait, it looks like there's a call to .toString.call.  It seems you can use Snap in a different way that doesn't trigger this code path:

https://groups.google.com/a/chromium.org/forum/#!msg/chromium-apps/qo0gNQ7fJ58/rRE90b6rMAAJ

pmac:
It's this call to "Function()" that's triggering it. https://github.com/adobe-webplatform/Snap.svg/blob/84fbff7d512c8145c522b71fc9c872cb0bcae49a/src/equal.js#L98

pmac:
Updated and squarshed.

PR #4336, opened 2016-09-09
[fix bug 1299336] Create fx mobile download desktop page.

jpetto:
## Description

Creates a new mobile download page intended to be reached through non-mozilla.org marketing channels.

Page requires a locale-specific App Store link and custom basket messages.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1299336

## Testing

Aside from the usual, ensure no regressions on pre-existing send to device implementations.

## Checklist


jpetto:
@alexgibson r? when you start your day Monday? Sorry for the below...

This PR is still somewhat in flux coming up on the go-live deadline of Monday, 9/12, and it's totally my fault. I didn't realize en-AU was not a supported locale in bedrock until today, which is why there are two solutions to the same problem present. The commented out view in `firefox/views.py` will almost surely be thrown out, but on the slim chance mhan decides a custom App Store link for Australia is not worth the drawbacks of using geolocation, I kept it around.

I'm waiting to hear back from [mhan in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1299336#c32), but I think review should be done first thing Monday morning regardless. I'll update this PR as soon as I hear any news (hopefully over the weekend).

Thanks, and again, sorry. 😞

jpetto:
Okay, we're rolling forward with the client-side App Store URL logic. I've removed the unused back-end code and added unit tests for the new send to device functionality.

bedrock/firefox/templates/firefox/mobile-download-desktop.html
```diff
@@ -0,0 +1,88 @@
...
+{#- This will appear as <meta property="og:image"> which can be used for social share -#}
+{% block page_image %}
+  {{ static('img/firefox/firefox-independent-1200.jpg') }}
+{% endblock %}
+
+{% block body_id %}firefox-mobile-download{% endblock %}
+
+{% block site_header %}{% endblock %}
+
+{% block site_css %}
```
alexgibson: Any reason why we're using `site_css` here and not `page_css`?

bedrock/settings/static_media.py
```diff
@@ -403,6 +403,18 @@
...
         'output_filename': 'css/firefox_mobile_download.css',
     },
+    'firefox_mobile_download_desktop': {
+        'source_filenames': (
+            'css/sandstone/sandstone-resp.less',
+            'css/tabzilla/tabzilla-static.less',
+            'css/base/simple_footer.less',
+            'css/base/send-to-device.less',
+            'css/firefox/new/horizon/common.less',
+            'css/firefox/new/horizon/scene1.less',
```
alexgibson: If we're using this horizon design more then maybe we should split it out in it's own file (not in this PR, but a separate bug). But for now, rather than include all the CSS from `/firefox/new/`, could we take only what's needed?

docs/send-to-device.rst
```diff
@@ -66,6 +66,21 @@ Usage
     var form = new Mozilla.SendToDevice();
     form.init();
 
+Geolocation Callback
+--------------------
+
+You can piggy-back on the widget's geolocation call by providing a callback function to be executed when the lookup has completed::
```
alexgibson: Nice 👍

media/css/firefox/mobile-download-desktop.less
```diff
@@ -0,0 +1,211 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../sandstone/lib.less";
+@import "../sandstone/fira-fonts.less";
```
alexgibson: Including the Fira fonts as well as Open Sans in the CSS means some browsers (*cough* IE) will download both sets of fonts. I think we should either be using Sandstone (with Open Sans) or Pebbles (with Fira), but we should generally be avoiding mixing CSS that uses both font families.

media/js/styleguide/docs/send-to-device.js
```diff
@@ -6,5 +6,10 @@ $(function() {
     'use strict';
 
     var form = new Mozilla.SendToDevice();
+
+    form.geoCallback = function(countryCode) {
+        alert('The country code is: ' + countryCode);
```
alexgibson: This can be removed?

media/js/firefox/mobile-download-desktop.js
```diff
@@ -0,0 +1,34 @@
+(function($) {
+    'use strict';
+
+    var appStoreURLs = {
+        'gb': 'https://app.adjust.com/2uo1qc?campaign=Heartbeat&adgroup=iOS_App_Store&creative=badge_de&fallback=https%3A%2F%2Fitunes.apple.com%2Fgb%2Fapp%2Fapple-store%2Fid989804926%3Fpt%3D373246%26ct%3Dadjust_tracker%26mt%3D8',
```
alexgibson: Should this be `&creative=badge_gb`? Seems there are lots of occurrences of `badge_de` below too.

alexgibson:
I'm sure this is a deliberate decision and for some reason that could make sense, but I have no idea why this is a new page and not just updating `/firefox/mobile-download/`? I get that it's aimed at desktop users for the form, but that could be hidden for mobile users pretty easy to only show the app store buttons. I'm sure you've thought all this through and probably brought it up already, but just stating for the record 😉

alexgibson:
I'm a little confused about the locales being targeted by this page. We only have a [small list of locales](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L973-L976) that are set up to generate emails via this widget. There doesn't seem to be any logic in the page to deal with this, and the list of locales in the JS also doesn't seem to match?

alexgibson:
A few questions / things to optimize, but otherwise the page looks good 👍 

My biggest questions is with regard to the locales that this page seems to be targeting, and how this differs from existing implementations of the form. Not quite sure what's correct, but if this page is different / special in some regard then it would be good to document it somewhere.

bedrock/firefox/templates/firefox/mobile-download-desktop.html
jpetto: Ah, nope. Just a short-deadline hack copied from the `mobile-download.html` template. Will fix.

bedrock/settings/static_media.py
jpetto: Yep, will cherry-pick only what's needed.

media/css/firefox/mobile-download-desktop.less
jpetto: Hm, since the widget includes Sandstone, I think we'll have to go that route. Hopefully the stakeholders will be okay with deviating from the design a little.

media/js/firefox/mobile-download-desktop.js
jpetto: I noticed this, too. All these links are taken from a doc provided by mhan. I assumed he put the links together correctly, but perhaps not. I can ask for confirmation in the bug.

media/js/styleguide/docs/send-to-device.js
jpetto: I wanted to put *something* in that function (and our linter wont allow `console.log` statements), but I suppose an `alert` is rather annoying. Will remove.

jpetto:
> I'm a little confused about the locales being targeted by this page. We only have a small list of locales that are set up to generate emails via this widget (it's also in the docs). There doesn't seem to be any logic in the page to deal with this, and the list of locales in the JS also doesn't seem to match?

The initial launch of this page will be targeted to en-* users only. The locales added here are just prep for phase 2, which will include a whole new set of basket messages.

alexgibson:
> The initial launch of this page will be targeted to en-* users only. The locales added here are just prep for phase 2, which will include a whole new set of basket messages.

Until those messages are in place, might it be worth restricting the locales this widget is shown for? Could the page get localized without us realizing?

media/js/styleguide/docs/send-to-device.js
alexgibson: You can add eslint rules to allow specific statements, including `console.log()`, however that causes issues in older browsers (which is why it's set to error by default)

jpetto:
> The initial launch of this page will be targeted to en-* users only. The locales added here are just prep for phase 2, which will include a whole new set of basket messages.
> 
> Until those messages are in place, might it be worth restricting the locales this widget is shown for? Could the page get localized without us realizing?

I asked mhan if he wanted this page restricted to en-* locales on the bedrock side, and he said no. There aren't any links to this page, aside from the marketing links mhan is pushing out (which he's limiting to en-*).

There aren't any new strings on this page, so the only localization effort is in creating the new basket messages.

media/js/styleguide/docs/send-to-device.js
```diff
@@ -2,9 +2,16 @@
...
 
+/* eslint-disable no-console */
+
 $(function() {
     'use strict';
 
     var form = new Mozilla.SendToDevice();
+
+    form.geoCallback = function(countryCode) {
+        console.log('The country code is: ' + countryCode);
```
jpetto: Because it's a local dev only page to demonstrate functionality, I left this in.

alexgibson: 👍

alexgibson:
r+ (on the basis that this page is en-US only for now)

There are a some things we need to address in the longer term:

- The send-to-device widget is currently showing for all locales. We should restrict it to the subset that is capable of receiving the newsletter once we know what this is.
- Most of the adjust links seem to point to `badge_de`. It is unclear if this is a bug or not, but seeing as the page is en-US only this should not matter for now.

PR #4337, opened 2016-09-12
[fix bug 1301721] isFirefoxUpToDate should only use the major version until FF 49 is released

alexgibson:
## Description
- Switches checks for `_isFirefoxUpToDate()` to only use the major version number until Firefox 49 is released (see bug for context).

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1301721

## Testing
- Check in both current & older Firefox (or change UA) to ensure no regressions.

### DIFF
--- a/media/js/base/mozilla-client.js
+++ b/media/js/base/mozilla-client.js
@@ -125,6 +125,52 @@ if (typeof Mozilla === 'undefined') {
     };
 
     /**
+     * Determine if user version is up to date with latest version from product details.
+     *
+     * @private
+     * @param  {Boolean} strict - if false compare the major version number only.
+     * @param  {Array} userVerArr - the user version number.
+     * @param  {Array} latestVerArr - the latest version number from product details.
+     * @return {Boolean} true if user version number is equal to or greater than product details version.
+     */
+    Client._compareVersion = function (strict, userVerArr, latestVerArr) {
+        var currentUserNumber = 0;
+        var currentLatestNumber = 0;
+        var isUpToDate = false;
+
+        // Make sure both latest and user array lengths match.
+        while (latestVerArr.length < userVerArr.length) {
+            latestVerArr.push('0');
+        }
+        while (userVerArr.length < latestVerArr.length) {
+            userVerArr.push('0');
+        }
+
+        // Only check the major version in non-strict comparison mode.
+        if (!strict) {
+            latestVerArr.length = 1;
+        }
+
+        // Step through the array from product details and compare to the user array.
+        for (var j = 0; j < latestVerArr.length; j++) {
+            currentUserNumber = Number(userVerArr[j]);
+            currentLatestNumber = Number(latestVerArr[j]);
+
+            if (currentUserNumber < currentLatestNumber) {
+                isUpToDate = false;
+                break;
+            } else if (currentUserNumber > currentLatestNumber) {
+                isUpToDate = true;
+                break;
+            } else {
+                isUpToDate = true;
+            }
+        }
+
+        return isUpToDate;
+    };
+
+    /**
      * Detect whether the user's Firefox is up to date or outdated. This data is mainly used for security notifications.
      *
      * @private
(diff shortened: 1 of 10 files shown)


kyoshino:
Yep, that's the trick used in [new/scene1.js](https://github.com/mozilla/bedrock/blob/master/media/js/firefox/new/scene1.js#L54-L56) 👍

jpetto:
All changes here look good. Tested in latest Fx and with overridden UA string and all is as expected.

Should we also remove the [`data.isUpToDate` check on scene 1](https://github.com/mozilla/bedrock/blob/master/media/js/firefox/new/scene1.js#L63) of `/firefox/new/`? If I understand the bug correctly, the code as-is would result in Fx release users on OS X or Linux incorrectly not being offered the "Refresh" button and survey.

FWIW, looking at the URLs affected in this PR on production, Fx 48.0 on OS X is treated as up to date. Is this expected?

⌛ 

Hm, after following the ridiculous path to get to a Firefox release download while using Firefox 👿 , I see 48.0.2 is now available for OS X (even though while using 48.0 I was not offered an update in the about screen). Could something have changed on the product details/release end that would make these changes unnecessary? (I definitely have not tested on Linux.)

jpetto:
Downloaded 48.0.1 and I was able to reproduce the bug. I'm sure there's a reason 48.0.1 is affected but 48.0 is not...

Also, about screen in 48.0.1 says it's up to date. 😖

alexgibson:
48.0.2 is available for download but automatic updates are throttled, so this change is still required.

Strange re: 48.0 behaving differently. Not sure what's up there will require some debugging in the morning.

jpetto:
If it helps the debugging, adrian in #webdev tested in 48.0 on Ubuntu and experienced the same behavior as I.

alexgibson:
I don't think the platform should make much difference but thanks for double checking. I think there's a bug in our client code, will see if I can track it down and fix while we're here.

alexgibson:
@jpetto - thanks for the review, I updated `scene.js` to also use `client._isFirefoxUpToDate(false)`

I also tracked down the bug in our client code (looks like due to a typo in one of our unit tests 😢). I fixed the typo and updated the logic so the tests all pass. Testing in 48.0 (against 48.0.1 as latest) it now seems to work as expected. I'd appreciate some extra review here, thanks.

media/js/base/mozilla-client.js
```diff
@@ -158,16 +158,32 @@ if (typeof Mozilla === 'undefined') {
...
+            for (var j = 0; j < latestVerArr.length; j++) {
+                userVerNumber = Number(userVerArr[j]);
+                latestVerNumber = Number(latestVerArr[j]);
+
+                if (userVerNumber < latestVerNumber) {
                     isUpToDate = false;
                     break;
+                } if (userVerNumber > latestVerNumber) {
+                    isUpToDate = true;
+                    break;
```
jpetto: The new logic here is sound as far as I can tell/test. However, it's really hard to follow this function - the complexity has exceeded reasonable variable names. If you have the time (and freshness of mind), it would be great to add some detailed comments.

tests/unit/spec/base/mozilla-client.js
```diff
@@ -458,7 +458,7 @@ describe('mozilla-client.js', function() {
 
         it('should consider outdated if user version is slightly less than latest version and strict option is true', function() {
             test(true, false, '46.0.1').toBeFalsy();
-            test(true, false, '45.0').toBeFalsy();
+            test(true, false, '46.0').toBeFalsy();
```
jpetto: Maybe add a comment here stating that it's important to test the scenario of user version 46.0 against latest version 46.0.1. (I couldn't come up with a non-awkward phrase to describe that general test case. 😕)

media/js/firefox/new/scene1.js
```diff
@@ -60,7 +60,8 @@
             if (client.isFirefoxDesktop) {
                 client.getFirefoxDetails(function(data) {
                     // data.accurate will only be true if UITour API is working.
-                    if (data.channel === 'release' && data.isUpToDate && data.accurate) {
+                    // bug 1301721 only use major Firefox version until 49.0 is released
+                    if (data.channel === 'release' && client._isFirefoxUpToDate(false) && data.accurate) {
```
alexgibson: I think calling it again here explicitly is fine, as it also serves as a reminder that we need to change it back again in the future.

media/js/base/mozilla-client.js
```diff
@@ -158,16 +158,32 @@ if (typeof Mozilla === 'undefined') {
...
-            for (var j = 0; j < userVerArr.length; j++) {
-                if (Number(userVerArr[j]) < Number(latestVerArr[j] || 0)) {
+            for (var j = 0; j < latestVerArr.length; j++) {
+                userVerNumber = Number(userVerArr[j]);
+                latestVerNumber = Number(latestVerArr[j]);
+
+                if (userVerNumber < latestVerNumber) {
                     isUpToDate = false;
                     break;
+                } if (userVerNumber > latestVerNumber) {
```
alexgibson: Bah, good catch

media/js/base/mozilla-client.js
alexgibson: Yeah this function is a bit overly long it does seem complicated. I think I may actually split the version check here out into a separate function.

tests/unit/spec/base/mozilla-client.js
alexgibson: I think the test description here sums already sums it up about as elequently as another comment could accomplish?

alexgibson:
@jpetto - split out the two functions and added comments, so hopefully the logic is a bit easier to follow. Tests still passing 👍

media/js/firefox/new/scene1.js
alexgibson: On second thought, I will remove this bug just leave the comment 👍

alexgibson: Come to think of it, we probably don't need the additional check in here at all (showing it for the major release version number is consistent with the rest of the page logic). I'll just remove it.

tests/unit/spec/base/mozilla-client.js
jpetto: Agreed. 👍

jpetto:
I'd say this one is good to go. r+

Nice (grueling) work! 😩

PR #4343, opened 2016-09-14
Fix unicode error in redirects

pmac:
Discovered via Sentry: https://sentry.prod.mozaws.net/marketing/bedrock-prod-eu-west/issues/349078/


pmac:
@jgmize r?

PR #4344, opened 2016-09-14
Fix bug 1300090: Exempt css files from softlinkstatic.py

pmac:
The content of the hashed CSS files is different from the non-hashed versions. In the former the references to other files (e.g. images) have been changed to their hashed versions.  We want to keep these versions since they are served with far-future cache headers and thus reduce our server loads.


pmac:
@jgmize r?

PR #4345, opened 2016-09-14
[bug 1292612] Remove Hello from title and desc

craigcook:
## Description
Failed to remove a few mentions of Hello when updating these pages. These are string changes so I've tagged them with `fx49_remove_hello` (cc: @flodolo and @peiying2)

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1292612

## Checklist
- [x] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/family/index.html
+++ b/bedrock/firefox/templates/firefox/family/index.html
@@ -6,7 +6,9 @@
 
 {% block page_title_prefix %}{% endblock %}
 {% block page_title %}
-{% if l10n_has_tag('tracking_protection') %}
+{% if l10n_has_tag('fx49_remove_hello') %}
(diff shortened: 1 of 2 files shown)


bedrock/firefox/templates/firefox/family/index.html
```diff
@@ -6,7 +6,9 @@
 
 {% block page_title_prefix %}{% endblock %}
 {% block page_title %}
-{% if l10n_has_tag('tracking_protection') %}
+{% if l10n_has_tag('fx49_remove_hello') %}
+  {{ _('Firefox – Desktop browser, Android, iOS, OS, Sync, Private Browsing') }}
```
jpetto: Should we drop "OS" from this string as well? (Not sure if L10n already started here...)

craigcook: Hah good catch. Since we're changing strings anyway...

PR #4346, opened 2016-09-15
[fix bug 1303076] Fix plugincheck to avoid enumerating on navigator.mimeTypes

alexgibson:
## Description
- Fixes plugincheck page in Nightly by avoiding enumerating directly on navigator.mineTypes when checking for Java version.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1303076

## Testing
- You need to have Java plugin installed to reproduce the error.
- Test that plugincheck loads in Nightly (and release for good measure)

### DIFF
--- a/media/js/plugincheck/lib/plugincheck.js
+++ b/media/js/plugincheck/lib/plugincheck.js
@@ -14,17 +14,19 @@
      */
     function setJavaVersion() {
         var type;
-        var installedVersion;
+        var installedVersion = '0.0.0'; // fallback is better than throwing an error
         var mimes = navigator.mimeTypes['application/x-java-applet'].enabledPlugin;
 
-        for (var mime in mimes) {
-            if(mimes.hasOwnProperty(mime) && mime.indexOf('jpi-version') > -1) {
-                type = mime;
-                // + 3 to strip of the leading 1. in java versions
-                installedVersion = type.substring(type.indexOf('=') + 3)
-                    .replace('_', '.');
-            }
+        var mime = Object.keys(mimes).filter(function(key) {
+            return mimes[key].type.indexOf('jpi-version') > -1;
+        });
+
+        if (mime.length > 0) {
+            type = mimes[mime[0]].type;
+            // + 3 to strip of the leading 1. in java versions
+            installedVersion = type.substring(type.indexOf('=') + 3).replace('_', '.');
         }
+
         return installedVersion;
     }
 


media/js/plugincheck/lib/plugincheck.js
```diff
@@ -16,9 +16,11 @@
         var type;
         var installedVersion;
         var mimes = navigator.mimeTypes['application/x-java-applet'].enabledPlugin;
+        var properties = Object.getOwnPropertyNames(mimes);
```
alexgibson: So this is a bit of a hack in that instead of enumerating over `navigator.MimeTypes` we're instead enumerating over an Array of own properties. This works, but I do wonder if there is a way to avoid enumerating entirely to fix this. I'd be open to suggestions?

alexgibson: Ok I think I have a better solution

alexgibson:
[Bug 1270364](https://bugzilla.mozilla.org/show_bug.cgi?id=1270364) has been backed out of Nightly for causing breakage (like what we've seen here), however they still want to re-land the patch after Firefox 51, so this still needs fixing.

To reproduce the issue whoever reviews this may need to install a slightly older Nightly to test: http://ftp.mozilla.org/pub/firefox/nightly/2016/09/2016-09-14-03-02-00-mozilla-central/

media/js/plugincheck/lib/plugincheck.js
```diff
@@ -14,17 +14,19 @@
      */
     function setJavaVersion() {
         var type;
-        var installedVersion;
+        var installedVersion = '0.0.0'; // fallback is better than returning undefined?
```
alexgibson: Using a string like `unknown` would cause an error (like the one that broke the page here). It needs to be a version number in the form of a string. I figured this is better than a broken page.

media/js/plugincheck/lib/plugincheck.js
```diff
@@ -14,17 +14,19 @@
...
-        for (var mime in mimes) {
-            if(mimes.hasOwnProperty(mime) && mime.indexOf('jpi-version') > -1) {
-                type = mime;
-                // + 3 to strip of the leading 1. in java versions
-                installedVersion = type.substring(type.indexOf('=') + 3)
-                    .replace('_', '.');
-            }
+        var mime = Object.keys(mimes).filter(function(key) {
+            return mimes[key].type.indexOf('jpi-version') > -1;
+        });
```
alexgibson: I *think* so yeah

media/js/plugincheck/lib/plugincheck.js
```diff
@@ -14,17 +14,19 @@
...
+
+        if (mime) {
```
alexgibson: Nice catch 👍

media/js/plugincheck/lib/plugincheck.js
```diff
@@ -14,17 +14,19 @@
...
+            type = mimes[mime].type;
```
alexgibson: Good call too

media/js/plugincheck/lib/plugincheck.js
jpetto: Switching to `unknown` worked for me locally, but totally understand if it should be a version number for dependent functions.

alexgibson: Hmm, let me investivate

alexgibson: You are quite right, however the page then lists the plugin as up-to-date. We could go that route, but alternatively `0.0.0` results in the plugin being listed as out-of-date. Would it be better to err on the side of caution just in case the plugin is vulnerable?

jpetto: Oh, sure enough. Listing as up to date isn't good. `0.0.0` it is. 😁

alexgibson:
@jpetto thanks for catching my slip-ups. Updated!

jpetto:
Looks good! 🔌 📥 ✔️

PR #4355, opened 2016-09-19
[fix bug 1295181] Test /all links on /firefox/new/.

jpetto:
## Description

Adds two variations to scene 1 of `/firefox/new`. Each variation adds a link underneath the primary download button that:

1. sends the user to `/firefox/all`
2. opens  a modal containing direct download links for all platforms, app store badges, and another primary download button

Understood that the JS is very much a hack right now for the modal variation. If that variation ends up winning, we'll need to figure out a more robust implementation.

Code to serve variations w/out Optimizely will come in a follow-up PR.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1295181

## Testing

The usual. Ensure new links only display when download button is present (i.e. not when using up-to-date Firefox) and only for en-US. Double check old IE.

https://bedrock-demo-jpetto-fxnew.us-west.moz.works/en-US/firefox/new/?v=1
https://bedrock-demo-jpetto-fxnew.us-west.moz.works/en-US/firefox/new/?v=2

## Checklist

### DIFF
--- a/bedrock/firefox/templates/firefox/includes/firefox-footer-links.html
+++ b/bedrock/firefox/templates/firefox/includes/firefox-footer-links.html
@@ -27,7 +27,7 @@
 
 {% if show_desktop %}
   <ul class="fx-footer-links os_desktop os_other">
-    <li><a href="{{ firefox_url('desktop', 'all', channel) }}">{{ _('Firefox for Other Platforms & Languages') }}</a></li>
+    <li><a id="fx-footer-links-desktop-all" href="{{ firefox_url('desktop', 'all', channel) }}">{{ _('Firefox for Other Platforms & Languages') }}</a></li>
     <li><a href="{{ firefox_url('desktop', 'notes', channel) }}">{{ _('Release Notes') }}</a></li>
   </ul>
 {% endif %}
--- a/media/css/base/mozilla-modal.less
+++ b/media/css/base/mozilla-modal.less
@@ -70,6 +70,7 @@ html.noscroll body {
                     .image-replaced();
                     border-radius: 50%;
                     border: 2px solid #fff;
+                    .transition(transform 0.15s ease);
 
                     &:hover,
                     &:focus {
--- a/media/css/firefox/new/all-test-variations.less
+++ b/media/css/firefox/new/all-test-variations.less
@@ -1,0 +1,190 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../../sandstone/lib.less";
+
+#fx-modal-wrapper {
+    display: none;
+}
+
+#fx-modal {
+    margin-top: 40px;
+}
+
+#fx-modal-content {
+    .clearfix();
+    background: #fff;
+    border-radius: 6px;
+    color: @textColorPrimary;
+    margin-top: 20px;
+    padding: 40px 20px;
+    text-align: center;
+
+    ul {
+        list-style: none;
+    }
+
+    li {
+        display: inline-block;
+        margin: 0;
+    }
+
+    h3,
+    h4 {
+        color: @textColorPrimary;
+    }
+
+    h3 span {
+        .font-size(18px);
+        display: block;
+        margin-top: 5px;
+        text-align: center;
+    }
+
+    .fx-privacy-link {
+        a:link,
+        a:visited {
+            color: @linkBlue;
+        }
+    }
+}
+
+.fx-modal-section {
+    .border-box();
+    padding: 20px 0;
+    width: 50%;
+}
+
+#fx-modal-secondary-cta {
+    float: left;
+    padding-right: 40px;
+}
+
+#fx-modal-primary-ctas {
+    border-left: 1px solid @textColorLight;
+    float: right;
+    padding-left: 40px;
+}
+
+#fx-modal-mobile-downloads {
+    text-align: center;
+
+    li:first-child {
+        margin-right: 10px;
+    }
+}
+
+#fx-modal-direct-downloads {
+    padding: 10px 30px 0;
+    text-align: left;
+
+    li {
+        .border-box();
+        margin-bottom: 10px;
+        width: 50%;
+
+        a {
+            .font-size(@largeFontSize);
+            background: url('/media/img/firefox/new/down-arrow.svg') left center no-repeat;
+            background-size: 20px 16px;
+            padding-left: 24px;
+        }
+    }
+}
+
+@media only screen and (max-width: @breakDesktop) {
+    #fx-modal-primary-ctas {
+        padding-left: 20px;
+        width: 60%;
+    }
+
+    #fx-modal-secondary-cta {
+        padding-right: 20px;
+        width: 40%;
+    }
+}
+
+@media only screen and (max-width: @breakTablet) {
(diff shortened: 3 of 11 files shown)


jpetto:
@alexgibson r?

bedrock/firefox/templates/firefox/includes/firefox-footer-links.html
```diff
@@ -27,7 +27,7 @@
 
 {% if show_desktop %}
   <ul class="fx-footer-links os_desktop os_other">
-    <li><a href="{{ firefox_url('desktop', 'all', channel) }}">{{ _('Firefox for Other Platforms & Languages') }}</a></li>
-    <li><a href="{{ firefox_url('desktop', 'notes', channel) }}">{{ _('Release Notes') }}</a></li>
+    <li><a id="fx-footer-links-desktop-all" href="{{ firefox_url('desktop', 'all', channel) }}">{{ _('Firefox for Other Platforms & Languages') }}</a></li>
+    <li><a id="fx-footer-links-release-notes" href="{{ firefox_url('desktop', 'notes', channel) }}">{{ _('Release Notes') }}</a></li>
```
alexgibson: Is this `id` being used?

bedrock/firefox/templates/firefox/new/horizon/scene1.html
```diff
@@ -120,9 +125,42 @@ <h2 class="good">{{_('Freedom is yours') }}</h2>
...
+            <li class="ios">
+              <a href="{{ firefox_ios_url('mozorg-fxnew_page_scene1_modal-appstore-button') }}" data-link-type="download" data-download-os="iOS">
+                <img src="{{ l10n_img('firefox/ios/btn-app-store.svg') }}" alt="{{ _('Download on the App Store') }}" width="152" height="45">
+              </a>
+            </li>
+          </ul>
+        </section>{#-- /#fx-modal-primary-ctas --#}
+
+        <section id="fx-modal-secondary-cta" class="fx-modal-section">
+          <h3>Download Firefox<span id="fx-modal-user-platform"></span></h3>
```
alexgibson: Nit: space here after `Download Firefox`?

media/css/base/mozilla-modal.less
```diff
@@ -70,6 +70,7 @@ html.noscroll body {
                     .image-replaced();
                     border-radius: 50%;
                     border: 2px solid #fff;
+                    transition: transform 0.15s ease;
```
alexgibson: Nit: use `.transition()` mixin? It makes no difference on this page, but other's could benefit from it?

media/img/firefox/new/horizon/down-arrow.svg
```diff
@@ -0,0 +1,14 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 100 100">
+  <path
```
alexgibson: You may have done this intentionally, but this svg could be minified to save on whitespece?

media/js/firefox/new/all-test-variations.js
```diff
@@ -0,0 +1,132 @@
...
+
+        return li;
+    };
+
+    var setupTest = function(version) {
+        var $dlButton = $('#download-button-desktop-release');
+        var $newLink;
+
+        // make sure desktop download button exists and user is on a recognized platform
+        if ($dlButton.length && $dlButton.find('.unrecognized-download:visible').length === 0) {
```
alexgibson: Should we add a check here to make sure it's the desktop download button that's visible, and not the mobile button?

alexgibson:
The svg arrows look a little squished in IE:

arrows

alexgibson:
This may be by design, but the link font size is a little smaller than the other privacy link:

font-size

alexgibson:
Seems to function as expected 👍 

Code is a little crazy but we knew this experiment would be such at least temporarily. The modal looks a bit broken in IE 6/7 but the platform links within it are still functional so I think that's fine? Other than the review items above, I think this looks good. r+wc 🌵

bedrock/firefox/templates/firefox/includes/firefox-footer-links.html
jpetto: It is not. I added this last minute to be consistent with its sibling, which I'm still not sure is the "right" choice. Happy to remove the unused id.

media/css/base/mozilla-modal.less
jpetto: I'm of the opinion that we deprecate that `.transition` mixin. All major browsers now support the non-prefixed syntax, it's a visual enhancement... What do you think? File a bug to drop the mixin, or should I use it here?

alexgibson: Personally I think Webkit is a bit of a different issue to evergreen desktop browsers. There are so many mobile webkit devices out there that are not updated, it doesn't seem like much hardship to keep the last vendor prefix in there. But I could be persuaded that we're not too concerned about old devices...

media/js/firefox/new/all-test-variations.js
jpetto: Ah, the CSS for the page hides the desktop download button on iOS and Android, but it wasn't preventing the variation JS from running. Will add a check here to make sure user is not on iOS or Android.

jpetto:
Yeah, font size on the link is as requested. Not too concerned with IE 6 & 7 UI issues (as long as the links work).

media/img/firefox/new/horizon/down-arrow.svg
jpetto: Ah, good catch. Will compress.

bedrock/firefox/templates/firefox/new/horizon/scene1.html
jpetto: Will fix. 👍

jpetto:
Marking as Do Not Merge until we verify no other teams need to run tests on this page at the same time.

jpetto:
Rebased with master and added Traffic Cop configuration. Would be good to have another once over to make sure all is well before merging (which is slated for Tuesday, 10/18).

media/css/firefox/new/all-test-variations.less
```diff
@@ -0,0 +1,190 @@
...
+    text-align: left;
+
+    li {
+        .border-box();
+        margin-bottom: 10px;
+        width: 50%;
+
+        a {
+            .font-size(@largeFontSize);
+            background: url('/media/img/firefox/new/horizon/down-arrow.svg') left center no-repeat;
```
alexgibson: Nit: now that we've removed the old /new template, this asset could just live at `/media/img/firefox/new/`

media/js/firefox/new/all-test-variations.js
```diff
@@ -0,0 +1,133 @@
...
+    var version = Number($('#masthead').data('version'));
+
+    // adds sort value & (optional) new text to direct download li
+    var prepLi = function(li) {
+        var $li = $(li);
+        var $liA = $li.find('a');
+        var newOsName;
+
+        switch ($liA.text()) {
+        case 'Windows':
```
alexgibson: This needn't block this PR, but just to note we're adding another download button type in #4413 for sha-1 downloads. I'll likely need to update the logic for what's displayed in this test after this merges to master.

jpetto: Yeah, I figured that new button would impact quite a few things. Hopefully wont be too much work to add it here when the time comes.

PR #4356, opened 2016-09-20
[no bug] Refactor pebbles framework and convert to Sass

craigcook:
## Description
This update solidifies the Pebbles framework, breaking it into smaller components and fleshing out a few pieces it needed. The second commit converts it from Less to Sass, with some further modularization and even a little extra refactoring where I found more things I could tweak. I kept it in two commits for easier dev and review but it might be cleaner to squash them before merging.

The at2x mixin is a little over-engineered in order to accept a complete file path with extension. Seems Sass doesn't split strings as handily as Less does so it requires a clunky loop to find the dot and count characters to either side of it. A simpler option would be to pass a file extension as another argument, but this more complex mixin makes for simpler usage and parity with the equivalent Less mixin. I'm open to discussion.

## Testing
At present we only have four pages using Pebbles (or eight if you count all four home page variants):
* the en-US home page
* /newsletter/firefox/
* /newsletter/mozilla/
* /about/leadership/

The Sass should all compile without error and those pages should all render as expected with hardly any visible differences. You'll likely need to delete your local static CSS and rerun collectstatic to make sure you have all the right pieces in the right places.

This also adds a bare-bones test page at /styleguide/pebbles/ (only in dev mode) that has nothing but the global Pebbles styling that comes with the base template.

## Checklist
- [ ] Requires l10n changes.


media/css/pebbles/includes/mixins/_at2x.scss
```diff
@@ -0,0 +1,56 @@
...
+    $hr-suffix: '-high-res';
+
+    // Set a counter and get the length of the image path.
+    $position: -1;
+    $strpath: '#{$path}';
+    $length: str-length($strpath);
+
+    // Loop through the image path and find the position of
+    // the dot to determine the file extension.
+    @for $i from $length through $length - 10 {
```
alexgibson: Took me a minute to figure out what the `$length - 10` was about. Maybe add a comment to make this a little easier to follow?

media/css/pebbles/includes/mixins/_grid.scss
```diff
@@ -0,0 +1,16 @@
+// Column spans for standard column widths
+// Usage:
+//  .foo { @include span(4); }
+
+@mixin span($columns) {
+    @include border-box;
+    float: left;
+    width: $column-width * $columns;
+    padding-left: $gutter-width/2;
```
alexgibson: Nit: space either side of `/` here and below?

media/css/pebbles/includes/mixins/_typography.scss
```diff
@@ -0,0 +1,81 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+// Font families
+@mixin fira-sans {
+    font-family: $font-fira-sans;
+}
+
+@mixin fira-sans-light {
+    font-family: 'Fira Sans Light', X-LocaleSpecific-Light, $font-fira-sans;
```
alexgibson: This may not be something we need to address here, but now we're using Fira do we still need all the extra `X-LocaleSpecific` logic in bedrock? It was added to help overcome some shortcomings in Open Sans (iirc), but Fira now has much better support for different character sets.

media/css/pebbles/includes/mixins/_typography.scss
```diff
@@ -0,0 +1,81 @@
...
+@mixin font-size($size) {
+    // When unit is pixels, pass it through and convert to rem
+    @if (unit($size) == 'px') {
+        font-size: $size;
+        font-size: remify($size);
+    }
+
+    // When unit is not pixels, show an error
+    @else {
+        @error 'This mixin only accepts sizes in pixels. You declared `font-size(#{$size})`.';
```
alexgibson: Love this 👍

media/css/pebbles/includes/mixins/_typography.scss
```diff
@@ -0,0 +1,81 @@
...
+
+@mixin font-size-huge { // For arbitrarily huge titles
+    @include font-size(50px);
+
+    @media #{$mq-tablet} {
+        @include font-size(90px);
+    }
+}
+
+@mixin font-size-h1 {
```
alexgibson: Not a blocker, but is it wise to include a reference to an actual element in the class name? I can picture things like `<h3 class="font-size-h1">Blah</h3>` which feels a bit wrong :/

alexgibson:
homepage header is missing some styling:

homepage-header

alexgibson:
As is the homepage footer:

homepage-footer

bedrock/styleguide/templates/styleguide/websites/pebbles.html
```diff
@@ -0,0 +1,155 @@
...
+        <button class="button green">green button</button>
+        <button class="button orange">orange button</button>
+        <button class="button dark">dark button</button>
+      </p>
+
+      <p>
+        <button class="button hollow">hollow button</button>
+        <button class="button hollow red">hollow red button</button>
+        <button class="button hollow dark">hollow dark button</button>
+        <span style="background: gray; padding: 10px 20px; display: inline-block;">
```
craigcook: Yeah... I just wanted to make the button visible, but didn't want to add a class for it, especially since this page has no style sheet of its own. I'll just remove the boxes and some of the buttons will be really hard to see ;)

Of course, this page doesn't *really* need to exist at all, and hopefully won't exist for long because we should start documenting this stuff in the design system.

media/css/pebbles/includes/mixins/_typography.scss
craigcook: Fira does have a really broad character set, but some languages that Fira still can't render (because it's a totally different writing system) will probably still want to specify their own font. Arabic, Japanese, etc. I think we should keep it.

media/css/pebbles/includes/mixins/_typography.scss
craigcook: Yeah, I debated this in my head a bit. I think the next best option is "font-size-level1" and so on. I think that's still clear what it means but isn't specifically an HTML element. I'll make the switch.

media/css/pebbles/includes/mixins/_at2x.scss
craigcook: To be honest, I'm not sure what it does either. I took this code from retinajs and the "from x through z" syntax is unfamiliar to me, seems it's a Sass thing. According to [their docs](http://sass-lang.com/documentation/file.SASS_REFERENCE.html#_10) I think the `$length - 10` is just a way to ensure the end point is always a lower number than the start point so the loop will always increment. Am I reading that right? I will happily add an explanatory comment once we figure out how to explain it :)

alexgibson: Hmm, wouldn't that make the function fail if the filename was greater than 10 characters?

alexgibson: Oh wait, it's just looking for the `.` right? It should work OK, just feels a bit like some random value picked from thin air.

craigcook:
Note: if #4365 merges before this does (which it should because it's higher priority), it'll cause conflicts in this PR and require some updates. Vice versa if this one merges first.

alexgibson:
Changes look good to me. r+ to merge once squashed 🎉

alexgibson:
Ok this now needs rebase after #4365 merged. I'll take another look once that's done before merging for good measure. Nearly there!

media/css/pebbles/includes/mixins/_at2x.scss
craigcook: Yeah, 10 seems like an arbitrary number. 1 might work just as well, or 100, or 23, or 42, etc. You might get a negative integer for short filenames but that's still lower so still increments the loop. I can do some testing with different paths to see if there's a good reason for using 10 instead of 1 but I'll add a comment in the meantime.

craigcook:
Rebased and refactored the home page after #4365 merged, then squashed commits. I think we're good to go but it warrants another quick pass, especially to make sure I didn't break the home page too badly.

alexgibson:
Masthead nav on the `about/leadership/` page is missing some padding.

leadership-header

alexgibson:
Aside from the one regression this looks good. r+wc 🗿

PR #4357, opened 2016-09-20
Enable x-content-type-options header

pmac:
Also default x-xss-protection to enabled. These changes will further improve our HTTP Observatory score.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -421,7 +421,8 @@ DISABLE_SSL = config('DISABLE_SSL', default=True, cast=bool)
 # SecurityMiddleware settings
 SECURE_HSTS_SECONDS = config('SECURE_HSTS_SECONDS', default='0', cast=int)
 SECURE_HSTS_INCLUDE_SUBDOMAINS = False
-SECURE_BROWSER_XSS_FILTER = config('SECURE_BROWSER_XSS_FILTER', default=False, cast=bool)
+SECURE_BROWSER_XSS_FILTER = config('SECURE_BROWSER_XSS_FILTER', default=True, cast=bool)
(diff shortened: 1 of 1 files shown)


pmac:
@jgmize r?

jgmize:
My understanding of the x-xss-protection header is that it can be useful for *disabling* xss for the browsers that support it, but I don't see the point in sending an extra header to enable something that's enabled by default. The subset of users that "X-XSS-Protection: 1" could theoretically help are only those who have actively *disabled* xss protection on those browsers that allow that (Firefox isn't one of them, AIUI).

jgmize:
Also, after scanning through https://bugzilla.mozilla.org/show_bug.cgi?id=471020 I'm finding it very difficult to justify adding the `X-Content-Type-Options: nosniff` header on www.mozilla.org. While I'd like to have a higher score on the HTTP Observatory, I think we should consider the tradeoffs before we add more headers to every single HTTP response. @marumari do you have any thoughts on the value of these headers on a site with no UGC?

pmac:
I've been told to improve our score on the observatory. So this will do that. The downsides to both of these additions seem very near zero. I vote we add them, but am happy to wait for @marumari to weigh in.

marumari:
It's obviously less important for sites without UGC, but even things like parsed GET parameters (such as utm_source) are essentially user generated.

XCTO has few downsides, provided we're setting the proper MIME types on our requests. It's unlikely to break anything.

XXSSP, I wouldn't worry about if you're working towards implementing CSP.  The Observatory exempts sites with good CSP policies from the XXSSP test.

As for the data sent, it looks like it would be about ~768 bytes for XTCO on mozilla.org, or an additional .15%.  That's if it's on HTTP/1.1.  If it's HTTP/2.0 (which most modern browsers use) additional headers are essentially free due to HPACK header compression.

jgmize:
Thanks @marumari and @pmac, all valid points.

PR #4358, opened 2016-09-20
Ping DMS after rnasync

jgmize:


jgmize:
@pmac r?

bin/cron.py
```diff
@@ -89,6 +89,7 @@ def update_security_advisories():
 
 
 @scheduled_job('interval', minutes=5)
+@ping_dms
```
pmac: Should also remove the one decorating `update_product_details`. No need to ping twice every half hour.

jgmize:
As discussed in IRC, we're going to leave the DMS snitch only on the product details update until we have better handling of recovering from invalid p-d data.

PR #4348, opened 2016-09-16
[fix bug 1299947] Firefox pre-release download pages

alexgibson:
This is still a WIP but opening now for l10n.

TODO list for myself:

- [x] iOS TestFlight graphic.
- [x] Finish responsiveness.
- [x] RTL styles.
- [x] Functional tests.

## Description
- Replaces the `/firefox/channel/` page with three new pre-release pages based on platform:

  - `/firefox/channel/desktop/`
  - `/firefox/channel/android/`
  - `/firefox/channel/ios/`

- The old `/firefox/channel/` URL redirects to `/firefox/channel/desktop/`

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1299947

## Testing
- Demo: https://www-demo1.allizom.org/firefox/channel/
- Test downloads, release notes and /all page links are correct for each channel / platform
- Test RTL styling.
- Test no-JS styling.

## Checklist
- [x] Requires l10n changes.
- [ ] Related functional & integration tests passing.

PR #4359, opened 2016-09-21
Fix Bug 1304448 - Figure out a way from mozilla.org to be able to install Firefox aurora on Google play

kyoshino:


jgmize:
@kyoshino thanks for working on this, but /firefox/channel/ page is going away as part of #4348, so it may make more sense to move this work to the new /firefox/channel/android/ page instead. Also, while I support making the play store link the primary button, I think we should preserve the direct download link as a secondary so that people who want to test aurora on devices that cannot use the google play store can still download it from www.mozilla.org.

kyoshino:
Sorry I was away! My PR won't be affected by #4348 since it just modifies the download button part. The problem is, we don't have /firefox/android/aurora/all/ yet which would provide direct APK links. Suppose 3 things to do:

1. Implement /firefox/android/aurora/all/ and /firefox/android/nightly/all/ ([Bug 1301650](https://bugzilla.mozilla.org/show_bug.cgi?id=1301650))
2. Add links to these pages to /firefox/channel/android/ (could be part of #4348, labelling "All Languages and Builds")
3. Change the primary button to link Play Store (this PR)

I'll work on the first thing first.

kyoshino:
I've combined this PR into #4382 because of the overlap in scope.

PR #4360, opened 2016-09-22
Bug 1304281

koshin43:
trees on the new Nightly "Whats New" page obstruct interacting with text on the page

https://bugzilla.mozilla.org/show_bug.cgi?id=1304281

## Checklist
- [ ] Requires l10n changes.

Fixed. A change in the left-tree width element fixes it

### DIFF
--- a/media/css/firefox/new/horizon/common.less
+++ b/media/css/firefox/new/horizon/common.less
@@ -61,7 +61,8 @@
     height: 602px;
     left: -375px;
     position: absolute;
-    width: 880px;
+    width: 909px;
+    pointer-events: none;
 }
 
 .trees-right {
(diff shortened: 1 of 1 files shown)


media/css/firefox/new/horizon/common.less
```diff
@@ -61,7 +61,7 @@
     height: 602px;
     left: -375px;
     position: absolute;
-    width: 909px;
+    width: 880px;
```
alexgibson: This is a shared template being used on multiple pages. While this does fix the bug, rather than alter the width of the graphic on multiple pages it may be better just to add `pointer-events: none;` to both `.trees-left` and `.trees-right`.

koshin43: Yeah thanks. That makes more sense. I changed it

alexgibson:
I squashed your commits, updated the commit message and merged to master in de15bccfb7f3fceccb52fd0e4c8a2b6e4a42f50d. Thanks again!

PR #4361, opened 2016-09-22
[no bug] Create traffic cop JS lib.

jpetto:
## Description

A new JS lib to perform simple traffic redirection for A/B/x testing.

Note that this is a first pass, minimum-to-be-useful version. I imagine we'll complicate this later on (or maybe not!), but for now, it just needs to redirect traffic.

## Bugzilla link

Nope.

## Testing

Create a new JS bundle that includes this lib, the newly added `mozilla-cookie-helper.js` lib (thanks MDN!), and a new JS file that configures and initializes the cop. Put that bundle in `{% extrahead %}` on the page and give it a whirl.

## Checklist


media/js/base/mozilla-traffic-cop.js
```diff
@@ -0,0 +1,174 @@
...
+    this.variations = config.variations;
+
+    // store total percentage of users targeted
+    this.totalPercentage = 0;
+
+    this.redirectVariation;
+
+    // calculate and store total percentage of variations
+    for (var v in this.variations) {
+        this.totalPercentage += this.variations[v];
```
alexgibson: As `config` is passed in, should we wrap this in a conditional that checks `hasOwnProperty` to prevent looping over all enumerable properties (just for good measure)? e.g.

```
if (this.variations.hasOwnProperty(v)) {
    this.totalPercentage += this.variations[v];
}
```

media/js/base/mozilla-traffic-cop.js
```diff
@@ -0,0 +1,174 @@
...
+        }
+    }
+};
+
+/*
+ * Ensures variations were provided and in total capture between 1 and 99%
+ * of users.
+ */
+Mozilla.TrafficCop.prototype.verifyConfig = function() {
+    if (!this.id) {
```
alexgibson: Could we check that `this.id` is a string here?

media/js/base/mozilla-traffic-cop.js
```diff
@@ -0,0 +1,174 @@
...
+        return false;
+    }
+
+    if (!this.variations) {
```
alexgibson: Could we check that `this.variations` is an Object and contains at least one key/value pair?

tests/unit/karma.conf.js
```diff
@@ -17,6 +17,7 @@ module.exports = function(config) {
             'media/js/base/mozilla-accordion-gatrack.js',
             'media/js/base/mozilla-accordion.js',
             'media/js/base/mozilla-client.js',
+            'media/js/base/mozilla-cookie-helper.js', // mozilla-traffic-cop.js
```
alexgibson: Rather than rely on this lib being present in the Karma bundle for the `mozilla-traffic-cop.js` tests to pass, I'd suggest removing it here and then mocking the `mozilla-cookie-helper` functions to return expected values for each test. This separates the dependency for testing and also means we're not creating things like cookies in the test runner.

media/js/base/mozilla-traffic-cop.js
```diff
@@ -0,0 +1,174 @@
...
+/*
+ * Checks to see if user is currently viewing a variation.
+ */
+Mozilla.TrafficCop.prototype.isVariation = function(queryString) {
+    var isVariation = false;
+    queryString = queryString || window.location.search;
+
+    // check queryString for presence of variation
+    for (var v in this.variations) {
+        if (queryString.indexOf(v) > -1) {
```
alexgibson: Should we also check that `v=b` is preceded by either `?` or `&`, to try and avoid any false positives with existing query parameters?

docs/mozilla-traffic-cop.rst
```diff
@@ -0,0 +1,104 @@
...
+--------------
+
+To add a Traffic Cop to a page, create a new JS bundle for the experiment that
+includes:
+
+- `mozilla-cookie-helper.js`
+- `mozilla-traffic-cop.js`
+- A new JS file that configures and initializes the experiment
+
+Place this new bundle in the `extrahead` block of the page
```
alexgibson: Should we also intentionally not run the lib in unsupported browsers like IE < 9? We could use a conditional comment or something similar. Seems like it could improve the reliability of test data if we avoid putting users who get no CSS and minimal JS into the testing pots.

alexgibson: Something that also may be worth considering is having the ability to turn off experiments without requiring a code change / prod push. Could we advocate placing the bundle within a switch perhaps?

alexgibson:
Say I have some variants set up like so:

```
var cop = new Mozilla.TrafficCop({
    id: 'exp_home-page-redirect-test',
    variations: {
        'v=b': 25,
        'v=c': 25
    }
});
```

If 50% of users go into the control version (no redirect), should they be prevented from going into a variation upon next visit to the page (assuming the experiment is still running)? Currently it seems like a repeat visit will enter the user into the dice roll again.

media/js/base/mozilla-traffic-cop.js
jpetto: Good suggestion. 👍

media/js/base/mozilla-traffic-cop.js
jpetto: Yessir. Will do.

media/js/base/mozilla-traffic-cop.js
jpetto: Yep, adding `Number` type check in the constructor, and will remove this check here (as the `totalPercentage` check below should be sufficient).

media/js/base/mozilla-traffic-cop.js
jpetto: Good catch.

tests/unit/spec/base/mozilla-traffic-cop.js
```diff
@@ -0,0 +1,191 @@
...
+ */
+
+/* global describe, beforeEach, afterEach, it, expect */
+
+describe('mozilla-traffic-cop.js', function () {
+    'use strict';
+
+    beforeEach(function() {
+        // stub out Mozilla.Cookie lib
+        Mozilla.Cookies = {
```
alexgibson: Instead of creating a dummy object which still hangs around after the test finishes, you can use `sinon` to create temporary stubs for each function i.e.

```
window.Mozilla.Cookies = sinon.stub();
window.Mozilla.Cookies.setItem = sinon.stub();
window.Mozilla.Cookies.getItem = sinon.stub();
window.Mozilla.Cookies.hasItem = sinon.stub();
```

bedrock/base/templates/base-new.html
```diff
@@ -7,13 +7,13 @@
   <head>
     <meta charset="utf-8">{# Note: Must be within first 512 bytes of page #}
 
+    {% block experiments %}{% endblock %}
```
alexgibson: Perhaps this should come after the main javascript `site` bundle? Doing it as early as possible is good, but if we want to give people the opportunity to opt-out of experiments (using DNT for example), then moving it after `site` bundle would enable that.

As a side thought, should it be up to each experiment to honor DNT, or should we bake this into the traffic cop lib itself? Open to suggestions as to what you think is best.

alexgibson:
This is looking great. Aside from the two points above I think this is pretty solid work. r+wc 👮

bedrock/base/templates/base-new.html
jpetto: This is an excellent point. I think the lib should respect DNT and halt initialization if DNT is enabled. Will update.

media/js/base/mozilla-traffic-cop.js
```diff
@@ -72,6 +72,11 @@ Mozilla.TrafficCop.noVariationCookieValue = 'novariation';
 Mozilla.TrafficCop.prototype.init = function() {
     var redirectUrl;
 
+    // respect the DNT
+    if (window._dntEnabled() === true) {
```
alexgibson: For an extra bit of resilience would it be worth adding check to make sure `_dntEnabled` is defined? You could also omit the `=== true` part.

alexgibson:
One last comment and looks like this needs a rebase, but otherwise I think we're good to go.

media/js/base/mozilla-traffic-cop.js
jpetto: Added a `typeof` check and an extra test. 👍

tests/unit/spec/base/mozilla-traffic-cop.js
```diff
@@ -40,6 +40,14 @@ describe('mozilla-traffic-cop.js', function () {
             cop.init();
             expect(cop.verifyConfig).toHaveBeenCalled();
         });
+
+        it('should initialize if DNT helper function is not available', function() {
+            var dntBak = window._dntEnabled;
+            window._dntEnabled = false;
```
alexgibson: If we want to assert that the lib still initializes if the `dntEnabled` function is not available, would it be better to stub it and return `undefined` instead of setting it to `false`?

jpetto: I'll defer to you here. I'm not sure on best practices for wiping out an existing global function like this.

If we did `window._dntEnabled = sinon.stub().returns(undefined);`, wouldn't the `typeof` check still return `function`?

alexgibson: The issue is here we're trying to assert one thing, but actually testing something a bit different. You're correct though, using a stub would still return `true` for the funtion check. Just changing this line to `window._dntEnabled = undefined;` should work though, right?

jpetto: Yep, that'll definitely work, and matches what we want to test better than setting to `false`.

PR #4362, opened 2016-09-23
Bug 1295248 technology landing page

alexgibson:
## Description
- Technology landing page @ `/technology/`.
- Two variations for testing, the second via `/technology/?v=b`.
- en-US only for now until we finalize strings.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1295248

## Testing
Demo: https://bedrock-demo-agibson.us-west.moz.works/en-US/technology/.

To see the blog feed you will need to update requirements, then run:

- `manage.py migrate`
- `manage.py update_blog_feeds`

## Checklist
- [ ] Requires l10n changes.


RCMainak:
Both versions looks awesome. I like the /about/technology/?v=b more it's more cleaner in my opinion if that matter :tongue:
Nicely done :+1:

alexgibson:
OK, I think this is now close enough to start code review. Still waiting on a few small visual updates to the developer links on ?v=b, but otherwise this is pretty much complete.

media/css/mozorg/technology.less
```diff
@@ -0,0 +1,1039 @@
...
+        }
+    }
+
+    .dev-links {
+        background-position: top 8px left;
+        background-repeat: no-repeat;
+        min-height: 100px;
+
+        &.mdn {
+            background-image: url('/media/img/mozorg/technology/icons-dev/mdn.svg');
```
alexgibson: Note to self: I still need to turn these into a sprite

bedrock/mozorg/templates/mozorg/technology/index.html
```diff
@@ -0,0 +1,255 @@
...
+        </ul>
+      </footer>
+    </section>
+
+    <section id="rust" class="odd">
+      <div class="container">
+        <div class="inner-container">
+          <h2>{{ _('Inventing a safer programming language') }}</h2>
+          <p>{{ _('Supported by Mozilla, Rust allows browsers, systems and more to run much faster and more safely.') }}</p>
+          <a rel="external" href="#" class="button">{{ _('Learn more') }}</a>
```
alexgibson: Note to self: need a URL here.

alexgibson: Proposed link as the blog entry isn't ready: https://developer.mozilla.org/docs/Mozilla/Rust

alexgibson:
Just had a last minute request to change the URL for this page from `about/technology` to just `technology`. Will update this on Monday.

jpetto:
Front-end code looks great. Tested in IE 8 - 10 and modern browsers.

Python bits make sense and work as expected, though maybe @jgmize wants to take a look there to see if we're missing any style points?

Do we need any extra GA stuff added (`data-*` attributes or JS), or is that all automagic these days?

## Nit-picky details that could be cleaned up (but probably aren't blockers?)

- On smaller screens, the developer resources look a little odd being left aligned. (Hard to convey in a screen cap.)

- The blog feed parser appears to have been tripped up by an HTML entity (em dash) in a post title. If any of these nit-picks is a blocker, it's this one.

screen shot 2016-10-03 at 2 16 17 pm

- Also in the image above, the ellipses are inconsistent - one in square brackets and one not.

- Ellipses placement also happens to be circumstantially odd (right after a comma and a period), but that's really get into the weeds.

alexgibson:
> Do we need any extra GA stuff added (data-* attributes or JS), or is that all automagic these days?

I believe because all the CTA buttons are external links for this page, we don't need any additional data attributes for GA in order to track clicks.

> On smaller screens, the developer resources look a little odd being left aligned. (Hard to convey in a screen cap.)

Fair suggestion, I'll look to centering the developer section on small screens 👍

alexgibson:
> The blog feed parser appears to have been tripped up by an HTML entity (em dash) in a post title. If any of these nit-picks is a blocker, it's this one.

The Jinja2 docs suggest I can use [unescape](http://jinja.pocoo.org/docs/dev/api/#jinja2.Markup.unescape) to resolve HTML entities. I'll defer to @pmac If my fix is appropriate though, there may be something better.

alexgibson:
> Ellipses placement also happens to be circumstantially odd (right after a comma and a period), but that's really get into the weeds.

After a bit of digging it looks like we don't insert the ellipses in the summary ourselves. This is already in the respective blog feeds (e.g. https://hacks.mozilla.org/feed/), so we can't control exactly where it occurs. The Hacks blog seems to use a slightly different convention using square brackets to the other blogs in the feed. We could try and do our own ellipses, or try and standardize them, but that may be a little more fiddly than it's worth.

pmac:
> We could try and do our own ellipses, or try and standardize what's already there, but that may be a little more fiddly than it's worth.

That's what I decided as well, and I mentioned it last week in a standup and Jen was cool with that too. We can always try to do something about it later if it ends up being a big deal to someone.

jpetto:
> We could try and do our own ellipses, or try and standardize what's already there, but that may be a little more fiddly than it's worth.
> 
> That's what I decided as well, and I mentioned it last week in a standup and Jen was cool with that too. We can always try to do something about it later if it ends up being a big deal to someone.

Agree that it's not worth the effort at this point.

jgmize:
The [circleci failure](https://circleci.com/gh/mozilla/bedrock/2357) is due to `requirements/docker-dev.txt` having gone missing since being added in #4010. The file is missing in master as well.

pmac:
Doh! I did that. I'll remove the reference. PR incoming.

jgmize:
@pmac I pushed a fix to this branch before I saw your comment about an incoming PR.

pmac:
@jgmize great. that works too. thanks!

bedrock/mozorg/management/commands/update_blog_feeds.py
```diff
@@ -0,0 +1,48 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument('-q', '--quiet', action='store_true', dest='quiet', default=False,
+                            help='If no error occurs, swallow all output.'),
+        parser.add_argument('--database', default='default',
+                            help=('Specifies the database to use, if using a db. '
+                                  'Defaults to "default".')),
+        parser.add_argument('--articles', default=5, type=int,
+                            help='Number of articles to store from each feed. Defaults to 5.')
+
+    def handle(self, *args, **options):
```
pmac: That's a good call.

PR #4365, opened 2016-09-23
[fix bug 1302772] new home page

craigcook:
## Description
What was variant A of home page prototype B is officially the new mozorg home page!

* rename home-en-US.html to home.html
* rename old home.html to home-voices.html
* rename associated CSS/JS
* rename img/home/new to img/home/2016
* delete some old images no longer in use

Strings for the new home page will live in 'mozorg/home/home-2016'

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1302772

## Testing
When testing this locally you'll probably need to clear your local static files and rerun `collectstatic` to make sure you pick up the renamed assets.

The new home page should be shown to all locales in dev mode. If DEV = False, the default locale (en-US) gets the new page and all other locales get the old home page until their strings are complete: `lang_file_is_active('mozorg/home/home-2016')`

## Checklist
- [x] Requires l10n changes.


craigcook:
Note that `tests/functional/test_home.py` references the old home page, but those tests are being skipped and have been for some time. We'll update home page tests in a later PR.

bedrock/mozorg/views.py
```diff
@@ -254,11 +254,10 @@ def home_tweets(locale):
 def home(request, template='mozorg/home/home.html'):
     locale = l10n_utils.get_locale(request)
 
-    # en-US gets new template
-    # not using locale-specific templates naming scheme yet as page/test results
-    # are still in flux
-    if locale == 'en-US':
-        template = 'mozorg/home/home-en-US.html'
+    if lang_file_is_active('mozorg/home/home-2016', l10n_utils.get_locale(request)):
```
flodolo: This works for me, but my understanding from the discussion with Jen (and the bug) was that we wanted to drop the old homepage and redirect to en-US for incomplete locale (on Oct 28).

TheoChevalier:
Any chance we could add a CSS rule for non-English locales, or wider the title a little bit? Or reduce font-size? It will be quite hard to make it fit on two lines AND add a break after the comma.

Current French translation (can’t guarantee it will be the final one, but right now the wrapping is quite unfortunate)
capture d ecran 2016-09-26 a 12 22 51

bedrock/mozorg/tests/test_views.py
```diff
@@ -48,20 +48,20 @@ def test_gets_default_mobilizer_url(self, resp_mock):
...
+    # testing home page
     def test_en_us(self, render_mock):
         req = RequestFactory().get('/')
         req.locale = 'en-US'
         views.home(req)
         render_mock.assert_called_once_with(req,
-            'mozorg/home/home-en-US.html', ANY)
+            'mozorg/home/home.html', ANY)
 
     def test_non_en_us(self, render_mock):
```
alexgibson: Instead of testing non-en-US here (which could fail as soon as `es-ES` is translated?) you can mock `lang_file_is_active` by pasting `@patch.object(views, 'lang_file_is_active', lambda *x: False)` above `def test_non_en_us`. You could rename the test to something like `test_home_old_template` too.

media/css/mozorg/home/home.less
```diff
@@ -2,1107 +2,1010 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-@import "../../sandstone/lib.less";
-@import "home-grid.less";
+@import '../../pebbles/lib.less';
+@import "../../firefox/horizon-lib.less";
```
alexgibson: Nit: single quotes for consistency

media/css/mozorg/home/home.less
```diff
@@ -2,1107 +2,1010 @@
...
 
-        .featured-event {
-            width: 100%;
-            margin-right: 0;
+        &.visible {
+            display: block;
+        }
+    }
+}
+
```
alexgibson: Nit: remove extra space

bedrock/mozorg/views.py
alexgibson: We could still technically do that here though, without any code changes right? Seems a shame to not show any locales the new page until Oct 28 even though they have translated. This way at least they can roll in a bit early?

flodolo: Absolutely yes, just wanted to make sure it was considered 🍰

craigcook:
Note: if #4356 merges before this does (which it shouldn't because it's lower priority), it'll trigger merge conflicts in this PR and require some updates. And vice versa if this merges first (which it should).

craigcook:
I've adjusted the sizes a bit to hopefully work better for most locales as a baseline (wider box, slightly smaller font) and some extra rules to enlarge the font just for English. I expect other locales could handle the larger font just as well, but assuming long strings for the baseline is safest. We can make other locale-specific tweaks if we need to.

alexgibson:
LGTM, r+ 🍺

PR #4368, opened 2016-09-27
[fix bug 1297399] Add preferences screen to email confirm page.

jpetto:
## Description

Add the subscription management form to the newsletter confirmation page.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1297399

## Testing

Ensure form works on `/newsletter/confirm/` and `/newsletter/existing/`.

## Checklist
- [ ] Requires l10n changes.


jpetto:
@pmac - Still a WIP, but when you have time, can you sanity check the approach here?

pmac:
I've not done a thorough review nor run it locally, but the direction seems solid. Glad we can mostly just use the same form on both pages.

Another way we could go is to use a single view for both of these pages, but pass in a parameter from `urls.py` that lets it know to use a different template for confirm vs. existing. Either way works though.

flodolo:
Are the strings in this PR final? I have a request to expose newsletter to other locales, and I would definitely prefer to include this one in the file before doing it.

jpetto:
> Are the strings in this PR final? I have a request to expose newsletter to other locales, and I would definitely prefer to include this one in the file before doing it.

[According to the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1297399#c15), the strings are final.

jpetto:
Closing in favor of #4378.

PR #4373, opened 2016-09-27
[fix bug 1305799] Update Mitchell's bio.

jpetto:
## Description

Updates Mitchell's bio to match that on the press center blog.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1305799

## Testing

Ensure no typos?

Page is not localized, so shouldn't be an issue changing strings.

## Checklist

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/leadership.html
+++ b/bedrock/mozorg/templates/mozorg/about/leadership.html
@@ -90,20 +90,34 @@
         <div class="person-bio">
           <p>
           {% trans %}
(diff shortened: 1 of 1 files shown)


jpetto:
@alexgibson - r?

bedrock/mozorg/templates/mozorg/about/leadership.html
```diff
@@ -90,21 +90,35 @@ <h3 class="group-title" id="mgmt-corp"><a href="{{ url('foundation.moco') }}">{{
...
+          {% endtrans %}
+          </p>
+
+          <p>
+            <em>
+            {% trans %}
+              TIME Magazine profiled Mitchell in its global list of “100 Most Influential People” under “Scientists and Thinkers”. Bloomberg listed her as one of the “25 Most Influential People on the Web”. She was honored as winner of the Anita Borg Institute’s 2009 Women of Vision Award and received the Aenne Burda Award for Creative Leadership in 2010. In 2012, Mitchell was inducted into the founding group of the <a rel="external" href="http://www.internethalloffame.org/inductees/mitchell-baker">Internet Society’s Hall of Fame</a>. She has appeared on NBC “Meet the Press”, BBC “HardTalk”, “The Charlie Rose Show” and “CNN Global Office”. She has spoken at high-level events like the World Economic Forum, Mobile World Congress, The Web Summit, Collision, Singularity University and Bits&Pretzels.
```
alexgibson: Nit: the `href` here should probably not be part of the string and passed in as a variable instead.

alexgibson:
r+ - aside from the one nit above the change looks good. Because we're not localizing this page, I think it's fine not to block merging and leave the `href` as part of the string for now. We can always update it in the next change.

PR #4376, opened 2016-09-29
Give ability to run supervisor for dev

pmac:
This will allow us to run supervisord in www-dev that will run the site and a process to keep l10n updated. It allows us to run this without also running the other cron tasks that update the DB in every container, since we use a separate clock container for that. Demo instances will also run the DB update process via supervisor.

The idea is that if you run `docker/run-supervisor.sh` and `DEV=True` in your environment, it will run the site and the `bin/cron_l10n.py` script. If you also set `DB_CRON=True` in your environment, it will add `bin/cron_db.py` to the processes. This should only really be used in demos and other single process instances. Other instances (like www-dev.allizom.org) use a real DB and have a separate docker container running just the `cron_db.py` script.

### DIFF
--- a/.bedrock_demo_env
+++ b/.bedrock_demo_env
@@ -4,6 +4,7 @@ CSP_EXTRA_FRAME_SRC=*.mozaws.net
 DEBUG=False
 DATABASE_URL=sqlite:///bedrock.db
 DEV=True
+DB_CRON=True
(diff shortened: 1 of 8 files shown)


pmac:
@glogiotatidis mind taking a look at this too? 

cc @flodolo

PR #4378, opened 2016-09-29
[fix bug 1297399] Show pref form on email confirmation.

jpetto:
## Description

Add subscription confirmation handling to `/newsletter/existing/`.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1297399

## Testing

Make sure newsletter preferences can be updated when confirming and not confirming subscription.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/newsletter/templates/newsletter/existing.html
+++ b/bedrock/newsletter/templates/newsletter/existing.html
@@ -15,20 +15,39 @@
 
   {% block main_feature %}
     <div id="main-feature">
+    {% if did_confirm == '1' %}
+      <h1>{{ _('Thanks for Subscribing!') }}</h1>
+
+      {% if l10n_has_tag('newsletter_confirm_2016') %}
+        <p>
+          {{ _('You’re awesome!') }}
+          {{ _('And we’re not just saying that because you trusted us with your email address.') }}
+          {{ _('Please be sure to add mozilla@e.mozilla.org to your address book to ensure we always reach your inbox.') }}
+        </p>
+
+        <p>
+          {{ _('Mozilla touches on a variety of important issues.') }}
+          {{ _('Open your inbox (and your heart) even more — take a look at other topics we cover.') }}
+        </p>
+      {% else %}
+        <p>{{ _('Your newsletter subscription has been confirmed.') }}</p>
+        <p>{{ _('Please be sure to add our sending address: mozilla@e.mozilla.org to your address book to ensure we always reach your inbox.') }}</p>
+      {% endif %}
+    {% else %}
       <h1>{{ _('Manage your Email Preferences') }}</h1>
 
-    {% if switch('newsletter-maintenance-mode') %}
-      <p class="billboard">
-        {{ _('This page is in maintenance mode and is temporarily unavailable.') }}<br>
-        {{ _('To update your email preferences, please check back in a little while. Thanks!') }}
-      </p>
-    {% else %}
-      <p>
-        {{ _('We love sharing updates about all the awesome things happening at Mozilla.') }}<br>
-        {{ _('Set your preferences below to make sure you always receive the news you want.') }}
-      </p>
+      {% if switch('newsletter-maintenance-mode') %}
+        <p class="billboard">
+          {{ _('This page is in maintenance mode and is temporarily unavailable.') }}<br>
+          {{ _('To update your email preferences, please check back in a little while. Thanks!') }}
+        </p>
+      {% else %}
+        <p>
+          {{ _('We love sharing updates about all the awesome things happening at Mozilla.') }}<br>
+          {{ _('Set your preferences below to make sure you always receive the news you want.') }}
+        </p>
+      {% endif %}
     {% endif %}
-
     </div>
   {% endblock %}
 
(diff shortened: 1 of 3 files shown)


jpetto:
@pmac - Trying the inverse approach - moving complexity into the template instead of the view layer. The template isn't quite a 🍝 dish yet, but is on its way. However, I do find this a simpler solution to the request at hand than #4368. As of now, I'm in favor of this approach. We can refactor if we end up needing the preferences form on more pages. What do you think?

If this is the way to go, should we get rid of the `/confirm` template and view logic?

pmac:
I agree that this looks pretty great. I think this will be fine and we need not worry about needing to include the form in other pages. I was thinking that we'd keep the confirm logic in the confirm view and just send the redirect to `/existing/{token}/?confirmed=1` on success (or rate limit), and display an error on error, but your way works just as well. Though, I guess we really shouldn't show them the whole form if there was an error confirming...

What do you think?

jpetto:
> I was thinking that we'd keep the confirm logic in the confirm view and just send the redirect to /existing/{token}/?confirmed=1 on success (or rate limit), and display an error on error

This is even better. Makes both the `existing` template and view layer less complex.

jpetto:
@pmac - Should be ready for review when you have the time. Thanks!

bedrock/newsletter/views.py
```diff
@@ -219,12 +222,18 @@ def confirm(request, token):
...
-    return l10n_utils.render(
-        request,
-        'newsletter/confirm.html',
-        {'success': success,
-         'generic_error': generic_error,
-         'token_error': token_error})
+    # Assume rate limit error means user already confirmed and clicked confirm
+    # link twice in quick succession
+    if success or rate_limit_error:
+        print '*** success or rate limit! ***'
```
pmac: remove this before merge

bedrock/newsletter/views.py
```diff
@@ -427,17 +436,18 @@ def existing(request, token=None):
             newsletter_languages[lang].append(newsletter)
     newsletter_languages = mark_safe(json.dumps(newsletter_languages))
 
-    # We also want a list of the newsletters the user is already subscribed
-    # to
+    # We also want a list of the newsletters the user is already subscribed to
     already_subscribed = mark_safe(json.dumps(user['newsletters']))
 
     context = {
+        'did_confirm': request.GET.get('confirm', None),
```
pmac: Let's send a bool to the template.

```python
    'did_confirm': request.GET.get('confirm', None) == '1',
```

Then you can just

```jinja
{% if did_confirm %}
...
```

bedrock/newsletter/views.py
```diff
@@ -219,12 +222,18 @@ def confirm(request, token):
...
-        request,
-        'newsletter/confirm.html',
-        {'success': success,
-         'generic_error': generic_error,
-         'token_error': token_error})
+    # Assume rate limit error means user already confirmed and clicked confirm
+    # link twice in quick succession
+    if success or rate_limit_error:
+        print '*** success or rate limit! ***'
+        return HttpResponseRedirect("%s?confirm=1" % (reverse('newsletter.existing.token', kwargs={'token': token})))
```
pmac: NIT: superfluous parens around `reverse`.

bedrock/newsletter/views.py
```diff
@@ -219,12 +222,18 @@ def confirm(request, token):
             # but just in case:
             generic_error = True
 
-    return l10n_utils.render(
-        request,
-        'newsletter/confirm.html',
-        {'success': success,
-         'generic_error': generic_error,
-         'token_error': token_error})
+    # Assume rate limit error means user already confirmed and clicked confirm
```
pmac: Thanks for this comment 👍

bedrock/newsletter/templates/newsletter/existing.html
```diff
@@ -15,20 +15,39 @@
 
   {% block main_feature %}
     <div id="main-feature">
-      <h1>{{ _('Manage your Email Preferences') }}</h1>
+    {% if did_confirm == '1' %}
```
pmac: With my suggested change below, this would just be `{% if did_confirm %}`. It's just a bit safer to avoid sending user input (from the url) straight to the template context.

bedrock/newsletter/views.py
jpetto: D'oh!

bedrock/newsletter/views.py
jpetto: Ah, yes. Nice.

bedrock/newsletter/tests/test_views.py
```diff
@@ -132,7 +132,7 @@ def test_will_show_confirm_copy(self, get_newsletters, mock_basket_request):
                 render.return_value = HttpResponse('')
                 self.client.get(url)
         request, template_name, context = render.call_args[0]
-        self.assertEqual(context['did_confirm'], '1')
+        self.assertEqual(context['did_confirm'], True)
```
pmac: NIT: use `assertTrue` here, or even just `assert context['did_confirm']`.

PR #4384, opened 2016-10-03
Fix Bug 1307091 - Change the redirect from en-US/firefox/nightly/ to the new pages

kyoshino:


bedrock/firefox/redirects.py
```diff
@@ -43,7 +43,8 @@ def firefox_mobile_faq(request, *args, **kwargs):
     redirect(r'^firefox/accountmanager/?$', 'https://developer.mozilla.org/Persona'),
 
     # bug 841846
```
alexgibson: Nit: please append the bug number along with the existing one here so we have some paper trail :)

tests/redirects/map_globalconf.py
```diff
@@ -797,7 +797,7 @@
     url_test('/persona{/{about,developer-faq}/,}', 'https://developer.mozilla.org/Persona'),
 
     # bug 841846
```
alexgibson: Same here please

bedrock/firefox/redirects.py
kyoshino: Yeah noticed that ;) Fixed!

kyoshino:
Thanks!

PR #4385, opened 2016-10-03
Remove and clone locale dir if not a repo in l10n_update

pmac:


pmac:
This got an r+ when it was #4244. Let's just see if the tests pass and I think we can get this in.

pmac:
This context for this is my work on getting l10n to update automatically on dev in #4376. Turns out the `locale` dir in the image deployed to the dev instance does not include the `.git` repo (and rightly so). So when updates are needed we should clone it and go from there.

PR #4386, opened 2016-10-03
Switch to webdev user before copying locale

jgmize:
@pmac r?

### DIFF
--- a/docker/dockerfiles/bedrock_l10n
+++ b/docker/dockerfiles/bedrock_l10n
@@ -1,2 +1,5 @@
 FROM ${FROM_DOCKER_REPOSITORY}:${GIT_COMMIT}
 COPY . /app/locale
+USER root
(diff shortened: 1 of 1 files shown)


docker/dockerfiles/bedrock_l10n
```diff
@@ -1,2 +1,5 @@
 FROM ${FROM_DOCKER_REPOSITORY}:${GIT_COMMIT}
 COPY . /app/locale
+USER root
+RUN chown -R webdev /app/locale
```
pmac: change to `chown -R webdev.webdev /app/locale` and I think we're good. The current code does change the owner but the group stays `root`.

jgmize: Good catch. amended to also change the group

pmac:
Thanks Josh!

PR #4387, opened 2016-10-04
Remove reference to requirements/docker-dev.txt

pmac:
The file was merged into requirements/docker.txt


pmac:
This will be fixed in #4362.

PR #4388, opened 2016-10-04
[fix bug 1304086] Test home page link styles

craigcook:
## Description
Testing some variant styling of section links on the home page, hoping they draw more attention and get more clicks. The variants have solid circles around the icons along with trailing arrows, with some animation on hover (the dots embiggen and the arrow moves). Variant B has a fade-in animation when the links scroll into view. The changes are strictly visual, no new strings.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1304086

## Testing
?v=a for Variant A (no fade)
?v=b for Variant B (fade in)

This uses our new traffic cop JS library to handle the redirects, with 25% to variant A, 25% to variant B, and the remaining 50% to the control. You'll need to disable DNT to trigger the redirect, and clear cookies (or start a new session) to get a different variant.

## Checklist
- [ ] Requires l10n changes.


media/css/mozorg/home/home-variant.scss
```diff
@@ -0,0 +1,104 @@
...
+            display: inline-block;
+            width: 45px;
+            height: 45px;
+            padding: 5px;
+            background-color: silver;
+            border-radius: 50%;
+            position: absolute;
+            left: -25px;
+            top: -12px;
+            transition: left 150ms ease, top 150ms ease, padding 150ms ease;
```
jpetto: Maybe it's my confirmation bias, but I swear this transition is a *little* choppy, particularly in Chrome. How awful would it be to de-couple the circular background from the icon and transition a `transform: scale()`?

This is a test, so I don't want to be too picky. It could wait on a winner being chosen before being cleaned up if it's a significant undertaking.

media/css/mozorg/home/home-variant.scss
```diff
@@ -0,0 +1,104 @@
...
+        color: #d93a33;
+    }
+
+    #who .icon {
+        background-color: #d93a33;
+
+        use {
+            fill: #fff;
+        }
+    }
```
jpetto: Could nest under a single `#who` selector?

media/css/mozorg/home/home-variant.scss
```diff
@@ -0,0 +1,104 @@
...
+                display: none;
+            }
+
+            &:hover .icon,
+            &:focus .icon {
+                padding: 7px;
+                left: -27px;
+                top: -14px;
+            }
+        }
```
jpetto: Could nest under a single `.icon-links` selector?

bedrock/settings/static_media.py
```diff
@@ -623,6 +623,13 @@
         ),
         'output_filename': 'css/home.css',
     },
+    'home-variant': {
+        'source_filenames': (
+            'css/newsletter/moznewsletter-subscribe.less',
+            'css/mozorg/home/home-variant.scss',
+        ),
+        'output_filename': 'css/home-variant.css',
```
jpetto: For consistency, `css/home-variant-bundle.css`? (Same with pre-existing `css/home.css` above...)

craigcook: doh, you're right of course. Will fix.

media/css/mozorg/home/home-variant.scss
craigcook: Good idea. I think repainting the padding while also shifting position makes for a tiny bit of jank, but I didn't want to scale the whole icon (and also wanted to avoid wrapping it a span). But I think I can use a pseudo-element for the solid circle and the image can just be on top of it. I think it'll work and should hopefully be smoother, and a transform-origin means I won't have to mess with repositioning.

media/css/mozorg/home/home-variant.scss
```diff
@@ -0,0 +1,121 @@
...
+                left: -25px;
+                position: absolute;
+                top: -12px;
+                width: 55px;
+            }
+
+            &:hover:before,
+            &:focus:before {
+                transform: scale(1.15);
+                transform-origin: 50% 50%;
```
jpetto: Is it worth using the mixins here for vendor prefixes?

jpetto:
Now those are some smooth transitions! One question on using mixins for vendor prefixes, but otherwise looks good.

media/css/mozorg/home/home-variant.scss
craigcook: Couldn't hurt. Silly old webkit...

PR #4389, opened 2016-10-04
Improve code and tests for blog feed fetching/parsing

pmac:
Refs Bug 1303794

### DIFF
--- a/bedrock/mozorg/models.py
+++ b/bedrock/mozorg/models.py
@@ -1,11 +1,13 @@
 from django.conf import settings
 from django.core.cache import cache
 from django.db import models
+from django.db import transaction
 from django.db.models.signals import post_save
 from django.db.utils import DatabaseError
 from django.dispatch import receiver
 from django.utils.html import strip_tags
 
+from feedparser import parse
 from jinja2 import Markup
 
 from picklefield import PickledObjectField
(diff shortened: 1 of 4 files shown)


pmac:
@jgmize r?

bedrock/mozorg/models.py
```diff
@@ -129,3 +131,42 @@ def blog_title(self):
     @property
     def blog_link(self):
         return settings.BLOG_FEEDS[self.blog_slug]['url']
+
+    @classmethod
+    def parse_feed(cls, feed_id, feed_options):
```
jgmize: NIT: this `classmethod` doesn't actually use `cls` and could just be a regular function instead.

pmac: They all could. I just couldn't find a better place for them, and adding them to the model class made imports cleaner. I could move them to a utils.py, but I think that the models file imports from utils, and so I can't import the model there w/o some inline thing... I'm conflicted on the best location, but this seems okay. I can move it again tho.

jgmize: Fair enough. I don't want to nitpick this any further; this is still an improvement over the current code and I'd rather focus our efforts on other things. As always, NITs can be safely ignored and aren't blockers for merge.

PR #4391, opened 2016-10-05
Force queryset eval in view to catch errors

pmac:


pmac:
@jgmize r?

jgmize:
Thanks @pmac

PR #4393, opened 2016-10-05
Add docs for feature flipping with deis config

jgmize:
@pmac r?

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -145,7 +145,7 @@ of the service. We maintain a `page on the Mozilla wiki detailing our use of Opt
 To work with/test these Optimizely switches locally, you must add the switches to your local environment. For example::
 
     # to switch on firefox-new-optimizely you'd add the following to your ``.env`` file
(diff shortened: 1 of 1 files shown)


pmac:
Other than my one suggestion this looks good. Update if you think it's necessary, then merge at will. r+

docs/install.rst
```diff
@@ -157,6 +157,16 @@ You then must set an Optimizely project code in ``.env``::
     You are not required to set up Optimizely as detailed above. If not configured,
     bedrock will treat the switches as set to ``off``.
 
+To do the equivalent in one of the bedrock apps deployed with `Deis <http://deis.io/>`_, use `deis config <http://docs.deis.io/en/latest/using_deis/config-application/`_. To continue the example above with a Deis app named ``bedrock-demo-switch``::
+
+    deis config:set SWITCH_FIREFOX_NEW_OPTIMIZELY=True -a bedrock-demo-switch
```
jgmize: I was just following the example immediately above-- I like `on` better though, so I'll go ahead and update both.

PR #4394, opened 2016-10-06
Bug 1305661 functional tests for home / technology page

alexgibson:
Updated functional tests for the new home page. This also required creating a page object for the Technology page in order to test the main site navigation, so I added a test for the technology page blog feed as a separate commit.

### DIFF
--- a/tests/functional/newsletter/test_newsletter_embed.py
+++ b/tests/functional/newsletter/test_newsletter_embed.py
@@ -27,8 +27,7 @@ from pages.smarton.base import SmartOnBasePage
 
 @pytest.mark.nondestructive
 @pytest.mark.parametrize(('page_class', 'url_kwargs'), [
-    pytest.mark.skipif(pytest.mark.smoke((HomePage, None)),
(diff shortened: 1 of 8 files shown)


jpetto:
Code looks good and tests are passing. 👍

Just to confirm, if the blog feed job were to fail for some reason, thereby sending no articles to the template (and causing the 2 technology page tests to fail), we wouldn't be blocked on a push, correct?

alexgibson:
Yes correct, any test failures will block a push.

alexgibson:
I should add, the test run has to fail 3 times consecutively for the push not to go through.

jpetto:
Hm, do we want to be blocked on a push based on external services? I know it's not very likely for all the blog feeds to go down/be inaccessible, but it's not something entirely under our control.

alexgibson:
We already have several test scores of this nature for various pages. I think we cache things pretty heavily, so if something fails it's more likely a failure on our side than a reliance on external data.

jpetto:
Good enough for me. 😁

PR #4397, opened 2016-10-07
[fix bug 1305682] Remove old /firefox/new/ template

alexgibson:
## Description
- Removes old `/firefox/new/` template and tidies up some of the file locations now we no longer need to maintain two separate templates.
- Also moves a bunch of assets other pages we're referencing into the `/firefox/template/` shared directory.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1305682

## Testing
- As this removes / renames a fair few files it might be worth testing this with a fresh `/static` directory and make sure both scene1 and scene2 display correctly still. Some of the files / includes have also been used on other page so, check those other pages too.

Demo server: https://bedrock-demo-agibson.us-west.moz.works/en-US/

## Checklist
- [x] Requires l10n changes.


bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -2,147 +2,121 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
-{% add_lang_files "firefox/new" %}
+{% add_lang_files "firefox/new/horizon" %}
```
flodolo: Yes, at this point it's better to keep the existing lang file name even if the template changes.

alexgibson:
@craigcook I know you're busy, but any chance you might have time to look at this one soon? I'd like to get it merged so I can start preparing for [bug 1308973](https://bugzilla.mozilla.org/show_bug.cgi?id=1308973)

craigcook:
Looks good to me. I've tested /firefox/new (both scenes), the home page, and the nightly whatsnew page that all use horizon styles/assets and everything looks as it should, and as you noted we can update the nightly whatsnew page so it's more independent. Are there any other pages we need to double-check for dependencies? If not this is an r+.

alexgibson:
> Are there any other pages we need to double-check for dependencies? If not this is an r+.

I've done a fairly thorough search and this seems to be it, unless you can find any more I may have missed.

craigcook:
The only lingering references to 'firefox/new/horizon' I can find are in two CSS comments. I think we're good 🍒

alexgibson:
Thanks for the review, @craigcook!

PR #4402, opened 2016-10-10
[fix bug 1305954] Create functional tests for Leadership page

alexgibson:
## Description
- Adds functional tests for opening / closing / cycling modal on `/about/leadership` page.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1305954

## Testing
- I had to make some additions to the HTML template to facilitate easier testing, so this change needs to be tested locally.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/leadership.html
+++ b/bedrock/mozorg/templates/mozorg/about/leadership.html
@@ -66,8 +66,8 @@
 
     <h3 class="group-title" id="mgmt-corp"><a href="{{ url('foundation.moco') }}">{{ _('Mozilla Corporation') }}</a></h3>
 
-    <div class="gallery">
-      <article id="mitchell-baker" class="vcard has-bio" itemscope itemtype="http://schema.org/Person">
+    <div class="gallery mgmt-corp">
+      <article id="mitchell-baker" data-id="mitchell-baker" class="vcard has-bio" itemscope itemtype="http://schema.org/Person">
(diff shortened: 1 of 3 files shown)


bedrock/mozorg/templates/mozorg/about/leadership.html
```diff
@@ -66,8 +66,8 @@ <h2 class="section-title">{{ _('Management Teams') }}</h2>
 
     <h3 class="group-title" id="mgmt-corp"><a href="{{ url('foundation.moco') }}">{{ _('Mozilla Corporation') }}</a></h3>
 
-    <div class="gallery">
-      <article id="mitchell-baker" class="vcard has-bio" itemscope itemtype="http://schema.org/Person">
+    <div class="gallery mgmt-corp">
+      <article id="mitchell-baker" data-id="mitchell-baker" class="vcard has-bio" itemscope itemtype="http://schema.org/Person">
```
alexgibson: Note for reviewer: we can't rely on using the ID for testing what's displayed in the modal as the [page JS removes it](https://github.com/mozilla/bedrock/blob/master/media/js/mozorg/about-leadership.js#L41) to prevent duplicate ID's on the page, so instead I'm adding a data attribute to use for reference.

jpetto:
Tests look solid and pass. Using a new `data-id` attribute for targeting feels like a good solution. Yay more coverage! r+

PR #4404, opened 2016-10-10
Validate squashed tag exists before using

jgmize:
@pmac r?


pmac:
Syntax looks fine. I'm going to take your word on it being the right thing to do.

r+

PR #4405, opened 2016-10-11
[fix bug 1309149] oldIE.scss is missing download button component

alexgibson:
## Description
- This should fix the test failures that blocked yesterday's push.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1309149


alexgibson:
@craigcook r?

craigcook:
D'oh, an oversight when I split the download button rules into their own component. Good catch. r+

PR #4406, opened 2016-10-11
[bug 1307567] Remove inline styles from legacy pages

alexgibson:
## Description
- Removes a bunch of inline styles from legacy pages.
- The only thing that should be left after this PR in terms of inline styles are those contained in SVG files.
- There are a couple of pages that we're loading web fonts from `fonts.googleapis.com`. As it stands I haven't found replacements for these fonts, just removed references to them (they're being blocked by CSP already). I'm open to suggestions for what we should do here, should we house the fonts on mozorg, find suitable system fallback fonts, or do nothing?

Update: It looks like both Crimson Text and Lora web font's both have SIL OFL licenses, so we can't redistribute them. I don't think we should add `fonts.googleapis.com` to the CSP whitelist.


bedrock/firefox/templates/firefox/geolocation.html
```diff
@@ -21,12 +21,21 @@
 <div id="main-feature">
   <h1>{{ _('Location-Aware Browsing') }}</h1>
   <p>
-  {% trans url='#geo-demo', id='try-geolocation', style='display:none' %}
-    Firefox can tell websites where you’re located so you can find info
-    that’s more relevant and more useful. It’s about making the Web smarter
-    &ndash; and is done in a way that totally respects your privacy.
-    <a href="{{ url }}" id="{{ id }}" style="{{ style }}">Give it a try!</a>
-  {% endtrans %}
+    {% if l10n_has_tag('remove-inline-styles') %}
```
alexgibson: @flodolo - this is a very old page to which I imagine traffic is pretty low. How bad would it be to break a string here if we we're to remove the l10n tag?

flodolo: The only change is the removal of `style="{{ style }}"`, right? I think I can script that, without using a new l10n tag.

alexgibson: Correct, thanks @flodolo 👍

bedrock/newsletter/templates/newsletter/hacks.mozilla.org.html
```diff
@@ -6,80 +6,12 @@
 <html>
```
craigcook: I don't think this page is used at all now? At one time it was shown in an iframe on the Hacks blog but they don't have it in the new design (and I'm not sure this newsletter is still alive at all). I think we can probably just delete this (including the related JS/CSS bits).

bedrock/privacy/templates/privacy/archive/firefox-2008-06.html
```diff
@@ -37,8 +37,8 @@ <h1 itemprop="name">Mozilla Firefox Privacy Policy</h1>
...
+              <dd><span class="underline">Firefox 1.0 – 2.x</span>. For these earlier versions of Firefox, “Talkback” is Firefox’s crash reporting feature. Talkback also collects Personal Information (including your name, email address) and Potentially Personal Information (including your IP address, your computer’s name, and the processes you were running at the time of the crash). You can selectively disable the sending of this information. Additionally, you have the option to include the URL of the site you were visiting when Firefox crashed, a comment, and your email address in the report. Mozilla only makes Non-Personal Information and Potentially Personal Information in the public reports available online at <a href="http://www.talkback-public.mozilla.org/">www.talkback-public.mozilla.org/</a>.</dd>
```
craigcook: Ouch 😬 I wonder if the underline is really essential? If the point is just to set this off visually but it doesn't specifically have to be underlined, then `<b>` or `<i>` would suffice here.

bedrock/security/templates/security/older-vulnerabilities.html
```diff
@@ -715,7 +715,7 @@ <h3 id="nov-2003">November 2003 Update</h3>
       <td valign="top">60</td>
       <td valign="top">Networking<br>
       </td>
-      <td style="vertical-align: top;">1.5 1.4.2</td>
+      <td valign="top">1.5 1.4.2</td>
```
craigcook: The alignment on these table cells is inconsistent throughout the page. I'd vote for removing the `valign`s everywhere rather than replacing these random inline styles. Maybe just set `td { vertical-align: top; }` for the whole table?

media/css/mozorg/mpl-2-0.scss
```diff
@@ -0,0 +1,63 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../pebbles/includes/lib';
+
+body {
```
craigcook: I know this was all just copied over but there's some silly/redundant styling here. Is it worth tidying it up or should we just leave it alone? I might even be inclined to remove the CSS entirely and just let browser defaults handle it.

bedrock/privacy/templates/privacy/archive/firefox-2008-06.html
craigcook: Also, the `<u>` element is back in HTML5 (after being removed in XHTML) but now it's meant for annotations and mispellings, not for emphasis. So visually a `<u>` would have the same effect but it's probably not semantically correct.

bedrock/firefox/templates/firefox/geolocation.html
```diff
@@ -21,11 +21,11 @@
...
-  {% trans url='#geo-demo', id='try-geolocation', style='display:none' %}
+  {% trans url='#geo-demo', id='try-geolocation' %}
     Firefox can tell websites where you’re located so you can find info
     that’s more relevant and more useful. It’s about making the Web smarter
     &ndash; and is done in a way that totally respects your privacy.
-    <a href="{{ url }}" id="{{ id }}" style="{{ style }}">Give it a try!</a>
+    <a href="{{ url }}" id="{{ id }}">Give it a try!</a>
   {% endtrans %}
   </p>
 </div>
```
craigcook: Just noticed the demo popup at the bottom of this page has `style="display:none"`, we can replace that with `class="hidden"`

craigcook:
Looks good, just a few minor nitpicks. It's tempting to overhaul some of that ancient code but perhaps that can of worms should remain closed for now. The only thing that really needs fixing is that last inline style on the geolocation page. r+wc 🍗

bedrock/newsletter/templates/newsletter/hacks.mozilla.org.html
alexgibson: Ah I had no idea this wasn't in use any longer, will remove it 👍

alexgibson:
@craigcook I started down the route of trying to fix up old CSS and improve the markup, but quickly came to realise it would be considerably more work than was highlighted in this diff. I think just concentrating on fixing the bug at hand is probably the best course of action for now. As these pages are very old I don't think it's worth spending much time trying to fix things, unless there is something obviously broken. We could always file some good first bugs for contributors to take on (kind of a wontfix, but patches welcome kind of thing).

alexgibson:
@craigcook updated

PR #4410, opened 2016-10-12
Fix Bug 1309692 - Unique page title is missing on the refreshed Foundation pages

kyoshino:
Simply use the page title for `<h1>` on each Foundation page.

### DIFF
--- a/bedrock/foundation/templates/foundation/about.html
+++ b/bedrock/foundation/templates/foundation/about.html
@@ -8,10 +8,6 @@
 
 {% set body_id = "about" %}
 
-{% block article_title %}
-  {{ _('About the Mozilla Foundation') }}
-{% endblock %}
-
 {% block article %}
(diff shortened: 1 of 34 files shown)


alanmoo:
I think we might want to take a different approach to this. See https://bugzilla.mozilla.org/show_bug.cgi?id=1309692

craigcook:
Putting a DNM on this for the moment while we sort things in the bug. I agree each page should have some kind of sensible title, probably in an h2, and they can all share the common header.

craigcook:
Seems like consensus in the bug is to add `h2`s for each page. 

@kyoshino Do you want to update this PR, or if you're otherwise occupied I can close this and do a new PR to add those headings.

alexgibson:
Closing this as it's stale, and it seems like we want a slightly different approach in the bug.

PR #4400, opened 2016-10-07
Add a git repo utility for working with git external data

pmac:
* Convert l10n_update to use new git utility
* Convert update_security_advisories to use new git utility

This will support easily adding a command to update product-details from git.

PR #4412, opened 2016-10-12
Fix bug 1301489: Add yml file compatibility to security advisories

pmac:


pmac:
This is ready for review, but I'd like to wait to merge it until #4400 is merged so I can rebase this onto it.

@glogiotatidis r?

bedrock/security/tests/test_utils.py
```diff
@@ -47,8 +55,80 @@ def test_parse_front_matter_only():
...
+                 'desc': 'A different site that is totally not bugzilla'},
+            ]
+        }),
+    ])
+
+
+YML_ADVISORY = StringIO(dedent("""\
+    announced: September 13, 2016
+    fixed_in:
+    - Firefox 49
```
glogiotatidis: nit: indent in.

bedrock/security/templates/security/advisory.html
```diff
@@ -33,8 +33,10 @@ <h1 itemprop="name" class="title-shadow-box">{{ _('Mozilla Foundation Security A
           <dt>Risk</dt>
           <dd>{{ advisory.extra_data.risk }}</dd>
         {% endif %}
+        {% if advisory.impact %}
         <dt>Impact</dt>
```
glogiotatidis: nit: indent in

bedrock/security/utils.py
```diff
@@ -63,6 +64,47 @@ def parse_md_file(file_name):
...
+        advisory['id'] = cve
+        advisory['impact_class'] = advisory['impact'].lower().split(None, 1)[0]
+        for bug in advisory['bugs']:
+            bug['url'] = parse_bug_url(bug['url'])
+        html.append(render_to_string('security/partials/cve.html', advisory))
+
+    return '\n\n'.join(html)
+
+
+def parse_bug_url(url):
```
glogiotatidis: ah I see how that's named bug['url'] elsewhere. Feel free to ignore

bedrock/security/models.py
```diff
@@ -67,7 +67,7 @@ def save(self, force_insert=False, force_update=False,
 class SecurityAdvisory(models.Model):
     id = models.CharField(max_length=8, primary_key=True, db_index=True)
     title = models.CharField(max_length=200)
-    impact = models.CharField(max_length=100)
+    impact = models.CharField(max_length=100, null=True)
```
glogiotatidis: Why not `default=''` instead of `null=True`. Making `CharFields` nullable creates a situation where this field will eval to False both when it's `null` and when it's empty string `''` which may result in unexpected results?

pmac: Also, I'm curious about the difference between `default=''` and `blank=True`. I feel like the latter is the official way to go, but I'm not sure whether there's a significant difference.

pmac: Seems like there's not a big difference, but I think blank=true makes the most sense. I had to adjust some other things (avoid explicitly setting the field value to `None` e.g.), but I think this is the Right Way™️. Good catch.

glogiotatidis: And Django defaults to `""` for CharFields according to https://github.com/django/django/blob/master/django/db/models/fields/__init__.py#L774

bedrock/security/management/commands/update_security_advisories.py
```diff
@@ -115,8 +121,8 @@ def add_or_update_advisory(data, html):
     kwargs = {
         'id': mfsa_id,
         'title': data.pop('title'),
-        'impact': data.pop('impact'),
-        'reporter': data.pop('reporter', None),
+        'impact': data.pop('impact', None) or '',
```
pmac: @glogiotatidis this basically means that default is `''`, so I think it doesn't matter much in practice.

bedrock/security/models.py
pmac: So I think we're good then right? It all seems to be working well in this iteration.

pmac:
Alright. I'm satisfied, and this will let me set a different branch for www-dev, which the security team wants so they have a place to test formatting. Thanks @glogiotatidis !

PR #4413, opened 2016-10-13
[fix bug 1304538] Add sha-1 download buttons for IE 6-8

alexgibson:
## Description
- Still WIP but opening this early to make sure I'm on the right path in terms of implementation.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1304538

## Testing
Demo:
- https://www-demo1.allizom.org/en-US/firefox/new/
- https://www-demo1.allizom.org/en-US/firefox/all/
- https://www-demo1.allizom.org/en-US/firefox/channel/desktop/

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -4,6 +4,7 @@ from operator import itemgetter
 from urllib import urlencode
 
 from django.conf import settings
+from bedrock.base.waffle import switch
 from product_details import ProductDetails
 from lib.l10n_utils.dotlang import _lazy as _
 
@@ -11,6 +12,11 @@ from lib.l10n_utils.dotlang import _lazy as _
 # TODO: port this to django-mozilla-product-details
 class _ProductDetails(ProductDetails):
     bouncer_url = 'https://download.mozilla.org/'
+
+    # Note allizom.org is the production endpoint for SHA-1.
+    # It uses this because that's the only SHA-1 certificate
+    # we have that's usable. (SHA-1 certs can no longer be issued).
+    sha1_bouncer_url = 'https://download-sha1.allizom.org/'
 
     def _matches_query(self, info, query):
         words = re.split(r',|,?\s+', query.strip().lower())
@@ -25,6 +31,7 @@ class FirefoxDesktop(_ProductDetails):
 
     # Human-readable platform names
     platform_labels = OrderedDict([
+        ('winsha1', 'Windows (XP/Vista)'),
         ('win', 'Windows'),
         ('win64', 'Windows 64-bit'),
         ('osx', 'OS X'),
@@ -62,6 +69,9 @@ class FirefoxDesktop(_ProductDetails):
 
     def __init__(self, **kwargs):
         super(FirefoxDesktop, self).__init__(**kwargs)
+
+    def get_bouncer_url(self, platform):
+        return self.sha1_bouncer_url if not switch('disable-sha1-downloads') and platform == 'winsha1' else self.bouncer_url
 
     def platforms(self, channel='release'):
         platforms = self.platform_labels.copy()
@@ -121,9 +131,10 @@ class FirefoxDesktop(_ProductDetails):
         for builds in all_builds:
             if locale in builds and version in builds[locale]:
                 _builds = builds[locale][version]
-                # Append 64-bit builds
+                # Append 64-bit builds and Sha-1
                 if 'Windows' in _builds:
                     _builds['Windows 64-bit'] = _builds['Windows']
+                    _builds['Windows (XP/Vista)'] = _builds['Windows']
                 if 'Linux' in _builds:
                     _builds['Linux 64-bit'] = _builds['Linux']
                 return version, _builds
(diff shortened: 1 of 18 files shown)


alexgibson:
@pmac r?

This is up on the demo server linked above and seems to be working. While we continue to test and iron out who should get served what, I think this is pretty much ready for code review?

alexgibson:
Note I need to update some special casing logic in this branch to handle the experiment that was just merged in #4355.

media/css/firefox/new/all-test-variations.less
```diff
@@ -90,6 +90,21 @@
...
             padding-left: 24px;
         }
+
+        &.winsha1 {
+            display: none;
+        }
+    }
+}
+
+.windows.sha1 {
```
jpetto: This should be `.windows.sha-1`, right?

alexgibson:
I've thought of one more thing we might need to add to this PR before it merges. Will update.

alexgibson:
> I've thought of one more thing we might need to add to this PR before it merges. Will update.

As discussed on IRC, the edge-case I though of was how to handle derivative products like SeaMonkey or PaleMoon. This is really getting into the weeds though, so until we know for certain how to handle these (if even needed?), perhaps we should leave things as-is and update later.

bedrock/firefox/tests/test_helpers.py
```diff
@@ -52,6 +52,7 @@ def check_dumb_button(self, doc):
         eq_(pq(links[5]).attr('href'), settings.APPLE_APPSTORE_FIREFOX_LINK
                                             .replace('/{country}/', '/'))
 
+    @patch('bedrock.base.waffle.switch', Mock(return_value=False))
```
alexgibson: This also feels a little bit weird not having it enabled by default in dev mode. We could alternatively enable it explicitly on stage/prod. Thoughts?

pmac: Try mocking `switch` where it's imported instead of where it's defined:

```python
@patch('bedrock.firefox.firefox_details.switch', Mock(return_value=False))
def test_button_force_direct(self):
    ...
```

pmac: Good point about disabled in dev... I think it's fine if it's disabled in most dev envs actually. We could manually enable it for `www-dev.a.o` or any demos, or do like you suggested and default it to off. I still think it should be mostly on in prod mode... Still another option is to avoid the whole `switch` system here and just use a setting we create specifically for this and set it to default to having the sha1 URL on by default. That might be the best thing here, but I'm happy with what we have now really.

alexgibson: > Try mocking switch where it's imported instead of where it's defined:

> @patch('bedrock.firefox.firefox_details.switch', Mock(return_value=False))
def test_button_force_direct(self):
    ...

Hmm, so even this seems to fail still locally :(

bedrock/firefox/firefox_details.py
```diff
@@ -12,6 +13,11 @@
 class _ProductDetails(ProductDetails):
     bouncer_url = 'https://download.mozilla.org/'
 
+    # Note allizom.org is the production endpoint for SHA-1.
+    # It uses this because that's the only SHA-1 certificate
+    # we have that's usable. (SHA-1 certs can no longer be issued).
+    sha1_bouncer_url = 'https://download.mozilla.org/' if switch('disable-sha1-downloads') else 'https://download-sha1.allizom.org/'
```
pmac: That's the thing though, you'd have to actively set the switch to have the tests fail. The default switch value should pass (like it's doing in circle). I agree with you that if someone did set the variable it would potentially be confusing as to why the test was failing though. Fixing it right is preferable.

alexgibson: No we're doing the opposite here? Enabling the switch disables the sha-1 URL, hence why tests fail for me locally (since all switches default to on for dev).

pmac: AH! DEV! Of course you're right. I'd forgotten. Please ignore me here and do the `get_bouncer_url()` thing or the like.

alexgibson:
@pmac - that did it 👍 Tests are passing for me locally now. Appreciate you helping me through that one, thanks.

alexgibson:
I've deployed this to demo1 and set `SWITCH_DISABLE_SHA1_DOWNLOADS=off`: https://www-demo1.allizom.org/en-US/

alexgibson:
@pmac - it just occurred to me we need to do this for Thunderbird links as well. I've added another commit to replicate what we've done for Firefox. Seems to be working fine, but would appreciate another look.

pmac:
@alexgibson I would be pretty shocked if people on XP were downloading Thunderbird in any significant numbers, but I do like being thorough. I'm looking at the changes now.

alexgibson:
> @alexgibson I would be pretty shocked if people on XP were downloading Thunderbird in any significant numbers, but I do like being thorough. I'm looking at the changes now.

Without doing this download buttons would not be visible at all for Thunderbird on XP/Vista (since regular bounder links get hidden), so I think this is required really.

pmac:
lgtm. r+

alexgibson:
@pmac - looks like we have the [green light from QA](https://bugzilla.mozilla.org/show_bug.cgi?id=1304538#c26). The only thing Stephen hasn't been able to test is IE6 on XP SP2 due to [bug 1284925](https://bugzilla.mozilla.org/show_bug.cgi?id=1284925). In order to test this platform could we do a manual push to stage before going to prod perhaps?

pmac:
I think we're good to merge then. We have the switch. That should allow us
to test without actually affecting downloads.

alexgibson:
> I think we're good to merge then. We have the switch. That should allow us to test without actually affecting downloads.

Ok, we can do it that way if it's easier. I'll merge a little later and then turn the switch on in prod to disable sha-1. We can then tag a commit for a prod push and once we successfully test in stage we can turn the switch back off again.

We should also double-check with @jpetto to make sure merging/pushing this isn't going to adversely effect the test data for [bug 1288517](https://bugzilla.mozilla.org/show_bug.cgi?id=1288517).

PR #4414, opened 2016-10-13
Bug 1309050: Update Twitter API to use one app credentials

pmac:
No need for support for multiple accounts and keys.

### DIFF
--- a/bedrock/mozorg/cron.py
+++ b/bedrock/mozorg/cron.py
@@ -12,6 +12,10 @@ from bedrock.mozorg.util import get_tweets
 
 @cronjobs.register
 def update_tweets():
+    if not settings.TWITTER_APP_KEYS['consumer_key']:
+        # twitter auth not configured
+        return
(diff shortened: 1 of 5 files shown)


pmac:
@glogiotatidis r?

pmac:
I've already created a new twitter account (so that if the credentials are ever compromised it won't matter since no one will be following the account) and added the new credential keys to the dev, stage, and prod apps in deis.

glogiotatidis:
Please also update example `local.py` https://github.com/mozilla/bedrock/blob/2fb16c05fbb847b57e0fbbea8a5b51d51d554e43/bedrock/settings/local.py-dist#L15

glogiotatidis:
💯  nice work @pmac! r+wc

pmac:
Ah yes! Good catch. We should actually just remove `local.py-dist` I think. The docs now recommend using a `.env` file.

PR #4416, opened 2016-10-17
Fix some issues with YML formatted security advisories

pmac:
* [Add optional description before all CVEs](https://bugzilla.mozilla.org/show_bug.cgi?id=1310736)
* [Ensure CVEs are ordered correctly](https://bugzilla.mozilla.org/show_bug.cgi?id=1310734)

### DIFF
--- a/bedrock/security/utils.py
+++ b/bedrock/security/utils.py
@@ -72,12 +72,15 @@ def parse_yml_file(file_name):
         if mfsa_id:
             data['mfsa_id'] = mfsa_id
 
-    return data, generate_yml_advisories_html(data['advisories'])
+    return data, generate_yml_advisories_html(data)
 
 
-def generate_yml_advisories_html(advisories):
+def generate_yml_advisories_html(data):
     html = []
-    for cve, advisory in advisories.items():
+    if 'description' in data:
+        html.append(markdown(data['description']))
+
+    for cve, advisory in data['advisories'].iteritems():
         advisory['id'] = cve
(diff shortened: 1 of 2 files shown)


pmac:
@albill This should cover the current issues. Please let me know a bug number I can associate with these changes.

pmac:
@glogiotatidis r?

bedrock/security/utils.py
```diff
@@ -72,12 +72,15 @@ def parse_yml_file(file_name):
...
 
 
-def generate_yml_advisories_html(advisories):
+def generate_yml_advisories_html(data):
     html = []
-    for cve, advisory in advisories.items():
+    if 'description' in data:
+        html.append(markdown(data['description']))
+
+    for cve, advisory in data['advisories'].iteritems():
```
pmac: Yeah. I'm always torn now with these. Should I get the efficiency now with Python 2, or make the transition to 3 easier for later. I usually opt for the former because the 2to3 tool is pretty good at that kinda thing, but it is a concern. For bedrock I think python3 is still a long way off, but hopefully I'm wrong about that.

pmac:
Thanks! I'm just waiting on bugs with which to associate this work and I'll merge after I've updated the commit messages.

albill:
https://bugzilla.mozilla.org/show_bug.cgi?id=1310734

and
https://bugzilla.mozilla.org/show_bug.cgi?id=1310736

are the two bugs for the issues addressed.

pmac:
Thanks Al! I've updated the commit messages and I believe we're ready for merge. I'll do that when CircleCI says it's ready.

PR #4418, opened 2016-10-17
[fix bug 1295181] Update audience size/segmentation for fx /new all test

jpetto:
## Description

Shrinking audience significantly to reduce risk. According to pg, should give us enough data in 2 days at this new rate.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1295181#c29

## Testing

Make sure double control variation works as expected. (Easiest way is probably to boost the `v=3` variation percentage.)

## Checklist


alexgibson:
r+

PR #4419, opened 2016-10-18
[fix bug 1308973] Make it easier for pre-release users to download Firefox from /firefox/new/

alexgibson:
## Description
- Firefox pre-release users who visit `/firefox/new/` will now see a download button so they can more easily switch to release channel.
- Pre-release users also see a "You’re using a pre-release version of Firefox" message.
- Minor refactor to some of `scene1.js` to make things a bit cleaner.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1308973

## Testing
- Test all page states, ensure no regressions.

## Checklist
- [x] Requires l10n changes.

…


alexgibson:
@peiying2 - there are two new strings in this PR, however I still need to confirm that they are final.

bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -49,30 +49,66 @@
...
                 <li><a class="more" href="{{ url('firefox.android.index') }}">{{_('Learn more about Firefox for Mobile')}}</a></li>
               </ul>
             </div>
+            <div class="desktop-pre-release-links-wrapper latest-links-wrapper">
+              <ul>
+                <li><a class="more" href="https://support.mozilla.org/products/firefox/install-and-update-firefox">
+                  {% if l10n_has_tag('firefox_new_pre_release') %}
+                    {{_('Update Firefox to the latest version')}}
+                  {% else %}
+                    {{_('Update your Firefox')}}
```
alexgibson: @peiying2 - I'm assuming this is ok to use as a fallback string? I got it from [download_button.lang](https://transvision.mozfr.org/?sourcelocale=en-US&locale=en-US&repo=mozilla_org&search_type=entities&recherche=mozilla_org/download_button.lang:5f8fd432&perfect_match=perfect_match)

alexgibson: It's ok, looks like we don't need this string after all.

alexgibson:
@peiying2 ok there should be only one new string here and this can be considered final. I've used an l10n tag.

Removing the do-not-merge label.

jpetto:
Code looks solid, `scene1.js` is smaller than ever, and pre-release users can download a bit more easily. 👏👏

PR #4420, opened 2016-10-18
Bug 1311037 - The language switcher doesn't take the anchor into account

pascalchevrel:
## Description
Bug 1311037 - The language switcher doesn't take the anchor into account
## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1311037

## Testing
Manual testing

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/js/base/global.js
+++ b/media/js/base/global.js
@@ -76,6 +76,7 @@ function initLangSwitcher() {
             'languageSelected': $language.val(),
             'previousLanguage': previousLanguage
         });
+        $('#lang_form').attr('action', window.location.hash || '#');
         $('#lang_form').submit();
     });
 }


jpetto:
Aside from a small nit on avoiding jQuery, looks good to me. r+wc

media/js/base/global.js
```diff
@@ -76,6 +76,7 @@ function initLangSwitcher() {
             'languageSelected': $language.val(),
             'previousLanguage': previousLanguage
         });
+        $('#lang_form').attr('action', $(location).attr('hash') || '#');
```
pascalchevrel: No other reason than I saw the site uses jquery so I used it :) I will update the patch to use javascript directly.

jpetto:
Noting that the anchor is not retained in IE <= 9, but I don't think it's worth pursuing that functionality.

Thanks @pascalchevrel!

PR #4421, opened 2016-10-18
Move switch function for use in python code

pmac:
* Avoid import issues by providing a separate module for this function
* Document using feature flipping within Python

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -137,6 +137,14 @@ environment variable ``SWITCH_THE_DUDE``. If the value of that variable is any o
 is set to one of those "true" values, then all switches will be considered "on" unless they are
 explicitly "off" in the environment.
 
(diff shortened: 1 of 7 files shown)


docs/install.rst
```diff
@@ -137,6 +137,14 @@ environment variable ``SWITCH_THE_DUDE``. If the value of that variable is any o
 is set to one of those "true" values, then all switches will be considered "on" unless they are
 explicitly "off" in the environment.
 
+You may also use these switches in Python in ``views.py`` files. For example::
+
+    from bedrock.base import waffle
```
alexgibson: As we just use `{% if switch('dude-and-walter') %}` in the templates, for consistency we could do from `from bedrock.base.waffle import switch` here in the example?

pmac:
Change sounds good to me. I've made it and the tests are running again. Thanks for the r? Alex!

PR #4422, opened 2016-10-18
[fix bug 1311052] Traffic Cop will ignore URL hash.

jpetto:
## Description

Ensure URL hash is handled appropriately in redirect URL when present in the original URL.

As the variant experience could be completely different from the control - hash element may not exist, we may want a "fresh" (no page scrolling) UX - I decided to drop the hash instead of adding it back to the variation URL.

However, I'm not completely sold on this approach. I suppose it could be a config option with the default being no hash, but that smells like over-engineering. My gut is saying leave it off until a specific case arises where we need it.

It would be relatively simple to add the hash on to the variation URL, so, thoughts?

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1311052

## Testing

Change a local test to have 100% variation coverage and open some private windows.

## Checklist

### DIFF
--- a/media/js/base/mozilla-traffic-cop.js
+++ b/media/js/base/mozilla-traffic-cop.js
@@ -146,7 +146,21 @@ Mozilla.TrafficCop.prototype.generateRedirectUrl = function(url, querystring) {
     // conjure a random number between 1 and 100 (inclusive)
     var rando = Math.floor(Math.random() * 100) + 1;
 
-    url = url || window.location;
+    // make sure to disregard hash in original URL
+    if (!url) {
+        // build URL root, e.g. https://www.mozilla.org
+        url = window.location.protocol + '//' + window.location.hostname;
+
+        // handy for local testing
+        if (window.location.port !== '') {
+            // adds port, e.g. (https://www.mozilla.org):443
+            url += ':' + window.location.port;
+        }
+
+        // add suffix, e.g. (https://www.mozilla.org)/en-US/about/
+        url += window.location.pathname;
+    }
+
     querystring = querystring || window.location.search;
 
     // check to see if user has a cookie from a previously visited variation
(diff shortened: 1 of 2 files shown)


media/js/base/mozilla-traffic-cop.js
```diff
@@ -146,7 +146,21 @@ Mozilla.TrafficCop.prototype.generateRedirectUrl = function(url, querystring) {
...
     var rando = Math.floor(Math.random() * 100) + 1;
 
-    url = url || window.location;
+    // make sure to disregard hash in original URL
+    if (!url) {
+        // build URL root, e.g. https://www.mozilla.org
+        url = window.location.protocol + '//' + window.location.hostname;
+
+        // handy for local testing
+        if (window.location.port !== '') {
```
alexgibson: How sure are we that this won't introduce some unexpected results in some browsers? (I have a funny recollection of IE doing some strange things even when no port is specified). Could a simpler option be to use a regex to strip the hash? Either way, an accompanying test would be good here 👍

media/js/base/mozilla-traffic-cop.js
```diff
@@ -146,7 +146,21 @@ Mozilla.TrafficCop.prototype.generateRedirectUrl = function(url, querystring) {
     // conjure a random number between 1 and 100 (inclusive)
     var rando = Math.floor(Math.random() * 100) + 1;
 
-    url = url || window.location;
+    // make sure to disregard hash in original URL
```
alexgibson: While I understand the premise that a hash might be unwanted in some test, I do wonder if we should err on the side of "if it's not broken, don't fix it"? 

For example, if a user is sent a URL like this: https://www.mozilla.org/en-US/firefox/channel/desktop/#beta, it should probably stick around, even if they go into some sort of experiment. If we have a case where it causes issues then maybe we could reconsider? I can't honestly decide what's best, it's a bit hard to say. But I'm happy to go with your judgement 😸

jpetto:
Refactored to retain the hash and added tests.

PR #4423, opened 2016-10-19
[bug 1311423] Re-enable flashtalking pixel on /firefox/new/

alexgibson:
This is pretty much just a reversal of https://github.com/mozilla/bedrock/commit/6f30fd090b658a34a1ef706e8a21f7819be2f8c3

@jpetto r?

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -1169,3 +1169,6 @@ if CSP_EXTRA_FRAME_SRC:
 
 # support older browsers (mainly Safari)
 CSP_FRAME_SRC = CSP_CHILD_SRC
+
+if config('SWITCH_FLASHTALKING', default=DEV, cast=bool):
+    CSP_IMG_SRC += ('servedby.flashtalking.com',)
--- a/bedrock/firefox/templates/firefox/new/scene2.html
+++ b/bedrock/firefox/templates/firefox/new/scene2.html
@@ -51,5 +51,8 @@
 {% endblock %}
 
 {% block js %}
+  {% if switch('flashtalking') %}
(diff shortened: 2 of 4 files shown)


alexgibson:
One thing I can't remember is if we used to have a delay on here before triggering the download. It looks like we haven't had one since before the pixel was removed.

bedrock/settings/base.py
```diff
@@ -1121,6 +1121,7 @@ def lazy_email_backend():
     '*.tiles.mapbox.com',
     'api.mapbox.com',
     'creativecommons.org',
+    'servedby.flashtalking.com',
```
pmac: I'd like to put this behind the switch as well. We can't import the `switch()` helper here, but we can check the environment. Put the following after the rest of the CSP configs and it should work.

```python
if config('SWITCH_FLASHTALKING', default=DEV, cast=bool):
    CSP_IMG_SRC += ('servedby.flashtalking.com',)
```

alexgibson: Ooh nice call

alexgibson:
> One thing I can't remember is if we used to have a delay on here before triggering the download. It looks like we haven't had one since before the pixel was removed.

Seeing as we've moved the pixel to a separate file (using a switch) that gets inserted prior to the main JS being requested & executed, I think it should give enough time for the pixel request to happen OK (if it was even an issue before). I don't think we should consider delaying the download unless it was really necessary.

jpetto:
I don't see any reason to delay the download, either.

r+, though will wait to merge until the morning to make sure all is ready.

jpetto:
Appears we are good to merge. 🙊

PR #4424, opened 2016-10-20
Some small improvements to MFSA YAML styling and function

pmac:
A real effort to make these things look nice is coming in a near-future sprint.

### DIFF
--- a/bedrock/security/templates/security/partials/cve.html
+++ b/bedrock/security/templates/security/partials/cve.html
@@ -1,19 +1,21 @@
-<h4 id="{{ id }}" class="level-heading cve">
-  <a href="#{{ id }}"><span class="anchor">#</span>{{ id }}: {{ title }}</a>
-</h4>
-<dl class="summary">
-  <dt>Reporter</dt>
-  <dd>{{ reporter|safe }}</dd>
-  <dt>Impact</dt>
-  <dd><span class="level {{ impact_class }}">{{ impact }}</span></dd>
-</dl>
-<h5>Description</h5>
-{{ description|markdown }}
-{% if bugs %}
-  <h5>References</h5>
-  <ul>
-    {% for bug in bugs -%}
-      <li><a href="{{ bug.url }}">{{ bug.desc|safe }}</a></li>
-    {%- endfor %}
-  </ul>
-{% endif %}
+<section class="cve">
+  <h4 id="{{ id }}" class="level-heading">
+    <a href="#{{ id }}"><span class="anchor">#</span>{{ id }}: {{ title }}</a>
+  </h4>
+  <dl class="summary">
+    <dt>Reporter</dt>
+    <dd>{{ reporter|safe }}</dd>
+    <dt>Impact</dt>
+    <dd><span class="level {{ impact_class }}">{{ impact }}</span></dd>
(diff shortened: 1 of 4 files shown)


pmac:
Unfortunately I also corrected the indentation in `security.less`. Please forgive the diff :(

@alexgibson r?

bedrock/security/templates/security/partials/cve.html
```diff
@@ -1,19 +1,21 @@
...
-{{ description|markdown }}
-{% if bugs %}
-  <h5>References</h5>
-  <ul>
-    {% for bug in bugs -%}
-      <li><a href="{{ bug.url }}">{{ bug.desc|safe }}</a></li>
-    {%- endfor %}
-  </ul>
-{% endif %}
+<div class="cve">
```
alexgibson: Nit: could make this a bit more semantic by using a `<section>` element instead of a `<div>`.

bedrock/security/templates/security/partials/cve.html
```diff
@@ -1,19 +1,21 @@
...
+  <dl class="summary">
+    <dt>Reporter</dt>
+    <dd>{{ reporter|safe }}</dd>
+    <dt>Impact</dt>
+    <dd><span class="level {{ impact_class }}">{{ impact }}</span></dd>
+  </dl>
+  <h5>Description</h5>
+  {{ description|markdown }}
+  {% if bugs %}
+    <h5>References</h5>
```
alexgibson: Nit: This could probably be an `<h6>`.

media/css/security/security.less
```diff
@@ -57,6 +57,26 @@
...
+            dt {
+                .font-size(@baseFontSize);
+                padding: 6px;
+            }
+            dd {
+                .font-size(@smallFontSize);
+                padding: 6px 6px 6px 120px;
+                .level {
+                    font-size: inherit;
+                    line-height: 12px;
```
alexgibson: `line-height` should be unit-less, a value of `1` here should work.

alexgibson:
At small viewports some of the long headings break out of the container. Adding `word-wrap: break-word;` to the headings should fix things (not ideal, but a quick fix).

wrod-wrap

bedrock/security/templates/security/partials/cve.html
pmac: It's not actually a sub heading of "Description", it's just indented because of the `if` tag. In that case `h5` is correct right?

alexgibson: Ah you're quite right, I misread 👍

pmac:
Updated things from your suggestions. The `section` element inherited some unfortunate styles, so I had to do a bit more tweaking to make those go away, but I think it looks pretty good.

I also removed the border above the section titles and increased the top margin.

media/css/security/security.less
```diff
@@ -57,6 +57,27 @@
         .font-size(32px);
         letter-spacing: normal;
     }
+    .cve {
+        margin: 2.5em 0 0 2.5em;
```
alexgibson: Should this be `1em` on the left, to keep the content in line? `2.5em` feels a bit odd, especially at smaller screen sizes.

PR #4425, opened 2016-10-20
[fix bug 1302246] Add URLs for mobile download desktop template.

jpetto:
## Description

Updates pre-existing `/firefox/mobile-download-desktop.html` template to be accessible and extended by whatsnew page for Fx 50+.

URLs added are:
- `/firefox/50.0/whatsnew/`
- `/firefox/mobile-download/desktop/`

Also does some cleanup/shuffling of old whatsnew files.
## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1302246
## Testing

For both URLs:
- **en-US**: Should get custom headline and widget
- **other basket supported locales** (listed in `mobile-download-desktop.html` template): Should get fallback headline and widget
- **non-basket supported locales**: Should get fallback headline and no widget
## Checklist


jpetto:
Waiting to merge while new basket message is being tested on the [demo]().

jpetto:
Basket message testing was a success, good to review and merge.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-50.html
```diff
@@ -0,0 +1,36 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% add_lang_files "firefox/sendto" "download" %}
+
+{% extends "firefox/mobile-download-desktop.html" %}
+
+{% block page_css %}
```
alexgibson: Do we need to add a `{% block gtm_page_id %}data-gtm-page-id="/firefox/whatsnew/"{% endblock %}` block for this page?

alexgibson:
Looks like the `/firstrun/learnmore` template needs updating to [match the template re-ordering](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/firstrun/learnmore/learnmore.html#L5) in this PR

bedrock/firefox/templates/firefox/whatsnew/whatsnew-50.html
jpetto: Ah, yes, good call. Will also add the robots `<meta>` tag.

bedrock/firefox/templates/firefox/mobile-download-desktop.html
```diff
@@ -49,21 +54,32 @@
...
             <div class="content">
+            {# Hide the widget if user is not in a supported basket locale #}
+            {% if show_send_to_device %}
               {{ send_to_device(include_title=False, message_set='fx-mobile-download-desktop') }}
+            {% endif %}
 
               <ul id="mobile-download-buttons">
                 <li class="android">
-                  {{ google_play_button(alt_href='https://app.adjust.com/2uo1qc?campaign=Heartbeat&adgroup=Google_Play_Store&creative=badge') }}
+                  {{ google_play_button(alt_href='https://app.adjust.com/2uo1qc?campaign=Heartbeat&adgroup=Google_Play_Store&creative=badge', anchor_attributes={'data-link-type': 'download', 'data-download-os': 'Android'}) }}
```
alexgibson: Hmm, you're quite right. Not quite sure why that is or if it's something we should probably fix, but this works for now.

jpetto: I *think* the idea is that if it's not the standard Google Play link, we should also re-consider the tracking attributes (which happened to be the same in this instance). I would say if we have one more situation like this (alternate URL, same tracking attrs), we re-work the helper.

alexgibson: Looks like it is mainly due to this button https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templates/mozorg/contribute/tasks/stumbler.html#L24

Agreed we should probably fix it in the macro if this becomes an issue. Making the stumbler badge not dependent on the macro would probably make the most sense as it seems to be the exception to most other occurrences.

bedrock/firefox/templates/firefox/mobile-download-desktop.html
```diff
@@ -29,13 +23,24 @@
...
 
+{#
+  Determine if user is in locale supported by specific basket messages used by
+  this template.
+
+  This template is used for multiple URLs (/firefox/mobile-download/desktop and
+  /firefox/50.0/whatsnew/), so it's easier to do this logic here rather than
+  creating some generalized view logic.
+#}
+{% set show_send_to_device = LANG.startswith('en-') or LANG in ['de','es-ES','fr','id','pl','pt-BR','ru'] %}
```
jpetto: Yeah, settings is probably a better place for this array. Will update.

alexgibson:
Code looks good. As discussed on IRC, there's one last open question as to whether or not we could just use the existing list of locles for send-to-device. The list here just seems to be a subset of locales, and newsletters are translated into generic languages for `es` and `en` etc.

https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L1007-L1009

jpetto:
[Checking in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1302246#c37) to see if we can use our pre-existing locale list instead of creating a new one.

jpetto:
A request has come in to change some of the tracking URLs (not sure which at this point), and we're still waiting on an answer to the locale list question, so Do No Merge remains in effect.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-50.html
```diff
@@ -0,0 +1,44 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% set send_to_device_message_set = 'fx-50-whatsnew' %}
+{% set app_store_url = 'https://app.adjust.com/2uo1qc?campaign=whats_new&adgroup=iOS_App_Store&creative=badge_en&fallback=https%3A%2F%2Fitunes.apple.com%2Fapp%2Fapple-store%2Fid989804926%3Fpt%3D373246%26ct%3Dadjust_tracker%26mt%3D8' %}
+{% set play_store_url = 'https://app.adjust.com/2uo1qc?campaign=whats_new&adgroup=Google_Play_Store&creative=badge_en' %}
+
+{% extends "firefox/mobile-download-desktop.html" %}
```
pmac: I believe this needs to be the first directive in the file (after the license comment).

jpetto: I wasn't sure if they needed to be present before the `extends`. I see now that they do not. 😁

jpetto:
Requirements are now complete, so this PR is again ready for review. I kept commits distinct for (hopefully) easier pick up. Will squash prior to merge.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-50.html
```diff
@@ -0,0 +1,46 @@
...
+{% block headline %}
+
+{# If user is in a locale with translated basket messages... #}
+{% if show_send_to_device %}
+  {# Custom headlines for en-* users #}
+  {% if LANG.startswith('en-') %}
+    <h4 class="up-to-date"><span>Your Firefox is up to date.</span></h4>
+    <h2>Now get on the go.</h2>
+
+    <h1>Send Firefox to your phone<br> and unleash your Internet.</h1>
```
alexgibson: Not a big deal as this is pretty much a desktop only page, but you could set the `<br>` here to `display:none;` at smaller view ports to avoid the odd looking line break in the copy. Feel free to ignore if you think this isn't important.

alexgibson:
Probably an edge case, but if I hit enter on the send-to-device form with JS disabled then there's no nice confirm message to say that the POST works (just a blank screen). This is by design IIRC (for original use cases of the form), but I wonder here if it would be worth hiding the form for users with JS disabled, and just displaying the store badges instead? This is what we do on both the `/android` and `/ios` pages.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-50.html
jpetto: It's a nice bit of cleanup that's easy to do. 👍

PR #4426, opened 2016-10-21
Don't load Pontoon on dev for older browsers

alexgibson:
While Pontoon is only loaded on dev environments, it still throws JS errors in older browsers such as IE.

pontoon-error

This can be a bit annoying when trying to debug real issues, so let's not load it for older browsers.


jpetto:
🅰➕

PR #4430, opened 2016-10-21
[fix bug 1307998] Add MOSS pages.

jpetto:
## Description

Adds MOSS pages.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1307998

## Testing

Nothing out of the ordinary. It's on [a demo](https://bedrock-demo-jpetto-moss.us-west.moz.works/en-US/moss/) for your convenience.

## Checklist
- [ ] Requires l10n changes.


jpetto:
It's possible I went a little too far in trying to keep all (okay, most) copy visible without horizontal scrolling (in a text editor). My feelings will not be hurt if it's not to the team's liking.

jpetto:
Troy found a couple copy edits, so marking as do not merge for now.

jpetto:
Fixed up copy and resolved conflicts. Should be good to go pending review.

bedrock/mozorg/templates/mozorg/moss/_alittlebitmore.html
```diff
@@ -0,0 +1,30 @@
+<section class="moss-section">
+  <h2 class="moss-section-heading">{{ _('A little bit more') }}</h2>
```
craigcook: Should this section include links? Are there wiki pages for these?

bedrock/mozorg/templates/mozorg/moss/index.html
```diff
@@ -0,0 +1,292 @@
...
+
+      <p class="moss-copy">
+      {% trans %}
+        Tor is free software and an open network that helps you defend against
+        network surveillance.
+      {% endtrans %}
+      </p>
+
+      <p class="moss-ctas">
+        <a href="https://www.torproject.org/" rel="external" class="moss-copy">https://www.torproject.org</a>
```
craigcook: Do these links really need to display the protocol? I think just "torproject.org" would be better for readability. I suppose this is a creative/copy decision but showing protocols here seems clunky and unnecessary. If the issue was already raised and this is what they want, so be it, but if it wasn't raised before I think we should ask now.

bedrock/mozorg/templates/mozorg/moss/index.html
```diff
@@ -0,0 +1,292 @@
...
+
+  <p class="moss-copy">
+  {% trans %}
+    Check out a few of the Mozilla Open Source Support past recipients.
+  {% endtrans %}
+  </p>
+
+  <ul class="moss-list moss-list-thirds">
+    <li>
+      {{ high_res_img('mozorg/moss/tor.png', {'alt': 'Tor', 'width': '300', 'height': '169'}) }}
```
craigcook: If this image with its alt text is acting as the title for the content, it should probably be wrapped in an `h3`. Then the dollar amount below is more content rather than a title so I'd probably just make it a paragraph (which could still use the `moss-subhead` class for styling).

bedrock/mozorg/templates/mozorg/moss/mission-partners.html
```diff
@@ -0,0 +1,319 @@
...
+
+  {% trans url1='https://wiki.mozilla.org/MOSS/Mentors' %}
+    If you need help completing your application, see the
+    <a href="{{ url1 }}">list of mentors</a>.
+  {% endtrans %}
+  </p>
+
+  <div class="moss-ctas">
+    <p class="moss-copy">
+      I’m ready!
```
craigcook: String should be wrapped for l10n.

bedrock/mozorg/templates/mozorg/moss/mission-partners.html
```diff
@@ -0,0 +1,319 @@
...
+
+  <dl class="moss-dl">
+    <dt class="moss-copy">
+      {{ _('What if my project doesn’t have a legal organizational home?') }}
+
+    {% trans %}
+      I see this is disfavored, and exceptional circumstances are required.
+    {% endtrans %}
+
+      {{ _('What might count as “exceptional”?') }}
```
craigcook: Yeah, I think maybe you did go a little overboard with the "no long lines" stuff 😉 I'm finding the switching between `{% trans %}` and `{{ _(''} }}` within the same paragraph less readable than a consistently formatted stack of long lines. I think it might be better to pick one method for enclosing strings and stick with it, at least within a given block of text.

bedrock/mozorg/templates/mozorg/moss/secure-open-source.html
```diff
@@ -0,0 +1,232 @@
...
+    {{ _('What your project should include') }}
+  </h2>
+
+  <p class="moss-copy">
+    {{ _('The SOS Fund has a very limited set of solid rules:') }}
+  </p>
+
+  <ul class="moss-bullet-list">
+    <li>
+      <p class="moss-copy">
```
craigcook: Maybe the nested paragraphs aren't necessary, could this just be `<li class="moss-copy">` ?

media/css/mozorg/moss/common.scss
```diff
@@ -0,0 +1,471 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../../pebbles/includes/lib';
+@import '../../pebbles/components/newsletter';
+
+
+$color-mozred-bright: saturate(lighten($color-mozred, 5%), 10%);
```
craigcook: Wonder if we should just define this in _variables.scss if we're going to use the same color in different places (home and MOSS so far).

media/css/mozorg/moss/common.scss
```diff
@@ -0,0 +1,471 @@
...
+
+.todo {
```
craigcook: Leftover from dev?

bedrock/mozorg/templates/mozorg/moss/_alittlebitmore.html
jpetto: D'oh, yep. Will add.

bedrock/mozorg/templates/mozorg/moss/index.html
jpetto: Yesss, good suggestion. 👍

bedrock/mozorg/templates/mozorg/moss/mission-partners.html
jpetto: Noted. I'll make sure all strings within a given element are uniformly wrapped for L10n (which basically means `{% trans %}` blocks for anything more than a few words).

bedrock/mozorg/templates/mozorg/moss/index.html
jpetto: This was some last-minute copy changing. It used to say "Visit {project}" with a link to the project. I'm going to say the protocol is not necessary. Will update.

bedrock/mozorg/templates/mozorg/moss/secure-open-source.html
jpetto: Good suggestion.

media/css/mozorg/moss/common.scss
jpetto: Yeah, probably not a bad idea. Will refactor.

jpetto:
> I'm not a big fan of all the moss- classes, makes it feel really one-offish and un-reusable. But then of course it is one-offish and I expect these pages will just live in a bubble until maybe we can come back to them with a redesign. So in that case I suppose a bunch of name-spaced classes may be the better way to isolate these pages, I just wish we would stop doing one-off pages like this. Some day when we have a design system...

Totally, completely, 100% agree. I think asked you in IRC a week ago about the status of the design system - I was working on these pages at that time. I really didn't want to use tag selectors, and creating yet another unique `.copy` class felt a little more wrong that using prefixes. I don't know. I'm frustrated, too. I'm happy to drop the `moss-` prefixes if you feel they're distracting/excessive. Wouldn't be much more than a find/replace.

craigcook:
New header looks much less broken! r+

PR #4431, opened 2016-10-25
[fix bug 1311957] Make /technology page v=b default and localize content

alexgibson:
## Description
- Makes `/technology` page version B default and makes ready for localization.
- Refactors page to use Pebbles/Sass instead of Sandstone/Less.
- Adds some RTL styles.
- Adds link to `/technology` page in the main navigation for mozorg pages, replacing links to `/contribute`.
## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1311957
## Testing
- Demo URL: https://bedrock-demo-agibson.us-west.moz.works/technology/
- Could do with some responsive testing given the desktop-first -> mobile-first refactor.
## Checklist
- [x] Requires l10n changes.


alexgibson:
Note: I still need to confirm that we are ok waiting on localization of the strings in base templates.

alexgibson:
Ok, we're only going to show the new navigation link for `en` locales for now, and we can then remove the conditional when localization has caught up. Marking this as ready.

bedrock/base/templates/base-pebbles.html
```diff
@@ -102,9 +102,15 @@
             <li class="first about-item">
               <a href="{{ url('mozorg.about') }}" data-link-type="nav" data-link-name="About">{{_('About')}}</a>
             </li>
+            {% if LANG.startswith('en-') %}
+            <li class="technology-item">
+              <a href="{{ url('mozorg.technology') }}" data-link-type="nav" data-link-name="Web Innovations" data-alt-text="{{ _('Innovations') }}">{{ _('Web Innovations') }}</a>
```
flodolo: Do we really need two separate strings for data-link-name and label?

alexgibson: I included this as I was worried that "Web Innovations" may well be too long to fit in the nav for some locales at smaller screen widths. If you think I'm being over cautious here I can be persuaded to drop it?

flodolo: How is the data-link-name attribute used in practice? All other menus only have one string, to avoid confusion I would personally stick to that

alexgibson: My plan was just to shelve it here for now, and then use it when the time comes if needed (when we remove the `en` conditional).

flodolo: I'll leave the decision to @peiying2

The string won't be testable anyway on the stage server, so we can also expose both of them saying that they're potential menu labels.

bedrock/base/templates/base-pebbles.html
```diff
@@ -102,9 +102,16 @@
             <li class="first about-item">
               <a href="{{ url('mozorg.about') }}" data-link-type="nav" data-link-name="About">{{_('About')}}</a>
             </li>
+            {% if LANG.startswith('en-') %}
```
jpetto: Would this be better as `l10n_has_tag` (here and below), or would the `.lang` file dance be too convoluted?

media/css/mozorg/technology.scss
```diff
@@ -0,0 +1,881 @@
...
+        }
+    }
+}
+
+#iot {
+    padding-top: 216px;
+    z-index: 0;
+
+    .inner-container {
+        padding-top: 20px * 2;
```
jpetto: Nit - just make it `40px`?

media/css/mozorg/technology.scss
```diff
@@ -0,0 +1,881 @@
...
+    }
+}
+
+#context-graph {
+    background-color: #f9ddc7;
+    padding-top: 213px;
+    z-index: 0;
+
+    .inner-container {
+        padding: 20px * 2 0;
```
jpetto: Same - no need for the maths?

media/css/mozorg/technology.scss
```diff
@@ -0,0 +1,881 @@
...
+            width: 100%;
+        }
+    }
+
+    @media #{$mq-tablet} {
+        padding-top: 0;
+
+        &:after {
+            @include background-size(400px 322px);
+            background-position: top (20px * 2) left (20px * 4);
```
jpetto: Superfluous maths?

media/css/mozorg/technology.scss
```diff
@@ -0,0 +1,881 @@
...
+        &:after {
+            @include background-size(446px 359px);
+        }
+    }
+}
+
+#rust {
+
+    .inner-container {
+        padding-top: 20px * 2;
```
jpetto: Moar maths (and below...)

bedrock/base/templates/base-pebbles.html
alexgibson: We can't use an l10n tag in the base templates, so we're doing `en` only for now until enough locales translate the strings.

media/css/mozorg/technology.scss
alexgibson: Doh yeah, these are remnants of a mass find/replace for `@baseLine` when converting from Sandstone to Pebbles. Will update.

alexgibson:
@jpetto thanks for the review, updated

jpetto:
r+ 🎃

flodolo:
String is now being localized, leaving the tag so I won't forget about removing the conditional.

alexgibson:
Thanks @flodolo - I filed a bug for removing the conditional here https://bugzilla.mozilla.org/show_bug.cgi?id=1316892

PR #4438, opened 2016-10-27
[fix bug 1313371] Update app-dev newsletter title & description.

jpetto:
## Description

Update title and description of the `app-dev` newsletter.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1313371

## Testing

As this newsletter is now English only, and since the Firefox Friends newsletter is also English only and [currently untranslated](https://www.mozilla.org/fr/newsletter/existing/80a58e10-88ee-4369-aabe-881da32a1aba/), I don't think we need to block on/code around L10n.

## Checklist

### DIFF
--- a/bedrock/newsletter/views.py
+++ b/bedrock/newsletter/views.py
@@ -71,9 +71,9 @@ NEWSLETTER_STRINGS = {
         'description': _lazy(u'A monthly newsletter on how to get involved with Mozilla on your campus. '),
         'title': _lazy(u'Firefox Student Ambassadors')},
     u'app-dev': {
-        'description': _lazy(u'News for developers about Firefox OS, Firefox Marketplace and the '
-                             u'Open Web apps ecosystem.'),
-        'title': _lazy(u'Firefox Apps & Hacks')},
+        'description': _lazy(u'A developer\u2019s guide to highlights of Web platform '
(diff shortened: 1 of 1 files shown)


flodolo:
These strings are localized 👎 

Since it's [not a shared file](https://github.com/jpetto/bedrock/blob/3926e3cf4b62dc7106920bc4268cedf37368f23c/bedrock/newsletter/views.py#L37) anymore, we could technically use tags, but I'm not sure it makes a lot of sense in this context.

jpetto:
> Since it's not a shared file anymore, we could technically use tags, but I'm not sure it makes a lot of sense in this context.

I think you're saying we're good to go forward with this change, displaying English strings while L10n catches up? Is that accurate @flodolo?

flodolo:
I'm more saying that I don't think the newsletter team wants to show the obsolete localized name/description while we localize it, but it would be worth to double check it.

jpetto:
Ah, I see. [Asked for confirmation in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1313371#c6).

jpetto:
Jessilyn confirmed that we should show the new strings to all users, even while untranslated. Ready for review and merge.

PR #4441, opened 2016-10-28
Update basket_data.json

jgmize:
Ran https://github.com/mozilla/bedrock/blob/master/bedrock/newsletter/utils.py#L83

@jpetto r?


jpetto:
👍 r+

PR #4448, opened 2016-11-02
[fix bug 1313714] Update newsletter id on /newsletter.

jpetto:
## Description

Updates the newsletter id used on [mozilla.org/newsletter/](https://www.mozilla.org/en-US/newsletter/).

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1313714

## Testing

Make sure I didn't miss updating any dependencies.

## Checklist

### DIFF
--- a/bedrock/newsletter/templates/newsletter/index.html
+++ b/bedrock/newsletter/templates/newsletter/index.html
@@ -3,7 +3,7 @@
 {% set_lang_files "mozorg/newsletters" %}
 
 {% set newsletter_title = _('Mozilla Newsletter') %}
-{% set newsletter_id = 'mozilla-and-you' %}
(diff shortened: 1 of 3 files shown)


alexgibson:
Unrelated to this PR, but I notice if you do a POST to `/newsletter (e.g. non-AJAX), then you see "Thanks!" twice:

many-thanks

Moving [this line](https://github.com/mozilla/bedrock/blob/master/bedrock/newsletter/templates/newsletter/one_newsletter_signup.html#L37) inside `{% block success_confirm %}` seems to be the fix?

jpetto:
Good catch on the headline. Opted to just remove the `success_confirm` block in `index.html` (as it really didn't need to override anything).

alexgibson:
r+ 🛂

PR #4452, opened 2016-11-03
Fix bug 1315082: Reset git repos on failed db updates

pmac:
Reset git repos from git-based data import commands (e.g. update_security_advisories) when data import fails but the git update succeeds. In this way the next run of the command will try again and see all of the changed files.

### DIFF
--- a/bedrock/mozorg/management/commands/update_product_details_files.py
+++ b/bedrock/mozorg/management/commands/update_product_details_files.py
@@ -28,10 +28,6 @@ class Command(BaseCommand):
         super(Command, self).__init__(stdout, stderr, no_color)
 
     def add_arguments(self, parser):
-        parser.add_argument('-f', '--force', action='store_true', dest='force', default=False,
-                            help=('Download product details even if they have '
-                                  'not been updated since the last fetch.'))
-
         parser.add_argument('-q', '--quiet', action='store_true', dest='quiet', default=False,
                             help='If no error occurs, swallow all output.'),
         parser.add_argument('--database', default='default',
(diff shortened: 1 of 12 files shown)


pmac:
@jgmize this is a start on improving the git-based commands. I need to test, but figured I'd get this in so we could discuss direction.

jgmize:
Seems like a plausible approach to me. Looking forward to seeing your testing approach for the failure scenario this is supposed to address.

pmac:
You and me both

bedrock/mozorg/management/commands/update_product_details_files.py
```diff
@@ -53,13 +54,20 @@ def handle(self, *args, **options):
...
+        if options['force'] or modified is None:
             files_to_load = self.file_storage.all_json_files()
         else:
             files_to_load = self.filter_filenames(modified)
 
         if files_to_load:
-            self.load_changes(options, files_to_load)
+            try:
+                self.load_changes(options, files_to_load)
+            except Exception:
```
metadave: is it worth logging the exception here?

pmac: Could do. I'll add that. Thanks!

pmac:
@jgmize I've updated this with the new DB-based directly. It may well be ready to go. This also decently simplifies the product-details and security-advisory commands.

bedrock/utils/migrations/0001_initial.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name='GitRepoState',
```
pmac: I have added a call to `manage.py migrate` to the clock process startup.

pmac:
@jgmize ready for another (and possibly final) look.

pmac:
@jgmize I changed the call in supervisor to `run-clock.sh`. Should be good now I think.

PR #4453, opened 2016-11-05
Update license to MPL 2

xfq:
## Description

Update the license info in `README.md` from MPL/GPL/LGPL tri-license to MPL 2.0.


pmac:
Thanks very much!

PR #4456, opened 2016-11-08
[fix bug 1279291] Construct and send attribution code from download buttons

alexgibson:
## Description
- Constructs attribution data based on utm parameters and referrer information for relay to the Firefox stub installer. 
- Data is first signed and encoded via an XHR request to the `stub_attribution_code` service, before being appended to Bouncer download URLs as query parameters. 
- Data returned from the service is also stored in a session cookie, to save multiple requests when navigating pages.
- Stub attribution requests are made only if the following criteria are satisfied:

  1. User is on Windows (since only windows users get the stub installer).
  2. User's browser does not need sha-1 bouncer.
  3. User does not have DNT enabled.
  4. User falls within a predefined sample rate limit (set via an env variable). See https://bugzilla.mozilla.org/show_bug.cgi?id=1278981#c6

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1279291

## Testing
- Demo: https://www-demo4.allizom.org/en-US/
- Local testing: to test this locally you'll have to set `STUB_ATTRIBUTION_HMAC_KEY='foo'` in your `.env` file (the value for the key can be anything).

## Checklist
- [ ] Requires l10n changes.


alexgibson:
@pmac before we get into code review and QA, I'd like to get your thoughts here in terms of approach?

bedrock/base/templates/base-pebbles.html
```diff
@@ -3,7 +3,7 @@
 <!doctype html>
 {# Note the "windows" class, without javascript platform-specific
      assets default to windows #}
-<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %}>
+<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %} {% if settings.STUB_ATTRIBUTION_HMAC_KEY %}data-stub-attribution-rate="{{ settings.STUB_ATTRIBUTION_RATE }}"{% endif %}>
```
pmac: I think all of the switching off of the JS should happen based on the `STUB_ATTRIBUTION_RATE` setting, per [my comment in the other bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1278981#c6). This would mean that the JS could be included but the ajax endpoint would return a 403 if the rate were > 0 but no HMAC key were set, but that might possibly be useful for testing. The main thing though is that the inclusion of the JS and these values should be based on the RATE setting, not HMAC.

bedrock/base/templates/base-pebbles.html
```diff
@@ -154,7 +154,16 @@ <h2 class="masthead-logo"><a href="{{ url('mozorg.home') }}">{{ high_res_img('sa
     {% block site_js %}
       {% javascript 'common' %}
     {% endblock %}
+
+    {# Bug 1279291 #}
+    {% block stub_attribution %}
+      {% if settings.STUB_ATTRIBUTION_HMAC_KEY %}
```
pmac: also `STUB_ATTRIBUTION_RATE` here.

media/js/base/stub-attribution.js
```diff
@@ -0,0 +1,240 @@
...
+    StubAttribution.updateBouncerLinks = function(data) {
+        /**
+         * If data is missing or the browser does not meet requirements for
+         * stub attribution, then do nothing.
+         */
+        if (!data.attribution_code || !data.attribution_sig || !StubAttribution.meetsRequirements()) {
+            return;
+        }
+
+        $('a[href*="https://download.mozilla.org/"]').each(function() {
```
pmac: we only need to update the windows link. none of the other platforms will use these attributes.

media/js/base/stub-attribution.js
```diff
@@ -0,0 +1,240 @@
...
+    StubAttribution.appendToDownloadURL = function(url, data) {
+
+        if (!data.attribution_code || !data.attribution_sig) {
+            return url;
+        }
+
+        // append stub attribution query params.
+        $.each(data, function(key, val) {
+            if (key === 'attribution_code' || key === 'attribution_sig') {
+                url += (url.indexOf('?') > -1 ? '&' : '?') + key + '=' + val;
```
pmac: don't think you need to look for the `?`, if a link to `download.m.o` has no other query params it's worthless and we've got bigger problems.

pmac:
I do like the general direction. I was hoping you didn't have to parse the query params and could just grab them all and tack it onto the ajax request along with the referrer. You're welcome to do that still, but it looks like we've already got tools to find and parse UTM params, so that's cool. Like I said in a line comment, let's not use the HMAC setting for the JS at all. For one thing the RATE setting is the proper one in my opinion, and for another I'm not comfortable with a sensitive setting like that being referenced in the template code at all; a small mistake could expose the secret.

alexgibson:
> For one thing the RATE setting is the proper one in my opinion, and for another I'm not comfortable with a sensitive setting like that being referenced in the template code at all; a small mistake could expose the secret.

This is a great point, will update.

media/js/base/stub-attribution.js
alexgibson: Yes I don't think it's strictly necessary here, but also it does no real harm leaving this here for completeness imo.

media/js/base/stub-attribution.js
alexgibson: I've added this now. I made the call to append the params to both 32 and 64bit Windows URLs. The 64bit links don't seem to use the stub installer currently, but I'm making the assumption this could change in the future.

alexgibson:
@pmac _ I added a commit to make the Bouncer URL configurable via an env variable. I figured this was probably the simplest way for testing on demo? We can always remove this prior to merging.

pmac:
@alexgibson seems like a fine change. The bit in the JS is a little unfortunate, but perhaps we can find a way of picking those elements out of the DOM w/o having to match the href? But on the whole it may well come in handy again in the future to be able to set the bouncer URL.

alexgibson:
> The bit in the JS is a little unfortunate, but perhaps we can find a way of picking those elements out of the DOM w/o having to match the href?

Yeah, thinking about it we could just drop the href from the selector and look for `.download-link`. It was this way originally to avoid the sha-1 bouncer URL's, but this should be safe now considering we're also checking the data-attributes for `win` and `win64`.

alexgibson:
@pmac - this  hopefully feels a little cleaner. I still had to add a check to make sure the link isn't transitional, but this at least avoids needing to depend on the Bouncer end point.

pmac:
Very nice!

alexgibson:
@pmac it occurred to me that using a session cookie might be too long for some people, particularly those who don't restart their browsers that often. I changed this PR to use a cookie that expires in 7 days, but even that might be too long. Any thoughts on what a sensible expiry date should be?

pmac:
@alexgibson great catch. I'd think 24 hours would be plenty. After that I'm not sure we can say that the old source is what drove the download if they do finally complete it. Is it written in such a way that coming from a new source, even if there is already a cookie will overwrite or update it? This might be a good question for @chrismore et al.

alexgibson:
> Is it written in such a way that coming from a new source, even if there is already a cookie will overwrite or update it?

Not currently, the existing cookie would have to expire. This also saves on the number of requests we make to auth, so I'd prefer not to have to make things overwrite unless really necessary. 24 hours for the cookie should suffice I hope.

chrismore:
24 hours sounds fine to me! Thanks

ejregithub:
@alexgibson - dark launch of the service has been scheduled for the morning of Thursday Feb 2nd.  Can you please look at a rebase here / put this into prod, behind switches,  in time for that date, when Paul will flip the switches to on?

pmac:
Just FYI @alexgibson, I changed how bedrock encodes the data (oremj and I decided to switch) in PR #4609, which has now merged. So I rebased `demo__4` branch on current master and pushed it. The rebase was clean, but I didn't double check anything. But, if you want you could potentially start from the new `demo__4` branch which is rebased, or do your own rebase on your branch and we can push the demo again when you're happy.

alexgibson:
> Just FYI @alexgibson, I changed how bedrock encodes the data (oremj and I decided to switch) in PR #4609, which has now merged. So I rebased demo__4 branch on current master and pushed it. The rebase was clean, but I didn't double check anything. But, if you want you could potentially start from the new demo__4 branch which is rebased, or do your own rebase on your branch and we can push the demo again when you're happy.

I've updated rebased and updated this branch, and re-pushed it to demo 4 - will give it some testing. I also added some extra logic to support the modal download links that we're just added to /new.

alexgibson:
@jpetto - I think this is now ready again for review. Appreciate if you could r?

media/js/base/stub-attribution.js
```diff
@@ -0,0 +1,264 @@
...
+        var utmCount = 0;
+
+        for (var utm in params) {
+            if (params.hasOwnProperty(utm)) {
+                utmCount += 1;
+            }
+        }
+
+        // if there are no utm params and no referrer, do nothing.
+        if (utmCount === 0 && (typeof referrer === 'undefined' || referrer === '')) {
```
jpetto: Apologies if this has been covered elsewhere, or is totally out of scope, but do we care about links coming from marketing emails or other non-web page sources? I'm not sure those have a referrer...

media/js/base/stub-attribution.js
```diff
@@ -0,0 +1,264 @@
...
+    /**
+     * Determine if the current page is scene2 of /firefox/new/.
+     * This is needed as scene2 auto-initiates the download. There is little point
+     * trying to make an XHR request here before the download begins, and we don't
+     * want to make the request a dependency on the download starting.
+     * @return {Boolean}.
+     */
+    StubAttribution.isFirefoxNewScene2 = function(location) {
+        location = typeof location !== 'undefined' ? location : window.location.href;
+        return location.indexOf('/firefox/new/?scene=2') > -1;
```
jpetto: Since we're dealing with external links, are we sure `scene=2` will always directly follow the `?`? Could we hit a scenario where `_utm` params are injected before the `scene=2` key/val pair?

jpetto:
Assuming this was discussed along the way, but what's the expected flow for this attribution? Guess is:

Windows user lands on external site X, which provides a `_utm`'d link to any page on mozilla.org *except* for `/firefox/new/?scene=2`. User lands on said page, gets the attribution info in a cookie, then clicks a download link on that page/during that visit pointing to `/firefox/new/?scene=2`, where attribution is pulled from cookie and injected prior to download kicking off? (Alternatively, user lands on a page like `/firefox/channel` with direct links and has attribution code injected there directly.)

media/js/base/stub-attribution.js
```diff
@@ -0,0 +1,264 @@
...
+        // target download buttons and other-platforms modal links.
+        $('.download-list .download-link, .download-platform-list .download-link').each(function() {
+            var version;
+            // If this is a transitional download link do nothing.
+            if (this.href && this.href.indexOf('/firefox/new/?scene=2') === -1) {
+
+                version = $(this).data('downloadVersion');
+                // Currently only Windows 32bit uses the stub installer, but this could
+                // easily change in the future so we'll make that bet now.
+                if (version && (version === 'win' || version === 'win64')) {
```
alexgibson: Hmm, no - but perhaps my comment didn't make sense. I'm betting that they will want attribution data for both 32 and 64 bit builds eventually (even though it only forks with 32 bit installer currently), so I'm adding that case now. I'll update the comment to make it a little clearer.

media/js/base/stub-attribution.js
alexgibson: Currently the ask is to capture utm params, and failing those are not present, to fall back to referrer. If neither exist we do nothing. As for emails, if those links don't use utm params for tracking then we could add some special casing for those. but until someone asks for it, I think we're fine.

media/js/base/stub-attribution.js
alexgibson: Good, call - will fix 👍

alexgibson:
> Windows user lands on external site X, which provides a _utm'd link to any page on mozilla.org except for /firefox/new/?scene=2. User lands on said page, gets the attribution info in a cookie, then clicks a download link on that page/during that visit pointing to /firefox/new/?scene=2, where attribution is pulled from cookie and injected prior to download kicking off? (Alternatively, user lands on a page like /firefox/channel with direct links and has attribution code injected there directly.)

Correct yeah. I should have probably talked you through this up front (sorry). But the fact you managed to work it all out OK just from the code / comments gives me some faith that this isn't too arcane 👍

alexgibson:
Thanks for the review @jpetto - I pushed a commit with some fixes.

jpetto:
Nice! LGTM 👍

PR #4462, opened 2016-11-11
Add gulp clean task for /static when in local development

alexgibson:
Add a little `gulp static:clean` task as a handy alternative to using `git clean` for static media (which is arguably simpler, but this seems a little less dangerous for accidents occurring). I stopped short of running this automatically when just running `gulp`, but it's something we *could* also do if we thought it was beneficial enough (although it could create issues of its own). But it seems like this is mainly useful for PR's that remove code, which are obvious enough to spot.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -5,6 +5,7 @@
 /* global __dirname, require */
 
 const gulp = require('gulp');
+const del = require('del');
 const karma = require('karma');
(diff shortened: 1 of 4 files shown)


jpetto:
LGTM. Much safer than `git clean` (which I've run from the wrong dir before 😞).

r+ ✨

PR #4466, opened 2016-11-15
[fix bug 1313708] Yahoo Search Test /firstrun

alexgibson:
## Description
- Yahoo /firstrun funnelcake experiment. Both copy and imagery are still subject to change.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1313708

## Testing
- `/en-US/firefox/50.0/firstrun/?f=92` shows the Yahoo page.
- `/en-US/firefox/50.0/firstrun/?f=91` shows the control page.

## Checklist
- [ ] Requires l10n changes.


alexgibson:
Ok, this now has final copy / design. Should be good for review, but leaving the dnm label until everything is approved on demo.

alexgibson:
@jpetto r?

alexgibson:
We have the go-ahead to merge / push this once reviewed.

media/css/firefox/firstrun/yahoo-retention.less
```diff
@@ -0,0 +1,201 @@
...
+            &:last-child:after {
+                content: '';
+                margin: 0;
+            }
+        }
+    }
+
+    a:link,
+    a:visited {
+        .transition(color .1s ease-in-out);
```
jpetto: As this is firstrun only, do we need the mixin?

media/js/firefox/firstrun/yahoo-retention.js
```diff
@@ -0,0 +1,84 @@
...
+            window.dataLayer.push({
+                'event': 'yh-click',
+                'eventLabel': 'Yahoo Search Click'
+            });
+        });
+    }
+
+    function initSearchUI(e) {
+        e.preventDefault();
+        shouldPromoteYahooSearch().then(function(result) {
```
jpetto: I imagine nearly all users only click this button once, so it's probably not worth caching the result of this promise. Just thinking out loud. 😁

alexgibson: They could still keep this page open for some reason. I'd rather err on the side of caution and leave this in here.

jpetto: I was just wondering if, after clicking the button once, we should remember the result of the promise so as not to re-run the UITour stuff on subsequent clicks. I suppose the user could make profile changes that would invalidate a previous result though...

jpetto: Agreed.

alexgibson:
@jpetto thanks for the review, updated!

PR #4473, opened 2016-11-17
Add Code of Conduct notice

pmac:
Copied from:
https://github.com/mozilla/addons-server/blob/master/CODE_OF_CONDUCT.md

Thanks to @andymckay

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -8,3 +8,13 @@ Following these guidelines will very much increase the chances of a pleasant
 code-review experience.
 
 [bedrock docs]: http://bedrock.readthedocs.org/en/latest/contribute.html
+
+Code of Conduct
+---------------
+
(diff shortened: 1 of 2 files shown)


pmac:
@mozilla/webdev-bedrock andym posted about this in #webdev and it seemed like a good idea to me. Probably also belongs in [sugardough](https://github.com/mozilla/sugardough). Thoughts?

pmac:
Could also just be an addition to the README I guess.

alexgibson:
I like this. Even though GitHub doesn't do anything with the file in particular, I agree it's useful to have around. Another place for it to live could be in CONTRIBUTING.md?

pmac:
CONTRIBUTING.md does make a lot of sense. I think posting it to both README and CONTRIBUTING might be fine. I'll update and see how it looks.

alexgibson:
LGTM, r+

PR #4474, opened 2016-11-17
Fix bug 1318429: Include basket api key with email pref center update

pmac:
A valid API key is requrired if they're interacting with a private newsletter
(e.g. the NDA Mozillians one)


jpetto:
Will this have any negative impact when sent along with/for non-private newsletters?

pmac:
@jpetto no. it'd be best if we didn't send the API key across the wire when it isn't necessary, but we can't really know that reliably before hand with this particular call. But we only send it with this one and not subscription in general, so it should be fine.

PR #4476, opened 2016-11-18
[fix bug 1316375] EOY 2016 fundraiser home page takeover

craigcook:
## Description
Adds two variants of a home page fundraising takeover, using our traffic cop script to split the audience 50/50. This is due to go live on 1 December, though it could/should merge before then and remain hidden behind the switch until the appointed time.

Switches:
`mozorg-home-fundraising-2016` to toggle the takeover.
`experiment-home-fundraising-2016` to toggle the traffic cop redirect.

Uses the l10n tag `mozorg-home-fundraising-2016`. The text strings should be the same as last year so we hope to reclaim those translations rather than start over.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1316375

## Testing
Version A is the default, version B takes a ?v=b parameter.
You should only see the takeover once per session, the first time you land on the home page. To see it again you need to open a new tab/window or clear your cookies.

Demo A: https://bedrock-demo-craigcook.us-west.moz.works/
Demo B: https://bedrock-demo-craigcook.us-west.moz.works/?v=b

## Checklist
- [x] Requires l10n changes.


craigcook:
Some ToDos before this merges:

- [ ] Verify correct UTM parameters
- [ ] Add/update tests

craigcook:
Oh, this also cleans up code from the previous Firefox copy testing, https://bugzilla.mozilla.org/show_bug.cgi?id=1313717

flodolo:
@TheoChevalier Just to confirm, this is something you're going to work on or should we?

TheoChevalier:
@flodolo I will work on it, if that’s fine with you guys

flodolo:
Absolutely yes. Ping me if you need anything.

media/js/mozorg/home/takeover-2016.js
```diff
@@ -0,0 +1,124 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+$(function () {
```
alexgibson: Could we run this code straight away instead of waiting for dom ready?

alexgibson:
RTL could do with a few tweaks:

rtl

alexgibson:
May not be worth worrying about, but the radio fields in IE9 are a little off:

radio-fields

alexgibson:
Aside from a few nits / suggestions this all looks good 👍. The main thing I think we need to work out is if we should be using the different currencies available for the takeover, as this could be a bit confusing for people.

media/js/mozorg/home/experiment-home-fundraising-2016.js
```diff
@@ -6,12 +6,9 @@
...
 
     var cop = new Mozilla.TrafficCop({
-        id: 'experiment-home-fx-copy',
+        id: 'experiment-home-fundraising-2016',
         variations: {
-            'v=a': 25,
-            'v=b': 25,
-            'v=c': 25,
-            'v=d': 25 // double control group
+            'v=b': 50
```
alexgibson: Afaik Peter has suggested this is what we should be doing. Maybe ask him for confirmation?

bedrock/mozorg/views.py
```diff
@@ -254,20 +254,27 @@ def home_tweets(locale):
...
+    donate_params['preset_list'] = donate_params['presets'].split(',')
 
     if lang_file_is_active('mozorg/home/index-2016', locale):
-        # check for variant - fx copy experiment
+        # check for variant - fundraising takeover
         v = request.GET.get('v', None)
 
-        if (locale.startswith('en-') and v in ['a', 'b', 'c']):
-            template = 'mozorg/home/home-{0}.html'.format(v)
+        if (locale.startswith('en-') and v == 'b'):
```
craigcook: No, both versions should be localized (same strings, only different designs). I'll fix this.

media/js/mozorg/home/takeover-2016.js
```diff
@@ -0,0 +1,124 @@
...
+    $donateCustom.keyup(clearOptions);
+
+    // the name of the parameter that indicates the amount the user
+    // wish to donate has to be called amount. Seeing that we have
+    // radio buttons with that name as well as an input field, we need
+    // to disable the input field prior to form submission to avoid a
+    // scenario where we have ?amount=50&amount=
+    $form.on('submit', function() {
+        if ($donateCustom.val() === '') {
+            $donateCustom.attr('disabled', 'disabled');
```
craigcook: That's a tricky one. We need to disable it on submit to prevent sending the blank amount, but upon using the back button there isn't any sort of event to trigger a reset. My first thought is a timer that just occasionally checks if the field is disabled and re-enables it if so, but that seems like a bad idea... any thoughts on a simple solution or should we just let it slide?

media/js/mozorg/home/takeover-2016.js
craigcook: Perhaps... this was reused from last year's takeover so I didn't rethink it much. Though with the script loaded at the end of the body would we gain much by firing it before DOMready?

tests/pages/home.py
```diff
@@ -10,10 +10,23 @@
...
     _download_button_locator = (By.ID, 'nav-download-firefox')
     _get_firefox_link_locator = (By.ID, 'fx-download-link')
 
     @property
+    def is_take_over_displayed(self):
+        return self.is_element_displayed(*self._take_over_locator)
+
+    def wait_for_page_to_load(self):
+        super(BasePage, self).wait_for_page_to_load()
+        if self.is_take_over_displayed:
```
alexgibson: Adding something like this before this line should do the trick:

```
el = self.find_element(By.TAG_NAME, 'html')
self.wait.until(lambda s: 'eoy-takeover' in el.get_attribute('class'))
```

Then at the end of `takeover-2016.js`:

```
$('html').addClass('eoy-takeover');
```

alexgibson:
This is only minor, but there's some slight overlap with the copy bleeding onto the photo's in version b using RTL. Happy to let this one slide, not a blocker.

rtl-b

bedrock/mozorg/templates/mozorg/home/includes/takeover-2016.html
```diff
@@ -0,0 +1,79 @@
...
+    <span class="snap shot-two"></span>
+    <span class="snap shot-three"></span>
+    <span class="snap shot-four"></span>
+    <span class="snap shot-five"></span>
+    <span class="snap shot-six"></span>
+    <span class="snap shot-seven"></span>
+  </div>
+
+  <div class="header-strip">
+    <p class="content"><button type="button" id="close_takeover" role="button">{{ _('Continue to mozilla.org') }}</button></p>
```
alexgibson: No need for `role="button"` here any longer?

alexgibson:
Looking good, the currency logic seems to be working well 👍 r+wc

TheoChevalier:
@craigcook The French version shows “€xxx”, but it should be “xxx €”
capture d ecran 2016-11-24 a 12 52 26

Could we use the same thing as on the snippet and the donate website for currency formatting? https://github.com/mozmar/snippets/blob/master/campaigns/mofo-eoy-2016/snippet.html#L317-L321

Also, small layout issue with Russian:
capture d ecran 2016-11-24 a 12 57 32

craigcook:
@TheoChevalier The currency symbol is part of the string now, which I realize is a new string and last year's translations won't work as-is. To change the order one would need to update the translated string, e.g. `%(symbol)s%(sum)s` becomes `%(sum)s %(symbol)s`.

To avoid changing all those strings that JS function may do the trick, but it'll take a bit more restructuring of the form to do it. I'll work on it.

On a slight tangent, what I *don't* like about using JS for this is that it's relying on script to inject essential content into the HTML. If JavaScript fails for any reason those buttons are empty and unusable. That's probably acceptable here because we're not showing the takeover at all if JS fails, but I really wish the form on donate.mozilla.org was using progressive enhancement.

craigcook:
@alexgibson I saw that overlap in RTL but it happens right at the edge of a breakpoint and I didn't want to reposition everything for that. I think we can let it slide.

media/js/mozorg/home/takeover-2016.js
```diff
@@ -19,6 +19,18 @@
     var $optionLabels = $('label[class!="own-amount"]', $donateOptionsSection);
     var $options = $('input[type="radio"]', $donateOptionsSection);
 
+    var locale = $form.data('locale');
+    var currencyCode = $form.data('currency');
+    var numberFormat = new Intl.NumberFormat(locale, {
```
alexgibson: Is this going to throw errors in browsers that don't support these functions? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat

craigcook: Yeah, I haven't tested in older browsers but I imagine it would. Thinking I'll just wrap this in some feature detection and let it fall back to the inline text.

craigcook:
I've added a basic feature detection for Intl.NumberFormat so older browsers should just get the template text without JS replacement. If locales update that string it'll show their desired format even in older browsers because that's what will be in the template, but with the old strings in an old browser they'll get something like "€x" ... not ideal but probably acceptable as a worst-case fallback.

I've also slightly reworked the include to send the variant in a UTM parameter, which was requested in the bug.

With that, I think we're good to merge this pending a final r+. Taking the DNM flag off. I've left the commits separate for review but will squash before merging.

media/js/mozorg/home/takeover-2016.js
```diff
@@ -0,0 +1,139 @@
...
+    var $takeoverModal = $('#fundraising_takeover');
+    var $overlay = $('.page-overlay');
+    var $closeModal = $('#close_takeover');
+    var $form = $('#fundraising-form');
+    var $donateOptionsSection = $('.donate-options');
+    var $donateCustom = $('#donate-custom');
+    var $optionLabels = $('label[class!="own-amount"]', $donateOptionsSection);
+    var $options = $('input[type="radio"]', $donateOptionsSection);
+
+    if ('Intl' in window && 'NumberFormat' in Intl && $takeoverModal.length > 0) {
```
alexgibson: Instead of putting this inline at the top of the file, suggest making it a function which is then called from inside `initTakeOver()`?

bedrock/mozorg/templates/mozorg/home/home-voices.html
```diff
@@ -359,6 +359,14 @@ <h2 class="primary go">{{ _('Mozilla Developer Network') }}</h2>
   <button id="scroll-prompt">{{ _('Discover more') }}</button>
 </main>
 
+{% if l10n_has_tag('mozorg-home-fundraising-2016') and switch('mozorg-home-fundraising-2016') %}
```
alexgibson: Are there many locales still on the old home page? Just curious if adding this is worth the extra complexity. If there are still some locales, it's definitely worth doing.

alexgibson:
A couple of questions / suggestions but otherwise looks good, r+wc

bedrock/mozorg/templates/mozorg/home/home-voices.html
flodolo: We still have a bunch of them :-\

At some point we should remove that template and clean up things.

bedrock/mozorg/templates/mozorg/home/home-b.html
```diff
@@ -6,10 +6,11 @@
...
 {% add_lang_files "mozorg/home/index-2016" %}
 
-{% block fx_copy %}
-  <h2>The Web is yours</h2>
+{% block body_class %}{{ super() }} variant-b{% endblock %}
 
-  <p>
-    Explore it your way with the speed and security of Firefox.
-  </p>
+{% block home_takeover %}
```
craigcook: Doh, yes it should. Fixed.

PR #4490, opened 2016-11-24
Bug 1320119 - Update donate currencies

TheoChevalier:
## Description
Follow up of bug 1316375, updating the currencies after the changes landed in https://github.com/mozilla/donate.mozilla.org/commit/7e9b6825f8fe2a15971e57f2edec335df8f768fd

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1320119

## Testing
- Check donate link on homepage footer

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -698,6 +698,12 @@ DONATE_PARAMS = {
         'presets': '100,50,25,15',
         'default': '50'
     },
+    'bn-IN': {
+        'currency': 'inr',
+        'symbol': u'₹',
+        'presets': '1000,500,250,150',
+        'default': '500'
+    },
     'brx': {
         'currency': 'inr',
         'symbol': u'₹',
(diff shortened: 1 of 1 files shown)


alexgibson:
This will likely need a rebase and updating after #4476 merges, which adds a symbol property to `DONATE_PARAMS`. Either that or @craigcook could add these directly to his PR.

TheoChevalier:
@alexgibson good catch, I rebased against Craig’s PR and added symbols. I guess I will have to do a final rebase after #4476 is merged

craigcook:
#4476 finally merged so you should be able to rebase this from master to tidy it up.

TheoChevalier:
Ah true, forgot about that. Will do

TheoChevalier:
@craigcook ready for review

bedrock/settings/base.py
```diff
@@ -698,6 +698,11 @@ def lazy_email_backend():
         'presets': '100,50,25,15',
         'default': '50'
     },
+    'bn-IN': {
+        'currency': 'inr',
```
jpetto: No symbol needed here?

TheoChevalier: doh, fixed

craigcook:
A minor tweak would be to add any locales that position the symbol to the right of the number to this rule in `takeover-2016.scss`: https://github.com/mozilla/bedrock/blob/master/media/css/mozorg/home/takeover-2016.scss#L241

May only be bn-IN that needs it?

craigcook:
One possible adjustment to the CSS but otherwise this looks good. r+ 💸

PR #4494, opened 2016-11-28
Fixed Search box responsive styling on /firefox/all page

RCMainak:
## Description
Fixed Search box responsive styling on /firefox/all page
## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1301363

### DIFF
--- a/media/css/firefox/all.less
+++ b/media/css/firefox/all.less
@@ -79,8 +79,6 @@ body.nightly {
 #language-search {
     padding-top: @baseLine;
     padding-bottom: @baseLine;
-    margin-left: @gridGutterWidth * 1.5;
-    margin-right: @gridGutterWidth * 1.5;
     input[type=search] {
         .span(3);
         float: none;
@@ -284,7 +282,7 @@ body.nightly {
             vertical-align: middle;
         }
         .search-column {
-            width: 340px;
+            width: 460px;
             padding: 5px 0;
         }
         .other-column {
@@ -346,16 +344,17 @@ body.nightly {
     }
 
     #language-search {
-        margin-left: @gridGutterWidth;
-        margin-right: @gridGutterWidth;
         input[type=search] {
-            .span(3);
+            width: 100%;
             float: none;
             margin-left: 0;
             vertical-align: middle;
         }
+        button {
+            width: 100%;
+            margin-top: 10px;
+        }
         .search-column {
-            margin: 0 0 (@baseLine / 2) 0;
             float: none;
             width: auto;
         }


media/css/firefox/all.less
```diff
@@ -276,6 +276,7 @@ body.nightly {
     }
 
     #language-search {
+        width: @widthTablet - (@gridGutterWidth * 2 * 2.25);
```
alexgibson: Rather than make this smaller to fit it's container I think it might be better to make `.search-column` larger at this breakpoint, to make better use of available space. It could probably remain the same size as it was at desktop width.

media/css/firefox/all.less
```diff
@@ -349,11 +350,14 @@ body.nightly {
...
         margin-right: @gridGutterWidth;
         input[type=search] {
-            .span(3);
+            width: 100%;
             float: none;
             margin-left: 0;
             vertical-align: middle;
         }
+        button {
+            width: 100%;
```
alexgibson: This could also do with a `margin-top: 10px` to give a little spacing between the input and the button.

media/css/firefox/all.less
RCMainak: there is something wrong with the width of `#language-search`. `.search-column` sdoesn't seems to slove the problem I'm trying to solve.

alexgibson: `.search-column` changes from `460px` width at desktop break-point to `340px` width at tablet break-point. The language-search form has a total maximim width of `710px` here, so there should be plenty of room to avoid having to change the width to such a small value here.

RCMainak: nah! That was not my intention. At tablet break-point (around 760-800 px) the width of `#language-search` is more than it's parents which is breaking the view. I couldn't figure out where the width is coming from so, thought about fixing the width by that wizardry :stuck_out_tongue_winking_eye: 

I haven't tested it on an actual devices. will attach a image when I get back home.

alexgibson: The width for language search is set here at tablet viewport: https://github.com/mozilla/bedrock/blob/master/media/css/firefox/all.less#L278-L297.

That certainly needs fixing too (nice spot), but my comment review comment still applies even after fixing this.

alexgibson: So to try and re-iterate what I think needs doing here:

1.) fix the overall width of `#language-search` at tablet break-point so it doesn't overflow it's parent container width (do this by changing the width of the child elements of `#language-search`, instead of trying to override the width here.
2.) Try setting .`search-column` to a slightly more user friendly, not so small width at tablet break-point.

alexgibson: Hmm, actually - please ignore my suggestion on making the search box wider. It looks like the extra space is reserved for a second (conditional) column. I'm not entirely convinced this column is ever used / shown, but maybe it is best not to disturb it here.

Please just set the width here explicitly to `660px` (this should override the inherited width) :)

RCMainak:
It should look good.

alexgibson:
r+, thanks!

PR #4497, opened 2016-11-29
[fix bug 1319207] Add custom privacy URL for Firefox Focus in de locale.

jpetto:
## Description

Creates a custom privacy URL for Firefox Focus visitors hitting the `de` locale.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1319207

## Testing

Ensure that visiting `/de/privacy/firefox-focus/` redirects to `/de/privacy/firefox-klar/`. Also ensure any non-de locale hitting `/privacy/firefox-klar/` is redirected to `/{locale}/privacy/firefox-focus/`.

## Checklist

### DIFF
--- a/bedrock/privacy/redirects.py
+++ b/bedrock/privacy/redirects.py
@@ -1,0 +1,10 @@
+from bedrock.redirects.util import no_redirect, redirect
+
+redirectpatterns = (
+    # bug 1319207
+    # 'Firefox Focus' cannot be used in de locale due to legal constraints
+    redirect(r'^de/privacy/firefox-focus/?', '/de/privacy/firefox-klar/', locale_prefix=False),
+    # special de URL should not be accessible from other locales
+    no_redirect(r'^de/privacy/firefox-klar/?', locale_prefix=False),
+    redirect(r'^privacy/firefox-klar/?', 'privacy.notices.firefox-focus'),
+)
--- a/bedrock/privacy/templates/privacy/index.html
+++ b/bedrock/privacy/templates/privacy/index.html
@@ -83,7 +83,14 @@
         <li class="policy-firefox-os"><a href="{{ url('privacy.notices.firefox-os') }}">{{ _('Firefox OS') }}</a></li>
         <li class="policy-firefox-cloud"><a href="{{ url('privacy.notices.firefox-cloud') }}">{{ _('Firefox Cloud Services') }}</a></li>
         <li class="policy-marketplace"><a href="https://marketplace.firefox.com/privacy-policy">{{ _('Firefox Marketplace') }}</a></li>
-        <li class="policy-firefox-focus"><a href="{{ url('privacy.notices.firefox-focus') }}">{{ _('Firefox Focus') }}</a></li>
+        <li class="policy-firefox-focus">
+        {% if LANG == 'de' %}
+          {# bug 1319207 - As product names are not exposed to L10n, best to specify unique name for de #}
+          <a href="{{ url('privacy.notices.firefox-klar') }}">Firefox Klar</a>
+        {% else %}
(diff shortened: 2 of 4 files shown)


jpetto:
@pmac r?

bedrock/privacy/templates/privacy/index.html
```diff
@@ -83,7 +83,13 @@ <h1 itemprop="name">{{ doc.h1.string }}</h1>
         <li class="policy-firefox-os"><a href="{{ url('privacy.notices.firefox-os') }}">{{ _('Firefox OS') }}</a></li>
         <li class="policy-firefox-cloud"><a href="{{ url('privacy.notices.firefox-cloud') }}">{{ _('Firefox Cloud Services') }}</a></li>
         <li class="policy-marketplace"><a href="https://marketplace.firefox.com/privacy-policy">{{ _('Firefox Marketplace') }}</a></li>
-        <li class="policy-firefox-focus"><a href="{{ url('privacy.notices.firefox-focus') }}">{{ _('Firefox Focus') }}</a></li>
+        <li class="policy-firefox-focus">
+        {% if LANG == 'de' %}
+          <a href="{{ url('privacy.notices.firefox-klar') }}">Firefox Klar</a>
```
jpetto: I threw this in as a nicety to send `de` users directly to the `-klar` URL. I could make the URL a variable with `set` to make things a little cleaner/not introduce an unwrapped string.

flodolo: We don't usually expose brand names to locales. In this case, it would also be a waste of resources, since all locales would have to confirm it, with the exception of German.

bedrock/privacy/redirects.py
```diff
@@ -0,0 +1,9 @@
+from bedrock.redirects.util import redirect
+
+redirectpatterns = (
+    # bug 1319207
+    # 'Firefox Focus' cannot be used in de locale due to legal constraints
+    redirect(r'^de/privacy/firefox-focus/?', '/de/privacy/firefox-klar/', locale_prefix=False),
+    # special de URL should not be accessible from other locales
+    redirect(r'^(?P<locale>(?!de)\w{2,3}(?:-\w{2})?/)privacy/firefox-klar/?', 'privacy.notices.firefox-focus', locale_prefix=False),
```
pmac: instead of this complex regex you could use the `no_redirect` helper which is also in the `bedrock.redirects.util` lib. It stops redirect processing and forces the normal Django process. So this could be:

```python
from bedrock.redirects.util import no_redirect, redirect

redirectpatterns = (
    # bug 1319207
    # 'Firefox Focus' cannot be used in de locale due to legal constraints
    redirect(r'^de/privacy/firefox-focus/?', '/de/privacy/firefox-klar/', locale_prefix=False),
    # special de URL should not be accessible from other locales
    no_redirect(r'^de/privacy/firefox-klar/?', locale_prefix=False),
    redirect(r'^privacy/firefox-klar/?', 'privacy.notices.firefox-focus'),
)
```

Both of these work, but I find the one with `no_redirect` a bit more readable. Just a suggestion though. I'm happy with what you have since it works :)

bedrock/privacy/templates/privacy/index.html
jpetto: @flodolo So are you suggesting we leave this code as-is, to ensure the correct product name is used for `de`?

flodolo: Yes, I think it's the right think to do.

bedrock/privacy/redirects.py
jpetto: Agree using `no_redirect` with a simpler subsequent `redirect` is more readable. Will update!

jpetto:
Waiting on confirmation of when to merge/push. Looking like later this afternoon.

PR #4500, opened 2016-11-30
Remove non-fx redirect from firstrun/whatsnew

pmac:
An unthrottled Fx update that opens /whatsnew/ causes
massive spike in traffic to bedrock, so these pages
can not be uncached at the CDN.


alexgibson:
This looks good to me. Tested locally and both firstrun and whatsnew pages no longer redirect in Chrome. Translations look good on locale pages.

I'd appreciate if perhaps another back-end dev could take a look code wise, but this looks pretty straight forward.

alexgibson:
r+

PR #4501, opened 2016-11-30
[fix bug 1319624] Add Julie Hanna to MoCo board

craigcook:
## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1319624


bedrock/mozorg/templates/mozorg/about/leadership.html
```diff
@@ -1141,6 +1141,12 @@ <h3 class="group-title" id="board-corp">{{ _('Mozilla Corporation') }}</h3>
       </li>
       <li class="vcard" itemscope itemtype="http://schema.org/Person">
         <figure class="headshot">
+          <img class="photo" itemprop="image" alt="" src="{{ static('img/mozorg/about/leadership/julie-hanna.jpg') }}">
```
jpetto: Should this use the `high_res_img` helper?

craigcook: Yup, it should. Seems I had a few headshots that hadn't gotten the high-res treatment because they were placeholders until the last minute, but we have high-res pics for those folks now. Updated this one and the others (still only have small pics for some people).

jpetto:
LGTM. r+ 📸

PR #4502, opened 2016-11-30
[fix bug 1321033] Hello EOL legal doc cleanup/archival.

jpetto:
## Description

Archives the current Hello [privacy policy](https://www.mozilla.org/privacy/firefox-hello/) and [terms of service](https://www.mozilla.org/about/legal/terms/firefox-hello/). Adds redirects from decommissioned URLs to the new archive pages.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1321033

## Testing

Content on archive pages is copy/paste from HTML generated by existing legal docs markdown, so double check tag completion/nesting. Make sure redirects work as expected.

## Checklist

### DIFF
--- a/bedrock/legal/redirects.py
+++ b/bedrock/legal/redirects.py
@@ -4,4 +4,7 @@ from bedrock.redirects.util import redirect
 redirectpatterns = (
     # bug 1243240
     redirect(r'^about/legal/report-abuse/?$', 'legal.report-infringement'),
+
+    # bug 1321033
+    redirect(r'^about/legal/terms/firefox-hello', 'privacy.archive.hello-2014-11'),
 )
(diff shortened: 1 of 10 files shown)


jpetto:
@craigcook r?

alexgibson:
Drive by comment - there's a defunct functional test url here that can be removed as part of this change: https://github.com/mozilla/bedrock/blob/master/tests/functional/test_generated_pages.py#L32

jpetto:
> Drive by comment - there's a defunct functional test url here that can be removed as part of this change: https://github.com/mozilla/bedrock/blob/master/tests/functional/test_generated_pages.py#L32

Updated. Thanks @alexgibson!

craigcook:
Looks like we need to update `legal/index.html` as well. Getting a NoReverseMatch error for 'legal.terms.firefox-hello' referenced in that file. I think we should just delete Hello from that page.

jpetto:
> Looks like we need to update legal/index.html as well. Getting a NoReverseMatch error for 'legal.terms.firefox-hello' referenced in that file. I think we should just delete Hello from that page.

Good catch. Agree it should be removed.

PR #4503, opened 2016-11-30
Add switch for enabling the 2015 annual report

pmac:
Switch can be enabled with the following env var:

```
SWITCH_SOM_2015=on
```


pmac:
@jpetto r?

jpetto:
LGTM. r+

PR #4506, opened 2016-12-01
[fix bug 1321440] unbreak fundraising takeover buttons

craigcook:
## Description
So in my enthusiasm for localizing the currency with the fancy new Intl API I foolishly broke the entire form. I was replacing the label contents with the localized text and didn't even notice that I was also replacing the hidden radio buttons.

This fixes my stupid mistake by wrapping the label text in a span and replacing THAT instead.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1321440

## Testing
Verify the form works and submits the selected value to donate.mozilla.org

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/home/includes/takeover-2016.html
+++ b/bedrock/mozorg/templates/mozorg/home/includes/takeover-2016.html
@@ -28,11 +28,10 @@
         <ul>
         {% for amount in donate_params.preset_list %}
           <li>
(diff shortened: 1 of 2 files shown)


craigcook:
@alexgibson r? This needs to go to production before we can switch on the takeover.

alexgibson:
Verified that the right values are passed along to the form 👍 r+

PR #4508, opened 2016-12-01
[fix bug 1320808] Add Helen Turvey to MoFo board

craigcook:
## Description
Due to go live Monday, 5 December but uses the switch `leadership-turvey-announce` so we don't have to do a prod push from Hawaii.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1320808


jpetto:
r+ 🏛

PR #4512, opened 2016-12-02
Add links to 2015 annual report and tax filings

tbm:
## Description

Add links to the 2015 annual report and tax filings which are currently missing.

## Bugzilla link

Not reported to Bugzilla. Is this something I should do next time?

## Testing

None, seems obvious enough...


craigcook:
These are new strings though the page isn't localized so there shouldn't be any impact. r+!

PR #4513, opened 2016-12-02
Remove spurious whitespace in links from 2015 annual report

tbm:
## Description

Remove spurious whitespace in links from 2015 annual report.

## Bugzilla link

Not reported to Bugzilla. Is this something I should do next time?

## Testing

None, seems obvious enough...


craigcook:
Looks good. These updates change the strings but the page isn't currently localized (and I'm not sure it ever will be). Tagging l10n just in case. r+ 👾

PR #4514, opened 2016-12-02
Add utility view for a/b tests

pmac:
Simple view for use when you want one alternate via a query param:

```
/your/url/?v=b
```

Also clean up google search experiment views using this new utility.


pmac:
@jpetto whaddya think? Look okay? Useful for future?

pmac:
@jpetto updated and much more useful and flexible now. There's a class-based-view Mixin available as well for adding this to existing CBVs. Let me know what you think of this approach.

bedrock/utils/views.py
```diff
@@ -0,0 +1,33 @@
...
+from lib.l10n_utils import LangFilesMixin
+
+
+class VariationMixin(object):
+    template_name_variations = None
+    template_context_variations = None
+
+    def get_template_names(self):
+        names = super(VariationMixin, self).get_template_names()
+        if self.template_name_variations:
```
jpetto: I haven't yet seen an instance where an experiment used fewer custom templates than variations, but I don't think this is a bad scenario for which to prepare. The downside is the values for `template_name_variations` and `template_context_variations` will often be the same/repeated, but that's not a big sacrifice and shouldn't be too error prone.

bedrock/mozorg/views.py
```diff
@@ -362,21 +362,7 @@ def internet_health(request):
...
-    version = request.GET.get('v', None)
-
-    if (locale != 'en-US' or version != 'b'):
-        version = None
-
-    return l10n_utils.render(request, 'mozorg/about/manifesto.html', {'version': version})
+about = VariationTemplateView.as_view(template_name='mozorg/about.html',
+                                      template_context_variations=['b'])
+about_manifesto = VariationTemplateView.as_view(template_name='mozorg/about/manifesto.html',
+                                                template_context_variations=['b'])
```
jpetto: Could these now be put back into `urls.py`?

jpetto:
Oh, would be nice to have docs on this as well, but that can come in a follow-up PR.

bedrock/utils/tests/test_views.py
```diff
@@ -0,0 +1,33 @@
...
+    view.request = req
+    assert view.get_context_data()['variation'] == ''
+    assert view.get_template_names() == ['mozorg/book-c.html']
+
+    view = views.VariationTemplateView(template_name_variations=['1'],
+                                       template_name='about.index.html')
+    req = rf.get('/', data={'v': '1'})
+    view.request = req
+    assert 'variation' not in view.get_context_data()
+    assert view.get_template_names() == ['about.index-1.html']
```
pmac: I didn't test that very much because I'm not sure they'll be useful to gether, but I also don't think it's necessary to prevent them from being used together. Since they rely on the same query param I don't think it'd be partiularly useful to use the context when you have a separate template.

bedrock/mozorg/views.py
pmac: Indeed they could. I'll do that.

pmac:
Docs is a good point. I'll try to add them before we merge this. I'll likely forget to do it otherwise.

pmac:
@jpetto docs added and mozorg views moved to urls.py. Should hopefully be ready for final review. Thanks again!

pmac:
Hmm... wondering if it'd be good to also add a facility for specifying a list of allowed locales as well? That seems a common requirement.

jpetto:
> Hmm... wondering if it'd be good to also add a facility for specifying a list of allowed locales as well? That seems a common requirement.

Good idea. I'd say 99% of the time, our tests are restricted to `en-US` or `en-*`. Maybe make `en-US` the default, and allow overriding with a list?

pmac:
@jpetto I made "all locales" the default, since I couldn't figure out a way to allow all locales otherwise. Also I like it that it needs to be explicit about only supporting en-US so we can see it at a glance. I've also updated the docs to reflect this. LMK what you think.

pmac:
I've also just added to this an extra argument to the `switch()` template helper. It will now take a `locales` argument, which should be a list of the locales in which this switch should be active. It will always return `False` for all other locales, and consult the env var for the listed locales. I just noticed that switches very commonly were accompanied by something like `LANG == 'en-US'`. There are also some `LANG.startswith('en')`, but I'm not sure yet how to accomplish this without harming other uses. If you think this is useful, not helpful, or have a suggestion for how it could work better, please let me know.

@jpetto @alexgibson @craigcook

pmac:
Adding those locale groups. Don't merge for now.

pmac:
@jpetto @alexgibson I've added support for "locale groups". It works like this: You can specify a list of locales as part of a switch call. In that list you can use specific locales (e.g. "de" or "pt-BR"), or a locale group name (e.g. "en" or "pt"). We don't support "en", but it will expand to all supported "en" locales (i.e. en-US, en-GB, and en-ZA at present). I think this is nicely flexible.

Note that the locale bits only work in the template function, not the one to be used in views. You'll still have to do your own locale logic in view code, which I think is fine since you have the full power of Python there.

pmac:
Now also added locale group support to the variation view. This is getting pretty good I think :)

alexgibson:
Sorry I only just caught up with this thread and read through the doc changes - it all looks great, should be really useful! 💯

bedrock/settings/base.py
```diff
@@ -182,6 +182,19 @@ def get_dev_languages():
 }
 
 
+def lazy_lang_group():
+    from django.conf import settings
+
+    groups = {}
+    langs = settings.DEV_LANGUAGES if settings.DEV else settings.PROD_LANGUAGES
+    for lang in langs:
+        if '-' in lang:
```
jpetto: Surely an edge case, but what would happen here for `ja` and `ja-JP-mac`? Seems `ja` would never be derived, as sending `ja` in would expand only to `ja-JP-mac`, right?

I can't see this being a practical issue any time soon, but could be a weird bug somewhere down the line.

Typing out loud here, but what would you think about removing groups of one, e.g. `ja-JP-mac`, `fy-NL` and `ga-IE`?

docs/coding.rst
```diff
@@ -111,6 +111,89 @@ Make sure to namespace your templates by putting them in a directory
...
+    ]
+
+Any request to the page in for example French would not use the alternate template even if a
+valid variation were given in the URL.
+
+.. note::
+
+    If you'd like to add this functionality to an existing Class-Based View, there is
+    a mixin that implements this pattern that should work with most views:
+    `bedrock.utils.views.VariationMixin`.
```
jpetto: Would be nice to have a code sample using the mixin for us front-enders.

bedrock/settings/base.py
pmac: Oooo... this is a great catch! I'll look into this.

docs/coding.rst
pmac: I don't think that's necessary really. The mixin is really only for views that are already customized and using a view subclass. I think it's advanced enough to warrant asking a backend dev at a minimum.

pmac:
@jpetto okay. I think we may finally have cracked it. I've added a test for the `LANG_GROUPS` lazy setting as well.

bedrock/base/tests/test_settings.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+@override_settings(DEV=False, PROD_LANGUAGES=('de', 'fr', 'nb-NO', 'ja', 'ja-JP-mac',
+                                              'en-US', 'en-ZA'))
+def test_lang_groups():
+    # should not contain 'nb' and 'ja' group should contain 'ja'
+    assert dict(settings.LANG_GROUPS) == {
+        'ja': ['ja-JP-mac', 'ja'],
+        'en': ['en-US', 'en-ZA'],
+    }
```
jpetto: 👍 👍

jpetto:
Looking great! r+

🚨 Woop woop! 🚨

PR #4515, opened 2016-12-04
Skip broken Fennec link for 'be' until bug 1321262 is resolved

alexgibson:
@pmac r?

No rush to look at this, I know you're traveling! (It's 4.30am and I'm awake, so thought I might as well try and do something productive).

P.S - I ran this locally and it passes.


alexgibson:
@pmac - thanks for the suggestion, updated 👍

PR #4518, opened 2016-12-14
Fixes bug 1323417 - Update the number of languages on firefox/all to 90

TheoChevalier:
## Description
We now have more locales, let’s be proud of it :) We’re still not localizing this page, so I’m not adding a l10n tag.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1323417

## Testing
Description & Tweet on https://www.mozilla.org/en-US/firefox/all/ are updated.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/all.html
+++ b/bedrock/firefox/templates/firefox/all.html
@@ -29,7 +29,7 @@
   {%- elif channel == 'esr' -%}
     {{ _('Firefox ESR is intended for groups who deploy and maintain the desktop environment in large organizations. <a href="%s">Learn more.</a>')|format(url('firefox.organizations.organizations')) }}
   {%- else -%}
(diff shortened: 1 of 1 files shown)


flodolo:
(CC @kyoshino for Japanese)

kyoshino:
I just checked [all.lang](https://github.com/mozilla-l10n/www.mozilla.org/blob/master/ja/firefox/all.lang) and found a placeholder `%s` which is apparently not used at this time. I guess the number should be retrieved from the [settings](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L538).

flodolo:
Given that languages use different plural forms based on the number, better not to use a placeholder in case we start localizing this page.

kyoshino:
And, yeah, there's only the Japanese l10n for this page, I'm happy to update the translation shortly.

PR #4519, opened 2016-12-14
[bug 1319244] Test developer and general search

craigcook:
## Description
This adds developer and general custom search to several pages:

Developer search
* /firefox/developer?v=b
* /firefox/channel/desktop?v=b
* /technology

General search
* /contribute
* /404

It also makes some updates to the activist search, moving the code snippet from an include to a macro so it's more reusable and refactors the JS a bit so it's also more reusable.

New switches added:
* `mozorg-search-general`
* `mozorg-search-developer`

Also see https://bugzilla.mozilla.org/show_bug.cgi?id=1323315

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1319244

## Testing
* Search should only be visible in English locales.
* Ensure search appears and functions as expected on all target pages (and that I didn't break the earlier activist search on /about etc).
* We're using Traffic Cop for a 50/50 split on both /firefox/developer and /firefox/channel/desktop so it should only be visible on those pages with a `?v=b` param.

## Checklist
- [ ] Requires l10n changes.


craigcook:
@bensternthal r? when you get a chance?

bensternthal:
Tested locally and everything works as expected. Glanced at the code and it seems good to me. Might be worth having @pmac check the python / django parts.

R+

bedrock/firefox/templates/firefox/channel/desktop.html
```diff
@@ -2,17 +2,31 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% from 'macros.html' import google_custom_search with context %}
```
pmac: The `extends` call should be first in the template. While this works it's best to move the `from` call here to below the `extends`.

bedrock/base/templates/404.html
```diff
@@ -3,6 +3,8 @@
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 {% set_lang_files "mozorg/404" %}
 
+{% from 'macros.html' import google_custom_search with context %}
```
pmac: This should be below the `extends` call. The `set_lang_files` above should be as well really. I know it's not your change, but we should clean that up now if you don't mind.

bedrock/base/templates/macros.html
```diff
@@ -474,3 +474,31 @@ <h2 class="thank-you hidden">{{ _('Your download link was sent.') }}</h2>
...
+    search_class: A class added to the bar for use as a style hook to customize the presentation.
+
+    term_list:    A comma separated list of suggested terms displayed in a loop via JS.
+
+    default_term: A static suggested term to avoid an empty label if JS fails. This should
+                  usually be the same as the last term in the list (last to avoid clumsy repetition
+                  when JS kicks in to start the cycle).
+
+  Search engines -
+    General: 014783244707853607354:ifn1iyzxd1a
```
pmac: Seems like a good opportunity to have these IDs in settings. Would be a bit more of a pain to go around the code changing them in every use location instead of just in settings. I'm thinking you could have something like this in settings:

```
GOOG_CUSTOM_SEARCH_TYPES = {
    'general': '014783244707853607354:ifn1iyzxd1a',
    'developer': '014783244707853607354:s-9zxgjd2o0',
    'activist': '014783244707853607354:bbpl3emdsii',
}
```

Then just have each call to the macro pass in "general", "developer", or "activist" and get the real value out of the setting: `{{ settings.GOOG_CUSTOM_SEARCH_TYPES[search_type] }}`. That would clean up the calls to the macro quite a bit and make it more obvious just looking at the template which one is being used.

bedrock/firefox/templates/firefox/channel/desktop.html
```diff
@@ -86,3 +100,9 @@
   </footer>
 </section>
 {% endblock %}
+
+{% block js %}
+  {% if LANG.startswith('en-') and switch('mozorg-search-developer') %}
```
pmac: This `if` statement and the one from the search block above (line 25) don't match. Seems like they should match.

bedrock/firefox/templates/firefox/developer.html
```diff
@@ -2,6 +2,8 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% from 'macros.html' import google_custom_search with context %}
```
pmac: Also move this one below `extends`.

bedrock/firefox/templates/firefox/developer.html
```diff
@@ -26,8 +28,18 @@
   {% stylesheet 'firefox_developer' %}
 {% endblock %}
 
+{% block experiments %}
+  {% if LANG.startswith('en-') and switch('experiment-search-developer') %}
```
pmac: More non-matching `if` blocks in this file.

bedrock/firefox/views.py
```diff
@@ -499,6 +499,26 @@ def sync(request):
     return l10n_utils.render(request, 'firefox/sync.html', {'version': version})
 
 
+def channel_desktop(request):
```
pmac: I've got a change coming in pr #4514 that will remove the need for these, but I can fix them if this lands first.

bedrock/foundation/templates/foundation/index.html
```diff
@@ -2,6 +2,8 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% from 'macros.html' import google_custom_search with context %}
```
pmac: Another to move below `extends`.

bedrock/base/templates/404.html
```diff
@@ -13,6 +15,18 @@
...
 
+{% block js %}
+  {% if LANG.startswith('en-') and switch('mozorg-search-general') %}
+    {% javascript 'custom-search' %}
+  {% endif %}
+{% endblock %}
+
+{% block site_header_unwrapped %}
+  {% if LANG.startswith('en-') and switch('mozorg-search-general') %}
+    {{ google_custom_search(settings.GOOG_CUSTOM_SEARCH_TYPES['general'], 'moz-search-general', _('Internet Health'), _('Volunteer, Events, Add-ons, Firefox, Internet Health')) }}
```
pmac: Hmm... this works, but I was more thinking you'd just pass `'general'` here and expand it in the macro code.

craigcook: oh, yeah, I suppose that would be better... but this is probably temporary anyway, probably to be removed in a few weeks.

bedrock/firefox/urls.py
```diff
@@ -30,15 +30,15 @@
...
+    url(r'^firefox/channel/desktop', views.channel_desktop, name='firefox.channel.desktop'),
     page('firefox/channel/android', 'firefox/channel/android.html'),
     page('firefox/channel/ios', 'firefox/channel/ios.html'),
     page('firefox/desktop', 'firefox/desktop/index.html'),
     page('firefox/desktop/fast', 'firefox/desktop/fast.html'),
     page('firefox/desktop/customize', 'firefox/desktop/customize.html'),
     page('firefox/desktop/tips', 'firefox/desktop/tips.html'),
     page('firefox/desktop/trust', 'firefox/desktop/trust.html'),
-    page('firefox/developer', 'firefox/developer.html'),
+    url(r'^firefox/developer', views.developer, name='firefox.developer'),
```
craigcook: @pmac Looks like this URL has broken the "/firefox/developers/notes" redirect for release notes, but I'm not sure what the correct regex should be here.

PR #4520, opened 2016-12-15
[fix bug 1323707] Remove conditional logic that hides download buttons from Firefox users

alexgibson:
## Description
- Simplifies download button logic on a few remaining pages (we already covered the home page and main download page in different bugs):
  - `/firefox/desktop/`
  - `/firefox/desktop/tips/`
  - `/firefox/releasenotes/`

I decided to leave `/products/` as is, since the download bar is more of a conditional UI, and the additional changes to `/desktop` already make it easier for Firefox pre-release users to download.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1323707

## Testing
- Test to make sure no visual regressions, download buttons show correctly etc.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/releases/release-notes.html
+++ b/bedrock/firefox/templates/firefox/releases/release-notes.html
@@ -267,24 +267,16 @@
           <a rel="external" href="{{ firefox_url('desktop', 'all', channel=channel_name|lower()) }}">{{ _('All Firefox downloads') }}</a>
         </div>
         <div class="download-buttons">
-          <div class="firefox-latest-cta">
-            <p class="version-message">Congrats! You’re using the latest version of Firefox.</p>
-          </div>
-          <div class="non-firefox-cta">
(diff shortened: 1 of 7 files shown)


alexgibson:
> What do you think about the display logic on the Private Browsing and Sync pages? Should we clean those up as well?

I think both those pages are a bit more complicated than it warrants changing here. As those pages are only really built for a single CTA at a time, showing both CTA buttons and download buttons together would likely require some additional design/UX work so as not to adversely effect conversions. This is something we could revisit when we have more time/resources perhaps (certainly for the Sync page).

> Also, it's a previous regression, but would you mind fixing the color of the privacy link under the download button in the footer of /desktop/tips/? It's currently white/invisible.

Nice catch, will update.

jpetto:
Agree we can wait on the Private Browsing & Sync pages. Those can be round 2. 😉

PR #4521, opened 2016-12-15
[bug 1316375] Set a cookie for the EOY takeover

craigcook:
## Description
We were previously using sessionStorage to disable the takeover once you'd seen it, but it meant you'd get the takeover again with any new session. It was decided that was potentially annoying to repeat visitors, so this switches to a permanent cookie with a 7 day expiration.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1316375

## Testing
Verify that the cookie is being set, is set to expire in 7 days, and that the takeover isn't shown again when the cookie is present. The takeover should also not be shown if cookies are disabled (otherwise you'd see it on every refresh).

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/js/mozorg/home/takeover-2016.js
+++ b/media/js/mozorg/home/takeover-2016.js
@@ -37,14 +37,28 @@
         });
     }
 
-    // sets the mozorg.takeover state to true indicating that the user has
-    // seen the takeover modal during their current session.
-    function storeSession() {
+    // Check for cookie support (stolen from Modernizr)
+    function testCookies() {
         try {
-            sessionStorage.setItem('mozorg.takeover.seen', 'true');
-        } catch (ex) {
-            // yum, errors taste nice ;)
+            // Create cookie
+            document.cookie = 'cookietest=1';
+            var ret = document.cookie.indexOf('cookietest=') !== -1;
+            // Delete cookie
+            document.cookie = 'cookietest=1; expires=Thu, 01-Jan-1970 00:00:01 GMT';
+            return ret;
         }
+        catch (e) {
+            return false;
+        }
+    }
+
+    // Sets a cookie so you won't see the takeover again for seven days
+    function seenTakover() {
+        var date = new Date();
+        date.setTime(date.getTime() + (7 * 24 * 60 * 60 * 1000));
+        var expires = date.toUTCString();
+
+        Mozilla.Cookies.setItem('eoy-takeover', 'true', expires);
     }
 
     // Shows and hides the takeover modal
@@ -67,18 +81,19 @@
         $options.removeAttr('checked');
     }
 
-    // Shows the take over modal, sets the required entry in
-    // session storage and adds an `esc` key handler to close
-    // the modal.
+    // Shows the take over modal, sets the seven day cookie
+    // and adds an `esc` key handler to close the modal.
     function initTakeOver() {
         // format the currency in browsers that support it
         if (intlNumberFormatSupported) {
             formatCurrency();
         }
+
         // show the takeover
         toggleModal();
-        // set mozorg.takeover.seen to true
-        storeSession();
+
+        // set the cookie
+        seenTakover();
 
         // trap key events, listening for escape.
         // we only want to do this here so, we do not attach the event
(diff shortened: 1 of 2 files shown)


craigcook:
@alexgibson r?

media/js/mozorg/home/takeover-2016.js
```diff
@@ -37,13 +37,16 @@
...
     }
 
-    // sets the mozorg.takeover state to true indicating that the user has
-    // seen the takeover modal during their current session.
-    function storeSession() {
+    // Sets a cookie so you won't see the takeover again for seven days
+    function seenTakover() {
+        var date = new Date();
+        date.setTime(date.getTime() + (7 * 24 * 60 * 60 * 1000));
+        var expires = date.toGMTString();
```
alexgibson: According to MDN, [`toGMTString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toGMTString) is deprecated. I guess [`toUTCString()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toUTCString) is the way to go?

media/js/mozorg/home/takeover-2016.js
```diff
@@ -67,18 +70,21 @@
...
         if (intlNumberFormatSupported) {
             formatCurrency();
         }
-        // show the takeover
-        toggleModal();
-        // set mozorg.takeover.seen to true
-        storeSession();
+
+        // Don't show the takeover if cookies are disabled or not supported
+        if (document.cookie) {
```
alexgibson: Rather than do this check here would it be better to move it to where the call to `initTakoOver()` is made? e.g.

```
if (document.cookie && !Mozilla.Cookies.hasItem('eoy-takeover')) {

}
```

alexgibson: Hmm, actually I'm not sure if `document.cookie` is enough here, would this return a false positive if no cookies exist (irrespective of being enabled or not). Perhaps simply `Mozilla.Cookies.hasItem('eoy-takeover')` would be enough?

alexgibson: Fwiw, this is the Modernizr feature detect for cookies if you wanted to add a helper function: https://github.com/Modernizr/Modernizr/blob/master/feature-detects/cookies.js

craigcook: I wanted to check for cookie support because if a browser doesn't support cookies (or has them blocked/disabled) it would always see the takeover... so no support for cookies, no takeover. The `!Mozilla.Cookie.hasItem('eoy-takeover')` checks for if that specific cookie is absent, but doesn't know if it's absent because the browser doesn't support cookies at all or just hasn't seen the takeover yet.

So I think we do want both checks, but maybe `document.cookie` isn't the best way to do that? It looks like Modernizr is testing for support by just setting a cookie and then detecting it, I'm sure we could rework this slightly to do the same thing. I like the idea of baking that into our cookie helper (heh, baking) but perhaps not in this PR.

media/js/mozorg/home/takeover-2016.js
```diff
@@ -37,13 +37,31 @@
...
-    // sets the mozorg.takeover state to true indicating that the user has
-    // seen the takeover modal during their current session.
-    function storeSession() {
+    // Check for cookie support (stolen from Modernizr)
+    function testCookies() {
         try {
-            sessionStorage.setItem('mozorg.takeover.seen', 'true');
+            // Create cookie
+            document.cookie = 'cookietest=1';
+            var ret = document.cookie.indexOf('cookietest=') != -1;
```
alexgibson: ESLint says this should be `!==`

media/js/mozorg/home/takeover-2016.js
```diff
@@ -37,13 +37,31 @@
...
+        }
+    }
+
+    // Sets a cookie so you won't see the takeover again for seven days
+    function seenTakover() {
+        var date = new Date();
+        date.setTime(date.getTime() + (7 * 24 * 60 * 60 * 1000));
+        var expires = date.toUTCString();
+
+        try {
```
alexgibson: Now we're using cookies and a feature detect, there should be no need for this `try/catch`, right?

media/js/mozorg/home/takeover-2016.js
```diff
@@ -133,9 +152,9 @@
     // Modal is waffled so first ensure it exists.
     if (arrayIndexOfSupported && $takeoverModal.length > 0) {
         try {
-            // only show the takeover modal if the user has not already seen it
-            // during this session.
-            if (sessionStorage.getItem('mozorg.takeover.seen') !== 'true') {
+            // only show the takeover modal if the user has not
+            // already seen it (has no cookie).
+            if (testCookies && !Mozilla.Cookies.hasItem('eoy-takeover')) {
```
alexgibson: Should this be `testCookies()` with parenthesis?

PR #4522, opened 2016-12-15
Bug 1317260: Update CA Cert Policy document

pmac:
This updates the file based on the markdown from the repo and documents updating it in future in the template itself. It also deletes the sub pages, and adds tests and redirects for the same.


gerv:
This PR currently seems to not remove the "inclusion" document, and doesn't update the index.html document to contain the new policy?

gerv:
@pmac Did you forget to push an updated version of your branch?

pmac:
The `index.html` file is for the https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/ page. The policy itself lives at https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/policy/ and is in the `policy.html` file. It looks to me like the `inclusion.html` file was removed. It didn't appear as though changes to the `index.html` file were required as it only linked to the policy (among other things). Do we no longer need a separate "certs" landing page?

gerv:
> The policy itself lives at https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/policy/ and is in the policy.html file

Hmm. It seems github doesn't load the diff for every file; there was a "Load Diff" button that I had to press to see this one. Apologies for misreading everything.

I think we probably want the policy at the URL https://www.mozilla.org/about/governance/policies/security-group/certs/policy/ - if that's what's going to happen, great. Yes, we still need https://www.mozilla.org/en-US/about/governance/policies/security-group/certs/ .

Ship it! :-)

pmac:
Great! Thanks for the review! I just want someone on the team to give the new code a quick look and we should be able to get this to prod today.

bedrock/mozorg/templates/mozorg/about/governance/policies/security/certs/policy.html
```diff
@@ -8,31 +8,778 @@
...
+    other software modifications, by making such changes a distributor may affect
+    its ability to use Mozilla trademarks in connection with its versions of the
+    software; see the Mozilla trademark policy for more information.</p>
+  <p>Please contact Mozilla at
+    <a href="mailto:certificates@mozilla.org">certificates@mozilla.org</a> for more
+    information about this policy and answers to related questions.</p>
+  <p>We reserve the right to change this policy in the future. We will do so only
+    after consulting with the public Mozilla community, in order to ensure that
+    all views are taken into account.</p>
+  <p><a name="inclusion"></a></p>
```
craigcook: Nit: named anchors should be avoided these days. It'll still work in browsers but it's not valid in HTML5 and can present accessibility problems. This should be an ID on the `h2` below to provide an anchor for linking: `<h2 id="inclusion">`

bedrock/mozorg/templates/mozorg/about/governance/policies/security/certs/policy.html
```diff
@@ -8,31 +8,778 @@
...
+        behalf and make decisions regarding all matters relating to CA
+        certificates included in our products. CAs or others objecting to a
+        particular decision may appeal to the <a
+            href="https://wiki.mozilla.org/Modules/Activities#Governance">Mozilla governance module
+          owner or
+          peer(s)</a>,
+        who will make a final decision.</p>
+    </li>
+  </ol>
+  <p><a name="maintenance"></a></p>
```
craigcook: Another named anchor that should be an ID on the heading.

bedrock/mozorg/templates/mozorg/about/governance/policies/security/certs/policy.html
```diff
@@ -8,31 +8,778 @@
...
+        <li>a representative of the CA or a representative of Mozilla may
+          request that a root certificate be removed by submitting a bug
+          report into the mozilla.org Bugzilla system, as described in the
+          <a href="https://wiki.mozilla.org/CA:Root_Change_Process">Root Change
+            Process</a>.
+        </li>
+      </ul>
+    </li>
+  </ol>
+  <p><a name="enforcement"></a></p>
```
craigcook: Here too.

bedrock/mozorg/templates/mozorg/about/governance/policies/security/certs/policy.html
pmac: Agreed, but this is generated by markdown, and that's how it is in the original. I can update here, but it'll switch back on update unless we adjust the original.

pmac: Also markdown doesn't allow for adding attributes to the HTML I don't think, so these extra `<a>` are in the raw markdown like that.

gerv:
@craigcook I suspect those are the output of the automated markdown converter. Are you going to require us to patch it before checking in this patch? Or do you want a manual post-processing step? Or are you willing to let it go? :-)

gerv:
Actually no, I lie, you are right, they are there in the Markdown. Apologies.

Can we roll with it this time, and I'll fix the markdown for the next release?

pmac:
Updated. Thanks again @craigcook.

gerv:
Or you can manually fix it, and I'll make sure it's fixed in the markdown for 2.4. That works too :-)

craigcook:
Works for me!

pmac:
@gerv I've manually fixed it for now, and we can figure out how best to do this in the source .md file next time. And like you said, it'll happen rarely, so it's not too onerous if we just have to fix the anchors when we update 🌴

PR #4524, opened 2016-12-15
[bug 1319244] Fix firefox/developer URL

craigcook:
There are redirects for /firefox/developer/notes and /firefox/developer/system-requirements that were broken by the new view. I think this tweak to the regex fixes it, at least it does for me locally (the redirects work in a browser and the tests pass).


bedrock/firefox/urls.py
```diff
@@ -38,7 +38,7 @@
     page('firefox/desktop/customize', 'firefox/desktop/customize.html'),
     page('firefox/desktop/tips', 'firefox/desktop/tips.html'),
     page('firefox/desktop/trust', 'firefox/desktop/trust.html'),
-    url(r'^firefox/developer', views.developer, name='firefox.developer'),
+    url(r'^firefox/developer/?$', views.developer, name='firefox.developer'),
```
pmac: Very nearly spot on. This works fine, but really you don't need the `?` in the regex. Django takes care of redirecting to the URL with the appended `/` if the user leaves it off, and that is good so that the page isn't accessible via both URLs (with and without the `/`). What caused the issue in the first place is that the regex was open ended. Leaving it at just `^firefox/developer` means "match any URL that starts with this". The fix is adding `/$` to the end, which switches it to mean that it should only match that full string and nothing more. So it was also apparently matching e.g. `firefox/developer/notes/` and overriding other views that start with `firefox/developer/...`. The `page()` helper that was used before ends up with a regex of `r'firefox/developer/$'`.

PR #4525, opened 2016-12-16
Bug  1320080 newsletter icon in thai page cut off

alisha17:

### DIFF
--- a/static/css/mozorg/home/home.scss
+++ b/static/css/mozorg/home/home.scss
@@ -800,7 +800,7 @@ main {
                 background-position: left top;
                 float: left;
                 margin-left: -380px;
-                padding:9px 0 0 140px;
+                padding:10px 0 0 140px;
                 width: 350px;
                 height: 115px;
                 h3,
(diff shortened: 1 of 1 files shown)


alexgibson:
It looks like you've accidentally added a whole new file here, instead of just the changes needed to fix the bug?

alexgibson:
Ah, it seems like you've manually added this file to the `/static` directory. You should be editing the file that lives in `/media/css/mozorg/home/home.scss`.

alexgibson:
I'd also suggest that rather than fixing this just for the home page, as more robust solution could be to apply the fix to the [newsletter component](https://github.com/mozilla/bedrock/blob/master/media/css/pebbles/components/_newsletter.scss), that is shared across many pages.

alisha17:
Okay. I will fix it as suggested.

alexgibson:
Thanks! Feel free to ask here or in the bug if you need any help, or drop into #www on IRC

alexgibson:
Closing in favor of #4539

PR #4526, opened 2016-12-16
Fix bug 1315082: Store git repo state in the db

pmac:
This reverts commit d962ef4b771f8ee13ea071143a237f5a8c54e6e6 which itself reverted an earlier attempt at this but which caused some tests to fail. Let's try it again.


pmac:
This is the reincarnation of #4452. Tests pass now. Hopefully we can get it in again soon.

PR #4529, opened 2016-12-18
Fix Bug 1324305 - Firefox Releases page does not list 50.1.0

kyoshino:
## Description

Firefox 50.1.0, the latest stability release, is not listed on the [release notes index page](https://www.mozilla.org/en-US/firefox/releases/) because the relevant code is not handling the irregular versioning properly. This PR is a workaround for the issue.

## Bugzilla link

[Bug 1324305](https://bugzilla.mozilla.org/show_bug.cgi?id=1324305)

## Testing

Firefox 50.1.0 should be displayed on the index page. No other changes.

## Checklist
- [ ] Requires l10n changes.


bedrock/releasenotes/views.py
```diff
@@ -211,6 +211,9 @@ def releases_index(request, product):
     if product == 'Firefox':
         major_releases = firefox_desktop.firefox_history_major_releases
         minor_releases = firefox_desktop.firefox_history_stability_releases
+        # Firefox 50.1.0 is not an ESR version but should be fallen under 50.0,
+        # so put it in the list as an exception (Bug 1324305).
+        esr_major_versions.append(50)
```
jgmize: Why not just eliminate esr_major_versions entirely and always use the same logic, something along the lines of this: https://github.com/mozilla/bedrock/commit/def6e1239b7b2a0d09550f5fb231afd2961ab953 (untested, I just put this together to clarify what I meant)

jpetto: Agree it would be preferable to provide a fix for potential future point releases as well (e.g. 54.1.0).

bedrock/releasenotes/tests/data/firefox_history_major_releases.json
```diff
@@ -1,42 +1 @@
...
-}
+{"1.0": "2004-11-09", "1.5": "2005-11-29", "2.0": "2006-10-24", "3.0": "2008-06-17", "3.5": "2009-06-30", "3.6": "2010-01-21", "4.0": "2011-03-22", "5.0": "2011-06-21", "6.0": "2011-08-16", "7.0": "2011-09-27", "8.0": "2011-11-08", "9.0": "2011-12-20", "10.0": "2012-01-31", "11.0": "2012-03-13", "12.0": "2012-04-24", "13.0": "2012-06-05", "14.0.1": "2012-07-17", "15.0": "2012-08-28", "16.0": "2012-10-09", "17.0": "2012-11-20", "18.0": "2013-01-08", "19.0": "2013-02-19", "20.0": "2013-04-02", "21.0": "2013-05-14", "22.0": "2013-06-25", "23.0": "2013-08-06", "24.0": "2013-09-17", "25.0": "2013-10-29", "26.0": "2013-12-10", "27.0": "2014-02-04", "28.0": "2014-03-18", "29.0": "2014-04-29", "30.0": "2014-06-10", "31.0": "2014-07-22", "32.0": "2014-09-02", "33.0": "2014-10-14", "33.1": "2014-11-10", "34.0": "2014-12-01", "35.0": "2015-01-13", "36.0": "2015-02-24", "37.0": "2015-03-31", "38.0": "2015-05-12", "39.0": "2015-07-02", "40.0": "2015-08-11", "41.0": "2015-09-22", "42.0": "2015-11-03", "43.0": "2015-12-15", "44.0": "2016-01-26", "45.0": "2016-03-08", "46.0": "2016-04-26", "47.0": "2016-06-07", "48.0": "2016-08-02", "49.0": "2016-09-20", "50.0": "2016-11-15"}
```
jpetto: A nit-picky request, but if you have time, could you pretty print the new JSON? Just makes it easier to scan.

bedrock/releasenotes/views.py
kyoshino: Agree that the logic should be future-proof. There are some glitches on older versions with @jgmize's patch, so I'm figuring out the best way.

kyoshino:
Now I take a straightforward approach that iterates over the minor versions rather than the tricky pattern matching based on the major versions. It should be future-proof.

Will investigate why the tests are failing. Those are passing locally.

bedrock/releasenotes/views.py
```diff
@@ -203,32 +203,45 @@ def latest_sysreq(request, channel, product):
     return HttpResponseRedirect('/' + '/'.join(path) + '/')
 
 
+def get_major_version(version):
+    """
+    Return a numeric major version number from the given version string. For
+    example, 50.0.2 => 52.0, 50.1.0 => 50.1.
```
jgmize: I think `50.0.2 => 52.0` should be replaced with `50.0.2 => 50.0`.

kyoshino: Oops, will fix the comment 👍

kyoshino:
Hmm, I still cannot solve the test failures. Apparently the patched `thunderbird_desktop._storage.data` is not restored after the tests in `bedrock/releasenotes/tests/test_base.py` is finished. From my quick test, the tests in `bedrock/bedrock/thunderbird/tests/test_base.py` are unexpectedly using the data from `bedrock/releasenotes/tests/data/` and therefore all the tests under `bedrock/bedrock/thunderbird/tests/` are failing. So weird 😕

alexgibson:
ping @kyoshino - any updates here?

alexgibson:
Closing as Firefox 50.0.1 was released some time ago, and this branch is now stale. Feel free to reopen if this is something that we still want to fix. Thanks @kyoshino

kyoshino:
The patch itself worked, but I was struggling with the fragile tests. Will work on this again when I get time.

PR #4535, opened 2016-12-20
[fix bug 1323841] Add Ashley Boyd to leadership page

craigcook:
## Description
Adds a new MoFo VP behind the switch `leadership-boyd-announce`. We'll be formally announcing on the morning of 3 January so we'll need to flip the switch at that time. Alas, this is apparently the best picture we have for her right now.

This update also adds permalinks to each bio to make them easier to link to (it's just the page URL with the corresponding hash). I wanted a link icon so ended up updating the icons font, which then meant I needed to update a few other spots that use it. Also, WOFF2!

The leadership page is not localized.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1323841


jpetto:
LGTM! r+ 👦D

PR #4536, opened 2016-12-20
Bug 1008390: Add query param to all locale redirects

pmac:
This may well break a bunch of integration tests. Just trying this in a demo so far.


pmac:
I don't think this is going to happen in the way it's done in this PR. Closing for now.

PR #4538, opened 2016-12-20
Remove listing of #mv IRC channel to prevent support requests.

jdashg:
Someone needs to fix the UX flow so it's obvious that the proper place for support questions is #firefox.

Until that time, we can stem the bleeding by removing this listing, since this is something that should be on Mana, not our public "contact us" for the office.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/contact/spaces/mountain-view.html
+++ b/bedrock/mozorg/templates/mozorg/contact/spaces/mountain-view.html
@@ -23,7 +23,6 @@
         </p>
 
         <ul class="extra">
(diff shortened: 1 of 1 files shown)


pmac:
Seems like something that should be done to all or none of the spaces. What do you think @craigcook @alexgibson @jpetto?

alexgibson:
This seems like something we should fix properly, but I'd like to understand the route people are going to make all the effort to drop into IRC, if none of the information here can satisfy them: https://www.mozilla.org/en-US/contact/

No objections to removing the IRC links either way though.

craigcook:
I agree, I'll remove all the IRC channels in a follow-up PR.

craigcook:
Replaced with #4541

PR #4539, opened 2016-12-21
bug1320080-Newletter icon in thai page fixed

alisha17:
Bug 1320080

The newsletter icon in thai homepage was cut off. By adding the height in the css, it was fixed.

### DIFF
--- a/media/css/newsletter/moznewsletter-subscribe.less
+++ b/media/css/newsletter/moznewsletter-subscribe.less
@@ -48,6 +48,7 @@
         padding: 10px 0 20px 180px;
         text-align: left;
         text-shadow: none;
+        height: 135px;
 
         h3 {
             .font-size(40px);
--- a/media/css/newsletter/moznewsletter-subscribe.less
+++ b/media/css/newsletter/moznewsletter-subscribe.less
@@ -48,7 +48,7 @@
         padding: 10px 0 20px 180px;
         text-align: left;
         text-shadow: none;
-        height: 135px;
+        min-height: 135px;
 
         h3 {
             .font-size(40px);
--- a/media/css/mozorg/home/home.scss
+++ b/media/css/mozorg/home/home.scss
@@ -756,7 +756,6 @@ main {
         @include background-size(100px, 100px);
         background-position: center top;
         margin-top: 0;
-        min-height: 0;
         padding-left: 0;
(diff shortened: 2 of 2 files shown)


flodolo:
Looks like you already have an open pr for this? #4525 
You can keep using that, just adding new commits to that branch.

alisha17:
But that now I have created a new one. What should I do? :> 
Should I close the previous pull request ?

flodolo:
Let's wait for @alexgibson to see what he suggests. There's usually value in using the existing one, to keep all conversation in one place, but he might be OK with using a new PR.

RCMainak:
I believe the standard is one commit / PR per bug.
@alisha17 If you make some mistake in your previous PR it's always wise to fix that PR rather then creating a new PR. That way you'll get to know about git more which is really fun thing to do. :smile: 

 All though Alex or other maintainers might have better advice.

alisha17:
My code was not getting committed in forked repository. I tried in many ways, but was unsuccessful. So, I removed that repo and again forked it. Later, on again checking clearly, I saw I was updating the file which is in .gitignore. I know it's my bad. Apologies! :>

RCMainak:
@alisha17 I would suggest you to get on #www channel on IRC. People there can help you with any problem you're facing related to bedrock & GitHub. I faced the same problems, people there helped me a lot & I learned a lot from them.

media/css/newsletter/moznewsletter-subscribe.less
```diff
@@ -48,6 +48,7 @@
         padding: 10px 0 20px 180px;
         text-align: left;
         text-shadow: none;
+        height: 135px;
```
alexgibson: While this fixes the bug for short strings, it breaks things for really long strings as it gives the element a fixed height. Please use `min-height: 135px;` instead. You may also need to [remove this line in home.scss](https://github.com/mozilla/bedrock/blob/master/media/css/mozorg/home/home.scss#L759) so `min-height` does not get overwritten.

Please just add another commit to your branch with this change, and github should pick up the changes here.

alexgibson:
I squashed your two commits and merged to master manually in https://github.com/mozilla/bedrock/commit/51849f019941b0d302178dd183af305ff3f8bc4d

Thanks for the fix, @alisha17! 💯 🎉

PR #4540, opened 2016-12-21
[fix bug 1323829] Update non-Fx header copy on /firefox/products/

alexgibson:
## Description
Updates copy on /products to be consistent with the newer download page messaging.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1323829

## Checklist
- [x] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/family/index.html
+++ b/bedrock/firefox/templates/firefox/family/index.html
@@ -31,9 +31,13 @@
         {{ download_firefox(alt_copy=_('Download'), dom_id='download-button-old-fx') }}
       </div>
       <div class="message" id="dlbar-nonfx">
-        {{ _('Choose Independent.') }}
-        {{ _('Choose Firefox.') }}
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/family/index.html
```diff
@@ -4,6 +4,8 @@
 
 {% extends "firefox/base-resp.html" %}
 
+{% add_lang_files "firefox/new/horizon" %}
```
alexgibson: @flodolo including the lag file here from the download page for "Browse Freely". Assuming this is the easiest (best?) solution.

flodolo: I wouldn't do this, file comes with baggage, including activation status and tags.

`Download Firefox` is available in download_button.lang (shared), so we're good there. 

The only new string is `Browse Freely`, we should just add the new string here behind a tag. We can import it from other files for most languages, but using the tag would prevent us from showing the string in English.

PR #4544, opened 2016-12-23
Switch to using a Jenkins Multibranch Pipeline for deploy

pmac:
The new Jenkins Multibranch Pipeline will now start the bedrock build jobs instead of having CircleCI do it. This wil mean having to push to a "prod" branch to trigger a prod deployment, but it should be much better and faster than waiting for Circle.

### DIFF
--- a/bin/tag-release.sh
+++ b/bin/tag-release.sh
@@ -14,4 +14,5 @@ done
 echo "tagged $tag_value"
 if [[ "$1" == "--push" ]]; then
   git push "$moz_git_remote" "$tag_value"
+  git push "$moz_git_remote" master:prod
 fi
--- a/docker/jenkins/check_if_tag.sh
+++ b/docker/jenkins/check_if_tag.sh
@@ -1,10 +1,10 @@
 #!/bin/bash
 
 # Used to trigger downstream Jenkins jobs
-TRIGGER_FILE=.commit_is_tag
+TRIGGER_FILE=".commit_is_tag"
 rm -rf $TRIGGER_FILE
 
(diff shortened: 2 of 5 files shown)


pmac:
After this I plan to port the demo builds into the jenkinsfile itself. Then get going on the rest.

bin/tag-release.sh
```diff
@@ -14,4 +14,5 @@ done
 echo "tagged $tag_value"
 if [[ "$1" == "--push" ]]; then
   git push "$moz_git_remote" "$tag_value"
+  git push "$moz_git_remote" master:prod
```
pmac: This assumes that you've tagged the HEAD commit of master, which is what should have happened. Something is wrong if this is wrong.

docker/jenkins/check_if_tag.sh
```diff
@@ -1,10 +1,10 @@
...
 
 # Used to trigger downstream Jenkins jobs
-TRIGGER_FILE=.commit_is_tag
+TRIGGER_FILE=".commit_is_tag"
 rm -rf $TRIGGER_FILE
 
-if git describe --tags --exact-match $GIT_COMMIT 2> /dev/null;
-then
+TAG=$(git describe --tags --exact-match $GIT_COMMIT 2> /dev/null)
+if [[ -n "$TAG" ]] && [[ "$TAG" =~ ^[0-9]{4}-[0-9]{2}-[0-9]{2}(\.[0-9])?$ ]]; then
```
pmac: Taken from the [standup project](https://github.com/mozilla/standup/blob/master/bin/deploy.sh#L11).

PR #4550, opened 2016-12-28
[fix bug 1325198] Test /all links on /new, part 2.

jpetto:
## Description

2nd round of testing for alternate download links on `/firefox/new/`. `en-US` only.

**Version 1:** - Link added underneath *Firefox Privacy* link that opens up a modal (winner from last test).

**Version 2:** - Majority of content from modal in v1, displayed directly on the page (underneath primary CTA).

I believe we want to run this test ASAP, but should confirm with other teams that no other upcoming experiments (i.e. Optimizely/funnelcake) would conflict.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1325198

## Testing

Ensure UI works across browsers (SVGs wont be displayed on old IE). Make sure direct download links work, and that experiment works as intended.

## Checklist


craigcook:
Some layout issues when the refresh button is shown and the download button container floats left. I'm not sure what circumstances show the refresh button. I got it locally but don't see it on demo, but maybe this scenario isn't possible in production?

fxnew-v2-refresh

craigcook:
I definitely get the refresh button in production in my current version. If you have trouble triggering it locally you can just manually add the `show-refresh` class to the html element.

media/css/firefox/new/all-test-variations-pt2.less
```diff
@@ -0,0 +1,250 @@
...
+}
+
+#test-direct-downloads {
+    padding: 10px 30px 0;
+    text-align: left;
+
+    li {
+        .border-box();
+        margin-bottom: 10px;
+        width: 50%;
```
craigcook: I think these may also need to float left instead of being inline-block. In smaller viewports "Windows (XP/Vista)" wraps to two lines and "OSX" is misaligned. I think floating the list items lets the text align top but there may be another fix too. 

fxnew-v2-listalign

alexgibson:
> I definitely get the refresh button in production in my current version. If you have trouble triggering it locally you can just manually add the show-refresh class to the html element.

You need to enable UITour in order to see the refresh button, and then view the page using the current release version of Firefox (may also need to update product details locally)

media/js/firefox/new/all-test-variations-pt2.js
```diff
@@ -0,0 +1,127 @@
...
+            newOsName = 'Linux 32-bit';
+            $li.attr('data-sort', 2);
+            break;
+        case 'Windows 64-bit':
+            $li.attr('data-sort', 3);
+            break;
+        case 'Linux 64-bit':
+            $li.attr('data-sort', 4);
+            break;
+        case 'OS X':
```
alexgibson: Not related to this change, but I guess we should really file a bug to rename `OS X` to `macOS` sometime soon?

media/js/firefox/new/all-test-variations-pt2.js
```diff
@@ -0,0 +1,127 @@
...
+
+        return li;
+    };
+
+    var setupTest = function(version) {
+        var $dlButton = $('#download-button-desktop-release');
+        var $html = $('html');
+
+        // make sure desktop download button exists, user is not on Android or iOS, and user is on a recognized platform
+        if (!$html.hasClass('ios') && !$html.hasClass('android') && $dlButton.length && $dlButton.find('.unrecognized-download:visible').length === 0) {
```
alexgibson: Is there any reason why we don't want to show this for iOS and Android users? At the very least, iOS users would also see there is an Android version available, and vice-versa?

media/js/firefox/new/all-test-variations-pt2.js
```diff
@@ -0,0 +1,127 @@
...
+            // pull the nojs links out of the primary CTA
+            var $directLis = $dlButton.find('.nojs-download li').remove();
```
alexgibson: I totally get that this is for a test and is quick and dirty, but a lot of this DOM-fu could probably be avoided if we wrote a macro specifically for this use-case?

media/css/firefox/new/all-test-variations-pt2.less
```diff
@@ -0,0 +1,250 @@
...
+        list-style: none;
+    }
+
+    li {
+        display: inline-block;
+        margin: 0;
+    }
+}
+
+#test-mobile-downloads {
```
alexgibson: It looks like these mobile links need to be block-level and have a set height/width to show up on IE < 9

bedrock/settings/static_media.py
```diff
@@ -1278,6 +1285,20 @@
         ),
         'output_filename': 'js/firefox_new_scene2-bundle.js',
     },
+    'experiment_firefox_new_scene1_all_links-pt2': {
```
alexgibson: This should be an underscore before `pt2`? e.g. `experiment_firefox_new_scene1_all_links_pt2`

media/js/firefox/new/all-test-variations-pt2.js
```diff
@@ -0,0 +1,127 @@
...
+                        'event': 'alternate-version',
+                        'link-name': 'Systems & Languages'
+                    });
+                });
+
+                // place the new link underneath the main download button
+                $dlButton.append($newLink);
+            // move forest up on the page a bit (original value = -254px)
+            } else if (version === 2) {
+                $('.forest-container').css('margin-top', '-310px');
```
alexgibson: Moving this content up is fine as long as the download buttons are shown, but if you set the HTML classes `linux arm` for example, then the install link can no longer be clicked because the horizon assets cover the content.

media/js/firefox/new/all-test-variations-pt2.js
alexgibson: Could this maybe be simplified to say "as long as there is a download button visible, show the extra content"?

media/js/firefox/new/all-test-variations-pt2.js
```diff
@@ -0,0 +1,127 @@
...
+            $li.attr('data-sort', 3);
+            break;
+        case 'Linux 64-bit':
+            $li.attr('data-sort', 4);
+            break;
+        case 'OS X':
+            $li.attr('data-sort', 5);
+            break;
+        default:
+            $li.attr('data-sort', 99);
```
alexgibson: Currently `Windows (XP/Vista)` is falling through to this ordering. Do we need to special case it? Ideally, Windows users should only see the link that is appropriate to them providing we have JS.

bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -61,6 +69,12 @@
               <div class="mobile download-button-wrapper" id="download-button-wrapper-mobile" data-upgrade-subtitle="{{_('Get it free on Google Play')}}">
                 {{ download_firefox(dom_id='download-button-mobile') }}
               </div>
+
+            {% if version == '2' %}
+              <div id="test-inpage-wrapper">
```
jpetto: Ah, yeah, tested in Beta and Dev Edition. 😞 Moving this block to the last element in `main-content` seems to do the trick.

media/js/firefox/new/all-test-variations-pt2.js
jpetto: Before the special XP/Vista link was added, Verdi requested the specific ordering (I believe) based on popularity. That being the case, I think XP/Vista being last is probably correct.

I think keeping the XP/Vista link visible for all users stays true to the original intent of the test, but am open to discussion if there's a potential for harm having all 3 Windows downloads displayed. Are you thinking we should do some detection and show either both Windows 32 and Windows 64 OR only XP/Vista?

media/js/firefox/new/all-test-variations-pt2.js
jpetto: Good call. There are too many display variations to be this cute about spacing. Will remove the special casing here.

bedrock/settings/static_media.py
jpetto: 👍

media/js/firefox/new/all-test-variations-pt2.js
alexgibson: This is a bit of a special edge case, as it's not really like a specific platform link like the others - it just points to a specific bouncer end-point that supports sha-1. For Windows XP users they still want "Windows 32 bit" for example, but the main link won't work at all as it points to the sha-256 bouncer URL. When they click "Windows XP/Vista", they still get the 32 bit Windows installer. It's a little confusing with 2 buttons pointing to the same product imho, but will defer to what you think is most appropriate.

alexgibson: If it might be of any help, I seem to remember adding special casing for the original experiment in the sha-1 PR: https://github.com/mozilla/bedrock/pull/4413/

media/js/firefox/new/all-test-variations-pt2.js
jpetto: Probably a good idea. Hopefully Mac users don't get confused. 😒

media/js/firefox/new/all-test-variations-pt2.js
jpetto: I think the reason we went desktop only here was to make sure the links are actionable. Clicking on the Google Play button while on iOS would probably be more frustrating than helpful.

I agree we may want to promote/display information about Android on iOS (and vice versa), but I think that's a different task than we're performing here.

(FWIW, my preference/hope is for a single link to a freshened up `/firefox/all/` instead of the handful of links we're testing here.)

media/js/firefox/new/all-test-variations-pt2.js
jpetto: Agreed, but I'm holding off on that until we have a final direction. I really hope we go with a single link to an improved `/firefox/all/` rather than injecting direct links here.

media/js/firefox/new/all-test-variations-pt2.js
jpetto: Ah, I see the difference now. Thanks for the link to the previous PR - will pilfer.

media/css/firefox/new/all-test-variations-pt2.less
jpetto: Based on conversation below, the XP/Vista link will be merged with the Windows 32-bit link, which will skirt us around this issue for now.

media/css/firefox/new/all-test-variations-pt2.less
jpetto: 👍 Will fix.

jpetto:
Pushed an update that I believe addresses all the comments. @craigcook + @alexgibson - one more look?

alexgibson:
Still seems to be an alignment issue for experiment `v=1` when both download and refresh buttons are shown:

refresh-alignment

alexgibson:
Other than the one layout issue, the changes all look good to me. r+wc 👍

alexgibson:
I don't know if it's even worth spending much time on this for an experiment, but I noticed in IE7 (cringe), the download button in the modal is not displayed:

ie7

Haven't dug in, but it might be something that is magically fixed by using the `inline-block;` mixin.

jpetto:
Fixed the ?v=1 link UI and the IE 7 display bug (float container just needed `position: relative`). Hopefully we're good now. 😣

PR #4551, opened 2016-12-28
[fix bug 1325324] Reduce font stack

craigcook:
## Description
Removes EOT and TTF from `@font-face` definitions, leaving EOT for oldIE for now (we should consider letting oldIE just fall back to system fonts). I haven't deleted the actual font files because I think some other sites may still be referencing them. I also left the full stacks in tabzilla.less since I imagine some sites that still embed the old tabzilla are also relying on it for Open Sans.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1325324


jgmize:
+1 for letting old IE fall back to system fonts.
-1 on keeping old fonts and tabzilla files, as old tabzilla is not dependent on bedrock as deployed today, it is sourced from the old CDN domain which does not use the production bedrock deployed in AWS.

craigcook:
Hmm... so if I update tabzilla's font rules, does it mean no external sites will get the updated styling because the changes would never be sent to the old CDN? And if that's the case, is there any reason not to just delete the whole `tabzilla.less` file?

There are also a bunch of other font files that aren't in use anywhere on mozilla.org so it should be safe to remove them from git.

craigcook:
I've updated tabzilla and deleted a bunch of redundant font files.

craigcook:
Also updated the old IE styles. IE < 9 just gets a sans-serif system font (probably Arial on most Windows installs).

media/css/base/mozilla-share-cta.less
```diff
@@ -6,9 +6,7 @@
 
 @font-face {
     font-family: 'Font Awesome';
-    src: url('/media/fonts/fontawesome-webfont.eot?v4.01#iefix') format('embedded-opentype'),
-         url('/media/fonts/fontawesome-webfont.woff?v4.01') format('woff'),
-         url('/media/fonts/fontawesome-webfont.ttf?v4.01') format('truetype');
+    src: url('/media/fonts/fontawesome-webfont.woff?v4.01') format('woff');
```
craigcook: Good call. There are some other pages using the full Font Awesome, I'll update those as well.

Of course, in many of these cases it might be preferable to use a custom font with only the glyphs we need. Like for this sharing widget we only need four icons but we're loading the entire 84k font (woff2 will be a little smaller). We could also switch to SVG instead of icon fonts but that's a bigger can of worms.

craigcook: Latest Font Awesome is up to 98k 😞

jpetto:
r+ 🌱 ✂️

alexgibson:
Huge win 💯

PR #4553, opened 2016-12-29
[no bug] refactor button classes

craigcook:
## Description
Our last big button refactor introduced a set of modifier classes for button variants, but those class names weren't explicitly tied to the element. E.g. `.green` for a green button. They're not completely generic because they do have to be combined with a `.button` class, but on their own the modifier classes don't say much. It's a small thing but it was bugging me.

This just updates those modifier classes to have more explicit names, changing `.green` to `.button-green` and so on (and this is one case where strictly presentational classes are totally fine, in my opinion).

@alexgibson @jpetto  Thoughts?


jpetto:
I like the change. 👍

alexgibson:
I'm all for better semantics, but to get a handle on how this change came about, is it fixing a specific problem that exists or is it just purely a naming thing? It seems like a large change that touches a lot of places, so just curious how much review it warrants before I wade in 😄.

craigcook:
> I'm all for better semantics, but to get a handle on how this change came about, is it fixing a specific problem that exists (e.g. an inheritance issue or bug) or is it just purely a naming thing? It seems like a large change that touches a lot of places, so just curious how much review it warrants before I wade in

Purely a naming thing. There's nothing actually *broken* with the existing names, I just find them a little too generic for my tastes. It's something I've wanted to change for months and found myself with some spare time to finally do it, but I'm not committed to it. This does indeed touch a lot of spots, pretty much every button on the site that isn't the basic blue, so I'll happily punt if it seems risky.

In theory, the greatest risk is missing a button on a page somewhere. It would still have its old class in the markup and wouldn't get the styling from the new name in CSS, in which case it would most likely fall back to the basic blue `.button` style without modification. That means we could possibly show a blue button on a blue background; not ideal but not the end of the world (and easily fixed once we find it).

One possible scenario that could cause actual (visual) breakage is if a button is customized in one special place (like the download button in the home page header) and that local customization is hanging from the old class, but we update the button in the HTML and don't update the accompanying CSS that customizes it. The home page was the only place I found where that was happening and I addressed it there, but it's possible someone has done the same thing somewhere else that I didn't find in my searching.

alexgibson:
I think it's a change certainly worth making, just wanted to be sure what to test for specifically. The only other scenario I can think of is if some piece of JS used one of the old classes as a selector hook, which could result in some form of broken functionality. I think a cursory grep through the code base should be enough in terms of review 👍

jpetto:
Looks like there might be some additional files that need updating:

## LESS
- [`firefox/ios.less`](https://github.com/craigcook/bedrock/blob/refactor-button-classes/media/css/firefox/ios.less)
- [`firefox/australis/fx38_0_5/firstrun.less`](https://github.com/craigcook/bedrock/blob/refactor-button-classes/media/css/firefox/australis/fx38_0_5/firstrun.less)
- [`firefox/accounts.less`](https://github.com/craigcook/bedrock/blob/refactor-button-classes/media/css/firefox/accounts.less)
- [`firefox/whatsnew/whatsnew-42.less`](https://github.com/craigcook/bedrock/blob/refactor-button-classes/media/css/firefox/whatsnew/whatsnew-42.less#L40)

## JS
- [`firefox/whatsnew/whatsnew-42.js`](https://github.com/craigcook/bedrock/blob/refactor-button-classes/media/js/firefox/whatsnew/whatsnew-42.js)
- [`firefox/whatsnew/whatsnew-zh-tw-49.js`](https://github.com/craigcook/bedrock/blob/refactor-button-classes/media/js/firefox/whatsnew/whatsnew-zh-tw-49.js)

Those are all I could find with a pretty rudimentary regex search.

media/js/firefox/whatsnew/whatsnew-42.js
```diff
@@ -14,7 +14,7 @@
         });
     }
 
-    Mozilla.HighlightTarget.init('.button.dark');
+    Mozilla.HighlightTarget.init('.button.button-dark');
 
     $('.button.dark').on('highlight-target', function() {
```
alexgibson: Looks like you missed this one

alexgibson:
I found a couple more places here that need updating:

https://github.com/mozilla/bedrock/blob/master/media/js/plugincheck/tmpl/plugincheck.ui.tmpl.js

https://github.com/mozilla/bedrock/blob/master/bedrock/thunderbird/templates/thunderbird/includes/download-button.html

media/js/firefox/whatsnew/whatsnew-42.js
alexgibson: I think this makes sense to try and separate. There is also the option for never using ID's in our CSS and reserving them only for JS hooks, which is a rule a lot of other sites tend to follow these days. Our base CSS is a little too dependent on ID's to make this an easy change though.

bedrock/thunderbird/templates/thunderbird/includes/download-button.html
```diff
@@ -7,7 +7,7 @@
     <h4>{{ _('Download Thunderbird') }} — {{ locale_name|safe }}</h4>
     <ul>
       {% for plat in builds -%}
-        <li><a href="{{ plat.download_link_direct or plat.download_link }}" class="download-link button green">{{ plat.os_arch_pretty or plat.os_pretty }}</a></li>
+        <li><a href="{{ plat.download_link_direct or plat.download_link }}" class="download-link button button-green">{{ plat.os_arch_pretty or plat.os_pretty }}</a></li>
```
alexgibson: This only seems to fix the dumb download links for Thunderbird. The main download buttons are still showing up as blue.

alexgibson:
Spotted one more regression with the Thunderbird download buttons. There are likely a few more things that we might miss here, but hopefully nothing too major. I think at this point we might be better off going ahead with this change and then fixing any minor regressions that may occur. It would have been great to have some form of visual regression testing set up around our download buttons to make testing changes like this easier. Another project for this year maybe?

PR #4557, opened 2017-01-03
[fix bug 1316892] Expose nav links to /technology and /internet-health for all locales

alexgibson:
## Description
- Exposes links for new landing pages in main nav for all locales (enough locales now have the translated strings).
- Hides the nav download button at tablet breakpoints to accommodate the new longer strings. I would have liked a way around this, but testing different locales it just becomes too squished to fit everything on 😢. The other option would be to hide the wordmark, but I settled on it being more important for the header (particularly for the home page, which also features a separate Firefox download CTA). 

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1316892

## Testing
- We have a few variations on the masthead for different pages, so double check to make sure I didn't miss any more?

## Checklist
- [x] Requires l10n changes.


alexgibson:
Thanks @craigcook updated

PR #4565, opened 2017-01-04
[fix bug 1311643] Remove legacy 'virtual-pageview' GA event from scene2 of /firefox/new/

alexgibson:
## Description
- Removes legacy GA download event, simplifying the JS for scene2 of the download page considerably.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1311643

## Testing
https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/js/firefox/new/scene2.js
+++ b/media/js/firefox/new/scene2.js
@@ -2,55 +2,29 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-(function($, dataLayer) {
+(function($) {
     'use strict';
 
     var isIELT9 = window.Mozilla.Client.platform === 'windows' && /MSIE\s[1-8]\./.test(navigator.userAgent);
     var $directDownloadLink = $('#direct-download-link');
     var $platformLink = $('#download-button-wrapper-desktop .download-list li:visible .download-link');
-    var $stage = $('#stage');
-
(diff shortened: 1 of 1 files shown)


alexgibson:
Tested in all major browsers on the demo server above, as well as IE 6-11 and Edge. Downloads still seems to be working as expected 👍

media/js/firefox/new/scene2.js
```diff
@@ -2,55 +2,29 @@
...
+
+        // Pull download link from the download button and add to the 'click here' link.
         // TODO: Remove and generate link in bedrock.
-        $directDownloadLink.attr('href', $platformLink.attr('href'));
+        $directDownloadLink.attr('href', downloadURL);
+
+        // if user is not on an IE that blocks JS triggered downloads, start the
+        // platform-detected download after window (read: images) have loaded.
+        // only auto-start the download if a visible platform link is detected.
+        if (!isIELT9 && $platformLink.length) {
```
jpetto: Since we're verifying the truthiness of `$platformLink.length` above, could this check just be `if (!isIELT9)`?

alexgibson: Ah, good catch - I moved this inside the conditional above bug forgot to simplify this one 👍

alexgibson:
Thanks @jpetto - updated

PR #4567, opened 2017-01-04
Add Stylelint to gulp and CI

alexgibson:
Taking a look at adding Stylelint to our gulp/CI. The full list of rules that can be applied is here: http://stylelint.io/user-guide/rules/

Right now this is just linting our Sass files, but I'll take a look to see if it can do the Less files too. I'm trying to avoid purely stylistic things, but if there are things that we would normally pick out in code review anyway we should probably add them.

Like ESLint, you can make allowances for certain rules on a case by case basis using CSS comments: http://stylelint.io/user-guide/configuration/#turning-rules-off-from-within-your-css

We can also make certain rules only report as warnings, so not to break a build: http://stylelint.io/user-guide/configuration/#severities-error--warning

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,0 +1,43 @@
+{
+  "rules": {
+    "color-no-invalid-hex": true,
+    "font-family-no-duplicate-names": true,
+    "font-family-name-quotes": ["always-where-recommended", { "severity": "warning" }],
+    "function-name-case": "lower",
+    "function-url-no-scheme-relative": true,
+    "function-url-quotes": ["always", { "severity": "warning" }],
+    "number-no-trailing-zeros": [true, { "severity": "warning" }],
+    "length-zero-no-unit": [true, { "severity": "warning" }],
+    "unit-case": "lower",
+    "unit-no-unknown": true,
+    "property-case": "lower",
+    "property-no-unknown": true,
+    "keyframe-declaration-no-important": true,
+    "declaration-no-important": true,
+    "declaration-block-no-ignored-properties": true,
+    "declaration-block-no-shorthand-property-overrides": true,
+    "declaration-block-single-line-max-declarations": [1, { "severity": "warning" }],
+    "declaration-block-trailing-semicolon": ["always", { "severity": "warning" }],
+    "block-no-empty": true,
+    "selector-no-empty": true,
+    "selector-pseudo-class-no-unknown": true,
+    "selector-pseudo-element-no-unknown": true,
+    "selector-pseudo-element-case": "lower",
+    "selector-type-case": "lower",
+    "selector-type-no-unknown": true,
+    "selector-max-empty-lines": 0,
+    "media-feature-name-case": "lower",
+    "media-feature-name-no-unknown": [true, {
+      ignoreMediaFeatureNames: ["min--moz-device-pixel-ratio"]
+    }],
+    "media-feature-no-missing-punctuation": true,
+    "stylelint-disable-reason": "always-before",
+    "comment-no-empty": true,
+    "max-nesting-depth": [5, { "severity": "warning" }],
+    "no-invalid-double-slash-comments": true,
+    "no-unknown-animations": true,
+    "no-extra-semicolons": true,
+    "no-missing-end-of-source-newline": [true, { "severity": "warning" }],
+    "no-eol-whitespace": [true, { "severity": "warning" }]
+  }
+}
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -9,12 +9,20 @@ const del = require('del');
 const karma = require('karma');
 const eslint = require('gulp-eslint');
 const watch = require('gulp-watch');
+const gulpStylelint = require('gulp-stylelint');
 
-const lintPaths = [
+const lintPathsJS = [
     'media/js/**/*.js',
     '!media/js/libs/*.js',
     'tests/unit/spec/**/*.js',
     'gulpfile.js'
+];
+
+const lintPathsCSS = [
+    'media/css/**/*.scss',
+    'media/css/**/*.less',
+    'media/css/**/*.css',
+    '!media/css/libs/*'
 ];
 
 gulp.task('media:watch', () => {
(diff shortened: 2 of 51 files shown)


alexgibson:
Would very much like input from @craigcook (and also @jpetto if he has any free mozorg time)

.stylelintrc
```diff
@@ -0,0 +1,43 @@
...
+    "selector-type-no-unknown": true,
+    "selector-max-empty-lines": 0,
+    "media-feature-name-case": "lower",
+    "media-feature-name-no-unknown": true,
+    "media-feature-no-missing-punctuation": true,
+    "stylelint-disable-reason": "always-after",
+    "comment-no-empty": true,
+    "max-nesting-depth": [5, { "severity": "warning" }],
+    "no-duplicate-selectors": [true, { "severity": "warning" }],
+    "no-invalid-double-slash-comments": true,
```
jpetto: Based on the lack of related errors when running locally, this appears to be smart enough to ignore double slash comments in `.less` and `.scss` files.

.stylelintrc
```diff
@@ -0,0 +1,43 @@
...
+    "number-no-trailing-zeros": [true, { "severity": "warning" }],
+    "length-zero-no-unit": [true, { "severity": "warning" }],
+    "unit-case": "lower",
+    "unit-no-unknown": true,
+    "property-case": "lower",
+    "property-no-unknown": true,
+    "keyframe-declaration-no-important": true,
+    "declaration-no-important": true,
+    "declaration-block-no-duplicate-properties": [true, { "severity": "warning" }],
+    "declaration-block-no-ignored-properties": true,
```
jpetto: Wondering how this will play with workarounds for quirks in IE 9. Might illuminate some phony voodoo though. 😁

.stylelintrc
```diff
@@ -0,0 +1,43 @@
...
+    "function-url-no-scheme-relative": true,
+    "function-url-quotes": ["always", { "severity": "warning" }],
+    "number-no-trailing-zeros": [true, { "severity": "warning" }],
+    "length-zero-no-unit": [true, { "severity": "warning" }],
+    "unit-case": "lower",
+    "unit-no-unknown": true,
+    "property-case": "lower",
+    "property-no-unknown": true,
+    "keyframe-declaration-no-important": true,
+    "declaration-no-important": true,
```
jpetto: I really hope we can do away with `!important`, but I have a hunch we'll need it, at least for a while...

gulpfile.js
```diff
@@ -33,17 +39,27 @@ gulp.task('js:test', done => {
...
+});
+
 gulp.task('static:clean', () => {
     return del(['static/**', '!static', '!static/.gitignore']);
 });
 
 gulp.task('default', () => {
     gulp.start('media:watch');
-    gulp.watch(lintPaths, ['js:lint']);
+    gulp.watch(lintPathsJS, ['js:lint']);
```
jpetto: Should we add the CSS linter to the default task? Or is it too slow/cumbersome?

.stylelintrc
alexgibson: Yeah this only applies to CSS files, where it can cause issues. It ignores both Sass and Less comments, so maybe this is actually redundant here...

.stylelintrc
alexgibson: For now I plan to make exceptions for it in the download button styles, but let it throw an error everywhere else...

gulpfile.js
alexgibson: It currently takes around 10 seconds to lint everything, which is kinda a long time... could still add it though? Will see how it play out 👍

.stylelintrc
alexgibson: Yeah, will delve into the errors it's throwing currently - might need to drop this to a warning perhaps, and add exceptions for where it's required.

.stylelintrc
craigcook: Are we linting post-processed .css files as well, or only the pre-processed source?

alexgibson: > Are we linting post-processed .css files as well, or only the pre-processed source?

No, this would probably be a next step after this.

.stylelintrc
craigcook: We might sometimes have to update the exceptions list (there are several `!important`s in the search stuff because Google) but hopefully that won't get out of hand.

alexgibson: Exceptions are added directly around the effected areas of CSS, so adding one for the Google search thing is pretty easy. I'd prefer to do this and try and keep the rule a strict "no" by default.

craigcook: Ah, that makes sense. I was envisioning some kind of separate "ignore" list we'd have to maintain, which sounds like a bad idea. But just some special comment at the point of instance is much nicer. So in that case I also vote for keeping this as a full-blown error; `!important` should only be allowed under very special circumstances.

gulpfile.js
alexgibson: I changed our default gulp task to lint both CSS and JS files, but for changed files only instead of the whole folders. This speeds things up considerably 👍

.stylelintrc
alexgibson: I've added some overrides to usages of `!important` where it's probably need it, so it's easier to see some practical examples.

.stylelintrc
alexgibson: I'm actually going to leave this rule in, as we do have some regular `.css` files in `/media`.

jpetto:
This is looking really good to me. Do we just need to fix all current linter errors before it's ready to merge?

alexgibson:
> This is looking really good to me. Do we just need to fix all current linter errors before it's ready to merge?

I'm thinking I'll fix all the errors, and maybe a bunch of easy warnings. Then we could leave the rest up as "good-first-bugs"?

jpetto:
> I'm thinking I'll fix all the errors, and maybe a bunch of easy warnings. Then we could leave the rest up as "good-first-bugs"?

Sounds like a plan!

alexgibson:
OK, i think i have this in a pretty good place. Once #4571 is merged I'll rebase and CircleCI should pass. Once this is merged, I'll file a tracking bug so contributors can help fix the linting warnings 👍

alexgibson:
Ok this is passing in CircleCI 🎈 

I think this is now ready for code review.

media/css/base/menu-resp.less
```diff
@@ -186,11 +185,6 @@ html[dir="rtl"] #nav-main .submenu {
         padding: 0;
     }
 
-    .js #nav-main li .submenu {
-        xleft: 80px;
-        xtop: auto;
```
craigcook: Wow, where did this even come from? Did I do that? Probably been here for years...

alexgibson: No idea - it was so mysterious I even searched in case there was some random proprietary thing I didn't know existed. Came up blank

PR #4569, opened 2017-01-05
Add git remote CLI switch for tag-release.sh script

pmac:
Improve docs for said script.

### DIFF
--- a/bin/tag-release.sh
+++ b/bin/tag-release.sh
@@ -1,6 +1,23 @@
 #!/bin/bash
+set -euo pipefail
 
 moz_git_remote="${MOZ_GIT_REMOTE:-origin}"
+do_push=false
+
+# parse cli args
+while [[ $# -ge 1 ]]; do
+  key="$1"
+  case $key in
+    -p|--push)
+      do_push=true
+      ;;
+    -r|--remote)
+      moz_git_remote="$2"
+      shift # past argument
+      ;;
+  esac
+  shift # past argument or value
+done
 
 # ensure all tags synced
 git fetch --tags
@@ -12,7 +29,7 @@ while ! git tag -a $tag_value -m "tag release $tag_value" 2> /dev/null; do
   tag_value="${date_tag}.${tag_suffix}"
 done
 echo "tagged $tag_value"
-if [[ "$1" == "--push" ]]; then
+if [[ "$do_push" == true ]]; then
   git push "$moz_git_remote" "$tag_value"
   git push "$moz_git_remote" master:prod
 fi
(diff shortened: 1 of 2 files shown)


pmac:
@alexgibson @jpetto @craigcook I've improved the tag/push CLI tool and the docs therefor. Let me know what you think.

alexgibson:
👍

bin/tag-release.sh
```diff
@@ -1,6 +1,23 @@
...
+set -euo pipefail
 
 moz_git_remote="${MOZ_GIT_REMOTE:-origin}"
+do_push=false
+
+# parse cli args
+while [[ $# -ge 1 ]]; do
+  key="$1"
+  case $key in
+    -P|--push)
```
jpetto: Is this intended to be an uppercase `-P` (while a lowercase `-r` below)?

docs/pipeline.rst
```diff
@@ -66,6 +66,19 @@ pushes, this is all handled by the pipeline, and is subject to change according
     #. If any staging tests fail above, check retries, e.g. `bedrock_test_stage_eu_west`_
 #. Tag and push the deployment by running ``bin/tag-release.sh --push``
 
+.. note::
+
+    By default the ``tag-release.sh`` script will push to the ``origin`` git remote. If you'd
+    like for it to push to a different remote name you can either pass in a ``-r`` or
+    ``--remote`` argument, or set the ``MOZ_GIT_REMOTE`` environment variable. So the following
```
jpetto: Is it possible to have `MOZ_GIT_REMOTE` picked up from `.env`? If not, I would probably just change the name of my remote to match the default (to avoid having to specify each time), which is maybe what we want anyway.

pmac: hmm... possible, but probably overkill for this. Easiest thing to do is to set an alias for your command in your `.bashrc` file:

```bash
alias bedrock-push='bin/tag-release.sh --push -r upstream`
```

Then you just have to type `bedrock-push` whenever you want to do a push. Alternately, just set the env var in your `.bashrc` (`export MOZ_GIT_REMOTE=upstream`).

bin/tag-release.sh
pmac: I did intentionally use a capital here. It's a mode that affects things in the remote repo, so I wanted to make it seem a bit more intentional, but that's probably overkill...

jpetto: I don't mind the capital P (for the reason you stated) - just wanted to make sure it was intentional. 👍

pmac:
I changed the short push option to lowercase `-p` for consistency, and updated the deployment documentation to better reflect the current reality.

pmac:
If y'all are happy with this I still need someone to approve a review before it's mergable, or just approve and merge it :) Thanks for the r?

alexgibson:
All looks good to me, thanks Paul!

PR #4581, opened 2017-01-13
Update FxA docs.

jpetto:
## Description

Adds bit to the FxA docs to get things working w/CSP.

### DIFF
--- a/docs/firefox-accounts.rst
+++ b/docs/firefox-accounts.rst
@@ -18,9 +18,10 @@ a non-production environment requires some additional Firefox profile configurat
 Local Development
 -----------------
 
-#. Set ``FXA_IFRAME_SRC`` in your ``.env`` file to point to the FXA development server:
+#. Add the following config settings to your ``.env`` file:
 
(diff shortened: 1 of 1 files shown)


docs/firefox-accounts.rst
```diff
@@ -18,9 +18,10 @@ a non-production environment requires some additional Firefox profile configurat
 Local Development
 -----------------
 
-#. Set ``FXA_IFRAME_SRC`` in your ``.env`` file to point to the FXA development server:
+#. Add the following config settings to your ``.env`` file:
 
     ``FXA_IFRAME_SRC=https://stomlinson.dev.lcip.org/``
+    ``CSP_EXTRA_FRAME_SRC=https://stomlinson.dev.lcip.org/``
```
alexgibson: Should this and `FXA_IFRAME_SRC above actually have single quotes around the URLs?

jpetto: I don't have any quotes around them in my `.env` file and it works, so I think it's okay?

PR #4583, opened 2017-01-16
Bug 1331479 - /foundation localization fixes

TheoChevalier:
Fixing hardcoded strings + donate URL: https://bugzilla.mozilla.org/show_bug.cgi?id=1331479

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/foundation/templates/foundation/base-resp.html
+++ b/bedrock/foundation/templates/foundation/base-resp.html
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% set_lang_files "foundation/index" %}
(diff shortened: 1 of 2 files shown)


bedrock/foundation/templates/foundation/base-resp.html
```diff
@@ -2,6 +2,7 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% set_lang_files "foundation/base-resp" %}
```
TheoChevalier: @flodolo do you have a preference here? Maybe a different name? Didn’t want to add the sidebar strings to main.lang since those are opt-in pages

TheoChevalier: Actually, probably better to put them into index.lang and then set this lang file here?

PR #4584, opened 2017-01-17
[fix bug 1331053] Add pop up on homepage leading to Open Design blog

alexgibson:
## Description
- Adds a stickup popup to the home page linking to the Open Design blog.
- Popup only shows for tell desktop viewports. Other viewports will see it embedded in the page right after the hero section.
- Popup only shows once per tab session so not to annoy people any further.
- Shown for `-en` locales only.
- Also removes the home page takeover code / switches.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1331053

## Testing
- Popup should be sticky on tall desktop viewports (test closing the popup).
- Popup should be embedded on mobile viewports or short desktop viewports (close button should not be visible).

## Checklist
- [ ] Requires l10n changes.


alexgibson:
OK, I've rebased this branch and also taken care of removing the home page takeover code / switches.  Once we turn off the takeover in prod we can push this out.

media/css/mozorg/home/brand-popup.scss
```diff
@@ -0,0 +1,111 @@
...
+        position: relative;
+        text-decoration: none;
+        text-transform: uppercase;
+
+        span {
+            @include transition(background-color 150ms ease-out);
+            background-color: #ebc044;
+            padding: 10px 15px;
+            -webkit-box-decoration-break: clone;
+            box-decoration-break: clone;
```
craigcook: ooo this is a new one on me. I expect it'll come in handy a lot if we have to keep doing this highlight effect.

alexgibson: New to me too (found it on MDN) - it doesn't work in IE or Edge yet, but the degradation is probably acceptable.

PR #4586, opened 2017-01-18
[bug 1331048] Hide navbar styling behind switch

craigcook:
## Description
While the takeover was behind a switch, the navbar changes weren't. This puts those style changes behind the same switch as the takeover (using a class for the color; kinda kludgey but it only needs to last a day).

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1331048

## Testing
Ensure the navbar shows the old wordmark and coloring when the switch is off.

## Checklist
- [ ] Requires l10n changes.


jpetto:
Looks good to me. r+ 🐟

PR #4588, opened 2017-01-18
[bug 1331048] Remove brand switch from main navigation

alexgibson:
We need to merge / push this before tomorrow as when we flip the switch off again, the old branding will come back.


alexgibson:
I added a second commit that adds some margin around the top and sides of the modal to make it a bit more obvious that it's a modal. We had a few reports from people who we're confused they couldn't interact with the rest of the screen, and didn't realize the modal was a takeover as such.

Would be interested in thoughts and review from @craigcook

PR #4589, opened 2017-01-18
Build and deploy via Jenkinsfile

pmac:
This is a big one. Let's go over it really well.


pmac:
@glogiotatidis @davehunt I'd appreciate any feedback you might have.

pmac:
Of note is that this new process pays no attention to git tags. It will push to stage then prod for any push to the `prod` branch. The tags are now just for book-keeping (which I still like). If anyone has any ideas on whether we can or should continue to require tagging in this pipeline please speak up. When looking at it I just couldn't find a scenario that would mean that a push to the prod branch should not be deployed, but we can easily block on it if we decide that is desirable. 

@jgmize you may have an opinion on the above as well.

Jenkinsfile
```diff
@@ -10,25 +34,267 @@ def ircNotification(stage, status) {
...
+def integrationTestJob(propFileName, appName, region) {
+    def testsBaseDir = 'docker/jenkins/properties/integration_tests'
+    def testsFileExt = '.properties'
+    return {
+        node {
+            unstash 'scripts'
+            unstash 'tests'
+            def fullFilename = "${testsBaseDir}/${propFileName}${testsFileExt}"
+            def testScript = "docker/jenkins/run_integration_tests.sh ${fullFilename}".toString()
+            withCredentials([[$class: 'UsernamePasswordMultiBinding',
```
davehunt: Ooh, I didn't know about `UsernamePasswordMultiBinding`...

Jenkinsfile
```diff
@@ -10,25 +34,267 @@ def ircNotification(stage, status) {
...
+                    ircNotification('EU-West Registry Push', 'failure')
+                    throw err
+                }
+            }
+        }
+        node {
+            unstash 'scripts'
+            unstash 'tests'
+            // prep for next stage
+            sh 'docker/jenkins/build_integration_test_image.sh'
```
davehunt: If the next stage happens on a different executor with a different workspace, will the effects of running this script be preserved?

docker/jenkins/run_integration_tests.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash -xe
-GIT_COMMIT=${GIT_COMMIT:-$(git rev-parse HEAD)}
-cp docker/dockerfiles/bedrock_integration_tests Dockerfile
-docker build -t bedrock_integration_tests:${GIT_COMMIT} --pull=true .
+
+# $1 should be the properties file for this run
+source "$1"
```
davehunt: 👍 nice

Jenkinsfile
pmac: I'd like to see a list of all of these, but I just found this out on the web somewhere when looking for `withCredentials` and `Jenkinsfile`. 🤷‍♂️

Jenkinsfile
pmac: This creates a docker image, so as long as it executes on the same machine with the same docker daemon it will be fine. If we move to a distributed executor jenkins setup then we may encounter problems. But I think those problems could be overcome with a shared docker repository for all jenkins nodes.

docker/jenkins/run_integration_tests.sh
pmac: Thanks! This is the best way I could figure out to inject environment variables via a file. I looked briefly at parsing the properties files as-is in the Groovy and using `withEnv`, but this seemed cleaner, easier, and less error prone.

pmac:
Based on some of the ideas in @glogiotatidis' work in mozilla/kuma#4102 I'm refactoring some of this. I'll add the do-not-merge tag.

pmac:
> I'd like to see more use of our jenkins global functions library in the spirit of re-using and sharing of code and practices among our projects but before this we'll need to expand it with more commands and flexibility.

Completely agree. I'm really happy that you started that project and provided a good example of its use. I think we'll be able to move things into there quickly.

Jenkinsfile
```diff
@@ -148,45 +148,48 @@ if ( config.branches.containsKey(env.BRANCH_NAME) ) {
...
-                node {
-                    unstash 'scripts'
-                    withEnv(["DEIS_PROFILE=${region.deis_profile}",
-                             "DOCKER_REPOSITORY=${appname}",
-                             "DEIS_APPLICATION=${appname}"]) {
-                        try {
-                            retry(3) {
-                                sh 'docker/jenkins/push2deis.sh'
+            // ensure no deploy/test cycle happens in parallel for an app/region
+            lock (stageName) {
```
glogiotatidis: 👍

Jenkinsfile
```diff
@@ -1,35 +1,215 @@
...
- *
- * @param stage step of build/deploy
- * @param result outcome of build (will be uppercased)
-*/
-def ircNotification(stage, status) {
-    sh "bin/irc-notify.sh --stage '${stage}' --status '${status}'"
-}
+#!groovy
+
+@Library('github.com/mozmar/jenkins-pipeline@master')
```
pmac: Will do. Thanks!

Jenkinsfile
```diff
@@ -1,35 +1,215 @@
...
+        }
+        node {
+            unstash 'scripts'
+            unstash 'tests'
+            // prep for next stage
+            sh 'docker/jenkins/build_integration_test_image.sh'
+            utils.ircNotification(config, [stage: 'Docker Image Pushes', status: 'complete'])
+        }
+    }
+
```
glogiotatidis: Insert a `milestone()` right here to ensure that " that older builds of a job will not overwrite a newer build.", i.e. that we won't be pushing an old build to the server?

https://jenkins.io/blog/2016/10/16/stage-lock-milestone/

pmac:
Okay. Docs are updated and steps are [looking good in BlueOcean](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/deploy-via-jenkinsfile/46/pipeline). Ready for merging? Could be.

PR #4590, opened 2017-01-18
Fix bug 1332008: Add /protocol redirect to open design blog

pmac:


pmac:
Haven't seen any objections in bugzilla. I think we're good to merge and can always update later if need be. Thanks Alex!

PR #4594, opened 2017-01-20
Skip home page tests in IE due to brand popup

alexgibson:
We hit some intermittent failures on the last prod push [1], [2] in IE as it seem to have trouble detecting the homepage popup that appears at the bottom of the screen. I could sink a ton of time into debugging this to try and make it more resilient, but this probably isn't worth it as it's only going to be around for a couple of weeks. This PR skips said tests in IE.

We need to remember to unskip these tests again when the popup gets removed.

[1] https://ci.us-west.moz.works/job/bedrock_integration_tests_runner/11163/
[2] https://ci.us-west.moz.works/job/bedrock_integration_tests_runner/11169/


alexgibson:
@pmac r? Would be good to get this in so the next push isn't disrupted by this  :(

PR #4595, opened 2017-01-21
Bug 1322959 - IHR Privacy-Security page, fix syntax error (unwrapped string)

flodolo:


craigcook:
D'oh! Sorry I missed that. :man_facepalming:

flodolo:
One trick that I use on new templates is to:
* Extract the .lang file, copy it in one of the locales
* Remove all translations and replace them with a single character like `.`
* Check the page

It's pretty fast (search for lines not starting with `;`, replace the text), unless there are Python variables, in which case you need to keep them manually (if you're missing a variable the string will automatically fallback to English).

We will probably need to push this out on Monday, since we're going to have translations of that page available.

PR #4580, opened 2017-01-13
[fix bug 1329661] Create ad-campaign tailored Firefox download pages

alexgibson:
## Description
- Adds two campaign specific download page themes to /new.
- URLs and final copy are still TBD.
- en-US only.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1329661

## Testing
Demo URLs:
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/?xv=breakfree
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/?xv=wayofthefox

## Checklist
- [ ] Requires l10n changes.
- [ ] Related functional & integration tests passing.

PR #4596, opened 2017-01-23
[fix bug 1331069] Replace Flashtalking Tracking Pixel With DoubleClick

alexgibson:
## Description
- Replaces Flashtalking Tracking Pixel With DoubleClick
- Updates switch name for the pixel to be more generic. We'll have to remember to turn this on again in staging / prod, and remove the old switch.
- We'll need to make sure the flip `SWITCH_TRACKING_PIXEL` is turned on on stage / prod prior to pushing this. The double click switch (`SWITCH_DOUBLE_CLICK`) should be turned on Feb 1st (this will toggle off Flash Talking).

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1331069

## Testing
- Demo: https://www-demo1.allizom.org/en-US/firefox/new/?scene=2
- Test that the correct pixel is triggered depending on the switch?

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -1302,5 +1302,14 @@ GOOG_CUSTOM_SEARCH_TYPES = {
     'activist': '014783244707853607354:bbpl3emdsii',
 }
 
-if config('SWITCH_FLASHTALKING', default=DEV, cast=bool):
-    CSP_IMG_SRC += ('servedby.flashtalking.com',)
+# Bug 1331069 - Replace Flashtalking Tracking Pixel With DoubleClick
+FLASH_TALKING_URL = 'https://servedby.flashtalking.com/spot/8/6247;40428;4669/?spotName=Mozilla_Download_Conversion'
+DOUBLE_CLICK_URL = ('https://ad.doubleclick.net/ddm/activity/src=6417015;type=deskt0;cat=mozil0;dc_lat=;dc_rdid=;'
+                    'tag_for_child_directed_treatment=;ord=1;num=1?&_dc_ck=try')
+
+TRACKING_PIXEL_URL = DOUBLE_CLICK_URL if config('SWITCH_DOUBLE_CLICK', default=DEV, cast=bool) else FLASH_TALKING_URL
+
+if config('SWITCH_TRACKING_PIXEL', default=DEV, cast=bool):
+    CSP_IMG_SRC += (
+        'ad.doubleclick.net' if config('SWITCH_DOUBLE_CLICK', default=DEV, cast=bool) else 'servedby.flashtalking.com',
+    )
--- a/media/js/firefox/new/pixel.js
+++ b/media/js/firefox/new/pixel.js
@@ -8,12 +8,16 @@
     // pixel status bug https://bugzilla.mozilla.org/show_bug.cgi?id=1311423
     function addPixel() {
         if (!window._dntEnabled()) {
-            var $pixel = $('<img />', {
-                width: '1',
-                height: '1',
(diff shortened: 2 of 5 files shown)


bedrock/settings/base.py
```diff
@@ -1302,5 +1302,6 @@ def lazy_email_backend():
     'activist': '014783244707853607354:bbpl3emdsii',
 }
 
-if config('SWITCH_FLASHTALKING', default=DEV, cast=bool):
-    CSP_IMG_SRC += ('servedby.flashtalking.com',)
+if config('SWITCH_TRACKING_PIXEL', default=DEV, cast=bool):
+    CSP_IMG_SRC += ('ad.doubleclick.net',
```
alexgibson: @pmac - this isn't critical as we're going to remove the flash talking pixel pretty quickly, but it would be nice if there was a simple way to use the switch here too instead of add both domains to the white list. Is this possible?

pmac: sure.. you could do something like

```python
CSP_IMG_SRC += (
    'ad.doubleclick.net' if config('SWITCH_DOUBLE_CLICK', default=DEV, cast=bool) else 'servedby.flashtalking.com',
)

Comment (alexgibson) [inline bedrock/settings/base.py]: Sorry, I totally didn't see the existing example of using a switch already right above this line of code (wasn't sure if it was OK from `base.py`). Thanks!

Comment (pmac) [inline bedrock/settings/base.py]: yeah... just can't import the `switch()` function here... so it's not quite as easy, but you can manually grab the switch environment variable. It's possible that I could make the switch function usable here, but I've not worked on that.

**Comment by alexgibson** [comment]
Ok, switch is in place for toggling which pixel is used. Hopefully the implementation is simple / sane enough. This is ready for review.

**Comment by alexgibson** [comment]
@craigcook I'm going to leave the do-not-merge label on this for now. I'll need to rebase this once #4580 merges, which is probably higher priority.

We also need to time merging / pushing this to prod with the switch changes listed above.

File: bedrock/settings/base.py [hunk 2]
```diff
@@ -1302,5 +1302,14 @@ def lazy_email_backend():
     'activist': '014783244707853607354:bbpl3emdsii',
 }
 
-if config('SWITCH_FLASHTALKING', default=DEV, cast=bool):
-    CSP_IMG_SRC += ('servedby.flashtalking.com',)
+# Bug 1331069 - Replace Flashtalking Tracking Pixel With DoubleClick
+FLASH_TALKING_URL = 'https://servedby.flashtalking.com/spot/8/6247;40428;4669/?spotName=Mozilla_Download_Conversion'
+DOUBLE_CLICK_URL = ('https://ad.doubleclick.net/ddm/activity/src=6417015;type=deskt0;cat=mozil0;dc_lat=;dc_rdid=;'
```
pmac: 👍

alexgibson:
Rebased and updated now that #4580 merged. I've also set `SWITCH_TRACKING_PIXEL` to `on` in both stage and prod ready for whenever this gets pushed.

media/js/firefox/new/pixel.js
```diff
@@ -8,12 +8,16 @@
     // pixel status bug https://bugzilla.mozilla.org/show_bug.cgi?id=1311423
     function addPixel() {
         if (!window._dntEnabled()) {
-            var $pixel = $('<img />', {
-                width: '1',
-                height: '1',
-                src: 'https://servedby.flashtalking.com/spot/8/6247;40428;4669/?spotName=Mozilla_Download_Conversion'
-            });
-            $('body').append($pixel);
+            var href = $('#strings').data('trackingPixel');
```
craigcook: TIL: dashes are converted to camelCase in dataset. https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#Name_conversion

PR #4597, opened 2017-01-23
[fix bug 1330605] Make the winner of /new download link test the new default

alexgibson:
## Description
- Adds a modal button to allow users to download Firefox for platforms other than the one they are currently using.
- Removes old A/B tests.
- Adds some new strings behind an l10n tag (cc @flodolo @peiying2)

TODO:
- [x] Add unit tests for platform link helper
- [x] Add functional test for opening modal
- [x] GA requirements
- [x] Add a few no-JS styles for better readability.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1330605

## Testing
- Demo: https://www-demo5.allizom.org/en-US/firefox/new/
- Test platform links are as expected, test modal responsiveness, RTL. Make sure no errors are thrown if a locale doesn't have the new l10n tag?

## Checklist
- [x] Requires l10n changes.


alexgibson:
OK aside from GA requirements this is pretty much complete (i think). Would appreciate if @pmac had time to take a look at the Python bits?

alexgibson:
GA requirements have been added, removing the do-not-merge label.

craigcook:
Some kind of weird alignment on the two-column list in Chrome and Safari but I can't find anything obvious in the CSS. I think maybe Webkit is applying the bottom margin of the last item in the left column to the top of the right column, which just seems silly (Firefox and IE get it right). Setting the list-items to `inline-block` fixes it but that seems like a hack. Setting `margin: 5px 0;` also fixes it and might be a more sound approach.

chrome-platforms-align

craigcook:
IE10 seems to have trouble with the arrow SVG. It has a viewbox of 100 x 100 so there's space around the arrow, so I think IE is scaling the entire box and not just the arrow inside it. I don't have IE11 or Edge VMs to test in, maybe they do better (Firefox and Chrome are fine). I think we can fix it in the SVG file though and make all the browsers happy.

ie10-arrow-squish

media/css/firefox/new/other-platforms.less
```diff
@@ -0,0 +1,281 @@
...
+    .download-platform-list {
+        .multi-column(auto, 2, 20px);
+        list-style: none;
+        margin: 20px auto;
+        text-align: left;
+        width: 80%;
+
+        li {
+            display: block;
+            margin: 0 0 10px;
```
craigcook: Setting this to `margin: 5px 0;` seems to fix the funky alignment in Webkit.

media/css/firefox/new/other-platforms.less
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+
+        a:link,
+        a:visited {
+            .font-size(@largeFontSize);
+            background-image: url('/media/img/firefox/new/down-arrow-blue.svg');
+            background-position: left center;
+            background-repeat: no-repeat;
+            background-size: 20px 16px;
```
craigcook: Use the `.background-size(20px 16px)` mixin here? Though at this point I wonder if we need to worry about prefixing background-size for ancient webkits. Maybe time to let it go.

media/css/firefox/new/other-platforms.less
```diff
@@ -0,0 +1,281 @@
...
+                margin: 10px 0 0 -24px;
+            }
+        }
+
+        a:link,
+        a:visited {
+            .font-size(@largeFontSize);
+            background-image: url('/media/img/firefox/new/down-arrow-blue.svg');
+            background-position: left center;
+            background-repeat: no-repeat;
```
craigcook: Could combine these in a shorthand `background` declaration, unless there's a reason verbosity is better in this instance (sometimes it is).

media/css/firefox/new/other-platforms.less
alexgibson: This fixes the issue in Webkit, but then breaks alignment in Firefox sadly. Seems `display: inline-block;` is the best cross browser remedy after a bit of searching online.

alexgibson: Heh, so it turns out IE doesn't like `display: inline-block;` either. The best solution seems to be avoid doing anything with margins at all, and instead use `line-height`.

alexgibson:
> IE10 seems to have trouble with the arrow SVG. It has a viewbox of 100 x 100 so there's space around the arrow, so I think IE is scaling the entire box and not just the arrow inside it. I don't have IE11 or Edge VMs to test in, maybe they do better (Firefox and Chrome are fine). I think we can fix it in the SVG file though and make all the browsers happy.

Looks like the image was missing `height` and `width` properties in the SVG file, which IE craps out about when missing. I'll add them back 👍

media/css/firefox/new/other-platforms.less
craigcook: Bah, multi-column is weird. I tried it with padding instead of margin but it was the same thing. `line-height` doesn't seem correct either but it might be the best we've got. Just wish every browser did it the same...

bedrock/firefox/templatetags/helpers.py
```diff
@@ -221,6 +221,66 @@ def download_firefox(ctx, channel='release', platform='all',
 
 
 @library.global_function
+@jinja2.contextfunction
+def download_firefox_desktop_list(ctx, channel='release', dom_id=None, locale=None,
```
pmac: isn't this code very similar to that for the buttons?

pmac: Looks like perhaps we can factor most of this into a function that both this and the `download_firefox` helper can use.

media/css/firefox/new/other-platforms.less
alexgibson: `line-height` seems to have fixed the issue in all the browsers I've tested in (checked IE 11 and Edge as well).

alexgibson:
@pmac - I added a commit that refactors the code into a `desktop_builds` function (I tried to follow the same convention as has already been done for android and iOS). Things seem to be working, and tests are passing. Would appreciate some careful review here as I'm kinda paranoid about touching this code 😉

alexgibson:
Rebased and squashed, thanks @pmac 👍

PR #4600, opened 2017-01-25
[fix bug 1333938] Fix Failing Bedrock/Newsletter Tests

heatherbooker:
## Description
Currently, cloning and installing Bedrock [as per the docs](https://bedrock.readthedocs.io/en/latest/install.html) leads to 11 tests in bedrock/newsletter failing. Adding one line to .env-dist fixes it.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1333938

## Testing
It was broken, and now it's not! Yay!

## Checklist
- [ ] Requires l10n changes.
- [X] Related functional & integration tests passing.


jpetto:
I've verified the bug and the fix, however I'm curious to know if there's a better way to get around the issue.

@pmac or @jgmize - Would it be preferable to patch this setting in the tests themselves? I believe this new setting is required for viewing/testing newsletter pages locally anyway, so this may be the best solution.

muhammednagy:
this happened with me too i spend 3 hours trying to know the reason

pmac:
Great find! I was seeing these too from time to time and hadn't yet spent the time to figure out why.

I'm torn on how to solve it though. I'm tempted to take the switch out, or at least change it to a setting with a default of "not maintenance mode" regardless of `DEV` setting. It is kindof nice to have around in case we need it.

pmac:
I think I'll go ahead and merge this, but then fix it more properly in a followup. Thanks for the help with this everyone!

heatherbooker:
Thanks to everyone and especially thank you so much to @jpetto for helping me with my first contribution! So exciting! :)

PR #4605, opened 2017-01-27
[bug 1332477] Privacy day promo

craigcook:
## Description
Adds an easter egg to the Internet Health - Privacy and Security page. Enter the Konami code (up up down down left right left right b a) to trigger a kinetic cascade of cartoon kazoos.

The konami code activates the easter egg, and once active any other click or keypress should deactivate and reset it. Entering the konami code again starts it over. The library I used supports touch events so it should work fairly well by swiping on mobile: up up down down left right left right tap tap. Small screens get fewer kazoos to hopefully improve performance and ease the download burden (should only download load two gifs instead of five). This was kind of hastily thrown together and I'm iffy on the mobile stuff. If it proves too troublesome we can make the egg desktop-only.

Everything is behind the switch `data-privacy-day-2017`

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1332477

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/internet-health/privacy-security/

## Checklist
- [ ] Requires l10n changes.


media/js/mozorg/internet-health/privacy-day-egg.js
```diff
@@ -0,0 +1,30 @@
...
+
+(function($, KonamiCode) {
+    'use strict';
+
+    var $document = $(document);
+    var $body = $('body');
+    var $kzp = $('#kzp');
+
+    new KonamiCode(function() {
+        $kzp.fadeIn('fast', function() {
```
alexgibson: By no means a blocker, but it would be nice to avoid jQuery animations and just go full CSS?

media/css/mozorg/internet-health/privacy-day-egg.scss
```diff
@@ -0,0 +1,231 @@
...
+    #kz15,
+    #kz16 {
+        display: block;
+    }
+}
+
+// Trigger the animation on a body class
+body.easter {
+    .kz {
+        animation: kz-fall infinite linear, kz-spin infinite linear;
```
alexgibson: Should use animation mixins here and below?

media/css/mozorg/internet-health/privacy-day-egg.scss
```diff
@@ -0,0 +1,231 @@
...
+
+    #kz16 {
+        left: 26%;
+        animation-name: kz-fall, kz-backspin;
+        animation-delay: 9.5s;
+        animation-duration: 8.2s, 8.25s;
+    }
+}
+
+@keyframes kz-fall {
```
alexgibson: Need webkit keyframes too?

media/css/mozorg/internet-health/privacy-day-egg.scss
craigcook: We have a mixin for `animation` but none for all the longhand properties like `animation-delay`. I could add those mixins to Pebbles but if we start using PostCSS and autoprefixer they would quickly become redundant. So for this one-off easter egg I think I'll just hardcode all the -webkit prefixes instead of adding a bunch of little-used and short-lived mixins.

media/js/mozorg/internet-health/privacy-day-egg.js
craigcook: It took a little bit of refactoring so the opacity, display, and animation all have different classes. I've also had to allow time for the transition to complete before completely hiding it and stopping the animation. I think an accidental side-benefit of the fadeIn was that it imposed a delay before adding the key binding to remove the easter egg. Without some delay your finger is still finishing a keystroke and ends the animation before it even begins. I think a brief delay before calling `endEgg()` solves that but there may be a better way.

PR #4606, opened 2017-01-27
Upgrade Whitenoise to 3.3.0

pmac:
Gives us a couple of new and improved things:

* Adds the new "[immutable](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Revalidation_and_reloading)" Cache-Control header extension
* Uses whitenoise for dev so all of the files work


root_files/projects/calendar/caldata
```diff
@@ -0,0 +1 @@
+../../../static/caldata
```
pmac: This is a symlink to the caldata files. This used to be handled by a custom whitenoise subclass, but this is much more straight-forward and less code. I'd have just moved the files but they need to be served by both URLs.

wsgi/app.py
```diff
@@ -21,11 +21,9 @@
 
 # must be imported after env var is set above.
 from django.core.wsgi import get_wsgi_application
-from bedrock.base.static import BedrockWhiteNoise
 from raven.contrib.django.raven_compat.middleware.wsgi import Sentry
 
 application = get_wsgi_application()
-application = BedrockWhiteNoise(application)
```
pmac: New whitenoise prefers to be a Django middleware rather than a wsgi one.

pmac:
cc @escattone in case MDN wants to do the same

PR #4607, opened 2017-01-27
[fix bug 1332663] Fix CSS linting errors for CSRF failure page

heatherbooker:
## Description
Fix CSS linting errors for the CSRF failure page, as laid out in the bug report.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1332663

## Testing
There are now no error messages when the following command is run:
```bash
./node_modules/gulp-stylelint/node_modules/.bin/stylelint "media/css/csrf-failure.less"
```

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/css/csrf-failure.less
+++ b/media/css/csrf-failure.less
@@ -36,35 +36,115 @@
 }
 
 @-webkit-keyframes hinge {
-  0% { -webkit-transform: rotate(0); -webkit-transform-origin: top left; -webkit-animation-timing-function: ease-in-out; }
-  20%, 60% { -webkit-transform: rotate(80deg); -webkit-transform-origin: top left; -webkit-animation-timing-function: ease-in-out; }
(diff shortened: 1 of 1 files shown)


heatherbooker:
I'm curious whether it would be preferable to have each style on a new line:
```less
0% {
  -webkit-transform: rotate(0);
  -webkit-transform-origin: top left;
  -webkit-animation-timing-function: ease-in-out;
}
```
I tried to change as little as possible in order to resolve the linting errors, but it seems that the way I left it:
```less
0% {
  -webkit-transform: rotate(0); -webkit-transform-origin: top left; -webkit-animation-timing-function: ease-in-out;
}
 ```
may be a bit strange. Let me know and I'll be happy to make changes!

alexgibson:
> I'm curious whether it would be preferable to have each style on a new line

Yes please, if you could put each style on a separate line that would be great.

heatherbooker:
Ok @alexgibson , I added those newlines! I also sent another PR #4627 which has a rule to find lines like these. Hope that helps!

PR #4610, opened 2017-01-27
[fix bug 1332677] Fix CSS linting errors for styleguide

heatherbooker:
## Description
Fix CSS linting errors for the styleguide, as laid out in the bug report.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1332677

## Testing
There are now no error messages when the following command is run:
```bash
./node_modules/gulp-stylelint/node_modules/.bin/stylelint "media/css/styleguide/**/*.less"
```

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/css/styleguide/identity-firefox.less
+++ b/media/css/styleguide/identity-firefox.less
@@ -239,8 +239,8 @@
             position: absolute;
             top: 0;
             right: 75px;
-            background: url(/media/img/styleguide/download-arrow.png) center right no-repeat;
-            padding-right: 40px
+            background: url("/media/img/styleguide/download-arrow.png") center right no-repeat;
+            padding-right: 40px;
         }
     }
     .clear {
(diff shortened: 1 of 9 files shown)


heatherbooker:
(I have the [same question in PR#4607](https://github.com/mozilla/bedrock/pull/4607))

I'm curious whether it would be preferable to have each style on a new line:
```less
#swatch_f4f4f4 .swatch span {
    background-color: #f4f4f4;
    color: #000;
}
```
I tried to change as little as possible in order to resolve the linting errors, but it seems that the way I left it:
```less
#swatch_f4f4f4 .swatch span {
    background-color: #f4f4f4; color: #000;
}
 ```
may be a bit strange. Let me know and I'll be happy to make changes!

alexgibson:
> I'm curious whether it would be preferable to have each style on a new line

Yes please, this is definitely preferable. We should really add a CSS linting rule for this (thanks for the reminder). If you could make the changes you suggest, that would be great.

heatherbooker:
Ok @alexgibson , I added those newlines! I also sent another PR #4627 which has a rule to find lines like these. Hope that helps!

PR #4615, opened 2017-01-31
[fix bug 1319244] Remove search test code

craigcook:
## Description
Removes all the bits and pieces relating to the cross-site search testing. 

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1319244

## Checklist
- [ ] Requires l10n changes.


bedrock/base/templates/macros.html
```diff
@@ -474,27 +474,3 @@ <h2 class="thank-you hidden">{{ _('Your download link was sent.') }}</h2>
     </footer>
   </section>
 {%- endmacro %}
-
-
-{% macro google_custom_search(search_type, type_class, default_term, term_list) %}
```
alexgibson: I seem to remember some warning from either @pmac or @jgmize about removing a shared macro sometimes causing issues with pages that are still cached. Not sure if this is still the case, but flagging here?

alexgibson:
There still seem to be some google related things left in the CSP white list, I'm guessing we should probably remove those too?

craigcook:
D'oh, good catch on the CSP bits. Updated.

Not sure what to do about removing the macro... if removing it breaks things cached pages, does leaving it break things on uncached pages?

bedrock/base/templates/macros.html
pmac: Hmm... I don't remember that. Cached in what way? We should be able to remove any macro as long as it's no longer used by any templates.

pmac:
This looks good to me. If there are issues I believe our selenium tests should catch them.

bedrock/base/templates/macros.html
alexgibson: All I remember was being asked to remove the macro in a second commit later on for fear of some kind of temporary breakage on deployment. Perhaps this is not the same thing.

alexgibson:
Looks like this needs a rebase, bot otherwise r+ from me

PR #4618, opened 2017-01-31
[fix bug 133513] Add internet health report link to home page

craigcook:
## Description
Adds a link to internethealthreport.org to the "impact" section of the home page.

The new string is behind the l10n tag `mozorg-home-health-report`

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1335131

## Checklist
- [x] Requires l10n changes.


craigcook:
cc @peiying2

alexgibson:
Section heights are a little uneven with the new link in place:

spacing

craigcook:
Fixed the height issue with a touch of flexbox. Any browsers that don't support flexbox but do support the media query will get the basic floats and still show a white gap, but I think we can live with that. Also tidied up the CSS a bit, addressed the linting errors, and renamed `#who` to `#impact` to better match the content.

I can explore ways to optimize the animation a bit but I don't want to hold up this PR. I don't see any jankiness on my machine but I don't doubt it exists. It's a transition on the `transform` property and I wonder if something else might be less paint-intensive.

media/css/mozorg/home/home.scss
```diff
@@ -397,13 +388,14 @@ main {
     }
 }
 
-// float layout for wider viewports
+// flex/float layout for wider viewports
 @media #{$mq-tablet} {
-    #who-innovate-wrapper {
+    #impact-innovate-wrapper {
         @include clearfix;
+        @include flexbox;
```
alexgibson: This is totally optional, but would another solution to be to use `display: table;` here and then `display: table-cell` for each column (dropping the `float`s)? This may arguably be a little dirty, but it has much wider support than Flexbox?

craigcook: Good idea, wish I'd thought of it! Changed to fake table layout, which should work in IE9/10 as well.

alexgibson:
Looks good, I'll merge once the pipes are green 👍

PR #4627, opened 2017-02-02
Add "declaration-block-semicolon-newline-after" rule to Stylelint

heatherbooker:
## Description
As mentioned in #4610 (and #4607), Stylelint could use a rule to ensure newlines between CSS declarations. This change adds the rule "declaration-block-semicolon-newline-after" ([view on Stylelint](https://stylelint.io/user-guide/rules/declaration-block-semicolon-newline-after/)) with value "always -multi-line".
Ie, with this new rule, now this:
```less
#swatch_f4f4f4 .swatch span {
    background-color: #f4f4f4; color: #000;
}
 ```
will trigger a warning; and this:
```less
#swatch_f4f4f4 .swatch span {
    background-color: #f4f4f4;
    color: #000;
}
```
will not.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -18,6 +18,7 @@
     "declaration-block-no-shorthand-property-overrides": true,
     "declaration-block-single-line-max-declarations": [1, { "severity": "warning" }],
     "declaration-block-trailing-semicolon": ["always", { "severity": "warning" }],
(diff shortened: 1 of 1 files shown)


alexgibson:
Thanks for this, looks great! 👍 

I think we may need to change the severity to a warning for the time being?

heatherbooker:
Awesome, I changed it to be just a warning! :+1:

PR #4630, opened 2017-02-02
Update the stub-attribution-code view to use a 5 min cache header

pmac:
The data includes a timestamp, so we can't cache for long.


pmac:
The stub service has a 10min validation of the timestamp produced, so it can't be cached for long.

PR #4638, opened 2017-02-07
Jenkinsfile deployment improvement

pmac:
* Use the full app URL with suffix for integration tests
* Better failure message for deployment testing stage
* Fix demo deploy success notification
* Output usable URL for stage and prod deis apps during deployment
* Deploy IRC notifications use BlueOcean build URLs
* Only push requested apps to private registry

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -114,7 +114,7 @@ if ( config.branches.containsKey(env.BRANCH_NAME) ) {
             node {
                 unstash 'scripts'
                 try {
-                    utils.pushPrivateReg(region.registry_port)
+                    utils.pushPrivateReg(region.registry_port, branchConfig.apps)
                 } catch(err) {
                     utils.ircNotification(config, [stage: stageName, status: 'failure'])
                     throw err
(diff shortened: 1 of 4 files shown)


pmac:
@jgmize improvements, including the change to use the `-deis` suffix URLs.

Jenkinsfile
```diff
@@ -114,15 +114,15 @@ if ( config.branches.containsKey(env.BRANCH_NAME) ) {
...
+                    utils.pushPrivateReg(region.registry_port, branchConfig.apps)
                 } catch(err) {
                     utils.ircNotification(config, [stage: stageName, status: 'failure'])
                     throw err
                 }
             }
         }
         for (appname in branchConfig.apps) {
-            def appURL = "https://${appname}.${region.name}.moz.works"
+            def appURL = "https://${appname}${branchConfig.app_name_suffix}.${region.name}.moz.works"
```
jgmize: The "-deis" `app_name_suffix` isn't globally required, and in fact in the two places it is required it's due to an artifact of a past misunderstanding and should probably be fixed. There are some nuances there that are probably best discussed over a higher bandwidth channel though.

PR #4642, opened 2017-02-07
Switch to UglifyJS and CleanCSS compressors

pmac:


pmac:
Been meaning to do this for a while. Will hopefully help @craigcook with some CSS minification issues he's having. It'll be on [my demo](https://bedrock-demo-pmac.us-west.moz.works) shortly.

pmac:
This has been rebased and run through deployment and integration testing on the [integration tests app](https://bedrock-integration-tests.us-west.moz.works/).

@jpetto r?

jpetto:
Actually, one consistency question - should we get file names/namespaces coordinated between `pipeline_compressors.py` and `storage.py`?

Definitely not a blocker.

pmac:
That's a good point. I used the longer `pipeline_compressors.py` because trying to use `compressors.py` resulted in some import errors for things that import things from just `compressors`, and then using the wrong code. I think changing `storage.py` to match would be a good idea to avoid future issues. I'll do that now.

pmac:
@jpetto updated.

PR #4579, opened 2017-01-12
open iframe on first run, sign in, animate

ericawright:
**work-in-progress**
@pmac 

## Description
Add animation to onboarding experience/ firstrun page.
Change sign-up form to sign-in instead
Animation drops user off on `about:newtab` page

Based on https://dl.dropboxusercontent.com/u/105549/mvp1.framer/index.html

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1322718
https://bugzilla.mozilla.org/show_bug.cgi?id=1333410

PR #4643, opened 2017-02-07
[fix bug 1337448] FxA redirect after login.

jpetto:
## Description

#4579 [introduced a regression](https://github.com/mozilla/bedrock/pull/4579/files#diff-dc9018ca2e4502c42779505aa8eeaea4L181) whereby users logging in to the FxA iframe stopped being redirected to the FxA settings page. This resulted in a significant uptick in errors on the FxA side (which is how the bug was discovered), as well as poor UX (the user saw a "Working..." error message in the iframe that never went away).

This PR fixes that regression, and adds tests.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1337448

## Testing

On demo3, using a profile configured to work with the FxA staging server, log in to Sync on the current firstrun page [1]. You should be redirected to the FxA accounts settings URL.

Also on demo3, using the same profile as above, log in to Sync on the soon to be released funnelcake firstrun page [2]. You should be redirected to the new tab view.

(Working with FxA on a demo is detailed in [the docs](http://bedrock.readthedocs.io/en/latest/firefox-accounts.html).)

[1] https://www-demo3.allizom.org/en-US/firefox/51.0/firstrun/
[2] https://www-demo3.allizom.org/en-US/firefox/51.0/firstrun/?f=99

## Checklist

### DIFF
--- a/media/js/base/global.js
+++ b/media/js/base/global.js
@@ -3,7 +3,7 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 /* exported triggerIEDownload, initDownloadLinks, initMobileDownloadLinks,
-   maybeSwitchToDistDownloadLinks, initLangSwitcher */
+   maybeSwitchToDistDownloadLinks, initLangSwitcher, doRedirect */
 
 // download buttons
 
@@ -79,6 +79,13 @@ function initLangSwitcher() {
         $('#lang_form').attr('action', window.location.hash || '#');
         $('#lang_form').submit();
     });
+}
+
+// client-side redirects (handy for testing)
+function doRedirect(destination) {
+    if (destination) {
+        window.location.href = destination;
+    }
 }
 
 // Create text translation function using #strings element.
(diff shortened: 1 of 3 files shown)


media/js/base/global.js
```diff
@@ -81,6 +81,13 @@ function initLangSwitcher() {
     });
 }
 
+// client-side redirects (handy for testing)
+function doRedirect(destination) {
```
jpetto: This function helps us write unit tests against scripts that perform a redirect. Might be worth using this function in other places where we currently do `window.location.href = `, but that's another PR.

jpetto:
@schalkneethling - As Alex is out this week, and this is a fairly important fix, can you find time to review? Thanks!

cc @shane-tomlinson

schalkneethling:
@jpetto Code looks good, js test all pass, and tested the redirection on demo3. r+

shane-tomlinson:
@jpetto, @schalkneethling - you guys are awesome. Thanks!

PR #4644, opened 2017-02-08
Use demo app alias URLs in IRC notifications

pmac:
e.g. www-demo1.allizom.org


pmac:
Verified this works by pushing demos. Merging.

PR #4645, opened 2017-02-08
[fix bug 1329642] A/B test Rust video embed at top of Web Innovations page

craigcook:
## Description
Creates a variant of the /technology page showcasing a Rust promo video in the page header. This will be an A/B test with 50% of en-US. English only, especially since the video is in English, so no l10n.

Uses the switch `experiment-technology-rust-video` to enable the traffic cop.

## Bugzilla link

## Testing
Add `?v=b` to the URL to see the variant page.
https://bedrock-demo-craigcook.us-west.moz.works/en-US/technology/?v=b

## Checklist
- [ ] Requires l10n changes.


media/js/mozorg/technology-rust-video-test.js
```diff
@@ -0,0 +1,80 @@
...
+    var firstScriptTag = document.getElementsByTagName('script')[0];
+    firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
+
+    function onYouTubeIframeAPIReady() {
+        // show video modal when user clicks "watch now" button
+        $('.hero-rust-video a.video-play').attr('role', 'button').click(function(e) {
+            e.preventDefault();
+
+            var $videoContainer = $('#video-container');
+            var video = $videoContainer.find('.video')[0];
```
jpetto: A pretty easy performance improvement would be moving these two vars to just before `function onYouTubeIframeAPIRead()`. As is, the DOM is re-queried each time the user clicks the button (which admittedly will probably only be once per visit, but it's a simple copy pasta).

media/css/mozorg/technology.scss
```diff
@@ -239,6 +239,138 @@ main {
     }
 }
 
+
+/*--------------------------------------------------------------------------*/
+// Rust video header
+.hero-rust-video {
+    background: #000 url('/media/img/mozorg/technology/icon-rust-hero.svg') -25% 30px no-repeat;
+    @include background-size(500px, 500px);
```
craigcook: Huh, somehow I got it in my head that the mixin requires a comma... will fix.

PR #4651, opened 2017-02-09
Remove /firefox/panorama redirect and test

jgmize:
Target is now 404


pmac:
Looks like there's another test for this in `map_globalconf.py` :( That's the test that failed.

jgmize:
> Looks like there's another test for this in map_globalconf.py :( That's the test that failed.

oops. amended to remove the other test.

PR #4653, opened 2017-02-09
Change how we do docker

pmac:
* More simple and streamlined base image build file
* Use a builder image to decrease the size and security footprint of running containers
* Make demos more prod-like
* Add l10n for prod builds via docker commit
* Add demo data via docker commit
* No longer include the media directory in production images
* Specifically include only those files required for the running site


pmac:
Not yet producing usable images, as you can see by my attempts in the pipeline. I've been pushing this to the `deploy-via-jenkinsfile` branch so that the pipeline will run. There are still some pages throwing 500s I believe due to missing files. On the plus side this is letting me fix issues with code referencing files in the `media` dir instead of `static`. But a review can commence as I believe these things will be minor, and if there's a larger issue with the direction of this then I'd like to know sooner than later.

pmac:
@jgmize things are now changed again and simplified. I'm still getting some errors, so I need to tune further, but the approach should be stable.

pmac:
Turns out Django is not really cool with not having the source media directories around, so I've switched back to including `media` in the image and using the `--link` switch for `collectstatic`. Shouldn't be much different in size to not including media, but that's the deal with the pages that I mentioned last week were broken. I believe this is now just about ready for merge. I'll squash commits in prep.

pmac:
@jgmize okay. ready for real now I think. I'd appreciate a look by you as well @glogiotatidis. I think this will be a real win, however there is no squashing at all right now. Please let me know if I should add some back, and if so where.

Jenkinsfile
```diff
@@ -39,12 +40,13 @@ if ( config.branches.containsKey(env.BRANCH_NAME) ) {
                 }
             }
             utils.ircNotification(config, [stage: 'Test & Deploy', status: 'starting'])
-            try {
-                utils.buildDockerImage(dockerfile: 'bedrock_base', update: true)
-                utils.buildDockerImage(dockerfile: 'bedrock_code', fromDockerfile: 'bedrock_base')
-            } catch(err) {
-                utils.ircNotification(config, [stage: 'Docker Build', status: 'failure'])
-                throw err
+            lock ("bedrock-docker-${env.GIT_COMMIT}") {
```
pmac: This is to make sure that the same commit pushed to 2 (or more) branches is only built once. This shouldn't happen often of course, but it does allow us to push `master`, then tag and push `prod` without much fear.

jenkins.yml
```diff
@@ -18,6 +21,8 @@ branches:
       - bedrock-stage
       - bedrock-prod
     integration_tests:
+      - headless
```
pmac: After a discussion with @alexgibson he showed me that we used to run the firefox and headless tests against production as well. This adds that back.

jenkins.yml
```diff
@@ -1,3 +1,6 @@
+environment:
+  SELENIUM_VERSION: 2.52.0
```
pmac: We now have the ability to define arbitrary environment variables in the YAML just like the fancy hosted services! Take that "so called" Circle CI 😝

Jenkinsfile
```diff
@@ -39,12 +40,13 @@ if ( config.branches.containsKey(env.BRANCH_NAME) ) {
...
+                try {
+                    sh 'docker/jenkins/build_images.sh --prod --test'
```
metadave: naive question: are both `--prod` and `--test` supposed to be specified here?

pmac: yes. what those switches really mean are "build the prod code image" and "build the testing image", which are separate things. There's also a `--demo` flag that will build the image suitable for deploying to a demo having synced all the external data to a squlite db.

docker/jenkins/build_images.sh
```diff
@@ -0,0 +1,83 @@
...
+set -exo pipefail
+
+if [[ -z "$GIT_COMMIT" ]]; then
+  GIT_COMMIT=$(git rev-parse HEAD)
+fi
+
+BUILD_IMAGE_TAG="mozorg/bedrock_build:${GIT_COMMIT}"
+CODE_IMAGE_TAG="mozorg/bedrock_code:${GIT_COMMIT}"
+DOCKER_REBUILD=false
+# dev mode will build the dev image containing a db file full of data
```
jgmize: s/dev/demo/g

docker/dockerfiles/bedrock_l10n
```diff
@@ -1,5 +1,8 @@
-FROM ${FROM_DOCKER_REPOSITORY}:${GIT_COMMIT}
-COPY . /app/locale
+FROM mozorg/bedrock_code:${GIT_COMMIT}
+
+COPY ./locale ./locale
+
+# Change User
 USER root
-RUN chown -R webdev.webdev /app/locale
+RUN chown webdev.webdev -R .
```
jgmize: please change this back as discussed

pmac:
Something is [failing](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/deploy-via-jenkinsfile/68/pipeline). I'm investigating.

pmac:
@jgmize we tracked the problem down to the Jenkins restart and the need to update the allowed scripts after that. Giorgos said he's working on a way to have that happen automatically.

So we're unblocked and things are working well!

* [prod-like deploy](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/deploy-via-jenkinsfile/71/pipeline/)
* [demo deploy](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/demo__pmac-jenkins/10/pipeline)

pmac:
@alexgibson pointed out that we're no longer running the integration smoke tests before deployment. I've added that back as part of this PR and am testing it now.

pmac:
@jgmize okay. [ready again](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/deploy-via-jenkinsfile/78/pipeline). now runs the smoke tests like it should always have been.

pmac:
Turns out we didn't need a separate integration_tests image since the image for running the unit tests was a superset of that one. I'm now using the bedrock_test image for running both the local docker selenium tests as well as the unit tests and other checks.

Jenkinsfile
```diff
@@ -1,6 +1,6 @@
 #!groovy
 
-@Library('github.com/mozmar/jenkins-pipeline@20170126.1')
+@Library('github.com/mozmar/jenkins-pipeline@15a2f1aeb08e58e2ee325c072b596d198771ccf9')
```
jgmize: Out of curiosity, why the switch from the dated tag to the hash?

pmac:
I also now have cleanup for the running docker selenium and bedrock instances added back.

Jenkinsfile
pmac: There just wasn't a dated tag for the version I wanted yet. I can tag now and switch to it.

pmac: Tagged and updated.

PR #4656, opened 2017-02-09
[bug 1337446] Add common js dependency to thunderbird start page for GA 

Sancus:
…nd add to tb start page

## Checklist
- [ ] Requires l10n changes.


Sancus:
http://i.imgur.com/0adNLVZ.png

Confirm this correctly pings GA servers on https://bedrock-demo-craigcook.us-west.moz.works/en-US/thunderbird/release/start/

craigcook:
So this little adventure has brought to light how our GA bits are split across two JavaScript bundles and also that `core-datalayer-init.js` seems to require jQuery for no particular reason. I think we could combine all the necessary scripts into one "google-analytics" bundle that wouldn't depend on anything else.

PR #4664, opened 2017-02-13
Hide and show skip button for new /firstrun

ericawright:
@shane-tomlinson @craigcook

## Description

- the skip button fades on the form in the iframe being engaged, and solidifies on it being unengaged. 

- the skip button dissapears altogether one a form has been submitted
## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1338348

### DIFF
--- a/media/js/firefox/firstrun/new-firstrun.js
+++ b/media/js/firefox/firstrun/new-firstrun.js
@@ -9,8 +9,22 @@
         var scene = document.getElementById('scene');
         var skipbutton = document.getElementById('skip-button');
 
-        var hideSkipButton = function () {
-            skipbutton.style.display = 'none';
+        var hideOrShowSkipButton = function (data) {
+            switch(data.data.url) {
+            case 'signin':
+            case 'signup':
+            case 'reset_password':
+                skipbutton.disabled = false;
+                skipbutton.classList.remove('skipbutton-hidden');
+                break;
+            default:
+                skipbutton.classList.add('skipbutton-hidden');
+                break;
+            }
+        };
+
+        var disableSkipButton = function () {
+            skipbutton.disabled = true;
         };
 
         var onVerificationComplete = function () {
@@ -32,7 +46,8 @@
                 gaEventName: 'firstrun-fxa',
                 onVerificationComplete: onVerificationComplete,
                 onLogin: onVerificationComplete,
-                onSignupMustVerify: hideSkipButton
+                onFormEnabled: disableSkipButton,
+                onNavigated:  hideOrShowSkipButton
             });
         });
 
--- a/media/css/firefox/firstrun/new-firstrun.less
+++ b/media/css/firefox/firstrun/new-firstrun.less
@@ -144,10 +144,22 @@ img {
         height: 24px;
         padding: 3px 10px;
 
-        &:hover {
+        &:not([disabled]):hover {
             background-color: #0670cc;
             border-color: #005bab;
         }
+    }
+
+    button[disabled] {
+      background-color: #ebebeb;
+      border-color: #b1b1b1;
+      color: #6a6a6a;
+      opacity: .5;
+      cursor: default;
+    }
+
+    .skipbutton-hidden {
+      display: none;
     }
 
     #sunrise {
--- a/media/js/base/mozilla-fxa-iframe.js
+++ b/media/js/base/mozilla-fxa-iframe.js
@@ -129,6 +129,18 @@ Mozilla.FxaIframe = (function() {
         case 'resize':
             _onResize(data);
             break;
+        // user has entered email address and password
+        case 'form_enabled':
+            _onFormEnabled(data);
+            break;
+        // user has deleted email address or password
+        case 'form_disabled':
+            _onFormDisabled(data);
+            break;
+        // user has chnaged pages within the iframe
+        case 'navigated':
+            _onNavigated(data);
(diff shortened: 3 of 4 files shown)


shane-tomlinson:
@ericawright - It looks good to me, the event names used in [the corresponding fxa-content-server PR](https://github.com/mozilla/fxa-content-server/pull/4717#event-960227857). We are now waiting for testing/deployment from Ops.

shane-tomlinson:
@jpetto - would you be the correct person to review this PR?

jpetto:
I certainly can. I'll review tomorrow morning if still needed.

shane-tomlinson:
Thanks @jpetto! @ericawright, @jpetto - the content server patch needed to support this (https://github.com/mozilla/fxa-content-server/issues/4717) is now running in production thanks to @jrgm.

media/js/firefox/firstrun/new-firstrun.js
```diff
@@ -9,8 +9,31 @@
...
+            skipbutton.style.borderColor = '#0670cc';
+            skipbutton.style.color = '#fff';
+            skipbutton.style.opacity = '1';
+        };
+
+        var disableSkipButton = function () {
+            skipbutton.style.backgroundColor = '#ebebeb';
+            skipbutton.style.borderColor = '#b1b1b1';
+            skipbutton.style.color = '#6a6a6a';
+            skipbutton.style.opacity = '.5';
```
jpetto: To keep concerns separated, let's use CSS classes to change the appearance of the button instead of manipulating with JS.

ericawright:
@jpetto thank you, I copied the other `postMessage` tests - hopefully that was what you're lookinging for.

media/js/firefox/firstrun/new-firstrun.js
```diff
@@ -9,8 +9,27 @@
...
         var skipbutton = document.getElementById('skip-button');
 
-        var hideSkipButton = function () {
-            skipbutton.style.display = 'none';
+        var hideOrShowSkipButton = function (data) {
+            switch(data.data.url) {
+            case 'signin':
+            case 'signup':
+            case 'reset_password':
+                skipbutton.removeClass('skipbutton-hidden');
```
jpetto: This throws an error, as `removeClass` is a jQuery function. You could either change `skipbutton` to a jQuery object (which would require a bit more of a refactor), *or*, since `firstrun` pages are basically guaranteed to run on a modern version of Firefox, you could use the [classList API](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList). (I'd go for the latter 😉.)

ericawright: whoops - thank you

jpetto:
Hey @ericawright  - Let me know if you're too busy to finish this one up. I think we want to get it out fairly soon. Thanks!

ericawright:
@jpetto I fixed it yesterday evening - sorry I thought it would notify you

media/js/firefox/firstrun/new-firstrun.js
jpetto: Still need to fix this `removeClass`, and the `addClass` on line 20 below.

ericawright:
hold on  I missed some

jpetto:
Okay, so the code looks good, but there appears to be a bit of a gap in the handling. If I attempt to log in or create an account, as soon as I hit the button (Sign In or Create account), the "skip" button again becomes active. This doesn't break anything functionally, but does make for an odd UI.

Additionally, if I click "Forgot password?" and enter an email, but then click "Remember password? Sign in.", the "skip" button remains in the disabled state.

Hopefully these cases can be addressed with the existing postMessage capabilities of the iframe.

ericawright:
I see what you mean. Yes, I can take care of that

ericawright:
@jpetto unfortunately there isn't any data coming through about which element is interacted with. I was hoping `form_engaged` might either send element data or only trigger from text input - but it only sends a message on the first interaction with a text element. So, no I can't see a way to stop the button from being visually activated again on submission...

jpetto:
> @jpetto unfortunately there isn't any data coming through about which element is interacted with. I was hoping form_engaged might either send element data or only trigger from text input - but it only sends a message on the first interaction with a text element. So, no I can't see a way to stop the button from being visually activated again on submission...

I just [commented in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1338348#c3) to see what folks would like to do here. I think we need more info sent along with the `form_disabled` postMessage so we know what action to take.

ericawright:
@jpetto After some discussion with Michael Verdi we have decided to not re-enable the button after the user removes their info - this should stop the flicker on submit. 
additionally - the button will be actually disable not just visually

jpetto:
Okay - I'll watch for your next push. 👍

ericawright:
@jpetto changes have been made

jpetto:
> note: I left in the tests and code for onFormEnabled even though I am no longer using it - should these be removed?

If it's still a valid postMessage available from the FxA iframe, let's leave the code & tests in place.

jpetto:
Updated [demo3](https://www-demo3.allizom.org/en-US/firefox/51.0/firstrun/?f=99) and all is functioning as expected. The skip button is now slightly aggressively disabled (e.g. switching from create account to sign in), but I think we're covered for the most common use cases.

PR #4665, opened 2017-02-13
Fix Bug 1336042 - Stop presenting 32 and 64 bit version on the same level on the download page

kyoshino:
## Description

Organize platform labels on the /all/ download pages, so that 64-bit builds will be more prominent.

## Bugzilla link

[Bug 1336042](https://bugzilla.mozilla.org/show_bug.cgi?id=1336042)

## Testing

Check the /all/ download pages for each channel and make sure the labels are displayed as:

> Windows 64-bit | Windows 32-bit | macOS | Linux 64-bit | Linux 32-bit


kyoshino:
The tests are failing because the order of download buttons has also changed. By the way, we should probably figure out first [how to offer the winsha1 builds](https://bugzilla.mozilla.org/show_bug.cgi?id=1339226) since it's no longer supported by Firefox 53+.

alexgibson:
> By the way, we should probably figure out first how to offer the winsha1 builds since it's no longer supported by Firefox 53+.

Bug 1339226 is being handled in bouncer, so I don't think it needs to block this.

alexgibson:
Table headings are rather difficult to read at smaller viewport sizes:

table-headings

alexgibson:
ping @kyoshino - any updates on this one?

kyoshino:
The win64 auto migration begins today. Will revisit this ASAP.

kyoshino:
@alexgibson Sorry I'm super late here! This is now simply reordering the columns so 32-bit and 64-bit can be clearly distinguished. I've also combined the language columns to make more room for other columns. So now the table has 5 columns in this order, aside from Language:

> Windows 64-bit | Windows 32-bit | macOS | Linux 64-bit | Linux 32-bit

PR #4667, opened 2017-02-13
Allow user to optionally specify duration of Traffic Cop cookie.

jpetto:
## Description

MDN is *probably* going to use Traffic Cop for an upcoming test, and will need the ability to have longer lasting cookies. This is in preparation for that use-case.

## Bugzilla link

Nope.

## Testing

Make sure tests pass? Could also create a local example and ensure cookie remains after closing the tab.

## Checklist

### DIFF
--- a/docs/mozilla-traffic-cop.rst
+++ b/docs/mozilla-traffic-cop.rst
@@ -11,8 +11,10 @@ Mozilla Traffic Cop
 ``mozilla-traffic-cop.js`` handles redirecting users to different A/B/x
 variations through a query parameter.
 
-**``mozilla-traffic-cop.js`` requires ``mozilla-cookie-helper.js``, which is
-included as default in the ``site.js`` bundle for most pages.**
+.. note::
+
+    ``mozilla-traffic-cop.js`` requires ``mozilla-cookie-helper.js``, which is
+    included by default in the ``site.js`` bundle for most pages.
 
 How It Works
 ------------
@@ -31,8 +33,9 @@ will choose the appropriate variation and redirect the user.
 If the percentage exceeds that of the config, no redirect will occur. (This
 could be considered the "control".)
 
-If a variation is chosen, a cookie is set (for the session only) that will
-send the user back to the same variation if/when the page is again visited.
+If a variation is chosen, a cookie is set that will send the user back to the
+same variation if/when the page is again visited. (See below for setting cookie
+duration.)
 
 Any query string parameters present when a user initially lands on a page will be
 propagated to the variation redirect.
@@ -45,9 +48,11 @@ Each instance of the Traffic Cop requires two pieces of configuration:
 - A string ID that is unique to other currently running tests (to avoid confusion when reading cookies)
 - A variations object that lists all variations by query string along with the associated visitor percentage
 
-An implementation might look like::
+An implementation might look like:
 
-    var cop = new Mozilla.TrafficCop({
+.. code-block:: javascript
+
+    var eddie = new Mozilla.TrafficCop({
         id: 'exp_firefox_new_headline',
         variations: {
             'v=1': 25,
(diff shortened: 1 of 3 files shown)


docs/mozilla-traffic-cop.rst
```diff
@@ -56,20 +61,44 @@ An implementation might look like::
...
 In the above example, the string "exp_firefox_new_headline" will be used as the
 cookie name should a variation be chosen.
 
 The test will have 3 variations, each targeting 25% of users.
 
+There is also an optional configuration value to specify how long the cookie
+associated with a visitor for an individual experiment will last:
+``cookieExpires``. This value must be a ``Number`` and represents the number of
+hours the cookie will last. If omitted, the cookie will only last for the
+current browsing session.
```
alexgibson: This doesn't need to be part of this change, but i do wonder if defaulting to a session cookie is actually the best option for the majority of experiments. A session cookie could last for a long time potentially, especially if a user rarely restarts their browser which can be quite a common behavior. Defaulting to 24 hours could perhaps be a better limit?

docs/mozilla-traffic-cop.rst
```diff
@@ -56,20 +61,44 @@ An implementation might look like::
...
 -------------------------
 
 Variations are sorted in the order provided, and percentages are tallied to
-create tiers. Take the following config::
+create tiers. Take the following config:
+
+.. code-block:: javascript
 
-    var cop = new Mozilla.TrafficCop({
+    var rex = new Mozilla.TrafficCop({
```
alexgibson: I got `eddie` and `lou`, but I'm lost on `rex`? 😄

tests/unit/spec/base/mozilla-traffic-cop.js
```diff
@@ -118,6 +118,54 @@ describe('mozilla-traffic-cop.js', function () {
...
+        it('should return null when no hours are specified', function() {
+            var cop = new Mozilla.TrafficCop(config);
+            expect(cop.cookieExpiresDate()).toBeFalsy();
+        });
+
+        it('should return a date when hours are specified', function() {
+            config.cookieExpires = 72;
+            var cop = new Mozilla.TrafficCop(config);
+            expect(cop.cookieExpiresDate()).toBeTruthy();
+            expect(Object.prototype.toString.call(cop.cookieExpiresDate())).toEqual('[object Date]');
```
alexgibson: If we made it possible to pass in an optional `date` parameter to `cookieExpiresDate()` for testing, would this make it easier to assert that the returned value is what we expect it to be given the value in `config`?

tests/unit/spec/base/mozilla-traffic-cop.js
```diff
@@ -118,6 +118,54 @@ describe('mozilla-traffic-cop.js', function () {
...
+            id: 'test123',
+            variations: {
+                'v=1': 40,
+                'v=2': 30
+            }
+        };
+
+        it('should return null when no hours are specified', function() {
+            var cop = new Mozilla.TrafficCop(config);
+            expect(cop.cookieExpiresDate()).toBeFalsy();
```
alexgibson: Would it be better to assert `null` here? A falsy value could also be `0` for example, which could result in a cookie that expired immediately?

docs/mozilla-traffic-cop.rst
jpetto: I ate some yogurt while looking out the window and thought about this. I think 24 hours is the right default. It's long enough that the user (hopefully) wouldn't be surprised (or even notice) if they got a different variation of a page, but not so long that they'll see the same variation for the life of the experiment.

docs/mozilla-traffic-cop.rst
jpetto: Rex Banner is a [deep cut](https://www.youtube.com/watch?v=YzSgzHu4Q5g).

tests/unit/spec/base/mozilla-traffic-cop.js
jpetto: Good call.

tests/unit/spec/base/mozilla-traffic-cop.js
jpetto: Oh, most definitely. Good suggestion.

tests/unit/spec/base/mozilla-traffic-cop.js
```diff
@@ -118,6 +118,81 @@ describe('mozilla-traffic-cop.js', function () {
...
+            variations: {
+                'v=1': 40,
+                'v=2': 30
+            }
+        };
+
+        it('should return a date 24 hours into the future when no hours are specified', function() {
+            var cop = new Mozilla.TrafficCop(config);
+            var expiry = cop.cookieExpiresDate(new Date(2017, 0, 1, 12, 30));
+            expect(expiry).not.toBe(null);
```
alexgibson: Given that we're testing the date below, I don't think this assert seems that useful?

tests/unit/spec/base/mozilla-traffic-cop.js
```diff
@@ -118,6 +118,81 @@ describe('mozilla-traffic-cop.js', function () {
...
+            expect(expiry.getFullYear()).toBe(2017);
+            expect(expiry.getMonth()).toBe(0);
+            expect(expiry.getDate()).toBe(2); // 24 hours default cookie length
+        });
+
+        it('should return an appropriate future date when hours are specified', function() {
+            config.cookieExpires = 72;
+            var cop = new Mozilla.TrafficCop(config);
+            var expiry = cop.cookieExpiresDate(new Date(2017, 0, 1, 12, 30));
+            expect(expiry).not.toBe(null);
```
alexgibson: Likewise I think we can remove this one too.

tests/unit/spec/base/mozilla-traffic-cop.js
```diff
@@ -118,6 +118,81 @@ describe('mozilla-traffic-cop.js', function () {
...
+            expect(Object.prototype.toString.call(expiry)).toEqual('[object Date]');
```
alexgibson: We can probably omit this assert too, given the date checks below?

jpetto: Heh, yeah, I was writing in implementation mode instead of testing mode. Removed. 👍

PR #4668, opened 2017-02-14
[fix bug 1336150] Remove old home page template

alexgibson:
## Description
- Removes old 2015 home page, now that the majority of locales (58) are on the new page.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1336150

## Testing
- Make sure I didn't miss any left over assets, or delete anything that's still needed?

## Checklist
- [x] Requires l10n changes.


bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -9,8 +9,6 @@
 
 {% block gtm_page_id %}data-gtm-page-id="Homepage"{% endblock %}
 
-{% block page_image %}{{ static('img/home/page-image.png') }}{% endblock %}
```
alexgibson: I'm removing this as the image is of the old home page design. It will now fall back to [this image](https://www.mozilla.org/media/img/mozorg/mozilla-256.jpg), which is also out of date as it's the old logo. We should probably file a bug to update this separately.

bedrock/mozorg/views.py
```diff
@@ -251,20 +251,8 @@ def home_tweets(locale):
...
-        template = 'mozorg/home/home-voices.html'
-
-    return l10n_utils.render(
-        request, template, {
-            'has_contribute': lang_file_is_active('mozorg/contribute'),
-            'tweets': home_tweets(locale),
-            'mobilizer_link': settings.MOBILIZER_LOCALE_LINK.get(
-                locale, settings.MOBILIZER_LOCALE_LINK['en-US'])})
+def home(request):
+    return l10n_utils.render(request, 'mozorg/home/home.html')
```
alexgibson: I thought about removing this view and turning the URL back to a regular `page` rule, but that would mean updating all the internal home page links on other pages, of which there are a shit ton. I decided it wasn't worth the effort, but can be persuaded otherwise.

media/css/tabzilla/tabzilla.less
```diff
@@ -599,7 +599,7 @@ html[dir="rtl"] {
     height: 150px;
     color: #fff;
     background-color: #95368c;
-    background-image: url(/media/img/home/voices/promos/gear/gear-high-res.png);
+    background-image: url(/media/img/tabzilla/gear-high-res.png);
```
alexgibson: It looks like the old tabzilla gear store promo was referencing an image from the old home page (yikes). I moved the image to a different directory.

bedrock/mozorg/views.py
craigcook: Also given how often the home page changes with regular A/B tests, plus another probably-imminent redesign, we'd only have to change it back to a view soon anyway. Best to leave it.

alexgibson:
@craigcook - it would be great of we could get this merged before the site wide footer changes. One less thing to pick apart.

PR #4671, opened 2017-02-16
[fix bug 1340087] Update funnelcake logic to include IDs for windows only

alexgibson:
## Description
- Updates bedrock's logic for including funnelcake query params in download links, so it now only happens for Windows and en-US. This avoids scenarios where the download could 404 for other platforms, should anyone come across a link like below:

https://www.mozilla.org/en-US/firefox/new/?f=99

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1340087

## Testing
- I added some unit tests, but this could still use some careful review.
- Demo URL: https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/?f=99

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -209,6 +209,10 @@ class FirefoxDesktop(_ProductDetails):
         _locale = 'ja-JP-mac' if platform == 'osx' and locale == 'ja' else locale
         _platform = 'win' if platform == 'winsha1' else platform
 
+        # Bug 1340087 - Only include funnelcake params for Windows 32bit en-US builds by default.
+        include_funnelcake_param = (funnelcake_id and _platform in settings.FUNNELCAKE_PLATFORMS and
+                                    _locale in settings.FUNNELCAKE_LOCALES)
+
         # Nightly and Aurora have a special download link format
         # see bug 1324001
         if channel in ['alpha', 'nightly']:
@@ -246,13 +250,13 @@ class FirefoxDesktop(_ProductDetails):
                 suffix = 'latest'
 
             _version = ('beta-' if channel == 'beta' else '') + suffix
-        elif not funnelcake_id:
+        elif not include_funnelcake_param:
             # Force download via SSL. Stub installers are always downloaded via SSL.
             # Funnelcakes may not be ready for SSL download
             _version += '-SSL'
 
         # append funnelcake id to version if we have one
-        if funnelcake_id:
+        if include_funnelcake_param:
             _version = '{vers}-f{fc}'.format(vers=_version, fc=funnelcake_id)
 
         # Check if direct download link has been requested
--- a/bedrock/firefox/tests/test_firefox_details.py
+++ b/bedrock/firefox/tests/test_firefox_details.py
@@ -399,6 +399,58 @@ class TestFirefoxDesktop(TestCase):
         """latest_major_version should return 0 when no int."""
         eq_(firefox_desktop.latest_major_version('release'), 0)
 
+    @override_settings(FUNNELCAKE_LOCALES=('en-US',), FUNNELCAKE_PLATFORMS=('win',))
+    def test_funnelcake_direct_links_en_us_win_only(self):
+        """
+        Ensure funnelcake params are included for Windows en-US builds only.
+        """
+        url = firefox_desktop.get_download_url('release', '45.0', 'win', 'en-US', force_direct=True, funnelcake_id='64')
+        ok_('product=firefox-stub-f64' in url)
+
+        url = firefox_desktop.get_download_url('release', '45.0', 'win', 'en-US', force_direct=True, force_full_installer=True, funnelcake_id='64')
+        ok_('product=firefox-45.0-f64' in url)
+
+        url = firefox_desktop.get_download_url('release', '45.0', 'win', 'en-US', force_direct=True, force_funnelcake=True, funnelcake_id='64')
+        ok_('product=firefox-latest-f64' in url)
+
+        url = firefox_desktop.get_download_url('release', '45.0', 'win', 'de', force_direct=True, funnelcake_id='64')
+        ok_('product=firefox-45.0-f64' not in url)
+
+        url = firefox_desktop.get_download_url('release', '45.0', 'osx', 'en-US', force_direct=True, funnelcake_id='64')
+        ok_('product=firefox-45.0-f64' not in url)
+
+    @override_settings(FUNNELCAKE_LOCALES=('en-US', 'de'), FUNNELCAKE_PLATFORMS=('linux', 'osx'))
+    def test_funnelcake_direct_links_locales_linux_osx_(self):
+        """
+        Ensure funnelcake params are included for Linux and OSX en-US builds only.
+        """
+        url = firefox_desktop.get_download_url('release', '45.0', 'win', 'en-US', force_direct=True, funnelcake_id='64')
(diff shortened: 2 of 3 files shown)


bedrock/firefox/firefox_details.py
```diff
@@ -246,13 +249,13 @@ def get_download_url(self, channel, version, platform, locale,
...
             _version = ('beta-' if channel == 'beta' else '') + suffix
-        elif not funnelcake_id:
+        elif not _include_funnelcake_param:
             # Force download via SSL. Stub installers are always downloaded via SSL.
             # Funnelcakes may not be ready for SSL download
             _version += '-SSL'
 
         # append funnelcake id to version if we have one
-        if funnelcake_id:
+        if _include_funnelcake_param:
```
alexgibson: Note: I'm only changing the logic for links where `force_direct` is `True`. I could also potentially apply this logic for transitional links below (e.g. `/firefox/new/?scene=2`), so it would only be `/firefox/new/?scene=2&f=99` for Windows. Thoughts?

I decided against doing this initially since we're running an experiment that shows a new design for the download page based on funnelcake params. If someone visits `/firefox/new/?f=99` on macOS for example, they will still link to `/firefox/new/?scene=2&f=99` when clicking the button. This means they will get the correct corresponding templates still, only their download now won't break.

If we think this is an edge case and makes things more confusing than it's worth, I'd be happy to update and only include the `f` param for Windows / en-US on transitional links too.

pmac: As long as downloads never break I'm happy. Having the "dumber" buttons that just link to scene 2 avoid adding the "f" param to non-win platforms seems more complex. Requesting the funnelcake at scene2 seems fine as long as scene2 is smart enough, and it sounds like it is.

bedrock/firefox/firefox_details.py
```diff
@@ -209,6 +209,9 @@ def get_download_url(self, channel, version, platform, locale,
         _locale = 'ja-JP-mac' if platform == 'osx' and locale == 'ja' else locale
         _platform = 'win' if platform == 'winsha1' else platform
 
+        # Bug 1340087 - Only include funnelcake params for Windows en-US builds.
+        _include_funnelcake_param = funnelcake_id and _platform == 'win' and _locale == 'en-US'
```
pmac: I think for ease of upading this should we be wrong or some funnelcake someday expand compat that we should have settinngs for which platforms and locales are funnelcake enabled. Something like `FUNNELCAKE_PLATFORMS` and `FUNNELCAKE_LOCALES` that are pulled from the environment using the CSV parser and with these defaults. It might be overkill and shouldn't block this, but it would make us not have to push to prod if we're wrong about any of this.

alexgibson: I like this suggestion 👍

alexgibson:
@pmac - I added a commit to use environment variables, seems to work 👍 

Will push this latest commit to the demo server too.

bedrock/firefox/firefox_details.py
```diff
@@ -209,6 +209,10 @@ def get_download_url(self, channel, version, platform, locale,
         _locale = 'ja-JP-mac' if platform == 'osx' and locale == 'ja' else locale
         _platform = 'win' if platform == 'winsha1' else platform
 
+        # Bug 1340087 - Only include funnelcake params for Windows 32bit en-US builds by default.
+        include_funnelcake_param = (_platform in settings.FUNNELCAKE_PLATFORMS and
```
pmac: have `funnelcake_id` be first. This will mean that the settings lists aren't checked until they need to be if a `funnelcake_id` exists.

bedrock/firefox/tests/test_firefox_details.py
```diff
@@ -399,6 +399,25 @@ def test_latest_major_version_no_int(self):
         """latest_major_version should return 0 when no int."""
         eq_(firefox_desktop.latest_major_version('release'), 0)
 
+    def test_funnelcake_direct_links_en_us_win_only(self):
```
pmac: should now add `@override_settings` for the funnelcake settings. Could also set those to something other than the defaults to prove they work.

bedrock/settings/base.py
```diff
@@ -1304,3 +1304,7 @@ def lazy_email_backend():
     CSP_IMG_SRC += (
         'ad.doubleclick.net',
     )
+
+# Bug 1340087 - Funnelcake experiments default to Windows 32bit and en-US builds only.
+FUNNELCAKE_PLATFORMS = config('FUNNELCAKE_PLATFORMS', default='win', cast=Csv())
```
pmac: Looks to me like you do have this right, but you're welcome to add such a test or alter the existing test to use more than one value.

alexgibson:
@pmac - added some review fixes and extra tests, thanks.

bedrock/firefox/tests/test_firefox_details.py
```diff
@@ -399,6 +399,58 @@ def test_latest_major_version_no_int(self):
         """latest_major_version should return 0 when no int."""
         eq_(firefox_desktop.latest_major_version('release'), 0)
 
+    @override_settings(FUNNELCAKE_LOCALES='en-US', FUNNELCAKE_PLATFORMS='win')
```
pmac: These should always be lists, not strings. The `cast=Csv()` ensures that the values of these settings will be lists, even if they're single element. But using `@override_settings` means that these won't pass through the cast function.

bedrock/firefox/tests/test_firefox_details.py
```diff
@@ -399,6 +399,58 @@ def test_latest_major_version_no_int(self):
...
+        ok_('product=firefox-latest-f64' in url)
+
+        url = firefox_desktop.get_download_url('release', '45.0', 'win', 'de', force_direct=True, funnelcake_id='64')
+        ok_('product=firefox-45.0-f64' not in url)
+
+        url = firefox_desktop.get_download_url('release', '45.0', 'osx', 'en-US', force_direct=True, funnelcake_id='64')
+        ok_('product=firefox-45.0-f64' not in url)
+
+    @override_settings(FUNNELCAKE_LOCALES=('en-US', 'de'), FUNNELCAKE_PLATFORMS=('linux', 'osx'))
+    def test_funnelcake_direct_links_locales_linux_osx_(self):
```
pmac: 💯

bedrock/firefox/tests/test_firefox_details.py
alexgibson: Ah good catch, will update

alexgibson:
Updated and squashed commits 👍

bedrock/firefox/tests/test_firefox_details.py
```diff
@@ -399,6 +399,58 @@ def test_latest_major_version_no_int(self):
         """latest_major_version should return 0 when no int."""
         eq_(firefox_desktop.latest_major_version('release'), 0)
 
+    @override_settings(FUNNELCAKE_LOCALES=('en-US'), FUNNELCAKE_PLATFORMS=('win'))
```
pmac: unfortunately, due to how python deals with parentheticals, these are still just strings. In order for them to be tuples it'll have to be `('en-US',)`, or you could make them lists with `['en-US']`. It's confusing, but it's how the syntax works with parentheses. The tests may pass though, since `'win' in 'win'` is `True`, but it's not testing what we want.

alexgibson: Interesting, TIL something

PR #4678, opened 2017-02-23
[fix bug 1341425] Task completion rate survey on mozilla.org

alexgibson:
## Description
- Adds task completion survey link to all mozorg pages (with exceptions for download or in-product pages).
- Popup will only be shown once for visitors (3 day cookie expiry limit).
- If popup is interacted with we won't show it again for much longer (30 day cookie expiry limit).
- Shown for `en` locales only.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1341425

## Testing
- Demo: https://www-demo4.allizom.org
- We're trying to apply a rule of thumb for showing the survey whereby the visitor must have made a conscious choice to visit the website to see it. We should not show the survey to users on in-product triggered pages, like `/firstrun`, `/whatsnew` or `/plugincheck`. We also want to avoid getting in the way of people downloading Firefox. If you can think of any other pages that fall into this category please suggest.

## Checklist
- [ ] Requires l10n changes.


alexgibson:
OK - I've run the full suite of integration tests against this branch and it looks good: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/6/pipeline/

Before this is reviewed/merged, it would be great if we could get #4680 in first. I can then rebase and give the integration tests one last run through.

alexgibson:
OK, I thin this should be good for review, but I'm leaving the DNM label on for now until we know everything is ready as far as GA goes.

alexgibson:
Sorry for the label noise above, please ignore - this is ready for review (I was toying with creating a new label so it's more obvious when something is ready, but decided we already have the `WIP` label so it's probably overkill.

media/js/mozorg/task-completion-rate-survey.js
```diff
@@ -0,0 +1,84 @@
...
+    /**
+     * Task completion rate surevey (bug 1341425).
+     */
+    if ($survey.length) {
+
+        if (typeof Mozilla.Cookies === 'undefined' || !Mozilla.Cookies.enabled()) {
+            return;
+        }
+
+        /* TODO: Re-enable once finished testing on demo */
```
alexgibson: Note to self: I still need to re-enable this before merging.

alexgibson:
Rebased this against latest master and re-ran integration tests, looks good: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/10/pipeline

media/css/pebbles/components/_survey-message.scss
```diff
@@ -0,0 +1,77 @@
...
+#survey-message {
+    @include border-box;
+    background: #fff0b3;
+    bottom: 0;
+    min-height: 30px;
+    overflow: hidden;
+    padding: 10px 20px;
+    position: fixed;
+    text-align: center;
+    transform: translateY(100%);
```
craigcook: Mixin here as well. 

As an aside, at this point all the browsers have supported transform unprefixed since 2014, and transition since 2013. I wonder if it's time to start phasing out some of those prefix mixins. Perhaps add PostCSS autoprefixer that would be easier to maintain in one place. Something to bikeshed in Toronto maybe.

media/css/base/survey-message.less
```diff
@@ -0,0 +1,76 @@
...
+#survey-message {
+    .border-box;
+    background: #fff0b3;
+    bottom: 0;
+    min-height: 30px;
+    overflow: hidden;
+    padding: 10px 20px;
+    position: fixed;
+    text-align: center;
+    transform: translateY(100%);
```
craigcook: Should use the mixin here to be consistent with mixins below.

media/css/pebbles/components/_survey-message.scss
alexgibson: I'm gonna make a call and not add these mixins. The survey still functions without this OK.

craigcook: Yeah, it was more for consistency than compatibility. I'm not especially concerned about catering to 5-year-old webkits. 👍

PR #4679, opened 2017-02-23
[fix bug 1340599] Remove hard-coded mozilla-labs forum links

alexgibson:
## Description
- Removes out of date mozilla-labs forum links to dead projects.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1340599


bedrock/mozorg/templates/mozorg/about/forums/forums.html
```diff
@@ -93,90 +93,6 @@
 
 <!--END_LIST-->
 
-<section id="mozilla-labs" class="forum-group">
```
craigcook: There's a link to this ID in the "Categories" list at the top of the page. Should remove that.

alexgibson: Nice catch 👍

PR #4680, opened 2017-02-23
[bug 1335120] Re-enable home page functional tests for IE

alexgibson:
I meant to re-enable these tests in #4663 but totally forgot.

### DIFF
--- a/tests/functional/test_home.py
+++ b/tests/functional/test_home.py
@@ -7,8 +7,6 @@ import pytest
 from pages.home import HomePage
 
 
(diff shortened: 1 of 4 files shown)


tests/functional/test_home.py
```diff
@@ -7,16 +7,13 @@
 from pages.home import HomePage
 
 
-@pytest.mark.skip_if_internet_explorer(reason='https://ci.us-west.moz.works/job/bedrock_integration_tests_runner/11169/')
-@pytest.mark.skip_if_firefox(reason='Download button is not shown to Firefox browsers.')
```
alexgibson: Also removing this skip, since we no longer hide the home page download button for Firefox users 😸

Issue #4682, opened 2017-02-23
Add more options to jenkins.yml


pmac:
We'd like to provide developers with the ability to change more options on a build per branch via the jenkins.yml file. For example:

* Turn on or off running unittests or smoke tests
* Turn on or off deployment and pushing to the docker hub
* Turn on or off automatic creation of the deis app if it doesn't already exist

pmac:
Also add a section to the jenkins.yml for a pre-defined branch that will just run the unit and smoke tests but not deploy. This will complement the pre-defined branch for deploying and running all of the integration tests.

pmac:
This is done.

PR #4684, opened 2017-02-24
Add more options to jenkins.yml for deployments

pmac:
* Add ability to toggle smoke tests
* Add ability to toggle deployment
* Add ability to toggle integration tests
* Add ability to toggle pushing docker images to public registry
* Add branch name to deploy image
* Add bash utility for setting git environment variables

### DIFF
--- a/docker/bin/push2privateregistries.sh
+++ b/docker/bin/push2privateregistries.sh
@@ -7,12 +7,14 @@
 #
 set -ex
 
+DOCKER_TAG="${BRANCH_NAME}-${GIT_COMMIT}"
+
 # Push to private registry
 for PRIVATE_REGISTRY in ${PRIVATE_REGISTRIES//,/ };
 do
     for DEIS_APP in ${DEIS_APPS//,/ };
     do
-        docker tag $FROM_DOCKER_REPOSITORY:${GIT_COMMIT} $PRIVATE_REGISTRY/$DEIS_APP:${GIT_COMMIT}
+        docker tag $FROM_DOCKER_REPOSITORY:${DOCKER_TAG} $PRIVATE_REGISTRY/$DEIS_APP:${GIT_COMMIT}
         docker push $PRIVATE_REGISTRY/$DEIS_APP:${GIT_COMMIT}
     done
 done
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -53,34 +53,38 @@ if ( config.branches.containsKey(env.BRANCH_NAME) ) {
         }
     }
 
-    milestone()
-    stage ('Test Images') {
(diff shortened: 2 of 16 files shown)


pmac:
This addresses most of #4682.

docker/bin/push2privateregistries.sh
```diff
@@ -7,12 +7,14 @@
...
 
+DOCKER_TAG="${BRANCH_NAME}-${GIT_COMMIT}"
+
 # Push to private registry
 for PRIVATE_REGISTRY in ${PRIVATE_REGISTRIES//,/ };
 do
     for DEIS_APP in ${DEIS_APPS//,/ };
     do
-        docker tag $FROM_DOCKER_REPOSITORY:${GIT_COMMIT} $PRIVATE_REGISTRY/$DEIS_APP:${GIT_COMMIT}
+        docker tag $FROM_DOCKER_REPOSITORY:${DOCKER_TAG} $PRIVATE_REGISTRY/$DEIS_APP:${GIT_COMMIT}
```
pmac: Don't need to use the branch-name tag in the private registry since it's already tagged per deis app name.

pmac:
I'm saving the bit about allowing deployment of demo-like apps via the `branches` config area of `jenkins.yml` for later as it's a much larger change. I'm thinking we'll eventually have a single pipeline flow that will work out whether to do a demo-style or prod-style deployment on the fly either via branch name or via `demo: true` in the YAML. I think that part is for after a docker development environment is done.

pmac:
Oh, and I pushed it to the `run-integration-tests` branch and the [test results look good](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/8/pipeline).

pmac:
@jgmize I'm not actually sure if assigning to you sends you email, so I'm pinging you manually as well for an r?

@glogiotatidis I'd appreciate it if you had a look as well if you have time.

pmac:
@jgmize you said in a meeting that containers are being left running on Jenkins. Turns out I broke cleanup when I changed `local` to `smoke` for the name of the functional tests. The last commit in this PR fixes that. Perhaps we can bump up review and merge of this in the queue?

jgmize:
@pmac and I discussed the edge case I ran into in https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/demo__pr4684/1/pipeline/ and adding an `--rm` to the `docker run` commands should eliminate the name conflict.

pmac:
@jgmize I've added said changes in a new commit. Thanks for the consult!

PR #4687, opened 2017-02-27
[fix bug 1342036] Remove Firefox OS feed links

alexgibson:
One thing I'm unsure on is if I need to do anything with migrations, as this is removing functionality that used the DB? https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/migrations/0001_initial.py#L14


pmac:
Looks good. We do need to add a migration to remove that table however. You should be able to do that simply by running:

```bash
python manage.py schemamigration firefox --auto
```

That should notice that the model is gone and create the appropriate migration file. Update this PR with the results and ping me again.

Thanks for doing this! 🎉

alexgibson:
@pmac updated, thanks 👍

pmac:
@alexgibson I thought of something important. I know this code is not used in theory, but just to be safe we should do this in 2 commits. We should remove the code first, then add the migration in a subsequent commit, and then do the deployment in 2 steps. This will ensure that the database table isn't removed before all of the code is removed that requires it. If you're willing to split your commit please do, if you'd rather I do it I'm very happy to do so as well.

Hopefully this makes sense. If not ping me tomorrow and we can discuss.

alexgibson:
@pmac - thanks, I split this PR up into two separate commits. 

I wasn't quite clear if you meant two separate PRs, one which we'd merge & deploy first, or if this single PR would be sufficient. Either way is fine, just let me know 👍

pmac:
This is perfect. Thanks again Alex!

PR #4689, opened 2017-02-27
Fix Bug 1336524 - The sign up button should be aligned properly with the mail input box - governance page

RCMainak:
## Description
Adjusted the height of the mail input box of the /about/governance/ page to align properly with the sign up button.
## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1336524

### DIFF
--- a/media/css/sandstone/sandstone-resp.less
+++ b/media/css/sandstone/sandstone-resp.less
@@ -901,7 +901,7 @@ nav.menu-bar {
     input[type=email] {
         .border-box();
         width: 100%;
-        padding: 15px 10px;
+        padding: 19px 10px;
     }
 
     .field-privacy {


RCMainak:
@alexgibson not sure what is wrong with the tests. Can you please take a look when you got some time.

alexgibson:
@RCMainak - it looks like the checks passed, but CircleCI failed to ping GitHub for some reason. I've triggered a rebuild either way so hopefully it should work this time.

RCMainak:
CircleCI really doesn't like me :smile_cat:

media/css/sandstone/sandstone-resp.less
```diff
@@ -902,6 +902,7 @@ nav.menu-bar {
         .border-box();
         width: 100%;
         padding: 15px 10px;
+        height: 53px;
```
alexgibson: Rather than introduce a set `height` here, it might be slightly nicer just to change the `padding` directly above to `18px 10px`?

RCMainak:
Of course, padding it more appropriate in this case. I think 19px will be better than 18px. I can change it if you don't like it.

PR #4691, opened 2017-02-28
Fix Bug 1343115 - Update Tokyo office address & map on Mozilla Spaces page

kyoshino:


kyoshino:
Thanks @alexgibson! Updated the pin coordinates as well.

PR #4692, opened 2017-02-28
[fix bug 1343033] Add Yahoo tracking pixels to scene2 of /firefox/new/

alexgibson:
## Description
- Adds 2 new Yahoo tracking pixels to scene 2 of /firefox/new/ (in addition to double click).

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1343033

## Testing
Demo: 
- https://www-demo1.allizom.org/en-US/firefox/new/
- https://www-demo1.allizom.org/en-US/firefox/new/?xv=breakfree
- https://www-demo1.allizom.org/en-US/firefox/new/?xv=wayofthefox
- https://www-demo1.allizom.org/en-US/firefox/new/?f=100

## Checklist
- [ ] Requires l10n changes.


alexgibson:
@jpetto - r on the front end?
@pmac - also interested in thoughts on the approach here for dealing with multiple pixels.

RCMainak:
Wow! The second & third ones are awesome. The other ones are good as well. Really like the second & third one :smile:

pmac:
@alexgibson this looks good to me. the only hiccup could be if one of the pixel URLs gets a comma in it some day. Could use some other character or string as a separator, maybe `::` or `;;`, but that's probably overkill. Nice work 👍

pmac:
One more point: do you think it's possible that we'd want to turn off any of these pixels individually? e.g. perhaps we'd want to leave Yahoo and remove doubleclick? That would require a deployment at present, but could potentially be reworked to be able to be done via env vars. I don't think that's a requirement for this, but it's a thought I had so I figured I'd bring it up for discussion.

media/js/base/mozilla-pixel.js
```diff
@@ -0,0 +1,57 @@
...
+
+        pixels = pixels.split(',');
+
+        for (var i = 0; i < pixels.length; i++) {
+            $pixel = $('<img />', {
+                width: '1',
+                height: '1',
+                src: pixels[i].trim()
+            });
+            $pixel.addClass('moz-px');
```
jpetto: Small nit - could do this in the attribute object above?

bedrock/firefox/templates/firefox/new/break-free/scene2.html
```diff
@@ -9,7 +9,7 @@
 {% block body_class %}scene-2{% endblock %}
 
 {% block string_data %}
-  data-tracking-pixel={{ settings.TRACKING_PIXEL_URL }}
+  data-pixels={% for pixel in settings.TRACKING_PIXELS %}{{ pixel }}{% if not loop.last %},{% endif %}{% endfor %}
```
jpetto: Should this attribute be wrapped in quotes?

media/js/base/mozilla-pixel.js
alexgibson: It turns out older versions of IE freak out because `class` is a reserved word.

bedrock/firefox/templates/firefox/new/break-free/scene2.html
alexgibson: Good call 👍

alexgibson:
@jpetto @pmac added some review fixes, thanks for the suggestions.

media/js/base/mozilla-pixel.js
jpetto: Sigh, of course. 😞

jpetto:
r+ from me 👁

alexgibson:
I'm going to leave the DNM label on this PR until we hear back from Yahoo that the pixels are working as expected.

bedrock/settings/base.py
```diff
@@ -1273,13 +1273,20 @@ def lazy_email_backend():
...
+AVAILABLE_TRACKING_PIXELS = {
+    'doubleclick': ('https://ad.doubleclick.net/ddm/activity/src=6417015;type=deskt0;cat=mozil0;dc_lat=;dc_rdid=;'
+                    'tag_for_child_directed_treatment=;ord=1;num=1?&_dc_ck=try'),
+    'yahoo_purple': 'https://sp.analytics.yahoo.com/spp.pl?a=10000&.yp=10022313',
+    'yahoo_green': 'https://sp.analytics.yahoo.com/spp.pl?a=10000&.yp=10022314',
+}
+ENABLED_PIXELS = config('ENABLED_PIXELS', default='doubleclick,yahoo_purple,yahoo_green', cast=Csv())
+TRACKING_PIXELS = [AVAILABLE_TRACKING_PIXELS[x] for x in ENABLED_PIXELS if x in AVAILABLE_TRACKING_PIXELS]
 
 if config('SWITCH_TRACKING_PIXEL', default=DEV, cast=bool):
```
pmac: consider changing this to something like:

```python
if 'doubleclick' in ENABLED_PIXELS:
    CSP_IMG_SRC += ('ad.doubleclick.net',)

if any([x.startswith('yahoo') for x in ENABLED_PIXELS]):
    CSP_IMG_SRC += ('sp.analytics.yahoo.com',)
```

I'd keep all of that nested in the switch, but this helps us keep the CSP just a tiny bit more lean.

alexgibson:
Thanks @pmac updated

ejregithub:
We are set to go on this to production - Yahoo confirmed their tests are as expected.

alexgibson:
Ok, we have confirmation that the pixels are working as expected, removing DNM label.

PR #4701, opened 2017-03-06
Bug 1335569 - Revert unwanted string change in IH privacy-security

flodolo:
Currently there's an English string exposed in production, since the page was localized.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/internet-health/privacy-security.html
+++ b/bedrock/mozorg/templates/mozorg/internet-health/privacy-security.html
@@ -453,7 +453,7 @@
       <p class="report-cta">
         <span>
         {% trans healthreport='https://internethealthreport.org' %}
-          Read on about Digital Inclusion in our <a href="{{ healthreport }}">Internet Health Report</a>.
+          Keep reading about Privacy and Security in our <a href="{{ healthreport }}">Internet Health Report</a>.
         {% endtrans %}
         </span>
       </p>


bedrock/mozorg/templates/mozorg/internet-health/privacy-security.html
```diff
@@ -452,9 +452,15 @@ <h3 class="topic-action-title">{{ _('What Mozilla is doing about government surv
       <h2>{{ _('Hungry for more?') }}</h2>
       <p class="report-cta">
         <span>
-        {% trans healthreport='https://internethealthreport.org' %}
-          Read on about Digital Inclusion in our <a href="{{ healthreport }}">Internet Health Report</a>.
-        {% endtrans %}
+        {% if l10n_has_tag('internet_health_subpage_rebrand') %}
+          {% trans healthreport='https://internethealthreport.org' %}
+            Read on about Digital Inclusion in our <a href="{{ healthreport }}">Internet Health Report</a>.
```
alexgibson: I'm confused here - This string says Digital Inclusion, yet the file being changed is the Privacy and Security page? Was this string change accidental to begin with perhaps? /cc @craigcook

alexgibson: Looking at the commit history I would hazard a guess that this string should not have been changed (it looks like a `<span>` was added for decoration, but the text might have been a copy/pasta). Will defer to Craig, but i would assume that the fix should be to just revert the string change.

craigcook: Correct, this shouldn't have changed. I was sloppy. We should revert this to the original string.

flodolo:
Updated.

PR #4703, opened 2017-03-06
[fix bug 1344240] Move /all link from page footer to inside modal on /firefox/new/

alexgibson:
## Description
- Moves the link to `/all` from the footer to inside modal on `/firefox/new/` (footer links are due to be removed in [Bug 1342235](https://bugzilla.mozilla.org/show_bug.cgi?id=1342235))
- This doesn't fix the discover-ability problem for other pages once the footer links are removed, but it should fix things for this most critical page. Arguably the links in the footer we're way too hard to find anyway.
- This PR uses existing strings in `main.lang` and `download_button.lang`, so it shouldn't require any l10n conditional tags.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1344240

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/new/onboarding/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/onboarding/scene1.html
@@ -81,7 +81,7 @@
                 <li><a class="more" href="{{ url('firefox.family.index') }}">{{_('Discover all the ways to take Firefox on the go')}}</a></li>
               </ul>
             </div>
-            <button id="other-platforms-modal-link" class="hidden" type="button">{{_('Download Firefox for another platform')}}</button>
+            <button id="other-platforms-modal-link" class="hidden" type="button">{{_('Firefox for Other Platforms & Languages')}}</button>
             <section id="benefits">
               <ul>
                 <li>
--- a/bedrock/firefox/templates/firefox/new/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/scene1.html
@@ -85,7 +85,7 @@
               </ul>
             </div>
             {% if l10n_has_tag('firefox_new_other_platforms') %}
-              <button id="other-platforms-modal-link" class="hidden" type="button">{{_('Download Firefox for another platform')}}</button>
+              <button id="other-platforms-modal-link" class="hidden" type="button">{{_('Firefox for Other Platforms & Languages')}}</button>
             {% endif %}
           </div>{#-- /.main-content --#}
         </div>{#-- /.inner-container --#}
--- a/bedrock/firefox/templates/firefox/includes/download-list.html
+++ b/bedrock/firefox/templates/firefox/includes/download-list.html
@@ -13,3 +13,5 @@
     </li>
   {%- endfor %}
 </ul>
+
+<p><a href="{{ firefox_url('desktop', 'all', 'release') }}">{{_('Available Languages') }}</a></p>
--- a/bedrock/firefox/templates/firefox/new/base.html
+++ b/bedrock/firefox/templates/firefox/new/base.html
@@ -52,9 +52,7 @@
 {% block email_form %}{% endblock %}
 
 {% block site_footer %}
-  {% with fx_footer_links=True %}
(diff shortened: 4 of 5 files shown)


bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -85,7 +85,7 @@
               </ul>
             </div>
             {% if l10n_has_tag('firefox_new_other_platforms') %}
```
craigcook: If we're replacing this string with a preexisting global string, should we remove this tag?

flodolo:
Please tags pull requests with l10n ;-)

bedrock/firefox/templates/firefox/new/scene1.html
flodolo: Since it's in a shared file, we couldn't have a tag there. At this point, remove the tag.

alexgibson: I believe this tag *might* still be needed, as it is also used for the other strings (further down the template) that occur within the modal. Removing it here would expose the modal link, regardless to if the modal is translated.

@flodolo - if enough locales have translated all the modal strings, I agree this can be removed.

flodolo: Yes, it's a very old string

flodolo:
Removing tag at this point, we don't need it.

bedrock/firefox/templates/firefox/new/scene1.html
alexgibson: > Yes, it's a very old string

Sorry what does this mean, that all locales have the modal strings translated? (there's more than just one that craig's pointing out here).

alexgibson: Here: https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/new/scene1.html#L120-L150

bedrock/firefox/templates/firefox/new/onboarding/scene1.html
```diff
@@ -81,7 +81,7 @@
                 <li><a class="more" href="{{ url('firefox.family.index') }}">{{_('Discover all the ways to take Firefox on the go')}}</a></li>
               </ul>
             </div>
-            <button id="other-platforms-modal-link" class="hidden" type="button">{{_('Download Firefox for another platform')}}</button>
+            <button id="other-platforms-modal-link" class="hidden" type="button">{{_('Firefox for Other Platforms & Languages')}}</button>
```
alexgibson: I just realized I still need to include the link to /all for this template, which I totally forgot to do

craigcook: Oh, hah, I totally missed that too. Can I revoke my r+? 🤦‍♂️

bedrock/firefox/templates/firefox/new/scene1.html
craigcook: I think some other strings on the page may need to stay behind the tag, but this specific instance of that tag can be removed.

alexgibson: No, because removing this here would expose the modal link for everyone, but then if the modal itself wasn't also translated, clicking the link would likely break?

flodolo: Sorry, now I get the confusion… 

I'm saying that we can drop the l10n_has_tag only here, since the string will be translated, not in in the whole page.

alexgibson: > I'm saying that we can drop the l10n_has_tag only here, since the string will be translated, not in in the whole page.

And I'm saying we can't drop it here alone, since it would cause the page to break ;)

flodolo: But, given that you're not adding this tag at all here, we shouldn't touch it at all

alexgibson:
@craigcook - thanks, I updated to move the /all link inside the `download-list.html` include, so it's automatically inserted in both templates.

PR #4704, opened 2017-03-07
[fix bug 1342235] Refactor footer links/layout

hoosteeno:
## Description
This PR replaces the content of the footer on every page of the site. After this PR is live, the footer should look _similar_ on every page. There will be minor variation to responsive behavior depending on the CSS that applies, since the CSS grids vary, too.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1342235

## Testing
The following URLs represent 4 different change types to be tested:
* Component CSS: https://bedrock-demo-hoosteeno.us-west.moz.works/en-US/
* Previously "simple footer": https://bedrock-demo-hoosteeno.us-west.moz.works/en-US/firefox/new/
* Fixed-width Sandstone: https://bedrock-demo-hoosteeno.us-west.moz.works/en-US/styleguide/
* Responsive Sandstone: https://bedrock-demo-hoosteeno.us-west.moz.works/en-US/about/

Testing checklist:
- [ ] Tested above URLs in Firefox responsive view on wide/narrow screens, LTR and RTL
- [ ] Tested above URLs in Chrome responsive view on wide/narrow screens, LTR and RTL
- [ ] Tested above URLs in IE11 wide and narrow, LTR and RTL
- [ ] Tested above URLs in iOS Safari, LTR and RTL

## Checklist
- [x] Requires l10n changes.


flodolo:
Just to make expectations clear: this has heavy l10n impact, to the point that we need to block until content is localized for a good part of our users (content for footer is in shared files).

Side question: would it be safe to remove the strings that are removed in this PR?

cc @peiying2

chrismore:
@flodolo can we do an en-US conditional block until it is localized so that en-US is not blocked until it is translated? I know the team has the ability to push out templates before they are localized and other locales get an alternative template.

flodolo:
> can we do an en-US conditional block until it is localized so that en-US is not blocked until it is translated?

Yes, that would be OK. Not sure how feasible it is from a code point of view, given how much this PR removes.

chrismore:
@alexgibson @hoosteeno do you know the question above about not blocking on L10N?

craigcook:
I would suggest making a separate file for the new footer and restoring the old footer(s) as they were. Then the conditional determines which file to include:

```
{% include 'includes/new-footer.html' %}
{% else %}
{% include 'includes/old-footer.html' %}
{% endif %}
```
Once locales are up to date we can remove the old footers.

Is this an experiment that is subject to change or are these strings locked and final? We also will probably be making further footer updates very soon in conjunction with global navigation (not undoing these changes but possibly adding/removing other bits). It might make sense to delay l10n a few weeks until the whole footer is more concrete.

bedrock/base/templates/includes/site-footer.html
```diff
@@ -1,60 +1,67 @@
...
                   <li><a href="https://www.instagram.com/mozillagram/" data-link-type="footer" data-link-name="Instagram (@mozillagram)">{{ _('Instagram') }}<span> (@mozillagram)</span></a></li>
                 </ul>
               </li>
+            </ul>
+          </div>
+          <div class="col col-3">
+            <strong>{{ _('Firefox') }}</strong>
+            <ul class="firefox-links">
+              <li><a href="{{ url('firefox.new') }}" data-link-type="footer" data-link-name="Download Firefox now">{{ _('Download Firefox now') }}</a></li>
+              <li><a href="{{ url('firefox') }}" data-link-type="footer" data-link-name="Firefox Web Browser">{{ _('Firefox Web Browser') }}</li></a>
```
alexgibson: This link is a duplicate of the link above, changing only keywords. I'm prepared to accept that these strings are geared towards robots, but creating duplicate links doesn't really do anything for our users. Can I suggest either dropping one of these, or rewording to cover more keywords better in a single instance?

alexgibson:
I second @craigcook's suggestion here - unless we're prepared to wait for translators and confident that these strings won't change again, doing this PR for `en` locales is the only real way forward.

If we don't want these strings translated until we're confident they do what we need (and make sense for the upcoming redesigned footer), we could also go as far as hard-coding them (i.e. not wrapping the strings for l10n)

bedrock/base/templates/includes/site-footer.html
hoosteeno: The firefox landing page (e.g. "firefox") currently redirects to the firefox download page ("new") -- yes. But that is a problem we aim to solve. We should have a page at /firefox that talks about firefox and doesn't just redirect to a green button.

That said, I'd like to send the "download" link to scene2, so the download begins, which the link promises. I can make that fix -- sound good?

hoosteeno:
@craigcook, I can do what you suggest. But the CSS changes are surgical and wide-ranging. If I drop all untranslated strings from non-en-US locales, the grid will probably break. Should I also include CSS conditionally per locale? 

@all, what do you think about making the "in" locales include our tier-1 markets (US, Germany, China)? That'll net us more value from this SEO exercise, and hopefully translating the small number of words into those languages is possible in a shorter time?

craigcook:
Hmm... The new footer could have some other style hook that all the new CSS could hang from while all the old CSS still works for the old footer. That may be the simplest way to keep the variations separate.

flodolo:
> @all, what do you think about making the "in" locales include our tier-1 markets (US, Germany, China)? That'll net us more value from this SEO exercise, and hopefully translating the small number of words into those languages is possible in a shorter time?

I would object that you say locale but talk about countries. Having said that, it's not possible, because these are shared strings: we would need to expose them, and wait for the requested languages to be ready. At that point, what would you do? Exclude other complete languages, create a huge whitelist?

If the content is not stable, and might change, I would strongly suggest to keep it en-US only until it's more likely to be stable.

bedrock/base/templates/includes/site-footer.html
alexgibson: The Firefox content hub is still a far off thing, and may not end up living at /firefox. Until that time, I don't think we should add duplicate links just for SEO. We could always add it later once the page exists and we know the URL.

Please keep the link to /new aimed at scene1. Pointing a regular link at scene2 will not work in older browsers, which is something our dedicated download buttons mitigate for us. I don't think we should add a normal link to scene2 like this from every page on the site.

alexgibson:
We're going to try and land this at the same time we launch the new global navigation for mozorg, as the two have interrelated changes (both will be for `en` locales only until we settle on something we're confident with). I'm going to close this and pick up the changes. Thanks for your efforts thus far @hoosteeno 👍

PR #4709, opened 2017-03-08
Bug 1344220: Enable stub installer for win64

pmac:
All locales for beta and release, and en-US only for aurora and nightly.

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -219,12 +219,14 @@ class FirefoxDesktop(_ProductDetails):
             fc_locales = config('FUNNELCAKE_%s_LOCALES' % funnelcake_id, default='', cast=Csv())
             include_funnelcake_param = _platform in fc_platforms and _locale in fc_locales
 
+        stub_langs = settings.STUB_INSTALLER_LOCALES.get(channel, {}).get(_platform, [])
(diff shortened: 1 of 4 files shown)


pmac:
Reworks configuration of stub installer platforms and locales as well.

@jgmize r?

bedrock/firefox/firefox_details.py
```diff
@@ -223,8 +223,10 @@ def get_download_url(self, channel, version, platform, locale,
         # see bug 1324001
         if channel in ['alpha', 'nightly']:
             prod_name = 'firefox-nightly' if channel == 'nightly' else 'firefox-aurora'
-            # Use the stub installer for 32-bit Windows
-            if _platform == 'win' and locale == 'en-US' and not force_full_installer:
+            # Use the stub installer for approved platforms
+            stub_langs = settings.STUB_INSTALLER_LOCALES[channel].get(_platform, [])
```
jgmize: Do we know for a fact that 'alpha' and 'nightly' will always be in `settings.STUB_INSTALLER_LOCALES`? If not we should change the dict access to a `.get(channel, {})` to avoid an error in that case.

pmac: Good point. Can't hurt. I thought tests should cover that, but I guess that setting is usually mocked out.

PR #4711, opened 2017-03-09
Bug 1345467: Allow per experiment configuration of funnelcake platforms and locales

pmac:
My proposal for how to most easily control funnelcake configuration with environment variables.

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -4,8 +4,11 @@ from operator import itemgetter
 from urllib import urlencode
 
 from django.conf import settings
-from bedrock.base.waffle import switch
+
+from decouple import Csv, config
 from product_details import ProductDetails
+
+from bedrock.base.waffle import switch
 from lib.l10n_utils.dotlang import _lazy as _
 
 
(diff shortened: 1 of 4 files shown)


pmac:
@jgmize and @alexgibson let's discuss. This needs tests and things, but I wanted your opinion on the direction before committing more time.

alexgibson:
I think this looks good in terms of mapping funnelcake IDs to locales/platforms. The only thing I'm unsure on is with regard to the default fallback of en-US/win being on all the time, if no specific ID's are specified in the env variable. I imagine this en-US/win default will be permanently on 95% of the time, which *could* still lead to potential errors. On the other side, it probably leaves Chris the most freedom and makes doing experiments easier from his point of view. So I guess we have to weigh up how strict we want to be.

pmac:
> I think this looks good in terms of mapping funnelcake IDs to locales/platforms. The only thing I'm unsure on is with regard to the default fallback of en-US/win being on all the time, if no specific ID's are specified in the env variable. I imagine this en-US/win default will be permanently on 95% of the time, which could still lead to potential errors. On the other side, it probably leaves Chris the most freedom and makes doing experiments easier from his point of view. So I guess we have to weigh up how strict we want to be.

This was exactly my thinking. I figured it might make it easier to swallow for the common case to remain active. I'd LOVE to make it a requirement that we specifically setup every FC ID though. If we can make that work I'm very happy to remove the defaults.

jgmize:
@cmore how are funnelcake ids currently tracked and mapped to platforms and languages? Is there a spreadsheet or wiki page, perhaps?

pmac:
@jgmize there is indeed. I've talked to cmore and there is both [a spreadsheet](https://docs.google.com/spreadsheets/d/1gZ-GFyubAAlIxh-u-5SR52rqKKfBRLRuwrjSq97QkWA/edit#gid=0) and a bug is filed for each new funnelcake. cmore agreed to CC Eric Renaud for each of these and we'll therefore know to add the locales and platforms to the app environments before their start dates.

Currently active experiments are funnelcakes 101 - 104, which I'll configure when this PR is merged to avoid any experiment interruption.

pmac:
Thanks Josh. I'm pre-configuring the bedrock instances for FCs 101 - 104 now. Ran:

```bash
$ deis-both config:set -a bedrock-dev \
FUNNELCAKE_101_PLATFORMS=win \
FUNNELCAKE_102_PLATFORMS=win \
FUNNELCAKE_103_PLATFORMS=win \
FUNNELCAKE_104_PLATFORMS=win \
FUNNELCAKE_101_LOCALES=en-US \
FUNNELCAKE_102_LOCALES=en-US \
FUNNELCAKE_103_LOCALES=de \
FUNNELCAKE_104_LOCALES=de
```

PR #4718, opened 2017-03-13
Add one retry to integration test runs

pmac:
We're seeing random failures, and this would help prevent having to run the full deployment again.


pmac:
Ran this change in the [run-integration-tests](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/11/pipeline) branch.

PR #4724, opened 2017-03-15
Fix bug 1347663: Add etag headers to responses

pmac:
Not sure why we weren't doing this before. Bedrock lends itself perfectly to etags and our CDN provider is ready for their use:

https://support.cloudflare.com/hc/en-us/articles/218505467-Does-CloudFlare-support-ETag-headers-

Django docs for the feature enabled:

https://docs.djangoproject.com/en/1.8/ref/middleware/#module-django.middleware.common

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -96,6 +96,8 @@ USE_L10N = True
 
 USE_TZ = True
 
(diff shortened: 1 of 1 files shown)


jgmize:
Useful docs I found while researching this PR: https://www.mnot.net/cache_docs/

PR #4660, opened 2017-02-12
fix Bug 1329931 : Decommission and redirect Firefox OS pages.

RCMainak:
## Description
Decommission and redirect Firefox OS pages.
## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1329931
## Checklist
- [ ] Requires l10n changes.
- [ ] Related functional & integration tests passing.

PR #4725, opened 2017-03-15
Fix Bug 1347697 - /about/history/details/ throwing 500s

kyoshino:
A regression from #4660.


pmac:
Just bummed it took me this long to notice! Thanks for the quick fix and merge Kohei and Alex!

PR #4732, opened 2017-03-23
[fix bug 1349760] Add link to /firefox/android/all to homepage

alexgibson:
## Description
- Adds link to /firefox/android/all/ to homepage Firefox CTA

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1349760

### DIFF
--- a/bedrock/mozorg/templates/mozorg/home/home.html
+++ b/bedrock/mozorg/templates/mozorg/home/home.html
@@ -176,6 +176,7 @@
                 </a>
               </li>
             </ul>
+            <p><a class="android-systems-link hidden" href="{{ firefox_url('android', 'all', 'release') }}">{{_('Systems &amp; Languages')}}</a></p>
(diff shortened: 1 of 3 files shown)


bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -176,6 +176,7 @@ <h2 class="section-headline">
                 </a>
               </li>
             </ul>
+            <p><a class="android-systems-link hidden" href="{{ firefox_url('android', 'all', 'release') }}">{{_('Systems &amp; Languages')}}</a></p>
```
alexgibson: This re-uses a string from `download_button.lang` (cc @peiying2 @flodolo)

flodolo: This should be OK, context doesn't change and it's a shared file

PR #4738, opened 2017-03-29
[fix bug 1346724] Implement global navigation prototype

alexgibson:
## Description
- Adds global navigation to nearly all mozorg pages for `en` locales only.
- Does not yet add nav to the majority of the Firefox app, since that requires more thought and UX attention. There's a [list of pages here](https://bugzilla.mozilla.org/show_bug.cgi?id=1346724#c18) that don't get the new nav just yet.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1346724

## Testing
- I've tested this in all modern browsers and a good range of mobile browsers.
- This touches many mozorg pages, and some may look a bit odd - but the key thing is we don't break anything.
- Non `en` locales should look as normal with the old navigation.
- ~I'm still working on running this branch against the full suite of integration tests, but it would be good to start review~ Branch is passing: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/18/pipeline/.

## Checklist
- [ ] Requires l10n changes.


bedrock/settings/static_media.py
```diff
@@ -1017,6 +1019,8 @@
             'js/base/mozilla-client.js',
             'js/base/mozilla-image-helper.js',
             'js/base/nav-main-resp.js',
+            'js/base/class-list-polyfill.js',
```
alexgibson: I'm adding this to support IE 9 & 10, although I could be tempted to drop the polyfill if we don't think it's worth it.

tests/functional/test_navigation.py
```diff
@@ -9,24 +9,25 @@
 
 @pytest.mark.nondestructive
 def test_navigation(base_url, selenium):
-    page = HomePage(selenium, base_url).open()
-    internet_health_page = page.navigation.open_internet_health()
+    locale = 'de'
```
alexgibson: I'm switching these tests to run against `de`, to make sure we don't accidentally break the old navigation whilst it's still around.

media/js/base/mozilla-global-nav.js
```diff
@@ -0,0 +1,273 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    var _closeButton = document.getElementById('nav-drawer-close-button');
```
alexgibson: Note: the reason I'm using plain JS here is with the assumption that this will eventually be used on more than just mozorg, so I'm trying to avoid too many dependencies.

bedrock/base/templates/includes/global-nav.html
```diff
@@ -0,0 +1,279 @@
+<nav role="navigation" class="moz-global-nav" id="moz-global-nav">
+    <div class="nav-horizontal-menu">
+
+      <button type="button" class="nav-button-menu nav-hidden" id="nav-button-menu" aria-controls="moz-global-nav-drawer">
+        <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 12 25" width="12" height="25">
+          <title>{{_('Menu')}}</title>
+          <g>
+            <path class="p1" fill="#231f20" d="M3 0h4.62v5.55H3zM3"/>
```
alexgibson: In Chrome I see these SVG paths are throwing an error `Error: <path> attribute d: Unexpected end of attribute. Expected number, "…0h4.62v5.55H3zM3".` I had to hack this SVG by hand, so I probably screwed up somewhere. I may need to see if I can get a version of the image as three separate paths.

alexgibson: Ok this is now fixed

bedrock/base/templates/includes/global-nav.html
```diff
@@ -0,0 +1,277 @@
...
+            <path class="p" d="M213.27 78.73c-11.19 0-18.14 8.3-18.14 22.72 0 13.22 6.1 23.39 18 23.39 11.36 0 18.82-9.15 18.82-23.73-.03-15.43-8.33-22.38-18.68-22.38z"/>
+          </svg>
+        </a>
+      </h2>
+
+      <ul class="nav-primary-links">
+        <li class="item-orange"><a href="{{ url('firefox.new') }}" data-id="firefox">{{ _('Firefox') }}</a></li>
```
craigcook: I'm not sure how I feel about these presentational classes. I know that at least for now they really are presentational and are only being used as hooks for colors, but maybe it would be better to classify these items based on the section/category of their contents, then define a theme color for each one in the CSS.

media/css/pebbles/components/_global-nav.scss
```diff
@@ -0,0 +1,594 @@
...
+                transform: translateX(-3px);
+            }
+
+            .p3 {
+                transform: translateX(4px);
+            }
+        }
+
+        &:focus {
+            outline: none;
```
alexgibson: Yeah, I think I need to use a different styling for focus here. I'll change the `fill` color to `#646464` to match the wordmark.

media/css/pebbles/components/_global-nav.scss
```diff
@@ -0,0 +1,594 @@
...
+        transition: fill 0.1s ease-in-out;
+
+        a {
+            display: block;
+            height: 25px;
+        }
+
+        &:hover .p,
+        &:focus .p,
+        &:active .p {
```
alexgibson: nice catch

media/css/pebbles/components/_global-nav.scss
```diff
@@ -0,0 +1,594 @@
...
+        visibility: hidden;
+    }
+}
+
+/* -------------------------------------------------------------------------- */
+// Side drawer promo thumbnails.
+
+.moz-global-nav-drawer {
+
+    .nav-promo {
```
alexgibson: I was in such a rush to get these in I completely forgot to do this, thanks for the reminder 👍

media/css/pebbles/components/_global-nav.scss
```diff
@@ -0,0 +1,594 @@
...
+    a:hover,
+    a:focus,
+    a:active {
+        color: #000;
+        text-decoration: underline;
+        transition: color 0.1s ease-in-out;
+    }
+
+    // Three-dotted nenu button
+    .nav-button-menu {
```
alexgibson: So I'm generally siding these days that buttons [shouldn't get a hand cursor](https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b) in general. Here I agree it's not immediately obvious that this is a button however, so I think an extra visual cue could be worthwhile.

bedrock/base/templates/includes/global-nav.html
alexgibson: I was torn on this as well, as doing it like this does make things easier for reuse for other sites, which may not have the same navigation items as mozorg. Perhaps I'm thinking too far ahead though, we can always change this further down the line to something less specific but not overly presentational.

alexgibson:
@craigcook thanks for the review thus far, updated.

media/css/pebbles/components/_global-nav.scss
craigcook: I would probably take that side of the argument for things that very clearly look like interactive buttons (a form submit, or a modal close button), but yeah, this element is something unfamiliar so the extra hint may be helpful.

media/css/base/global-nav.less
```diff
@@ -0,0 +1,602 @@
...
+    opacity: 0;
+    position: absolute;
+    right: auto;
+    top: 0;
+    transition: opacity 0.2s ease-in-out;
+    z-index: 10000;
+}
+
+html.moz-nav-open {
+    .moz-global-nav-page-mask {
```
jpetto: Nit - don't really need to nest here.

media/js/base/mozilla-global-nav.js
jpetto: Plain JS 4evar. 🙌

media/css/pebbles/components/_global-nav.scss
```diff
@@ -0,0 +1,600 @@
...
+        }
+    }
+
+    .download-link:link,
+    .download-link:visited {
+        @include font-size(14px);
+        background: inherit;
+        border-radius: 0;
+        border: 1px solid $color-firefox-light-orange;
+        color: $color-firefox-light-orange;
```
craigcook: Let's try darkening the text a tad for better contrast. Maybe `color: darken($color-firefox-light-orange, 5%);`

bedrock/settings/static_media.py
jpetto: The polyfill is pretty small. I'm good with keeping it around.

alexgibson:
> Could definitely be my VM, but IE 11 crashes when using the nav. Seems to happen after opening the nav a second time.

Hmm, I can replicate this in any of the versions of IE11 I have running on Browser Stack. Can you give some more detail / debug any more?

alexgibson:
> Could definitely be my VM, but IE 11 crashes when using the nav. Seems to happen after opening the nav a second time.

I've tested this on every combination of IE11 on both Browser Stack and Sauce Labs, and also on my real Windows laptop on both IE11 and Edge, and cannot reproduce anything. After discussing on IRC, I think we may have to chalk this up to VM issues.

alexgibson:
Also tried this on my better-half's Windows 8.1 laptop and IE11 was fine 👍

bedrock/base/templates/includes/global-nav.html
```diff
@@ -0,0 +1,277 @@
...
+
+      <ul class="nav-primary-links">
+        <li class="item-firefox"><a href="{{ url('firefox.new') }}" data-id="firefox">{{ _('Firefox') }}</a></li>
+        <li class="item-internet-health"><a href="{{ url('mozorg.internet-health') }}" data-id="internet-health">{{ _('Internet Health') }}</a></li>
+        <li class="item-technology"><a href="{{ url('mozorg.technology') }}" data-id="technology">{{ _('Technology') }}</a></li>
+        <li class="item-about-us"><a href="{{ url('mozorg.about') }}" data-id="about-us">{{ _('About Us') }}</a></li>
+        <li class="item-get-involved"><a href="{{ url('mozorg.contribute.index') }}" data-id="get-involved">{{ _('Get Involved') }}</a></li>
+      </ul>
+
+      {{ download_firefox(alt_copy=_('Download Firefox'), dom_id='global-nav-download-firefox') }}
```
alexgibson: I need to try and limit the length of text used here, so if I omit this `alt_copy` then the button could say "Firefox for Android" for example, depending on the platform.

alexgibson:
Thanks @jpetto and @craigcook - I think all review comments thus far should now be addressed. 

I also switched to using `CSS.Supports` for `transition` feature detection in the JS, so this now matches how we do feature detection in the CSS. Previously, using `'transition' in document.documentElement.style` was causing a `transitionend` events to leak in IE, since the feature query in the CSS prevented IE from getting transitions to begin with, which resulted in the event never firing/unbinding.

alexgibson:
@jpetto @craigcook care to take a second look here? Aim is to have this out today

craigcook:
Looking good to me. I did notice while testing IE10 that (in my VM at least) the download button wraps under *just* before the tablet breakpoint. It doesn't happen in my Mac browsers so I suspect it's down to minor differences in front rendering making things just a tad wider on Windows. But given changing fonts plus l10n I think our tolerances may be too close and we should tighten the spacing at an earlier breakpoint.

navbar-breakpoint

alexgibson:
Nice catch @craigcook - updated.

alexgibson:
Thanks @jpetto @craigcook squashed 👍

PR #4741, opened 2017-03-30
[fix bug 1336573] Jul-Dec 2016 transparency report

craigcook:
## Description
Adds a transparency report for July-December 2016 and moves the previous one to "past reports." No l10n required.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1336573

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/en-US/about/policy/transparency/jul-dec-2016/

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
+++ b/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
@@ -7,6 +7,7 @@
 <nav class="content nav-previous">
   <h3>{{ _('Previous Reports') }}</h3>
   <ul>
+    <li><a href="{{ url('mozorg.about.policy.transparency.jan-jun-2016') }}">January-June 2016</a></li>
     <li><a href="{{ url('mozorg.about.policy.transparency.jan-dec-2015') }}">January-December 2015</a></li>
   </ul>
 </nav>
--- a/bedrock/mozorg/templates/mozorg/about/policy/transparency/jul-dec-2016.html
+++ b/bedrock/mozorg/templates/mozorg/about/policy/transparency/jul-dec-2016.html
@@ -4,10 +4,10 @@
 
 {% extends "mozorg/about/policy/transparency/report-base.html" %}
 
-{% block page_title %}{{ _('Transparency Report - January–June, 2016') }}{% endblock %}
+{% block page_title %}{{ _('Transparency Report - July–December, 2016') }}{% endblock %}
 
 {% block report_title %}
-  <h1>{{ _('Transparency Report') }} <span>{{ _('Reporting Period: January 1, 2016 to June 30, 2016') }}</span></h1>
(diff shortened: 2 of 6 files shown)


craigcook:
Marked DNM for the moment awaiting final signoff in the bug.

bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
```diff
@@ -8,5 +8,6 @@
   <h3>{{ _('Previous Reports') }}</h3>
   <ul>
     <li><a href="{{ url('mozorg.about.policy.transparency.jan-dec-2015') }}">January-December 2015</a></li>
+    <li><a href="{{ url('mozorg.about.policy.transparency.jan-jun-2016') }}">January-June 2016</a></li>
```
alexgibson: Nit: should past reports be listed by most-recent first?

bedrock/mozorg/templates/mozorg/about/policy/transparency/jul-dec-2016.html
```diff
@@ -0,0 +1,256 @@
...
+          shared_responsibility='https://blog.mozilla.org/blog/2016/09/13/cybersecurity-is-a-shared-responsibility/',
+          security_vuln='https://blog.mozilla.org/netpolicy/2016/09/19/improving-government-disclosure-of-security-vulnerabilities/',
+          speaker_gov='http://cyberlaw.stanford.edu/events/government-hacking-vulnerabilities-equities-process',
+          speaker_hack='http://cyberlaw.stanford.edu/events/government-hacking-rule-41',
+          gag_orders='https://blog.mozilla.org/mozilla-brief-of-joint-amicus-in-msft-v-doj/',
+          eu_privacy='https://blog.mozilla.org/netpolicy/files/2017/01/Mozilla-Filing_E-Privacy-Survey.pdf',
+          india_cloud='https://blog.mozilla.org/netpolicy/files/2016/07/Mozilla-Submissions-TRAI-Consultation-on-Cloud-Computing.pdf'
+      %}
+        In the Reporting Period, we highlighted the need to see
+        <a href="{{ shared_responsibility }}" rel="external">cybersecurity as a shared responsibility</a>,
```
alexgibson: This is a tiny nit and I doubt these particular links will likely ever change, but I've recently started to question if we should be including `rel="external"` inside trans blocks in general. The URLs can change independently of translations, however this bakes in the assumption that the URL will only ever be external and provides little real user benefit. I think here it's likely fine, but just flagging in general.

craigcook: Fair point. Honestly I'm not even sure there's much benefit to `rel="external"` at all, I've just made it a habit. In this case I don't think this page will ever be translated and the links will probably never change but I'll remove them anyway.

bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
craigcook: Good idea. I shall make it so.

PR #4745, opened 2017-04-03
Fix external /firefox/channel redirect tests

alexgibson:
@pmac - I'm guessing whomever owns these domains updated the external redirects to point to the new channel pages?

Failing in master: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/master/190/pipeline/


pmac:
Huh. I didn't realize (or remember) that we tested these. Doesn't seem like something we should be testing since we have no control over them and they're not part of bedrock. Perhaps we should just remove them?

alexgibson:
> Huh. I didn't realize (or remember) that we tested these. Doesn't seem like something we should be testing since we have no control over them and they're not part of bedrock. Perhaps we should just remove them?

Fine by me - assume these are legacy things from the mcom-test repo?

alexgibson:
@pmac updated

alexgibson:
@pmac fixed the error caught by circle 👍

PR #4747, opened 2017-04-03
/firefox/products/ variation & experiment

jpetto:
## Description

Adds a variation for `/firefox/products/`. This page is sort of a version 0 for the landing page of the Firefox Content Hub.

We are creating this as a variation to test conversions against a similar page prior to moving to the `/firefox/` URL.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1348146

## Testing

50% of `en-US` traffic should go to `?v=a` (control) and 50% to `?v=b` (variation). It's up on a demo:

https://bedrock-demo-jpetto.us-west.moz.works/en-US/firefox/products/?v=b

## Checklist


bedrock/firefox/templates/firefox/family/index-b.html
```diff
@@ -0,0 +1,173 @@
...
+      <h3>{{ _('Browse faster with Firefox') }}</h3>
+
+      <p>
+      {% trans %}
+        Firefox is faster than ever before, thanks to a new multi-process
+        architecture that improves browser speed and page-load time.
+      {% endtrans %}
+
+      {% trans url1=url('mozorg.home'), url2='https://blog.mozilla.org/firefox/tag/quantum' %}
+        <a href="{{ url1 }}">Mozilla</a>, the nonprofit that builds Firefox,
```
flodolo: This is a peculiar way to spell *nonprofit*: there are only a couple of cases in mozilla.org for pages exposed to localization (Foundation landing page, DNT), all the rest are *non-profit*

bedrock/firefox/templates/firefox/family/index-b.html
```diff
@@ -0,0 +1,173 @@
...
+               url4='https://addons.mozilla.org/category/search',
+               url5='https://addons.mozilla.org/category/vpn' %}
+        What does your browser history say about you? Who has access to your browser
+        history? Plenty of browsers offer an incognito mode, but only Firefox offers
+        advanced tracking protection: Just
+        <a href="{{ url1 }}">open a private browsing window</a> and enjoy safe
+        browsing again. Enhance your private browsing experience with dozens of
+        <a href="{{ url2 }}">privacy add-ons</a> — for
+        <a href="{{ url3 }}">ad blocking</a>,
+        <a href="{{ url4 }}">private search</a>, <a href="{{ url5 }}">vpn</a> and
```
flodolo: [Not sure](https://en.wikipedia.org/wiki/Acronym#Case) if this should be vpn or VPN.

bedrock/firefox/templates/firefox/family/index-b.html
```diff
@@ -0,0 +1,173 @@
...
+      <h3>{{ _('Use Firefox Everywhere') }}</h3>
+
+      <p>
+      {% trans url1=url('firefox.ios'), url2=url('firefox.desktop.index'),
+               url3=url('firefox.android.index'), url4=url('firefox.all'),
+               url5=url('firefox.sync') %}
+        Looking for the best browser for desktop and mobile? Firefox is available on
+        <a href="{{ url1 }}">iOS</a>, <a href="{{ url2 }}">Windows</a>,
+        <a href="{{ url3 }}">Android</a>, <a href="{{ url4 }}">linux</a>, and
+        <a href="{{ url2 }}">MacOS</a>. Need a better browser in your language?
```
flodolo: macOS

bedrock/firefox/templates/firefox/family/index-b.html
```diff
@@ -0,0 +1,173 @@
...
+<section class="content-section" id="better">
+  <div class="content">
+    <div class="copy">
+      <h3>{{ _('Choose Firefox, A Better Browser') }}</h3>
+
+      <p>
+      {% trans url1=url('mozorg.home'), url2=url('firefox.new') %}
+        Most browsers are designed to limit your choices and protect a corporate
+        agenda, but not Firefox. Firefox is the only independent web browser,
+        proudly <a href="{{ url1 }}">built by a nonprofit</a> on a mission to make
```
flodolo: Same here (nonprofit)

bedrock/firefox/templates/firefox/family/index-b.html
```diff
@@ -0,0 +1,173 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import fxfamilynav with context %}
+
+{% extends "base-resp.html" %}
```
alexgibson: Should this extend `firefox/base-resp.html` instead? Currently the page is getting the new global nav, which seems out of place with the family navigation.

bedrock/firefox/templates/firefox/family/index-b.html
```diff
@@ -0,0 +1,173 @@
...
+        architecture that improves browser speed and page-load time.
+      {% endtrans %}
+
+      {% trans url1=url('mozorg.home'), url2='https://blog.mozilla.org/firefox/tag/quantum' %}
+        <a href="{{ url1 }}">Mozilla</a>, the nonprofit that builds Firefox,
+        invented a new programming language to
+        <a rel="external" href="{{ url2 }}">make Firefox faster</a>.
+      {% endtrans %}
+
+      {% trans %}
```
alexgibson: Does each sentence need to be a separate trans block? Seems like the paragraph could just be one?

media/css/firefox/family/index-b.less
```diff
@@ -0,0 +1,305 @@
...
+    }
+}
+
+#intro {
+    &.content-section {
+        padding-top: 60px;
+    }
+
+    .content {
+        .at2x('/media/img/firefox/family/index-b/intro.png', 632px, 550px);
```
alexgibson: I know this is something we may not have, but it would be really great to have OS specific images here and not just macOS?

alexgibson:
We shouldn't forget to [skip the product page functional tests](https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/test_products.py) when the experiment is running.

bedrock/firefox/templates/firefox/family/index-b.html
jpetto: Ah, didn't see this until after a rebase. Yep, will switch the inheritance.

bedrock/firefox/templates/firefox/family/index-b.html
jpetto: I'm 99% sure this page will not be localized in its current state. It's just an `en-US` test to see how downloads stack up. I probably shouldn't have even wrapped the strings...

media/css/firefox/family/index-b.less
jpetto: Agreed, but not worth it for a short lived page/experiment.

bedrock/firefox/templates/firefox/family/index-b.html
jpetto: Good catch - will update to "non-profit".

bedrock/firefox/templates/firefox/family/index-b.html
jpetto: Hm, will change to "VPN".

media/css/firefox/family/index-b.less
alexgibson: Well, I dunno - could a large visual like this negatively impact conversions if people think it's for macOS only?

jpetto: I mean, to you and I, it's clearly an un-badged iMac, but the intent is "beautiful desktop computer as backdrop to our browser". I can check with Justin though - see what he thinks.

alexgibson: Yeah, but to my dad for example it could just be "nothing like his computer" 😉

jpetto: Are you daring me to put a [Dimension](http://www.desktop-driver.com/wp-content/uploads/2013/02/Dell-Dimension-4100-450x336.jpg) in there instead? Because I will.

jpetto: I asked for Justin & Rachel's thoughts in [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1348146#c24).

alexgibson: > Are you daring me to put a Dimension in there instead? Because I will.

LOL, maybe add a Windows XP logo too?

jpetto:
> I'll be honest, I'm a little lost on what the experiment is trying to find out. On the new experiment page there is a single CTA (download button), as opposed to individual CTA's off to the individual product pages on the regular /products page. Are we just concerned with how these perform?

It's not obvious, and was just changed. Initially, we were going to run this page at `/firefox/` and test against `/new/`. But, because the intent of this page is so different from `/new/`, we decided that wasn't a good test.

If you search for "firefox" (in Google, in the US), the top result is `/firefox/products/` (we don't know why), which does not do a great job informing the visitor about the product.

So, this is a two-step approach. 1) Make sure this page doesn't do significantly worse than the current `/firefox/products/` page, and then, 2) move this page [1] (assuming it wins) over to `/firefox/`. (Guessing we'd also add a redirect from `/firefox/products/` to `/firefox/`.) We are almost certain reclaiming `/firefox/` will cause us to take a hit on downloads (instead of redirecting to `/new/`), but apparently the SEO gains are worth it.

> Would it make sense to run this against /new instead? (which is essentially what eliminating the /firefox redirect would be doing).

The general assumption is that any page with a different intent than `/new/` will perform worse in terms of downloads (at least in the short-ish term). The hope is that providing information/education about the product, as well as having better SEO, will benefit us in the long run.


[1] To further complicate matters, this page is version 0 of the Firefox hub, and will be changed as soon as the revision is ready. We should still have time to test the page in this PR though.

jpetto:
@flodolo - Just to be clear, we should not begin localization of this page at this time. This is an en-US only experiment that already has a version 2 in the works.

alexgibson:
@jpetto thanks for the detailed explaination 👍

flodolo:
> @flodolo - Just to be clear, we should not begin localization of this page at this time. This is an en-US only experiment that already has a version 2 in the works.

I know, but changes are usually incremental, so I prefer to point weird things early 🦄

jpetto:
> I know, but changes are usually incremental, so I prefer to point weird things early

It is most definitely appreciated. 😁

media/img/firefox/family/index-b/download-arrow.svg
```diff
@@ -0,0 +1 @@
+<svg width="22" height="21" viewBox="0 0 22 21" xmlns="http://www.w3.org/2000/svg"><title>Download Icon</title><g stroke="#FFF" stroke-width="2" fill="none" fill-rule="evenodd" stroke-linecap="round" stroke-linejoin="round"><path d="M1 20h20M6.5 10.5l4.528 4.528M15.5 10.5l-4.528 4.528M11 1v13.533"/></g></svg>
```
alexgibson: Nit: remove `<title>`?

bedrock/firefox/templates/firefox/family/index-b.html
```diff
@@ -0,0 +1,163 @@
...
+      <h3>{{ _('Use Firefox Everywhere') }}</h3>
+
+      <p>
+      {% trans url1=url('firefox.ios'), url2=url('firefox.desktop.index'),
+               url3=url('firefox.android.index'), url4=url('firefox.all'),
+               url5=url('firefox.sync') %}
+        Looking for the best browser for desktop and mobile? Firefox is
+        available on <a href="{{ url1 }}">iOS</a>,
+        <a href="{{ url2 }}">Windows and macOS</a>,
+        <a href="{{ url3 }}">Android</a>and <a href="{{ url4 }}">linux</a>. Need
```
alexgibson: Is there a reason Linux goes to `/all` and not `/desktop`?

PR #4749, opened 2017-04-04
Fix bug 1353407: Upgrade to Django 1.8.18

pmac:


pmac:
Doesn't really need review, we just need to make sure tests pass. But if you happen to know of anything @glogiotatidis that I should have done then please put it here. But an approved review is required for merging regardless.

PR #4736, opened 2017-03-27
fix bug 1345338 Update titles+metadata on mozilla.org/firefox/*

FrancescoSTL:
## Description
Updating the following pages with [these](https://docs.google.com/spreadsheets/d/1kFpqkkssd-keOyVS13azQS_iJ8ESNtvqjZ5TLy5hBq0/edit#gid=0) meta-decriptions and titles:

- https://www.mozilla.org/en-US/firefox/products/
- https://www.mozilla.org/en-US/firefox/desktop/
- https://www.mozilla.org/en-US/firefox/ios/
- https://www.mozilla.org/en-US/firefox/private-browsing/
- https://www.mozilla.org/en-US/firefox/sync/
- https://www.mozilla.org/en-US/firefox/releases/
- https://www.mozilla.org/en-US/firefox/android/

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1345338#c12

## Testing
Ensure various meta titles and descriptions are updated as denoted [here](https://docs.google.com/spreadsheets/d/1kFpqkkssd-keOyVS13azQS_iJ8ESNtvqjZ5TLy5hBq0/edit#gid=0).

## Checklist
- [x] Requires l10n changes.

PR #4751, opened 2017-04-06
Bug 1345338 follow-up: update fallback title in firefox/channel iOS page

flodolo:
Currently falling back to a description, which doesn't make sense.

### DIFF
--- a/bedrock/firefox/templates/firefox/channel/ios.html
+++ b/bedrock/firefox/templates/firefox/channel/ios.html
@@ -10,7 +10,7 @@
 {% if l10n_has_tag('firefox_title_meta_bug1345338') %}
   {{ _('Try New Features in a Pre-Release iOS Browser | Firefox') }}
 {% else %}
-  {{_('Test beta versions of Firefox for iOS via Apple’s TestFlight program and help make our mobile browser for iPhone, iPad and iPod touch even better.')}}
+  {{ _('Download and test future releases of Firefox for desktop, Android and iOS.') }}
 {% endif %}
 {% endblock %}
 


bedrock/firefox/templates/firefox/channel/ios.html
```diff
@@ -9,8 +9,6 @@
 {% block page_title %}
 {% if l10n_has_tag('firefox_title_meta_bug1345338') %}
   {{ _('Try New Features in a Pre-Release iOS Browser | Firefox') }}
-{% else %}
-  {{_('Test beta versions of Firefox for iOS via Apple’s TestFlight program and help make our mobile browser for iPhone, iPad and iPod touch even better.')}}
```
craigcook: Looks like this was a mis-paste, there should have been a previous title as the fallback. Just removing this means it would fall back to an empty title. I think we need to restore the old title string but not sure what the l10n impact of that is (if it's already been removed from .lang files)

flodolo: There was no title before, I assume it falls back to a default title? 
https://github.com/mozilla/bedrock/pull/4736/files#diff-a56dc7204957db25760b0dfbf4e0979d

Note that it's not the only case, this page was missing a desc
https://github.com/mozilla/bedrock/pull/4736/files#diff-a56dc7204957db25760b0dfbf4e0979d

craigcook: You're right, I was looking at the iOS product page, not the channel page. Seems this one never had a unique title, it was just inheriting it from `channel/base.html`, which is what it will do again for locales without the tag. My mistake!

craigcook: I take that back again... the fallback title is gone from base.html so without this string the page has no title at all. Not ideal :/

craigcook: Here's the old title, maybe we need to restore this to base.html
https://github.com/mozilla/bedrock/blob/9b6dd4401ca5f7c66a2423c217beced6fbfb8e9b/bedrock/firefox/templates/firefox/channel/base.html#L10

flodolo: Argh, indeed. Restoring

flodolo: Restored title in base, amended commit message. The alternative would be to have an else here with that title, not sure which one is preferable.

craigcook: Looking at the desktop and android pages, they both have an else with the old string. We could do the same here (it looks like all these pages share the same index.lang file?) or update the other two templates to remove the else and use the same base fallback... either approach works but we should be consistent.

Sorry what seemed to be a simple fix has turned so messy :/

flodolo: Updated. Honestly, I'm not sure how much sense it makes to spend more time on this, there's a comment on the bug hinting that we need to change them all.

In the meantime I just want to make we don't start falling back to a crappy title just for localized pages.

craigcook: Agreed, it's not worth spending more time on. This is good. Thanks for your patience!

PR #4752, opened 2017-04-06
[fix bug 1353587] Global nav incremental improvements

alexgibson:
## Description
- Nav open state now locks scrolling the page (desktop viewport only).
- Close button in the drawer menu now animates (desktop viewport only).
- Improves a11y support for nav using WAI-ARIA roles.
- Deprecate custom GA events in favor of dynamic data-attributes (already tested on a demo with Peter).
- Updated styling for nav download button.
- Small style tweaks to main nav for mobile viewports.
- Pages under each top-level section are now highlighted on page load.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1353587

## Testing
- Test both pebbles and sandstone pages for any CSS regressions.
- Test wide desktop (locked nav) and smaller screens (regular page scrolling).

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/base/templates/base-pebbles.html
+++ b/bedrock/base/templates/base-pebbles.html
@@ -79,7 +79,7 @@
     {% endblock %}
   </head>
 
-  <body {% if self.body_id() %}id="{% block body_id %}{% endblock %}" {% endif %}class="html-{{ DIR }} {% block body_class %}{% endblock %}"{% block body_attrs %}{% endblock %}>
+  <body {% if self.body_id() %}id="{% block body_id %}{% endblock %}" {% endif %}class="html-{{ DIR }} {% block body_class %}{% endblock %}" {% block body_attrs %}{% endblock %}>
     <div id="strings"
       data-global-close="{{ _('Close') }}"
       data-global-next="{{ _('Next') }}"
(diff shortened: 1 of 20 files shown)


alexgibson:
@jpetto - OK I believe this is now final enough for code review. Leaving the dnm label on for now just in case.

alexgibson:
> Looks like the nav highlight still needs to be applied to /contact/ URLs. Also, should all URLs under /about/ be given the highlight? For example, /about/manifesto/?

Good catch, will update

> Would be nice if the "open" dots cursor (currently the hand) matched the "close" dots cursor (current the arrow). (Love the animation on the "close" dots!)

Will do 👍 

> There's probably a reason we can't lock scrolling to just the expanded nav on mobile, but I can't remember what it is.

`overflow: scroll` has pretty patchy support on mobile, especially on older devices. For something as important as the site navigation, I'd want to spend significantly more time than I have to testing on mobile devices. Maybe one day we can do it, but for now I'd rather play the safe bet and not make it unusable for certain devices.

alexgibson:
@jpetto thanks for the review, pushed some fixes.

alexgibson:
@jpetto - OK, Yulia asked for more changes to this, but I'm going to punt on these until the next iteration to concentrate on hub work. I think we're good to merge?

alexgibson:
Oh, let me squash commits first

PR #4755, opened 2017-04-07
[Fix bug 1352821] Add additional analytics tool

FrancescoSTL:
## Description
Adding analytics tool as described in https://bugzilla.mozilla.org/show_bug.cgi?id=1352821

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1352821

### DIFF
--- a/bedrock/base/templates/base-pebbles.html
+++ b/bedrock/base/templates/base-pebbles.html
@@ -3,7 +3,7 @@
 <!doctype html>
 {# Note the "windows" class, without javascript platform-specific
      assets default to windows #}
-<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_PROJECT_ID }}"{% endif %} {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %} {% if settings.STUB_ATTRIBUTION_RATE %}data-stub-attribution-rate="{{ settings.STUB_ATTRIBUTION_RATE }}"{% endif %}>
+<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.STATCOUNTER_PROJECT_ID %}data-statcounter-project-id="{{ settings.STATCOUNTER_PROJECT_ID }}"{% endif %} {% if settings.STATCOUNTER_SECURITY_ID %}data-statcounter-security-id="{{ settings.STATCOUNTER_SECURITY_ID }}"{% endif %} {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_PROJECT_ID }}"{% endif %} {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %} {% if settings.STUB_ATTRIBUTION_RATE %}data-stub-attribution-rate="{{ settings.STUB_ATTRIBUTION_RATE }}"{% endif %}>
   <head>
     <meta charset="utf-8">{# Note: Must be within first 512 bytes of page #}
 
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -3,7 +3,7 @@
 <!doctype html>
 {# Note the "windows" class, without javascript platform-specific
      assets default to windows #}
-<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_PROJECT_ID }}"{% endif %} {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %} {% if settings.STUB_ATTRIBUTION_RATE %}data-stub-attribution-rate="{{ settings.STUB_ATTRIBUTION_RATE }}"{% endif %}>
+<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.STATCOUNTER_PROJECT_ID %}data-statcounter-project-id="{{ settings.STATCOUNTER_PROJECT_ID }}"{% endif %} {% if settings.STATCOUNTER_SECURITY_ID %}data-statcounter-security-id="{{ settings.STATCOUNTER_SECURITY_ID }}"{% endif %} {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_PROJECT_ID }}"{% endif %} {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %} {% if settings.STUB_ATTRIBUTION_RATE %}data-stub-attribution-rate="{{ settings.STUB_ATTRIBUTION_RATE }}"{% endif %}>
   <head>
     <meta charset="utf-8">{# Note: Must be within first 512 bytes of page #}
 
--- a/bedrock/base/templates/base.html
+++ b/bedrock/base/templates/base.html
@@ -3,7 +3,7 @@
 <!doctype html>
 {# Note the "windows" class, without javascript platform-specific
      assets default to windows #}
-<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_PROJECT_ID }}"{% endif %} {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %} {% if settings.STUB_ATTRIBUTION_RATE %}data-stub-attribution-rate="{{ settings.STUB_ATTRIBUTION_RATE }}"{% endif %}>
+<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.STATCOUNTER_PROJECT_ID %}data-statcounter-project-id="{{ settings.STATCOUNTER_PROJECT_ID }}"{% endif %} {% if settings.STATCOUNTER_SECURITY_ID %}data-statcounter-security-id="{{ settings.STATCOUNTER_SECURITY_ID }}"{% endif %} {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_PROJECT_ID }}"{% endif %} {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %} {% if settings.STUB_ATTRIBUTION_RATE %}data-stub-attribution-rate="{{ settings.STUB_ATTRIBUTION_RATE }}"{% endif %}>
   <head>
     <meta charset="utf-8">{# Note: Must be within first 512 bytes of page #}
 
--- a/bedrock/base/templates/includes/stat_counter.html
+++ b/bedrock/base/templates/includes/stat_counter.html
@@ -1,0 +1,7 @@
+<noscript>
+  <div class="statcounter">
+    <a title="site stats" href="http://statcounter.com/" target="_blank">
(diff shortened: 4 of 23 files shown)


FrancescoSTL:
@jpetto @alexgibson any chance for a review on this?

alexgibson:
We need to wait on pushing this to prod until the Websites Privacy Notice has been updated. There should be a separate PR coming to take care of that.

bedrock/base/templates/includes/stat_counter.html
```diff
@@ -0,0 +1,7 @@
+<noscript>
+  <div class="statcounter">
+    <a title="site stats" href="http://statcounter.com/" target="_blank">
+      <img class="statcounter"src="//c.statcounter.com/11297457/0/d8541690/1/" alt="sitestats">
```
jpetto: We should keep our specific ID (which is `d8541690`, afaict) in the `settings` object to avoid having it hard-coded in these templates.

I suggest copying the way we do [`OPTIMIZELY_PROJECT_ID` in `base.py`](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L1113). (Feel free to place `STATCOUNTER_ID` just after `OPTIMIZELY_PROJECT_ID`.)

media/js/libs/stat-counter.js
```diff
@@ -0,0 +1,9 @@
+// If doNotTrack is not enabled, it is ok to add StatCounter
+// @see https://bugzilla.mozilla.org/show_bug.cgi?id=1217896 for more details
+if (typeof window._dntEnabled === 'function' && !window._dntEnabled()) {
+    var sc_project=11297457;
+    var sc_invisible=1;
+    var sc_security="d8541690";
+    var scJsHost = (("https:" == document.location.protocol) ? "https://secure." : "http://www.");
+    document.write("<script type='text/javascript' src='"+scJsHost+"statcounter.com/counter/counter.js'></script>");
+}
```
jpetto: Let's update this file to mimic `optimizely-snippet.js`.

- Move to `/media/js/base/`
- Place the `STATCOUNTER_ID` in a `data-*` attribute on the `<html>` element in the base templates
- Update to use the `data-statcounter-id` attribute instead of having our ID hard-coded here

FrancescoSTL:
Updated the PR @jpetto. Would we need to do the same with `sc_project`?

chrismore:
@alexgibson the PR for the privacy policy is here: https://github.com/mozilla/legal-docs/pull/773/files

As you mentioned, this should be merged and pushed before this overall is pushed.

jpetto:
Hm, this still isn't quite right. I see the task list as follows:

1. The only changes to the 3 base templates (`base-resp.html`, `base.html`, and `base-pebbles.html`) is to add `data-*` attributes for both the `project` and `security` values for StatCounter. E.g. `{% if settings.STATCOUNTER_PROJECT_ID %}data-statcounter-project-id="{{ settings.STATCOUNTER_PROJECT_ID }}"{% endif %}` to the `<html>` element. Look at how it's done for `OPTIMIZELY_PROJECT_ID` for reference.

2. Rename `STATCOUNTER_ID` to `STATCOUNTER_PROJECT_ID` in `settings/base.py`. Add `STATCOUNTER_SECURITY_ID` as well. Both of these should default to an empty string (just like `OPTIMIZELY_PROJECT_ID`).

3. Remove the `data-statcounter-id` attribute from the `` element in `includes/stat_counter.html`. (This is now taken care of in step 1 above.)

3. In all of the templates that require the StatCounter code (anywhere you've added `{% block stat_counter %}{{ super() }}{% endblock %}`), place the code currently (and incorrectly) in the base templates (wrapped in a switch) into the existing `js` block, making sure to retain any existing JavaScript bundles. If the template does not have a `js` block, feel free to add one. For example, on `/firefox/desktop/customize.html`, the code would look like:

```
{% block js %}
   <!--[if IE 9]>
   {% javascript 'matchmedia_addlistener' %}
   <![endif]-->
   {% javascript 'firefox_desktop_customize' %}

   {% if switch('statcounter-experiment') %}
      {% javascript 'stat_counter' %}
      {% include 'includes/stat_counter.html' %}
   {% endif %}
 {% endblock %}
```

4. Update `base/stat-counter.js` to pick up both the `project` and `security` IDs from the page's `<html>` element.

media/js/base/stat-counter.js
```diff
@@ -0,0 +1,11 @@
+var STATCOUNTER_ID = document.getElementsByTagName('html')[0].getAttribute('data-statcounter-id');
```
alexgibson: We should also wrap this in an IIFE, to avoid global variables.

media/js/base/stat-counter.js
```diff
@@ -0,0 +1,11 @@
...
+
+// If doNotTrack is not enabled, it is ok to add StatCounter
+// @see https://bugzilla.mozilla.org/show_bug.cgi?id=1217896 for more details
+if (typeof window._dntEnabled === 'function' && !window._dntEnabled()) {
+    var sc_project=11297457;
+    var sc_invisible=1;
+    var sc_security=STATCOUNTER_ID;
+    var scJsHost = (("https:" == document.location.protocol) ? "https://secure." : "http://www.");
+    document.write("<script type='text/javascript' src='"+scJsHost+"statcounter.com/counter/counter.js'></script>");
```
alexgibson: I know this may be in statcounter's original snippet, but `document.write` should generally be avoided for performance reasons. I'd suggest using `createElement` like we do for [other third party snippets](https://github.com/mozilla/bedrock/blob/master/media/js/base/optimizely-snippet.js)?

chrismore:
Thanks @jpetto and @alexgibson. @FrancescoSTL will be back online this evening to make revisions to the PR based on the feedback above.

media/js/base/stat-counter.js
FrancescoSTL: The analytic tool relies upon the sc_security and sc_project in the global scope. They are referenced from within `counter.js` (loaded by StatCounter) as `window.sc_project` and `window.sc_security`

media/js/base/stat-counter.js
FrancescoSTL: Sure :)

FrancescoSTL:
Fixed changes as mentioned above. Tested with `.env-dist` variables and all seems to be in order as expected.

@alexgibson I was unable to address the IFFE per comment [here](https://github.com/mozilla/bedrock/pull/4755#discussion_r111064105), however, I was able to replace `document.write` with `createElement`.

@jpetto, @alexgibson r?

media/js/base/stat-counter.js
alexgibson: If the script needs to reference some global variables you can still make them available on window e.g `var sc_security = window.sc_security = 'foo'` (Adding a comment to explain this would also be good). This should still allow you to use an IIFE.

FrancescoSTL:
If I'm not mistaken, CIrcleCI is only failing due to the env variable

alexgibson:
CircleCI is reporting several linting errors in the JS that need fixing:

```
/home/ubuntu/bedrock/media/js/base/stat-counter.js
   7:9   error  Identifier 'sc_project' is not in camel case       camelcase
   7:9   error  'sc_project' is assigned a value but never used    no-unused-vars
   8:9   error  'sc_invisible' is assigned a value but never used  no-unused-vars
   8:9   error  Identifier 'sc_invisible' is not in camel case     camelcase
   9:9   error  Identifier 'sc_security' is not in camel case      camelcase
   9:9   error  'sc_security' is assigned a value but never used   no-unused-vars
  12:22  error  Strings must use singlequote                       quotes
  12:31  error  Expected '===' and instead saw '=='                eqeqeq
  12:64  error  Strings must use singlequote                       quotes
  12:84  error  Strings must use singlequote                       quotes
  13:35  error  Strings must use singlequote                       quotes
```
If any of these can't be changed (for example stat counter needs underscores in variable names), you can add exceptions on a case by case basis. See the ESLint docs for more info: http://eslint.org/

Also please see my reply regarding the IIFE and how to handle global vars.

FrancescoSTL:
Addressed the CircleCI issues and IIFE, thanks @alexgibson! r? @alexgibson @jpetto

chrismore:
@alexgibson @jpetto We need to get this merged and pushed on Monday or Tuesday of this upcoming week. We are unblocked by legal as seen in this bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1353178

Looks like @FrancescoSTL got the latest changes to the PR in, so hopefully we can merge soon.

Thanks everyone.

bedrock/base/templates/base-pebbles.html
```diff
@@ -3,7 +3,7 @@
...
+<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.STATCOUNTER_PROJECT_ID %}data-statcounter-project-id="{{ settings.STATCOUNTER_PROJECT_ID }}"{% endif %} {% if settings.STATCOUNTER_SECURITY_ID %}data-statcounter-security-id="{{ settings.STATCOUNTER_SECURITY_ID }}"{% endif %} {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_SECURITY_ID }}"{% endif %} {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_PROJECT_ID }}"{% endif %} {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %} {% if settings.STUB_ATTRIBUTION_RATE %}data-stub-attribution-rate="{{ settings.STUB_ATTRIBUTION_RATE }}"{% endif %}>
```
jpetto: Looks like some copy pasta here - `data-optimizely-project-id` is included twice, with the first incorrectly referencing `settings.OPTIMIZELY_SECURITY_ID`.

bedrock/base/templates/base-resp.html
```diff
@@ -3,7 +3,7 @@
...
+<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.STATCOUNTER_PROJECT_ID %}data-statcounter-project-id="{{ settings.STATCOUNTER_PROJECT_ID }}"{% endif %} {% if settings.STATCOUNTER_SECURITY_ID %}data-statcounter-security-id="{{ settings.STATCOUNTER_SECURITY_ID }}"{% endif %} {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_SECURITY_ID }}"{% endif %} {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_PROJECT_ID }}"{% endif %} {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %} {% if settings.STUB_ATTRIBUTION_RATE %}data-stub-attribution-rate="{{ settings.STUB_ATTRIBUTION_RATE }}"{% endif %}>
```
jpetto: Same here - `data-optimizely-project-id` is included twice, initially with an invalid `settings` key.

bedrock/base/templates/base.html
```diff
@@ -3,7 +3,7 @@
...
+<html class="windows x86 no-js" lang="{{ LANG|replace('en-US', 'en') }}" dir="{{ DIR }}" data-latest-firefox="{{ latest_firefox_version }}" data-esr-versions="{{ esr_firefox_versions|join(' ') }}" {% if settings.STATCOUNTER_PROJECT_ID %}data-statcounter-project-id="{{ settings.STATCOUNTER_PROJECT_ID }}"{% endif %} {% if settings.STATCOUNTER_SECURITY_ID %}data-statcounter-security-id="{{ settings.STATCOUNTER_SECURITY_ID }}"{% endif %} {% if settings.OPTIMIZELY_PROJECT_ID %}data-optimizely-project-id="{{ settings.OPTIMIZELY_SECURITY_ID }}"{% endif %} {% if settings.GTM_CONTAINER_ID %}data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %} {% block gtm_page_id %}{% endblock %} {% if settings.STUB_ATTRIBUTION_RATE %}data-stub-attribution-rate="{{ settings.STUB_ATTRIBUTION_RATE }}"{% endif %}>
```
jpetto: `data-optimizely-project-id` assigned to an invalid `settings` key - `OPTIMIZELY_SECURITY_ID`.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-base-42.html
```diff
@@ -72,6 +72,11 @@ <h1 data-alt="{{ _('The world’s most private browsing. Only from Firefox.') }}
 
 {% block email_form %}{% endblock%}
 
-{% block js %}{% endblock %}
+{% block js %}
+  {% if switch('statcounter-experiment') %}
+    {% javascript 'stat_counter' %}
+    {% include 'includes/stat_counter.html' %}
+  {% endif %}
+{% endblock %}
```
jpetto: No need to update this template, as it is only used as a base for `whatsnew-42.html` and `whatsnew-zh-tw-49.html` (which are updated in this PR to contain the StatCounter stuff).

media/js/base/stat-counter.js
```diff
@@ -0,0 +1,20 @@
+(function () {
```
jpetto: Nit pick - add `'use strict'` as first directive in the IIFE.

bedrock/settings/base.py
```diff
@@ -1225,6 +1231,7 @@ def lazy_email_backend():
 CSP_IMG_SRC = CSP_DEFAULT_SRC + (
     'data:',
     '*.optimizely.com',
+    '*.statcounter.com',
```
jpetto: For ease of clean-up, lets move these CSP exceptions into a conditional like we do for `ad.doubleclick.net`. Remove `*.statcounter.com` and `www.statcounter.com` from the canonical lists and move into an `if` block at the bottom of the file (underneath the `if config('SWITCH_TRACKING_PIXEL', default=DEV, cast=bool):` block). Should look like:

```
# See https://bugzilla.mozilla.org/show_bug.cgi?id=1352821
if config('SWITCH_STATCOUNTER_EXPERIMENT', default=DEV, cast=bool):
    CSP_IMG_SRC += ('*.statcounter.com',)
    CSP_SCRIPT_SRC += ('www.statcounter.com',)
```

FrancescoSTL:
@jpetto fixed all issues above. Should be ready to merge if nothing more needs fixing, as we've been unblocked by legal. R?

jpetto:
Good points @pmac. I see a little benefit to the switch, as this is an experiment and may need to be toggled quickly and/or repeatedly (and we don't want the snippet code served if the experiment is off). I think we can roll as-is. (I'll handle the extra config'ing. 😄)

media/js/base/stat-counter.js
```diff
@@ -0,0 +1,22 @@
...
+    'use strict'
```
jpetto: Missing semi-colon here is the reason tests are failing. Will patch manually and merge.

pmac:
@jpetto very much agreed. I was just saying that the switch could be the ID setting instead of a separate `SWITCH_*` thing. But this will definitely work as I said.

chrismore:
Thanks @jpetto and @pmac ! Ping me in #www when this merge is deployed on production.

PR #4757, opened 2017-04-10
Bug 1353794: Add app to consume and store wordpress data

pmac:
Adds a generic view to include blog posts in a template context as well as methods on the model manager to easily filter by tags and blogs. If a blog post has featured media, you can get the URL to whatever size you want via a helper method on the model object.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -430,6 +430,7 @@ INSTALLED_APPS = (
     'bedrock.thunderbird',
     'bedrock.shapeoftheweb',
     'bedrock.utils',
+    'bedrock.wordpress',
     # last so that redirects here will be last
     'bedrock.redirects',
 
@@ -543,7 +544,18 @@ BLOG_FEEDS = {
     },
     'internetcitizen': {
         'url': 'https://blog.mozilla.org/internetcitizen/',
-        'name': 'Internet Citizen'
+        'name': 'Internet Citizen',
+    },
+}
+# same as above, but uses the JSON REST API and thus gets
+# more data (e.g. tags and featured image urls).
+# TODO: Once this rolls out for the Firefox blog, we should
+# convert the others and remove the old XML feed blogs above.
+WP_BLOGS = {
+    'firefox': {
+        'url': 'https://blog.mozilla.org/firefox/',
+        'name': 'The Firefox Frontier',
+        'num_posts': 20,
     },
 }
 
--- a/bedrock/wordpress/api.py
+++ b/bedrock/wordpress/api.py
@@ -1,0 +1,67 @@
+from __future__ import print_function, unicode_literals
+from django.conf import settings
+
+import requests
+from raven.contrib.django.raven_compat.models import client as sentry_client
+
+
+def _request(api_url, limit=None, page=1):
+    # 100 is max per page from WP
+    per_page = limit or 100
+    resp = requests.get(api_url, params={'per_page': per_page, 'page': page}, timeout=5)
+    resp.raise_for_status()
+    data = resp.json()
+    if limit is None and page == 1:
+        num_pages = int(resp.headers.get('x-wp-totalpages', 1))
+        if num_pages > 1:
+            for i in range(2, num_pages + 1):
+                data.extend(_request(api_url, page=i))
+
+    return data
+
+
+def _api_url(feed_url, data_type, data_id):
+    api_url = '{}/wp-json/wp/v2/{}/'.format(feed_url.rstrip('/'), data_type)
+    if data_id:
+        api_url += str(data_id)
+    return api_url
+
+
+def get_wp_data(feed_id, data_type, data_id=None, limit=None):
+    try:
+        feed_config = settings.WP_BLOGS[feed_id]
+        if data_type == 'posts' and limit is None:
+            limit = feed_config.get('num_posts', 20)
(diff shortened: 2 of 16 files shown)


pmac:
@jpetto @alexgibson I think this will meet your needs. You should after this lands be able simply change your view class for these pages to `BlogPostView`, set the blog name and tag(s) you want, and then `blog_posts` will be in the context for you to loop over.

Alternately you can just hard-code the template to start with and I can make a final pass to make it use the blog feed content :) An example view is below just for reference:

```python
class FirefoxPerformanceView(BlogPostView):
    blog_slug = 'firefox'
    blog_tags = ['browser', 'jank']
    template_name = 'firefox/performance.html'
```

alexgibson:
This sounds like it should suit our needs very well, thanks @pmac 💯

bedrock/wordpress/api.py
```diff
@@ -0,0 +1,64 @@
+from __future__ import print_function, unicode_literals
+from django.conf import settings
+
+import requests
+
+
+def _request(api_url, limit=None, page=1):
+    # 100 is max per page from WP
+    per_page = limit or 100
+    resp = requests.get(api_url, params={'per_page': per_page, 'page': page}, timeout=5)
```
glogiotatidis: I think `resp.raise_for_status()` would be helpful here and combined with the comment below in the exception handling block

bedrock/wordpress/api.py
```diff
@@ -0,0 +1,64 @@
...
+        if data_type == 'posts' and limit is None:
+            limit = feed_config.get('num_posts', 20)
+        api_url = _api_url(feed_config['url'], data_type, data_id)
+        if data_id:
+            data = _request(api_url, limit=1)
+        else:
+            data = _request(api_url, limit=limit)
+
+        return data
+    except Exception:
```
glogiotatidis: Maybe ping Sentry with the exception so it's actually logged somewhere that blog update fails?

bedrock/wordpress/models.py
```diff
@@ -0,0 +1,113 @@
...
+    tags = JSONField()
+
+    objects = BlogPostManager()
+
+    class Meta:
+        get_latest_by = 'date'
+        ordering = ['-date']
+
+    def __unicode__(self):
+        return '%s: %s' % (self.blog_name, self.title)
```
glogiotatidis: Make sure you're returning a unicode string `return u'%s: %s' % (self.blog_name, self.title)`

bedrock/wordpress/api.py
```diff
@@ -0,0 +1,64 @@
...
+        return None
+
+
+def get_posts_data(feed_id, num_posts=None):
+    posts = get_wp_data(feed_id, 'posts', limit=num_posts)
```
glogiotatidis: Does it make sense to [bleach](https://pypi.python.org/pypi/bleach) the data or we trust the source?

bedrock/wordpress/models.py
```diff
@@ -0,0 +1,113 @@
...
+                        date=post['date_gmt'],
+                        modified=post['modified_gmt'],
+                        title=post['title']['rendered'],
+                        excerpt=post['excerpt']['rendered'],
+                        link=post['link'],
+                        featured_media=post['featured_media'],
+                        tags=post['tags'],
+                    )
+                    count += 1
+                except DatabaseError:
```
glogiotatidis: What kind of errors do you expect here? I'm concerned that we may delete all `posts` in line 37 and then for some reason not be able to add new, thus leaving the db with zero posts and no notification that something is wrong.

bedrock/wordpress/models.py
glogiotatidis: Line https://github.com/mozilla/bedrock/pull/4757/files#diff-71ab2ee0e7addf70ea176c9c9c18fdc7R19 should also need `u`

Or better use `unicode_literals` as we do in other places to avoid running behind those `str` vs `unicode` errors.

bedrock/wordpress/tests/test_views.py
```diff
@@ -0,0 +1,43 @@
...
+
+import pytest
+import responses
+
+from bedrock.wordpress.models import BlogPost
+from bedrock.wordpress.tests.test_models import setup_responses, TEST_WP_BLOGS
+from bedrock.wordpress.views import BlogPostsView
+
+
+@responses.activate
```
glogiotatidis: 👍

bedrock/wordpress/models.py
pmac: I'm not sure what kind, but the idea was to try to prevent the scenario you mention. I thought by catching some strange error with a single post we could store the ones we could instead of canceling the whole thing. But, perhaps I should let the exception bubble up since that might cancel the transaction and thus not delete everything?

bedrock/wordpress/api.py
pmac: Hmm... good point... I think we trust the authors, but we shouldn't trust the security of the WP instance, and it could be a vector for script injection here if it were compromised. I'll add more bleach.

glogiotatidis:
r+!

bedrock/settings/base.py
```diff
@@ -430,6 +430,7 @@ def set_whitenoise_headers(headers, path, url):
     'bedrock.thunderbird',
```
glogiotatidis: 💯

PR #4759, opened 2017-04-10
[no bug] update legal docs

craigcook:
## Description
Updates legal docs submodule to latest.

I still have some misgivings about the de ->de-DE renaming. What say you @pmac ?


craigcook:
Oh I seem to have misread the original PR on legal-docs; it was renaming de-De -> de, not the other way around. So I think this in fact fixes an error that previously existed.

pmac:
I agree with your assessment. It also seems like we're not currently publishing the test pilot docs on the site, so I think the point is moot anyway. r+

PR #4761, opened 2017-04-12
Add pebbles base template for /firefox app

alexgibson:
@jpetto @craigcook - it occurred to me that we don't yet have a base template in the Firefox app that uses pebbles. I think it might be a good idea to create a common base template which all the new hub pages can inherit from. This gives us the following things:

- All pages have the correct Favicon / social meta data.
- For non-en locales, we'll likely have to fix up some sort of minimal header since they don't get the new global nav. For now I've just updated the default header in `./base-pebbles.html` to use the new logo. This did mean some slight refactoring of the existing header CSS on other pages, as there are so many versions of the same `#masthead` CSS :/

Thoughts?

Testing:
--------
- New base template for en-US should get the global nav, other locales the old nav (but with new logo).
- Test other pebbles page headers (non en) to make sure there are no regressions.

### DIFF
--- a/bedrock/firefox/urls.py
+++ b/bedrock/firefox/urls.py
@@ -118,4 +118,6 @@ urlpatterns = (
 
     url('^firefox/stub_attribution_code/$', views.stub_attribution_code,
         name='firefox.stub_attribution_code'),
+
+    page('firefox/foo', 'firefox/base-pebbles.html'),
 )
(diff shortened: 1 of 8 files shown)


bedrock/firefox/urls.py
```diff
@@ -118,4 +118,6 @@
 
     url('^firefox/stub_attribution_code/$', views.stub_attribution_code,
         name='firefox.stub_attribution_code'),
+
+    page('firefox/foo', 'firefox/base-pebbles.html'),
```
alexgibson: You can view the base template using this temporary URL (leaving this here so I don't forget to remove it again before merging).

PR #4763, opened 2017-04-12
Improve Wordpress blog support, port pages, and remove old system

pmac:
This is done in 3 commits:

1. Improve new wordpress support to handle existing pages and use cases. Also add all blogs to new system so the data will be there and ready when the updated views land.
2. Port the Technology and Internet Health pages (the only ones currently displaying blog posts) to the new system.
3. Remove the old system.

It's done this way so that we can push this change to prod in those three steps. It needs to be done this way to ensure a completely seamless transition.


bedrock/wordpress/models.py
```diff
@@ -111,6 +121,17 @@ def blog_link(self):
     def blog_name(self):
         return settings.WP_BLOGS[self.wp_blog_slug]['name']
 
+    def get_featured_tag(self, tags):
```
pmac: @alexgibson this is what I've got so far. `blog_tags` is also added to the context. Usage would be:

```jinja
<span class="featured-tag">{{ article.get_featured_tag(blog_tags) }}</span>
```

alexgibson: LGTM, thanks @pmac 💯

bedrock/wordpress/models.py
```diff
@@ -92,16 +107,11 @@ def get_absolute_url(self):
         return self.link
 
     def htmlify(self):
-        summary = strip_tags(self.excerpt).strip()
-        if summary.lower().endswith('continue reading'):
-            summary = summary[:-16]
-
-        return Markup(summary)
+        return Markup(self.excerpt)
```
glogiotatidis: Note that [according to the docs](https://docs.djangoproject.com/en/1.11/ref/utils/#django.utils.html.strip_tags) `strip_tags` is not to be considered safe.

pmac: Hmm... I can experiment with replacing this with bleach.

pmac: Updated to use bleach.

pmac:
@glogiotatidis with the update to bleach do you think we're good to merge?

PR #4764, opened 2017-04-12
[bug 1355512] Add Zilla Slab fonts

craigcook:
## Description
Adds Zilla Slab font files in both WOFF and WOFF2 and adds mixins/variables to Pebbles. This does't actually apply this styling to any pages yet but we'll start using Zilla on the new Firefox pages and will update some other pages shortly thereafter.

Some thoughts on the highlight variant... Where it's used in the mockups I've seen so far calls for additional filled space before or after the text. This can't be done with padding if you're using the highlight font because the space won't be filled; it must be done with space characters. That can't be done with spaces in HTML unless you use `&nbsb;` (which is presentational markup and generally icky and junks up strings for l10n). The better option is a pseudo-element in CSS and `content: '\a0\a0\a0';` (or however many spaces you need). Of course, if we have to use pseduo-elements for empty space it means we can't use pseudo-elements for any other purpose on the same element, like adding an icon. Space characters instead of padding also messes up text wrapping (additional lines will have no space).

So we'll need to be thoughtful about how and where we use the knock-out/highlight style vs a background color with filled text. If we end up using it as sparingly as I think we might, it could be better to move the highlight `@font-face` into its own include and only load that font when needed on specific pages.

Also note this adds files for a bunch of other weights/styles but we're so far only calling on regular, bold, and bold highlight.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1355512

## Testing
This PR doesn't apply Zilla anywhere so you'll need to add it to a page to really see it in action. Easiest test is to just change `$base-font-stack: $font-open-sans;` to `$base-font-stack: $font-zilla-slab;` in `pebbles/includes/_variables.scss`.


alexgibson:
We're going to have to be careful when using this on body copy, as it automatically transforms `mozilla` into `moz://a`:

bodycopy

alexgibson:
This PR looks good - agree we need to be careful when using the highlight weight. If it doesn't really end up being used on these new hub pages, we should switch it to its own include before they go out. As noted above, I think we're going to be unable to use this on body copy, unless there's a way to stop it turning `Mozilla` into `moz://a` (which is just our logo, not our company name).

PR #4767, opened 2017-04-13
[fix bug 1354710] Firstrun pages need to pass funnelcake parameters to FxA

alexgibson:
## Description
- Pass along funnelcake ID's along in the utm_source to the /firstrun FxA iframe.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1354710

## Testing
- If page contains funnelcake query param e.g. `?f=110`, then `utm_source` should equal `firstrun_f110`.
- If pages does not have funnelcake query param, `utm_source` should just be `firstrun`.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/firstrun/firstrun-horizon.html
+++ b/bedrock/firefox/templates/firefox/firstrun/firstrun-horizon.html
@@ -55,8 +55,10 @@
           {% endif %}
         </header>
         <div class="fxaccounts-container">
+          {# Bug 1354710: firstrun pages need to pass funnelcake parameters to FxA #}
+          {% set utm_source = 'firstrun' if not funnelcake_id else 'firstrun_f' + funnelcake_id %}
           <div class="fxaccounts" id="fxa-iframe-config" data-host="{{ settings.FXA_IFRAME_SRC }}" data-mozillaonline-host="{{ settings.FXA_IFRAME_SRC_MOZILLAONLINE }}">
-            <iframe id="fxa" scrolling="no" data-src="{{ settings.FXA_IFRAME_SRC }}?utm_campaign=fxa-embedded-form&amp;utm_medium=referral&amp;utm_source=firstrun&amp;utm_content=fx-{{ version }}&amp;entrypoint=firstrun&amp;service=sync&amp;context=iframe&amp;style=chromeless&amp;haltAfterSignIn=true"></iframe>
+            <iframe id="fxa" scrolling="no" data-src="{{ settings.FXA_IFRAME_SRC }}?utm_campaign=fxa-embedded-form&amp;utm_medium=referral&amp;utm_source={{ utm_source }}&amp;utm_content=fx-{{ version }}&amp;entrypoint=firstrun&amp;service=sync&amp;context=iframe&amp;style=chromeless&amp;haltAfterSignIn=true"></iframe>
           </div>
           <p>{{ _('With your Firefox Account you can sync your bookmarks, passwords, open tabs and more, and access them everywhere you use Firefox.') }}</p>
         </div>
--- a/bedrock/firefox/templates/firefox/firstrun/new-firstrun.html
+++ b/bedrock/firefox/templates/firefox/firstrun/new-firstrun.html
@@ -12,6 +12,8 @@
   {% javascript 'firefox_new_firstrun' %}
 {% endblock %}
 
+{% block global_nav %}{% endblock %}
+
 {% block content %}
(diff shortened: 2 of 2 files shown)


bedrock/firefox/templates/firefox/firstrun/new-firstrun.html
```diff
@@ -12,6 +12,8 @@
   {% javascript 'firefox_new_firstrun' %}
 {% endblock %}
 
+{% block global_nav %}{% endblock %}
```
alexgibson: I noticed this template does not inherit from the firefox app, so it was getting the nav incorrectly.

bedrock/firefox/templates/firefox/firstrun/firstrun-horizon.html
```diff
@@ -55,8 +55,10 @@ <h2 class="larger">{{ _('Sign in to Firefox and you’re good to go.') }}</h2>
           {% endif %}
         </header>
         <div class="fxaccounts-container">
+          {# Bug 1354710: firstrun pages need to pass funnelcake parameters to FxA #}
+          {% set utm_source = 'firstrun' if not funnelcake_id else 'firstrun_f' + funnelcake_id %}
```
jpetto: I wonder if it's worth asking in the bug if this change should remain indefinitely, or should be removed after this round of funnelcakes. I'm guessing the former, but 🤷‍♂️.

bedrock/firefox/templates/firefox/firstrun/new-firstrun.html
```diff
@@ -12,6 +12,8 @@
...
+
```
jpetto: I *think* [the bug is saying](https://bugzilla.mozilla.org/show_bug.cgi?id=1354710#c5) they want to use this template as well. If so, we'd need the same change as applied above, as well as some view logic.

Though they could mean they want to use the standard firstrun template, but incorporate the FxA UI/UX from FC 99...which appears to be triggered via the `entrypoint` param? Not clear to me.

bedrock/firefox/templates/firefox/firstrun/new-firstrun.html
jpetto: Oooh, nice catch.

bedrock/firefox/templates/firefox/firstrun/new-firstrun.html
alexgibson: I thought the same thing, but this template is only shown for f=99 and f=100. So I'm assuming they just mean the FxA UI that was shown in that past experiment.

bedrock/firefox/templates/firefox/firstrun/firstrun-horizon.html
alexgibson: OK, I'll ask in the bug.

bedrock/firefox/templates/firefox/firstrun/new-firstrun.html
jpetto: If that is the case (using `firstrun-horizon.html` with the FxA UI from f=99), wouldn't we need to conditionally update the `entrypoint` param as well? (I'm totally guessing at that `entrypoint` param here - it's the first noticeable difference I saw between the implementations.)

alexgibson: I think we're confusing things here maybe, but the bug isn't really clear. I believe the bug is saying they are showing the same FxA UI for the new experiments, so I don't think we need to touch the old template. The only ask is to change the `utm_source` from what I can tell.

bedrock/firefox/templates/firefox/firstrun/firstrun-horizon.html
alexgibson: [Reply in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1354710#c15) is that doing this for all funnelcakes would be useful.

bedrock/firefox/templates/firefox/firstrun/new-firstrun.html
alexgibson: I've asked in the bug for clarification on this.

alexgibson: [Reply in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1354710#c18) is that what we've done here should satisfy the requirements OK.

PR #4769, opened 2017-04-13
[fix bug 1349729] Add Optimizely snippet to /firefox/new/ variations.

jpetto:
## Description

Experiments need to be run on two variants of `/firefox/new/`.

Also, confirmed with cmore that Optimizely only needs to be on scene 1, and only for en-US audiences (for the time being, anyway).

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1349729#c21

## Testing

Make sure Optimizely loads as expected, and no typos or anything.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/new/break-free/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/break-free/scene1.html
@@ -4,6 +4,12 @@
 
 {% extends "firefox/new/break-free/base.html" %}
 
+{% block optimizely %}
+  {% if switch('firefox-new-optimizely', ['en-US']) %}
+    {% include 'includes/optimizely.html' %}
+  {% endif %}
+{% endblock %}
+
 {% block content %}
 <div class="main-content content">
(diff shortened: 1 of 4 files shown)


bedrock/firefox/templates/firefox/new/break-free/scene1.html
```diff
@@ -4,6 +4,12 @@
 
 {% extends "firefox/new/break-free/base.html" %}
 
+{% block optimizely %}
+  {% if switch('firefox-new-variant-optimizely') %}
```
jpetto: Adding a new switch name here so we don't have to load Optimizely on the canonical `/firefox/new/` while these experiments are running.

jpetto:
@craigcook - Well, things *did* change. Sorry for the double review. 😒

craigcook:
Still looks good to me. I repeat my r+ 🍳

PR #4770, opened 2017-04-17
Fix bug 1290962: Include locale in Fx download scene2 URL

pmac:
Add option to download_firefox helper to include the locale in the URL for the scene2 page.


pmac:
@alexgibson I know you're very busy with other things, but this would be a good one to fix. We're being pinged in the bug.

PR #4775, opened 2017-04-19
[fix bug 1355333] Update lang files for /whatsnew 50.

jpetto:
## Description

Updates lang files for `whatsnew-50.html` (as per @flodolo's request).


peiying2:
@jpetto, there is no line break in L81.

PR #4786, opened 2017-04-26
Update Firefox base template for new hub pages

alexgibson:
- Add global nav default highlight attribute
- Use correct Twitter / Facebook accounts in meta data
- Remove superfluous /firefox/new lang file


alexgibson:
@jpetto quick r?

PR #4791, opened 2017-04-28
Update Mozilla Location Service API key

hannosch:
Starting April 19th we've seen a massive increase in API calls with the old key. I'd like to rotate the key, to see if someone else is using the old key and possibly restrict its usage.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -1219,7 +1219,7 @@ STATSD_PREFIX = config('STATSD_PREFIX', default='bedrock')
 
 FIREFOX_MOBILE_SYSREQ_URL = 'https://support.mozilla.org/kb/will-firefox-work-my-mobile-device'
 
-MOZILLA_LOCATION_SERVICES_KEY = 'ec4d0c4b-b9ac-4d72-9197-289160930e14'
+MOZILLA_LOCATION_SERVICES_KEY = 'a9b98c12-d9d5-4015-a2db-63536c26dc14'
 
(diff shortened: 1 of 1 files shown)


alexgibson:
I would hazard a guess that this is because it's being used in a /whatsnew page: https://www.mozilla.org/en-US/firefox/53.0/whatsnew/.

Is this causing issues for you?

hannosch:
It's not a major issue, I've only noticed two weeks later after all :)

It's just that before this key was used for about 70,000 requests per day. Now it's used for 20 million requests per day. A 286x increase looks fairly suspicious. So far we got a total of about 60 million requests per day, the extra 20 million are noticeable.

If this is legitimate use, maybe there's some way to cache the country lookup on the client side. Most other users store the country value in local storage for 30 days and prefer that value before issuing a new request.

alexgibson:
> Most other users store the country value in local storage for 30 days and prefer that value before issuing a new request.

I think this is probably a good suggestion - I'll file a bug for this, thanks!

Do you still want to rotate the key here?

hannosch:
> Do you still want to rotate the key here?

Yeah, rotating the key doesn't hurt. It should make sure this is actually the what's new page and not someone else. This isn't time critical at all though.

alexgibson:
Tested locally and works as expected, thanks @hannosch 👍 

I [filed a bug for caching our responses](https://bugzilla.mozilla.org/show_bug.cgi?id=1360616). I'm not sure how much effect that would have for this particular case, as most users see this page once and then don't return. But I think it may be worth doing anyway for other pages.

PR #4795, opened 2017-05-01
Fix bug 1300373: redirect helper never return protocol relative URLs

pmac:
Manually prevent the redirect helper function from returning protocol relative URLs (e.g. //google.com/) as this could allow unexpected redirection to other domains and we never need to redirect to an unknown protocol.

Also add shortcuts that skip Django's `reverse` function when the input is obviously a URL.

### DIFF
--- a/tests/redirects/map_htaccess.py
+++ b/tests/redirects/map_htaccess.py
@@ -130,7 +130,13 @@ URLS = flatten((
     url_test('/en-US/firefox/help/', 'https://support.mozilla.org/'),
 
     # Bug 1255882
+    url_test('/some/url///', '/some/url/'),
+    url_test('////', '/en-US/'),
+    url_test('/en-US///', '/en-US/'),
     url_test('/de/firefox/about/', '/de/about/'),
+
+    # bug 1300373
+    url_test('/%2fgoogle.com//', '/google.com/'),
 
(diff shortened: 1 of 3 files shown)


pmac:
@jgmize here's another go at it. This should definitely fix all the things and there are more tests.

tests/redirects/map_htaccess.py
```diff
@@ -130,8 +130,14 @@
     url_test('/en-US/firefox/help/', 'https://support.mozilla.org/'),
 
     # Bug 1255882
+    url_test('/some/url///', '/some/url/'),
+    url_test('////', '/en-US/'),
+    url_test('/en-US///', '/en-US/'),
     url_test('/de/firefox/about/', '/de/about/'),
 
+    # bug 1300373
+    url_test('/%2fgoogle.com//', '/google.com/'),
```
floatingatoll: //%2f also?

pmac: Currently it looks like `https://www.mozilla.org//%2fgoogle.com//` redirects to an innocuous 404. I think we're okay with that one. I can add a test to make sure though.

floatingatoll:
As long as it stays at www.mozilla.org, it's fine to handle it however you
like.

PR #4803, opened 2017-05-03
[fix bug 1353581] New Firefox platform pages

craigcook:
## Description
New Firefox product pages for desktop, Android, and iOS. Feature pages (https://github.com/mozilla/bedrock/pull/4802) should merge first so we can update links on these pages.

English only for first release; not ready for l10n (some strings aren't considered final and will probably be changing in another iteration). Other locales should continue to get the old product pages for now.

## Bugzilla link
[Desktop](https://bugzilla.mozilla.org/show_bug.cgi?id=1353581)
[Android](https://bugzilla.mozilla.org/show_bug.cgi?id=1354600)
[iOS](https://bugzilla.mozilla.org/show_bug.cgi?id=1354603)

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/en-US/firefox/desktop/
https://bedrock-demo-craigcook.us-west.moz.works/en-US/firefox/android/
https://bedrock-demo-craigcook.us-west.moz.works/en-US/firefox/ios/

## Checklist
- [ ] Requires l10n changes.
- [ ] Related functional & integration tests passing.

PR #4804, opened 2017-05-04
[fix bug 1352139] Add Firefox Hub landing page.

jpetto:
## Description

Adds Fx hub landing page.

Working w/pmac to help get redirects in order...

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1352139

## Testing

Aside from the usual browser/responsive/UI checks, ensure `en-US` gets new `/firefox/` page, while all other locales get redirected to `/firefox/new/`.

https://bedrock-demo-jpetto.us-west.moz.works/en-US/firefox/

## Checklist
- [ ] Requires l10n changes.


bedrock/firefox/redirects.py
```diff
@@ -515,4 +515,10 @@ def firefox_channel(*args, **kwargs):
     # Bug 1277196
     redirect(r'^firefox(?:\/\d+\.\d+(?:\.\d+)?(?:a\d+)?)?/firstrun/learnmore/?$', 'firefox.features'),
     redirect(r'^firefox/windows-10/welcome/?$', 'https://support.mozilla.org/kb/how-change-your-default-browser-windows-10'),
+
+    # Bug 1352139
+    # Only en-US visitors get new /firefox/ page (for now...)
+    # This does not work...
+    # PMAC: halp pls
+    redirect(r'^((?!en\-US))\w{2,3}(?:-\w{2})?/firefox/?$', '{locale}/firefox/new/'),
```
pmac: I think we can just delete this and move the redirect to the new `/firefox/` view.

bedrock/firefox/urls.py
```diff
@@ -88,8 +89,7 @@
     url(tracking_protection_re, views.TrackingProtectionTourView.as_view(),
         name='firefox.tracking-protection-tour.start'),
 
-    # This dummy page definition makes it possible to link to /firefox/ (Bug 878068)
-    url('^firefox/$', views.fx_home_redirect, name='firefox'),
+    # PMAC: this removed line seems like it was superfluous before these changes, but i'm really not sure
```
pmac: seems fine

bedrock/firefox/views.py
```diff
@@ -526,3 +524,23 @@ def ios_testflight(request):
...
+class FirefoxHubView(BlogPostsView):
+    blog_posts_limit = 1
+    blog_posts_template_variable = 'articles'
+    blog_slugs = 'firefox'
+    blog_tags = ['home']
+    template_name = 'firefox/hub/home.html'
+
+    # Is this necessary with a redirect added to firefox/redirects.py?
+    # (The redirect in firefox/redirects.py doesn't work atm, so...)
+    def dispatch(self, request, *args, **kwargs):
```
pmac: I wouldn't override `dispatch`, I'd do `get`. It's not a big deal, but that would let the View class handle POST requests differently (in that they'd get a 405). The could would be roughly the same though.

bedrock/firefox/views.py
```diff
@@ -526,3 +524,23 @@ def ios_testflight(request):
...
+        locale = l10n_utils.get_locale(request)
+
+        # until we get localizable copy, non en-US locales continue to get
+        # redirect, though now it's a 302
+        if not locale == 'en-US':
```
pmac: NIT: This is a little confusing to read. Would be better as `if locale != 'en-US':`.

tests/redirects/map_globalconf.py
```diff
@@ -984,7 +985,8 @@
 
     # from mcom-tests
     url_test('/firefox/', '/en-US/firefox/'),
-    url_test('/en-US/firefox/', '/firefox/new/'),
+    # PMAC: changed locale below from en-US to fr, but test isn't passing
```
pmac: It will pass when you delete that redirects from `redirects.py`. Also you should keep the `en-US` test to verify the new destination.

tests/redirects/map_globalconf.py
```diff
@@ -22,6 +22,7 @@
     url_test('/pt/', '/pt-BR/'),
 
     # bug 795970 - lowercase to uppercase, e.g. en-us to en-US
+    # PMAC: Will these need changing with our reclamation of /en-US/firefox?
```
pmac: I don't think so. They still pass right?

bedrock/firefox/redirects.py
pmac: Which you've already done. The redirect you're doing in the view means we don't need this.

bedrock/firefox/urls.py
```diff
@@ -27,7 +27,8 @@
 
 
 urlpatterns = (
-    redirect(r'^firefox/$', 'firefox.new', name='firefox', locale_prefix=False),
+    # PMAC: after fixing firefox/redirects.py, this new url() should be ok/all that's needed?
```
pmac: Yes

bedrock/firefox/redirects.py
jpetto: Okay, cool. I wasn't sure if redirects were preferred in these files over in the view, or if it mattered at all.

jpetto:
@alexgibson Should be good to review. Demo is all updated. I can rebase with `firefox-hub-branch` after #4803 merges to work out the conflicts. (I have no idea why the `critters` are conflicting...)

media/css/firefox/hub/home.scss
```diff
@@ -0,0 +1,225 @@
...
+            float: none;
+            justify-content: center;
+            margin: 20px 0;
+            width: auto;
+        }
+    }
+}
+
+/* -------------------------------------------------------------------------- */
+// Features
```
alexgibson: Is the CSS for this section needed still seeing as we have `_card-blocks.scss`?

media/css/firefox/hub/home.scss
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    @media #{$mq-tablet} {
+        .card-content-wrapper {
+            @include at2x('/media/img/firefox/hub/sync-large.jpg', 100%, auto);
+        }
+    }
+}
+
+#newsletter-form {
```
alexgibson: Is this needed now we have `_outro.scss`?

media/css/hubs/_cards.scss
```diff
@@ -0,0 +1,309 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
alexgibson: Is this file doing the same thing as `_card_blocks.scss`?

media/css/hubs/_cards.scss
```diff
@@ -0,0 +1,309 @@
...
+            margin-right: auto;
+            padding-bottom: $card-padding * 3;
+            padding-right: $card-padding * 2;
+            padding-top: 60px;
+            width: 40%;
+        }
+    }
+}
+
+.three-card-monte {
```
alexgibson: It seems like there some duplication in this file that does the same thing as `_card_blocks.scss`, should we all converge on the same code? (both product and feature pages are using the same shared HTML/CSS).

alexgibson: Also I see why you've made this file in `/css/hubs/`, but I do wonder if this is perhaps a bit preemptive as we don't really know what the other hub pages will include just yet. Thoughts?

alexgibson:
Should these links be the burnt orange color?

link-color

media/css/hubs/_cards.scss
jpetto: I think we should converge this code, but I'd rather not dive into that now. We can do this during the refactor/code cleanup next sprint?

media/css/firefox/hub/home.scss
jpetto: Good catch. Removing.

media/css/hubs/_cards.scss
alexgibson: sgtm 🌮

jpetto:
> Should these links be the burnt orange color?

I guess? I don't have design direction in the Sketch file. I'll make them burnt orange with an underline (as burnt orange in a non-bold weight is a bit understated).

media/css/firefox/hub/home.scss
```diff
@@ -0,0 +1,205 @@
...
+    .form-contents,
+    .form-submit,
+    .fx-privacy-link a {
+        @include open-sans();
+    }
+
+    .fx-privacy-link a {
+        color: #9c432a;
+        text-decoration: none;
+    }
```
alexgibson: Nit: it would be nice if this link had an underline on hover/focus?

PR #4808, opened 2017-05-05
Restructure jenkins deployment. Demos build in main pipeline.

pmac:
* Fix demo deploy and update docs.
* Make it work with deis1 and deis2
* Enforce max 24 char app names
* Change demo prefix to "www-"
* Fix SSL redirect for Deis v2 clusters. Allow settings per cluster.
* Improve env file handling and combination for demo deploys

This is a big one, but makes deployment configuratin much nicer and more flexible. You now use .yml files specific to a branch. They all live under `jenkins/branches/branch.yml`. Demo builds are now prefixed with `demo/`, and so special demo configs are under `jenkins/branches/demo/branch.yml`. You can set a specific groovy file to run in the YAML by setting the `pipeline.script` value. And the big deal is that demos are now fully customizable. You can change any setting you want by adding an env file to e.g. `jenkins/branches/demo/branch.env`, and if you have a branch specific YAML file you can deploy your demo to any of our 4 deis clusters and run any of our tests (unit, smoke, integration) against it.

I've tried to update the docs as best I can. Please let me know what you think.


Jenkinsfile
```diff
@@ -1,216 +1,64 @@
...
-        setGitEnvironmentVariables()
-        setConfigEnvironmentVariables(config)
-        // save the files for later
-        stash name: 'scripts', includes: 'bin/,docker/'
-        stash name: 'tests', includes: 'tests/,requirements/'
-        stash 'workspace'
+def loadBranch(String branch) {
+    // load the utility functions used below
+    utils = load 'jenkins/utils.groovy'
+    is_demo_branch = ( env.BRANCH_NAME ==~ /^demo\/[\w-]+$/ )
```
jgmize: I see that you've [documented this change](https://github.com/mozilla/bedrock/pull/4808/files#diff-b4f2733694997b4f1983468feb1dc9ffR140) and this is intentional, and I agree with the new prefix aesthetically, but I'd suggest modifying the regex slightly to accommodate the existing branches with the current `demo__` prefix, perhaps something like `/^demo[\/_\w-]+$/` (untested)

pmac: Agreed. The deprecation will be tough, but I don't think it'll be because of the branch names, I think it'll be the app names that will be the issue. Unless we can upgrade Deis soon and go back to the old naming scheme I think a lot of demo URLs that perhaps other people outside our team are watching will cease to be updatable if we merge this. So I think perhaps it should be blocked on that Deis upgrade and I should remove the app renaming and 24 character check. You have thoughts on that?

pmac:
I've gone ahead and switched the app-name limit to 63 characters in expectation that the deis workflow upgrade will land before we need to deploy demos to those clusters. I also switched back to a demo app prefix of "bedrock-demo-" in order to rock less of the boat.

Jenkinsfile
pmac: I'm conflicted on supporting both demo branch naming schemes. It'd complicate the code a bit, and I really like the new way. Also branches are cheap and I feel like I can get the word out quickly.

pmac: I'll be sending out an email after this merges anyway to explain the new features:

* New demo branch naming scheme
* New branch-specific config files
* New branch-specific env files for updating the demo config without deis access

pmac:
@glogiotatidis I've just now realized that merely assigning someone as a reviewer in a PR is not sufficient to send that person a notification email from Github. Sorry about that. This PR is the one I was building when we talked last week. I'd love your feedback.

Jenkinsfile
jgmize: The [Deis upgrade](https://github.com/mozmar/infra/issues/213) is complete, unblocking this PR.

jgmize: I don't have any strong objections to breaking backwards compatibility, but I would recommend getting feedback from @alexgibson @jpetto and/or @craigcook as it would impact them.

metadave:
Deis Workflow has been [upgraded in Virginia and Tokyo](https://github.com/mozmar/infra/issues/213#issuecomment-299993174)

Jenkinsfile
pmac: It will only impact branches based on master after this lands. Existing demo branches will continue to deploy fine. But yes, feedback is always good.

jenkins/bin/cat_envs.py
```diff
@@ -0,0 +1,63 @@
+#!/usr/bin/env python
```
pmac: @jgmize @glogiotatidis thought about splitting this out into a separate small library as I'm guessing other projects might want to use it. Thoughts?

glogiotatidis:
> Edit: or maybe it does? I dunno. Regardless thanks for your r? when you get some time. Happy to hop on vidyo to discuss if you'd like as well.

It does!

docker/bin/push2deis.sh
```diff
@@ -8,7 +8,22 @@
...
+DEIS_BIN="${DEIS_BIN:-deis}"
+
+if [[ "$DEIS_BIN" == "deis2" ]]; then
+    if [[ "$DEMO_MODE" == "true" ]]; then
+        DOCKER_REPO="mozorg/bedrock_demo:${GIT_COMMIT}"
+    else
+        DOCKER_REPO="mozorg/bedrock:${BRANCH_NAME}-${GIT_COMMIT}"
+    fi
+else
+    DOCKER_REPO="$DOCKER_PRIVATE_REPO:${GIT_COMMIT}"
```
glogiotatidis: side talk: Do we plan to support private registries in deis2 somehow or are we going to use docker hub?

pmac: @jgmize should correct me if I'm wrong, but last I talked to him about it we were not planning to have private registries for deis2 as the efficiency gains for deis1 weren't that great.

jgmize: So far we've only been using public quay.io and docker hub, as the main use cases we had for private docker registries disappeared, and in the new infra the public registries perform better as each node can pull the image directly from the public registry in parallel, without having to first wait for the complete image to be copied to a cluster-local registry.

docker/bin/prep_demo.sh
```diff
@@ -0,0 +1,25 @@
...
+
+ENV_FILES=(
+  "jenkins/branches/demo/default.env"
+  "jenkins/regions/${DEIS_PROFILE}.env"
+  "jenkins/branches/${BRANCH_NAME}.env"
+)
+
+# reads which ever of the above files exist in order and combines values
+# pre-installed in jenkins
+ENV_VALUES=( $(envcat "${ENV_FILES[@]}") )
```
pmac: @jgmize @glogiotatidis I did go ahead and publish that env file combining utility as [envcat](https://github.com/pmac/envcat). I couldn't figure out how to install it in a nice way during this script run, so I've installed it globally onto the jenkins box. Please correct me if this isn't the right way to go here.

jenkins/default.groovy
```diff
@@ -0,0 +1,150 @@
...
+                                    }
+                                }
+                                sh 'docker/bin/push2deis.sh'
+                            }
+                        } catch(err) {
+                            utils.ircNotification([stage: stageName, status: 'failure'])
+                            throw err
+                        }
+                    }
+                    utils.ircNotification([message: appURL, status: 'shipped'])
```
pmac: I've slightly altered the notifications so that the 🚢 message means that the app has been deployed, and the 🎉 message below indicates that the integration tests have passed. Just FYI.

docker/bin/prep_demo.sh
jgmize: That's fine for a short term solution, but would you mind filing an issue (or even better a PR) against https://github.com/mozmar/ee-infra-jenkins/ to add envcat?

PR #4813, opened 2017-05-08
Fix Bug 1360812 - Font display issue with Fulah (missing glyphs on bold text)

kyoshino:
## Description

Add bold, semibold and italic font subsets for the Fulah language to solve the awkward text issue especially on the current [home page](https://www.mozilla.org/ff/). Also add missing WOFF2 subsets.

## Bugzilla link

[Bug 1360812](https://bugzilla.mozilla.org/show_bug.cgi?id=1360812)

## Testing

A [demo site](https://bedrock-demo-fulah-fonts.us-west.moz.works/ff/) is available, which has already been reviewed by a native localizer. Please just make sure the CSS is fine.

## Checklist
- [ ] Requires l10n changes.


fitojb:
This was one of the reasons I was appaled that somebody decided to replace Fira (which has great glyph coverage) with Open Sans (which has not)…

kyoshino:
We could eventually replace Open Sans with Noto Sans. I'll file a new bug for that. The font requires tests by localizers because it still has some issues with some languages.

PR #4823, opened 2017-05-09
[fix bug 1363411] Update Firefox hub page titles

alexgibson:
## Description
Updates Firefox hub page titles as per the original spec.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1363411

## Testing
Titles should match this doc https://docs.google.com/spreadsheets/d/1Xx5HxJhV3L304g67e3757uNWeuGqS6wu8eg-H93umdw/edit?ts=58fed2cd#gid=0

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/base-pebbles.html
+++ b/bedrock/firefox/templates/firefox/base-pebbles.html
@@ -4,7 +4,8 @@
 
 {% extends "base-pebbles.html" %}
 
-{% block page_title_prefix %}{{_('Mozilla Firefox Web Browser')}} — {% endblock %}
+{% block page_title_prefix %}{% endblock %}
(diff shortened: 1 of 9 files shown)


alexgibson:
> This has probably already been asked, but why is /features/private-browsing the only feature page with the | Firefox suffix? Some SEO reason?

I think it's just because the other two already mention Firefox as a keyword?

jpetto:
> This has probably already been asked, but why is /features/private-browsing the only feature page with the | Firefox suffix? Some SEO reason?
> 
> I think it's just because the other two already mention Firefox as a keyword?

Seems reasonable. SEO don't care about consistency.

PR #4829, opened 2017-05-10
[fix bug 1363543] /new variations to support ad campaigns - private, general, competitor

alexgibson:
## Description
- Three new variations for ad campaigns, based on previous templates.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1363543

## Testing
- Copy pasta errors are likely the biggest error prone thing here

https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/?xv=moreprotection
https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/?xv=workingout
https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/?xv=youdoyou

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/css/firefox/new/fx-lifestyle/browse-up-to-you.less
+++ b/media/css/firefox/new/fx-lifestyle/browse-up-to-you.less
@@ -5,6 +5,7 @@
 @import '../../../sandstone/lib.less';
 
 #intro {
+    background-color: #868b8f;
     background-image: url('/media/img/firefox/new/fx-lifestyle/browse-up-to-you/intro-high-res.jpg');
     background-position: center top;
     background-repeat: no-repeat;
--- a/media/css/firefox/new/fx-lifestyle/conformity-not-default.less
+++ b/media/css/firefox/new/fx-lifestyle/conformity-not-default.less
@@ -5,6 +5,7 @@
 @import '../../../sandstone/lib.less';
 
 #intro {
+    background-color: #784e66;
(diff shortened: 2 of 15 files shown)


media/css/firefox/new/fx-lifestyle/browse-up-to-you.less
```diff
@@ -5,6 +5,7 @@
 @import '../../../sandstone/lib.less';
 
 #intro {
+    background-color: #868b8f;
```
alexgibson: Added some background colors that match the pallet of each image. I noticed while working out in the garden that the pages are unreadable until the images load, which will likely hurt slower connections :/

media/css/firefox/new/fx-lifestyle/conformity-not-default.less
```diff
@@ -14,6 +15,10 @@
         p {
             .font-size(22px);
         }
+
+        .copy {
+            width: 500px;
```
alexgibson: Had to adjust the width here to accommodate slightly more copy, given that the images are all fixed height :/

alexgibson: We could totally change these to use `background-size: cover` which would be more flexible, but given that these are in english only I decided against it for now.

media/css/firefox/new/fx-lifestyle/browse-up-to-you.less
jpetto: Nice addition.

🏋️ 🏡

PR #4831, opened 2017-05-11
Remove app name suffix from deployment

pmac:
Was only needed for the "-deis" suffix on stage and prod but those are going away.

I've added a few more fixes to this one in subsequent commits:

* switch to dockerized envcat so we don't have to worry with installing it globally on the jenkins box
* more testing success message to a single one at the end of the build to make it a bit less chatty on IRC
* Remove old demo config file

### DIFF
--- a/bin/envcat
+++ b/bin/envcat
@@ -1,0 +1,3 @@
+#!/bin/bash
+
+docker run --rm -v $(pwd):/envcat mozmeao/envcat:latest "$@"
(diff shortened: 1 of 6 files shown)


pmac:
@jgmize r?

jgmize:
Removed bedrock-prod and bedrock-stage DNS entries in https://github.com/mozmar/infra/issues/254 so `curl https://bedrock-{stage,prod}.{eu,us}-west.moz.works` no longer returns invalid cert errors.

pmac:
Great. All that's required now is an approving review and merge. Thanks!

jgmize:
https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/demo%2Fpmac-new-pipeline/29/pipeline

PR #4835, opened 2017-05-12
Fix bug 1364452: Add ability to specify alternate canonical paths per locale

pmac:
I'd like to have this be more automatic based on an alternate template and the
languages activated for it so that locales would automatically migrate to the new URL
as locales are activated, but this should fix the immediate issue.


pmac:
@jpetto I think this will do it for now. PG thinks it will solve the immediate issue. We'll be introducing another of these on Tuesday next week though apparently with the deprecation of /firefox/products/. So we'll need to land this and add this technique to that view as well. All of this can wait until Monday though.

PR #4836, opened 2017-05-15
Fix Bug 1359102 - Release notes for Nightly need to display the Firefox Nightly logo + wording fixes

kyoshino:
## Description

* Change the logo and headline on the [Firefox Nightly notes](https://www.mozilla.org/en-US/firefox/nightly/notes/).
* Redesign the page by adopting the current [download page](https://www.mozilla.org/en-US/firefox/nightly/all/)'s theme.
* Repurpose Aurora's `space` visual style which is no longer used.

## Bugzilla link

[Bug 1359102](https://bugzilla.mozilla.org/show_bug.cgi?id=1359102)

## Testing

Check out the [demo server](https://bedrock-demo-nightly-notes.us-west.moz.works/en-US/firefox/nightly/notes/) to see everything looks right.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/css/sandstone/lib.less
+++ b/media/css/sandstone/lib.less
@@ -51,8 +51,6 @@
 @linkBlue:           rgb(0,150,221);
 @linkBlueHover:      rgb(0,83,159);
 @linkSkyBlue:        rgb(103,167,208);
-@linkSpaceBlue:      rgb(47,138,202);
-@linkSpacePurple:    rgb(147,111,188);
 
 @mozillaRed: #c13832;
 
--- a/bedrock/firefox/templates/firefox/all.html
+++ b/bedrock/firefox/templates/firefox/all.html
@@ -65,7 +65,7 @@
 
 {% block body_id %}firefox-all{% endblock %}
 {% block body_class -%}
-  sky {{ platform }} {{ channel }} {% if platform == 'desktop' and channel in ['alpha', 'nightly'] %}blueprint{% endif %}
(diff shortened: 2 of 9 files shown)


media/css/sandstone/lib.less
```diff
@@ -52,7 +53,7 @@
 @linkBlueHover:      rgb(0,83,159);
 @linkSkyBlue:        rgb(103,167,208);
 @linkSpaceBlue:      rgb(47,138,202);
-@linkSpacePurple:    rgb(147,111,188);
+@linkSpaceOrange:    rgb(255,173,31);
```
alexgibson: I'm not a UX or design person, but I find it a bit hard to distinguish what is and isn't a link here?

links

I think we should probably simplify this theme so both the body copy and headings are the same color (white), and then choose a single link color. Rather than add more new colors to our now (already out of date) sandstone pallet, I think we should probably choose from the existing link colors. If we change the page to have a dark themed footer (as seen on the [homepage](https://www.mozilla.org/en-US/), then we can just make the link color blue to match for example.

Thoughts?

alexgibson:
This PR also needs a rebase.

media/css/sandstone/lib.less
kyoshino: Agree. The headings should probably remain while, while the links could be `@linkSpaceBlue`.

kyoshino:
Rebased and revised the colour scheme.

PR #4839, opened 2017-05-16
[fix bug 1363708] Make Firefox hub pages 100% fluid width

alexgibson:
## Description
- Makes the existing Firefox hub pages 100% fluid width (based on feedback from creative).
- Consolidates some of the variables used in hub pages into `_lib.scss`

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1363708

## Testing
Demo: https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/

- Test responsiveness for grade-A browsers.
- Test responsiveness for older browsers that use fallbacks.

Note: I've barely touched the CSS for individual pages / components. What we wrote largely just seems "to work" 🎉 🌮 .

I'm also leaving the footer for now, as this is being redesigned soon anyway.

## Checklist
- [ ] Requires l10n changes.


craigcook:
This looks good as far as the sizing and responsiveness, but the side drawer in the global nav is wonky on wide browsers. I wonder if it's worth working on that here in combination with the liquification or if we should get this out first and tackle the drawer separately. Thoughts?

craigcook:
Another thought... I know we're focused on hub pages first but I wonder how much effort it would be to convert all of Pebbles to fluid width. Another thing we might not want to dive into in this PR but this does seem to be incurring some new debt.

alexgibson:
> but the side drawer in the global nav is wonky on wide browsers.

Can you explain what you mean by 'wonky'? Do you mean it floats off to the side? If that's the case, I explained this to Christa already. This is still the direction they want to go.

alexgibson:
> Another thought... I know we're focused on hub pages first but I wonder how much effort it would be to convert all of Pebbles to fluid width. Another thing we might not want to dive into in this PR but this does seem to be incurring some new debt.

I'm not doing that here and now, sorry. We'll have to save it for another time.

craigcook:
> Can you explain what you mean by 'wonky'? Do you mean it floats off to the side? 

Yeah, that expanse of whitespace between the content and the drawer. There are ways we might address it, but you're right, it means changing how the drawer works, way out of scope for this PR. We'll just have to live with it.

PR #4840, opened 2017-05-17
[fix bug 1363955] Switch Pebbles to Zilla Slab

craigcook:
## Description
This sets Zilla Slab as the base font for all Pebbles-based pages, with Open Sans as a secondary accent font. This started just to fix the fake italics on Firefox hub pages, but adding two more weights of Zilla on top of the four weights of Open Sans was too much. This way it's two fonts in and two fonts out. And we might as well make the leap to global Zilla sooner rather than later (just in Pebbles; no Zilla in Sandstone). 

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1363955

## Testing
https://bedrock-demo-craigcook.us-west.moz.works

This only impacts Pebbles-based pages, so that's the home page, all the new Firefox pages, /technology, /internet-health (and subpages), /about/leadership, /moss, /newsletter/mozilla, /newsletter/firefox, /newsletter/developer.... I think that's it.

### DIFF
--- a/media/css/pebbles/components/_modal.scss
+++ b/media/css/pebbles/components/_modal.scss
@@ -18,6 +18,18 @@ html.noscroll {
     #modal {
         position: absolute;
     }
+
+    @media #{$mq-tablet} {
+        height: auto;
+
+        body {
+            height: auto;
+        }
+
+        #modal {
+            position: fixed;
+        }
+    }
 }
 
 
--- a/media/css/pebbles/global.scss
+++ b/media/css/pebbles/global.scss
@@ -8,6 +8,7 @@
 
 @import
     // Fonts
+    'includes/fonts/zilla-slab',
     'includes/fonts/open-sans',
 
     // Base elements - general HTML elements
--- a/media/css/base/global-nav.less
+++ b/media/css/base/global-nav.less
@@ -324,8 +324,7 @@
     }
 
     .intro {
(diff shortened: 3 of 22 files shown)


media/css/pebbles/global.scss
```diff
@@ -8,6 +8,7 @@
 
 @import
     // Fonts
+    'includes/fonts/zilla-slab',
```
alexgibson: As we're adding this `@import` here, the existing imports can now also be removed from the existing hub pages?

- `media/css/firefox/features/common.scss`
- `media/css/firefox/hub/home.scss`
- `media/css/firefox/product-page.scss`

media/css/pebbles/components/_modal.scss
```diff
@@ -18,6 +18,18 @@ html.noscroll {
...
+
+    @media #{$mq-tablet} {
+        height: auto;
+
+        body {
+            height: auto;
+        }
+
+        #modal {
+            position: fixed;
```
alexgibson: I noticed that setting `body` to `position: relative;` also fixes the display issue, but I think this fix is probably better to make the modal behavior consistent 👍

media/css/pebbles/global.scss
craigcook: Good catch! Fixed.

PR #4845, opened 2017-05-22
Upgrade gunicorn and switch to meinheld worker

jgmize:
This should eliminate some gunicorn worker timeouts we're currently seeing in the k8s deployments, as it did for [snippets](https://github.com/mozmar/snippets-service/pull/234), [careers](https://github.com/mozmar/lumbergh/pull/194), and [MDN](https://github.com/mozilla/kuma/pull/4197). See https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/meinheld/6/pipeline/ for successful demo deployment and full test suite run.


metadave:
+1 from the SRE side of things

pmac:
Haven't figured out what's causing the test failures in the headless suite. They don't seem to be the same tests failing every time, which leads me to believe it's some kind of race condition or threading failure.

PR #4852, opened 2017-05-23
Tag all images as "latest" during prod push

pmac:
This will allow other jobs (configurator e.g.) to use
the ":latest" images for testing and other things.


docker/bin/push2dockerhub.sh
```diff
@@ -27,8 +27,6 @@ fi
 docker push $DOCKER_REPOSITORY:${DOCKER_TAG}
 
 if [[ "$GIT_TAG_DATE_BASED" == true ]]; then
-    docker tag $FROM_DOCKER_REPOSITORY:${DOCKER_TAG} $DOCKER_REPOSITORY:$GIT_TAG
```
jgmize: wait, why isn't this necessary any more?

pmac: I was thinking it was now happening in the `docker_build.sh` script now, but it could be a different name. Safer to leave it in I guess.

pmac:
updated

PR #4853, opened 2017-05-23
[fix bug 1364236] Pass original referrer from traffic cop to GA

jpetto:
## Description

Adds functionality to pass original referrer ([set in a cookie by Traffic Cop](https://github.com/mozilla/trafficcop/pull/5)) to GTM after a Traffic Cop redirect has taken place.

**Note** - Version of Traffic Cop in this PR matches that pending in [`mozilla/traffic-cop #6`](https://github.com/mozilla/trafficcop/pull/6)

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1364236

## Testing

Biggest things to check:

1. Correct info is being passed to GTM in `core-datalayer-page-id.js`
2. Changing namespace around `dntEnabled` hasn't broken anything
3. General sanity around changes

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/js/base/core-datalayer-page-id.js
+++ b/media/js/base/core-datalayer-page-id.js
@@ -6,7 +6,7 @@ if (typeof Mozilla == 'undefined') {
     var Mozilla = {};
 }
 
-(function() {
+(function(Mozilla) {
     // init dataLayer object
     var dataLayer = window.dataLayer = window.dataLayer || [];
     var Analytics = {};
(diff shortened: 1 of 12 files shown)


media/js/base/core-datalayer-page-id.js
```diff
@@ -22,12 +23,37 @@ if (typeof Mozilla == 'undefined') {
...
+    Analytics.getTrafficCopReferrer = function() {
+        // if referrer hasn't been set yet, try to grab it and then remove the cookie
+        if (Mozilla.Cookies && Mozilla.Cookies.hasItem('mozilla-traffic-cop-original-referrer')) {
+            trafficCopOriginalReferrer = Mozilla.Cookies.getItem('mozilla-traffic-cop-original-referrer');
+
+            // referrer shouldn't persist
+            Mozilla.Cookies.removeItem('mozilla-traffic-cop-original-referrer');
+        }
+
+        return trafficCopOriginalReferrer;
```
alexgibson: Hmm, this works but I found it a bit confusing when first reading the code. Calling `getTrafficCopReferrer()` twice (line 45 and 48 below) would imply that the cookie gets deleted the first time, however the original referrer is maintained because of the way the variable is scoped. Would this be easier to follow if we avoided side-effects, and let this function return the same thing each time?

PR #4854, opened 2017-05-24
[fix bug 1367270] Reword blog CTA in news feed

hoosteeno:
## Description
Rewords the blog CTA in the news feed from "The Firefox Frontier" to "Read More".

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1367270

## Testing
Passing: 
```(venv) bedrock$ py.test lib bedrock/firefox```

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/base/templates/includes/news-feed.html
+++ b/bedrock/base/templates/includes/news-feed.html
@@ -24,7 +24,7 @@
               <p>{{ article.htmlify() }}</p>
             </div>
             <p class="entry-cta">
-              <a href="{{ article.link }}?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=fx-content-hub" class="cta-link" data-link-name="{{ article.blog_title }}" data-link-type="link" data-link-position="Fresh from the blog">{{ article.blog_name }}</a>
+              <a href="{{ article.link }}?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=fx-content-hub" class="cta-link" rel="nofollow" data-link-name="{{ article.blog_title }}" data-link-type="link" data-link-position="Fresh from the blog">{{ _('Read more') }}</a>
             </p>
           </article>
         </li>

### CODE REVIEW

File: bedrock/base/templates/includes/news-feed.html [hunk 1]
```diff
@@ -24,7 +24,7 @@ <h3 class="entry-title">{{ article.blog_title }}</h3>
               <p>{{ article.htmlify() }}</p>
             </div>
             <p class="entry-cta">
-              <a href="{{ article.link }}?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=fx-content-hub" class="cta-link" data-link-name="{{ article.blog_title }}" data-link-type="link" data-link-position="Fresh from the blog">{{ article.blog_name }}</a>
+              <a href="{{ article.link }}?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=fx-content-hub" class="cta-link" rel="nofollow" data-link-name="{{ article.blog_title }}" data-link-type="link" data-link-position="Fresh from the blog">{{ _('Read More') }}</a>
```
alexgibson: Nit: should this just be `Read more` (lowercase m)? This seems consistent with casing for other cta links on the hub pages.

Also, should we also change `data-link-name="{{ article.blog_title }}"` to `data-link-name="Read more"`? The value for this data attribute should generally match the link text.

flodolo:
Please remember to tag pull requests with strings if they need to be translated.
Also note that we don't have such a string, and it's unclear to me if this template is used anywhere in localized pages.

alexgibson:
> Please remember to tag pull requests with strings if they need to be translated.
Also note that we don't have such a string, and it's unclear to me if this template is used anywhere in localized pages.

This PR does not require l10n changes, thanks @flodolo

hoosteeno:
@alexgibson good call on "Read more". 

Regarding data link, it is set to the article title so we know which article drew the click. It was set that way before this change, when the link text was "The Firefox Frontier", too. I think setting it to "Read more" would mean we grouped clicks on all 3 "Read more" links on the page, but the intent is to measure clicks per article. So I think we should leave as-is.

alexgibson:
Ah, right you are! I misread it for `blog_name`.

PR #4855, opened 2017-05-24
[bug 1353587] Make global nav accommodate text wrapping

craigcook:
## Description
The first iteration of the global nav had some hard-coded widths designed around the specific English links and didn't really handle links wrapping when they get lonk, which is almost guaranteed to happen once we start localizing those labels. This makes a few tweaks to the nav so when links wrap to a second line it doesn't look too horrible.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1353587

## Testing
Since the navbar is only English right now you'll have to change some content locally to really test the wrapping. Try adding longer text in existing links as well as additional links, both in the main nav and subnav.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/css/pebbles/components/_global-nav.scss
+++ b/media/css/pebbles/components/_global-nav.scss
@@ -9,9 +9,9 @@
 
 .moz-global-nav {
     @include open-sans;
-    position: relative;
     background: #fff;
     overflow: hidden;
+    position: relative;
 
     a:link,
     a:visited {
@@ -100,7 +100,7 @@
             height: 25px;
         }
 
-        @media #{$mq-nav-desktop} {
+        @media #{$mq-desktop} {
             margin-left: 17px;
 
             a {
@@ -114,16 +114,16 @@
     // Horizontal link menu container
     .nav-horizontal-menu {
         @include clearfix;
-        padding: 10px 0 10px;
-        position: relative;
-        overflow: hidden;
         margin-left: 5px;
+        overflow: hidden;
+        padding: 10px 0;
+        position: relative;
 
-        @media #{$mq-nav-tablet} {
+        @media #{$mq-tablet} {
             padding-top: 25px;
         }
 
-        @media #{$mq-nav-desktop} {
+        @media #{$mq-desktop} {
             margin-left: 17px;
         }
     }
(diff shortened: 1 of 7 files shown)


alexgibson:
The other thing that occurs to me, is now that both navs wrap better, we *could* start using the regular desktop and tablet breakpoints for when we do thinks like hide the primary links, instead of using custom breakpoints for the nav. 

We may need to do some additional tweaks to accommodate this better (e.g. only applying 40px right padding to nav links for wide desktop), but it may make our breakpoint code easier to follow in the long run. Thoughts?

craigcook:
D'oh, of course I forgot the Sandstone version. And I'll work on switching to the standard breakpoints.

craigcook:
I've updated the Sandstone/Less nav and switched to our regular breakpoints (also removing the special nav breakpoints since they were no longer used anywhere).

alexgibson:
Looks like some of the mobile-first padding values are back to front on the .less version of the global nav? (if only we could just use the same, uniform vale across the board 🙄)

padding

alexgibson:
@craigcook would you mind also pushing this to a demo? I can give things a check over for IE using browserstack, thanks.

alexgibson:
We should probably also update the sub-nav sticky break point to match these changes https://github.com/mozilla/bedrock/blob/master/media/js/hubs/sub-nav.js#L29

craigcook:
Switched padding around in Less (dangit), changed sub-nav.js breakpoint to 760px, and pushed to https://bedrock-demo-craigcook-nav.us-west.moz.works/

alexgibson:
Changes look good, although the right padding on links are still different between the scss and less versions (40px vs 20px)

media/css/pebbles/components/_global-nav.scss
```diff
@@ -515,6 +531,7 @@ html.moz-nav-open .moz-global-nav-page-mask {
 
         .download-title {
             font-weight: bold;
+            white-space: nowrap;
```
alexgibson: I see this here in the scss version, but not the less version? Should they both be the same?

craigcook: Hmm... well both should be the same, but now I'm not sure which way to go and maybe I should remove it from the SCSS. What do you think: let the button wrap or not?

alexgibson: Well, for really long strings this could create overlapping text at some screen sizes, right? If that's a possibility, I'd say wrapping may be a bit more robust?

craigcook: Removed!

PR #4858, opened 2017-05-25
[fix bug 1367793] Global nav links should behave like links

alexgibson:
## Description
- Makes the top horizontal links on the global link behave like, duh - links.
- No longer auto expand/collapse sub menu's on drawer toggle (since only direct clicks from the user now trigger state change).

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1367793

## Testing
- Ensure links behave like links, naturally.
- Ensure sidebar still works as normal via the menu button.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/base/templates/includes/global-nav.html
+++ b/bedrock/base/templates/includes/global-nav.html
@@ -22,19 +22,19 @@
           {% else %}
             {% set firefox_url = url('firefox.new') %}
           {% endif %}
-          <a href="{{ firefox_url }}" data-id="firefox" data-link-type="nav" data-link-position="top" data-link-name="expand" data-link-group="Firefox">{{ _('Firefox') }}</a>
+          <a href="{{ firefox_url }}" data-link-type="nav" data-link-position="top" data-link-name="Firefox">{{ _('Firefox') }}</a>
         </li>
         <li class="item-internet-health">
-          <a href="{{ url('mozorg.internet-health') }}" data-id="internet-health" data-link-type="nav" data-link-position="top" data-link-name="expand" data-link-group="Internet Health">{{ _('Internet Health') }}</a>
+          <a href="{{ url('mozorg.internet-health') }}" data-link-type="nav" data-link-position="top" data-link-name="Internet Health">{{ _('Internet Health') }}</a>
         </li>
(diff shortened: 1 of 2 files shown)


alexgibson:
Marking as do-not-merge for now, as Peter may want to A/B test this first.

alexgibson:
@jpetto OK, we have the green light from Christa to move forward with this change. Ready for review / merge.

bedrock/base/templates/includes/global-nav.html
```diff
@@ -25,16 +25,16 @@ <h2 class="nav-logo">
           <a href="{{ firefox_url }}" data-id="firefox" data-link-type="nav" data-link-position="top" data-link-name="expand" data-link-group="Firefox">{{ _('Firefox') }}</a>
```
jpetto: Should the Firefox link above have its attributes updated as well?

bedrock/base/templates/includes/global-nav.html
```diff
@@ -25,16 +25,16 @@ <h2 class="nav-logo">
...
         </li>
         <li class="item-internet-health">
-          <a href="{{ url('mozorg.internet-health') }}" data-id="internet-health" data-link-type="nav" data-link-position="top" data-link-name="expand" data-link-group="Internet Health">{{ _('Internet Health') }}</a>
+          <a href="{{ url('mozorg.internet-health') }}" data-link-type="nav" data-link-position="top" data-link-name="Internet Health" data-link-group="Internet Health">{{ _('Internet Health') }}</a>
```
jpetto: Do you think we still need `data-link-group`? It's now a repeat of `data-link-name`...

bedrock/base/templates/includes/global-nav.html
alexgibson: Ah, nice catch

bedrock/base/templates/includes/global-nav.html
alexgibson: Yeah I think this is redundant now, will just keep `data-link-name`.

alexgibson:
@jpetto thanks for the review - updated

PR #4860, opened 2017-05-26
[fix bug 1363304] Lazy load Firefox feature thumbnail images

alexgibson:
## Description
- Lazy loads /firefox/features/ thumbnails using [IntersectionObserver API](https://developer.mozilla.org/docs/Web/API/Intersection_Observer_API).
- Non supporting browsers load images normally (we could fallback to some JS library that does lazy load, but for now I think this is a good enough improvement).

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1363304

## Testing
- Demo: https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/features/
- Intersection Observer doesn't ship in Firefox until 54, so you'll need to test using Beta/Nightly for now.
- Also test on Chrome and Edge which both support Intersection Observer.
- Test noscript fallback.

## Checklist
- [ ] Requires l10n changes.


alexgibson:
@jpetto would be good to get your thoughts here in terms of approach. Not urgent, just when you have some free time.

alexgibson:
@jpetto we have some more feature pages coming down the pipe currently, so would be good to try and get this in sometime soon?

media/js/firefox/features-landing.js
```diff
@@ -0,0 +1,47 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    var supportsInsersectionObserver = typeof IntersectionObserver !== 'undefined';
+
+    function callback(changes, observer) {
```
jpetto: Nit - would be nice for this function to have a slightly less generic name. Maybe `observerCallback`?

media/js/firefox/features-landing.js
```diff
@@ -0,0 +1,47 @@
...
+        changes.forEach(function(change) {
+            if (change.intersectionRatio > 0) {
+                change.target.src = change.target.dataset.src;
+                change.target.onload = function() {
```
jpetto: Do we need to wait for the image to load before removing the `data-src` attribute?

media/css/firefox/features/common.scss
```diff
@@ -158,10 +160,32 @@ main {
...
         margin-bottom: 20px;
         max-width: 420px;
+        opacity: 1;
+        transition: opacity 0.3s;
         width: 100%;
     }
+
+    img[data-src] {
+    	opacity: 0;
+        display: none;
```
jpetto: There is no smaller nit than calling out an extra space.

bedrock/firefox/templates/firefox/features/index.html
```diff
@@ -203,7 +242,10 @@ <h2 class="section-title"><span>{{ _('Independence features') }}</span></h2>
       </li>
       <li class="features-list-item">
         <a href="{{ url('firefox.features.independent') }}" data-link-type="link" data-link-name="Protect Your Rights" data-link-position="independence 3">
-          <img src="{{ static('img/firefox/features/thumbnails/independent.jpg') }}" alt="">
+          <img src="{{ static('img/firefox/features/thumbnails/placeholder.png') }}" data-src="{{ static('img/firefox/features/thumbnails/independent.jpg') }}" alt="">
+          <noscript>
+            <img src="{{ static('img/firefox/features/thumbnails/independent.jpg') }}" alt="">
+          </noscript>
```
jpetto: This is getting close to a macro...

alexgibson: Hmm, I like this suggestion 👍

media/js/firefox/features-landing.js
alexgibson: This is just for the little fade-in, seems a bit nicer than just swapping out the image as it plops in.

media/js/firefox/features-landing.js
alexgibson: Good suggestion

media/js/firefox/features-landing.js
jpetto: Ah, yeah, agreed.

jpetto:
Btw, tested in Fx Nightly, Safari, Chrome, Edge, IE 11, and IE 9 and all did the right thing.

alexgibson:
@jpetto added a commit with some fixes. Thanks for suggesting the macro 👍

alexgibson:
Squashed :shipit:

PR #4863, opened 2017-05-29
[fix bug 1366892] remove Thunderbird version of Aurora from channel page

Sancus:
@alexgibson @craigcook Hey front-end friends, is there a class or better way to center the section .span(6)/.span_narrow(6) in this patch?

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1366892
## Testing
Testing merely involves checking if the Thunderbird Earlybird image and text is gone from the page at
/en-US/thunderbird/channel/
## Checklist


Sancus:
This is updated, I went with Width: 45% because 50% made the text flow differently. Also there didn't seem to be any need for the 760px - 1000px breakpoint anymore, so I removed it, and I also fixed the comments so they're actually accurate about what the break points are.

The <760px version already uses .span-all which includes width: auto; for really small screens, so it wasn't necessary to add it.

alexgibson:
Please remember to update functional tests for this page:

https://github.com/mozilla/bedrock/blob/master/tests/functional/thunderbird/test_thunderbird.py
https://github.com/mozilla/bedrock/blob/master/tests/pages/thunderbird/channel.py

Sancus:
@alexgibson Thanks! Did not even realize those were there. I've updated the PR accordingly, since we're just removing content, we're also just removing the test for that piece.

jgmize:
https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/36/pipeline

PR #4870, opened 2017-06-01
Fix Bug 1364470 - Update Country Codes in Hreflang tags

kyoshino:
## Description

Some locales on mozorg are not supported by Google and other search engines, and those `<link rel="alternate">` lines lead to errors on the SEM dashboard. This PR removes the known 7 locales from the alternate locale list.

## Bugzilla link

[Bug 1364470](https://bugzilla.mozilla.org/show_bug.cgi?id=1364470)

## Testing

Check the homepage, for example, to see the 7 locales are not included in the alternate locale list, while the language switcher at the bottom of the page still offers these locales.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/base/templates/includes/canonical-url.html
+++ b/bedrock/base/templates/includes/canonical-url.html
@@ -10,7 +10,7 @@
         {% if code == 'en-US' -%}
         <link rel="alternate" hreflang="en" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="English">
         <link rel="alternate" hreflang="en-CA" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="English (Canada)">
-        {% else -%}
+        {% elif code|length != 3 -%}{#- Bug 1364470: Drop ISO 639-2 and -3 locales not supported by Google -#}
         <link rel="alternate" hreflang="{{ code }}" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="{{ label|safe }}">
         {% endif -%}
       {% endfor -%}


bedrock/base/templates/includes/canonical-url.html
```diff
@@ -10,7 +10,7 @@
         {% if code == 'en-US' -%}
         <link rel="alternate" hreflang="en" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="English">
         <link rel="alternate" hreflang="en-CA" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="English (Canada)">
-        {% else -%}
+        {% elif code not in ['ast', 'cak', 'dsb', 'hsb', 'kab', 'lij', 'son'] -%}{#- Bug 1364470: drop locales not supported by Google -#}
```
pmac: I think we should keep this list in settings. It feels kindof hidden here. Settings are already in the template context, so it should be a simple addition to `settings/base.py` and change this to `elif code not in settings.LANGS_NON_GOOGLE` (or whatever the setting name ends up being).

pmac: I'd actually suggest something more like `settings.LANGS_NO_ISO_639_1` to be more specific.

kyoshino: I guess we can simply drop all three character locales here rather than defining a list.

pmac: Looks like the list isn't quite the same. These are the 3 letter locales in `settings.PROD_LANGUAGES`:

```python
['ach', 'ast', 'cak', 'dsb', 'hsb', 'kab', 'lij', 'ltg', 'mai', 'son']
```

But perhaps ach, ltg, and mai just aren't on any of these pages?

kyoshino: Yeah, so I've just updated the template to simply check the length of locale codes.

kyoshino: Found `ach` on [/firefox/new/](https://www.mozilla.org/en-US/firefox/new/) but anyway the code length check is enough I guess...

PR #4874, opened 2017-06-02
Remove Fx home blog feed functional test

jpetto:
## Description

As the home page blog feed presence is potentially very fickle, best to remove the test for now so as not to block test runs should a blog post not be available.

Kept the [property/region on the page class](https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/home.py#L24) in the event we want to reinstate a version of this test later on.

## Bugzilla link

Nope.

## Testing

Ensure tests still pass and no other code should be removed.

## Checklist
- [ ] Requires l10n changes.


alexgibson:
I'd say let's just remove the property selectors from the page object too, easy to add back if we need?

PR #4878, opened 2017-06-05
Fix Bug 1370337 - On /about/forums, replace mdn entries with links to Discourse

kyoshino:
## Description

Update the [Forums page](https://www.mozilla.org/en-US/about/forums/) to modify MDN forum entries so those have a link to Discourse instead of Google Groups.

## Bugzilla link

[Bug 1370337](https://bugzilla.mozilla.org/show_bug.cgi?id=1370337)

## Testing

Visit `/about/forums/` to see the 3 MDN forums have been replaced with the custom entries, while others remain the same.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/forums/forums.html
+++ b/bedrock/mozorg/templates/mozorg/about/forums/forums.html
@@ -4,6 +4,9 @@
 
 {% extends "mozorg/about-base.html" %}
 
+{# Bug 1370337: MDN forums have moved from Google Groups to Discourse #}
+{% set mdn_forums = ['mozilla.mdn', 'mozilla.dev.mdc', 'mozilla.dev.mdn'] %}
+
 {% block page_title %}{{ _('Forums') }}{% endblock %}
 {% block body_id %}about-forums{% endblock %}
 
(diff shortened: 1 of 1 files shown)


bedrock/mozorg/templates/mozorg/about/forums/forums.html
```diff
@@ -73,13 +76,23 @@
         <li id="{{ forum.dashed }}">
           <div class="summary">
             <h4>{{ forum.id }}<a href="#{{ forum.dashed }}">#</a></h4>
-            <p class="description">{{ forum.description|safe }}</p>
+            <p class="description">
+              {%- if forum.id in mdn_forums -%}
```
alexgibson: Curious why we're putting all this special casing in the template for this. Is there any reason why we can't just update [the source](https://github.com/mozilla/community-data/blob/master/forums/raw-ng-list.txt)?

kyoshino: We can update the source's description but the links below have to be hardcoded anyway, I guess.

alexgibson: I wonder if it's worth revising what they are trying to do here, so it fits better with Gerv's system. This works, but trying to keep all the info that gets updated in one place seems like the way to go.

alexgibson: Ah, I see what you mean about the hard coded links below, yes that probably makes sense. Hmm, I'm not sure what's best here. This works does work...

bedrock/mozorg/templates/mozorg/about/forums/forums.html
```diff
@@ -73,13 +76,23 @@
...
+                {{ _('Discussions about MDN Web Docs have moved to Discourse.') }}
```
alexgibson: Hmm, the bug does say that the [appropriate forum name](https://bugzilla.mozilla.org/show_bug.cgi?id=1370337#c0) should still be used in each description.

I wonder if another solution may be to remove these old forums from the original source file, and then we could maintain a hard-coded list here in the template. Might that work?

kyoshino:
Thanks! 😺

PR #4879, opened 2017-06-05
[fix bug 1366397] BATM - Basic animated landing page

craigcook:
## Description
Adds two new versions of the Firefox download page for the BATM campaign: one with simple animation and one static with a graphic. Also sets up an A/B/C test across three versions: the two new ones and one that already exists. Targeting English only so no l10n but there's a possibility of a German version if this test goes well.

~~Some of the JS is clunky and repetitive and could use some refactoring so marking this WIP for the moment but otherwise I think this is ready for an initial code review.~~

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1366397

## Testing
[Demo - animated](https://bedrock-demo-craigcook.us-west.moz.works/en-US/firefox/new/?xv=batmprivate&v=a)
[Demo - static](https://bedrock-demo-craigcook.us-west.moz.works/en-US/firefox/new/?xv=batmprivate&v=b)

## Checklist
- [ ] Requires l10n changes.


media/js/firefox/new/scene1-batm-anim.js
```diff
@@ -0,0 +1,84 @@
...
+        $('html').addClass('ready');
+    });
+
+    // Set desktop media query
+    if (typeof matchMedia !== 'undefined') {
+        mqDesktop = matchMedia('(min-width: 1000px)');
+    }
+
+    // Fire the points for desktop and up
+    if (mqDesktop.matches) {
```
alexgibson: This will throw an error in browsers that don't support `matchMedia` since the conditional on line 30 will leave it as `undefined`. You may need to move all the `matchMedia` logic inside the conditional that handles the feature detection.

bedrock/firefox/templates/firefox/new/batm/machine-a.html
```diff
@@ -0,0 +1,73 @@
...
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import google_play_button with context %}
+
+{% extends "firefox/new/batm/base.html" %}
+
+{% block experiments %}
+  {% if switch('experiment-firefox-new-batm-anim', ['en-US']) %}
+    {% javascript 'experiment_firefox_new_batm_anim' %}
```
alexgibson: Hmm, I don't think this traffic cop experiment will work the way this is set up. You need to trigger the experiment from a page that isn't already one of the variations. Traffic cop won't do this in order to [avoid potential infinite redirects](https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-traffic-cop.js#L98-L100). Is that right, @jpetto?

I'm a little confused why they want to do an experiment like this anyway. Can't they already control which variations they are sending traffic to directly from the ads, and then adjust as they monitor performance?

media/js/firefox/new/scene1-batm-anim.js
```diff
@@ -0,0 +1,84 @@
...
+        var point1Next = point1.find('.next');
+        var point2Next = point2.find('.next');
+
+        // Show the first point
+        point1.show();
+
+        // Hide point 1, show point 2
+        point1Next.on('click', function() {
+            point1.fadeOut(100);
+            point2.fadeIn(100);
```
alexgibson: Not a blocker, but it would be nice to move focus to the heading of the next popup when it opens?

bedrock/firefox/views.py
```diff
@@ -552,6 +552,10 @@ def new(request):
                 template = 'firefox/new/batm/nimble.html'
             elif experience == 'batmresist':
                 template = 'firefox/new/batm/resist.html'
+            elif experience == 'batma':
```
alexgibson: It would be nice to [add some tests](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/tests/test_views.py#L494-L504) for these new variations?

media/css/firefox/new/batm-machine.scss
```diff
@@ -0,0 +1,535 @@
...
+        opacity: 1;
+        transform: translateY(0);
+    }
+}
+
+html,
+body {
+    background: #ec8c32;
+    background: linear-gradient(135deg, #ec8c32 0%,#e5472c 100%);
+    min-height: 100%;
```
alexgibson: This isn't new to these variations, but I notice there's a 2px solid white top border on the pages left from the old tabzilla. Probably good to remove?

bedrock/firefox/templates/firefox/new/batm/machine-a.html
craigcook: I believe the idea is for the ads to link explicitly to `xv=batma` and Traffic Cop would then serve up of three variants. But if they can do that kind of randomization on the ad server, then we wouldn't need Traffic Cop at all. I'll check with Fabio.

bedrock/firefox/templates/firefox/new/batm/machine-a.html
```diff
@@ -0,0 +1,73 @@
...
+  {% endif %}
+{% endblock %}
+
+{% block optimizely %}
```
craigcook: I was told at the last minute to make sure Optimizely was included on these pages but I have no idea what their plan for that is. Presumably so they can do tests within tests, which sounds like madness to me. I'll bring this up with Fabio as well. We definitely shouldn't use Traffic Cop and Optimizely simultaneously.

media/css/firefox/new/batm-machine.scss
craigcook: Gone!

bedrock/firefox/views.py
craigcook: Oh yeah, tests. I really must learn how to write those some day. I can at least copy-paste.

bedrock/firefox/templates/firefox/new/batm/machine-a.html
craigcook: I've been assured there will be no nested tests, and the two snippets are behind separate switches so we could make sure they're never both on, but I think removing the optimizely snippet entirely will eliminate any risk. Once the traffic cop test is conclusive and we have a winning page, we can remove traffic cop and add optimizely so they can do more testing with the winner if they want to.

alexgibson: But I'm not sure the traffic cop test will work the way you have it set here? Maybe @jpetto may have some thoughts here.

media/js/firefox/new/experiment-firefox-new-batm-anim.js
```diff
@@ -0,0 +1,18 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    var cop = new Mozilla.TrafficCop({
+        id: 'experiment_firefox_new_batm_anim',
+        variations: {
+            'xv=batmprivate': 33,   // Control: basic page, no machine
+            'xv=batma': 33,         // animated machine
```
alexgibson: If batma is controlling the traffic cop experiment, then why is it a variation here? Perhaps this is what's confusing me. The experiment bundle also seems to be included in the batmb template.

craigcook: Yeah, sorry, I need to clean this up further. batmprivate is the control (I just found this out).

media/js/firefox/new/scene1-batm-anim.js
```diff
@@ -0,0 +1,96 @@
...
+
+        // Hide point 1, show point 2
+        point1Next.on('click', function() {
+            point1.fadeOut(100);
+            point2.fadeIn(100).attr('tabindex',-1).focus();
+
+            window.dataLayer.push({
+                'event': 'in-page-interaction',
+                'eAction': 'button click',
+                'eLabel': 'Point 1 "Next" button click'
```
craigcook: These GA bits were a late addition. I need to confer with Peter to make sure they'll work as-is or if I need to adjust.

craigcook:
The "batmresist" and "batmnimble" variants aren't currently receiving any traffic so I've removed them for cleanliness. There's an Optimizely test running on the "batmfree" variant at the moment so it has to remain in place for the moment, and we'll just make sure that test is turned off before we turn on traffic cop for these new variants. 

The plan is to run this animation test for a while to determine which page template is the winner (animated, static, or text only) then re-do those headline variants on the winning template for a separate test, planned to start around the 19th.

media/js/firefox/new/experiment-firefox-new-batm-anim.js
```diff
@@ -0,0 +1,18 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    var cop = new Mozilla.TrafficCop({
+        id: 'experiment_firefox_new_batm_anim',
+        variations: {
+            'xv=batmprivate': 33,   // Control: basic page, no machine
```
alexgibson: This still won't work. Hitting this page never triggers the traffic cop experiment, since it's listed as a variation (at least, it never works for me locally). You can't have `batmprivate` listed here if it's the page that acts as the traffic cop.

media/js/firefox/new/scene1-batm-anim.js
```diff
@@ -0,0 +1,96 @@
...
+        var point1Next = point1.find('.next');
+        var point2Next = point2.find('.next');
+
+        // Show the first point
+        point1.show();
+
+        // Hide point 1, show point 2
+        point1Next.on('click', function() {
+            point1.fadeOut(100);
+            point2.fadeIn(100).attr('tabindex',-1).focus();
```
alexgibson: Nit: space before `-1`

media/js/firefox/new/scene1-batm-anim.js
```diff
@@ -0,0 +1,96 @@
...
+                'event': 'in-page-interaction',
+                'eAction': 'button click',
+                'eLabel': 'Point 1 "Next" button click'
+            });
+        });
+
+        // Hide point 2, show point 3
+        point2Next.on('click', function() {
+            point2.fadeOut(100);
+            point3.fadeIn(100).attr('tabindex',-1).focus();
```
alexgibson: Nit: space before `-1`

media/js/firefox/new/scene1-batm-anim.js
```diff
@@ -0,0 +1,96 @@
...
+                'event': 'in-page-interaction',
+                'eAction': 'button click',
+                'eLabel': 'Point 2 "Next" button click'
+            });
+        });
+
+        // Hide all points and just show the clicked one
+        $('.point .open').on('click', function() {
+            allPoints.fadeOut();
+            $(this).next('.label').fadeIn(100).attr('tabindex',-1).focus();
```
alexgibson: Nit: space before `-1`

bedrock/firefox/templates/firefox/new/batm/resist.html
```diff
@@ -1,13 +0,0 @@
-{# This Source Code Form is subject to the terms of the Mozilla Public
```
alexgibson: Are we sure these variations will no longer be needed?

bedrock/settings/static_media.py
```diff
@@ -1386,6 +1392,19 @@
...
+    'firefox_new_scene1_batm_anim': {
+        'source_filenames': (
+            'js/firefox/new/scene1-batm-anim.js',
+        ),
+        'output_filename': 'js/firefox_new_scene1_batm_anim-bundle.js',
+    },
+    'experiment_firefox_new_batm_anim': {
+        'source_filenames': (
+            'js/base/mozilla-traffic-cop.js',
+            'js/mozorg/experiment-firefox-new-batm-anim.js',
```
alexgibson: Should this be `js/firefox/new/experiment-firefox-new-batm-anim.js`?

media/js/firefox/new/experiment-firefox-new-batm-anim.js
alexgibson: Hmm, this is also going to cause an issue since (once you remove this line) it will redirect to `?xv=batmprivate&xv=batma`. This isn't something traffic cop is set up to do, as it was designed to display variations based on the original URL, not redirect to another URL entirely.

craigcook: D'oh, now I get it... in the past we've done a "double control" where one variant was the current page, but that was still a redirect. I misunderstood how Traffic Cop was working in those instances.

So, ok... we want people to land at a URL, and 33% stay where they landed, 33% get sent to A and 33% get sent to B... but if where they landed is already an `?xv=` URL Traffic Cop won't know that and will try to add another query string onto it. So... this experiment is impossible without changes to Traffic Cop. 

I'm not sure what to do about this. We probably need to talk to Fabio and figure out what the next step is and if we can rethink the test. I wonder if we'll have to use Optimizely instead 😞

alexgibson: The only solution I can think of that wouldn't require changes to traffic cop would be to show these new template variations using a `v` parameter instead, so `?xv=batmprivate&v=batma` etc.

bedrock/firefox/templates/firefox/new/batm/resist.html
craigcook: They're only testing alternative headlines and according to Fabio that test isn't even planned to begin for a few weeks. It's possible these templates could come back as they are if the control is the winner, but if A or B is the winner we'd have to update these variants for the next headline test. We can leave them in place even though they're not getting traffic, they're not hurting anyone, I was just thought I'd clear out templates that aren't being used. 

I'm probably just piling more onto this already-too-complicated PR than I should. I'll put these back and we'll pretend it never happened :)

media/js/firefox/new/experiment-firefox-new-batm-anim.js
craigcook: Hm... that could work. I think there's GTM stuff configured specifically for "xv" but maybe we can work around that.

craigcook: I've updated to use old-fashioned `v=a` and `v=b` params and Traffic Cop seems to be working (also once I finally fixed the file page on the JS, doh).

bedrock/firefox/templates/firefox/new/batm/machine-a.html
```diff
@@ -0,0 +1,61 @@
...
+        <div class="label" data-point="Point 3, Opt out with Firefox">
+          <h2>{{ _('Opt out with Firefox') }}</h2>
+          <p>{{ _('Firefox doesn’t sell access to your personal information. From privacy tools to tracking protection, you control who sees what.') }}</p>
+        </div>
+      </li>
+    </ul>
+  </section>
+
+  <div class="download-cta">
+    <h1 id="main-header-copy" data-experience="batma">{{_('Browse against the machine')}}</h1>
```
alexgibson: Should this now be `data-experience="batmprivate"`?

bedrock/firefox/templates/firefox/new/batm/machine-b.html
```diff
@@ -0,0 +1,47 @@
...
+{% block page_css %}
+  {% stylesheet 'firefox_new_batm_machine' %}
+{% endblock %}
+
+{% block body_class %}scene-1 machine-static{% endblock %}
+
+{% block content %}
+<main role="main">
+  <header class="page-hero">
+    <h1 id="main-header-copy" data-experience="batmb"><span>{{ _('Browse against the machine') }}</span></h1>
```
alexgibson: Should this now be `data-experience="batmprivate"`?

media/js/firefox/new/scene1-batm-anim.js
```diff
@@ -0,0 +1,102 @@
...
+                    'eLabel': 'Show ' + point3.data('point')
+                });
+            }).attr('tabindex', -1).focus();
+        });
+
+        // Hide all points and just show the clicked one
+        $('.point .open').on('click', function() {
+            allPoints.fadeOut();
+            $(this).next('.label').fadeIn(100, function() {
+                window.dataLayer.push({
```
alexgibson: Not a blocker, but could we avoid some repetition by making these GA calls a function and passing in a parameter?

PR #4881, opened 2017-06-08
[fix bug 1365741] Fx54 /whatsnew: Add QR code for all zh-TW visitors.

jpetto:
## Description

*Most* of the changes here are new files that are direct copies of existing files. See below.

This is the first of many requests for the Fx54 WNP. This is the only one with clear requirements at the moment, and since Fx54 release is Tuesday, 6/13, it's best to get in what we know.

**This request:** Serve all zh-TW users the QR code instead of the send to device widget. Experiment run on Fx53 showed the QR code to perform better.

This PR also sets up files/directories in anticipation of other Fx54 WNP change requests, which essentially means copying files from the previous WNP iteration. 

Also removes completed experiment from Fx50 zh-TW template.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1365741#c1

## Testing

Ensure `/zh-TW/firefox/54.0/whatsnew/` gets the custom template showing the QR code. All other locales should get the default `/fx54/whatsnew-54.html` template, which at this point is a straight copy of the previous WNP.

## Checklist


bedrock/firefox/templates/firefox/whatsnew/fx54/whatsnew-54.html
```diff
@@ -0,0 +1,126 @@
...
+{% set play_store_url = 'https://app.adjust.com/2uo1qc?campaign=whats_new&amp;adgroup=Google_Play_Store&amp;creative=badge_en' %}
+
+{% block content %}
+<main role="main">
+  <div class="main-container">
+    <div class="horizon">
+      <div class="stars">
+        <div class="inner-container">
+          <header id="masthead">
+            <div id="tabzilla">
```
alexgibson: I know this is just reusing an old template, but can we just remove the old logo here? Seems a shame to be showing it on /whatnew

PR #4882, opened 2017-06-08
[fix bug 1365741] Fx54 /whatsnew: Add experiment for de.

jpetto:
## Description

Basically running the same experiment we ran for zh-TW on Fx 53 whatsnew, but now for de. 50% should get `?v=a`, which is the control with the send to device widget. 50% get `?v=b`, which is the QR code.

Also:

- reverts `data-link-name` on the update wordmark to `tabzilla` (at pg's request)
- moves app store badges further "below the fold" if either send to device widget or QR code are present

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1365741#c4

## Testing

In a browser with DNT disabled, make sure `/de/firefox/54.0/whatsnew/` loads the variations outlined above.

## Checklist
- [ ] Requires l10n changes.


jpetto:
Closing in favor of #4883

PR #4883, opened 2017-06-08
[fix bug 1365741] Fx54 /whatsnew: Add experiments for de, ja, & zh-CN.

jpetto:
## Description

**Experiment 1: de**
Basically running the same experiment we ran for zh-TW on Fx 53 whatsnew, but now for de. 50% should get `?v=a`, which is the control with the send to device widget. 50% get `?v=b`, which is the QR code.

**Experiment 2: ja & zh-CN**
`?v=c` is the control, and has no primary CTA (just app store badges). `?v=d` is the QR code.

Also:

- reverts `data-link-name` on the updated wordmark to `tabzilla` (at pg's request)
- moves app store badges further "below the fold" if either send to device widget or QR code are present (these are the "primary CTAs")

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1365741#c4
https://bugzilla.mozilla.org/show_bug.cgi?id=1365741#c6

## Testing

In a browser with DNT disabled, make sure `/de/firefox/54.0/whatsnew/` loads the variations outlined above. Also make sure `/ja/firefox/54.0/whatsnew/` and `/zh-CN/firefox/54.0/whatsnews/` load their specific variations outlined above.

Variations code in the view is a little weird, as it's a single view, but two different experiments. Breaking up by `?v=a/?v=b` and `?v=c/?v=d` seems to work just fine.

## Checklist

### DIFF
--- a/bedrock/firefox/templates/firefox/whatsnew/fx54/whatsnew-54-ja-zhcn-base.html
+++ b/bedrock/firefox/templates/firefox/whatsnew/fx54/whatsnew-54-ja-zhcn-base.html
@@ -1,0 +1,22 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/whatsnew/fx54/whatsnew-54.html" %}
+
+{% add_lang_files "firefox/whatsnew_50" "firefox/sendto" %}
+
+{% block experiments %}
+  {% if switch('firefox-whatsnew-54-ja-zhcn', ['ja', 'zh-CN']) %}
+    {% javascript 'experiment_whatsnew_54_ja_zhcn' %}
+  {% endif %}
+{% endblock %}
+
+{% block body_content %}
+  {% if variation == 'd' %} {# QR code #}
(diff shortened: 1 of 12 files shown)


jpetto:
Only marking as DO NOT MERGE while experiment percentages are verified. Can & should still be reviewed as-is.

bedrock/firefox/templates/firefox/whatsnew/fx54/whatsnew-54-ja-zhcn-base.html
```diff
@@ -0,0 +1,22 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/whatsnew/fx54/whatsnew-54.html" %}
+
+{% add_lang_files "firefox/sendto" "download" %}
```
flodolo: "download" should be removed, file doesn't exist anymore. And it should definitely load "firefox/whatsnew_50"

jpetto:
[Experiment percentages have been verified](https://bugzilla.mozilla.org/show_bug.cgi?id=1365741#c12). Good to merge after final review/squash.

PR #4886, opened 2017-06-09
[no bug] Use shared lang files in all Internet Health pages

flodolo:
My gut reaction is: let's never do this again. 

We ended up duplicating strings all over the place by adding a new page to a connected system every couple of weeks. 

This tries to fix the mess, and it would be blocking us from exposing new strings for the hub (we need it merged and pushed to prod), since it moved shared strings (nav menu, newsletter) to a shared file.

CC @peiying2

### DIFF
--- a/bedrock/mozorg/templates/mozorg/internet-health/index.html
+++ b/bedrock/mozorg/templates/mozorg/internet-health/index.html
@@ -4,7 +4,7 @@
 
 {% extends "base-pebbles.html" %}
 
-{% add_lang_files "mozorg/internet-health/index" %}
+{% add_lang_files "mozorg/internet-health/shared" %}
 
 {% block page_title %}{{ _('Protect Internet Health and Privacy with Mozilla') }}{% endblock %}
 {% block page_title_suffix %}{% endblock %}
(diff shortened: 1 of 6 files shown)


bedrock/mozorg/templates/mozorg/internet-health/index.html
```diff
@@ -4,7 +4,7 @@
 
 {% extends "base-pebbles.html" %}
 
-{% add_lang_files "mozorg/internet-health/index" %}
+{% add_lang_files "mozorg/internet-health/shared" %}
```
alexgibson: I'm not sure this will work, see: https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/views.py#L307-L310

flodolo: It's the same name as the template, there's no need to add the lang file. Can you give it a try locally? It seems to work just fine for me with DEV=False: if the lang file is activated I get the new hub, if it's not I fall back to the old internet-health page

flodolo: You can use this branch to avoid creating the files by hand 
https://github.com/mozilla-l10n/www.mozilla.org/tree/Jun8

alexgibson: You're right, tested this locally and it does work 💯

PR #4903, opened 2017-06-14
[fix bug 1359982] Add location tag to download button

alexgibson:
## Description
- Add an optional `download_location` parameter to the button helper, which can be used to specify the location of a button on any given page, passed to GA via a data attribute.
- Adds data attributes to all buttons on new style hub pages, plus the global and sub nav components (We're is not concerned with older legacy pages right now).
- Also adds data attributes to `/firefox/new/`
~- Still need to confirm this is what Peter expects to see, so do not merge for now.~

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1359982

## Testing
https://www-demo3.allizom.org/en-US/firefox/

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/bedrock/firefox/templates/firefox/new/onboarding/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/onboarding/scene1.html
@@ -50,11 +50,11 @@
             <section id="download-buttons">
               {# handles desktop users (windows, os x, linux) #}
               <div class="download-button-wrapper desktop" id="download-button-wrapper-desktop">
-                {{ download_firefox(alt_copy=_('Free Download'), locale_in_transition=True) }}
+                {{ download_firefox(alt_copy=_('Free Download'), locale_in_transition=True, download_location='primary cta') }}
               </div>
               {# handles Android and iOS users #}
               <div class="mobile download-button-wrapper" id="download-button-wrapper-mobile" data-upgrade-subtitle="{{_('Get it free on Google Play')}}">
-                {{ download_firefox(dom_id='download-button-mobile') }}
+                {{ download_firefox(dom_id='download-button-mobile', download_location='primary cta') }}
               </div>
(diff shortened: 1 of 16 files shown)


alexgibson:
Ok, still waiting on confirmation of events for the pages in question from Peter, but I think this should be ready for code review.

bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -142,10 +142,10 @@ <h2 class="good">{{_('Freedom is yours') }}</h2>
 
         <ul class="other-platforms-mobile">
           <li class="android">
-            {{ google_play_button() }}
+            {{ google_play_button(anchor_attributes={'data-location': 'other'}) }}
```
jpetto: Should this be `data-download-location`?

bedrock/firefox/templates/firefox/new/onboarding/scene1.html
```diff
@@ -120,10 +120,10 @@ <h2 class="forward">{{_('Go Forward') }}</h2>
 
         <ul class="other-platforms-mobile">
           <li class="android">
-            {{ google_play_button() }}
+            {{ google_play_button(anchor_attributes={'data-location': 'other'}) }}
```
jpetto: Should be `data-download-location`?

bedrock/firefox/templates/firefox/new/scene1.html
alexgibson: Ah yes, good catch - thanks

alexgibson:
@jpetto - got confirmation from PG in the bug that this is all looking good and ready when we are. I've added another commit covering the mobile buttons on scene2 and the new Focus page. One last check and then I think we're good to merge?

jpetto:
LGTM!

PR #4907, opened 2017-06-15
Fix bug 1369738: Add XML sitemap

pmac:
Add a management command to generate a sitemap and update deployment to generate and include it in the builds.

Thanks to @kyoshino in PR #1333 for most of the basis of this work.


pmac:
Still evaluating the output of this. [Demo server is up with example](https://bedrock-demo-pmac.us-west.moz.works/sitemap.xml). Will remove labels when it's approved. But the code is basically complete and could be reviewed if someone gets time.

hoosteeno:
There are 2 open questions about the sitemap that I don't know the answer to now, but I think we should know the answer to before we launch it.

1) Is there any benefit to including locale information _both_ inline _and_ in the sitemap? They are posited as alternatives by some online sources. Is either one better? What if they get out of sync? 
2) Where should x-default point? To the locale-free URL or to the default locale (en-US)?

There is a heap of conflicting information about all this and I have not untangled it yet. If you find authoritative, _recent_ information about it please share.

pmac:
> Is there any benefit to including locale information both inline and in the sitemap? They are posited as alternatives by some online sources. Is either one better? What if they get out of sync?

This is what Peter German (don't know his account here) and I have been discussing. I'm of the opinion that duplication could lead to errors and confusion, while PG has been advocating for including the localized URLs in the sitemap as well as the site in order to be thorough. I'm genuinely unsure which is the better option. For now I do have it working to generate the localized "hreflang" elements in the sitemap, but that could easily be removed.

> Where should x-default point? To the locale-free URL or to the default locale (en-US)?

I've seen conflicting information on that as well. Google's webmaster docs seem to suggest that those pointing to a redirect are just fine and expected, while other docs out there suggest the opposite. It's easy to do either way. The current state of the PR is that it has changed the `x-default` URLs to use the default site locale (en-US), but that differs from what we're doing in prod now.

hoosteeno:
> I'm of the opinion that duplication could lead to errors and confusion

I agree. I also don't think that leaving them off for v1 will cause harm. If you can leave the hreflang content out of the sitemap, without throwing away the code, I think that's the best option for this first release.

> The current state of the PR is that it has changed the x-default URLs to use the default site locale (en-US)

Let's keep it this way. After reading quite a few blog posts and knowledgebase articles and webmasters-stack answers about this, I think en-US is our default. In other words, it should look kind of like this:

https://github.com/hoosteeno/hreflangs/blob/master/en-US/index.html

pmac:
> I agree. I also don't think that leaving them off for v1 will cause harm. If you can leave the hreflang content out of the sitemap, without throwing away the code, I think that's the best option for this first release.

👍 

> Let's keep it this way. After reading quite a few blog posts and knowledgebase articles and webmasters-stack answers about this, I think en-US is our default. In other words, it should look kind of like this:

Cool. If you look at the source for the [home page of my demo](https://bedrock-demo-pmac.us-west.moz.works/en-US/) I believe you'll see it setup just like that.

pmac:
Okay, the sitemap on my demo (deploying now) is greatly simplified to only include the default URLs. Some of our URLs do not include the locale (e.g. /credits/) and that is also reflected in the sitemap. I changed the name of the JSON file to `default-urls.json` since it is no longer all of the URLs. Let me know what you think.

hoosteeno:
I just used this sitemap as the basis of a crawl using ScreamingFrog[0]. It found 7 404 pages, which we should just exclude:

```
https://www.mozilla.org/en-US/styleguide/pebbles/
https://www.mozilla.org/en-US/styleguide/docs/mozilla-accordion/
https://www.mozilla.org/en-US/styleguide/docs/send-to-device/
https://www.mozilla.org/en-US/styleguide/docs/mozilla-pager/
https://www.mozilla.org/en-US/foundation/annualreport/2014/
https://www.mozilla.org/en-US/press/press-inquiry/
https://www.mozilla.org/en-US/styleguide/all-download-buttons/
```

All the rest were 200s, 301s, or 302s. 👍 

[0] https://builtvisible.com/check-your-xml-sitemap-errors/

pmac:
@hoosteeno ones like `/en-US/firefox/desktop/fast/` are interesting at the moment. They do work in every language BUT en-US, while the new one that works ONLY in en-US is listed. I could potentially catch those and put them in the map in an alternate language (e.g. `de`). Thoughts?

bedrock/settings/base.py
```diff
@@ -258,6 +258,59 @@ def lazy_langs():
...
+    r'^firefox/notes/$'
+    r'^teach/$'
+    r'^thunderbird/(release)?notes/$'
+    r'^about/legal/impressum/$',
+    r'^security/announce/'
+]
+
+# Pages we do want indexed but don't show up in automated URL discovery
+# or are only available in a non-default locale
+EXTRA_INDEX_URLS = [
```
pmac: @hoosteeno Just added these manually as they redirect in "en-US", but not in "de".

hoosteeno: I think this approaches a hack (due to how we've rolled out en-US), but I don't see any alternative now. Let's see how it goes.

pmac: I don't disagree with any of that :shipit:

pmac:
To be clear @hoosteeno, are you happy enough with the current state of this one to get it merged and into prod?

pmac:
This is ready for review and merge now. We'll tweak more later unless new issues come up in review.

pmac:
Also, for future reference, the version of this branch that includes all locales in the sitemap is in a [branch on my fork](https://github.com/pmac/bedrock/tree/add-sitemap-xml-all-locales), should we need it.

hoosteeno:
> To be clear @hoosteeno, are you happy enough with the current state of this one to get it merged and into prod?

Yes, let's ship it. :shipit:

PR #4909, opened 2017-06-19
[fix bug 1363499] Experiment for funnelcakes 114-118.

jpetto:
## Description

Adds experiment for funnelcakes 114 - 118 to `/firefox/new` for en-US, Windows, desktop, non-Firefox users geolocated in the US.

(I checked with Hanno regarding the increased hits to `location.services.mozilla.com` and he said it wont be a problem.)

Must add following funnelcake config to env:

```
FUNNELCAKE_114_PLATFORMS=win
FUNNELCAKE_114_LOCALES=en-US
FUNNELCAKE_115_PLATFORMS=win
FUNNELCAKE_115_LOCALES=en-US
FUNNELCAKE_116_PLATFORMS=win
FUNNELCAKE_116_LOCALES=en-US
FUNNELCAKE_117_PLATFORMS=win
FUNNELCAKE_117_LOCALES=en-US
FUNNELCAKE_118_PLATFORMS=win
FUNNELCAKE_118_LOCALES=en-US
```

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1363499

## Testing

Visit `/en-US/firefox/new/` on Windows (or with `windows` hard-coded in `window.site.getPlatform()`) in a browser with DNT disabled. Should get one of 5 variations 95% of the time.

**DO NOT MERGE** - Funnelcakes are not yet active.

https://bedrock-demo-jpetto.us-west.moz.works/en-US/firefox/new/

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/js/firefox/new/experiment-search-retention-funnelcakes.js
+++ b/media/js/firefox/new/experiment-search-retention-funnelcakes.js
@@ -1,0 +1,99 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    var experimentId = 'experiment_search_retention_funnelcake';
+    var geoNonmatchCookie = experimentId + '_nonUS';
+
+    // IE 9 can't handle the truth
+    var isIELT10 = /MSIE\s[1-9]\./.test(navigator.userAgent);
+
+    // swiped from mozilla-client.js
+    var ua = navigator.userAgent;
+    var isLikeFirefox = /Iceweasel|IceCat|SeaMonkey|Camino|like\ Firefox/i.test(ua);
+    var isFirefox = /\s(Firefox|FxiOS)/.test(ua) && !isLikeFirefox;
+    var isMobile = /^(android|ios|fxos)$/.test(window.site.platform);
+
+    // check if cookies are enabled
+    var hasCookies = (typeof Mozilla.Cookies !== 'undefined' || Mozilla.Cookies.enabled());
+    // check if user already was served a variation
+    var hasVariationCookie = hasCookies && Mozilla.Cookies.hasItem(experimentId);
+    // check if user already failed geolookup
+    var hasGeoNonmatchCookie = hasCookies && Mozilla.Cookies.hasItem(geoNonmatchCookie);
+    // check if user is on windows
+    var isWindows = window.site.platform === 'windows';
+    // check if current URL has a funnelcake param (in the unlikely event of navigating directly)
+    var isFunnelcake = /^.*\?.*f=\d{3}.*/.test(document.location.search);
+    // check if DNT is detectable and off
+    var dntOk = (typeof Mozilla.dntEnabled === 'function' && !Mozilla.dntEnabled());
+
+    // 0. if visitor previously failed the geolookup, skip everything
+    if (hasGeoNonmatchCookie) {
+        return;
+    }
+
+    // 1. if visitor already has a cookie for this experiment, skip the expensive
+    // checks below and run experiment (uses variation in the cookie)
+    if (hasVariationCookie) {
+        runExperiment();
+    }
+    // 2. experiment criteria prior to expensive geolookup check:
+    //      a. is not currently on a funnelcake URL (sanity infinite loop check)
+    //      b. is on Windows
+    //      c. on IE 10 or greater
+    //      d. is not on Firefox
+    //      e. is not on a mobile browser
+    //      f. DNT is detectable and off
+    else if (!isFunnelcake && isWindows && !isIELT10 && !isFirefox && !isMobile && dntOk) {
+        // 2. check geolocation for US
+        var xhr = new XMLHttpRequest();
+
+        xhr.onload = function(r) {
+            // make sure status is in the acceptable range
+            if (r.target.status >= 200 && r.target.status < 300) {
+                var country;
+
+                try {
+                    country = JSON.parse(r.target.response).country_code.toLowerCase();
+                } catch (e) {
+                    country = 'none';
+                }
+
+                if (country === 'us') {
+                    runExperiment();
+                } else {
+                    if (hasCookies) {
+                        // store cookie for two days to avoid repeated lookups
+                        // that we know will fail
+                        var d = new Date();
+                        d.setHours(d.getHours() + 48);
+                        Mozilla.Cookies.setItem(geoNonmatchCookie, country, d);
+                    }
+                }
+            }
+        };
+
+        xhr.open('GET', '/country-code.json');
+        // must come after open call above for IE 10 & 11
+        xhr.timeout = 2000;
+        xhr.send();
+    }
+
+    function runExperiment() {
+        var landsman = new Mozilla.TrafficCop({
+            id: experimentId,
+            variations: {
+                'f=114': 9,
+                'f=115': 9,
+                'f=116': 34,
+                'f=117': 9,
+                'f=118': 34
+            }
+        });
+
+        landsman.init();
+    }
+})(window.Mozilla);
--- a/bedrock/firefox/templates/firefox/new/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/scene1.html
@@ -6,6 +6,12 @@
 
 {% extends "firefox/new/base.html" %}
 
+{% block experiments %}
+  {% if switch('firefox-new-search-retention-funnelcakes', ['en-US']) %}
+    {% javascript 'experiment_firefox_new_fc_search_retention' %}
+  {% endif %}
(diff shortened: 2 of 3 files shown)


media/js/firefox/new/experiment-search-retention-funnelcakes.js
```diff
@@ -0,0 +1,66 @@
...
+            if (r.target.status >= 200 && r.target.status < 300) {
+                var country;
+
+                try {
+                    country = r.target.response.country_code.toLowerCase();
+                } catch (e) {
+                    // unset var above is good enough
+                }
+
+                if (country === 'us') {
```
alexgibson: Hmm, If I'm in the UK and hit this page on a Windows machine, would this query the API on every page load? Would it be worth caching this response in a cookie perhaps?

media/js/firefox/new/experiment-search-retention-funnelcakes.js
```diff
@@ -0,0 +1,66 @@
...
+                    runExperiment();
+                }
+            }
+        };
+
+        xhr.open('GET', 'https://location.services.mozilla.com/v1/country?key=a9b98c12-d9d5-4015-a2db-63536c26dc14');
+        xhr.timeout = 3000;
```
alexgibson: Think we could drop this even lower, 2 seconds maybe?

media/js/firefox/new/experiment-search-retention-funnelcakes.js
```diff
@@ -0,0 +1,66 @@
...
+    var fcRe = /^.*\?.*f=\d{3}.*/;
+
+    // 0. if visitor already has a cookie for this experiment, skip the expenisve checks below
+    if (hasCookie) {
+        runExperiment();
+    }
+    // 1. check to make sure user is on windows and not fx
+    // 1.1 extra check around DNT so we don't do unnecessary geolookups
+    // 1.2 make sure user isn't on a variation
+    else if (!fcRe.test(document.location.search) && window.site.platform === 'windows' && !isFirefox && !isMobile && typeof Mozilla.dntEnabled === 'function' && !Mozilla.dntEnabled()) {
```
alexgibson: This is a bit of a long-winded conditional. Could we use some variables to make it shorter and a bit more readable?

media/js/firefox/new/experiment-search-retention-funnelcakes.js
```diff
@@ -0,0 +1,66 @@
...
+                    // unset var above is good enough
+                }
+
+                if (country === 'us') {
+                    runExperiment();
+                }
+            }
+        };
+
+        xhr.open('GET', 'https://location.services.mozilla.com/v1/country?key=a9b98c12-d9d5-4015-a2db-63536c26dc14');
```
alexgibson: This throws an error in older IE browsers due to same origin policy, do we need to exclude those?

screen shot 2017-06-20 at 10 36 10 am

media/js/firefox/new/experiment-search-retention-funnelcakes.js
jpetto: I wavered on this value quite a bit. Unsure of what is acceptably performant here. I could do 2s though. Can always up it if the service seems to be taking too long.

media/js/firefox/new/experiment-search-retention-funnelcakes.js
jpetto: Good suggestion. Will add a cookie.

media/js/firefox/new/experiment-search-retention-funnelcakes.js
jpetto: Good call. Will clean this up a bit.

jpetto:
@alexgibson Addressed feedback, but there are still issues with IE. I can't get 10 or 11 to load the geolookup script. I get `SEC7118: XMLHttpRequest for https://location.services.mozilla.com/v1/country?key=a9b98c12-d9d5-4015-a2db-63536c26dc14 required Cross Origin Resource Sharing (CORS).`. FWIW, this also happens for the send to device widget on `/fr/firefox/android`.

I checked the headers, and `location.services.mozilla.com` is sending `Access-Control-Allow-Origin: *` Further research required I suppose.

jpetto:
@alexgibson The funnelcakes were not ready, so hoping you can do final review tomorrow morning.

My biggest concern at this point is that @pmac's new `/country_code.json` endpoint hasn't been tested against CloudFlare. We'll just have to get it out to stage to ensure it's working.

If code looks good, we're fine to merge, as we have to flip a switch for this PR to have any effect.

pmac:
I have now done preliminary testing and it is working very well! I did the following:

```bash
$ curl https://www.mozilla.org/country-code.json
```

I did that from home, our servers in Oregon, and our servers in Ireland, and got the expected results! I believe we're good to go for a real run of this thing.

pmac:
Just to be extra thorough I did the same as above with my VPN connected to Sweden and then Japan and got correct results. I believe we're good to go. This only works when going through the primary domains (and thus CloudFlare). If you want to see an example failure you can:

```bash
$ curl https://bedroc-prod.us-west.moz.works/country-code.json
```

alexgibson:
I just tested

```
$ curl https://www.mozilla.org/country-code.json
```

Gives me:  `{"country_code": "GB"}` 👍

media/js/firefox/new/experiment-search-retention-funnelcakes.js
```diff
@@ -0,0 +1,99 @@
...
+    var experimentId = 'experiment_search_retention_funnelcake';
+    var geoNonmatchCookie = experimentId + '_nonUS';
+
+    // IE 9 can't handle the truth
+    var isIELT10 = /MSIE\s[1-9]\./.test(navigator.userAgent);
+
+    // swiped from mozilla-client.js
+    var ua = navigator.userAgent;
+    var isLikeFirefox = /Iceweasel|IceCat|SeaMonkey|Camino|like\ Firefox/i.test(ua);
+    var isFirefox = /\s(Firefox|FxiOS)/.test(ua) && !isLikeFirefox(ua);
```
alexgibson: This is currently throwing an error `TypeError: isLikeFirefox is not a function`

media/js/firefox/new/experiment-search-retention-funnelcakes.js
```diff
@@ -0,0 +1,99 @@
...
+            if (r.target.status >= 200 && r.target.status < 300) {
+                var country;
+
+                try {
+                    country = JSON.parse(r.target.response).country_code.toLowerCase();
+                } catch (e) {
+                    country = 'none';
+                }
+
+                if (country === 'us') {
```
alexgibson: Running locally this conditional is truthy for me, despite not being in the US. I'm guessing this is expected? (the prod end point seems to work fine for me).

media/js/firefox/new/experiment-search-retention-funnelcakes.js
```diff
@@ -0,0 +1,99 @@
...
+        runExperiment();
+    }
+    // 2. experiment criteria prior to expensive geolookup check:
+    //      a. is not currently on a funnelcake URL (sanity infinite loop check)
+    //      b. is on Windows
+    //      c. on IE 10 or greater
+    //      d. is not on Firefox
+    //      e. is not on a mobile browser
+    //      f. DNT is detectable and off
+    else if (!isFunnelcake && isWindows && !isIELT10 && !isFirefox && !isMobile && dntOk) {
```
alexgibson: Not a blocker for this experiment, but more a thought for the future. It feels like we're sticking a lot of logic in the head of one of our most critical pages here. It would be great to refactor this into some sort of JS lib (funnelcake helper?) so that it could have unit tests?

media/js/firefox/new/experiment-search-retention-funnelcakes.js
pmac: This is by design for the country-code.json endpoint when `DEV=True`. It will by default return `US`, but you can change what it returns with the `DEV_GEO_COUNTRY_CODE` setting. Whatever that is set to will be what the endpoint returns in DEV mode (which includes demos and www-dev). The only time the new endpoint works is when it's behind the CloudFlare CDN, so accessing it via the direct URLs (e.g. bedrock-prod.us-west.moz.works) will result in an error (which is actually a convenient way of testing the error state).

media/js/firefox/new/experiment-search-retention-funnelcakes.js
jpetto: D'oh. Should just be `!isLikeFirefox` on this line. Will fix.

media/js/firefox/new/experiment-search-retention-funnelcakes.js
jpetto: Totally agree. As we do more of these experiments, we'll get a better idea of common functionality we can factor out - but still try to keep the JS in the head as minimal & targeted as possible.

PR #4912, opened 2017-06-19
[fix bug 1366397] BATMAn - add variant param to download button

craigcook:
## Description
Adds the `v=` param to download buttons, and also loads the correct JS on the B version.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1366397

## Testing
Ensure the buttons on both page variants are getting both the `xv=batmprivate` and `v=a/b` parameters and that it's being passed through scene 2.

## Checklist
- [ ] Requires l10n changes.

### DIFF
--- a/media/js/firefox/new/scene1-batm-anim.js
+++ b/media/js/firefox/new/scene1-batm-anim.js
@@ -5,7 +5,9 @@
 (function($) {
     'use strict';
 
-    var exp = $('#main-header-copy').data('experience');
+    var title = $('#main-header-copy');
+    var exp = title.data('experience');
+    var variant = title.data('variant');
     var mqDesktop;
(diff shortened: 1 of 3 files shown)


media/js/firefox/new/scene1-batm-anim.js
```diff
@@ -17,7 +18,7 @@
     $('.download-link').each(function(i, link) {
         if (exp && link.href.indexOf('scene=2') > -1) {
```
alexgibson: Nit: suggest adding `if (exp && variant && link.href...)`

media/js/firefox/new/scene1-batm-anim.js
```diff
@@ -6,6 +6,7 @@
     'use strict';
 
     var exp = $('#main-header-copy').data('experience');
+    var variant = $('#main-header-copy').data('variant');
```
alexgibson: Nit: you can store `$('#main-header-copy')` as a variable to avoid querying the DOM twice here.

PR #4915, opened 2017-06-20
Switch to a bedrock geoip service using our CDN headers

pmac:
https://support.cloudflare.com/hc/en-us/articles/200168236-What-does-CloudFlare-IP-Geolocation-do-

### DIFF
--- a/bedrock/base/views.py
+++ b/bedrock/base/views.py
@@ -3,12 +3,48 @@ import logging
 
 from django.conf import settings
 from django.http import HttpResponse, HttpResponseBadRequest
+from django.views.decorators.cache import never_cache
 from django.views.decorators.csrf import csrf_exempt
-from django.views.decorators.http import require_POST
+from django.views.decorators.http import require_POST, require_safe
+
+from bedrock.mozorg.util import HttpResponseJSON
+from lib import l10n_utils
 
 from raven.contrib.django.models import client
 
-from lib import l10n_utils
+
+@require_safe
+@never_cache
+def geolocate(request):
+    """Return the country code provided by our CDN
+
+    https://support.cloudflare.com/hc/en-us/articles/200168236-What-does-CloudFlare-IP-Geolocation-do-
+
+    Mimics the responses from the Mozilla Location Service:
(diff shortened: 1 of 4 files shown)


pmac:
This should work with all of our code that currently works with mozilla location services. It won't work for demos or www-dev or any other instance that isn't behind CloudFlare, so that is definitely a down side. I could make it so that it could support both depending on a setting, but that would be a bit more code (not a lot, but more, and this seemed time sensitive). I can always followup later with that improvement and tests for the view.

pmac:
Hmm... thinking about it, I should add the service via this PR, then change the endpoints in another so that we can test the thing manually before putting it into use. I'll split it out now.

pmac:
Okay, now this PR only adds the new geolocation endpoint (`/country-code.json`). We can get this into stage and prod (the two instances behind CloudFlare) for testing if we do indeed want/need it. After that a PR to change over to using should be easy.

bedrock/base/views.py
```diff
@@ -3,12 +3,30 @@
...
+
+    https://support.cloudflare.com/hc/en-us/articles/200168236-What-does-CloudFlare-IP-Geolocation-do-
+    """
+    country_code = request.META.get('HTTP_CF_IPCOUNTRY', 'XX')
+    if country_code == 'XX' or len(country_code) != 2:
+        return HttpResponseJSON({'country_code': None}, status=404)
+
+    return HttpResponseJSON({
+        'country_code': country_code,
+    })
```
jpetto: Wondering if we could put an easy short-circuit here for local dev. Maybe have it look for a config setting? Or if `DEBUG=True` just return `us`?

pmac: Could definitely do that.... let me think on this...

pmac:
@jpetto went ahead and added a couple tests just to be safe :) Thanks for the r?

jpetto:
:shipit:

pmclanahan:
Circle having problems. I'm going to override the check and merge since this doesn't affect any prod pages.

PR #4945, opened 2017-07-06
[fix bug 1369144] Adjust Fx Hub headers & sub-headers

alexgibson:
## Description
- Updates headings in Pebbles to fit the [new typographic scale](http://craigcook.github.io/etc/typescale/).
- Changes all labels to sentance case
- Updates product headings to incorporate the word "Browser".
- Moves header CSS to a shared component.
- Updates nav and subnav CSS to accommodate RTL.

Note: I've lowered the base size of Zilla from 20px to 18px for the body copy on all hub pages. If we stuck exactly to the scale linked above, the base size should be 16px. For the regular Zilla weight however, 18px *looks* the same size as 16px in Open Sans. Zilla at 16px looks around 14px, and feels just too small. The two just aren't well balanced enough imho :/

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1369144
https://bugzilla.mozilla.org/show_bug.cgi?id=1375636

## Testing
https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/

This change effects hub pages mainly, although the heading size changes do effect all Pebbles pages, so a cursory check of all existing pages is worthwhile. Impact should be minimal as it's only font size changes.

### DIFF
--- a/bedrock/base/templates/includes/news-feed.html
+++ b/bedrock/base/templates/includes/news-feed.html
@@ -6,7 +6,7 @@
 {% if LANG.startswith('en') and articles %}
 <section class="news">
   <div class="content">
-    <h2 class="section-title"><span>{{ _('Fresh from our Blog') }}</span></h2>
+    <h2 class="section-title"><span>{{ _('Fresh from our blog') }}</span></h2>
     <ul class="news-feed">
       {% for article in articles %}
         <li>
--- a/bedrock/mozorg/templates/mozorg/developer/index.html
+++ b/bedrock/mozorg/templates/mozorg/developer/index.html
@@ -40,10 +40,8 @@
   <header class="header-intro">
     <div class="content">
       <div class="header-container">
-        <div class="header-content">
-          <h1><span>{{ _('Mozilla Developers') }}</span></h1>
-          <p class="tagline">{{ _('Developer tools, resources, videos and more') }}</p>
-        </div>
+        <h1><span>{{ _('Mozilla Developers') }}</span></h1>
+        <p class="tagline">{{ _('Developer tools, resources, videos and more') }}</p>
       </div>
     </div>
   </header>
@@ -76,7 +74,7 @@
 
   <section class="resources">
     <div class="content">
-      <h2 class="section-title"><span>{{ _('Developer Resources') }}</span></h2>
+      <h2 class="section-title"><span>{{ _('Developer resources') }}</span></h2>
 
       <div class="card-group">
         <div class="card card-image-large resource-mdn">
@@ -147,7 +145,7 @@
 
   <section class="events">
     <div class="content">
-      <h2 class="section-title"><span>{{ _('Events to Check Out') }}</span></h2>
+      <h2 class="section-title"><span>{{ _('Events to check out') }}</span></h2>
 
       <ul class="news-feed">
         <li>
@@ -177,7 +175,7 @@
 
   <section class="projects">
     <div class="content">
-      <h2 class="section-title"><span>{{ _('Projects We Love') }}</span></h2>
+      <h2 class="section-title"><span>{{ _('Projects we love') }}</span></h2>
 
       <div class="card-group">
         <div class="card card-double">
@@ -282,7 +280,7 @@
 
   <section class="videos">
     <div class="content">
-      <h2 class="section-title"><span>{{ _('Videos to Watch') }}</span></h2>
+      <h2 class="section-title"><span>{{ _('Videos to watch') }}</span></h2>
 
       <div class="card-group">
         <div class="card card-double">
@@ -357,7 +355,7 @@
 
   <section class="social">
     <div class="content">
-      <h2 class="section-title"><span>{{ _('Stay Connected') }}</span></h2>
+      <h2 class="section-title"><span>{{ _('Stay connected') }}</span></h2>
 
       <ul class="social-groups">
         <li>
--- a/media/css/hubs/_common.scss
+++ b/media/css/hubs/_common.scss
@@ -13,7 +13,7 @@ $color-link-blue-dark: #175a77;
 // Section titles
 
 .section-title {
-    @include font-size(24px);
+    @include font-size-level4;
     margin: 0 0 20px;
 
     span {
@@ -21,7 +21,7 @@ $color-link-blue-dark: #175a77;
         background-color: #000;
         color: #fff;
         line-height: 1.3;
-        padding: 0 5px 0 60px;
+        padding: 0 5px 0 40px;
     }
 }
 
@@ -29,7 +29,7 @@ $color-link-blue-dark: #175a77;
 // CTA link
 
 a.cta-link {
-    @include font-size(18px);
+    @include font-size-level6;
     border-bottom: 2px solid $color-link-blue;
     color: $color-link-blue;
     display: inline-block;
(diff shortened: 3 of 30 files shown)


alexgibson:
@craigcook this is still WIP and needs feedback from Christa, but I would very much appreciate your expertise here in terms of direction.

bedrock/base/templates/includes/news-feed.html
```diff
@@ -6,7 +6,7 @@
 {% if LANG.startswith('en') and articles %}
 <section class="news">
   <div class="content">
-    <h2 class="section-title"><span>{{ _('Fresh from our Blog') }}</span></h2>
+    <h2 class="section-title"><span>{{ _('Fresh from our blog') }}</span></h2>
```
alexgibson: This string is shared among several pages, but is only shown to English locales.

bedrock/mozorg/templates/mozorg/developer/index.html
```diff
@@ -81,7 +79,7 @@ <h3 class="card-title">{{ _('Firefox DevTools') }}</h3>
 
   <section class="resources">
     <div class="content">
-      <h2 class="section-title"><span>{{ _('Developer Resources') }}</span></h2>
+      <h2 class="section-title"><span>{{ _('Developer resources') }}</span></h2>
```
alexgibson: These titles are changing to sentence case. I don't believe the /developer hub page is being localized yet, so this *should* be fine?

bedrock/base/templates/includes/news-feed.html
flodolo: Confirmed, it was never extracted

bedrock/mozorg/templates/mozorg/developer/index.html
flodolo: Confirmed, not localized

craigcook:
Zilla Slab is a really compact font and Open Sans is a really open font (it's even in the name) so they're not really equivalent at the same sizes. Open Sans just takes up more space and was designed for screen readability at small sizes. Zilla Slab was designed as a "display" font intended for big headlines and wasn't optimized for readability at regular body sizes. We really shouldn't even be using it for body text but that's what we've got. So if we size everything big for Zilla Slab and someone blocks the font to fall back to Helvetica or Arial, everything will look too big. There's just no winning. 

We could add an 18px level in our scale between 4 and 5 but I really like the idea of keeping them aligned with HTML heading levels (the original inspiration). Maybe we can make level5 18px so level6 becomes 16px as the smallest possible heading. Then the 14px and 12px sizes probably shouldn't have names based on heading levels, maybe `$font-size-small` and `$font-size-extra-small`? In my ideal world we would never use the `font-size` mixin to size text arbitrarily, we would only use a predetermined set of standard sizes.

For further type nerding, I should also mention that when I first experimented with Zilla I found setting serif as the fallback meant you got Times and that just looked hideously incompatible. So even though Zilla is technically a serif face (slab serif) I found it better to fall back to a sans-serif in the font stack. Chunky slab serifs mixed with no serifs looks better than chunky slab serifs mixed with thin, pointy serifs.

alexgibson:
Ok, I've updated this as follows:

- The base font size for all pebbles pages has now changed from `100%` (implied 16px as this is what browsers default to) to `18px`.
- `remify` function now calculates rems using `$rems: $pixels / 18px;`
- Top navigation menus and site footer and newsletter forms (which use Open Sans), are now explicitly set to `16px`.

This allows all new pages to automatically use `18px` as the base font size for Zilla. We just need to remember going forward to set Open Sans to 16px where appropriate.

media/css/pebbles/base/elements/_document.scss
```diff
@@ -6,12 +6,12 @@
 
 
 html {
-    font-size: 100%;
+    font-size: 18px;
```
craigcook: Been thinking about this a bit more and doing some testing. If we declare `18px` explicitly on both `html` and `body` it will override user preferences. But if we set `html` to `112.5%` (which comes out to 18px for the vast majority who don't change the 16px default) then use the font-size mixin on `body` (`@include font-size(18px);`) we should get the overall embiggening while still respecting the user's settings. Thoughts?

alexgibson: Yeah I was a little wary of doing this as well, but I think your suggestion sounds sensible. I'll play with it and see how it looks.

alexgibson:
@craigcook updated

alexgibson:
@craigcook I know you had some comments still to make, but it would be great to try and get this one merged soon if you can find some time. Thanks

media/css/pebbles/base/elements/_document.scss
```diff
@@ -6,12 +6,12 @@
...
 
 html {
-    font-size: 100%;
+    font-size: 112.5%;
     background: #fff;
 }
 
 body {
-    font-size: 100%;
+    @include font-size(18px)
```
craigcook: MIssing a `;` here. I'm surprised stylelint doesn't complain...

media/css/hubs/_masthead.scss
```diff
@@ -146,14 +138,51 @@
...
 .oldwin #nav-download-firefox,
 .oldmac #nav-download-firefox {
     display: none;
 }
+
+html[dir="rtl"] {
+    #masthead {
+
+        .masthead-logo {
+            @media #{$mq-tablet} {
```
craigcook: Odd nesting, though it still compiles as expected. Should probably put the selector inside the media query just for cleanliness.

media/css/hubs/_common.scss
```diff
@@ -13,23 +13,23 @@ $color-link-blue-dark: #175a77;
...
+        padding: 0 5px 0 40px;
     }
 }
 
 /* -------------------------------------------------------------------------- */
 // CTA link
 
 a.cta-link {
-    @include font-size(18px);
+    @include font-size(14px);
```
craigcook: I propose adding a `font-size-level6` that is 14px and bumps up to 16px on tablet and larger (see gist https://gist.github.com/craigcook/23addac6555f239900e1e9c600553fa4). If we do that we can use it here.

media/css/pebbles/includes/mixins/_typography.scss
```diff
@@ -48,41 +48,73 @@
...
+    }
+
+    @media #{$mq-desktop-wide} {
+        @include font-size(60px);
     }
 }
 
 @mixin font-size-level2 {
-    @include font-size(28px);
+    @include font-size(32px);
```
craigcook: I'm also open to better names for these mixins if we want to start using them more strictly. The "level" names were inspired by heading levels (h1-h6) so maybe that's good enough. Just sometimes a little weird when we have an h2 explicitly set to level4 or an h4 set to level2.

media/css/hubs/_common.scss
alexgibson: The original ask for this copy was to be visually smaller than the body copy, but 16px in bold looks the same size as 18px regular. One downside of only ever using these mixins is it's now impossible to define a set font size that doesn't change depending on the viewport size, unless it's something that inherits the base size directly.

Still, I'm going to change this now as you suggest else I feel we'll just be spinning wheels on what's really a minor detail.

craigcook: Hm, yeah. 16px Open Sans is about the same size as 18px Zilla Slab just because Zilla is so compact and unsuited for body text. So maybe this is a place to use `font-size(14px)` after all. But I'm inclined to just let this go as-is for now, we've fussed over this PR too much already.

PR #4946, opened 2017-07-06
[fix bug 1378992] Add privacy notice for Cliqz

craigcook:
## Description
Updates the legal-docs submodule.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1378992

PR #4947, opened 2017-07-07
[fix bug 1378992] Add Firefox + Cliqz privacy page

alexgibson:
## Description
- Creates the template and URL for the privacy notice added in #4946.
- Also added a link to the new notice in `/privacy/`

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1378992

## Testing
- Make sure you update git sub-modules to include the latest commits from legal-docs.
- Visit `/firefox-cliqz/` and make sure the page displays correctly.

### DIFF
--- a/bedrock/privacy/templates/privacy/index.html
+++ b/bedrock/privacy/templates/privacy/index.html
@@ -81,6 +81,7 @@
         <li class="policy-websites"><a href="{{ url('privacy.notices.websites') }}">{{ _('Mozilla Websites, Communications &amp; Cookies') }}</a></li>
         <li class="policy-firefox"><a href="{{ url('privacy.notices.firefox') }}">{{ _('Firefox Browser') }}</a></li>
         <li class="policy-firefox-os"><a href="{{ url('privacy.notices.firefox-os') }}">{{ _('Firefox OS') }}</a></li>
+        <li class="policy-firefox-cliqz"><a href="{{ url('privacy.notices.firefox-cliqz') }}">{{ _('Firefox + Cliqz') }}</a></li>
         <li class="policy-firefox-cloud"><a href="{{ url('privacy.notices.firefox-cloud') }}">{{ _('Firefox Cloud Services') }}</a></li>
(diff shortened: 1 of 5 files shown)


bedrock/privacy/templates/privacy/index.html
```diff
@@ -81,6 +81,7 @@ <h1 itemprop="name">{{ doc.h1.string }}</h1>
         <li class="policy-websites"><a href="{{ url('privacy.notices.websites') }}">{{ _('Mozilla Websites, Communications &amp; Cookies') }}</a></li>
         <li class="policy-firefox"><a href="{{ url('privacy.notices.firefox') }}">{{ _('Firefox Browser') }}</a></li>
         <li class="policy-firefox-os"><a href="{{ url('privacy.notices.firefox-os') }}">{{ _('Firefox OS') }}</a></li>
+        <li class="policy-firefox-cliqz"><a href="{{ url('privacy.notices.firefox-cliqz') }}">{{ _('Firefox + Cliqz') }}</a></li>
```
alexgibson: @flodolo this adds a new string. The new notice page already has a german translation ready, so maybe we are ok to add this without a tag?

alexgibson: Scratch that, it doesn't have a german translation. Should I use a tag here?

alexgibson: Turns out there is a german translation, but it's labelled incorrectly: https://github.com/mozilla/bedrock/pull/4946/files

I've asked in the bug if this is intentional or an error.

alexgibson: cc @peiying2

flodolo: I don't think we want a tag here, we need to show it to all locales nonetheless. I'm not even sure we want to localize it, given that it explicitly says "Firefox + Cliqz" in the privacy doc.

alexgibson:
Updated the legal-docs sub module so it correctly shows the German translation.

PR #4957, opened 2017-07-10
[fix bug 1371870] Redesign Firefox Dev Edition page

craigcook:
## Description
Implements a new Developer Edition page, with all new graphics and copy. There are a few minor things still in progress and pending final approval from stakeholders, so marking Do Not Merge for the moment.

This is English-only for now. There are more updates for this page planned for some time in the near future so we should hold off extracting strings until we have better direction on what is changing and when. 

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1371870

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/en-US/firefox/developer/


craigcook:
Approved, this is ready for review/merge.

bedrock/firefox/templates/firefox/products/developer.html
```diff
@@ -0,0 +1,269 @@
...
+{% endblock %}
+
+{% block page_css %}
+  {% stylesheet 'firefox_developer_2017' %}
+{% endblock %}
+
+{% block body_id %}firefox-developer{% endblock %}
+{% block body_class %}firefox-developer{% endblock %}
+
+{% block body_attrs %}{{ super() }} data-global-nav-current-link="firefox"{% endblock %}
```
alexgibson: This doesn't seem to be needed since the page doesn't have global nav?

media/css/firefox/developer-2017.scss
```diff
@@ -0,0 +1,478 @@
...
+                        border-right: 0;
+                    }
+                }
+            }
+        }
+    }
+
+    // Conflicting download button in global nav is confusing.
+    // Hide it for now pending a better solution.
+    #global-nav-download-firefox,
```
alexgibson: Think you can remove these lines?

media/css/pebbles/base/elements/_document.scss
```diff
@@ -13,7 +13,6 @@ html {
 body {
     font-size: 100%;
     background: #fff;
-    border-top: 2px solid #fff;
```
alexgibson: 👍

alexgibson:
You'll need to update the [selectors](https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/developer.py#L16-L17) for the functional test here:

https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/test_developer.py#L13

You can delete this test:

https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/test_developer.py#L22
https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/developer.py#L30-L42

media/css/firefox/developer-2017.scss
```diff
@@ -0,0 +1,478 @@
...
+            &:active {
+                background-color: transparentize($color-devteal, .9);
+                border-color: lighten($color-devteal, 10%);
+                color: lighten($color-devteal, 10%);
+            }
+        }
+    }
+
+    // Download button
+    .download-button {
```
alexgibson: If this is another new download button style, would it be worth consolidating into buttons.scss?

bedrock/firefox/templates/firefox/products/developer.html
craigcook: Doh. Removing the global nav was a last-second change, I need to clean up after myself.

media/css/firefox/developer-2017.scss
craigcook: Alas, this is very much a one-off page at this point and a v2 update is coming in the nearish future so I really don't know how long this button style might last. It's not being used anywhere else as yet. I don't know if it's worth adding it as a variation in buttons.scss if there's a good chance we'd be removing it soon. 

I would *like* to rework this page as a standard hub-style product page and standardize what a dev edition download button looks like globally, but FxMarketing and DevMarketing are different units who don't seem to talk to each other so for now this page is a brand new snowflake.

craigcook:
@peiying2 Since this merged I've received word that any forthcoming copy changes probably won't happen for some time so we can begin localization whenever you like. If there's anything that needs to be fixed in the template let me know and I can do it in a followup PR.

PR #4959, opened 2017-07-11
[bug 1370577] Create Thank You Snippet landing page. 

ScottDowne:


ScottDowne:
Not ready to merge, but I'm new to Bedrock so I wanted to get something up early in case I was doing something pretty wrong.

jpetto:
I wont get into actual review territory yet, aside from mentioning that if there's only a single page being created, there's no need to create a `base.html` page. We generally only do that when there are multiple pages that share a structure & some content. Single, one-off pages can be self-contained in a single HTML file. 😄

ScottDowne:
@jpetto Thanks for the look. I think this is ready to start technical review.

However, it's still getting some design feedback, and not sure how it's usually done here, but there are going to be some more css changes, but mostly size and position related changes. Other than that it is ready.

Thoughts on this?

ScottDowne:
Also, how hard would it be to get a sort of review staging link from this  for feedback? Is that a thing in Bedrock I can leverage? If so how hard?

ScottDowne:
Don't think I can remove the WIP progress label myself, but yeah, this is good to go.

jpetto:
@ScottDowne I attempted to push this to a demo, but [it failed](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/demo%2fjpetto-tysnippet/1/pipeline/). I believe the issue is that you're using ES6 classes, which our uglifier cannot handle. We can't ship ES6 to mozilla.org yet. 😞

ScottDowne:
Oh, no prob, I'll fix that up!

ScottDowne:
Fixed. I just ran it through Babel, I think it was just the usage of `class` but it would make sure I got some other stuff I missed too. Code is still pretty readable, but happy to make some formatting tweaks.

jpetto:
The demo is up:

https://bedrock-demo-jpetto-tysnippet.us-west.moz.works/en-US/thank-you/
https://bedrock-demo-jpetto-tysnippet.us-west.moz.works/en-US/thank-you-b/

The build is failing at the moment due to CSS and JS linting errors. You can see the details [here](https://circleci.com/gh/mozilla/bedrock/6029?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). (You may need to collapse and expand the error sections to see the actual messages. I think this is a bug with Circle's UI.)

How soon is this slated to go live?

ScottDowne:
@jpetto I think I got the bulk of the linting (I don't think I can re trigger a build, so I might have missed a few things)

Also did some updates based on design feedback.

I don't know the intended release time for this, my guess is mid next week at the absolute earliest, but don't quote me on this ;)

alexgibson:
> I think I got the bulk of the linting (I don't think I can re trigger a build, so I might have missed a few things)

Just fyi, you can [run the JS/CSS linting checks locally](https://github.com/mozilla/bedrock/blob/master/circle.yml#L25-L27) rather than waiting on them to be run in Circle.

jpetto:
@ScottDowne We've just merged #4996 to get a new URL structure set up for these kinds of one-off pages.

This will require you to do a bit of refactoring in this PR, which I am very happy to help with if needed. Ideally, we'll set this up looking forward, understanding more one-off pages will be added in the future.

Here's what I see as the task list:

- Move the new URLs to `bedrock/etc/urls.py`, and make more descriptive, e.g. `etc/firefox/snippet-thank-you-a`. (You can check the `urls.py` files in other apps for guidance, or just ask here.)
- Move the templates to `bedrock/etc/templates/etc/firefox`.
- Move the `.scss` file to `media/css/etc/firefox`.
- Move the `.js` file to `media/js/etc/firefox`.
- Move the CSS & JS bundles to the bottom of their respective sections, and give them more descriptive names, e.g. `firefox-snippet-thank-you-a` (in an attempt to keep all `etc/` bundles grouped together).

Ping me if you need anything!

ScottDowne:
@jpetto Cool, I think I got these done. Everything I did made sense and pages seem to be working.

I did however changed the file structure name a bit from your proposed to `/etc/firefox/retention/thank-you/` where it made sense. I felt it was a bit better than `/snippet`, as I could see other snippet related things potentially having a thank you too. Totally open to naming this whatever though.

It ended up with the final url of `/etc/thank-you-b/` and I am wondering if that should be `/etc/firefox/retention/thank-you-b/` or something.

bedrock/settings/static_media.py
```diff
@@ -1072,6 +1072,18 @@
...
+            'css/etc/firefox/retention/thank-you-a.scss',
+        ),
+        'output_filename': 'css/firefox-retention-thank-you-a-bundle.css',
+    },
+    'firefox-retention-thank-you-b': {
+        'source_filenames': (
+            'css/etc/firefox/retention/thank-you-b.scss',
+        ),
+        'output_filename': 'css/firefox-retention-thank-you-b-bundle.css',
+    }
```
jpetto: Can we prefix these bundle names with `etc-`? I think that'll be a good pattern going forward to avoid confusion with bundles used on `/firefox/` URLs.

bedrock/settings/static_media.py
```diff
@@ -1839,4 +1851,10 @@
         ),
         'output_filename': 'js/matchmedia_addlistener-bundle.js',
     },
+    'firefox-retention-thank-you': {
+        'source_filenames': (
+            'js/etc/firefox/retention/confetti.js',
+        ),
+        'output_filename': 'js/firefox-retention-thank-you-bundle.js',
+    }
```
jpetto: Same here - let's prefix with `etc-`.

bedrock/etc/urls.py
```diff
@@ -2,5 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from bedrock.mozorg.util import page
+
 urlpatterns = (
+    page('thank-you', 'firefox/retention/thank-you-a.html'),
+    page('thank-you-b', 'firefox/retention/thank-you-b.html')
```
jpetto: Yep, I agree the URLs should be `/etc/firefox/retention/thank-you-X`.

bedrock/etc/templates/firefox/retention/thank-you-a.html
```diff
@@ -0,0 +1,86 @@
...
+
+{% block site_header %}{% endblock %}
+
+{% block content %}
+<main>
+  <canvas id="confetti-canvas"></canvas>
+  <section class="thank-you-section">
+    <div class="thank-you-header">
+      <div>It's all</div>
+      <div>thanks to you</div>
```
jpetto: Are these pages intended to be localized or English only? If the former, we'll need to wrap the strings, and probably do a little markup refactoring to handle line formatting in other locales. If the latter, what should happen if someone hits the URLs with a different locale, e.g. `fr`? We generally avoid showing English strings to other locales, but if the related campaign is only targeted at English speakers (meaning these URLs would generally not be discoverable by other locales), we can probably not worry about it.

bedrock/etc/templates/firefox/retention/thank-you-a.html
```diff
@@ -0,0 +1,86 @@
...
+{% extends "base-pebbles.html" %}
+
+{% block page_title %}
+  {{ _('Thank You') }}
+{% endblock %}
+
+{% block page_title_suffix %}{% endblock %}
+
+{% block page_desc %}
+  {{ _('Thank you page.') }}
```
jpetto: If we don't need these pages localized, I'd suggest unwrapping all currently wrapped strings.

bedrock/etc/templates/firefox/retention/thank-you-a.html
```diff
@@ -0,0 +1,86 @@
...
+        </div>
+      </div>
+      <div class="thank-you-action">
+        <div class="desc">Make the Internet a safer place</div>
+        <div class="time">4 min read</div>
+        <div>
+          <a href="https://www.mozilla.org/internet-health/privacy-security/">Learn more</a>
+        </div>
+      </div>
+    </div>
```
jpetto: Not to be overly nit-picky, but is there a more semantic structure here than a bunch of `<div>`s?

media/css/etc/firefox/retention/thank-you-a.scss
```diff
@@ -0,0 +1,216 @@
...
+    @media #{$mq-phone-wide} {
+        padding: 0 30px;
+    }
+
+    @media #{$mq-tablet} {
+        padding: 0;
+    }
+    .thank-you-actions-container {
+        display: flex;
+        flex-wrap: wrap;
```
jpetto: Do we need to worry about browser support? IE 10 and 11 may not look right.

media/css/etc/firefox/retention/thank-you-a.scss
```diff
@@ -0,0 +1,216 @@
...
+        margin: 0 -9px;
+        .thank-you-action {
+            background: #FFFFFF;
+            padding: 24px;
+            margin: 9px;
+            width: 100%;
+            @media #{$mq-tablet} {
+              width: calc(50% - 66px);
```
jpetto: I know it's nit-picky, but can we standardize indentation to 4 spaces in this file?

media/css/etc/firefox/retention/thank-you-a.scss
```diff
@@ -0,0 +1,216 @@
...
+            }
+            .desc, .time {
+                font-family: antoniobold;
```
jpetto: Should supply a fallback font in the event the fonts wont/can't load?

media/css/etc/firefox/retention/thank-you-b.scss
```diff
@@ -0,0 +1,226 @@
...
+         url('/media/fonts/opensans-regular.woff') format('woff');
+}
+
+@font-face {
+    font-family: 'Open Sans';
+    font-weight: bold;
+    font-style: normal;
+    src: url('/media/fonts/opensans-semibold.woff2') format('woff2'),
+         url('/media/fonts/opensans-semibold.woff') format('woff');
+}
```
jpetto: Suggest making a common `.scss` file for these two to share instead of repeating rules.

media/css/etc/firefox/retention/thank-you-b.scss
```diff
@@ -0,0 +1,226 @@
...
+          text-shadow: -5px 0 0 rgb(215, 102, 40), -8px 0 0 rgba(32, 35, 36, 0.3);
+        }
+        div:first-child {
+          color: #FFFFFF;
+          text-shadow: -5px 0 0 rgb(148, 209, 175), -8px 0 0 rgba(32, 35, 36, 0.3);
+        }
+    }
+
+    p {
+        font-size: 1.3rem;
```
jpetto: Can drop this `font-size`, as it's repeated below.

media/css/etc/firefox/retention/thank-you-b.scss
```diff
@@ -0,0 +1,226 @@
...
+        padding: 0;
+    }
+    .thank-you-actions-container {
+        padding-left: 50px;
+
+        @media #{$mq-desktop} {
+          width: 793px;
+          margin-right: auto;
+        }
+        margin-top: 94px;
```
jpetto: Move this up by `padding-left` for aesthetics?

bedrock/etc/templates/firefox/retention/thank-you-a.html
ScottDowne: English only.

media/css/etc/firefox/retention/thank-you-a.scss
ScottDowne: Forgot to double check this, but after a quick look, we're probably fine as is. IE 11 seems to be fine, and while IE10 doesn't quite display it the same, I don't think it's broken, just lacking an enhancement.

media/css/etc/firefox/retention/thank-you-a.scss
ScottDowne: Totally reasonable :)

bedrock/etc/templates/firefox/retention/thank-you-a.html
ScottDowne: Can do.

media/css/etc/firefox/retention/thank-you-a.scss
ScottDowne: sans-serif might be a little better, so I'll use that as the fallback.

ScottDowne:
@jpetto Updated!

bedrock/etc/templates/firefox/retention/thank-you-b.html
```diff
@@ -0,0 +1,87 @@
...
+
+{% block js %}
+  {% javascript 'firefox-retention-thank-you' %}
+{% endblock %}
+
+{% block body_id %}thank-you{% endblock %}
+
+{% block page_content %}
+
+{% endblock %}
```
jpetto: Same - should be able to remove this empty block.

media/css/etc/firefox/retention/thank-you.scss
```diff
@@ -0,0 +1,161 @@
...
+         url('/media/fonts/opensans-regular.woff') format('woff');
+}
+
+@font-face {
+    font-family: 'Open Sans';
+    font-weight: bold;
+    font-style: normal;
+    src: url('/media/fonts/opensans-semibold.woff2') format('woff2'),
+         url('/media/fonts/opensans-semibold.woff') format('woff');
+}
```
ScottDowne: I couldn't get semibold to work with any default, and after trying again now, still not sure how I can get semibold to work, and it was specifically requested during design feedback that I use it.

I don't see semi bold defined anywhere, and looking at the code, it looks like it's added on a case by case basis.

I am going to remove the regular non bold one though, as that's already there. I'm also going to be more specific with semibold and set it as 600, instead of overwriting the existing bold.

ScottDowne: If you have a tip on how I can get a default semibold, though, I'll update that.

ScottDowne:
@jpetto Updated, good feedback :)

Still not 100% on the semibold thing, but I added a question, and did do a small fixup for what I could around that font. Let me know if you see where I could do more.

bedrock/etc/urls.py
```diff
@@ -2,5 +2,9 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from bedrock.mozorg.util import page
+
 urlpatterns = (
+    page('firefox/retention/thank-you-a', 'firefox/retention/thank-you-a.html'),
+    page('firefox/retention/thank-you-b', 'firefox/retention/thank-you-b.html')
```
jpetto: I know it's a little weird, but our convention on bedrock for template folders is `{app name}/templates/{app name}/whatever`. Can you move the HTML templates from `/templates/firefox/retention` into `/templates/etc/firefox/retention`?

bedrock/settings/static_media.py
```diff
@@ -1839,4 +1853,10 @@
         ),
         'output_filename': 'js/matchmedia_addlistener-bundle.js',
     },
+    'firefox-retention-thank-you': {
```
jpetto: Nit/tidiness pick - prefix bundle name with `etc-`.

ScottDowne:
@jpetto Thanks again :)

Fixed up.

ScottDowne:
Given that these links are fairly hidden, is there any downside to making them live whenever this passes, or should we coordinate the timing for landing this?

jpetto:
I don't see any reason to wait on merging (obviously 😉). These should go live tomorrow.

ScottDowne:
Awesome, thanks! 🚀 🎉

jpetto:
These are live!

https://www.mozilla.org/en-US/etc/firefox/retention/thank-you-a/
https://www.mozilla.org/en-US/etc/firefox/retention/thank-you-b/

flodolo:
Why is this marked as l10n?

jpetto:
> Why is this marked as l10n?

Not sure, as the page is not meant to be localized. Removing the label.

PR #4960, opened 2017-07-11
[fix bug 1378137] Create bedrock shared lib for lazy-loading images

alexgibson:
## Description
- Standardizes image lazy loading across pages using a shared JS lib.
- Adds unit tests.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1378137

## Testing
- Ensure no regressions on both `/firefox/features/` and `/developer/`


tests/unit/spec/base/mozilla-lazy-load.js
```diff
@@ -0,0 +1,150 @@
...
+        });
+
+        afterEach(function(){
+            $target.remove();
+        });
+
+        it('should register an IntersectionObserver correctly', function() {
+            Mozilla.LazyLoad.loadAll('.observer-list-test img');
+            expect($target.find('.image1').attr('src')).toEqual('/img/foo.png');
+            expect($target.find('.image2').attr('src')).toEqual('/img/foo.png');
```
jpetto: Do you think it's worth testing to make sure `data-src` is removed after updating `src`?

media/js/base/mozilla-lazy-load.js
```diff
@@ -0,0 +1,94 @@
...
+                this.removeAttribute('data-src');
+            };
+        });
+    };
+
+    /**
+     * Initiates LazyLoad via feature detecting support for IntersectioObserver.
+     * @param _selector (String) - CSS selector for target images e.g. '.list-item img'.
+     */
+    LazyLoad.init = function(selector) {
```
jpetto: Potential future enhancement - allow `init` to accept either a string or an array? I could see two different "sets" of images wanting to be lazily loaded.

media/js/base/mozilla-lazy-load.js
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        return observer;
+    };
+
+    /**
+     * Fallback for older browser by simply loading all images upon page load.
+     * @param _selector (String) - CSS selector for target images e.g. '.list-item img'.
+     */
+    LazyLoad.loadAll = function(_selector) {
```
jpetto: What do you think about a more descriptive method name here that indicates it's a fallback? Definitely not a blocker.

bedrock/base/templates/macros.html
```diff
@@ -2,6 +2,15 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% macro lazy_image(image_url, placeholder_url) -%}
+<div class="image-container">
```
jpetto: I might be overthinking it, but what if the class here was something more unique, e.g. `lazy-image-container`? Then we wouldn't have to pass a selector in at all - you'd just know that any `` inside an element with class `lazy-image-container` should get observed. (This is already the case on the developer page...)

media/js/base/mozilla-lazy-load.js
jpetto: (I now realize this is incompatible with my next comment 😝.)

tests/unit/spec/base/mozilla-lazy-load.js
alexgibson: This was surprisingly tricky to test, but I think I've found something not too overly complicated.

media/js/base/mozilla-lazy-load.js
alexgibson: Hmm, ok will rename to `loadAllFallback`

bedrock/base/templates/macros.html
alexgibson: Nice suggestion, I will make it so 👍

alexgibson:
@jpetto thanks for the review, updated!

media/js/base/mozilla-lazy-load.js
alexgibson: I think this could be a good addition when we have a future use case for it. Right now, it should still be ok to call this helper lib twice on once page. It shouldn't have any significant performance impact on the page by doing this, as the browser handles all the scheduling/callbacks 😄

alexgibson: Ah scratch that actually, this is not quite set up to do multiple instances yet. But like I say, maybe a future addition.

alexgibson:
@jpetto squashed, thanks

alexgibson:
@jpetto rebased and fixed conflict

PR #4967, opened 2017-07-11
lang file ok tag avoids untranslated skip

pmac:
This allows untranslated strings with the {ok} tag to override the same translated string from another lang file in the set.

### DIFF
--- a/lib/l10n_utils/dotlang.py
+++ b/lib/l10n_utils/dotlang.py
@@ -52,6 +52,7 @@ def parse(path, skip_untranslated=True, extract_comments=False):
         comment = None
 
         for line in lines:
+            l10n_tag = None
             if u'�' in line:
                 mail_error(path, line)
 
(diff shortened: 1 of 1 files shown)


pmac:
I'm pushing this to [my demo](https://bedrock-demo-pmac.us-west.moz.works/de/firefox/features/) now. @peiying2 I think this will fix it. I'll add this PR to the bug once it's filed. Hopefully this fixes the issue.

Related: What exactly do the `{ok}` and `{l10n-extra}` tags in lang files signify? Should the system be doing even more with them?

jgmize:
@pmac would you mind adding a test for this case?

pmac:
> @pmac would you mind adding a test for this case?

planning on it. just getting it up on the demo and starting the conversation.

pmac:
gotta fix the tests as well :)

peiying2:
@pmac https://bugzilla.mozilla.org/show_bug.cgi?id=1380229.  Don't know what {l10n-extra} means.

pmac:
PR updated with test fixes and additions.

flodolo:
`{l10n-extra}` was supposed to mark an extra string in a .lang file, so that it could be available in a specific locale's file without being removed by updates (lang_update).

Having said that, we should remove it. We've never used it, and support in tools was removed during a complete rewrite at the beginning of 2016.

PR #4971, opened 2017-07-12
[fix bug 1370587] Move sync page to Fx hub

jpetto:
## Description

Moves the Sync page into the Fx hub.

Some relevant specifics:

- dropped the functional tests due to the fact that there are no longer any consistent buttons (aside from the main download button, which is handled in another test)
- dropped variation code as experiments are long finished (afaik)
- updated all links and redirects from `firefox.sync` to `firefox.features.sync`
- removed a bunch of images that i don't think are used any longer
- migrated the `en` only file as-is for retained strings/blocks

There will be a second phase to address L10n things, but didn't want to push ship date or add complexity to this PR.

**DO NOT MERGE** while I ask Ty about styling on warning messages.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1370587

## Testing

https://www-demo5.allizom.org/en-US/firefox/features/sync/ (need to have UITour set up to work w/demo5)

Make sure all 6 states display/function as expected (probably easiest to just hack the JS file to force a state):

1. Firefox 31+ (signed-in to Sync) <-- default
2. Firefox 31+ (signed-out of Sync)
3. Firefox 29 or 30
4. Firefox 28 or older
5. Firefox for Android
6. Not Firefox (any other browser)


alexgibson:
> dropped the functional tests due to the fact that there are no longer any consistent buttons (aside from the main download button, which is handled in another test)

Can I suggest adding in this new page to the [existing parameterized tests](https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/features/test_features.py) for /feature pages, just so we have some base coverage? This is worthwhile doing imo, seeing that even a missing image can cause a 500.

bedrock/firefox/templates/firefox/features/sync.html
```diff
@@ -0,0 +1,165 @@
...
+        </div>
+
+        <div class="show-fx-29-30 show-fx-28-older">
+          <a href="https://support.mozilla.org/kb/update-firefox-latest-version" class="button button-hollow" data-interaction="button click" data-element-action="Sync CTA" id="cta-update">
+            {{ _('Update your Firefox') }}
+          </a>
+        </div>
+
+        <div class="show-not-fx">
+          {{ download_firefox(button_color='button-arrow', download_location='secondary cta') }}
```
alexgibson: Is `secondary cta` a value Peter is mapping? It wasn't in the original list of [proposed values](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templatetags/helpers.py#L118-L119) so seems unique to this page atm.

alexgibson:
I believe some of the images being removed here are the source files from the animation that's still at the bottom on `/firefox/desktop/customize/`. The CSS file that uses base46 encodes all the assets, so removing them here won't break anything - but the original files will be gone. I find it unlikely we'll update that animation though, so I'm not opposed to deleting them if you think it's worth doing now.

bedrock/firefox/templates/firefox/features/sync.html
```diff
@@ -0,0 +1,165 @@
...
+
+{% block features_list %}
+<section>
+  <div class="content">
+    <h2 class="section-title"><span>{{ _('Sync') }}</span></h2>
+
+    <div id="sync-intro-copy-cta-wrapper">
+      <div id="sync-intro-copy">
+        {# Warning Text #}
+        <div class="show-fx-29-30 show-fx-28-older warning">
```
alexgibson: I know this is legacy stuff from the old page, but do we still need to cover detecting all these old FF versions separately? Could we just say `if FF < 31` then show an update message?

bedrock/firefox/templates/firefox/features/sync.html
```diff
@@ -0,0 +1,165 @@
...
+          </ul>
+
+          <p class="button-footer">
+            <a href="https://support.mozilla.org/kb/how-do-i-set-up-firefox-sync" class="ga-link" data-interaction="outbound link click">
+              {{ _('Need help?') }}
+            </a>
+          </p>
+        </div>
+
+        <div class="show-fx-31-signed-out show-fx-android">
```
alexgibson: This button seems to get show for Android, yet UITour is only supported on Firefox for Desktop?

alexgibson: As a related note, there seems to be no iOS specific messaging on this page (presumably because the old page also didn't cover it).

alexgibson:
- There's a [URL](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/family/index.html#L142) that needs updating in `/firefox/products/` still.

- I notice there's also an [old hard-coded URL here](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/desktop/customize.html#L249) that may need updating too..

bedrock/firefox/templates/firefox/features/sync.html
```diff
@@ -0,0 +1,165 @@
...
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import google_play_button with context %}
+
+{% add_lang_files "firefox/shared" "firefox/sync" %}
+
+{% extends "firefox/features/base.html" %}
+
+{% block page_title %}
+  {% if l10n_has_tag('firefox_title_meta_bug1345338') %}
```
jpetto: It's a good question that's probably a "no". There are a few L10n related changes that will be coming, most notably doing away with/integrating the `sync-en.html` template. The L10n stuff will be handled in a follow-up PR in the interest of getting this out.

bedrock/firefox/templates/firefox/features/sync.html
jpetto: Yeah, I can update/simplify the logic a bit. Combining 29-30 and 28-older came about yesterday after copy review and I just avoided the scope creep on the JS side. Better to be clean though.

bedrock/firefox/templates/firefox/features/sync.html
jpetto: I thought UITour was only desktop as well, but Kirby tested yesterday on Android and said it worked. I'll ask Craig to verify, just to be safe.

And yeah, I thought about the lack of iOS handling, too, but again didn't want to expose scope creep when we promised to get this out this week. (Meta problems.) I'll bring it up for the follow-up L10n related PR.

alexgibson: I'm not sure how someone would even configure their Android device to work against a demo server for UITour (even if it did support it). But this is news to me if it is somehow supported now.

bedrock/firefox/templates/firefox/features/sync.html
```diff
@@ -0,0 +1,165 @@
...
+          </p>
+        </div>
+      {% endblock %}
+      </div>{#--/#sync-intro-copy--#}
+
+      <div id="sync-intro-cta">
+        <div class="show-fx-31-signed-in">
+          <ul class="primary-buttons">
+            <li>
+              {{ google_play_button(id='cta-android') }}
```
jpetto: 👍

bedrock/firefox/templates/firefox/features/sync.html
```diff
@@ -0,0 +1,165 @@
...
+            </li>
+            <li>
+              <a href="{{ firefox_ios_url('mozorg-sync_page-appstore-button') }}" data-link-type="download" data-download-os="iOS" id="cta-ios">
```
jpetto: 👍

bedrock/firefox/templates/firefox/features/sync.html
jpetto: This is a very, very good point, and illustrative of why I shouldn't be taking large chunks of real-time feedback in Slack.

bedrock/firefox/templates/firefox/features/sync.html
jpetto: Will change to `other`.

jpetto:
> I believe some of the images being removed here are the source files from the animation that's still at the bottom on /firefox/desktop/customize/. The CSS file that page uses base46 encodes all the assets, so removing them here won't break anything - but the original files will be gone. I find it unlikely we'll update that animation though, so I'm not opposed to deleting them if you think it's worth doing now.

Yeah, I'm happy to be cavalier about removing these images, as the `/customize` page is EOL.

jpetto:
> Also, not a blocker for this PR - but it would be great to rewrite some of that JS logic so it can have tests? Another one for the backlog maybe.

I'll add this to the L10n follow-up PR list. I think it is valuable, but I have a hunch some logic may be changing soon.

alexgibson:
Looks like you can remove this file:

https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/sync.py

As there's no newsletter on this new sync page, we also need to remove it from this test:

https://github.com/mozilla/bedrock/blob/master/tests/functional/newsletter/test_newsletter_embed.py#L23

https://github.com/mozilla/bedrock/blob/master/tests/functional/newsletter/test_newsletter_embed.py#L46

.gitignore
```diff
@@ -30,6 +30,7 @@ venv
 *.mmdb
 james.ini
 test-results.xml
+tests/functional/assets
```
alexgibson: There's already `tests/functional/assets/style.css` down below - maybe just update that to omit the `style.css`?

flodolo:
I need someone to help me out here: has anything actually changed for l10n in this PR? It doesn't look like it, but better ask…

jpetto:
> I need someone to help me out here: has anything actually changed for l10n in this PR? It doesn't look like it, but better ask…

Nope, not in this PR. There will be a follow-up to this one that will integrate the two templates (`sync.html` and `sync-en.html`), wrapping the new English strings in `sync-en.html`.

flodolo:
Thanks, removing the tag then.

PR #4974, opened 2017-07-13
[fix bug 1380137] Update firefox/new and /firstrun designs

alexgibson:
## Description
- Updates both `/firefox/new/` and `/firstrun/` with new templates that we're previously run as funnelcake experiments (This is going to be part of the new onboarding flow for Firefox 57)
- Also removes some legacy Yahoo! `/firstrun` funnelcake templates that are no longer needed (this helped simplify the tests considerably)
- Upon signin to FxA on /firstrun, page now redirects to `about:home` instead of `about:newtab` ([Bug 1381051](https://bugzilla.mozilla.org/show_bug.cgi?id=1381051))

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1380137
https://bugzilla.mozilla.org/show_bug.cgi?id=1381051

## Testing
Demo:
https://www-demo1.allizom.org/en-US/firefox/new/
https://www-demo1.allizom.org/en-US/firefox/54.0/firstrun/

Neither of these pages are new, so have already been code reviewed to a fairly good level. However, each was originally a funnelcake experiment - so some cross-browser testing is still worthwhile to make sure nothing was originally missed, or regressed since it was coded.

Both the sign-in and sign-up flow should be tested for `/firstrun` on the demo server, using a fresh profile [configured to run against the demo environment](http://bedrock.readthedocs.io/en/latest/firefox-accounts.html#demo-server-testing). ~Also make sure to [configure UITour](http://bedrock.readthedocs.io/en/latest/uitour.html#local-development) to enable the transition to `about:newtab`~


bedrock/firefox/views.py
```diff
@@ -387,22 +387,14 @@ def get_context_data(self, **kwargs):
...
-            # New onboarding animation funnecake
-            if locale == 'en-US' and funnelcake in ['99', '100', '111', '112', '113']:
-                template = 'firefox/firstrun/new-firstrun.html'
-            # Yahoo search retention funnelcake test
-            elif locale == 'en-US' and funnelcake == '92':
-                template = 'firefox/firstrun/yahoo-retention.html'
-            elif locale == 'en-US' and funnelcake == '90':
-                # ravioli/katie couric promo
-                template = 'firefox/firstrun/ravioli.html'
+            if lang_file_is_active('firefox/firstrun/onboarding', locale):
```
alexgibson: @flodolo new template here that will require translating. Happy to rename the .lang file to anything you prefer.

bedrock/firefox/views.py
```diff
@@ -528,15 +515,15 @@ def new(request):
             elif experience in ['batmfree', 'batmprivate', 'batmnimble', 'batmresist']:
                 template = 'firefox/new/batm/scene2.html'
             else:
-                template = 'firefox/new/scene2.html'
+                template = 'firefox/new/onboarding/scene2.html'
+        elif lang_file_is_active('firefox/new/onboarding', locale):
```
alexgibson: @flodolo a second one here

bedrock/firefox/views.py
```diff
@@ -568,7 +555,9 @@ def new(request):
             elif experience == 'batmresist':
                 template = 'firefox/new/batm/resist.html'
             else:
-                template = 'firefox/new/scene1.html'
+                template = 'firefox/new/onboarding/scene1.html'
+        elif lang_file_is_active('firefox/new/onboarding', locale):
```
alexgibson: @flodolo a third one here (but can probably share the same lang file as `'firefox/new/onboarding/scene2.html'`

alexgibson:
> Just to confirm: the old firefox/(new|firstrun)/horizon is replaced by firefox/(new|firstrun)/onboarding?

Correct

> Given how little content there is in each file, I would use firefox/new/onboarding for all 3 of the templates.

I'll make it so 👍

alexgibson:
@flodolo added a commit so all templates use the same lang file

alexgibson:
OK, I think this is now complete enough it should be ready for review.

jpetto:
Oh, also tested `/firstrun`, both sign-in and sign-up and worked as expected.

alexgibson:
> Wondering if we should let whomever is in charge of the search retention funnelcake know that the page is going to be changing?

Good suggestion - I left a [comment in the relevant bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1363499#c52)

alexgibson:
@jpetto thanks for the review. I think we're good to go here, Verdi has [confirmed in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1380137#c4) that we're good to ship

PR #4977, opened 2017-07-14
[fix bug 1380825] Firstrun page should add  param upon opening FxA's settings page

alexgibson:
## Description
- Adds `?service=sync` as a temporary workaround to avoid an unwanted redirect back to /signin.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1380825

## Testing
Demo: https://www-demo3.allizom.org/en-US/firefox/54.0/firstrun/

1. Launch a new profile using [settings required](http://bedrock.readthedocs.io/en/latest/firefox-accounts.html#demo-server-testing) for testing against a demo instance of bedrock / fxa server.
2. Sign in with an existing Fx Account.
3. Verify that upon successful signin you are redirected to https://accounts.firefox.com/settings?service=sync

### DIFF
--- a/media/js/base/mozilla-fxa-iframe.js
+++ b/media/js/base/mozilla-fxa-iframe.js
@@ -217,7 +217,9 @@ Mozilla.FxaIframe = (function() {
             // hide the FxA iframe.
             _userCallback('onLogin', data);
         } else {
-            Mozilla.Utils.doRedirect(_host + '/settings');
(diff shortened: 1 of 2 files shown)


jpetto:
Works as expected! Signed in on demo3 and was redirected to https://accounts.stage.mozaws.net/settings?service=sync.

Glad we could find an easy temporary workaround.

PR #4978, opened 2017-07-14
[bug 1378137] Fix lazy-image placeholder color on /firefox/features/

alexgibson:
Minor visual regression from #4960

### DIFF
--- a/media/css/firefox/features/common.scss
+++ b/media/css/firefox/features/common.scss
@@ -176,7 +176,7 @@ main {
         display: none;
     }
 
-    .image-container {
+    .lazy-image-container {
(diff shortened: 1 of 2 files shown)


jpetto:
D'oh! Should have caught this during review.

jpetto:
Need to update [on the new developer page](https://github.com/mozilla/bedrock/blob/master/media/css/mozorg/developer/index.scss#L137) as well?

alexgibson:
It's only a subtle thing, I totally missed it too

alexgibson:
> Need to update on the new developer page as well?

Good catch, updated!

PR #4979, opened 2017-07-15
[fix bug 1380044] Firefox download page with full BATM animation

craigcook:
## Description
Updates the basic animation variant with a fully animated video (7 second loop). Also removes the old static variant. Sets up a 50/50 traffic cop test on `xv=batmprivate` using the switch `experiment-firefox-new-batm-anim` and I'll have a www-config PR to flip it.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1380044

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/en-US/firefox/new/?xv=batmprivate&v=b


craigcook:
https://github.com/mozmar/www-config/pull/24

media/js/firefox/new/experiment-firefox-new-batm-anim.js
```diff
@@ -0,0 +1,16 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    var cop = new Mozilla.TrafficCop({
+        id: 'experiment_firefox_new_batm_anim',
+        variations: {
+            'v=b': 50   // animated machine
```
craigcook: Good call.

craigcook:
The video isn't playing in Safari for some mysterious reason. If I swap in a different video served from our video host (https://videos.cdn.mozilla.net/uploads/mozillaorg/Mozilla_2014_i_am.mp4, the video from the about page) it plays just fine. If I put that same video into /media/img it doesn't play. So this leads me to believe this isn't a problem with the machine video file encoding, I think it's something in bedrock. 

I'm going to ask Rainer to put the videos on the CDN and update the urls and remove them from bedrock, but that will have to wait until morning. Marking Do Not Merge for the moment.

jpetto:
I think it's also worth filing a bug to figure out why we can't serve these from bedrock. It could be a P2 or whatever, but I think worth a look.

craigcook:
After a few more folks had a chance to see the demo page and provide feedback we're going to make some design and copy changes. The general feedback was that with so many moving elements there is simply too much going on with the overlay dialog boxes. The idea now is to eliminate those boxes and reduce it to a single blurb at the bottom of the page, allowing the animation to stand much more on its own.

Closing this PR for now and I'll open a fresh one with the revisions when they're done.

PR #4981, opened 2017-07-17
[fix bug 1380490] L10n/JS updates for Sync page.

jpetto:
## Description

- Cleans up old L10n tags that are no longer needed (confirmed w/Peying)
- Integrates `sync-en.html` template into `sync.html` & wraps previously en only strings
- Big JS refactor for easier unit testing
- Accounts for Fx iOS users (same messaging as Fx Android)

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1380490

## Testing

Ensure no regressions. Sanity check JS refactoring.


media/js/firefox/features-sync.js
```diff
@@ -1,104 +1,126 @@
...
-
-                // Query if the UITour API is working before we use the API
-                Mozilla.UITour.getConfiguration('sync', function (config) {
-
-                    // Variation #1: Firefox 31+ signed IN to Sync (default)
-                    if (config.setup) {
-                        swapState('state-fx-31-signed-in');
-                        state = 'Firefox 31 or Higher: Signed-In';
+    SyncPage.setBodyClass = function(stateClass) {
+        var classes = SyncPage.body.getAttribute('class');
```
alexgibson: I don't think this works in older browsers like IE, should be `SyncPage.body.className;`?

media/js/firefox/features-sync.js
```diff
@@ -1,104 +1,126 @@
...
+        classes = classes.replace('state-default', stateClass);
+        SyncPage.body.setAttribute('class', classes);
```
alexgibson: Again should this be `SyncPage.body.className = classes;`?

media/js/firefox/features-sync.js
```diff
@@ -1,104 +1,126 @@
...
+                'event': 'sync-click',
+                'browser': SyncPage.state
+            });
+
+            Mozilla.UITour.showFirefoxAccounts(SyncPage.params.utmParamsFxA());
+        };
+
+        // Firefox Sync sign in flow button
+        // Only displayed for Fx 31+
+        SyncPage.ctaSync.addEventListener('click', SyncPage.ctaSyncClick);
```
alexgibson: If this button is shown only for Fx 31+ should we only bind the event if that conditional is true?

bedrock/settings/static_media.py
```diff
@@ -1365,6 +1365,7 @@
         'source_filenames': (
             'js/base/uitour-lib.js',
             'js/firefox/features-sync.js',
+            'js/firefox/features-sync-init.js',
```
jpetto: Yeah, good call. Will do.

media/js/firefox/features-sync.js
jpetto: `getAttribute` and `setAttribute` work back to IE 6, so I think we're good. 😉

media/js/firefox/features-sync.js
jpetto: Good call.

media/js/firefox/features-sync.js
alexgibson: Not for `body` (try it!)

alexgibson: To add more clarity, 'class' is not a valid attribute. As per the spec, it's actually a property `className`. Modern browsers support it, but not every browser

jpetto: Oh, wow. Weak!

(Since my IE 6 VM reached EOL, I don't have way to test there...)

PR #4986, opened 2017-07-20
Use Whitenoise as wsgi middleware, not Django

pmac:
When used as a Django middleware all asset responses include all headers intended for page responses (e.g. CSP and XSS protection).  Switch to WSGI middleware to avoid sending extra data as well as to avoid confusing Safari.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -410,8 +410,6 @@ BASIC_AUTH_CREDS = config('BASIC_AUTH_CREDS', default=None)
 
 MIDDLEWARE_CLASSES = [
     'django.middleware.security.SecurityMiddleware',
-    # must be after SecurityMiddleware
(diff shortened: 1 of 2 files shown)


pmac:
@jgmize proposing this change partially due to our belief that Safari failing to display an mp4 file was due to some extra and unexpected headers sent with static files due to how we're using Whitenoise. This change greatly reduces the header load of all static asset responses. Any feedback would be appreciated.

cc @craigcook

pmac:
@glogiotatidis mind taking a look? r?

pmac:
It doesn't seem to be important. The default Django dev server does a fine job with the static media, but that's a good idea should that change.

jgmize:
belated 👍 -- thanks @pmac & @glogiotatidis

PR #4988, opened 2017-07-20
[fix bug 1370574] Add send tabs feature page.

jpetto:
## Description

The page is a close fork of the recently updated Sync page. For now, CSS and JS are shared between the two, as the layout and functionality are nearly identical.

For reference, the states detected are:

- Fx 31+ signed in to Sync
- Fx 31+ signed out of Sync
- Fx < 31
- Fx Android or iOS
- not Fx

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1370574

## Testing

Ensure old redirect from `/firefox/features/send-tabs` to `/firefox/features/sync` is no longer active. Make sure copy generally makes sense for all states.


bedrock/firefox/templates/firefox/features/send-tabs.html
```diff
@@ -0,0 +1,151 @@
...
+        <img src="{{ static('img/firefox/features/thumbnails/send-tabs/still-emailing.jpg') }}" alt="">
+        <h3>{{ _('Free up your inbox') }}</h3>
+        <p>
+          {{ _('Still email links to yourself? Send tabs between devices instead.') }}
+        </p>
+      </li>
+
+      <li class="features-list-item">
+        <img src="{{ static('img/firefox/features/thumbnails/send-tabs/device-hop.jpg') }}" alt="">
+        <h3>{{ _('Device hop like a pro') }}</h3>
```
alexgibson: I know `Device hop` can be a bit confusing for some translations. Would it be worth adding an l10n comment to provide an explanation?

bedrock/firefox/templates/firefox/features/send-tabs.html
```diff
@@ -0,0 +1,151 @@
...
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import google_play_button with context %}
+
+{% add_lang_files "firefox/shared" "firefox/sync" %}
+
+{% extends "firefox/features/base.html" %}
+
+{% block page_title %}
+  {{ _('Take your browser tabs to go | Firefox') }}
```
jpetto: Good call.

bedrock/firefox/templates/firefox/features/send-tabs.html
```diff
@@ -0,0 +1,151 @@
...
+        <div class="show-fx-30-older warning">
+          <p>
+            <span>{{ _('It looks like you’re running an older version of Firefox.') }}</span>
+          </p>
+        </div>
+
+        <div class="show-fx-31-signed-in">
+          <strong>{{ _('Logged in and connected!') }}</strong>
+          <p>
+            {{ _('You’re all set to start sending tabs.') }}
```
jpetto: Sigh, yes. This is what happens when you skip a UI/UX phase and just go with "fork the Sync page and update the copy". Hopefully we get priority to do a phase 2.

flodolo:
Uhm, why is this not tagged as l10n? Also, you're modifying a page that is already localized (features/index).

bedrock/firefox/templates/firefox/features/send-tabs.html
```diff
@@ -0,0 +1,151 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import google_play_button with context %}
+
+{% add_lang_files "firefox/shared" "firefox/sync" %}
```
jpetto: There's one shared string ("It looks like you’re running an older version of Firefox.") - I guess I'm not sure if that warrants importing the Sync lang file or not.

@peiying2 or @flodolo  - Do you have an opinion here? There's one shared string between this template and `/firefox/features/sync.html`.

flodolo: Definitely don't import a lang file for a shared string, we can copy over the translation

jpetto:
> Uhm, why is this not tagged as l10n? Also, you're modifying a page that is already localized (features/index).

Sorry! I'll wrap `features/index` changes with an L10n tag conditional.

bedrock/firefox/templates/firefox/features/index.html
```diff
@@ -79,61 +79,73 @@ <h2 class="section-title"><span>{{ _('Modern') }}</span></h2>
         </p>
       </li>
       <li class="features-list-item">
-        <a href="{{ url('firefox.android.index') }}" data-link-type="link" data-link-name="Firefox for Android" data-link-position="modern 5">
+        <a href="{{ url('firefox.features.send-tabs') }}" data-link-type="link" data-link-name="Tabs That Travel" data-link-position="modern 5">
```
flodolo: This block should show up only if firefox/features/send-tabs is activated. With all the shared strings and included files, that's the only safe way to do it…

flodolo: Actually, since these strings don't appear to be used in other templates, and don't need to put them in shared, we should be able to put them in firefox/features/index.lang and use a tag.

jpetto:
Came up with a mildly hack-y solution to the `data-link-position` issue on `/firefox/features/`. [Asked pg](https://bugzilla.mozilla.org/show_bug.cgi?id=1370574#c29) if he sees an issue with the approach. Adding DNM label until he responds.

bedrock/firefox/templates/firefox/features/index.html
```diff
@@ -78,6 +78,9 @@ <h2 class="section-title"><span>{{ _('Modern') }}</span></h2>
           </a>
         </p>
       </li>
+    {% set modern_iterator = 4 %}
+    {% if l10n_has_tag('firefox-features-send-tabs') %}
+      {% set modern_iterator = 5 %}
```
alexgibson: While this works and preserves the data attributes as they are in the current form, my gut says this is only prolonging the existence of the same issue for a bit longer. What happens when there's another tag for example, would just become more complicated again? While I'm not opposed to merging it this way, my preference would be to either:

- Remove this data attribute altogether (how useful really is it?)
- Rework the feature list items into a macro, which could generate this number dynamically (this would likely be a fair bit of work, so I wouldn't expect it to happen here).
- Find some alternative solution for GA that doesn't rely so much on a specific number.

jpetto: You are totally right that this is a non-scalable fix. I do think re-examining how we do GA in this section is the next step here. I've asked Peter to give his thoughts in the bug.

jpetto:
@alexgibson Let me know what you think about this JS approach.

media/js/firefox/features/landing.js
```diff
@@ -6,4 +6,10 @@
     'use strict';
 
     Mozilla.LazyLoad.init();
+
+    // data-link-position is dynamic due to L10n trickle-in
+    // may eventually be needed for other feature types on this page
```
alexgibson: This is a nice solution 👍 

My only question is whether it may be worth doing this now for all three lists? It may not be obvious why two lists have the attribute but the first does not. This is maybe going a bit beyond the scope of this bug, but it might not be that much more code to write?

jpetto:
I think there are new images coming already, so still do not merge.

jpetto:
New images have been added. Still need to squash prior to merge, but good for final review.

PR #4989, opened 2017-07-20
[fix bug 1382107] Developer Edition string updates

craigcook:
## Description
Changes a CTA to avoid a duplicate string. Also changes some headlines and CTAs to sentence case. A few headings appear to be title case but only because the feature name is treated as a proper noun (e.g. "Responsive Design Mode").

The page hasn't yet been exposed to l10n so I don't think we need tags for these changes.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1382107


flodolo:
Confirmed, no need for tags.

flodolo:
I'm just going ahead and merge this one, since it's only strings. Thanks Craig 👍

PR #4991, opened 2017-07-21
Updates Traffic Cop to latest release (customCallback ability).

jpetto:
## Description

Brings in the latest version of Traffic Cop. We can now do experiments that execute arbitrary JS instead of performing a redirect.

## Issue / Bugzilla link

https://github.com/mozilla/trafficcop/issues/7

## Testing

Make sure existing experiments work with no regressions.

The following URL (updated with latest version of TC) has an experiment running for IE 10 and greater (with DNT off, geolocated in US). You should get a funnelcake variation within a few tries:

https://bedrock-demo-jpetto-tc.us-west.moz.works/en-US/firefox/new/

I tested this on a VM and the experiment ran successfully.

### DIFF
--- a/media/js/base/mozilla-traffic-cop.js
+++ b/media/js/base/mozilla-traffic-cop.js
@@ -26,11 +26,19 @@ if (typeof Mozilla === 'undefined') {
  *
  *
  * @param Object config: Object literal containing the following:
- *      String id (required): Unique-ish string for cookie identification.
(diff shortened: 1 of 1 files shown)


alexgibson:
> The following URL (updated with latest version of TC) has an experiment running for IE 10 and greater (with DNT off, geolocated in US).

I tested the demo URL using a VM on BrowserStack and got ?f=118

PR #4996, opened 2017-07-25
[fix bug 1383115] Add new bedrock/etc app for one-off pages.

jpetto:
## Description

Adds scaffolding for new `bedrock/etc` app.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1383115

## Testing

Make sure all files added are required, and no-indexing is set up correctly.

### DIFF
--- a/bedrock/etc/models.py
+++ b/bedrock/etc/models.py
@@ -1,3 +1,0 @@
-from django.db import models
-
-# Create your models here.
--- a/bedrock/etc/models.py
+++ b/bedrock/etc/models.py
@@ -1,0 +1,3 @@
+from django.db import models
+
+# Create your models here.
--- a/bedrock/etc/redirects.py
+++ b/bedrock/etc/redirects.py
@@ -1,3 +1,0 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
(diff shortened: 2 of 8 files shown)


bedrock/etc/models.py
```diff
@@ -0,0 +1,3 @@
+from django.db import models
```
pmac: Can delete this. flak8 will likely complain.

pmac: The line in this file that is, not the file itself.

bedrock/etc/redirects.py
```diff
@@ -0,0 +1,3 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
```
pmac: Don't really need this file. We can always add it later should we grow some redirects for these. The only required potentially empty file for an app is `models.py`.

bedrock/etc/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
pmac: You can remove this file until we have tests.

bedrock/urls.py
```diff
@@ -31,6 +31,7 @@
     url(r'', include('bedrock.mozorg.urls')),
     url(r'', include('bedrock.newsletter.urls')),
     url(r'', include('bedrock.teach.urls')),
+    url(r'', include('bedrock.etc.urls')),
```
pmac: If these should all be at `/etc/*` then this should be:

```python
url(r'^etc/', include('bedrock.etc.urls')),
```

PR #5001, opened 2017-07-26
[fix bug 1381776] Implement site wide notification for out-of-date Firefox users

alexgibson:
## Description
- Adds a new JS library for displaying site wide notifications to visitors (en-US only for now).
- Shows a notification to Firefox users who are out of date, prompting them to update.
- Test 4 variations on messaging.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1381776

## Testing
Demo: https://bedrock-demo-agibson.us-west.moz.works
Demo instructions: https://bugzilla.mozilla.org/show_bug.cgi?id=1381776#c11

Note: I've tested this as far back as Firefox 3.6 and it seems to work as expected.

Things to verify:
- Notification is shown for out-of-date Firefox only.
- Once a user sees a specific copy variation, navigating pages or refreshing the page will result in seeing the same variation only.
- Once a user interacts with the notification (clicking CTA, or closing with the 'x'), they will not see notifications again (21 day cookie expiry).
- Clicking the update button takes the user to scene2 of /firefox/new/
- Clicking close button removes the notification.
- Test responsiveness for both Pebbles and Sandstone versions.
- Notification should not show on main download pages, or /firstrun & /whatsnew templates.

~TODO - run this branch against integration tests to make sure it's not going ot break anything while visible.~


alexgibson:
OK, still waiting on confirming GA but this is close enough to code-complete to start review.

media/js/base/mozilla-notification-banner.js
```diff
@@ -0,0 +1,267 @@
...
+    /**
+     * Initializes a new notification banner
+     * @param {Object} options - Object literal containing the following:
+     * {String} id (required) - Unique identifier used to track variations a visitor has seen e.g. 'fx-out-of-date-banner-copy1-direct-1'.
+     * {String} name (required) - Generic name for the notification type e.g. 'fx-out-of-date'.
+     * {String} experimentName (optional) - Generic name for tracking a specific experiment in GA e.g. 'fx-out-of-date-banner-copy1'.
+     * {String} experimentVariant (optional) - Identifier for experiment variation in GA e.g. 'direct-1'.
+     * {String} heading (required) - Copy string for notification heading.
+     * {String} message (required) - Copy string for notification message / subheading.
+     * {String} confirm (required) - Copy sring for main CTA button.
```
flodolo: typo: sring (in a few places)

flodolo:
I take this is English only?

alexgibson:
> I take this is English only?

For now yes, hence the hard-coded strings.

media/js/base/mozilla-notification-banner.js
alexgibson: Thanks

media/js/base/mozilla-client.js
```diff
@@ -239,7 +239,7 @@ if (typeof Mozilla === 'undefined') {
         };
 
         var onRetrieved = function (accurate, version, channel, distribution) {
-            document.removeEventListener('mozUITourResponse', listener);
+            document.removeEventListener('mozUITourResponse', listener, false);
```
alexgibson: This was throwing an error in really old Firefox versions as `useCapture` wasn't always optional. I'm setting it explicityle here to the default value, which is `false`.

media/css/firefox/family-nav.less
```diff
@@ -549,6 +549,8 @@
         background-image: none;
         box-shadow: 0 1px 3px rgba(50, 50, 50, 0.4);
         position: fixed;
+        top: 0;
```
alexgibson: This fixes a bug in the family sticky nav when an element comes before the `#outerwrapper`, which was caught in this test failure: https://saucelabs.com/beta/tests/7847e6fc2eee49099e3e7b84a772fd13/watch#24

alexgibson:
This branch now passes agains the full integration test suite: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/43/pipeline

media/js/base/mozilla-notification-banner.js
```diff
@@ -0,0 +1,267 @@
...
+        return 'querySelector' in document &&
+               'querySelectorAll' in document &&
+               'addEventListener' in window &&
+               'createDocumentFragment' in document;
+    };
+
+    /**
+     * Initializes a new notification banner
+     * @param {Object} options - Object literal containing the following:
+     * {String} id (required) - Unique identifier used to track variations a visitor has seen e.g. 'fx-out-of-date-banner-copy1-direct-1'.
```
alexgibson: Note: I'm going to follow up with proper docs for this library, but I'll do that in a follow up PR.

media/js/base/mozilla-global-nav.js
```diff
@@ -61,6 +61,12 @@
          * @param {String} id - optional accordion id to expand on open.
          */
         toggleDrawer: function(accordionId) {
+
+            // make sure we're at the top of the window before opening drawer.
```
alexgibson: This is a workaround to fix a bug in the global nav when there is an alement that comes above it in the page

alexgibson:
GA is all working and confirmed by Peter, so removing the do-not-merge label.

media/js/base/mozilla-notification-banner-init.js
```diff
@@ -0,0 +1,97 @@
...
+    if (client._isFirefoxDesktop) {
+        client.getFirefoxDetails(function(details) {
+            // User must be out of date and on release channel.
+            if (!details.isUpToDate && details.channel === 'release') {
+
+                // Check that cookies are enabled before seeing if one already exists.
+                if (typeof Mozilla.Cookies !== 'undefined' && Mozilla.Cookies.enabled()) {
+                    var cookie = Mozilla.NotificationBanner.getCookie();
+
+                    if (cookie) {
```
jpetto: What do you think about doing this check before the UITour checks? If a cookie was set, it means the above checks already passed, right? Might be a little more responsive for the user to not have to wait for UITour on subsequent page loads.

media/js/base/mozilla-notification-banner.js
```diff
@@ -0,0 +1,269 @@
...
+     */
+    NotificationBanner.create = function(options) {
+        var _notification = document.createDocumentFragment();
+        var _container = document.createElement('div');
+        var _content = document.createElement('div');
+        var _contentContainer = document.createElement('div');
+        var _heading = document.createElement('h2');
+        var _message = document.createElement('p');
+        var _confirm = document.createElement('a');
+        var _close = document.createElement('button');
```
jpetto: I wish HTML Templates had better IE/Edge support...

media/js/base/mozilla-notification-banner-init.js
```diff
@@ -0,0 +1,97 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+$(function() {
```
jpetto: Do we need to wait for `document.ready` here?

media/js/base/mozilla-notification-banner.js
```diff
@@ -0,0 +1,269 @@
...
+    /**
+     * Gets notification cookie if it exists.
+     * @return {String} cookie value if it exists else null.
+     */
+    NotificationBanner.getCookie = function() {
+        return Mozilla.Cookies.getItem(NotificationBanner.COOKIE_CODE_ID);
+    };
+
+    /**
+     * Generates cookie expiry data defaulting to 21 days.
```
jpetto: I always do the other typo - "date" instead of "data". 😝

media/js/base/mozilla-notification-banner.js
```diff
@@ -0,0 +1,269 @@
...
+
+    /**
+     * Handles interaction with notification close button and removes from the DOM.
+     */
+    NotificationBanner.close = function() {
+        var notification = document.querySelector('.notification-banner');
+        var close = notification.querySelector('.notification-banner-close');
+
+        if (notification) {
+            close.removeEventListener('click', NotificationBanner.close, false);
```
jpetto: Should we unbind the `confirm` listener here as well? There's no way to get back to it once you close the notification, right?

media/js/base/mozilla-notification-banner-init.js
alexgibson: Yeah, this is not very obvious but we need the notification GA event to fire after [this event](https://github.com/mozilla/bedrock/blob/master/media/js/base/core-datalayer-init.js#L26-L28). I'll add a comment here to make this more obvious as to why.

media/js/base/mozilla-notification-banner-init.js
alexgibson: Nice suggestion 👍

media/js/base/mozilla-notification-banner.js
alexgibson: Good spot

media/js/base/mozilla-notification-banner.js
alexgibson: sigh yeah... but even if they did this would still be no use for old Firefoxen :/

media/css/pebbles/components/_notification-banner.scss
```diff
@@ -0,0 +1,124 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../pebbles/includes/lib';
+
+/* -------------------------------------------------------------------------- */
+// Notification Banner
+
+.notification-banner {
+    @include background-size(auto, 100%);
```
alexgibson: Ah yes, I think this is leftover. I can nix it 👍

media/js/base/mozilla-notification-banner-init.js
alexgibson: Hmm, on second thought I'm not sure if this is right. A user might have a cookie for 21 days after they have updated without interacting with the banner for example - which would cause us to show the notification even if they are up to date.

alexgibson:
@jpetto thanks for the review, I pushed some fixes. Let me know if you have any further questions / suggestions 👍

media/js/base/mozilla-notification-banner-init.js
jpetto: Ah, yes, good example scenario.

bedrock/firefox/templates/firefox/australis/base.html
```diff
@@ -53,3 +53,6 @@
 {% block site_footer %}{% endblock %}
 {% block email_form %}{% endblock %}
 {% block js %}{% endblock %}
+
+{# Bug 1381776 #}
+{% block update_notification %}{% endblock %}
```
jpetto: I would like to keep this on the radar, as implied opt-in feels risky, and explicit opt-out in templates can be difficult to manage and easy to forget. (I can see a quick-turnaround /new variation requested that forgets to/doesn't know to explicitly opt-out.)

jpetto:
@alexgibson Squash and then we'll merge?

alexgibson:
@jpetto thanks squashed 👍 

This is going to be turned on Tuesday via the switch, but can merge now.

PR #5004, opened 2017-07-26
[fix bug 1380841] Add Firefox update to plugincheck page

craigcook:
## Description
The plugin check functionality is deprecated but there are still old Firefoxes with in-product links to the page, which now presents misleading information. This creates a new plugincheck page encouraging users to update Firefox. This will entirely replace the old page once we have decent locale coverage, at which time we can delete the old plugincheck page and assets.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1380841

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/en-US/plugincheck/

There are three states with different messaging:
* Latest Firefox (or pre-release)
* Outdated Firefox
* Not Firefox

It should detect minor versions as well - e.g. Firefox 54.0 should be treated as out of date because 54.0.1 is the latest - but that could be tricky to test if you don't have a stale browser handy.

### DIFF
--- a/bedrock/settings/static_media.py
+++ b/bedrock/settings/static_media.py
@@ -859,6 +859,12 @@ PIPELINE_CSS = {
         ),
         'output_filename': 'css/plugincheck-bundle.css',
     },
+    'plugincheck-update': {
+        'source_filenames': (
+            'css/plugincheck/plugincheck-update.scss',
+        ),
+        'output_filename': 'css/plugincheck-update.css',
+    },
     'press': {
         'source_filenames': (
             'css/press/press.less',
@@ -1699,6 +1705,12 @@ PIPELINE_JS = {
             'js/plugincheck/check-plugins.js',
         ),
         'output_filename': 'js/plugincheck-bundle.js',
+    },
+    'plugincheck-update': {
+        'source_filenames': (
+            'js/plugincheck/plugincheck-update.js',
+        ),
+        'output_filename': 'js/plugincheck-update-bundle.js',
     },
     'press_speaker_request': {
         'source_filenames': (
--- a/bedrock/mozorg/templates/mozorg/plugincheck-update.html
+++ b/bedrock/mozorg/templates/mozorg/plugincheck-update.html
@@ -1,0 +1,114 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "base-pebbles.html" %}
+
+{% block page_image %}{{ static('img/plugincheck/page-image.png') }}{% endblock %}
+
+{% block page_title %}
+  {{ _('More browser features, fewer plugin updates | Firefox') }}
+{% endblock %}
+{% block page_title_suffix %}{% endblock %}
+
+{% block page_desc %}
+  {{ _('Browse with fewer interruptions. Features from NPAPI plugins are now automatically supported in Firefox. Download the latest Firefox now!') }}
+{% endblock %}
+
+{% block body_class %}{% endblock %}
+{% block body_id %}plugincheck{% endblock %}
+
+{% block page_css %}
+  {% stylesheet 'plugincheck-update' %}
+{% endblock %}
+
+
+{% block site_header_logo %}{% endblock %}
+
+{% block content %}
+<main role="main">
+  <section class="section">
+    <div class="content">
+
+      <div class="message message-firefox-current">
+        {# L10n: <br> for visual formatting #}
+        <h1>{{ _('Good news - you’re using<br> the latest version of Firefox!') }}</h1>
+        <p class="tagline">
+        {% trans npapi='https://blog.mozilla.org/futurereleases/2015/10/08/npapi-plugins-in-firefox/' %}
+          Even better news - Firefox now <a href="{{ npapi }}">supports features</a>
+          that used to require additional plugins.
+        {% endtrans %}
+        </p>
+
+        <p class="support">
+          <a href="https://support.mozilla.org" data-link-type="support" data-link-position="current" data-link-name="Mozilla Support">
+            {{ _('Need help? Visit Mozilla Support') }}
+          </a>
+        </p>
+      </div>
+
+      <div class="message message-firefox-old">
+        {# L10n: <br> for visual formatting #}
+        <h1>{{ _('Better features,<br> fewer plugin updates') }}</h1>
+        <p class="tagline">
+        {% trans npapi='https://blog.mozilla.org/futurereleases/2015/10/08/npapi-plugins-in-firefox/',
+                 blocklist='https://support.mozilla.org/kb/add-ons-cause-issues-are-on-blocklist' %}
+          Firefox has been <a href="{{ npapi }}">expanding to support features</a>
+          that used to require extra plugins. Now, they’re kept up-to-date automatically
+          and managed through <a href="{{ blocklist }}">blocklisting</a> for added security.
+          So you can stay current with just one Firefox update.
+        {% endtrans %}
+        </p>
+
+        <div class="cta">
+          <p>{{ _('Download the latest version:') }}</p>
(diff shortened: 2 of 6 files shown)


bedrock/settings/static_media.py
```diff
@@ -1671,14 +1669,7 @@
     },
     'plugincheck': {
         'source_filenames': (
-            'js/libs/mustache.js',
```
craigcook: Not sure if this lib is used anywhere else... should I delete?

craigcook: Baleted!

bedrock/mozorg/templates/mozorg/plugincheck.html
```diff
@@ -2,18 +2,27 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
-{% extends "mozorg/base-resp.html" %}
+{% extends "base-pebbles.html" %}
 
-{% add_lang_files "mozorg/plugincheck-redesign" %}
+{% add_lang_files "mozorg/plugincheck-redesign" "firefox/shared" %}
```
flodolo: Uhm, why are you adding this lang file?

craigcook: For the newsletter signup box. We could also just add those strings as new ones here, but they're already translated in the shared file.

flodolo: Adding a shared file introduces a lot of complexity (you can't activate the file unless the shared one is complete). If it's only for the newsletter strings, it's better to expose them and import translations if available.

craigcook:
> Is it just me or you're drastically changing the template's content? Wouldn't it be clearer to use a different template and fall back to the old one if not localized?

Yeah, it's pretty much replacing the entire page. We could indeed do a separate template and let locales fall back to the old one, but the plugincheck functionality is being deprecated and keeping the old page alive would be misleading to users. Most of the old text needs to be deleted entirely so it seemed simpler to do it in the single template with a few new strings behind a tag.

bedrock/mozorg/templates/mozorg/plugincheck.html
craigcook: I've removed the extra lang file and wrapped the newsletter strings in the same tag, and it falls back to the standard "Keep up with all things Firefox." headline that appears in main.lang

flodolo:
> We could indeed do a separate template and let locales fall back to the old one, but the plugincheck functionality is being deprecated and keeping the old page alive would be misleading to users.

The new page will remain online for longer though, which means any new locales working on this page will have to deal with a ton of obsolete strings.

I still think that having a different template would be a cleaner approach. We can set a short-ish deadline on which disabling the old one if you think that's better.

craigcook:
Alright, I'll make a new template so we can generate a clean new lang file rather than make a mess trying to delete large chunks of the existing one.

craigcook:
I've switched to a forked template and non-English locales can continue to get the old page until the new one is activated. We should give locales a fairly short deadline to do that, maybe four weeks? Then we can delete the old template and any locales that still haven't updated will default to English.

flodolo:
> I've switched to a forked template and non-English locales can continue to get the old page until the new one is activated. We should give locales a fairly short deadline to do that, maybe four weeks? Then we can delete the old template and any locales that still haven't updated will default to English.

Thanks, this works for me. @peiying2 ?

I'd even suggest to file a bug with a date as soon as this one merges.

alexgibson:
Before I move ahead with reviewing this, are we sure this is what is best for existing visitors? It looks like we've switched back to preserving the old page / functionality. Is this really he best option given the page might be telling people their Flash is up-to-date when it may be insecure? It's not up to me, but I'd rather we even showed English strings as opposed to tell people false information.

craigcook:
I'm a little torn. Keeping the old template alive seems better for localizers, but worse for users, and in most matters I will always prioritize user benefit over developer convenience. Every day that page is allowed to live is another day of providing bad information to the public. It needs to be deleted as soon as possible. 

I can go back to my original approach of drastically changing the single template rather than forking a new one (I stashed it in another branch, it wouldn't take much effort to restore) but that does seem to complicate l10n. 

What if we just purge all locales and do a single template with new strings, no tags or fallbacks? Relaunch the page in English and locales can translate as they would any other new page.

flodolo:
I'm not sure how much traffic that page generates these days, but in the past it used to be one of the most visited. If that's still the case, I think it's good to have a buffer period.

flodolo:
> It's not up to me, but I'd rather we even showed English strings as opposed to tell people false information.

You're forgetting the most important bit: a lot of people won't understand what that page says if it's in English. Is that really better?

alexgibson:
> You're forgetting the most important bit: a lot of people won't understand what that page says if it's in English. Is that really better?

What I'm questioning is whether it's better to tell people something they understand but is false (and also potentially damaging to thier security), or to not tell them at all. It's a tough line for sure, but I'm not sure that preserving the old page is really the best solution.

alexgibson:
OK, I'm just going to review this as is. I'm not sure it's the best solution, but it's better than nothing.

craigcook:
What if instead of showing English we redirect locales to /firefox/new until they're translated?

bedrock/settings/static_media.py
```diff
@@ -1700,6 +1706,12 @@
         ),
         'output_filename': 'js/plugincheck-bundle.js',
     },
+    'plugincheck-update': {
+        'source_filenames': (
+            'js/mozorg/plugincheck-update.js',
```
alexgibson: Nit: suggest moving this file to `js/plugincheck/` for consistency as long as the other files are still around?

flodolo:
Who should own this decision? Eric?

I'm happy to shorten the buffer, but we really need to give one to our localizers.

media/js/mozorg/plugincheck-update.js
```diff
@@ -0,0 +1,23 @@
...
+
+(function($) {
+    'use strict';
+
+    var client = window.Mozilla.Client;
+    var body = $('body');
+
+    if (client._isFirefoxDesktop) {
+        client.getFirefoxDetails(function(details) {
+            if (details.isUpToDate) {
```
alexgibson: Do we need to check for ESR users here too?

alexgibson: And also that they are on release channel?

alexgibson: Actually looking at the logic for `Mozilla.Client`, it detects ESR sucessfully too (using UITour) and compares that against the latest ESR version number in product details, so you can actually omit the ESR check and it should "just work".

We probably don't want to show this to out-of-date Beta/Nightly users though, so maybe just check that the channel is either 'release' or 'esr'?

alexgibson: Hmm, actually - we would want ESR users to update their ESR version, and not switch to release channel, right? So in that case, perhaps we do excluse ESR users altogether?

craigcook: All good questions. Pre-release browsers seem to report as "up to date" even if they're not up to date on their own channel (e.g. I haven't restarted Nightly in four days but the page tells me it's up to date). We initially decided not to do anything special for pre-release channels because this page is mostly aimed at old Firefoxes that still have the in-product link. But then... if you're on a year old copy of Dev Edition that should still report as outdated, but you'll get a download button for release channel... I'm not sure what to do about that scenario.

We do probably need to do something special for ESR. I agree that if someone is on an old ESR they should probably update to the latest ESR, not release. We can use the same messaging for ESR, we just need to give them the right download. Does the button logic already do this?

ejregithub: Messaging to ESR users is a good idea, assuming the button logic is confirmed to function as Craig is thinking it does ^^ ?

alexgibson: After discussion on Slack, consensus seems to be that we should avoid calling out ESR in general, since they are unlikely to be able to update anyway.

craigcook: I shall make it so. We'll only do special messaging for release channel. ESR or pre-release channels get the generic default.

PR #5006, opened 2017-07-27
[fix bug 1383063] /new variation to support SEM program (search ads)

alexgibson:
## Description
- Adds three new variations of Firefox download pages
- Do-not-merge until Fabio has taken a look, but good for review.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1383063

## Testing
https://www-demo3.allizom.org/en-US/firefox/new/?xv=fast
https://www-demo3.allizom.org/en-US/firefox/new/?xv=secure
https://www-demo3.allizom.org/en-US/firefox/new/?xv=nonprofit
https://www-demo3.allizom.org/en-US/firefox/new/?xv=compatible


alexgibson:
I haven't had any further feedback from Fabio, so I think this is good to merge (if they want these things out fast, they lose the luxury of taking their time). We can always update with tweaks later if needed.

alexgibson:
@jpetto thanks for the commits! I squashed your two commits and reworded the message slightly, so you still get the credit. Should be good for merge if all looks ok?

PR #5010, opened 2017-07-27
[fix bug 1364268] Add modification section to participation guidelines

craigcook:
## Description
Adds a section to the CPG about future updates/amendments/modifications. L10n for this page is handled differently than most bedrock pages so we've opted not to wrap the new text in an l10n tag (cc @peiying2)

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1364268


peiying2:
@craigcook could you add L10n label there.

craigcook:
@peiying2 Is this safe to merge and push to production?

peiying2:
@craigcook sorry Craig, overlooked this request. Please do.  The localized version is already on production.

PR #5018, opened 2017-08-01
[fix bug 1380044] Firefox download page with full BATM animation

craigcook:
## Description
Replaces the previous simple animation with a fully animated video (7 second loop). This is a new variant of the "batmprivate" landing page and sets up a 50%/50% Traffic Cop test. Corresponding PR for the switch https://github.com/mozmeao/www-config/pull/24

The desire is to get this localized in German right away so we should expose strings, at least for de-DE. ccing @peiying2 

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1380044

## Testing
["private" variant](https://bedrock-demo-craigcook.us-west.moz.works/en-US/firefox/new/?xv=batmprivate)
["machine" variant](https://bedrock-demo-craigcook.us-west.moz.works/en-US/firefox/new/?xv=batmprivate&v=b)


craigcook:
Marking Do Not Merge for the moment until we get signoff from stakeholders.

flodolo:
I confess it's really hard to make sense of this PR from a localization point of view

Which templates should be localized from this folder?
https://github.com/craigcook/bedrock/tree/2537bb8047c147a96a53dbebbba60046bc33a9f0/bedrock/firefox/templates/firefox/new/batm

machine.html and scene2.html are used in the demo link above, what about all the rest? Note that all templates should reference a common lang file.

How do you determine if the localized file should be displayed? If you try to load the variation, and the lang file is not activated, you'd fall back to en-US, wouldn't you?

craigcook:
It's an A/B test so I think they're only interested in the "private" variant, the "machine" variant, and scene2. Of course, they all rely on other templates as well so in the end I suppose we really do need to extract the whole folder, and they can all share a common lang file. What should I name it? Would `{% add_lang_files "firefox/new/batm/shared" %}` work?

All this BATM stuff has been for a paid ad campaign targeting English markets (I think even US specifically) so l10n hasn't been a concern, but now they want to run a few ads in Germany. I don't know that there's any intention of showing these pages to other locales beyond that, at least not yet, so I'd hate for volunteers in other locales to waste time translating a page they'll never see. Is it possible to expose strings to just one locale? Or since it sounds like we have an agency translator working on this for DE, maybe we should avoid exposing them to the community at all? I'm not sure what to do :/

craigcook:
I've just been informed that ONLY the "batmprivate" variant will be tested in German, and NOT the machine animation. But it does still depend on other templates so we might as well proceed with stringifying everything in a shared file. The traffic cop redirect is only for en-US so other locales should never see the animated machine page. 

An agency in Germany is working on the translation so I don't think we want to expose the strings to community localizers. But we do need to extract them and generate a lang file for de-DE, I just don't know how best to proceed with that.

flodolo:
It's possible to expose this content only to German, but @peiying2 need to coordinate to avoid the German translator jumping on it.

The shared file can be called "firefox/new/batm", but I still need a clear pointer to the URL that will be displayed, to figure out which strings are needed.

craigcook:
This is the page being tested in German: https://bedrock-demo-craigcook.us-west.moz.works/en-US/firefox/new/?xv=batmprivate ([already in production](https://www.mozilla.org/firefox/new/?xv=batmprivate)).

As for which files contain strings, we might as well extract the entire folder. I'll update the lang file reference to `firefox/new/batm`. The bulk of the text is in scene1.html and most of the variant templates only test different headlines. But now I realize those three headings on scene1 are really American and "punny" so I'm going to see if we can come up with more neutral alternatives for translation. 

Sorry this has turned into such a mess. I think there's been poor communication at a few different levels and it's taking us time to sort it out.

craigcook:
OK, so the [German translation has already been completed](https://docs.google.com/document/d/1GF3c5mQzKgC5gTu-hij1zWP9-udbfO5xiQxcGiCBYyI/), we just need to get the original strings into Pontoon so they can get them into a German lang file. I thought they were waiting for us to deliver strings so they could _begin_ translation. Sorry for my confusion, clearly I've been out of the loop.

So I think the plan is:
1) Extract all the strings in all the templates in firefox/new/batm into a shared lang file.
2) Expose them only to DE and inform the DE community that an agency has done the translation so they needn't bother.

bedrock/firefox/templates/firefox/new/batm/scene1.html
```diff
@@ -19,14 +21,17 @@
...
       <h2>{{_('Click it to the man')}}</h2>
       <p>{{ _('Because Firefox is independent, and built by a non-profit, we can do things that others can’t, like build new products and features without a hidden agenda. We fight for your online rights, keep corporate powers in check, and make the internet accessible to everyone, everywhere.') }}</p>
     </li>
     <li>
+      {# L10n: This is a pun on the American slang "get out of Dodge", meaning to leave town before the situation gets worse. Alternative headline: "Beautifully transparent" #}
       <h2>{{_('Get out of dodgy')}}</h2>
       <p>{{_('Firefox doesn’t sell access to your personal information. In fact, we’re obsessed with protecting your privacy. That’s why we’ve made Firefox Private Browsing more powerful than the others. From privacy tools to tracking protection, you control who sees what.')}}</p>
     </li>
     <li>
+      {% L10n: This is American slang to sit down and rest, here being used as a pun to say Firefox is "lighter". Alternative headline: "Beautifully fast" %}
```
flodolo: This should be `{# … #}`.

flodolo:
One more note: I don't seem to be able to load this page locally for any locale besides en-US, even if the repo is set to DEV.

bedrock/firefox/templates/firefox/new/batm/scene1.html
craigcook: oops!

craigcook:
Those strings don't appear on that particular page but they're on other variants in the same folder. I thought the idea was to just stringify the whole folder in a single shared file, but I don't think the agency has any intent to translate them, which I now realize means the shared file can't be complete... so I'll exclude the templates we don't actually need for this German test.

craigcook:
And it seems all these download page variants are explicitly en-US in the view (https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/views.py#L512) so I'll have to update that for de...

peiying2:
@craigcook we need to get this link fixed.  Oliver Agency has completed the translation, but the staging link continues pointing to the old one. Now I got an email from Germany saying the page on product needs to come down ASAP due to quality of work.  I know the agency would like to verify it on staging first, but the replacement content is better already.  Should I start an email thread on this?  The person from the agency only works during regular EU business hours.  I usually see a ping but he is offline during PDT biz hours.

peiying2:
Alternatively, I can have Stephan work on fixing the current live page (onboarding) while we sort out this one.

craigcook:
Sorry, I had neglected to update the demo server yesterday. I'm updating now.

craigcook:
Demo server updated and German is working: https://bedrock-demo-craigcook.us-west.moz.works/de/firefox/new/?xv=batmprivate

peiying2:
@craigcook thanks for the link. From l10n side, it is production ready.  Stephan is to work on the current live page on Thursday.

bedrock/firefox/views.py
```diff
@@ -542,8 +542,11 @@ def new(request):
                 template = 'firefox/new/batm/scene2.html'
             else:
                 template = 'firefox/new/onboarding/scene2.html'
-        elif lang_file_is_active('firefox/new/onboarding', locale):
-            template = 'firefox/new/onboarding/scene2.html'
+        elif lang_file_is_active('firefox/new/onboarding' 'firefox/new/batm', locale):
+            if experience == 'batmprivate':
+                template = 'firefox/new/batm/scene2.html'
+            else:
+                template = 'firefox/new/onboarding/scene2.html'
```
jpetto: I'm not sure this logic is sound. If `experience` is *not* `batmprivate`, then we don't care if the `firefox/new/batm` lang file is active, correct?

Should this instead be two `elif`s?

```
# aside from en-US, this will only be seen for 'de' locale
elif lang_file_is_active('firefox/new/batm', locale) and experience == 'batmprivate':
    template = `firefox/new/batm/scene2.html`
elif lang_file_is_active('firefox/new/onboarding', locale):
    template = `firefox/new/onboarding/scene2.html`
````

I had a little trouble following the localization conversation, but it sounded to me like this page will only be seen for `de` and `en-US`, which is why I added the comment in the code snippet above. Disregard if I've misunderstood the L10n stuff.

File: bedrock/firefox/views.py [hunk 3]
```diff
@@ -579,21 +582,22 @@ def new(request):
...
+            elif experience == 'batmprivate':
+                if variant == 'b':
+                    template = 'firefox/new/batm/machine.html'
+                else:
+                    template = 'firefox/new/batm/private.html'
+            else:
+                template = 'firefox/new/onboarding/scene1.html'
+        elif lang_file_is_active('firefox/new/onboarding' 'firefox/new/batm', locale):
+            if experience == 'batmprivate':
+                template = 'firefox/new/batm/private.html'
```
jpetto: Same logic question here as above.

bedrock/firefox/templates/firefox/new/batm/machine.html
```diff
@@ -0,0 +1,45 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import google_play_button with context %}
```
jpetto: Are we using this macro/showing this button anywhere?

bedrock/firefox/templates/firefox/new/batm/machine.html
```diff
@@ -0,0 +1,45 @@
...
+{% endblock %}
+
+{% block body_class %}scene-1 machine-animated{% endblock %}
+
+{% block content %}
+<main role="main">
+  <section class="machine">
+    <video id="machine" preload="auto" autoplay loop muted>
+      <source src="https://assets.mozilla.net/video/batm-animation.webm" type="video/webm">
+      <source src="https://assets.mozilla.net/video/batm-animation.mp4" type="video/mp4">
```
jpetto: It works in Safari! (Do we know why it works this time?)

bedrock/firefox/templates/firefox/new/batm/machine.html
```diff
@@ -0,0 +1,45 @@
...
+    {{ download_firefox(alt_copy=_('Download Firefox'), locale_in_transition=True, button_color='button-arrow') }}
+  </div>
+
+</main>
+{% endblock %}
+
+{% block js %}
+  <!--[if IE 9]>
+    {% javascript 'matchmedia_addlistener' %}
+  <![endif]-->
```
jpetto: Do we still need this lib included? Looks like all the `matchMedia` code was removed from the JS file...

bedrock/firefox/templates/firefox/new/batm/machine.html
craigcook: It was the hosting. MP4s served directly from bedrock with all our extra security headers were being flagged as unsafe content in Safari. Serving them from this S3 bucket doesn't have all those extra headers so Safari is happy to treat it as safe content... seems counter-intuitive, but oh well.

bedrock/firefox/templates/firefox/new/batm/machine.html
craigcook: Good catch! Will remove.

bedrock/firefox/views.py
craigcook: Yeah, this was the bit I was least confident about. It seems to get the expected result but just felt wrong and I wasn't sure what I was missing... now I realize I was missing the `and` bit, rather than nesting a second if. I'll update to your version which is much more sound. Thanks!

bedrock/firefox/templates/firefox/new/batm/machine.html
craigcook: Just copypasta. This macro is imported on all the other BATM variants but may not be needed on any of them. Everything still seems to work without it so it could be a holdover from some past variant that had an Android button.

bedrock/firefox/templates/firefox/new/batm/machine.html
jpetto: Extra security headers === unsafe content? Ooookay. 🙄

bedrock/firefox/templates/firefox/new/batm/private.html
```diff
@@ -2,18 +2,20 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% add_lang_files "firefox/new/batm" %}
+
 {% from "macros.html" import google_play_button with context %}
```
jpetto: Can we get rid of these where possible? No sense if having extra cruft, eh?

craigcook: Yeah, I can remove it from all the BATM variants, just didn't want to touch every file if I didn't have to. But if it's just this deletion that shouldn't be too bad.

bedrock/firefox/templates/firefox/new/batm/private.html
```diff
@@ -2,18 +2,18 @@
...
-    {% include 'includes/optimizely.html' %}
+{% block experiments %}
+  {% if switch('experiment-firefox-new-batm-anim', ['en-US']) %}
+    {% javascript 'experiment_firefox_new_batm_anim' %}
   {% endif %}
 {% endblock %}
 
 {% block main_header_copy %}
-<p id="main-header-copy" data-experience="batmprivate">
+<p id="main-header-copy" data-experience="batmprivate" data-variant="a">
```
jpetto: Just noticed - do we need this `data-variant="a"` attribute? I don't think it ever does anything, as there's no JS bundle in this template...

bedrock/firefox/views.py
```diff
@@ -579,19 +582,19 @@ def new(request):
...
                 template = 'firefox/new/batm/resist.html'
+            elif experience == 'batmprivate':
+                if variant == 'b':
+                    template = 'firefox/new/batm/machine.html'
+                else:
+                    template = 'firefox/new/batm/private.html'
             else:
                 template = 'firefox/new/onboarding/scene1.html'
+        elif lang_file_is_active('firefox/new/batm', locale) and experience == 'batmprivate':
+            template = 'firefox/new/batm/private.html'
```
jpetto: Looking at this again, do we need to do the same `if variant == 'b'` check in this instance?

bedrock/firefox/templates/firefox/new/batm/private.html
craigcook: I think this is for GA but I could be wrong.

bedrock/firefox/views.py
craigcook: It's only `private.html` that is getting localized; the animated version is en-US only. If they decide to test that page in German at some later date we can update the view but I think this is probably OK.

media/js/firefox/new/scene1-batm.js
```diff
@@ -5,11 +5,16 @@
...
+    var exp = title.data('experience');
+    var variant = title.data('variant');
 
     $('.download-link').each(function(i, link) {
-        if (exp && link.href.indexOf('scene=2') > -1) {
-            // specify v=1 template for scene 2
+        if (exp && variant && link.href.indexOf('scene=2') > -1) {
+            // append experience and variant
+            link.href = link.href.replace('scene=2', 'scene=2&xv=' + exp + '&v=' + variant);
+        } else if (exp && !variant && link.href.indexOf('scene=2') > -1) {
```
craigcook: This is repetitive and I suspect there may be a more efficient way to do it.

bedrock/firefox/templates/firefox/new/batm/private.html
jpetto: I think this still needs adjustment. Currently, if you land on `/firefox/new/?xv=batmprivate`, the button to scene 2 contains `/firefox/new/?scene=2&xv=batmprivate&v=a`, which, I believe, would be bad for GA (as it would be artificially inflating visitors to the "a" variation).

You could remedy this by only including this data attribute if `v=a` exists in the querystring by changing the line above to:

`<p id="main-header-copy" data-experience="batmprivate"{% if request.GET.get('v', None) == 'a' %} data-variant="a"{% endif %}>`

media/js/firefox/new/scene1-batm.js
```diff
@@ -5,11 +5,16 @@
...
+            // append experience
             link.href = link.href.replace('scene=2', 'scene=2&xv=' + exp);
```
jpetto: Yeah, I think we could be a little clearer and more performant with something like:

```
var newParams = 'scene=2';

// conditionally construct new params for download links
if (exp) {
    newParams += '&xv=' + exp;
}

if (variant) {
    newParams += '&v=' + variant;
}

$('.download-link').each(function(i, link) {
    link.href = link.href.replace('scene=2', newParams);
});
```

PR #5022, opened 2017-08-02
Fix Bug 1386702 - Remove XP/Vista downloads from Desktop Channel page

kyoshino:
## Description

Remove the XP/Vista SHA-1 download link from non-Release download buttons, show the unsupported message instead.

## Issue / Bugzilla link

[Bug 1386702](https://bugzilla.mozilla.org/show_bug.cgi?id=1386702)

## Testing

The simplest way to test this is changing the class name on `<html>` to `windows sha-1 x86 js loaded` and see what happens with the download buttons. On the channel page, all the buttons will be replaced with the "Your system doesn't meet the requirements to run Firefox" message, while the regular Release channel buttons on `/firefox/` etc. won't be affected.

### DIFF
--- a/bedrock/firefox/templates/firefox/includes/download-button.html
+++ b/bedrock/firefox/templates/firefox/includes/download-button.html
@@ -37,7 +37,7 @@
 {% set download_class = download_class ~ ' download-button-desktop' if not (show_android or show_ios) else download_class %}
 {% set download_class = download_class ~ ' download-button-ios' if not (show_desktop or show_android) else download_class %}
 
-<div id="{{ id }}" class="{{ download_class }}">
+<div id="{{ id }}" class="{{ download_class }}" data-channel="{{ channel or 'release' }}">
   {# no JS #}
   <div class="nojs-download">
     {{ alt_buttons(builds) }}
--- a/media/css/sandstone/buttons.less
+++ b/media/css/sandstone/buttons.less
@@ -402,7 +402,8 @@ ul.download-list {
     display: none !important;
 }
 
-.windows.sha-1 .download-button .os_winsha1 {
+.windows.sha-1 .download-button[data-channel="release"] .os_winsha1,
+.windows.sha-1 .download-button:not([data-channel="release"]) .unsupported-download {
     display: block !important;
 }
 
--- a/bedrock/firefox/templatetags/helpers.py
+++ b/bedrock/firefox/templatetags/helpers.py
@@ -39,6 +39,11 @@ def desktop_builds(channel, builds=None, locale=None, force_direct=False,
             if plat_os == 'win64':
                 continue
 
+        # Output Windows XP/Vista link only for the regular Release channel,
+        # which actually serves Firefox 52 ESR.
+        if channel != 'release' and plat_os == 'winsha1':
+            continue
+
         # And generate all the info
         download_link = firefox_desktop.get_download_url(
             channel, version, plat_os, locale,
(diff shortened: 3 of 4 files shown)


media/css/sandstone/buttons.less
```diff
@@ -402,7 +402,8 @@ ul.download-list {
     display: none !important;
 }
 
-.windows.sha-1 .download-button .os_winsha1 {
+.windows.sha-1 .download-button[data-channel="release"] .os_winsha1,
+.windows.sha-1 .download-button:not([data-channel="release"]) .unsupported-download {
```
alexgibson: I'm pretty sure this is going to break older IE's. Support for `not` is patchy at best, I'd rather find a solution that does not rely on this.

bedrock/firefox/templates/firefox/includes/download-button.html
```diff
@@ -37,7 +37,7 @@
 {% set download_class = download_class ~ ' download-button-desktop' if not (show_android or show_ios) else download_class %}
 {% set download_class = download_class ~ ' download-button-ios' if not (show_desktop or show_android) else download_class %}
 
-<div id="{{ id }}" class="{{ download_class }}">
+<div id="{{ id }}" class="{{ download_class }}" data-channel="{{ channel or 'release' }}">
```
alexgibson: Sadly IE6 does not support attribute selectors afaik

alexgibson:
A couple more points to consider:

- Bouncer automatically redirects all sha-1 product download links to FF52 ESR.
- Unless there is an alternative and simple solution for this, I'm not sure if someone visiting the page with JS disabled is reason enough to make this change?

kyoshino:
Hmm, forgot this should be considering IE6. I don't have an XP/Vista machine for testing, but IIUC the SHA-1 ESR download button is displayed for XP users even when JavaScript is enabled, which is wrong for Beta, DevEdition and Nightly (but okay for the Release channel.) Will fix the IE6 issue.

alexgibson:
Yes all download buttons are displayed without JS disabled since there is no way to tell if a user is on XP, but like I say I'd say this so really a minor issue imo.

kyoshino:
In terms of the small XP/Vista user base, it's a minor issue for sure, but serving ESR 52 to people trying to download Beta, DevEdition or Nightly is an obvious bug. [site.js](https://github.com/mozilla/bedrock/blob/master/media/js/base/site.js#L166-L169) is detecting those users, and this PR simply uses the `sha-1` class name.

alexgibson:
> but serving ESR 52 to people trying to download Beta, DevEdition or Nightly is an obvious bug

Actually this was intentional on the bouncer side, afaik.

kyoshino:
Serving ESR 52  to people trying to download the *Release* build is intentional, right?

alexgibson:
> Serving ESR 52  to people trying to download the Release build is intentional, right?

All users on XP/Vista who see the sha-1 bouncer links will download ESR52, irrespective of the produce yes.

kyoshino:
I know that's how the bouncer is designed, but showing a wrong download button here is a different topic.

alexgibson:
> I know that's how the bouncer is designed, but showing a wrong download button here is a different topic.

Right. Like I say if there's a simple solution for this then I'm all for it. But pre-release audiences are a minority, and also more tech savy. Hence, I don't think this is worth an overly complicated change that is hard to test/QA.

kyoshino:
Actually I can simplify the patch by getting rid of the Python bits, because the unsupported message is just a magic of HTML/CSS. Then the *wrong* button will be displayed when JS is disabled, but I think that's okay.

kyoshino:
Given the XP/Vista support EOL is approaching, I'm closing this PR and bug as WONTFIX.

PR #5025, opened 2017-08-03
[fix bug 1387034] L10n layout issue on /firstrun page

alexgibson:
## Description
- Fixes an overlapping link issue on `/firstrun` that only effects longer strings on certain locales.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1387034

## Testing
- Testing `de` or `fr` locales should no longer have the issue.

### DIFF
--- a/media/css/firefox/firstrun/onboarding.less
+++ b/media/css/firefox/firstrun/onboarding.less
@@ -84,24 +84,15 @@ img {
     height: 60%;
     float: left;
     margin: 79px 0;
-    position: relative;
-
-    img {
-        position: relative;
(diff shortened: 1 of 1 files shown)


media/css/firefox/firstrun/onboarding.less
```diff
@@ -84,24 +84,15 @@ img {
     height: 60%;
     float: left;
     margin: 79px 0;
-    position: relative;
-
-    img {
```
alexgibson: This selector appears to be unused

PR #5028, opened 2017-08-04
Remove euw from deployments

jgmize:
https://github.com/mozmeao/infra/issues/336


pmac:
Something went wrong at Circle. Rerunning now.

PR #5035, opened 2017-08-07
[fix bug 1388144] Update legal docs, Firefox Cloud Services

craigcook:
## Description
Update legal-docs submodule to latest, with updates to Cloud Services policies. This needs to go live on 8 August for the Firefox 55 release.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=bug-1388144


pmac:
Not merging because 1) circle isn't done, and 2) not sure if we should wait until tomorrow. If we don't have to wait please merge at will.

PR #5039, opened 2017-08-08
[fix bug 1387984] Implement notification for out-of-date Firefox users on /firstrun & /whatsnew

alexgibson:
## Description
- Displays out-of-date Firefox notification to users on all /firstrun and /whatsnew templates (en-US only).
- Notification only shows for Firefoxen that are at least 2 major versions out of date and within a 5% sample limit.
- CTA click opens modal with instructions to update/restart.
- Creates new base templates which all /firstrun and /whatsnew pages inherit from.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1387984

## Testing
Demo:
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/54.0/firstrun/
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/54.0/whatsnew/
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/firstrun/
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/whatsnew/

Things to check:
- Notification shows as expected on Firefox <= 53 within defined sample rate.
- Regular notification shown on other pages still works as expected with no regressions.
- Modal dialog shows correct information depending on if the platform is MacOS / Windows.
- Notification doesn't break onboarding animation / layout.


alexgibson:
Ok, still waiting on final confirmation of GA bits, but this is close enough to complete to be ready for code review.

bedrock/firefox/templates/firefox/firstrun/firstrun.html
```diff
@@ -1,99 +0,0 @@
-{# This Source Code Form is subject to the terms of the Mozilla Public
```
alexgibson: I'm removing this template as it appears to have been orphaned some time ago.

alexgibson:
Ok, Peter has confirmed there are no additional GA requirements here. This should be good to merge once reviewed.

bedrock/firefox/templates/firefox/australis/base.html
```diff
@@ -54,5 +60,10 @@
 {% block email_form %}{% endblock %}
 {% block js %}{% endblock %}
 
-{# Bug 1381776 #}
-{% block update_notification %}{% endblock %}
+{# Bug 1387984 #}
+{% block update_notification %}
+  {% if switch('firefox-update-notification-modal', ['en-US']) %}
```
alexgibson: This PR uses a different switch name to the regular notification, so it can be toggled on/off independently. It will need enabling once this merges/goes to prod.

media/js/base/mozilla-notification-banner.js
```diff
@@ -16,6 +16,7 @@ if (typeof Mozilla === 'undefined') {
     NotificationBanner.COOKIE_CODE_ID = 'mozilla-notification-banner';
     NotificationBanner.COOKIE_EXPIRATION_DAYS = 21; // default cookie expiry 21 days
     NotificationBanner.COOKIE_INTERACTION_VALUE = 'interacted';
+    NotificationBanner.SAMPLE_RATE = 0.05; // 5% sample rate limit
```
jpetto: Would this be better specified as an option either per instance or per config? Feels a bit odd to have this library specific.

jpetto:
`/firefox/54.0/firstrun/` could use a bottom margin, as the FxA iframe is flush against the bottom of the viewport when the banner is shown (on my 15", anyway).

jpetto:
The post "skip this step" onboarding animation seems to suffer just a little bit with the banner open, but I don't think it's bad enough to spend time on.

jpetto:
The modal `<h2>` and `<h3>` copy is white/invisible on the `firefox/firstrun/firstrun-horizon.html` template.

jpetto:
Tested on all `/firstrun` and `/whatsnew` pages on both OS X and Windows. Tested existing notification banner on other pages and didn't notice any regressions.

New base template approach is looking good.

media/js/base/mozilla-notification-banner-modal-init.js
```diff
@@ -0,0 +1,112 @@
...
+            'closeLabel': 'Close'
+        }
+    ];
+
+    /**
+     * Determine if Firefox cliient is out of date.
+     * @return {Boolean} - Returns true if client is at least 2 major versions out of date.
+     */
+    var _isClientOutOfDate = function() {
+        var clientVersion = client._getFirefoxMajorVersion();
```
alexgibson: Bah, caught again :/

media/js/base/mozilla-notification-banner.js
alexgibson: This is just a default value, it can easily be overridden per instance if required. I don't think we have any plans to display more than one notification on the same page at the same time (at least, I really hope not from a UX perspective), and we don't have any requirements to have different sample rates per config either, so until we do i think this is probably acceptable. Let me know if you think differently though?

alexgibson: Hmm, on second thought - maybe it would just be a bit cleaner. Let me mull it over 😉

alexgibson:
> /firefox/54.0/firstrun/ could use a bottom margin, as the FxA iframe is flush against the bottom of the viewport when the banner is shown (on my 15", anyway).

Nice catch

> The post "skip this step" onboarding animation seems to suffer just a little bit with the banner open, but I don't think it's bad enough to spend time on.

Yeah, my main aim here was just not to kill the page completely (even doing that sucked up way more time than I initially planned for).

> The modal <h2> and <h3> copy is white/invisible on the firefox/firstrun/firstrun-horizon.html template.

It's an easy fix, I'll do this even though it may not be seen currently 👍

bedrock/settings/static_media.py
```diff
@@ -1610,6 +1611,14 @@
         ),
         'output_filename': 'js/firefox-update-notification-bundle.js',
     },
+    'firefox-update-notification-modal': {
```
alexgibson: I'll make the CSS bundle consistent and add `-firefox` to it. This also makes it consistent with the other notification for regular pages.

media/js/base/mozilla-notification-banner.js
alexgibson: Ok, this can't be per config, since we need to know the sample rate before a config is selected. What i can do it make a couple of convenience methods to get/set a sample rate which can be configured per instance.

alexgibson:
@jpetto thanks for the review, I've added a commit with some fixes for your suggestions and pushed the latest to demo. I think this is ready for another look.

media/js/base/mozilla-notification-banner.js
```diff
@@ -12,6 +12,7 @@ if (typeof Mozilla === 'undefined') {
 
     var NotificationBanner = {};
     var _options = {};
+    var _sampleRate = 0;
```
jpetto: Small nit - a comment would be nice detailing that `0` actually means 100%.

jpetto:
@alexgibson squish time?

alexgibson:
Thanks @jpetto - added a comment and squashed.

PR #5041, opened 2017-08-09
Fix bug 1344220: Enable stub installer for all locales for all windows

pmac:


bedrock/settings/base.py
```diff
@@ -679,25 +679,19 @@ def lazy_email_backend():
...
         'win': STUB_INSTALLER_ALL,
-        # should come back in Fx 55
-        # bug 1348610
-        # 'win64': ['en-us'],
+        'win64': STUB_INSTALLER_ALL,
     },
     'nightly': {
         'win': STUB_INSTALLER_ALL,
-        'win64': ['en-us'],
+        'win64': STUB_INSTALLER_ALL,
```
jpetto: Not a blocker for this PR, but since these are now all the same, maybe good to refactor/remove the logic at some point?

pmac: Agreed. I think we can remove the whole thing, but didn't want to do it quite yet in case we need to quickly roll back.

jpetto: Yep, agree doing so now would be a little too optimistic. 😉

PR #5042, opened 2017-08-09
Wrap Firefox retention thank you pages in l10n blocks.

ScottDowne:
## Description
Localize the Firefox retention thank you pages.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1388860

### DIFF
--- a/bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
+++ b/bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
@@ -4,14 +4,16 @@
 
 {% extends "base-pebbles.html" %}
 
+{% add_lang_files "firefox/retention/thank-you" %}
+
 {% block page_title %}
-  Thank You
+  {{ _('Thank You') }}
 {% endblock %}
 
 {% block page_title_suffix %}{% endblock %}
 
 {% block page_desc %}
-  Thank you page.
+  {{ _('Thank you page.') }}
 {% endblock %}
 
 {% block page_css %}
@@ -25,33 +27,37 @@
   <canvas id="confetti-canvas"></canvas>
   <header>
     <h1>
-      <span class="white">It's all</span></br>
+    {% if LANG.startswith('en') %}
+      <span class="white">It’s all</span><br/>
       thanks to you
+    {% else %}
+      {{ _('It’s all thanks to you') }}
+    {% endif %}
     </h1>
     <p>
-      Choosing Firefox helps Mozilla make the Internet<br/>a better place. Here's what you can do next:
+      {{ _('Choosing Firefox helps Mozilla make the Internet a better place. Here’s what you can do next:') }}
     </p>
   </header>
   <section>
     <article>
-      <h1>Make Firefox your default browser</h1>
-      <div class="time">1 min action</div>
-        <a href="https://support.mozilla.org/kb/make-firefox-your-default-browser">Set as your default</a>
+      <h1>{{ _('Make Firefox your default browser') }}</h1>
+      <div class="time">{{ _('1 min action') }}</div>
+        <a href="https://support.mozilla.org/kb/make-firefox-your-default-browser">{{ _('Set as your default') }}</a>
     </article>
     <article>
-      <h1>Get Firefox on your phone</h1>
-      <div class="time">2 min install</div>
-      <a href="https://support.mozilla.org/kb/install-firefox-android-device-using-google-play">Download Firefox</a>
+      <h1>{{ _('Get Firefox on your phone') }}</h1>
+      <div class="time">{{ _('2 min install') }}</div>
+      <a href="https://support.mozilla.org/kb/install-firefox-android-device-using-google-play">{{ _('Download Firefox') }}</a>
     </article>
     <article>
-      <h1>Tell your friends about Firefox</h1>
-      <div class="time">1 min share</div>
-      <a href="https://twitter.com/home?status=%F0%9F%94%A5%2B%F0%9F%A6%8A=%F0%9F%99%8C%F0%9F%8F%BE%20Join%20me%20in%20the%20fight%20for%20an%20open%20web%20by%20choosing%20Firefox!%20https%3A//mozilla.org/firefox/new">Send a tweet</a>
+      <h1>{{ _('Tell your friends about Firefox') }}</h1>
+      <div class="time">{{ _('1 min share') }}</div>
+      <a href="{{ 'https://www.twitter.com/intent/tweet?url=%s&text=%s%s'|format('https://mozilla.org/firefox/new'|urlencode, '%F0%9F%94%A5%2B%F0%9F%A6%8A=%F0%9F%99%8C%F0%9F%8F%BE%20', _('Join me in the fight for an open web by choosing Firefox!')|urlencode)|e }}">{{ _('Send a tweet') }}</a>
     </article>
     <article>
-      <h1>Make the Internet a safer place</h1>
-      <div class="time">4 min read</div>
-      <a href="https://www.mozilla.org/internet-health/privacy-security/">Learn more</a>
+      <h1>{{ _('Make the Internet a safer place') }}</h1>
+      <div class="time">{{ _('4 min read') }}</div>
+      <a href="https://www.mozilla.org/internet-health/privacy-security/">{{ _('Learn more') }}</a>
     </article>
   </section>
 </main>
@@ -59,7 +65,7 @@
   <div class="content">
     {{ email_newsletter_form(newsletters='mozilla-foundation',
        title=_('Stay in touch for more cool stuff'),
-       subtitle=_('Get the latest & greatest from Firefox<br/>delivered straight to your inbox.'),
+       subtitle=_('Get the latest & greatest from Firefox delivered straight to your inbox.'),
        button_class='button-light',
        spinner_color='#fff')
     }}
--- a/bedrock/etc/templates/etc/firefox/retention/thank-you-b.html
+++ b/bedrock/etc/templates/etc/firefox/retention/thank-you-b.html
@@ -4,14 +4,16 @@
 
 {% extends "base-pebbles.html" %}
 
+{% add_lang_files "firefox/retention/thank-you" %}
+
 {% block page_title %}
-  Thank You
+  {{ _('Thank You') }}
 {% endblock %}
 
 {% block page_title_suffix %}{% endblock %}
 
 {% block page_desc %}
-  Thank you page.
+  {{ _('Thank you page.') }}
 {% endblock %}
 
 {% block page_css %}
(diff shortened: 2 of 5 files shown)


bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
```diff
@@ -25,41 +25,45 @@
   <canvas id="confetti-canvas"></canvas>
   <header>
     <h1>
-      <span class="white">It's all</span></br>
+    {% if LANG.startswith('en') %}
+      <span class="white">It’s all</span>
       thanks to you
+    {% else %}
+      It’s all thanks to you
```
jpetto: This string needs to be wrapped.

bedrock/etc/templates/etc/firefox/retention/thank-you-b.html
```diff
@@ -26,48 +26,58 @@
...
     <header>
       <h1>
-        <span class="white">It's all</span></br>
-        thanks</br>
+      {% if LANG.startswith('en') %}
+        <span class="white">It’s all</span>
+        thanks<br/>
         to you
+      {% else %}
+        It’s all thanks to you
```
jpetto: Same here - string needs to be wrapped.

bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
```diff
@@ -25,41 +25,45 @@
...
+      {{ _('Choosing Firefox helps Mozilla make the Internet a better place. Here’s what you can do next:') }}
     </p>
   </header>
   <section>
     <article>
-      <h1>Make Firefox your default browser</h1>
-      <div class="time">1 min action</div>
-        <a href="https://support.mozilla.org/kb/make-firefox-your-default-browser">Set as your default</a>
+      <h1>{{ _('Make Firefox your default browser') }}</h1>
+      <div class="time">{{ _('1 min action') }}</div>
```
jpetto: @peiying2 or @flodolo: Do these strings depicting time need any special treatment? These strings are meant to communicate the time ("1 min" or "4 min") necessary to complete the task ("action" or "read"). Would an L10n comment be helpful here?

jpetto:
One other primary concern is the font. Based on the file size (~20KB), I'm guessing Antonio supports only Latin characters, which means it won't work for many locales. We can specify Open Sans as the default, then target Antonio at some or all Latin-based languages in the CSS (using the `lang` attribute on the `<html>` element). Alternatively, we could drop Antonio altogether.

Also, the use of `text-transform: uppercase` does not play well with all character sets, so suggest applying that only when Antonio can be used.

ScottDowne:
@jpetto Thanks for checking!

I'm fine with backup to Open Sans where we cannot use Antonio.

I'll fix up the lines I forgot to wrap.

As an alternative to `text-transform: uppercase` I can just put the strings as uppercase and let the translators change this or not? Does that work?

flodolo:
> As an alternative to text-transform: uppercase I can just put the strings as uppercase and let the translators change this or not? Does that work?

Normally that would be the right approach, but not here: the entire page is basically uppercase, it would be a pain to translate. Call me old school, but I personally question a design with so much content uppercase.

ScottDowne:
@flodolo OK, so jpetto's suggestion of only applying uppercase to English would work? If so, I'll do that.

flodolo:
> @flodolo OK, so jpetto's suggestion of only applying uppercase to English would work?

Yes, that would be for the best.

ScottDowne:
@flodolo @jpetto  So I have it working with Open Sans. However, any strong opinions on adding Open Sans Condensed for this? It seems to work a bit better, is closer visually to Antonio, and closer to the designs. I can get a version of it here: https://google-webfonts-helper.herokuapp.com/fonts/open-sans-condensed?subsets=cyrillic,greek-ext,latin,greek,cyrillic-ext,latin-ext

Not sure which of the charsets I should include, though.

Thoughts?

flodolo:
> Not sure which of the charsets I should include, though.

I don't think the charsets supported are enough (don't see Chinese, Hindi to cite a couple).

We've broken mozilla.org for some languages multiple times just by moving away from Open Sans.

ScottDowne:
@flodolo Gotchya, ok, I'll stick with regular open sans.

ScottDowne:
@jpetto Pushed up fixes.

I did a bit of css to make sure changes to compensate for open sans being different width and heights than Antonio.

bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
peiying2: @jpetto we will add comments here.

flodolo:
We should load a single .lang file for both templates (via `add_lang_files`), e.g. `firefox/retention/thank-you`.

BTW, it's still completely unclear to me which locales are the target for these pages, since they look orphaned. What's the URL for testing? Who's going to get to these pages and under which conditions?

peiying2:
@flodolo from what I heard, the traffic is directed from snippets. Does this mean these pages are limited to the snippets supported locales?  We are still waiting for the clarity on this.

peiying2:
Snippet team will coordinate the timing of the page launch. They have filed this ticket. But in terms of which languages are needed, the snippet team won't drive the decision.  They will support whatever is needed on mozilla.org: https://bugzilla.mozilla.org/show_bug.cgi?id=1394945

peiying2:
Through email thread, the decision is made and it is official: de, es-es, fr, id, pl, pt-br, ru, zh-tw.

Deadline for these two pages and snippet: Sept 15.

flodolo:
> We should load a single .lang file for both templates (via add_lang_files), e.g. firefox/retention/thank-you.

Can  someone fix this so we can extract the strings? Deadline is already way too close for my like
I also need a place to point localizers to.

flodolo:
Found the URLs in the meantime
https://www-dev.allizom.org/etc/firefox/retention/thank-you-a/
https://www-dev.allizom.org/etc/firefox/retention/thank-you-b/

I'm likely going to push to this branch to move things.

flodolo:
@jpetto 
Fixed, rebased against master and squashed. Can you take a last look?

bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
```diff
@@ -25,41 +27,45 @@
...
+      <div class="time">{{ _('2 min install') }}</div>
+      <a href="https://support.mozilla.org/kb/install-firefox-android-device-using-google-play">{{ _('Download Firefox') }}</a>
     </article>
     <article>
-      <h1>Tell your friends about Firefox</h1>
-      <div class="time">1 min share</div>
-      <a href="https://twitter.com/home?status=%F0%9F%94%A5%2B%F0%9F%A6%8A=%F0%9F%99%8C%F0%9F%8F%BE%20Join%20me%20in%20the%20fight%20for%20an%20open%20web%20by%20choosing%20Firefox!%20https%3A//mozilla.org/firefox/new">Send a tweet</a>
+      <h1>{{ _('Tell your friends about Firefox') }}</h1>
+      <div class="time">{{ _('1 min share') }}</div>
+      <a href="https://twitter.com/home?status=%F0%9F%94%A5%2B%F0%9F%A6%8A=%F0%9F%99%8C%F0%9F%8F%BE%20Join%20me%20in%20the%20fight%20for%20an%20open%20web%20by%20choosing%20Firefox!%20https%3A//mozilla.org/firefox/new">{{ _('Send a tweet') }}</a>
```
flodolo: Actually, this should be localized :-\

bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
```diff
@@ -25,41 +27,45 @@
...
+      <div class="time">{{ _('2 min install') }}</div>
+      <a href="https://support.mozilla.org/kb/install-firefox-android-device-using-google-play">{{ _('Download Firefox') }}</a>
     </article>
     <article>
-      <h1>Tell your friends about Firefox</h1>
-      <div class="time">1 min share</div>
-      <a href="https://twitter.com/home?status=%F0%9F%94%A5%2B%F0%9F%A6%8A=%F0%9F%99%8C%F0%9F%8F%BE%20Join%20me%20in%20the%20fight%20for%20an%20open%20web%20by%20choosing%20Firefox!%20https%3A//mozilla.org/firefox/new">Send a tweet</a>
+      <h1>{{ _('Tell your friends about Firefox') }}</h1>
+      <div class="time">{{ _('1 min share') }}</div>
+      <a href="{{ 'https://www.twitter.com/intent/tweet?url=%s&text=%s%s'|format('https://mozilla.org/firefox/new'|urlencode, '%F0%9F%94%A5%2B%F0%9F%A6%8A=%F0%9F%99%8C%F0%9F%8F%BE%20', _('Join me in the fight for an open web by choosing Firefox!')|urlencode)|e }}">{{ _('Send a tweet') }}</a>
```
flodolo: I copied the approach from existing pages, not sure if it's the best way to do it.

flodolo:
(thanks to @TheoChevalier for catching the rogue tweet, we should really rethink the whole en-US first approach, as in the page should be fully localizable from the start)

bedrock/etc/templates/etc/firefox/retention/thank-you-b.html
```diff
@@ -26,48 +28,54 @@
...
+        {{ _('Choosing Firefox helps Mozilla make the Internet a better place.') }}
       </p>
     </header>
   </section>
   <section class="right-panel">
     <h1>
-      Here's more super-easy things</br>
-      you can do to keep supporting</br>
-      our mission:
+      {{ _('Here’s more super-easy things you can do to keep supporting our mission:') }}
```
peiying2: Here's .... things -> Here are ... things

flodolo: Since this page is already live for en-US, I'm tempted to just ignore this and leave it as is. This would delay exposing strings, and I would get it out of the way, unless we agree quickly on fixing it.

bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
```diff
@@ -25,41 +27,45 @@
...
+      <div class="time">{{ _('1 min action') }}</div>
+        <a href="https://support.mozilla.org/kb/make-firefox-your-default-browser">{{ _('Set as your default') }}</a>
     </article>
     <article>
-      <h1>Get Firefox on your phone</h1>
-      <div class="time">2 min install</div>
-      <a href="https://support.mozilla.org/kb/install-firefox-android-device-using-google-play">Download Firefox</a>
+      <h1>{{ _('Get Firefox on your phone') }}</h1>
+      <div class="time">{{ _('2 min install') }}</div>
+      <a href="https://support.mozilla.org/kb/install-firefox-android-device-using-google-play">{{ _('Download Firefox') }}</a>
```
alexgibson: Not related to this PR, but why do we send users to a SUMO article for Android here? A mozorg page like `/firefox/android/` might be better, although we have `/firefox/ios/` too. Probably worth another bug as opposed to piling on this PR.

bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
alexgibson: Seems to work for me 👍

alexgibson:
I filed a bug for the mobile download link: https://bugzilla.mozilla.org/show_bug.cgi?id=1395475

PR #5043, opened 2017-08-09
[fix bug 1384362] Add WebVR to technology page

craigcook:
## Description
Replacing A-Frame with WebVR, with new strings begind the l10n tag `mozorg_tech_webvr`. I also took the opportunity to switch to standardized font sizes.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1384362


media/css/mozorg/technology.scss
```diff
@@ -878,11 +807,11 @@ html[dir="rtl"] #technologies > section {
         }
 
         p {
-            @include font-size(16px);
+            @include font-size-level5;
```
alexgibson: I wish this wasn't so akward with Zilla :/ 

This looks fine on desktop, but on mobile viewports this goes to `14px`, which looks about `12px` feels a little tiny. On the other hand, if we set this to `font-size-level4` and bump everything else up accordingly, it feels a little too large on desktop...

After mulling it over a bit, I think what you have done here might be the better option - but thought I'd flag it in case you had any further thoughts.

bedrock/mozorg/templates/mozorg/technology.html
```diff
@@ -51,14 +51,23 @@
...
           <a rel="external" href="https://aframe.io/" class="button">{{ _('Explore A-Frame') }}</a>
+        {% endif %}
         </div>
       </div>
       <footer>
         <ul>
-        {% if l10n_has_tag('emerging_tech_links') %}
+        {% if l10n_has_tag('mozorg_tech_webvr') %}
+          <li><a rel="external" href="https://research.mozilla.org/mixed-reality/">{{ _('Learn about WebVR') }}</a></li>
+          <li><a rel="external" href="https://aframe.io/">{{ _('Explore A-Frame') }}</a></li>
```
alexgibson: It looks like we already have the string `Explore A-Frame` translated above. Could we avoid adding another conditional here, seeing as `Learn about WebVR` is also in the l10n block below?

alexgibson:
Oops sorry, i totally missed that jpetto had been flagged for review here already.

media/css/mozorg/technology.scss
craigcook: Yeah... maybe we should stick with a hard `font-size(16px)` here since 14px really is too small for body copy in Zilla. Perhaps standard body copy should be the common exception to the "use size mixins" rule because I really don't want it to ever go smaller than 16px. Or we could even add another mixin like `font-size-body` that starts at 16px and goes to 18px on desktop? Maybe that's overthinking? Really this page just needs to be redesigned to follow our new type conventions.

alexgibson: Yeah, I think if we change anything now then a fixed 16px rule sounds like the best bet. Totally up to you though 👍

craigcook: Yeah, I'm gonna make this 16px. The best long-term solution will be to redesign the page.

PR #5044, opened 2017-08-09
Firefox retention thank you pages fix canvas after tab change.

ScottDowne:
## Description
On FF nightly, if you open one of the Retention pages, tab away, and tab back after a few seconds, the canvas stops drawing.

This resets it. I'm trying to get a better understanding as to why, as it's something that's new in nightly and not on stable. Might even be a reproducible bug. I'll keep digging, but for now I don't see why not have people looking at this patch in it's current state.

### DIFF
--- a/media/js/etc/firefox/retention/confetti.js
+++ b/media/js/etc/firefox/retention/confetti.js
@@ -45,6 +45,7 @@
             defaultOpacity = 1;
         } else {
             window.addEventListener('resize', resizeWindow, false);
+            window.addEventListener('focus', resizeWindow, false);
         }
 
         var confettiSize = 6;


jpetto:
This bug seems to be fixed in Nightly - or at least I'm not seeing it Mac on 57.0a1 (2017-08-15). Let me know if you're still seeing this and/or if you think this patch is still worthwhile.

ScottDowne:
Still happening, I'm on Linux. Might be more of an OS related thing than I originally thought.

jpetto:
Gotcha. I don't see any harm in merging this.

jpetto:
Hm, actually, what do you think about using the [Page Visibility API](https://developer.mozilla.org/en-US/docs/Web/API/Page_Visibility_API) here instead? Should have better performance, and I don't think would trigger events when app switching.

alexgibson:
Closing due to inactivity, but feel free to reopen if this is something you would still like to fix. Thanks!

PR #5045, opened 2017-08-10
Fix bug 1389178: Dev Edition notes redirect to beta

pmac:
They were still redirecting to aurora notes, which is now stuck at an old version as well. Sys Requirments redirect had the same problem.

### DIFF
--- a/bedrock/releasenotes/tests/test_base.py
+++ b/bedrock/releasenotes/tests/test_base.py
@@ -358,10 +358,10 @@ class TestNotesRedirects(TestCase):
                    '/firefox/23.0beta/releasenotes/')
 
     @patch('bedrock.releasenotes.views.firefox_desktop.latest_version',
-           Mock(return_value='24.0a2'))
+           Mock(return_value='23.0b1'))
     def test_desktop_developer_version(self):
         self._test('/firefox/developer/notes/',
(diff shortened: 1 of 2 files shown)


pmac:
Left the aurora redirects in there just because they should do something and there may well still be links floating around.

bedrock/releasenotes/tests/test_base.py
```diff
@@ -358,10 +358,10 @@ def test_desktop_beta_version(self):
                    '/firefox/23.0beta/releasenotes/')
 
     @patch('bedrock.releasenotes.views.firefox_desktop.latest_version',
-           Mock(return_value='24.0a2'))
+           Mock(return_value='23.0b1'))
     def test_desktop_developer_version(self):
         self._test('/firefox/developer/notes/',
-                   '/firefox/24.0a2/auroranotes/')
+                   '/firefox/23.0beta/releasenotes/')
```
jpetto: Do you think we should keep a test for aurora on desktop?

pmac: I don't see a desktop aurora test. Are you suggesting we add one?

jpetto: I just meant keeping a test for a `/firefox/xx.0ax/auroranotes/` url (as it's being removed in this change).

PR #5048, opened 2017-08-12
[bug 1329642] Remove Rust video test from technology page

craigcook:
## Description
Many months ago we did an A/B test promoting Rust at the top of the technology landing page. I honestly don't recall the findings of that test but I'm going to assume it was inconclusive and/or had negligible impact since we never made the Rust header permanent. This PR cleans up old test code.

## Issue / Bugzilla link


craigcook:
Alright, guess that Rust header was actually quite popular and should have been made permanent at some point long ago. However, the video is only in English, so... we're going to show the Rust version to English locales and the original to all others, but we can at least tidy up the traffic cop parts. Closing this PR.

PR #5049, opened 2017-08-14
[fix bug 1388871] Promote update banner winning copy.

jpetto:
## Description

Promotes winning copy to the default for site-wide banner and alternate firstrun/whatsnew banner.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1388871

## Testing

Essentially ensure no regressions introduced in terms of banner functionality. Also make sure strings are showing up properly in banners.

**Note that banners will be en-US only while new strings are localized.**

https://www-demo5.allizom.org/en-US/

*banner only has 5% display frequency for `/firstrun` and `/whatsnew`*
https://www-demo5.allizom.org/en-US/firefox/54.0/firstrun
https://www-demo5.allizom.org/en-US/firefox/54.0/whatsnew

1. Download an old version of Fx. 53.0.2 is available [here](https://ftp.mozilla.org/pub/firefox/releases/53.0.2/)
2. Once you have installed / opened the browser, open a new tab and type 'about:config'. Accept the warning that is displayed.
3. On the about:config page, search for `uitour`.
4. Verify that `browser.uitour.enabled` is already set to `true`.
5. Next create a new preference (Right click -> New -> String) called `browser.uitour.testingOrigins`.
6. Give that preference a value of `https://www-demo5.allizom.org`
7. Open a new tab and visit `https://www-demo5.allizom.org`

### DIFF
--- a/media/js/base/mozilla-notification-banner-init.js
+++ b/media/js/base/mozilla-notification-banner-init.js
@@ -7,65 +7,35 @@ $(function() {
     'use strict';
 
     var client = window.Mozilla.Client;
+    var utils = window.Mozilla.Utils;
 
-    var options = [
-        {
-            'id': 'fx-out-of-date-banner-copy1-direct-1',
-            'name': 'fx-out-of-date',
-            'experimentName': 'fx-out-of-date-banner-copy1',
-            'experimentVariant': 'direct-1',
-            'heading': 'Your browser security is at risk.',
-            'message': 'Update Firefox now to protect yourself from the latest malware.',
-            'confirm': 'Update now',
-            'confirmAction': 'Update Firefox',
-            'confirmLabel': 'Firefox for Desktop',
-            'url': '/firefox/new/?scene=2',
-            'close': 'Close',
-            'closeLabel': 'Close'
-        },
-        {
-            'id': 'fx-out-of-date-banner-copy1-direct-2',
-            'name': 'fx-out-of-date',
-            'experimentName': 'fx-out-of-date-banner-copy1',
-            'experimentVariant': 'direct-2',
-            'heading': 'Your Firefox is out-of-date.',
-            'message': 'Get the most recent version to keep browsing securely.',
-            'confirm': 'Update Firefox',
-            'confirmAction': 'Update Firefox',
-            'confirmLabel': 'Firefox for Desktop',
-            'url': '/firefox/new/?scene=2',
-            'close': 'Close',
-            'closeLabel': 'Close'
-        },
-        {
-            'id': 'fx-out-of-date-banner-copy1-foxy-1',
-            'name': 'fx-out-of-date',
-            'experimentName': 'fx-out-of-date-banner-copy1',
-            'experimentVariant': 'foxy-1',
-            'heading': 'Psst… it’s time for a tune up',
-            'message': 'Stay safe and fast with a quick update.',
-            'confirm': 'Update Firefox',
-            'confirmAction': 'Update Firefox',
-            'confirmLabel': 'Firefox for Desktop',
-            'url': '/firefox/new/?scene=2',
-            'close': 'Close',
-            'closeLabel': 'Close'
-        },
-        {
-            'id': 'fx-out-of-date-banner-copy1-foxy-2',
-            'name': 'fx-out-of-date',
-            'experimentName': 'fx-out-of-date-banner-copy1',
-            'experimentVariant': 'foxy-2',
-            'heading': 'Time to browse better!',
-            'message': 'Get the latest version of Firefox for extra speed and safety.',
-            'confirm': 'Update now',
-            'confirmAction': 'Update Firefox',
-            'confirmLabel': 'Firefox for Desktop',
-            'url': '/firefox/new/?scene=2',
-            'close': 'Close',
-            'closeLabel': 'Close'
-        }
-    ];
+    var headingText;
+    var messageText;
+    var confirmText;
+    var closeText;
+
+    // try to get localized copy
+    // if any of the below fail, the banner will detect missing strings and
+    // will not initialize
+    if (typeof utils !== 'undefined') {
+        headingText = utils.trans('global-fx-out-of-date-banner-heading');
+        messageText = utils.trans('global-fx-out-of-date-banner-message');
+        confirmText = utils.trans('global-fx-out-of-date-banner-confirm');
+        closeText = utils.trans('global-close');
+    }
+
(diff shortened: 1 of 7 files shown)


flodolo:
Are the strings final? We'll need them in a shared file, so the sooner we expose them, the better.

jpetto:
> Are the strings final? We'll need them in a shared file, so the sooner we expose them, the better.

As far as I know, they are final. The [request in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1388871#c2) seems pretty clear. I say expose away.

peiying2:
@craigcook @jpetto could you attach a screenshot on where these strings show up? It won't be easy to test these strings in context but at least we should have an en-US one as a reference.

jpetto:
Hopefully these will help:

"Standard" bar on most pages/URLs:

screen shot 2017-08-14 at 5 27 25 pm

Modal-invoking banner on `/whatsnew` and `/firstrun` pages:

screen shot 2017-08-14 at 5 27 01 pm

Modal open after clicking "Update Firefox" in banner above:

screen shot 2017-08-14 at 5 27 14 pm

peiying2:
Thanks!!

flodolo:
Sorry, once again this is really confusing from a localization point of view. I see 4 strings exposed in this PR (for the banner), where are the strings from the last screen? The last thing I want is to have this amount of text into a shared file, and that's what we need to use for the other 4

jpetto:
> Sorry, once again this is really confusing from a localization point of view. I see 4 strings exposed in this PR (for the banner), where are the strings from the last screen? The last thing I want is to have this amount of text into a shared file, and that's what we need to use for the other 4

The modal content is located in [base/templates/includes/firefox-update-instructions.html](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/includes/firefox-update-instructions.html). This modal is only displayed on `/firstrun` and `/whatsnew` pages, so the strings wouldn't have to be in a globally shared file. **However**, this modal is still being tested for effectiveness, so let's hold off on localizing that particular screen until we have enough traffic to determine if it's a good piece of UI. (Apologies, I'm coming in late on this project.)

flodolo:
> However, this modal is still being tested for effectiveness, so let's hold off on localizing that particular screen until we have enough traffic to determine if it's a good piece of UI. (Apologies, I'm coming in late on this project.)

Just to make sure we're on the same page: you're saying to hold off on all strings?

On a side note, we should load a better lang file for that template, e.g. `firefox/update-instructions.lang`.

jpetto:
Just hold off on strings in `firefox-update-instructions.html`. The 4 new strings in the base templates should be localized.

flodolo:
Where is "Firefox for Desktop" used?

jpetto:
> Where is "Firefox for Desktop" used?

Good catch! After reading through the notification lib comments, I see it's only used for GA. I did some refactoring to make this more clear, and removed the superfluous strings from the base templates.

media/js/base/mozilla-notification-banner-init.js
```diff
@@ -7,65 +7,39 @@ $(function() {
...
+    var confirmText;
+    var closeText;
+
+    // try to get localized copy
+    if (typeof utils !== 'undefined') {
+        headingText = utils.trans('global-fx-out-of-date-banner-heading');
+        messageText = utils.trans('global-fx-out-of-date-banner-message');
+        confirmText = utils.trans('global-fx-out-of-date-banner-confirm');
+        closeText = utils.trans('global-close');
+    // fall back to English copy if we can't get to the trans function for some reason
```
alexgibson: Would it be better not to show the notification at all, rather than show untranslated English strings? The way the lib works, if any of the required things are not supplied, the notification should not show at all anyway.

media/js/base/mozilla-notification-banner-init.js
```diff
@@ -7,65 +7,39 @@ $(function() {
...
+    } else {
+        headingText = 'Your Firefox is out-of-date.';
+        messageText = 'Get the most recent version to keep browsing securely.';
+        confirmText = 'Update Firefox';
+        closeText = 'Close';
+    }
+
+    var options = [{
```
alexgibson: As we now have only one choice of messaging, there's no need for this to be in an array. You can remove the call to `Mozilla.NotificationBanner.getOptions(options)` further down and just call `Mozilla.NotificationBanner.init(choice);` directly, passing in the object literal.

media/js/base/mozilla-notification-banner.js
```diff
@@ -313,11 +313,11 @@ if (typeof Mozilla === 'undefined') {
      * {String} heading (required) - Copy string for notification heading.
      * {String} message (required) - Copy string for notification message / subheading.
      * {String} confirm (required) - Copy string for main CTA button.
-     * {String} confirmAction (required) - String for action of CTA button in GA (this should always be English).
-     * {String} confirmLabel (required) - String for labelling CTA button in GA (this should always be English).
      * {String} url (required) - URL for main CTA link.
      * {String} close (required) - Copy string for close button.
-     * {String} closeLabel (required) - String for labelling close button in GA (this should always be English).
+     * {String} gaConfirmAction (required) - String for action of CTA button in GA (this should always be English).
```
alexgibson: 👍

media/js/base/mozilla-notification-banner-init.js
```diff
@@ -7,65 +7,35 @@ $(function() {
...
-    ];
+    var headingText;
+    var messageText;
+    var confirmText;
+    var closeText;
+
+    // try to get localized copy
+    // if any of the below fail, the banner will detect missing strings and
+    // will not initialize
+    if (typeof utils !== 'undefined') {
```
alexgibson: Curious, do you know of any pages where this might be an issue? `Mozilla.Utils` is served in the same base bundle as `Mozilla.Client`, yet we don't normally worry about this case for the latter lib. I'm not saying change anything as it works and may well be worth doing, but I'm curious is there are any fringe pages I'm not aware of, where we could start encouraging more stringent checks when using `Mozilla.Client`?

jpetto: To be honest, I am not sure. I haven't done a real analysis of what would happen if that `common` bundle didn't load in relation to other scripts on the page. The check here isn't hurting anything and *looks* like a safety measure, but could be superfluous.

It would be great to have/find the time to do a JS audit to see what our cross-bundle dependencies are. Maybe a good personal deliverable...

PR #5050, opened 2017-08-14
Update the list of official Twitter accounts

TheoChevalier:
Just noticed we were not promoting the official German and French Twitter handles, this is fixing it.
Filing a bug to get approval from Liz or Sierra


TheoChevalier:
Got greenlights from Liz: https://bugzilla.mozilla.org/show_bug.cgi?id=1390199#c1

PR #5051, opened 2017-08-15
[bug 1329642] Technology page, make Rust header permanent for English

craigcook:
## Description
The Rust video header got good results in the previous test so we can make it permanent. However, the video itself is only in English and there are no plans/budget to localize it. Thus other locales continue getting the old text header and only English gets the Rust header.

<del>This PR also removes the `emerging_tech_links` l10n tag since it looks like [we have good coverage for those strings](https://l10n.mozilla-community.org/langchecker/?locale=all&website=0&file=mozorg/technology.lang).</del>

## Issue / Bugzilla link


flodolo:
I'm fine with removing the tag (no need to update the PR), but note that good coverage is not really the criteria: if a page is activated, but doesn't have the tag, we're going to show English content, so we should deactivate the page.

craigcook:
I was perhaps a bit too hasty in removing the tag. It looks like those strings were only added a few weeks ago, but I thought they were several months old. When I looked at coverage (number of locales) I had that assumption set in my mind, and figured if those locales still hadn't updated after six months they likely never would and it was better to deactivate them.

So I think it may be better to put the tag back and give the remaining locales a bit more time to update. It's my own fault for assuming without verifying. Sorry for being overeager.

flodolo:
> So I think it may be better to put the tag back and give the remaining locales a bit more time to update. It's my own fault for assuming without verifying. Sorry for being overeager.

OK, at that point it shouldn't be tagged L10N anymore, should it?

craigcook:
> OK, at that point it shouldn't be tagged L10N anymore, should it?

Correct, I'm replacing the tags in the templates and will remove the l10n label. This PR should have no l10n impact. Sorry!

bedrock/mozorg/templates/mozorg/technology.html
```diff
@@ -206,7 +200,7 @@
 {% block email_form %}{% endblock %}
 
 {% block js %}
-  {% if not l10n_has_tag('emerging_tech_links') %}
+  {% if LANG.startswith('en-') or not l10n_has_tag('emerging_tech_links') %}
```
jpetto: I see that this works, but it looks kind of backwards. If the visitor's locale does *not* have the L10n tag `emergine_tech_links`, they get the video JS? It's tough to understand why/how this tag affects including JS. Maybe add a comment to clarify?

bedrock/mozorg/views.py
```diff
@@ -261,13 +260,21 @@ def contribute_friends(request):
...
 
+    def get_template_names(self):
+        locale = l10n_utils.get_locale(self.request)
+
+        if locale.startswith('en-'):
+            template_name = 'mozorg/technology-en.html'
+        else:
+            template_name = 'mozorg/technology.html'
+
+        return [template_name]
```
jpetto: Do you think we should add a couple simple tests for this view logic?

bedrock/mozorg/templates/mozorg/technology.html
craigcook: There's a button to the play video in the fallback content for locales without the tag and we only need the JS for the video. I can add a comment to clarify.

However, the video is in English so I would be happy to remove that video button for all locales, regardless of tag. That would mean the fallback content would have no CTA at all for those few remaining locales but maybe we can live with that?

bedrock/mozorg/views.py
craigcook: Good call. I shall make it so.

bedrock/mozorg/templates/mozorg/technology.html
craigcook: I think instead of the video button for the fallback, we can have the same Rust link as the CTA and use the existing "Learn more" string.

jpetto: Agreed.

PR #5057, opened 2017-08-18
[fix bug 1386112] Open Fx Hub to all locales.

jpetto:
- Decommission /firefox/family

## Description

As the `/firefox/` pages now have sufficient L10n coverage, we can open up the hub to all locales and drop the old templates for the following URLs:

- `/firefox/android/`
- `/firefox/android/faq/` (redirecting to SUMO)
- `/firefox/developer/`
- `/firefox/features/`
- `/firefox/ios/`
- `/firefox/products/`

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1386112

## Testing

We had quite a few pages pointing to `/firefox/products/` - make sure I haven't missed anything or made an inappropriate link to `/firefox/`.

Make sure there's not a use/user case that would require us to keep any of the deleted files.

Also make sure all assets/code from removed templates have been banished.

https://www-demo4.allizom.org/


flodolo:
I thought we would be using that bug also for getting rid of old templates like 
firefox/templates/firefox/ios.html (replaced by firefox/products/ios.html)?

jpetto:
> I thought we would be using that bug also for getting rid of old templates like
> firefox/templates/firefox/ios.html (replaced by firefox/products/ios.html)?

Good call. Will update this PR.

bedrock/firefox/tests/test_views.py
```diff
@@ -654,66 +654,66 @@ def test_get_context_data_one_star(self):
 @override_settings(DEV=False)
 @patch('bedrock.firefox.views.l10n_utils.render')
 class TestFirefoxFeatures(TestCase):
```
alexgibson: Is this test needed now that all locales show the same template?

alexgibson: Same question for the other tests below

bedrock/firefox/views.py
```diff
@@ -613,14 +612,7 @@ def ios_testflight(request):
 
 
 def features_landing(request):
```
alexgibson: Does this URL need to be a view still, could it just be a regular page defined in the URLs file?

bedrock/firefox/views.py
```diff
@@ -678,14 +670,7 @@ class FirefoxProductDesktopView(BlogPostsView):
     blog_tags = ['browser', 'featured']
 
     def get_template_names(self):
```
alexgibson: Does this method still need to be used here? Could you just do `template_name = 'firefox/products/desktop.html'`?

alexgibson: Same question for other method names below

bedrock/firefox/tests/test_views.py
```diff
@@ -654,66 +654,66 @@ def test_get_context_data_one_star(self):
...
         view.request.locale = 'en-US'
         eq_(view.get_template_names(), ['firefox/products/desktop.html'])
 
     @patch.object(views, 'lang_file_is_active', lambda *x: False)
-    def test_old_template(self, render_mock):
+    def test_non_en_us(self, render_mock):
         view = views.FirefoxProductDesktopView()
         view.request = RequestFactory().get('/firefox/desktop/')
         view.request.locale = 'fr'
-        eq_(view.get_template_names(), ['firefox/desktop/index.html'])
```
alexgibson: Should this old `'firefox/desktop/index.html'` template be removed too? 

It also has several old sub-pages that still exist, such as `/desktop/fast/` etc. Could/should we get rid of those too?

bedrock/firefox/urls.py
```diff
@@ -84,9 +84,6 @@
     page('firefox/private-browsing', 'firefox/private-browsing.html', alt_canonical_paths={
```
alexgibson: Seeing as we're removing most of the other legacy pages in this PR, can `firefox/private-browsing` be retired as well?

bedrock/firefox/tests/test_views.py
alexgibson: Incidentally, if we do remove those pages - would we also be free to remove the old family nav itself?

jpetto: This is exactly the point where I said to myself "let me just file another bug and submit another PR". I'm pretty sure we can remove those, but I wanted to stop blowing up the scope of this request (already went from 37 files to 112 files). I'll file a new bug for further cleanup.

bedrock/firefox/urls.py
jpetto: I'll make this part of the follow-up bug/PR. 👍

bedrock/base/templates/includes/global-nav.html
```diff
@@ -75,19 +70,17 @@ <h3 class="summary" data-id="firefox">
               <ul class="nav-menu-secondary-links">
                 <li>
                 {% if LANG == 'en-US' %}
```
jpetto: I debated this when I reviewed the PR that introduced it. My reasoning was that since the strings were already wrapped, and we (at the time, anyway) seemed to be going in the direction of showing the global nav to all, I figured it was appropriate future-proofing.

I think leaving it in place for now makes sense.

bedrock/firefox/views.py
jpetto: Can definitely clean these up.

bedrock/firefox/views.py
jpetto: Yeah, I'll move to a regular ol `page` in `urls.py`.

bedrock/firefox/tests/test_views.py
jpetto: I was going to leave them in as safeguards against future changes to the view, but since I'm going to make these just `page` directives, I don't see a reason to keep the tests.

jpetto:
Yep, as far as I know, we're good to push this out. bedrock feels lighter already.

peiying2:
@jpetto to answer your question on IRC, all have reached 95%+ except for developer.lang at 93%.

PR #5060, opened 2017-08-21
added meta info

vschs007:
added meta info

##


alexgibson:
Thanks for this pull request.

The meta keywords tag is no longer used by Google, and is rarely used by other search engines. When it is used, it doesn't carry much weight. This PR adds the meta keywords tag to the home page only, and with a single keyword 'mozorg', which is unlikely to be a term regular users search for.

Because of the above I'm going to close this pull request.

PR #5062, opened 2017-08-22
[fix bug 1386622] Update /new strings for advanced installs.

jpetto:
## Description

Updates link messaging for advanced install options and other languages.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1386622#c27

## Testing

Make sure new UX is sane, and old UX (for un-translated locales) isn't broken.

https://www-demo5.allizom.org/firefox/new/

### DIFF
--- a/bedrock/firefox/templates/firefox/includes/download-list.html
+++ b/bedrock/firefox/templates/firefox/includes/download-list.html
@@ -13,5 +13,3 @@
     </li>
   {%- endfor %}
 </ul>
-
-<p><a href="{{ firefox_url('desktop', 'all', 'release') }}">{{_('Available Languages') }}</a></p>
--- a/bedrock/firefox/templates/firefox/new/onboarding/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/onboarding/scene1.html
@@ -85,7 +85,27 @@
                 <li><a class="more" href="{{ url('firefox.family.index') }}">{{_('Discover all the ways to take Firefox on the go')}}</a></li>
               </ul>
             </div>
-            <button id="other-platforms-modal-link" class="hidden" type="button">{{_('Firefox for Other Platforms & Languages')}}</button>
+
+            <div id="other-platforms-languages-wrapper" class="hidden">
+              <button id="other-platforms-modal-link" type="button">
+              {% if l10n_has_tag('firefox_new_install_options') %}
+                {{ _('Advanced install options & other platforms') }}
+              {% else %}
+                {{ _('Firefox for other platforms') }}
+              {% endif %}
+              </button>
+
+              <p>
+                <a href="{{ url('firefox.all') }}">
+                {% if l10n_has_tag('firefox_new_install_options') %}
+                  {{ _('Other languages') }}
+                {% else %}
+                  {{ _('Available Languages') }}
+                {% endif %}
+                </a>
+              </p>
+            </div>
+
             <section id="benefits">
               <ul>
                 <li>
@@ -118,7 +138,13 @@
   <div id="other-platforms">
     <div class="container">
       <section class="section-other-platforms">
-        <h4>{{_('Firefox for other platforms') }}</h4>
+        <h4>
+        {% if l10n_has_tag('firefox_new_install_options') %}
+          {{ _('Advanced Install Options & Other Platforms') }}
+        {% else %}
+          {{_('Firefox for other platforms') }}
+        {% endif %}
+        </h4>
 
         {{ download_firefox_desktop_list() }}
 
--- a/bedrock/firefox/templates/firefox/new/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/scene1.html
@@ -105,9 +105,11 @@
                 <li><a class="more" href="{{ url('firefox.family.index') }}">{{_('Discover all the ways to take Firefox on the go')}}</a></li>
               </ul>
             </div>
-            {% if l10n_has_tag('firefox_new_other_platforms') %}
-              <button id="other-platforms-modal-link" class="hidden" type="button">{{_('Firefox for Other Platforms & Languages')}}</button>
-            {% endif %}
+
+            <div id="other-platforms-languages-wrapper" class="hidden">
+              <button id="other-platforms-modal-link" type="button">{{_('Firefox for Other Platforms & Languages')}}</button>
+            </div>
+
           </div>{#-- /.main-content --#}
         </div>{#-- /.inner-container --#}
       </div>
@@ -138,7 +140,6 @@
     </aside>
   </div>
 
-  {% if l10n_has_tag('firefox_new_other_platforms') %}
   <div id="other-platforms">
     <div class="container">
       <section class="section-other-platforms">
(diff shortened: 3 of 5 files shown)


jpetto:
@flodolo Any concerns from the L10n side on this one?

flodolo:
Not sure about the changes to `templates/firefox/new/scene1.html` (that maps to firefox/new/horizon).

We're deprecating that file in favor of firefox/new/onboarding, en-US is not using it, we shouldn't really be adding new strings to it.

jpetto:
Hm, good point on `firefox/new/scene1.html`. I thought to update it as it's still potentially being served, but only to locales that don't have `firefox/new/onboarding/scene1.html` translated. It makes sense that no more effort should be expended adding strings to the old template.

Sounds like I should revert `firefox/new/scene1.html`.

flodolo:
It looks like this PR doesn't apply to master?

bedrock/firefox/templates/firefox/new/onboarding/scene1.html
```diff
@@ -117,10 +131,20 @@ <h2 class="forward">{{_('Go Forward') }}</h2>
...
-      <section class="section-other-platforms">
-        <h4>{{_('Firefox for other platforms') }}</h4>
+      {#
+        when l10n tag 'firefox_new_install_options' is removed, remove
+        'no-lang-link' class below and from the CSS file
+      #}
+      <section class="section-other-platforms{% if l10n_has_tag('firefox_new_install_options') %} no-lang-link{% endif %}">
+        <h4>
+        {% if l10n_has_tag('firefox_new_install_options') %}
+          {{ _('Advanced Install Options & Other Platforms') }}
```
flodolo: Correction: we might have a problem. 

The same string is used both for the link and for the title. Sadly that's a known limitation of lang file (string is the ID).

Any chance that we can differentiate them? 

P.S. is it just me or title case is normally not used in links?

bedrock/firefox/templates/firefox/includes/download-list.html
```diff
@@ -14,4 +14,5 @@
   {%- endfor %}
 </ul>
 
-<p><a href="{{ firefox_url('desktop', 'all', 'release') }}">{{_('Available Languages') }}</a></p>
+{# The <p> below should be removed when l10n tag 'firefox_new_install_options' is removed #}
```
alexgibson: Instead of using CSS & l10n tag trickery to conditionally show this link, there might be a simpler option. It looks like the current template has both of these strings already translated:

- `Firefox for other platforms`
- `Available Languages`

Could we use these in the fallback instead, removing the link from inside the modal altogether and display two links one above the other? Might that be simpler?

bedrock/firefox/templates/firefox/new/onboarding/scene1.html
```diff
@@ -85,7 +85,21 @@
                 <li><a class="more" href="{{ url('firefox.family.index') }}">{{_('Discover all the ways to take Firefox on the go')}}</a></li>
               </ul>
             </div>
-            <button id="other-platforms-modal-link" class="hidden" type="button">{{_('Firefox for Other Platforms & Languages')}}</button>
+
+            <div id="other-platforms-languages-wrapper" class="hidden">
+            {% if l10n_has_tag('firefox_new_install_options') %}
+              <button id="other-platforms-modal-link" type="button">
```
alexgibson: This is a nit and not something I expect to be solved in this PR (I'm not sure of the right answer), but from a UX perspective it feels a bit weird for the first link to open a modal and then one directly underneath act like a normal link. It feels like this element should be styled more like a button. Maybe something for Rachel to think about when the redesign happens?

bedrock/firefox/templates/firefox/new/onboarding/scene1.html
```diff
@@ -117,10 +131,20 @@ <h2 class="forward">{{_('Go Forward') }}</h2>
...
+        <h4>
+        {% if l10n_has_tag('firefox_new_install_options') %}
+          {{ _('Advanced Install Options & Other Platforms') }}
+        {% else %}
+          {{_('Firefox for other platforms') }}
+        {% endif %}
+        </h4>
 
-        {{ download_firefox_desktop_list() }}
+        {{ download_firefox_desktop_list('release') }}
```
alexgibson: Not sure if this is required as it should be the default value anyway, but should this be `channel='release'`?

bedrock/firefox/templates/firefox/includes/download-list.html
jpetto: I like this approach. Definitely less complex.

bedrock/firefox/templates/firefox/new/onboarding/scene1.html
jpetto: Totally agree. There's UX needed here now. Is a modal really best? What if it was just a section you could scroll/anchor link to? We'd have more room for context and design...

bedrock/firefox/templates/firefox/new/onboarding/scene1.html
jpetto: Ah, this was an artifact I meant to remove. And yes, should be a keyword argument if provided.

bedrock/firefox/templates/firefox/new/onboarding/scene1.html
jpetto: Sorry, I'm not following. Are you saying that the string in the `<h4>` cannot be the same as the string used in the `<button>`?

flodolo: Ideally, yes. For example: what if you want to use a different case for the two contexts (e.g. all lowercase in the button)?

bedrock/firefox/templates/firefox/new/onboarding/scene1.html
```diff
@@ -85,7 +85,27 @@
...
+                {{ _('Advanced install options & other platforms') }}
+              {% else %}
+                {{ _('Firefox for other platforms') }}
+              {% endif %}
+              </button>
+
+              <p>
+                <a href="{{ url('firefox.all') }}">
+                {% if l10n_has_tag('firefox_new_install_options') %}
+                  {{ _('Other Languages') }}
```
alexgibson: Should we make this link sentance case to match the one above?

bedrock/firefox/templates/firefox/new/onboarding/scene1.html
```diff
@@ -85,7 +85,27 @@
...
+            <div id="other-platforms-languages-wrapper" class="hidden">
+              <button id="other-platforms-modal-link" type="button">
+              {% if l10n_has_tag('firefox_new_install_options') %}
+                {{ _('Advanced install options & other platforms') }}
+              {% else %}
+                {{ _('Firefox for other platforms') }}
+              {% endif %}
+              </button>
+
+              <p>
```
alexgibson: Could do with a teeny bit more breathing room between links?

screen shot 2017-08-24 at 18 18 44

jpetto:
@alexgibson - Think I've got all feedback addressed.

@flodolo - How are things looking from an L10n standpoint?

PR #5065, opened 2017-08-24
Add docs for JS lazy-load and notification libs

alexgibson:
## Description
- Adds docs for both `mozilla-lazy-load.js` and `mozilla-notification-banner.js`

## Issue / Bugzilla link
None

## Testing
- Check for typos, make sure docs compile ok?

### DIFF
--- a/docs/mozilla-lazy-load.rst
+++ b/docs/mozilla-lazy-load.rst
@@ -1,0 +1,76 @@
+.. This Source Code Form is subject to the terms of the Mozilla Public
+.. License, v. 2.0. If a copy of the MPL was not distributed with this
+.. file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+.. _mozillalazyload:
+
+=========================
+Mozilla Image Lazy Loader
+=========================
+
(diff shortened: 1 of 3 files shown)


alexgibson:
@jpetto by no means urgent, but a proof-read would be appreciated when you find some time?

docs/mozilla-lazy-load.rst
```diff
@@ -0,0 +1,69 @@
...
+    Mozilla.LazyLoad.init();
+
+Options
+~~~~~~~
+
+If you don't want to use the macro provided above and instead use your own HTML, you can pass an alternate custom CSS selecter::
+
+    Mozilla.LazyLoad.init('.my-own-lazy-loading-container');
+
+This will instantiate the plugin to observe intersection of elements using the DOM elements matching the selector provided.
```
jpetto: Might be nice to clarify *which* selector you are customizing here - it's not immediately obvious that it's on the wrapper element. Perhaps a code example?

alexgibson:
@jpetto thanks for the review, updated.

PR #5071, opened 2017-08-29
Fix bug 1394877: Use FIREFOX_DEVEDITION product-details key

pmac:
No longer use FIREFOX_AURORA product-details key.


pmac:
I have already deployed and [run integration tests against this](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/demo%2fpmac/21/pipeline/). Let's get this out quickly if we can.

bedrock/firefox/firefox_details.py
```diff
@@ -53,6 +54,7 @@ class FirefoxDesktop(_ProductDetails):
     version_map = {
         'nightly': 'FIREFOX_NIGHTLY',
         'alpha': 'FIREFOX_AURORA',
```
jpetto: Is there a reason to keep this key/value? Should the value now be `FIREFOX_DEVEDITION`?

pmac: I think so. Good catch.

pmac: Turns out I can't remove this as some code still references "alpha", but I did need to change it to reference `FIREFOX_DEVEDITION` instead of `FIREFOX_AURORA`.

pmac:
I've updated the items you mentioned in a new commit. I'll squarsh and merge when you give the 👍

thanks

PR #5072, opened 2017-08-29
Update product-details for code docker image

pmac:
Add p-d to the code image


pmac:
Some smoke tests (for the download pages) require up-to-date product-details. We should fix this, but for now this fixes deployment.

PR #5078, opened 2017-08-30
[fix bug 1392860] Fx Accounts membership experiments in /firstun

alexgibson:
## Description
- Adds a new `/firstrun` experiment page (en-US only) that introduces the idea of membership, whereby users self-identify their path of interest. Depending on the chosen path, users will receive newsletters via basket after signup for a Firefox Account.
- Clicking on a button to select a path passes a basket ID along to the Firefox Accounts server via a `utm_campaign` query param value.
- The full flow for this experiment has been [successfully tested and verified on demo](https://bugzilla.mozilla.org/show_bug.cgi?id=1392860#c55)

~**do not merge** for now, pending a final test plan~

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1392860

## Testing
- In order to test the Firefox Accounts signup form on the demo server, you will need to create a custom Firefox profile using these instructions: http://bedrock.readthedocs.io/en/latest/firefox-accounts.html#demo-server-testing
- Manually verify that the correct basket ID's are being passed along in the `utm_campaign` param when loading the FxA iframe signup form after clicking a button.

Demo: 
- https://www-demo4.allizom.org/en-US/firefox/55.0/firstrun/?v=a (control)
- https://www-demo4.allizom.org/en-US/firefox/55.0/firstrun/?v=b (membership)


alexgibson:
We're still waiting on final GA requirements, but I think this is now far along enough it should be good for code review.

bedrock/firefox/views.py
```diff
@@ -394,7 +394,7 @@ def get_template_names(self):
             template = 'firefox/dev-firstrun.html'
         elif show_40_firstrun(version):
             if locale == 'en-US' and exp in ['a', 'b']:
-                template = 'firefox/firstrun/index-{0}.html'.format(exp)
+                template = 'firefox/firstrun/membership-{0}.html'.format(exp)
```
alexgibson: Note: unit tests are currently failing due to this. The previous experiment is going to be stopped & removed prior to this merging, so I'm just leaving it like this for now.

media/js/firefox/firstrun/experiment-firstrun-membership.js
```diff
@@ -0,0 +1,17 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    var cop = new Mozilla.TrafficCop({
+        id: 'experiment_firstrun_membership',
+        variations: {
+            'v=a': 5,  // double-control
```
alexgibson: TODO: need to verify target audience sizes

alexgibson:
@schalkneethling - ok, this one is still do not merge for now until we verify a test plan, but it can be considered code complete.

media/js/firefox/firstrun/membership.js
```diff
@@ -0,0 +1,214 @@
...
+        for (var i = 0; i < button.length; i++) {
+            button[i].removeEventListener('mouseover', onButtonMouseOver, false);
+            button[i].removeEventListener('mouseout', onButtonMouseOut, false);
+            button[i].removeEventListener('click', onButtonClick, false);
+        }
+
+        laterLink.removeEventListener('click', onLaterLinkClick, false);
+    }
+
+    function animateElement(selector, keyframes, options) {
```
alexgibson: This is my first forray into the Web Animations API. This is just for a single experiment right now, but I may refactor some of what's here into a nice little JS animation library in the future (as some form of personal deliverable)

media/js/firefox/firstrun/membership.js
```diff
@@ -0,0 +1,214 @@
...
+            var keyframes = [
+                { opacity: 0 },
+                { opacity: 1 }
+            ];
+
+            var options = {
+                duration: 200,
+                fill: 'forwards',
+                easing: 'ease-in',
+            };
```
schalkneethling: I'm wondering, the `keyframes` and `options` here are almost identical to those in `fadeOutStep1`. I would suggest that you have a config object as follows:

```
var config = {
    keyframes: [
        { opacity: 0 },
        { opacity: 1 }
    ],
    options: {
        duration: 200,
        fill: 'forwards',
        easing: 'ease-out',
    }
}
```

media/js/firefox/firstrun/membership.js
```diff
@@ -0,0 +1,214 @@
...
+
+            step2.classList.remove('hidden');
+
+            animateElement('.step-2', keyframes, options).then(function() {
```
schalkneethling: Then here you can do:

```
config.options.easing = 'ease-in';
animateElement('.step-2', config.keyframes, config.options).then(function() {
...
```

media/js/firefox/firstrun/membership.js
```diff
@@ -0,0 +1,214 @@
...
+                { opacity: 0 }
+            ];
+
+            var options = {
+                duration: 200,
+                fill: 'forwards',
+                easing: 'ease-out',
+            };
+
+            animateElement('.step-1', keyframes, options).then(function() {
```
schalkneethling: Here you can simply:

```
animateElement('.step-2', config.keyframes, config.options).then(function() {
...
```

media/js/firefox/firstrun/membership.js
```diff
@@ -0,0 +1,214 @@
...
+
+        laterLink.removeEventListener('click', onLaterLinkClick, false);
+    }
+
+    function animateElement(selector, keyframes, options) {
+        return new Promise(function(resolve, reject) {
+            var element = document.querySelector(selector);
+
+            if (element) {
+                var animation = element.animate(keyframes, options);
```
schalkneethling: Would it be worth wrapping the calls to animate inside a `requestAnimationFrame`?

bedrock/firefox/templates/firefox/firstrun/membership-b.html
```diff
@@ -0,0 +1,81 @@
...
+
+{% block page_title %}{{ _('Welcome to Firefox') }}{% endblock %}
+
+{% block site_css %}
+  {% stylesheet 'firefox_firstrun_membership' %}
+{% endblock %}
+
+{% block js %}
+  {% javascript 'firefox_firstrun_membership' %}
+{% endblock %}
```
schalkneethling: Is this placed here intentionally to be blocking?

bedrock/firefox/tests/test_base.py
```diff
@@ -360,12 +360,41 @@ def test_fx_firstrun_56_0(self, render_mock):
 
     @override_settings(DEV=True)
     def test_fx_firstrun_57_0(self, render_mock):
-        """Should use 57 quantum firstrun template"""
+        """Should use 57 quantum dev edition firstrun template"""
```
schalkneethling: Thanks for correcting this 👍

media/js/firefox/firstrun/membership.js
alexgibson: The neat thing about the Web Animations API is the browser can make it's own optimizations without the need for things such as `requestAnimationFrame`. There's a great intro to it on MDN: https://developer.mozilla.org/docs/Web/API/Web_Animations_API/Using_the_Web_Animations_API

bedrock/firefox/templates/firefox/firstrun/membership-b.html
alexgibson: This isn't blocking per-say. The place where it appears in the source order is defined in the base template (at the bottom before closing body tag). Saying that, I'll move it to the bottom of this template too, just to avoid confusion.

media/js/firefox/firstrun/membership.js
alexgibson: Well, they are similar but not quite. The keyframes are reversed, and the easing also changes. I could use a shared config object and then override things where needed, but I'm not sure this would make the code any easier to follow. Thoughts?

bedrock/settings/static_media.py
```diff
@@ -361,6 +361,14 @@
         ),
         'output_filename': 'css/firefox_firstrun-bundle.css',
     },
+    'firefox_firstrun_membership': {
+        'source_filenames': (
+            'css/base/notification-banner.less',
```
schalkneethling: Is the CSS here related to the notification here? https://github.com/mozilla/bedrock/pull/5078/files#diff-3d3fd080659e5ac82e85416beb285e4eR80 - If so, do you still need the CSS then?

alexgibson: Nice catch 👍

media/js/firefox/firstrun/membership.js
schalkneethling: Ah, I was wondering whether it was doing this internally. 👍

media/js/firefox/firstrun/membership.js
schalkneethling: True. I still like the idea of a shared `config` object, it’s also a few lines less code ;) I do not have strong feelings on this so, I would say, either way is fine.

bedrock/firefox/templates/firefox/firstrun/membership-b.html
schalkneethling: Ah ok. I was wondering whether the ordering mattered in Jinja or whether you can "implement" these blocks in any order and they will still be stuck into the normal source location.

I do agree though that it would look less confusing if done at the end of the file. 👍

media/css/firefox/firstrun/membership.less
```diff
@@ -0,0 +1,258 @@
...
+    span {
+        border-radius: 50%;
+        border: 3px solid #fff;
+        display: inline-block;
+        height: 13px;
+        margin: 0 5px;
+        overflow: hidden;
+        text-indent: -999em;
+        transition: background-color .2s ease-in-out;
+        width: 13px;
```
schalkneethling: Nit: probably worth adding `cursor: pointer;`?

media/js/firefox/firstrun/membership.js
```diff
@@ -0,0 +1,214 @@
...
+            });
+        }
+    }
+
+    // replaces utm_campaign param in FxA iframe data-src with data-id value from button click.
+    function updateUTMParam(id) {
+        var utmCampaign = id ? encodeURIComponent(id) : 'fxa-embedded-form';
+        var iframe = document.getElementById('fxa');
+        var dataSrc = iframe.getAttribute('data-src');
+        iframe.setAttribute('data-src', dataSrc.replace('utm_campaign=fxa-embedded-form', 'utm_campaign=' + utmCampaign));
```
schalkneethling: As this is going to run in recent versions of Firefox only, any reason not to do the above as follows:

```
var dataSrc = iframe.dataset.src
iframe.dataset.src = dataSrc.replace('utm_campaign=fxa-embedded-form', 'utm_campaign=' + utmCampaign);
```

media/css/firefox/firstrun/membership.less
```diff
@@ -0,0 +1,258 @@
...
+            background: transparent;
+            border-radius: 20px;
+            border: 3px solid @colorPhotonBlue;
+            color: @colorPhotonBlue;
+            font-weight: bold;
+            margin: 20px 0;
+            padding: 12px 8px;
+            text-transform: uppercase;
+            width: 100%;
+            transition: background .2s ease-in-out, color .2s ease-in-out;
```
schalkneethling: Nit: probably worth adding cursor: pointer;?

bedrock/firefox/templates/firefox/firstrun/membership-b.html
```diff
@@ -0,0 +1,81 @@
...
+
+    <section class="path-selector">
+      <h2>{{ _('Pick your Firefox path') }}</h2>
+      <p>{{ _('Get the Firefox that suits you best.') }}</p>
+      <ul>
+        <li><button type="button" data-index="0" data-id="membership-idealo" data-label="Independence">{{ _('Independence') }}</button></li>
+        <li><button type="button" data-index="1" data-id="membership-comm" data-label="Privacy">{{ _('Privacy') }}</button></li>
+        <li><button type="button" data-index="2" data-id="membership-tech" data-label="Technology">{{ _('Technology') }}</button></li>
+      </ul>
+      <a href="{{ url('newsletter.mozilla') }}" class="later">{{ _('Remind me later') }}</a>
```
schalkneethling: So, I know this is supposed to go to `/en-US/newsletter/mozilla/` but, currently when you click the link and does nothing, and stays on the page. It is probably because of the `event.preventDefault()` here https://github.com/mozilla/bedrock/pull/5078/files#diff-a5df29ee5206cbcc954abe5b3291fe96R123

media/js/firefox/firstrun/membership.js
alexgibson: Yeah it does save a couple of lines of code, but one thing I'm trying to do these days is avoid writing functions that produce side-effects (i.e. altering the config object). This is relatively safe as there would be only two functions that consume it, but if this page we're more complicated it could lead to possible errors. I'm going to leave it be for now 👍

bedrock/firefox/templates/firefox/firstrun/membership-b.html
alexgibson: Doh, nice catch!

alexgibson:
Thanks for the review @schalkneethling! I added a commit with some review fixes.

alexgibson:
Ok, updated with the final audience segments as [confirmed in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1392860#c59). This one should be good to merge, thanks again @schalkneethling

PR #5081, opened 2017-08-31
[fix bug 1393132] Remove old /firefox/new/ and /firefox/firstrun/ templates

alexgibson:
## Description
- Removes legacy `/new` and `/firstrun` templates, reorganizing & tidying up both directories.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1393132

## Testing
- Run with a clean `/static` cache via `gulp static:clean`
- Make sure both `/new` and `/firstrun` pages load and work as expected.
- Make sure the copy A/B test on `/firstrun` still works as expected.
- Check to make sure I didn't delete anyhting that was still needed, or miss something that can be removed?

Demo:
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/55.0/firstrun/
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/55.0/firstrun/?v=a
- https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/55.0/firstrun/?v=b


media/js/firefox/onboarding/user-actions.js
```diff
@@ -1,66 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
-* License, v. 2.0. If a copy of the MPL was not distributed with this
-* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
-(function(Mozilla, $) {
```
alexgibson: As far as I can tell, this JS file seems orphaned from some past experiment

alexgibson:
OK, I think this is ready for review. The diff looks worse than it is (mostly just renaming and moving files), but this could still do with a second pair of eyes.

bedrock/firefox/templates/firefox/new/scene2.html
```diff
@@ -29,29 +29,18 @@
...
             <a href="{{ url('mozorg.home') }}" data-link-type="nav" data-link-name="tabzilla">Mozilla</a>
           </div>
-          <h2>
-          {% if LANG == 'en-US' %}
-            <a href="{{ url('firefox') }}">{{ high_res_img('firefox/template/header-logo-inverse.png', {'alt': 'Firefox', 'width': '185', 'height': '70'}) }}</a>
-          {% else %}
-            {{ high_res_img('firefox/template/header-logo-inverse.png', {'alt': 'Firefox', 'width': '185', 'height': '70'}) }}
-          {% endif %}
-          </h2>
+          <h2>{{ high_res_img('firefox/template/header-logo-inverse.png', {'alt': 'Firefox', 'width': '185', 'height': '70'}) }}</h2>
```
jpetto: Same question on dropping the link. Just making sure it's intended.

bedrock/firefox/templates/firefox/firstrun/index.html
```diff
@@ -13,7 +13,7 @@
 {% endblock %}
 
 {% block site_css %}
-  {% stylesheet 'firefox_firstrun_onboarding' %}
+  {% stylesheet 'firefox_firstrun' %}
 {% endblock %}
```
jpetto: I see one block didn't get copied over *coughoptimizelycough*. I'm not mad about it, however it does still exist in the `new/base.html` template. Might be worth putting in `firstrun/base.html` for future requests?

jpetto: Not directly related to this PR, but wondering if we should implement the open graph stuff (`og:url` in particular) from the old `firstrun-horizon.html` template over to `firstrun/base.html`?

bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -51,31 +45,17 @@
...
               <div class="download-button-wrapper desktop" id="download-button-wrapper-desktop">
-                {# **WARNING**
-
-                  The `locale_in_transition` parameter should be used very carefully. It's included
-                  here because this template and scene2 share a lang file, and are therefore
-                  translated into the same list of locales. Adding this to another download button
-                  would significantly restrict the builds of Firefox available for download.
-
-                  Bug 1290962
-                #}
```
jpetto: Is this warning no longer valid?

bedrock/settings/static_media.py
```diff
@@ -1211,18 +1176,11 @@
     'firefox_firstrun': {
         'source_filenames': (
             'js/base/mozilla-fxa-iframe.js',
+            'js/base/uitour-lib.js',
```
jpetto: I assume this is necessary, but I'm missing what for.

bedrock/firefox/templates/firefox/new/base.html
```diff
@@ -4,23 +4,17 @@
 
 {% from "macros.html" import google_play_button with context %}
 
-{% add_lang_files "firefox/new/horizon" %}
+{% add_lang_files "firefox/new/onboarding" %}
```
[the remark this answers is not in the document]
flodolo: The diff looks very confusing, but firefox/new/horizon is being removed, while firefox/new/onboarding moves to firefox/new

Since both templates in onboarding where referencing an extra lang file (firefox/new/onboarding), this move is transparent to us.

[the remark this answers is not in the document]
alexgibson: This should all be good, you can verify by changing locales for each page on the demo (or locally, if you run sync_all)

bedrock/firefox/templates/firefox/new/base.html
```diff
@@ -4,23 +4,17 @@
...
 {% block page_title_prefix %}{{_('Download Firefox')}} — {% endblock %}
 {% block page_title %}{{_('Free Web Browser')}}{% endblock %}
-{% block page_desc %}
-{% if l10n_has_tag('macos_branding') %}
-  {{ _('Download Mozilla Firefox, a free Web browser. Firefox is created by a global non-profit dedicated to putting individuals in control online. Get Firefox for Windows, macOS, Linux, Android and iOS today!') }}
-{% else %}
-  {{ _('Download Mozilla Firefox, a free Web browser. Firefox is created by a global non-profit dedicated to putting individuals in control online. Get Firefox for Windows, OS X, Linux, Android and iOS today!') }}
-{% endif %}
-{% endblock %}
+{% block page_desc %}{{_('Download Mozilla Firefox, a free Web browser. Firefox is created by a global non-profit dedicated to putting individuals in control online. Get Firefox for Windows, Mac OS X, Linux, Android and iOS today!')}}{% endblock %}
```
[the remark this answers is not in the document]
flodolo: Ugh, looks like [this was never updated](https://transvision.mozfr.org/?recherche=Windows%2C+Mac+OS+X%2C+Linux%2C+Android+and+iOS+today&repo=mozilla_org&sourcelocale=en-US&locale=it&search_type=strings_entities) (it it used?). I would leave it out of this PR anyway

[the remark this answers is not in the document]
alexgibson: Sigh, it looks like the macOS renaming missed this file originally. I'll update here with a new l10n tag /cc @flodolo

bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -30,13 +30,7 @@
...
               <a href="{{ url('mozorg.home') }}" data-link-type="nav" data-link-name="tabzilla">Mozilla</a>
             </div>
-            <h2>
-            {% if LANG == 'en-US' %}
-              <a href="{{ url('firefox') }}">{{ high_res_img('firefox/template/header-logo-inverse.png', {'alt': 'Firefox', 'width': '185', 'height': '70'}) }}</a>
-            {% else %}
-              {{ high_res_img('firefox/template/header-logo-inverse.png', {'alt': 'Firefox', 'width': '185', 'height': '70'}) }}
-            {% endif %}
-            </h2>
+            <h2>{{ high_res_img('firefox/template/header-logo-inverse.png', {'alt': 'Firefox', 'width': '185', 'height': '70'}) }}</h2>
```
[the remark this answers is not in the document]
alexgibson: Good catch

bedrock/firefox/templates/firefox/new/scene1.html
alexgibson: It most ceetainly is valid, I'll re-add it to the new default template

bedrock/firefox/templates/firefox/new/scene2.html
alexgibson: Will update this one as well

bedrock/firefox/templates/firefox/new/base.html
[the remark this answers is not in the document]
flodolo: I would prefer to leave this one only for the old templates removal, to make things a little clearer.

bedrock/settings/static_media.py
alexgibson: Hmm, at one time it was needed for the end page transition, but now it goes to `about:home` it may not be needed. I'll check 👍

bedrock/firefox/templates/firefox/new/base.html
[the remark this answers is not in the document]
alexgibson: OK! 👍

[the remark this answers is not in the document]
flodolo: (having said that, if you want, we can still add the string behind a tag)

bedrock/firefox/templates/firefox/firstrun/index.html
alexgibson: I'll take a look at this 👍

bedrock/firefox/templates/firefox/new/base.html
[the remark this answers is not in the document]
alexgibson: Ok, I'll add a tag with the new string (it's a simple fix, and avoids creating another bug)

bedrock/firefox/templates/firefox/firstrun/index.html
alexgibson: I've added optimizely to `/firstrun/base.html`.

As for the open graph strings, I do agree it makes sense for them to belong in `firstrun/base.html`, so we don't have to duplicate them for every page. This does mean adding `{% add_lang_files "firefox/new/onboarding" %}` to the base template though, so I want to make sure @flodolo is ok with the change.

flodolo: Definitely not OK, we're removing that file. Please add the strings behind a tag, we'll import them if they're already translated in the old file.

alexgibson: Ok! 👍

bedrock/firefox/templates/firefox/firstrun/base.html
```diff
@@ -2,12 +2,42 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
 
+{% add_lang_files "firefox/new/onboarding" %}
+
 {% extends "firefox/base.html" %}
 
 {% block extra_meta %}<meta name="robots" content="noindex">{% endblock %}
 
+{% if l10n_has_tag('firstrun_meta_info') %}
```
alexgibson: I've added an l10n tag for the meta info here in `firstrun/base.html`

I admit, I'm a bit confused why we need this as the new default page is the only template inheriting from this file. It is possible that other firstrun templates will want to inherit from this in the future however, so I think adding these strings to a shared file might make more sense. Perhaps I'm over-thinking this though

bedrock/firefox/templates/firefox/new/base.html
```diff
@@ -4,23 +4,23 @@
...
 
 {% block page_title_prefix %}{{_('Download Firefox')}} — {% endblock %}
 {% block page_title %}{{_('Free Web Browser')}}{% endblock %}
 {% block page_desc %}
-{% if l10n_has_tag('macos_branding') %}
-  {{ _('Download Mozilla Firefox, a free Web browser. Firefox is created by a global non-profit dedicated to putting individuals in control online. Get Firefox for Windows, macOS, Linux, Android and iOS today!') }}
-{% else %}
-  {{ _('Download Mozilla Firefox, a free Web browser. Firefox is created by a global non-profit dedicated to putting individuals in control online. Get Firefox for Windows, OS X, Linux, Android and iOS today!') }}
-{% endif %}
+  {% if l10n_has_tag('macos_branding') %}
```
alexgibson: Added an l10n tag for the string change on `/new` here.

alexgibson:
@jpetto - thanks for the review, I think I should have addressed everything. We still need to clear up how to manage those strings in the base template for firstrun, but this should be ready for another look.

bedrock/firefox/templates/firefox/firstrun/base.html
```diff
@@ -6,8 +6,27 @@
...
```
alexgibson: Ok, for some reason (that I've not quite worked out yet), this l10n tag does not work here anyway. I'm still not quite clear on why we need it though, seeing as the new /firstrun template is the only translated template inheriting from it. Perhaps @flodolo can help explain the reasoning here for why we need it?

bedrock/firefox/templates/firefox/firstrun/index.html
flodolo: I should not comment while I have two feet out of the door (sorry, complicated period…).

`{% add_lang_files "firefox/new/onboarding" %}` is fine. I've thought that you were adding the old template (horizon).

alexgibson:
OK, l10n issues should now be resolved, this should be good to go once reviewed.

bedrock/firefox/templates/firefox/firstrun/base.html
```diff
@@ -2,12 +2,31 @@
...
+{% add_lang_files "firefox/new/onboarding" %}
+
 {% extends "firefox/base.html" %}
 
 {% block extra_meta %}<meta name="robots" content="noindex">{% endblock %}
 
+{% block page_title_prefix %}{% endblock %}
+{% block page_title %}{{ _('Welcome to Firefox') }}{% endblock %}
+{% block page_title_suffix %}{% endblock %}
+{% block page_desc %}{{_('Download Mozilla Firefox, a free Web browser. Firefox is created by a global non-profit dedicated to putting individuals in control online. Get Firefox for Windows, Mac OS X, Linux, Android and iOS today!')}}{% endblock %}
```
flodolo: I think we need to update this one too (behind the same tag)?

alexgibson: Good catch, updated

bedrock/firefox/templates/firefox/firstrun/base.html
```diff
@@ -2,12 +2,37 @@
...
+    {{_('Download Mozilla Firefox, a free Web browser. Firefox is created by a global non-profit dedicated to putting individuals in control online. Get Firefox for Windows, Mac OS X, Linux, Android and iOS today!')}}
+  {% endif %}
+{% endblock %}
+
+{#- This will appear as <meta property="og:title"> which can be used for social share -#}
+{% block page_og_title %}{{ _('Chart your own course') }}{% endblock %}
+
+{#- This will appear as <meta property="og:description"> which can be used for social share -#}
+{% block page_og_desc %}{{_('Firefox is non-profit, non-corporate, non-compromised. Choosing Firefox isn’t just choosing a browser. It’s a vote for personal freedom.')}}{% endblock %}
+
```
jpetto: Any reason to omit the `page_og_url` block here? Seems it would be good to have in the event someone tries to share this URL.

alexgibson: Ah, looks like the onboarding template omitted this - will add it back, nice spot

alexgibson:
@jpetto - thanks, updated with `page_og_url` and squashed

alexgibson:
Merging this, as it was approved above by @jpetto - thanks!

PR #5009, opened 2017-07-27
[fix bug 1383305] Add Fx55 WNP experiment.

jpetto:
## Description

Adds an experiment for locales *not* in the following list:

`('de', 'en-GB', 'en-US', 'en-ZA', 'es-ES', 'fr', 'id', 'ja', 'ja-JP-mac', 'pl', 'pt-BR', 'ru', 'zh-CN', 'zh-TW')`

`?v=a`: control
`?v=b`: QR code displayed, no app store badges displayed
`?v=c`: QR code displayed, app store badges moved up

Locales in the exclusion list should get the default behavior for that locale, carried over from WNP for Fx54.

This PR also ensures the country-specific store badge links work when the Send To Device widget is absent. (Previously this functionality piggy-backed on the Send To Device geolocation lookup.)

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1383305

## Testing

Make sure DNT is disabled to be entered in to the experiment. Can also hit the variations directly:

- https://bedrock-demo-jpetto.us-west.moz.works/it/firefox/55.0/whatsnew/?v=a (control)
- https://bedrock-demo-jpetto.us-west.moz.works/it/firefox/55.0/whatsnew/?v=b (only QR code)
- https://bedrock-demo-jpetto.us-west.moz.works/it/firefox/55.0/whatsnew/?v=c (QR code and store badges)

`?v=c` has custom adjust links, which look almost the same as the existing, except have `_v2` appended to the `&creative=` query param. The demo has a hack to spoof geolocation to test these  alternate links:

- https://bedrock-demo-jpetto.us-west.moz.works/it/firefox/55.0/whatsnew/?v=a&country=it (should see `&creative=badge_it` in store badge URL)
- https://bedrock-demo-jpetto.us-west.moz.works/it/firefox/55.0/whatsnew/?v=c&country=it (should see `&creative=badge_it_v2` in store badge URL)

PR #5085, opened 2017-09-01
[fix bug 1383305] Promote Fx55 WNP experiment winner.

jpetto:
## Description

Promotes the winning variation from the test begun in #5009.

For all locales that do not get the Send to Device widget a QR code will now be displayed instead of the app store badges.

Two questions pending in the bug:

1. should the locales with the Send to Device widget keep the app store badges or not?
2. which QR code image should be promoted?

The above shouldn't block review though - biggest change would be deleting HTML and swapping an image. (Hoping to get this out tomorrow, hence the semi-early review ask.)

**Update**

App store badges removed for all conditions. JS complexity drastically reduced!

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1383305#c20

## Testing

Make sure no regressions and no files missed that could be deleted.


flodolo:
Is this supposed to be localized?

alexgibson:
Let me know if you want me to take another look should there be any updates

flodolo:
> Is this supposed to be localized?

Never mind, it looks like it's just things moving around.

alexgibson:
> @alexgibson Can you take a look at the second commit here? Thanks!

Changes look good to me and locales seem to be doing the right thing still. Hooray for less code!

jpetto:
@alexgibson Squashed!

PR #5086, opened 2017-09-01
[bug 1395342] Update strings for send-to-device widget

craigcook:
## Description
Updates strings so we can begin supporting SMS in multiple countries. New strings are behind the tag `sendto_updates_bug1395342`

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1395342

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -290,7 +290,13 @@
       <h2 class="thank-you hidden">{{ _('Your download link was sent.') }}</h2>
       <form id="send-to-device-form" action="{{ url('firefox.send-to-device-post') }}" method="post"{% if platform == 'select' %} class="dropdown"{% endif %}>
         <ul class="error-list hidden">
-          <li class="sms">{{ _('Sorry. This number isn’t valid. Please enter a U.S. phone number.') }}</li>
+          <li class="sms">
+          {% if l10n_has_tag('sendto_updates_bug1395342') %}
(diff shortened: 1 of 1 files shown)


bedrock/base/templates/macros.html
```diff
@@ -290,7 +290,13 @@ <h2 class="form-heading">
       <h2 class="thank-you hidden">{{ _('Your download link was sent.') }}</h2>
       <form id="send-to-device-form" action="{{ url('firefox.send-to-device-post') }}" method="post"{% if platform == 'select' %} class="dropdown"{% endif %}>
         <ul class="error-list hidden">
-          <li class="sms">{{ _('Sorry. This number isn’t valid. Please enter a U.S. phone number.') }}</li>
+          <li class="sms">
+          {% if l10n_has_tag('sendto_updates_bug1395342') %}
```
alexgibson: Will an l10n tag work here, considering this is a shared macro?

flodolo: We're actually loading a file in the templates using sendto, so it should work (but worth testing)
https://github.com/mozilla/bedrock/search?utf8=%E2%9C%93&q=firefox%2Fsendto&type=

alexgibson: Thanks @flodolo - I tested this on the current `/whatsnew` page and it does seem to be working 👍

flodolo:
I've exposed strings, please warn if anything changes from now on ;-)

PR #5087, opened 2017-09-04
[fix bug 1392680] Enable /new/?xv=fast for l10n to support German ad campaign

alexgibson:
## Description
- Enables `firefox/new/?xv=fast` for translation
- Checks for lang file `'firefox/new/sem'`

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1392680

## Testing
- Check that this deck of cards is still standing (thankfully we have some tests)

### DIFF
--- a/bedrock/firefox/templates/firefox/new/sem/base.html
+++ b/bedrock/firefox/templates/firefox/new/sem/base.html
@@ -2,6 +2,9 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{# Bug 1392680: (limited) localization of ?xv=fast variation only #}
+{% add_lang_files "firefox/new/sem" %}
+
 {% extends "firefox/base-resp.html" %}
 
 {% block extra_meta %}
@@ -9,8 +12,8 @@
 {% endblock %}
 
 {% block page_title_prefix %}{% endblock %}
-{% block page_title %}Free Web Browser for Android, iOS, Desktop | Firefox{% endblock %}
-{% block page_desc %}Millions of people around the world trust Firefox Web browsers on Android, iOS and desktop computers. Fast. Private. Download now!{% endblock %}
+{% block page_title %}{{ _('Free Web Browser for Android, iOS, Desktop') }} | {{ _('Firefox') }}{% endblock %}
+{% block page_desc %}{{ _('Millions of people around the world trust Firefox Web browsers on Android, iOS and desktop computers. Fast. Private. Download now!') }}{% endblock %}
 
 {% block page_css %}
   {% stylesheet 'firefox_new_sem' %}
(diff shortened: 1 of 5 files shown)


alexgibson:
@jpetto r?

This PR left me feeling a little uneasy. Even for a small change, it required way more re-jigging than I feel comfortable with. Thankfully we have some fairly comprehensive unit tests for all the different templates, but I think long term we need to find a new solution here (especially if translating for German is going to become a regular thing). I don't think it needs to necessarily block this PR, but I'd love to know your thoughts on a better solution. I have some ideas, but it's probably a chunk of work we'll not get time for until after 57.

flodolo:
This is only one step:
* Templates don't have a single string wrapped.
* They should both references a single lang file, likely ` firefox/new/fast.lang` at this point, following batm.

alexgibson:
> Templates don't have a single string wrapped.

Bah, I should have checked - thanks.

I'll update

alexgibson:
Ok, strings are now wrapped (lang file is `firefox/new/sem`)

flodolo:
Extracted files locally and everything seems to work, no hard coded strings. Please warn me if strings change for any reason before merge.

alexgibson:
> Please warn me if strings change for any reason before merge.

Will do, thanks @flodolo

bedrock/firefox/views.py
```diff
@@ -542,14 +544,18 @@ def new(request):
                 template = 'firefox/new/batm/scene2.html'
```
jpetto: Should be able to remove `batmprivate` from the list above (line 543) as it's now checked for above.

bedrock/firefox/views.py
```diff
@@ -509,7 +509,11 @@ def new(request):
     locale = l10n_utils.get_locale(request)
 
     if scene == '2':
-        if locale == 'en-US':
+        if lang_file_is_active('firefox/new/batm', locale) and experience == 'batmprivate':
+            template = 'firefox/new/batm/scene2.html'
+        elif lang_file_is_active('firefox/new/sem', locale) and experience == 'fast':
```
jpetto: Until we have a better solution, maybe add a comment that the above are only intended to be `en-US` and `de`? (I think that's also the case for `batmprivate`...)

bedrock/firefox/templates/firefox/new/sem/base.html
```diff
@@ -2,15 +2,17 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% add_lang_files "firefox/new/sem" %}
```
jpetto: Do you think it's worth a comment along with this lang file import detailing the limited localization? Might be good to know it's only `en-US` and `de` at a quick glance for when we forget the details next week. 😝

jpetto:
Agree 100% that we need a more elegant, scalable, and sane way of dealing with these views. I'll file an issue with my initial thoughts.

alexgibson:
@jpetto thanks for the review, I added a second commit with some fixes.

alexgibson:
Squashed!

PR #5089, opened 2017-09-04
Fix Bug 1396550 - Firefox for Android Nightly is pointing at obsolete api-15 folder

kyoshino:
## Description

The download link of Firefox Nightly for Android is broken due the an Android API update. Firefox 56+ requires Android 4.1 (API 16) instead of Android 4.0.3 (API 15). This PR will fix the issue. This kind of complexity will be solved once Android Nightly gets bouncer URLs :sweat_smile:

Updated the relevant CSS files as well, but I think it's no longer used since all the Android download buttons are currently pointing Google Play.

## Issue / Bugzilla link

[Bug 1396550](https://bugzilla.mozilla.org/show_bug.cgi?id=1396550)

## Testing

* Visit `/firefox/android/nightly/all/` and make sure the ARM download link works properly. The label should say "ARM devices (Android 4.1+)".
* Visit `/firefox/android/beta/all/`. The label should say "ARM devices (Android 4.1+)".
* Visit `/firefox/android/all/`. The label should still say "ARM devices (Android 4.0.3+)".

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -296,10 +296,10 @@ class FirefoxDesktop(_ProductDetails):
 
 
 class FirefoxAndroid(_ProductDetails):
-    # Architecture names defined in bouncer and these human-readable names
+    # Human-readable architecture names
     platform_labels = OrderedDict([
-        ('api-15', _('ARM devices\n(Android 4.0.3+)')),
-        ('x86', _('Intel devices\n(Android 4.0.3+ x86 CPU)')),
+        ('arm', _('ARM devices\n(Android %s+)')),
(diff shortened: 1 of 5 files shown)


flodolo:
> There are translated string changes however.

I can't spot any localized test in this PR?

kyoshino:
Um yes, `ARM devices\n(Android 4.0.3+) ` and `Intel devices\n(Android 4.0.3+ x86 CPU)` have been changed to `ARM devices\n(Android %s+)` and `Intel devices\n(Android %s+ x86 CPU)` respectively, but those are not exposed to l10n anyway. So l10n isn't a blocker.

pmac:
Cool. Those are the ones to which I was referring. If those aren't yet localized anyway then let's go. Thanks Kohei!

PR #5091, opened 2017-09-05
[fix bug 1391829] Add Dev Edition to Firefox secondary nav

alexgibson:
## Description
- Adds "Developer Edition" to secondary nav on Firefox
- ~en-US only for now, until we have enough locales translated with the new string.~

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1391829

### DIFF
--- a/bedrock/base/templates/includes/global-nav.html
+++ b/bedrock/base/templates/includes/global-nav.html
@@ -105,6 +105,11 @@
                   </a>
                 </li>
                 <li>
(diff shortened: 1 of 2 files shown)


flodolo:
We actually already have the string, same context too
https://github.com/mozilla-l10n/www.mozilla.org/blob/master/en-US/main.lang#L31-L33

alexgibson:
Ah, super - thanks

alexgibson:
> We should also add it to the sidebar/flyout/hamburger menu (sorry I've neglected this PR).

Bah, totally missed that - will update, thanks

alexgibson:
@craigcook updated

PR #5093, opened 2017-09-05
[fix bug 1366198] Switch Pebbles to fluid width by default

alexgibson:
## Description
This is an attempt to make all Pebbles pages 100% fluid width by default (as opposed to the Fx-hub specific overrides we're doing currently on a per-page basis). I figured it is best to try and re-pay this debt early, instead of making it even harder long-term.

**Note:** I didn't go to the full extent of making all existing pages 100% fluid width. Some pages contain fixed sections still (I decided it wasn't worth the effort), but they should still wrap nicely using our regular breakpoints. The newsletter component is also a rats nest of fixed width dependencies so I didn't touch it. I think it probably needs rewriting from scratch.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1366198

## Testing
- Test pebbles-base pages for responsive regressions. It may be worth running `gulp static:clean` before testing/reviewing.


alexgibson:
Pull #5092 should probably be merged before this one (I didn't test against the old IH template)

alexgibson:
Closing this so we can revisit at a later date

PR #5096, opened 2017-09-06
[fix bug 1395788] /whatsnew page for 56 release

alexgibson:
## Description
- Adds a new `/whatsnew` page for Firefox 56.
- 56 `/whatsnew` page should be shown when updating from FF < 56 -> 56.x.
- Older mobile promo `/whatsnew` should be shown when updating from 56.x -> 56.y

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1395788

## Testing
Demo: https://www-demo3.allizom.org/firefox/56.0/whatsnew/
- Test responsiveness, non-english locales, RTL.


bedrock/firefox/templates/firefox/whatsnew/base-pebbles.html
```diff
@@ -0,0 +1,32 @@
...
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% extends "firefox/base-pebbles.html" %}
+
+{% block extra_meta %}<meta name="robots" content="noindex">{% endblock %}
+
+{% block gtm_page_id %}data-gtm-page-id="/firefox/whatsnew/"{% endblock %}
+
+{#- This will appear as <meta property="og:title"> which can be used for social share -#}
+{% block page_og_title %}{{ _('A better, faster, private browser for today') }} | {{ _('Firefox') }}{% endblock %}
```
alexgibson: I took these strings from the `/firefox` page, as the meta sharing strings on older whatsnew pages are sorely out of date and no one ever gives us new ones. This template is what will hopefully be a new base template for all future `/whatsnew` pages, so it might be worth putting these strings into a shared lang file?

flodolo: My understanding is that we're targeting only German for 56. Isn't that the case (@peiying2)?

bedrock/firefox/templates/firefox/whatsnew/whatsnew-56.html
```diff
@@ -0,0 +1,63 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/whatsnew/base-pebbles.html" %}
```
flodolo: Please add a simple lang file reference to the template

```
{% add_lang_files "firefox/whatsnew_56" %}
```

Depending on the answer (German only?), we might need to add it to the base template too.

flodolo: Never mind the last part, this template has its own title and desc

bedrock/firefox/templates/firefox/whatsnew/base-pebbles.html
flodolo: OK, after trying to extract the strings I have a better picture. This has nothing to do with 56 being (or not) only for German.

Before exposing them to a shared file, I think it would be good to get some confirmation that it's useful. In other words, I don't want to add these strings to discover that no template will ever use them in the future.

If we start getting whatsnew pages using them, we can move them to main.lang

alexgibson: > Before exposing them to a shared file, I think it would be good to get some confirmation that it's useful. In other words, I don't want to add these strings to discover that no template will ever use them in the future.

Well, we need to start transitioning `/whatsnew` pages to use Pebbles instead of Sandstone, and this new base template enables that so future pages can build upon it. The meta strings included here are just for social sharing, but having them here means we no longer need to copy/paste them into each new template. I can't promise they won't ever change in the future, but I can at least run these strings past our copy writer today. 

> If we start getting whatsnew pages using them, we can move them to main.lang

Makes sense

flodolo:
I have the strings extracted locally, waiting for Pei to confirm which languages are involved.

flodolo:
> Please wait, we may have a string change coming still (I did say above I'd confirm when we're good to go)

They're not exposed for localization, only extracted locally. Basically, if you could keep further changes in a separate commit before the final squash, that would be enough.

alexgibson:
@flodolo - final string changes are in, I think we should be good to go here.

bedrock/firefox/templates/firefox/whatsnew/base-pebbles.html
flodolo: I missed these strings were for OG sharing, not the title/desc of the page. Adding them to main.lang, to avoid referencing the lang file in base.

flodolo:
> @flodolo - final string changes are in, I think we should be good to go here.

OK, updated my file.

alexgibson:
@flodolo OK, [as per the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1395788#c21) no further string changes needed here 👍

bedrock/firefox/templates/firefox/whatsnew/whatsnew-56.html
```diff
@@ -0,0 +1,84 @@
...
+    {% trans %}
+      We’ve been laying the groundwork for our biggest release of the year, coming in November.
+      The New Firefox will feature even more improvements to speed and performance, with a modern,
+      intuitive design to match.
+    {% endtrans %}
+    </p>
+
+    <ul>
+      <li>
+        <h3 class="new-tab">{{ _('Find it faster') }}</h3>
```
TheoChevalier: Don’t we want the search icon for this one, and the new-tab icon for the third item?

alexgibson: Thanks for the catch 👍

alexgibson:
Ok, this should be ready for review. We're waiting on one updated image for the browser screenshot in the middle section, but otherwise this should be complete.

alexgibson:
@jpetto or @craigcook this one is ready for review and needs to be in production before the 26th

alexgibson:
I just got some new images to drop into this page (no design changes), will update asap.

alexgibson:
@jpetto ok, there's some kind of inter-team snafu going on. The images may or may not change, but don't let that hold up review. If need be, @craigcook could take a last check over tomorrow if there are any changes.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-56.html
```diff
@@ -0,0 +1,84 @@
...
+      </li>
+    </ul>
+  </div>
+</section>
+</div>
+{% endblock %}
+
+{% block email_form %}{% endblock%}
+
+{% block js %}{% endblock %}
```
jpetto: Verifying that the update notification banner *should* be included in this template.

media/css/firefox/whatsnew/whatsnew-56.scss
```diff
@@ -0,0 +1,388 @@
...
+        color: #4a4a4a;
+        display: inline-block;
+        font-weight: bold;
+        margin: 0 auto;
+        padding: 12px 0 10px 60px;
+    }
+}
+
+html[dir="rtl"] .up-to-date-header {
+    h1 {
```
jpetto: Nit - extra nesting.

media/css/firefox/whatsnew/whatsnew-56.scss
```diff
@@ -0,0 +1,388 @@
...
+        padding: 12px 60px 10px 0;
+    }
+}
+
+/* -------------------------------------------------------------------------- */
+// Content intro
+
+.intro {
+    background: #0060df;
+    background: url('/media/img/firefox/whatsnew_56/fox-tail-header.png') top -340px center no-repeat,
```
jpetto: Nit - do we need a fallback `background` declaration here? Should we specify `background-color` and `background-image`?

media/css/firefox/whatsnew/whatsnew-56.scss
```diff
@@ -0,0 +1,388 @@
...
+    @media #{$mq-desktop-wide} {
+        .intro-wrapper:after {
+            height: 319px;
+            width: 547px;
+        }
+    }
+}
+
+html[dir="rtl"] {
+    .intro {
```
jpetto: Nit - extra nesting.

media/css/firefox/whatsnew/whatsnew-56.scss
```diff
@@ -0,0 +1,388 @@
...
+        left: auto;
+        margin-right: -65px;
+        right: 50%;
+    }
+
+    &:after {
+        left: 50%;
+        margin-left: -140px;
+        right: auto;
+    }
```
jpetto: Noticed that this results in the part of the image showing the tab being offscreen to the left. Not sure if a big deal or not.

media/css/firefox/whatsnew/whatsnew-56.scss
```diff
@@ -0,0 +1,388 @@
...
+        width: 280px;
+    }
+
+    @media #{$mq-tablet} {
+        min-height: 435px;
+        padding-top: 0;
+        width: 50%;
+
+        li {
+            @include font-size-level3;
```
jpetto: I'm sure you're following design direction, but this copy looks really big in the page!

media/css/firefox/whatsnew/whatsnew-56.scss
alexgibson: Ah this was copy pasta from the `/quantum` page, where we did need a fallback. Removing here 👍

media/css/firefox/whatsnew/whatsnew-56.scss
alexgibson: Yeah agree, this is as per the design though (and it does look a bit too small at `font-size-level4`)

media/css/firefox/whatsnew/whatsnew-56.scss
alexgibson: The design is supposed to position the browser image slightly offset from the viewport, but yeah here the tab part is not visible. Ideally we would have a RTL browser screenshot here, but I don't think we'll likely get one in time. I don't think it's really a big deal though.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-56.html
alexgibson: It's a standard `/whatsnew` page so yes, the banner should be included (this is handled in the `/whatsnew` base template)

alexgibson:
@jpetto thanks for the review. I added two new commits - one with review feedback and another updating the middle browser screenshot. Maybe if @craigcook can give this a quick glance over today we can squash & merge?

alexgibson:
@craigcook squashed, thanks for taking a second look

PR #5098, opened 2017-09-06
Bug 1392501 homepage experiment q32017

jpetto:
## Description

**Analytics requirements yet to come.**

Adds homepage variation that will be served to 100% of en-US visitors when switch is enabled.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1392501

## Testing

To test on the demo, ensure the above URL is added to `browser.uitour.testingOrigins` on `about:config`. ([docs](http://bedrock.readthedocs.io/en/latest/uitour.html#local-development))

To access the different promos:

1. Use Firefox signed in to Sync to see the Pocket promo
2. Use Firefox signed out of Sync *OR* Firefox on iOS/Android to see the Sync promo
3. Use non-Firefox to see the download Firefox promo

Ensure rendering is good in all browsers (IE 9+, Safari, Chrome).


jpetto:
@alexgibson I merged #5093 into a copy of this branch locally and didn't notice any issues, but a second look couldn't hurt.

bedrock/mozorg/templates/mozorg/home/home-b.html
```diff
@@ -0,0 +1,110 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import google_play_button with context %}
+
+{% extends "mozorg/home/home.html" %}
+
+{% add_lang_files "mozorg/home/index-2016" "firefox/features/index" "firefox/sync" "firefox/whatsnew_38" "firefox/whatsnew_50" %}
```
flodolo: Ugh, this is really bad…

Any of these files could activate the home page.

jpetto: > Ugh, this is really bad…

Is this technically bad, or just messy?

> Any of these files could activate the home page.

Not sure what this means.

flodolo: Bedrock shows a template on locale X based on the fact that the lang file is activated (has `##active##` at the beginning).

Here you are importing a ton of lang files for one template. Result: if even one of them is complete and activated, the entire template is active, even if we don't have all the strings translated.

Take `firefox/sync.lang`, is a template as old as it can be, with 89 languages activate. 

Now set your local install as PROD, and load the home page for 'ach': it should load almost completely in English, instead of redirecting to en-US.

bedrock/mozorg/templates/mozorg/home/home-b.html
```diff
@@ -0,0 +1,110 @@
...
+            <p>{{ _('Discover unlimited potential, endless possibility and a wide open Web.') }}</p>
+          {% endif %}
+
+          {{ download_firefox(dom_id='download-intro', button_color='button-arrow', download_location='primary cta') }}
+        </div>{#--/.show-default --#}
+
+        {# shown for Fx desktop signed out of sync, and Fx for Android & iOS #}
+        <div class="show-fx-signed-out show-fx-mobile">
+          {# lang file: firefox/sync #}
+          <h2>{{ _('Your web, how you like it, everywhere') }}</h2>
```
flodolo: Side note: we're removing firefox/sync in another open PR, and this string is not available in the new firefox/features/sync template

bedrock/mozorg/templates/mozorg/home/home-b.html
```diff
@@ -0,0 +1,110 @@
...
+
+          {# lang file: firefox/whatsnew_38 #}
+          <p>{{ _('With your Firefox Account you can sync your bookmarks, passwords, open tabs and more, and access them everywhere you use Firefox.') }}</p>
+
+          {# lang file: firefox/sync #}
+          {# for desktop users, link below is wired to open about:accounts #}
+          <a href="{{ url('firefox.features.sync') }}" class="button" id="fxa-sign-in">{{ _('Get started with Sync') }}</a>
```
flodolo: Same here, this is a string we're going to remove soon

alexgibson:
Page is missing a max-width?

screenshot-2017-9-11 internet for people not profit

This may well be intentional as-per the design, but should the page widths / margins be consistent with the other new branded pages?

bedrock/mozorg/templates/mozorg/home/home-b.html
```diff
@@ -0,0 +1,69 @@
...
+          <h2>{{ _('Your web, how you like it, everywhere') }}</h2>
+
+          <p>{{ _('With your Firefox Account you can sync your bookmarks, passwords, open tabs and more, and access them everywhere you use Firefox.') }}</p>
+
+          {# for desktop users, link below is wired to open about:accounts #}
+          <a href="{{ url('firefox.features.sync') }}" class="button" id="fxa-sign-in">{{ _('Get started with Sync') }}</a>
+        </div>{#--/.show-fx-signed-out--#}
+
+        <div class="show-fx-signed-in">
+          <h2>Pocket for Firefox</h2>
```
alexgibson: Unwrapped strings here and below?

bedrock/mozorg/templates/mozorg/home/home-b.html
```diff
@@ -0,0 +1,69 @@
...
+          <p>
+            Save articles, videos, pages and more for later.
+          </p>
+
+          <p><a href="https://getpocket.com/firefox/" rel="external noopener noreferrer" class="button">Get pocket</a></p>
```
alexgibson: Curious, why does this link need `noopener noreferrer`?

bedrock/mozorg/templates/mozorg/home/home-b.html
```diff
@@ -0,0 +1,69 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "mozorg/home/home.html" %}
```
alexgibson: Maybe I'm missing some part of the test, but why does this new template inherit from the old home page template? Would this make it hard to remove the old one later?

media/css/mozorg/home/home-b.scss
```diff
@@ -0,0 +1,576 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../../pebbles/includes/lib';
+@import '../../hubs/buttons';
+
+body {
+    @include open-sans;
```
alexgibson: I totally welcome using `open-sans` for body copy here :) (I'm hoping we can switch the rest of the new pages to do this by default later)

media/css/mozorg/home/home-b.scss
```diff
@@ -0,0 +1,576 @@
...
+
+h2,
+h3 {
+    @include zilla-slab;
+}
+
+// hide download button in the nav
+#nav-download-firefox,
+#global-nav-download-firefox {
+    display: none;
```
alexgibson: I'm guessing this is intentional as per the design, but fwiw If I'm signed into Sync on Nightly, then it's a little hard to download Firefox directly from the home page. Not a huge issue for most users, but I we've had bug reports like this in the past when conditional states on different pages cause frustration.

media/css/mozorg/home/home-b.scss
```diff
@@ -0,0 +1,576 @@
...
+    }
+}
+
+@media #{$mq-desktop} {
+    .moz-global-nav .nav-primary-links {
+        width: calc(100% - 220px);
+    }
+}
+
+// flexbox layout for savvy browsers to push nav items to the right
```
alexgibson: Is this needed?

media/css/mozorg/home/home-b.scss
```diff
@@ -0,0 +1,576 @@
...
+    .content {
+        display: block;
+        padding: 40px 20px;
+        position: relative;
+        width: 100%;
+        z-index: 2;
+    }
+
+    h2 {
+        @include font-size(32px);
```
alexgibson: Nit: try to use [font size mixins](https://github.com/mozilla/bedrock/blob/master/media/css/pebbles/includes/mixins/_typography.scss) for consistency, instead of using arbitrary font sizes. Same applies to occurences below

media/js/mozorg/home/home-b.js
```diff
@@ -0,0 +1,110 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function($, Mozilla, dataLayer) {
+    'use strict';
+
+    var HomePageB = {
```
alexgibson: This is totally fine for now, but I see this repeating/sharing of code with the sync/send-tabs pages something we could refactor into a reusable component in the future? :)

media/js/mozorg/home/home-b.js
```diff
@@ -0,0 +1,110 @@
...
+
+    HomePageB.ctaSyncClick = function(e) {
+        e.preventDefault();
+
+        HomePageB.dataLayer.push({
+            'event': 'sync-click',
+            'browser': HomePageB.state
+        });
+
+        // TODO: do we need to set custom utm params here?
```
alexgibson: Noting TODO

media/js/mozorg/home/home-b.js
```diff
@@ -0,0 +1,110 @@
...
+        Mozilla.UITour.showFirefoxAccounts(HomePageB.params.utmParamsFxA());
+    };
+
+    HomePageB.init = function(config) {
+        HomePageB.client = config.client;
+        HomePageB.fxVersion = HomePageB.client.FirefoxMajorVersion;
+        HomePageB.params = new window._SearchParams();
```
alexgibson: Hmm so, not necessarily introduced by this PR. But I noticed `'js/base/search-params.js'` is only available on all pages becuase it's part of the stub-attribution bundle. Given that can be toggled on/off independently, would it be better to move `'js/base/search-params.js'` to the common bundle?

alexgibson:
Thank you text lacks contrast:

screen shot 2017-09-11 at 16 43 17

bedrock/mozorg/views.py
```diff
@@ -198,7 +199,14 @@ def get_context_data(self, **kwargs):
 
 
 def home(request):
-    return l10n_utils.render(request, 'mozorg/home/home.html')
+    locale = l10n_utils.get_locale(request)
+
+    if locale == 'en-US' and switch('experiment-home-q32017'):
```
alexgibson: I'm sure there's a good reason, but why hide the new page behind a switch if it's going out to 100% en-US?

jpetto: I figured just in case it bombs in terms of engagement and we want to revert quickly.

media/js/mozorg/home/home-b.js
jpetto: If this logic lives on (which, to be frank, I hope it *doesn't*), yes, definitely worth refactoring into something more reusable.

media/css/mozorg/home/home-b.scss
jpetto: Sigh, yeah, the million edge cases are why we all loathe conditional content, particularly around download buttons. Perhaps I could check for release channel as well?

bedrock/mozorg/templates/mozorg/home/home-b.html
jpetto: Just being overly cautious. Probably not needed. Will remove.

bedrock/mozorg/templates/mozorg/home/home-b.html
jpetto: It's en-US only, and these strings are not set to be localized, so I don't see a reason to wrap them.

bedrock/mozorg/templates/mozorg/home/home-b.html
jpetto: I think because I approached this as a test, but it's not really a test, but it's kind of a test. There's still a fair amount of code shared between the two, and I'm not convinced that this new page is going to be *the* page at this point, so it seemed leaving the inheritance intact was the right move for now.

It wouldn't be very difficult to migrate this to a standalone template if that ends up being our direction.

bedrock/mozorg/templates/mozorg/home/home-b.html
alexgibson: Up to you, but I've fallen for this way too many times now in my own PRs

bedrock/mozorg/templates/mozorg/home/home-b.html
alexgibson: I like your optimism about this new page not lasting long (we've all heard this before), but if it's easy to pick apart still then 👍

media/css/mozorg/home/home-b.scss
jpetto: Ah, now that it's en-US only, nope, no longer needed. Nice spot.

media/css/mozorg/home/home-b.scss
jpetto: Yeah, I was being lazy and copy/pasted from `home.scss`. Will update to use the pre-defined font sizes.

media/css/mozorg/home/home-b.scss
alexgibson: Totally up to you and how complicated it gets, maybe something we could fix later?

media/js/mozorg/home/home-b.js
jpetto: Good catch. I'll move it from the `stub-attribution` bundle to the `common` bundle.

media/css/mozorg/home/home-b.scss
jpetto: Yep, after looking at it, going to leave as-is. Will comment in the JS file that it should be re-visited if/when the logic is re-addressed.

media/js/mozorg/home/home-b.js
```diff
@@ -77,20 +86,27 @@
...
                     Mozilla.UITour.getConfiguration('sync', function (config) {
 
-                        // Variation #2: Firefox 31+ signed IN to Sync
+                        // Variation #2: Firefox 31+ signed IN to Sync (shows Pocket promo)
                         if (config.setup) {
                             HomePageB.setBodyClass('state-fx-signed-in');
                             HomePageB.state = 'Firefox 31 or Higher: Signed-In';
 
+                            HomePageB.ctaPocket = document.getElementById('pocket-cta');
+                            HomePageB.ctaPocket.addEventListener('click', HomePageB.ctaPocketClick);
```
alexgibson: oh, just caught this - should add the default `false` as the last argument here and below?

PR #5109, opened 2017-09-13
[fix bug 1399585] IH Hub remove outdated content.

jpetto:
## Description

Removes outdated content from the IH hub home page.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1399585

## Testing

Make sure I didn't miss removing anything?


craigcook:
Do we know how much longer this "June campaign" page is supposed to live? Do we need to just remove these bits or should we remove the whole page and go back to the original? I've asked this question in the bug since I'm not sure who was requesting this update in the first place.

Assuming we do keep the campaign page with just these changes, once we remove that big header image the black box around the top section looks odd. I wonder if we should remove that background color and the extra padding so the page is white?

alexgibson:
+1 for removing the non-translated copy if possible. The June campaign thing was a bit of a mess, as they didn't think about l10n at all.

jpetto:
Will be moving common voice from promo to main content behind L10n tag. Blocked on new robot image.

craigcook:
Closing this PR and will open a fresh one later.

PR #5110, opened 2017-09-13
Some small changes to FF retention pages in css and a link.

ScottDowne:
## Description
Two small changes.
1. Updating a link that was requested. Should be straight forward enough.
2. If the page width is wide enough for thank-you-b https://www.mozilla.org/en-US/etc/firefox/retention/thank-you-b/ the "it's all thanks to you" blerb is no longer centered inside the confetti. The css changes should address this.

### DIFF
--- a/bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
+++ b/bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
@@ -47,7 +47,7 @@
     <article>
       <h1>{{ _('Get Firefox on your phone') }}</h1>
       <div class="time">{{ _('2 min install') }}</div>
-      <a href="https://support.mozilla.org/kb/install-firefox-android-device-using-google-play">{{ _('Download Firefox') }}</a>
+      <a href="https://www.mozilla.org/firefox/mobile-download/desktop/">{{ _('Download Firefox') }}</a>
     </article>
     <article>
       <h1>{{ _('Tell your friends about Firefox') }}</h1>
--- a/bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
+++ b/bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
@@ -47,7 +47,7 @@
     <article>
       <h1>{{ _('Get Firefox on your phone') }}</h1>
       <div class="time">{{ _('2 min install') }}</div>
-      <a href="https://www.mozilla.org/firefox/mobile-download/desktop/">{{ _('Download Firefox') }}</a>
+      <a href="{{ url('firefox.mobile-download-desktop') }}">{{ _('Download Firefox') }}</a>
     </article>
     <article>
       <h1>{{ _('Tell your friends about Firefox') }}</h1>
@@ -57,7 +57,7 @@
     <article>
       <h1>{{ _('Make the Internet a safer place') }}</h1>
       <div class="time">{{ _('4 min read') }}</div>
-      <a href="https://www.mozilla.org/internet-health/privacy-security/">{{ _('Learn more') }}</a>
+      <a href="{{ url('mozorg.internet-health.privacy-security') }}">{{ _('Learn more') }}</a>
     </article>
   </section>
 </main>
--- a/media/css/etc/firefox/retention/thank-you-b.scss
+++ b/media/css/etc/firefox/retention/thank-you-b.scss
@@ -32,7 +32,8 @@ main {
             header {
                 box-sizing: border-box;
                 text-align: center;
-                max-width: 490px;
+                max-width: 793px;
+                width: 100%;
                 margin-left: auto;
                 margin-right: auto;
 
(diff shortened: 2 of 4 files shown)


bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
```diff
@@ -47,7 +47,7 @@
     <article>
       <h1>{{ _('Get Firefox on your phone') }}</h1>
       <div class="time">{{ _('2 min install') }}</div>
-      <a href="https://support.mozilla.org/kb/install-firefox-android-device-using-google-play">{{ _('Download Firefox') }}</a>
+      <a href="https://www.mozilla.org/firefox/mobile-download/desktop/">{{ _('Download Firefox') }}</a>
```
jpetto: Should use our `url` helper here and below: `{{ url('firefox.mobile-download-desktop') }}`

media/css/etc/firefox/retention/thank-you-b.scss
```diff
@@ -41,10 +42,17 @@ main {
...
+                    max-width: 490px;
                     margin-left: auto;
-                    width: 793px;
+                    margin-right: auto;
+                }
+
+                @media #{$mq-desktop} {
                     height: 737px;
+                    margin-left: auto;
+                    margin-right: 0;
```
jpetto: No need to repeat `margin-left: auto`, as it will inherit from above.

jpetto:
Right column text is cut off at desktop breakpoint:

screen shot 2017-09-14 at 5 02 17 pm

bedrock/etc/templates/etc/firefox/retention/thank-you-a.html
ScottDowne: Oh interesting. 1. I didn't even notice that link was in moz.org but also 2. that's a cool helper!

jpetto: Yeah, keeps all links pointing to the same environment, which is pretty critical for not going crazy. 😉

ScottDowne: I found another link I used on moz.org too, so updating that too.

ScottDowne:
The right column breakpoint had me looking at a few other things that drew me to this: https://www.mozilla.org/de/etc/firefox/retention/thank-you-b/

Now that the page is being localized, some of my css is just not very robust.

In the case of English I'll look the same, but for German, I got it to go from this:

bad-german

To this:

good-german

ScottDowne:
@jpetto Updates! 🚀

jpetto:
A couple more minor-ish visual issues:

1. For some locales (like `fr`) on `thank-you-b`, the columns are lopsided. Could maybe use flexbox for supporting browsers ([using `@supports`](https://github.com/mozilla/bedrock/blob/55fc4980db2cbe1396d41abcb7727e0f9c0222dd/media/css/mozorg/home/home.scss#L110))?

screen shot 2017-09-18 at 1 38 41 pm

2. The newsletter area still has padding for a logo when below 760px wide. (This is a pre-existing bug, but would be nice to fix in this PR.)

screen shot 2017-09-18 at 1 39 20 pm

ScottDowne:
By lopsided, you mean it should instead look like this?

lopsided

jpetto:
Lopsided meaning the turquoise left side isn't as tall as the white background right side.

May not be a huge issue, but looks a little funky.

ScottDowne:
Ah yeah, I just saw that and was fixing anyway.

ScottDowne:
Updated a handful of other little things I found + what you found.

1. lopsided backgrounds, I didn't need to use flexbox but considered it. We can still use flexbox if my solution has something wrong that I missed, but I wanted to avoid `@supports` if I could get the same results without flexbox with not a lot of work.
1. The `a` tags had some styles I broke when I moved them into the `h1`, fixed.
2. Removed the `br` tags I added as they were adding that extra newline I didn't intend.
3. A few style things in the top most `h1` on the right panel I broke.
4. Fixed the signup container's top spacing.

ScottDowne:
Some errors, but I havn't yet made any sense of.

jpetto:
Re-ran the build and all tests are now passing. (Sometimes our CI falls down.)

Version A looks really good, but version B isn't looking great at < 1000px, particularly in German or Spanish. There's not really any consistent alignment - the header is centered, and the body copy is kind of all over the place.

screen shot 2017-09-19 at 2 20 14 pm

Even at 1200px+ wide, the right column's alignment is pretty haggard.

screen shot 2017-09-19 at 2 18 28 pm

ScottDowne:
@jpetto Yeah it's been bugging me, and it's not a great layout. It really just wants to be css grid layout.

I can come up with something, but would you be fine with me fixing it in a followup patch? Might make things easier to take the existing changes as is as they are better and add some significant improvements.

jpetto:
> I can come up with something, but would you be fine with me fixing it in a followup patch? Might make things easier to take the existing changes as is as they are better and add some significant improvements.

Yeah, I suppose. Jean says it's really important to get these updated links out, and she will ensure only en-US users see `thank-you-b`, so I can merge as-is.

ScottDowne:
Thanks! I'm working on a fix for the layout now. Hopefully should have another PR by end of day.

PR #5112, opened 2017-09-15
[fix bug 1399173] Add link to /firefox/ in the old site navigation

alexgibson:
## Description
- Adds a link to `/firefox/` in the old nav, now that the Firefox hub is fully localized.
- This adds the string `"Firefox"`, but should be ok with l10n?

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1399173

## Testing
- Make sure i didn't miss any other occurences of the old nav template?

### DIFF
--- a/bedrock/base/templates/base-pebbles.html
+++ b/bedrock/base/templates/base-pebbles.html
@@ -107,7 +107,10 @@
           <nav class="masthead-nav-main" id="nav-main">
             <span class="toggle" role="button" aria-controls="nav-main-menu" aria-expanded="false" tabindex="0">{{_('Menu')}}</span>
             <ul class="nav-main-menu" id="nav-main-menu">
-              <li class="first internet-health-item">
+              <li class="first firefox-item">
+                <a href="{{ url('firefox') }}" data-link-type="nav" data-link-name="Firefox">{{ _('Firefox') }}</a>
+              </li>
+              <li class="internet-health-item">
                 <a href="{{ url('mozorg.internet-health') }}" data-link-type="nav" data-link-name="Internet Health">{{ _('Internet Health') }}</a>
               </li>
               <li class="technology-item">
--- a/bedrock/base/templates/base-resp.html
+++ b/bedrock/base/templates/base-resp.html
@@ -115,7 +115,10 @@
           <nav id="nav-main" role="navigation">
             <span class="toggle" role="button" aria-controls="nav-main-menu" aria-expanded="false" tabindex="0">{{_('Menu')}}</span>
             <ul id="nav-main-menu">
-              <li class="first internet-health-item">
+              <li class="first firefox-item">
+                <a href="{{ url('firefox') }}" data-link-type="nav" data-link-name="Firefox">{{ _('Firefox') }}</a>
+              </li>
+              <li class="internet-health-item">
                 <a href="{{ url('mozorg.internet-health') }}" data-link-type="nav" data-link-name="Internet Health">{{ _('Internet Health') }}</a>
               </li>
               <li class="technology-item">
--- a/bedrock/base/templates/base.html
+++ b/bedrock/base/templates/base.html
@@ -102,7 +102,10 @@
         {% block site_header_nav %}
         <nav id="nav-main" role="navigation">
           <ul>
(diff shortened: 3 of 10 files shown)


alexgibson:
Successful test run: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/74/pipeline/

jpetto:
`/it/` breaks the line on the home page:

screen shot 2017-09-15 at 1 47 28 pm

A few locales (`it`, `de`, `fr`) line break on the hub pages:

screen shot 2017-09-15 at 1 47 14 pm

alexgibson:
@jpetto thanks for the review. Added some fixes

alexgibson:
Just noticed one more issue - will update shortly

alexgibson:
Ok, should be good for another look

jpetto:
Hm, a few more visual issues/comments:

- When under 1000px wide, it looks like the nav is accounting for the hidden download button. Should we expand the nav for this breakpoint to avoid two-line navs for some locales?

screen shot 2017-09-18 at 1 06 33 pm

- At *exactly* 760px wide, the menu is invisible on non `en-US` pages, e.g. `/it/`, `/fr/firefox/`, `/de/technology/`. (This appears to be a pre-existing bug in [`nav-main-resp.js`](https://github.com/mozilla/bedrock/blob/master/media/js/base/nav-main-resp.js#L224).)

screen shot 2017-09-18 at 12 58 06 pm

A couple minor nits:

- When between 760px and 1000px, should we switch to the little "m" logo as we do on the fully localized nav? Would give us a bit more room and allow the nav and subnav to line up nicely.

- At > 1000px on the Fx Hub pages, the main nav and sub nav don't line up on the left side. (I'd call this one very minor.) 

screen shot 2017-09-18 at 1 08 33 pm

I hate to say a pre-existing bug is a blocker, but I think having no nav for some visitors is of this variety. Everything else could probably be done in a follow-up PR if this needs to go out/you need to move on to 57 work.

alexgibson:
I'll have a look and see what I can do here. Much of this isn't really linked to this PR / bug, but I agree it needs improving.

jpetto:
> I'll have a look and see what I can do here. Much of this isn't really linked to this PR / bug, but I agree it needs improving.

I'll let you decide what else, if anything, you want to include with this PR. I know you're/we're slammed with 57 stuff, so no need to block this on incremental improvements that can be made later.

alexgibson:
@jpetto I fixed the top two items, but unfortunately I don't have the bandwidth to start tweaking both the nav and sub nav visually. nice spot on the 760px width bug 😩

jpetto:
Since you're going to squash this anyway, can I ask for the addition of one keystroke? In `nav-main-resp.js`, then handler for leaving small mode needs to check for `>= 760` (not just `> 760`).

Thanks!

alexgibson:
> In nav-main-resp.js, then handler for leaving small mode needs to check for >= 760 (not just > 760).

Hmm, so - uncovering this bug reveals another unfortunate and long-running flaw in Sandstone's breakpoints. Changing these values by 1px fixes the invisible nav on Pebbles pages, but now introduces a 1px offset in when the nav enters/leaves small mode on Sandstone pages. This is because breakpoints in sandstone-resp.less are defined like so:

/* Tablet Layout: 760px */
@media only screen and (min-width: @breakTablet) and (max-width: @breakDesktop) {

}

/* Mobile Layout: 320px */
@media only screen and (max-width: @breakTablet) {

}

As you can see, at exactly 760px wide, both media queries are satisfied :/

I'm loathed to risk tweaking this when it effects so many pages and has been like this for years. That said, I can't seem to notice the nav functionality actually breaking on Sandstone pages when we change the values in this PR. Functionally, it still seems to work OK.

I'm not sure what the best course of action is. It feels like overkill to form a Pebbles version of `nav-main-resp.js`. But I'm not sure what the next best option is. Thoughts?

alexgibson:
OK, as discussed on IRC I've reverted the changes to `nav-main-resp.js` as this is complicated enough to warrant it's own bug. I'll file one after this merges.

PR #5115, opened 2017-09-15
Fix bug 1400369: Send-to-device widget support multiple countries

pmac:
This adds a setting for specifying supported countries as they'll be added as support is added in Salesforce.

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -276,7 +276,7 @@
     {% set ios_link = firefox_ios_url('mozorg') %}
   {% endif %}
   {% set android_link = settings.GOOGLE_PLAY_FIREFOX_LINK %}
-  <section id="send-to-device" class="{% if include_logo %}logo {% endif %}{% if include_title %}title{% else %}no-title{% endif %}" data-key="{{ settings.MOZILLA_LOCATION_SERVICES_KEY }}" {% if spinner_color %}data-spinner-color="{{ spinner_color }}{% endif %}">
+  <section id="send-to-device" class="{% if include_logo %}logo {% endif %}{% if include_title %}title{% else %}no-title{% endif %}" data-countries="{{ send_to_device_countries() }}"{% if spinner_color %} data-spinner-color="{{ spinner_color }}"{% endif %}>
     <div class="form-container">
       {% if include_title %}
         <h2 class="form-heading">
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -1181,6 +1181,7 @@ SEND_TO_DEVICE_LOCALES = ['de', 'en-GB', 'en-US', 'en-ZA',
 
 # country code for /country-code.json to return in dev mode
 DEV_GEO_COUNTRY_CODE = config('DEV_GEO_COUNTRY_CODE', default='US')
+SEND_TO_DEVICE_COUNTRIES = config('SEND_TO_DEVICE_COUNTRIES', default='US', cast=Csv())
 
 RNA_SYNC_URL = config('RNA_SYNC_URL',
                       default='https://nucleus.mozilla.org/rna/sync/')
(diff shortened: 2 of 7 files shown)


pmac:
This is now updated with tests and a small bit of refactoring to make things make more sense (to me at least). Let me know what you think. Should be fully backward compatible with the "US only" way we were before.

media/js/base/send-to-device.js
```diff
@@ -49,34 +50,48 @@ if (typeof Mozilla === 'undefined') {
...
+    SendToDevice.prototype.getLocation = function() {
         var self = this;
-        var key = this.$widget.data('key');
 
-        // should location.services.mozilla.com be slow to load,
+        // if a dev has added ?geo=<country code> to the URL
+        // we can skip the geo lookup and act as if it worked
+        if (window.location.href.indexOf('?geo=') !== -1) {
+            var urlRe = /\?geo=([a-z]{2})/i;
+            var match = urlRe.exec(window.location.href);
```
jpetto: Could maybe test against `window.location.search` here (and above) instead of the full href?

bedrock/settings/base.py
```diff
@@ -1181,6 +1181,7 @@ def lazy_email_backend():
 
 # country code for /country-code.json to return in dev mode
 DEV_GEO_COUNTRY_CODE = config('DEV_GEO_COUNTRY_CODE', default='US')
+SEND_TO_DEVICE_COUNTIRES = config('SEND_TO_DEVICE_COUNTIRES', default='US', cast=Csv())
```
jpetto: `s/COUNTIRES/COUNTRIES` 😉 

(Also change in `helpers.py`)

bedrock/base/templates/macros.html
```diff
@@ -276,7 +276,7 @@
     {% set ios_link = firefox_ios_url('mozorg') %}
   {% endif %}
   {% set android_link = settings.GOOGLE_PLAY_FIREFOX_LINK %}
-  <section id="send-to-device" class="{% if include_logo %}logo {% endif %}{% if include_title %}title{% else %}no-title{% endif %}" data-key="{{ settings.MOZILLA_LOCATION_SERVICES_KEY }}" {% if spinner_color %}data-spinner-color="{{ spinner_color }}{% endif %}">
+  <section id="send-to-device" class="{% if include_logo %}logo {% endif %}{% if include_title %}title{% else %}no-title{% endif %}" data-countries="{{ send_to_device_countries() }}"{% if spinner_color %} data-spinner-color="{{ spinner_color }}{% endif %}">
```
jpetto: Not introduced in this PR, but easy enough to fix here. The final double quote should be moved inside the final `{% endif %}`. Currently an extra double quote is generated when `spinner_color` is false.

media/js/base/send-to-device.js
```diff
@@ -49,34 +50,48 @@ if (typeof Mozilla === 'undefined') {
...
+            if (match) {
+                SendToDevice.COUNTRY_CODE = match[1].toUpperCase();
+                self.updateMessaging();
+                if (typeof self.geoCallback === 'function') {
+                    self.geoCallback(SendToDevice.COUNTRY_CODE);
```
jpetto: If we're switching to upper case (which is fine, IMO), we'll need to change the scripts that are currently relying on lowercase:

- [`mobile-download-desktop.js`line 42](https://github.com/mozilla/bedrock/blob/master/media/js/firefox/mobile-download-desktop.js#L42)
- [`whatsnew-50.js` line 46](https://github.com/mozilla/bedrock/blob/master/media/js/firefox/whatsnew/whatsnew-50.js#L46)

bedrock/settings/base.py
pmac: LOL! Great catch 😆

media/js/base/send-to-device.js
pmac: I can switch to lower. No good reason to use one or the other as long as it's consistent.

pmac:
@jpetto thanks for the review. All items addressed.

PR #5117, opened 2017-09-20
FF Retention page css grid layout.

ScottDowne:
## Description

Fixing the right panel in FF Retention page b to be a grid like layout.

### DIFF
--- a/bedrock/etc/templates/etc/firefox/retention/thank-you-b.html
+++ b/bedrock/etc/templates/etc/firefox/retention/thank-you-b.html
@@ -41,34 +41,37 @@
       </p>
     </header>
   </section>
-  <section class="right-panel">
-    <h1>
-      {{ _('Here’s more super-easy things you can do to keep supporting our mission:') }}
-    </h1>
-    <article>
-      <span class="time">{{ _('1 min') }}</span>
-      <h1>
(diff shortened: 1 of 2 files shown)


ScottDowne:
@jpetto This seems to get what we want, but I'm not sure how I feel about it.

It uses the css `display: table` style to achieve this. Which is meh, but probably the best supported method I could find. Thoughts?

jpetto:
This is definitely an improvement at viewports around 700px, but below that (particularly down around 400px), things are actually worse, as the text in German is cut off.

To simplify things, I'd advise going with a stacked layout when below the tablet breakpoint, then enable this new grid-like layout after that.

What do you think?

(Apologies for the late review - I was on PTO, and then jumped right back in to the 57 launch stuff.)

ScottDowne:
I forgot about this. I think the campaign is over, and I dunno if we need this still.

If I'm doing more work on this, it might have to wait a week for me to be less busy again.

jpetto:
No problem! If the campaign is indeed over, feel free to close this PR. (Can always open again at a later date.)

Thanks!

alexgibson:
I'm going to close this PR as it seems inactive now, but feel free to reopen if this is something you would still like to fix 👍

PR #5121, opened 2017-09-21
[fix bug 1396029] Redesign Dev Edition page for Quantum

craigcook:
## Description
New dev edition product page, with new graphics and copy. I'm using `if lang_file_is_active` in the view and other locales will continue getting the old page until they're updated. This also makes minor updates to the old page, just swapping out the logo and fixing the fallback newsletter form.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1396029

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/firefox/developer/


bedrock/base/templates/includes/global-nav.html
```diff
@@ -105,7 +105,7 @@ <h3 class="summary" data-id="firefox">
                   </a>
                 </li>
                 <li>
-                  <a href="{{ url('firefox.products.developer') }}" data-link-type="nav" data-link-name="Developer Edition" data-link-group="Firefox">
+                  <a href="{{ url('firefox.developer') }}" data-link-type="nav" data-link-name="Developer Edition" data-link-group="Firefox">
```
craigcook: Changed this url structure to match the other Fx product pages for consistency.

bedrock/firefox/templates/firefox/products/developer-quantum.html
```diff
@@ -0,0 +1,236 @@
...
+        <h2 class="section-title"><span>{{ _('New Tools') }}</span></h2>
+
+        <div class="section-body">
+          <h3 class="body-title">{{ _('Firefox DevTools') }}</h3>
+
+          <img class="hero-image" src="{{ static('img/firefox/products/developer/quantum/hero-debugger-ani.gif') }}" alt="">
+          <p>
+          {% trans %}
+            The new Firefox DevTools are powerful, flexible, and best of all,
+            hackable. This includes a best-in-class javascript debugger, which
```
peiying2: javascript -> Javascript

craigcook: good catch! Fixing.

flodolo: Just to be sure: JavaScript

media/css/firefox/developer-quantum.scss
```diff
@@ -0,0 +1,596 @@
...
+    @media #{$mq-tablet} {
+        padding: 40px 0;
+        text-align: left;
+
+        &:after {
+            @include background-size(3000px 62px);
+            @include section-curve;
+            background-image: url('/media/img/firefox/products/developer/quantum/curve-download-bottom.svg');
+            height: 62px;
+            bottom: -62px;
```
schalkneethling: This leaves a thin black border between the nightly down and the newsletter section. Changing this to `bottom: -61px;` solves the problem.

craigcook:
This is ready for review but shouldn't go to production until Tuesday, 26 September, so marking DNM for now and we can merge to master closer to launch day.

media/css/firefox/developer-quantum.scss
```diff
@@ -0,0 +1,598 @@
...
+
+
+//*------------------------------------------------------------------*/
+// Performance
+.section-performance {
+    .section-title {
+        color: #3078fb;
+
+        span {
+            border-color: #3078fb;
```
schalkneethling: Wonder if it is worth moving this to a variable as well?

schalkneethling:
This looks great @craigcook!

A couple of general observations I thought I would call out.

Because you are hiding the download button in the global nav, the navigation looks pretty awkward especially at tablet width. Not sure how much effort it would be to allow the navigation items to use the new available space?

screen shot 2017-09-22 at 12 01 40

At around tablet width, the bottom curve slightly obscures the copy in the third column of the  features section:

screen shot 2017-09-22 at 12 02 38

From the code I noticed that the basic styling of the email form error is intentional. On the dark background there is not much of a choice really but, perhaps we can give the block some padding, perhaps even tweak the opacity a bit?

screen shot 2017-09-22 at 12 23 43

Not sure what the design requirement is but, would it improve the display if we centered the "Thank you" copy after successful newsletter submission. Both horizontally within the container as well as center the text itself?

screen shot 2017-09-22 at 12 24 42

peiying2:
The last devtools block doesn't have enough height. https://bugzilla.mozilla.org/show_bug.cgi?id=1402280

schalkneethling:
Other than the couple of observations and comment, this is looking awesome r+w/c 🍛

schalkneethling:
@craigcook Quick question regarding the sign up button for the newsletter. For the firstrun and what's new, it is shown to use the same type of gradient as the download button at the bottom of the page:

screen shot 2017-09-22 at 16 32 14

If this is not the case for the product page, perhaps I should not do that so it is consistent. Thoughts?

schalkneethling:
@craigcook Also, are we not going to have these pages behind a waffle flag? I am thinking that it would probably be easier when opening pull requests for firstrun and what's new if we can have this one merged to master, before I open pull requests for those two.

If not, that I would have to include all of the shared static assets in both of those pull requests that already exist in this one. Thoughts?

craigcook:
> @craigcook Quick question regarding the sign up button for the newsletter. For the firstrun and what's new, it is shown to use the same type of gradient as the download button at the bottom of the page:

I opted not to use that gradient because I think download buttons should be distinct from other buttons. Let's go with the hollow buttons in all newsletter forms. We can start trying to identify common patterns in these new branded pages and enforce some consistency of our own.

media/css/firefox/developer-quantum.scss
craigcook: Good catch, these should be $color-dev-mediumblue.

craigcook:
Updated to address comments and put the page behind the switch `fx-product-page-dev-quantum`. If I did that right, that means we should be able to merge this early and get the shared assets into master.

bedrock/firefox/views.py
```diff
@@ -677,6 +678,17 @@ class FirefoxHubView(BlogPostsView):
     template_name = 'firefox/hub/home.html'
 
 
+def FirefoxProductDeveloperView(request):
+    locale = l10n_utils.get_locale(request)
+
+    if lang_file_is_active('firefox/products/developer-quantum', locale) and switch('fx-product-page-dev-quantum'):
```
craigcook: Here's the switch.

PR #5123, opened 2017-09-21
Switch from NPM to Yarn for NodeJS deps

pmac:
And switch docker base to mozmeao/base:pythode-2.7-6.11


pmac:
This upgrades our versions of python and node, and hopefully addresses some reliability issues we've seen in CircleCI.

pmac:
FYI: the `mozmeao/base:pythode-*` images are from https://github.com/mozmeao/docker-pythode

pmac:
This looks like a big change but that's mostly because I'm deleting hundreds of files with the `node_modules` folder that was checked in so that we'd have `npm-lockdown` at install time. It'll be nice to have that gone.

pmac:
I've pushed this to my demo and run the tests and everything is green:

https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/demo%2fpmac/28/pipeline/

pmac:
/cc @alexgibson @jpetto 

This shouldn't affect much apart from the recommendation to use Yarn instead of npm, but nothing has changed that would disallow someone from using npm locally. It will just mean using `yarn add` to manage dependencies so that the `yarn.lock` will stay up to date.

docker/dockerfiles/bedrock_base
```diff
@@ -1,21 +1,10 @@
...
 
 WORKDIR /app
 EXPOSE 8000
 CMD ["./bin/run.sh"]
 
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends \
-                    gettext build-essential python-{dev,pip,setuptools} \
-                    libxml2-dev libxslt1.1 libxslt1-dev git libpq-dev
+RUN apt-install gettext build-essential libxml2-dev libxslt1.1 libxslt1-dev git libpq-dev
```
safwanrahman: any particular reason to remove `update`?

safwanrahman: Oh! sorry, Its `apt-install` that auto update

pmac: Exactly. It's just been consolidated into a [convenience script](https://github.com/mozmeao/docker-pythode/blob/master/apt-install) that does all of that. Just makes the dockerfile a bit cleaner looking.

pmac:
@jgmize I've updated this to use the `mozmeao/base:python-2.7` image as the base and included the official node commands in the `bedrock_build` dockerfile.

pmac:
And I just [ran a build of it in my demo](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/demo%2fpmac/29/pipeline/) and it passed.

PR #5128, opened 2017-09-23
Fix bug 1401708: The Beast awakens

pmac:
Update the Book of Mozilla to include the new verse heralding the release of Firefox 57 (bug 1370613).

### DIFF
--- a/bedrock/mozorg/templates/mozorg/book.html
+++ b/bedrock/mozorg/templates/mozorg/book.html
@@ -96,6 +96,24 @@ And the beasts brought <em class="f">fire</em> and light to the darkness.
 </p>
 
 
+{% if switch('firefox-57-release') %}
+<!-- 14th November 2017: Firefox 57 is officially released -->
+<!-- The major release of projects Quantum, Flow, and Photon, bringing Rust-based software,
+     major architectural changes, and a new look to the masses. -->
+
+<p class="moztext">
+The Beast adopted <em>new raiment</em> and studied the ways of <em>Time</em> and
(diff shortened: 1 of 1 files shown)


pmac:
I've marked this as blocked because I'm not sure if we should go ahead and deploy this or wait for November and release day. I'll try to find out.

pmac:
@alexgibson @jpetto @craigcook please let me know if any of you have opinions on the wording of the comment above the new verse. It looks from the others like it's an important part of this and I want to get it right :)

bedrock/mozorg/templates/mozorg/book.html
```diff
@@ -96,6 +96,21 @@
 </p>
 
 
+<!-- 14th November 2017: Firefox 57 is officially released -->
+<!-- The major release of projects Quantum, FLow, and Photon, bringing Rust-based software and major architectural changes to the masses. -->
```
flodolo: Flow. Also worth mentioning raiment (theme/aspect)

alexgibson:
Should we put this behind a switch? I'm thinking we may need to align on some sort of standard switch for the release of 57 that we can use across different pages (e.g. `firefox-57-release` or similar)

pmac:
That sounds perfect @alexgibson. The consensus seems to be that the page should not be updated until the 57 release. I'll add the switch to this PR.

bedrock/mozorg/templates/mozorg/book.html
pmac: Thanks flod!

alexgibson:
LGTM 👍

PR #5130, opened 2017-09-25
[fix bug 1393649] Update Firefox page-image for Quantum 57

alexgibson:
## Description
- Updates the Firefox page image for Quantum
- New image is behind the switch `FIREFOX_57_RELEASE` (still need to confirm if this is necessary)
- Removes `firefox-independent-1200.jpg` and it's references, which is just a duplicate of the existing `page-image.png`.
- This PR does not touch pages that have defined their own specific page image, which may need updating independently.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1393649

## Testing
- Are there any other pages I've missed updating that should be covered here?
- Make sure I didn't miss any references to the removed image.


alexgibson:
Ok, rebased this and also confirmed it does need to be behind a switch.

alexgibson:
@schalkneethling if you have some time for an r?

PR #5131, opened 2017-09-25
Fix Bug 1396007, adds revised Firefox privacy policy page

schalkneethling:
## Description
This adds the new revised Firefox privacy policy page located at privacy/firefox

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1396007

## Testing
No special testing scenarios.

### DIFF
--- a/bedrock/privacy/templates/privacy/notices/firefox-quantum.html
+++ b/bedrock/privacy/templates/privacy/notices/firefox-quantum.html
@@ -1,0 +1,67 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "base-pebbles.html" %}
+
+{% block page_title %}
+  {{ doc.body.section.h2.get_text() }}
+{% endblock %}
+
+{% block page_css %}
+  {% stylesheet 'firefox_privacy' %}
+{% endblock %}
+
+{% block body_id %}firefox-notice{% endblock %}
+
+{% block string_data %}
+  data-tabpanel-open-text="{{ _('Show All') }}"
+  data-tabpanel-close-text="{{ _('Hide All') }}"
+  data-choices-copy="{{ _('Choose how you want to share this data in Firefox.') }}"
+  data-choices-button="{{ _('Choose…') }}"
+{% endblock %}
+
+{% block content %}
+<article class="section-content" itemtype="http://schema.org/Article">
+
+{% set header = doc.body.section.extract() %}
+{% set lead_in = doc.body.section.extract() %}
+{% set footnote = doc.body.select('#footnote')[0].extract() %}
+
+<header class="privacy-head">
+  <div class="privacy-head-wrapper">
+    <h1 itemprop="name">{{ header.select('h2 span[class="privacy-header-firefox"]')|join|safe }}
+    {{ header.select('h2 span[class="privacy-header-policy"]')|join|safe }}</h1>
+
+    <time datetime="{{ header.select('[datetime]')[0]['datetime'] }}" itemprop="dateModified">{{ header.select('[datetime]')[0].string }}</time>
+  </div>
+</header>
+
+<section class="privacy-lead-in">
+  <div class="content-girdle">
+    {{ lead_in.select('header')|join|safe }}
+    <div class="optional-lead-in">
+      {{ lead_in.select('div p:nth-of-type(1)')|join|safe }}
+    </div>
+    {{ lead_in.select('div p:nth-of-type(2)')|join|safe }}
+  </div>
+</section>
+
+<main itemprop="articleBody">
+  {% set sections = doc.body.find_all('section', recursive=False) %}
+  {% for section in sections %}
+    {{ section|safe }}
+  {% endfor %}
+</main>
+
+<section class="privacy-footnote">
+  <div class="content-girdle">
+    {{ footnote.div|join|safe }}
+  </div>
+</section>
+</article>
+{% endblock %}
+
+{% block js %}
+  {% javascript 'firefox_privacy' %}
+{% endblock %}
--- a/media/css/privacy/firefox-privacy.scss
+++ b/media/css/privacy/firefox-privacy.scss
@@ -1,0 +1,319 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import
+    '../pebbles/includes/lib',
+    '../pebbles/includes/fonts/open-sans';
+
+article {
+    @include open-sans;
+}
+
+$light-gray-background: #f2f2f2;
+
+.privacy-head,
+.privacy-lead-in,
+.privacy-footnote {
+    background-color: $light-gray-background;
+    color: #333;
+}
+
+/* POLICY HEADER */
+.privacy-head {
+    padding: 20px;
+    text-align: left;
+
+    @media #{$mq-tablet} {
+        padding: 20px 30px;
+    }
+
+    @media #{$mq-desktop} {
+        position: sticky;
+        top: 0;
+        padding: 0;
+        text-align: center;
+        z-index: 2;
+
+        .privacy-head-wrapper {
+            margin: 0 auto;
+            padding-left: 17px;
+            max-width: 1420px;
+        }
+    }
+
+    h1 {
+        display: inline-block;
+        @include font-size-level2
+
+        @media #{$mq-desktop} {
+            margin: 40px 0;
+        }
+
+        a {
+            color: inherit;
+            text-decoration: none;
+        }
+
+        &:before {
+            display: inline-block;
+            content: '';
+            @include at2x('/media/img/firefox/template/logo-large.png', 30px, 33px);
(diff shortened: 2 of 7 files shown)


alexgibson:
When I pull this down and update legal docs submodule I get the error:

```
UndefinedError 'None' has no attribute 'extract'
```

Are there any special steps I need to take to get this running?

alexgibson:
With the accompanyment of the global navigation, the header looks a little odd. I think it should either sit flush inline with the mozilla wordmark logo, or alternatively be centered llike the rest of the page (that may be not exact to the design, but I think either option would be a little tidier?)

header

alexgibson:
On longer titles, the show/hide button runs directly underneath the text which looks a little cramped

title

media/css/privacy/firefox-privacy.scss
```diff
@@ -0,0 +1,282 @@
...
+
+        a {
+            color: inherit;
+            text-decoration: none;
+        }
+
+        &:before {
+            display: inline-block;
+            content: '';
+            @include at2x('/media/img/firefox/australis/fx38.0.5/firefox-logo.png', 30px, 33px);
```
alexgibson: This is a logo that belongs to a different (and very old) page that could be removed at any time. I'd suggest using one from the shared /firefox/template/ directory instead.

media/css/privacy/firefox-privacy.scss
```diff
@@ -0,0 +1,282 @@
...
+            width: 30px;
+            height: 33px;
+            vertical-align: middle;
+
+            @media #{$mq-tablet} {
+                background-size: 40px 43px;
```
alexgibson: Nit: use `background-size()` mixin here and below?

media/css/privacy/firefox-privacy.scss
```diff
@@ -0,0 +1,282 @@
...
+        @include font-size-level3
+        font-weight: 600;
+        text-align: left;
+
+        @media #{$mq-tablet} {
+            font-weight: normal;
+        }
+
+        @media #{$mq-desktop} {
+            margin-bottom: (20px * 2);
```
alexgibson: Nit: simplify and use `40px` (same for other occurences below)

media/css/privacy/firefox-privacy.scss
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    h3.collapsed {
+        &:after {
+            background-image: url('/media/img/privacy/arrowhead-down-16.svg');
+        }
+    }
+
+    section {
+        section {
```
alexgibson: Nit: unnecessary nesting

media/js/privacy/firefox-privacy.js
```diff
@@ -0,0 +1,288 @@
...
+
+            Mozilla.UITour.openPreferences('privacy-reports');
+        }
+    });
+
+    /* add a class to indicate that js is enabled. This will trigger
+    the appropriate styling to be applied */
+    mainContent.classList.add('interactive-widget');
+
+    initPage();
```
alexgibson: All this vanilla JS is great 👍 , but I'd suggest using a feature detect just to stop older browsers throwing a bunch of errors?

media/js/privacy/firefox-privacy.js
```diff
@@ -0,0 +1,288 @@
...
+        var initialTopicContent = initialTopic.querySelector('div');
+        var initialTopicHeading = initialTopic.querySelector('h3');
+
+        initialTopicContent.classList.remove('hidden');
+        initialTopicContent.setAttribute('aria-hidden', false);
+
+        initialTopicHeading.classList.remove('collapsed');
+
+        if (Mozilla.Client.isFirefoxDesktop) {
+            addDataChoicesWidget(initialTopicContent);
```
alexgibson: Would it be worth adding this behind a call to `Mozilla.UITour.ping()`? That was the button will only show if UITour is working.

media/js/privacy/firefox-privacy.js
```diff
@@ -0,0 +1,288 @@
...
+            targetElem.textContent = tabpanelCloseText;
+        }
+    }
+
+    /**
+     * Listens for click events on items inside the main content area.
+     * It then calls the appropriate function based on the event target.
+     * This avoids adding multiple event handlers on individual elements.
+     */
+    mainContent.addEventListener('click', function(event) {
```
alexgibson: Event delegation definitely has benefits for some scenarios, but you don't have that many events here and it does make the code a bit harder to read with everythign in a single handler like this. Not suggesting you change anythin now though, as this does all seem to work :)

media/js/privacy/firefox-privacy.js
```diff
@@ -0,0 +1,288 @@
...
+     *
+     * {
+     *     elem: document.getElementById('strings'),
+     *     property: 'isExpanded',
+     *     value: 'value'
+     * }
+     *
+     * @returns For get operations, it returns the value of the data attribute
+     */
+    function dataAttributeHandler(data) {
```
alexgibson: This seems to work fine and I'm all for using newer features, but could all this somewhat complex code be avoided by just using `getAttribute` and `setAttribute`? I know they are older, but less code and greater simpliciy has some benefit still. You by no means have to change this now that it's written though :)

alexgibson:
On RTL locales the header should probably be reversed also:

rtl

bedrock/privacy/templates/privacy/notices/firefox-quantum.html
```diff
@@ -0,0 +1,67 @@
...
+{% block content %}
+<article class="section-content" itemtype="http://schema.org/Article">
+
+{% set header = doc.body.section.extract() %}
+{% set lead_in = doc.body.section.extract() %}
+{% set footnote = doc.body.select('#footnote')[0].extract() %}
+
+<header class="privacy-head">
+  <div class="privacy-head-wrapper">
+    <h2 itemprop="name">{{ header.select('h2 span[class="privacy-header-firefox"]')|join|safe }}
```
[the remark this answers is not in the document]
schalkneethling: Well, I was thinking that. Then I saw the Mozilla logo is marked up as a `H2` so, then I felt it was "wrong" to have a `H1` follow a `H2`. With that said though, this is inside an `article` and the `header` of a `section` so, as far as a HTML5 outline it is totally fine.

`H1` it is.

media/css/privacy/firefox-privacy.scss
```diff
@@ -0,0 +1,282 @@
...
+
+* {
+    @include open-sans;
+}
+
+@font-face {
+    font-family: 'Open Sans';
+    font-weight: 600;
+    font-style: normal;
+    src: url('/media/fonts/opensans-semibold.woff2') format('woff2'),
```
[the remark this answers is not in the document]
schalkneethling: Hmmm so, in the invision doc it specifically asked for semibold but, I felt a little uncomfortable as it is not part of the default stack. I then saw some other pages sometimes include additional fonts so, thought it might be a thing we do from time to time.

With what Craig said yesterday though, I am kinda thinking twice about this now, especially seeing that you brought it up. Perhaps it is worth dropping that extra weight and request, and just using bold.

media/js/privacy/firefox-privacy.js
schalkneethling: Yeah, I considered that but then, the temptation of writing something to make it work won ;p

media/js/privacy/firefox-privacy.js
schalkneethling: Oh, that is even better. Thanks!

media/css/privacy/firefox-privacy.scss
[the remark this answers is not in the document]
alexgibson: It is up to you, but my bet is most folks won't even notice 😉

media/js/privacy/firefox-privacy.js
alexgibson: I'd still keep the check for `Mozilla.Client.isFirefoxDesktop` fwiw (to avoid calling UITour entirely for other browser)

media/css/privacy/firefox-privacy.scss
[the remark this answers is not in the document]
schalkneethling: Yup, I removed it and went with plain old bold.

media/css/privacy/firefox-privacy.scss
schalkneethling: Remnant from when I was using `$gutter-width` ;)

schalkneethling:
Thanks for the review @alexgibson - updated

media/js/privacy/firefox-privacy.js
```diff
@@ -280,9 +284,11 @@
         }
     });
 
-    /* add a class to indicate that js is enabled. This will trigger
-    the appropriate styling to be applied */
-    mainContent.classList.add('interactive-widget');
-
-    initPage();
+    // do not execute in IE8 and lower
+    if (document.addEventListener) {
```
alexgibson: Nit: change this to `'addEventListener' in window` (is this single detect enough?)

Also, I believe even with this check IE will throw an error since you have `mainContent.addEventListener('click')` above.

schalkneethling: Will do.

> is this single detect enough?
Yes, this will filter out IE8 and below

> Also, I believe even with this check IE will throw an error since you have mainContent.addEventListener('click') above.

True, I can move that into the `if` conditional to avoid that. Good call.

schalkneethling:
@alexgibson updated and rebased.

PR #5136, opened 2017-09-26
Bug 1400369: Send-to-device sends country and language when available

pmac:
This is so the SMS infra can pick the proper message that is deliverable in said country and language.


pmac:
Overlooked the sending of the info to basket last PR. This should fix things and fully prepare us for these new international messages.

PR #5140, opened 2017-09-27
Bug 1386184 - Firstrun page should skip the FxA form if the user is signed in

gasolin:
## Description

Use Mozilla.UITour.getConfiguration("sync") API to get FxA signin status and skip the FxA form if the user is signed in
http://searchfox.org/mozilla-central/source/browser/components/uitour/UITour.jsm#1605

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1386184

### DIFF
--- a/media/js/firefox/firstrun/firstrun.js
+++ b/media/js/firefox/firstrun/firstrun.js
@@ -41,24 +41,33 @@
 
         skipbutton.onclick = onVerificationComplete;
 
-        Mozilla.Client.getFirefoxDetails(function(data) {
-            Mozilla.FxaIframe.init({
-                distribution: data.distribution,
-                gaEventName: 'firstrun-fxa',
-                onVerificationComplete: onVerificationComplete,
-                onLogin: onVerificationComplete,
-                onFormEnabled: disableSkipButton,
-                onNavigated:  hideOrShowSkipButton
-            });
+        Mozilla.UITour.getConfiguration('sync', function(config) {
+            if (config.setup) {
+                // skip showing the sign in form if the user is signed in
+                document.getElementById('sky').addEventListener('transitionend', function() {
+                    onVerificationComplete();
+                });
+            } else {
+                document.getElementById('sky').addEventListener('transitionend', function(event) {
+                    if (event.propertyName === 'opacity') {
+                        scene.dataset.modal = 'true';
+                    }
+                }, false);
+
+                Mozilla.Client.getFirefoxDetails(function(data) {
+                    Mozilla.FxaIframe.init({
+                        distribution: data.distribution,
+                        gaEventName: 'firstrun-fxa',
+                        onVerificationComplete: onVerificationComplete,
+                        onLogin: onVerificationComplete,
+                        onFormEnabled: disableSkipButton,
+                        onNavigated:  hideOrShowSkipButton
+                    });
+                });
+            }
         });
 
         scene.dataset.sunrise = 'true';
-
-        document.getElementById('sky').addEventListener('transitionend', function(event) {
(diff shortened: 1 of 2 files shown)


media/js/firefox/firstrun/firstrun.js
```diff
@@ -41,15 +41,22 @@
...
-        Mozilla.Client.getFirefoxDetails(function(data) {
-            Mozilla.FxaIframe.init({
-                distribution: data.distribution,
-                gaEventName: 'firstrun-fxa',
-                onVerificationComplete: onVerificationComplete,
-                onLogin: onVerificationComplete,
-                onFormEnabled: disableSkipButton,
-                onNavigated:  hideOrShowSkipButton
-            });
+        Mozilla.UITour.getConfiguration('sync', function(config) {
```
alexgibson: This throws an error `TypeError: Mozilla.UITour is undefined`. Please make sure to test your changes locally. Thanks!

gasolin: Thanks for review. I figure out that I need to run both `manage.py runserver` & `gulp` **at same time** to run bedrock properly.

Ref to `http://bedrock.readthedocs.io/en/latest/uitour.html#local-development`
While test please make sure to add default prefs in `browser/app/profile/firefox.js`
```
pref("browser.uitour.testingOrigins", "http://127.0.0.1:8000");
pref("browser.uitour.requireSecure", false);
```
and test the page on the custom firefox build.

Then open `http://127.0.0.1:8000/en-US/firefox/57.0/firstrun/` to see how it goes.
(without login sync -> normal signin page, with login to sync -> redirect to about:home directly after sunrise)

In local test I can't see the FxA iframe though, but it should be work.

alexgibson:
The animation seems broken. I attached a gif to show you what I maen:

firstrun

I think the issue it you're trying to step through the ending animation too early (the first step never has a chance to happen).

gasolin:
Thanks for catching that! I added a listener to wait for `sky` transitionend. Now it redirects after all animations are complete.

alexgibson:
The sunrise happens ok now, but the glitch in the ending animation still reminas.

gasolin:
I saw the glitch in the ending animation but can't find the cause yet. Any suggestion is welcome
http://g.recordit.co/eT5eIygseE.gif

alexgibson:
Sorry I don't have time to dig into this right now as we're busy with FF 57 work, but I can try and take a look as soon as I can find some spare time. Fwiw, this page is currently being redesigned for 57.

gasolin:
@alexgibson if the page is being redesigned for v57, do you think this issue isstill  worth to solve?

alexgibson:
This page will still be around I would think for older Firefox versions, at least for a while. So it’s still worth fixing if this is something that’s important to the onboarding team. It will probably also need implementing on the newly designed page as well.

gasolin:
@alexgibson I don't know there's a new page for v57 but I think we only need implement this on the newly designed page. Could you point me where or how to show the newly designed page in bedrock?

alexgibson:
The new page hasn’t been coded just yet.

alexgibson:
@gasolin if you’re having trouble finishing this bug, please speak to Eric Renaud on the websites team and we can prioritise finishing this off.

gasolin:
This is an optional fix which just affect a small portion of user. (they can still press the `skip auth` button) And if this page is not target for v57, we don't need to fix it in priority.

gasolin:
I did some more test and found the animation works fine on Chrome based browser. the performance timeline shows gc and relayout is happen during the timing of glitch. Once the new page is available, I'll test there if that still happens.

PR #5141, opened 2017-09-27
[fix bug 1401402] Update channel page with current logos

alexgibson:
## Description
- Updates to channel page:
  - Add Quantum style page header
  - Add new Beta logo
  - Add new Developer Edition logo
  - Add new page image for social sharing
  - Change border-radius of all default buttons to `100px` (I figure this is relatively safe, and better than even more overriding?)

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1401402

## Testing
- Test new styling for any regressions
- Make sure I didn't remove any logos that are still in use?


media/css/firefox/channel.less
```diff
@@ -75,12 +88,18 @@
...
         }
     }
 }
 
+.html-rtl #masthead .logo {
+    float: left;
+    margin: 10px 12px 0 0;
+}
+
+
```
schalkneethling: Are the two blank lines above intentional?

schalkneethling:
In IE9 large viewport, the developer and nightly logos are not scaled appropriately.

virtualbox_windows 7 - ie9_27_09_2017_15_20_31

And smaller viewports they are scaled appropriately but, at mobile phone viewports, the developer and nightly logos looks pixelated:

virtualbox_windows 7 - ie9_27_09_2017_15_21_07

schalkneethling:
IE10 has the same behavior as noted above. I also noticed that in both IE9 and 10, the page heading overlaps the logo.

virtualbox_windows 7 - ie10_27_09_2017_15_29_47

schalkneethling:
Strangely, same behavior in Ie11 and Edge. It seems that at large viewport widths there is no `background-size` being set for those two logos.

alexgibson:
@schalkneethling thanks for the review, added a second commit with some fixes.

media/css/firefox/channel.less
```diff
@@ -484,38 +500,41 @@ html[lang^="en"] {
 }
 
 #beta header {
-    .at2x('/media/img/firefox/template/beta-logo.png', 180px, 189px);
+    .at2x('/media/img/logos/firefox/beta-logo.png', 180px, 178px);
+    .background-size(180px 178px);
```
schalkneethling: Nit: Use `background-size` mixin for this and the other instances?

schalkneethling:
Tinypng was able to squeeze the tail image down to 41.6kb and the non high-res logo to 7.6kb - attaching them here.

[tinified.zip](https://github.com/mozilla/bedrock/files/1337490/tinified.zip)

alexgibson:
> Tinypng was able to squeeze the tail image down to 41.6kb and the non high-res logo to 7.6kb - attaching them here.

I think these images have already been compressed once, I think that's likely enough?

alexgibson:
> Tinypng was able to squeeze the tail image down to 41.6kb and the non high-res logo to 7.6kb - attaching them here.

Ok, this new image has some fairly visible artifacts. I'm not going to compress it a second time

media/css/firefox/channel.less
schalkneethling: Oh bleh it;s LESS :) I was looking for the `@include background-size...`

PR #5143, opened 2017-09-27
[fix bug 1396029] Minor updates to Dev Edition page

craigcook:
## Description
A few superficial tweaks to the Dev Edition product page.

- Updated hero screenshot.
- Some spacing and alignment fixes.
- <s>Updated meta description (uses the l10n tag `dev_edition_quantum_updates`).</s>
- New newsletter title copy (English-only).
- Removed the `fx-product-page-dev-quantum` switch, no longer needed (I'll have a cleanup PR to delete it in www-config as well).

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1396029


bedrock/firefox/templates/firefox/products/developer-quantum.html
```diff
@@ -7,7 +7,13 @@
 {% extends "firefox/base-pebbles.html" %}
 
 {% block page_title %}{{_('Firefox Developer Edition')}}{% endblock %}
-{% block page_desc %}{{ _('Built for those who build the Web. The only browser made for developers like you.') }}{% endblock %}
+{% block page_desc %}
+{% if l10n_has_tag('dev_edition_quantum_updates') %}
+  {{ _('Firefox Developer Edition is the blazing fast browser that offers cutting edge developer tools and the latest features like CSS Grid support and framework debugging.') }}
```
flodolo: Can we add this without a period? We just exposed that string as part of whatsnew/firstrun, and we could avoid the tag.

craigcook: Oh that's good news. Period and tag removed.

schalkneethling:
This looks great @craigcook - Only one small concern. The line height on the heading and copy for the newsletter form looks cramped to me at small viewports. Thoughts?

screen shot 2017-09-28 at 15 12 55

schalkneethling:
Other than that, it's an r+w/c from me 🚀

craigcook:
> This looks great @craigcook - Only one small concern. The line height on the heading and copy for the newsletter form looks cramped to me at small viewports. Thoughts?

It's an h4 so it was getting heading line-height, even though it looks like it should be a paragraph. I've bumped it up and also adjusted the width of the form on mobile.

craigcook:
Also changed the all-caps titles to sentence case, by request. Since we're not using Antonio those titles are extra shouty in Open Sans.

schalkneethling:
> Also changed the all-caps titles to sentence case, by request. Since we're not using Antonio those titles are extra shouty in Open Sans.

@craigcook  I wonder, should I make these updates to the firstrun and whatsnew pages as well? It would be consistent, and these do look a lot better aesthetically as well.

craigcook:
> @craigcook I wonder, should I make these updates to the firstrun and whatsnew pages as well? It would be consistent, and these do look a lot better aesthetically as well.

I already did 😁

PR #5144, opened 2017-09-27
Conditional use new or old template for Firefox privacy policy

schalkneethling:
## Description
Prevents errors when loading old format of Firefox privacy policy

## Issue / Bugzilla link
No issue

## Testing
Using /en-US/privacy/firefox/ should load new template using /hu/privacy/firefox/ should load old template with old privacy policy

### DIFF
--- a/bedrock/privacy/views.py
+++ b/bedrock/privacy/views.py
@@ -70,8 +70,17 @@ class PrivacyDocView(LegalDocView):
         return doc
 
 
-firefox_notices = PrivacyDocView.as_view(
-    template_name='privacy/notices/firefox-quantum.html',
+class FirefoxPrivacyDocView(PrivacyDocView):
+    template_name = 'privacy/notices/firefox.html'
+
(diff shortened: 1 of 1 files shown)


bedrock/privacy/views.py
```diff
@@ -70,8 +70,18 @@ def get_legal_doc(self):
         return doc
 
 
-firefox_notices = PrivacyDocView.as_view(
-    template_name='privacy/notices/firefox-quantum.html',
+class FirefoxPrivacyDocView(LegalDocView):
```
pmac: I think I would add a default to the class itself, then override that in the method if the selector matches:

```python
class FirefoxPrivacyDocView(LegalDocView):
    template_name = 'privacy/notices/firefox.html'
    ...
```
Then just remove that else statement in the `get_legal_doc` method.

pmac: Also just subclass `PrivacyDocView` and remove line 76.

PR #5149, opened 2017-09-28
Fix Bug 1394042, redirect cloud services pn to Fx pn

schalkneethling:
## Description
As requested in the bug, all traffic to '/privacy/firefox-cloud/' should be redirected to '/privacy/firefox/'

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1394042

## Testing

Requests for '/privacy/firefox-cloud/' should redirect to '/privacy/firefox/'

### DIFF
--- a/bedrock/releasenotes/tests/test_models.py
+++ b/bedrock/releasenotes/tests/test_models.py
@@ -39,7 +39,7 @@ class TestReleaseNotesURL(TestCase):
         mock_reverse.assert_called_with('firefox.desktop.releasenotes', args=('42.0', 'release'))
 
 
-@override_settings(RELEASE_NOTES_PATH=RELEASES_PATH)
+@override_settings(RELEASE_NOTES_PATH=RELEASES_PATH, DEV=False)
 class TestReleaseModel(TestCase):
     def setUp(self):
         release_cache.clear()
@@ -57,7 +57,7 @@ class TestReleaseModel(TestCase):
         assert 'custom url' == rel.get_bug_search_url()
 
     def test_equivalent_release_for_product(self):
-        """Based on the test files the equivalent release for 56 should be 56.0.2."""
+        """Based on the test files the equivalent release for 56 should be 56.0.2"""
         rel = models.get_release('firefox', '56.0', 'release')
         android = rel.equivalent_release_for_product('Firefox for Android')
         assert android.version == '56.0.2'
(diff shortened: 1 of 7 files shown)


bedrock/releasenotes/tests/test_models.py
```diff
@@ -57,10 +57,10 @@ def test_get_bug_search_url(self):
         assert 'custom url' == rel.get_bug_search_url()
 
     def test_equivalent_release_for_product(self):
-        """Based on the test files the equivalent release for 56 should be 56.0.2."""
+        """Based on the test files the equivalent release for 56 should be 56.0.3"""
         rel = models.get_release('firefox', '56.0', 'release')
         android = rel.equivalent_release_for_product('Firefox for Android')
-        assert android.version == '56.0.2'
+        assert android.version == '56.0.3'
```
schalkneethling: This test failed for me locally so I updated the version number based on the expected assertion.

schalkneethling:
@pmclanahan After having done the actual redirect, I looked at `urls.py` and now I am wondering. Should I just change

`url(r'^/firefox-cloud/$', views.firefox_cloud_notices, name='privacy.notices.firefox-cloud'),`

to be

`url(r'^/firefox-cloud/$', views.firefox_notices, name='privacy.notices.firefox'),`

Or is the explicit redirect the preferred method? Thanks!

bedrock/releasenotes/tests/test_models.py
pmac: Ah! That's my bad. That is dependent on `DEV=False`. The fix here is not to change the version but to add `@override_settings(DEV=False)` to that test method, or just change the decorator on the class to `@override_settings(RELEASE_NOTES_PATH=RELEASES_PATH, DEV=False)`.

pmac:
Apart from the failing test this looks good. They specifically requested a redirect, and having the same content at multiple URLs is bad for the Google Juice. Just fix up that test for me (sorry about that) and we'll be good to go!

schalkneethling:
Thanks @pmac - updated.

pmac:
There's one more thing I forgot. I think we should also remove the entry for firefox-cloud in privacy/urls.py and the associated template file. If we leave them now they'll probably sit in the repo for quite a while before someone realizes they're no longer needed.

schalkneethling:
Done, and done ;p Thanks @pmac

PR #5150, opened 2017-09-28
Fix Bug 1391847, changes to secondary nav for /developer and /technology

schalkneethling:
## Description
Adds new sub navigation items that are consistent between the /developer and /technology pages

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1391847

## Testing
/technology and /developer should have the same sub navigation but:

1) when on /technology, the first item in the subnav will be Developer Hub and link to /developer
2) when on /developer, the first item in the subnav will be Technology Hub and link to /technology

### DIFF
--- a/bedrock/base/templates/includes/global-nav.html
+++ b/bedrock/base/templates/includes/global-nav.html
@@ -248,14 +248,22 @@
                   </a>
                 </li>
                 <li>
-                  <a href="{{ url('mozorg.developer') }}" data-link-type="nav" data-link-name="Mozilla Developers" data-link-group="Technology">
-                    {{ _('Mozilla Developers') }}
-                  </a>
+                    <a href="{{ url('mozorg.developer') }}" data-link-type="nav" data-link-name="Developer Hub" data-link-group="Technology">{{ _('Developer Hub') }}</a>
                 </li>
                 <li>
-                  <a href="https://developer.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=MDN" data-link-type="nav" data-link-name="MDN" data-link-group="Technology">
-                    <abbr title="{{ _('Mozilla Developer Network') }}">{{ _('MDN') }}</abbr>
-                  </a>
+                  <a href="https://research.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=Research" data-link-type="nav" data-link-name="Research" data-link-group="Technology">{{ _('Research') }}</a>
+                </li>
+                <li>
+                  <a href="https://vr.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=VR" data-link-type="nav" data-link-name="Virtual Reality" data-link-group="Technology">{{ _('Virtual Reality') }}</a>
+                </li>
+                <li>
+                  <a href="https://games.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=Games" data-link-type="nav" data-link-name="Games" data-link-group="Technology">{{ _('Games') }}</a>
+                </li>
+                <li>
+                  <a href="https://developer.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=Documentation" data-link-type="nav" data-link-name="Documentation" data-link-group="Technology">{{ _('Documentation') }}</a>
+                </li>
+                <li>
+                  <a href="https://hacks.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=HacksBlog" data-link-type="nav" data-link-name="Hacks Blog" data-link-group="Technology">{{ _('Hacks blog') }}</a>
                 </li>
               </ul>
 
--- a/bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
+++ b/bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
@@ -7,7 +7,22 @@
 {% block sub_nav_class %}technology{% endblock %}
 
 {% block sub_nav_primary_links %}
-<li><a href="{{ url('mozorg.technology') }}" data-link-name="Web Innovations" data-link-type="nav" data-link-position="subnav">{{ _('Web Innovations') }}</a></li>
-<li><a href="{{ url('mozorg.developer') }}" data-link-name="Mozilla Developers" data-link-type="nav" data-link-position="subnav">{{ _('Mozilla Developers') }}</a></li>
-<li><a href="https://developer.mozilla.org" data-link-name="MDN" data-link-type="nav" data-link-position="subnav">{{ _('MDN') }}</a></li>
+<li>
+  <a href="{{ url('mozorg.developer') }}" data-link-name="Developer Hub" data-link-type="nav" data-link-position="subnav">{{ _('Developer Hub') }}</a>
+</li>
+<li>
+  <a href="https://research.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=dev-tech-sub-nav&amp;utm_content=Research" data-link-name="Research" data-link-type="nav" data-link-position="subnav">{{ _('Research') }}</a>
+</li>
+<li>
+  <a href="https://vr.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=dev-tech-sub-nav&amp;utm_content=VR" data-link-name="Virtual Reality" data-link-type="nav" data-link-position="subnav">{{ _('Virtual Reality') }}</a>
+</li>
+<li>
+  <a href="https://games.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=dev-tech-sub-nav&amp;utm_content=Games" data-link-name="Games" data-link-type="nav" data-link-position="subnav">{{ _('Games') }}</a>
+</li>
+<li>
+  <a href="https://developer.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=dev-tech-sub-nav&amp;utm_content=Documentation" data-link-name="Documentation" data-link-type="nav" data-link-position="subnav">{{ _('Documentation') }}</a>
+</li>
+<li>
+  <a href="https://hacks.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=dev-tech-sub-nav&amp;utm_content=HacksBlog" data-link-name="Hacks Blog" data-link-type="nav" data-link-position="subnav">{{ _('Hacks blog') }}</a>
+</li>
 {% endblock %}
--- a/bedrock/mozorg/templates/mozorg/technology.html
+++ b/bedrock/mozorg/templates/mozorg/technology.html
@@ -29,6 +29,8 @@
 {% endblock %}
 
 {% block content %}
+{% include 'mozorg/developer/includes/sub-nav.html' %}
+
 <main role="main">
   {% block page_hero %}
   <div class="main-heading-container">
--- a/media/css/mozorg/technology.scss
+++ b/media/css/mozorg/technology.scss
@@ -4,6 +4,7 @@
 
 @import '../pebbles/includes/lib';
 @import '../pebbles/components/newsletter';
+@import '../hubs/sub-nav';
 
 body {
     border-top: none;


bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
```diff
@@ -7,7 +7,26 @@
...
 
 {% block sub_nav_primary_links %}
-<li><a href="{{ url('mozorg.technology') }}" data-link-name="Web Innovations" data-link-type="nav" data-link-position="subnav">{{ _('Web Innovations') }}</a></li>
-<li><a href="{{ url('mozorg.developer') }}" data-link-name="Mozilla Developers" data-link-type="nav" data-link-position="subnav">{{ _('Mozilla Developers') }}</a></li>
-<li><a href="https://developer.mozilla.org" data-link-name="MDN" data-link-type="nav" data-link-position="subnav">{{ _('MDN') }}</a></li>
+<li>
+  {% if 'technology' in request.path %}
+    <a href="{{ url('mozorg.developer') }}" data-link-name="Developer Hub" data-link-type="nav" data-link-position="subnav">{{ _('Developer Hub') }}</a>
+  {% else %}
+    <a href="{{ url('mozorg.technology') }}" data-link-name="Technology Hub" data-link-type="nav" data-link-position="subnav">{{ _('Technology Hub') }}</a>
```
alexgibson: Not quite sure where this request came from, but fwiw I thin the subnav shold probably be consistent across all pages. `Technology Hub` isn't a term we use either (the whole `hub` word isn't really used publically).

bedrock/mozorg/templates/mozorg/technology.html
```diff
@@ -29,6 +29,7 @@
 {% endblock %}
 
 {% block content %}
+{% include 'mozorg/developer/includes/sub-nav.html' %}
```
alexgibson: I believe this page is localized, so it may need some for of l10n tag (cc @flodolo)

alexgibson: The sub nav on this page also isn't sticky, so it's missing some JS.

alexgibson:
Also noticed a small typo in the commit message: `changes to secondary nav doe /developer and`

bedrock/mozorg/templates/mozorg/technology.html
schalkneethling: Ah yes, I did not specifically notice that but, was wondering whether there was some JS required. Will add in what is missing.

bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
alexgibson: `/technology/` still has a link on every page as it's a primary nav item, so two links are maybe a bit redundant. See what they say, but I honestly got confused when I loaded things up before reading the PR :)

alexgibson:
Oh I almost forgot, we shouls also make sure these new sub-links are reflected in the global nav side drawer, so people get the same experience on mobile.

bedrock/mozorg/templates/mozorg/technology.html
schalkneethling: Ok so, maybe I should create a new one called `dev-tech-sub-nav.html`. Then I can do:

```
 new subnav
{% else %}
  current nav
{% endif %}
```

bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
schalkneethling: :) Yeah, let's see what the feedback there is. Quick to revert that change.

bedrock/mozorg/templates/mozorg/technology.html
alexgibson: Not quite sure i follow, why would be need a new sub nav? I think an l10n tag should be enough, but I'll defer to @flodolo

schalkneethling:
> Oh I almost forgot, we should also make sure these new sub-links are reflected in the global nav side drawer, so people get the same experience on mobile.

@alexgibson So there is an entry for Technology in the drawer but not Developer so, two questions:

1) Should I place the links under the technology drawer section only or
2) Not add it there, and create a new entry called Developer?

alexgibson:
> So there is an entry for Technology in the drawer but not Developer so, two questions:

The link to `/technology` is in there since the primary nav links are now visible on mobile. What follows should just match the sub nav.

bedrock/mozorg/templates/mozorg/technology.html
schalkneethling: ah ok, I might misunderstand what an l10n flag is :-/ I was thinking we might want to show the new navigation items only to `en-*` until it has been localized in enough locales. While it is being localized, we show the current entries to everyone else.

bedrock/base/templates/includes/global-nav.html
```diff
@@ -248,14 +248,22 @@ <h3 class="summary" data-id="technology">
...
+                    <a href="{{ url('mozorg.developer') }}" data-link-name="Developer Hub" data-link-type="nav" data-link-position="subnav">{{ _('Developer Hub') }}</a>
                 </li>
                 <li>
-                  <a href="https://developer.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=MDN" data-link-type="nav" data-link-name="MDN" data-link-group="Technology">
-                    <abbr title="{{ _('Mozilla Developer Network') }}">{{ _('MDN') }}</abbr>
-                  </a>
+                  <a href="https://research.mozilla.org/" data-link-name="Research" data-link-type="nav" data-link-position="subnav">{{ _('Research') }}</a>
+                </li>
+                <li>
+                  <a href="https://vr.mozilla.org/" data-link-name="Virtual Reality" data-link-type="nav" data-link-position="subnav">{{ _('Virtual Reality') }}</a>
```
schalkneethling: Just notices this does not need the `data-location` but does need the `data-group` attribute

bedrock/mozorg/templates/mozorg/developer/index.html
```diff
@@ -31,7 +31,13 @@
 {% endblock %}
 
 {% block content %}
-{% include 'mozorg/developer/includes/sub-nav.html' %}
+
+{% if LANG.startswith('en-') %}
+  {% include 'mozorg/developer/includes/dev-tech-sub-nav.html' %}
```
alexgibson: I'm not entirely sure we need a second nav (afaik the developer page is not localized yet).

bedrock/mozorg/templates/mozorg/technology.html
```diff
@@ -29,6 +29,13 @@
 {% endblock %}
 
 {% block content %}
+
+{% if LANG.startswith('en-') %}
+  {% include 'mozorg/developer/includes/dev-tech-sub-nav.html' %}
+{% else %}
+  {% include 'mozorg/developer/includes/sub-nav.html' %}
```
alexgibson: This page is localized, so a fallback like this probably isn't going to work (since this page does not contain the strings for the existing sub nav). I'd like to get @flodolo's input here, but an l10n tag is probably the easiest solution.

bedrock/mozorg/templates/mozorg/developer/includes/dev-tech-sub-nav.html
```diff
@@ -0,0 +1,32 @@
...
+
+{% block sub_nav_class %}technology{% endblock %}
+
+{% block sub_nav_primary_links %}
+<li>
+  {% if 'technology' in request.path %}
+    <a href="{{ url('mozorg.developer') }}" data-link-name="Developer Hub" data-link-type="nav" data-link-position="subnav">{{ _('Developer Hub') }}</a>
+  {% else %}
+    <a href="{{ url('mozorg.technology') }}" data-link-name="Technology Hub" data-link-type="nav" data-link-position="subnav">{{ _('Technology Hub') }}</a>
+  {% endif %}
```
schalkneethling: We may, or may not use this.

bedrock/base/templates/includes/global-nav.html
```diff
@@ -248,14 +248,22 @@ <h3 class="summary" data-id="technology">
...
-                  <a href="{{ url('mozorg.developer') }}" data-link-type="nav" data-link-name="Mozilla Developers" data-link-group="Technology">
-                    {{ _('Mozilla Developers') }}
-                  </a>
+                    <a href="{{ url('mozorg.developer') }}" data-link-name="Developer Hub" data-link-type="nav" data-link-position="subnav">{{ _('Developer Hub') }}</a>
                 </li>
                 <li>
-                  <a href="https://developer.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=MDN" data-link-type="nav" data-link-name="MDN" data-link-group="Technology">
-                    <abbr title="{{ _('Mozilla Developer Network') }}">{{ _('MDN') }}</abbr>
-                  </a>
+                  <a href="https://research.mozilla.org/" data-link-name="Research" data-link-type="nav" data-link-position="subnav">{{ _('Research') }}</a>
```
alexgibson: We should probably make sure all these outbound links retain the same utm params that we're used previously here.

bedrock/mozorg/templates/mozorg/developer/index.html
```diff
@@ -31,7 +31,13 @@
...
+{% else %}
+  {% include 'mozorg/developer/includes/sub-nav.html' %}
+{% endif %}
```
schalkneethling: We may, or may not use this.

bedrock/mozorg/templates/mozorg/technology.html
```diff
@@ -29,6 +29,13 @@
...
+{% endif %}
```
schalkneethling: We may, or may not use this.

bedrock/base/templates/includes/global-nav.html
schalkneethling: Good catch Alex

bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
alexgibson: If we do keep this link, we should probably make it consistent with the link above and just use "Technology"

bedrock/mozorg/templates/mozorg/technology.html
schalkneethling: That is very true, hmmmm. So what is the l10n tag you are referring to? I probably even know, just need a refresher.

alexgibson: Sorry my bad, I should have explained a bit further. One similar to this https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/new/scene1.html#L103-L107

schalkneethling: So, that makes more sense but, would we not need the additional file anyhow? The reason I say that is because, the /developer page currently has a subnav while the technology page does not.

So, for the /developer page I can do a if l10n tag show the new subnav else the old but, for /technology there is no current subnav so, it should be if l10n tag include the new subnav else nothing.

schalkneethling:
@alexgibson ok, updated.

flodolo:
Can someone verify if this actually affects l10n? I've lost track of the navigation we're using, given the number of en-US experiments going on
https://www-dev.allizom.org/de/technology/

If this is a navigation shared across several pages, it can't be beyond a tag, and needs to be placed into a shared file (and not exposed for a while).

P.S. Per IRC discussion from yesterday, pings should go to @peiying2 ;-)

bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
```diff
@@ -7,7 +7,26 @@
...
-<li><a href="https://developer.mozilla.org" data-link-name="MDN" data-link-type="nav" data-link-position="subnav">{{ _('MDN') }}</a></li>
+<li>
+  {% if 'technology' in request.path %}
+    <a href="{{ url('mozorg.developer') }}" data-link-name="Developer Hub" data-link-type="nav" data-link-position="subnav">{{ _('Developer Hub') }}</a>
+  {% else %}
+    <a href="{{ url('mozorg.technology') }}" data-link-name="Technology Hub" data-link-type="nav" data-link-position="subnav">{{ _('Technology') }}</a>
+  {% endif %}
+</li>
+<li>
+  <a href="https://research.mozilla.org/" data-link-name="Research" data-link-type="nav" data-link-position="subnav">{{ _('Research') }}</a>
```
alexgibson: I do wonder if we should also have some utm params for these external links in the secondary nav? (the "support" link to SUMO in the Firefox sub nav has some as an example).

bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
```diff
@@ -7,7 +7,26 @@
 {% block sub_nav_class %}technology{% endblock %}
 
 {% block sub_nav_primary_links %}
-<li><a href="{{ url('mozorg.technology') }}" data-link-name="Web Innovations" data-link-type="nav" data-link-position="subnav">{{ _('Web Innovations') }}</a></li>
-<li><a href="{{ url('mozorg.developer') }}" data-link-name="Mozilla Developers" data-link-type="nav" data-link-position="subnav">{{ _('Mozilla Developers') }}</a></li>
-<li><a href="https://developer.mozilla.org" data-link-name="MDN" data-link-type="nav" data-link-position="subnav">{{ _('MDN') }}</a></li>
+<li>
+  {% if 'technology' in request.path %}
```
alexgibson: I know we're waiting on some feedback, but flagging that we may not need this logic

alexgibson:
> P.S. Per IRC discussion from yesterday, pings should go to @peiying2 ;-)

Sorry, I missed this aspect of the convo, my bad

alexgibson:
> If this is a navigation shared across several pages, it can't be beyond a tag, and needs to be placed into a shared file (and not exposed for a while).

Ok, so @schalkneethling it sounds like we need to drop the l10n tag approach. This change may just have to sit and wait a few weeks while localization happens

bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
schalkneethling: And they also not have in the drawer, perhaps I should use the same here? but perhaps keep the `data-link-position="subnav"` here?

bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
schalkneethling: Having seen that there is a link in the primary nav now, I agree that this is not needed but, yeah, let's wait on feedback in the bug to be certain.

bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
alexgibson: Both the sub nav and the drawer have specific data attributes, so I would just apply the same ones that you see in existing examples.

bedrock/mozorg/templates/mozorg/developer/includes/sub-nav.html
```diff
@@ -7,7 +7,26 @@
...
+    <a href="{{ url('mozorg.developer') }}" data-link-name="Developer Hub" data-link-type="nav" data-link-position="subnav">{{ _('Developer Hub') }}</a>
+  {% else %}
+    <a href="{{ url('mozorg.technology') }}" data-link-name="Technology Hub" data-link-type="nav" data-link-position="subnav">{{ _('Technology') }}</a>
```
alexgibson: data-link-name="Technology"

alexgibson:
So, it looks like there are still a few updates outstanding here, but this is close. I'd suggest double checking all the strings in the sub navigation so we can tell @peiying2 if things are ready to extract or not.

bedrock/base/templates/includes/global-nav.html
```diff
@@ -248,14 +248,22 @@ <h3 class="summary" data-id="technology">
...
+                  <a href="https://research.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=Research" data-link-type="nav" data-link-name="Research" data-link-group="Technology">{{ _('Research') }}</a>
+                </li>
+                <li>
+                  <a href="https://vr.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=VR" data-link-type="nav" data-link-name="Virtual Reality" data-link-group="Technology">{{ _('Virtual Reality') }}</a>
+                </li>
+                <li>
+                  <a href="https://games.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=Games" data-link-type="nav" data-link-name="Games" data-link-group="Technology">{{ _('Games') }}</a>
+                </li>
+                <li>
+                  <a href="https://developer.mozilla.org/?utm_source=www.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=global-nav&amp;utm_content=MDN" data-link-type="nav" data-link-name="Documentation" data-link-group="Technology">{{ _('Documentation') }}</a>
```
alexgibson: Should this be `utm_content=Documentation`?

schalkneethling:
Thanks @alexgibson - updated

peiying2:
@schalkneethling do you have a test link for me to check where the strings show up?

schalkneethling:
> do you have a test link for me to check where the strings show up?

@peiying2 I will push this to a demo and let you know the URL.

alexgibson:
@peiying2 is this PR good to merge yet?

peiying2:
@alexgibson good to merge from l10n perspective.

PR #5151, opened 2017-09-28
Add better cache invalidation to release notes

pmac:
Use the etag returned by nucleus in the cache keys for the release objects cache. This allows us to set long expire times while keeping fast invalidation by checking the etag file once per minute.


pmac:
Interested in your thoughts on this @glogiotatidis and/or @jgmize. Since we're storing a hash of the data in the repo (could also have used the git hash for HEAD) I figured it'd be a good way to invalidate the cache and give us the ability to cache for much longer during the normal periods when the data isn't changing.

pmac:
Hmm... tests failing because the redirect tests check the `*/notes/` redirects and the release note data isn't there during the test run... not sure what to  do about that one. Will have to think more about it next week.

pmac:
This should really be two PRs, but I wanted to start on the "release notes data for latest notes redirects" and I needed the improved caching. So they're separate commits, but one PR. Apologies for the size...

pmac:
@glogiotatidis I've fixed and added tests. I decided to remove the redirect tests for the e.g. `/firefox/notes/` redirects since they rely on external data and the views for those are well tested in the unit tests. Those are already well mocked, so adding more mocked tests doesn't seem useful.

bedrock/releasenotes/models.py
```diff
@@ -114,15 +115,30 @@ def version_obj(self):
...
-        if self.product == 'Firefox for iOS':
-            return reverse('firefox.ios.releasenotes', args=(self.version, prefix))
+            urlname = 'firefox.android.releasenotes'
+        elif self.product == 'Firefox for iOS':
+            urlname = 'firefox.ios.releasenotes'
+        else:
+            urlname = 'firefox.desktop.releasenotes'
+
+        prefix = 'aurora' if self.channel == 'Aurora' else 'release'
+        return reverse(urlname, args=(self.version, prefix))
```
glogiotatidis: code style nit: you use lists instead of tuples of args elsewhere.

bedrock/releasenotes/models.py
```diff
@@ -270,10 +286,45 @@ def get_release_or_404(version, product):
...
+def get_all_releases(product, channel='release'):
     file_prefix = get_file_id(product, channel, '*')
-    releases = glob(os.path.join(release_notes_path(), file_prefix + '.*'))
+    cache_key = get_cache_key('all:%s:%s' % (product, channel))
+    releases = cache.get(cache_key)
+    if not releases:
+        releases = glob(os.path.join(release_notes_path(), file_prefix + '.json'))
+        if releases:
+            releases = (get_release_from_file(r) for r in releases)
+            releases = sorted((r for r in releases if r.is_public),
```
glogiotatidis: 👍 great use of generators

bedrock/releasenotes/views.py
```diff
@@ -217,25 +180,23 @@ def releases_index(request, product):
...
-                    note.version = release.version
-                    notes[note.id] = note
+            if note.is_public and note.tag:
+                note.link = link + '#note-' + str(note.id)
+                note.version = release.version
+                notes[note.id] = note
 
     # Sort by date in descending order
-    notes = sorted(notes.values(), key=lambda x: x.modified, reverse=True)
+    notes = sorted(notes.values(), key=attrgetter('modified'), reverse=True)
```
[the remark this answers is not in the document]
pmac: `itemgetter` is also great :)

bedrock/releasenotes/tests/test_models.py
```diff
@@ -160,22 +160,24 @@ def test_get_release_none_load(self, grff_mock, grfn_mock):
...
-    def test_get_release_from_file_no_cache(self, grffs_mock, cache_mock):
+    def test_get_release_from_file_no_cache(self, grffs_mock, cache_key_mock, cache_mock):
         cache_mock.get.return_value = None
         grffs_mock.return_value = 'donnie'
         assert models.get_release_from_file('walter') == 'donnie'
-        cache_mock.get.assert_called_with('walter')
+        cache_key_mock.assert_called_with('walter')
         grffs_mock.assert_called_with('walter')
-        cache_mock.set.assert_called_with('walter', 'donnie', 300)
+        cache_mock.set.assert_called_with(cache_key_mock(), 'donnie', 7200)
```
glogiotatidis: Maybe use `models.LONG_RN_CACHE_TIMEOUT`

PR #5152, opened 2017-09-28
Fix exception when equavalent releases exist but none are public

pmac:
Was getting IndexError trying to get the first item in an empty list because we were not checking for an empty list after filtering out non-public releases in the equivalent_release_for_product method.

### DIFF
--- a/bedrock/releasenotes/tests/test_models.py
+++ b/bedrock/releasenotes/tests/test_models.py
@@ -68,10 +68,18 @@ class TestReleaseModel(TestCase):
         android = rel.equivalent_release_for_product('Firefox for Android')
         assert android is None
 
+    @patch.object(models, 'get_release_from_file')
+    def test_equivalent_release_for_product_no_public_match(self, grff_mock):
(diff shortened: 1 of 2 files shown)


bedrock/releasenotes/tests/test_models.py
```diff
@@ -68,10 +68,18 @@ def test_equivalent_release_for_product_none_match(self):
...
+        rel = models.Release(dict(product='Firefox', version='56.0', is_public=True, channel='Release'))
+        grff_mock.return_value = models.Release(dict(product='Firefox for Android', is_public=False, version='56.0.2', channel='Release'))
+        android = rel.equivalent_release_for_product('Firefox for Android')
+        assert android is None
+
     def test_note_fixed_in_release(self):
         rel = models.get_release('firefox', '55.0a1')
         note = rel.notes[11]
-        assert note.fixed_in_release.get_absolute_url() == '/en-US/firefox/55.0a1/releasenotes/'
+        with self.activate('en-US'):
```
pmac: Also fixed this flaky test.

PR #5161, opened 2017-09-29
[fix bug 1404444] Quantum page overflow

craigcook:
## Description
Although overflow should have been hidden it's still possible to move the viewport with some methods of scrolling because of the super-wide wave image. This sets that pseudo-element to 100% width to avoid overflow, and updates the SVG to correct the offset.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1404444

## Testing
If your mouse has a scrollwheel you can probably middle-click to get a 2D drag-scrolly widget thing (no idea what that's called). It should only move vertically, not horizontally. Compare to production to see the bug.


dholbert:
> If your mouse has a scrollwheel you can probably middle-click to get a 2D drag-scrolly widget thing (no idea what that's called

That feature is called "autoscroll", BTW. (Enabled by default on Windows, disabled on other platforms.)  It's not documented well, but there's a bit of info about it at https://askubuntu.com/questions/908/firefox-middle-mouse-button-scroll .

Thanks for the fix!

alexgibson:
Has this fix altered the offset of the wave? Notice it wasn’t exactly centered. There are also other waves further down fwiw

craigcook:
> Has this fix altered the offset of the wave? Notice it wasn’t exactly centered. There are also other waves further down fwiw

Yeah, I adjusted the SVG itself to offset it when centered. 

The other waves don't seem to trigger the bug, so maybe the real problem was overflow on just the intro section... may warrant further investigation.

craigcook:
Oh, this is embarassing. The bug is on the 56 whatsnew page, not the Quantum page. I fixed this in the wrong place 🤦‍♂️

PR #5165, opened 2017-10-02
Fix Bug 1404852, updates Fx Cliqz notice to match new Fx policy design

schalkneethling:
## Description
Changes Firefox+Cliqz privacy notice to match the new Firefox privacy notice. Page can be found at:
https://bedrock-demo-privacy.us-west.moz.works/en-US/privacy/firefox-cliqz/

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1404852

## Testing
Ensure that for locales that are not localized, this falls back to the current layout and copy for that lang. Although, I believe for this one specifically, there is only en-US and de


bedrock/privacy/templates/privacy/notices/firefox-cliqz-quantum.html
```diff
@@ -0,0 +1,72 @@
...
+{% set header = doc.body.section.extract() %}
+{% set lead_in = doc.body.section.extract() %}
+{% set footnote = doc.body.select('#footnote') %}
+{% if footnote %}
+    {% set footnote = footnote[0].extract() %}
+{% endif %}
+
+<header class="privacy-head">
+  <div class="privacy-head-wrapper">
+    <h1 itemprop="name">{{ header.select('h2 span[class="privacy-header-firefox"]')|join|safe }}
```
alexgibson: Shouldn't the header read `Firefox + Cliqz Privacy Notice`? (like the page title)

bedrock/privacy/templates/privacy/notices/firefox-cliqz-quantum.html
```diff
@@ -0,0 +1,72 @@
...
+  {{ doc.body.section.h2.get_text() }}
+{% endblock %}
+
+{% block page_css %}
+  {% stylesheet 'firefox_privacy' %}
+{% endblock %}
+
+{% block body_id %}firefox-notice{% endblock %}
+
+{% block string_data %}
```
alexgibson: Rather than add these explicitly for every new template, it may be a better to make a common base tempalte from which all redesigned privacy notices can inherit from? You could apply the same to the CSS/JS bundles which are the same for each page.

bedrock/privacy/views.py
```diff
@@ -80,15 +80,24 @@ def get_legal_doc(self):
         return doc
 
 
+class FirefoxCliqzPrivacyDocView(PrivacyDocView):
+    template_name = 'privacy/notices/firefox-cliqz.html'
+
+    def get_legal_doc(self):
```
alexgibson: Should this method have some sort of unit test?

alexgibson:
Not related to this PR itself, but the sticky header takes up a lot of vertical space. Could we trim the padding a bit maybe?

screenshot-2017-10-2 firefox cliqz privacy notice

(I also don't really see what purpose the sticky header serves tbh, but we'll let it slide as this is what they asked for) 😉

bedrock/privacy/templates/privacy/notices/firefox-cliqz-quantum.html
schalkneethling: Yup, as I opened this I was thinking it would be worth having a new base at this point that this one and the previous one can extend. Will do that.

bedrock/privacy/templates/privacy/notices/firefox-cliqz-quantum.html
schalkneethling: Doh! Indeed.

bedrock/privacy/views.py
schalkneethling: This one actually "implements" the function inherited from `PrivacyDocView` so, there should already be a test for that?

schalkneethling:
> Not related to this PR itself, but the sticky header takes up a lot of vertical space. Could we trim the padding a bit maybe?

I reckon we can do that although, there is currently a comment, that I asked for an expanded response to, where some of the changes we made during the previous code review phase for the Fx page is being questioned. So let's see.

bedrock/privacy/views.py
alexgibson: Hmm, maybe I'm looking in the wrong place - but I can see a test for the original function either?

schalkneethling: @pmac Thoughts? ^^

pmac: It would be best if there were a unit test for both of these. I didn't push it because we manually verified it works and we'll be able to delete these work-arounds once all of the docs in the legal-docs repo are updated with the new format. If you'd like to push for tests for these though I can't really fault that logic. It's a function of how quickly we need this to land really.

alexgibson: I don't mind merging this without one if that is best, but it feels like we probably should have some form of test? Could make it a followup PR if that works 👍

alexgibson: It also feels like we're going to go through this for every Privacy notice, so some form of reusable & testable function may come in handy?

pmac: Hmm... is that right? If that's true then I'm happy to do a followup PR to clean these up and add a reusable and tested base class for these.

schalkneethling: Yes I agree. As more of these are converted, there will be more and more that needs this same kinda fallback as some locales are being localized so, having a base class would be great. Thanks @pmac 

I am waiting for the PR on legal-docs to be merged and, for the relevant people to sign of on the page. As soon as I have those, I can remove the DNM label from this one.

bedrock/privacy/templates/privacy/base.html
```diff
@@ -0,0 +1,35 @@
...
+  data-tabpanel-close-text="{{ _('Hide All') }}"
+  data-choices-copy="{{ _('Choose how you want to share this data in Firefox.') }}"
+  data-choices-button="{{ _('Choose…') }}"
+{% endblock %}
+
+{% set header = doc.body.section.extract() %}
+{% set lead_in = doc.body.section.extract() %}
+{% set footnote = doc.body.select('#footnote') %}
+{% if footnote %}
+    {% set footnote = footnote[0].extract() %}
```
alexgibson: Nit: indentation

schalkneethling:
Thanks @alexgibson ~ updated.

PR #5170, opened 2017-10-02
Fix bug 1405107: Add an en-US entry to alternet hreflang tags

pmac:
Hoping this helps. Does seem like an omission.


pmac:
Yes. It's a bit of a stab in the dark. I talked to Justin about this yesterday. Basically nothing from www.mozilla.org shows up on any page of en-US google search results for basic terms like "browser". This is very strange. We're trying to figure out why by fixing some things that seem incorrect.

alexgibson:
> Basically nothing from www.mozilla.org shows up on any page of en-US google search results for basic terms like "browser". 

Weird. This does feel like a shot in the dark, but I don't see how it could do harm.

PR #5176, opened 2017-10-04
Fix bug 1405436: Add redirects for a lot of 404s with links in the wild

pmac:


bedrock/redirects/middleware.py
```diff
@@ -9,6 +9,7 @@ def __init__(self, resolver=None):
 
     def process_request(self, request):
         try:
+            print request.path_info
```
jpetto: 🖨

pmac: DOH! Great catch!

pmac:
Errant print statement removed. Thanks again Jon!

PR #5180, opened 2017-10-05
Bug 1395342: Configure send-to-device SMS countries per message-set

pmac:
Not all messages will be available in the same list of countries.  This ensures we're able to provide the best experience by showing the SMS version only to people in the countries for which we have an SMS message that works for them.

### DIFF
--- a/bedrock/base/templatetags/helpers.py
+++ b/bedrock/base/templatetags/helpers.py
@@ -15,8 +15,12 @@ from bedrock.utils import expand_locale_groups
 
 
 @library.global_function
-def send_to_device_countries():
-    return '|%s|' % '|'.join(cc.lower() for cc in settings.SEND_TO_DEVICE_COUNTRIES)
+def send_to_device_sms_countries(message_set):
+    try:
+        countries = settings.SEND_TO_DEVICE_MESSAGE_SETS[message_set]['sms_countries']
+    except KeyError:
+        countries = ['US']
+    return '|%s|' % '|'.join(cc.lower() for cc in countries)
 
 
 @library.global_function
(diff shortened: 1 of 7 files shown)


bedrock/base/templatetags/helpers.py
```diff
@@ -15,8 +15,12 @@
 
 
 @library.global_function
-def send_to_device_countries():
-    return '|%s|' % '|'.join(cc.lower() for cc in settings.SEND_TO_DEVICE_COUNTRIES)
+def send_to_device_countries(message_set):
```
jpetto: As the logic is now getting a bit complex between locales and countries, email and SMS, I wonder if we should rename this to `send_to_device_sms_countries`?

I don't want to blow up the scope of this PR, but there are a few other places we could make this distinction a bit clearer as well. Maybe a follow-up PR.

bedrock/settings/base.py
```diff
@@ -1177,7 +1177,50 @@ def lazy_email_backend():
 
 # country code for /country-code.json to return in dev mode
 DEV_GEO_COUNTRY_CODE = config('DEV_GEO_COUNTRY_CODE', default='US')
-SEND_TO_DEVICE_COUNTRIES = config('SEND_TO_DEVICE_COUNTRIES', default='US', cast=Csv())
+SEND_TO_DEVICE_MESSAGE_SETS = {
+    'default': {
+        'sms_countries': config('STD_COUNTRIES_DEFAULT', default='US', cast=Csv()),
```
jpetto: Should we have a public place for referencing which countries support SMS for each message set?

In addition, would be good to update the portion of [the docs](https://github.com/mozilla/bedrock/blob/master/docs/send-to-device.rst) referring to SMS (and GeoDude?):

>The form allows sending via SMS or Email, although the SMS copy & messaging is shown only to those in the US. Geo-location is handled in JavaScript using GeoDude.

pmac: Great catch. I'll update.

pmac: As for public, I was just going to rely on the published configs that are part of the configurator. Hopefully that's enough?

pmac:
Updated function name, env var names, and docs.

PR #5181, opened 2017-10-05
Fix wrong product returned for latest notes

pmac:
Currently for example /firefox/notes/ could redirect to the latest Firefox for Android if it were newer.


pmac:
Thanks!

PR #5185, opened 2017-10-06
[fix bug 1406139] Add unbranded search funnelcake experiment.

jpetto:
## Description

Adds funnelcakes for unbranded search experiment. en-US locale/US geo only, so should not conflict with de only Cliqz funnelcakes.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1406139

## Testing

Ensure experiment works as expected - 30% chance of a funnelcake when using Windows/non-Fx with DNT disabled.

### DIFF
--- a/bedrock/firefox/templates/firefox/new/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/scene1.html
@@ -10,6 +10,10 @@
   {% if switch('firefox-new-cliqz-funnelcakes', ['de']) %}
     {% javascript 'experiment_firefox_new_cliqz_funnelcakes' %}
   {% endif %}
+
+  {% if switch('firefox-new-unbranded-search-funnelcakes', ['en-US']) %}
+    {% javascript 'experiment_firefox_new_unbranded_search_funnelcakes' %}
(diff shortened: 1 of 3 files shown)


jpetto:
This should not go out until 56.0.1 is released, so **DO NOT MERGE** for now. Should be good for code review though.

(It's behind a switch, so we could merge early, but I'd rather hold back in the event a change is requested between now and 56.0.1 release.)

schalkneethling:
> 30% chance of a funnelcake when using Windows/non-Fx with DNT disabled.

Excuse my lack of understanding here ;p but, from what I can see in the code here, there is nothing that would prevent this from executing in Firefox. The line above suggests that this should only run on Windows and non Firefox browsers. @jpetto

jpetto:
Ah, yeah, much of the logic is buried in the [mozilla-traffic-cop-funnelcake-geo.js](https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-traffic-cop-funnelcake-geo.js#L68) library.

Should probably write some docs for that lib...

PR #5188, opened 2017-10-09
Fix Bug 1387466 - Adjust download page strings to remove "(incl. Win 64)" now that Win64 builds are default on all channels

kyoshino:
## Description

Update the links from the [channel](https://www.mozilla.org/en-US/firefox/channel/desktop/) page to /all/ pages since the 64-bit version is now default. Also use the term "platforms" instead of "builds" to be consistent with the [/new/](https://www.mozilla.org/en-US/firefox/new/) page. So this basically replaces "**All Languages and Builds (incl. Win 64)**" with "**All Languages and Platforms**" (which I think was the original copy.)

This Requires a l10n update, but we can use "All Languages and Builds" as the fallback string, which has already been used on the Android channel page.

## Issue / Bugzilla link

[Bug 1387466](https://bugzilla.mozilla.org/show_bug.cgi?id=1387466)

## Testing

Just visit the channel page and make sure the new link copy is displayed.

### DIFF
--- a/bedrock/firefox/templates/firefox/channel/android.html
+++ b/bedrock/firefox/templates/firefox/channel/android.html
@@ -52,7 +52,7 @@
     <div class="container">
       <ul>
         <li><a href="{{ firefox_url('android', 'notes', 'beta') }}">{{_('Release Notes')}}</a></li>
-        <li><a href="{{ firefox_url('android', 'all', 'beta') }}">{{_('All Languages and Builds')}}</a></li>
+        <li><a href="{{ firefox_url('android', 'all', 'beta') }}">{% if l10n_has_tag('firefox_channel_bug1387466') %}{{_('All Languages and Platforms')}}{% else %}{{_('All Languages and Builds')}}{% endif %}</a></li>
       </ul>
     </div>
   </footer>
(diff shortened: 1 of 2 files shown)


bedrock/firefox/templates/firefox/channel/android.html
```diff
@@ -52,7 +52,7 @@
     <div class="container">
       <ul>
         <li><a href="{{ firefox_url('android', 'notes', 'beta') }}">{{_('Release Notes')}}</a></li>
-        <li><a href="{{ firefox_url('android', 'all', 'beta') }}">{{_('All Languages and Builds')}}</a></li>
+        <li><a href="{{ firefox_url('android', 'all', 'beta') }}">{% if l10n_has_tag('firefox_channel_bug1387466') %}{{_('All Languages and Platforms')}}{% else %}{{_('All Languages and Builds')}}{% endif %}</a></li>
```
cpeterso: Where will the `firefox_channel_bug1387466` l10n tag be defined? (I don't know how bedrock l10n works.)

kyoshino: It will be manually added to .lang files by @peiying2 or @flodolo [like this](https://github.com/mozilla-l10n/www.mozilla.org/commit/69bcc5f49fbeb8273cb0a929bbaf5e8058fad15b). In English, tags are simply ignored and the primary (not fallback) string will always be used.

PR #5189, opened 2017-10-09
Move release notes back to the database

pmac:
This is prep for our move to using a local DB for bedrock. This also helps a lot with release querying and reduces the code quite a bit, as well as making this more efficient due to the caching built into Django's ORM.


pmac:
This should stay as two commits since it should be deployed in 2 phases. The first commit sets up the model and starts the updater loading the data into the database. The 2nd commit switches out the machinery to use the database instead of the filesystem.

bedrock/releasenotes/models.py
```diff
@@ -188,142 +248,48 @@ def equivalent_desktop_release(self):
             return self.equivalent_release_for_product('Firefox')
 
 
+@memoize(LONG_RN_CACHE_TIMEOUT)
 def get_release(product, version, channel=None):
-    channels = [channel] if channel else Release.CHANNELS
+    print '!!! Getting Release'
```
jgmize: please delete this print statement

PR #5191, opened 2017-10-10
Bug 1395342: Send to device number field not US only

pmac:
This removes the form validation that I failed to remove before.  It also enables the proper message when basket says that the number is invalid.

### DIFF
--- a/bedrock/firefox/forms.py
+++ b/bedrock/firefox/forms.py
@@ -5,47 +5,20 @@
 import re
 
 from django import forms
-from django.core.validators import EMPTY_VALUES
-from django.utils.encoding import smart_text
-
-from lib.l10n_utils.dotlang import _lazy as _
-
-
-class USPhoneNumberField(forms.CharField):
-    """A form field that validates input as a U.S. phone number.
-
-    Note: The default 'invalid' error message serves as a placeholder and is not
-    currently in use or localized.
-    """
-    default_error_messages = {
-        'invalid': _("Sorry. This number isn't valid. Please enter a U.S. phone "
-                     'number or <a href="%s">'
-                     'download directly from Google Play.</a>') % 'http://mzl.la/OgZo6k',
-    }
-
-    def __init__(self, *args, **kwargs):
-        kwargs.setdefault('max_length', 14)
-        super(USPhoneNumberField, self).__init__(*args, **kwargs)
-
-    def clean(self, value):
-        super(USPhoneNumberField, self).clean(value)
-        if value in EMPTY_VALUES:
-            return ''
-
-        value = re.sub(r'\D+', '', smart_text(value))
-        if len(value) == 10:
-            value = '1' + value
-        elif len(value) != 11 or value[0] != '1':
-            raise forms.ValidationError(self.error_messages['invalid'])
-
-        return value
 
 
 class SendToDeviceWidgetForm(forms.Form):
-    number = USPhoneNumberField(required=False)
+    number = forms.CharField(max_length=20, min_length=4, required=False)
     email = forms.EmailField(max_length=100, required=False)
     platform = forms.ChoiceField(choices=(
         ('ios', 'ios'),
         ('android', 'android'),
         ('all', 'all'),
     ), required=False)
+
+    def clean_number(self):
+        number = self.cleaned_data['number']
+        if number:
+            number = re.sub(r'\D+', '', number)
+
+        return number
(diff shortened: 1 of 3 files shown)


bedrock/firefox/forms.py
```diff
@@ -42,7 +42,7 @@ def clean(self, value):
 
 
 class SendToDeviceWidgetForm(forms.Form):
-    number = USPhoneNumberField(required=False)
```
jpetto: Doesn't look like it's used any more - should we keep the `USPhoneNumberField` class around for possible future use or drop it to clean the code base?

bedrock/firefox/tests/test_views.py
```diff
@@ -223,7 +223,7 @@ def test_send_android_sms_with_country(self):
         })
         ok_(resp_data['success'])
         self.mock_send_sms.assert_called_with('post', 'subscribe_sms', data={
-            'mobile_number': '15558675309',
+            'mobile_number': '5558675309',
```
jpetto: Just making sure it's okay to send a number to basket without a country code.

bedrock/firefox/views.py
```diff
@@ -220,9 +220,13 @@ def send_to_device_ajax(request):
 
                 try:
                     basket.request('post', 'subscribe_sms', data=data)
-                except basket.BasketException:
-                    return HttpResponseJSON({'success': False, 'errors': ['system']},
-                                            status=400)
+                except basket.BasketException as e:
+                    if e.desc == 'mobile_number is invalid':
```
jpetto: To clarify, this error will come directly from basket, correct? Could/should we do any validation (e.g. make sure length is sufficient) on our end first to avoid the round trip?

bedrock/firefox/views.py
```diff
@@ -220,9 +220,13 @@ def send_to_device_ajax(request):
...
+                        return HttpResponseJSON({'success': False, 'errors': ['number']},
+                                                status=400)
```
jpetto: Thinking we should change this to `status=200` so we continue to show the "Sorry, we can’t send SMS messages to this phone number. " error message. As is, an invalid number displays the "An error occurred in our system. Please try again later." message, which is really accurate (or helpful).

bedrock/firefox/tests/test_views.py
pmac: Yup. There's new code in basket that interprets the number along with the country code and will format appropriately.

bedrock/firefox/tests/test_views.py
```diff
@@ -198,19 +198,19 @@ def test_send_android_sms(self):
...
             'msg_name': views.SEND_TO_DEVICE_MESSAGE_SETS['default']['sms']['android'],
             'lang': 'en-US',
         })
 
     def test_send_android_sms_non_en_us(self):
         resp_data = self._request({
             'platform': 'android',
-            'phone-or-email': '5558675309',
+            'phone-or-email': '015558675309',
```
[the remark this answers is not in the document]
pmac: Not necessarily. The basket side will format it appropriately for the country, but we're doing geolocation so that we can interpret local numbers as well. I mainly needed this so that I could have a test that made sure that a longer phone number would pass validation.

bedrock/firefox/views.py
pmac: It's really tough to do that now. I can maybe find out what the max length of any phone number in the world is, but there is a max on the form field that will prevent any ridiculous entries.

bedrock/firefox/views.py
pmac: Oh, so the JS will interpret a 400 response as an "error" and send it though the error callback which won't display the "number" message?

pmac: I've fixed this so that the view and test use a 200 response so that the JS will interpret the error codes.

bedrock/firefox/forms.py
pmac: I think you're right. It might be useful in future, but that's why we have git history right :)

bedrock/firefox/views.py
jpetto: I was more curious about numbers being too short - like someone accidentally hit enter after putting in only 3 digits or something.

pmac: Hmm... perhaps... I'll look into it.

pmac:
This is now updated with checks and tests for the number being between 4 and 20 characters, as well as stripping all punctuation from the input before shipping to basket.

PR #5194, opened 2017-10-11
Fix Bug 1407349, expand max-content to prevent choose button from overlapping text

schalkneethling:
## Description
In some locales, notably French, the constricted `max-width` of the content is causing the Choose button to overlap the text. I always felt that `600px` was going to be to narrow, so here I extended it to `700px` and tested `en`, `fr`, `de`, `it` and there is now enough breathing room.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1407349

## Testing
Ensure that the Choose button does not overlap the text in various locales, most notably Franch.

### DIFF
--- a/media/css/privacy/firefox-privacy.scss
+++ b/media/css/privacy/firefox-privacy.scss
@@ -130,7 +130,7 @@ main {
     width: 94%;
 
     @media #{$mq-desktop} {
-        max-width: 600px;
+        max-width: 700px;
     }
 
     h2 {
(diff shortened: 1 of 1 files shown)


alexgibson:
Hmm, this does fix the problem for the French locale - but what if the string is longer again, the bug still comes back? Would setting a variable width on the `<p>` make sense, like 70% so it flows into two lines when needed?

schalkneethling:
> Would setting a variable width on the <p> make sense, like 70% so it flows into two lines when needed?

Definitely a good idea. I will add that and update the PR.

schalkneethling:
@alexgibson Thanks for the review - updated.

PR #5195, opened 2017-10-11
Fix Bug 1405076, firefox landing page layout and content updates

schalkneethling:
## Description
Updates the /firefox page to mirror the current /firefox/quantum with additional copy updates.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1405076

## Testing
Ensure that is switch is not enabled the current content hub version of the page is shown.

### DIFF
--- a/bedrock/firefox/templates/firefox/firefox-quantum.html
+++ b/bedrock/firefox/templates/firefox/firefox-quantum.html
@@ -1,0 +1,170 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/base-pebbles.html" %}
+
+{% block page_css %}
+  {% stylesheet 'firefox-quantum-landing' %}
+{% endblock %}
+
+{% block page_title %}{{ _('2x faster and 30% less memory') }} | {{ _('Firefox Browser') }}{% endblock %}
+{% block page_desc %}{{ _('Download now to start browsing the fastest Firefox ever.') }}{% endblock %}
+{% block page_image %}{{ static('img/firefox/template/page-image-quantum.png') }}{% endblock %}
+
+{% block content %}
+{% include 'firefox/includes/hub/sub-nav.html' %}
+<main role="main">
+  <header class="main-header" data-scroll-tracking="Main Header">
+    <div class="header-image">
+      {{ high_res_img('firefox/quantum/new-tab.png', {'alt': '', 'width': '922', 'height': '662'}) }}
+    </div>
+    <div class="content">
+      <div class="header-content">
+        <h1>{{ _('<strong>Firefox</strong> Quantum') }}</h1>
+        <h2>{{ _('New. Fast. For good.') }}</h2>
+        <div class="cta-wrapper">
+          {{ download_firefox(alt_copy=_('Download Now'), dom_id='firefox-desktop-page-primary', download_location='primary cta', button_color='quantum-download') }}
+        </div>
+      </div>
+    </div>
+  </header>
+  <div class="key-features-section main" data-scroll-tracking="First Key Features">
+    <div class="content">
+      <section class="key-feature fast">
+        <div class="key-feature-content">
+          <h2>{{ _('2x faster') }}</h2>
+          <p>
+          {% trans url='https://blog.mozilla.org/firefox-de/firefox-quantum-superschnell-und-schont-den-speicher/' if LANG == 'de' else 'https://blog.mozilla.org/firefox/quantum-performance-test/' %}
+            Powered by a new, cutting-edge engine, Firefox has <a href="{{ url }}">doubled its speed</a>
+            from last year. Because the Internet waits for no one.
+          {% endtrans %}
+          </p>
+        </div>
+      </section>
+      <section class="key-feature performance">
+        <div class="key-feature-content">
+          <h2>{{ _('Lean, mean speed machine') }}</h2>
+          <p>
+          {% trans url='https://blog.mozilla.org/firefox-de/firefox-quantum-superschnell-und-schont-den-speicher/' if LANG == 'de' else 'https://blog.mozilla.org/firefox/quantum-performance-test/' %}
+            Firefox Quantum’s new engine uses <a href="{{ url }}">30% less memory</a> than Chrome,
+            so other programs won’t slow down during browsing. Now that’s a win-win.
+          {% endtrans %}
+          </p>
+        </div>
+      </section>
+      <section class="key-feature design">
+        <h2>{{ _('Beautiful, intelligent design') }}</h2>
+        <div class="feature-carousel-container">
+          <ul class="feature-carousel">
+            <li class="video-bookmarking">
+              <figure>
+                <video preload="metadata" muted controls poster="{{ static('img/firefox/quantum/carousel/placeholder-bookmarks.png') }}">
+                  <source src="https://assets.mozilla.net/video/quantum/bookmarking.webm" type="video/webm">
+                  <source src="https://assets.mozilla.net/video/quantum/bookmarking.ogv" type="video/ogg; codecs='theora, vorbis'">
+                  <source src="https://assets.mozilla.net/video/quantum/bookmarking.mp4" type="video/mp4">
+                  <img src="{{ static('img/firefox/quantum/carousel/placeholder-bookmarks.png') }}" alt="">
+                </video>
+                <figcaption>{{ _('Bookmarks + Pocket') }}</figcaption>
+              </figure>
+            </li>
+            <li class="video-new-tab">
+              <figure>
+                <video preload="metadata" muted controls poster="{{ static('img/firefox/quantum/carousel/placeholder-new-tab.png') }}">
+                  <source src="https://assets.mozilla.net/video/quantum/new-tab.webm" type="video/webm">
+                  <source src="https://assets.mozilla.net/video/quantum/new-tab.ogv" type="video/ogg; codecs='theora, vorbis'">
+                  <source src="https://assets.mozilla.net/video/quantum/new-tab.mp4" type="video/mp4">
+                  <img src="{{ static('img/firefox/quantum/carousel/placeholder-new-tab.png') }}" alt="">
+                </video>
+                <figcaption>{{ _('New Tab') }}</figcaption>
+              </figure>
+            </li>
+            <li class="video-screenshots">
+              <figure>
+                <video preload="metadata" muted controls poster="{{ static('img/firefox/quantum/carousel/placeholder-screenshots.png') }}">
+                  <source src="https://assets.mozilla.net/video/quantum/screenshots.webm" type="video/webm">
+                  <source src="https://assets.mozilla.net/video/quantum/screenshots.ogv" type="video/ogg; codecs='theora, vorbis'">
+                  <source src="https://assets.mozilla.net/video/quantum/screenshots.mp4" type="video/mp4">
+                  <img src="{{ static('img/firefox/quantum/carousel/placeholder-screenshots.png') }}" alt="">
+                </video>
+                <figcaption>{{ _('Screenshots') }}</figcaption>
+              </figure>
+            </li>
+          </ul>
+          <button class="feature-carousel-previous" type="button">{{ _('Previous') }}</button>
+          <button class="feature-carousel-next" type="button">{{ _('Next') }}</button>
+          <p>
+          {% trans %}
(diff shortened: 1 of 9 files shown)


bedrock/firefox/templates/firefox/firefox-quantum.html
```diff
@@ -0,0 +1,169 @@
...
+  <header class="main-header" data-scroll-tracking="Main Header">
+    <div class="header-image">
+      {{ high_res_img('firefox/quantum/new-tab.png', {'alt': '', 'width': '922', 'height': '662'}) }}
+    </div>
+    <div class="content">
+      <div class="header-content">
+        <h1>{{ _('<strong>Firefox</strong> Quantum') }}</h1>
+        <h2>{{ _('New. Fast. For good.') }}</h2>
+        <div class="cta-wrapper">
+          {{ download_firefox(alt_copy=_('Download Now'), dom_id='firefox-desktop-page', download_location='firefox-header') }}
```
alexgibson: I think this should be `download_location='primary cta'`

bedrock/firefox/templates/firefox/firefox-quantum.html
```diff
@@ -0,0 +1,169 @@
...
+          </p>
+        </div>
+      </section>
+    </div>
+  </div>
+  <section class="secondary-download-cta" data-scroll-tracking="Secondary Download CTA">
+    <div class="content">
+      <div class="secondary-download-cta-content">
+        <h2>{{ _('Start browsing with <strong>Firefox</strong> Quantum') }}</h2>
+        {{ download_firefox(alt_copy=_('Download Now'), dom_id='firefox-desktop-page', download_location='firefox-secondary') }}
```
alexgibson: This should be `download_location='other'`

bedrock/firefox/views.py
```diff
@@ -737,11 +738,14 @@ class FirefoxFocusView(BlogPostsView):
 
 
 class FirefoxHubView(BlogPostsView):
-    blog_posts_limit = 1
-    blog_posts_template_variable = 'articles'
-    blog_slugs = 'firefox'
-    blog_tags = ['home']
-    template_name = 'firefox/hub/home.html'
+    if switch('firefox-57-release'):
```
alexgibson: It would be nice to have some unit tests that ensure the correct template is loaded for both states of the switch

media/css/firefox/quantum-landing.scss
```diff
@@ -0,0 +1,805 @@
...
+
+#outer-wrapper {
+    overflow-x: hidden;
+}
+
+h1, h2, h3, h4, h5, h6 {
+    font-weight: normal;
+}
+
+a.download-link.button.button-green {
```
alexgibson: Rather than overriding styles, it would be nice here to make a new button style and `@import` it into the stylesheet. Similar to what we've done [here](https://github.com/alexgibson/bedrock/blob/bug-1397381-quantum-download-page-redesign/media/css/firefox/new/quantum/_buttons.scss#L10-L31). Give it a button name like `quantum`. This will make it easier for us to consolidate styles between pages in the future.

schalkneethling:
> My biggest question on the page is that we're missing the Firefox sub-navigation?
@alexgibson That is my bad, I completely forgot that you have to explicitly import it to the page. Will fix that.

media/css/firefox/quantum-landing.scss
schalkneethling: So, the reason I did the override is because I am using the `download_firefox` helper. Now, although you cannot specify a custom class, I see there is a parameter called `button_color` with a default value of `button-green` so, I guess I can use that as a way to specify a custom class, say

`{{ download_firefox(alt_copy=_('Download Now'), dom_id='firefox-desktop-page', download_location='other', button_color='quantum-download') }}`

bedrock/firefox/tests/test_views.py
```diff
@@ -754,6 +754,22 @@ def test_get_context_data_one_star(self):
         self.assertFalse('donate_stars_url' in ctx)
 
 
+class TestFirefoxHubView(TestCase):
```
alexgibson: This passes for me locally:

```python
@override_settings(DEV=False)
@patch('bedrock.firefox.views.l10n_utils.render')
class TestFirefoxFeatures(TestCase):
    @patch.object(views, 'lang_file_is_active', lambda *x: True)
    @patch('bedrock.firefox.views.switch', Mock(return_value=True))
    def test_new_firefox_page_loads_with_active_switch(self, render_mock):
        view = views.FirefoxHubView()
        view.request = RequestFactory().get('/firefox/')
        view.request.locale = 'en-US'
        eq_(view.get_template_names(), ['firefox/firefox-quantum.html'])

    @patch.object(views, 'lang_file_is_active', lambda *x: False)
    @patch('bedrock.firefox.views.switch', Mock(return_value=False))
    def test_firefox_hub_page_loads_with_inactive_switch(self, render_mock):
        view = views.FirefoxHubView()
        view.request = RequestFactory().get('/firefox/')
        view.request.locale = 'fr'
        eq_(view.get_template_names(), ['firefox/hub/home.html'])
```

bedrock/firefox/templates/firefox/firefox-quantum.html
```diff
@@ -0,0 +1,170 @@
...
+          </p>
+        </div>
+      </section>
+    </div>
+  </div>
+  <section class="secondary-download-cta" data-scroll-tracking="Secondary Download CTA">
+    <div class="content">
+      <div class="secondary-download-cta-content">
+        <h2>{{ _('Start browsing with <strong>Firefox</strong> Quantum') }}</h2>
+        {{ download_firefox(alt_copy=_('Download Now'), dom_id='firefox-desktop-page', download_location='other', button_color='quantum-download') }}
```
alexgibson: This is a duplicate ID: `dom_id='firefox-desktop-page'` - it should be different to the one in the primary cta section at the top of the page

bedrock/firefox/tests/test_views.py
alexgibson: Hmm, actually - you could probably split this test out further so the switch is tested independently of the locale activation

alexgibson:
Download now is not aligned vertically

screen shot 2017-10-12 at 15 47 03

alexgibson:
Is this heading a little large maybe? Alternatively, could you shrink the width so "Firefox Quantum" is on the second line together?

screen shot 2017-10-12 at 15 47 40

bedrock/firefox/tests/test_views.py
schalkneethling: Oh so you mean test that if one of the two is not True then the current page should show. Will do.

bedrock/firefox/templates/firefox/firefox-quantum.html
schalkneethling: Doh!

schalkneethling:
> Is this heading a little large maybe? Alternatively, could you shrink the width so "Firefox Quantum" is on the second line together?

Agree

schalkneethling:
> Download now is not aligned vertically

I made the "Firefox Privacy Notice" a smaller font size to match the page you pointed out the other day that also uses the same button style.

alexgibson:
> I made the "Firefox Privacy Notice" a smaller font size to match the page you pointed out the other day that also uses the same button style.

I don't see the same issue here: https://www-demo3.allizom.org/en-US/firefox/new/

schalkneethling:
> I don't see the same issue here: www-demo3.allizom.org/en-US/firefox/new

@alexgibson I believe the last update solved the problem. Apologies, my mind is all over the place today...

peiying2:
@schalkneethling could you post test URL here please?

tests/functional/newsletter/test_newsletter_embed.py
```diff
@@ -71,6 +77,9 @@ def test_newsletter_successful_sign_up(page_class, base_url, selenium):
     assert page.newsletter.sign_up_successful
 
 
+# /firefox Quantum page behind switch, unskip when the switch is removed
+@pytest.mark.skipif(
```
alexgibson: Same here:

```python
@pytest.mark.parametrize('page_class', [
    HomePage,
    ContributePage,
    pytest.mark.skipif(HomePage, reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1293779')])
```

tests/functional/newsletter/test_newsletter_embed.py
```diff
@@ -58,6 +61,9 @@ def test_newsletter_default_values(page_class, url_kwargs, base_url, selenium):
     assert page.newsletter.is_privacy_policy_link_displayed
 
 
+# /firefox Quantum page behind switch, unskip when the switch is removed
+@pytest.mark.skipif(
```
alexgibson: Instead of skipping all the tests, skip just the failing one:

```python
@pytest.mark.parametrize('page_class', [
    HomePage,
    ContributePage,
    pytest.mark.skipif(HomePage, reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1293779')])
```

tests/functional/newsletter/test_newsletter_embed.py
```diff
@@ -24,6 +24,9 @@
 from pages.smarton.base import SmartOnBasePage
 
 
+# /firefox Quantum page behind switch, unskip when the switch is removed
+@pytest.mark.skipif(
+    reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1405076')
```
alexgibson: Use this to skip only the failing test:

```python
pytest.mark.skipif((FirefoxHomePage, None), reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1405076')
```

tests/functional/firefox/test_home.py
```diff
@@ -7,6 +7,9 @@
 from pages.firefox.home import FirefoxHomePage
 
 
+# /firefox Quantum page behind switch, unskip when the switch is removed
```
alexgibson: I don't think we really need the comments here, as we have the `reason` referencing the bug.

tests/functional/newsletter/test_newsletter_embed.py
schalkneethling: Guessing the above should be:

```
pytest.mark.skipif(FirefoxHomePage, reason='https://bugzilla.mozilla.org/show_bug.cgi?id=1405076')])
```

alexgibson: Ah yes, my bad - thanks!

media/css/firefox/components/_buttons.scss
```diff
@@ -0,0 +1,27 @@
...
+a.button:visited {
+    &.quantum-download {
+        background-color: #4ee331;
+        padding: .3em 40px .5em;
+        border-color: #4ee331;
+        border-radius: 100px;
+
+        .download-title {
+            color: #0040ab;
+            @include font-size(14px);
```
alexgibson: Nit: try and and avoid using arbitraty font sizes. Suggest `@include font-size-small;` here?

media/css/firefox/quantum-landing.scss
```diff
@@ -0,0 +1,787 @@
...
+#outer-wrapper {
+    overflow-x: hidden;
+}
+
+h1, h2, h3, h4, h5, h6 {
+    font-weight: normal;
+}
+
+.fx-privacy-link a {
+    @include font-size(14px);
```
alexgibson: Suggest `@include font-size-extra-small;` here? Could this also be moved to `_buttons.scss`?

media/css/firefox/components/_buttons.scss
```diff
@@ -0,0 +1,27 @@
...
+            font-weight: 700;
```
alexgibson: Any reason not just to use `font-weight: bold;`?

schalkneethling: Nope, I was using the same code as from the reference page that uses the same button style but, bold would work just as well.

schalkneethling:
Thanks for the detailed review @alexgibson - updated.

bedrock/firefox/views.py
```diff
@@ -744,7 +745,17 @@ class FirefoxHubView(BlogPostsView):
     blog_posts_template_variable = 'articles'
     blog_slugs = 'firefox'
     blog_tags = ['home']
-    template_name = 'firefox/hub/home.html'
+
+    def get_template_names(self):
+        locale = l10n_utils.get_locale(self.request)
+
+        if switch('firefox-57-release') and lang_file_is_active(
+                'firefox/firefox-quantum.lang', locale):
```
alexgibson: Appologies I missed this before, it should be `firefox/firefox-quantum`? (no `.lang`)

schalkneethling: 🐍

PR #5200, opened 2017-10-12
IRC notification if either smoke or unit tests fail during deploy

pmac:


pmac:
This should fix it @alexgibson. Let me know what you think.

alexgibson:
Looks good to me, thanks @pmac!

PR #5206, opened 2017-10-13
Fix Bug 1401245, send tabs a/b test

schalkneethling:
## Description
Implements an A/B test to determine if there is a higher engagement on send tabs with a video in the intro as apposed to a static image.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1401245

## Testing
Ensure traffic is directed to v=a and v=b and that the correct template is loaded.


bedrock/settings/static_media.py
```diff
@@ -1239,6 +1245,13 @@
         ),
         'output_filename': 'js/firefox-features-sync-bundle.js',
     },
+    'experiment-send-tabs': {
+        'source_filenames': {
```
alexgibson: This should be curly braces `(`

bedrock/settings/static_media.py
```diff
@@ -307,6 +307,12 @@
         ),
         'output_filename': 'css/firefox-features-sync-bundle.css',
     },
+    'experiment-send-tabs': {
+        'source_filenames': {
```
alexgibson: Same here `(`

jpetto:
Closing in favor of #5242

PR #5212, opened 2017-10-18
[fix bug 1397519] Firefox desktop page Quantum redesign

alexgibson:
## Description
- Adds redesigned `/firefox/` page for Quantum 57 release
- Redirects `/firefox/desktop/` to `/firefox/`.
- Both things are behind the switch `firefox-57-release`.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1397519

## Testing
Demo: https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/

- Test major evergreen browser, plus IE > 9.
- Test responsiveness.
- Test RTL styles.
- Test redirect fot /firefox/desktop/ works as expected.


alexgibson:
OK, strings here should all be final.

peiying2:
@alexgibson I couldn't extract the strings using the script, including a patch I normally do. Maybe it is because it didn't pass the test.

alexgibson:
@peiying2 the tests shouldn’t make a difference to being able to extract afaik.

alexgibson:
Successful test run: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/105/pipeline/

alexgibson:
OK, I believe everything should now be in place here. This is ready for review.

bedrock/firefox/templates/firefox/hub/home-quantum.html
```diff
@@ -0,0 +1,390 @@
...
+                  <a href="https://www.youtube.com/watch?v=YIywpvHewc0">
+                    <img src="{{ static('img/firefox/hub/quantum/quantum-vs-chrome-poster.png') }}" alt="">
+                  </a>
+                </video>
+              </div>
+            </div>
+          </div>
+        </div>
+      </div>
+    </div>
```
jpetto: `<div>` tree! Maybe add some comments after to denote which parent is being closed?

media/js/firefox/quantum/features-scroller.js
```diff
@@ -0,0 +1,216 @@
...
+                handler: function(direction) {
+                    if (direction === 'up') {
+                        setActiveFeature(this.element.id);
+                        playVideo(this.element);
+                    } else {
+                        pauseVideo(this.element);
+                    }
+                },
+                offset: '-50%'
+            }));
```
jpetto: Not sure I see a way around it (aside from not auto-playing!), but having two waypoints per feature is a bummer.

alexgibson:
> Wonder if folks will complain about the sticky nav not working, even in Edge. Suppose we can cross that bridge if/when we come to it.

The latest version of Edge (16) supports both `position:sticky;` and `display: grid;`, so that should be 1st class support on the latest versions of every major desktop browser. If that's not enough, I think we can make a case that using native browser features gives a much smoother experience than resorting to a JS implementation.

> Chrome (on macOS) has some visual glitches when scrolling over the gradients, but I don't think there's much we can do there.

I don't see any scrolling through the page on Chrome here, but likely not much we can do there yeah.

bedrock/firefox/views.py
```diff
@@ -749,6 +749,15 @@ class FirefoxProductDesktopView(BlogPostsView):
     blog_tags = ['browser', 'featured']
     template_name = 'firefox/products/desktop.html'
 
+    def render_to_response(self, context, **response_kwargs):
+        if switch('firefox-57-release'):
+            return HttpResponsePermanentRedirect(reverse('firefox'))
```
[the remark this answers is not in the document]
alexgibson: Once the switch is enabled we can remove this logic and just do a regular 301 redirect in `redirects.py` - so i think this is OK?

media/js/firefox/quantum/features-scroller.js
alexgibson: Yeah I thought about a way to do this too, but if we're to support auto play for both scrolling up and down the web page, two waypoints are needed: one for when the videos pass through the top of the viewport, and one for when the pass through the bottom of the viewport. This seems to work well enough, even though it's more overhead than i would have liked. In a perfect world, I would have not chosen to auto-play the videos at all (I think it's arrogant to assume users want this).

alexgibson:
@jpetto thanks for the review! I added a commit with some updates.

alexgibson:
@jpetto squashed - I think this should be good to merge. Thanks again!

PR #5214, opened 2017-10-18
Fix 1409873: Remove dead link on Faye contributor page

pmac:


pmac:
The domain has lapsed and is now a placeholder. As the bug suggests, we should remove it before it becomes something unfortunate and we're linking to something we shouldn't.

PR #5215, opened 2017-10-18
[fix bug 1397523] Update UI of fx feature pages for 57.

jpetto:
## Description

Applies 57 redesign to all `/firefox/features/` pages.

Also attempts to abstract re-used styles from `/firefox/new/` and `/firefox/quantum/`.

Also removes `sync-old.html` template/logic (as we have 95% coverage on strings in `sync.html`).

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1397523

## Testing

https://bedrock-demo-jpetto.us-west.moz.works/en-US/firefox/features/

In addition to testing the new templates in modern browsers and old IEs, and mobile/desktop, and RTL, make sure no regressions to `/firefox/new/` and `/firefox/quantum/`.

Once again, the`/firefox/features/sync/` and `/firefox/features/send-tabs/` pages should get the full round of tests: fx31+ signed in, fx31+ signed out, old fx, not fx, fx android, fx iOS.

[Integration tests are passing](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/101/pipeline/).


peiying2:
@jpetto let me know when this PR is ready for string extraction.

flodolo:
> @jpetto let me know when this PR is ready for string extraction.

Is there a list of string changes somewhere? Because it's honestly impossible to figure out with this amount of changes and new files.

flodolo:
> Is there a list of string changes somewhere? Because it's honestly impossible to figure out with this amount of changes and new files.

OK, it's only two strings behind the tag firefox-features-57-update

jpetto:
> OK, it's only two strings behind the tag firefox-features-57-update

Yep, the only new strings are behind `l10n_has_tag` conditionals. Many strings have been dropped though.

flodolo:
> Yep, the only new strings are behind l10n_has_tag conditionals. Many strings have been dropped though.

That's not good :-\ Do you have a list of the strings? The alternative would be extracting all templates, go string by string, and figure out if each string is actually missing or in the shared file.

jpetto:
> That's not good :-\ Do you have a list of the strings? The alternative would be extracting all templates, go string by string, and figure out if each string is actually missing or in the shared file.

I don't have a list, but I can give you a general pattern.

On the landing page (`/firefox/features/`), many of the feature blurbs have been stripped down to only the heading. The description and link text has been dropped for all but 5. Also, the Android and iOS "features" have been removed entirely.

The other major change is the "More Performance/More Freedom" section currently at the bottom of all feature pages on prod has been removed.

flodolo:
We're extracting the 2 new strings. For removal, we need to wait that this goes to production anyway.

alexgibson:
Blog titles could do with a little extra line height?

screen shot 2017-10-23 at 10 47 41

alexgibson:
RTL `.key-feature` elements are mis-aligned:

alexgibson:
This is just a suggestion, but seeing as we know these pages are going to be just SEO fodder as soon as they get removed from the nav - would it be worth adding a "Back to landing page" type of link on each detail page? At least then, users who do land on one could at least discover more then a single feature.

media/css/quantum/_common.scss
```diff
@@ -0,0 +1,165 @@
...
+/* -------------------------------------------------------------------------- */
+// Key features sections
+
+.key-features-section {
+    background: #fff;
+
+    h2 {
+        @include font-size-level3;
+        color: $color-quantum-blue;
+        font-weight: bold;
```
alexgibson: On the `/firefox/quantum/` page design, these titles we're previously all in regular. This change leaves a bit of an odd mix of weights on the page?

media/css/quantum/_common.scss
```diff
@@ -0,0 +1,165 @@
...
+}
+
+h1, h2, h3, h4, h5, h6 {
+    font-weight: normal;
+}
+
+/* -------------------------------------------------------------------------- */
+// Key features sections
+
+.key-features-section {
```
alexgibson: Would this be worth putting in it's own file to import? I'm curious how regularly this will be used, or if pages would just end up importing unused CSS.

media/css/firefox/features/quantum/common.scss
```diff
@@ -0,0 +1,140 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../../../pebbles/includes/lib';
+@import '../../../hubs/mixins';
+@import '../../../hubs/sections';
+@import '../../../hubs/masthead';
+@import '../../../hubs/header';
+@import '../../../hubs/sub-nav';
+@import '../../../quantum/common';
+@import '../../../quantum/buttons';
```
alexgibson: Funny story - I was originally told the button on `/new` would be a one-off special case, but it's already sneaking in to other places unsuprisingly. I don't think it's going to be the new default button style though (at least the one on `/firefox/` is more akin to the regular buttons and suits a white background). I don't suggest we do anything about this now, but it's something I think we might need to try and sort out with our designers :/

alexgibson:
Not part of this change, but the `thanks` title is a bit out sync in alignment.

screen shot 2017-10-23 at 11 20 18

bedrock/firefox/templates/firefox/features/quantum/index.html
```diff
@@ -0,0 +1,172 @@
...
+          <h3>{{ _('Browse faster') }}</h3>
+        </a>
+      </li>
+    </ul>
+  </div>
+</section>
+
+<section class="fx-product-benefits-section">
+  <h2 class="content">
+  {% if l10n_has_tag('firefox-features-57-update') %}
```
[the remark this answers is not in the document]
jpetto: AFAIK, we don't have an appropriate fallback string. I know it's not really semantic, but I kept the `<h2>` to maintain the visual separation. I'll switch to a `<div>` containing the conditional `<h2>` to be a bit more semantic.

bedrock/firefox/templates/firefox/features/quantum/index.html
```diff
@@ -0,0 +1,172 @@
...
+    </ul>
+  </div>
+</section>
+{% endblock %}
+
+{% block news %}{% endblock %}
+
+{% block js %}
+  {{ super() }}
+  {% javascript 'firefox-features-landing' %}
```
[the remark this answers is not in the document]
jpetto: Ah, good catch. There's also some GA stuff in there, but I don't think it applies to the new layout. Will remove.

media/css/firefox/features/quantum/common.scss
jpetto: 😒

media/css/quantum/_common.scss
jpetto: Will refactor into its own file.

media/css/quantum/_common.scss
jpetto: Will remove `bold` for consistency.

media/css/quantum/_lib.scss
```diff
@@ -0,0 +1,20 @@
...
+
+@import '../pebbles/includes/lib';
+
+
+$color-quantum-blue: #0a84ff;
+
+@mixin tail-gradient-heading() {
+    @include font-size-level3;
+    background: #0060df;
+    background: url('/media/img/firefox/quantum/fox-tail-fresh-from-our-blog.png') bottom -300px center no-repeat,
```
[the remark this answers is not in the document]
jpetto: Agreed. Will just use ad-hoc for now.

media/css/quantum/_newsletter.scss
```diff
@@ -126,6 +125,7 @@
     }
 
     @media #{$mq-tablet} {
+        display: grid;
```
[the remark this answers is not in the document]
jpetto: Will add float-y fallback.

media/css/firefox/features/quantum/common.scss
```diff
@@ -0,0 +1,140 @@
...
+
+
+// update style of nav/sub nav download buttons
+#nav-download-firefox,
+#global-nav-download-firefox,
+#sub-nav-download-firefox {
+    .download-link.button.button-green {
+        @include font-size-small;
+        @include transition(background-color .1s ease-in-out);
+        background-color: #d7d7db;
```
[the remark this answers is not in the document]
jpetto: I'll make it green. 👍

jpetto:
@alexgibson I think I've addressed all feedback. Thanks!

bedrock/settings/static_media.py
```diff
@@ -1221,13 +1250,6 @@
         ),
         'output_filename': 'js/firefox-features-hub-bundle.js',
     },
-    'firefox-features-landing': {
```
alexgibson: Isn't this bundle still needed by the old page?

media/js/firefox/features/landing.js
```diff
@@ -1,21 +0,0 @@
-/* This Source Code Form is subject to the terms of the Mozilla Public
```
alexgibson: File still needed by the old template?

media/css/quantum/_buttons.scss
```diff
@@ -2,10 +2,35 @@
...
+@import 'lib';
 
-$color-quantum-bright-green: #4ee331;
-$color-quantum-blue: #0a84ff;
+
+// update style of nav/sub nav download buttons
+#nav-download-firefox,
+#global-nav-download-firefox,
+#sub-nav-download-firefox {
+    .download-link.button.button-green {
```
alexgibson: I'm guessing we're not going to change the default button styles for the nav on all pages yet, but It would be nice to make this a new button style, instead of just overriding `button-green`?

media/css/quantum/_buttons.scss
```diff
@@ -20,7 +45,7 @@ a.button:visited {
         padding: .9em 40px;
 
         .download-title {
-            color: #0040ab;
+            color: #fff;
```
alexgibson: I can't comment on the line below here, but the hover color feels too bright and makes the button text hard to read? I think Christa mentioned she was going to work on hover states, so she might have an alternative? I also notice there's a 2px border on hover, which may or may not be required.

media/css/quantum/_buttons.scss
```diff
@@ -20,7 +45,7 @@ a.button:visited {
         padding: .9em 40px;
```
alexgibson: We should probably double check default padding for buttons too, as this special casing may no longer br required? 

Also, according to the doc that was shared in Slack (🙄), default button text should be `16px`?

alexgibson:
In RTL the key-feature images are not aligned:

https://screenshots.firefox.com/uoLBssKzw2wL3ohb/127.0.0.1

bedrock/settings/static_media.py
jpetto: 🤦‍♂️

media/css/quantum/_buttons.scss
jpetto: For now, assuming any template that includes this file will want the full "quantum" UI experience. Can/will revisit in upcoming PRs if this proves to not be the case.

media/css/quantum/_buttons.scss
jpetto: I personally think this padding looks good, and will update to 16px for the text. We'll see if Christa has more direction when she gets in.

media/css/quantum/_buttons.scss
jpetto: Going to go 2% darker until we have real direction.

PR #5222, opened 2017-10-19
Fix Bug 1399665 - Update Developer Edition, DevTools and Roadshow tiles on /Developer/

kyoshino:
## Description

Update the [Mozilla Developers](https://www.mozilla.org/en-US/developer/) page as follows:

* Hero: Replace the DevTools promo with Quantum Developer Edition, use a new image
* DevTools: Change the link to MDN
* Developer Roadshow: Change the link to GitHub

## Issue / Bugzilla link

[Bug 1399665](https://bugzilla.mozilla.org/show_bug.cgi?id=1399665)

## Testing

Visit the Developer page and make sure those 3 tiles have been updated.


kyoshino:
Got new images. Should be ready for review!

craigcook:
Should these string changes be behind an l10n tag? Also, I think we're supposed to stop using "Firefox Quantum" after 57 releases, so we would need to change at least that string again in the very near future. I'll confirm if that's the plan.

kyoshino:
Looks like this page hasn't localized yet, so we don't have to care about l10n.

craigcook:
You're right, it hasn't yet been localized. My mistake.

PR #5223, opened 2017-10-19
Bug 1397532 - New quantum firstrun animation.

ericawright:
r=@craigcook 

## Description
Redesign firstrun "wow moment"

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1397532

 - Whoever is responsible for ` /media/css/base/notification-banner.less` (see line 157) will need to add new "special hackz" for this page ;D

### DIFF
--- a/bedrock/firefox/templates/firefox/firstrun/firstrun_quantum.html
+++ b/bedrock/firefox/templates/firefox/firstrun/firstrun_quantum.html
@@ -1,0 +1,103 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% add_lang_files "firefox/new/quantum" %}
+
+{% extends "firefox/firstrun/base.html" %}
+
+{% block body_id %}firstrun-quantum-onboarding{% endblock %}
+
+{% block page_title %}
+  {{ _('Welcome to Firefox') }}
+{% endblock %}
+
+{% block site_css %}
+  {% stylesheet 'firefox_firstrun_quantum' %}
+{% endblock %}
+
+{% block js %}
+  {% javascript 'firefox_firstrun_quantum' %}
+{% endblock %}
+
+{% block global_nav %}{% endblock %}
+{% block site_header %}{% endblock %}
+
+{% block content %}
+  <div id="scene" xmlns:xlink= "http://www.w3.org/1999/xlink">
+
+     <svg id="svg1" class="st1 background" xmlns="http://www.w3.org/2000/svg" viewBox="-150 -210 1750 1300" preserveAspectRatio="none">
+       <path d="M1370.3-182.6C1266.9-68.3,1137.6,49.5,938.2,19.7C713.4-13.8,523.4-126.7,308.9-176.4
+         C44.3-223.7-119.5-7.6-133,165.8c0,0-92.4,416.6,212.7,598.7c254,151.6,913.4,346.3,1353.8,165.9
+         c146.8-60.1,133.2-686.2,87.5-1015.2C1502.5-218.3,1401.5-217.1,1370.3-182.6z"/>
+    </svg>
+
+    <svg class="st2 background" xmlns="http://www.w3.org/2000/svg" viewBox="-300 100 2000 800" preserveAspectRatio="none">
+      <path d="M-152.8,139.4C13.7,87.4,256,100.5,578,267.6c274.7,142.6,278.9,20,620.9-61.4
+    	c467.6-111.3,534,417.6,429,555.7c-447,414-1490.7,140.5-1707.2,12S-397.3,215.8-152.8,139.4z"/>
+    </svg>
+
+    <svg class="st3 background" xmlns="http://www.w3.org/2000/svg" viewBox="-160 265 1650 580" preserveAspectRatio="none">
+      <path d="M-132.3 346.9c84.6-94.2 379.7-122 662.4 24.7 168.4 87.4 171 12.3 380.6-37.6 286.6-68.2 657.8 200.7 544.3 333-375.6 314.4-1222.2 78.8-1354.9 0s-317.7-225.1-232.4-320.1z"/>      
+    </svg>
+
+    <svg class="st4 background" xmlns="http://www.w3.org/2000/svg" viewBox="-270 550 2100 290" preserveAspectRatio="none">
+      <path d="M1784.3 638c-234-191.8-398.9 43.1-824.9 197.8C565 979 307.4 476.2 26.2 564.6c-281.2 88.4-314.8 588 182 913.2s1345.9 140.9 1549.6-140C1915 1121 1949 773 1784.3 638z"/>
+    </svg>
+
+    <svg class="st5 background" xmlns="http://www.w3.org/2000/svg" viewBox="-278 -170 1900 750" preserveAspectRatio="none">
+      <path d="M71.8 503.1c1315 237.8 1451-225.9 1443-438S778.8 405 415.4 282.9c-316.6-106.4-497-380-684.8-337s-973.8 319.4 341.2 557.2z"/>      
+    </svg> 
+
+    <svg class="st6 background" xmlns="http://www.w3.org/2000/svg" viewBox="-300 -150 2050 900" preserveAspectRatio="none">
+      <path d="M495.3-365.1C66.6-148.4-687.9 864.2-325.8 729.7s582.1-353.8 847.5-390c453.8-61.9 814 456.7 1344.2 334.6 530.2-122-579.6-1305.2-1370.6-1039.4z"/>
+    </svg>
+
+    <svg class="st7 background" xmlns="http://www.w3.org/2000/svg" viewBox="-230 -200 1380 790" preserveAspectRatio="none">
+      <path d="M-583.2-65.9c56.5 448.6 1076.5 1063.5 1574.5 248S470.2 26.6 3.8-335.4s-643.6-179.1-587 269.5z"/>
+    </svg>
+
+    <svg class="st8 background" xmlns="http://www.w3.org/2000/svg" viewBox="-155 -80 2000 580" preserveAspectRatio="none">
+      <path d="M1327.5-88.5c-1315-237.8-1451 225.9-1443 438S620.5 9.7 983.9 131.8c316.6 106.4 497 380 684.8 337s973.8-319.4-341.2-557.3z"/>
+    </svg>
+
+    <svg class="white-overlay" viewBox="0 0 1440 800" preserveAspectRatio="none">
+      <rect x="0" y="0" width="1440" height="800"/>
+    </svg>
+
+    <svg class="cls-2" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 1440 800" preserveAspectRatio="none">
+      <defs>
+        <linearGradient id="light-gradient" gradientUnits="userSpaceOnUse" x1="765.723" y1="594.394" x2="769.406" y2="229.758" gradientTransform="matrix(1 0 0 -1 0 800)">
+          <stop offset="0" stop-color="#002989"/>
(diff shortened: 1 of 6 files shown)


media/img/aboutRights.xhtml
```diff
@@ -0,0 +1,102 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
flodolo: Wrong file, wrong place?

flodolo:
Side question: do we expect more content or this is the final plan for the Firstrun page?

craigcook:
> Side question: do we expect more content or this is the final plan for the Firstrun page?

I think the copy will be pretty minimal so not much more than this, but I'm not sure if the copy here is actually final. I'll find out and we can try to get these strings frozen as soon as possible.

media/img/aboutRights.xhtml
ericawright: yep, this shouldn't be in here at all, thank you

ericawright:
@craigcook @flodolo there is no copy change as compared to the previous firstrun animation. So this copy is final.

bedrock/firefox/templates/firefox/firstrun/firstrun_quantum.html
```diff
@@ -0,0 +1,78 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% add_lang_files "firefox/new/onboarding" %}
```
flodolo: We shouldn't use this file, it's already been replaced by `firefox/new/quantum` and scheduled for removal.

bedrock/firefox/templates/firefox/firstrun/firstrun_quantum.html
```diff
@@ -0,0 +1,78 @@
...
+
+{% extends "firefox/firstrun/base.html" %}
+
+{% block body_id %}firstrun-quantum-onboarding{% endblock %}
+
+{% block page_title %}
+  {{ _('Welcome to Firefox') }}
```
flodolo: We need to copy all these strings in `firefox/new/quantum.lang` (not in Bedrock, on our side)

media/css/firefox/firstrun/firstrun_quantum.less
```diff
@@ -0,0 +1,405 @@
...
+    min-width: 100%;
+    object-fit: cover;
+    position: fixed;
+    right: 0;
+    top: 0;
+    width: 100%;
+}
+
+#firefox-logo {
+    background: url('/media/img/firefox/firstrun-57/firefox_and_wordmark.png');
```
craigcook: This URL should be `img/firefox/firstrun-quantum/firefox_and_wordmark.png` to fix the build error

ericawright:
@craigcook this is close enough that it can be looked at now. 
I'm using `Mozilla.UITour` to see if the user is logged in - a few will be according to https://bugzilla.mozilla.org/show_bug.cgi?id=1386184. And if they are logged in already to simply play the animation for a bit and move on. Unfortunately this seems to require a bit of config when running locally.
Some visual changes - colour and positioning will likely still happen for a while. 
one performance issue I'm still working on is during the spin-out, the fps drops drastically and I'm seeing a few long repaints. My theory is that this has to do with the sizes of the svgs (as that was the issue earlier) If you have any ideas to fix it let me know.

bedrock/settings/static_media.py
```diff
@@ -1301,6 +1316,28 @@
         ),
         'output_filename': 'js/firefox_firstrun-bundle.js',
     },
+    'firefox_firstrun_quantum': {
+        'source_filenames': (
+            'js/base/mozilla-fxa-iframe.js',
```
craigcook: I think you need to add `'js/base/uitour-lib.js',` here to invoke UITour

ericawright: yes, that explains it, thank you

media/js/firefox/firstrun/firstrun_quantum.js
```diff
@@ -0,0 +1,70 @@
...
+                onVerificationComplete: onVerificationComplete,
+                onLogin: onVerificationComplete,
+                onFormEnabled: disableSkipButton,
+                onNavigated: hideOrShowSkipButton
+            });
+        });
+    };
+
+    document.onreadystatechange = function () {
+        let syncConfig;
```
craigcook: UglifyJS doesn't seem down with `let` and is throwing build errors. Switching this to a regular `var` passes the test and still seems to work.

media/js/firefox/firstrun/firstrun_quantum.js
```diff
@@ -0,0 +1,70 @@
...
+        Mozilla.UITour.getConfiguration('sync', config => {
```
craigcook: Uglify also isn't happy with `=>` here. I think we need to go with an old fashioned function.

ericawright:
@craigcook passing the tests now. Visual and performance changes will still be incomming.

media/css/firefox/firstrun/firstrun_quantum.less
```diff
@@ -0,0 +1,542 @@
...
+
+:root {
+  --animation-curve: linear;
+  // --animation-curve: cubic-bezier(.73,.08,.94,.33);
+  // --backwards: cubic-bezier(.49,-0.41,.96,.67);
+}
+
+body {
+    margin: 0;
+    overflow: hidden;
```
craigcook: This is preventing any scrolling, but scrolling is probably essential for smaller windows. You might not be able to scroll down to the button, for example. We should probably build in some responsiveness (see other comments elsewhere) and at least allow for some vertical scrolling when needed. I'm not sure how to make that work with the wavy background though since that pretty much has to overflow. Maybe it could be in a wrapper sized to the viewport with `overflow: hidden` and `position: fixed;`, then the content on a separate layer on top of it? Worth a try maybe?

media/css/firefox/firstrun/firstrun_quantum.less
```diff
@@ -0,0 +1,542 @@
...
+.fxaccounts-container {
+    bottom: 0;
+    color: #fff;
+    height: 485px;
+    left: 0;
+    margin: auto;
+    position: absolute;
+    right: 0;
+    top: 0;
+    width: 980px;
```
craigcook: This fixed width seems a bit too wide for safety and might assume larger windows than some people will have. We don't need to be too concerned about really small viewports since this isn't shown on mobile, but we could add a media query or two to make things more compact for smaller windows. The old page just fits at 750 x 500 so that seems like a decent minimum size to aim for. We could go fully responsive and stack boxes in really skinny windows but that might be overkill.

media/css/firefox/firstrun/firstrun_quantum.less
```diff
@@ -0,0 +1,542 @@
...
+        margin: 0 0 10px;
+    }
+}
+
+#skip-button {
+    -moz-appearance: none;
+    background-color: #0996f8;
+    border-radius: 3px;
+    border: 1px solid #0670cc;
+    box-shadow: 0 0 0 0 transparent;
```
craigcook: I think `box-shadow: none;` also removes a shadow, but this may not be needed since I'm not sure this button has a shadow to remove.

bedrock/firefox/templates/firefox/firstrun/firstrun_quantum.html
```diff
@@ -0,0 +1,111 @@
...
+          <stop offset="0" stop-color="#ff272d"/>
+          <stop offset=".5" stop-color="#ff8e00"/>
+          <stop offset="1" stop-color="#ffde00"/>
+        </linearGradient>
+      </defs>
+      <rect x="0" y="0" width="1440" height="800"/>
+    </svg>
+
+    <div class="fxaccounts-container">
+      {% block page_content %}
```
craigcook: I don't think we need this block, at least not at this point. Looks like it was carried over from the previous page that had a few variant templates for an A/B test, but we don't have such a test for this page yet.

bedrock/firefox/templates/firefox/firstrun/firstrun_quantum.html
```diff
@@ -0,0 +1,111 @@
...
+      <div id=firefox-logo></div>
```
craigcook: Nit: attribute should be quoted.

bedrock/settings/static_media.py
```diff
@@ -396,6 +396,21 @@
...
     },
+    'firefox_firstrun_quantum': {
+        'source_filenames': (
+            'css/base/notification-banner.less',
+            'css/firefox/firstrun/firstrun_quantum.less',
+            'css/base/mozilla-fxa-iframe.less',
+        ),
+        'output_filename': 'css/firefox_firstrun_quantum-bundle.css',
+    },
+    'firefox_firstrun_membership': {
```
craigcook: This was removed in https://github.com/mozilla/bedrock/pull/5243. Rebasing from master might clean this up automatically but if not we should remove it here.

bedrock/settings/static_media.py
```diff
@@ -1301,6 +1316,29 @@
...
     },
+    'firefox_firstrun_quantum': {
+        'source_filenames': (
+            'js/base/uitour-lib.js',
+            'js/base/mozilla-fxa-iframe.js',
+            'js/firefox/firstrun/firstrun_quantum.js',
+        ),
+        'output_filename': 'js/firefox_firstrun_quantum-bundle.js',
+    },
+    'experiment-firstrun-membership': {
```
craigcook: This and the `'firefox_firstrun_membership'` block below were also removed from master and shouldn't be readded here.

bedrock/firefox/templates/firefox/firstrun/firstrun_quantum.html
```diff
@@ -0,0 +1,111 @@
...
+
+{% block page_title %}
+  {{ _('Welcome to Firefox') }}
+{% endblock %}
+
+{% block site_css %}
+  {% stylesheet 'firefox_firstrun_quantum' %}
+{% endblock %}
+
+{% block experiments %}
```
craigcook: I suspect this came over from the previous template, but was removed in https://github.com/mozilla/bedrock/pull/5243/ and probably shouldn't be here.

media/css/firefox/firstrun/firstrun_quantum.less
ericawright: I've set each image to `position: fixed`, and changed the overflow-y on a height of 710px.

ericawright:
@vladikoff can you ensure that I'm using the correct address (utm_campaign etc) for the fxa iframe? Would you like any new params for this new page?
`firstrun_quantum.html` line 96

bedrock/firefox/templates/firefox/firstrun/firstrun_quantum.html
```diff
@@ -0,0 +1,103 @@
...
+
+    <div class="fxaccounts-container">
+      <div id="firefox-logo"></div>
+      <div id="left-divider">
+        <p id="title">{{_('Already using Firefox?')}}</p>
+        <p class="content">{{_('Sign in to your account and we’ll sync the bookmarks, passwords and other great things you’ve saved to Firefox on other devices.')}}</p>
+        <a href="{{ url('firefox.features.sync') }}" target="_blank">{{_('Learn more about Firefox Accounts')}}</a>
+      </div>
+      <div class="fxaccounts" id="fxa-iframe-config" data-host="{{ settings.FXA_IFRAME_SRC }}" data-mozillaonline-host="{{ settings.FXA_IFRAME_SRC_MOZILLAONLINE }}">
+        <iframe id="fxa" scrolling="no" data-src="{{ settings.FXA_IFRAME_SRC }}signin?utm_campaign=fxa-embedded-form&amp;utm_medium=referral&amp;utm_source={{ utm_source }}&amp;utm_content=fx-{{ version }}&amp;entrypoint=firstrun&amp;service=sync&amp;context=iframe&amp;style=chromeless&amp;haltAfterSignIn=true"></iframe>
```
vladikoff: params look good !

PR #5225, opened 2017-10-20
Fix bug 937865: Use new firefox-*-latest-ssl bouncer products

pmac:
This should move most of our download links to using bouncer aliases and away from using version numbers in the URLs. This means that users will download the latest versions as soon as bouncer updates the aliases (e.g. `firefox-latest-ssl`) instead of having to wait for the site to update the data from product-details and the CDN cache to time out. We've been using such aliases for Nightly and Aurora (dev edition) for a while now, but the releng team has recently updated bouncer ([bug 1387622](https://bugzilla.mozilla.org/show_bug.cgi?id=1387622)) to have these new aliases and we should use them.

Only esr_next should use the version number now. It could move to an alias in the future as well pending [bug 1408868](https://bugzilla.mozilla.org/show_bug.cgi?id=1408868)


pmac:
@chrismore this does make it so that only stub installers will get the `-f##` suffixes for funnelcakes. So that limits funnelcakes to only `win` or `win64`. Does that sound okay?

bedrock/firefox/tests/test_firefox_details.py
```diff
@@ -116,7 +114,7 @@ def setUp(self):
     def test_get_download_url(self):
         url = firefox_desktop.get_download_url('release', '17.0.1', 'osx', 'pt-BR', True)
```
jpetto: It seems the `version` argument is now only used for the `esr_next` channel. Is there a benefit to keeping the `version` argument in tests for other channels?

bedrock/firefox/tests/test_firefox_details.py
```diff
@@ -460,49 +456,11 @@ def test_funnelcake_direct_links_en_us_win_only(self):
...
-
-        url = firefox_desktop.get_download_url('release', '45.0', 'linux', 'de', force_direct=True, funnelcake_id='64')
-        ok_('product=firefox-45.0-f64' in url)
-
-        url = firefox_desktop.get_download_url('release', '45.0', 'linux', 'en-US', force_direct=True, funnelcake_id='64')
-        ok_('product=firefox-45.0-f64' in url)
-
-        url = firefox_desktop.get_download_url('release', '45.0', 'linux', 'fr', force_direct=True, funnelcake_id='64')
-        ok_('product=firefox-45.0-f64' not in url)
+        ok_('product=firefox-stub-f64' not in url)
```
jpetto: Would it be beneficial/appropriate to test `win64` here?

bedrock/firefox/tests/test_firefox_details.py
pmac: It's just the way it's called from the `_get_filtered_builds` method. It's potentially possible to remove it, but I'd rather do a more thorough cleanup when/if we can also use an alias for `esr_next`.

chrismore:
@pmac re: the -f[0-9]+ funnelcake IDs appended to the end of product names.

90% of funnelcakes are full installs that have the same bouncer product name as the stub installer. Why? It is easier to have releng create a unique funnelcake build instead of a unique funnelcake build + stub installer. The stub installer by default remotely queries for the latest build, so to create one less moving part, the stub installers use full installers. About 10% of funnelcakes are stub installers if the stub installer is part of the overall user experience being tested, but it doesn't happen that often. 

Here are two examples:

actually downloads the stub installer funnelcake:

https://download.mozilla.org/?product=firefox-stub-f100&os=win&lang=en-US

downloads the full installer for funnelcake 91 even though says "stub" in the URL.

https://download.mozilla.org/?product=firefox-stub-f91&os=win&lang=en-US

Since on bouncer, the -fXXX funnelcake product names are ad hoc, they can point to where ever the releng team decides to point them. 

Another reason for the full vs stub mix when the product name still says "firefox-stub" is to not have to have additional logic on www.mozilla.org that says IF -fXXX then do this product name OR else. The logic is then just on the bouncer side if the URL is confusing.

Should we get Nick Thomas to give some input here?

p.s. I would have replied sooner, but I was on PTO on Friday.

pmac:
@chrismore thanks! that all makes sense to me. So does that mean that we are indeed fine making it so that the `-f###` suffix is only applied to the `-stub` product names? That does make it more simple for me, but it's not how it is at present. If all of the bouncer names for funnelcakes will be called `firefox-stub-f###` then that works great for me. If you need something else just let me know.

chrismore:
@pmac is the default product name for Windows going to remain firefox-stub? In theory, if the product name was "firefox-browser-6-5000", the releng team can map that to whatever binary funnelake or not, so it should be fine with all funnelcakes being to the -f### appended to the end of the -stub product names.

pmac:
@chrismore yes, for windows in every channel (except ESR) the default is `firefox-stub` (or e.g. `firefox-beta-stub`). Thanks!

chrismore:
Makes all sense to me!

pmac:
Some updates:

I discussed some of this with Rail and bouncer handles all of the redirection to the ESR build we use for sha1 downloads (users on WinXP/Vista). People on those platforms always get this ESR regardless of the product they're requesting. So I'm removing the winsha1 from all non-release channel download buttons since it seems disingenuous to have a Win XP download for those buttons since they won't/can't get those channels. But we don't need to use the `firefox-sha1-latest` product names as it's used internally in bouncer.

pmac:
Tests are updated and a [successful integration-tests run](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/104/pipeline/) has happened. I think it's ready for another look.

pmac:
@kyoshino if you have time I'd appreciate your thoughts. I believe you're the most familiar with this area of the code.

PR #5226, opened 2017-10-20
Fix bug 836587: Add "Source code for this page" link to footer

pmac:
It will be a link to the current template source on github.

Original inspiration for this commit from:

https://github.com/l-hedgehog/bedrock/commit/36f011c0

### DIFF
--- a/bedrock/base/templates/includes/site-footer.html
+++ b/bedrock/base/templates/includes/site-footer.html
@@ -16,7 +16,11 @@
           <li><a href="{{ url('mozorg.about') }}" data-link-type="footer" data-link-name="About">{{ _('About') }}</a></li>
           <li><a href="{{ url('mozorg.contact.contact-landing') }}" data-link-type="footer" data-link-name="Contact Us">{{ _('Contact Us') }}</a></li>
           <li><a href="{{ donate_url('footer') }}" class="donate" data-link-type="footer" data-link-name="Donate">{{ _('Donate') }}</a></li>
-          <li><a href="https://wiki.mozilla.org/Webdev/GetInvolved/mozilla.org" data-link-type="footer" data-link-name="Contribute to this site">{{ _('Contribute to this site') }}</a></li>
+          <li><a href="https://wiki.mozilla.org/Webdev/GetInvolved/mozilla.org" data-link-type="footer" rel="nofollow" data-link-name="Contribute to this site">{{ _('Contribute to this site') }}</a></li>
+          {# TODO remove the "and" clause after tranlsation #}
+          {% if template_source_url and (LANG.startswith('en') or switch('footer-source-link')) %}
+            <li><a href="{{ template_source_url }}" data-link-type="footer" rel="nofollow" data-link-name="Source code for this page">{{ _('Source code for this page') }}</a></li>
+          {% endif %}
           <li>
             <ul class="social-links">
               <li><a class="twitter" href="https://twitter.com/mozilla" data-link-type="footer" data-link-name="Twitter (@mozilla)">{{ _('Twitter') }}<span> (@mozilla)</span></a></li>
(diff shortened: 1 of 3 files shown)


pmac:
We'll obviously have to block for l10n for the new string (unless someone has ideas). But this seems pretty straight forward.

/cc @l-hedgehog @alexgibson @jpetto

pmac:
Thoughts anyone? Shall I put up a demo? That seems like a good idea regardless. I'll link a demo here shortly.

pmac:
Okay. You can now see this in action on [my demo](https://bedrock-demo-pmac.us-west.moz.works/en-US/). There is an "edit this page" link in the footer of most pages there.

alexgibson:
I think this looks great, nice work @pmac!

bedrock/base/templates/includes/site-footer.html
```diff
@@ -17,6 +17,9 @@
           <li><a href="{{ url('mozorg.contact.contact-landing') }}" data-link-type="footer" data-link-name="Contact Us">{{ _('Contact Us') }}</a></li>
           <li><a href="{{ donate_url('footer') }}" class="donate" data-link-type="footer" data-link-name="Donate">{{ _('Donate') }}</a></li>
           <li><a href="https://wiki.mozilla.org/Webdev/GetInvolved/mozilla.org" data-link-type="footer" data-link-name="Contribute to this site">{{ _('Contribute to this site') }}</a></li>
+          {% if template_source_url %}
+            <li><a href="{{ template_source_url }}" data-link-type="footer" data-link-name="Edit this page">{{ _('Edit this page') }}</a></li>
```
alexgibson: This adds a new string to the shared footer, so we'll either need to show it for `en` only temporarily, or let this sit while the new string is translated?

jpetto:
Agree that this is a great addition. ⭐️

bedrock/base/templates/includes/site-footer.html
```diff
@@ -17,6 +17,9 @@
           <li><a href="{{ url('mozorg.contact.contact-landing') }}" data-link-type="footer" data-link-name="Contact Us">{{ _('Contact Us') }}</a></li>
           <li><a href="{{ donate_url('footer') }}" class="donate" data-link-type="footer" data-link-name="Donate">{{ _('Donate') }}</a></li>
           <li><a href="https://wiki.mozilla.org/Webdev/GetInvolved/mozilla.org" data-link-type="footer" data-link-name="Contribute to this site">{{ _('Contribute to this site') }}</a></li>
+          {% if template_source_url %}
+            <li><a href="{{ template_source_url }}" data-link-type="footer" data-link-name="Edit this page">{{ _('Source code for this page') }}</a></li>
```
alexgibson: Should we add `rel="nofollow"` to this link, probably useful only for humans?

pmac:
After discussing with @alexgibson on IRC I've changed the link wording to "Source code for this page". That's more accurate as well as possibly less confusing if someone has never seen Github before.

pmac:
Okay. I think we all approve of the change and the new string. 

@peiying2 this PR contains a single new string, but it is in the footer and therefore needs to go into `main.lang`. You may get started on that when ready. Please let me know when it's well-enough translated to get this enabled in production.

pmac:
@alexgibson I've updated it such that the link will show immediately for English locales but hidden with a switch for everything else so that we can enable it only after it's translated broadly enough to enable. Mind taking a quick look at those bits?

flodolo:
I strongly suggest to file a bug to don't forget about removing this conditional.

pmac:
@flodolo would a bug be the best place for the l10n team to update us on the status of this, or should I find it in the l10n tools and keep an eye out myself? If opening a bug with you and Peiying CC'd is the best thing then I'm more than happy to do so.

flodolo:
It should be a bug, since it needs to be addressed by developers. I'd even suggest to assign it to someone, so it ends up in a queue and doesn't get forgotten. Feel free to ping any of us in a couple of weeks to check the status.

PR #5228, opened 2017-10-23
[fix bug 1406299] Only show notification if /whatsnew page URL is out of date

alexgibson:
## Description
Adds a check to make sure a ` /firstrun` or `/whatsnew` page is at least 2 major releases out of date before trying to show a notification.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1406299

## Testing
Out of date notification should only be shown if using an old firefox AND the /firstrun or /whatsnew page is also old (2 major versions out of date).


alexgibson:
@jpetto thanks for the review. i pushed a fix to rename the file 👍

PR #5230, opened 2017-10-24
Packages - adding TocToc - [npm]

tiagomoraismorgado88:
## Packages - adding TocToc - [npm]

**this PR does basically aim at**:
- *adding toctoc nodejs extension*

---

**main motivation behind such does basically lie within:**
- *turning it easy to generate tables of contents out of readme files*

---

</img> </img>

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,755 @@
+{
+  "name": "bedrock",
+  "version": "0.1.0",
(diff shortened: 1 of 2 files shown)


pmac:
Thanks for the contribution, but I don't believe we need this for our readme at the moment.

PR #5233, opened 2017-10-30
[fix bug 1405329] Referral loop thank you page

craigcook:
## Description
Adds a new thank you page variant with options to email, tweet, or share on Facebook. Users will arrive at this page after giving feedback via heartbeat. If they gave a score of 4 or 5, indicated by a `score` URL param (e.g. `?score=5`), they get the thank you page. Lower scores (e.g. `?score=1`) are redirected to SUMO.

Some strings are hard-coded English, but if the test proves successful we may want to open it up to l10n and will need to rework those strings in that case.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1405329

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/en-US/etc/firefox/retention/thank-you-referral/?score=5


jpetto:
At <= tablet viewport, all email links have the Twitter icon:

screen shot 2017-10-30 at 3 23 12 pm

bedrock/etc/templates/etc/firefox/retention/thank-you-referral.html
```diff
@@ -0,0 +1,80 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "base-pebbles.html" %}
```
jpetto: If this is for 57, should the open graph image have the new logo? (When testing the Facebook link, I see it's the old/current logo.)

media/css/etc/firefox/retention/thank-you-referral.scss
```diff
@@ -0,0 +1,200 @@
...
+
+    header {
+        padding: 70px 9px 0;
+        width: 100%;
+        max-width: 1038px;
+    }
+
+    section {
+        display: flex;
+        flex-wrap: wrap;
```
jpetto: As long as this page is only viewed by Fx users (which I believe it is, because heartbeat), this should be fine.

bedrock/etc/templates/etc/firefox/retention/thank-you-referral.html
craigcook: Good call, I'll update.

craigcook:
> At <= tablet viewport, all email links have the Twitter icon

D'oh, I have the icons inside in the desktop media query. 🤦‍♂️

bedrock/etc/templates/etc/firefox/retention/thank-you-referral.html
craigcook: Oh actually it's the Firefox hub page being shared, not this page, so it'll be getting get the new logo in #5212 already.

jpetto: Ah, right. Nice!

PR #5238, opened 2017-11-01
[fix bug 1400328] Update home page for 57.

jpetto:
## Description

Updates the home page for 57 launch.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1400328

## Testing

The usual - evergreen browsers, IE 9+, RTL, responsiveness.

[integration tests](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/107/pipeline/)


peiying2:
@jpetto let me know when it is ready for string extraction.

flodolo:
I think the logic is broken here.

1) You're updating `home/home-new.html` (+2 strings), which is our current template, without using tags.
2) You're extending that template in `home/home-quantum.html`, using the same lang file (`mozorg/home/index-2016`). Which means, any language that has that lang file activated, will start showing that content, in English.

I see two ways out of here:
* Add all the new strings behind a tag.
* Duplicate `home-new.html` as `home-quantum.html`, and use a different lang file for it (`mozorg/home/index-quantum.lang`).

I'd prefer the latter, seems cleaner.

Final consideration: it's Nov 3 where I write (Friday), that means these strings will be exposed for less than a week by the time we finished extracting strings (next week). I hope that's clear to whoever planned these changes to the home page.

bedrock/mozorg/templates/mozorg/home/home-quantum.html
```diff
@@ -0,0 +1,181 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import google_play_button with context %}
+{% extends "base-pebbles.html" %}
+
+{% add_lang_files "mozorg/home/index-2016" %}
```
jpetto: @flodolo - Do we need this?

flodolo: You need to update to

```
{% add_lang_files "mozorg/home/index-quantum" %}
```

jpetto:
Okay, templates have been re-arranged. @peiying2 - strings are final and ready for extraction.

flodolo:
@jpetto Can you update the demo server?

jpetto:
> Can you update the demo server? Strings are still showing up in English.

@flodolo Demo server has been updated!

flodolo:
Reporting from IRC to make sure it doesn't go unnoticed: on mobile, the download button overlaps the menu.

jpetto:
@craigcook We already have a change request in the bug (updated Fx logo with wordmark is coming soon?), but we can update that in a follow-up PR. Peter should QA GA tomorrow, but I think we're good for review now.

Thanks!

alexgibson:
> (updated Fx logo with wordmark is coming soon?)

Fwiw this "wordmark" isn't really a wordmark (I'm not sure if brand really get what one is). The ask is to use "The new Firefox" as an image, hard-coded in English. They want to treat it like a product name, but it's really just a title. We pushed back agains this on the /firefox page, with the help of Flod. Open sans next to the logo using real text is the way to go here.

jpetto:
> Fwiw this "wordmark" isn't really a wordmark (I'm not sure if brand really get what one is). The ask is to use "The new Firefox" as an image, hard-coded in English. They want to treat it like a product name, but it's really just a title. We pushed back against this on the /firefox page, with the help of Flod. Open sans next to the logo using real text is the way to go here.

Sheesh, thanks for the context. Seems as though this decision has not been realized on the creative end (as I've gotten comments in two different bugs about this).

media/css/mozorg/home/home-quantum.scss
```diff
@@ -0,0 +1,742 @@
...
+    h2 {
+        @include font-size-level2;
+        margin-bottom: 1em;
+    }
+
+    p {
+        @include font-size-level5;
+        font-weight: normal;
+        line-height: 1.5;
+        margin: 0 auto 1em;
```
craigcook: @jpetto Sorry, just spotted a small bug I missed earlier. This should switch to `margin: 0 0 1em;` at tablet width, otherwise it ends up centered when it hits its max-width when everything else is left-aligned. I'd open a PR myself but I know we'll be making a few more home updates on Monday so just making a note here.

PR #5239, opened 2017-11-01
[fix bug 1403122] Add Firefox Rocket privacy notice

craigcook:
## Description
Adds a new page for the Firefox Rocket privacy notice. It looks like it's been a while since we updated the legal-docs submodule so this commit picks up a lot of other recent changes in addition to this new policy.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1403122


bedrock/privacy/templates/privacy/notices/firefox-rocket.html
```diff
@@ -0,0 +1,43 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
```
jpetto: It seems this new template is identical to `firefox-quantum.html`, which is identical to `firefox-cliqz-quantum.html`. Should we DRY things up a bit here by making a new base for these pages to inherit from?

craigcook: Good idea. It really is just a new base template. I'll work on that.

chianger:
Can we have this merged now?

PR #5247, opened 2017-11-05
Fix bug 1405042: Newsletter country update page use user-meta basket API

pmac:
Using the update_user like it was doing was clearing out newsletter subscriptions.


pmac:
Relies on https://github.com/mozmeao/basket/pull/52

PR #5249, opened 2017-11-08
Move to CircleCI 2 for PR testing

pmac:


pmac:
This seems to be working well on CircleCI 2.0. It's a good bit faster than our current config for 1.0. Let me know what you think.

/cc @jgmize @metadave

.circleci/config.yml
```diff
@@ -0,0 +1,61 @@
...
+      - run:
+          name: Git Submodule
+          command: |
+            git submodule sync
+            git submodule update --init --recursive
+            echo "$CIRCLE_SHA1" > .current-git-hash
+      - restore_cache:
+          keys:
+            - v1-docker-layers-{{ checksum ".current-git-hash" }}
+            - v1-docker-layers-
```
jgmize: when does this key get saved?

jgmize: found the answer in https://circleci.com/docs/2.0/caching/#restoring-cache

pmac: I agree it's odd, but the system does seem like a good one. It's because once a cache is saved its immutable, so you save a very specific key but then you can restore more loosely. I kinda like it actually. It'd be better if we could use their "docker layer caching" but it's a premium paid add-on and we'll hopefully soon be done with Circle :)

pmac:
Thanks! I'll have to juggle required checks in github before merging this and then likely guide people through it after, so I'll wait until morning to merge. That should also give @glogiotatidis time to weigh in if he'd like to.

metadave:
👍

pmac:
Since this will be a bit disruptive to our workflow (it changes what checks Circle sends to github, thus makes things different for merge requirements) it needs to wait until after the launch of Quantum to merge. Added the DNM label and will remove next week.

docker/bin/docker_build.sh
```diff
@@ -40,7 +40,7 @@ DOCKER_IMAGE_TAG="${DOCKER_REPO}/bedrock_${DOCKERFILE}:${DOCKER_TAG}"
 
 # generate the dockerfile
 rm -f "$FINAL_DOCKERFILE"
-cat "docker/dockerfiles/bedrock_$DOCKERFILE" | envsubst '$GIT_COMMIT' > "$FINAL_DOCKERFILE"
+sed -e "s/\${GIT_COMMIT}/${GIT_COMMIT}/g" "docker/dockerfiles/bedrock_$DOCKERFILE" > "$FINAL_DOCKERFILE"
```
glogiotatidis: tip, not relevant to the Circle v2 change:

Docker now supports `--build-arg=` so we could change  the Dockerfile to 

```
ARG GIT_COMMIT
FROM mozorg/bedrock_base:${GIT_COMMIT}
...
```

and forget about env variable substitution!

pmac: Nice! Thanks! I'll check that out.

PR #5252, opened 2017-11-08
[bug 1410144] Update privacy policy branding

craigcook:
## Description
This updates the privacy policy template with a branded header and also makes it a bit more general-purpose to work with products besides desktop Firefox. There was some refactoring of the JS to separate the part that injects the UItour widget.

## Issue / Bugzilla link
[Template reskin](https://bugzilla.mozilla.org/show_bug.cgi?id=1410144)
[Remove UItour widget from Rocket page](https://bugzilla.mozilla.org/show_bug.cgi?id=1414424)

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/privacy/firefox/


bedrock/privacy/templates/privacy/base-quantum.html
```diff
@@ -9,10 +9,10 @@
 {% endblock %}
 
 {% block page_css %}
-  {% stylesheet 'firefox_privacy' %}
+  {% stylesheet 'privacy_quantum' %}
 {% endblock %}
 
-{% block body_id %}firefox-notice{% endblock %}
+{% block body_class %}privacy-policy {% if switch('firefox-57-release') %}has-quantum{% endif %}{% endblock %}
```
craigcook: Just so we get an automatic logo update when Quantum launches. (Note to self: clean this up later.)

media/css/privacy/privacy-quantum.scss
```diff
@@ -6,127 +6,104 @@
...
-            padding-left: 17px;
-            max-width: 1024px;
-        }
+        padding: 60px 20px;
     }
 
-    h1 {
-        display: inline-block;
-        @include font-size-level4;
+    a:link,
```
alexgibson: nit: would be good to add a hover state here for the link in the header text?

media/css/privacy/privacy-quantum.scss
```diff
@@ -6,127 +6,104 @@
     '../pebbles/includes/lib',
     '../pebbles/includes/fonts/open-sans';
 
-article {
+body.privacy-policy {
```
alexgibson: Is the aditional class on `body` needed here?

media/js/privacy/privacy-quantum.js
```diff
@@ -1,37 +1,13 @@
 (function() {
     'use strict';
 
-    var mainContent = document.querySelector('main');
+    var mainContent = document.querySelector('.privacy-body');
```
alexgibson: Hmm, we should probably also move these DOM queries to within a basic JS feature detect?

media/js/privacy/privacy-quantum-firefox.js
```diff
@@ -0,0 +1,55 @@
+(function() {
+    'use strict';
+
+    if (document.querySelectorAll) {
```
alexgibson: This should be `'querySelectorAll' in document`. Instead of wrapping it around the variable declarations here however, it may be simpler/neater to write a small function like:

```javascript
function supportsBaselineJS() {
    return 'querySelectorAll' in document &&
           'addEventListener' in window;
}
```

You could then use it like:

```javascript
if (supportsBasicJS()) {
    // execute page JS here
}
```

You could also apply this to the other JS file.

alexgibson:
Nav also still shows the old style download button?

media/js/privacy/privacy-quantum-firefox.js
craigcook: I think I might have it right this time... Schalk wrote all this originally so I'm just butchering his stuff without really knowing what I'm doing 😟

media/js/privacy/privacy-quantum-firefox.js
```diff
@@ -0,0 +1,61 @@
...
+                return;
+            }
+
+            Mozilla.UITour.openPreferences('privacy-reports');
+        });
+    }
+
+    // Don't execute if features aren't supported and client isn't desktop Firefox
+    if (supportsBaselineJS() && Mozilla.Client.isFirefoxDesktop) {
+        var topicHeaders = document.querySelectorAll('.privacy-body .content-girdle > section');
```
alexgibson: If these are to be considered page scoped variables, it's good practise to leave the declarations at the top of the file for readability. So at the top of the file just`var topicHeaders;`. Then do the assignments here e.g. `topicHeaders = document.querySelectorAll('.privacy-body .content-girdle > section');`

You also still need to do this for the other file.

I do see the pill-shaped button now (probably my bad last time).

PR #5255, opened 2017-11-09
[bug 1397381] Update /firefox/new/ headline

alexgibson:
## Description
- Updates the headline on `/firefox/new/` to match the headline on `/firefox/`.
- Uses l10n tag `firefox_quantum_new_headline`.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1397381#c51

## Testing
Check for copy pasta?

### DIFF
--- a/bedrock/firefox/templates/firefox/new/quantum/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/quantum/scene1.html
@@ -27,8 +27,13 @@
 
 {% block main_content %}
 <header>
+{% if l10n_has_tag('firefox_quantum_new_headline') %}
+  <h1>{{ _('The new <strong>Firefox</strong>') }}</h1>
+  <h2>{{ _('Fast for good.') }}</h2>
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/new/quantum/scene1.html
```diff
@@ -27,8 +27,13 @@
 
 {% block main_content %}
 <header>
+{% if LANG.startswith('en') %}
```
flodolo: Doesn't it make more sense to put it behind a tag, if the goal is to expose it later? We can just wait to expose the string on our side, and there will be no code changes needed.

Only downside: en-GB and en-ZA would show the old tagline. Truth is, it's highly unlikely they'll translate the page, so they'll fall back to en-US anyway.

alexgibson: Ok will use a tag 👍

PR #5224, opened 2017-10-20
[fix bug 1397520] Add /firefox/mobile/ page.

jpetto:
## Description

Adds `/firefox/mobile/` page.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1397520

## Testing

https://bedrock-demo-jpetto-fxmobile.us-west.moz.works/firefox/mobile/

Should get the standard run-through: evergreen browsers, IE 9+, RTL, responsiveness.

The [existing redirect](https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/redirects.py#L146) from `firefox/mobile/` to `/firefox/android/` should continue to exist when the switch is off.

Interaction notes:

- The Firefox modal will show the Send to Device widget for all [pre-defined locales](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L1226). To see the QR code for Firefox, change the locale in the URL to a locale that does not suppor the widget, e.g. `/it/`.

- The Focus modal should always show the QR code for desktop.

- If on Android or iOS, both modals (Firefox and Focus) should show app store badges only. You can test on an iOS or Android device, or use the dev tools to change "osx" (or "windows") on the `<html>` element's class attribute to either "android" or "ios".

[integration test run](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/106/pipeline/)

PR #5260, opened 2017-11-10
Bug 1415544 - show QR code to alternative mobile app store for Fx Chi…

l-hedgehog:
…na repack.

## Description
Certain pages show a QR code to Firefox at iOS App Store or Google Play Store, Beijing office would like to change it to provide the our customized build of Firefox for Android if an user visits such pages with Fx China repack.

We're using the url of Tencent app store since we believe WeChat is the QR code scanner of choice in China, and Tencent's own app store works best with it.

PR already updated after the merge of #5224.

## Issue / Bugzilla link
https://bugzil.la/1415544

## Testing
* Setup UITour permission for test 1
* Set "distribution.id" with Services.prefs.getDefaultBranch("distribution.").setCharPref("id", "MozillaOnline"); in Firefox's browser console.
* Visit /firefox/{54-56}/whatsnew/ or the future /firefox/mobile/, the QR code shown should link to http://a.app.qq.com/o/simple.jsp?pkgname=org.mozilla.firefox


alexgibson:
The QR code in this PR forwards me to an unsecure address: http://a.app.qq.com/o/simple.jsp?pkgname=org.mozilla.firefox

Any reason why we can't use the https alternative? https://a.app.qq.com/o/simple.jsp?pkgname=org.mozilla.firefox

There seem to be some mixed content issues, but this would be better than an unsecure redirect?

l-hedgehog:
Hi @alexgibson, did you see my comment in https://bugzil.la/1415544#c14 ? Our contact at Tencent claimed their https endpoint is unsupported.

alexgibson:
> Hi @alexgibson, did you see my comment in https://bugzil.la/1415544#c14 ? Our contact at Tencent claimed their https endpoint is unsupported.

This may be a problem :/ (I also filed https://bugzilla.mozilla.org/show_bug.cgi?id=1419022 for our existing QRCodes - thanks for alerting!)

alexgibson:
Hi @l-hedgehog, any update here on finding a suitable https endpoint? Can we just point directly to the relevant app stores if Tencent don't support https?

l-hedgehog:
Hi @alexgibson , without access to Google Play Store, we have many Android app stores here in China. For each release we upload our build to all major stores, and we're providing direct apk links on our own website.

WeChat, which we believe is the most popular QR code scanner among Chinese users, blocks downloading of any apk not hosted at Tencent app store. This kinda makes it the most relevant app store.

To minimize the non-secure connection, it looks like I could create a new page at https://www.firefox.com.cn/ and use JavaScript to:
* redirect users on iOS to Apple iOS App Store;
* redirect users using WeChat on Android to Tencent App Store (http);
* leave everybody else on this page and let them try downloading the apk directly (maybe just a clone of /mobile/).

By doing this, we're limiting the http endpoint to WeChat on Android users, and we would be able to better understand what QR code scanners are being used and make adjustments accordingly.

I don't think using https://a.app.qq.com/o/simple.jsp?pkgname=org.mozilla.firefox instead of its http equivalent helps, since the apk is served over http anyway.

l-hedgehog:
> I could create a new page at https://www.firefox.com.cn/ and use JavaScript to ...

This page and the corresponding QR code went live two days ago at http://www.firefox.com.cn/mobile/

> and we would be able to better understand what QR code scanners are being used

The initial result from yesterday:
* 55%+ of visits on Android are using WeChat, with UC Browser being a distant second.
* Similarly, ~2/3 of visits on iOS are using WeChat.

l-hedgehog:
Hi @alexgibson 

I've updated the PR to use the new QRCode pointing to https://www.firefox.com.cn/mobile/qrcode-redirect/ which then detects and redirects based on user agent, as described in my previous comments.

This will reduce the exposed users (those redirected to unsecure Tencent app store) by half.

PR #5268, opened 2017-11-11
[fix bug 1400328] 57 home page - update dino img & fix CSS bug.

jpetto:
## Description

Fixes [CSS bug](https://github.com/mozilla/bedrock/pull/5238#discussion_r150369980) for desktop+ viewports and replaces choppy dino PNG with SVG.


bedrock/mozorg/templates/mozorg/home/home-quantum.html
```diff
@@ -76,7 +76,7 @@
   <section class="intro-section" id="mozilla">
     <div class="content">
       <div class="intro-image">
-        <img src="{{ static('img/home/quantum/dino.png') }}" alt="" />
+        <img src="{{ static('img/home/quantum/dino.svg') }}" alt="" />
```
craigcook: I think this needs width and height attributes. Chrome and IE are showing the dino tiny for some reason, maybe some default sizing when size can't be calculated from the image data. So the SVG file probably needs width and height attributes as well, since it looks like some browsers don't grok viewBox. Firefox shows it at the viewBox size on wide desktop sizes, which is probably too big (the `max-size: 100%;` kicks in and scales it down at smaller breakpoints).

PR #5269, opened 2017-11-11
[fix bug 1416496] Redirect mobile product pages to new mobile page.

jpetto:
## Description

Redirects `/firefox/android/`, `/firefox/focus/`, and `/firefox/ios/` to `/firefox/mobile/` for 57 launch.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1416496

## Testing

Ensure correct pages load with switch on and off.

### DIFF
--- a/bedrock/firefox/tests/test_views.py
+++ b/bedrock/firefox/tests/test_views.py
@@ -1053,3 +1053,50 @@ class TestFirefoxQuantumPageRedirect(TestCase):
         resp = views.quantum(req)
         eq_(resp.status_code, 301)
         ok_(resp.url.endswith('/en-US/firefox/'))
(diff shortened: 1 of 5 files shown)


jpetto:
> Looks good, but this is likely to break functional tests for the existing mobile product pages?

Hm, right. Should we just remove those tests? Adding a `skipif` doesn't seem to make much sense if the pages will be gone for good after we flip the switch.

alexgibson:
I’ve been adding skips so far simply because it’s quicker. We can always remove the tests along with the actual page code after release.

jpetto:
Cool - just added the `skipif`s.

PR #5265, opened 2017-11-10
Updates new Fx home page copy to be more inclusive.

jpetto:
## Description

Updates copy to be more inclusive.

@peiying2 has been made aware via IRC.

PR #5274, opened 2017-11-13
[fix bug 1416805] Update copy on Fx home page.

jpetto:
## Description

Updates copy based on request from MoFo.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1416805

### DIFF
--- a/bedrock/firefox/templates/firefox/hub/home-quantum.html
+++ b/bedrock/firefox/templates/firefox/hub/home-quantum.html
@@ -358,12 +358,12 @@
         <div class="browse-for-good-section-content">
           <h2 class="section-title">{{ _('Browse for good') }}</h2>
           <p>
-          {% if l10n_has_tag('firefox-home-copy-update-20171110') %}
+          {% if l10n_has_tag('firefox-home-copy-update-20171114') %}
             {% trans %}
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/hub/home-quantum.html
```diff
@@ -358,7 +358,14 @@ <h2 class="section-title">{{ _('Sync your devices') }}</h2>
...
           <p>
-          {% if l10n_has_tag('firefox-home-copy-update-20171110') %}
+          {% if l10n_has_tag('firefox-home-copy-update-20171114') %}
+            {% trans %}
+              Firefox is made by Mozilla, the non-profit champions of a healthy internet.
+              Mozilla also tackles issues like privacy, misinformation and trolling by
+              investing in fellowships, campaigns and new technologies designed to make
+              the internet healthier.
+            {% endtrans %}
+          {% elif l10n_has_tag('firefox-home-copy-update-20171110') %}
```
flodolo: We didn't expose strings for #5265 yet. You can get rid of the this section, and only leave new string and old one (with girls instead of women).

jpetto: 👍

PR #5275, opened 2017-11-13
bug 1397532 change rtl treatment of wave and add white background

ericawright:
@craigcook 

## Description
Small visual tweaks for the firstrun page. 
Wave 7 stays to the left for rtl and ltr locales.
Add white background to show through for mix-blend-mode. so we never see colour off of the waves.
connect messaging for fxa-iframe hiding and showing the skip button when appropriate
shrink iframe to tested size to enable multi-column layout

### DIFF
--- a/media/css/firefox/firstrun/firstrun_quantum.less
+++ b/media/css/firefox/firstrun/firstrun_quantum.less
@@ -23,7 +23,10 @@
 body {
     margin: 0;
     overflow: hidden;
-    font-family: 'Fira Sans Light', sans-serif;
+}
+
+* {
+  font-family: 'Fira Sans Light', sans-serif;
 }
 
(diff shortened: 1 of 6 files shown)


media/css/firefox/firstrun/firstrun_quantum.less
```diff
@@ -23,7 +23,10 @@
 body {
     margin: 0;
     overflow: hidden;
-    font-family: 'Fira Sans Light', sans-serif;
+}
+
+* {
```
craigcook: D'oh, I guess this did need to be universal after all. Sorry for that.

ericawright: no worries, I had forgotten

PR #5281, opened 2017-11-14
Fix bug 1416825: Add iOS link to release notes

pmac:
This was removed in 900fc99d24044e66b82d9c85717c665dea2e39f6. We can now add it back since iOS release notes are now being maintained.


pmac:
@alexgibson logo update added

PR #5284, opened 2017-11-14
[fix bug 1413995] Add Reggie Watts /new campaign page.

jpetto:
## Description

Adds Reggie Watts campaign download page.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1413995

## Testing

The usual - IE 9+, evergreen browsers. `en-US` only.


craigcook:
At extra-wide sizes the edges of the photo are obvious. I wonder if we can get a slightly wider crop of that image and blend the edges out to a solid color that can match the background.

reggie-image-edges

And at smaller sizes it's weird not to have his picture on the page when the tagline is "Reggie fast" ... like, who is Reggie in that context? The tagline doesn't make sense without the picture. I don't suppose we got any design guidance for this page in small viewports? Maybe we can come up with something quick and simple that still gets his face on it without going into a whole new mockup/review cycle.

media/css/firefox/new/reggie-watts.scss
```diff
@@ -0,0 +1,289 @@
...
+                margin-bottom: 0;
+            }
+        }
+
+        h3 {
+            margin-bottom: 20px;
+        }
+
+        p {
+            line-height: 30px;
```
craigcook: Should usually avoid fixed-unit line-heights. If you want to bump this up from the global setting `line-height: 1.66;` would get you 30px at an 18px font size.

jpetto:
@craigcook Updated the CSS and got a new image at 1600px wide.

The sub-head changed, so the copy now makes sense with and without the photo of Mr. Watts.

craigcook:
r+! 🎤

PR #5285, opened 2017-11-15
Home page Quantum cleanup

craigcook:
## Description
This started as a quick fix for RTL but while I was already touching the home page I figured I would remove the `firefox-57-release` switch, and while I was doing that I figured I'd go ahead and delete the old home template as this is going to be the home page for the near future at least until the next redesign. I also removed the Quantum promo from the old home page and made some RTL tweaks there. Some locales still get the old home page but lots have already localized the new one.

## Issue / Bugzilla link
[RTL bug 1417121](https://bugzilla.mozilla.org/show_bug.cgi?id=1417121)

## Testing
Check RTL layout on both new and old home pages. You'll need to switch to `Dev=False` to get the old page.


craigcook:
Hrm probably need to update some tests, too..

bedrock/mozorg/tests/test_views.py
```diff
@@ -350,14 +350,15 @@ def test_technology_locale_template(self, render_mock):
...
-    def test_home_enUS(self, render_mock):
+    def test_home_quantum_template(self, render_mock):
         request = RequestFactory().get('/')
         request.locale = 'en-US'
         views.home(request)
-        render_mock.assert_called_once_with(request, 'mozorg/home/home-new.html')
+        render_mock.assert_called_once_with(request, 'mozorg/home/home-quantum.html')
 
-    def test_home_non_enUS(self, render_mock):
+    @patch.object(views, 'lang_file_is_active', lambda *x: False)
```
craigcook: No idea if I did this right, I copied from elsewhere, but from context it seems like this makes the test behave as if the lang file is NOT active and loads the fallback template. Am I guessing right? It passes locally so I hope this works.

media/css/mozorg/home/fx-quantum-promo.scss
```diff
@@ -1,62 +0,0 @@
-// This Source Code Form is subject to the terms of the Mozilla Public
-// License, v. 2.0. If a copy of the MPL was not distributed with this
-// file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
-@import '../../pebbles/includes/lib';
-
-.quantum-promo {
-    background-color: #45caf3;
-    @include at2x('/media/img/mozorg/promo/quantum-promo-banner-phone.png', 760px, 200px);
```
jpetto: Can delete this image (and the one below) as well?

bedrock/mozorg/tests/test_views.py
jpetto: Looks good to me 😄

jpetto:
RTL has a weird hover state on the nav, which currently exists in production:

screen shot 2017-11-15 at 10 16 03 am

Not related to this PR, but maybe an easy fix? If not, we can file a bug and get this out.

media/css/mozorg/home/fx-quantum-promo.scss
craigcook: Doh! Good catch.

PR #5287, opened 2017-11-15
[fix bug 1416725] Update Firefox site footer links

alexgibson:
## Description
- Updates site footer links to reflect Quantum page changes.
- All strings are already in [main.lang](https://raw.githubusercontent.com/mozilla-l10n/www.mozilla.org/master/en-US/main.lang), so l10n should not be required?

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1416725

## Testing
Make sure translated strings show up as expected?

### DIFF
--- a/bedrock/base/templates/includes/site-footer.html
+++ b/bedrock/base/templates/includes/site-footer.html
@@ -35,19 +35,11 @@
           <a href="{{ url('firefox') }}" data-link-type="footer" data-link-name="Firefox">{{ _('Firefox') }}</a>
         </h5>
         <ul class="firefox-links">
-        {% if LANG.startswith('en') %}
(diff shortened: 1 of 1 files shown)


alexgibson:
One thought - should we pass this change by our SEO guru's? We no longer use the word "browser" in our sub navigation links, so this change does seem consistent from a user perspective. I don't know if our period of browser-itis actually yielded any measurable results, but it seems like it's no longer appearing everywhere in our copy.

I'm happy to go with this change as-is, as it feels like the right thing to do. But I'm also not adverse to seeking further approval.

jpetto:
I agree this is the right thing to do, and is consistent with new language being used. We can always revert?

jpetto:
Waiting to merge just in case L10n has input, and to see if maybe Peter wants to weigh in on if having "desktop" in there was a benefit.

jpetto:
After convo in Slack reminded me we removed "browser" from the secondary nav, I think we're good to go here.

PR #5295, opened 2017-11-16
[bug 1417888] Disable firstrun animation for some platforms

craigcook:
## Description
This disables the background wave animation in Linux and older Windowses (less than Win8), as well as a simplified end transition. Macs and newer Windowses (Win8 and up) should appear the same as before.

This is hopefully a temporary patch to address the most pressing concerns reported on Linux and older Windows machines that lack GPU acceleration until we can regroup and look into revising the page for better performance across all platforms and devices.

The diff looks a bit bigger than it really is because a lot of lines just changed indentation.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1417888

## Testing
https://www-demo4.allizom.org/firefox/57.0/firstrun/

This uses UITour so you'll need to whitelist test domains in `browser.uitour.testingOrigins`

Test the page in Firefox 57+ on Windows 7 and verify correct FxA form functionality with no background animation. On either signing in or skipping, there should be a simple fade-to-white end transition before loading the new tab page. We should test Linux as well since I don't have access to a Linux machine.

Test the page in Firefox 57+ on Windows 8 or better (as well as Mac) and verify no regressions from what's currently in production. You should get the rolling waves background animation and the spin-and-fade end transition before loading the new tab page.

### DIFF
--- a/media/js/firefox/firstrun/firstrun_quantum.js
+++ b/media/js/firefox/firstrun/firstrun_quantum.js
@@ -5,6 +5,11 @@
 (function (Mozilla) {
     'use strict';
 
+    // Add class to indicate Windows 8 and later
+    if (window.site.platform === 'windows' && window.site.platformVersion >= 6.2) {
+        document.documentElement.classList.add('win8up');
+    }
+
     var beginAnimation = function (syncConfig) {
         var redirectDest = 'about:home';
 
@@ -31,10 +36,8 @@
 
         var onVerificationComplete = function () {
             scene.dataset.signIn = 'true';
-            document.getElementById('wave1').addEventListener('animationend', function(event) {
-                if (event.animationName === 'Expand1') {
-                    window.location.href = redirectDest;
-                }
+            document.getElementById('white-overlay').addEventListener('transitionend', function() {
+                window.location.href = redirectDest;
             }, false);
         };
 
--- a/bedrock/firefox/templates/firefox/firstrun/firstrun_quantum.html
+++ b/bedrock/firefox/templates/firefox/firstrun/firstrun_quantum.html
@@ -30,7 +30,7 @@
       <rect x="0" y="0" width="1440" height="800"/>
     </svg>
 
-    <svg id="wave1" class="wave1 background" xmlns="http://www.w3.org/2000/svg" viewBox="-150 -210 1750 1300" preserveAspectRatio="none">
+    <svg class="wave1 background" xmlns="http://www.w3.org/2000/svg" viewBox="-150 -210 1750 1300" preserveAspectRatio="none">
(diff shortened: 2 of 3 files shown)


craigcook:
Or more accurately, this *enables* the animation only for Mac or Windows 8+, any others should get the simplified version.

media/js/firefox/firstrun/firstrun_quantum.js
```diff
@@ -31,10 +36,8 @@
 
         var onVerificationComplete = function () {
             scene.dataset.signIn = 'true';
-            document.getElementById('wave1').addEventListener('animationend', function(event) {
-                if (event.animationName === 'Expand1') {
-                    window.location.href = redirectDest;
-                }
+            document.getElementById('white-overlay').addEventListener('transitionend', function() {
```
craigcook: Changed this because "wave1" no longer animates on all platforms, but the white overlay does transition.

media/js/firefox/firstrun/firstrun_quantum.js
```diff
@@ -5,6 +5,11 @@
 (function (Mozilla) {
     'use strict';
 
+    // Add class to indicate Windows 8 and later
+    if ((window.site.getPlatform() === 'windows') && (parseFloat(window.site.getPlatformVersion()) >= 6.2)) {
+        document.documentElement.className += ' win8up';
```
alexgibson: As this is Firefox only, you can use `document.documentElement.classList.add('win8up');`

media/js/firefox/firstrun/firstrun_quantum.js
```diff
@@ -5,6 +5,11 @@
 (function (Mozilla) {
     'use strict';
 
+    // Add class to indicate Windows 8 and later
+    if ((window.site.getPlatform() === 'windows') && (parseFloat(window.site.getPlatformVersion()) >= 6.2)) {
```
alexgibson: You can simplify this to `if (window.site.platform === 'windows' && window.site.platformVersion >= 6.2)`

Natim:
capture

Here is what I see when connecting to your demo page. It is probably because I am already connected to Sync.

In the current implementation I see the animation and then I am redirected to a new tab page.

I don't know if your page should do the same in case the animation don't run. Just though of mentioning it.

nical:
I can confirm that this removes the animation on my linux laptop.

media/js/firefox/firstrun/firstrun_quantum.js
ericawright: in that case "wave1" id can be removed, it wasn't used for anything else.

craigcook:
@alexgibson Updated!

PR #5296, opened 2017-11-16
[fix bug 1415295] Add fundraiser banner to home page.

jpetto:
## Description

Adds fundraiser banner to the top of the home page. Also updates "Our Impact"/"Our Innovations" sections with new items.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1415295

## Testing

Ensure banner looks good on old IE and modern browsers.

I'm using `localStorage` to only show the banner once per session, so maybe test in private browsing mode. (And lmk if we should extend the hiding beyond the session.)


jpetto:
@peiying2 / @flodolo - There's some...interesting stuff going on in the string for the donate section. Please ask in IRC/Slack/here if you have any questions.

bedrock/mozorg/templates/mozorg/home/home-quantum.html
```diff
@@ -32,6 +32,71 @@
...
+          <label for="onetime">
+            <input type="radio" value="single" id="onetime" name="frequency" checked> {{ _('One-time') }}
+          </label>
+
+          <input type="hidden" name="currency" value="{{ donate_params.currency }}" />
+          <input type="hidden" name="presets" value="{{ donate_params.presets }}" />
+
+          {# GA params #}
+          <input type="hidden" name="ref" value="EOYFR2015" />
+          <input type="hidden" name="utm_campaign" value="EOYFR2015" />
```
TheoChevalier: I know it’s a WIP, but heads up that we will want an updated tag here. `EOYFR2017` should be okay, unless @vojtechmzl would prefer something else

jpetto:
Okay, aside from having Peter review the GA stuff (which should hopefully be quick), we're ready for review here. This needs to go out Friday 12/1.

@alexgibson Have time for a review tomorrow?

alexgibson:
> @alexgibson Have time for a review tomorrow?

Sure, I'll review this first thing

alexgibson:
We're getting a bit notification overloaded with both this and the out of date banner. Think we should disable the latter while this is running? :/

screenshot-2017-11-30 internet for people not profit

alexgibson:
> I'm using localStorage to only show the banner once per session, so maybe test in private browsing mode. (And lmk if we should extend the hiding beyond the session.)

Could we maybe use a session cookie instead of `sessionStorage`? Just an idea, not a blocker. It could be a bit more performant, and also extends longer than the life of the current tab?

media/css/mozorg/home/fundraiser2017.scss
```diff
@@ -0,0 +1,206 @@
...
+}
+
+#fundraiser {
+    background: #075f60 url('/media/img/home/fundraiser2017/background.png') center left no-repeat;
+    background-size: cover;
+    color: #fff;
+    padding: 40px 20px;
+    text-align: center;
+
+    // hide by default if JS is available to avoid flicker
```
alexgibson: The only noticable downside to this is that there is a page jump when the donate section appears. It's not a big thing, but on slow connections (esp mobile) it could be a bit jarring maybe? This may be a little late design-wise, but it does make me wonder if it could be nicer to treat this like a fully-fledged section of the page, underneath the nav, and not like a notification or popup?

media/js/mozorg/home/fundraiser-eoy2017.js
```diff
@@ -0,0 +1,31 @@
...
+    var wasClosed = false;
+
+    // see if visitor previously dismissed the banner
+    try {
+        wasClosed = sessionStorage.getItem(storageKey);
+    } catch(ex) {
+        // fail silently
+    }
+
+    if (!wasClosed) {
```
alexgibson: This isn't the fault of this PR, but fwiw I can't get the actual donate.mozilla.org website to work on any version of IE. Edge seems to work fine, as does Chrom / Safari / Firefox. Should we not show the form for any version of IE?

media/css/mozorg/home/fundraiser2017.scss
jpetto: You aren't wrong on any count above, including being late design-wise. ;)

I think it's better to be jarring coming in for users that haven't seen it yet than going out for users who have previously dismissed it.

media/js/mozorg/home/fundraiser-eoy2017.js
jpetto: Hm, appears to be a TLS issue with IE 9 and 10 (IE 11 works for me). I can get the site to work in IE 9 and 10, but only if I go into the security section and check TLS 1.1.

I [filed a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1422119), but will try to reach out via chat to see if this is something we should restrict here or if MoFo can fix on their end.

PR #5298, opened 2017-11-18
Make update_wordpress command efficient with DB writes

pmac:
Used to delete all entries before adding again from the server.  Now uses the server post ID and modified time to only update those that need updating and removing older posts.

### DIFF
--- a/bedrock/wordpress/models.py
+++ b/bedrock/wordpress/models.py
@@ -6,20 +6,43 @@ import operator
 import random
 
 from django.conf import settings
-from django.db import models, transaction
+from django.db import models
 from django.db.models import Q
 from django.db.utils import DatabaseError
+from django.utils.dateparse import parse_datetime
+from django.utils.timezone import make_aware, utc
 
 import bleach
 from django_extensions.db.fields.json import JSONField
 from jinja2 import Markup
 from raven.contrib.django.raven_compat.models import client as sentry_client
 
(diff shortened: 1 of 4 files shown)


pmac:
This is in support of more efficient DB updates so that we can use a sqlite db in prod. This will make it easier to tell when things have updated with a checksum of the db file.

bedrock/wordpress/models.py
```diff
@@ -58,38 +81,59 @@ def filter_by_blogs(self, *blog_slugs):
...
+                else:
+                    self.create(**post)
+            except DatabaseError:
+                sentry_client.captureException()
+                raise
+
+            update_count += 1
+
+        # clean up after changes
+        if update_count:
```
glogiotatidis: There're may be expired posts without any updates, right? In case a blog post was just deleted?

pmac: Could be. It's not critical that these be deleted in a timely manner. This is here just to make sure the DB doesn't grow much larger than it needs to. We pull in 20 posts by default, but the most we every use on a page is 5 at the moment. The blogs are active enough that this should be okay IMHO. If you think it's worth a check every time I'm fine with doing that as well.

glogiotatidis: I would check every time just because things change but I won't insist on that, if you feel it's ok keep is the way it is.

PR #5302, opened 2017-11-20
[fix bug 1416706] Remove old Firefox Desktop page template

alexgibson:
## Description
- Removes old `/firefox/desktop/` page template, which is now pernamently redirected to `/firefox`.
- Also removes the `firefox-57-release` switch from the Firefox home page view, global nav, and Firefox sub-nav.

Note: the old mobile product pages are being removed in [bug 1418500](https://bugzilla.mozilla.org/show_bug.cgi?id=1418500) 

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1416706

## Testing
- Make sure i didn't miss anything, or remove something still in use?


alexgibson:
Successful test run: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/122/pipeline/

bedrock/firefox/views.py
```diff
@@ -851,7 +834,7 @@ class FirefoxHubView(BlogPostsView):
     def get_template_names(self):
         locale = l10n_utils.get_locale(self.request)
 
-        if switch('firefox-57-release') and lang_file_is_active('firefox/hub/home-quantum', locale):
+        if lang_file_is_active('firefox/hub/home-quantum', locale):
```
jpetto: Should we do away with this whole "hub" nomenclature?

alexgibson: Hmm, I don't think we can rename lang files without breaking things?

jpetto: Oh, sorry! I was talking about the bedrock directory structure/file naming, not really the lang files (I just had to pick spot in the file where GitHub would let me comment). Seems a little odd to have the Firefox home page in a `hub` directory.

alexgibson:
Thanks for the review @jpetto - I removed the `{% if LANG.startswith('en-') %}` conditionals from the old hub page. For now, I think  I'm going to leave the filenames / directories as they are (so that they are consistent with the lang file). Once we can remove the old page and no longer need the conditional around, we can tidy up where things live?

PR #5304, opened 2017-11-20
[fix bug 1419125] Ensure string values in utmParamsFxA.

jpetto:
## Description

Forces all values in the object returned by `utmParamsFxA` to be strings, as `UITour.showFirefoxAccounts` fails if any values are numeric.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1419125

## Testing

Ensure no regressions?

### DIFF
--- a/media/js/base/search-params.js
+++ b/media/js/base/search-params.js
@@ -81,5 +81,11 @@ _SearchParams.prototype.utmParamsFxA = function(pathname) {
 
     /* eslint-enable camelcase */
 
+    // ensure all values are strings, as no numeric values are allowed
+    // into UITour.showFirefoxAccounts
+    Object.keys(utms).forEach(function(key) {
+        utms[key] = utms[key].toString();
+    });
+
     return utms;
 };
--- a/tests/unit/spec/base/search-params.js
+++ b/tests/unit/spec/base/search-params.js
@@ -59,6 +59,15 @@ describe('search-params.js', function() {
             expect(utms.utm_content).toEqual('/firefox/sync/');
         });
 
+        it('should return an object of string utm_ values for FxA', function () {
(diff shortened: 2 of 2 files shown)


media/js/base/search-params.js
```diff
@@ -81,5 +81,11 @@ _SearchParams.prototype.utmParamsFxA = function(pathname) {
 
     /* eslint-enable camelcase */
 
+    // ensure all values are strings, as no numeric values are allowed
+    // into UITour.showFirefoxAccounts
+    Object.keys(utms).forEach(function(key) {
+        utms[key] = utms[key].toString();
```
alexgibson: Could you avoid enumerating through the keys twice if you did `utms[key] = val.toString();` on line 51 inside `_SearchParams.prototype.utmParams`?

jpetto: Since `_SearchParams.prototype.utmParams` is also used by our stub attribution script, I was trying to avoid regressions there (as I'm not super familiar and this bug is currently causing issues in prod). Also, I'm not sure it's "correct" to force strings for all values all the time - FxA seems to be the only place strings are required.

What do you think?

alexgibson: Hmm, ok I guess? I thought `toString()` should have been pretty safe, but I guess we could always err on the side of caution? How confident are we that this function will only ever get called via a browser that supports `forEach` on an Array?

jpetto: Pretty confident? It should only be called for Firefox, right? Happy to go back to an earlier solution if you see gaps in my logic though.

alexgibson: Good enough for me 🍡

PR #5306, opened 2017-11-21
[fix bug 1418500] Remove orphaned mobile product pages.

jpetto:
## Description

Removes old mobile product pages and related assets. Also updates redirects and links related to the deprecated pages.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1418500

## Testing

Make sure all things to be removed were, and that nothing removed is still needed. (That reads like a riddle.)

Tried to run integration tests, but the [job](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/128/pipeline/) seems to be frozen (2nd attempt).


bedrock/firefox/templates/firefox/hub/home.html
```diff
@@ -131,10 +131,10 @@ <h2 class="card-heading">{{ _('Take Firefox to go!') }}</h2>
               </a>
 
               <p>
-              {% trans url1=url('firefox.ios'), url2=url('firefox.android.index') %}
+              {% trans url=url('firefox.mobile') %}
                 Get your favorite bookmarks, passwords & more on any
-                <a href="{{ url1 }}">iOS</a> and
-                <a href="{{ url2 }}">Android</a> device with Firefox Sync.
+                <a href="{{ url }}">iOS</a> and
```
alexgibson: This is going to break translations I think. You might need to keep the same variable names to avoid altering the string?

jpetto: I was wondering if this would result in a string change. I'll go back to two variables here.

PR #5316, opened 2017-11-29
feat(fxa): Add the email-first experiment to the firstrun page

shane-tomlinson:
@alexgibson - here's the WIP I mentioned in bugzilla.

## Description

Add the FxA email-first experiment to the quantum firstrun page.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1405822
https://docs.google.com/document/d/1seUQEXOR-JGD6ff77iAP8lzDjW-BPengXugUzo-p2UQ

## Testing

I have yet to add tests, and have lots of questions.

* Are `a` and `b` typical names for a variation? I see a combination of `a`, `b`, and `1`, `2`, `3`, etc.
* When an A/B test changes copy, do you use a single template for both variants or create separate templates? I went with a single template and if statements to control which copy is displayed.
* The [way I set the template context](https://github.com/mozilla/bedrock/compare/master...shane-tomlinson:fxa-email-first-experiment?expand=1#diff-784e0d4de42af29595e70315a6be5337R422) seems like it should be extracted, but I'm not sure how you usually handle that.
* Since the A/B test is limited to en-US, do I need to surround the treatment group text with  `_('')`
* If the new strings should be translated now, is there a big string table or are they extracted from the templates?
* I have a content-server fix that reduces the large gap between the FxA submit button and the "submit" button. This shrinks the iframe quite a bit and makes the box appear to be too high. Would it be alright to use flexbox to vertically center the iframe?

fixes mozilla/fxa-bugzilla-mirror#381

### DIFF
--- a/bedrock/firefox/templates/firefox/firstrun/email-first-experiment/index-a.html
+++ b/bedrock/firefox/templates/firefox/firstrun/email-first-experiment/index-a.html
@@ -1,0 +1,5 @@
+{% extends "firefox/firstrun/firstrun_quantum.html" %}
+
+{% block iframe %}
+  <iframe id="fxa" scrolling="no" data-src="{{ settings.FXA_IFRAME_SRC }}signin?utm_campaign=fxa-embedded-form&amp;utm_medium=referral&amp;utm_source=firstrun-login&amp;utm_content=fx-{{ version }}&amp;entrypoint=firstrun&amp;service=sync&amp;context=iframe&amp;style=chromeless&amp;haltAfterSignIn=true"></iframe>
+{% endblock %}
--- a/bedrock/settings/static_media.py
+++ b/bedrock/settings/static_media.py
@@ -1267,6 +1267,13 @@ PIPELINE_JS = {
         ),
         'output_filename': 'js/firefox_firstrun_quantum-bundle.js',
     },
+    'experiment_firstrun_email_first': {
+        'source_filenames': (
+            'js/base/mozilla-traffic-cop.js',
+            'js/firefox/firstrun/experiment-firstrun-email-first.js',
+        ),
+        'output_filename': 'js/firefox_firstrun_quantum-email_first_experiment-bundle.js',
+    },
     'firefox_firstrun_cliqz_funnelcake': {
         'source_filenames': (
             'js/firefox/firstrun/firstrun-cliqz-funnelcake.js',
--- a/media/css/firefox/firstrun/firstrun_quantum.less
+++ b/media/css/firefox/firstrun/firstrun_quantum.less
@@ -135,6 +135,13 @@ body {
     border: none;
     position: relative;
     width: 318px;
+
+    // Bug 1405822: Ensures the iframe + skip button
+    // is never smaller than the text on the left hand side.
+    min-height: 321px;
+    &.treatment {
+        min-height: 220px;
+    }
 }
 
 #fxa-iframe-config {
--- a/media/js/firefox/firstrun/experiment-firstrun-email-first.js
+++ b/media/js/firefox/firstrun/experiment-firstrun-email-first.js
@@ -1,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    /**
+     * An experiment to show FxA's "email-first" flow where
+     * the user first enters their email and then /signin
+     * or /signup is displayed depending on whether the
+     * account exists.
+     *
+     * 5% go to each bucket, 10% total.
+     */
+    var cop = new Mozilla.TrafficCop({
+        id: 'experiment_firstrun_email_first',
+        variations: {
+            'v=a': 5,  // control
+            'v=b': 5   // treatment - email first
+        }
+    });
+
+    cop.init();
+})(window.Mozilla);
--- a/media/js/firefox/firstrun/firstrun_quantum.js
+++ b/media/js/firefox/firstrun/firstrun_quantum.js
@@ -18,6 +18,7 @@
         scene.dataset.animate = 'true';
         var hideOrShowSkipButton = function (data) {
             switch(data.data.url) {
+            case '':
             case 'signin':
             case 'signup':
             case 'reset_password':
--- a/bedrock/firefox/templates/firefox/firstrun/email-first-experiment/index-b.html
+++ b/bedrock/firefox/templates/firefox/firstrun/email-first-experiment/index-b.html
@@ -1,0 +1,13 @@
+{% extends "firefox/firstrun/firstrun_quantum.html" %}
+
+{% block fxa_title %}
+  {{ _('Take Firefox with You') }}
+{% endblock %}
+
+{% block fxa_content %}
+  {{ _('Get your bookmarks, history, passwords and other settings on all your devices.') }}
+{% endblock %}
+
+{% block iframe %}
+  <iframe class="treatment" id="fxa" scrolling="no" data-src="{{ settings.FXA_IFRAME_SRC }}?action=email&amp;utm_campaign=fxa-embedded-form&amp;utm_medium=referral&amp;utm_source=firstrun-emailfirst&amp;utm_content=fx-{{ version }}&amp;entrypoint=firstrun&amp;service=sync&amp;context=iframe&amp;style=chromeless&amp;haltAfterSignIn=true"></iframe>
+{% endblock %}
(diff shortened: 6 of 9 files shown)


alexgibson:
> Are a and b typical names for a variation? I see a combination of a, b, and 1, 2, 3, etc.

We don't have a strict rule here. Personally I use `a`, `b` here etc.

> When an A/B test changes copy, do you use a single template for both variants or create separate templates? I went with a single template and if statements to control which copy is displayed.

Doing context switching logic in a single template works fine, but can sometimes get a bit messy / confusing to read. We usually go for split templates, mainly because they are super easy to read / follow - and also real easy to clean up again once an experiment has finished.

> The way I set the template context seems like it should be extracted, but I'm not sure how you usually handle that.

You could probably avoid altering the context in the view if you went for split templates for each variation.

> Since the A/B test is limited to en-US, do I need to surround the treatment group text with _('')

No not for an english only experiment. The only reason why it might be a good idea to do it now, would be if the changes we're likely to become the new default. But hardcoded strings should be fine for now.

> If the new strings should be translated now, is there a big string table or are they extracted from the templates?

We have an extrnal repo where strings are extracted to for translation. I don't think we need to worry about this just yet.

> I have a content-server fix that reduces the large gap between the FxA submit button and the "submit" button. This shrinks the iframe quite a bit and makes the box appear to be too high. Would it be alright to use flexbox to vertically center the iframe?

Given Firefox has supported flexbox for some time, I think this should be fine :)

alexgibson:
To handle a split template, you could probably do something like this in the view:

```python
elif show_57_firstrun(version):
    if locale == 'en-US' and exp in ['a', 'b']:
        template = 'firefox/firstrun/email-experiment/index-{0}.html'.format(exp)
    else:
        template = 'firefox/firstrun/firstrun_quantum.html'
```

This is then also pretty easy to write a test for, as you just need to verify the right template is shown for the appropriate URLs

media/css/firefox/firstrun/firstrun_quantum.less
```diff
@@ -135,6 +135,11 @@ body {
     border: none;
     position: relative;
     width: 318px;
+    &.treatment {
```
shane-tomlinson: I didn't want to affect styling of the original page, so I limited this change to the treatment group.

media/js/firefox/firstrun/firstrun_quantum.js
```diff
@@ -18,6 +18,7 @@
         scene.dataset.animate = 'true';
         var hideOrShowSkipButton = function (data) {
             switch(data.data.url) {
+            case '':
```
shane-tomlinson: FxA's "enter email" page is the index (/) and has a url of `''` (a blank string).

bedrock/firefox/templates/firefox/firstrun/email-first-experiment/index-a.html
```diff
@@ -0,0 +1,107 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
```
shane-tomlinson: These two templates are copies of firefox_quantum.html with the necessary changes.

media/js/firefox/firstrun/experiment-firstrun-email-first.js
```diff
@@ -0,0 +1,25 @@
...
+
+    /**
+     * An experiment to show FxA's "email-first" flow where
+     * the user first enters their email and then /signin
+     * or /signup is displayed depending on whether the
+     * account exists.
+     *
+     * 5% go to each bucket, 10% total.
+     */
+    var cop = new Mozilla.TrafficCop({
```
shane-tomlinson: Do you test TrafficCop files?

shane-tomlinson:
Depends on a tiny content server patch to look good: https://github.com/mozilla/fxa-content-server/pull/5765

bedrock/firefox/templates/firefox/firstrun/email-first-experiment/index-a.html
alexgibson: Instead of copying the entire template for each variation, I suggest you just extend `firefox/firstrun/firstrun_quantum.html`.

```
{% extends "firefox/firstrun/firstrun_quantum.html" %}
```
In `firstrun_quantum.html` you can then put jinja blocks around each piece of HTML you wish to change in the test:

{% block iframe %}
  // original iframe html
{% endblock %}

In the variations you can then override the content using:

```
{% block iframe %}
  <iframe id="fxa" scrolling="no" data-src="{{ settings.FXA_IFRAME_SRC }}signin?utm_campaign=fxa-embedded-form&amp;utm_medium=referral&amp;utm_source=firstrun-login&amp;utm_content=fx-{{ version }}&amp;entrypoint=firstrun&amp;service=sync&amp;context=iframe&amp;style=chromeless&amp;haltAfterSignIn=true"></iframe>
{% endblock %}

File: media/css/firefox/firstrun/firstrun_quantum.less (hunk 1 above)
Comment (alexgibson) [reply media/css/firefox/firstrun/firstrun_quantum.less]: Makes sense! I suggest here you add a comment with a reference to the bug number, so our future selves can work out why this is here :)

File: media/js/firefox/firstrun/experiment-firstrun-email-first.js (hunk 4 above)
Comment (alexgibson) [reply media/js/firefox/firstrun/experiment-firstrun-email-first.js]: The traffic cop library is already pretty heavily unit tested, so I think this is fine as-is.

File: bedrock/settings/static_media.py [hunk 5]
```diff
@@ -1277,6 +1277,8 @@
     },
     'firefox_firstrun_quantum': {
         'source_filenames': (
+            'js/base/mozilla-traffic-cop.js',
+            'js/firefox/firstrun/experiment-firstrun-email-first.js',
```
alexgibson: These tow files should be in their own bundle, and then included in the `{% block experiments %}` block so that they excecute in the `<head>` of the page for a faster redirect.

alexgibson: We should also wrap the experiment JS in a switch, so we can toggle the experiment on/off independently of a deployment. e.g.

```
{% block experiments %}
  {% if switch('experiment-firstrun-email', ['en-US']) %}
    {% javascript 'experiment-firstrun-email' %}
  {% endif %}
{% endblock %}
```

Note: the `en-US` locale bit means the bundle will only be included for that locale.

media/css/firefox/firstrun/firstrun_quantum.less
shane-tomlinson: Done

bedrock/settings/static_media.py
shane-tomlinson: Done

bedrock/firefox/templates/firefox/firstrun/email-first-experiment/index-a.html
shane-tomlinson: Done

shane-tomlinson:
@alexgibson - if the latest version looks good to you, could you toss this on a demo server and I'll have @davismtl and @ryanfeeley have a look?

alexgibson:
This is now up on a demo:

https://www-demo1.allizom.org/en-US/firefox/57.0/firstrun/?v=a
https://www-demo1.allizom.org/en-US/firefox/57.0/firstrun/?v=b

alexgibson:
@shane-tomlinson to config demo1 to point to your auth server, you can create a file in your PR here called: `jenkins/branches/demo/1.env`

In that file set the following values:

```
FXA_IFRAME_SRC=https://stomlinson.dev.lcip.org/ CSP_EXTRA_FRAME_SRC=https://stomlinson.dev.lcip.org/
```

Once that is done, I can deploy the branch to demo and fingers crossed it should work

shane-tomlinson:
@alexgibson - pushed with that change!

jenkins/branches/demo/1.env
```diff
@@ -0,0 +1,2 @@
+FXA_IFRAME_SRC="https://stomlinson.dev.lcip.org/"
```
alexgibson: I think you need to omit the double quotes here: `FXA_IFRAME_SRC=https://stomlinson.dev.lcip.org/`

shane-tomlinson: Removed. I had the double quotes in my own .env file when testing against stomlinson.dev.lcip.org, so kept them. Wasn't sure if that was necessary or not.

alexgibson:
@shane-tomlinson sorry for the delay, it looks like your demo is now configured and working :)

shane-tomlinson:
@alexgibson - thanks for putting this up on demo1, @ryanfeeley tested this yesterday in a big group setting and gave the 👍. We haven't yet released our small styling fix to prod, that should happen today. When you have a moment, can you do a final review?

Once both sides are approved, what's the timeframe needed to get this in front of users?

Thanks again!

jenkins/branches/demo/1.env
```diff
@@ -0,0 +1,2 @@
+FXA_IFRAME_SRC=https://stomlinson.dev.lcip.org/
```
alexgibson: Please remove this file before we merge

PR #5318, opened 2017-11-29
[bug 1343823] Remove old /whatsnew templates

alexgibson:
## Description
- Removes a boat load of old `/whatsnew` pages.
- The `/whatsnew` template for firefox 54 (mobile promo) becomes the new default for all fallback URLs.
- Renames `firefox/whatsnew_50.lang` to `firefox/whatsnew.lang`.
- Adds functional tests for default `/whatsnew` page.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1343823

## Testing
Demo: https://www-demo3.allizom.org/firefox/whatsnew/

- Check default `/whatsnew` and check for any regressions?
- Make sure I didn't miss removing any supreflous assets, or did remove anything that's still needed?


flodolo:
Do you plan to remove also old /firstrun? I'm wondering about that entire australis folder

alexgibson:
> Do you plan to remove also old /firstrun? I'm wondering about that entire australis folder

Not in this PR. There are still some unanswered questions I need to find out, relating to which versions of Firefox supports the iframe. We may need to do something special there for a fallback.

bedrock/firefox/templates/firefox/whatsnew/index.html
```diff
@@ -21,7 +21,7 @@
 {% block site_header %}{% endblock %}
 
 {% block page_css %}
-  {% stylesheet 'firefox_whatsnew_54' %}
+  {% stylesheet 'firefox_whatsnew' %}
```
alexgibson: @flodolo is there anything you'd like to do here to tidy up lang files? The file here contains the following above:

```
{% add_lang_files "firefox/whatsnew_50" "firefox/sendto" %}
```

flodolo: Yes, I was just noticing that. I would rename it to firefox/whatsnew for making it less confusing, then we can get rid of firefox/whatsnew50

alexgibson:
Successful test run:

https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/133/pipeline

alexgibson:
This should be good for review. Will wait to get the thumbs up from @flodolo with regard to l10n before we merge.

flodolo:
It looks correct to me.

We still need to keep `firefox/australis/firefox_tour.lang` (used in remaining /australis files).
Rename `firefox/whatsnew_50.lang` to `firefox/whatsnew.lang`.
Remove `firefox/whatsnew_42.lang`.

Technically, whatsnew_50 is a fallback, but it shouldn't be actively used. Is that correct? The only active WNP should be 56 at this point. If that's the case, we're not blocking and I can rename the file any time.

alexgibson:
> Technically, whatsnew_50 is a fallback, but it shouldn't be actively used. Is that correct? The only active WNP should be 56 at this point. If that's the case, we're not blocking and I can rename the file any time.

Technically, any `/whatsnew URL is still active at any point. But what we're doing here is making the whatsnew template for 54 the default for every version < 54. 

So users who currently open:

https://www.mozilla.org/en-US/firefox/50.0/whatsnew/
https://www.mozilla.org/en-US/firefox/42.0/whatsnew/
https://www.mozilla.org/en-US/firefox/38.0/whatsnew/

Will now see:

https://www-demo3.allizom.org/en-US/firefox/50.0/whatsnew/
https://www-demo3.allizom.org/en-US/firefox/42.0/whatsnew/
https://www-demo3.allizom.org/en-US/firefox/38.0/whatsnew/

Does that make sense?

bedrock/firefox/templates/firefox/whatsnew/index.html
```diff
@@ -21,16 +21,13 @@
...
 {% endblock %}
 
 {% set show_send_to_device = LANG in settings.SEND_TO_DEVICE_LOCALES %}
 
 {% set send_to_device_message_set = 'fx-50-whatsnew' %}
 
-{% set app_store_url = 'https://app.adjust.com/2uo1qc?campaign=whats_new&adgroup=iOS_App_Store&creative=badge_en&fallback=https%3A%2F%2Fitunes.apple.com%2Fapp%2Fapple-store%2Fid989804926%3Fpt%3D373246%26ct%3Dadjust_tracker%26mt%3D8' %}
```
alexgibson: These don't seem to be used any longer?

flodolo:
> Does that make sense?

Yes. As said, l10n is not a blocker here. I'll copy whatsnew_50 as whatsnew, then we can remove 50 and 42 files when this is merged.

alexgibson:
Great, thanks @flodolo

flodolo:
I'm removing the l10n tag: file is already available on both trunk and prod, and other files marked as obsolete internally.

alexgibson:
@jpetto thanks for the review, updated and squashed.

PR #5319, opened 2017-11-29
Remove OpenHatch

vmbrasseur:
## Description
OpenHatch has [closed its doors](https://blog.openhatch.org/2017/celebrating-our-successes-and-winding-down-as-an-organization/) and is no longer providing assistance to new contributors. Removing it from the page lest we lead new contribs down a blind alley.

## Issue / Bugzilla link
None. Just something I spotted while reading the Open Innovation page.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/internet-health/open-innovation.html
+++ b/bedrock/mozorg/templates/mozorg/internet-health/open-innovation.html
@@ -126,10 +126,15 @@
             <li>
               <h4>{{ _('Contribute to a project') }}</h4>
               <p>
-              {% trans contribute=url('mozorg.contribute.index'),
-              %}
-                Contribute to an existing open source project. Become a <a href="{{ contribute }}">Mozillian!</a>
-              {% endtrans %}
+              {% if l10n_has_tag('open-innovation-update-20171130') %}
+                  {% trans contribute=url('mozorg.contribute.index') %}
+                     Contribute to an existing open source project. Become a <a href="{{ contribute }}">Mozillian!</a>
+                  {% endtrans %}
+              {% else %}
+                  {% trans contribute=url('mozorg.contribute.index'),
+                           open_hatch='https://openhatch.org/' %}
+                     Contribute to an existing open source project. Become a <a href="{{ contribute }}">Mozillian!</a> Or use <a href="{{ open_hatch }}">OpenHatch</a> to match your skillset to a project.
+              {% endif %}
               </p>
             </li>
             <li>
--- a/bedrock/mozorg/templates/mozorg/internet-health/open-innovation.html
+++ b/bedrock/mozorg/templates/mozorg/internet-health/open-innovation.html
@@ -134,6 +134,7 @@
                   {% trans contribute=url('mozorg.contribute.index'),
                            open_hatch='https://openhatch.org/' %}
                      Contribute to an existing open source project. Become a <a href="{{ contribute }}">Mozillian!</a> Or use <a href="{{ open_hatch }}">OpenHatch</a> to match your skillset to a project.
+                  {% endtrans %}
               {% endif %}
               </p>
             </li>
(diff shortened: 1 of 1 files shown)


bedrock/mozorg/templates/mozorg/internet-health/open-innovation.html
```diff
@@ -127,10 +127,8 @@ <h3 class="topic-action-title">{{ _('How you can support open source') }}</h3>
               <h4>{{ _('Contribute to a project') }}</h4>
               <p>
               {% trans contribute=url('mozorg.contribute.index'),
-                       open_hatch='https://openhatch.org/'
```
jpetto: We can get rid of the comma on the preceding line, and bring up the closing brackets to that same line.

bedrock/mozorg/templates/mozorg/internet-health/open-innovation.html
```diff
@@ -127,10 +127,8 @@ <h3 class="topic-action-title">{{ _('How you can support open source') }}</h3>
...
               %}
                 Contribute to an existing open source project. Become a <a href="{{ contribute }}">Mozillian!</a>
-                Or use <a href="{{ open_hatch }}">OpenHatch</a> to match your skillset to a project.
```
jpetto: As this is changing a localized string, we're going to need to put this change behind an `l10n_has_tag` conditional (to give localizers a chance to update the translations). See [here](https://github.com/mozilla/bedrock/blob/40677449e280a7c171be006b0242cc2400d4684d/bedrock/firefox/templates/firefox/developer-quantum-whatsnew.html#L75) for an example.

Basically, we need to put the updated `{% trans %}` block (with the OpenHatch URL removed) in the `if` block, and the current `{% trans %}` block in the `else` block.

I recommend using `open-innovation-update-20171130` as the tag in the `l10n_has_tag` call.

Let me know if you have any questions or if I can help!

vmbrasseur: Thanks, @jpetto. I'll try to get to this tomorrow.

flodolo: Passing by note: underscore is preferred to hyphen in tags ;-)

It can also be as simple as update_20171130, since its scope is the file itself (and template using it).

vmbrasseur: The underscore is in a line which is being removed, so I won't fret about that. It may be worth adding this tip to a styleguide, though.

vmbrasseur:
New code pushed.

Tonchi:
Energize WarpSpeed@IBM.Comp.186,000Miles/Second/GoGoUGoAGoGoIngGoingGone/

Tonchi:
RunAndSave@FullSpeedingWhatGoesInIsGoodKeepRunningToTtillAllGone/Program.FragmentsSaveAsRunRAMinROM.comi

flodolo:
@Tonchi 
No clue what you're doing, but please stop ;-)

bedrock/mozorg/templates/mozorg/internet-health/open-innovation.html
```diff
@@ -126,12 +126,15 @@ <h3 class="topic-action-title">{{ _('How you can support open source') }}</h3>
...
-              {% endtrans %}
+              {% if l10n_has_tag('open-innovation-update-20171130') %}
+                  {% trans contribute=url('mozorg.contribute.index') %}
+                     Contribute to an existing open source project. Become a <a href="{{ contribute }}">Mozillian!</a>
+                  {% endtrans %}
+              {% else %}
+                  {% trans contribute=url('mozorg.contribute.index'),
+                           open_hatch='https://openhatch.org/' %}
+                     Contribute to an existing open source project. Become a <a href="{{ contribute }}">Mozillian!</a> Or use <a href="{{ open_hatch }}">OpenHatch</a> to match your skillset to a project.
+              {% endif %}
```
jpetto: Missing the closing `{% endtrans %}` before this `{% endif %}`. After that fix, I think we'll be good!

vmbrasseur: Oops! You're totally right, of course. Thank you!

vmbrasseur: All fixed, I believe.

alexgibson:
Hmm, I don't seem to be able to merge this without @jpetto blessing also :)

PR #5323, opened 2017-12-01
fix(copy):Change passwords to logins

MeridelW:
This is my second commit. Please replace "passwords" with "logins" to be consistent with product.


craigcook:
I see references to both "logins" and "passwords" in Firefox (in usage they refer to slightly different things; a "login" is a whole set of credentials, including a password). I think "password" is a simpler and more commonly understood word for this line of copy, and is of course still technically accurate. Passwords are indeed synced along with other information.

This page is also localized in dozens of languages so this change would need to be marked as a new string behind an l10n tag, preserving the old string as a fallback until all locales can update.

For these reasons I'm going to close this pull request, but we may revisit the copy on this page in the near future. Thanks for your contribution!

PR #5324, opened 2017-12-01
[fix bug 1287200] Update jQuery to latest release.

jpetto:
## Description

- Updates jQuery to latest release
- Adds new `common-ie8` bundle and conditional logic for IE 8 and below
- Updates Waypoints lib to the latest release (for compatibility with latest jQuery)
- Removes unused JS files

IE 8 specific bundle initializes download buttons and enables GTM and stub attribution. The following JS blocks/files will not be served to IE <= 8:

- `experiments` block
- `js` block (this is the big one)
- `update-notification` block
- `pontoon.js` script

jQuery migrate revealed some places where our jQuery is now using deprecated methods/techniques. I'd fix these in a follow-up PR so as to limit the scope here as much as possible.

### Shared files between `common` and `common-ie8`

- `js/base/mozilla-client.js`
- `js/base/core-datalayer.js`
- `js/base/core-datalayer-init.js`
- `js/base/search-params.js`

These files are needed for GTM and/or stub attribution. For now, there's no cost here, but we will need to remain aware of the browser baseline when updating any of the shared files so as to not introduce errors in IE <= 8.

Ideally, we'll be able to remove GTM and stub attribution code from IE <= 8, but it's almost 2018 and 10% of downloads are from IE <= 8, so this may never manifest. This means that we will need to keep an open mind to forking these shared files into modern browser/IE <= 8 versions.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1287200

## Testing

https://www-demo5.allizom.org/

SSL issues with named demos right now, so also available on my demo below (however, FxA isn't testable there):

https://bedrock-demo-jpetto.us-west.moz.works/

[Functional tests are passing](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/160/pipeline/).

#### All browsers
- Check download buttons on `/`, `/firefox/`, and `/firefox/new/`
- Verify GTM is functioning (pg verified on Slack on 2018/01/03)
- Verify stub attribution is working (not sure who we need here)

#### IE > 8, modern browsers
- Check JS on a few pages using Waypoints:
  - `/firefox/` (sub-nav and in-page nav)
  - `/teach/smarton/tracking/` (sub-nav)
- Check newsletter form functionality (any URL should do)
- Check `/contact` pages
- Check FxA iframe on `/firstrun`


jpetto:
@alexgibson & @craigcook I think this is now ready to be reviewed. Would be great if both of you could review, as there's a pretty wide impact.

media/js/base/base-page-init.js
```diff
@@ -5,22 +5,47 @@
...
+        }
+
+        // note that document.ready happened to inform window.load
+        ready = true;
+    });
 
     $(window).on('load', function () {
-        $('html').addClass('loaded');
+        // if document.ready happened already, fire onWindowLoad
+        if (ready) {
```
alexgibson: This is changing the behavior of the loaded class that we rely on for initiating functional tests. I'm not sure we want to trigger it on DOM ready this way, since assets such as images may not have loaded yet, or JS may not have fully excecuted. The test run looks green even with this change which is good, but it may be more prone to intermittent failures this way due to slow network conditions?

alexgibson:
Have we checked how this might impact the legacy `tabzilla.js`?

media/js/base/base-page-init.js
jpetto: My intent was to *not* change the behavior. The current implementation waits for both DOM ready and window load, which is what I'm trying to replicate here by making sure both `ready` and `loaded` are `true` prior to adding the `loaded` class. Maybe I'm missing something obvious though?

alexgibson: I'm not sure if this is perhaps over-complicating it a bit, but maybe I'm missing something? The previous behavior added a window `load` event handler inside a DOM ready handler. Given that DOM ready always before load, all that really matters here is that the loaded class gets added on the `load` event, right?

jpetto: I'm not sure ready will always fire before load in 3.0. From the [GitHub thread linked yesterday](https://github.com/jquery/jquery/issues/3194#issuecomment-228556922):

> We recently made ready handlers fire asynchronously. This has advantages that are hard to give up. The disadvantage is that the ready handler can sometimes fire after the load event if the load event fires quickly enough.

This is probably a pretty edge-y case, but if guy from jQuery core team says it can happen, I don't mind being on the safe side.

alexgibson: Ah I see,

To make this a bit easier to read/follow, would the [pattern here](https://github.com/jquery/jquery/issues/3194#issuecomment-228562930) be worth considering? It seems a bit clearer and easier to follow without needing the `load` even handler nested inside a ready handler?

alexgibson: Actually, ignore my last comment. We wouldn't want to wait for both events to be satisfied before excecuting everything, right? So what you have is likely fine (and tests pass) 👍

jpetto:
> I haven't dug into this past looking at the code yet, but the other thing that springs to mind is the legacy tabzilla.js. Might it be impacted by this change?

Hm, good question. Will figure out how to test this.

alexgibson:
> Hm, good question. Will figure out how to test this.

I did notice that the minimum required version of jQuery in Tabzilla seems to be [1.7.1](https://github.com/mozilla/bedrock/blob/master/bedrock/tabzilla/templates/tabzilla/tabzilla.js#L75), which is a file also being removed in this PR. I wonder if we should fix [bug 1388296](https://bugzilla.mozilla.org/show_bug.cgi?id=1388296) and neuter tabzilla.js as a precursor to this change? We could handle it in a separate PR, thoughts?

bedrock/settings/static_media.py
```diff
@@ -1088,6 +1088,14 @@
         ),
         'output_filename': 'js/common-bundle.js',
     },
+    'common-ie8': {
+        'source_filenames': (
+            'js/libs/jquery-1.11.3.min.js',
+            'js/ie8/mozilla-utils.js',
```
alexgibson: One thing maybe missing here is GA. Do we need to make sure both `js/base/core-datalayer.js`, and `core-datalayer-init.js` are IE8 compatible?

bedrock/base/templates/base-pebbles.html
```diff
@@ -154,17 +154,27 @@ <h2 class="masthead-logo"><a href="{{ url('mozorg.home') }}"><img src="{{ static
...
+        {% javascript 'common' %}
+      <!--<![endif]-->
+
+      <!--[if lte IE 8]>
+        {% javascript 'common-ie8' %}
+      <![endif]-->
     {% endblock %}
 
     {# Bug 1279291 #}
+    <!--[if !lte IE 8]><!-->
```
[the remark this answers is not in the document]
alexgibson: One other thought - should we exclude old IE from the `experiments` blocks? It seems like these users would be poor data for muddying experiment resultsd anyway, given many pages get next to no CSS styling.

media/js/ie8/mozilla-utils.js
```diff
@@ -0,0 +1,45 @@
...
+
+    var Utils = {};
+
+    /**
+     * Bug 393263 A special function for IE < 9.
+     * Without this hack there is no prompt to download after they click. sigh.
+     * @param {link} direct link to download URL
+     * @param {userAgent} optional UA string for testing purposes.
+     */
+    Utils.triggerIEDownload = function(link, userAgent) {
```
alexgibson: Should we remove these functions from the regular `mozilla-utils.js` and then run the unit tests against this file?

alexgibson:
> Note that I've excluded stub attribution JS from IE 8 and below. I'm on the fence on this decision. Keeping stub attribution around at this point would be pretty easy - just adding search-params.js to the IE 8 bundle. However, this may prove challenging in the future when updating jQuery and/or the stub attribution JS. May be best to get analytics input here.

I'm not at all opposed to limiting stub attribution to IE9 and above. I guess we should probably ask someone how important data this is though.

alexgibson:
Should we close #4222 in favor of this PR?

bedrock/base/templates/base-pebbles.html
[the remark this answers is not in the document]
jpetto: These are all great suggestions.

bedrock/settings/static_media.py
```diff
@@ -1071,7 +1071,7 @@
     # bundle should offset the extra weight.
     'common': {
         'source_filenames': (
-            'js/libs/jquery-1.11.3.min.js',
+            'js/libs/jquery-3.2.1.min.js',
```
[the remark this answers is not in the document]
jpetto: Hm, I don't see the `newsletter_form` bundle used anywhere. Deleting!

`tracking-protection-tour` bundle seems fine w/`3.2.1`.

media/js/base/base-page-init.js
```diff
@@ -5,22 +5,52 @@
...
+    // https://github.com/jquery/jquery/issues/3194#issuecomment-228556922
+    var loaded = false;
+    var ready = false;
 
-    /* Bug 1264843: In partner distribution of desktop Firefox, switch the
-       downloads to corresponding partner build of Firefox for Android. */
-    if (client.isFirefoxDesktop) {
-        client.getFirefoxDetails(utils.maybeSwitchToDistDownloadLinks);
+    function onWindowLoad() {
+        $('html').addClass('loaded');
```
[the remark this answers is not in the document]
jpetto: Good suggestion.

media/js/ie8/mozilla-utils.js
jpetto: Good call. Will do.

jpetto:
@alexgibson Though the demo server is having SSL issues, this should be again ready for review.

media/js/base/mozilla-utils.js
```diff
@@ -12,31 +12,8 @@ if (typeof Mozilla === 'undefined') {
...
-    // ie functionality if on ie
+    // a11y assistance
     Utils.initDownloadLinks = function() {
-        $('.download-link').each(function() {
-            var $el = $(this);
-            $el.click(function() {
-                Utils.triggerIEDownload($el.data('direct-link'));
-            });
-        });
         $('.download-list').attr('role', 'presentation');
```
alexgibson: I know a11y is important, but it looks like this was added a long time ago. Is this `role` still useful enough on its own to keep this function around?

jpetto: It's a good question that I didn't really spend much time thinking about. (I was more focused on keeping scope for this PR as limited as possible.)

**I am not an a11y expert!**

That being said, I believe the intent here is to keep screen readers from treating the `<ul>` as an actual list to be iterated over, as the visitor *should* only care about the one `<li>` displayed. However, all other `<li>`s are set to `display: none;`, which (I believe) should prevent them from being recognized by screen readers.

Unless you disagree with the above assessment, I'll remove this function. Let me know!

alexgibson: Remove away 👍

jpetto:
> One minor nit aside, code-wise this lgtm. I'd like to give it some more thorough testing once the demo server is accessible again however, especially on older IE. I can also help test whether stub attribution is still working ok.

I updated my demo for getting around the SSL issues (link added to the description). We can't test FxA as-is, but I did run a test last week before the SSL issues and it was working as expected.

> 
> As for GA, has Peter managed to confirm that everything looks good on IE <= 8?

Yep, Peter confirmed GA looked good on IE 8 yesterday.

> 
> Last question is, how confident are we that this won't impact tabzilla.js?

Pretty confident. If jQuery isn't available on load, `tabzilla.js` grabs `jquery-1.7.1-min.js`...which I just realized is loaded from `mozorg.cdn.mozilla.net/media/js/libs`, meaning there's no reason to keep `jquery-1.7.1.min.js` around in bedrock. Will remove!

alexgibson:
When testing in IE 8 and below on `/firefox/new/`, downloads appear to work fine as expected. One thing I do notice though is that the "click here" link on scene2 now defaults to `/firefox/all/`, since the scene2 JS no longer executes. This still works, as in it's possible for users to download Firefox still, so maybe it's ok? Alternatively, we could special case the download page JS so that "click here" link is still populated with a direct download link. Thoughts?

alexgibson:
In order to test that stub attribution still works, I believe we need to set an environment variable on the demo with a dummy key value, e.g. `STUB_ATTRIBUTION_HMAC_KEY='foo'` (it can be any value for testing the bedrock side of things)

alexgibson:
When visiting Sandstone pages on IE8 and below, the newsletter forms no longer expand, so the forms can no longer submit successfully: https://bedrock-demo-jpetto.us-west.moz.works/en-US/about/

Not sure how concerned we are about these signups (or how many there are from these visitors?), but could it be something that we can fix in the oldIE stylesheet perhaps?

alexgibson:
The language drop down in the site footer also no longer works in old IE. I believe there's a hidden submit button that makes it with without JS. Could we perhaps force that button to show for oldIE too, to maintain functionality?

alexgibson:
Aside from these things, I think this is in pretty good shape. We don't necessarily need to address them all if we decide it's not worth it, but some might be fixable with a few specific CSS rules?

Nice work @jpetto 👍

jpetto:
@alexgibson I've updated the lang switcher to have the button visible for IE <= 8, as well as adjusted the newsletters on Sandstone pages to be expanded (for the same IE <= 8 audience).

I think I'm fine leaving the scene 2 link pointing to `/firefox/all/` for now. I think that's pretty edge case-y, and has a perfectly workable fallback. Unless we get complaints, I'm happy not writing (more) IE 8 specific JS.

Both demos (mine and demo 5 (which again has working SSL)) have been updated with a stub attribution env value.

Thanks!

alexgibson:
@jpetto updates look good to me. Squash time?

PR #5327, opened 2017-12-05
Git repos avoid using remotes

pmac:
Strange state when initial setup of git repos happened under a different config than the running container. This should fix this by using repo URLs directly.

### DIFF
--- a/bedrock/utils/git.py
+++ b/bedrock/utils/git.py
@@ -23,17 +23,13 @@ GIT = getattr(settings, 'GIT_BIN', 'git')
 
 
 class GitRepo(object):
-    def __init__(self, path, remote_url=None, remote_name=None, branch_name='master'):
+    def __init__(self, path, remote_url=None, branch_name='master'):
         self.path = Path(path)
         self.path_str = str(self.path)
         self.remote_url = remote_url
         self.branch_name = branch_name
-        if not remote_name:
-            remote_name = 'bedrock-dev' if settings.DEV else 'bedrock-prod'
-
-        self.remote_name = remote_name
-        db_latest_key = '%s:%s:%s:%s' % (self.path_str, remote_url or '',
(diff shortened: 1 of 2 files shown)


jpetto:
@pmac Could you provide a bit more context on the issue this is solving?

pmac:
> @pmac Could you provide a bit more context on the issue this is solving?

Certainly,

I noticed this problem when starting to include the external data git-repos in the built docker images that we ship. The issue is that we have settings for the git branch that a deployment should pull from. When the repo is already setup and in the image this can be an issue because the old way doesn't ensure that the right branch is checked out if the repo is already there and the proper remote exists. It assumes that all is well in that case and just does a `git pull`. So you might change the setting on a deployment to a testing branch to see how some different data reacts, or even a different repo (your fork e.g.) and it wouldn't work because the code would think all was okay. Before we included the repos in the docker images it was working fine because it cloned the repo every time the command was run for the first time. Changing a setting in Deis restarts all of the containers which means they start fresh and re-clone the repos. It's more efficient if we include the repos in the images, but it did introduce this issue. It doesn't come up too often, but it can be an issue if we need to switch things up from the values with which the image was built.

I hope that was clear. Let me know if something is still confusing.

jpetto:
Ah, okay. So the fix is to avoid the use of remote names to ensure that a pre-configured remote (with a potentially wrong branch) isn't used. Is that accurate?

pmac:
> Ah, okay. So the fix is to avoid the use of remote names to ensure that a pre-configured remote (with a potentially wrong branch) isn't used. Is that accurate?

Potentially wrong branch or even repo URL, yes. That's exactly it. Also this just seems like the more correct way to go about this. Git remotes are just conveniences for CLI use and shouldn't been needed or used in git scripting IMHO.

PR #5328, opened 2017-12-05
[bug 1422899] Remove old plugincheck template

alexgibson:
## Description
- Removes old `/plugincheck/` page template and assets.

## Issue / Bugzilla link

## Testing
- Check I didn't miss anything or remove something that's still needed?
- Check that translations still work ok?


bedrock/settings/static_media.py
```diff
@@ -785,17 +785,9 @@
...
-        ),
-        'output_filename': 'css/plugincheck-bundle.css',
-    },
-    'plugincheck-update': {
-        'source_filenames': (
-            'css/plugincheck/plugincheck-update.scss',
+            'css/plugincheck/plugincheck.scss',
         ),
-        'output_filename': 'css/plugincheck-update.css',
+        'output_filename': 'css/plugincheck.css',
```
jpetto: Should be `css/plugincheck-bundle.css`?

alexgibson:
Thanks @jpetto, updated.

>Not specifically related to this PR (this happens in prod for me right now), but it seems to be impossible to get the message-firefox-current content. Can file a follow-up bug to avoid exploding scope here.

It looks like you need to be using an up to date Firefox on release channel to see that message. Not quite sure why pre-releases aren't in that same bucket. There seems to be a bug about that here: https://bugzilla.mozilla.org/show_bug.cgi?id=1393035

flodolo:
For reference, we removed the lang file in 
https://github.com/mozilla-l10n/langchecker/pull/699

PR #5337, opened 2017-12-06
[bug 1422899] Remove old home page template

alexgibson:
## Description
- Removes old home page template

## Issue / Bugzilla link

## Testing
- Probably a good idea to run `gulp static:clean` prior to testing this PR.
- Check I didn't remove any other assets that may still be needed?
- Check translations still appear as expected.


craigcook:
> Noting that merging this will mean I have to make some updates to #5356 

Actually that was a painless rebase, I expected merge conflicts but all my edits were in the include not in the home template. Huzzah!

flodolo:
> There are only a handful of locales that haven't fully updated and still get the old home page. Are we OK with them falling back to English? Perhaps this will help encourage them to update?

At this point, yes.

PR #5340, opened 2017-12-06
[fix bug 1423237] Add net neutrality protest banner

craigcook:
## Description
Adds banners to the home page and /firefox/new/ page (just the default, none of the special variants). The 'close' button hides the banner and sets a cookie, just recycling the same code as the fundraising banner.

It's behind the switch `netneutrality-12-2017` and we'll activate it early on Dec 12 and take it down a few days later (the vote is scheduled for Dec 14).

This is explicitly en-US only, no l10n.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1423237

## Testing
Test responsiveness, verify the close cookie works, check for any dumb mistakes I might have made.


bedrock/firefox/templates/firefox/new/quantum/base.html
```diff
@@ -40,6 +40,10 @@
 {% block site_header %}{% endblock %}
 
 {% block content %}
+{% if LANG == 'en-US' and switch('netneutrality-12-2017') %}
```
jpetto: Can do the lang check in the switch call (here and in other templates):

`{% if switch('netneutrality-12-2017', ['en-US']) %}`

bedrock/mozorg/templates/mozorg/home/home-quantum.html
```diff
@@ -251,14 +253,16 @@ <h3>{{ _('Need help?') }}</h3>
...
+  {% elif switch('home-fundraiser-EOY2017') and l10n_has_tag('home-fundraiser-EOY2017') %}
     {% javascript 'home-fundraiser-eoy2017' %}
   {% endif %}
 {% endblock %}
 
 {% block update_notification %}
-  {# do not show the update banner if the fundraiser banner is active #}
-  {% if not switch('home-fundraiser-EOY2017') %}
+  {# do not show the update banner if the fundraiser banner (or net neutrality banner) is active #}
+  {% if not switch('home-fundraiser-EOY2017') or not switch('netneutrality-12-2017') %}
```
jpetto: I think this should be an `and` instead of an `or` - both switches need to be off for the update banner to be shown.

media/js/mozorg/home/netneutrality-12-2017.js
```diff
@@ -0,0 +1,33 @@
...
+
+            if (cookiesOK) {
+                d = new Date();
+                d.setTime(d.getTime() + (cookieDurationDays * 24 * 60 * 60 * 1000)); // 2 day expiration
+                Mozilla.Cookies.setItem(storageKey, true, d.toUTCString(), '/');
+            }
+        });
+    }
+
+})(window.jQuery, window.Mozilla);
```
jpetto: Can we drop one of these identical `.js` files and just use one for all bundles?

media/css/mozorg/home/netneutrality-12-2017.scss
```diff
@@ -0,0 +1,136 @@
...
+        #promo-headline {
+            float: right;
+
+            h1 {
+                padding-left: 0;
+                padding-right: 1em;
+            }
+        }
+    }
+}
```
jpetto: The CSS for these pages looks almost identical - could we use a single file for all bundles? I know it's a quick turnaround and page lifespan, so not a huge issue as-is.

jpetto:
screen shot 2017-12-07 at 1 03 19 pm

That button proximity makes me claustrophobic.

media/js/mozorg/home/netneutrality-12-2017.js
craigcook: Yeah there's a lot of duplication but I was trying to keep it totally self-contained for easy removal... but maybe that wasn't the best strategy. I'll work on combining/simplifying.

media/css/mozorg/home/netneutrality-12-2017.scss
```diff
@@ -0,0 +1,174 @@
...
+        padding-top: 4px;
+        width: 50%;
+
+        p {
+            margin-bottom: 2em;
+        }
+    }
+}
+
+html[dir="rtl"] {
```
jpetto: Probably don't need any RTL stuff?

craigcook: copypasta! 🍝

PR #5341, opened 2017-12-07
[bug 1422899] Remove old dev edition product page

alexgibson:
## Description
- Removes old firefox developer product page template
- Moves dev edition /firstrun and /whatsnew templates to the same directory, as they share much of the same code (and lang files).

Note: I have left the old dev edition /whatsnew and /firstrun templates, as I'm not sure it makes sense to show the quantum template to older versions.

## Issue / Bugzilla link

## Testing
- Probably a good idea to run gulp static:clean prior to testing this PR.
- Check I didn't remove any other assets that may still be needed?
- Check both the /firstrun and /whatsnew pages still look as expected.
- Check translations still appear as expected.


alexgibson:
Successful run for the integration test changes here: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/148/pipeline

media/js/firefox/dev-firstrun.js
```diff
@@ -93,252 +74,6 @@ function onYouTubeIframeAPIReady() {
 
     }
 
-    // make sure we strip any HTML tags before injecting door hanger text
-    function getText(string) {
-        return $('<div/>').html(utils.trans(string)).text();
-    }
-
-    // shows the Dev Tools doorhanger step
-    function showDevToolsDoorhanger() {
```
alexgibson: Note: I decided we should remove this old UITour functionality as much of it is now broken during the releases that led up to Quantum. We could keep it around, as it's served only to old version of dev edition - but this makes it difficult to test and maintain without downloading old versions of the browser.

craigcook: Agreed, this seems unnecessary at this point. If someone with an older version lands on firstrun all the basic content is still there.

PR #5344, opened 2017-12-08
[no bug] Update Fx logo on /about/history slides.

jpetto:
## Description

Noticed this old logo while working on something else. This page is very legacy, but easy enough to update the logo.

## Issue / Bugzilla link

Newp.

## Testing

Check for silly mistakes?

### DIFF
--- a/media/css/mozorg/history-slides.less
+++ b/media/css/mozorg/history-slides.less
@@ -507,8 +507,11 @@
     }
 
     .logo {
+        .at2x('/media/img/logos/firefox/logo-quantum.png', 132px, 132px);
         .transition(opacity .25s ease-in);
-        background: transparent url('/media/img/mozorg/history/fxlogo.png') no-repeat;
+        background-color: transparent;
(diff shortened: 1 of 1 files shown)


media/css/mozorg/history-slides.less
```diff
@@ -507,8 +507,10 @@
     }
 
     .logo {
+        .at2x('/media/img/logos/firefox/logo-quantum.png', 132px, 132px);
```
craigcook: It seems the Less mixin only outputs the background-size inside the high dpi media query, which means low-res screens aren't scaling the image. I think we just need to explicitly add background-size here as well.

image

jpetto: Ah, d'oh! Added (but can't concretely verify outside of dev tools, as I only have a high-dpi screen atm).

craigcook: That fixed it! And I vote we leave the Android one alone. We really need to get rid of this page anyway.

PR #5349, opened 2017-12-19
Avoid testing all download URLs more than once per deployment

pmac:
I'm not sure if this will help with our deployment woes or not, but it does seem inefficient at best to test every download URL on the site 6* per deployment. I'll explain:

The download tests find all download links on the various `*/all/` pages. That comes out to around 3,000 unique URLs, almost all of which point to bouncer (download.mozilla.org). That means that we're sending 3,000 requests to bouncer for each deployment, per app, per region. Since we deploy stage and prod to 3 regions during a prod deployment that comes out to 6 deployments and 18,000 requests. These are just `HEAD` requests, so the CDN isn't sending the download bytes, but it does seem wasteful to check those for every app in every region, especially since the checks are all coming from the Jenkins box.

### DIFF
--- a/docker/bin/run_integration_tests.sh
+++ b/docker/bin/run_integration_tests.sh
@@ -25,6 +25,10 @@ case $1 in
     PLATFORM="Windows 7"
     MARK_EXPRESSION="sanity and not headless"
     ;;
+  download)
+    DRIVER=
+    MARK_EXPRESSION=download
+    ;;
   headless)
     DRIVER=
     MARK_EXPRESSION=headless
--- a/jenkins/branches/master.yml
+++ b/jenkins/branches/master.yml
@@ -12,6 +12,7 @@ integration_tests:
   usw:
     - firefox
     - headless
+    - download
   tokyo:
     - headless
   frankfurt:
--- a/jenkins/branches/prod.yml
+++ b/jenkins/branches/prod.yml
@@ -13,6 +13,7 @@ apps:
 integration_tests:
   usw:
     - headless
+    - download
     - firefox
     - chrome
     - ie
--- a/jenkins/branches/run-integration-tests.yml
+++ b/jenkins/branches/run-integration-tests.yml
@@ -8,6 +8,7 @@ apps:
 integration_tests:
   usw:
     - headless
+    - download
     - firefox
     - chrome
     - ie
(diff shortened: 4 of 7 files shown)


pmac:
In testing this I've found something strange. It does seem like the download tests behave differently than the other headless tests. I am locally getting consistent failures when collecting tests for the l10n download tests (`test_download_l10n.py`), but I don't seem to get failures running the tests of the other pages and the redirects. My initial hypothesis is that this difference is somehow related to the requests happening as part of test setup for the download tests, and as part of the tests themselves for the others. Still investigating though.

Tonchi:
[TONCHI]

Tonchi:
>>>_[TONCHI_PASTE_&_GO_FIND_}_<<<

alexgibson:
I haven't started reviewing the code here just yet, as it sounds like you're still investigating (nice detective work!). I will add this though as some other items for consideration:

- Now that we've worked out this isn't a deployment related issue, should we reset the retry's from 3 back to 1? It doesn't seem like 3 made any difference here, and I'm wary that this might pave over other intermittent (or possibly genuine) failures in the future?
- This PR does seem like something worth doing, as I agree testing the same links 6 times per deployment is excessive. It does feel like we're just trying to work around the real issue still though. These tests we're working fairly reliably against both frankfurt and tokyo at one time. I wonder if someone on the SRE team would be able to trace back to the date we started having issues on, to see if any other infra changes we're made around that time that could be impacting this?

pmac:
Seems I've neglected to respond to your excellent feedback. Thought I had. Apologies. Here it is:

Re point 1: Agreed. I'll add that.

Re point 2: Also agreed, but this is still worth doing. We do need to keep investigating, and I'm not super convinced that this will actually stop the failures.

Thanks!

pmac:
I updated the retry on the tests to "2" down from "3". In looking at the test runs I'd expected to see 4 runs (one initial and 3 retries), but I only see 3. So I think the `retry` function is misnamed and is more a `max_number_of_tries`. In other words I think `retry(1)` is a noop.

docker/bin/run_integration_tests.sh
```diff
@@ -25,6 +25,10 @@ case $1 in
     PLATFORM="Windows 7"
     MARK_EXPRESSION="sanity and not headless"
     ;;
+  headlessplus)
```
alexgibson: Could we use a more descriptive name here perhaps? We could even just call this `download`, and then explicitly include both `headless` and `download` in the YAML files where appropriate?

pmac: I wrestled with that too. I don't know what to call it. Could just have both and add the download tests to those places... I guess that does make most sense.

PR #5358, opened 2017-12-20
Only run download test setup when tests are selected

pmac:


alexgibson:
This looks good and works as expected locally for me.

My only question is if we need to update this line also: https://github.com/mozilla/bedrock/blob/master/docker/bin/run_integration_tests.sh#L26

At one time I remember IE sessions we're running headless tests by mistake, but that was part of the old setup. This line could be redundant now, I'm not too sure without further investigation.

pmac:
I looked into it just now and I do not see any overlap between tests marked `sanity` and those marked either `headless` or `download`. So I think we're clear to just update that mark expression for the `ie8` tests, which I'll do now. Good catch!

alexgibson:
Looks good. Only question is do we now need to update this condition also? https://github.com/mozilla/bedrock/blob/master/tests/functional/test_generated_pages.py#L10

pmac:
I looked into it just now and I do not see any overlap between tests marked `sanity` and those marked either `headless` or `download`. So I think we're clear to just update that mark expression for the `ie8` tests, which I'll do now. Good catch!

pmac:
This is updated and ready final review and merge.

pmac:
I don't think we do. You can still run the suite using "not headless" to indicate not to run those tests and it should indeed skip that setup.

pmac:
But then we don't do that in our scripts, so I dunno. It's all very confusing and too tightly coupled.

The good part about the one you linked is that it doesn't really do anything, just returns a list of strings.

PR #5359, opened 2017-12-21
[fix bug 1393035] Plugin check doesn't recognize nightly as FF

alexgibson:
## Description
- Fixes plugincheck to treat pre-release versions of Firefox as up-to-date.
- Also adjusts the logic to only check against the major version number.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1393035#c4

## Testing
- Test pre-release versions of Firefox are considered as up-to-date.
- Test old versions are considered out of date.
- Test latest ESR is considered as up-to-date.

### DIFF
--- a/media/js/plugincheck/plugincheck.js
+++ b/media/js/plugincheck/plugincheck.js
@@ -10,9 +10,11 @@
 
     if (client.isFirefoxDesktop) {
         client.getFirefoxDetails(function(details) {
-            if (details.isUpToDate && details.channel === 'release') {
+            // set `strict` to `false`, so we only test against the major version number.
+            var isUpToDate = client._isFirefoxUpToDate(false, details.isESR, details.version);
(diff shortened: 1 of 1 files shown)


media/js/plugincheck/plugincheck.js
```diff
@@ -10,9 +10,10 @@
 
     if (client.isFirefoxDesktop) {
         client.getFirefoxDetails(function(details) {
-            if (details.isUpToDate && details.channel === 'release') {
+            var isUpToDate = client._isFirefoxUpToDate(false, details.isESR, details.version);
```
jpetto: Only thought here is to add a comment explaining when `isUpToDate` is true and when it's false (as the method call/signature doesn't make it obvious).

alexgibson:
@jpetto updated, thanks

PR #5372, opened 2018-01-04
[no bug] Stop using firefox/new/onboarding.lang in firstrun files

flodolo:


flodolo:
I'm removing a bunch of obsolete files, firefox/new/onboarding should be one of them, but I've discovered it's still used in these files.

We're only missing one string in firefox/new/quantum, so that's an easy fix on our side.

flodolo:
I'm removing a bunch of obsolete files from the l10n repository, `firefox/new/onboarding` should be one of them, but I've discovered it's still used in these files.

We're only missing one string in firefox/new/quantum, so that's an easy fix on our side.

PR #5373, opened 2018-01-04
Fix bug 1428150: Remove old tabzilla

pmac:


bedrock/infobar/tests.py
```diff
@@ -9,11 +9,11 @@
 from mock import patch
 
 from bedrock.mozorg.tests import TestCase
-from bedrock.tabzilla.views import template_last_modified
+from bedrock.infobar.views import template_last_modified
```
alexgibson: Is there a reason to keep this infobar view around still? Afaik it was just a feature of Tabzilla, same as the transbar right?

bedrock/mozorg/redirects.py
```diff
@@ -659,4 +659,12 @@ def decider(request, **kwargs):
 
     # Bug 1384370
     redirect(r'^developers/?$', 'mozorg.developer'),
+
+    # Bug 1428150
+    gone(r'^tabzilla/transbar\.jsonp$'),
+    gone(r'^tabzilla/tabzilla\.js$'),
+    gone(r'^tabzilla/media/js/tabzilla\.js$'),
+    redirect(r'tabzilla/media/css/tabzilla\.css$',
```
alexgibson: Is it worth having a test for this redirect? (or any of the 410's above?)

bedrock/tabzilla/templates/tabzilla/tabzilla.js
```diff
@@ -1,729 +0,0 @@
...
-            var setup = Infobar[value];
-            if (setup) {
-                setup.call();
-            }
-        });
-
-        // Careers teaser in error console.
-        $(window).load(function() {
-            try {
-                console.log("{% filter js_escape|safe %}{% include "includes/tabzilla-console-footer.html" %}{% endfilter %}");
```
alexgibson: Can we remove `includes/tabzilla-console-footer.html` in this PR too?

bedrock/infobar/tests.py
pmac: I'm not sure about that. I'll investigate.

bedrock/mozorg/redirects.py
pmac: Not sure. Probably not...

bedrock/tabzilla/templates/tabzilla/tabzilla.js
pmac: I think that might be the same one that's included in the head of every page? Might not be though. I'll check.

bedrock/infobar/tests.py
pmac: I can't find any reference to it anywhere in the code and I looked in New Relic and it gets a single hit maybe once every few days. I say we nuke it.

bedrock/mozorg/redirects.py
pmac: I've added tests for all of this

bedrock/tabzilla/templates/tabzilla/tabzilla.js
pmac: Yup. Deleted. Good catch.

pmac:
I've updated everything from feedback (finally). Have another look when you've got time.

pmac:
I thought about that. Do you think it'd be worth keeping the page just as something for people who might be looking for info to find? We could add a note about it's deprecation but continue to link to the external repos. Thoughts?

NIT: I always thought r+wc meant that the PR was good to merge as soon as a couple more things were addressed but didn't require further review. But marking it as "Changes requested" means it's not mergable without further review. Just wondering out loud if we should formalize the review process a bit more or if you really did want to review again?

alexgibson:
> NIT: I always thought r+wc meant that the PR was good to merge as soon as a couple more things were addressed but didn't require further review. But marking it as "Changes requested" means it's not mergable without further review. Just wondering out loud if we should formalize the review process a bit more or if you really did want to review again?

I just meant r+ with changes (as in removing the obsolete docs). But if you'd rather leave them for the time being I'll approve 👍

PR #5374, opened 2018-01-05
Remove old search experiment

alexgibson:
@jpetto r?


alexgibson:
Bah, good catch - updated!

PR #5378, opened 2018-01-08
[bug 1402967] Update Quantum link text on /technology

alexgibson:
## Description
- Updated Quantum CTA link on `/technology` to "Learn more", since the link goes to a product page and is not an actual download button.
- Also removes a bunch of l10n tags since [coverage on this page](https://l10n.mozilla-community.org/langchecker/?website=www.mozilla.org&file=mozorg/technology.lang&action=translate) seems complete.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1402967

## Testing
- Check I didn't miss any other occurences of obsolete l10n tags?

### DIFF
--- a/bedrock/mozorg/templates/mozorg/technology.html
+++ b/bedrock/mozorg/templates/mozorg/technology.html
@@ -47,37 +47,25 @@
     <section id="vr" class="odd">
       <div class="content">
         <div class="inner-container">
-        {% if l10n_has_tag('mozorg_tech_webvr') %}
           <h2>{{ _('Bringing Virtual Reality to the web platform') }}</h2>
           <p>{{ _('Using WebVR, developers, designers and artists can share VR experiences on the web.') }}</p>
           <a rel="external" href="https://vr.mozilla.org/" class="button">{{ _('Learn more') }}</a>
-        {% else %}
-          <h2>{{ _('Making Virtual Reality a reality for all') }}</h2>
-          <p>{{ _('Using A-Frame, developers, designers and artists are able to easily create accessible VR experiences.') }}</p>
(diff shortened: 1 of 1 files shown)


craigcook:
The code changes look good to me but pinging @peiying2 to verify if it's OK to remove these l10n tags. I see a handful of locales with the page activated that are missing one or more of these tags, so I think removing them now would show mixed English content. For example 'af' is active but has none of the tags so it still shows a bunch of outdated links and no Quantum section.

I'm not sure how best to proceed. We can keep the l10n tags and just update the link since "Learn more" is already a global string. But it means keeping a ton of tags around forever. Maybe we can send a notice to localizers to update this page soon or risk deactivation?

Of course, how long can it be before this page gets redesigned with all new copy anyway...

flodolo:
We can remove the tags, disable all pages and only enable those that are fully translated at that point.

alexgibson:
All strings here show coverage of between 98% and 102% of our l10n user base, so I think we're ok (as confirmed by @flodolo)

craigcook:
Alright. So we can just deactivate the page for any locale not 100%? Do we need to do that before this PR can be merged, or at least before we push it to production? Looks like that's about 18 languages losing this page, at least until they can update and reactivate. Unless I'm misunderstanding the process (which I probably am).

flodolo:
We should it before push to production, not blocking this PR.

flodolo:
Removed active tag in https://github.com/mozilla-l10n/www.mozilla.org/commit/30479c622f22d0efdac22fd84fbf9b82c29c7654

Will need to clean up tags after this is in prod, leaving it to @peiying2 for the next string export

PR #5379, opened 2018-01-08
Implement font-display: swap; to remove FOIT

alexgibson:
## Description
- Adds `font-display: swap;` descriptor to our generic `@font-face` stack, eliminating FOIT in browsers that support it (Currently in Chrome & FF 58+. It's also in Safari Tech Preview). This makes text content readable much earlier in the page loading process, especially for those on slower connections.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1427719

## Testing
- Demo: https://www-demo1.allizom.org/en-US/
- Hard-refreshing pages in both Chrome and Firefox (58+) is the easiest way to view the change in behavior on individual pages. I also did some basic tests on mobile using web page test, which you can see in the bug.
- Make sure I didn't miss any `@font-face` rules?

### DIFF
--- a/docs/l10n.rst
+++ b/docs/l10n.rst
@@ -350,30 +350,35 @@ Here's an example of ``intl.css``:
     @font-face {
       font-family: X-LocaleSpecific-Light;
       font-weight: normal;
+      font-display: swap;
       src: local(mplus-2p-light), local(Meiryo);
     }
 
     @font-face {
       font-family: X-LocaleSpecific-Light;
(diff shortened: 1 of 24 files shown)


stephaniehobson:
Icon fonts skipped ✅
Display fonts included ✅

Only missing @font-face rules I found was in tabzilla.less but that's going away.

R+

Thoughts on adding `font-display: block` to the icon fonts declarations?

PR #5380, opened 2018-01-09
[fix bug 1405409] Add microdata to /firefox/ page

alexgibson:
## Description
- Adds shared Schema.org microdata include to `/firefox/`.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1405409

## Testing
- Microdata should be present when inspecting the markup for `/firefox/` (but nothing visually noticable to the page).

### DIFF
--- a/bedrock/firefox/templates/firefox/home.html
+++ b/bedrock/firefox/templates/firefox/home.html
@@ -398,6 +398,8 @@
       {{ download_firefox(dom_id='footer-download', alt_copy=_('Download now'), locale_in_transition=True, download_location='other') }}
     </div>
   </section>
+
+  {% include 'firefox/includes/schemaorg-app.html' %}
 </main>
 {% endblock %}
 


flodolo:
This can't work. Tags are associated to strings, there would be no strings in this template to determine if tag should be added or not.

What's the goal of this data? Do we need it outside of en-*?

One possible solution is to have a lang file for `firefox/includes/schemaorg-app`, and check for file activation here. But there's only one string as far as I can see, and it doesn't seem to make a lot of sense.

If there's no plan to extend that file with other info, we could just copy the string from firefox/new to main.lang, and be done with it.

alexgibson:
> If there's no plan to extend that file with other info, we could just copy the string from firefox/new to main.lang, and be done with it.

I think this may work best, as I can't see any further info being added to this any time soon. I'll remove the l10n tag, thanks.

flodolo:
Switching tag, I'll copy the string.

alexgibson:
Thanks @flodolo!

alexgibson:
> What's the goal of this data? Do we need it outside of en-*?

As to the goal, you can read more about it in the bug. Personally, I'm dubious about the benefits of microdata in general, but it does no real harm adding this I think.

PR #5383, opened 2018-01-11
[fix bug 1423436] Whatsnew page promoting Rocket for Indonesia

craigcook:
## Description
Adds a custom whatsnew page promoting Rocket, exclusively in Indonesia, and only for versions 57.0.* and up. All other locales/versions should get their version-appropriate localized page.

Text is hard-coded in Indonesian so no l10n needed (I didn't even wrap the strings). If/when Rocket is ever available in other countries we'll need to rethink this but for now I think hard-coding strings is the way to go.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1423436

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/id/firefox/58.0/whatsnew/

- Verify the page is only shown when locale == id.
- Verify the page is only shown when version is 57.0.* or higher.


craigcook:
Also since I don't read Indonesian I'm asking in the bug for someone to verify the copy is correct, so maybe we should block merging until we get that signoff from a native speaker... Still ready for code review but marking DNM just in case text changes.

craigcook:
Copy verified and approved.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-rocket-id.html
```diff
@@ -0,0 +1,57 @@
...
+      </figure>
+    </div>
+  </div>
+</main>
+{% endblock %}
+
+{% block email_form %}{% endblock %}
+
+{% block js %}
+  {% javascript 'firefox_whatsnew' %}
```
alexgibson: Is this JS bundle needed?

bedrock/firefox/views.py
```diff
@@ -322,6 +322,15 @@ def show_57_firstrun(version):
     return version >= Version('57.0')
 
 
+def show_rocket_whatsnew(version):
```
alexgibson: Would this mean anyone updating from < 57.0 -> 57.0.1 misses the Chris Beard letter? I don't see any specific ask in the bug to target dot releases for 57 specifically, although it is a bit vague on info. There is a [bug for 58](https://bugzilla.mozilla.org/show_bug.cgi?id=1428419) though?

Also, this logic is targeting the `id` locale only. Should we be showing the mobile `/whatsnew` page for all locales on dot releases instead?  As this stands, every other locale would see the cbeard letter again. If we changed the logic to show the mobile `/whatsnew` page for all locales, we could then use a locale-specific template for this page variation?

Some tests for whatever we end up with here would be good too.

media/css/firefox/whatsnew/whatsnew-rocket.scss
```diff
@@ -0,0 +1,83 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import '../../pebbles/includes/lib';
+
```
alexgibson: Nit: this page isn't using our new standard margins? `@import '../../hubs/sections';`

bedrock/firefox/templates/firefox/whatsnew/whatsnew-rocket-id.html
```diff
@@ -0,0 +1,57 @@
...
+
+
+{% block content %}
+<main role="main" class="section">
+  <div class="content">
+    <a class="moz-logo" href="{{ url('mozorg.home') }}"><img src="{{ static('img/logos/mozilla/wordmark-dark.svg')}}" alt="Mozilla" width="101" height="34"></a>
+
+    <div class="rocket-content">
+      <div class="content-intro">
+        {# Rough translation: "You have the latest version of Firefox" #}
```
alexgibson: Useful 👍 (I'll do the same thing for the zh-TW variation)

bedrock/firefox/templates/firefox/whatsnew/whatsnew-rocket-id.html
craigcook: oh, probably not... just carried over from the template I copied 😁

bedrock/firefox/views.py
alexgibson: Could we maybe do some logic like:

- Users updating from < 57.0 see cbeard letter
- Users updating from > 57.0 see mobile promo

This is the same kind of logic we did for the 56 whatsnew page. As it stands here, people would see the cbeard letter multiple times, unless they have `id` as their preferred locale?

craigcook:
Alright, tweaked the logic a bit:

- For 57+ you see the Beard letter when oldversion is less than 57, for locales that have the letter (including `id`). Other locales get the basic Quantum whatsnew page.
- For 57+ you see the reskinned whatsnew page with mobile CTA when oldversion is greater than 57 (eg 57.0.4 > 57.0.3) for all locales except `id` which gets the Rocket page.

bedrock/firefox/views.py
```diff
@@ -313,6 +313,23 @@ def show_57_dev_whatsnew(version):
...
+    except ValueError:
+        return False
+
+    v57 = Version('57.0')
+    v58 = Version('58.0')
+
+    if oldversion:
+        return version >= v57 and version < v58 and oldversion < v57
+    else:
+        return version >= v57
```
alexgibson: Should this be `return version == v57`? 

Currently this always shows the cbeard letter e.g. for 57.0.4 and 58.0 etc. Some tests for this logic would also be good?

alexgibson:
There's a breakpoint at ~1000px where the background seems to repeat:

screen shot 2018-01-15 at 09 41 49

PR #5390, opened 2018-01-12
Bug 1430147: Restore page-image-quantum.png

pmac:
Add a symlink to page-image.png, and since it's the same file as page-image-quantum.png, the hashed name will be the same as it was originally, thus continuing to serve the image at the old URL.

### DIFF
--- a/media/img/firefox/template/page-image-quantum.png
+++ b/media/img/firefox/template/page-image-quantum.png
@@ -1,0 +1,1 @@
+page-image.png


pmac:
I tried a redirect in our normal way, but it didn't work because I think Whitenoise intercepts `/media/*` URLs before the rest of the app is involved.

pmac:
In my local tests this change results in the original hashed filename existing after a build.

Issue #5080, opened 2017-08-30
Upgrade Django to latest 1.11


pmac:
1.11 is the next LTS release and we should move to it. This covers upgrading Django and all other libs required to be updated to support it.

click4code:
Hi

I am interested in this issue. Please tell me how to proceed?

pmac:
This is going to be a lot of work I'm afraid. If you're an experienced Django developer then feel free to give it a shot, but I expect it to be a rather large project as some 3rd party library updates will potentially require major changes.

NdagiStanley:
Hi @pmac 
What's the progress of this issue?

Hasan-Ahmad:
I think we should upgrade to 2.0, it makes life a lot easier with its new syntax

NdagiStanley:
True. It's best we upgrade to version 2.0

NdagiStanley:
@pmac could we create a branch for this upgrade to `Django 2` so that we can collaboratively make this upgrade?
cc @Hasan-Ahmad @click4abhishek 

I'll shift the base of the PR I have raised to that branch.

Cheers.

angelpongo:
Thanks

StephenOrJames:
I would advise against Django 2 for now, especially since it drops support for Python 3 (this repository is still using Python 2). Everything needs to be made Python 3-compatible first, which is, itself, a large challenge (e.g. some packages need to be updated (e.g. some of the current packages are Python 2-only, but have later versions that support Python 3), imports fixed (explicit relative imports, `urllib`/`urlparse` to `urllib.request`/`urllib.parse`), `cStringIO.StringIO`/`StringIO.StringIO` to `io.BytesIO`/`io.StringIO` (as appropriate), etc.

angelpongo:
Thanks for the heads up

pmac:
We do want to get to Django 2 eventually, but for now the best path forward is 1.11. @StephenOrJames is right that it will be quite a lot of work to get to Python 3. I think it's going to be a good bit of work to get to 1.11 even.

BojanKogoj:
As a reminder, Django 1.8 LTS has ended a few days ago (1 April 2018) according to https://www.djangoproject.com/download/#supported-versions

If I understand correctly the plan is:
-  Django 1.11 with Python 2.7
-  Django 1.11 with Python 3+
-  Django 2+

NdagiStanley:
I guess @BojanKogoj. @pmac, right?

[...]

PR #5391, opened 2018-01-14
Upgrade to Django2

NdagiStanley:
## Description
Upgrade the project to Django 2.

## Issue link
#5080

## Testing
Run `py.test lib bedrock` then start the server by running `python manage.py runserver`

### DIFF
--- a/bedrock/externalfiles/__init__.py
+++ b/bedrock/externalfiles/__init__.py
@@ -7,9 +7,9 @@ import logging
 import os.path
 from time import mktime
 try:
-    from cStringIO import StringIO
-except ImportError:
     from StringIO import StringIO
+except ImportError:
+    from io import StringIO
 
 from django.conf import settings
 from django.core.cache import caches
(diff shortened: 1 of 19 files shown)


alexgibson:
Thank you for this PR @NdagiStanley - from the discussion in #5080 it seems like updating to Django 2 (and supporting Python 3) is still some way off our roadmap. Keeping this branch rebased and up-to-date over a long period of time seems an unlikely task, so we should probably consider closing this.

Thought @pmac ?

pmac:
Agreed. Appreciate the effort @NdagiStanley! But I think there are several dependencies that need to be updated to work with Py3 or worked around as well as changes to bedrock itself. We should give it a good attempt again soon though.

NdagiStanley:
Understood @alexgibson and @pmac.
So what are the next steps?

pmac:
I think next is to go through our requirements files and see which do and do not support Python 3.6. The ones that don't we'll have to see if they maybe have a separate Python3 release, or if not we'll have to see if that's something we can contribute to the package or if we can work around not having that package. After that is done the rest of the job should be much easier.

NdagiStanley:
@pmac That makes sense. So let's continue with that on #5080 , cool?
Having a project for this might help make the process easier. So that we have cards for specific packages and have columns for `Python 3 compatible`, `Needs contribution for porting to Python 3` and `Needs an alternative/ a work around`.

What do you think? @alexgibson thoughts?

PR #5400, opened 2018-01-18
[fix bug 1429321] Switch campaign page

alexgibson:
## Description
- Adds `/firefox/switch/` campaign page.
- ~Design is still incoming, this PR is just to expose strings for l10n.~

~**I will update here when strings should be considered final**~

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1429321

## Testing
Demo: https://bedrock-demo-agibson.us-west.moz.works/firefox/switch/


bedrock/firefox/templates/firefox/switch.html
```diff
@@ -0,0 +1,93 @@
...
+                    <li><a class="mail-yahoo" data-link-type="link" data-link-name="Send an email - Yahoo" href="{{ 'https://compose.mail.yahoo.com/?subject=%s&body=%s%s'|format(email_subject, email_body, share_url|urlencode)|e }}">{{ _('Yahoo Mail') }}</a></li>
+                    <li><a class="mail-outlook" data-link-type="link" data-link-name="Send an email - Outlook" href="{{ 'https://mail.live.com/mail/EditMessageLight.aspx?n=&subject=%s&body=%s%s'|format(email_subject, email_body, share_url|urlencode)|e }}">{{ _('Outlook.com') }}</a></li>
+                    <li><a class="mail-other" data-link-type="link" data-link-name="Send an email - Other" href="{{ 'mailto:?subject=%s&body=%s%s'|format(email_subject, email_body, share_url|urlencode)|e }}">{{ _('Other') }}</a></li>
+                  </ul>
+                </details>
+              </li>
+
+              <li class="share-opt-twitter">
+                {# L10n: the string below is for a referral Twitter message. #}
+                {% set tweet_text = _('🔥 Firefox makes switching from Chrome really fast. Try it out!') %}
```
alexgibson: @flodolo do you see any issues using an emoji in a translated string like this?

bedrock/firefox/templates/firefox/switch.html
```diff
@@ -0,0 +1,93 @@
...
+              </li>
+
+              <li class="share-opt-email">
+                <details>
+                  <summary>{{ _('Send an email') }}</summary>
+                  {# L10n: the string below is for a referral email subject line. #}
+                  {% set email_subject = _('Switch to Firefox')|urlencode %}
+
+                  {# L10n: the string below is for a referral email message body. Use \n\n for line breaks in the message as appropriate. #}
+                  {% set email_body = _('Hey, \n\nFirefox makes switching from Chrome really fast. I like it a lot, and you should try it. \n\nCheck it out and let me know what you think: ')|urlencode %}
```
alexgibson: @flodolo also interested in your thoughts here on how best to handle this?

bedrock/firefox/templates/firefox/switch.html
flodolo: Double checked with Théo, he [used emojis](https://github.com/mozilla-l10n/engagement-l10n/blob/master/en-US/social/2017/iot_survey.lang) in the past and we didn't have issues in our toolchain

bedrock/firefox/templates/firefox/switch.html
flodolo: Yes, this should work. What's after the colon though?

alexgibson: After the colon comes the URL, which is populated automatically. Would an additional comment help here?

flodolo: Yep, it'd definitely help

alexgibson:
Strings here are now final, ready for l10n.

bedrock/firefox/templates/firefox/switch.html
```diff
@@ -0,0 +1,87 @@
...
+                  <a data-link-type="link" data-link-name="Share to Facebook" href="https://www.facebook.com/sharer/sharer.php?s=100&p[url]={{ share_url|urlencode }}">
+                    {{ _('Share to Facebook') }}
+                  </a>
+                </li>
+                <li>
+                  {# L10n: the string below is for a referral email subject line. #}
+                  {% set email_subject = _('Switch to Firefox')|urlencode %}
+
+                  {# L10n: the string below is for a referral email message body. Use \n\n for line breaks in the message as appropriate. At the end of the string after the colon, a URL will be added automatically. #}
+                  {% set email_body = _('Hey, \n\nFirefox makes switching from Chrome really fast. I like it a lot, and you should try it. \n\nCheck it out and let me know what you think: ')|urlencode %}
```
flodolo: There are two issues here:
* The link doesn't really work, it creates an email with + instead of sèaces
* The main problem is that `\n` gets lost in the string extraction. It becomes
```
Hey, Firefox makes switching from Chrome really fast. I like it a lot, and you should try it. Check it out and let me know what you think: 
```

And it gets correctly read from the lang file (i.e. it shows up localized if I change it), which means `\n` is not only stripped by the extraction process, but also ignored by Django.

I think we need `<br>` here, or split the string in two parts. Note that this is blocking extracting strings.

flodolo: One more thing that I forgot: we don't support leading or trailing spaces in strings, they're stripped off.

PR #5408, opened 2018-01-22
Fix Bug 1431348 - Unpublished Release notes should not be 404 but should get a default template with a "coming soon" message instead

kyoshino:
## Description

Show a "coming soon" page for unpublished releases on production, where people get a 404 page at this time. The draft notes can still be previewed on stage.

**Do not merge yet** until the copy gets reviewed by @pascalchevrel.

## Issue / Bugzilla link

[Bug 1431348 - Unpublished Release notes should not be 404 but should get a default template with a "coming soon" message instead](https://bugzilla.mozilla.org/show_bug.cgi?id=1431348)

## Testing

This is actually difficult to test with actual release data depending on the timing. Currently [59.0beta](https://www.mozilla.org/en-US/firefox/59.0beta/releasenotes/) is unpublished so this page will be "coming soon" once this PR is merged, like [this screnshot](https://screenshots.firefox.com/MjqzpkbJdVq3rCpA/127.0.0.1).

### DIFF
--- a/bedrock/firefox/templates/firefox/releases/release-notes.html
+++ b/bedrock/firefox/templates/firefox/releases/release-notes.html
@@ -129,11 +129,18 @@
           <div class="version">
             <h2>{{ _('{version}')|f(version=release.version) }}</h2>
             <h3>{{ release.product }} {{ release.channel }}</h3>
-            <p>{{ _('{date}')|f(date=release.release_date|l10n_format_date) }}</p>
+            {% if release.is_public %}
+              <p>{{ _('{date}')|f(date=release.release_date|l10n_format_date) }}</p>
+            {% endif %}
           </div>
           <div class="description">
+            {% if release.is_public %}
               <h2>{{ _('Version {version}, first offered to {channel} channel users on {date}')|f(channel=release.channel, date=release.release_date|l10n_format_date, version=release.version) }}</h2>
               <p>{{ release.text|safe }}</p>
+            {% else %}
+              <h2>{{ _('The notes for this release are not yet publicly available, but are coming soon! Please check this page again later.') }}</h2>
+              {% block extra_draft_description %}{% endblock %}
+            {% endif %}
           </div>
           {% endblock %}
         </div>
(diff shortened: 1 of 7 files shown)


bedrock/firefox/templates/firefox/releases/release-notes.html
```diff
@@ -129,11 +129,19 @@ <h1>{{ _('See what’s new in Firefox!') }}</h1>
           <div class="version">
             <h2>{{ _('{version}')|f(version=release.version) }}</h2>
             <h3>{{ release.product }} {{ release.channel }}</h3>
-            <p>{{ _('{date}')|f(date=release.release_date|l10n_format_date) }}</p>
+            {% if release.is_public or settings.DEV %}
```
pmac: Instead of having to do `or settings.DEV` everywhere, let's put a new property method on the ProductRelease model that returns `self.is_public or settings.DEV`. Should keep the templates a bit more clean. In fact, we already read in the individual notes as `is_public = True` when the site is in DEV mode when the notes are imported, perhaps we should set all notes in the DB as `is_public = True` on import as well when DEV is true and just not have to worry with it at run time?

pascalchevrel:
> This is actually difficult to test with actual release data depending on the timing. Currently 59.0beta is unpublished so this page will be "coming soon" once this PR is merged, like this screnshot.

Once Nightly 60 is ready, we can use the nightly release notes for testing

pmac:
> This is actually difficult to test with actual release data depending on the timing. Currently 59.0beta is unpublished so this page will be "coming soon" once this PR is merged, like this screnshot.

Bedrock pulls release data from [a github repo](https://github.com/mozilla/release-notes/). You should be able to easily fork said repo, make whatever changes you want, update your data from that repo via settings changes, and see what happens. We can also deploy a demo pointed at your fork of the repo to test with if you'd like.

kyoshino:
Sorry for the delay! Just updated the copy and code. As @pmac suggested, all releases are now marked public in DEV mode. [Forked](https://github.com/kyoshino/release-notes/tree/test) the release-notes repo, and tested locally.

PR #5411, opened 2018-01-24
Add functional tests for /firefox/switch and locale-specific /whatsnew pages

alexgibson:
## Description
- Adds functional tests for:
  - `/firefox/switch/`
  - `/id/firefox/whatsnew/`
  - `/zh-TW/firefox/whatsnew/`

## Issue / Bugzilla link
N/A

## Testing
Successful test run: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/168/pipeline/

### DIFF
--- a/tests/functional/firefox/test_switch.py
+++ b/tests/functional/firefox/test_switch.py
@@ -1,0 +1,13 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.firefox.switch import FirefoxSwitchPage
+
+
+@pytest.mark.nondestructive
+def test_download_button_displayed(base_url, selenium):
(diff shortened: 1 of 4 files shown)


tests/functional/firefox/test_switch.py
```diff
@@ -0,0 +1,13 @@
...
+
+import pytest
+
+from pages.firefox.switch import FirefoxSwitchPage
+
+
+@pytest.mark.nondestructive
+def test_download_button_displayed(base_url, selenium):
+    page = FirefoxSwitchPage(selenium, base_url).open()
+    assert page.download_button.is_displayed
```
jpetto: Do you think we should update this method (or maybe add a new method?) like `page.download_button.is_displayed_above_the_fold_on_macbook_air`?

alexgibson: Selenium doesn't test if something is displayed within the visible viewport (just that is is not hidden), because it understands that such a thing is a fools errand 😉

jpetto: :trollface:

PR #5412, opened 2018-01-24
Add Oregon-B cluster to deployment

pmac:
For now this adds it as the last cluster for dev, stage, and prod deployments. It will be shuffeled to the primary cluster when us-west is decommissioned.


jgmize:
for https://github.com/mozmeao/infra/issues/704

PR #5420, opened 2018-01-29
Fix multiple objects returned error for release notes

pmac:
Drafts were not being deleted on refresh of data when DEV = False (as it is in production. This fixes that.


bedrock/releasenotes/models.py
```diff
@@ -124,7 +124,7 @@ def refresh(self):
         release_objs = []
         rn_path = os.path.join(settings.RELEASE_NOTES_PATH, 'releases')
         with transaction.atomic(using=self.db):
-            self.all().delete()
+            self.get_queryset(True).delete()
```
jgmize: would you mind changing the param to `include_drafts=True` so that it's more obvious why this is necessary?

pmac:
Updated. Thanks Josh!

PR #5421, opened 2018-01-30
Fix Bug 1425454 - The onboarding flow should end up focusing the awesomebar instead of the search box in the page

Mardak:
## Description
Use UITour method to navigate to about:newtab (normally `Error: Access to 'about:newtab' from script denied`) that is also being updated to focus the address bar.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1425454

## Testing
https://bug1425454.bmoattachments.org/attachment.cgi?id=8946540

### DIFF
--- a/media/js/firefox/firstrun/firstrun_quantum.js
+++ b/media/js/firefox/firstrun/firstrun_quantum.js
@@ -11,7 +11,7 @@
     }
 
     var beginAnimation = function (syncConfig) {
-        var redirectDest = 'about:home';
+        var redirectDest = 'about:newtab';
 
         var scene = document.getElementById('scene');
         var skipbutton = document.getElementById('skip-button');
@@ -38,7 +38,12 @@
         var onVerificationComplete = function () {
             scene.dataset.signIn = 'true';
             document.getElementById('white-overlay').addEventListener('transitionend', function() {
-                window.location.href = redirectDest;
+                // Specially navigate to about:newtab and focus address bar
+                if (redirectDest === 'about:newtab') {
+                    Mozilla.UITour.showNewTab();
+                } else {
+                    window.location.href = redirectDest;
+                }
             }, false);
         };
 


alexgibson:
Marking this as DNM until the UITour patch is uplifed to Firefox 58.

media/js/firefox/firstrun/firstrun.js
```diff
@@ -35,7 +34,12 @@
             document.getElementById('sunrise').addEventListener('transitionend', function(event) {
                 if (event.propertyName === 'transform') {
                     window.setTimeout(function () {
-                        window.location.href = redirectDest;
+                        // Specially navigate to about:newtab and focus address bar
+                        if (redirectDest === 'about:newtab') {
+                            Mozilla.UITour.showNewTab();
```
alexgibson: This results in `TypeError: Mozilla.UITour is undefined`. It looks like the UITour lib is not included in the [bundle for this page](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/static_media.py#L1218-L1224).

media/js/firefox/firstrun/firstrun.js
```diff
@@ -6,8 +6,7 @@
     'use strict';
 
     var animateSunrise = function () {
-        // Bug 1381051 sign in should redirect to about:home instead of about:newtab.
-        var redirectDest = 'about:home';
+        var redirectDest = 'about:newtab';
```
alexgibson: This particular JS file is used for `/firefox/56.0/firstrun/` and below. Just checking we're ok to redirect to `about:newtab` in older versions, as well as >= 57?

Mardak: Actually, no we don't want to change it for 56 below. about:newtab is still Tiles and about:home is the old about:home (which focuses search box). So I'll undo the changes to this file, and it then won't require the bundle change either.

Mardak:
I updated the PR reverting firstrun changes for pre-quantum. Here's a video showing the 56 behavior (about:home search focus) vs 58 behavior (about:newtab but no address bar focus -- the same behavior currently live of about:home no address bar focus) vs 60 behavior (about:newtab address babr focus):

https://bug1425454.bmoattachments.org/attachment.cgi?id=8947961

Mardak:
In particular, if we merge this now, people on 57+ will see no difference until updating to 60, which will have the address bar get focused. We could get this sooner with 59 firstrun users by uplifting to mozilla-beta.

Mardak:
@alexgibson based on past's comments in the bug, we should be able to merge these changes for Firefox 57+, and if we get it verified while mozilla-beta is still `EARLY_BETA` (before next Tuesday 13th), we can get it uplifted from 60 to 59 if we want this one release earlier.

alexgibson:
LGTM, thanks @Mardak 👍

PR #5426, opened 2018-02-01
Fix bug 1429429: Update headline on /firefox/mobile

stephaniehobson:
## Description

- Updated headline
- Changed subheading text
- Focus download button is green now
- Product backgrounds are now white
- Changed to black outlines around phones
- Larger product logos
- Spacing adjusted

## Issue / Bugzilla link
[Bug 1429429: Update headline on /firefox/mobile/](https://bugzilla.mozilla.org/show_bug.cgi?id=1429429)

## Testing
- [ ] RTL
- [ ] All 3 breakpoints
- [ ] Without localized tags
- [ ] With longer strings for translation


bedrock/firefox/templates/firefox/mobile.html
```diff
@@ -26,9 +26,20 @@
     <header class="main-header">
       <div class="content">
         <div class="header-content">
-          <h1>{{ _('Firefox for mobile') }}</h1>
-          <h3>{{ _('The speed you need with the privacy you want on all your devices.') }}</h3>
-          <p>{{ _('Choose one or download them all for iOS and Android.') }}</p>
+          <div class="header-title">
+              <h1>{{ _('Firefox for mobile') }}</h1>
+              {% if l10n_has_tag('fast-012018') %}
```
flodolo: It doesn't break anything, but please use underscore for tags

P.S. indentation seems off for the endif closing tag

media/css/firefox/mobile.scss
```diff
@@ -106,12 +119,43 @@ a.button.button-focus {
...
+
+    .header-title,
+    .header-info {
+        display: table-cell;
+        width: 50%;
+    }
+
+    .header-title {
+        padding-right: $header-product-tablet-padding / 2;
+        white-space: nowrap;
```
jpetto: I'm not sure this is going to work with localized text, as `nowrap` overrides the `width: 50%`. In the past, we've used a one-off `html[lang^="en-"]` selector to keep English copy from wrapping (in the event that's important here).

screen shot 2018-02-05 at 5 06 19 pm

media/css/firefox/mobile.scss
```diff
@@ -106,12 +119,43 @@ a.button.button-focus {
...
+    }
+
+    .header-info {
+        padding-bottom: 0.75em; // keeps bottom of text from going below header-title
```
jpetto: Will this perform as intended when all the strings are localized and have radically variable lengths?

media/css/firefox/mobile.scss
```diff
@@ -106,12 +119,43 @@ a.button.button-focus {
     }
 }
 
-html[dir="rtl"] .main-header {
-    @media #{$mq-desktop} {
-        text-align: right;
```
jpetto: I think we need to retain this (or something like it) for RTL.

bedrock/firefox/templates/firefox/mobile.html
```diff
@@ -26,9 +26,20 @@
...
-          <p>{{ _('Choose one or download them all for iOS and Android.') }}</p>
+          <div class="header-title">
+            <h1>{{ _('Firefox for mobile') }}</h1>
+            {% if l10n_has_tag('fast_012018') %}
+              <p class="header-tagline">{{ _('Fast for good') }}</p>
+            {% endif %}
+          </div>
+          <div class="header-info">
+            <h3>{{ _('The speed you need with the privacy you want on all your devices.') }}</h3>
+            {% if l10n_has_tag('available_012018') %}
```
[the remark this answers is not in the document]
flodolo: As far as I see, see are all new strings introduced in this PR, so they should be under the same tag.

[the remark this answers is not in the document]
stephaniehobson: 👍

media/css/firefox/mobile.scss
stephaniehobson: This is kind of the intended effect. It's a lot more like the comp I got from George than the default English display is:

screen shot 2018-02-06 at 13 45 36

But I agree this language and breakpoint look bad so I'll play around with it.

media/css/firefox/mobile.scss
stephaniehobson: It working as intended so I think I need to re-word the comment ;)

media/css/firefox/mobile.scss
stephaniehobson: My kingdom for container queries.

stephaniehobson: Also, how did you get it to show the l10n fallbacks? My local dev shows the untranslated strings for things missing a translation.

jpetto: You can either set `Dev=False` in your `.env` file (the classy way), or just hack it to `if not l10n_has_tag('...')`.

stephaniehobson:
`logo-focus-high-res.png` does not want to get any smaller :/

stephaniehobson:
Updated 🎉

flodolo:
Passing by note: don't forget to tag a PR with L10N if it has new strings.

media/css/firefox/mobile.scss
```diff
@@ -196,6 +231,7 @@ html[dir="rtl"] .main-header {
         .header-product-image {
             display: block;
             float: right;
+            margin-top: (($header-product-overflow + $header-product-tablet-padding) * -1);
```
jpetto: I'm wondering if this should be `0` at `$mq-desktop` (and then back to this calculated value at `$mq-desktop-wide`). Currently looks a bit odd/imbalanced at around 1000px wide.

(Not a blocker, and I wouldn't recommend doing a full creative review here. Skip if it's re-engaging a long process.)

screen shot 2018-02-07 at 1 25 42 pm

stephaniehobson:
ImageAlpha makes it a *bigger* file when I do it. Maybe email me yours?

media/css/firefox/mobile.scss
stephaniehobson: This is updated but not reflected in the diff at this point.

stephaniehobson:
Updated 🎉

PR #5429, opened 2018-02-01
Test frankfurt stage without deploying

jgmize:


jgmize:
https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/frankfurt-stage-test-only/5/pipeline

PR #5431, opened 2018-02-02
Fix attribution_code with non-ascii domain referrers

pmac:
These were causing exceptions due to encoding errors. The stub attribution service has a whitelist that doesn't include any such domains anyway. We now ignore those.


pmac:
See https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/383667/

PR #5432, opened 2018-02-02
Fix Bug 1430027 - On the download page, for Linux, highlight the 64-bit version (to limit the number of 32 downloads)

kyoshino:
## Description

* Highlight recommended builds on the [Firefox download page](https://www.mozilla.org/en-US/firefox/all/)
* l10n: there is 1 new string: _Recommended_

## Issue / Bugzilla link

[Bug 1430027 - On the download page, for Linux, highlight the 64-bit version (to limit the number of 32 downloads)](https://bugzilla.mozilla.org/show_bug.cgi?id=1430027)

## Testing

* [Demo server](https://bedrock-demo-firefox-all.us-west.moz.works/en-US/firefox/all/) is available
* The change should not affect download pages other than Firefox for desktop.

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -42,6 +42,12 @@ class FirefoxDesktop(_ProductDetails):
         ('linux', 'Linux 32-bit'),
     ])
 
+    # Recommended/modern vs traditional/legacy platforms
+    platform_classification = OrderedDict([
+        ('recommended', ('win64', 'osx', 'linux64')),
+        ('traditional', ('win', 'winsha1', 'linux')),
+    ])
+
     # Human-readable channel names
     channel_labels = {
         'nightly': _('Firefox Nightly'),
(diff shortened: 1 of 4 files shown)


flodolo:
> l10n: there are 2 new strings: Recommended and Legacy.

Are the strings final here?

kyoshino:
@flodolo based on the discussion in the bug, _Legacy_ won't be used and only _Recommended_ will be added as a new string.

flodolo:
> @flodolo based on the discussion in the bug, Legacy won't be used and only Recommended will be added as a new string.

Thanks, extracting that string then.

pascalchevrel:
@flod are we good now with l10n coverage?

flodolo:
Yes, I think we're good (also, cc @peiying2)

bedrock/firefox/firefox_details.py
```diff
@@ -341,7 +353,7 @@ class FirefoxAndroid(_ProductDetails):
         'x86': archive_url_base + '-%s/fennec-%s.multi.android-i386.apk',
     }
 
-    def platforms(self, channel='release'):
+    def platforms(self, channel='release', classified=False):
```
jpetto: Could we add a comment here explaining why this method has an unused `classified` parameter?

bedrock/firefox/views.py
```diff
@@ -265,7 +265,8 @@ def all_downloads(request, platform, channel):
 
     context = {
         'platform': platform,
-        'platforms': product.platforms(channel),
+        'platforms': product.platforms(channel, True),
+        'platform_cls': product.platform_classification,
```
jpetto: While this works for `FirefoxAndroid` (I believe through the `__getattr__` implementation in `ProductDetails`), I think it would be clearer to explicitly set `plaform_classification = None` (with a comment) in the `FirefoxAndroid` class in `firefox_details.py`. 

This will make understanding/reading the code quite a bit simpler for those unfamiliar with `ProductDetails`.

jpetto:
Thanks @kyoshino!

PR #5433, opened 2018-02-05
Update deprecated jQuery-related JS. (jQuery 3.2.1)

jpetto:
## Description

Updates jQuery JS that's now deprecated after the jQuery 3.2.1 update.

## Testing

Ensure all functionality remains intact/no regressions.


media/js/base/nav-main.js
```diff
@@ -14,7 +14,7 @@ $(document).ready(function() {
     var newItem;
     var focusedItem;
 
-    $('#nav-main > .has-submenus > li').bind('mouseover focusin', function() {
+    $('#nav-main > .has-submenus > li').on('mouseover focusin', function() {
```
alexgibson: Is this file still used anywhere? I can't see it in any bundles

media/js/mozorg/history-slides.js
```diff
@@ -34,6 +34,7 @@
 
     // set up the slideshow
     // @requires: sequencejs
+    // NOTE: this version of sequence relies on deprecated jQuery methods
```
alexgibson: Is this something that is likely to break in future jQuery updates? Wondering how we don't forget about it if so...

alexgibson:
Forgot to metion - a successful functional test run would be nice to see here before we merge?

media/js/base/nav-main.js
jpetto: Good catch. Removing!

media/js/mozorg/history-slides.js
jpetto: Unsure here. I don't know how soon the jQuery team plans to completely remove the deprecated methods relied on by this library.

Could we perhaps write functional tests for this page that will fail if we update jQuery to a version that no longer supports the methods used in the sequence library?

Additionally, we could file a bug to update this page to a newer library (or maybe even a new design), making it a priority by it being a blocker for us to stay on the latest version of jQuery.

alexgibson: Hmm, we do seem to have some [functional tests for this page](https://github.com/mozilla/bedrock/blob/master/tests/functional/test_history.py). Maybe that will be enough?

Either way, this is a very old page and probably in need of an update/redesign. Filing a bug may be worthwhile just to get it on the radar?

jpetto: Will file a bug 👍

jpetto:
[Functional tests are passing](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/176/pipeline).

PR #5435, opened 2018-02-06
Fix errors with redirects with merge_queries=True

pmac:
Found errors via Sentry:
https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/611704/

### DIFF
--- a/bedrock/redirects/tests/test_util.py
+++ b/bedrock/redirects/tests/test_util.py
@@ -181,6 +181,20 @@ class TestRedirectUrlPattern(TestCase):
         self.assertTrue(url.path, 'abides')
         self.assertEqual(query_dict, {'aggression': ['not_stand'], 'hates': ['the-eagles']})
 
+    def test_merge_query_no_url_query(self):
+        """
+        Should merge query params if requested and no params in URL
+        """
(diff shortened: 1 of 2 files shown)


bedrock/redirects/tests/test_util.py
```diff
@@ -181,6 +181,20 @@ def test_merge_query(self):
         self.assertTrue(url.path, 'abides')
         self.assertEqual(query_dict, {'aggression': ['not_stand'], 'hates': ['the-eagles']})
 
+    def test_merge_query_no_url_query(self):
+        """
+        Should merge query params if requested and no params in URL
+        """
+        pattern = redirect(r'^the/dude$', 'abides',
+                           query={'aggression': 'not_stand'}, merge_query=True)
```
jpetto: You know I have the utmost respect for all things Lebowski, but this test would be easier to read/follow with recognizable URLs. 😬

pmac: I just copied a similar test from above it. I'm sure the values in that test are my fault as well, but I don't think I should change them all.

jpetto: Fair enough 👍

PR #5437, opened 2018-02-07
[fix bug 1405822] Set Firefox Accounts email-first form as default on /firstrun

alexgibson:
## Description
- Sets the winner of email-first form experiment ([bug 1405822](https://bugzilla.mozilla.org/show_bug.cgi?id=1405822)) as the new default experience on /firstrun.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1405822

## Testing
Demo: https://www-demo3.allizom.org/en-US/firefox/58.0/firstrun/
- Ensure signup flow still works as intended (requires both [UITour](http://bedrock.readthedocs.io/en/latest/uitour.html#local-development) and FxA to be [configured on demo](http://bedrock.readthedocs.io/en/latest/firefox-accounts.html#demo-server-testing)]


alexgibson:
This is DNM for now, until we verify the full signup flow is working on the regular staging server.

peiying2:
@alexgibson let me know if the strings are final for l10n.

alexgibson:
> @alexgibson let me know if the strings are final for l10n.

Yes final, thanks

alexgibson:
Ok, this is now ready for review.

docs/firefox-accounts.rst
```diff
@@ -54,31 +54,13 @@ Local Development
...
-        user_pref("webchannel.allowObject.urlWhitelist", "https://accounts.firefox.com https://accounts.stage.mozaws.net https://content.cdn.mozilla.net https://input.mozilla.org https://support.mozilla.org https://install.mozilla.org https://stomlinson.dev.lcip.org/");
-
-        user_pref("general.warnOnAboutConfig", false);
-        user_pref("devtools.chrome.enabled", true);
-        user_pref("devtools.debugger.remote-enabled", true);
-
-#. Start Firefox using the new profile you created in step 2.
-#. Verify the FxA settings look correct by opening ``about:config`` and searching for
-   ``identity.fxaccounts``.
+#. Open ``about:profiles``.
```
alexgibson: @jpetto I added some handy new steps for testing the FxA form on demo that [Ryan pointed me to in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1405822#c28). I updated the docs here, but would be good if you could try these steps out too. Seems much simpler? :)

A didn't update the instructions for local testing, because they seem, well... broken for a long time (should we even be keeping them around)?

alexgibson: Also, fwiw I don't seem to get verification emails for Mailinator any longer, although other email addresses seem to work fine.

jpetto: I think Mailinator addresses are now flagged/blocked on the FxA side.

jpetto: Wow these new instructions are so much easier. Excellent.

I would say we can drop the local instructions for now. AFAIK, they *have* been broken for a very long time. It would be nice to get them fixed (as pushing to a demo for testing isn't the fastest), but I don't see a reason to keep broken instructions around. Maybe we file a bug for local testing?

PR #5438, opened 2018-02-07
[fix bug 1407278] Update 'click here' link text on /firefox/new/?scene=2

alexgibson:
## Description
- Updates the "click here" link copy on `/firefox/new/?scene=2` to be less mouse-pointer centric (also a bit more descriptive for screen readers).

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1407278

## Testing
- Ensure no regressions in new & fallback copy.

### DIFF
--- a/bedrock/firefox/templates/firefox/new/scene2.html
+++ b/bedrock/firefox/templates/firefox/new/scene2.html
@@ -38,9 +38,17 @@
 <p class="help">
   {# fallback_url is replaced by the platform download link via JS, but if
       something fails the user should still get a link to a working download path. #}
-  {% trans id='direct-download-link', fallback_url=url('firefox.all') %}
-    Your download should begin automatically. If it doesn’t, <a id="{{ id }}" href="{{ fallback_url }}">click here</a>.
-  {% endtrans %}
+
+  {% set id, fallback_url = 'direct-download-link', url('firefox.all') %}
+  {% if l10n_has_tag('firefox_new_try_again_link_020818') %}
+    {% trans %}
+      Your download should begin automatically. Didn’t work? <a id="{{ id }}" href="{{ fallback_url }}">Try downloading again</a>.
+    {% endtrans %}
+  {% else %}
+    {% trans %}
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/new/scene2.html
```diff
@@ -38,9 +38,17 @@ <h1>{{ _('Let’s do this!') }}</h1>
 <p class="help">
   {# fallback_url is replaced by the platform download link via JS, but if
       something fails the user should still get a link to a working download path. #}
-  {% trans id='direct-download-link', fallback_url=url('firefox.all') %}
-    Your download should begin automatically. If it doesn’t, <a id="{{ id }}" href="{{ fallback_url }}">click here</a>.
-  {% endtrans %}
+
+  {% set id, fallback_url = 'direct-download-link', url('firefox.all') %}
+  {% if l10n_has_tag('firefox_new_try_again_link') %}
```
stephaniehobson: Should this have a date or a bug number in it?

alexgibson: I don't believe we have any real conventions for formatting l10n tags, aside from using underscores to separate words. The only thing that matters technically, is that the tag name is unique in the lang file. I don't believe it's really something that localizers would use for reference. Either way, adding a date is a good extra measure, so I'll do that here 👍

alexgibson:
Thanks for the review @stephaniehobson!

I don't believe we need to do any additional for tracking the link for GA here (All link clicks are passed to GTM by default, and this page is pretty heavily tracked for events). Still, I'll I can try and confirm with Gareth on this.

peiying2:
@alexgibson let me know if the strings are final for l10n.

alexgibson:
> @alexgibson let me know if the strings are final for l10n.

Yes final, thanks

alexgibson:
Ok, I think we should probably merge this as-is. To answer the question on GA, I checked with Peter and we don't currently track the link. If we want to add this, I think it's likely outside of the scope of this copy change and may be better suited to a separate bug (it may require some more complex JS to track the platform etc).

PR #5439, opened 2018-02-07
[fix bug 1429849] Hide button in top nav on /firefox/mobile/

alexgibson:
## Description
- Hides button in the top nav (& sticky sub nav) on `/firefox/mobile/` which both download the desktop product.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1429849

## Testing
- Both global nav and sub nav buttons should be hidden.

### DIFF
--- a/media/css/firefox/mobile.scss
+++ b/media/css/firefox/mobile.scss
@@ -8,6 +8,7 @@
 @import '../hubs/masthead';
 @import '../hubs/sections';
 @import '../hubs/sub-nav';
+@import '../hubs/nav-download-button-remove';
 @import '../quantum/common';
 @import '../quantum/features-scroller';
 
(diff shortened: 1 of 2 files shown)


stephaniehobson:
Needed update and working as expected 👍 

Not a blocker: do you have any time to fix the white space this leaves in the nav-menu? This is probably only worth while if there are other use-cases for hiding the download button?

screen shot 2018-02-07 at 15 53 26

alexgibson:
@stephaniehobson thanks for the review and suggestion on hiding the white space :)

I think it's likely that we may want / need to do this on some other key pages in the future, so I created a new import for these overrides, to hopefully make this a bit more reusable. I also realized I needed to hide the download button on non-en-US localized pages, so I included that here too.

media/css/firefox/mobile.scss
```diff
@@ -8,6 +8,7 @@
 @import '../hubs/masthead';
 @import '../hubs/sections';
 @import '../hubs/sub-nav';
+@import '../hubs/nav-download-button-overrides';
```
stephaniehobson: Since this file removes the button and is not a good place to put other overrides can we name it something like `nav-download-button-remove`?

alexgibson:
@stephaniehobson thanks, updated!

PR #5443, opened 2018-02-08
[bug 1434764] Add TrafficCop experiment for waitface video

hoosteeno:
## Description
Add a trafficcop experiment to send half of visitors to an xv template to a different xv template, if they are non-Firefox desktop users in en-US with specific parameters.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1434764

## Testing
Unit testing of templates passes after adjusting tests.
Sparse functional testing works as expected -- 50% of chrome visits end at the 2nd template.

## Questions for reviewers
Please consider the way this PR works with params. I believe analytics will work as we want it to work with this setup, and user flow works as expected, but would appreciate scrutiny of this approach.

### DIFF
--- a/bedrock/firefox/templates/firefox/new/wait-face/base.html
+++ b/bedrock/firefox/templates/firefox/new/wait-face/base.html
@@ -12,6 +12,12 @@
   {% endif %}
 {% endblock %}
 
+{% block experiments %}
+  {% if switch('experiment-firefox-new-waitface', ['en-US']) %}
+    {% javascript 'experiment_firefox_new_waitface' %}
+  {% endif %}
+{% endblock %}
+
 {% block extra_meta %}
   <meta name="robots" content="noindex,follow">
 {% endblock %}
--- a/bedrock/firefox/tests/test_views.py
+++ b/bedrock/firefox/tests/test_views.py
@@ -754,14 +754,26 @@ class TestFirefoxNew(TestCase):
 
     # wait face video experiment bug 1431795
 
-    def test_wait_face_video_scene_1(self, render_mock):
-        req = RequestFactory().get('/firefox/new/?xv=2xfaster')
+    def test_wait_face_video_var_a_scene_1(self, render_mock):
+        req = RequestFactory().get('/firefox/new/?xv=waitface&v=a')
+        req.locale = 'en-US'
+        views.new(req)
+        render_mock.assert_called_once_with(req, 'firefox/new/wait-face/scene1.html')
+
+    def test_wait_face_video_var_a_scene_2(self, render_mock):
+        req = RequestFactory().get('/firefox/new/?scene=2&xv=waitface&v=a')
+        req.locale = 'en-US'
+        views.new(req)
+        render_mock.assert_called_once_with(req, 'firefox/new/wait-face/scene2.html')
+
+    def test_wait_face_video_var_b_scene_1(self, render_mock):
+        req = RequestFactory().get('/firefox/new/?xv=waitface&v=b')
         req.locale = 'en-US'
         views.new(req)
         render_mock.assert_called_once_with(req, 'firefox/new/wait-face/scene1-video.html')
 
-    def test_wait_face_video_scene_2(self, render_mock):
-        req = RequestFactory().get('/firefox/new/?scene=2&xv=2xfaster')
+    def test_wait_face_video_var_b_scene_2(self, render_mock):
+        req = RequestFactory().get('/firefox/new/?scene=2&xv=waitface&v=b')
         req.locale = 'en-US'
         views.new(req)
         render_mock.assert_called_once_with(req, 'firefox/new/wait-face/scene2.html')
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -519,8 +519,6 @@ def new(request):
                 template = 'firefox/new/sem/unsupported-browser/scene2.html'
             elif experience in ['batmfree', 'batmnimble', 'batmresist']:
                 template = 'firefox/new/batm/scene2.html'
-            elif experience == '2xfaster':
-                template = 'firefox/new/wait-face/scene2.html'
             else:
                 template = 'firefox/new/scene2.html'
         else:
@@ -535,7 +533,10 @@ def new(request):
         elif lang_file_is_active('firefox/new/sem', locale) and experience == 'fast':
             template = 'firefox/new/sem/fast/scene1.html'
         elif lang_file_is_active('firefox/new/wait-face', locale) and experience == 'waitface':
-            template = 'firefox/new/wait-face/scene1.html'
+            if variant == 'b':
+                template = 'firefox/new/wait-face/scene1-video.html'
+            else:
+                template = 'firefox/new/wait-face/scene1.html'
         elif lang_file_is_active('firefox/new/reggiewatts', locale) and experience == 'reggiewatts':
             template = 'firefox/new/reggie-watts/scene1.html'
         elif locale == 'en-US':
@@ -571,8 +572,6 @@ def new(request):
                 template = 'firefox/new/batm/nimble.html'
             elif experience == 'batmresist':
                 template = 'firefox/new/batm/resist.html'
-            elif experience == '2xfaster':
-                template = 'firefox/new/wait-face/scene1-video.html'
             else:
                 template = 'firefox/new/scene1.html'
         else:
--- a/media/js/firefox/new/experiment-firefox-new-waitface.js
+++ b/media/js/firefox/new/experiment-firefox-new-waitface.js
@@ -1,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+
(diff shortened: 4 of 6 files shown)


media/js/firefox/new/experiment-firefox-new-waitface.js
```diff
@@ -0,0 +1,46 @@
...
+
+/* Experiment: https://bugzilla.mozilla.org/show_bug.cgi?id=1434764
+ * URL: https://www.mozilla.org/en-US/firefox/new/?xv=waitface
+ * Audience: Desktop, Non-Fx, en-US with utm_medium=display&utm_source=doubleclick    
+ * Traffic: 100% audience, 50/50 split*/
+
+(function() {
+    'use strict';
+
+    var isDesktop = function() {
```
alexgibson: We have a [client detection library](https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-client.js) in our common page bundle for convenience. You should be able to do the following:

```javascript
var client = Mozilla.Client;

if (client.isDesktop && !client.isFirefox && paramsMatch()) {
    // run experiment
}
```

media/js/firefox/new/experiment-firefox-new-waitface.js
```diff
@@ -0,0 +1,46 @@
...
+        return ! /android|ios/.test(site.platform);
+    };
+
+    var notFirefox = function () {
+        return ! /\sFirefox/.test(navigator.userAgent);
+    };
+
+    var paramsMatch = function() {
```
alexgibson: Likewise here we also have a [search param library](https://github.com/mozilla/bedrock/blob/master/media/js/base/search-params.js) in our common bundle that makes it easy to get URL params. That may be helpful here.

media/js/firefox/new/experiment-firefox-new-waitface.js
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+        return true;
+    };
+
+    if (isDesktop() && notFirefox() && paramsMatch()) {
+        var cop = new Mozilla.TrafficCop({
+            id: 'experiment_firefox_new_waitface',
+            variations: {
+                novariation: 50,
```
alexgibson: This experiment is mixing both a redirect and a callback variation, which is something I'm not sure we've done before. The [docs](https://github.com/mozilla/trafficcop/blob/master/documentation.md) don't specifically say we should/should not do this, but one thing I'm not sure about is how we would track the double-control group here without a unique URL. Thoughts on this @jpetto?

media/js/firefox/new/experiment-firefox-new-waitface.js
```diff
@@ -0,0 +1,46 @@
...
+        return ! /android|ios/.test(site.platform);
+    };
+
+    var notFirefox = function () {
+        return ! /\sFirefox/.test(navigator.userAgent);
+    };
+
+    var paramsMatch = function() {
+        var params = {
+            'utm_medium': 'display',
```
alexgibson: Would we want to make sure these utm params are also passed along to scene2? (not sure it that's needed or useful here).

bedrock/firefox/views.py
```diff
@@ -540,7 +540,10 @@ def new(request):
         elif lang_file_is_active('firefox/new/sem', locale) and experience == 'fast':
             template = 'firefox/new/sem/fast/scene1.html'
         elif lang_file_is_active('firefox/new/wait-face', locale) and experience == 'waitface':
-            template = 'firefox/new/wait-face/scene1.html'
+            if variant == '2xfaster':
```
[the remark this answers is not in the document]
hoosteeno: I think so, but I wasn't comfortable changing how the download button works without discussing. Can I just add additional parameter handling to /bedrock/media/js/firefox/new/wait-face-scene1.js to make the download button push these params through?

media/js/firefox/new/experiment-firefox-new-waitface.js
hoosteeno: Regarding this and other questions like it: Mozilla.Client and many other utilities are loaded at the bottom of the document by a base template, and TC is loaded at the top, and those both seem like good things. Rather than change one of those (which might affect perceived performance), @jpetto and I agreed I should just pull out the relevant bits. But I can change it easily if you think I should.

media/js/firefox/new/experiment-firefox-new-waitface.js
hoosteeno: Yes, that's a fine idea. I can do that in the page's JS file.

media/js/firefox/new/experiment-firefox-new-waitface.js
alexgibson: Ah, yes - I neglected to consider where in the document this file is loaded. I think what you've done makes sense.

bedrock/firefox/views.py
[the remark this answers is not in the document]
alexgibson: > Can I just add additional parameter handling to /bedrock/media/js/firefox/new/wait-face-scene1.js to make the download button push these params through?

I think that should be do-able yeah

media/js/firefox/new/experiment-firefox-new-waitface.js
jpetto: Also note that in the past we've often supplied a control variation to account for the redirect time. E.g., you may want to add a `'v=control': 50` that will simply redirect to control template.

hoosteeno:
The latest PR decouples this experiment from the data-experience attribute in markup. It repackages parameters in order to pass them from scene1 to scene2. I left the data-experience attribute in place because it is such a common pattern throughout the site, it might be wanted on waitface again sometime.

media/js/firefox/new/experiment-firefox-new-waitface.js
```diff
@@ -0,0 +1,46 @@
...
+            }
+        }
+        return true;
+    };
+
+    if (isDesktop() && notFirefox() && paramsMatch()) {
+        var cop = new Mozilla.TrafficCop({
+            id: 'experiment_firefox_new_waitface',
+            variations: {
+                'v=control': 50, // 50% redirected to control
```
alexgibson: Nit: suggest sticking to our standard convention of using `v=a` and `v=b` for experiment variations.

media/js/firefox/new/experiment-firefox-new-waitface.js
```diff
@@ -0,0 +1,46 @@
...
+    var notFirefox = function () {
+        return ! /\sFirefox/.test(navigator.userAgent);
+    };
+
+    var paramsMatch = function() {
+        var params = {
+            'utm_medium': 'display', 
+            'utm_source': 'doubleclick'
+        };
+        for(var key in params) {
```
alexgibson: Nit: new line before `for` and space before braces.

media/js/firefox/new/experiment-firefox-new-waitface.js
```diff
@@ -0,0 +1,46 @@
...
+ * Traffic: 100% audience, 50/50 split*/
+
+(function() {
+    'use strict';
+
+    var isDesktop = function() {
+        return ! /android|ios/.test(site.platform);
+    };
+
+    var notFirefox = function () {
```
alexgibson: NIt: There's no real need for these to be functions. You could just do:

```
 var isDesktop = !/android|ios/.test(site.platform);
 var notFirefox = !/\sFirefox/.test(navigator.userAgent);
```

media/js/firefox/new/experiment-firefox-new-waitface.js
```diff
@@ -0,0 +1,46 @@
...
+        return ! /\sFirefox/.test(navigator.userAgent);
+    };
+
+    var paramsMatch = function() {
+        var params = {
+            'utm_medium': 'display', 
+            'utm_source': 'doubleclick'
+        };
+        for(var key in params) {
+            if (! (new RegExp('[\&\?]' + key + '=' + params[key])).test(location.search)) {
```
alexgibson: Nit: no need for an extra space after `!`.

bedrock/firefox/tests/test_views.py
```diff
@@ -755,7 +755,7 @@ def test_wait_face_scene_2(self, render_mock):
     # wait face video experiment bug 1431795
 
     def test_wait_face_video_scene_1(self, render_mock):
-        req = RequestFactory().get('/firefox/new/?xv=2xfaster')
+        req = RequestFactory().get('/firefox/new/?xv=waitface&v=2xfaster')
```
alexgibson: [Tests are currently failing in CI](https://circleci.com/gh/mozilla/bedrock/8908?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link). I think you just need to update the params like you did here, for the `test_wait_face_video_scene_2` test directly below.

bedrock/firefox/templates/firefox/new/wait-face/base.html
```diff
@@ -12,6 +12,12 @@
   {% endif %}
 {% endblock %}
 
+{% block experiments %}
+  {% if switch('experiment-firefox-new-waitface'), ['en-US'] %}
```
hoosteeno: Wow, this is a sneaky bug. The locale here is not applied, and no error is thrown, because the close paren comes too early.

alexgibson: Sneaky indeed, nice spot 👍

PR #5450, opened 2018-02-14
Delete mozilla.html

ptrpov:


alexgibson:
Can you please provide a related bug or issue as to why this template should be removed?

alexgibson:
The template is still in use, so removing this will likely break things: https://github.com/mozilla/bedrock/blob/master/bedrock/legal/urls.py#L22

PR #5456, opened 2018-02-16
[fix bug 1438464] Decommission Lightbeam pages

alexgibson:
## Description
- Removes Lightbeam app and pages under `/lightbeam/`

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1438464

## Testing
Make sure I didn't miss any stray references of internal URLs?


bedrock/settings/base.py
```diff
@@ -517,7 +516,6 @@ def set_whitenoise_headers(headers, path, url):
     '/privacy/',
     '/foundation/',
     '/teach/',
-    '/lightbeam/',
```
alexgibson: @pmac not sure if this is important or not, but I'm removing this from the `VaryNoCacheMiddleware` list for CloudFlare.

pmac: If those URLs are gone then this is fine 👍

PR #5460, opened 2018-02-19
[fix bug 1438307] Update branding on /firefox/accounts/ page

alexgibson:
## Description
- Updates branding on `/firefox/accounts/` page and converts CSS from Sandstone to Pebbles.
- Changes Firefox Accounts iframe from sign-up to email-only variation.

**Note:** No string changes. Do not merge for now until we have creative approval.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1438307

## Testing
Demo: https://www-demo3.allizom.org/en-US/firefox/accounts/

To test on demo, please make sure you have both [UITour enabled](http://bedrock.readthedocs.io/en/latest/uitour.html#local-development), and set up a [Firefox profile](http://bedrock.readthedocs.io/en/latest/firefox-accounts.html#demo-server-testing) as per the docs.


media/js/firefox/accounts.js
```diff
@@ -5,18 +5,9 @@
 (function(Mozilla, $) {
     'use strict';
 
-    var fxaEmail;
     var client = Mozilla.Client;
     var $main = $('main');
 
-    // check for fxaEmail in sessionStorage or URL params
-    try {
-        fxaEmail = sessionStorage.getItem('fxa-email');
```
alexgibson: This seems to be from some past experiment, and I can find no reference to any page setting this any longer - so can only assume it's now fine to remove.

jpetto: FWIW, I think this was from an experiment on `/firerox/sync/`.

media/css/base/mozilla-fxa-iframe.scss
```diff
@@ -0,0 +1,11 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+#fxa {
+    opacity: 0;
+
+    &.loaded {
+        opacity: 1;
+    }
+}
```
jpetto: Guessing this is intended to be used on other FxA pages at a later date?

media/css/firefox/accounts.scss
```diff
@@ -0,0 +1,326 @@
...
+// Firefox Accounts for container
+
+.fxa-cta {
+    background: #fff;
+    background: #fff;
+    border-radius: 6px;
+    margin: 0 auto;
+    max-width: 400px;
+    text-align: center;
+    padding-top: 40px;
```
jpetto: This makes the box visible even when the iframe is hidden at viewports < 1000px. Might need to hide this element as well?

Chrome screenshot:

screen shot 2018-02-20 at 11 20 36 am

jpetto: Might need some `padding-bottom` on Fx at < 1000px wide, as the button is flush against the bottom of the white box.

screen shot 2018-02-20 at 10 44 23 am

(Going between breakpoints results in a not great UX, but I believe this has been the case for quite some time due to how the iframe resizes itself. Probably still an edge case for now.)

media/css/firefox/accounts.scss
```diff
@@ -0,0 +1,326 @@
...
+        list-style-position: inside;
+
+        li {
+            margin-bottom: 10px;
+        }
+    }
+
+    @media #{$mq-desktop} {
+        @include background-size(350px, 153px);
+        background-position: top left;
```
jpetto: This should probably stay `top center` at/beyond `$mq-desktop` for states that do not show the form.

screen shot 2018-02-20 at 11 33 24 am

bedrock/settings/static_media.py
```diff
@@ -1133,7 +1132,6 @@
         'source_filenames': (
             'js/base/mozilla-fxa-iframe.js',
             'js/base/uitour-lib.js',
-            'js/firefox/sync-animation.js',
```
jpetto: Looks like this might be the end of this lib? Can we remove it?

media/css/base/mozilla-fxa-iframe.scss
alexgibson: I'm just following existing conventions here really. The `.less` version of this import is used by both `/firstrun/` and `/accounts/`. I'm guessing we'll want to do the same once `/firstrun/` is ported to Pebbles.

media/css/firefox/accounts.scss
alexgibson: Bah nice catch (I added this top padding only after testing all the other states). Will fix

alexgibson: I think setting a `min-height` on the iframe itself here might help, and also fix the issue when switching between breakpoints.

media/css/firefox/accounts.scss
alexgibson: Good catch 👍

bedrock/settings/static_media.py
alexgibson: Yup, will remove!

alexgibson:
Thanks for the review @jpetto - I added another commit with some fixes.

media/css/base/mozilla-fxa-iframe.scss
jpetto: 👍

alexgibson:
@jpetto this has now been approved by creative. Squashed and ready to merge!

PR #5463, opened 2018-02-20
Bug 1438302: Move /firefox/new/?scene=2 to /firefox/download/thanks/

pmac:
Having the page from which Firefox is actually downloaded be a "real" url and not a parameterized version of another URL will be better for a number of reasons, not least of which is that it will enable us to host a static version of the site that allows downloads.

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -13,6 +13,7 @@ from urlparse import urlparse
 from django.conf import settings
 from django.http import Http404, HttpResponsePermanentRedirect
 from django.utils.cache import patch_response_headers
+from django.utils.encoding import force_text
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_POST, require_GET
 from django.views.generic.base import TemplateView
@@ -469,6 +470,21 @@ class TrackingProtectionTourView(l10n_utils.LangFilesMixin, TemplateView):
     template_name = 'firefox/tracking-protection-tour.html'
 
 
+def download_thanks(request):
+    experience = request.GET.get('xv', None)
+    locale = l10n_utils.get_locale(request)
+
+    # `wait-face`, `reggiewatts` variations are currently localized for both en-US and de locales.
+    if lang_file_is_active('firefox/new/wait-face', locale) and experience == 'waitface':
+        template = 'firefox/new/wait-face/scene2.html'
+    elif lang_file_is_active('firefox/new/reggiewatts', locale) and experience == 'reggiewatts':
+        template = 'firefox/new/reggie-watts/scene2.html'
+    else:
+        template = 'firefox/new/scene2.html'
+
+    return l10n_utils.render(request, template)
+
+
 def new(request):
     # Remove legacy query parameters (Bug 1236791)
     if request.GET.get('product', None) or request.GET.get('os', None):
@@ -484,13 +500,12 @@ def new(request):
         variant = None
 
     if scene == '2':
-        # `wait-face`, `reggiewatts` variations are currently localized for both en-US and de locales.
-        if lang_file_is_active('firefox/new/wait-face', locale) and experience == 'waitface':
-            template = 'firefox/new/wait-face/scene2.html'
-        elif lang_file_is_active('firefox/new/reggiewatts', locale) and experience == 'reggiewatts':
-            template = 'firefox/new/reggie-watts/scene2.html'
-        else:
-            template = 'firefox/new/scene2.html'
+        # send to new permanent scene2 URL (bug 1438302)
+        thanks_url = reverse('firefox.download.thanks')
+        query_string = request.META.get('QUERY_STRING', '')
+        if query_string:
+            done_url = '?'.join([thanks_url, force_text(query_string, errors='ignore')])
+        return HttpResponsePermanentRedirect(done_url)
     # if no/incorrect scene specified, show scene 1
     else:
         if lang_file_is_active('firefox/new/wait-face', locale) and experience == 'waitface':
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -30,7 +30,7 @@ class _ProductDetails(ProductDetails):
 
 
 class FirefoxDesktop(_ProductDetails):
-    download_base_url_transition = '/firefox/new/?scene=2'
+    download_base_url_transition = '/firefox/download/thanks/'
 
     # Human-readable platform names
     platform_labels = OrderedDict([
--- a/bedrock/firefox/redirects.py
+++ b/bedrock/firefox/redirects.py
@@ -69,6 +69,9 @@ redirectpatterns = (
     # bug 878871
     redirect(r'^firefoxos', '/firefox/os/'),
 
+    # bug 1438302
+    no_redirect(r'^firefox/download/thanks/$'),
+
     # Bug 1006616
     redirect(r'^download/?$', 'firefox.new'),
 
--- a/bedrock/firefox/templates/firefox/new/scene2.html
+++ b/bedrock/firefox/templates/firefox/new/scene2.html
@@ -6,8 +6,12 @@
 
 {% extends "firefox/new/base.html" %}
 
-{# "noindex" pages should not have the canonical or hreflang tags: bug 1442331 #}
-{% block canonical_urls %}<meta name="robots" content="noindex,follow">{% endblock %}
+{% block canonical_urls %}
+  {# the "scene 2" page should set canonical to the /firefox/new/ page to give that
+     page the SEO weight of links directly to "scene 2".
+     see https://github.com/mozilla/bedrock/pull/5463 #}
+  <link rel="canonical" href="{{ settings.CANONICAL_URL }}/{{ LANG }}/firefox/new/">
+{% endblock %}
 
 {% block optimizely %}
   {% if switch('firefox-new-scene2-optimizely', ['en-US']) %}
--- a/bedrock/firefox/tests/test_helpers.py
+++ b/bedrock/firefox/tests/test_helpers.py
@@ -97,7 +97,7 @@ class TestDownloadButtons(TestCase):
         for link in links[1:5]:
             link = pq(link)
             href = link.attr('href')
-            eq_(href, '/fr/firefox/new/?scene=2')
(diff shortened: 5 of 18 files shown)


hoosteeno:
This is a great change. I don't think we should get any more committed to the (meaningless) "/new" in our download URL. I suggest /firefox/download/thanks/ for the page we show when the download has begun and we are thanking the user for downloading, with an eventual goal of renaming /firefox/new/ to /firefox/download/.

pmac:
> I suggest /firefox/download/thanks/ for the page we show when the download has begun and we are thanking the user for downloading, with an eventual goal of renaming /firefox/new/ to /firefox/download/.

I would love to eventually reclaim the `/firefox/download/` URL. I agree that this PR isn't the place to do that, but we should look at doing it soon.

pmac:
Great feedback @alexgibson! I agree with pretty much everything. It is indeed a BIG change, but it is one we've wanted to make I believe. I just now see a bit more impetus to go ahead and get it done. It does make sense that it would be confusing for `/new/` to be indexable but `/new/?scene=2` to say it's not. I also agree that getting this on a demo and tested is key. I'll do that today.

I do like @hoosteeno's suggestion of making the URL something that will be good in the final form if we are indeed going to change `/new/` to `/download/`.

rraue:
Btw. besides the mail I send, if there is to much risk in moving the url of scene2, the most important point from Seo is to remove the canonical and hreflang on this noindex page. If this is the problem, it would be solved, if moved or not.

pmac:
> Btw. besides the mail I send, if there is to much risk in moving the url of scene2, the most important point from Seo is to remove the canonical and hreflang on this noindex page. If this is the problem, it would be solved, if moved or not.

Interesting. I was under the impression that it was weird for the search bots if one version of a page didn't have "noindex" and one did. That is `/firefox/new/` does not have noindex, and `/firefox/new/?scene=2` does. Woudldn't those be considered the same page and only versions of the same page? I thought a `noindex` on one would affect the other if they're at the same base URL.

jpetto:
I'm in agreement with Alex & Justin above. I'd like to get this on a demo and test **very** thoroughly. I also think we should do better than `/firefox/new/done/`. I personally like `/firefox/download/thanks/`, but I could be persuaded otherwise.

pmac:
I've updated the PR to the new `/firefox/download/thanks/` URL. That seems to be everyone's favorite so far. It's also deploying to [my demo](https://bedrock-demo-pmac.us-west.moz.works/en-US/) and [running tests](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/demo%2fpmac/50/pipeline/) as I type. Should be up shortly.

pmac:
I also addressed the JS fixes Alex brought up for the scen1.js files.

pmac:
[Demo](https://bedrock-demo-pmac.us-west.moz.works/en-US/firefox/new/?xv=reggiewatts) is up.

alexgibson:
> Btw. besides the mail I send, if there is to much risk in moving the url of scene2, the most important point from Seo is to remove the canonical and hreflang on this noindex page. If this is the problem, it would be solved, if moved or not.

Should we split out removing the canonical and hreflang URLs from scene2 into a separate pull request? It seems like that may be a much smaller and safer change to make, which we could then deploy independently of this larger movement of things?

rraue:
I would love that Alex, as I really would like to get rid of this language bug in the serps, but am still not totally sure about this url change, like you people.

pmac:
> Should we split out removing the canonical and hreflang URLs from scene2 into a separate pull request? It seems like that may be a much smaller and safer change to make, which we could then deploy independently of this larger movement of things?

That's easy enough. I can throw together that PR pretty quickly. I'll do that.

pmac:
Filed #5470 for the hreflang and canonical thing

pmac:
Was just talking with Jon and now that the removal of hreflang tags from scene2 is done and this is no longer as critical I wanted to talk about possibly including in this change the move of `/firefox/new/` itself to `/firefox/download/`. I'm just not sure of all of the pros and cons from both technical and SEO standpoints. Opinions anyone?

alexgibson:
> Opinions anyone?

No objections from me, it seems like the next most logical change. It would be good to get @rraue's thoughts here though.

rraue:
I copy my Email on this topic here:

Its a tough question. I will try to be as precise as possible, because a lot of principles are here interfering, so it will be a bit more to read:

_Tl:dr: I would not move /new/ right now and would integrate it if so in a possible new informational infrastructure, if we will need it. Could be the outcome of Jennifers work, I guess_

- In general a 301 transfer since around a year all signals from the referring url. Before google changed that a 301 lost around 30% of its power. So changing urls got a lot safer. But still there are three remaining risks: 

- A well established document has more signals (technical ones like speed, change history, also but as well user signals like ctr, bounce rate aso) then a new one and in a lot of serps (search engine result pages) Google is playing “never change a running system”. There are studies saying that 75% of all top rankings are older then a year. A new url is a new document for Google, because of the 301 it gets it but will remain “sceptic” and only transfer over time the old signals. So we could see for some time (if we do everything right it should not be more then a couple of weeks) worse rankings even for the big keywords, because especially seeps with brand keywords are higher protected then just informational serps. We risk for some time to have less downloads.

- It could be, that we rank for a lot keywords, where we are equal strong then competitors and are on the better position because we have been longer there. The system for new documents is, that it needs to be better then the existing one. Being the same good is not good enough. There is the risk of losing some keywords like that. I don’t know the page right now good enough to predict the reaction. I would not assume the risk very hight.

- I am not sure if in the transition most of the rankings will jump to /firefox/ from of /new/ and I have no clue if this will jump back then. It should but I wouldn’t be to sure abut it.

- Google in general advises to not change urls, if they are well established and there is no need in the informational structure. But Google underestimates in my experience itself the network effects of a good informational structures both on domain level and more important in the internal workflow of organisations. But with one url I would go with Google and not change it, if there is no need.

- I wouldn’t change the url now, while we have this problems. In general I still think we should think about  a really good midterm plan to merge /firefox/, /firefox/new/ and /firefox/new/?scene=2. Its not really understandable what are the differences, not for users, not for crawlers. I get that a full page with all that functionality would be creating a monster, but it would make so much things easier as we decide who sees which button when and not the search engine anymore. But even if we do not do it, we should first have a informational plan for the whole domain, before we move important things. I really think, we should look here at our competitors and do some more ux and convertion experiments to find a future safe setup.

- But I guess the ctr in the serps would be better for a file called /download/ or just /firefox/.

- We could implement a breadcrumb in the footer and as js+ld in the head to get rid of the new in a lot of serp cases. On mobile Google usually shows the breadcrumb instead of the url. There we could call the new page download instead of new. Could work.

pmac:
Sounds like this is a no-go for now. Thanks all.

pmac:
I'd like to revisit this. Regardless of whether we ever move `/firefox/new/` I would like to move the page to which every button points to be a real page. This has recently become more important because we'd like to generate a fully static version of the site to use as a backup if bedrock were to be attacked in a way that takes it down. If we had an updated s3 bucket with a static version of the site waiting we could stay up well enough to at least provide info and allow people to continue to download firefox. However, because the actual download happens on `/firefox/new/?scene=2` we can't do that with a fully static hosted site as that would simply return `/firefox/new/` again. If we merge this then we could have `/firefox/download/thanks/` that would work great even when the site was static.

I'll rebase this PR and get it ready, but in the interim please let me know if you have objections to this plan.

bedrock/firefox/views.py
```diff
@@ -484,13 +500,12 @@ def new(request):
...
-            template = 'firefox/new/wait-face/scene2.html'
-        elif lang_file_is_active('firefox/new/reggiewatts', locale) and experience == 'reggiewatts':
-            template = 'firefox/new/reggie-watts/scene2.html'
-        else:
-            template = 'firefox/new/scene2.html'
+        # send to new permanent scene2 URL (bug 1438302)
+        thanks_url = reverse('firefox.download.thanks')
+        query_string = request.META.get('QUERY_STRING', '')
+        if query_string:
+            done_url = '?'.join([thanks_url, force_text(query_string, errors='ignore')])
```
jpetto: Any reason to keep the `scene=2` query param around? Maybe for keeping track of how many folks are still hitting the old URL?

pmac: I thought maybe it'd be useful for that, yes, but also I didn't want to go to the trouble to remove it since it's not hurting anything. I'm happy to though if we'd prefer it that way.

rraue:
I agree with pmac. Scene2 is ruled out as the cause for our language-serp problem, all canonicals work and we can do the move now. I even would like to test this page being indexed but with a canonical pointing to to /firefox/new/. We know now that our canonicals have been buggy, this is fixed, and I would like to have all internal links counting. 

It would be good to merge this together. Shall I file a bug for the test?

full60mm:
Yup never mind i obviously have a lot to learn! Smh ha thanx for the support guys/gals highly appreciate it!

rraue:
After we changed our canonical setup I am pretty sure, that we can reindex scene2 again without showing up in the search results, like it did last time. But we should be very careful and take it back when its otherwise. We had issues with that before and scene2 ranked instead of /new/. I am pretty sure this will not happen, but if so we need to take back the changes I recommend now to be safe again.

This needs to be done the same time this is pushed online to https://github.com/mozilla/bedrock/blob/697b5a537d75f45873b302a3777de00dbbc2cdaf/bedrock/firefox/templates/firefox/new/scene2.html

It should be indexable again, so we need to remove the noindex, follow
therefore it gets a canonical agin, this should be to /firefox/new/ and not to itself
The page must not have any hreflang 

I will then check in Analytics and the Search Console if the new page will get search traffic, if so, we need to switch back to the actual template with noindex, no canonical and no hreflang.

But if it gets indexed but the canonical excepted, all internal links will count for /firefox/new/ and that would improve our architecture a lot.

This needs to be done:
delete

pmac:
I've updated the `scene2.html` template (the main one, not the campaign versions) to have a canonical tag pointing to the current locale version of `/firefox/new/`. This should mean that all direct links to `/firefox/download/thanks/` will give their SEO juice to `/firefox/new/` and not show up in search results itself.

pmac:
If we're good with this I just need an r+ and merge :)

PR #5466, opened 2018-02-21
Bug 1438302: x-default hreflang tag set to locale-free redirector URL

pmac:


rraue:
I commented in bugzilla and I would not change this. We use the x-default right.

PR #5469, opened 2018-02-22
Add translations back to english Tecnology page

pmac:
Since en-* locales have a separate template it wasn't picking up the list of translations, and so it appeared from the English page that it wasn't translated. This also meant that the en-GB version was redirecting to the en-US since that locale isn't active for the "mozorg/technology-en" langfile.


pmac:
Something I found while investigating some SEO strangeness. This has manifested in the `<link rel="alternative"...` tags on the non-en pages to be wrong about the en pages.

PR #5471, opened 2018-02-23
Improve healthz and add readiness URLs

pmac:
Also add django-allow-cidr


pmac:
I'll wait for a bit to merge this as IIRC @jpetto had a question about this last week.

pmac:
See https://github.com/mozmeao/infra/issues/735 for details on why this is necessary.

PR #5472, opened 2018-02-26
[fix bug 1438633] Add /whatsnew page for Firefox 59 with FxA CTA

alexgibson:
## Description
- Adds `firefox/whatsnew/59.0/` page featuring two CTAs:
  1. Firefox Accounts form for users not already signed into Sync.
  2. Firefox Mobile QRCode promo for users already signed into Sync.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1438633

## Testing
Demo: https://www-demo3.allizom.org/en-US/firefox/59.0/whatsnew/

To test the page functionality, you'll need to set up a [test profile](http://bedrock.readthedocs.io/en/latest/firefox-accounts.html#demo-server-testing) and also enable [uitour](http://bedrock.readthedocs.io/en/latest/uitour.html) for the demo server.

- [ ] Test signup works as expected
- [ ] Test QR Code is displayed as expected for users already signed in.
- [ ] Test responsiveness
- [ ] Test RTL.


bedrock/firefox/templates/firefox/whatsnew/whatsnew-fxa.html
```diff
@@ -0,0 +1,56 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% add_lang_files "firefox/new/quantum" %}
```
flodolo: This isn't going to work. For example: you're using `Take Firefox with You`, which is behind a tag in this lang file. Like this, we would end up enabling the page and showing the string in English for some locales.

We should use a separate lang file, and import the strings (`firefox/whatsnew_59`?)

flodolo: Let’s use `firefox/whatsnew_59`. Not only there's the tag/activation problem, but also a string that comes from a completely different file.

alexgibson:
@flodolo - updated to use `firefox/whatsnew_59`. Note that the string `Your Firefox is up to date.` is not used here.

Also, please hold on creating this new lang file just for now. We need to confirm that these 41 locales are going to be enough for the team requesting this page (this was why I hadn't pinged yourself or Peiying about this just yet).

flodolo:
> Note that the string Your Firefox is up to date. is not used here.

Ah, missed the piece of logic there.

> Also, please hold on creating this new lang file just for now. We need to confirm that these 41 locales are going to be enough for the team requesting this page (this was why I hadn't pinged yourself or Peiying about this just yet).

I've already created it locally, since that was the only safe way to get the list of locales. There's a meeting schedule tonight, hopefully I'll get some clarity.

flodolo:
I've just merged the [files on www.mozilla.org](https://github.com/mozilla-l10n/langchecker/pull/775). Eric is going to get approval on the screenshots, then we'll need this in production to allow testing for QA.

alexgibson:
Still waiting on confirmation that all is good in terms of GA and tracking conversions, but otherwise this is good for code review.

media/js/firefox/whatsnew/whatsnew-fxa.js
```diff
@@ -0,0 +1,35 @@
...
+            });
+        });
+    }
+
+    function showMobilePromo() {
+        document.querySelector('.main-content').classList.add('signed-in');
+    }
+
+    Mozilla.UITour.getConfiguration('sync', function(config) {
+        clearTimeout(syncTimeout); // clear UITour fallback timeout.
```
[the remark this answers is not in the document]
alexgibson: Good suggestion 👍

alexgibson:
@jpetto thanks for the review, pushed some fixes.

alexgibson:
@jpetto I updated the QRCode in this pull request. If all looks good, we're all set to merge.

PR #5473, opened 2018-02-26
[fix bug 1436973] Decommission legacy /firefox/unsupported/ URLs

alexgibson:
## Description
- Remove legacy `/firefox/unsupported/` URLs and redirect to `/firefox/unsupported-systems/`

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1436973


bedrock/firefox/redirects.py
```diff
@@ -578,4 +578,9 @@ def firefox_channel(*args, **kwargs):
 
     # bug 1419244
     redirect(r'^firefox/mobile-download(/.*)?', 'firefox.mobile'),
+
+    # bug 960651, 1436973
+    redirect(r'(firefox|mobile)/([^/]+)/details(/|/.+\.html)?$', 'firefox.unsupported-systems',
+             locale_prefix=False),
+    redirect(r'^firefox/unsupported(/.*)', 'firefox.unsupported-systems')
```
pmac: I'm not sure the purpose of the parenthetical in this. I think the following should be sufficient:

```python
redirect(r'^firefox/unsupported/', 'firefox.unsupported-systems'),
```

It's a bit subtle in its exclusion of `/firefox/unsupported-systems/` and inclusion of all `/firefox/unsuported/*`, but it is correct and I think is fine. Up to you though if you want to change it as I don't think this is wrong per se.

alexgibson:
Thanks for the review @pmac - updated with your suggestion 👍

PR #5477, opened 2018-02-28
Bug 1388914 - Delete thunderbird content

stephaniehobson:
## Description

- Removed thunderbird app
  - Removed image, CSS, and JS assets for thunderbird app
  - Remove associated tests
- Remove /projects/calendar and /projects/calendar/holidays
  - Remove associated image, CSS, and JS assets
- Remove release notes
  - Remove associated tests
- Moved redirects from thunderbird app to bedrock redirects
  - Updated to point to thunderbird.net where appropriate
- Added redirects to thunderbird.net

## Issue / Bugzilla link

[Bug 1388914 - Delete thunderbird content](https://bugzilla.mozilla.org/show_bug.cgi?id=1388914)

## Testing

- thunderbird app files and assets are gone
- thunderbird assets are not gone from legal or privacy areas of site
- redirects send user to appropriate pages


bedrock/releasenotes/models.py
```diff
@@ -149,7 +149,7 @@ class ProductRelease(models.Model):
     CHANNELS = ('Nightly', 'Aurora', 'Beta', 'Release', 'ESR')
     PRODUCTS = ('Firefox', 'Firefox for Android',
                 'Firefox Extended Support Release', 'Firefox OS',
-                'Thunderbird', 'Firefox for iOS')
+                'Firefox for iOS')
```
stephaniehobson: I took Thunderbird out of the products list. Is that okay?

bedrock/settings/base.py
```diff
@@ -511,7 +506,6 @@ def set_whitenoise_headers(headers, path, url):
     '/contribute/',
     '/about/',
     '/contact/',
-    '/thunderbird/',
```
stephaniehobson: There's a comment saying this list must match a list at CloudFlare. What do we need to do to make that happen?

bedrock/releasenotes/models.py
kewisch: I'm not quite sure where this is displayed, but if this removes it from some mozilla.org page headers, I'd prefer if we could keep it in and have it link to Thunderbird.net. @Sancus probaby has a better idea what this is for.

Sancus: afaik, Thunderbird is not linked to by any other part of the mozilla.org site these days.

kewisch: Very well then, if this is the last link then it seems reasonable to remove.

flodolo:
We have several pages localized for Thunderbird. Are you going to import existing content from mozilla.org to the new place?

Sancus:
@flodolo 
All content (except for the orphaned email-providers page, which we're deleting) is duplicated at https://www.thunderbird.net including localized stuff, at the same paths. I am using the mozilla.org .lang files currently but I have a port to gettext in progress that I would like to get up on pontoon.mozilla.org as an independent project.

flodolo:
> All content (except for the orphaned email-providers page, which we're deleting) is duplicated at https://www.thunderbird.net including localized stuff, at the same paths. I am using the mozilla.org .lang files currently but I have a port to gettext in progress that I would like to get up on pontoon.mozilla.org as an independent project.

Thanks, that makes sense.

bedrock/releasenotes/models.py
stephaniehobson: I believe this is connected to the database some how. Probably needs an okay from @pmac.

bedrock/settings/base.py
stephaniehobson: This is probably a question for @jgmize

bedrock/releasenotes/models.py
pmac: Please don't remove this. It has to do with importing releases from our release notes server which would fail since this is used as a list of allowed values for a DB field. We'll have to remove this after Thunderbird notes are removed from said server (Nucleus).

bedrock/settings/base.py
pmac: That just means that the /thunderbird/ without locales (so the redirects) URLs are allowed to be cached with a "Vary: Accept-Language" header. It should be safe to remove this and we can remove the /thunderbird config from CloudFlare later.

bedrock/mozorg/redirects.py
```diff
@@ -146,8 +146,6 @@ def decider(request, **kwargs):
     redirect(r'^foundation/identity-guidelines/index.html', 'styleguide.home'),
     redirect(r'^foundation/identity-guidelines/mozilla-foundation.html',
              'styleguide.identity.mozilla-branding'),
-    redirect(r'^foundation/identity-guidelines/thunderbird.html',
```
alexgibson: Removing this redirect allows `foundation/identity-guidelines/thunderbird.html` to 404. Is this intentional?

bedrock/mozorg/templates/mozorg/projects/mozilla-based.html
```diff
@@ -28,7 +28,7 @@ <h3><a href="http://www.seamonkey-project.org/"><img src="{{ static('img/project
   <p>{{ _('The all-in-one application formerly known as the “Mozilla Application Suite”, containing a web browser, a mail and newsgroups client, an HTML editor, web development tools, and an IRC chat client.') }}</p>
   </li>
   <li>
-  <h3><a href="/projects/calendar/"><img src="{{ static('img/projects/calendar/calendar-logo.png') }}" width="64" height="64" alt="" /><span>{{ _('Lightning') }}</span></a></h3>
+  <h3><a href="https://www.thunderbird.net/calendar/"><img src="{{ static('img/projects/calendar/calendar-logo.png') }}" width="64" height="64" alt="" /><span>{{ _('Lightning') }}</span></a></h3>
```
alexgibson: Just noting that this URL is still 404.

alexgibson: It looks like the logo here `projects/calendar/calendar-logo.png` is still being used, but it's removed in this PR below.

bedrock/mozorg/templates/mozorg/projects/calendar.html
```diff
@@ -1,90 +0,0 @@
-{# This Source Code Form is subject to the terms of the Mozilla Public
```
alexgibson: The URL for this page and the one below can be removed:

https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/urls.py#L32-L34

It looks like we can also remove the view for holiday calendars:

https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/views.py#L155-L172

And the calendar data:

https://github.com/mozilla/bedrock/tree/master/media/caldata

bedrock/releasenotes/tests/test_base.py
```diff
@@ -285,16 +254,6 @@ def test_relnotes_index_firefox(self, render_mock):
         eq_(releases[6][1]['minor'],
             ['31.1.0', '31.1.1', '31.2.0', '31.3.0', '31.4.0', '31.5.0'])
 
-    @patch('bedrock.releasenotes.views.thunderbird_desktop', thunderbird_desktop)
-    def test_relnotes_index_thunderbird(self):
-        with self.activate('en-US'):
-            response = self.client.get(reverse('thunderbird.releases.index'))
-        doc = pq(response.content)
```
alexgibson: It looks like `pq` can be removed from the imports in this file:

```
F401 'pq' imported but unused
```

bedrock/redirects/redirects.py
```diff
@@ -2004,6 +2006,53 @@
     redirect('^status/minutes\.html$', 'https://wiki.mozilla.org/WeeklyUpdates'),
     redirect('^store$', 'https://store.mozilla.org'),
     redirect('^testdrivers$', 'http://wiki.mozilla.org/B2G_Testdrivers_Program'),
+    # bug 1124038
+    redirect(r'^thunderbird/organizations/(?:all-esr\.html|faq/?)$', 'https://www.thunderbird.net/organizations/'),
```
alexgibson: We need to update the [tests for these redirects](https://github.com/mozilla/bedrock/blob/master/tests/redirects/map_globalconf.py). We should probably also add new tests for any new redirects added here, so we can be confident they're working as expected.

You can find info on how to run the redirect tests locally here: http://bedrock.readthedocs.io/en/latest/redirects.html#testing-redirects

bedrock/thunderbird/redirects.py
```diff
@@ -1,58 +0,0 @@
-from bedrock.redirects.util import redirect
-
-
-redirectpatterns = (
-    # bug 1124038
-    redirect(r'^thunderbird/organizations/(?:all-esr\.html|faq/?)$', 'thunderbird.organizations'),
-
-    # bug 1124042
-    redirect(r'^thunderbird/features/email_providers\.html$', 'thunderbird.email-providers'),
```
Sancus: Should be redirected to https://www.thunderbird.net/en-US/email-providers/ as it turns out it is used in Thunderbird.

bedrock/settings/base.py
stephaniehobson: 👍

bedrock/mozorg/redirects.py
stephaniehobson: Initially this PR included removing the style guide as well but I scaled it back. I guess this is an artifact.

But we are planning to take the style guide out, with it gone I'm not sure there's an appropriate place to redirect this. Ideas?

bedrock/mozorg/templates/mozorg/projects/mozilla-based.html
stephaniehobson: I've moved it into the mozilla-based media assets folder and updated the reference.

Thank you for the very through review!

bedrock/thunderbird/redirects.py
stephaniehobson: Added 👍

bedrock/mozorg/redirects.py
stephaniehobson: I'll add it back and let the style guide PR deal with it.

bedrock/mozorg/templates/mozorg/projects/calendar.html
stephaniehobson: ✅

bedrock/releasenotes/tests/test_base.py
stephaniehobson: ✅

bedrock/mozorg/views.py
```diff
@@ -152,26 +152,6 @@ def contribute_studentambassadors_landing(request):
                              {'tweets': tweets})
 
 
-def holiday_calendars(request, template='mozorg/projects/holiday-calendars.html'):
-    """Generate the table of holiday calendars from JSON."""
-    calendars = []
-    json_file = find_static('caldata/calendars.json')
```
alexgibson: Looks like there's some cleanup that can be done after removing this view:

```
(venv) agibson-42230:bedrock agibson$ flake8 bedrock lib tests
bedrock/mozorg/views.py:5:1: F401 'json' imported but unused
bedrock/mozorg/views.py:8:1: F401 'find_static' imported but unused
bedrock/mozorg/views.py:21:1: F401 'static' imported but unused
3       F401 'find_static' imported but unused
```

stephaniehobson:
Added and updated the redirect tests to match expectations.

Two tests are legitimate failures, I'll discuss them on the bug.

bedrock/redirects/redirects.py
```diff
@@ -1644,6 +1644,10 @@
     redirect('^performance/leak-brownbag\.html$',
              'http://wiki.mozilla.org/Performance:Leak_Tools'),
     redirect('^privacy-policy(\.html|/.*)?$', '/privacy/websites/'),
+    redirect('^projects/calendar/$', 'https://www.thunderbird.net/calendar/'),
+    redirect('^projects/calendar/holidays/$', 'https://www.thunderbird.net/calendar/holidays/'),
+    # bug 885799, 952429
+    redirect(r'^projects/calendar/holidays\.html$', 'https://www.thunderbird.net/calendar/holidays/'),
```
pmac: Might also be a good idea to redirect the /media urls for the calendar files as well so that we don't get a bunch of bugs filed for missing ical files. Though we may also not want to continue to being in the business of handling all of that traffic from Lightening. Dunno. Would be something like the following if we wanted to do it:

```python
redirect(r'media/caldata/(?P<filename>.+\.ics)$', 'https://www.thunderbird.net/media/caldata/{filename}'),
```

Sancus: I don't really use Lightning. Does it directly link to these files in some way? If so we should probably change the URLs at the source? @kewisch

alexgibson:
If we're removing the calendar data, then it looks like we can also remove this script:

https://github.com/mozilla/bedrock/blob/master/scripts/check_calendars.py

And also the command to run it:

https://github.com/mozilla/bedrock/blob/master/bin/run-tests.sh#L4

bedrock/redirects/redirects.py
kewisch: People directly use these links for calendars. I could add an in-product transition mechanism, but this would need time to propagate. Are there stats on how often these are requested?

kewisch: Unfortunately not, at least not for ICS feeds. We have a such thing for caldav calendars and what I would do in product is add that same check for ics feeds.

Sancus:
Does that check calendars script just exist to make sure people don't commit malformed ics files or something? (wondering if I need to copy it over and run it for thunderbird.net)

pmac:
> Does that check calendars script just exist to make sure people don't commit malformed ics files or something? (wondering if I need to copy it over and run it for thunderbird.net)

That's it exactly. It's there to help us review PRs that update those files. It's just a sanity check for the format of the ICS files. I think taking it with you is a great idea.

bedrock/redirects/redirects.py
pmac: So should we continue hosting the calendar files for a while so that we don't break people's calendars until product changes are made to support redirection?

pmac: I'm cool with that. @alexgibson @stephaniehobson is that okay with you? We'd just leave those .ics files where they are for now but remove everything else, and the new location for the holiday calendars page and files will be the new thunderbird site. 

@kewisch I'll see what I can find out about the traffic. I'll file an issue in our infra repo and CC you there.

kewisch: Thanks! In reading that ticket I realized I wasn't quite clear in what Lightning can do, I apologize. The code is very well capable to follow 3xx redirects, but it will not "remember" the redirect and continue to request the original url on the next refresh. 

The product change I would like to do is to allow the user to update the url on a 3xx redirect, so that eventually these calendar file redirects will no longer be necessary.

pmac: Ah! That's cool. So, what should we do here in your opinion? Keep the files for now or go ahead with the redirects? The redirects won't bother us, but I understand if you'd like for us to complete the traffic estimate before we start redirecting to you.

pmac: I updated the ticket to try to explain the situation more accurately.

Sancus: Cloudflare caches the .ics files for us so I don't think the traffic requirements are that important. Nice to know, of course, because it will be difficult for us to obtain similar stats since we don't currently have any way to access Cloudflare logs.

pmac: Are you using cloudflare for thunderbird.net as well? And if so why would you not have access to the logs?

Sancus: Cloudflare doesn't provide log access unless you have an Enterprise account, afaik. That's what their support database says anyway. (https://support.cloudflare.com/hc/en-us/articles/216672448-Enterprise-Log-Share-Logpull-REST-API)

pmac: Ah. Didn't realize that.

Sancus: Did we ever get the CDN logs? I don't think this should block this PR being merged btw.

pmac: I filed this but haven't heard anything https://github.com/mozmeao/infra/issues/753

alexgibson:
@stephaniehobson just checking in here as I lost track of the convo's. Is this ready for another round of review, or are we waiting on other pieces?

stephaniehobson:
Added back the .ics files and associated code (kept it in a separate PR for review-ablilty but should be squashed before merge). Ready for re-review, but still hold off on merge so the Thunrderbird team can co-ordinate the switch on their side.

Sancus:
We are ready to go, just need a date for merge so that I can make sure everyone is aware of it in advance.

tests/redirects/map_globalconf.py
```diff
@@ -235,20 +235,27 @@
     url_test('/{m,{firefox/,}mobile}/faq/', 'https://support.mozilla.org/products/mobile'),
 
     # bug 885799, 952429
-    url_test('/projects/calendar/holidays.html', '/projects/calendar/holidays/'),
-    url_test('/en-US/projects/calendar/random/stuff/', '/projects/calendar/'),
-    # redirects don't catch real urls
-    url_test('/en-US/projects/calendar/', status_code=requests.codes.ok),
-    url_test('/en-US/projects/calendar/holidays/', status_code=requests.codes.ok),
+    url_test('/projects/calendar/holidays.html', 'https://www.thunderbird.net/en-US/calendar/holidays/', follow_redirects=True),
```
alexgibson: These redirect tests would be a bit faster (and less prone to failure via external factors) if we removed `follow_redirects=True` e.g.:

```
url_test('/projects/calendar/holidays.html', 'https://www.thunderbird.net/calendar/holidays/'),
```

bedrock/redirects/redirects.py
```diff
@@ -2005,6 +2009,59 @@
...
+    redirect(r'^thunderbird/about/(?P<page>board|press|staff)/',
+             'http://website-archive.mozilla.org/www.mozilla.org/thunderbird'
+             '/thunderbird/about/{page}/'),
+    # bug 1388914
+    redirect(r'^thunderbird/all/$', 'https://www.thunderbird.net/thunderbird/all/'),
+    redirect(r'^thunderbird/beta/(?P<path>.*)$', 'https://www.thunderbird.net/thunderbird/beta/{path}'),
+    redirect(r'^thunderbird/(?P<version>[^/]+)/releasenotes/$',
+             'https://www.thunderbird.net/thunderbird/{version}/releasenotes/'),
+    redirect(r'^thunderbird/(?P<version>[^/]+)/system-requirements/$',
+             'https://www.thunderbird.net/en-US/thunderbird/{version}/system-requirements/'),
```
alexgibson: This is likely from an old test, but we can probably remove the hard coded `/en-US/` here and thunderbird.net can manage the locale redirection itself?

alexgibson:
URL here that can be updated? https://github.com/mozilla/bedrock/blob/master/bedrock/security/templates/security/product-advisories.html#L24

alexgibson:
A couple of legacy test URLs here that can probably now be removed? https://github.com/mozilla/bedrock/blob/master/tests/redirects/test_urls.py#L72-L74

tests/redirects/map_globalconf.py
stephaniehobson: Happy to make the change. Can you explain to me why this won't make the test fail?

pmac: Notice he also changed the destination to remove the `en-US/` bit. Point is that we don't care what thunderbird.net does with the request, just that we're sending people to the proper beginning URL. So just test that bedrock is sending a 301 to the right URL and don't follow redirects. Not only is it slower to hit the destination, but it would also cause our tests to fail if that site were experiencing problems or changed their redirects.

stephaniehobson: Ah, okay, the test does not require a positive response to pass, it just looks that the request went to the correct place?

pmac: Right. These are redirect tests, so they just look at the response from bedrock to make sure it matches what we expect.

tests/redirects/map_globalconf.py
```diff
@@ -276,7 +283,7 @@
     # bug 1204579
     url_test('/thunderbird/2.0.0.0/eula/', '/about/legal/eula/thunderbird-2/'),
     url_test('/thunderbird/about/legal/', '/about/legal/terms/mozilla/'),
-    url_test('/thunderbird/download/', '/thunderbird/'),
+    url_test('/thunderbird/download/', 'https://www.thunderbird.net/en-US/', follow_redirects=True),
```
alexgibson: One more test where we can remove ``follow_redirects=True``

stephaniehobson:
Squashed 🦑

stephaniehobson:
Woot!! I'll discuss a launch plan in the bug.

flodolo:
We used to have a localize page
https://www-dev.allizom.org/thunderbird/release/start

That now redirects to a 404. Expected?

PR #5479, opened 2018-03-01
[fix bug 1442229] Exclude /firefox/tracking-protection/start/ from sitemap.xml

alexgibson:
## Description
- Removes `/firefox/tracking-protection/start/` from our XML sitemap.
- Also removes canonical/hreflang URLs from template, since this is a noindex page.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1442229

## Testing
- I wasn't sure how to test the generated sitemap locally?

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -294,6 +294,7 @@ NOINDEX_URLS = [
     r'^firefox/send-to-device-post',
     r'^firefox/feedback',
     r'^firefox/stub_attribution_code/',
+    r'^.+/tracking-protection/start/$',
     r'^.+/(firstrun|whatsnew)/$',
     r'^l10n_example/',
     r'^m/',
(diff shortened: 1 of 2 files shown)


rraue:
We should then also remove the comple hreflang links, as it is noindex.

bedrock/settings/base.py
```diff
@@ -294,6 +294,7 @@ def lazy_langs():
     r'^firefox/send-to-device-post',
     r'^firefox/feedback',
     r'^firefox/stub_attribution_code/',
+    r'^.+/tracking-protection/start/$',
```
pmac: I can't find the reason that this is made for `*/tracking-protection/start/` rather than explicitly for Firefox. Do we expect there to be a page like this for other sections of the site?

alexgibson: The URL can be either `/firefox/tracking-protection/start/` or something like `/firefox/58.0/tracking-protection/start/`

pmac: Ah! Cool. Thanks.

PR #5482, opened 2018-03-02
[fix bug 1442565] Missing languages in hreflang

alexgibson:
## Description
- Adds missing languages to `hreflang` for Google-juice.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1442565

## Testing
Demo: https://bedrock-demo-agibson.us-west.moz.works/en-US/firefox/new/

- [ ] Check for accidental copy pasta? 🍝

### DIFF
--- a/bedrock/base/templates/includes/canonical-url.html
+++ b/bedrock/base/templates/includes/canonical-url.html
@@ -12,6 +12,33 @@
         <link rel="alternate" hreflang="en" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="English">
         <link rel="alternate" hreflang="en-US" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="English (USA)">
         <link rel="alternate" hreflang="en-CA" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="English (Canada)">
+        {% elif code == 'es-ES' -%}
+        <link rel="alternate" hreflang="es" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="Español">
+        <link rel="alternate" hreflang="es-ES" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="{{ label|safe }}">
+        {% elif code == 'fy-NL' -%}
+        <link rel="alternate" hreflang="fy" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="Frysk">
+        <link rel="alternate" hreflang="fy-NL" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="{{ label|safe }}">
+        {% elif code == 'gu-IN' -%}
+        <link rel="alternate" hreflang="gu" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="ગુજરાતી">
+        <link rel="alternate" hreflang="gu-IN" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="{{ label|safe }}">
+        {% elif code == 'hi-IN' -%}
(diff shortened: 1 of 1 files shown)


bedrock/base/templates/includes/canonical-url.html
```diff
@@ -12,6 +12,33 @@
         <link rel="alternate" hreflang="en" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="English">
         <link rel="alternate" hreflang="en-US" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="English (USA)">
         <link rel="alternate" hreflang="en-CA" href="{{ settings.CANONICAL_URL + '/' + code + loop_canonical_path }}" title="English (Canada)">
+        {% elif code == 'es-ES' -%}
```
pmac: Only for `es-ES`? We support several `es-XX` locales. We should include `es` if any of them are active right? Perhaps we could improve the code list in the context to include all 2 letter locales?

alexgibson: Right, but (iiuc) the ask here is to provide Google with a hint to the default language (not locale) for Spanish, which should probably always be `es-ES` (same as we have already done for `en-US`). We're not doing this for every two letter locale i.e. `zh-CN` and `zh-CW`. Is that right, @rraue?

pmac: Right, but what I’m saying is that a page might have only `es-AR` enabled, and then `es` wouldn’t show up. It might be rare enough to not matter. But we could create a more official list of default locales for ones where we have more than one regional locale. This is probabally good enough for now, but we should keep in mind that it could result in these languages beeing missing if these specific regions aren’t activated.

alexgibson: Hmm, yeah - I don't know how likely that is - but I guess it could happen. If `es-ES` isn't enabled for a page however, would it be wise to still point Googlt to it? Or would it be better to leave it out?

pmac: I guess the question is really whether we want only a specific locale to be the default for a two letter code and not be there if that locale isn’t active, or to try to choose the best for a default from the list that is active?

alexgibson: Right, I guess that's a question for @rraue to answer.

rraue: You are both right. I assumed es-ES is always translated, if one of the other spanish locales is. That might be wrong and then we would create more missing confirmation links like in https://bugzilla.mozilla.org/show_bug.cgi?id=1442339. 

Of course es can be only present if es-ES exists. I took es-ES as the default for es because it has more pages translated then the other locales, but it might get us some issues. Not sure how to check that. 

The same of course is valid for the other added languages, the locale pointing to needs to exists so we link the language.

alexgibson: So what is the best recommended course of action here? Should we proceed as-is, or should we default to e.g. `es` being `es-ES` if available, else falling back to another `es` locale?

rraue: So for now, let us just include es when es-ES exists. Same logic for the other added languages. I need to understand our workflow for translations better to recomment a broader solution. This will at least solve the problem for our most important pages, as they exist in es-ES and we do not add any new issue.

So at least any spanish version for the firefox download ranks in spanish speaking countries not having a locale. We rank for "descargar firefox" with the english version of /new/ on no. 1 in Colombia, Peru, Ecuador, Venezuela and so on and this would be fixed.

pmac: We could go all out and check the lang files and see if e.g. es-MX is ever activated without es-ES, but my guess is that you’re probably right that that’s rare, and it will certainly not be the case on the really popular pages. I think we can go ahead with this and then improve later if this really does come up.

alexgibson: Thanks both, sgtm 🌮

PR #5489, opened 2018-03-06
Bug 1441735: Firefox Quantum for Enterprise beta sign up

stephaniehobson:
**l10n**: strings are not final

Can I get a sanity check on this please? 

I extended /firefox/new/wait-face from /firefox/enterprise/index.html which is in a different directory. It works but maybe it's bad practice?

I'm waiting on a new hero image.

## Description
- Edit wait-face template for use with a different CTA.
- Extend wait-face template for enterprise beta signup.

## Issue / Bugzilla link
[Bug 1441735](https://bugzilla.mozilla.org/show_bug.cgi?id=1441735)


bedrock/firefox/templates/firefox/enterprise/index.html
```diff
@@ -0,0 +1,50 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/new/wait-face/base.html" %}
```
flodolo: This is not going to work, because you're going to also inherit this line, and load a lang file that has nothing to do with this page
https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/new/wait-face/base.html#L5

flodolo:
Removing l10n label, this wasn't exposed to localization.

PR #5494, opened 2018-03-08
[fix bug 1442129] Opt-out confirmation page tweaks

alexgibson:
## Description
- Fixes subhead font-size being smaller than body copy on small screens.
- Use title case for for button.
- Add placeholder text to email field.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1442129

### DIFF
--- a/bedrock/newsletter/templates/newsletter/opt-out-confirmation.html
+++ b/bedrock/newsletter/templates/newsletter/opt-out-confirmation.html
@@ -50,9 +50,9 @@
               <div class="field-contents">
                 {{ form.email.errors }}
                 <label for="id_email">{{ _('Your email address:') }}</label>
-                {{ form.email }}
+                {{ field_with_attrs(form.email, placeholder=_('yourname@example.com'))|safe }}
               </div>
-              <button class="button button-hollow button-dark" type="submit">{{ _('Manage preferences') }}</button>
+              <button class="button button-hollow button-dark" type="submit">{{ _('Manage Preferences') }}</button>
             </div>
           </div>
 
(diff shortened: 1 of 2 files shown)


media/css/newsletter/newsletter-opt-out-confirmation.scss
```diff
@@ -33,9 +33,15 @@ header {
     }
 
     .tagline {
-        @include font-size-level4;
+        @include font-size(18px);
```
jpetto: If we need this to match `<body>` copy size, could we just drop these size declarations altogether? Wouldn't that fall back to match the rest of the copy size?

bedrock/newsletter/templates/newsletter/opt-out-confirmation.html
```diff
@@ -50,9 +50,9 @@ <h1>{{ self.page_title() }}</h1>
               <div class="field-contents">
                 {{ form.email.errors }}
                 <label for="id_email">{{ _('Your email address:') }}</label>
-                {{ form.email }}
+                {{ field_with_attrs(form.email, placeholder=_('yourname@example.com'))|safe }}
               </div>
-              <button class="button button-hollow button-dark" type="submit">{{ _('Manage preferences') }}</button>
+              <button class="button button-hollow button-dark" type="submit">{{ _('Manage Preferences') }}</button>
```
jpetto: Do we need to alert L10n here?

media/css/newsletter/newsletter-opt-out-confirmation.scss
alexgibson: Yup will remove 👍

bedrock/newsletter/templates/newsletter/opt-out-confirmation.html
alexgibson: This page isn't localized, so shouldn't impact l10n

alexgibson:
Thanks @jpetto updated

PR #5495, opened 2018-03-08
Fix Bug 1441818 - New advanced download page should not put Windows 32 & Linux 32 at the same level

kyoshino:
## Description

Reorganize the recommended and traditional builds on [`firefox/new/`](https://www.mozilla.org/en-US/firefox/new/) like #5432.

## Issue / Bugzilla link

[Bug 1441818 - New advanced download page should not put Windows 32 & Linux 32 at the same level](https://bugzilla.mozilla.org/show_bug.cgi?id=1441818)

## Testing

Open `/firefox/new/`, click on "Advanced install options & other platforms" and see the builds reordered like [this screenshot](https://screenshots.firefox.com/05SqX3GR1KuC1BxI/www-local.allizom.org).

### DIFF
--- a/media/css/firefox/new/_other-platforms.scss
+++ b/media/css/firefox/new/_other-platforms.scss
@@ -65,15 +65,29 @@ html[dir="rtl"] #other-platforms-modal-link {
         }
     }
 
+    .download-platform-lists {
+        margin: 20px 40px;
+    }
+
     .download-platform-list {
         list-style: none;
         margin: 20px auto;
         text-align: center;
 
+        &.recommended li {
+            font-weight: bold;
+        }
+
+        &.traditional {
+            border-width: 1px 0;
+            border-color: #D7D7DB;
+            border-style: solid;
+        }
+
         li {
             display: block;
             line-height: 2;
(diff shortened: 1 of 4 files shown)


pascalchevrel:
@alexgibson @pmac could one of you review this PR? Thanks

media/css/firefox/new/_other-platforms.scss
```diff
@@ -65,15 +65,33 @@ html[dir="rtl"] #other-platforms-modal-link {
...
+            font-weight: bold;
+        }
+
+        &.traditional {
+            border-width: 1px 0;
+            border-color: #D7D7DB;
+            border-style: solid;
+
+            li {
+                opacity: .7;
```
alexgibson: With the opacity set here this text fails a11y contrast checks. I'd suggest relying on just the bold font weight to highlight recommended vs traditional.

pascalchevrel: @kyoshino could you update your patch please? Thanks

kyoshino: Okay, removed the `opacity` property.

PR #5496, opened 2018-03-08
[fix bug 1441597] Update copy for FxA copy test.

jpetto:
## Description

Updates copy for one variation.


craigcook:
CircleCI is having issues at the moment but this is just a copy change, not too worried about it breaking the build :)

PR #5500, opened 2018-03-09
[fix bug 1399654] Firstrun - change skip button to gray

craigcook:
## Description
Changes the color of the skip button so it doesn't overshadow or compete with the signup button (which recently changed from gray to blue).

I opted for blue text and border instead of dark gray so it wouldn't look disabled, but it's still secondary to the more prominent signup button. Happy to adjust colors if someone has a better suggestion.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1399654

## Testing
https://bedrock-demo-craigcook.us-west.moz.works/firefox/58.0/firstrun/
Older versions, too: https://bedrock-demo-craigcook.us-west.moz.works/firefox/55.0/firstrun/

### DIFF
--- a/media/css/firefox/firstrun/firstrun-quantum.scss
+++ b/media/css/firefox/firstrun/firstrun-quantum.scss
@@ -191,19 +191,21 @@ body {
 #skip-button {
     -moz-appearance: none;
     @include font-size(14px);
-    background-color: #0996f8;
-    border-radius: 3px;
-    border: 1px solid #0670cc;
-    color: #fff;
+    background-color: #fcfcfc;
+    border-radius: 2px;
+    border: 1px solid #0a84ff;
+    color: #0a84ff;
     cursor: pointer;
     display: block;
     margin: 10px auto 0 auto;
     min-height: 24px;
     padding: 5px 10px;
+    transition: background-color 150ms, color 150ms, border-color 150ms;
 
     &:not([disabled]):hover {
-        background-color: #0670cc;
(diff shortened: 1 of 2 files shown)


alexgibson:
Code-wise this looks good to me, although maybe we should pass the updated button styles past someone else for approval?

The other question I thought about was if we're OK to change something like this mid-way through a set of copy/image variation tests. I'll defer to @hoosteeno here.

hoosteeno:
> The other question I thought about was if we're OK to change something like this mid-way through a set of copy/image variation tests. 

Would it change on any of the controls or variations? If not, then it won't confound our experiments. If so, it might confound our experiments. In either case, we should compare before/after non-launch-week WoW averages to learn whether it had any impact (or, alternatively, put it in an experiment).

alexgibson:
> Would it change on any of the controls or variations? If not, then it won't confound our experiments. If so, it might confound our experiments. In either case, we should compare before/after non-launch-week WoW averages to learn whether it had any impact (or, alternatively, put it in an experiment).

Yes, this change would show for all `/firstrun` pages, including experiment variations. I think we can likely add a *no-not-merge* label on this until the current experiments are finished, if that would be preferable. It's only a small change, and would hopefully only be benefitial to any variation of an experiment, but it could technically color the data somewhat.

craigcook:
If all other firstrun experiments have concluded this should be safe to merge.

PR #5501, opened 2018-03-09
Switch liveness and readiness views to django-watchman

pmac:
Not using the Watchman Dashboard at present since it requires Django templates and we only support Jinja for now. We could import the templates later if we deem it necessary, but the JSON output should be sufficient for the foreseeable future.


metadave:
nice!

PR #5504, opened 2018-03-12
Mark test_localized_download_links as xfail due to bug 1445077

jgmize:
## Description
test_localized_download_links is currently failing due to a legitimate 404 for a localized nightly build

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1445077

## Testing
This should mark the failing test as an expected failure.


jgmize:
For future reference it should be possible to skip specific parameters: https://docs.pytest.org/en/2.9.0/skipping.html#skip-xfail-with-parametrize

PR #5513, opened 2018-03-16
Bug 1445070: Update look and feel of manifesto pages

stephaniehobson:
## Description

Updating look and feel of manifesto pages in preparation to add addendum. The goal of this PR is to convert the pages from sandstone to pebbles before the design and translation work on the addendum begins.

## Issue / Bugzilla link

[Bug 1445070](https://bugzilla.mozilla.org/show_bug.cgi?id=1445070)

## Testing

- I have left some #l10n strings in place and commented out in the HTML until we have the final designs. I don't want to delete them in case we need them.
- The diff looks big because I removed a wrapper div but there are no string changes.
- What is `__twitterIntentHandler` and why did the old code check for it? I've left it in place for now.
- The most substantial change was to the javascript.


bedrock/mozorg/templates/mozorg/manifesto-base.html
```diff
@@ -0,0 +1,41 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{%- set_lang_files "main" -%}
```
flodolo: If the strings are only coming from main.lang, you don't need this line: [main.lang is a shared file](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L255).

stephaniehobson:
This is up on the demo server now: https://bedrock-demo-shobson.us-west.moz.works/en-US/about/manifesto/

bedrock/mozorg/templates/mozorg/manifesto-base.html
stephaniehobson: 👍

stephaniehobson:
I have an okay from the design side for this now.

bedrock/mozorg/templates/mozorg/about/manifesto.html
```diff
@@ -55,275 +55,263 @@
...
     </div>
   </article>
 </main>
 
 {% endblock %}
 
-{% block email_form %}{% endblock %}
-
 {% block js %}
   <!--[if IE 9]>
```
alexgibson: This conditional comment and script is included in the [pebbles base template](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/base-pebbles.html#L155-L157) by default, so we should be ok to remove this.

bedrock/settings/static_media.py
```diff
@@ -556,6 +556,13 @@
         ),
         'output_filename': 'css/manifesto-bundle.css',
     },
+    'manifesto': {
+        'source_filenames': (
+            'css/mozorg/manifesto.scss',
+            'css/base/mozilla-modal.less',
```
alexgibson: Is this still needed?

media/css/mozorg/manifesto.scss
```diff
@@ -0,0 +1,152 @@
...
+
+@import '../pebbles/includes/lib';
+@import '../pebbles/components/highlight';
+@import '../pebbles/components/newsletter';
+@import '../hubs/sections';
+@import '../hubs/buttons';
+@import '../hubs/masthead';
+
+.principle {
+    box-sizing: border-box;
```
alexgibson: Nit: use `border-box` mixin?

media/css/mozorg/manifesto.scss
```diff
@@ -0,0 +1,152 @@
...
+    }
+}
+
+@media #{$mq-desktop} {
+    .principle {
+        padding: 0 $desktop-content-padding $desktop-content-padding $desktop-content-padding;
+    }
+}
+
+@supports (display: grid) {
```
alexgibson: Not a blocker as the content is still perfectly readable in browsers that don't support CSS grid, but would it be worth considering a fallback to something like Flexbox for slightly older browsers?

media/css/mozorg/manifesto.scss
```diff
@@ -0,0 +1,152 @@
...
+
+.videos {
+    a {
+        display: block;
+        position: relative;
+        padding-top: 50px;
+        min-height: 18px;
+    }
+    img {
+        position: absolute;
```
alexgibson: Nit: alphabetising the CSS here and elsewhere would be nice

media/js/mozorg/manifesto.js
```diff
@@ -148,6 +16,7 @@ $(function() {
         // Check if the official Twitter widget is activated. If so, we don't
         // have to open a popup window ourselves as it will be opened by the
         // widget code.
+        // https://dev.twitter.com/docs/intents
```
alexgibson: It looks like this link has changed to https://dev.twitter.com/web/intents

As for `window.__twitterIntentHandler`, I think we should be ok to remove this. We wouldn't embed a twitter widget in pages on mozorg since it violated our websites privacy policy. Not entirely sure why this was here to begin with, as I see no history of the widget ever being on the page.

media/js/mozorg/manifesto.js
```diff
@@ -168,73 +37,34 @@ $(function() {
...
 
         if ($this.hasClass('tweet')) {
             // Open Twitter in a sub window
             event.preventDefault();
             openTwitterSubwin(section, href);
+        } else if ($this.hasClass('principle-number')) {
+            // nothing
         } else {
             // Open the link in a new tab
             $this.attr({
```
alexgibson: As this page no longer has modal states, I totally wouldn't object if we killed this catch-all condition that opens *every* link in a new window. Thoughts?

alexgibson:
On RTL pages, the principal number overlaps the principal test, and the video/image thumbnails are stil left aligned.

screen shot 2018-03-20 at 09 41 01

alexgibson:
The main contant container on `/manifesto/` could use the same top margin spacing as on `/manifesto/details/`

heading-margins

media/css/mozorg/manifesto.scss
stephaniehobson: Not for a design that's only going to live a week ;)

media/js/mozorg/manifesto.js
stephaniehobson: Is it okay if I leave this for the PR with the addendum in it? I want to spend some time to understand the analytics before I mess with them.

stephaniehobson:
Margins updated, RTL solved by ditching the dangling principle numbers. Nits, picked.

Ready for re-review :)

stephaniehobson:
Push again 2 minutes after I posted that because I didn't get all my changes in the first push >_<

media/css/mozorg/manifesto.scss
alexgibson: Ah, I wasn’t aware the design was going to change drastically again. Not worth spending too much time over 👍

PR #5516, opened 2018-03-20
[fix bug 1447269] Headings should not be smaller than base font size on small screens

alexgibson:
## Description
- Updates our `font-size-level` mixins in Pebbles, so that headings never become smaller than our base font size for body copy (something I have to frequently fight against).
- I decided against sneaking this into the homepage redesign PR, since it touches other pages.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1447269

## Testing
- This touches a lot of pages, but it only really makes a couple of pixels difference to certain bits of copy. So should hopefully be a fairly safe change.

### DIFF
--- a/media/css/pebbles/includes/mixins/_typography.scss
+++ b/media/css/pebbles/includes/mixins/_typography.scss
@@ -96,11 +96,7 @@
 }
 
 @mixin font-size-level4 {
-    @include font-size(16px);
-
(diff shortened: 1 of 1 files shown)


jpetto:
Agree this is safe. Worst case I can see is some font sizes look a little wrong on small viewports...which is already the case, so 🚢

PR #5519, opened 2018-03-21
Always send lang with newsletter preferences updates

pmac:
Otherwise basket will guess at a best language.


bedrock/newsletter/views.py
```diff
@@ -417,6 +417,8 @@ def existing(request, token=None):
             if not remove_all:
                 kwargs['newsletters'] = ",".join(newsletters)
             if kwargs:
+                # always send lang so basket doesn't try to guess
+                kwargs['lang'] = data['lang']
```
jpetto: AFAICT, this form field can never be blank, so this should be good.

PR #5520, opened 2018-03-21
[fix bug 1417888] Remove /firstrun bg animation.

jpetto:
- Also fixes bug 1447724 - text artifacts caused by svg animation

## Description

Improves `/firstrun` page performance by removing animation on SVG waves. Also fixes text artifacts near animated Sync SVG images.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1417888
https://bugzilla.mozilla.org/show_bug.cgi?id=1447724

## Testing

Ensure page elements are visible and transitions are working on all supported OSs.

### DIFF
--- a/media/css/firefox/firstrun/firstrun-quantum.scss
+++ b/media/css/firefox/firstrun/firstrun-quantum.scss
@@ -242,6 +242,9 @@ body {
     p {
         background: url('/media/img/firefox/firstrun/sync/sync-devices-icons-anim.svg') bottom center no-repeat;
         padding-bottom: 210px;
+        // avoids odd text rendering bugs triggered by animated SVG bg
+        // https://bugzilla.mozilla.org/show_bug.cgi?id=1447724
+        will-change: opacity, transform;
     }
 
     .content {
@@ -292,11 +295,8 @@ html[dir="rtl"] #main-header {
 
 /* -------------------------------------------------------------------------- */
 // Animations & transitions
-
-// Simplified interactions for some platforms
 #scene[data-sign-in=true] {
     .fxaccounts-container {
-        animation: HideVisibility 0s .1s 1 forwards;
         opacity: 0;
         transition: opacity .5s;
     }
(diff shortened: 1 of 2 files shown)


jpetto:
While working/testing this fix, I noticed performance _still_ isn't great on the page. Removing the SVG waves altogether seemed to free up ~500MB of memory (2017 MBP), so it's possible there's more that can be done here (or a problem with Firefox itself rendering these SVGs).

I don't necessarily think this means we shouldn't merge this change, but it probably _does_ mean we should look at a redesign/re-work of this page.

stephaniehobson:
Today was actually the first time I have been able to view this page on my local machine (not sure why).

Unfortunately, the tricks I used to view it earlier (sign out of sync, restart, open it in private browsing) are not working now. I see the blue waves but the animation does not run and the page content does not display.

I get this error in the console but I think the animation is supposed to run anyway:

> Will-change memory consumption is too high. Budget limit is the document surface area multiplied by 3 (894720 px). Occurrences of will-change over the budget will be ignored.

Can someone else please pick up this review? @craigcook @alexgibson.

media/css/firefox/firstrun/firstrun-quantum.scss
```diff
@@ -312,325 +312,44 @@ html[dir="rtl"] #main-header {
...
-
-            .content {
-                transition-delay: 2s;
-            }
-
-            a {
-                transition-delay: 2.2s;
-            }
-        }
+        opacity: .85;
```
stephaniehobson: I see some improvements when I bump this to 1

alexgibson:
It sounds like you need to configure UITour to work locally. Docs with steps needed are here: http://bedrock.readthedocs.io/en/latest/uitour.html

media/css/firefox/firstrun/firstrun-quantum.scss
```diff
@@ -242,6 +242,9 @@ body {
     p {
         background: url('/media/img/firefox/firstrun/sync/sync-devices-icons-anim.svg') bottom center no-repeat;
         padding-bottom: 210px;
+        // avoids odd text rendering bugs triggered by animated SVG bg
+        // https://bugzilla.mozilla.org/show_bug.cgi?id=1447724
+        will-change: opacity, transform;
```
craigcook: Seems to do the trick 👍

media/css/firefox/firstrun/firstrun-quantum.scss
craigcook: I don't get a huge performance boost locally but I also don't know if the slight transparency is a huge benefit either, so maybe any little bit helps?

PR #5521, opened 2018-03-22
[fix bug 1439771] Remove waitface experiment files (also bug 1431795)

jpetto:
## Description

Removes `/new/?xv=waitface` experiment files.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1439771
https://bugzilla.mozilla.org/show_bug.cgi?id=1431795

## Testing

Ensure no regressions and that all necessary files have been removed.


bedrock/firefox/views.py
```diff
@@ -494,26 +489,14 @@ def new(request):
...
-                template = 'firefox/new/wait-face/scene1.html'
+            template = 'firefox/new/wait-face/scene1.html'
         elif lang_file_is_active('firefox/new/reggiewatts', locale) and experience == 'reggiewatts':
             template = 'firefox/new/reggie-watts/scene1.html'
         else:
             template = 'firefox/new/scene1.html'
 
-    return l10n_utils.render(request, template, {'v': variant})
+    # sending empty context so we don't have to update the tests every time we need a variant value
+    return l10n_utils.render(request, template, {})
```
jpetto: @pmac - Is this horrible? Any other suggestions here? It seems silly to have to re-add `ANY` to all tests of this view when we need a variation sent in `context`.

flodolo:
(please remember to flag the PR as L10N if it's localized)

bedrock/firefox/views.py
pmac: Well, I see you no longer need variant in the context, but you also don't need to not have it... if you expect to need variant again then I see no compelling reason for you to remove it. You can just not delete the bits from lines 479-485, and maybe add a comment. I think that'd be less confusing.

jpetto:
> (please remember to flag the PR as L10N if it's localized)

@flodolo I don't think this should impact L10N - as far as I'm aware, the templates being removed were for en-US only experiments.

flodolo:
> @flodolo I don't think this should impact L10N - as far as I'm aware, the templates being removed were for en-US only experiments.

www.mozilla.org//de/firefox/new/?xv=waitface

jpetto:
The `/new/?xv=waitface` page is not being removed or changed in this PR. This PR removes two different experiments based on that page.

flodolo:
> The /new/?xv=waitface page is not being removed or changed in this PR. This PR removes two different experiments based on that page.

Ugh, I clearly have no idea of how it works :-\ You're right, this doesn't have l10n impact then.

jpetto:
@alexgibson Rebased!

PR #5523, opened 2018-03-23
[fix bug 1403474] Add back link to quantum privacy pages.

jpetto:
## Description

Adds a link back to `/privacy/` on all Quantum-themed privacy pages.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1403474

## Testing

Ensure I did the L10n correctly?

### DIFF
--- a/bedrock/privacy/templates/privacy/base-quantum.html
+++ b/bedrock/privacy/templates/privacy/base-quantum.html
@@ -2,6 +2,7 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% add_lang_files "privacy/index" %}
 {% extends "base-pebbles.html" %}
 
 {% block page_title %}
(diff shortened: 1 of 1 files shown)


bedrock/privacy/templates/privacy/base-quantum.html
```diff
@@ -2,6 +2,7 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% add_lang_files "privacy/index" %}
```
alexgibson: This seems to work ok, but looking in this template there are other strings here already e.g. `"Choose how you want to share this data in Firefox."`, which for some reason live in `.main.lang`. Just want to make sure we're still ok using this older lang file for these newer templates. Thoughts @flodolo ?

flodolo: Redirecting to @peiying2

alexgibson: @peiying2 I'm not quite clear if this means we are ok to use the older `privacy/index.lang` file, or not?

peiying2: @alexgibson yes, let's use the older template privacy/index.lang to keep it simple. Thx.

alexgibson: Thanks!

PR #5525, opened 2018-03-26
[fix bug 1444000] Add ?xv=portland /new variation.

jpetto:
## Description

Adds a `/new/?xv=portland` variation.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1444000

## Testing

- [x] Ensure no regressions.
- [x] Ensure custom adjust link is used for iOS 
- [x] Ensure custom adjust link is used for Android
- [x] Check older browsers


media/css/firefox/new/portland.scss
```diff
@@ -0,0 +1,290 @@
...
+        width: 100%;
+    }
+
+    h3 {
+        @include font-size-level2;
+        font-family: 'ITC Franklin Gothic LT';
+        margin-bottom: 20px;
+        text-transform: uppercase;
+
+        &:before {
```
stephaniehobson: Alphabetize?

media/js/firefox/new/portland-font-tracker.js
```diff
@@ -0,0 +1,4 @@
+// provided by fonts.com
+/* eslint-disable */
+eval(function(p,a,c,k,e,r){e=function(c){return c.toString(a)};if(!''.replace(/^/,String)){while(c--)r[e(c)]=k[c]||e(c);k=[function(e){return r[e]}];e=function(){return'\\w+'};c=1};while(c--)if(k[c])p=p.replace(new RegExp('\\b'+e(c)+'\\b','g'),k[c]);return p}('4 7=g.b;4 2=3.p(\'r\');2.9=\'a/5\';2.c=\'d\';2.e=(\'6:\'==3.u.h?\'6:\':\'i:\')+\'//j.k.l/t/1.5?m=n&o=\'+7;(3.8(\'q\')[0]||3.8(\'s\')[0]).f(2);',31,31,'||mtiTracking|document|var|css|https|projectId|getElementsByTagName|type|text|MTIProjectId|rel|stylesheet|href|appendChild|window|protocol|http|fast|fonts|net|apiType|css|projectid|createElement|head|link|body||location'.split('|'),0,{}))
```
april: This code should not be compressed/minified in the repository.

It is evaluating arbitrary third-party code that is essentially impossible to read. If we want to minify it, we should do it as part of the Bedrock build process.

jpetto: Unfortunately this is how the code is provided from the third party. I don't believe we have access to an unminified version. 😕

april: Could we ask fonts.com for the unminified version? I would be hesitant to run this kind of code on our website, when run from a 3rd party.

jpetto: Hm, I'm going to try a different approach that loads external CSS as opposed to external JS. Will push a commit shortly.

april:
Anyways, aside from not using `https` and a vague feeling that we should be trying to use open fonts whenever possible, I don't have any security objections to this.

Nice fix to remove the JavaScript loading, @jpetto.

PR #5533, opened 2018-03-29
[fix bug 1448875] Skip /zh-CN/ home page hreflang

alexgibson:
## Description
- Skip `/zh-CN/` hreflang on the home page, as it redirects to `http://www.firefox.com.cn/`

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1448875

## Testing
- Visit the homepage and you should not see either `hreflang="zh-CN"` or `hreflang="zh"` listed as alternate links.
- Visit any other page, such as `/firefox/` and both `hreflang="zh-CN"` or `hreflang="zh"` should be listed.


alexgibson:
@rraue sanity check on the logic here?

rraue:
@alexgibson Sanity checked. Exactly like that!

PR #5543, opened 2018-03-30
Use python's sha256 instead of requiring a shell program

pmac:
This should mean easier use of db download script for developers.

### DIFF
--- a/bin/db_s3_utils.py
+++ b/bin/db_s3_utils.py
@@ -1,4 +1,5 @@
 import json
+from hashlib import sha256
 from os import getenv
 from subprocess import check_output, CalledProcessError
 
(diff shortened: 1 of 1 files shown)


bin/db_s3_utils.py
```diff
@@ -1,19 +1,30 @@
...
 JSON_DATA_FILE = getenv('AWS_DB_JSON_DATA_FILE', 'bedrock_db_info.json')
 DB_FILE = 'bedrock.db'
 CACHE = {}
+BLOCKSIZE = 65536
+
+
+def _sha256_sum(filename):
+    hasher = sha256()
+    with open(filename, 'rb') as fh:
+        for chunk in iter(lambda: fh.read(BLOCKSIZE), b""):
```
glogiotatidis: elegant :tophat:

pmac: Wish I could take credit for thinking of this. Found an example like it on StackOverflow and looked up `iter` and this use (with the 2nd parameter) in the docs. I really like it too. And I always love finding odd corners of Python like this. There seem to be a nearly infinite number of things to discover :)

PR #5544, opened 2018-03-30
Bug 1435821: Add Mitre CVE JSON Feed

pmac:


albill:
Can we get this merged? It has been almost two weeks and we want to get this feed up for the Mitre folks to read.

pmac:
Sure. I’ll get it reviewed. I’ve been waiting on some word that the format was okay. I thought we were still waiting on dveditz to verify that it contained the proper data.

bedrock/security/management/commands/update_security_advisories.py
```diff
@@ -117,6 +118,38 @@ def add_hofers(filename, data):
         )
 
 
+def parse_cve_id(cve_id):
+    cve_year, cve_order = cve_id.split('-')[1:]
+    return int(cve_year), int(cve_order)
+
+
+def add_or_update_cve(data):
```
jpetto: Do you think it's worth adding tests for these two methods? Might be nice just to have an example of expected input data format.

bedrock/security/models.py
```diff
@@ -132,3 +132,89 @@ def year_quarter(self):
...
+                }
+            })
+
+        return {
+            'data_type': 'CVE',
+            'data_format': 'MITRE',
+            'data_version': '4.0',
+            'CVE_data_meta': {
+                'ID': self.id,
+                'ASSIGNER': 'cve@mitre.org'
```
dveditz: The ASSIGNER is 'security@mozilla.org'

bedrock/security/models.py
```diff
@@ -132,3 +132,89 @@ def year_quarter(self):
...
+                    ]
+                }
+            },
+            'problemtype': {
+                'problemtype_data': [
+                    {
+                        'description': [
+                            {
+                                'lang': 'eng',
+                                'value': self.title,
```
dveditz: Problemtype is not the title, they are wanting something more generic like a CWE "Use after free", "buffer overflow", etc. see https://cwe.mitre.org/data/definitions/659.html for a partial list.

Our current yaml doesn't have a clean existing source of this. Some of the right phrases exist in the titles, but sometimes not.

I'm confused by the spec, which says "Mandatory: None", but also that it's a CNA requirement (and it shows up in the minimal examples). I'm tempted to drop it and see what complaints we get, but I suppose title is as good as anything if we don't drop it.

bedrock/security/models.py
```diff
@@ -132,3 +132,89 @@ def year_quarter(self):
...
+        """Return a MITRE format data structure for the CVE
+
+        See https://github.com/CVEProject/automation-working-group/blob/master/cve_json_schema/DRAFT-JSON-file-format-v4.md
+        """
+        product_data = []
+        for prod_name, versions in self.product_versions().iteritems():
+            product_data.append({
+                'product_name': prod_name,
+                'version': {
+                    'version_data': [{'version_value': vers} for vers in versions],
```
dveditz: We need version_data to be
  [{'version_value': vers, 'version_affected': '<' }  for ...

bedrock/security/models.py
pmac: We could come up with a yaml entry for `cve_problem_type` or the like. And I can use that if defined or `title` if not. Would that be sufficient?

pmac:
@dveditz just pushed a fix for the assigner and version_data. Let me know what you think about the title thing and we can hopefully get this launched soon. Thanks for the review!

bedrock/security/models.py
dveditz: That sounds good. And if the CVE people complain we could easily change it to be cve_problem_type if present and nothing otherwise.

dveditz:
The other changes look good, thanks!

bedrock/security/models.py
pmac: Works for me. I'll get this updated and deployed to the demo early next week for you to check out, then hopefully we can merge this shortly thereafter.

bedrock/security/management/commands/update_security_advisories.py
```diff
@@ -117,6 +118,39 @@ def add_hofers(filename, data):
...
+    for cve_id, advisory in data['advisories'].iteritems():
+        cve_year, cve_order = parse_cve_id(cve_id)
+        update_advisory_bugs(advisory)
+        cve_title = advisory.get('cve_problemtype', advisory.get('title')) or ''
```
pmac: @dveditz I've just pushed this change. You can add a `cve_problemtype` field to any advisory in the `advisories` list and it will override the title for that one in the CVE feed. We can always tweak later if need be.

albill:
Go for it.

albill:
That worked. It is populated now.

PR #5549, opened 2018-04-02
[fix bug 1449758] Use high res logo on FB containers page.

jpetto:
## Description

Just updates the logo to be high res (when applicable).


jpetto:
(Should have caught this in review - my bad.)

PR #5555, opened 2018-04-05
[fix bug 1450106] Add FB container firstrun template.

jpetto:
## Description

Does two primary things:

1. Appends a funnelcake ID to all download clicks for *all* visitors landing on `/firefox/facebookcontainer/`. Through `env` config, actually receiving this funnelcake is limited to `en-US` and `win/win64` platforms only.

2. Adds a `/firstrun` template that is loaded by the funnelcake build mentioned above. This template focuses on the Facebook container add-on.

I've put both the JS necessary to append the funnelcake ID (step 1 above) *and* display the `/firstrun` template behind the same switch, just in case we need to quickly disable the whole thing.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1450106
For reference, funnelcake bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1450463

## Testing

With `DEV=True` (which it probably already is), visit `/firefox/facebookcontainer/` in a non-Fx browser. Clicking the download button should bring you to `/download/thanks/?f=135`.

To actually receive/install the funnelcake, you must be on Windows. Installing/running the funnelcake (with no pre-existing profile!) should result in Fx opening to `https://www.mozilla.org/en-US/firefox/59.0.2/firstrun/?xv=facebook-container`. (Obviously this will display the current `/firstrun` template, as the new one is not yet in prod.)

To manually test the new `/firstrun` template, visit `/firefox/59.0.2/firstrun/?xv=facebook-container`. The button there should link to the Facebook container extension.

https://bedrock-demo-jpetto.oregon-b.moz.works/en-US/firefox/facebookcontainer/
https://bedrock-demo-jpetto.oregon-b.moz.works/en-US/firefox/59.0.2/firstrun/?xv=facebook-container


jpetto:
**DO NOT MERGE** for now while I try to get actual high res images. (The ones in this PR are just straight copies so I could have the code finalized.)

craigcook:
I see none of the strings are wrapped for l10n. Are we very confident this page will never be translated? I know it isn't for now, but if it's wildly successful and someone decides next week it needs to roll out to Germany... maybe good to wrap strings anyway?

PR #5556, opened 2018-04-05
Move addendum title behind l10n tag

stephaniehobson:
## Description
The heading "The Mozilla Manifesto Addendum" was added quickly behind an English only flag for launch. Now that we have time, this heading should be available for translation to all locales.

## Issue / Bugzilla link
[Bug 1445070 - Addendum to manifesto](https://bugzilla.mozilla.org/show_bug.cgi?id=1445070)

## Testing
The title should show up when translated. If the addendum section is labeled the principles section should also be labeled as the manifesto.


alexgibson:
Oh, it would also be great to reference the bug number in the commit message? Thanks @stephaniehobson 💯

alexgibson:
r+ - squash commits and reference the bug number in the commit, then we're good to merge?

stephaniehobson:
@alexgibson The bug number is in the commit title and message... can you give me a bit more detail on where you think it is missing from?

alexgibson:
> can you give me a bit more detail on where you think it is missing from?

I see the bug number in the GitHub issue here, but not in either of the commits. Maybe you have a local commit that needs pushing?

stephaniehobson:
Is this not the commit message?

screen shot 2018-04-11 at 12 29 20

flodolo:
> Is this not the commit message?

No, see Commits tab in the same screen (there should be one, with the bug reference)

schermata 2018-04-12 alle 07 08 15

PR #5557, opened 2018-04-05
[fix bug 1449758] Add Focus CTA for mobile browsers

craigcook:
## Description
The extension is only available for desktop Firefox. This adds an alternative CTA for mobile browsers promoting Firefox Focus.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1449758#c26

## Testing
The mobile tagline and CTA should only be shown when Android or iOS are detected, and the platform-appropriate store badge should appear.
https://bedrock-demo-craigcook.oregon-b.moz.works/firefox/facebookcontainer/

### DIFF
--- a/bedrock/firefox/templates/firefox/facebookcontainer/index.html
+++ b/bedrock/firefox/templates/firefox/facebookcontainer/index.html
@@ -2,6 +2,8 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% from "macros.html" import google_play_button with context %}
+
 {% extends "firefox/base-pebbles.html" %}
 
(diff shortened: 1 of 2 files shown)


jpetto:
Code looks good, but I think the UI could use a little bit of love. The bunched together copy doesn't look great:

screen shot 2018-04-06 at 10 07 59 am

Standardizing the margins/padding could go a long way. Perhaps some visual indicator that the mobile copy is a notice/warning of some sort? As-is, it would be easy to skip the copy and just click the button, thinking you're getting the extension.

craigcook:
You're right, the message is totally lost and all I see is the button. I've made some changes to try to call out this as something separate from the page around it. 

mobile-cta-box

I wonder if we should even hide *all* the content on mobile, even the headline and video.

Issue #5561, opened 2018-04-07
Integration test timeouts on bedrock-dev


jgmize:
Example from https://ci.us-west.moz.works/view/Bedrock/job/bedrock_multibranch_pipeline/job/master/943/testReport/junit/tests.functional.firefox/test_whatsnew/Test_bedrock_dev_oregon_b___firefox___test_send_to_device_fails_when_missing_required_fields/ 

```
tests/functional/firefox/test_whatsnew.py:25: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/lib/python2.7/site-packages/pypom/page.py:93: in open
    self.wait_for_page_to_load()
tests/pages/base.py:22: in wait_for_page_to_load
    self.wait.until(lambda s: 'loaded' in el.get_attribute('class'))
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = <selenium.webdriver.support.wait.WebDriverWait (session="4609e5e181c44ce49386db84f2228fa3")>
method = <function <lambda> at 0x7f5289061f50>, message = ''

    def until(self, method, message=''):
        """Calls the method provided with the driver as an argument until the \
            return value is not False."""
        screen = None
        stacktrace = None
    
        end_time = time.time() + self._timeout
        while True:
            try:
                value = method(self._driver)
                if value:
                    return value
            except self._ignored_exceptions as exc:
                screen = getattr(exc, 'screen', None)
                stacktrace = getattr(exc, 'stacktrace', None)
            time.sleep(self._poll)
            if time.time() > end_time:
                break
>       raise TimeoutException(message, screen, stacktrace)
E       TimeoutException: Message:

end_time   = 1523035849.181287
message    = ''
method     = <function <lambda> at 0x7f5289061f50>
screen     = None
self       = <selenium.webdriver.support.wait.WebDriverWait (session="4609e5e181c44ce49386db84f2228fa3")>
stacktrace = None
value      = False

/usr/local/lib/python2.7/site-packages/selenium/webdriver/support/wait.py:80: TimeoutException
```

[...]

PR #5562, opened 2018-04-09
Test stage on pushes to master for #5561

jgmize:
## Description
I'd like to test the hypothesis that the browser test timeouts could be caused by something specific to when settings.DEV=True by testing the stage environment, where DEV=False, on every push to master instead of only on push to the prod branch. 

## Issue / Bugzilla link

#5561

## Testing

Stage will be tested on push to master

### DIFF
--- a/jenkins/branches/master.yml
+++ b/jenkins/branches/master.yml
@@ -7,8 +7,8 @@ regions:
   - tokyo
   - frankfurt
 apps:
-  - bedrock-dev
   - bedrock-stage
+  - bedrock-dev
 integration_tests:
   oregon-b:
     - firefox
--- a/jenkins/branches/master.yml
+++ b/jenkins/branches/master.yml
@@ -8,6 +8,7 @@ regions:
   - frankfurt
 apps:
   - bedrock-dev
+  - bedrock-stage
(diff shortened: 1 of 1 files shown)


alexgibson:
Not sure if this is easily possible, but would a more accurate test be to set Dev=False on the same environment where we’re seeing the issues?

pmac:
> Not sure if this is easily possible, but would a more accurate test be to set DEV=False on the same environment where we’re seeing the issues? Looking at the prod branch, we don’t seem to be having the same issues against stage, but i’m not sure if that tells us the cause is something directly related to the Dev setting.

We could do a brief test like that. The l10n team uses www-dev with DEV=True so that they can see unactivated translations there to check their work. We definitely have to keep that, but we could do a time-boxed test I think no problem.

pmac:
And for this PR I'd much rather test another instance than stage for this. The dev image is built with dev data (l10n, relnotes, etc.) so I don't really want it on stage if we can help it. It should overwrite at the next DB update, but still. I like either Alex's suggestion of trying a test with DEV=False on dev, or setting DEV=False on a demo and trying that.

PR #5565, opened 2018-04-09
Fix #5551: Store Newsletter data from basket in the DB

### LINKED ISSUE #5551: Move basket newsletter data from cache to DB
The [email_newsletter_form](https://github.com/mozilla/bedrock/blob/2fb16c05fbb847b57e0fbbea8a5b51d51d554e43/bedrock/newsletter/templatetags/helpers.py#L20) macro [used across the site](https://github.com/mozilla/bedrock/search?utf8=%E2%9C%93&q=email_newsletter_form&type=) currently attempts to fetch newsletter data from basket.mozilla.org whenever the cache expires. The current cache timeout is 1 hour, and happens on a per instance basis. We currently [mitigate the risk of a complete basket outage](https://github.com/mozilla/bedrock/pull/4214) by using locally stored data in the event of an error, but if the error is a timeout or basket is experiencing any performance issues then this could impact the performance of pages across the site, including the home page. Now that bedrock is [SQLitened](https://mozilla.github.io/meao/2018/03/28/bedrock-the-sqlitening/) it would be faster and more reliable to always fetch the newsletter data from the local sqlite file and update it in a single global cron job similar to our other external data sources.

pmac:
No longer rely on requests to basket and cache for newsletter data required to render pages. Only post info to basket during user interaction.


bedrock/newsletter/forms.py
```diff
@@ -59,16 +59,6 @@ def get_lang_choices(newsletters=None):
     return sorted(lang_choices, key=itemgetter(1))
 
 
-def newsletter_title(newsletter):
```
pmac: Seems like this was vestigial. Can't find it used anywhere.

bedrock/newsletter/management/commands/update_newsletter_data.py
```diff
@@ -0,0 +1,30 @@
...
+    def add_arguments(self, parser):
+        parser.add_argument('-q', '--quiet', action='store_true', dest='quiet', default=False,
+                            help='If no error occurs, swallow all output.'),
+
+    def handle(self, *args, **options):
+        newsletters = basket.get_newsletters()
+        if not newsletters:
+            raise CommandError('No data from basket')
+
+        Newsletter.objects.all().delete()
```
jgmize: I was concerned about race conditions here at first, but then I remembered that this is being run in a separate container that isn't serving traffic, and that the web containers will update the entire DB atomically, so this shouldn't be an issue.

PR #5571, opened 2018-04-11
Move langfile data to the database for use and distribution

pmac:


jgmize:
@pmac and I discussed this and decided against this approach.

PR #5572, opened 2018-04-12
[bug 1453144] Facebook Container - fix typo

craigcook:
## Description
“Facebook” is a proper noun and should always be capitalized in English. We need to get this string updated so l10n can begin.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1453144


flodolo:
Is it possible to push this fix to production before the weekend? Otherwise we risk enabling pages with that string untranslated.

jpetto:
> Is it possible to push this fix to production before the weekend? Otherwise we risk enabling pages with that string untranslated.

We'll get it out today.

jpetto:
> Is it possible to push this fix to production before the weekend? Otherwise we risk enabling pages with that string untranslated.

This is now in production.

PR #5575, opened 2018-04-12
Fix bug 1447772: Require fraud report attachment to be images

pmac:
Also move legal tests so that py.test will find them


bedrock/legal/forms.py
```diff
@@ -97,9 +97,7 @@ class FraudReportForm(forms.Form):
             }
         )
     )
-    input_attachment = forms.FileField(
-        required=False,
-    )
+    input_attachment = forms.ImageField(required=False)
```
jpetto: From the docs, it appears this auto-validates format to ensure the file is of a type understood by Pillow. Is that right?

pmac: That is my understanding as well. And there is a test that tries to upload a text file and fails.

pmac:
@jpetto think I should change the display name on the form from "Attachment" to e.g. "Image Attachment"? There are probably l10n issues with that though...

bedrock/legal/forms.py
jpetto: I guess my question is: will this allow things like .pdf and .tiff? Do we want to allow those kinds of images? Should we force a restriction to the three "standards": .png, .gif, and .jpg?

pmac:
The error message when you upload a non-image does say "Upload a valid image. The file you uploaded was either not an image or a corrupted image." UX isn't the best, but it's not bad either. Probably fine.

jpetto:
> think I should change the display name on the form from "Attachment" to e.g. "Image Attachment"? There are probably l10n issues with that though...

I do think this is a good idea, but a bummer for L10n. Could be put behind a tag though?

pmac:
Oh. It seems that page is localizable, but not localized. So I can probably change the string.

pmac:
Okay. I updated the help note to prepend `Images only.`

bedrock/legal/forms.py
pmac: Might allow .tiff, but doesn't allow .pdf.

pmac: These are the supported formats http://pillow.readthedocs.io/en/latest/handbook/image-file-formats.html

pmac: I believe we could filter on image type if we need to.

jpetto: That's quite the list! I think we're pretty well covered. We can limit at a later date if needed.

glogiotatidis:
lgtm, nice work!

PR #5578, opened 2018-04-13
Move tests.py to tests/test_base.py so pytest will find and run them

pmac:
Seems we've been ignoring these tests for quite a while.


pmac:
The externalfiles app had changed how it worked (it used to directly fetch files from the SVN server, but now does a git pull) but the tests weren't updated and we missed that since they weren't running. So yeah, they are for functionality that we no longer have.

pmac:
Thanks for the review!

PR #5582, opened 2018-04-13
Fix #5581: Switch lang switcher to client-side

### LINKED ISSUE #5581: Remove the ?lang feature
Instead of using a query parameter to change the language we should do this redirect client side. This will mean fewer unintended language redirections and simpler middleware.

pmac:
This is not yet in a mergable state IMHO, but I would like for people to let me know what they think of the proposal and technique. There must be a better way to parse the current URL, but this does work. Needs tests, but there don't seem to currently be tests for the lang switcher function.

### DIFF
--- a/media/js/base/mozilla-utils.js
+++ b/media/js/base/mozilla-utils.js
@@ -37,17 +37,24 @@ if (typeof Mozilla === 'undefined') {
         });
     };
 
+    Utils.switchPathLanguage = function(location, newLang) {
+        // get path without locale
+        var urlpath = location.pathname.slice(1).split('/').slice(1).join('/');
+        return '/' + newLang + '/' + urlpath + location.search;
+    };
+
     // language switcher
     Utils.initLangSwitcher = function() {
         var $language = $('#page-language-select');
         var previousLanguage = $language.val();
         $language.on('change', function() {
+            var newLanguage = $language.val();
             window.dataLayer.push({
                 'event': 'change-language',
-                'languageSelected': $language.val(),
+                'languageSelected': newLanguage,
                 'previousLanguage': previousLanguage
             });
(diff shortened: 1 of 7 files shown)


jpetto:
The approach here looks good and sane to me. Can probably do more cleanup (like removing tests around `FF_EXEMPT_LANG_PARAM_URLS`), but I think it's close.

FWIW, I'm not sure there's a better approach to parsing the current URL that'll work in old IE.

pmac:
Tests are cleaned up. I feel like there should be a test for the new JS as well, but I'm a bit of a noob as far as that goes. If anyone can point me in the right direction I'm happy to give it a go.

media/js/base/mozilla-utils.js
```diff
@@ -42,12 +42,16 @@ if (typeof Mozilla === 'undefined') {
...
             window.dataLayer.push({
                 'event': 'change-language',
-                'languageSelected': $language.val(),
+                'languageSelected': newLanguage,
                 'previousLanguage': previousLanguage
             });
-            $('#lang_form').attr('action', window.location.hash || '#').submit();
+            // get path without locale
+            var urlpath = window.location.pathname.slice(1).split('/').slice(1).join('/');
+            var newpath =  '/' + newLanguage + '/' + urlpath + window.location.search;
```
jpetto: It might be nice to add a helper method (and associated test) for this URL parsing/rebuilding functionality. Maybe something like `Utils.switchLangauge(pathName, newLang, queryString)`. That function would then be pure and easily testable.

jpetto:
Should probably do a `run-integration-tests` run to ensure that our functional tests (particularly [this one](https://github.com/mozilla/bedrock/blob/master/tests/functional/test_l10n.py#L13)) still pass. I *think* it'll still pass, but better to make sure.

PR #5584, opened 2018-04-17
Fix #5583: Support loading config from the DB

### LINKED ISSUE #5583: Move feature switch distribution to the DB
It would be easier, faster, less disruptive, and more reliable to distribute at least the feature switches via the database. It would work like so:

1. A management command would pull/update the [www-config repo](https://github.com/mozmeao/www-config)
2. Said management command would load the config values from the `*.env` file appropriate to the app (dev, stage, or prod) into a database table
3. The switch functions in bedrock would look for a configuration by name in the environment, then database, then return the default (value of `settings.DEV`).

I'd also like to look into the possibility of distributing other settings changes in this way if possible. I know there are 3rd party django apps out there for storing general settings in the DB, but I'm not sure if any are simple enough for what we want and also support loading overrides from the environment.

pmac:
* Convert from python-decouple to everett for config loading
  * Gives us more flexibility on loading config values and will allow us to extend config storage options
* Create a model for storing configs
* Create a management command for loading configs from git repo


pmac:
After thinking more about this and talking with others I don't think this is the right way to go. It's cool to be able to distribute settings in this way, but it really only benefits a specific kind of setting, i.e. mostly waffle switches. So I think I'll keep those separate and start a new thing for feature flips and other similar settings, while keeping the more traditional role of the www-config repo intact. Closing this in favor of a soon-to-come PR in that direction.

PR #5585, opened 2018-04-17
[fix bug 1436737] Remove blog thumbnail links from global nav side drawer

alexgibson:
## Description
- Removes blog thumbnail links from the global nav side drawer which have not been updated in over a year.
- Removing this old content reduces the number of image requests for every page load.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1436737

## Testing
- Ensure no regressions
- Make sure I didn't miss anything?


alexgibson:
> Only question from me is - should we keep the CSS around in case we again run promos like this? It's a rather small amount of code, but I suppose we can also dig it up out of history if needed. Your call!

I say let's bin it. Like you say, we always have git history. No one has bothered to update these since it launched, so I find it unlikely that'll change. The nav may be completely redesigned soon anyway 👍

PR #5587, opened 2018-04-17
Bug 1388914: Redirect /thunderbird to thunderbird.net

stephaniehobson:
Catch the case where there is no slash

## Description

Catch the case where there is no trailing slash.

## Issue / Bugzilla link

[Bug 1388914 - Delete thunderbird content](https://bugzilla.mozilla.org/show_bug.cgi?id=1388914)

## Testing

Does it redirect?

### DIFF
--- a/bedrock/redirects/redirects.py
+++ b/bedrock/redirects/redirects.py
@@ -2061,7 +2061,7 @@ redirectpatterns = (
              'https://www.thunderbird.net/thunderbird/{version}/releasenotes/'),
     redirect(r'^thunderbird/(?P<version>[^/]+)/system-requirements/$',
              'https://www.thunderbird.net/thunderbird/{version}/system-requirements/'),
-    redirect(r'^thunderbird/(?P<path>.*)', 'https://www.thunderbird.net/{path}'),
+    redirect(r'^thunderbird/?(?P<path>.*)', 'https://www.thunderbird.net/{path}'),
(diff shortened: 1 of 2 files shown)


bedrock/redirects/redirects.py
```diff
@@ -2061,6 +2061,7 @@
              'https://www.thunderbird.net/thunderbird/{version}/releasenotes/'),
     redirect(r'^thunderbird/(?P<version>[^/]+)/system-requirements/$',
              'https://www.thunderbird.net/thunderbird/{version}/system-requirements/'),
+    redirect(r'^thunderbird', 'https://www.thunderbird.net/'),
```
alexgibson: Instead of adding another redirect here I *think* you can just add a `?` after `thunderbird/` in the existing rule below?

```python
redirect(r'^thunderbird/?(?P<path>.*)', 'https://www.thunderbird.net/{path}'),
```

tests/redirects/map_globalconf.py
```diff
@@ -239,6 +239,7 @@
     url_test('/en-US/projects/calendar/random/stuff/', 'https://www.thunderbird.net/calendar/'),
 
     # bug 1388914
+    url_test('/thunderbird', 'https://www.thunderbird.net/'),
```
alexgibson: You can combine these two tests like so:

```python
url_test('/thunderbird{,/}', 'https://www.thunderbird.net/'),
```

PR #5593, opened 2018-04-20
Fix bug 1299817: Add img tag to set HSTS header for mozilla.org

pmac:
Bedrock sets HSTS for www.mozilla.org, but only a request from the root domain itself can set it for the root domain. This will set HSTS for all visitors to the site.

### DIFF
--- a/bedrock/base/templates/base-pebbles.html
+++ b/bedrock/base/templates/base-pebbles.html
@@ -187,5 +187,9 @@
       <script src="https://pontoon.mozilla.org/pontoon.js"></script>
     {% endif %}
     <!--<![endif]-->
+
+    {# Bug 1299817 #}
+    {# Sets the HSTS header for the root domain #}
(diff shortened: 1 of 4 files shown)


jpetto:
As long as the only important factor is the image coming over the network (and not rendered in any way), I think we just need to add `class="hidden"` to avoid layout issues.

pmac:
@jpetto hidden class added.

jpetto:
D'oh! I just remembered we have a new base template this should be added to: `/bedrock/base/templates/base-pebbles-basic.html`.

pmac:
> D'oh! I just remembered we have a new base template this should be added to: /bedrock/base/templates/base-pebbles-basic.html.

But that is just a child of `base-pebbles.html`, so it should inherit it.

jpetto:
So it is. Code review before coffee - not all the cylinders were firing. Looks good!

PR #5596, opened 2018-04-24
Exclude from CVE feed if feed flag is false

pmac:


pmac:
I still need tests for all of this, but for now this feed is experimental and in some bit of flux.

PR #5607, opened 2018-04-30
[fix bug 1457024] Add /whatsnew view for Fx60.

jpetto:
## Description

Adds a new template for `/firefox/60.0/whatsnew/` for almost all locales.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1457024

## Testing

https://www-demo4.allizom.org/en-US/firefox/60.0/whatsnew/

There are many states, detailed in the `whatsnew-60.js` file. They are:

1. User is not logged in to Sync (or UITour JS fails/times out):
    Display the FxA iframe

2. User is logged in to Sync, but does not have any mobile devices set up:
    Display App/Play store badges for Firefox Mobile

    - Send to Device widget is available in the user's locale:
        Display the Send to Device widget
    - Send to Device widget is *not* available in the user's locale:
        Display a QR code for Firefox Mobile

3. User is logged in to Sync, and *does* have a mobile device set up:
    Display a QR code for Firefox Focus
    Display App/Play store badges for Focus

To test the above states on demo4, you will need to have a profile set up to test FxA on demo servers as detailed [in the docs](http://bedrock.readthedocs.io/en/latest/firefox-accounts.html#demo-server-testing). You will need to further customize this profile to [allow UITour to work on demo4](http://bedrock.readthedocs.io/en/latest/uitour.html#local-development).


bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
```diff
@@ -0,0 +1,124 @@
...
+
+ {% block content %}
+ <main>
+   {% with custom_message=_('Congrats! You’re using the latest version of Firefox.') %}
+     {% include 'firefox/whatsnew/includes/header.html' %}
+   {% endwith %}
+   <div class="main-content {% if show_send_to_device %}has-send-to-device{% else %}no-send-to-device{% endif %}">
+     <div class="content-container visible" id="logo-container">
+       <div class="logo-wrapper">
+         {{ high_res_img('logos/firefox/logo-quantum.png', {'alt': 'Firefox', 'width': '100', 'height': '100', 'class': 'logo-fx showing', 'id': 'logo-fx'}) }}
```
alexgibson: The Firefox logo here has a class of `showing` by default, which means it's the only thing visible before the JS executes. This making for a very quick but noticable content jump on page load:

https://screenshots.firefox.com/JyJviwyhEwH0s2pp/127.0.0.1

Would hiding it by default along with the copy make sense, or cause more trouble than it's worth?

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
```diff
@@ -0,0 +1,124 @@
...
+
+     {#
+       If the user is logged in to Sync *and* has mobile devices sync'ed, display
+       QR code for Focus and App/Play store badges.
+     #}
+     <div class="content-container focus-container" id="focus-container">
+       <header class="content-header">
+         <h2>{{ _('Firefox Focus') }}</h2>
+         <p class="tagline">{{ _('Blocks ads with trackers. Automatically. Plus, it can even speed up page loading. Get the private browser that has your back.') }}</p>
+         <p>{{ _('Scan the QR code to get started.') }}</p>
```
alexgibson: Could this string be shown along side the other QR code state? The added context it provides would be useful, but I guess it depends which locales have this string translated.

alexgibson:
The no-JS view for non-send-to-device locales could do with a little extra attention? https://screenshots.firefox.com/5vLWPbdbQrILHqPw/127.0.0.1

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
```diff
@@ -0,0 +1,124 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+ {% from "macros.html" import google_play_button, send_to_device with context %}
+
+ {% add_lang_files "firefox/whatsnew_59" "firefox/mobile" "firefox/sendto" %}
```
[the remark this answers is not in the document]
flodolo: No, because people will get to this page based on a list given to RelEng that [takes it into account](https://bugzilla.mozilla.org/show_bug.cgi?id=1457024#c6).

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
jpetto: Good suggestions. Will do.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
jpetto: Adding the string above "Sync up with Firefox on mobile" is a little awkward, but I suppose better than not having it.

(I don't know why this QR code layout is different from that of Focus, but we don't really have time to re-open those discussions.)

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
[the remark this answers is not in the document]
jpetto: Yep, we should be fine on locales as RelEng is handling it.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
```diff
@@ -0,0 +1,126 @@
...
+      </div>
+    </div>
+    {#
+      If the user is not signed in to Sync, display the FxA iframe.
+    #}
+    <div class="content-container fxa-container" id="fxa-container">
+      <header class="content-header">
+        <h2>{{ _('Take Firefox with You') }}</h2>
+        <p class="tagline">{{ _('Get your bookmarks, history, passwords and other settings on all your devices.') }}</p>
+        <a href="{{ url('firefox.features.sync') }}" target="_blank" rel="noopener noreferrer">{{_('Learn more about Firefox Accounts')}}</a>
```
alexgibson: Should this nopw point to `url('firefox.accounts.features')`?

media/js/base/mozilla-client.js
```diff
@@ -320,6 +320,71 @@ if (typeof Mozilla === 'undefined') {
...
+            window.clearTimeout(timer);
+            onRetrieved(event.detail.data.setup, event.detail.data.desktopDevices, event.detail.data.mobileDevices, event.detail.data.totalDevices);
+        };
+
+        var onRetrieved = function (setup, desktopDevices, mobileDevices, totalDevices) {
+            document.removeEventListener('mozUITourResponse', listener, false);
+
+            var details = Client.SyncDetails = {
+                'setup': setup,
+                'desktopDevices': desktopDevices,
```
alexgibson: There's always a chance that these values could be undefined, since they are only supported in Firefox 50 and above. Should we fallback to `0` if not supported?

media/js/base/mozilla-client.js
```diff
@@ -320,6 +320,71 @@ if (typeof Mozilla === 'undefined') {
...
+
+            callback(details);
+        };
+
+        // Prepare fallback function in case the API doesn't work
+        var fallback = function () { onRetrieved(false, 0, 0, 0); };
+
+        // If Firefox is old or for Android, call the fallback function immediately because the API is not implemented
+        var userVer = Client._getFirefoxVersion();
+        if (parseFloat(userVer) < 35 || Client._isFirefoxAndroid()) {
```
alexgibson: Iirc, querying `sync` has been supported since Australis, which was Firefox 29. It was only `appinfo` that came later in 35.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
```diff
@@ -0,0 +1,126 @@
...
+{% block body_id %}firefox-whatsnew{% endblock %}
+{% block body_class %}{% endblock %}
+
+{% block site_header %}{% endblock %}
+
+{% block page_css %}
+  {% stylesheet 'firefox_whatsnew_60' %}
+{% endblock %}
+
+{% set show_send_to_device = LANG not in settings.SEND_TO_DEVICE_LOCALES %}
```
alexgibson: Remove `not`? (assume this is left over from debug 😄)

alexgibson:
I'm sure this has been bikeshedded already, but fwiw I find the position of the app store badges underneath the QRCode (like on the Focus layout) makes the copy more readable. It's a little confusing to see them directly underneath "Scan the QRCode..."

app-store-badges

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
jpetto: Ah, yes! I forgot that merged yesterday. Will rebase and update.

media/js/base/mozilla-client.js
jpetto: Ah, sweet. Will update.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
jpetto: Yyyyep, left over from ye olde debugging.

media/js/base/mozilla-client.js
jpetto: Definitely. Will do.

jpetto:
Adding back Do Not Merge label while someone from FxA QA's the iframe (based on pg's suggestion).

jpetto:
@shane-tomlinson Peter (our analytics person) mentioned someone from your team may want to review the FxA iframe in this PR to ensure the proper query params are being sent. Can you comment here letting us know if this is needed, and if so, the results of the review?

Thanks!

jpetto:
rfkelly approved the FxA iframe config in the bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1457024#c41

shane-tomlinson:
@jpetto - yesterday was holiday in the UK, so I didn't get a chance to see this until now.

Can you change `context` from `iframe` to `fx_firstrun_v2`? We added
`fx_firstrun_v2` a while back to give you a couple of additional commands, `form_engaged`
and `navigated` to help your team keep track of those things. Plus, I'd like to
ditch support for `context=iframe` soon so that we have less code to maintain.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx60.html
```diff
@@ -0,0 +1,126 @@
...
+      </div>
+    </div>
+    {#
+      If the user is not signed in to Sync, display the FxA iframe.
+    #}
+    <div class="content-container fxa-container" id="fxa-container">
+      <header class="content-header">
+        <h2>{{ _('Take Firefox with You') }}</h2>
+        <p class="tagline">{{ _('Get your bookmarks, history, passwords and other settings on all your devices.') }}</p>
+        <a href="{{ url('firefox.accounts-features') }}" target="_blank" rel="noopener noreferrer">{{_('Learn more about Firefox Accounts')}}</a>
```
l-hedgehog: I understand this `firefox.accounts-features` page is only localized/enabled for en-US? Is it better to use `firefox.accounts` or `firefox.features.sync` here for other locales?

jpetto: Ah, good catch! I will use `firefox.features.sync` for non en-US locales.

PR #5608, opened 2018-04-30
Fix bug 1457961 - Enable en-CA on production for mozilla.org

flodolo:


stephaniehobson:
@alexgibson Do we need to do anything else for this? Like we did for the en-GB pages?

alexgibson:
> @alexgibson Do we need to do anything else for this? Like we did for the en-GB pages?

Raphael [commented here](https://bugzilla.mozilla.org/show_bug.cgi?id=1457959#c4), but I'm not quite clear on what the resolution is. It seems like this will probly cause some issues in search, but then trying to counter the issue doesn't seem to be proposed either.

Thoughts @rraue?

alexgibson:
We also seem to special case `en-CA` in our canonical hreflang before this change:

https://github.com/mozilla/bedrock/blob/54082f23f902282b56a068f2b5a4a110835e5d44/bedrock/base/templates/includes/canonical-url.html#L14

Maybe we want to change that now `en-CA` is becoming a real locale?

rraue:
@alexgibson We should do what we did for the UK version in the title and the h1 at least on the root page, the firefox page and the new page. These three pages should exist to enable en-ca as a real locale. We need to change the hreflang like you mentioned and en-ca needs to be treated then like any other locale. 

If this will get us into more issues with hreflang we need a solution for ca and uk anyways, so I am not against it. It could be even then a advantage,a s we would be able to a/b test with ca and uk testing difefrnet solutions. But first lets hope for the best and google understanding the setup.

alexgibson:
> We should do what we did for the UK version in the title and the h1 at least on the root page, the firefox page and the new page. These three pages should exist to enable en-ca as a real locale. We need to change the hreflang like you mentioned and en-ca needs to be treated then like any other locale.

@flodolo are you comfortable working on this in this PR, or would you like one of us to do it?

flodolo:
> @flodolo are you comfortable working on this in this PR, or would you like one of us to do it?

Working on it, even if I'm really confused by this piece of code
https://github.com/mozilla/bedrock/blob/7b1d74ed2089f7b83cb232435011bf80fb93632a/bedrock/base/tests/test_urlresolvers.py#L122-L128

I would expect this to fail for en-MX already?

flodolo:
Ignore for now, trying to wrap my head around Django syntax. Will ping as soon as it's ready.

flodolo:
OK, I think I've tested all the pages.

alexgibson:
I think we're just missing the two home page variations:

https://github.com/mozilla/bedrock/blob/673438808cbfeb466394b38fd69e48bc17780bb6/bedrock/mozorg/templates/mozorg/home/home.html#L15-L21

https://github.com/mozilla/bedrock/blob/673438808cbfeb466394b38fd69e48bc17780bb6/bedrock/mozorg/templates/mozorg/home/home-a.html#L13-L19

flodolo:
Thanks, squashed commits

alexgibson:
Thanks @flodolo

PR #5609, opened 2018-04-30
Fix bug 1456502: Add stub attribution data with no data

pmac:
Use "www.mozilla.org" as source and "(none)" as medium when we get no data as a way of marking downloads as having come from the site as opposed to elsewhere.


tests/unit/spec/base/stub-attribution.js
```diff
@@ -179,11 +179,19 @@ describe('stub-attribution.js', function() {
...
+        it('should return empty data if neither utm params and referrer are present', function() {
+            /* eslint-disable camelcase */
+            var data = {
+                utm_source: undefined,
+                utm_medium: undefined,
+                utm_campaign: undefined,
+                utm_content: undefined,
+                referrer: ''
+            };
             spyOn(window._SearchParams.prototype, 'utmParams').and.returnValue({});
```
alexgibson: Nit: `/* eslint-enable camelcase */` after the variable declaration here.

pmac:
@alexgibson updated. Thanks!

PR #5615, opened 2018-05-01
[fix bug 1343823] Redirect old /firstrun URLs to /firefox/new/

alexgibson:
## Description
- Removes legacy `/firstrun/` pages (older than FF 40) and redirects those URLs to `/firefox/new/`
- Removes non-responsive base template from bedrock \o/.
- Removes non-responsive version on Sandstone from bedrock \o/.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1343823

## Testing
- [ ] Redirects work as expected
- [ ] Check nothing else uses the deprecated base template.
- [ ] Ensure no regressions on `/firefox/40.0/firstrun/` (this PR switches the base template that page uses).


bedrock/firefox/views.py
```diff
@@ -343,26 +343,35 @@ def show_57_firstrun(version):
...
 
 class FirstrunView(l10n_utils.LangFilesMixin, TemplateView):
+    def get(self, *args, **kwargs):
+        version = self.kwargs.get('version') or ''
+
+        # redirect legacy /firstrun URLs to /firefox/new/
+        if redirect_old_firstrun(version):
+            return HttpResponsePermanentRedirect(reverse('firefox.new'))
+        else:
+            return super(FirstrunView, self).get(self.request, *args, **kwargs)
```
pmac: Should be `return super(FirstrunView, self).get(*args, **kwargs)`

bedrock/firefox/views.py
```diff
@@ -343,26 +343,35 @@ def show_57_firstrun(version):
     return version >= Version('57.0')
 
 
-def show_40_firstrun(version):
+def show_57_dev_firstrun(version):
+    version = version[:-2]
```
pmac: why only stripping the last two characters? Will it definitely be something like "40.1"? Could be "40.0.1" right? Or is that to get rid of the "a2" or whatever it is?

alexgibson: Hmm, this isn't actually part of this PR, b ut I think you're right yes it is to remove the `a2` part from the Dev Edition version number.

alexgibson:
Thanks @pmac updated

bedrock/firefox/views.py
pmac: Ah! Yes. I missed that the diff was just showing the line being moved. Carry on :)

alexgibson:
Thanks for the reviews - this PR has two approvals, so I'm gonna merge 👍

Issue #5417, opened 2018-01-26
Decouple static asset compilation and bundling from Django


pmac:
Practically speaking this means moving to a system more like what is described in [this blog post](http://owaislone.org/blog/modern-frontends-with-django/) (also [this one](http://owaislone.org/blog/webpack-plus-reactjs-and-django/)), using [Webpack](https://webpack.js.org/), [webpack-bundle-tracker](https://github.com/owais/webpack-bundle-tracker), and [django-webpack-loader](https://github.com/owais/django-webpack-loader/). This will have several advantages over [django-pipeline](https://django-pipeline.readthedocs.io/) which we currently use: 

1. **Modern Tooling:** We'd have easy access to a whole world of NodeJS and Webpack tools and plugins to make our JS and CSS output and developer experience better.
2. **Deployment Efficiency:** If our asset building didn't require our python dependencies they could be built simultaneously and deployed independently if so required.
3. **Developer Familarity:** NodeJS toolchains for front-end web development has simply won. We should provide the modern front-end developer the modern tools with which they're already familiar and proficient. 

Let's talk about how we'd like this to look @mozilla/webdev-bedrock. Also please speak up if you'd like to keep `django-pipeline`. I don't believe people feel that way, but I'm very much willing to be told that this is the wrong direction.

jpetto:
Started [a discussion](https://github.com/orgs/mozilla/teams/webdev-bedrock/discussions/1).

jpetto:
I'm all for this change, but would love to hammer out what the following scenarios would look like:

1. local dev (avoiding bundling for easier debugging)
2. local dev with prod bundles (testing hashes/bundles)
3. demo/stage/prod asset building

After reading the articles posted ([1](http://owaislone.org/blog/modern-frontends-with-django/), [2](http://owaislone.org/blog/webpack-plus-reactjs-and-django/)), it seems 1 & 2 above could be handled relatively easily with separate webpack configs.

Also interested in how we handle bundles in GitHub - do we build locally and submit w/PR, or rely on our push process to handle asset building?

pmac:
> Also interested in how we handle bundles in GitHub - do we build locally and submit w/PR, or rely on our push process to handle asset building?

[...]

PR #5618, opened 2018-05-01
Switch from django-pipeline to gulp tasks for asset processing

pmac:
* Remove django-pipeline
* Do all of the pipeline things in gulp:
  * compile less -> css
  * compile sass -> css
  * concatinate js and css bundles
  * minify js and css
* Also now does:
  * Write sourcemaps in dev mode
  * Add `npm start` that starts Django and gulp watch for file watching and auto-browser-refresh sweetness
* Add a preliminary docker dev workflow
  * run `docker-compose up` and one web and one assets container will spin up and run like the native install

I need to document all of this, but I'm waiting on that until we're happy with this direction and interfaces are finalized.


pmac:
This will fix #5417

pmac:
This is based on my previous work discussed in #5417, but adds bundling back. So this should now have zero end-user impact, but put us in a better position to iterate on our setup and improve developer workflows.

pmac:
Also I left the management commands that output the `static-bundle.json` file and modify all of the templates for you to inspect and to make it easy to update this PR as things change in master. They can be removed following this merging or even just before.

gulpfile.js
```diff
@@ -58,27 +155,64 @@ gulp.task('css:lint', () => {
...
-    });
+gulp.task('reload-other', ['assets'], browserSync.reload);
+gulp.task('reload-css', ['css:compile'], browserSync.reload);
+gulp.task('reload-js', ['js:compile'], browserSync.reload);
+gulp.task('reload', browserSync.reload);
+
+// --------------------------
+// DEV/WATCH TASK
+// --------------------------
+gulp.task('watch', ['browser-sync'], function () {
```
alexgibson: Nit: this function and a few others can turned into shorthand arrow functions `=>`

gulpfile.js
```diff
@@ -58,27 +155,64 @@ gulp.task('css:lint', () => {
...
+        '!node_modules/@mozilla-protocol/core/*'
+    ]).pipe(gulp.dest('static_build'));
+});
 
-    gulp.watch(lintPathsJS).on('change', file => {
-        return gulp.src(file.path)
-            .pipe(eslint())
-            .pipe(eslint.format());
+gulp.task('browser-sync', ['js:compile', 'css:compile'], () => {
+    var proxyURL = process.env.BS_PROXY_URL || 'localhost:8000';
```
alexgibson: Couple of `var`'s here (and elsewhere) that can be `const`?

pmac:
> Are we going to switch to testing asset compilation in circleCI as part of the test_js job, or is that still handled elsewhere?

I'm not sure we're testing asset compilation in circle at all right now, but that's a good point regardless. I'll make sure the asset docker build happens in CI just to ensure it happens without error.

pmac:
@alexgibson actually we do build the assets image as part of the "build_images.sh" script run, so if that fails it should fail and fail the build.

pmac:
And I've now updated based on your suggestions.

pmac:
Clearly I need to squash, but I'm not doing that yet in case we need to do another major change in how we handle the template bundles. Other than that I think this may be ready to go. I've updated the install docs and improved the code and readability in the gulpfile.

docs/install.rst
```diff
@@ -4,20 +4,46 @@
...
+.. note::
+
+    This method assumes you have `Docker installed for your platform <https://www.docker.com/community-edition#/download>`_.
+    If not please do that now or skip to the ``Local Installation`` section.
+
+This is the simplest way to get started developing for bedrock. If you're on Linux or Mac (and possibly Windows 10 with the
+Linux subsystem) you can run a script that will pull our production docker images and use those to build developer ones and
+start the servers you need to serve bedrock locally::
+
+    $ bin/docker-start.sh
```
alexgibson: I ran this locally and all the build part worked, but when I go to view the home page locally I get the following DB error: https://gist.github.com/alexgibson/54f7ff0b053dfdd898d1e7b697dfcd25

Note: other pages that don't have a newsletter such as `/firefox` work as expected.

docs/install.rst
```diff
@@ -67,16 +86,23 @@ You may also need to install the `Gulp <http://gulpjs.com/>`_ cli globally::
...
 This will be important during development so that you can easily know when
-you've broken something with a change. You should still have your virtualenv
+you've broken something with a change.
+
+Docker
+------
+
+We manage our local docker environment with docker-compose. All you need to do here is run::
+
+    $ docker-compose run test
```
alexgibson: Note for other reviewers: I initially got a bunch of errors when running the tests using this command, and needed to clean out my `.pyc` files using `find . -name \*.pyc -delete`. After doing this the tests passed OK.

gulpfile.js
```diff
@@ -58,27 +196,62 @@ gulp.task('css:lint', () => {
...
-                reporters: [{
-                    formatter: 'string',
-                    console: true
-                }]
-            }));
-    });
+    // --------------------------
+    // watch:css, less, and sass
+    // --------------------------
+    gulp.watch('media/css/**/*', ['reload-css']);
```
alexgibson: One other thing I notice is when changing a scss/less file, or a JS file - the watch command seems to be recompiling everything. This means it takes around 6 seconds to see the changes (on docker at least). It might be worth only compiling the files that have changed here?

alexgibson:
I wouldn't call source maps a blokcer, but when viewing the style editor in Dev Tools I see some errors e.g. on /firefox/

https://gist.github.com/alexgibson/0b4017cd57e926b622014d8282cd0048

docs/install.rst
pmac: I believe this is due to your local database file not being up-to-date. I believe most users of the docker method won't have any local db, but it's possible. I'll see what I can do.

docs/install.rst
pmac: This is similar to the database issue in that I don't think may people would have .pyc files who are using docker, but it is possible. I can maybe put a .pyc deletion step in a script that this runs.

pmac:
> I wouldn't call source maps a blocker, but when viewing the style editor in Dev Tools I see some errors e.g. on /firefox/

Ah! The URLs are wrong for the maps... hmm... I did have it set to include the source in the css file, perhaps that's the easiest way to go.

gulpfile.js
jpetto: FWIW, running locally/outside of Docker, this takes 10 seconds to re-compile the CSS and reload the browser on my machine. 😞

jpetto: Also FWIW, back when I was on MDN, we were able to avoid re-compiling everything because some smart person (probs @stephaniehobson) organized the CSS so that all includes/non page CSS files were kept in a separate folder. This meant the re-compile only had to target the "higher level" `.css/.scss/.less` files (as recompiling them would infer a recompile of any `.css/.scss/.less` files they `@import`'ed).

Even if we did feel like re-architecting our CSS folder structure, the sheer size of moz.org may prevent it from being a huge boost.

I'm not saying we should stick with Django Pipeline, but one nice thing it granted was only recompiling the files of the one page being loaded.

alexgibson: I think it should be possible to only compile the changed file here. Even if we can’t figure out auto compilation of imported files right away, it still might be preferable to endure rather than a really long compile time.

jpetto: IIRC, the issue with only compiling the changed file is if that changed file is a library/include. The files including that library/include wouldn't get recompiled, so you wouldn't see the change reflected on any pages until the including files were also modified/recompiled. (I think this is basically what pmac was getting at above.)

alexgibson: Yup, I get the problem. I seem to remember we had this same issue a while back, before we switched to Django Pipeline.

alexgibson: There seem to be some solutions out there possibly. This is the first one I found but there seems to be a few others as well: https://github.com/berstend/gulp-sass-inheritance

pmac: @jpetto I'm actually already only compiling the files used in pages, since I'm only compiling those less and sass files referenced in bundles.

pmac: It's just that we have kindof a ton of those :)

pmac: I also had it working a bit better before adding bundling since I could just recompile less or sass individually, but now I have to do the bundling step as well before the page will work again. I could potentially figure out to which bundle(s) a file belonged and just rebuild it. We could also just document that changing a file not directly included in a bundle will require the user to call `gulp css:compile` for the whole lot. Better?

jpetto: If working on an include/library file, the need to run any gulp task with *every file change* would be, IMO, a deal-breaking inconvenience. CSS dev is often many rapid changes and reloads. An extra command in that chain would be really expensive.

If we want to stick to compiling pre-page request, I think we need to find some lib like Alex linked above, or do something like Paul mentioned above, where all and only the set of files impacted by the actual file changed are recompiled.

pmac: I'm going to explore these things alex brought up, mostly the `gulp-cached` and then the `gulp-sass-inheritance`. I think `gulp-cached` will get us a long way actually.

pmac:
@alexgibson @jpetto  okay... I've done a bunch of work on this and I think it's a whole lot better. Give it a spin and let me know.

pmac:
Note the docs are updated. tl;dr is:

## Docker

```bash
$ bin/docker-bootstrap.sh
$ bin/docker-start.sh
```

## Local

```bash
$ npm start
```

pmac:
I've changed docs and startup procedures again btw. :) Should no longer be in flux (I hope). New tl;dr:

## Docker

```bash
$ make clean build
$ make run
```

Though you really only need the 2nd one since it will run `make build` for you if it needs to.

## Local

This is the same, but here it is again for completeness. 

```bash
$ npm start
```

Makefile
```diff
@@ -1,80 +1,99 @@
...
 	# test related things
 	-rm -f .coverage
-	-rm -rf results
-
-	# static files
-	git clean -fdx static
-
 	# docs files
 	-rm -rf docs/_build/
-
```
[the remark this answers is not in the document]
pmac: Interesting. I did not know that. Great catch.

docs/install.rst
```diff
@@ -95,38 +143,41 @@ To test a single app, specify the app by name in the command above. e.g.::
...
+    $ npm start
 
-If you have problems with gulp, or you for some reason don't want to use it you can set::
+If you are not inside a virtualenv, you can activate it by doing::
 
-    PIPELINE_COLLECTOR_ENABLED=True
+    $ source venv/bin/activate
 
-in your ``.env`` file or otherwise set it in your environment and it will collect media for you as you make changes. The reason that this is not the preferred method is that it is much slower than using gulp.
+Browsersync
```
pmac: @bensternthal I added this section about browsersync and how it works and potential issues. Do you think this is sufficient?

pmac:
This has been rebased and cleaned up and squashed. It's about as ready as it'll get I think. I'll follow this PR up with a cleanup one to remove the management commands I added to help me keep this updated as well as the old static asset bundle definitions.

gulpfile.js
```diff
@@ -2,52 +2,296 @@
...
+        })))
+        .pipe(gulp.dest(buildDir));
+});
+
+/**
+ * Watch and only compile those SASS files that have changed
+ */
+gulp.task('sass:watch', () => {
+    return watch(buildDir + '/css/**/*.scss', {base: buildDir})
+        // filter out internal imports (files starting with "_" )
```
alexgibson: From my local testing, this seems to have a side effect in that changes to imported scss files aren't currently picked up at all. The only way I can visibly see a change to an imported style sheet (e.g. `_sub_nav.scss`) is to also force a change in the parent stylesheet (e.g. `firefox/home.scss`). I know this is a somewhat tricky thing we've discessed earlier in this pull request, but this is going to be somewhat of a pain point for front-end dev :(

gulpfile.js
```diff
@@ -2,52 +2,296 @@
...
+        .pipe(gulpif(global.watching, cached('all', cachedOpts)))
+        .pipe(gulp.dest(buildDir));
+});
+
+/**
+ * Find all SASS files from bundles in the static_build directory and compile them.
+ */
+gulp.task('sass', ['assets'], () => {
+    return gulp.src(allBundleFiles('css', '.scss'), {base: buildDir, cwd: buildDir})
+        //filter out unchanged scss files, only works when watching
```
alexgibson: This seems much improved 👍

gulpfile.js
pmac: Agreed. It's definitely a missing piece. I tried gulp-sass-inheritance but couldn't get it working. Only thing I can think for now is to add a watcher specifically for the `_*.scss` files that will rebuild all of the bundles. It's not very efficient, but it would at least not force a dev to quit everything and restart. To be clear all you'd need to do is restart everything since it does a full build at start, but that's a pain as you said.

jpetto: I would go a little stronger than "somewhat of a pain point". 😝 IMO, this is a blocker for front-end dev.

I think it's definitely worth trying a separate watcher for `.scss` files prefixed with `_`. We'll see what performance is like rebuilding all bundles that way.

pmac: @alexgibson @jpetto I've just pushed an update that will rebuild all SASS bundles when a library file (one that starts with `_`) is changed. This isn't optimal, but does work and isn't too slow in mytesting. Let me know what you think.

pmac: @alexgibson I've just rebased this and pushed my demo again. I think this is ready for final review.

pmac:
@jgmize I've updated `set_git_env_vars.sh` to include a couple of new variables we discussed. Take a look if you'd like.

gulpfile.js
alexgibson: Agree this isn't optimal, but it does work and i don't think it should interfere with workflow to the extent this should be a blocker. I'd like to get the other front-end developers to kick the tyres of this PR though.

gulpfile.js
```diff
@@ -2,52 +2,307 @@
...
+/**
+ * Combine the CSS files after SASS/LESS compilation into bundles
+ * based on definitions in the `static-bundles.json` file.
+ */
+gulp.task('css:compile', ['sass', 'less'], () => {
+    return merge(staticBundles.css.map(bundleCssFiles));
+});
+
+
+gulp.task('css:watch', () => {
```
alexgibson: Nit: would be nice to have a comment for this task?

alexgibson:
Oh I forgot to mention - I did some manual testing on legacy browsers (IE 6/7/8) and could download Firefox sucessfully. There was a JS error in the console, but I believe this was the browser sync plugin when running locally. It would be good to push this to a demo server just to verify I'm correct here.

pmac:
Thanks Alex! Deploying [the demo](https://bedrock-demo-pipelinefree.oregon-b.moz.works/en-US/) now with the latest version and [running all integration tests](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/demo%2fpipelinefree/42/pipeline/).

Note: You can turn off the notification that pops up on reload in the gulpfile:

https://github.com/mozilla/bedrock/blob/8de3d812404766cc40cbd31c5d1f50dd3cd5a794/gulpfile.js#L338

Not sure if that would eliminate the JS error locally, but it could help should a CSS conflict arise someday.

pmac:
> P.S I'm particularly looking forward to the docker workflow, super exciting!

Me too! Pretty psyched with how well that's working, and hoping we can replicate to some other projects as well.

pmac:
Demo is freshly deployed and [all tests passed](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/demo%2fpipelinefree/42/pipeline/). Feel free to test all of the things.

https://bedrock-demo-pipelinefree.oregon-b.moz.works/en-US/

pmac:
@craigcook @stephaniehobson hi! I believe this is ready to go (minus one more rebase). Can you please find some time this week to evaluate the new local dev processes herein and let us know if you're okay with us merging this? It's blocking some other work and we'd like to get it in before the all-hands if possible, but we want to be sure that the new workflow makes sense to everyone. This PR includes updated docs that should hopefully get you going quickly with the differences.

docs/install.rst
```diff
@@ -67,23 +106,32 @@ You may also need to install the `Gulp <http://gulpjs.com/>`_ cli globally::
...
 This will be important during development so that you can easily know when
-you've broken something with a change. You should still have your virtualenv
+you've broken something with a change.
+
+Docker
+------
+
+We manage our local docker environment with docker-compose and Make. All you need to do here is run::
+
+    $ make test
```
jpetto: Is it possible to test a single app this way?

jpetto:
One more 👍 from me.

jgmize:
If there are no objections, I'd like to merge this tomorrow morning. See https://github.com/mozilla/bedrock/issues/5762#issuecomment-395235593 for context.

craigcook:
I haven't done a deep dive into all the code changes here but I just ran through the install process and everything went smoothly and the site is running as it should, no DB errors for me either. So that's a 👍 for me, for whatever it's worth.

docs/install.rst
stephaniehobson: Pls add this to the docs if it's not there already.

stephaniehobson:
Hm. Not just my massive Trash causing problems this time.

I upped the Docker allocation to 128 GB and I get the same error. Sorry I left this until everyone had gone home to test :(

jwhitlock:
@stephaniehobson that sounds a bit like https://stackoverflow.com/a/36584376/10612 . Try ``docker system prune``. And maybe wipe out any kuma images you are keeping around 🙁

stephaniehobson:
/me scoops kuma images protectively into her arms

PR #5619, opened 2018-05-02
Fix Bug 1458389: Added additional portland variants.

bensternthal:
## Description
This PR enables the following new URLs to render the 3 portland variations:

https://www.mozilla.org/en-US/firefox/new/?xv=forgood
https://www.mozilla.org/en-US/firefox/new/?xv=safe
https://www.mozilla.org/en-US/firefox/new/?xv=fast

New adjust links were also added for each.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1458389

## Testing
The 3 urls should all render the same content as their existing _siblings_:
portland=forgood
portland-safe=safe
portland-fast=fast

A few things in this PR to pay attention to.

1. I rarely code these days so ...
2. Because the urls provided by fabio were not great (example "fast") I did my best with the classnames for handling adjust and avoiding any possible conflict. 
3. I am setting a var in context to accomplish the above.. im not 100% sure about the previous usage of context here so take a close look.
4. If we are doing more variations like this we should think about a better implementation.

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -505,11 +505,11 @@ def download_thanks(request):
     elif locale == 'de' and experience == 'berlin':
         template = 'firefox/new/berlin/scene2.html'
     elif locale == 'en-US':
-        if experience == 'portland':
+        if experience in ['portland', 'forgood']:
             template = 'firefox/new/portland/scene2.html'
-        elif experience == 'portland-fast':
+        elif experience in ['portland-fast', 'fast']:
             template = 'firefox/new/portland/scene2-fast.html'
-        elif experience == 'portland-safe':
+        elif experience in ['portland-safe', 'safe']:
             template = 'firefox/new/portland/scene2-safe.html'
         else:
             template = 'firefox/new/scene2.html'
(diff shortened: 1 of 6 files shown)


bedrock/firefox/views.py
```diff
@@ -505,11 +505,11 @@ def download_thanks(request):
     elif locale == 'de' and experience == 'berlin':
         template = 'firefox/new/berlin/scene2.html'
     elif locale == 'en-US':
-        if experience == 'portland':
+        if experience == 'portland' or experience == 'forgood':
```
jpetto: Would be better as `experience in ['portland', 'forgood']`.

My brain was tired and distracted when I sent you in the `or` direction. Apologies. 😞

bedrock/firefox/views.py
```diff
@@ -555,14 +557,23 @@ def new(request):
...
+                template = 'firefox/new/portland/scene1-safe.html'
             else:
                 template = 'firefox/new/scene1.html'
         else:
             template = 'firefox/new/scene1.html'
 
     # no harm done by passing 'v' to template, even when no experiment is running
     # (also makes tests easier to maintain by always sending a context)
-    return l10n_utils.render(request, template, {'v': variant})
+    return l10n_utils.render(request, template, context)
```
jpetto: We could be a bit cleaner by passing the pre-defined `experience` in `context`:

`return l10n_utils.render(request, template, {'experience': experience})`

This would eliminate the rather repetitive calls if the `elif` chain, and give us consistency in the JS selectors (which would need updating along with the variable name in the templates).

bedrock/firefox/views.py
```diff
@@ -533,6 +533,8 @@ def new(request):
     if variant not in []:  # place expected ?v= values in this list
         variant = None
 
+    context = {'v': variant}
```
jpetto: Don't think this is needed.

bensternthal:
@jpetto ok i think i got everything.. if this looks good I'd like to get this into a push today. There is an ad campaign on hold until this goes out... appreciate the help with the pr.

PR #5675, opened 2018-05-10
Update/remove the LXR redirects

sylvestre:


alexgibson:
@sylvestre looks like tests are still failing here

PR #5692, opened 2018-05-14
Bug 1456355: Sticky download cta

stephaniehobson:
## Description
- Update look of header and footer download CTA areas
- Update first row of content to make up for CTA not being there anymore
- Add sticky download banner
  - Dismissable
  - Does not return on page reload
  - Does not appear with JS disabled because it can't accommodate 6 buttons.
- Waypoints added to improve accessibility
- Remove download button from global nav

## L10n
- Just a/b testing in en-US only

## Issue / Bugzilla link
[Bug 1456355 - [mozilla] - Homepage updates - April 30th](https://bugzilla.mozilla.org/show_bug.cgi?id=1456355)

## Testing
- I'm nervous about the possibility of this conflicting with the global
  nav in some way.
- Check analytics on the download buttons (especially the sticky one)
- Test with JS disabled
- Test in Windows
- mobile, tablet, desktop, and wide desktop
- Dismiss button on sticky download
  - appears
  - dismisses the sticky banner
  - banner does not return on page refresh (unless cookies disabled)
  - no JS errors if cookies are disabled


stephaniehobson:
Note that there's a known problem with the blue banner showing on extra wide desktop monitors.

stephaniehobson:
The wide screen bug is fixed.

media/css/mozorg/home/home-2018.scss
```diff
@@ -204,73 +289,86 @@ main {
...
-        }
+    .download-firefox-secondary-cta {
+        min-height: 460px;
 
-        .secondary-image {
-            float: right;
-            height: 477px;
-            width: 48%;
+        &:before {
+            background: url('/media/img/home/2018/browser.png') top left / contain no-repeat;
```
jpetto: Could use the `at2x` helper here?

media/js/mozorg/home/home.js
```diff
@@ -66,4 +70,66 @@
...
     initVideoEvents();
-})();
+
+    /*
+     * Sticky CTA
+     */
+
+
+    var stickyCTA = document.getElementById('download-firefox-sticky-cta');
+    $(stickyCTA).attr('aria-hidden', 'true');
```
jpetto: Could just do the jQuery-fication once to avoid multiple `$` wrappings/DOM lookups:

`var $stickyCTA = $('#download-firefox-sticky-cta');`

media/js/mozorg/home/home.js
```diff
@@ -66,4 +70,66 @@
...
+    });
+
+    // init dismiss button
+    function initDismissStickyCTA() {
+        // add button
+        var $dismissButton = $('<button>').addClass('sticky-dismiss').text('Dismiss download prompt.');
+        var $stickyWrapper = $(stickyCTA).find('.primary-wrapper');
+        $dismissButton.appendTo($stickyWrapper);
+        // listen for click
+        $dismissButton.on('click', function(){
```
jpetto: Could use the `.one` method here, as this element will only ever be clicked once.

media/js/mozorg/home/home.js
```diff
@@ -66,4 +70,66 @@
...
+            }
+        }
+    });
+
+    // init dismiss button
+    function initDismissStickyCTA() {
+        // add button
+        var $dismissButton = $('<button>').addClass('sticky-dismiss').text('Dismiss download prompt.');
+        var $stickyWrapper = $(stickyCTA).find('.primary-wrapper');
+        $dismissButton.appendTo($stickyWrapper);
```
jpetto: Just curious - why not add this element directly in the HTML?

media/js/mozorg/home/home.js
```diff
@@ -66,4 +70,66 @@
...
+            var d = new Date();
+            d.setTime(d.getTime() + (30 * 24 * 60 * 60 * 1000)); // 30 days
+            Mozilla.Cookies.setItem('sticky-home-cta-dismissed', 'true', d, '/');
+        }
+    }
+
+    // check if previously dismissed
+    // Check that cookies are enabled before seeing if one already exists.
+    if (typeof Mozilla.Cookies !== 'undefined' && Mozilla.Cookies.enabled()) {
+        if (Mozilla.Cookies.getItem('sticky-home-cta-dismissed')) {
```
jpetto: Not a huge deal, but would be a bit more performant to do the inverse here - if the CTA was _not_ previously dismissed, set up the Waypoint and `aria-` stuff, as well as `initDismissStickyCTA()`. (This would prevent the Waypoint from being generated and immediately torn down in the event of a prior dismissal.)

jpetto:
I have fixed my VMs and things are looking good in IE 9-Edge. 😀

bedrock/mozorg/templates/mozorg/home/home-a.html
```diff
@@ -187,7 +190,7 @@ <h3 class="primary-title">{{ _('Fast for good.') }}</h3>
           link_url='https://blog.mozilla.org/firefox/switch-chrome-firefox-just-minutes/?utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=homepage&utm_content=v-a'
       )}}
     </div>
-
+-->
```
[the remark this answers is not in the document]
stephaniehobson: I don't know? It'd be annoying to have to copy it into the correct format again and we're still iterating on the design. If the right person does the next iteration they'll know to get it from Git history though.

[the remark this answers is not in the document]
stephaniehobson: I get the feeling you're leaning towards removing it ;)

media/js/mozorg/home/home.js
stephaniehobson: If an element is only functional with JS, I try to avoid having it in the HTML, then I don't have to write CSS to hide it if the JS is not functional.

I'm happy to do it the moz.org way if there is a moz.org way though.

media/css/mozorg/home/home-2018.scss
jpetto: Iiiinteresting, and good to know.

media/js/mozorg/home/home.js
```diff
@@ -66,4 +70,68 @@
...
+        if (!Mozilla.Cookies.getItem('sticky-home-cta-dismissed')) {
+            // init the button
+            initStickyCTA();
+        } else {
+            $stickyCTA.remove();
+        }
+    } else {
+        // they can still dismiss it and get ARIA with cookies disabled
+        initStickyCTA();
+    }
```
jpetto: I'm of the mind that we shouldn't show the banner if dismissing it won't be remembered. I could be persuaded otherwise though.

(Maybe just don't show the "x" if `hasCookies === false`? Though that may be more work than just never showing the banner to begin with.)

media/js/mozorg/home/home.js
```diff
@@ -66,4 +70,68 @@
...
+        $dismissButton.appendTo($stickyWrapper);
+        // listen for click
+        $dismissButton.one('click', function(){
+            // dismiss
+            dismissStickyCTA(primaryTop);
+        });
+    }
+
+    // handle dismiss
+    function dismissStickyCTA(wayPoint) {
```
jpetto: The variable name `wayPoint` scares me a little because it's so close to conflicting with `Waypoint`. Not critical to change for this test, but 😬

media/js/mozorg/home/home.js
```diff
@@ -66,4 +70,68 @@
...
+                    $stickyCTA.removeAttr('aria-hidden');
+                } else {
+                    // hidden again as they scroll up
+                    $stickyCTA.attr('aria-hidden', 'true');
+                }
+            }
+        });
+
+        // add button
+        var $dismissButton = $('<button>').addClass('sticky-dismiss').text('Dismiss download prompt.');
```
jpetto: Don't have a problem with keeping this in JS in this instance, but if/when a string needs to be localized, definitely easier to put the markup directly in the `.html` file.

media/js/mozorg/home/home.js
```diff
@@ -66,4 +70,68 @@
...
     // Video card interactions.
     initVideoEvents();
-})();
+
+    /*
+     * Sticky CTA
+     */
+
+
+    var $stickyCTA = $(document.getElementById('download-firefox-sticky-cta'));
```
jpetto: Nit pick - could just do `$('#download-firefox-sticky-cta');`

stephaniehobson: I believe this is faster but I can't find a citation for that.

media/js/mozorg/home/home.js
stephaniehobson: I'm fine not showing it.

media/js/mozorg/home/home.js
stephaniehobson: 👍

media/js/mozorg/home/home.js
jpetto: I know `getElementById` is faster than `querySelector`, and both are faster than jQuery's `$` lookup, but I don't know about the `$` `+` `getElementById` combo.

PR #5695, opened 2018-05-15
[fix bug 1458036] Copy updates to /firefox/enterprise/ page

alexgibson:
## Description
- Updates `/firefox/enterprise/` page with links to both general release and ESR downloads.
- Page is not yet localized, so copy changes should be fine as-is.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1458036

## Testing
- [ ] Check for copy pasta errors
- [ ] Test responsiveness
- [ ] No regressions to existing content

Demo:
-------

https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/enterprise/

PR #5698, opened 2018-05-15
[bug 1458036] Clean up Enterprise switch

craigcook:
## Description
https://github.com/mozilla/bedrock/pull/5695 updated the tests for the Firefox Enterprise page but the old CTA was still visible and the new one was behind a switch, which meant the test looking for the new button wasn't finding it. Since that switch is no longer needed the ideal fix for the breaking test is to just remove the switch, making the new button visible for all.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1458036

### DIFF
--- a/bedrock/firefox/templates/firefox/organizations/organizations.html
+++ b/bedrock/firefox/templates/firefox/organizations/organizations.html
@@ -27,7 +27,6 @@
 {% block content %}
 
 {% block main_feature %}
-  {% if switch('firefox_enterprise', ['en-US']) %}
     <div class="pick">
       <h1 class="large">{{ _('Find the right Firefox for you') }}</h1>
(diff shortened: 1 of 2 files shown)


bedrock/firefox/templates/firefox/organizations/organizations.html
```diff
@@ -47,48 +46,7 @@ <h1 class="large">{{ _('Find the right Firefox for you') }}</h1>
         </div>
       </div>
     </div>
-  {% else %}
-    <hgroup id="main-feature" class="stacked">
```
craigcook: HTML5!

bensternthal:
Closes https://github.com/mozilla/bedrock/issues/5693

PR #5709, opened 2018-05-16
Fix default setting for Newrelic INI File in wsgi/app.py

pmac:
This was causing the app container to fail to start since New Relic was trying to start with a non existing config file called "False", thus causing the smoke tests to file since the app container did not stay running.

### DIFF
--- a/wsgi/app.py
+++ b/wsgi/app.py
@@ -11,7 +11,7 @@ except ImportError:
 
 
 if newrelic:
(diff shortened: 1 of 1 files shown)


pmac:
See https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/master/1019/pipeline/44 for issue evidence.

Issue #5629, opened 2018-05-03
iOS Adjust Flow Fix


bensternthal:
[Description of issue](https://docs.google.com/document/d/1Cf7860EXtW4q4BFFKYpwkEZWE4hV44CgZdIKTp19G04/edit)

ejregithub:
Met with Aaron/Jon/Devyani to clarify what can be done.  Jon is unblocked, able to create the fallback URL that will eliminate the 'no where state' when clicking cancel and to use js to automagically forward the user to the appstore.

alexgibson:
Closing old issue

PR #5724, opened 2018-05-21
[fix issue 5629] Improve adjust iOS flow.

jpetto:
## Description

Improve UX for iOS users following adjust.com links to download Firefox mobile. 

## Issue / Bugzilla link

#5629
https://bugzilla.mozilla.org/show_bug.cgi?id=1460039

## Testing

https://app.adjust.com/2uo1qc?campaign=whats_new&adgroup=send_to_device&creative=sms&fallback_lp=https%3A%2F%2Fwww-demo2.allizom.org%2Ffirefox%2Fnew%2F%3Fopen-app-store%3Dtrue

Both email and SMS the above link to yourself. Open the link on iOS, both in your mail client and messaging client.

You *should* see our `/firefox/new/` page displayed and immediately be given an iOS (not website) prompt to open the App Store. Accepting the dialog should take you to the App Store, while cancelling should leave you on `/firefox/new/.

(Note that there is a known [issue going from the Gmail iOS app to the Chrome iOS browser](https://bugzilla.mozilla.org/show_bug.cgi?id=1460039#c5). Someone from Adjust is looking at this, but I don't think we need to block on it.)

### DIFF
--- a/media/js/firefox/new/scene1.js
+++ b/media/js/firefox/new/scene1.js
@@ -120,4 +120,19 @@
         initOtherPlatformsModal();
     }
 
+    /**
+     * If user is on iOS and has followed an adjust.com link, redirect to
+     * the App Store.
+     * https://github.com/mozilla/bedrock/issues/5629
+     */
+    // TODO: is this a solid enough check?
+    if (window.site.platform === 'ios' && /open\-app\-store=true/.test(window.location.search)) {
+        var appStoreURL = $('.os_ios a').attr('href');
+
+        if (appStoreURL) {
+            // TODO: do we need a dataLayer call here?
+            window.location.href = appStoreURL;
+        }
+    }
+
 })(window.jQuery);
--- a/media/js/firefox/new/scene1.js
+++ b/media/js/firefox/new/scene1.js
@@ -125,12 +125,10 @@
      * the App Store.
      * https://github.com/mozilla/bedrock/issues/5629
      */
-    // TODO: is this a solid enough check?
     if (window.site.platform === 'ios' && /open\-app\-store=true/.test(window.location.search)) {
-        var appStoreURL = $('.os_ios a').attr('href');
+        var appStoreURL = $('.download-button .os_ios > a.download-link').attr('href');
 
         if (appStoreURL) {
-            // TODO: do we need a dataLayer call here?
             window.location.href = appStoreURL;
         }
     }


jpetto:
We are still waiting for additional testing to make sure this approach fixes the issue [1], but in the mean time would like confirmation of approach. The code change is fairly simple - did I miss any pitfalls or potentially introduce regressions?

[1] The flow worked as expected for Stephanie and I.

bensternthal:
Jon - I would say dont wait for me to test. If this works for you and Stephanie.. ship it!

media/js/firefox/new/scene1.js
```diff
@@ -120,4 +120,19 @@
...
     }
 
+    /**
+     * If user is on iOS and has followed an adjust.com link, redirect to
+     * the App Store.
+     * https://github.com/mozilla/bedrock/issues/5629
+     */
+    // TODO: is this a solid enough check?
+    if (window.site.platform === 'ios' && /open\-app\-store=true/.test(window.location.search)) {
+        var appStoreURL = $('.os_ios a').attr('href');
```
alexgibson: Maybe make this selector a little more specific just for good measure? `.download-button .ios_ios > .download-link`

media/js/firefox/new/scene1.js
```diff
@@ -120,4 +120,19 @@
...
+
+        if (appStoreURL) {
+            // TODO: do we need a dataLayer call here?
```
alexgibson: This could be a good idea if it may give us an idea of the drop-off rate. Could we attach some query params to the app store link maybe to attribute the referral?

media/js/firefox/new/scene1.js
```diff
@@ -120,4 +120,19 @@
         initOtherPlatformsModal();
     }
 
+    /**
+     * If user is on iOS and has followed an adjust.com link, redirect to
+     * the App Store.
+     * https://github.com/mozilla/bedrock/issues/5629
+     */
+    // TODO: is this a solid enough check?
```
alexgibson: The only other thing I can think to do here would be to also check the referrer came from adjust.com. It does worry me a little that this kind of auto-redirect thing could one day get us penalized in search rank. If someone shared the link or published / tweeted it, could it get picked up and indexed?

media/js/firefox/new/scene1.js
jpetto: I don't _think_ we need any extra query params here, as adjust will know (with pretty high accuracy) which adjust.com link the visitor is coming from.

media/js/firefox/new/scene1.js
jpetto: Hmmm, this is a very valid concern.

@rraue - do you have any thoughts here? For context, someone will land on `/firefox/new/?open-app-store=true` and be automatically redirected to the Apple App store. If this URL gets shared somehow, would/could there be a negative SEO impact?

jpetto: Unfortunately, (according to Aaron at Adjust), we cannot rely on `adjust.com` being present in `document.referrer`. 😞

alexgibson: It'd still be nice to try and get @rraue's thoughts here re and potential SEO impact?

jpetto:
> Another thought that occured to me, was if we could just set the adjust callback URL to /download/thanks/ and be done with any further code? (the scene2 page would try and auto redirect to the app store anyay, right?)

Thank you for being outside of the forest enough to see the trees. Yes, this is a much better solution that doesn't require any code changes. 🤦‍♂️ 

I'll update the bug with this recommendation.

PR #5727, opened 2018-05-22
Add more data to git repo state in the database

### LINKED ISSUE #5726: Add info on external-git repo current versions to /healthz-cron/
We import data from various external git repositories in bedrock (product-details, release-notes, security-advisories, feature-switches, etc.), and it would therefore be nice for an information page to include the names and currently synced git hashes of said repos for informational and debugging purposes.

The current git hash of the various repos is in the database in the `GitRepoState` model.

pmac:
This is so that we can display the current commit synced to the db for each external repo on the /healthz-cron/ page. We also link to the repo, the commit, and display how long ago the repo was last updated.

Fix #5726

### DIFF
--- a/bedrock/base/management/commands/update_www_config.py
+++ b/bedrock/base/management/commands/update_www_config.py
@@ -49,7 +49,7 @@ class Command(BaseCommand):
     def handle(self, *args, **options):
         self.quiet = options['quiet']
         repo = GitRepo(settings.WWW_CONFIG_PATH, settings.WWW_CONFIG_REPO,
-                       branch_name=settings.WWW_CONFIG_BRANCH)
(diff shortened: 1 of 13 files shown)


pmac:
Note: The external repo definitions won't show up in the list until they're updated after this lands. This is due to them needing to add some data (name, last update, url) to their entries in the DB, and this doesn't happen until an update. I'll put a demo up shortly so people can see the page updates.

pmac:
Here are the changes:

https://bedrock-demo-pmac.oregon-b.moz.works/healthz-cron/

It is expected that the `download_database` entry is red and failing because I have that instance set to do local DB updates and not download the one from dev since it has schema changes.

Issue #5730, opened 2018-05-23
MOSS content updates


ejregithub:
https://www.mozilla.org/en-US/moss/

- [ ] Remove Header Photo
- [ ] Replace URL on “Apply Now” Buttons to Fluxx on the following pages:
- https://www.mozilla.org/en-US/moss/ 
- https://www.mozilla.org/en-US/moss/foundational-technology/ 
- https://www.mozilla.org/en-US/moss/mission-partners/ 
- [ ] Change the application cycle to monthly (from quarterly)
- [ ] Add the MoFo/MoCo Disclosure language

alexgibson:
The header image removal in this issue is being handled by a contributor https://github.com/mozilla/bedrock/pull/5667

ejregithub:
I've asked for clarity as to when these changes can be made (if we can do them before June 1st or if they must be on that specific day).  We're meeting with the requesting team on March 29th to confirm the asks and to look at those for later dates.

ejregithub:
I've confirmed from Mehan at MoFo that the other content update requests need to wait til June 1st for go live, so let's put them behind a switch.

PR #5743, opened 2018-05-25
[fix issue 5730] MOSS content updates for June 1

ejregithub:
As stated this is for issue #5730.
It includes updates to the links to apply (Apply now buttons), disclosure copy and and change to the copy on frequency of evaluations.

The Do Not Merge tag has been applied due to the fact that the content should go live on June 1st (anytime) and not before

### DIFF
--- a/bedrock/mozorg/templates/mozorg/moss/index.html
+++ b/bedrock/mozorg/templates/mozorg/moss/index.html
@@ -50,12 +50,11 @@
   {% endtrans %}
   </p>
 
-    <p class="moss-copy">
-  {% trans %}
-    MOSS is an awards program of the Mozilla Corporation, which is ultimately responsible for selecting and
-    funding final award recipients. MOSS is administered in part by the Mozilla Foundation, the nonprofit owner of the Mozilla Corporation.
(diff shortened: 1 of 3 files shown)


bedrock/mozorg/templates/mozorg/moss/index.html
```diff
@@ -48,6 +48,13 @@ <h2 class="moss-section-heading">{{ _('About MOSS') }}</h2>
     resources on a vibrant regional open source community - in this case,
     in India.
   {% endtrans %}
+  </p>
+
+    <p class="moss-copy">
+  {% trans %}
+    MOSS is an awards program of the Mozilla Corporation, which is ultimately responsible for selecting and
```
stephaniehobson: Because we are adding new text to a page that has already been translated we need to wrap the new content in a `l10n_has_tag`. Make sure you include the `<p>` tags inside the if statement.

 There's a [code example in the docs](http://bedrock.readthedocs.io/en/latest/l10n.html#staging-copy-changes).

Let me know if that looks like too much coding too ;)

PR #5760, opened 2018-05-30
Add base template for Protocol design system (Fixes #5752)

### LINKED ISSUE #5752: Create bedrock base template for Protocol
This is an issue to track creation of a new base template for using Protocol in bedrock. There are still a couple of global pieces missing from Protocol, such as the main nav/footer, and the download button. Until these pieces exist in Protocol, we may need to reuse some Pebbles component CSS as a temporary measure.

alexgibson:
## Description
- Adds a new base template for Protocol.
- Ports the following components which do not yet exist in Protocol:
  - Global nav
  - Legacy masthead
  - Download button
  - Footer
- Adds a universal Protocol stylesheet for legacy IE browsers.
- Adds temporary test page for demo purposes.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/5752

## Testing
Demo page: https://bedrock-demo-agibson.oregon-b.moz.works/en-US/protocol-test/

### DIFF
--- a/media/css/protocol/components/_download-button.scss
+++ b/media/css/protocol/components/_download-button.scss
@@ -1,0 +1,234 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../../../../node_modules/@mozilla-protocol/core/protocol/css/includes/lib";
+
+// Download button color
+$color-button-green: #16da00;
+
+// The basis for all buttons on mozilla.org
+%base-button {
+    @include text-body-md;
+    @include open-sans;
+    background: none;
+    border-radius: 100px;
+    border-style: solid;
+    border-width: 2px;
+    cursor: pointer;
+    display: inline-block;
+    font-weight: bold;
+    padding: .9em 40px;
+    position: relative;
+    text-align: center;
+    text-decoration: none;
+    text-shadow: none;
+    transition: background-color .1s ease-in-out, border-color .1s ease-in-out;
+
+    &:active {
+        position: relative;
+        top: 1px;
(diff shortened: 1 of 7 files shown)


bedrock/mozorg/urls.py
```diff
@@ -315,4 +315,6 @@
     url(r'^microsummaries/0\.1$', views.namespaces, {'namespace': 'microsummaries'}),
     url(r'^projects/xforms/2005/type$', views.namespaces, {'namespace': 'xforms-type'}),
     url(r'^xbl$', views.namespaces, {'namespace': 'xbl'}),
+
+    page('protocol-test', 'mozorg/protocol.html'),
```
alexgibson: TODO: remove this page post-code review.

media/css/protocol/components/_download-button.scss
```diff
@@ -0,0 +1,234 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../../../../node_modules/@mozilla-protocol/core/protocol/css/includes/lib";
```
craigcook: Maybe I'm crazy, but after a little local experimenting I'm not sure we actually need to re-import this in every file. It's become such a habit for so long I haven't even questioned it before, but now I wonder if it's a legacy from Sandstone/Less that isn't needed in Sass. It needs more thorough testing. But I don't like the length and fragility of this relative path so if we don't have to repeat it everywhere that would be nice. Not relevant to this PR but something to think about.

alexgibson: Godd point - let's investigate further in Protocol itself 👍

alexgibson:
Thanks @craigcook - demo page removed :shipit:

Issue #5798, opened 2018-06-07
Change makefile after first prod deployment of #5618


pmac:
Once the build and deployment changes are pushed to production for the first time after the merge of #5618 all of the docker images (testing, building, etc.) will be updated on the docker hub. Once that happens developers should not need to build their own docker images locally unless something changes with a python or node dependency. So we should update the makefile and docs to reflect that the default method of getting started is to simply pull the images from dockerhub and start up the servers.

PR #5812, opened 2018-06-09
Move install instructions and Makefile to mostly just pull

pmac:
Most developers won't need to build their own Docker images. Pulling our pre-built ones from the Docker Hub will work perfectly for them.  This patch changes the install instructions to reflect this as well as changing the majority of the Makefile commands to depend on pull rather than build.


pmac:
Will fix #5798

PR #5819, opened 2018-06-17
Fix Bug 1465965: wrapped text in Jinja trans statement

astrastefania:
## Description
One of the two OSM link wasn't in the Jinja trans statement

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1465965

## Testing
Checked manually


astrastefania:
Great!

PR #5830, opened 2018-06-20
[fix #5829] xfail releasenotes timed cache test.

### LINKED ISSUE #5829: Improve/remove release notes version change test
[This test](https://github.com/mozilla/bedrock/blob/master/bedrock/releasenotes/tests/test__utils.py#L46) (`test_calls_function_when_version_changes`) is currently failing repeatedly in Circle, but is working fine locally. Most likely a race condition between `sleep` and the memoize timeout.

jpetto:
## Description

Marks failing* `/releasenotes` util test with `xfail`.

* Only seems to fail in Circle, but not in Jenkins or locally.

### DIFF
--- a/bedrock/releasenotes/tests/test__utils.py
+++ b/bedrock/releasenotes/tests/test__utils.py
@@ -1,3 +1,4 @@
+import pytest
 import time
 
 from mock import Mock, patch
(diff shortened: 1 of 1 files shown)


kyoshino:
Confirming that all the tests now have passed in #5832.

PR #5837, opened 2018-06-21
Fix source of set_git_env_vars.sh in deploy.sh

jgmize:
## Description

See error in https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/master/1077/pipeline/


jgmize:
This is a followup to an error introduced in https://github.com/mozilla/bedrock/pull/5836

PR #5842, opened 2018-06-25
Revert "Revert 'Add Jul-Dec 2017 Transparency Report'"

stephaniehobson:
Do not merge until July 3.

## Description
Reverts the revert of the Transparency Report.

This code has been previously reviewed and approved but had to be reverted when the client asked to wait until July 3 to launch.

## Issue / Bugzilla link
Closes #5747

Issue #5747, opened 2018-05-25
Transparency report updates


ejregithub:
https://www.mozilla.org/en-US/about/policy/transparency/

- [x] #1 - Create new transparency report for July-December 2017 with content [here](https://www.mozilla.org/en-US/about/policy/transparency/jan-jun-2017/)
- [x] #2 - Corresponding to #1: In https://www.mozilla.org/en-US/about/policy/transparency/ - Change text on button to “July-December 2017 Report”
- [x] #3 - add new Question to [FAQ](https://www.mozilla.org/en-US/about/policy/transparency/#faq ) according to spec [here](https://docs.google.com/document/d/1eXoZbq2apake2e2onL5YHbchu_7zpCb5bks6yoD7S6I/edit)

stephaniehobson:
We were asked to revert this and wait until July 3 to launch.

alexgibson:
This is up on demo https://bedrock-demo-agibson.oregon-b.moz.works/en-US/about/policy/transparency/jul-dec-2017/

jgmize:
https://www.mozilla.org/en-US/about/policy/transparency/ now has errors since https://github.com/mozilla/bedrock/pull/5848 went into production:

NoReverseMatch: Reverse for 'mozorg.about.policy.transparency.jul-dec-2017' with arguments '()' and keyword arguments '{}' not found. 0 pattern(s) tried: []

jgmize:
the errors appear to be the result of the switch `transparency-report-030618` being unset or false: https://github.com/mozilla/bedrock/pull/5848/files#diff-71245d0bee3444a747f3251f25fc59a4R317

PR #5848, opened 2018-06-26
Add Jul-Dec 2017 Transparency Report

alexgibson:
## Description
- Replaces #5842
- Adds a switch for toggling the new transparency report (not due to go live until July 3rd, so with a switch we can merge & enable when needed).
- Needs to be in prod **before July 3rd**.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/5747

## Testing
Demo: 
- https://bedrock-demo-agibson.oregon-b.moz.works/en-US/about/policy/transparency/
- https://bedrock-demo-agibson.oregon-b.moz.works/en-US/about/policy/transparency/jul-dec-2017/

Switch:
- Test new & old content displays as expected when toggling the switch `SWITCH_TRANSPARENCY_REPORT_030618`

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
+++ b/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
@@ -7,7 +7,9 @@
 <nav class="content nav-previous">
   <h3>{{ _('Previous Reports') }}</h3>
   <ul>
+    {% if switch('transparency-report-030618') %}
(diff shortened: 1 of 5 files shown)


alexgibson:
Execs want to see this on a demo, so marking as DNM for the time being.

alexgibson:
Ok, this is good to review/merge and will be behind a switch until the 3rd

alexgibson:
@craigcook I think the privacy folks would like this in production for the 3rd, which means it would need review & merge / push probably today?

PR #5920, opened 2018-07-11
[fix #5840] Update static-bundles when Docker restarts.

### LINKED ISSUE #5840: Changes to static-bundles.json not recognized when running locally via Docker
The new CSS/JS bundles created for `/whatsnew/` for Fx61 are 404'ing when running locally via Docker, but work fine when running via `npm start`.

STR: Running locally via Docker, create a new JS bundle (containing any `.js` file) and include it on any page. You'll see a 404 in the terminal window when loading that page. Running the site via `npm start` loads the new bundle correctly.

After a quick look, my guess is that this is due to `static-bundles.json` being a [part of the Docker image itself](https://github.com/mozilla/bedrock/blob/master/Dockerfile#L17-L18), and not being watched for changes. Having to re-build the Docker image each time a bundle is added is (IMO) too clunky.

jpetto:
## Description

Puts `static-bundles.json` inside `/media` so changes are picked up by Docker without rebuilding the image.

## Issue / Bugzilla link

#5840 

## Testing

Pull down this PR and run `make build`, then `make run`. Add a new bundle or two to `/media/static-bundles.json`, add those bundles to a page, re-start Docker, and you should see those bundles loaded on said page.


docs/install.rst
```diff
@@ -66,6 +66,16 @@ files and/or package.json file then simply running::
...
 
+**Asset bundles**
+
+If you make a change to ``media/static-bundles.json``, you'll need to restart Docker.
+
+.. note::
+
+    Sometimes stopping Docker doesn't actually kill the images. To be safe, after stopping docker, run
+    ``docker ps`` to ensure the containers were actually stopped. If they have not been stopped, you can force
+    them by running ``docker kill <container_name>``, e.g. ``docker kill bedrock_app_1``.
```
pmac: Or just `docker-compose kill`.

PR #5926, opened 2018-07-14
Use HTML char for ampersand

black-dragon74:
for better cross-browser compatibility.


alexgibson:
Thanks for the pull request. While this may the right thing to do technically, this does change a string on the page which means re-translating for every language. The ampersand isn't really doing any harm where it's situated here, so I don't think this is worth changing on its own. When we have a more substantial string change happening on this page it may be a better time to revisit.

Thanks for pointing it out!

PR #5929, opened 2018-07-16
Fix #5928: Run full release Docker build in PR tests

### LINKED ISSUE #5928: Compile CSS in CI on pull requests
I'm pretty sure we used to compile CSS in CI against each pull request. I'm guessing this changed as part of the django pipeline removal, but it would probably be useful to have this check still.

As an example, https://github.com/mozilla/bedrock/pull/5927 breaks the home page due to an error upstream - but this isn't picked up until someone pulls down the branch to test locally.

pmac:
CircleCI was only running a partial build of our images for pull-requests, but we need it to run the full release build so that we can make sure all assets build correctly before merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,7 +88,7 @@ docs: .docker-build-pull
 .docker-build-ci:
 	${MAKE} build-ci
 
-build-ci: .env .docker-build-pull submodules
+build-ci: .docker-build-pull
 	${DC_CI} build release
 #	tag intermediate images using cache
 	${DC_CI} build app assets builder app-base
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,6 +17,8 @@ RUN yarn global add gulp-cli@2.0.1
 COPY .eslintrc.js .stylelintrc gulpfile.js ./
 COPY ./media ./media
 
+RUN gulp build --production
+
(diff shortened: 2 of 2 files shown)


metadave:
:+1:

Makefile
```diff
@@ -88,7 +88,7 @@ docs: .docker-build-pull
 .docker-build-ci:
 	${MAKE} build-ci
 
-build-ci: .env .docker-build-pull submodules
```
pmac: I just noticed that these were superfluous since the `pull` step has them as dependencies.

pmac:
I've changed how I did this. I moved the production asset building to the final step in the `assets` image build instead of its own intermediary image build step. This simplifies the build slightly, lets us catch assets errors earlier in the build process (as well as running them in Circle CI), and means that the assets image will have pre-built assets for inspection and potential local time savings.

The reason my first attempt failed was because CircleCI apparently can not create symlinks in images. I've not fully investigated whether this is just a limitation of theirs or if django's `collectstatic -l` management command is doing something odd for linking, but that is why we can't just do the entire prod image build in Circle for now.

/cc @jgmize that's the deal if you're still curious ^^

PR #5932, opened 2018-07-17
Upload static media to s3 on deploy

pmac:
This copies the files out of the docker container and uploads them to the stage and prod s3 buckets when a prod deploy is happening (stage and prod are both deployed during a prod deployment). 

/cc @jpetto @jgmize @metadave

### DIFF
--- a/bin/upload-staticfiles.sh
+++ b/bin/upload-staticfiles.sh
@@ -1,0 +1,42 @@
+#!/bin/bash
+
+set -exo pipefail
+
+source docker/bin/set_git_env_vars.sh
+
+if [[ "$BRANCH_NAME" != "prod" ]]; then
+    # nothing to do if not prod deployment
+    exit 0
+fi
+
+CONTAINER_NAME="bedrock-${BRANCH_AND_COMMIT}"
+
+rm -rf static
+# have to rerun staticfiles without symlinks or we just copy broken symlinks
+docker-compose run --name "$CONTAINER_NAME" release docker/bin/build_staticfiles.sh --nolink
+docker cp "${CONTAINER_NAME}:/app/static" ./static
+docker rm -f "$CONTAINER_NAME"
+
+cd static
+for BUCKET in stage prod; do
+    # hashed filenames
+    aws s3 sync \
+        --exclude "*" \
+        --include "*.????????????.*" \
+        --acl public-read \
+        --cache-control "max-age=315360000, public, immutable" \
+        --profile bedrock-media \
+        . "s3://bedrock-${BUCKET}-media/"
+    # non-hashed filenames
+    # may not need to include non-hashed files
+    # TODO look into this if this is slow or makes the bucket too large
+    aws s3 sync \
+        --exclude "*.????????????.*" \
+        --acl public-read \
+        --cache-control "max-age=21600, public" \
+        --profile bedrock-media \
+        . "s3://bedrock-${BUCKET}-media/"
+done
+
+cd ..
+rm -rf static
--- a/.gitignore
+++ b/.gitignore
@@ -55,4 +55,5 @@ package-lock.json
 .vscode/
 static_build
 static_final
+static
 npm_debug.log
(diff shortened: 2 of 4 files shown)


bin/upload-staticfiles.sh
```diff
@@ -0,0 +1,42 @@
...
+for BUCKET in stage prod; do
+    # hashed filenames
+    aws s3 sync \
+        --exclude "*" \
+        --include "*.????????????.*" \
+        --acl public-read \
+        --cache-control "max-age=315360000, public, immutable" \
+        --profile bedrock-media \
+        . "s3://bedrock-${BUCKET}-media/"
+    # non-hashed filenames
```
metadave: are lines 30 and 31 the same comment repeated?

bin/upload-staticfiles.sh
```diff
@@ -0,0 +1,42 @@
...
+    # may not need to include non-hashed files
+    # TODO look into this if this is slow or makes the bucket too large
```
metadave: mdn syncs required an [additional flag](https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html) `--page-size` to help prevent timeouts on large buckets.

bin/upload-staticfiles.sh
```diff
@@ -0,0 +1,42 @@
...
+
+cd static
+for BUCKET in stage prod; do
+    # hashed filenames
+    aws s3 sync \
+        --exclude "*" \
+        --include "*.????????????.*" \
+        --acl public-read \
+        --cache-control "max-age=315360000, public, immutable" \
+        --profile bedrock-media \
```
metadave: any reason for using `--profile` vs the `AWS_PROFILE` env var?

pmac: that I didn't know about `AWS_PROFILE` and was just copying what IRL did:

https://github.com/mozilla/irlpodcast/blob/91b8bbb35da7f9ff3b1e844b7ce103bfd43eca0a/Jenkinsfile#L17

bin/upload-staticfiles.sh
pmac: according to the docs:

```
--page-size (integer) The number of results to return in each response to a list operation. The default value is 1000 (the maximum allowed). Using a lower value may help if an operation times out.
```

To my knowledge we're not doing any list operations. Or is it doing that behind-the-scenes during the sync?

bin/upload-staticfiles.sh
pmac: Happy to switch it if one is better than the other though.

bin/upload-staticfiles.sh
metadave: I seem to remember troubleshooting the MDN sync and seeing that `aws s3 sync` used list to compare files.

bin/upload-staticfiles.sh
pmac: They look different to me locally. Line 30 says what the next section is doing, and 31 says that we may not eventually need to do this and 32 says we should look into that someday if it's a problem.

pmac:
The buckets are now populated as @jpetto and I did a test run of this:

https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/static-media-to-s3/1/pipeline

bin/upload-staticfiles.sh
pmac: @metadave do you recall to what value they set `--page-size`?

metadave: [100](https://github.com/mdn/infra/blob/master/apps/mdn/utils/mdn_backup_cron/image/mdn_sync.sh#L11)

metadave: definitely not required, but keep it in mind in case there are sync issues.

pmac:
To be clear: what this PR does is upload the media to the buckets during deployment. It does not change the app in any way to make it use the s3 buckets for serving prod assets. My proposal is:

1. Merge this PR
2. Deploy production
3. Change settings in CloudFlare for staging to point `/media/*` to the staging s3 bucket
4. Test staging extensively to make sure this works as expected
5. Make the same changes in CloudFlare for prod
6. Profit! 💰💰💰💰

PR #5936, opened 2018-07-20
Use staticfiles.json map to set proper cache headers in s3

pmac:
Instead of trying to match filenames it is more accurate to use the JSON map Django gives us to move the hashed files to a new directory and do the `aws s3 sync` separately for each type of file.


pmac:
Related to https://github.com/mozilla/kitsune/pull/3272

PR #5939, opened 2018-07-23
[fix #5723] Add Pocket feed to the home page.

### LINKED ISSUE #5723: Automated Pocket Feed On The Mozilla.org Homepage
- [x] - [LDAP account set up request](https://mozilla.service-now.com/sp?id=ticket&table=incident&sys_id=69d005d6dbfadb40809e17e15b961977)

This issue captures pre-reqs to delivering an automated feed of stories from [mozillaHQ](https://getpocket.com/@MozillaHQ) pocket account. It will be unblocked once we reach a decision on the homepage. 

### User Accounts
The current mozilla account is a shared login among a few mozilla employees. This should be a proper ldap account with 2 factor authentication enabled.

1. Request from IT an LDAP user mozillahqpocket@mozilla.com  with Janis and any other folks who need access assigned (they have done this before).
1. Once that is done sign up for pocket via -> `https://getpocket.com/signup/ and choose the "Sign Up with Google" option.`
1. Reach out to either Justin or Matt to migrate the existing account over. 

### Retrieving Data
Below is the note from Matt regarding the API:

> Re: the [API](https://getpocket.com/developer/docs/v3/retrieve) - the endpoint you specified will return you "saves" - basically anything that has saved to the specified account. It will not return you the recommended posts you put on https://getpocket.com/@MozillaHQ. If you want those, you will either need to use the approach Nick outlined -or- if you want to use the API, I think we could provide you access to a non-publicized endpoint that would return the recommended posts in JSON format (similar to the response you referenced).

### Pocket Contacts
* matt@getpocket.com - Can assist with all things technical.
* justin@getpocket.com - Can assist with anything account or community related.

jpetto:
## Description

Adds automated Pocket feed to the homepage.

## Issue / Bugzilla link

#5723 

## Testing

https://bedrock-demo-jpetto.oregon-b.moz.works/en-US/


jpetto:
@pmac Still semi-WIP, but I think in good enough shape to take a first pass at a review.

jpetto:
Noting that Matt from Pocket said hitting their API once every five minutes is not a problem.

jpetto:
@pmac This should now be good for a review. There are a couple `TODO`s in there that I'd specifically like feedback on.

Thanks!

bedrock/pocketfeed/api.py
```diff
@@ -0,0 +1,59 @@
...
+        del article['quote']
+
+        check_article_image(article)
+
+
+def check_article_image(article):
+    """Determine if external image is available"""
+
+    # sanity check to make sure image provided by API actually exists and is https
+    if (article['image_src'] and re.match(r'^https://', article['image_src'], flags=re.I)):
```
pmac: NIT: superfluous parentheses around the if statement

bedrock/pocketfeed/management/commands/update_pocketfeed.py
```diff
@@ -0,0 +1,28 @@
...
+from bedrock.pocketfeed.models import PocketArticle
+
+
+class Command(BaseCommand):
+    def add_arguments(self, parser):
+        parser.add_argument('-q', '--quiet', action='store_true', dest='quiet', default=False,
+                            help='If no error occurs, swallow all output.'),
+
+    def handle(self, *args, **options):
+        error = None
```
pmac: NIT: No need to define error here if it's only used once below

bedrock/pocketfeed/models.py
```diff
@@ -0,0 +1,107 @@
...
+        return staticfiles_storage.url('img/pocket/pocket-feed-default.png')
+
+    # TODO: is this necessary? title's may have unicode characters, but i don't *think* will ever have markup
+    @property
+    def display_title(self):
+        return Markup(self.title).unescape()
+
+    @property
+    def image(self):
+        return self.image_src if self.image_src else self.fallback_image()
```
pmac: NIT: Could be simplified as `return self.image_src or self.fallback_image()`

bedrock/pocketfeed/models.py
```diff
@@ -0,0 +1,107 @@
...
+        if update_count:
+            expired_articles = self.exclude(pocket_id__in=article_ids)
+            del_count = expired_articles.count()
+            expired_articles.delete()
+
+        return update_count, del_count
+
+
+class PocketArticle(models.Model):
+    pocket_id = models.IntegerField()
```
pmac: As I said above we could avoid having extra autogenerated IDs by doing:

```python
class PocketArticle(models.Model):
    id = models.IntegerField(primary_key=True)
    ...
```

But it's not a requirement that we do. However, doing it like this might mean that our sync code is superfluous and we could just do a `PocketArticle.objects.all().delete()` during the refresh and it would work fine since the IDs wouldn't change. I'm not suggesting we do that since the better way is already coded, but we could if we ran into issues with the other code.

jenkins/branches/demo/default.env
```diff
@@ -10,3 +10,4 @@ SECRET_KEY=itsasekrit
 ALLOWED_HOSTS=*
 PROD_DETAILS_STORAGE=product_details.storage.PDDatabaseStorage
 GTM_CONTAINER_ID=GTM-MW3R8V
+LOCAL_DB_UPDATE=True
```
pmac: Remove this change before merge.

bedrock/pocketfeed/api.py
jpetto: Sometimes I'm still writing JavaScript. 😁

bedrock/pocketfeed/models.py
jpetto: I think I'll leave as-is for now, but good to keep in mind. 👍

pmac:
I think we're good to go here. My only other thought is that we might want a switch around the display of the feed so that we can turn it off whenever we need. I see that it will disappear cleanly if there is no data, but it'd be nice to roll it out to prod, then enable only on stage to see how it's going, then enable in prod, as well as having a quick way to take it down should an unfortunate article slip in there.

jpetto:
Hmm, based on history, I concur. A switch in the home template should suffice, yes?

pmac:
> Hmm, based on history, I concur. A switch in the home template should suffice, yes?

Yes exactly. I think that'll do nicely. 👍

PR #5945, opened 2018-07-25
Switch from IRC to Slack notifications from Jenkins

pmac:


pmac:
See the messages in `#www-notify` on our slack instance to see it working for my demo deployment from this branch.

pmac:
I installed [slack-cli](https://pypi.org/project/slack-cli/) on Jenkins and configured it with the slack token for the `meao-jenkins` bot I created. So all other Jenkins jobs should be able to do something similar should they wish to.

PR #5954, opened 2018-07-30
Use :latest docker tags for builds from "master"

pmac:
Add :prod-latest tags for production images. This means that local devs would get updated images with updated dependencies much more quickly just via a `make pull` instead of having to build fresh images themselves.


pmac:
@jgmize I can't think of a reason that changing the definition of `:latest` in this way will negatively impact anything, but please let me know if you can think of a reason to do this another way.

pmac:
@jgmize bump

PR #5957, opened 2018-07-31
Update security sidebar and main index to link to top level bug bounty page

pmac:
Updating sidebar and main index to link to top level bug bounty page because there was only a hidden link to the overall progrm. Fixed mailto: links and updated language for the first time in eight years for clarity.


albill:
Thanks.

Issue #5741, opened 2018-05-25
Microdata of SoftwareApplication and Product is missing required data


alexgibson:
https://bugzilla.mozilla.org/show_bug.cgi?id=1460253

/cc @ejregithub

rraue:
To test, if we can use js+ld inline I propose @type "organization" as the social media profiles in "sameAs" usually  show up in the knowledgraph box as seen at apple: https://screenshots.firefox.com/CoLyrbEUoiJ4MBtE/www.google.com in the right sidebar.

We currently have "only" twitter and facebook there. Lets see if we get added at least youtube or linkedin as well: https://screenshots.firefox.com/JIRS6iN9rGLyOFDe/www.google.com. If we know that it works, we can talk about the sense of it and which Networks there really should be.

The code for us would be on our homepage: 
<script type="application/ld+json">
	{
		"@context": "http://schema.org",
		"@type": "Organization",
		"name": "Mozilla",
		"url": "https://www.mozilla.org",
		"logo": "https://www.mozilla.org/media/img/mozorg/mozilla-256.4720741d4108.jpg",
		"sameAs": [
			"https://www.wikidata.org/wiki/Q9661",
			"https://twitter.com/mozilla?lang=de",
			"https://www.facebook.com/mozilla",
			"https://github.com/mozilla",
			"https://www.instagram.com/mozilla/",
                         "https://www.youtube.com/user/Mozilla";
			"https://wikipedia.org/wiki/Mozilla",
			"https://www.linkedin.com/company/mozilla-corporation"
		]
	}
</script>

I am just not sure with the url element if it should be localized or not. For the first test I think we should go withou locale even thought its a redirect.

rraue:
Reopening because of https://github.com/mozilla/bedrock/issues/6949

The test on the homepage was succesful, we can use js+ld instead of microdata and we should. To fix Softwareapllication on /new we need two things:

- [ ] Converting microdata to js+ld and correct all errors (currency)

- [ ] We need an aggregated rating for Firefox, else it will not be displayed in search results but most other downloadpages have this visual advantage

rraue:
The js+ld code including breadcrump and SiteNavigationElement:

```
<script type="application/ld+json">
  [{
    "@context": "http://schema.org",
    "@type": "BreadcrumbList",
    "itemListElement": [{
      "@type": "ListItem",
      "position": 1,
      "item": {
        "@id": "https://www.mozilla.org/en-US/",
[...]

PR #5980, opened 2018-08-08
Add inline js+ld microdata to home page (#5741)

alexgibson:
## Description
- Adds inline ld+json microdata to the English home page to test it's impact vs data-attributes.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/5741

## Testing
- Inline js-ld should be visible in the home page markup and should not throw a CSP error.


alexgibson:
@rraue - I added this based on your example in the issue. I stuck with Twitter, Instagram & YouTube, since these are the main social media channels we currently promote on the site. Does that work for you?

rraue:
Please take all sameAs in the code from my example, that’s exactly what this meta information is for. We don’t need it on the site, that’s ux but it helps crawlers to make connections and therefor use the right logo.

But as it’s for now just the test if it will be not blocked in backend, it’s not really important, but you know how long mvp code sometimes survives on our site.

And for the real solution for Firefox as product it will be really important
I actually on purpose searched for all official channels and will do this for Firefox as well.

Not all of this will appear in search in the knowledge graph, but will be indexed and will help us being the true source of our brand in the knowledge graph. The more official channels and reliable information are connected, the easier it gets to have this wonership.

alexgibson:
Updated 👍

PR #5984, opened 2018-08-08
Fix #5983: Remove DEIS_* settings

### LINKED ISSUE #5983: Remove use of DEIS_* variables in settings
We can do the following instead:

1. Replace `DEIS_DOMAIN` with `CLUSTER_NAME` which is set in deployment.
2. Replace `DEIS_APP` with parsing the prefix of `HOSTNAME` which looks like e.g. `bedrock-stage-web-75c9dd854-pwc7x`

pmac:
Move to `CLUSTER_NAME` from the environment and `APP_NAME`
derived from `HOSTNAME`.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -368,6 +368,13 @@ PUENTE = {
 
 
 def get_app_name(hostname):
+    """
+    Get the app name from the host name.
+
+    The hostname in our deployments will be in the form `bedrock-{version}-{type}-{random-ID}`
+    where {version} is "dev", "stage", or "prod", and {type} is the process type
+    (e.g. "web" or "clock"). Everywhere else it won't be in this form and will return None.
+    """
     if hostname.startswith('bedrock-'):
         app_mode = hostname.split('-')[1]
         return 'bedrock-' + app_mode
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -366,11 +366,20 @@ PUENTE = {
     }
 }
 
+
+def get_app_name(hostname):
+    if hostname.startswith('bedrock-'):
+        app_mode = hostname.split('-')[1]
+        return 'bedrock-' + app_mode
+
+    return None
+
+
 HOSTNAME = platform.node()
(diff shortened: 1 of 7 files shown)


bedrock/settings/base.py
```diff
@@ -366,11 +366,20 @@ def set_whitenoise_headers(headers, path, url):
     }
 }
 
+
+def get_app_name(hostname):
+    if hostname.startswith('bedrock-'):
```
jpetto: Maybe a comment here as to the naming convention/expected possible values? (You know, for the next time our infra changes :trollface:)

bedrock/settings/base.py
```diff
@@ -366,11 +366,20 @@ def set_whitenoise_headers(headers, path, url):
...
+        return 'bedrock-' + app_mode
+
+    return None
+
+
 HOSTNAME = platform.node()
-DEIS_APP = config('DEIS_APP', default='')
-DEIS_DOMAIN = config('DEIS_DOMAIN', default='')
+APP_NAME = get_app_name(HOSTNAME)
+CLUSTER_NAME = config('CLUSTER_NAME', default='')
```
jpetto: Is there anything we need to do to ensure this is set on the clusters prior to merging/deploying this PR?

pmac: I took [this line in the configmap](https://github.com/mozmeao/www-config/blob/git-sync-operator/iowa-a/bedrock-prod/configmap.yaml#L10) to mean that it's ready. @jgmize can you confirm that it's ready and available in every cluster including oregon?

bedrock/settings/base.py
```diff
@@ -366,11 +366,27 @@ def set_whitenoise_headers(headers, path, url):
     }
 }
 
+
+def get_app_name(hostname):
+    """
```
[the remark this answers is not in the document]
jpetto: Exactly! 🌟

bedrock/settings/base.py
pmac: @jgmize actually I don't see `CLUSTER_NAME` in oregon deployments. I'll set that via deis commands and then merge.

Issue #5775, opened 2018-06-04
Seo: Complete Sitemap


rraue:
- Helps to analyse and to keep track of changes not just internally, but like crawlers see it. 
- Webmastertools provide more data on documents in Sitemaps. Biggest bonus: crawling gets much faster.
-First step to implement hreflang in sitemaps (produces less errors then in head)

Brief: https://bugzilla.mozilla.org/show_bug.cgi?id=1441250

Part of https://github.com/mozilla/bedrock/issues/5652

PR #5987, opened 2018-08-08
Re #5775: Split sitemap by locale

pmac:
Also add DB-based pages to sitemap from release notes and security.

This PR does not address the last modified dates. That is potentially a lot of work and makes sense to address in a separate PR.


pmac:
@rraue this is deployed to [my demo](https://bedrock-demo-pmac.oregon-b.moz.works/sitemap_index.xml). Take a look and let me know. Details:

* Sitemap index: `/sitemap_index.xml`
* Sitemap for locale: `/{locale}/sitemap.xml`
* Sitemap for URLs with no locale: `/sitemap_none.xml`

As I said above I'm going to tackle the modified dates in a separate PR after this one merges.

rraue:
I found some urls with server errors in the /en-US/ sitemap. Will scan tomorrow all urls, but most errors should occur in locales if so in en-US as well.

Here is the list: https://docs.google.com/spreadsheets/d/1zyAyBlRKF3OgK_boEC5VnWg8kDYjU_CkrAX7TLExj-I/edit?usp=sharing

pmac:
Very odd. All of those 50X URLs work just fine for me. And those are prod urls. We shouldn't be seeing that. I'll check the rest out as well. I'm sure we can deal with the redirect ones.

rraue:
Maybe it’s a temporary cloud flare problem. I am hitting these 50x with my browser as well from Germany. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

rraue:
Another thing I recogniced: there is a sitemap for /hi/ but we use /hi-IN/ if I am not mistaken, they all appear as 301.

pmac:
> Another thing I recogniced: there is a sitemap for /hi/ but we use /hi-IN/ if I am not mistaken, they all appear as 301.

Ah yes. Legal docs. Those are still problematic for l10n :( For now I'll just manually add the `hi`  URLs to `hi-IN` during sitemap creation.

rraue:
You killed all 301s and 404s in the sitemaps, yay. Just found these 7 urls, all 302:

https://www.mozilla.org/hi-IN/privacy/
https://www.mozilla.org/hi-IN/about/legal/acceptable-use/
https://www.mozilla.org/hi-IN/privacy/firefox-os/
https://www.mozilla.org/hi-IN/privacy/firefox/
https://www.mozilla.org/hi-IN/about/legal/terms/services/
https://www.mozilla.org/hi-IN/privacy/websites/
https://www.mozilla.org/hi-IN/about/legal/terms/mozilla/

everything else looks fantastic.

pmac:
I filed an [issue on the legal-docs repo](https://github.com/mozilla/legal-docs/issues/1155) earlier today for this. The problem is that the `hi-IN` translation of the legal-docs pages isn't accessable because the doc claims to be in `hi`, but the site doesn't support that locale. Hopefully we'll get it fixed soon, but for now I'll remove them from the list.

Dockerfile
```diff
@@ -100,9 +100,6 @@ RUN ./bin/run-db-download.py --force
 # get fresh l10n files
 RUN ./manage.py l10n_update
 
-# generate the sitemaps
-RUN ./manage.py update_sitemaps
```
pmac: I removed this because the sitemaps take some time now that we're getting the translations and this step was only for the dev image where it's not really needed.

bin/bootstrap.sh
```diff
@@ -19,6 +19,3 @@ git submodule update --init --recursive
 
 # get fresh l10n files
 ./manage.py l10n_update
-
-# generate the sitemaps
-./manage.py update_sitemaps
```
pmac: Same as above: I removed this because it can break with improper settings and it's not needed for locally running the site.

pmac:
@jpetto this is now officially ready for review. Thanks!

rraue:
I was now able to test the sitemaps live and found some pages missing in the sitemap. Weirdly all locale /about/ pages aren't in the sitemap but the en-US version is. The rest are the pages always showing up when something is missing like old releasenotes and system requirements. I did a list of pages not in the sitemap but indexable and 200: https://docs.google.com/spreadsheets/d/15iRo0ospuuQqSgI6YUtBKdGvqfMUHj8Awzv8PE7ssX8/edit?usp=sharing

rraue:
And /firefox/all/ is in the sitemaps for all locales, but isn't indexable, so we should remove it.

pmac:
@rraue can you open a new issue for this stuff. This PR was closed over a week ago.

pmac:
Or reopen #5775

PR #5996, opened 2018-08-10
Fix bug 1482045: Update legal-docs repo

pmac:
These changes include additional translations for the Firefox Rocket
privacy policy.


pmac:
This will fix #5997

PR #6001, opened 2018-08-13
Fix #6000: Map legal-docs locales to bedrock prod locales

### LINKED ISSUE #6000: Fix legal-docs language code missmatch
The supported locale for Hindi on the site is `hi-IN`, but in other projects like legal-docs it's `hi`. Instead of changing the other projects we should have a way to map these in bedrock.

Also some languages available in legal docs (`sw` for example) aren't supported on the site, but do show up as available translations because there is a file in the legal-docs repo. When you try to view these tranlsations of the page however it's just a redirect to `en-US`. We should filter the list of available translations and only allow those that are in the `PROD_LANGUAGES` list. 

See https://github.com/mozilla/legal-docs/issues/1155

pmac:
Also filter out legal-docs translations that are not production bedrock locales.

see also: https://github.com/mozilla/legal-docs/issues/1155

### DIFF
--- a/bedrock/legal_docs/tests/test_base.py
+++ b/bedrock/legal_docs/tests/test_base.py
@@ -5,7 +5,7 @@
 from os.path import join
 
 from django.http import Http404, HttpResponse
-from django.test import RequestFactory
+from django.test import override_settings, RequestFactory
 
 from mock import patch, ANY
 from nose.tools import eq_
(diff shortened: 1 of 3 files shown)


pmac:
@jpetto this is ready for r? and it would be good to get it in soon since it fixes some legal-docs translations.

Issue #5943, opened 2018-07-25
/download/thanks page mail CTA content - a/b test 


ejregithub:
## Important Links

[Manapage with Test Write Up](https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=81691444)

[Deck with User Flow](https://docs.google.com/presentation/d/1Sq9DCPt5-clMx2rCQ079cM5D7ybkgNLXqMpF_Bb8N5g/edit?usp=sharing)

[Content and Copy](https://docs.google.com/document/d/1ijjDcATdeBGWrBdyTWmJ-4la4RIb4Tyepn3VHv7Dy1s/edit#)

[Sketch File](https://share.goabstract.com/efef0859-ed93-41ef-bbad-dfc90a4ce3ae)

### Prototypes
[Post-Download](https://invis.io/Q8NACUBYM9U)

## Success Criteria

- [ ] 'Expedite' delivery (this is from the Tiger Team)
- [ ] Implemented to 100% of en-US (across all content variations being tested)
- [ ] a/b test in place allocating 50/50 split (of the entire audience) to test content (headlines) in the email CTA

thejeffreytaylor:
ETA for copy treatments Thursday SF lunchtime.

dzingeek:
Working on the flow now. Just posted it to the team for feedback before next steps.

dzingeek:
**Link to deck w/user flow and rough wireframes.** 
https://docs.google.com/presentation/d/1Sq9DCPt5-clMx2rCQ079cM5D7ybkgNLXqMpF_Bb8N5g/edit?usp=sharing

thejeffreytaylor:
Copy here: https://docs.google.com/document/d/1ijjDcATdeBGWrBdyTWmJ-4la4RIb4Tyepn3VHv7Dy1s/edit#

michaelmccombie:
Here is a link to the InVision prototype for the pre-download workflow test:
https://invis.io/SDNAAE3ZCK4

craigcook:
@ejregithub  The title on this issue says it's an a/b test but the success criteria say it's 100% of en-US... we can run it as a test (meaning I'll put it behind a switch so we can turn it off) but if it's a/b are we testing it 50/50 against the regular thanks page? That may not be a meaningful test because the page with the form will obviously get more signups than the page without the form (which is zero).

[...]

PR #5999, opened 2018-08-13
Pre-download newsletter experiment on /firefox (Fixes #5944)

alexgibson:
## Description
- Adds experiment to show a newsletter signup modal when clicking "Download Firefox" on the `/firefox` page.

**do not merge** until we get both analytics and the new email privacy text in place.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/5944

## Testing
Demo:
  - https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/

Direct URL variants:
  - https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/?v=a (control)
  - https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/?v=b (newsletter + headline 1)
  - https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/?v=c (newsletter + headline 2)

PR #6008, opened 2018-08-13
[#5943] Email subscription on /download/thanks/

craigcook:
## Description
Adds a newsletter signup form on the /download/thanks/ page. 

This is a Traffic Cop A/B experiment with two headline variations, with a 50/50 split for all en-US users. The experiment uses the switch `firefox_download_scene2_email_experiment`

Targeting English-only for now so no l10n needed. If this test proves successful I imagine a version of this would become the new default /thanks/ page for all locales but would likely be redesigned by then.

Also adds a new email privacy page at /privacy/email/. The link to that page may become part of the standard form and we'll need to localize the page at that point.

## Testing
[Version A](https://bedrock-demo-craigcook.oregon-b.moz.works/en-US/firefox/download/thanks/?v=a)
[Version B](https://bedrock-demo-craigcook.oregon-b.moz.works/en-US/firefox/download/thanks/?v=b)
[Email privacy](https://bedrock-demo-craigcook.oregon-b.moz.works/en-US/privacy/email/)


craigcook:
@agibson This will surely cause some merge conflicts with https://github.com/mozilla/bedrock/pull/5999 though this one probably needs to merge first since it includes the /privacy/email page... hopefully the rebase won't be too hairy.

media/js/firefox/new/scene2-email.js
```diff
@@ -0,0 +1,52 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function($) {
+    'use strict';
+
+    var isIELT9 = window.Mozilla.Client.platform === 'windows' && /MSIE\s[1-8]\./.test(navigator.userAgent);
```
alexgibson: All this JS looks the same as the regular `scene2.js` code. Rather than have to maintain all this JS twice, I'd suggest putting the new modal-related JS code into a separate file that we can include in the template, in addition to the regular `scene2.js` file?

bedrock/firefox/views.py
```diff
@@ -573,6 +573,10 @@ def download_thanks(request):
             template = 'firefox/new/portland/scene2-safe.html'
         elif experience == 'betterbrowser':
             template = 'firefox/new/better-browser/scene2.html'
+        elif switch('firefox_download_scene2_email_experiment') and variant == 'a':
```
alexgibson: Maybe I'm a little confused on the plan, but I thought we wanted to roll this new page out as the default because it was the "right thing to do"? I had imagined the A/B copy test would be separate to doing that - but I'm guessing I didn't understand quite right.

alexgibson:
The design has changed since I last saw it, but is something supposed to be over to the right of the page? At first i thought I was missing an image file locally, but it seems to be the same on demo. Visually it feels like something should be there, or if not the newsletter could be centered on the page.

screenshot_2018-08-14 download the fastest firefox ever

media/js/firefox/new/experiment-download-scene2-email.js
```diff
@@ -0,0 +1,17 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function(Mozilla) {
+    'use strict';
+
+    var cop = new Mozilla.TrafficCop({
+        id: 'experiment-download-scene2-email-082018',
+        variations: {
+            'v=a': 50,
```
alexgibson: Which is the control group here? The new page isn't made default, but we're still putting 100% of users into the copy A/B test. Usually here we'd add a control group to compare conversions against, which normally defaults to the current experience in prod. It feels like that's not what we're testing here though, we're just testing the copy. Is there any reason why we can't make this new page default regardless of the experiment? Then we could treat say, version a, as the control?

bedrock/firefox/views.py
```diff
@@ -573,6 +573,10 @@ def download_thanks(request):
...
+            template = 'firefox/new/email/scene2-a.html'
+        elif switch('firefox_download_scene2_email_experiment') and variant == 'b':
```
alexgibson: A couple of Python unit test here would be good?

media/css/firefox/new/scene2_email.scss
```diff
@@ -0,0 +1,240 @@
...
+    #modal & {
+        @include open-sans;
+        display: block;
+        padding: 20px 40px;
+    }
+}
+
+
+
+
```
alexgibson: Nit blank lines

bedrock/firefox/templates/firefox/new/email/scene2-base.html
```diff
@@ -0,0 +1,114 @@
...
+    <div class="content-main">
+      <div class="download-button-wrapper desktop" id="download-button-wrapper-desktop">
+        {{ download_firefox(force_direct=true, dom_id='primary-download-button') }}
+      </div>
+
+      <h1 class="headline">{% block headline %}{% endblock %}</h1>
+      <p class="tagline">{{ _('Get informative tips, tricks and product announcements delivered to your inbox.') }}</p>
+
+      <div class="c-newsletter">
+        {{ email_newsletter_form(
```
alexgibson: Not something we necessarily need to fix right now, but what do people see if they don't get the newsletter translated?

media/js/firefox/new/scene2-email.js
craigcook: Good call. It is indeed just a copy of scene2.js with the modal added, no reason for the repetition. I'll update.

bedrock/firefox/views.py
craigcook: I'll get clarification on the test. I was under the impression this was going to be temporary and would need to be switch-offable, though I don't know what we're measuring against so maybe it really is the new permanent default (for en-US).

bedrock/firefox/templates/firefox/new/email/scene2-base.html
craigcook: Good question... since this is English only for now it's not an issue but I don't know what the longer term plan is. There's talk of soon redesigning the default /new as well as /thanks so things are likely to change. When we have a new default thanks page for all locales we'll definitely need a non-newsletter fallback. I'd suggest Fx mobile, like on the current page.

craigcook:
OK, made various updates...

* Switched to Protocol
* Simplified JS
* Made this the default page for English, no switch, with variant A as the control
* Added and updated tests

djst:
I don't know if you sorted this already, but we should roll out the test
50%/50% split. It is indeed "the right thing to do" so there's no need for
a control outside of these two variants. We are just trying to see which of
these two performs the best. As soon as we have a winner, we'll end the
test and roll out the winner to 100%. The previous thank you page without a
newsletter signup tout is not coming back. Hope that clarifies.

Den ons 15 aug. 2018 kl 00:53 skrev Craig Cook <notifications@github.com>:

bedrock/firefox/templates/firefox/new/email/scene2.html
```diff
@@ -11,6 +17,11 @@
   <link rel="canonical" href="{{ settings.CANONICAL_URL }}/{{ LANG }}/firefox/new/">
 {% endblock %}
 
+{% block page_image %}{{ static('img/firefox/template/page-image.png') }}{% endblock %}
```
alexgibson: Rather than add all these pieces of metadata on a per-page basis, would it make sense to create a `firefox/base-protocol.html` file that this template extends? (similar to how we already do for pebbles). That way all new Protocol pages under `/firefox/` could get this meta data for free?

media/css/firefox/new/scene2_email.scss
```diff
@@ -120,103 +121,43 @@
...
-        font-weight: bold;
-        margin: 0 0 .25em;
-    }
-
-    .privacy-check-label {
-        @include font-size-small;
-        padding-left: 2em;
-        display: block;
+            background: $color-white;
+            box-shadow: 4px 4px 0 0 #000;
```
alexgibson: Buttons are always a little contentious, but we seem to be making quite a few button related tweaks on Protocol pages that have not yet been adopted in the design system. Being forward thinking is good, but the more we do this the more papercuts we create for ourselves in the future to go back and undo all these custom styles. I guess this is up to our designers, but imho the UX here is not going to be negatively impacted if we just use the stock button for now :/

media/js/firefox/new/scene2-email.js
```diff
@@ -2,51 +2,43 @@
...
-                }, 1000);
-            });
-        }
-    }
-
-    var $modalWrapper = $('.email-privacy');
-    var $modalLink = $('.email-privacy-link');
-
-    $modalLink.on('click', function(e) {
+    // var $modalWrapper = $('.email-privacy');
```
alexgibson: Can this commented out code be removed now?

media/static-bundles.json
```diff
@@ -979,7 +978,10 @@
     },
     {
       "files": [
-        "js/base/mozilla-modal.js",
+        "protocol/js/protocol-modal.js",
+        "protocol/js/protocol-newsletter.js",
+        "js/newsletter/form-protocol.js",
```
alexgibson: Ugh, I totally forgot we still need to move that code into Protocol 😢

media/js/firefox/new/scene2-email.js
```diff
@@ -2,51 +2,43 @@
...
+    var strings = document.getElementById('strings');
+
+    trigger.addEventListener('click', function(e) {
         e.preventDefault();
-        Mozilla.Modal.createModal(this, $modalWrapper, {
-            title: $(this).text()
+        Mozilla.Modal.createModal(e.target, content, {
+            title: title.innerHTML,
+            className: 'mzp-t-firefox',
+            closeText: strings.dataset.close
```
alexgibson: Nit: use `Mozilla.Utils.trans()` here for browsers that don't support native `dataset`.

alexgibson:
> I don't know if you sorted this already, but we should roll out the test 50%/50% split. It is indeed "the right thing to do" so there's no need for a control outside of these two variants. We are just trying to see which of these two performs the best. As soon as we have a winner, we'll end the test and roll out the winner to 100%. The previous thank you page without a newsletter signup tout is not coming back. Hope that clarifies.

Right - but all tests should have a control group. If all we're testing is the copy, then one of the copy variations should be the control. So what we're doing is this:

- Make variation a of this new page the default for en-US .
- 50% of visitors to this page may see a different copy (variation b).
- Conversions are compared against the other 50% (variation a, control).

Hope that makes sense.

bedrock/firefox/templates/firefox/new/email/scene2.html
```diff
@@ -0,0 +1,132 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "base-protocol.html" %}
```
alexgibson: If I set `Dev=False`, I don't see the language switcher in the footer for this page, so i think we need to add `{% add_lang_files "firefox/new/quantum" %}` here?

alexgibson:
If I visit the `/thanks/` page in an old legacy IE, everything looks presentable enough, but the newsletter form won't expand, so I can't submit an email sucessfully. The form should work without JS still, so maybe we can add a CSS override to in oldIE.scss to make sure it's always expanded?

screen shot 2018-08-15 at 12 27 00

djst:
Yep, makes total sense.

Den ons 15 aug. 2018 kl 11:53 skrev Alex Gibson <notifications@github.com>:

media/css/firefox/new/scene2_email.scss
craigcook: I'm working on a PR to make this the base button style in Protocol. But yeah, this page would probably be fine with the current Protocol button until it's updated in Protocol (at which point this would automatically update once the new button is published upstream; the system works!) I'll remove this extra styling.

media/js/firefox/new/scene2-email.js
craigcook: doh! I'll clean up my mess.

media/css/firefox/new/scene2_email.scss
craigcook: Oh now I remember... the default button switches to white in a dark theme but Jennifer wanted it to be black here so it would be distinct from the form fields. So since I needed to invert the color anyway that's why I added the shadow... so it seems this still needs to be a non-standard button for the moment, but I'll at least remove the shadow and just invert the colors. 

And I'll make sure to include a theme class when I update the buttons in Protocol so we can still have a black button in a dark theme when needed.

craigcook:
> If I visit the /thanks/ page in an old legacy IE, everything looks presentable enough, but the newsletter form won't expand, so I can't submit an email sucessfully. The form should work without JS still, so maybe we can add a CSS override to in oldIE.scss to make sure it's always expanded?

Which version of IE? Did we ever fully drop all JS from IE8 and older? The form is only hidden when JS is available so evidently we're still adding the `.js` class in old IEs in order to hide things but not serving them the JS that makes them visible again.

alexgibson:
> Which version of IE? Did we ever fully drop all JS from IE8 and older? The form is only hidden when JS is available so evidently we're still adding the .js class in old IEs in order to hide things but not serving them the JS that makes them visible again.

We don't serve first class CSS or JS to < IE9.

craigcook:
OK, *all* browsers still get `site.js` which is what adds the `js` class to root, but lte IE8 don't get the newsletter JS. I think rather than trying to so any fancy JS hoops a simple override in the CSS is better. Here comes `!important`...

craigcook:
Hold on, waiting for a small copy change on the email privacy page, as well as final signoff from Michelle. Marking DNM for the moment.

PR #6009, opened 2018-08-14
Add press and privacy links to footer

craigcook:
## Description
Adds a general privacy link and a press center link, to replace those being removed from the About page. Since the new About page is English-only (for now), the Privacy link is also English-only to avoid string conflicts (otherwise it would show two identical strings for most locales). We need to get the "Website Privacy Notice" fully localized in main.lang so we can remove these conditionals.


stephaniehobson:
So we end up with "Privacy" and "Website Privacy" next to each other?

craigcook:
> So we end up with "Privacy" and "Website Privacy" next to each other?

Correct. It's... not ideal. But since we're losing the general privacy link on the new about page we need that general link elsewhere, but also need to link to the website policy specifically. The cookies link actually goes to a section of the same page but I think EU regulations require a specific "cookies" link. It's all pretty awkward but I think we do need all three links.

Also [bug 1483624](https://bugzilla.mozilla.org/show_bug.cgi?id=1483624) was just filed so we might need to find a spot to cram another link in...

craigcook:
On second thought let's do the community participation link separately...

PR #6018, opened 2018-08-15
Fix #5969: Remove Facebook container funnelcake.

### LINKED ISSUE #5969: Remove Facebook container funnelcake
Marketing has ceased to this funnelcake, so we can turn it off (mozmeao/www-config#125) and remove all associated code.

For reference, #5555 is the PR that added the funnelcake.

amychurchwell:
## Description
Turn off and remove associated code.
(Undo PR #5555)

## Issue / Bugzilla link
#5969


stephaniehobson:
One of the tests is failing. You should be able to click "Details" and see the full error message - which is HUGE and CONFUSING so I'll pull out the relevant bit here:

```
bedrock/firefox/views.py:413:1: F841 local variable 'variation' is assigned to but never used
bedrock/firefox/views.py:421:12: E111 indentation is not a multiple of four
1       E111 indentation is not a multiple of four
1       F841 local variable 'variation' is assigned to but never used
Makefile:79: recipe for target 'test-image' failed
```

Looks like it's just the Python linter being picky about indentation and a variable that isn't in use.

bedrock/firefox/views.py
```diff
@@ -409,9 +408,6 @@ def get_context_data(self, **kwargs):
     def get_template_names(self):
         version = self.kwargs.get('version') or ''
 
-        experience = self.request.GET.get('xv', None)
-        locale = l10n_utils.get_locale(self.request)
```
stephaniehobson: I think the most recent error is because this line needs to stay.

alexgibson:
As a side note, you can run the exact same tests locally that run inside of CircleCI using the command `make test` (assuming you're running docker). If you're running bedrock locally, there are a slightly different set of instructions [here](https://bedrock.readthedocs.io/en/latest/install.html#run-the-tests).

amychurchwell:
@alexgibson I believe it's good to go.

PR #6035, opened 2018-08-16
Fix #5851: Support overriding or ammending list of locales

### LINKED ISSUE #5851: Support locale-specific templates with no lang file activation
https://bugzilla.mozilla.org/show_bug.cgi?id=1415964

pmac:
This modifies how several parts of translation in bedrock work:

1. You can now provide a list of locales in `active_locales` in the template context that will be the list of available translations for the page.
2. You can now provide a list of locales in `add_active_locales` to add extra available translations for the page.
3. The `page()` and `LangFilesMixin` helpers are updated to support these new context items.
4. You can now create and use `en-US` locale-specific-templates.

### DIFF
--- a/bedrock/mozorg/util.py
+++ b/bedrock/mozorg/util.py
@@ -51,8 +51,14 @@ def page(name, tmpl, decorators=None, url_name=None, **kwargs):
         be applied to the view.
     @param url_name: The value to use as the URL name, default is to coerce
         the template path into a name as described above.
+    @param active_locales: A list of locale codes that should be active for this page
+        regardless of the state of the lang files. Useful for pages with locale-
+        specific templates or non-English text in the template. Ignores the lang
+        file activation tags.
+    @param add_active_locales: A list of locale codes that should be active for this page
+        in addition to those from the lang files.
     @param kwargs: Any additional arguments are passed to l10n_utils.render
-        after the request and the template name.
+        as the context.
     """
     pattern = r'^%s/$' % name if name else r'^$'
 
--- a/docs/l10n.rst
+++ b/docs/l10n.rst
@@ -452,12 +452,69 @@ When a request is made for a particular page, bedrock's rendering function
 automatically checks for a locale-specific template, and, if one exists, will
 render it instead of the originally specified (locale-agnostic) template.
 
+.. NOTE::
+
+    Creating a locale-specific template for en-US was not possible when this
+    feature was introduced, but it is now. So you can create your en-US-only
+    template and the rest of the locales will continue to use the default.
+
 .. IMPORTANT::
 
     Note that the presence of an L10n template (e.g.
     ``locale/de/templates/firefox/new.html``) will take precedence over
     a locale-specific template in bedrock.
 
+
+Specifying Active Locales in Views
+----------------------------------
+
+Normally we rely on activation tags in our translation files (.lang files)
+to determine in which languages a page will be available. This will almost always
+be what we want for a page. But sometimes we need to explicitly state the locales
+available for a page. The `impressum` page for example is only available in German
+and the template itself has German hard-coded into it since we don't need it to be
+translated into any other languages. In cases like these we can send a list of locale
+codes with the template context and it will be the final list. This can be accomplished
+in a few ways depending on how the view is coded.
+
+For a plain view function, you can simply pass a list of locale codes to `l10n_utils.render`
+in the context using the name `active_locales`. This will be the full list of available
+translations. Use `add_active_locales` if you want to add languages to the existing list:
+
+.. code-block:: python
+
+    def french_and_german_only(request):
+        return l10n_utils.render(request, 'home.html', {'active_locales': ['de', 'fr'])
+
+If you don't need a custom view and are just using the `page()` helper function in your `urls.py`
+file, then you can similarly pass in a list:
+
+.. code-block:: python
+
+    page('about', 'about.html', active_locales=['en-US', 'es-ES']),
+
+Or if your view is even more fancy and you're using a Class-Based-View that inherits from `LangFilesMixin`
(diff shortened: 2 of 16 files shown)


pmac:
I would love to hear your opinions on this @alexgibson @jpetto @craigcook @stephaniehobson. See the changes to the l10n docs in the PR for (hopefully) better descriptions of the new features.

pmac:
Tests are passing! Please let me know your thoughts and we can get this in. Should help many l10n tasks.

pmac:
FYI @peiying2

docs/l10n.rst
```diff
@@ -452,13 +452,70 @@ When a request is made for a particular page, bedrock's rendering function
...
+
+    def french_and_german_only(request):
+        return l10n_utils.render(request, 'home.html', {'active_locales': ['de', 'fr'])
+
+If you don't need a custom view and are just using the `page()` helper function in your `urls.py`
+file, then you can similarly pass in a list:
+
+.. code-block:: python
+
+    page('about', 'about.html', active_locales=['en-US', 'es-ES']),
```
stephaniehobson: In this case would we also have to put the strings in for translation to Spanish in the traditional way? And would we not be able to deploy the page until all the strings are translated in Ponton?

stephaniehobson: Or this some kind of magical theoretical page that is appropriate to display in both without translation?

lib/l10n_utils/__init__.py
```diff
@@ -55,55 +57,31 @@ def render(request, template, context=None, **kwargs):
...
     context['template_source_url'] = template_source_url(template)
 
-    # Get the available translation list of the current page
-    context.setdefault('translations', {})
-    context['translations'].update(translations_for_template(template))
-
-    # Look for localized template if not default lang.
-    if hasattr(request, 'locale') and request.locale != settings.LANGUAGE_CODE:
-
+    # if `locales` is given use it as the full list of active translations
```
stephaniehobson: Does this over-ride the language files then?

docs/l10n.rst
stephaniehobson: Further reading suggests different content will be served for each locale but I'm not able to figure out where one should store that different content to have it delivered. More documentation required maybe?

pmac: The idea is that it would be used with locale-specific templates that you just hard-code in the language of the template. So it's basically saying that the dev team is taking responsibility for the translation of these pages. It should rarely be used, but it's a means to prevent the l10n team from having to create empty .lang files with the `## active ##` tag at the top.

lib/l10n_utils/__init__.py
pmac: In this case yes. In this way you can have a page (e.g. `/about/legal/impressum/`) that is not available in en-US.

stephaniehobson:
These changes make sense to me 👍

bedrock/mozorg/util.py
```diff
@@ -51,8 +51,14 @@ def page(name, tmpl, decorators=None, url_name=None, **kwargs):
         be applied to the view.
     @param url_name: The value to use as the URL name, default is to coerce
         the template path into a name as described above.
+    @param active_locales: A list of locale codes that should be active for this page
+        regardless of the state of the lang files. Useful for pages with locale-
+        specific templates or non-English text in the template. Ignores the lang
+        file activation tags.
+    @param add_active_locales: A list of locale codes that should be active for this page
```
alexgibson: It would be nice to maybe add an example of using `add_active_locales` in the docs? I maybe can see us using this more often than `active_locales`, especially for locale-specific pages where templates may not have their own separate view or URL definition.

For example, with pages [like this one](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/whatsnew/index.id.html), instead of doing `
{% add_lang_files "firefox/whatsnew" %}` in the template, would we just be able to do the following in the view?

```python
add_active_locales = ['id']
```

Currently if we don't do the above hack, then the page is not accessible in production when `Dev=False`. We also have a similar issue when pages are en-US-only, but we forget to add it so that other translations are not available in the language switcher in the footer.

alexgibson:
Aside from one question / suggestion this all sounds good to me. nice work! 💯

bedrock/mozorg/util.py
alexgibson: I was just thinking of how it works preior to this PR, but I believe that's the case yeah, e.g. https://github.com/mozilla/bedrock/pull/6049

pmac: I do mention using `add_active_locales` in the docs and that you can use it in all the places I gave examples for `active_locales`:

https://github.com/mozilla/bedrock/pull/6035/files#diff-8c556c7af0e5b741b54a3567ed7e333cR513

You think I  should add some specific examples of `add_active_locales` as well?

pmac: > I was just thinking of how it works preior to this PR, but I believe that's the case yeah, e.g. #6049

This one is different actually. `about-en.html` is not a locale-specific-template, it's a fully separate template. The system can't know that it's meant as a variation of a template unless it's told. So I think adding the `add_langfile` tag to that one is appropriate, but the original template name lang file should be used when using the locale-specific-template feature of bedrock.

pmac: @alexgibson does the above make sense? If so I'd like to get this in before it gets too far out-of-date.

alexgibson: @pmac sorry I messed this, you're right ans yes that makes good sense. Thumbs up from me 👍

pmac:
I've rebased this and I think it's ready. Unless anyone objects I'd appreciate someone adding an approved review so we can land this. Thanks for all the review help!

PR #6044, opened 2018-08-20
Fix Bug 1339226 - Update download buttons for Windows XP/Vista users once Firefox 52 ESR reaches EOL

kyoshino:
## Description

* Firefox 52 ESR reaches the EOL on September 5. The Firefox download buttons have to be updated to show a warning to Windows XP/Vista users while allowing downloads
* See [my bug comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1339226#c17) for details and screenshots
* I don’t know how Protocol works so didn’t add any warning styling to `media/css/protocol/components/_download-button.scss`

## Localization

There are 2 new short strings. The latter is TBD

```html
<a href="%(url)s">Mozilla no longer provides security updates for Firefox on Windows XP or Vista</a>, but you can still download the final Windows 32-bit version below.
```
```html
You’re using an insecure, outdated operating system <a href="%(url)s">no longer supported by Firefox</a>.
```

## Issue / Bugzilla link

[Bug 1339226](https://bugzilla.mozilla.org/show_bug.cgi?id=1339226#c17)

## Testing

Testing with a physical/virtual XP/Vista machine would be great, but you can still see the warnings by changing the class name on `<html>` to `windows xpvista x86 js loaded`

### DIFF
--- a/bedrock/firefox/templates/firefox/new/scene2.html
+++ b/bedrock/firefox/templates/firefox/new/scene2.html
@@ -58,6 +58,9 @@
 {% block content %}
 <main class="main mzp-t-firefox mzp-t-dark">
   <section class="status">
+    <div class="mzp-l-content mzp-t-firefox windows-xpvista">
+      <p>{{ _('You’re using an insecure, outdated operating system <a href="%(url)s">no longer supported by Firefox</a>.')|format(url='https://support.mozilla.org/kb/end-support-windows-xp-and-vista') }}</p>
+    </div>
     <div class="mzp-l-content mzp-t-firefox">
       {# fallback_url is replaced by the platform download link via JS, but if
         something fails the user should still get a link to a working download path. #}
--- a/media/css/firefox/all.less
+++ b/media/css/firefox/all.less
@@ -25,14 +25,41 @@
         .span-all();
         margin-top: @baseLine;
     }
-    p.warning {
-        font-size: @largeFontSize;
-    }
     li {
         margin: 0;
         list-style-type: none;
         a:after {
             content: "\00A0\00BB";
+        }
+    }
+
+    .warning {
+        position: relative;
+        .border-box();
+        border: 2px solid #FFA500;
+        padding: 10px 15px;
+        color: @textColorSecondary;
+        background-color: #FAFAD2;
+        font-size: @largeFontSize;
+        text-align: left;
+
+        &.xpvista {
+            display: none;
+
+            .windows.xpvista & {
+                display: block;
+            }
+        }
+
+        a:link,
+        a:visited,
+        a:hover,
+        a:focus,
+        a:active {
+            color: @textColorSecondary;
+            text-decoration: underline;
+            .open-sans-light();
+            font-weight: bold;
         }
     }
 }
--- a/media/css/firefox/new/scene2.scss
+++ b/media/css/firefox/new/scene2.scss
@@ -41,6 +41,25 @@ $image-path: '/media/protocol/img';
     p {
         margin: 0;
     }
+
+    .windows-xpvista {
+        display: none;
+        max-width: none;
+        color: $color-black;
+        background-color: $brand-lemon-yellow;
+
+        .windows.xpvista & {
+            display: block;
+        }
+
+        a:link,
+        a:visited,
+        a:hover,
+        a:focus,
+        a:active {
+            color: $color-black;
+        }
+    }
 }
 
 .main {
--- a/media/js/base/site.js
+++ b/media/js/base/site.js
@@ -146,10 +146,17 @@
         // to avoid lots of flickering
         var platform = window.site.platform = window.site.getPlatform();
         var version = window.site.platformVersion = window.site.getPlatformVersion();
+        var _version = version ? parseFloat(version) : 0;
 
         if (platform === 'windows') {
+            // Add class to allow Windows XP/Vista users to download Firefox 52 ESR, though these
+            // legacy platforms are now officially unsupported
+            if (_version >= 5.1 && _version <= 6) {
+                h.className += ' xpvista';
+            }
+
             // Add class to support downloading Firefox for Windows 64-bit on Windows 7 and later
-            if (version && parseFloat(version) >= 6.1) {
+            if (_version >= 6.1) {
                 h.className += ' win7up';
             }
         } else {
--- a/bedrock/firefox/templates/firefox/all.html
+++ b/bedrock/firefox/templates/firefox/all.html
@@ -104,7 +104,9 @@
     <h1>{{ self.page_title()  }}</h1>
     <h2>{{ self.page_desc() }}{% if self.learn_more_url() %} <a href="{{ self.learn_more_url() }}">{{ _('Learn more') }}</a>{% endif %}</h2>
     {% if channel == 'nightly' %}
-    <p class="warning">{{ _('<strong>Warning</strong>: Nightly is updated daily. It is most appropriate for core Mozilla contributors and early adopters, not regular users.') }}</p>
+    <p class="warning nightly">{{ _('<strong>Warning</strong>: Nightly is updated daily. It is most appropriate for core Mozilla contributors and early adopters, not regular users.') }}</p>
+    {% elif channel == 'release' %}
+    <p class="warning xpvista">{{ _('<a href="%(url)s">Mozilla no longer provides security updates for Firefox on Windows XP or Vista</a>, but you can still download the final Windows 32-bit version.')|format(url='https://support.mozilla.org/kb/end-support-windows-xp-and-vista') }}</p>
     {% endif %}
     <ul>
       {% if channel != 'nightly' %}
--- a/bedrock/firefox/templates/firefox/new/scene1.html
+++ b/bedrock/firefox/templates/firefox/new/scene1.html
@@ -31,6 +31,7 @@
     <p class="version-message android-old">{{_('<a href="%(url)s">Update</a> your Firefox for the latest in speed and privacy.')|format(url='https://support.mozilla.org/kb/update-latest-version-firefox-android')}}</p>
     <p class="version-message desktop-old">{{_('<a href="%(url)s">Update</a> your Firefox for the latest in speed and privacy.')|format(url='https://support.mozilla.org/kb/update-firefox-latest-version') }}</p>
     <p class="version-message firefox-pre-release">{{_('You’re using a pre-release version of Firefox.') }}</p>
+    <p class="version-message windows-xpvista">{{_('You’re using an insecure, outdated operating system <a href="%(url)s">no longer supported by Firefox</a>.')|format(url='https://support.mozilla.org/kb/end-support-windows-xp-and-vista') }}</p>
   </div>
 </div>
 {% endblock %}
--- a/media/css/firefox/new/scene1.scss
+++ b/media/css/firefox/new/scene1.scss
@@ -109,6 +109,20 @@ html.android.firefox-old {
     }
 }
 
+html.windows.xpvista {
+    .version-message-container {
+        display: block;
+    }
+
+    .version-message.windows-xpvista {
+        display: inline-block;
+
+        &:before {
+            background-image: url('/media/img/firefox/new/icon-alert.svg');
+        }
+    }
+}
+
 .linux-arm-download-instructions {
     display: none;
     margin-top: 40px;
--- a/media/js/firefox/new/scene1.js
+++ b/media/js/firefox/new/scene1.js
@@ -67,6 +67,11 @@
     };
 
     var setFirefoxStatus = function() {
+        // Windows XP/Vista warning overrides the Firefox status message
+        if ($html.hasClass('xpvista')) {
+            return;
+        }
+
         var status = getFirefoxStatus();
         $html.addClass(status);
 


kyoshino:
@peiying2 The strings are now confirmed. Can you extract the stuff?

cpeterso:
Thanks, @kyoshino! The strings in your PR LGTM.

kyoshino:
Hey @peiying2 can you take a look at this?

peiying2:
Yes, working on this now.

cpeterso:
Hi @peiying2, do you have any updates on your review of Kohei's Windows XP download warning? Firefox ESR (52.9) will EOL support for Windows XP next week, so it would be good to have the download warning in place then. :)

alexgibson:
Hi @cpeterso - this PR still needs code review before it can be merged. Apologies no one has got to this yet, it snuck under our team's radar.

media/css/pebbles/components/_buttons-download.scss
```diff
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 @import '../includes/lib';
+@import '../includes/fonts/font-awesome';
```
alexgibson: It would be nice to avoid needing to load font-awesome just for one tiny icon. I'd suggest using a simple png instead here.

media/css/pebbles/components/_buttons-download.scss
```diff
@@ -93,13 +95,72 @@ ul.download-list {
...
+    .windows-xpvista-download,
     .linux-arm-download,
     .unrecognized-download,
     .unsupported-download,
     .unsupported-download-osx,
     .nojs-download {
         display: none;
     }
+
+    .windows-xpvista-download {
```
alexgibson: While this warning is certainly eye-catching, I do wonder if it's maybe a little excessive. I'd love to get @dzingeek's thoughts here from a design perspective. It doesn't look or feel like any other kind of warning we have on the site currently.

screen shot 2018-08-30 at 10 21 23

media/css/pebbles/components/_buttons-download.scss
```diff
@@ -93,13 +95,72 @@ ul.download-list {
...
+        position: relative;
+        @include border-box;
+        margin: 0 0 20px 0;
+        border-width: 2px 2px 2px 0;
+        border-style: solid;
+        border-color: orange;
```
alexgibson: Nit: please use hex colors

media/js/base/site.js
```diff
@@ -146,12 +146,17 @@
...
 
         if (platform === 'windows') {
             // Add class to support downloading Firefox for Windows 64-bit on Windows 7 and later
-            if (version && parseFloat(version) >= 6.1) {
+            if (_version >= 6.1) {
                 h.className += ' win7up';
             }
+            // Add class to support downloading Firefox 52 ESR on Windows XP/Vista
+            if (_version >= 5.1 && _version <= 6) {
+                h.className += ' xpvista';
```
alexgibson: Some unit tests here would be good.

media/css/pebbles/components/_buttons-download.scss
```diff
@@ -66,6 +67,7 @@ ul.download-list {
 
 .download-button {
     display: inline-block;
+    text-align: center;
```
alexgibson: This is going to change all text and messaging within the download button to centered, which looks odd on some pages where text is left or right aligned. Is there a reason to change it?

bedrock/firefox/templates/firefox/includes/download-button.html
```diff
@@ -44,6 +46,10 @@
   </div>
   {% if show_desktop %}
     {# unrecognized/unsupported platform #}
+    <p class="windows-xpvista-download">
+      <span class="icon" aria-hidden="true"></span>
+      {{ _('<a href="%(url)s">Mozilla no longer provides security updates for Firefox on Windows XP or Vista</a>, but you can still download the final Windows 32-bit version below.')|format(url='https://support.mozilla.org/kb/end-support-windows-xp-and-vista') }}
```
alexgibson: From an SEO point of view we're adding this text numerous times to just about every page on the website. I'm not sure if there's any real impact here, but I'd be curious to know if @rraue has any thoughts here?

The vast majority of pages on mozorg no longer serve first-class CSS & JS to IE < 9, with the only real exception being our main download page on `/firefox/new/`. Another option could be to add it only to `/firefox/new/`, as opposed to everywhere.

media/css/firefox/all.less
```diff
@@ -4,6 +4,14 @@
 
 @import "../sandstone/lib.less";
 
+@font-face {
```
alexgibson: Again, let's try and avoid using icon fonts. They are heavy to download, especially when only using a single icon.

bedrock/firefox/templates/firefox/includes/download-button.html
rraue: Less text, not being related to the actual page, in the body is better than more, I agree @alexgibson. On the firefox pages it doesn't harm anything as it is text related to the topic of the page.

alexgibson: I'd be interested in @cpeterso's thoughts here. Is this message really needed everywhere, or just on our main download pages? e.g. `/firefox/new/`, `/firefox/`, `/firefox/all/`?

media/css/pebbles/components/_buttons-download.scss
kyoshino: Thanks @alexgibson! Will replace Font Awesome with a small SVG file.

alexgibson: I think SVG is likely not to be shown on older systems, so I'd either find a png or drop the icon.

media/css/pebbles/components/_buttons-download.scss
kyoshino: Forgot to mention but this design is supposed to be tentative. The Protocol would probably have a style for such a warning?

dzingeek: Right good call Alex. Let me think a bit more on this and also bring @vincejoyiv into the mix, he might already have some ideas on this.

kyoshino: The warning above the button has to be prominent anyway IMO, because people are downloading the unsupported/insecure version. We intentionally allow to do it only for XP/Vista users.

media/js/base/site.js
kyoshino: IIRC we have never had unit tests for the class names on `<body>` but I agree to have some tests. So I’m making a change to `getPlatform()` in [site.js](https://github.com/mozilla/bedrock/blob/master/media/js/base/site.js) and update the relevant test accordingly. Basically Windows XP and Vista should be `oldwin`.

media/css/pebbles/components/_buttons-download.scss
kyoshino: It aims to align the download button and the Privacy Notice link. Without this, [it looks rather odd](https://screenshots.firefox.com/t286AJZDpqHVxol2/localhost).

media/js/base/site.js
kyoshino: Updated `getPlatform()` to detect XP/Vista as `oldwin` while using `windows xpvista` as the class name on `<body>` to continue showing the download button on these legacy platforms.

media/css/pebbles/components/_buttons-download.scss
alexgibson: I'm wondering if this message really needs to be part of the download button itself, or if it would be enough just to include this as a message on our main Firefox download pages. See https://github.com/mozilla/bedrock/pull/6044#discussion_r213969504

bedrock/firefox/templates/firefox/includes/download-button.html
alexgibson: I'm not sure if it can easily be repurposed, not at least while the existing banner is running.

cpeterso:
> I'd be interested in @cpeterso's thoughts here. Is this message really needed everywhere, or just on our main download pages? e.g. /firefox/new/, /firefox/, /firefox/all/?

It would be nice to make the XP message available everywhere a user might download Firefox, but most users who land on these pages will probably download Firefox regardless of any warnings. My only question is whether they will download an XP compatible version of Firefox:

@kyoshino : Without your code changes, what Firefox installer version will the /firefox/, /firefox/new/, and /firefox/all/ pages serve to XP users? Will they still get ESR 52.9 (XP compatible but not supported)? Or will they get the latest Firefox version (not XP compatible)?

kyoshino:
@cpeterso I thought the bouncer had been redirecting all downloads from XP/Vista to 52 ESR. That’s [Bug 1334538](https://bugzilla.mozilla.org/show_bug.cgi?id=1334538) maybe.

cpeterso:
> Is this message really needed everywhere, or just on our main download pages? e.g. /firefox/new/, /firefox/, /firefox/all/?

@alexgibson , what are the non-main download pages?

> I thought the bouncer had been redirecting all downloads from XP/Vista to 52 ESR. That’s Bug 1334538 maybe.

Oh yes. I think you're right! It would be nice to publish the XP warning messages before 62 ships next week, but the bouncer redirect means there is no big rush or hard deadline! XP users will still the appropriate installer. The XP warning message will still be useful longer term.

alexgibson:
> Oh yes. I think you're right! It would be nice to publish the XP warning messages before 62 ships next week, but the bouncer redirect means there is no big rush or hard deadline! XP users will still the appropriate installer. The XP warning message will still be useful longer term.

Ok great, it's good to know there's no longer a mad rush to get this in. It will also give a little more opportunity for our designers to take a look :) 

This warning message is obviously important to users on XP/Vista, but we need to take a minute to think about the best way to show it, and where to show it.

I'll try and summarise my thought here:

- Adding a warning box like this to the download button itself means it could literally show up in hundreds of places. This does cover all our bases so to speak, but it could also break UI's where the button is shown in a confined space, so it does make it more time consuming to QA.
- Adding the message to the download button will result in the markup appearing numerous times on just about every web page on mozilla.org. That's thousands of web pages (all locales). While the message is hidden, it will still be picked up by search engines, which is confusing for web pages that aren't directly related to Firefox itself.
- Aside from our main Firefox download page, most pages on mozilla.org no longer serve first-class CSS & JS to IE < 9, so the message may not be as effective to many users on XP/Vista, where they already get a very bare-bones stylesheet (text is readable, people can download, but little else).

If showing this along side every download button is a must then we can do this certainly, but it may take a little time to review. The alternative would be to add the message just to our main download funnel entry points (`/firefox/`, `/firefox/new/`, `/firefox/all/`), but it really depends on just how vital/effective we think this message will be.

Thanks for your patience on this while we work it out.

alexgibson:
> Instead of adding a warning to each download button, it’s perhaps better to have the warning on the thanks page, and force XP/Vista users to explicitly click a download button on the page while preventing auto download.

I'd be in favor of a message on the /thanks/ page as it keeps things really simple. I don't think we should prevent or change the download flow itself though, as long as the warning is clear enough.

kyoshino:
The string has already been translated and it says 

> Mozilla no longer provides security updates for Firefox on Windows XP or Vista, but you can still download the final Windows 32-bit version **below**.

So a download button should exist below the message 😅

alexgibson:
That's pretty unfortunate, then I don't think we should put the message on `/thanks/`. Forcing users to click download twice is a step too far i think :(

cpeterso:
Since we don't need to prevent XP users from downloading the Firefox installer, showing the XP warning on the /thanks/ page that is opened after the user clicks the download button would be OK.

alexgibson:
> Since we don't need to prevent XP users from downloading the Firefox installer, showing the XP warning on the /thanks/ page that is opened after the user clicks the download button would be OK. The XP warning message would need to be changed and retranslated to not refer to the download button, as noted above. :)

@peiying2 @kyoshino is this something we can do? I think it makes the most sense from a technical perspective, whilst still also meeting user needs. It's unfortunate the strings went out for translation before this was reviewed, so apologies.

kyoshino:
So we should probably strip “below” from the end of the string. @peiying2 I’m sorry to bother you but can you do this?

```
<a href=“%(url)s”>Mozilla no longer provides security updates for Firefox on Windows XP or Vista</a>, but you can still download the final Windows 32-bit version.
```

kyoshino:
Moved the warning message to the thanks page. ([screenshot](https://screenshots.firefox.com/nmtHQQbYxIB38cka/localhost))

bedrock/firefox/templates/firefox/new/scene2.html
```diff
@@ -34,6 +34,11 @@ <h1>{{ _('Let’s do this!') }}</h1>
   {{ download_firefox(force_direct=true, dom_id='primary-download-button') }}
 </div>
 
+<p class="windows-xpvista-download">
+  <span class="icon" aria-hidden="true"></span>
+  {{ _('<a href="%(url)s">Mozilla no longer provides security updates for Firefox on Windows XP or Vista</a>, but you can still download the final Windows 32-bit version.')|format(url='https://support.mozilla.org/kb/end-support-windows-xp-and-vista') }}
```
alexgibson: We're about to land a new default template for scene2 in #6092. Once that merges, we should rebase and make sure the message is visible on both templates.

media/css/firefox/new/scene2.scss
```diff
@@ -20,14 +20,71 @@
...
+    border-color: #FFA500;
+    padding: 5px 10px 5px 40px;
+    min-height: 20px;
+    color: $color-text-secondary;
+    background-color: #FAFAD2;
+    @include font-size(14px);
+    text-align: left;
+
+    .moz-sub-nav.stuck & {
+        display: none !important;
```
alexgibson: Stylelint is failing due to this `!important` declaration. Is this needed?

bedrock/firefox/templates/firefox/new/scene2.html
alexgibson: I also wonder if the copy here might be confusing still. The user has already initiated the download, do they need to download something else, where is the 32bit version?

Perhaps the existing string from scene1 would make more sense here?

kyoshino: Okay, adding the Blocked label then.

alexgibson: @kyoshino please see my suggestion on using the string from scene1 here instead.

alexgibson: @kyoshino I don't think this issue needs to be blocked here, we can reuse the existing strings we have already I think.

alexgibson:
@kyoshino if you no longer have time to work on this, please let me know and I can try and make the last changes we need here. Thanks

kyoshino:
@alexgibson Thanks for your ping and sorry I couldn’t work on this last week. Let me finish this up.

alexgibson:
Now that https://github.com/mozilla/bedrock/pull/6092 has merged, we should make sure to add the warning to both /thanks/ templates in this PR.

kyoshino:
* Updated the warning to match the new scene2 template, using the same message as scene1: “You’re using an insecure, outdated operating system no longer supported by Firefox.”
* Simplified the style, removing the icon
* Fixed the legible text on IE due to `color: inherit `

media/js/base/site.js
```diff
@@ -12,10 +12,10 @@
             ua = ua || navigator.userAgent;
 
             if (/Win(16|9[x58]|NT( [1234]| 5\.0| [^0-9]|[^ -]|$))/.test(ua) ||
-                    /Windows ([MC]E|9[x58]|3\.1|4\.10|NT( [1234]\D| 5\.0| [^0-9]|[^ ]|$))/.test(ua) ||
+                    /Windows ([MC]E|9[x58]|3\.1|4\.10|NT( [12345]\D| 6\.0| [^0-9]|[^ ]|$))/.test(ua) ||
```
alexgibson: This change is breaking downloads in IE < 9, since the class name is now `oldwin` instead of `windows` https://github.com/mozilla/bedrock/blob/master/media/js/ie8/mozilla-utils-ie8.js#L25

Thinking about it, I think changing this logic now might be a bit confusing in itself. Either we're supporting XP / Vista or we're not. We seem to want to continue allowing people to download it for now and just show a warning, so I think we should probably treat it as "supported" until the time comes when we officially want to stop people downloading it, like we do for other old systems. Right now I would read this code and assume XP/Vista should see the standard unsupported message, which isn't the case.

media/css/firefox/new/scene2_email.scss
```diff
@@ -41,6 +41,24 @@ $image-path: '/media/protocol/img';
     p {
         margin: 0;
     }
+
+    .windows-xpvista {
+        display: none;
+        color: $color-black;
+        background-color: #FAFAD2;
```
alexgibson: This isn't one of our [brand colors](https://protocol.mozilla.org/fundamentals/color.html). I'd suggest changing it to `$brand-lemon-yellow` and allowing the warning to run full width of the screen:

screenshot_2018-09-13 download firefox free web browser 1

media/js/base/site.js
```diff
@@ -146,12 +146,17 @@
...
         if (platform === 'windows') {
             // Add class to support downloading Firefox for Windows 64-bit on Windows 7 and later
-            if (version && parseFloat(version) >= 6.1) {
+            if (_version >= 6.1) {
                 h.className += ' win7up';
             }
+        } else if (platform === 'oldwin' && _version >= 5.1 && _version <= 6) {
+            // Add class to allow Windows XP/Vista users to download Firefox 52 ESR, though these
+            // legacy platforms are now officially unsupported
+            h.className += ' xpvista';
```
alexgibson: This is confusing since the system class name ends up as `windows xpvista`, yet `window.site.platform` returns `oldwin`

media/js/base/site.js
kyoshino: Hmm, okay. Then I'll revert this change to continue using `oldwin` for XP/Vista. I'll remove the test as well, which is mentioned earlier but makes things complicated.

kyoshino:
Rebased and squashed. Thanks for your review, @alexgibson!

PR #6067, opened 2018-08-22
Fix bug 1485175: Add anchors to certs policy page

pmac:
Also update the process in a comment on the page to include anchor generation in future updates.


pmac:
That is a good point, but these are specifically for a table-of-contents and are generated by the Markdown TOS plugin. Also they asked for the number based ones in the bug. I could push back there, but this seems like the path-of-least-resistance for now.

PR #6071, opened 2018-08-23
Truncate stub attribution code if over max length

pmac:
See https://github.com/mozilla-services/stubattribution/issues/82

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -7,7 +7,6 @@ import hashlib
 import hmac
 import re
 from collections import OrderedDict
-from time import time
 from urlparse import urlparse
 
 from django.conf import settings
@@ -78,12 +77,10 @@ def stub_attribution_code(request):
         return HttpResponseJSON({'error': 'Resource only available via XHR'}, status=400)
 
     response = None
-    rate = settings.STUB_ATTRIBUTION_RATE
-    key = settings.STUB_ATTRIBUTION_HMAC_KEY
-    if not rate:
+    if not settings.STUB_ATTRIBUTION_RATE:
         # return as though it was rate limited, since it was
         response = HttpResponseJSON({'error': 'rate limited'}, status=429)
-    elif not key:
+    elif not settings.STUB_ATTRIBUTION_HMAC_KEY:
         response = HttpResponseJSON({'error': 'service not configured'}, status=403)
 
     if response:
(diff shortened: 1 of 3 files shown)


bedrock/firefox/views.py
```diff
@@ -119,15 +117,33 @@ def stub_attribution_code(request):
...
+    """
+    Take the attribution codes and return the base64 encoded string
+    respecting max length and HMAC signature.
+    """
+    key = settings.STUB_ATTRIBUTION_HMAC_KEY
+    code = b64_encode_codes(codes)
+    while len(code) > settings.STUB_ATTRIBUTION_MAX_LEN:
+        # remove 5 char at a time
+        codes['campaign'] = codes['campaign'][:-5] + '_'
+        code = b64_encode_codes(codes)
```
jpetto: Could this potentially result in an infinite loop if `codes['campaign']` gets cut to zero characters, but the string remains longer than the max (due to some other param being long)?

pmac: ooooooo... great catch.. yes. I'll update.

pmac:
I'm in the process of changing this. Since the changes from the above-linked issue for the attribution service have landed and will go to prod very soon (likely today) I'm changing this to complement those changes.

bedrock/firefox/views.py
```diff
@@ -118,16 +115,39 @@ def stub_attribution_code(request):
...
+
+
+def sign_attribution_codes(codes):
+    """
+    Take the attribution codes and return the base64 encoded string
+    respecting max length and HMAC signature.
+    """
+    key = settings.STUB_ATTRIBUTION_HMAC_KEY
+    code = get_attrribution_code(codes)
+    print len(code)
```
jpetto: 🙅‍♂️

jpetto: 🖨

bedrock/firefox/views.py
```diff
@@ -118,16 +115,39 @@ def stub_attribution_code(request):
...
+def get_attrribution_code(codes):
+    """
+    Take the attribution codes and return the URL encoded string
+    respecting max length.
+    """
     code = '&'.join('='.join(attr) for attr in codes.items())
+    if len(codes['campaign']) > 5 and len(code) > settings.STUB_ATTRIBUTION_MAX_LEN:
+        # remove 5 char at a time
+        codes['campaign'] = codes['campaign'][:-5] + '_'
+        code = get_attrribution_code(codes)
```
jpetto: If I'm reading this right, if a param other than `utm_campaign` makes this string longer than `STUB_ATTRIBUTION_MAX_LEN`, we allow it? This is likely an edge case, but curious as to the intended handling.

bedrock/firefox/views.py
```diff
@@ -118,16 +115,39 @@ def stub_attribution_code(request):
...
+    patch_response_headers(response, 300)  # 5 min
+    return response
+
+
+def get_attrribution_code(codes):
+    """
+    Take the attribution codes and return the URL encoded string
+    respecting max length.
+    """
     code = '&'.join('='.join(attr) for attr in codes.items())
```
jpetto: Just clarifying that we're checking the length of `code` prior to base64 encoding?

bedrock/firefox/views.py
pmac: DOH!

bedrock/firefox/views.py
pmac: Yeah. My thinking is that if we can't seem to fix it just send it on and hope for the best. I didn't see the point in blocking it. Though maybe we should since we know the attribution service will?

bedrock/firefox/views.py
pmac: Right. This is what the new changes to the attribution service do. There's a 200 char limit on the decoded data.

bedrock/firefox/views.py
pmac: Updated to return 400 error when code would be > max rather than sending a code that would be too long anyway. We can adjust max via a setting so this should be decently future proof.

PR #6074, opened 2018-08-24
Add channel to system requirements page title

stephaniehobson:
## Description
Add channel to system requirements page title. Except if it's release, omit channel if it's release.

## Issue / Bugzilla link
mozilla/bedrock/issues#6053


stephaniehobson:
Two spaces are treated as on by HTML so I'm not worried about the extra space.

Issue #6004, opened 2018-08-13
Set up traffic cop for - /new variations for Safari visitors


ejregithub:
https://bugzilla.mozilla.org/show_bug.cgi?id=1479085

Content assets

Success Criteria:
- Traffic cop is live/in place to route traffic according to spec in the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1479085) to page [version One](https://github.com/mozilla/bedrock/issues/5950) and [version Two](https://github.com/mozilla/bedrock/issues/6003)

Issue #6003, opened 2018-08-13
Code V2 - /new variation for Safari visitors


ejregithub:
https://bugzilla.mozilla.org/show_bug.cgi?id=1479085

[Content assets](https://share.goabstract.com/c89ffc12-ca7b-4848-844d-04db048a5ba2)

Success Criteria:
- Version number 2 is live (not dependent on [Version 1](https://github.com/mozilla/bedrock/issues/5950) or Traffic cop)

craigcook:
https://bedrock-demo-craigcook.oregon-b.moz.works/en-US/firefox/new/?xv=safari&v=2

Issue #5950, opened 2018-07-27
Code V1 - /new variation for Safari visitors


ejregithub:
https://bugzilla.mozilla.org/show_bug.cgi?id=1479085

[Content assets](https://share.goabstract.com/c89ffc12-ca7b-4848-844d-04db048a5ba2)

stephaniehobson:
Time estimates:
Code V1: L
Code V2: M
Traffic Cop: S

stephaniehobson:
I'm going to take V2, the "comparison table" version with the goal of having it in for review before I leave EOD Wednesday.

craigcook:
https://bedrock-demo-craigcook.oregon-b.moz.works/en-US/firefox/new/?xv=safari&v=1

craigcook:
There was no copy deck, only the Sketch file.

thejeffreytaylor:
Please see Sung for all copy references. I can help on B versions if needed, but he and Kirby are the final authorities here. My involvement is I did some content design work on the tables and edited/re-wrote the copy.

PR #6080, opened 2018-08-27
[Fix bug 1479085] Add Firefox vs Safari download page

craigcook:
## Description
Adds two variants of a "Firefox vs Safari" comparison download page, with a Traffic Cop redirect (using the regular page as a control).

These are English only.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1479085
Issues #5950, #6003, #6004

## Testing
https://bedrock-demo-craigcook.oregon-b.moz.works/en-US/firefox/new/?xv=safari&v=1
https://bedrock-demo-craigcook.oregon-b.moz.works/en-US/firefox/new/?xv=safari&v=2


bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -10,6 +10,9 @@
   {% if switch('firefox_new_better_browser_experiment', ['en-US']) %}
     {{ js_bundle('firefox_new_better_browser_experiment') }}
   {% endif %}
+  {% if switch('firefox_new_compare_safari_experiment', ['en-US']) %}
+    {{ js_bundle('firefox_new_compare_safari_experiment') }}
```
craigcook: I have concerns about these experiments colliding. In theory, even with multiple experiments running, they should each only take effect when the URL matches the experiment, but I suspect we might need to wrap add some other condition so it doesn't even load unless the URL is one we're testing... `if xv == 'safari'` or something?

stephaniehobson: Maybe the better browser experiment has been running long enough?

craigcook: Possibly. Though we also know we'll need to have Safari, Edge, and Chrome pages running simultaneously, each with A and B variants, so we'll be in the same boat.

bedrock/firefox/templates/firefox/new/compare/base-1.html
```diff
@@ -0,0 +1,75 @@
...
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/base-protocol.html" %}
+
+{# "noindex" pages should not have the canonical or hreflang tags: bug 1442331 #}
+{% block canonical_urls %}<meta name="robots" content="noindex,follow">{% endblock %}
+
+{% block page_title_prefix %}{% endblock %}
+{% block page_title %}{{_('Free Web Browser for Android, iOS, Desktop | Firefox')}}{% endblock %}
```
stephaniehobson: Maybe we should update this to match the expected OS of the users? "Free Web Browsers for MacOS and iOS."

bedrock/firefox/templates/firefox/new/compare/base-1.html
```diff
@@ -0,0 +1,75 @@
...
+{% block page_desc %}{{_('Millions of people around the world trust Firefox Web browsers on Android, iOS and desktop computers. Fast. Private. Download now!')}}{% endblock %}
```
stephaniehobson: This can also be customized.

media/css/firefox/new/compare.scss
```diff
@@ -0,0 +1,487 @@
...
+
+$font-path: '/media/fonts';
+$image-path: '/media/protocol/img';
+
+@import "../../../../node_modules/@mozilla-protocol/core/protocol/css/includes/lib";
+@import "../../../../node_modules/@mozilla-protocol/core/protocol/css/components/button";
+@import "../../../../node_modules/@mozilla-protocol/core/protocol/css/components/download-button";
+@import "../../../../node_modules/@mozilla-protocol/core/protocol/css/components/call-out";
+@import "../../../../node_modules/@mozilla-protocol/core/protocol/css/components/hero";
+@import "../../../../node_modules/@mozilla-protocol/core/protocol/css/components/feature-card";
```
stephaniehobson: I don't think you're using feature card anymore.

media/css/firefox/new/compare.scss
```diff
@@ -0,0 +1,487 @@
...
+            width: 80px;
+
+            span {
+                display: block;
+            }
+        }
+    }
+}
+
+@media (max-width:479) {
```
stephaniehobson: Missing units, I don't think this MQ is being fired.

stephaniehobson: And that reminds me, we should have the breakpoints as variables in Protocol so we can write custom MQs if we need to. (It doesn't make sense to define the grid at the small mobile sizes and then over-ride everything back to the default behaviour at large ones. Though we can do that if we decide to take a hard line on acceptable MQs.)

craigcook: You wrote this 😉 Presumably a dev artifact, but I failed to notice it. Since it isn't taking effect anyway we can just remove it.

media/css/firefox/new/compare.scss
craigcook: It's used on v1

bedrock/firefox/views.py
```diff
@@ -571,6 +571,8 @@ def download_thanks(request):
             template = 'firefox/new/portland/scene2-safe.html'
         elif experience == 'betterbrowser':
             template = 'firefox/new/better-browser/scene2.html'
+        elif experience == 'compare':
+            template = 'firefox/new/compare/scene2.html'
```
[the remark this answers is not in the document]
craigcook: Oops, this is a leftover. Balete! 🥊

bedrock/firefox/templates/firefox/new/compare/base-1.html
```diff
@@ -0,0 +1,75 @@
...
+      <div class="mzp-l-content">
+        <div class="mzp-c-hero-body">
+          <h1 class="mzp-c-hero-title">{% block hero_title %}{% endblock %}</h1>
+
+          <div class="mzp-c-hero-desc">
+            <p>{% block hero_desc %}{% endblock %}</p>
+          </div>
+
+          <div class="mzp-c-hero-cta mzp-c-button-download">
+            {{ download_firefox(alt_copy=_('Download Firefox'), download_location='primary cta', button_color='mzp-c-button mzp-t-download') }}
```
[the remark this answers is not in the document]
craigcook: I'm not sure we have `data-link-name` in the download button. There's `data-display-name` that indicates the platform and I do see that, but I'm not finding link-name on any other instances.

media/css/firefox/new/compare.scss
stephaniehobson: I don't know when it broke but it was tested before I handed it off (I probably was the one who broke it). It improves the mobile experience though (I think) so should stay.

craigcook: We do have breakpoints as tokens, it's just poorly documented and hard to reference: https://github.com/mozilla/protocol-tokens/blob/master/tokens/_aliases.yml#L135 A token reference is on the to-do list.

So this can be rewritten as `@media (max-width: $screen-sm)`

media/css/firefox/new/compare.scss
stephaniehobson: Ah, yeah.

bedrock/firefox/templates/firefox/new/compare/base-1.html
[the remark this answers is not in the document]
stephaniehobson: 🤷‍♀️ I do not understand our analytics.

media/static-bundles.json
```diff
@@ -1155,6 +1161,20 @@
       ],
       "name": "about-transparency"
     },
+    {
+      "files": [
+        "js/base/mozilla-traffic-cop.js",
+        "js/firefox/new/better-browser-experiment.js"
+      ],
+      "name": "firefox_new_better_browser_experiment"
```
stephaniehobson: Looks like this snuck back in in the rebase.

craigcook: D'oh. There were merge conflicts, I think I resolved this wrong. Fixing.

stephaniehobson:
R+ 👒

Issue #5635, opened 2018-05-03
(8/15) Yandex Page


bensternthal:
Success Criteria
- [ ] Use https://yandex.ru/firefox/mozilla as link to the Yandex build download page
- [x] Show the conditional content to users identified in Russia, only, based on geolocation
- [ ] Launch w. a/b test in place as described in [manapage](https://mana.mozilla.org/wiki/pages/viewpage.action?pageId=80413405)

sandboxjake70:
testing comments

djst:
What is blocking this project?

bensternthal:
On hold due to things still being worked out between Moz & Yandex. I will forward you an email chain with more information, out of our hands for now.

alexgibson:
I just need a final copy deck and sketch files and then we're good to start dev work here.

mrkwvr:
Final design sketch file
[Yandex Landing Page-07102018.sketch.zip](https://github.com/mozilla/bedrock/files/2181502/Yandex.Landing.Page-07102018.sketch.zip)

alexgibson:
Link to final copy doc (in English): https://docs.google.com/document/d/1VsZsKD8wEzLJcaicg0I2uMYS0CZWr3V0ogACQ8CpY8E/edit#

alexgibson:
I'm going to add a blocker to this task until we have the final Russian translation.

alexgibson:
Other things we still need:

- The destination URL which we should point the dl button to (https://yandex.ru/firefox/mozilla is 404)
- The Yandex privacy policy link.

ejregithub:
@alexgibson - just the FYI that https://yandex.ru/firefox/mozilla is now in place.

alexgibson:
Localized copy: https://docs.google.com/document/d/1vL5BkSAuyIaIiPcNMyMznbCE_4Mha_nHIxKrbTJQzTE/edit

alexgibson:
We’re still waiting on one last string translation, but here’s the updated demo with the Yandex copy in Russian https://bedrock-demo-agibson.oregon-b.moz.works/ru/firefox/new/?geo=ru (edited)

alexgibson:
All strings are now in place, but we still need a GA & a fully formed test plan. Going to mark this as blocked.

alexgibson:
With this new `/ru/firefox/new/` page, I'm going to simplify a couple of things given the new Yandex geolocation requirements.

[...]

PR #6086, opened 2018-08-28
Enable Yandex functional tests (#5635)

alexgibson:
Yandex is now enabled in production, so removing the skip for functional tests.


loothoader:
It's just too russian....

Issue #6072, opened 2018-08-23
Make winner of a/b test on /download/thanks page the default, localize to target langs w. newsletter emails


ejregithub:
From the test /download/thanks page mail CTA content - a/b test #5943

Success Criteria:

- [x] Version A is made the default content and shown to: en. es, pt-br, fr, ru, ID, DE, PL
- [ ] Content is localized to en, es, pt-br, fr, ru, ID, DE, PL corresponding to newsletters available in those langs

stephaniehobson:
I am concerned that the new page does not include the "Try downloading again" prompt that the old one did.

ejregithub:
@stephaniehobson - can you please confirm if I'm seeing the right thing here?  I just went to https://www.mozilla.org/en-US/firefox/download/thanks/?v=a  (and the b version_ and was met with this banner in each:

<img width="1072" alt="screen shot 2018-08-23 at 5 05 53 pm" src="https://user-images.githubusercontent.com/20547886/44558083-0b35ac80-a6f7-11e8-8599-8622e2786cde.png">

craigcook:
There are a few additional updates pending to the page, namely with the after-subscribe thank you message. The desire is to hide the title and tagline and add a Firefox mobile promo below the thanks message. That was the original design but a didn't get implemented due to my oversight. It's not a major change but will take a little JavaScripting to listen for the ajax response.

stephaniehobson:
Okay, I see it now. I just had a bit of banner blindness.

alexgibson:
My 2 pence on this request (after some brief discussion with Amy yesterday):

- If we only localize this new page for the locales requested in the description, we'll have to wait for those translations being completed before we can merge this PR, which would likely delay this work rather than make it happen quicker. We'll also still be in a situation where every other locale is being served the old template, so we're automatically making technical debt for ourselves.
- If instead we chose to localize this page fully, we could merge straight away, because we could use an l10n conditional to serve up the old template e.g.

```python
if lang_file_is_active('firefox/download/thanks', locale):
  template_name = 'firefox/new/scene2-email.html'
else:
  template_name = 'firefox/new/scene2.html'
```

If we did this, locales could opt into the new page independently of any further code changes. We'd need to show some other content for locales that don't get the newsletter however. Some mobile download buttons may be a good choice here.

ejregithub:
@amychurchwell - I confirmed with Lauren that we can show the CTA for the es-ES newsletter to all es locales.

PR #6092, opened 2018-08-29
Fix: #6072 Make 'A' variant of firefox/download/thanks default.

amychurchwell:
## Description
_From the test /download/thanks page mail CTA content - a/b test #5943_

🚧 Make 'A' variant of firefox/download/thanks default.
`elif lang_file_is_active('firefox/new/scene2_email.html', locale)`

## Issue / Bugzilla link
#6072


amychurchwell:
Translation only required for languages the newsletter comes in: id, de, en, es, fr, pl, pt, ru, and zh-TW
No lang files should be provided outside of those.

bedrock/firefox/views.py
```diff
@@ -570,6 +570,8 @@ def download_thanks(request):
         template = 'firefox/new/wait-face/scene2.html'
     elif lang_file_is_active('firefox/new/reggiewatts', locale) and experience == 'reggiewatts':
         template = 'firefox/new/reggie-watts/scene2.html'
+    elif lang_file_is_active('firefox/new/scene2_email', locale):
```
alexgibson: It looks like we'll need to move this further down this crazy piece of spagetti, becuse both `de` and `en-US` are in the list of locales. It should be the final `else` on line 601:

```python
elif lang_file_is_active('firefox/new/scene2_email', locale):
    template = 'firefox/new/scene2_email.html'
else:
    template = 'firefox/new/scene2.html'
```
You'll then need to update some failing tests because the template name has changed. You can see the failing tests locally by running `docker-compose run test py.test bedrock/firefox`.

bedrock/firefox/views.py
```diff
@@ -597,7 +596,9 @@ def download_thanks(request):
         elif experience == 'betterbrowser':
             template = 'firefox/new/better-browser/scene2.html'
         else:
-            template = 'firefox/new/email/scene2.html'
+            template = 'firefox/new/scene2_email.html'
+    elif lang_file_is_active('firefox/new/scene2_email', locale) and show_newsletter:
```
alexgibson: Instead of using `show_newsletter` as a variable here, try passing it as a context variable to the jinja template:

```
return l10n_utils.render(request, template, {'show_newsletter': show_newsletter})
```

You can then use `show_newsletter` in the HTML itself, to conditionally display the newsletter.

bedrock/firefox/templates/firefox/new/scene2_email.html
```diff
@@ -2,16 +2,12 @@
...
 
+{% from "macros.html" import google_play_button with context %}
+
 {% add_lang_files "firefox/new/quantum" %}
 
 {% extends "firefox/base-protocol.html" %}
 
-{% block experiments %}
-  {% if switch('firefox_download_scene2_email_experiment', ['en-US']) %}
-    {{ js_bundle('firefox_download_scene2_email_experiment') }}
```
alexgibson: Good spot - I think you should be able to remove this bundle definition and the experiment files too 👍

bedrock/firefox/views.py
```diff
@@ -586,7 +585,7 @@ def download_thanks(request):
         elif experience == 'auf-deiner-seite':
             template = 'firefox/new/berlin/scene2-auf-deiner-seite.html'
         else:
-            template = 'firefox/new/scene2_email.html'
+            template = 'firefox/new/scene2.html'
```
alexgibson: Should this be `'firefox/new/scene2_email.html'` still?

bedrock/firefox/tests/test_views.py
```diff
@@ -729,7 +729,7 @@ def test_berlin_nonde_scene_2(self, render_mock):
         req = RequestFactory().get('/firefox/download/thanks/?xv=berlin')
         req.locale = 'en-US'
         views.download_thanks(req)
-        render_mock.assert_called_once_with(req, 'firefox/new/email/scene2.html')
+        render_mock.assert_called_once_with(req, 'firefox/new/scene2_email.html')
```
alexgibson: I think you can remove this test case, since it's related to variation b?

bedrock/firefox/tests/test_views.py
```diff
@@ -717,20 +711,14 @@ def test_berlin_scene_2(self, render_mock):
...
-        render_mock.assert_called_once_with(req, 'firefox/new/berlin/scene2.html')
+        render_mock.assert_called_once_with(req, 'firefox/new/berlin/scene2.html', ANY)
 
     def test_berlin_nonde_scene_1(self, render_mock):
         req = RequestFactory().get('/firefox/new/?xv=berlin')
         req.locale = 'en-US'
         views.new(req)
         render_mock.assert_called_once_with(req, 'firefox/new/scene1.html', ANY)
 
-    def test_berlin_nonde_scene_2(self, render_mock):
```
alexgibson: I think you need to keep this test, but update the assert to expect `firefox/new/scene2_email.html`

bedrock/firefox/views.py
```diff
@@ -560,6 +560,7 @@ def download_thanks(request):
     experience = request.GET.get('xv', None)
     locale = l10n_utils.get_locale(request)
     variant = request.GET.get('v', None)
+    show_newsletter = locale in ['en-US', 'en-GB', 'en-CA', 'en-ZA', 'es-ES', 'pt-BR', 'fr', 'ru', 'id', 'de', 'pl']
```
alexgibson: @ejregithub can you please double-check the list of locales here that should get shown the newsletter? We're including all English locales (assuming that's wanted). We could also show all the `es` locales too if wanted.

amychurchwell:
@alexgibson Fixed! 👍

bedrock/firefox/views.py
```diff
@@ -584,11 +585,9 @@ def download_thanks(request):
         elif experience == 'auf-deiner-seite':
             template = 'firefox/new/berlin/scene2-auf-deiner-seite.html'
         else:
-            template = 'firefox/new/scene2.html'
+            template = 'firefox/new/scene2_email.html'
```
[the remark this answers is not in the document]
craigcook: Hm though if we change that condition it could break a bunch of the "non-English" tests that use 'de' as the test locale. I don't know if those tests will return true for lang_file_is_active... maybe they do? It's true in dev mode but I think they might fail when testing prod...

Maybe we should change all those tests to use a locale less complicated than German, at least as long as we have all these special German pages.

peiying2:
@alexgibson is the new template name added to the beginning of the file? I only see firefox/new/quantum. 

@craigcook I am not following some of the conversations here, :(. I know there were times some German pages were coded without going through l10n team. Not sure if it is relevant to the discussion here. 

@flodolo pls comment. Would like to process this PR first thing Thursday morning.

flodolo:
I'm having a hard time following the reasoning too.

We're going to expose the strings from `firefox/new/scene2_email.html` in the existing lang file`firefox/new/quantum`, so they should be behind a tag, and you can't check for activation on a file.

alexgibson:
@peiying2 @flodolo `/firefox/download/thanks/` is now an entirely separate file / template to `/firefox/new/`, so I'm not sure we can put the new strings here behind a tag. The PR is checking for ` if lang_file_is_active('firefox/new/scene2_email', locale):` in the view.

bedrock/firefox/views.py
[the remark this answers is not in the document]
alexgibson: @craigcook oh great catch! A second pair of eyes here is always good 👀 👍 

We can mock the lang file active status in the tests should we need, but it should always be true for Dev.

bedrock/firefox/templates/firefox/new/scene2_email.html
```diff
@@ -2,16 +2,12 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
+{% from "macros.html" import google_play_button with context %}
+
 {% add_lang_files "firefox/new/quantum" %}
```
flodolo: Then this line should be removed.

alexgibson:
> I'm not exactly sure how the scene2 could be a completely separate template.

Do you think this may cause issues?

craigcook:
So... I just had one of those "trying to fall asleep but mind keeps running" brainwaves that might simplify a lot of things.

Since we're using the old strings as the non-newsletter fallback content in scene2_email.html, why do we even need two separate templates? Can we just add the newsletter to scene2.html with an l10n tag?

```
{% if show_newsletter and l10n_has_tag('download_thanks_newsletter_092018') %}
  // show newsletter CTA with new strings
{% else %}
  // show old strings with app store buttons
{% endif %}
```

If I'm right, this would simplify the lang files, the view, and the tests.

flodolo:
> Do you think this may cause issues?

Locales could end up with one file activated, and the other not activated. It's OK it the main new/quantum.lang is the one activated, not so sure if it's the new file.

alexgibson:
> Locales could end up with one file activated, and the other not activated. It's OK it the main new/quantum.lang is the one activated, not so sure if it's the new file.

Hmm, this is a good point. Perhaps keeping everything in one lang file would be a better solution :/

alexgibson:
@amychurchwell looks like we may have to rethink how we handle showing this new content in terms of l10n. I think Craig's suggestion may work, but we'd need to re-jig the existing `scene2_email.html` template. We should discuss this later when you're around :)

amychurchwell:
@alexgibson @craigcook Should be good to go! 👍

alexgibson:
@amychurchwell the changes here look good to me! I think the last thing you need to do here is rebase this branch against master. There have been a few other download templates that have been added since this branch was created, so tests will fail if we merge now to master.

@peiying2 @flodolo strings should now be simplified here, behind the l10n tag `firefox_new_try_again_link_020818`. Can you please take a look and let us know if you forsee any further issues. Thanks!

Issue #5844, opened 2018-06-25
Deprecate the old send-to-device widget


craigcook:
For the whatsnew page for Firefox 61 we introduced a refactored widget so we can send messages for either Firefox or Firefox Focus, and additional products in the future. The new widget avoids duplicate IDs so we can safely embed more than one widget on a page. It also follows the Protocol naming convention and is a step towards making it a proper Protocol organism (finalizing that will be another task for another repo).

The old widget is still in place on the whatsnew 60 page as well as the /firefox/mobile/ page, and possibly a few other places but those two may be it. We should update those instances to use the new widget macro, port the existing tests to new widget, update docs where necessary, and do any additional cleanup needed to get us to a more maintainable one-widget state.

alexgibson:
We combined the features from send-yourself into send-to-device, so now there's only one library in the repo.

PR #6100, opened 2018-08-30
Add spin.min.js to whatsnew_61, fix JS error in sendto widget

craigcook:
## Description
While testing #6090 @stephaniehobson noted this error thrown by the sendto widget:
`ReferenceError: Spinner is not defined`

This seems also make the widget submission fail (at least when using 'success@example.com' to test), which seems pretty serious.

The sendto widget was forked from the original send-to-device widget and renders a little spinner while the Ajax is working. However, the library that renders that spinner is included in the common.js bundle but is *not* included in the common-protocol.js bundle, so Protocol-based pages using the new widget are broken.

We're making an effort not to rely on third-party libraries with Protocol so I'd rather not just add the spinner to the common bundle and forget about it, so instead the quicker fix is to add spin.min.js only on the pages that need it (only whatsnew61 right now, as well as whatsnew62 but that's in the other PR). The real long-term fix for this issue is to refactor the sendto widget not to rely on that spinner, and that can be part of deprecating the old widget as well (see #5844). This is just a quick hack to hastily fix a bug in production.

## Testing
Load firefox/61.0/whatsnew while signed into sync (so you get the Fx mobile promo with widget), verify no JS errors and that the widget successfully submits (using 'success@example.com' should be sufficient).

### DIFF
--- a/media/static-bundles.json
+++ b/media/static-bundles.json
@@ -951,6 +951,7 @@
     {
       "files": [
         "js/base/mozilla-fxa-iframe.js",
+        "js/libs/spin.min.js",
         "js/base/send-yourself.js",
         "js/firefox/whatsnew/whatsnew-61.js",
         "js/firefox/whatsnew/whatsnew-61-init.js"


alexgibson:
Hmm, I can't reproduce this error for `/en-US/firefox/61.0/whatsnew/`. The template in question seems to be extending `firefox/whatsnew/base-pebbles.html` rather than Protocol. I think the error may just be in your branch for 62?

craigcook:
Hrm. Well this is interesting. If I fake it on the 61 page in production by using devtools to make the widget visible, I can get the broken response (it shows https://www.mozilla.org/en-US/firefox/send-to-device-post/) which I assumed was due to the missing spinner because it's the same thing that happened in 62 when spinner was missing... but if I'm actually signed in and get the widget legitimately it works just fine. 

So now when I'm in less of a panic and actually think it through, Whatsnew61 is indeed pebbles-based and loads common.js complete with spinner. I was making hasty assumptions and didn't properly investigate causes. Sorry!

PR #6101, opened 2018-08-30
Sitemap: exclude /firefox/*/all/, fix missing URLs, and add grants urls

### LINKED ISSUE #6048: Missing urls in sitemap
I was now able to test and crawl the sitemaps live and found some pages missing in the sitemap. Weirdly all locale /about/ pages aren't in the sitemap but the en-US version is. The rest are the pages always showing up when something is missing like old releasenotes and system requirements. I did a list of pages not in the sitemap but indexable and 200: https://docs.google.com/spreadsheets/d/15iRo0ospuuQqSgI6YUtBKdGvqfMUHj8Awzv8PE7ssX8/edit?usp=sharing

And /firefox/all/ is in the sitemaps for all locales, but isn't indexable, so we should remove it.

There are some old pages ending with .html, these pages do not need to be included, @dzingeek and @thejeffreytaylor we need to figure out what to do with them.

pmac:
Also add system-requirements URLs to sitemap.

Fixes #6048

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,7 +48,7 @@ tests/functional/driver.log
 tests/functional/assets
 .eslintcache
 .docker-build*
-root_files/sitemap_*.xml
+root_files/sitemap*.xml
 root_files/*/sitemap.xml
 root_files/sitemap.json
 node_modules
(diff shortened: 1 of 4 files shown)


pmac:
I've gotten the 👍 from @rraue. This is ready for r?

PR #6105, opened 2018-08-30
Update /firefox/mobile/ page with new strings

stephaniehobson:
## Description
Update /firefox/mobile/ page with new strings.

## Issue / Bugzilla link
bedrock/mozilla#6083


bedrock/firefox/templates/firefox/mobile.html
```diff
@@ -302,13 +302,24 @@ <h3>{{ _('Speed') }}</h3>
               </section>
 
               <section class="page-article-section feature-content" id="focus-lightweight" data-section="focus" data-ga-label="firefox focus lightweight">
-                <h3>{{ _('Lightweight') }}</h3>
```
alexgibson: When I scroll down the page, the side navigation still lists "Lightweight" as the feature heading, even though the copy here has been updated to centre on privacy. The accompanying image also mentions "Small, yet powerful at less than 5MB.", which now seems out of place.

screenshot_2018-09-04 mobile browsers for ios and android firefox.

I also don't really see the difference between this new section, and the first "Automatic privacy" section? They both seem to talk about roughly the same thing? :/

screenshot_2018-09-04 mobile browsers for ios and android firefox 1

ejregithub:
I've asked for Loren (the requester of this work) to review the comments and respond.

bedrock/firefox/templates/firefox/mobile.html
lwa-moz: Hi Alex - the two are similar, but highlighting different things. My take is that consumers will be interested in both features and brand positioning, which often feel like the same thing from an internal perspective.

"Automatic privacy" refers to features, calling out the automatic blocking of ads and ease of erasing your browser history. These features are available in other products, so to bolster this claim I'm suggesting we reiterate our brand positioning with the third section.

"On your side" refers to the mission that drives the creation of these features, and provides a deeper level of trust concerning the brand behind the product. Akin to talking about improved airbags in a new Toyota model, and then talking about how Toyota is a trusted brand when it comes to safety - similar ideas that enhance one another.

The side navigation should mirror the new subject heading.

The caption should be removed - you're correct that we are moving away from the "small yet powerful" language due to increased Focus Android filesize.

stephaniehobson:
Updated.

media/css/firefox/mobile.scss
```diff
@@ -663,8 +663,8 @@ html[dir="rtl"] {
             @include transform(translateY(100%));
         }
 
-        &[data-current="focus-lightweight"] ul:before {
-            @include transform(translateY(200%));
+        &[data-current="focus-oys"] ul:before {
+            @include transform(translateY(100%));
```
alexgibson: Looks like this should be `200%`

stephaniehobson:
Updated.

Issue #5946, opened 2018-07-25
(Due 8/28) WNP 62 - Sept 5th


ejregithub:
[Master project doc](https://docs.google.com/document/d/1wEpa-pEyQOIKwXSRligj6cUPbXQ-qN8wmP4LIe9q7ks/edit#heading=h.44tk8jl1sj1b)

Success Criteria

- [x] EN (all en locales) to be shown content for Pocket
- [x] Page is live on production August 29th to accommodate external QA

thejeffreytaylor:
Copy deck here: https://docs.google.com/document/d/1uf8y5PqUHWghuI_wUmdmQVNEU28XvO9uOhuQmvPtqc0/edit?usp=sharing

michaelmccombie:
Here is a link to the design files
https://share.goabstract.com/48f0d841-d03a-4546-a164-e90c1ea58e5d

dzingeek:
I will package up the assets up today.

thejeffreytaylor:
Copy decks for both versions will be up to spec by EOD today.

mrkwvr:
Design file and assets available here https://mozilla.box.com/s/0v0sgnllpiq9662kgtyf1mudm8g0vjgz

dzingeek:
Designs have been updated. We are going out with version A and if there is time we will test Version B.
https://share.goabstract.com/aef26895-1f7b-485f-a6db-f18efb2c6a44

dzingeek:
**Note:** The design and copy was approved by Janis & Devyani in the slack channel 08.21.18 around  10:35am

thejeffreytaylor:
Copy Decks: 

[Version A (to be deployed) ](https://docs.google.com/document/d/1y8mjxwaZ5tCuwoBP3KKpODCsEKOg0DAg1UsWzZFu3dY/edit)

[Version B](https://docs.google.com/document/d/1uf8y5PqUHWghuI_wUmdmQVNEU28XvO9uOhuQmvPtqc0/edit?usp=sharing) (for testing, if bandwidth allows)

dzingeek:
**One small tweak to the design:**
Adding Fx logo in header area since the copy no longer references Fx. Sketch file is up to date. 

See png for reference. 
![screen shot 2018-08-23 at 9 40 22 am](https://user-images.githubusercontent.com/3740746/44539518-3699a680-a6b9-11e8-941e-e5cc6af1d736.png)

dzingeek:
[...]

PR #6108, opened 2018-08-31
[#5946] CSS fixes for Whatsnew62

craigcook:
## Description
- [x] Fix header logo layout when up-to-date message is hidden.
- [ ] Fix visited link color in sendto widget.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/5946#issuecomment-417498724

## Testing
You'll need to view the page with an outdated Firefox so the "up to date" message is hidden; verify the header logos look acceptable (Firefox on the left, Mozilla/Pocket on the right in wide viewports). Short of using an actual old Firefox you can also fake it by removing the "show-up-to-date-messaging" class in devtools, or locally commenting out the `checkUpToDate` function in whatsnew-62.js.

To see the sendto widget you'll need to either be signed into a Firefox account or fake it by firing the `showFirefoxMobile()` event in whatsnew-62.js. There's a "learn more" link in the privacy statement below the form field. It should be blue when unvisited, purple when visited.

### DIFF
--- a/media/css/firefox/whatsnew/whatsnew-62.scss
+++ b/media/css/firefox/whatsnew/whatsnew-62.scss
@@ -116,20 +116,14 @@ $image-path: '/media/protocol/img';
     }
 
     @media #{$mq-md} {
-        &.show-up-to-date-message .page-header-inner {
-            align-items: center;
-            display: flex;
-            justify-content: space-between;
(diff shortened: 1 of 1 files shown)


stephaniehobson:
R+

PR #6118, opened 2018-09-04
Move transparancy report off sandstone

stephaniehobson:


stephaniehobson:
Buttons should stay equal heights now. Tightened up mobile spacing a bit too.

PR #6133, opened 2018-09-06
remove the download col for esr52

sylvestre:


pmac:
Looks like we should turn off the `esr` switch first?

pmac:
We can actually remove the entire surrounding if statement for the ESR switch. Both code-blocks are the same after this change.

sylvestre:
ok, I will update my PR
(I was asking what was the switch for, now, I know :p)

pmac:
Actually I think we need to do something else. Now that the switch has been flipped the page title doesn't make much sense as there is no choice to be made. I'll file an issue to update the page and remove the switch unless you really want to update this PR. Thanks for the help!

stephaniehobson:
@amychurchwell Do you want to take this review? We can address the page title separately but let's get this merged since @sylvestre was nice enough to save us the work.

stephaniehobson:
Yeah, we should remove the switch over there too.

PR #6144, opened 2018-09-07
Remove frankfurt from master deploys

jgmize:
Due to continuing issues with 504s: https://github.com/mozmeao/infra/issues/908

### DIFF
--- a/jenkins/branches/master.yml
+++ b/jenkins/branches/master.yml
@@ -5,7 +5,6 @@ smoke_tests: true
 regions:
   - oregon-b
   - tokyo
-  - frankfurt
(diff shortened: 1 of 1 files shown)


pmac:
As long as frankfurt has been removed from the dns for www-dev.allizom.org this seems good to merge 👍

metadave:
```
www-dev.allizom.org.    3600    IN    CNAME    allizom.ramzom.org.
```
points to `abb45ebd7fbb211e788530656fabf84c` ELB in us-west-2 (oregon-b)

I don't see any Frankfurt here. +1 @jgmize ?

Issue #6055, opened 2018-08-20
[August] Homepage content updates


ejregithub:
Success Criteria:

- [ ] homepage updated with [new content](https://docs.google.com/spreadsheets/d/14e4UMW1FIntdB-DU6WFq_tsUO21d2MJF0jsS2hP_Gok/edit#gid=0) 

* handy mock up in [Airtable](https://airtable.com/shrVQdWrXYQus2NqT). (Be sure to click on the Next Record link at the top of the page to get to the 8/10/2018 edition)

ishanrai05:
Hi, I would like to help with this issue.

alexgibson:
@ejregithub please note here there's a bit more work than just content updates. The home page editorial team is moving to using YouTube for all their videos, as exporting the different formats was taking too long in terms of resourcing. That means for this update, we'll need to write some new JS that uses the [YouTube Player API](https://developers.google.com/youtube/iframe_api_reference), so we can lazy-load the videos when people click to open the modal.

I can discuss this with @amychurchwell if/when she picks this up, if that helps.

amychurchwell:
@ejregithub we tend to avoid animations unless they are demonstrating product features. I'll create static images so we can get the changes out. If there is a need for the images to be animated on the homepage the team can discuss it further. 🙂

PR #6147, opened 2018-09-07
Fix: #6055 August Homepage Updates

amychurchwell:
## Description
Cards updated with new content. Implementation of YouTube API to play videos in modals.

## Issue / Bugzilla link
#6055


amychurchwell:
@alexgibson the YouTube API is loading videos inside of the card modal, but it currently doesn't show a video 'preview' on the card itself (just the fallback image.)
screen shot 2018-09-07 at 1 48 12 pm

bedrock/mozorg/templates/mozorg/home/home-en.html
```diff
@@ -71,10 +71,7 @@ <h2 class="mzp-c-card-title"><span>{{ title }}</span></h2>
   {% if video_url %}
   <div class="mzp-c-card-video-wrapper hidden">
     <figure class="mzp-c-card-video-content">
-      <video class="mzp-c-card-video" preload="none" controls playsinline data-ga-label="{{ ga_title }}">
-        <source src="{{ video_url }}.webm" type="video/webm">
-        <source src="{{ video_url }}.mp4" type="video/mp4">
-      </video>
+      <div class="video-play" data-id="{{ video_id }}"/>
```
alexgibson: Nit: a `<div>` is not self closing

bedrock/mozorg/templates/mozorg/home/home-en.html
```diff
@@ -49,7 +49,7 @@
   {% endif %}
 {% endblock %}
 
-{% macro card_block(class, tag_label, media_icon, title, ga_title, desc, image_url, include_highres_image, video_url, aspect_ratio, link_url, meta) -%}
+{% macro card_block(class, tag_label, media_icon, title, ga_title, desc, image_url, include_highres_image, video_url, video_id, aspect_ratio, link_url, meta) -%}
```
alexgibson: Do we need both `video_url` and `video_id` still? Could we just have one parameter, maybe `youtube_id`?

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,203 @@
...
+ }
+
+ (function(Waypoint) {
+     'use strict';
+
+     var tag = document.createElement('script');
+     tag.src = 'https://www.youtube.com/iframe_api';
+     var firstScriptTag = document.getElementsByTagName('script')[0];
+     firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
+    //
```
alexgibson: Nit: remove blank comment

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,203 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-(function(Waypoint) {
```
alexgibson: ESLint is complaining about the global variables used by the YouTube API, so you may need to add exceptions for them here at the top of this file:

```javascript
/* global YT */
/* eslint no-unused-vars: [2, { "varsIgnorePattern": "onYouTubeIframeAPIReady" }] */
```

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,203 @@
...
-                        // fail silently
-                    }
-                },
-                onDestroy: function() {
-                    video.pause();
-                }
-            });
-        }
-    }
+         for (var j = 0; j < videos.length; j++) {
```
alexgibson: Since we want to track play/pause events using the YouTube API, we can probably remove the old event listeners here.

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,203 @@
...
+     var firstScriptTag = document.getElementsByTagName('script')[0];
+     firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
+    //
+     function onYouTubeIframeAPIReady() {
+
+        // lazy load video when visitor clicks the placeholder.
+         var videoLink = document.querySelector('.video-play');
+         var videoId = videoLink.getAttribute('data-id');
+
+         new YT.Player(videoLink, {
```
alexgibson: Hmm, right now this is trying to initialize on page load, which triggers the video autoplay blocking feature in Nightly:

screen shot 2018-09-10 at 10 00 01

What we want to do is only load the video player when someone clicks a vide card to open the modal, so inside here we likely want to bind the event listener for the card click events, similar to how we do it for the [berlin page](https://github.com/mozilla/bedrock/blob/master/media/js/firefox/new/berlin-scene1.js#L31).

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,203 @@
...
+             case YT.PlayerState.PLAYING:
+                 state = 'video play';
+                 break;
+             case YT.PlayerState.PAUSED:
+                 state = 'video paused';
+                 break;
+             case YT.PlayerState.ENDED:
+                 state = 'video complete';
+                 break;
+             }
```
alexgibson: This is the function we want to use to track the video play/pause/end events. Right now ESLint is complaining, becuase the `state` variable is not being used here.

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,203 @@
...
      */
 
-    function trackVideoInteraction(title, state) {
-        window.dataLayer.push({
-            'event': 'video-interaction',
-            'videoTitle': title,
-            'interaction': state
-        });
-    }
+     function trackVideoInteraction(title, state) {
```
alexgibson: Nit: indentation here and below.

bedrock/mozorg/templates/mozorg/home/home-en.html
```diff
@@ -116,56 +113,56 @@ <h3 class="primary-title-sub">{{ _('Fast for good.') }}</h3>
       <div class="mzp-l-card-hero">
         {{ card_block(
             class='mzp-c-card-large',
-            tag_label=_('Podcast'),
-            media_icon='mzp-has-video',
-            title=_('IRL season 3: The grand bargain'),
-            ga_title='IRL season 3: The grand bargain',
-            desc=_('You click “accept” every day, but often don’t know what you’re giving away. Is it a fair trade, or should you demand a better bargain?'),
-            image_url='home/2018/cards/irl-season-3.png',
```
alexgibson: Nit: We should remove any old image assets no longer used on the page.

alexgibson:
@amychurchwell I'd also suggest getting this up on a demo server so MJ can review the content 👍

amychurchwell:
@alexgibson Pushed what I had even though it's not merge-worthy. Unable to access YouTube's [play](https://developers.google.com/youtube/iframe_api_reference#playVideo) & [pause](https://developers.google.com/youtube/iframe_api_reference#pauseVideo) functions under `onCreate: function()` to have the video auto-play on opening. There is possibly a scoping issue going on? 🤔 

screen shot 2018-09-10 at 4 54 41 pm

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,173 @@
...
+             player = new YT.Player(videoLink, {
+                 height: '703',
+                 width: '1200',
+                 videoId: videoId,
+                 playerVars: {
+                     modestbranding: 1, // hide YouTube logo.
+                     rel: 0, // do not show related videos on end.
+                 },
+                 events: {
+                     'videoInteraction': trackVideoInteraction,
```
alexgibson: `videoInteraction` isn't a supported event for youtube's API. You'll need to handle tracking GA events using the `onStateChange` event handler.

To get the video to auto-play, it looks like you'll also need to add an `onReady` event handler here also:

```javascript
'onReady': onPlayerReady,
```

And then something like:

```javascript
function onPlayerReady(event) {
    event.target.playVideo();
}
```

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,173 @@
...
+             var card = event.currentTarget;
+             var title = card.querySelector('.mzp-c-card-title').innerText;
+             var sibling = getNextEl(event.currentTarget);
+             var content = sibling.querySelector('.mzp-c-card-video-content');
+
+             var videoLink = document.querySelector('.video-play');
+             var videoId = videoLink.getAttribute('data-id');
+             var player;
+
+             player = new YT.Player(videoLink, {
```
alexgibson: Nit: you could just declare the variable on this line also, instead of line 43 above?

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,173 @@
...
-        }
-    }
+         for (var i = 0; i < videoCards.length; i++) {
+             videoCards[i].addEventListener('click', playVideo, false);
+         }
+     }
 
     // Lazyload images
-    Mozilla.LazyLoad.init();
+     Mozilla.LazyLoad.init();
```
alexgibson: Still got some indentation issues here and below. Installing [editorconfig](https://editorconfig.org/) in your editor may help with this.

media/js/mozorg/home/home.js
amychurchwell: The issue I was running into with `onPlayerReady()` was it was being called even _after_ the modal was closed. So I believe the start/stop needs to live in `onCreate:` and `onDestroy:` so it's dependent on the video modal. Wdyt?

alexgibson: Hmm, right. I figured we could only call the YouTube API after the modal was created so we’d still need to wait for the player to be in the ready state before trying to play? As for stopping it when the modal closes, would calling `destroy()` on the video work?

alexgibson: I found an older example of using this API inside a modal which might also help 😀

https://github.com/mozilla/bedrock/blob/master/media/js/firefox/dev-firstrun.js

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,171 @@
...
+                     state = 'video paused';
+                     break;
+                 case YT.PlayerState.ENDED:
+                     state = 'video complete';
+                     break;
+                 }
+
+                 if (state) {
+                     window.dataLayer.push({
+                         'event': 'onReady',
```
alexgibson: `event` should be labelled `video-interaction` instead of `onReady`

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,171 @@
...
-    }
-
-    /*
-     * Video Tracking
-     */
+ /* global YT */
+ /* eslint no-unused-vars: [2, { "varsIgnorePattern": "onYouTubeIframeAPIReady" }] */
+
+// YouTube API hook has to be in global scope
+ function onYouTubeIframeAPIReady() {
```
alexgibson: Nit: this JS file still has some indentation issues here and below.

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,171 @@
...
+             // init the button
+             initStickyCTA();
+         } else {
+             $stickyCTA.remove();
+         }
+     } else {
+         $stickyCTA.remove();
+     }
+
+ })(window.Waypoint);
```
alexgibson: It looks like the sticky CTA on this page still uses jQuery :(

In the console I see the following error on page load:

```
Error: No element option passed to Waypoint constructor
```

I think this is because you're missing `window.jQuery` here as the first argument passed to the IIFE.

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,171 @@
...
+             var title = card.querySelector('.mzp-c-card-title').innerText;
+             var sibling = getNextEl(event.currentTarget);
+             var content = sibling.querySelector('.mzp-c-card-video-content');
+
+             var videoLink = document.querySelector('.video-play');
+             var videoId = videoLink.getAttribute('data-id');
+
+             var player = new YT.Player(videoLink, {
+                 height: '390',
+                 width: '640',
```
alexgibson: When the video loads it's a set width, whereas it should probably span the full width of the modal container:

screenshot_2018-09-12 internet for people not profit

On small viewports it also runs out of the container:

screenshot_2018-09-12-internet-for-people -not-profit

media/js/mozorg/home/home.js
```diff
@@ -2,147 +2,171 @@
...
+     function initVideoEvents() {
+         var videoCards = document.querySelectorAll('.mzp-c-card.has-video-embed .mzp-c-card-block-link');
+
+         function playVideo(event) {
+             var card = event.currentTarget;
+             var title = card.querySelector('.mzp-c-card-title').innerText;
+             var sibling = getNextEl(event.currentTarget);
+             var content = sibling.querySelector('.mzp-c-card-video-content');
+
+             var videoLink = document.querySelector('.video-play');
```
alexgibson: This will always get the first element on the page that matches this selector, which works here but would break if we had multiple videos on the page. Should this be changed to `content.querySelector`?

bedrock/mozorg/templates/mozorg/home/home-en.html
```diff
@@ -116,56 +113,55 @@ <h3 class="primary-title-sub">{{ _('Fast for good.') }}</h3>
...
-            ga_title='Get the newest Firefox',
-            desc=_('The latest Firefox is here. Built to help you work smarter and faster.'),
-            image_url='home/2018/cards/firefox.png',
-            include_highres_image=True,
+            tag_label=_('Podcast'),
+            title=_('IRL: Paid attention'),
+            ga_title='IRL: Paid attention',
+            desc=_('There’s a new currency in town (and no, we’re not talking about Bitcoin). We’re talking about attention. Tune in to IRL.'),
+            image_url='home/2018/cards/irl-s3-e3.png',
             aspect_ratio='mzp-has-aspect-3-2',
```
alexgibson: Should this be `mzp-has-aspect-1-1`? The source image seems square, but visually it works with either aspect ratio. If they want to keep it 3:2, could we crop the image to save some bytes?

bedrock/mozorg/templates/mozorg/home/home-en.html
```diff
@@ -280,70 +270,67 @@ <h2 class="mzp-c-billboard-title">{{ _('Support a healthy internet.') }}</h2>
...
-            title=_('7 ways online rights have changed '),
-            ga_title='7 ways online rights have changed ',
-            desc=_('As 2018 hits the halfway point, here is our list of online issues we feel are most important.'),
-            image_url='home/2018/cards/online-rights.png',
-            include_highres_image=True,
+            tag_label=_('Partners'),
+            title=_('WeTransfer keeps community at heart'),
+            ga_title='WeTransfer keeps community at heart',
+            desc=_('WeTransfer, the independent file transfer and sharing service, makes file-sharing easy in Firefox while keeping user privacy at heart.'),
+            image_url='home/2018/cards/we-transfer.png',
```
alexgibson: Some image such as this one are pretty huge. Can we resize them more apropriately according to the size of to save on page weight? Resizing images can be a tedious task, but it's probably worthwhile still 😩 

The previous home page cards can probably act as general guidelines for sizing (we still need better docs for this in Protocol 😢)

bedrock/mozorg/templates/mozorg/home/home-en.html
```diff
@@ -68,13 +68,10 @@ <h2 class="mzp-c-card-title"><span>{{ title }}</span></h2>
...
   </a>
-  {% if video_url %}
+  {% if youtube_id %}
   <div class="mzp-c-card-video-wrapper hidden">
     <figure class="mzp-c-card-video-content">
-      <video class="mzp-c-card-video" preload="none" controls playsinline data-ga-label="{{ ga_title }}">
-        <source src="{{ video_url }}.webm" type="video/webm">
-        <source src="{{ video_url }}.mp4" type="video/mp4">
-      </video>
+      <div class="video-play" data-id="{{ youtube_id }}"><h2>test</h2></div>
```
alexgibson: Nit: remove `<h2>test</h2>`?

amychurchwell: @alexgibson thanks for the notes, Alex! 😓 Truth be told, I pushed what I had EOD so it is a bit rough in places.

Issue #6054, opened 2018-08-20
[Due Sept 17] Profile per install feature - DOWNGRADE informational page


ejregithub:
[Background doc](https://docs.google.com/document/d/1qW4J_ons1ZlKgY1zA_taaRQ_Nk4BK5Afsynhb7gAqRQ/edit#)

Success Criteria:

- [ ] Page w. content informing users in the Downgrade scenario
- [ ] unique URLs for scenario (with a unique query param, etc)
- [ ] page set as no index
- [ ] Timing: pages in place Sept 17th to be seen in 64 Nightly release cycle
- [ ] Must be localized (top 10 in Nightly) for launch

thejeffreytaylor:
[Copy deck here ](https://docs.google.com/document/d/1s2IbTeRMGOkKqu8kDiJW4_NrROAl9FVMVlBMbIzb5ao/edit)- still needs review.

dzingeek:
@thejeffreytaylor let's sync up on this as I have shifted things around a bit before we send out to other team for review. 

@ejregithub please hold off on getting other teams approval.

dzingeek:
[design file here](https://share.goabstract.com/f1e8d41d-bac9-41e5-b71b-a3346c956509) - let me know if you have any questions

thejeffreytaylor:
Copy is updated. There may be small changes to accommodate design requirements, but the feedback has been implemented.

dzingeek:
@craigcook **Design tweaks** 

**Header area:**
- [ ] make "An Important Change..." copy smaller and dark red. (see screenshot for reference)
- [ ] make "We detected that you..." copy larger and bold (see screenshot for reference)

**FxA area:**
- [ ] make "Enter your email to..." bold
- [ ] move "by proceeding, you agree.." copy below button (this may not be possible)

Note: can we not have the button on the same "line" as the entry filed?

![screen shot 2018-09-11 at 12 40 05 pm](https://user-images.githubusercontent.com/3740746/45383661-bb912500-b5c0-11e8-846a-503c5790bbb7.png)
![screen shot 2018-09-11 at 12 40 11 pm](https://user-images.githubusercontent.com/3740746/45383663-bb912500-b5c0-11e8-898f-b5fc6572e231.png)

craigcook:
https://bedrock-demo-craigcook.oregon-b.moz.works/firefox/profile-downgrade/

djst:
What's the status of this? Deadline is 9/17 according to card.

Issue #5974, opened 2018-08-06
/new experiment to add FxA CTA 


thejeffreytaylor:
[Experiment overview deck](https://docs.google.com/presentation/d/109QM--B0e1bndEov7wpCJWbvHmy26Hh-Wj6KOxObjmY/edit?usp=sharing). 

Success Criteria:
- [x] Uses the FxAccount iframeless form
- [ ] Metrics must track impact of FxA CTA on downloads
- [x] Experiment content is shown only to up-to-date Firefox users
- [x] Fx Accounts team is informed of launch, in advance of it

alexgibson:
This task is somewhat dependent on the status of https://github.com/mozilla/bedrock/pull/5713

thejeffreytaylor:
And is a sibling to #5944 - Once all these experiments are done, we need to make sure that our email collection efforts don't confuse users (Fxa vs. onboarding, etc.), and that the flows work such that users don't feel pestered or confused by the timing of these asks.

dzingeek:
Here is a .png of the user flow. @mrkwvr is working on the ui for the page now.

![new_fxa-experiment_uf](https://user-images.githubusercontent.com/3740746/44061489-6f66ba06-9f0d-11e8-85d5-4f96ab874cf3.png)

ejregithub:
Eric to ask jpetto for status on PR for iframeless

ejregithub:
I pinged Alex Davis about this two weeks ago, he's good with it.

ejregithub:
@stephaniehobson - as a prerequisite to this experiment running, could you please review @jpetto 's PR for the iframeless FxAccount form? https://github.com/mozilla/bedrock/pull/5713

mrkwvr:
Final design file https://mozilla.box.com/s/y3gjvsl7d1ny1a2tm0a3wbr1fib5l1q4

dzingeek:
Looks like we got approval from Lauren to move forward with this direction and copy via slack. 
08/23/18 @ 12:31pm

stephaniehobson:
#5713 is merged now.

The original requirements for the iframeless form were to put it on /firstrun which is only ever seen by Firefox users.

So, even though #5713 is merged it is not ready to be used on /new immediately.

stephaniehobson:
As @thejeffreytaylor pointed out, we now ask the users for email addresses on both the download and the download thanks pages with virtually identical forms.

[...]

Issue #5655, opened 2018-05-04
[Due Sept 17] Profile per install feature - MIGRATION scenario informational page


bensternthal:
[Background doc](https://docs.google.com/document/d/1qW4J_ons1ZlKgY1zA_taaRQ_Nk4BK5Afsynhb7gAqRQ/edit#)

Success Criteria:

- [ ] Page w. content informing users in the primary (new profile(s) created) scenario
- [ ] unique URLs for scenario (with a unique query param, etc)
- [ ] page set as no index
- [ ] Timing: pages in place Sept 17th to be seen in 64 Nightly release cycle
- [ ] Must be localized (top 10 in Nightly) for launch

ejregithub:
@peiying2 - flagging the need for l10n with you.  We'll get final copy asap.  The request from product is to have the content in place, localized to top 10 langs, for Sept 5th

thejeffreytaylor:
[Copy deck here](https://docs.google.com/document/d/1Ze-2tnOFJ6b4pMH2MilPC3LADM3_JWwT195JawXdIAU/edit?usp=sharing)

thejeffreytaylor:
Copy is updated. There may be small changes to accommodate design requirements, but the feedback has been implemented.

dzingeek:
@craigcook Use this copy doc 

**Migration Copy Only**
https://docs.google.com/document/d/1Xbqr6_qmPwXi5zWzss57lIyTpGGlI_iRWtOi4SLF3SQ/edit?usp=sharing

dzingeek:
@craigcook **Design tweaks** 

**Header area:**
- [ ] make "An Important Change..." copy smaller and dark red. (see screenshot for reference)
- [ ] make "We detected that you..." copy larger and bold (see screenshot for reference)

**FxA area:**
- [ ] make "Enter your email to..." bold
- [ ] move "by proceeding, you agree.." copy below button (this may not be possible)

Note: can we not have the button on the same "line" as the entry filed?

![screen shot 2018-09-11 at 12 40 05 pm](https://user-images.githubusercontent.com/3740746/45383661-bb912500-b5c0-11e8-846a-503c5790bbb7.png)
![screen shot 2018-09-11 at 12 40 11 pm](https://user-images.githubusercontent.com/3740746/45383663-bb912500-b5c0-11e8-898f-b5fc6572e231.png)

craigcook:
[...]

PR #6168, opened 2018-09-13
[#6054, #5655] Profile-per-install messaging

craigcook:
## Description
Adds two variant pages explaining the new changes in profile management to users with multiple installs. This change is targeted for Firefox 64 but is due to land in Nightly in a few days so we need to a) secure the URLs since they'll be in-product and b) allow plenty of time for l10n before the release of Fx64.

Note this also includes some changes in the iframeless FxA macro that can have impact on other current FxA work such as #5974 

## Bug/Issue
#6054 
#5655 
https://bugzilla.mozilla.org/show_bug.cgi?id=1474285

## Testing
https://bedrock-demo-craigcook.oregon-b.moz.works/firefox/profile-migrate/
https://bedrock-demo-craigcook.oregon-b.moz.works/firefox/profile-downgrade/

### DIFF
--- a/bedrock/firefox/templates/firefox/profile/profile-base.html
+++ b/bedrock/firefox/templates/firefox/profile/profile-base.html
@@ -68,7 +68,7 @@
     <div class="fxa-cta show-fxa-supported-signed-in show-fxa-supported-signed-out">
       <p class="show-fxa-supported-signed-in signed-in">{{ _('You’re signed in and ready to start using Sync.') }}</p>
       <div class="show-fxa-supported-signed-out signed-out">
-        {{ fxa_email_form(entrypoint='accounts-page', utm_source='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'accounts-page', 'medium': 'referral'}, intro_text=_('Get Your Firefox Account'), button_text=_('Sign Up Now'), button_class='mzp-c-button') }}
+        {{ fxa_email_form(entrypoint='accounts-page', utm_source='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'accounts-page', 'medium': 'referral'},) }}
       </div>
     </div>
   </aside>
--- a/bedrock/firefox/templates/firefox/profile/profile-base.html
+++ b/bedrock/firefox/templates/firefox/profile/profile-base.html
@@ -1,0 +1,82 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% from "macros.html" import fxa_email_form with context %}
+
+{% add_lang_files "firefox/profile-per-install" %}
+
+{% extends "firefox/base-protocol.html" %}
+
+{% block page_title %}{{ _('An Important Change in Firefox: Dedicated Profiles') }}{% endblock %}
+
+{% block page_css %}
+  {{ css_bundle('profile_per_install') }}
+{% endblock %}
+
+{% block body_class %}state-fxa-default{% endblock %}
+
+{% block site_header %}
+<header id="page-header" class="page-header">
+  <div class="mzp-l-content">
+    <a href="{{ url('firefox') }}">
+      {{ high_res_img('logos/firefox/logo-quantum-wordmark-large.png', {'alt': 'Firefox', 'width': '147', 'height': '55', 'class': 'logo-fx'}) }}
+    </a>
+
+    <a href="{{ url('mozorg.home') }}">
+      <img src="{{ static('img/logos/mozilla/wordmark-dark.svg') }}" alt="Mozilla" width="112" height="32" class="logo-moz" id="logo-moz">
+    </a>
+  </div>
+</header>
+{% endblock %}
+
+{% block content %}
+<article class="mzp-c-article mzp-t-firefox mzp-l-content">
+  <h1 class="mzp-c-article-title">{{ self.page_title() }}</h1>
+
+  {% block profile_state %}{% endblock %}
+
+  <h2>{{ _('What’s Changed') }}</h2>
+
+  <p>{{ _('Before, all Firefox installations were sharing a single profile by default. Now, Firefox provides dedicated profiles for each Firefox installation (this includes installations of Firefox Nightly, Firefox Beta, or Firefox ESR). This will make Firefox more stable when switching between Firefox installations on the same computer.') }}</p>
+
+  <h2>{{ _('How Dedicated Profiles Work') }}</h2>
+
+  <p>{{ _('Firefox saves information such as bookmarks, passwords and user preferences in a set of files called your profile. This profile is stored in a separate location from the Firefox program files.') }}</p>
+
+  <ul class="mzp-u-list-styled">
+    <li><p>{{ _('The installation you choose as your primary Firefox installation will keep your previously shared profile data like bookmarks, history and passwords.') }}</p></li>
+    <li><p>{{ _('You have not lost any personal data or customizations. Any previous profile data is safe and attached to the first Firefox installation that was opened after this change.') }}</p></li>
+    <li><p>{{ _('All of your Firefox installations will now have separate profiles.') }}</p></li>
+  </ul>
+
+  <h2>{{ _('What You Need to Do') }}</h2>
+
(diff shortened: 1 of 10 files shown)


craigcook:
This is pending final copy and creative signoff so DNM for now but should be ready for initial code review.

bedrock/firefox/templates/firefox/profile/profile-downgrade.html
```diff
@@ -0,0 +1,11 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% add_lang_files "firefox/profile-per-install" %}
+
+{% extends "firefox/profile/profile-base.html" %}
+
+{% block profile_state %}
+  <p class="mzp-c-article-intro">{{ _('We detected that you have installed and opened an older version of Firefox on your computer. You may need to take action.') }}</p>
```
stephaniehobson: I don't think we should be creating classes with the .mzp prefix in bedrock. If it's not in protocol leave off the `mzp` prefix.

craigcook: This is in Protocol: https://protocol.mozilla.org/patterns/organisms/article.html

craigcook: Though I now realize that doesn't include an article-intro in the example...

bedrock/firefox/templates/firefox/profile/profile-base.html
```diff
@@ -0,0 +1,82 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% from "macros.html" import fxa_email_form with context %}
+
+{% add_lang_files "firefox/profile-per-install" %}
```
craigcook: @peiying2 Strings are split across three different templates but two of them just have one small bit of text, so I think all three should share a single lang file. Will this work?

craigcook:
I backed out the changes to the FxA macro and they'll happen as part of #6173 instead.

bedrock/firefox/templates/firefox/profile/profile-base.html
peiying2: Let me check and get back to you.

media/css/firefox/profile-per-install.scss
```diff
@@ -0,0 +1,80 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+$font-path: '/media/fonts';
+$image-path: '/media/protocol/img';
+
+@import "../../../node_modules/@mozilla-protocol/core/protocol/css/includes/lib";
```
alexgibson: You can shorted these to:

```
@import "../../protocol/css/includes/lib";
@import "../../protocol/css/components/article";
```

media/css/protocol/components/_fxa.scss
```diff
@@ -0,0 +1,27 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+@import "../../../node_modules/@mozilla-protocol/core/protocol/css/includes/lib";
+
+// Hide elements marked for specific variations by default
```
alexgibson: Is there a reason this should be under `/protocol/`? I see there's one in the `/pebbles/` directory already, but these are really just state classes and nothing directly related to either library. Would it make sense for this to live in `/css/base/` where any page can use it?

media/css/firefox/profile-per-install.scss
```diff
@@ -0,0 +1,80 @@
...
+@import "../../../node_modules/@mozilla-protocol/core/protocol/css/components/article";
+@import "../protocol/components/fxa";
+
+.mzp-c-article {
+    margin-bottom: $layout-lg;
+
+    &.mzp-l-content {
```
alexgibson: Whilst this is largely leaning on existing article component styles with a couple of small overrides, it's not something we have clearly defined in Protocol yet. We should probably make sure we fix https://github.com/mozilla/protocol/issues/232 and then follow up here to apply these new styles? It's a small thing, but my concern is the more of these paper cuts we create then if we forget about them, we'll end up with a bigger wound than we anticipate?

media/css/firefox/profile-per-install.scss
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+//* -------------------------------------------------------------------------- */
+// Firefox Accounts
+
+.fxa-cta {
+   .signed-in {
+        @include at2x('/media/img/firefox/accounts/green-check.png', 30px, 30px);
+        background-position: left top;
```
alexgibson: Need a bidi style here:

screen shot 2018-09-20 at 10 26 34

media/js/base/mozilla-fxa-form.js
```diff
@@ -66,9 +65,6 @@ Mozilla.FxaForm = (function(Mozilla) {
     }
 
     if (fxaForm) {
-        // form is Firefox only right now
```
alexgibson: We probably don't want to remove this, right?

bedrock/firefox/templates/firefox/profile/profile-base.html
```diff
@@ -0,0 +1,82 @@
...
+{% block page_title %}{{ _('An Important Change in Firefox: Dedicated Profiles') }}{% endblock %}
+
+{% block page_css %}
+  {{ css_bundle('profile_per_install') }}
+{% endblock %}
+
+{% block body_class %}state-fxa-default{% endblock %}
+
+{% block site_header %}
+<header id="page-header" class="page-header">
```
alexgibson: We should also think about turning this into a Protocol component perhaps? This kind of  header features on a few different pages now, such as here and `/whatsnew/`, `/firefox/new/` etc?

craigcook: Yeah I think it probably should. I already swiped it from whatsnew for this page so it's a perfect example of reusable code.

media/css/firefox/profile-per-install.scss
craigcook: Yup, it's a problem. We keep getting designs that are almost-but-not-quite Protocol. The ideal solution is to make such design changes in Protocol first. That's how a design system is supposed to work. But we're not given the time to follow the best procedure so we hack and rack up technical debt.

This page should have been in production last week and we really can't wait any longer. So I can either go against the intended design and use only what's already in Protocol, or we can keep the tweaks to ship it as designed and take on the added debt.

Honestly I'm not sure which is best. If we de-tweak this page and change things Protocol later we'd probably still have to revisit the page to implement the changes, even if it's just adding a new layout class. So in that case maybe we should just keep the tweaks and take the debt.

Amy suggested we should keep a running list of all the Protocol pages in bedrock, and we can include what components are in use on them, with notes about any tweaks that were made to each one. Keeping such a list would have the dual benefits of 1) making sure we don't forget about pages that need to be updated when Protocol changes and 2) building a strong case for how much technical debt we're incurring when we need to ask for time to work on Protocol.

alexgibson: Yeah at this point I think we just have to run with the custom changes. For future work, I think we need to push back harder on custom designs if they want us to rush work in. These are both really basic pages visually, so probably could have been built with existing components.

media/css/protocol/components/_fxa.scss
craigcook: Good idea. I don't think we should make that change in this PR though since it will conflict with #6173

alexgibson: I don't see this file used in #6173, but I may be misunderstanding something?

craigcook: Ah you're right, that's a Pebbles page so it's using the Pebbles version. But I really meant I didn't want to move it to /base in this PR, since we'd move the Pebbles one to base as well.

alexgibson: Yeah no worries, I filed an issue so we can move both at the same time, which probably makes more sense anyway https://github.com/mozilla/bedrock/issues/6211

Issue #6082, opened 2018-08-27
[Sept 10] Update UI tour copy for identity popup changes in 63


ejregithub:
https://bugzilla.mozilla.org/show_bug.cgi?id=1484585

Success Criteria:
- [ ]  content (new strings) must be live in prod on Sept 10th

alexgibson:
Blocked due to ambiguity in the slide deck.

PR #6114, opened 2018-09-03
Update Tracking Protection tour strings for Content Blocking shield study (#6082)

alexgibson:
## Description
- Updates existing Tracking Protection tour for Content Blocking shield study, to launch during 63 Beta.
- Shield study will be en-US only for Firefox 63 Beta, but the ask is to localize these new strings ahead of rolling out to all locales post-study.
- Strings have been [tested and verified on demo](https://bugzilla.mozilla.org/show_bug.cgi?id=1484585#c10) by the Firefox team. Further QA will happen when this goes to production before the shield study is launched.

~I will update here when strings are final~ strings are final

**Needs to be in prod for Monday 10th September**

## Issue / Bugzilla link
#6082

## Testing
First, make sure to **test in Nightly** (strings have not yet been uplifted to Beta).

Next, whitelist UITour for the demo instance by creating a new preference called `browser.uitour.testingOrigins` in `about:config` and assign it a value of `https://bedrock-demo-agibson.oregon-b.moz.works/`.

Then you should be able to test for following URLs:

Private Browsing mode (make sure to test in a new PB window):
- https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/63.0/tracking-protection/start/?variation=0
- https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/63.0/tracking-protection/start/?variation=1
- https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/63.0/tracking-protection/start/?variation=2

Normal mode (make sure to test in a non PB window)
- https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/63.0/tracking-protection/start/?variation=0&newtab=true
- https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/63.0/tracking-protection/start/?variation=1&newtab=true
- https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/63.0/tracking-protection/start/?variation=2&newtab=true

PR #6169, opened 2018-09-13
Capitalize 'Restart Tour' in variation 0 of Tracking Protection tour (#6082)

alexgibson:
## Description
- A minor string change: "Restart tour" to "Restart Tour".
- Both strings are already in the template, so no l10n required.

## Issue / Bugzilla link
#6082

## Testing
- Test in Nightly.
- Page should be opened in a private browsing window with Content Blocking enabled.
- URL: `/firefox/63.0/tracking-protection/start/?variation=0`

### DIFF
--- a/bedrock/firefox/templates/firefox/tracking-protection-tour/variation-0.html
+++ b/bedrock/firefox/templates/firefox/tracking-protection-tour/variation-0.html
@@ -18,3 +18,9 @@ data-panel3-text-alt="{{ _('It’s easy to turn on Tracking Protection for the w
 data-panel3-step="{{ _('3 of 3') }}"
 data-panel3-button="{{ _('Got it!') }}"
 {% endblock %}
+
+{% block thanks %}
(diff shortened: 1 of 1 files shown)


stephaniehobson:
This says you're just capitalizing a string but the diff shows a new block being added?

alexgibson:
> This says you're just capitalizing a string but the diff shows a new block being added?

This change is copying the existing block from the base template, just changing the capitalization for this variation.

flodolo:
For the record, this broke the button for all language, because the "Restart Tour" string is not available in any lang file :-\

flodolo:
Correct: added in https://github.com/mozilla/bedrock/pull/6114, never exposed in https://github.com/mozilla-l10n/www.mozilla.org/pull/242, so it's all our fault.

In the future, it would be great to double check via [Transvision](https://transvision.mozfr.org/) more than templates.

Issue #6188, opened 2018-09-18
/new title and description test


rraue:
Right now the description of /firefox/new/ is too long and cut in Search Engines:
<img width="625" alt="screen shot 2018-09-18 at 16 10 29" src="https://user-images.githubusercontent.com/36700618/45693431-65425a00-bb5d-11e8-98a1-1a399fafdff7.png">
I would like to test if we will get a higher ctr in serps with a fitting description and even better rankings with a better title:
<img width="738" alt="screen shot 2018-09-18 at 16 13 39" src="https://user-images.githubusercontent.com/36700618/45693645-d97cfd80-bb5d-11e8-8366-94fdb9c2ddff.png">

Old title: "Download Firefox — Free Web Browser"
New title: "Download Mozilla Firefox — The Fast and Free Web Browser 🦊"

Old description: "Download Mozilla Firefox, a free Web browser. Firefox is created by a global non-profit dedicated to putting individuals in control online. Get Firefox for Windows, macOS, Linux, Android and iOS today!"
New description: "Download Firefox for Windows, Mac, Linux, Android, and iOS. Firefox is an independent, people-first browser made by Mozilla. Get it now, it’s free and fast. ✓"

I do not have a hypothesis about how it will influence ctr, clicks, and rankings but its not much work and even little impact would be huge, as it is our main organic landing page. I will monitor the test closely, so we can redo it when the results are worse after three days. If inconclusive or positive the test should run two weeks.

rraue:
To not run two tests on one page at the same time, this one is on hold and blocked until results for https://github.com/mozilla/bedrock/issues/6462 are conclusive.

rraue:
Closed because of new tests on https://github.com/mozilla/bedrock/issues/8697#issuecomment-603335297

PR #6194, opened 2018-09-19
/firefox/new/  title and description test

rraue:
## Description
Testing title and meta-description of /firefox/new/ with a switch, so the test can be turned off if results are negative
## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/6188

### DIFF
--- a/bedrock/firefox/templates/firefox/new/base.html
+++ b/bedrock/firefox/templates/firefox/new/base.html
@@ -10,18 +10,33 @@
 
 {% block page_title_prefix %}{{_('Download Firefox')}} — {% endblock %}
 
-{# Bug 1438302 Avoid duplicate content for en-CA and en-GB pages. #}
-{%- block page_title -%}
-  {% if LANG == 'en-CA' %}
-    Free Web Browser (CA)
-  {% elif LANG == 'en-GB' %}
-    Free Web Browser (UK)
-  {% else %}
-    {{_('Free Web Browser')}}
-  {% endif %}
-{%- endblock -%}
 
-{% block page_desc %}{{_('Download Mozilla Firefox, a free Web browser. Firefox is created by a global non-profit dedicated to putting individuals in control online. Get Firefox for Windows, macOS, Linux, Android and iOS today!')}}{% endblock %}
+{% if switch('firefox_new_title_experiment') %}
+    {# Bug 1438302 Avoid duplicate content for en-CA and en-GB pages. #}
+    {%- block page_title -%}
+    {% if LANG == 'en-CA' %}
+    The Fast and Free Web Browser 🦊 (CA)
+    {% elif LANG == 'en-GB' %}
+    The Fast and Free Web Browser 🦊 (UK)
+    {% else %}
+    {{_('The Fast and Free Web Browser 🦊')}}
+    {% endif %}
+    {%- endblock -%}
+    {% block page_title_suffix %}{% endblock %}
+    {% block page_desc %}{{_('Download Firefox for Windows, Mac, Linux, Android, and iOS. Firefox is an independent, people-first browser made by Mozilla. Get it now, it's free and fast. ✓')}}{% endblock %}
+{% else %}
+    {# Bug 1438302 Avoid duplicate content for en-CA and en-GB pages. #}
+    {%- block page_title -%}
+    {% if LANG == 'en-CA' %}
+    Free Web Browser (CA)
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/new/base.html
```diff
@@ -10,18 +10,33 @@
...
+    {% if LANG == 'en-CA' %}
+    The Fast and Free Web Browser 🦊 (CA)
+    {% elif LANG == 'en-GB' %}
+    The Fast and Free Web Browser 🦊 (UK)
+    {% else %}
+    {{_('The Fast and Free Web Browser 🦊')}}
+    {% endif %}
+    {%- endblock -%}
+    {% block page_title_suffix %}{% endblock %}
+    {% block page_desc %}{{_('Download Firefox for Windows, Mac, Linux, Android, and iOS. Firefox is an independent, people-first browser made by Mozilla. Get it now, it's free and fast. ✓')}}{% endblock %}
```
alexgibson: You need a curly apostrophe in `it’s` here. This is currently throwing an error

bedrock/firefox/templates/firefox/new/base.html
```diff
@@ -10,18 +10,33 @@
...
+{% else %}
+    {# Bug 1438302 Avoid duplicate content for en-CA and en-GB pages. #}
+    {%- block page_title -%}
```
alexgibson: You can't include a block twice in a template like this. You will need to do the switch conditonal inside each block. I'll work on an example.

bedrock/firefox/templates/firefox/new/base.html
```diff
@@ -10,18 +10,33 @@
...
+
+{% if switch('firefox_new_title_experiment') %}
+    {# Bug 1438302 Avoid duplicate content for en-CA and en-GB pages. #}
+    {%- block page_title -%}
+    {% if LANG == 'en-CA' %}
+    The Fast and Free Web Browser 🦊 (CA)
+    {% elif LANG == 'en-GB' %}
+    The Fast and Free Web Browser 🦊 (UK)
+    {% else %}
+    {{_('The Fast and Free Web Browser 🦊')}}
```
alexgibson: This is breaking translations, you'll need to use the existing `Free Web Browser` string as a fallback.

bedrock/firefox/templates/firefox/new/base.html
```diff
@@ -10,18 +10,33 @@
 
 {% block page_title_prefix %}{{_('Download Firefox')}} — {% endblock %}
 
-{# Bug 1438302 Avoid duplicate content for en-CA and en-GB pages. #}
-{%- block page_title -%}
-  {% if LANG == 'en-CA' %}
+
+{% if switch('firefox_new_title_experiment') %}
```
alexgibson: It would be nice to include a comment here linking to the issue, as a clue for what this all means to future adventurers.

bedrock/firefox/templates/firefox/new/base.html
```diff
@@ -10,18 +10,33 @@
...
+    {# Bug 1438302 Avoid duplicate content for en-CA and en-GB pages. #}
+    {%- block page_title -%}
+    {% if LANG == 'en-CA' %}
+    The Fast and Free Web Browser 🦊 (CA)
```
alexgibson: @pmac are there any encoding gotcha's about including emoji like this that you are aware of? I think this should be fine, but I may be unaware of edgecases.

pmac: I'm actually not sure. It should be like any other unicode character in theory, so I'd say no, but I don't think we've done this exact thing before. So I'd say it "should" be okay, but we should keep an eye on it.

alexgibson: Ok thanks. Yeah we're putting this behind a switch and only for a couple of locales. We can always flip it off if we notice anything weird. Thanks

alexgibson:
@rraue is this something you're still interesting in testing out?

rraue:
Yes, I still would like to test this, I am just not sure, if my coding skills are good enough for this. I will update the description, so it reflects the goal of this test and the changed title and description.

alexgibson:
@rraue happy to help finish this up here if you need

rraue:
@alexgibson That would be great, I will learn from your code. I don't know yet how to implement the switch right and the fallback you mentioned. I updated the description and changed there the wrong apostrophe as well.

Issue #6203, opened 2018-09-19
[9/25] - Update 'Evolving the Firefox Brand' card with Monitor content


ejregithub:
Description: Swap out Card 1 (Evolving the Firefox Brand) for Firefox Monitor content on 9/24 (to drive people directly to the Monitor site)

kellymj:
Headline: Protect yourself with Firefox Monitor

Blurb: Find out what hackers already know about you and get alerts when data breaches put you at risk.

URL: https://monitor.firefox.com/  (with UTM tags that your team adds)

ejregithub:
![sp_fx_monitor_moz org homepage 01](https://user-images.githubusercontent.com/20547886/45849622-89ca3d80-bce7-11e8-9978-2df0ac936511.png)

ejregithub:
@alexgibson - all assets are now here ^^

alexgibson:
Enabled once again, closing.

PR #6223, opened 2018-09-25
Add switch for Firefox Monitor homepage card (#6203)

alexgibson:


alexgibson:
@djst gave this a sanity check, and it's a small change we need to push asap, so I'm going to merge it.

Issue #5944, opened 2018-07-25
Experiment - ask for email prior to download


ejregithub:
[manapage w. test write up](https://mana.mozilla.org/wiki/display/FIREFOX/Ask+for+email+prior+to+download+test)

Success Criteria

- [ ] non-Firefox en-US visitors on Windows only
- [ ] Implemented on www.mozilla.org/firefox/

Prototype:
https://invis.io/SDNAAE3ZCK4

Sketch File:
https://share.goabstract.com/3473d71b-737b-480a-8dd0-658d84692668

thejeffreytaylor:
[Copy doc](https://docs.google.com/document/d/1ijjDcATdeBGWrBdyTWmJ-4la4RIb4Tyepn3VHv7Dy1s/edit?usp=sharing) for all states (not the usual copy deck template due to the nature of this project, but I can drop the text in there anytime upon request).

craigcook:
For further explanation, this is showing a sticky newsletter subscription form at the bottom of the Firefox product page that will appear some time after the initial page load (after scrolling a certain distance? Or after a certain amount of time? Was this decided?). 

It should only be shown when viewports are large enough to show it without obscuring too much of the page (so not shown on small screens). There should be a "close" button to dismiss the sticky form, and if closed it should still appear (non-sticky) further down the page if the user continues to scroll down to it. There's precedent for this treatment on the now-deleted SmartOn pages so some code could be resurrected from git history for reference.

The hypothesis is that making the form more noticeable will get more attention from people who are Firefox-curious but haven't yet downloaded. At present the newsletter form is near the bottom of a very long page so it's easily overlooked or ignored. Making the form more noticeable (but still unobtrusive) should increase the number of subscribers. If this proves successful it may be something we try on other pages as well.

Success criteria above says this is for Windows only. Is that accurate? Unless this is tied to some funnelcake experiment I think this can be shown to all platforms. Do we still want to target non-Firefox or should it be shown to all browsers?

michaelmccombie:
Here are links to the Prototype and Sketch file:

Prototype:
https://invis.io/SDNAAE3ZCK4

Sketch File:
https://share.goabstract.com/3473d71b-737b-480a-8dd0-658d84692668

alexgibson:
> Success Criteria

> non-Firefox en-US visitors on Windows only
> Implemented on www.mozilla.org/firefox/

To me these aren't success criteria, they sound more like requirements.

alexgibson:
[...]

PR #6224, opened 2018-09-25
Remove pre-download experiment code (#5944)

alexgibson:
## Description
- Removes experiment code added in https://github.com/mozilla/bedrock/pull/5999

DNM until I get the OK to remove this.

## Issue / Bugzilla link
#5944

## Testing
- Double check I didn't miss any stray references?


alexgibson:
@ejregithub are we ok to remove the experiment code here? This would never be reusable for anything past just an experiment, so if we're satisfied that the test is over, I see no real benefit in keeping it around.

djst:
Results suggest the experiment was fairly successful. No drop in downloads, increase in signups (though poor opt-in clicks from confirmation email). I think the next step is to iterate on the experiment rather than removing the code.

alexgibson:
> Results suggest the experiment was fairly successful. No drop in downloads, increase in signups (though poor opt-in clicks from confirmation email). I think the next step is to iterate on the experiment rather than removing the code.

Before we start to iterate on this, It would be good to see a throough analysis of the results before we draw any solid conclusions. The data I've seen suggests that there was a drop in downloads, which others have also commented on - so there may be some ambiguity here in how people interpret the data. We should also survey the quality of emails entered, to wokr out if there we're a high number of bugus emails entered (the high drop off in confirmations may be a clue to this).

alexgibson:
@djst just checking in on the status here. Now that the data has been more thoroughly analysed, are we good to go ahead and remove this left-over code?

alexgibson:
> If we'll iterate on pre-download experiments, will the new test be so different that we'll not need this code? If so, I'm fine with removing this.

Everything is in Git history, so nothing is lost even if we did decide to reuse any of this.

alexgibson:
> If we'll iterate on pre-download experiments, will the new test be so different that we'll not need this code? If so, I'm fine with removing this.

Everything is in Git history, so nothing is lost even if we did decide to reuse any of this.

alexgibson:
> If we'll iterate on pre-download experiments, will the new test be so different that we'll not need this code? If so, I'm fine with removing this.

Everything is in Git history, so nothing is lost even if we did decide to reuse any of this.

alexgibson:
> If we'll iterate on pre-download experiments, will the new test be so different that we'll not need this code? If so, I'm fine with removing this.

Everything is in Git history, so nothing is lost even if we did decide to reuse any of this

alexgibson:
Ok, this is rebased and ready for review.

stephaniehobson:
R+ ✨

stephaniehobson:
Waiting for master to turn green so I can do a push. I'll leave this unmerged for now.

PR #6233, opened 2018-09-26
Always use database product-detail storage

pmac:
Now that we're distributing sqlite databases with these data we should use it in non-prod situations as well. The file-based data is very old.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -99,9 +99,8 @@ PROD_DETAILS_URL = config('PROD_DETAILS_URL',
 # This ultimately controls how LANGUAGES are constructed.
 PROD_DETAILS_CACHE_NAME = 'product-details'
 PROD_DETAILS_CACHE_TIMEOUT = 60 * 15  # 15 min
-default_pdstorage = 'PDDatabaseStorage' if PROD else 'PDFileStorage'
 PROD_DETAILS_STORAGE = config('PROD_DETAILS_STORAGE',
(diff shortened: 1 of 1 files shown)


kyoshino:
Tested this locally. Somehow the latest releases are still missing even after runnning `bin/bootstrap.sh --force`: 62.0, 62.0.2, 60.2.0, 60.2.1. Do I have something else to do?

pmac:
I tried running via docker just now (`make pull run`) and I see the latest releases. I'll take a look at non-docker.

kyoshino:
I just checked my `.env` file but it was the same as `.env-dist`, didn't have any of `PROD_DETAILS_*` settings.

kyoshino:
And I dug into `bedrock.db` and it had the latest releases from [`firefox_history_stability_releases.json`](https://product-details.mozilla.org/1.0/firefox_history_stability_releases.json) up to 62.0.2. Still the latest release on `http://localhost:8000/en-US/firefox/releases/` is 61.0.2.

jgmize:
I tested this locally, and after running `bin/run-db-download --force` and `make pull run` as described above I'm getting 62.0.2 on http://localhost:8000/en-US/firefox/releases/

jgmize:
@kyoshino if you check http://localhost:8000/en-US/firefox/releases/ again are you still getting the older release? The fact that you have the latest release in the DB makes me suspect that you may have been looking at a locally cached version of the page, which could persist for up to 10 minutes unless you restarted the server process to clear the cache.

PR #6235, opened 2018-09-27
[fix #5764] Update MOSS pages

### LINKED ISSUE #5764: [9/17] MOSS updates
[New updates to MOSS site](https://docs.google.com/document/d/1As5KUmbvF9TTlGn_BvPxcf4fo9Mgk-9WVzaK4yENLMI/edit?ts=5b893b09)

craigcook:
## Description
Updates main MOSS page with new content.

This page is localized in Hindi (hi-IN) but so much text has changed that it seemed best to make a whole separate index file and generate a new lang file for `mozorg/moss/index-092018` so the Hindi page can fall back to the old content for the time being.

## Issue / Bugzilla link
#5764 

## Testing
https://bedrock-demo-craigcook.oregon-b.moz.works/moss/


bedrock/mozorg/urls.py
```diff
@@ -293,11 +293,12 @@
     page('internet-health/web-literacy', 'mozorg/internet-health/web-literacy.html'),
     page('internet-health/decentralization', 'mozorg/internet-health/decentralization.html'),
 
-    page('moss', 'mozorg/moss/index.html'),
+    url(r'moss/$', views.moss_view, name='mozorg.moss'),
```
alexgibson: I believe this should be `r'^moss/$'` else any URL that ends with `moss` would be valid? (e.g. `foo/moss/`)

bedrock/mozorg/views.py
```diff
@@ -202,3 +203,14 @@ def about_view(request):
         template_name = 'mozorg/about.html'
 
     return l10n_utils.render(request, template_name)
+
+
+def moss_view(request):
+    locale = l10n_utils.get_locale(request)
+
+    if lang_file_is_active('mozorg/moss/index-092018', locale):
```
alexgibson: Is this logic worth adding a test for?

bedrock/mozorg/urls.py
craigcook: D'oh! Good catch.

bedrock/mozorg/views.py
craigcook: Perhaps, but I'm hopeful it will be localized quickly so the logic may be short-lived. It's only translated in Hindi and I don't know how they're going about handling updates.

craigcook:
@peiying2 This page is only localized in Hindi but if I recall they went through a sort of non-traditional process to translate it. For these content updates I've done it all as a new template with a new lang file, allowing non-English locales to fall back to the old template. 

But now I'm having second thoughts and wondering if that's a good idea. Perhaps we should just remove the outdated Hindi translation and update the main `index.html`. Then hi-IN would just get the English page until they can translate it.

Thoughts?

peiying2:
@craigcook yes, let's keep outdated page up until the Hindi page is fully localized, then we can move to this page. Also I recall that time it was localized by an agency, not the community. I will see if we can do the same to help move things along.

craigcook:
@peiying2 Thanks! With that blessing, I think we're good to merge.

PR #6239, opened 2018-10-01
[fix bug 1483611] Remove interest section from newsletter preferences page

alexgibson:
https://bugzilla.mozilla.org/show_bug.cgi?id=1483611

### DIFF
--- a/bedrock/newsletter/templates/newsletter/existing.html
+++ b/bedrock/newsletter/templates/newsletter/existing.html
@@ -141,46 +141,6 @@
         </div>
       </div><!-- close #default-newsletters -->
 
-      <div class="field{% if did_confirm %} hidden{% endif %}" id="other-newsletters">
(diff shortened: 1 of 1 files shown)


alexgibson:
@pmac r? Unclear if there's anything else that needs doing here, or if it really is this simple.

pmac:
What you've done looks good. I'll check the view and see whether there is any cleanup we can do there.

jessilyndavis:
Yep! We're getting rid of these interest categories. As soon as they disappear from the email pref center, I'm going to remove these entries in basket completely.  So no worries that people get globally unsubscribed if they don't show at the moment. 

The bigger issue/question is: if I delete them from Basket (and eventually the fields from SFDC) - will it cause issues with this code?

pmac:
Nope. Once we remove this you should be able to remove them from basket and SFDC at will. It would only be a problem if someone were trying to subscribe or unsubscribe from them, which should be impossible after this.

PR #6247, opened 2018-10-02
Remove display: none from .content-container to fix #6246

### LINKED ISSUE #6246: Blank notification area at top of what's new page
I booted an old machine for the first time in a while, opened up Firefox, and landed on https://www.mozilla.org/en-US/firefox/60.0.2/whatsnew/?oldversion=59.0.2 with a blank grey notification bar at the top, like so:

screenshot-2018-10-1 see what s new with firefox

jgmize:
## Description

Remove display: none from .content-container stylesheet to prevent empty notifications on whatsnew

## Issue / Bugzilla link

#6246

## Testing

Load whatsnew pages such as https://www.mozilla.org/en-US/firefox/60.0.2/whatsnew/?oldversion=59.0.2 with an out of date browser and verify content appears.

### DIFF
--- a/media/css/firefox/whatsnew/whatsnew-60.scss
+++ b/media/css/firefox/whatsnew/whatsnew-60.scss
@@ -90,7 +90,6 @@ h1, h2, h3, h4, h5, h6 {
 
 .content-container {
     color: #fff;
-    display: none;
     padding: 0 10px;
     z-index: 10;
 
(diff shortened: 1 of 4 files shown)


jgmize:
There are several other sections with the `content-container` class that need the `display: none` by default so they can be shown conditionally; this causes more problems than it solves.

PR #6248, opened 2018-10-02
Make notification banner content container visible to fix #6246

### LINKED ISSUE #6246: Blank notification area at top of what's new page

jgmize:
## Description

Make notification banner content container visible 

## Issue / Bugzilla link

#6246 

## Testing

Load https://www.mozilla.org/en-US/firefox/60.0.2/whatsnew/?oldversion=59.0.2 in an out of date version of firefox.

### DIFF
--- a/media/css/base/notification-banner.less
+++ b/media/css/base/notification-banner.less
@@ -38,7 +38,7 @@
         font-style: italic;
     }
 
-    .content-container {
+    .notification-content-container {
         float: left;
         position: relative;
         width: 65%;
@@ -120,7 +120,7 @@
             width: @widthTablet;
         }
 
-        .content-container:after {
+        .notification-content-container:after {
             .background-size(192px 190px);
             height: 190px;
             right: -200px;
@@ -138,7 +138,7 @@
             width: auto;
         }
 
-        .content-container:after {
+        .notification-content-container:after {
             content: none;
         }
 
@@ -148,7 +148,7 @@
     }
 
     @media only screen and (max-width: @breakMobileLandscape) {
-        .content-container {
+        .notification-content-container {
             width: 80%;
         }
     }
@@ -179,4 +179,3 @@
         margin-bottom: 100px;
     }
 }
-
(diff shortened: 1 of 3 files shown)


jgmize:
pushed to demo branch, deploying in https://ci.us-west.moz.works/job/bedrock_multibranch_pipeline/job/demo%252Fjgmize/

jgmize:
https://bedrock-demo-jgmize.oregon-b.moz.works/en-US/firefox/60.0.2/whatsnew/?oldversion=59.0.2 shows me the following on an out of date Firefox:

image

alexgibson:
@jgmize thanks for spotting this 👍 

I've tested this locally and it does fix the immediate bug, however there are a couple of remaining issues still:

- The fix is relying on an in-page class called `visible`, which is not related to the notification itself, so the issue could still occur on other pages potentially.
- The notification is still inheriting some in-page styles from the `.content-container` class name (e.g. the text is in white, which leaves a color contrast/a11y problem).

I think the best solution here would probably be to update the `.content-container` class name in the notification to something less generic, like `.notification-content-container`. This should help avoid the likelihood of the notification inheriting any unwanted styles that may prevent it from displaying correctly.

Note: you would need to update the classname wherever it occurs in both the JS and CSS source files:

- [mozilla-notification-banner.js](https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-notification-banner.js#L152)
- [notification-banner.less](https://github.com/mozilla/bedrock/blob/master/media/css/base/notification-banner.less)
- [notification-banner.scss](https://github.com/mozilla/bedrock/blob/master/media/css/pebbles/components/_notification-banner.scss)

Let me know if you want to work on doing this. If you don't have time I can see if I can knock it out. Thanks!

jgmize:
Thanks @alexgibson, that's a much better solution. I force pushed your suggested changes to this PR branch and to demo/jgmize.

PR #6249, opened 2018-10-02
Legal: remove Firefox Marketplace

caugner:
## Description

Firefox Marketplace does not exist anymore, therefore it doesn't make sense to link the (non-existing) terms anymore.

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1496300

## Testing

(None.)


ejregithub:
I'm vetting this with Legal.

alexgibson:
@ejregithub fwiw, the link here times out and goes nowhere. I can't see that legal would want to keep it around.

PR #6253, opened 2018-10-02
Fix #6252 by using mozorg.moss.index name for url 

### LINKED ISSUE #6252: NoReverseMatch error for mozorg.moss.index
https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/4772958/

https://www.mozilla.org/en-US/internet-health/open-innovation/ is working correctly for me, and we've only seen a few of these errors. It may have something to do with https://github.com/mozilla/bedrock/blob/4d61c48dbf0f25da13fb412adb535979d77f0fd4/bedrock/mozorg/views.py#L211-L212

The correct name appears to 'mozorg.moss': https://github.com/mozilla/bedrock/blob/4d61c48dbf0f25da13fb412adb535979d77f0fd4/bedrock/mozorg/urls.py#L299

jgmize:
## Description

Fix intermittent NoReverseMatch error

## Issue / Bugzilla link

#6252

## Testing

check https://www.mozilla.org/en-US/internet-health/open-innovation/ and https://www.mozilla.org/en-US/grants/ (or local/dev/demo equivalent)

### DIFF
--- a/bedrock/mozorg/urls.py
+++ b/bedrock/mozorg/urls.py
@@ -296,7 +296,7 @@ urlpatterns = (
     page('internet-health/web-literacy', 'mozorg/internet-health/web-literacy.html'),
     page('internet-health/decentralization', 'mozorg/internet-health/decentralization.html'),
 
-    url(r'^moss/$', views.moss_view, name='mozorg.moss'),
+    url(r'^moss/$', views.moss_view, name='mozorg.moss.index'),
     page('moss/foundational-technology', 'mozorg/moss/foundational-technology.html'),
     page('moss/mission-partners', 'mozorg/moss/mission-partners.html'),
     page('moss/mission-partners-india', 'mozorg/moss/mission-partners-india.html'),


jgmize:
This should hopefully fix the regression introduced in https://github.com/mozilla/bedrock/pull/6235 -- alternatively, we could change https://github.com/mozilla/bedrock/pull/6235/files#diff-71245d0bee3444a747f3251f25fc59a4R296 to 'moss.org.index'

jgmize:
Just saw another [error in sentry](https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/4772958/) for https://www.mozilla.org/en-US/grants/ which also refers to 'moss.org.index', so this is an incomplete fix.

jgmize:
Force-pushed an update to fix the name instead of changing the references.

alexgibson:
Looks like this is breaking things in production, sigh :(

The pages only appear broken because of URLs appear within l10n tags, so it only effects certain locales.

PR #6279, opened 2018-10-05
Add 'tl' to list of production languages

pmac:
Fix bug 1482045


pmac:
This has to wait on https://github.com/mozilla-l10n/www.mozilla.org/pull/249 to merge and make it to prod langfiles so that we don't have untranslated pages showing as active.

PR #6281, opened 2018-10-05
Remove switch for election firstrun page

stephaniehobson:
## Description
I need to put the bundle download buttons behind a switch and I want
to re-use the name of this one instead of coming up with a 3rd name.

## Issue / Bugzilla link
mozilla/bedrock#6186


stephaniehobson:
Actually, nevermind.

Issue #6145, opened 2018-09-07
Update title of the ESR page


pmac:
It is currently "Find the right Firefox for you", but there is only one option now that legacy ESR is no longer available. We should update this title and remove the switch if/else block for the legacy button.

See #6133

anshulahuja98:
i can see updates on the issue link you attached, please inform the current status of this issue

stephaniehobson:
Fixed in #6290

PR #6290, opened 2018-10-13
updated esr title to 'Firefox Extended Support Release'

annezhou920:
## Description
Updated ESR page title from "Find the right Firefox for you" to "Firefox Extended Support Release"

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/6145

### DIFF
--- a/bedrock/firefox/templates/firefox/organizations/organizations.html
+++ b/bedrock/firefox/templates/firefox/organizations/organizations.html
@@ -28,7 +28,7 @@
 
 {% block main_feature %}
     <div class="pick">
-      <h1 class="large">{{ _('Find the right Firefox for you') }}</h1>
+      <h1 class="large">{{ _('Firefox Extended Support Release') }}</h1>
       <div class="row">
         <div class="col">
           <p>


pmac:
I do think this is an improvement, but there would then be a lot of the phrase "Firefox Extended Support Release" at the top of the page. I think we also need to update the download button to be more simple.

Do you have thoughts @alexgibson?

alexgibson:
This seems like a copy related question - @ejregithub can you perhaps get Clay to take a look at this? Seems like a good first-bug? :)

annezhou920:
@pmac how about just "Downloads" for the button?

@alexgibson please let me know what the correct copy should be and i can make the copy update :)

stephaniehobson:
How about this?
- Change the title, and remove the bottom margin
- Remove the bold text
- Leave the button

screen shot 2018-11-07 at 1 57 43 pm

@annezhou920 What do you think of that? Are you comfortable editing CSS? We just need to change the bottom border for `.pick h1` to `0`.

stephaniehobson:
Thanks for the speedy updates 💨 

There's a second margin declaration for the heading on line 121, that needs the bottom margin removed as well.

annezhou920:
@stephaniehobson thank you for the direction!

PR #6293, opened 2018-10-14
Compressed media/img PNGs with PNGQuant

vcz-fr:
## Description
Compressed all the PNGs I could find on media/img with PNGQuant. This merge request is pretty large. I can split it to reduce the review complexity.
Please check for visual differences.

## Issue / Bugzilla link
No issue; proposed improvement

## Testing
Manual visual testing for most of them.


alexgibson:
Thanks for this pull request - the majority of images in bedrock should have been pretty carefully optimised already, and seeing as PNGQuant is a lossy algorithm, I'm a bit hesitant to blanket re-compress the entire image directory (even though there are likely some good optimisations here). I'm just not sure we have time to manually QA all these changes.

A better approach might be to identify images that don't seem to have been optimised at all yet. If that's something you have time and willingness to work on, may that be a suitable alternative approach?

Issue #6287, opened 2018-10-12
[10/19] Kill "Powered by Mozilla" page


ejregithub:
Description: Legal recently rediscovered the "Powered by Mozilla" page. It looks pretty defunct, and Michael F. reached out to Mike Hoye who confirmed it's defunct. The "form" on the page to ask permission for the badge no longer goes anywhere. The page needs to be removed.

- Developer Hours [  ]

Success Criteria:
- [ ] The page is removed from mozilla.org by the assigned date (10/18)

ejregithub:
@amychurchwell - could you please this one up, letting me know how you would size it (in hours/days) for total dev time required including review/merge/push?

amychurchwell:
@ejregithub - Work-wise, >1 hour --assuming I am correct in seeing that there are hardly any dependencies on this page. Review/merge/push depends on how busy everyone but possibly today.

kyoshino:
Just FYI: The [linked form](https://bugzilla.mozilla.org/enter_bug.cgi?product=Marketing&format=poweredby) will also be removed from Bugzilla today or tomorrow. https://github.com/mozilla-bteam/bmo/pull/830

amychurchwell:
@kyoshino Thanks Kohei! This should be reviewed/merged today.

PR #6295, opened 2018-10-15
Remove about/powered-by

amychurchwell:
## Description
Remove the "Powered by Mozilla" page and all references to it.
Needs to be live 10/18

## Issue / Bugzilla link
#6287


bedrock/mozorg/redirects.py
```diff
@@ -229,9 +229,6 @@ def decider(request, **kwargs):
     # bug 854561
     redirect(r'^projects/mozilla-based(\.html|/)?', 'mozorg.projects.mozilla-based'),
 
-    # bug 851727
-    redirect(r'^projects/powered-by(\.html|/)?', 'mozorg.powered-by'),
```
craigcook: I think we probably still need this redirect but should change the destination. And we should add a new redirect for `/about/powered-by` that goes to the same place. I'm not sure what a good destination might be, though. Maybe just the home page?

amychurchwell: @craigcook I will redirect to the homepage for now. I'm sure a bug will be filed if that creates issues contextually within the site. 🙂

PR #6297, opened 2018-10-15
Firefox Rocket -> Lite privacy document changes

pmac:
This includes all of the necessary changes. All changes have been made in the legal-docs repo and this PR updates to that version. The go-live for this is set for Nov 1st.

Fix bug 1498613


pmac:
This is ready for review, but I've added DNM tag as we need to wait until Nov 1 for this to go live.

chianger:
reminder: logo update is necessary as well. assets are addressed here https://docs.google.com/document/d/1Ls2x0Gzcu4w_TohDkOFJYp7VOYlB8nmrl7w2ms75ioc/edit?ts=5bbd31b3#

pmac:
Logo update is now included in this PR.

pmac:
@alexgibson I've set you as reviewer, but feel free to pass it off if you want/need to. It's no rush, but it can be reviewed now, but go-live is not until Nov 1st.

chianger:
So to be precise we're pushing the apk to google play at 11am Nov-1st (UTC+8) == 8pm Oct-31 (UTC-7). May not be a PDT-friendly time schedule, so if it's easier you can merge it a few hours earlier.

media/css/privacy/privacy.less
```diff
@@ -193,7 +193,7 @@
       background-position: center -40px;
     }
 
-    &.policy-firefox-rocket a:before {
+    &.policy-firefox-lite a:before {
```
alexgibson: The image sprite that this selector references still contains the old rocket logo, is that expected? I don't see where this is used.

pmac: This is really just in prep for this being added to the sidebar of the main privacy page. It's not there yet, so we've got some time to update this. Honestly using a sprite is no longer advantageous since we have HTTP2 enabled, but that's beside the point :) We should fix up the sprite though, you're right. Do you know if that's part of another issue or should that be fixed in this one?

alexgibson: Yeah, this page is old and so is the sprite. I'd say don't worry about it right now, as it's not used in the page. We can update the logo later on, and maybe even get rid of the sprite while we're at it?

Issue #6276, opened 2018-10-04
Add an event label to Yandex download page signifying a Yandex referral


ejregithub:
Description(task): Add an event label to Yandex download page signifying a Yandex referral
[Bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1485419)

Analytics Pros specified changes we should make to the download button on the Yandex download page:

data-link-name="Yandex redirect"
data-link-type="Download Firefox (RU)"

//Optional CDs
data-link-position="primary cta"
data-link-group="<link group>" //any extra info
data-cta-position="<cta position>" //any extra info

PR #6307, opened 2018-10-17
Add data attributes to Yandex download button

amychurchwell:
## Description
Add data attributes to Yandex download button.

**data-link-name="Yandex redirect"
data-link-type="Download Firefox (RU)"**

_optional_
_data-link-position="primary cta"
data-link-group="" //any extra info
data-cta-position="" //any extra info_

## Issue / Bugzilla link
#6276

### DIFF
--- a/media/js/firefox/new/yandex/scene1.js
+++ b/media/js/firefox/new/yandex/scene1.js
@@ -131,6 +131,9 @@
             button[i].href = Mozilla.Utils.trans('buttonLink');
             button[i].removeAttribute('data-direct-link');
             button[i].querySelector('.download-title').innerHTML = Mozilla.Utils.trans('buttonText');
+            button[i].setAttribute('data-link-name', 'Yandex redirect');
+            button[i].setAttribute('data-link-type', 'Download Firefox (RU)');
+            button[i].setAttribute('data-link-position', 'primary cta');
         }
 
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/includes/download-button.html
```diff
@@ -67,6 +67,7 @@
            data-link-type="download"
```
amychurchwell: @alexgibson one of the requirements is to add `data-link-type="Download Firefox (RU)"` but I have concerns seeing as it is hard-coded to be `"download"`

Considering some sort of `{% if yandex %}data-link-type="Download Firefox (RU)"{% else %}data-link-type="download"{% endif %}` but am a little unsure how to define if it's `yandex`? Am also unsure if changing it to something other than `"download"` breaks it.

alexgibson:
Hi Amy, the issue seems to suggest we only want to append these data-attributes to the button if the user is geo-located in Russia (so we can distinguish from regular downloads). This means we can't use the Python helper to add the attributes like this, and we instead need to use JS.

The relevant code is here:

https://github.com/mozilla/bedrock/blob/master/media/js/firefox/new/yandex/scene1.js#L128-L134

Sorry if this wasn't quite clear from our talk yesterday

PR #6313, opened 2018-10-18
Fix Bug 1500250 - Broken layout on https://www.mozilla.org/fr/firefox/developer/

kyoshino:
## Description

Use flexbox instead of float to fix the broken layout. The reported page is French but this could happen in any locale.

## Issue / Bugzilla link

[Bug 1500250 - Broken layout on https://www.mozilla.org/fr/firefox/developer/](https://bugzilla.mozilla.org/show_bug.cgi?id=1500250)

## Testing

Visit /fr/firefox/developer/ and make sure the layout is fixed.

### DIFF
--- a/bedrock/firefox/templates/firefox/developer/includes/highlights.html
+++ b/bedrock/firefox/templates/firefox/developer/includes/highlights.html
@@ -1,6 +1,6 @@
 <section class="section section-highlights">
   <div class="content">
-    <div class="highlight-tools">
+    <div class="highlight-item">
       <h2 class="section-title"><span>{{ _('New Tools') }}</span></h2>
 
       <div class="section-body">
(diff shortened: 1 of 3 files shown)


kyoshino:
@craigcook Looks like I beat you 😸

craigcook:
Not related to this particular bug, but something I just noticed in testing... The section titles don't handle wrapping very well.

image

I think increasing the line-height to make room for the underline is the best fix.

craigcook:
One more thing I noticed...

When this page was first built it had exactly two items in that highlights section, that I named `.highlight-tools` and `.highlight-features` but when we added two more items those names became redundant. I think we should update the class names so all four are the same general-purpose element, maybe just `.hightlight-item`. 

Currently the `.highlight-tools` element has a bottom margin on mobile but `.highlight-features` doesn't, and that second class gets reused on the two new items. I think all four should get the bottom margin on mobile so there's some proper spacing.

Sorry for the scope creep!

PR #6314, opened 2018-10-19
update testflight logo

### LINKED ISSUE #5701: Update TestFlight Page With New Logo
https://bugzilla.mozilla.org/show_bug.cgi?id=1428365

philschatz:
## Description

This updates the testflight logo to use the quantum logo.

/cc @amychurchwell  as part of Hacktoberfest

## Issue / Bugzilla link

Fixes #5701


amychurchwell:
Looking good!
@dzingeek do we have a version of this graphic available with the new logo?
screen shot 2018-10-19 at 9 39 30 am

dzingeek:
@amychurchwell since the above graphic doesn't really do anything for the user and this page at some point should get a redesign, I suggest we just lose that graphic for now.

ksandhu16:
How can I resolve this issue?

philschatz:
The extra image is loaded in via CSS at [media/css/firefox/testflight.less lines 129-150](https://github.com/mozilla/bedrock/blob/master/media/css/firefox/testflight.less#L129-L150) and those lines should probably be removed along with the corresponding images.

I am having trouble running `make run` again (specifically the `docker-compose pull release app assets builder app-base` step) so I have not been able to verify that removing those lines fixes the problem.

amychurchwell:
@philschatz thanks again for your help on this issue.
Perhaps I am a little confused by your question. Are you running `make run` & then trying to run `docker-compose up app assets`? If that is the case, it is either one or the other. They are two different options: the longhand docker-compose ...etc. for those who do not have/want to use Make and need to call the docker & compose commands directly.

If you are using `make run` solo it should refresh with any saved changes, if not you can manually refresh. You could also try running `make pull` to pull down the latest Docker images.
More info: https://bedrock.readthedocs.io/en/latest/install.html#make-it-run

RE: CSS
You are correct that those are some of the lines! We should remove those. We may also want to go ahead and remove lines [117-119](https://github.com/mozilla/bedrock/blob/master/media/css/firefox/testflight.less#L117-L119) and either remove padding-bottom and change margin-top to just margin. Or adjust padding-bottom to a more appropriate number. 🙂

philschatz:
I have updated the CSS, reduced the padding by the height of the image, and removed the iOS testflight images.

image


Thanks for helping me debug. I was running `make run` but was just noting that it failed on the `docker-compose line`. It works now, maybe it was an intermittent connection issue. Also, I accidentally kept visiting `http://localhost:8080` instead of port `:8000` which is why the styling did not show up for me.

Issue #6306, opened 2018-10-17
Urgent legal copy changes to some web pages


djst:

craigcook:
I think we'll need some more guidance on these items so we can take proper action to address them.

> On the landing page for IRL, change the line “IRL is an original podcast from Mozilla” by adding “Corporation” to the end.

This seems simple enough at a glance but I wonder if it has broader implications for the brand. We generally refer to "Mozilla" as a singular entity (which I guess is the whole problem here) so if we have to start saying "The Mozilla Corporation" emphatically everywhere we would normally say "Mozilla," that's a really far-reaching change way beyond one tagline on one website.

> At https://www.mozilla.org/en-US/foundation/ there is a description of the roles of the Foundation and the Corporation. Add to the description of the corporation that it owns and is responsible for mozilla.org websites other than foundation.mozilla.org (or whatever is accurate to say, that makes clear that www.mozilla.org and blog.mozilla.org are not foundation content).

The legal team should provide us with accurate copy here, or should we just draft something for them to review? And is it correct to say that the entirety of www.mozilla.org is "not Foundation content"? We have a fair amount of what I would consider Foundation content on the site. The manifesto for example, as well as a lot of governance policies and documentation.

As for blog.mozilla.org, that subdomain houses 80+ separate and distinct blogs with many thousands of unique pages, some of which might be also considered Foundation content.

I worry this could snowball into a massive undertaking to break up mozorg into separate sites. I guess I mostly want to know what exactly distinguishes "MoCo" content from "MoFo" content. I'm not a lawyer (obviously) but might it be sufficient to say that the Mozilla Corporation owns and manages these websites on behalf of the Foundation? So even if we publish "Foundation content" that web property still falls under MoCo ownership?

> The browser certificates seem to show ownership by Mozilla Corporation for www.mozilla.org but not blog.mozilla.org—it would be good to have blog.mozilla.org also registered so the security certificates show that it is owned by Mozilla Corporation.

This may be a job for webops. I'm not sure what cert is being used for the blog subdomain or how to get the address bar to show the organization name. For what it's worth, I'm not sure any of our many subdomains display the org name in the address bar (e.g. support.mozilla.org, careers.mozilla.org, foundation.mozilla.org, developer.mozilla.org, and so on).

[...]

PR #6320, opened 2018-10-19
6306 foundation legal updates

craigcook:
## Description
Adds a notice to the global footer to clarify the Mozilla Foundation is a not-for-profit parent of the Mozilla Corporation. Also updates home page copy to clarify the Foundation is a not-for-profit.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/6306

## Testing
https://bedrock-demo-craigcook.oregon-b.moz.works/en-US/


bedrock/base/templates/includes/protocol/site-footer.html
```diff
@@ -66,6 +66,11 @@ <h5>
             <li><a href="{{ url('privacy.notices.websites') }}#cookies" data-link-type="footer" data-link-name="Cookies">{{ _('Cookies') }}</a></li>
             <li><a href="{{ url('legal.index') }}" data-link-type="footer" data-link-name="Legal">{{ _('Legal') }}</a></li>
           </ul>
+        {% if LANG.startswith('en-') %}
+          <p class="mzp-c-footer-license">
+            {{ _('Visit Mozilla Corporation’s not-for-profit parent, the <a href="%s" data-link-type="footer" data-link-name="Mozilla Foundation">Mozilla Foundation</a>.'|format('https://foundation.mozilla.org')) }}
```
stephaniehobson: I think `data-link-type` and `data-link-name` should be outside of the translation block. We want to be able to measure the performance of the link using the same query no matter which language we're looking at.

alexgibson:
@peiying2 - can we please add the new footer string here to `main.lang`, and we can remove the `en` conditional later once translations have caught up? Thanks

PR #6359, opened 2018-10-24
Truncated strings in Content Blocking tour for some locales (Fixes #6358)

### LINKED ISSUE #6358: Truncated strings in Content Blocking tour for some locales
The Content Blocking tour features some strings that have [curly double quotes in the HTML](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/tracking-protection-tour/variation-1.html#L23). Some locales have translated these as standard (non-curly) quotes, which is resulting in truncated strings in the tour, due to malformed HTML in the template.

screen shot 2018-10-24 at 09 33 24

The string above should read `Hvis blokering af indhold forhindrer dig i at anvende et websted, du stoler på, så kan du vælge "Deaktiver blokering midlertidigt" i dette panel.`

I'm trying to work out if there's an easy fix for this in Jinja

alexgibson:
Fixes #6358

@pmac this is the only filter I could find that seems to fix the issue. Curious if this is ok, or if there's a better way?

You can test this URL in a **private browsing window** to see the result (step 3 should show the string in question). You'll also need to whitelist your local dev server for [UITour](https://bedrock.readthedocs.io/en/latest/uitour.html#local-development)

`/da/firefox/63.0/tracking-protection/start/?step=1&variation=1`

### DIFF
--- a/bedrock/firefox/templates/firefox/tracking-protection-tour/index.html
+++ b/bedrock/firefox/templates/firefox/tracking-protection-tour/index.html
@@ -30,20 +30,20 @@ data-panel1-button="{{ _('Next') }}"
 data-panel3-title="{{ _('Want to make changes?') }}"
 
 {% if l10n_has_tag('tp_tour_firefox_201809') %}
-  data-panel3-text="{{ _('It’s easy to turn off Tracking Protection for the website you’re on by clicking “Disable For This Session.”') }}"
+  data-panel3-text="{{ _('It’s easy to turn off Tracking Protection for the website you’re on by clicking “Disable For This Session.”')|forceescape }}"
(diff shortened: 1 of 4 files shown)


pmac:
If `forceescape` is working then great. I think you could also have used the [escape filter](http://jinja.pocoo.org/docs/2.10/templates/#escape) though I'm not exactly clear on the difference. I know the latter is more common and only escapes the few common issues (mainly `&, <, >, ‘, ”`).

alexgibson:
The `escape` filter is the first thing I tried, but it seemed to make no difference and I couldn't work out why. `forceescape` seems to do the trick though.

Issue #6356, opened 2018-10-23
MOSS Revisions


ejregithub:
[Change request doc](https://docs.google.com/document/d/1mZbzUryAZ-p-ZnbIZ7wo1bJWyyszYoFS8vjBqWosmiI/edit)

[Change requests 'visualized'](https://docs.google.com/presentation/d/1g4nXSjmhnHEgY98pF7UoV3KkdDOjKltUDEzWbO5WjLI/edit#slide=id.g439cb43d6c_0_37)

ejregithub:
Stakeholder approved on 10/24

PR #6363, opened 2018-10-24
MOSS revisions (#6356)

amychurchwell:
## Description
Copy changes, add icons.

## Issue / Bugzilla link
#6356

## Testing
https://bedrock-demo-achurchwell.oregon-b.moz.works/en-US/moss/

### DIFF
--- a/bedrock/mozorg/templates/mozorg/moss/index-092018.html
+++ b/bedrock/mozorg/templates/mozorg/moss/index-092018.html
@@ -25,8 +25,8 @@
 
   <p class="moss-copy">{{ _('MOSS awards are available in multiple tracks:') }}</p>
 
-  <ul class="moss-list moss-list-thirds">
-    <li>
+  <ul class="moss-list moss-list-thirds" id="moss-focus-list">
+    <li id="foundational-technology">
       <h3 class="moss-subhead">{{ _('Track I: Foundational Technology') }}</h3>
       <p class="moss-copy">{{ _('The Foundational Technology track supports open source projects that Mozilla relies on, either as an embedded part of our products or as part of our everyday work.') }}</p>
       <div class="moss-ctas">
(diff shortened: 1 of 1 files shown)


bedrock/mozorg/templates/mozorg/moss/index-092018.html
```diff
@@ -35,19 +35,19 @@ <h3 class="moss-subhead">{{ _('Track I: Foundational Technology') }}</h3>
       </div>
     </li>
 
-    <li>
-      <h3 class="moss-subhead">{{ _('Track II: Mission Partners') }}</h3>
+    <li id="mission-partners">
+      <h3 class="moss-subhead">{{ _('Tracks II & IV: Global Mission Partners') }}</h3>
```
alexgibson: I'm not sure if any locales have translated this page yet, but this template is behind an l10n activation tag in the view: https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/views.py#L216

If this page is being actively localized, all string changes will need to be behind an l10n conditional tag. If not, we're ok to make these changes.

Perhaps @peiying2 can advise here on translation status.

bedrock/mozorg/templates/mozorg/moss/index-092018.html
```diff
@@ -325,6 +325,22 @@ <h2 class="moss-section-heading">{{ _('Frequently Asked Questions') }}</h2>
       </p>
     </dd>
 
+    {% if l10n_has_tag('moss_faq_201810') %}
```
alexgibson: Depending on @peiying2's answer, if the string changes above are ok, there may be no need for an l10n conditional tag here.

ejregithub: The page was being localized, but done so *only* to Hindi and not actively (and hits to the page in Hindi over the past year are anemic).  We're okay to go w/o l10n tags

alexgibson: @ejregithub I think we should just check with @peiying2 here before assuming anyting.

ejregithub: @alexgibson - I found Lindsay's (owner of the MOSS content) response:  "We discussed this today, and we feel like (given the low numbers) we don't need the localization right now."

She also noted the want to get the content live for Mozfest, though that may've passed considering the timezone diffs . . (Perfect timing before our MozFest session tomorrow- thank you!)

alexgibson: Ok, if we can't wait to here from @peiying2 then @amychurchwell please remove the l10n tags.

amychurchwell: @alexgibson ok, will do!

Issue #6361, opened 2018-10-24
[10/29] Shield study for Private Browsing


ejregithub:
[Copy/idea doc](https://docs.google.com/document/d/16rDLIXSxhrnOZXTHLcSz2RR0aFg5hHii2mpCv0oos1g/edit?ts=5bcfadae)

Description

Success Criteria:
- [x] page is live and shown to shield study participants

Risks
- [ ] Timeline

Tasks
- ~~Copy~~ We're using their copy
- [x] Design
- [x] QA of creative (internal)
-  ~~QA of creative (stakeholder)~~ We're using their copy
- [x] Dev - Code page
- [x] QA of coded page (internal)
- [x] QA of coded page (stakeholder)

stephaniehobson:
Vince and I had a short follow up. Keeping in mind that there is a tight deadline and it's an experiment we decided a few things:

- We will use the copy provided (which is from SUMO)
- We will drop the keyboard shortcut explanation

Vince will:
- [ ] make images that are:
  - same aspect ratio of current images
  - match graphic treatment of current images
  - not include keyboard shortcuts

Stephanie will:
- [ ] add the experiment to the current page
- [ ] communicate with the Fx team to set a URL variable to trigger it
- [ ] add the 2 blocks continuing the same left/right pattern
- [ ] leave one of the marketing blurbs as the first thing on the page

vincejoyiv:
Image assets attached below.

![newprivatewindow_ 2x](https://user-images.githubusercontent.com/13005624/47530147-a3b6ed00-d85e-11e8-8942-021f3ce5e87b.jpg)
![openlinkpbm_ 2x](https://user-images.githubusercontent.com/13005624/47530148-a3b6ed00-d85e-11e8-94e5-a56577732c96.jpg)

stephaniehobson:
The PR is in.

It is up on demo here: https://bedrock-demo-shobson.oregon-b.moz.works/en-US/firefox/features/private-browsing/?v=a

With the deadline I think it is important to get it merged as long as there's nothing broken. If Porfirio has feedback we can address it after it's live.

RosanneScholl:
This is wonderful! Thank you so much for creating it on rather short notice. For the purposes of the research, is it possible to re-order the blocks so that the two instruction ones are first, and the value propositions are lower? That's the only change I think will be needed.

RosanneScholl:
[...]

PR #6364, opened 2018-10-26
#6361 Add instructions to privacy page

stephaniehobson:
** Deadline: Oct 29** Please push Friday if the review passes.

## Description
- Only for users arriving in an `en` locale with the `?v=a` query variable
- Add 2 more sections to features/private-browsing page to show users how to use the feature.
- Hide the download button

## Issue / Bugzilla link
mozilla/bedrock#6361

## Testing
- New squares show up in en with ?v=a 
- Download button does not appear in en with ?v=a 
- The page is unchanged under all other conditions


bedrock/firefox/templates/firefox/features/private-browsing.html
```diff
@@ -42,6 +51,26 @@ <h2>{{ _('Browse without a trace') }}</h2>
       </div>
     </section>
 
+    {% if (variation == 'a') and (request.locale in ['en-US', 'en-GB', 'en-CA', 'en-AU']) %}
```
alexgibson: Not a blocker, but you can specify locales in the `VariationTemplateView`, so you don't need to query locales directly in the template. I think for a short experimnet this is fine though.

bedrock/firefox/templates/firefox/features/private-browsing.html
```diff
@@ -42,6 +51,26 @@ <h2>{{ _('Browse without a trace') }}</h2>
...
+    <section class="key-feature private-browsing-feature-new">
+      <div class="key-feature-content">
+        <h2>Open a new, blank Private Window</h2>
+        <p>
+          Click the menu button <img src="{{ static('img/firefox/election/icon-menu.svg') }}" alt="Open menu" height="12" width="14"> and then click New Private Window.
```
alexgibson: Also not a blocker, just an idea - we could go one step further here in the future perhaps, by using UITour to directly open the menu and highlighting the Private Browsing icon, as opposed to just describing what to do. We used to do this on the old `/firefox/private-browsing/` page as the main CTA.

Issue #6372, opened 2018-10-26
[Email Pref Center] Newsletter Title & description needs updating "Mozillian Campaigns"


ejregithub:
Description:  It's best to read the bug, in which pmac has provided comments https://bugzilla.mozilla.org/show_bug.cgi?id=1501842

PR #6443, opened 2018-11-06
[fix bug 1501842] Update Mozillian Campaigns newsletter title and desc

alexgibson:
https://bugzilla.mozilla.org/show_bug.cgi?id=1501842
#6372


alexgibson:
@pmac r? I tried removing the entry here as you suggested in the bug, but that didn't seem to have the desired effect. Seeing as the newsletter is English only, I figured updating the string directly would still be ok here?

pmac:
It should fall back to the description in the basket JSON feed, but you'd have to have an up-to-date database containing the newsletter data update for it to show the right thing. This also works just fine though.

PR #6449, opened 2018-11-07
Add Mozilla Labs newsletter form to /technology (Fixes #6400)

### LINKED ISSUE #6400: Update newsletter CTA on /technology/
MVP: 
- make the newsletter sign up field on https://www.mozilla.org/en-US/technology/ point to the email list value 'mozilla-technology'
- replace copy in the newsletter module 
from:
Get the Mozilla newsletter for our latest tech news and more.
to:
Get the Mozilla Labs newsletter for updates on our latest tech and product innovations
------------------
post-MVP:
- replace image the newsletter module with design similar [to this](https://drive.google.com/drive/folders/17__1UBXPcfApV1hurWh9fwMx0IXrRUL0?usp=sharing) that reflects the other creative in the page

alexgibson:
## Description
- Updates newsletter form on `/technology` to subscribe to Mozilla Labs newsletter (English locales only).

## Issue / Bugzilla link
#6400
https://bugzilla.mozilla.org/show_bug.cgi?id=1503021

## Testing
- [ ] Subscription should work as expected
- [ ] Non-en locales stil get regular newsletter form as expected.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/technology.html
+++ b/bedrock/mozorg/templates/mozorg/technology.html
@@ -149,10 +149,22 @@
 
   <aside id="newsletter-subscribe">
     <div class="content">
-      {{ email_newsletter_form(newsletters='mozilla-foundation',
-         title=_('What’s next?'),
-         subtitle=_('Get the Mozilla newsletter for our latest tech news and more.'), button_class='button-light')
-      }}
+      {% if LANG.startswith('en-') %}
+        {{ email_newsletter_form(
+          newsletters='mozilla-technology',
+          title=_('What’s next?'),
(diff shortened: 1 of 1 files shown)


stephaniehobson:
I'm seeing the language drop down with the new newsletter:

screen shot 2018-11-07 at 12 11 51 pm

I know it's English only and it looks like the macro has an option to suppress the language drop down that's not working? Or maybe I'm misunderstanding what `include_language=False` does.

alexgibson:
@stephaniehobson I think what you're seeing there is the country selector, which is separate from the language drop down. If you check a non-English version of the page, you should see the additional select input.

stephaniehobson:
You're right, I'm confusing the two.

R+ 📰

PR #6469, opened 2018-11-13
Update/remove the LXR redirects

alexgibson:
Replaces #5675


alexgibson:
@pmac r?

Side querstion: is there something different about the redirect tests in `map_301.py`? Things are passing here in CircleCI, but I can't get the same result when running locally in docker. It's almost like my changes don't get picked up. I tried running `make build`, but then `make test` fails afterward with an error about not finding `conftest.py`.

pmac:
@alexgibson how are you running the redirect tests? It can be all about your local settings and the test command. The command run on circle is:

```
py.test -r a -m smoke tests/redirects
```

alexgibson:
I tried again with a fresh `.env`, but the tests here still fail for me locally when running `make test`. It almost seems like the local changes to the redirects aren't being picked up.

I tried running `make build` before running `make test`,  but I seem to get the following error just before the redirect tests start:

```
Traceback (most recent call last):
  File "/venv/lib/python2.7/site-packages/_pytest/config.py", line 338, in _importconftest
    mod = conftestpath.pyimport()
  File "/venv/lib/python2.7/site-packages/py/_path/local.py", line 668, in pyimport
    raise self.ImportMismatchError(modname, modfile, self)
ImportMismatchError: ('conftest', '/Users/agibson/Documents/GitHub/bedrock/tests/conftest.py', local('/app/tests/conftest.py'))
ERROR: could not load /app/tests/conftest.py
```

The only way to fix it seems to be if I run `make pull` again, but of course that wipes out my `make build`.

pmac:
@alexgibson AH! Found it. This line is including the `test` directory in the image, but the directory name is `tests` 😭 Fix that typo and all should be well I think.

https://github.com/mozilla/bedrock/blob/aede772ee2d887d4d8d2d4a435b4e336e988c50f/docker-compose.yml#L58

pmac:
Do you want to include that fix in this PR or shall I file a separate one?

alexgibson:
@pmac I added a commit, let me know if I got that right

alexgibson:
Fwiw, I still can't get things to work locally it seems. I get the same error `ERROR: could not load /app/tests/conftest.py` when running `make test`.

alexgibson:
@pmac so if you pull this branch locally and then run `make test` you don't get the error above?

alexgibson:
Ok, I seem to have fixed it by running `make clean` and then `make test` works as expected. No idea what had gone wrong :/

pmac:
If `make clean` worked maybe it was an old `.pyc` file sitting around. Glad you got it working regardless!

PR #6477, opened 2018-11-14
[fix #6473] Hide Fx download button in nav on Dev Edition product page

### LINKED ISSUE #6473: Hide Fx download button in new nav on Dev Edition product page
The Dev Edition product page at https://www.mozilla.org/firefox/developer/ uses some CSS to hide the standard download button in the global nav, but the new global nav has different classes so the new button isn't hidden. We just need to add the appropriate class name to the page style sheet.

craigcook:
## Description
The Dev Edition product page at https://www.mozilla.org/firefox/developer/ uses some CSS to hide the standard download button in the global nav, but the new global nav has different classes so the new button isn't hidden. We just need to add the appropriate class name to the page style sheet. 

## Issue / Bugzilla link
#6473

### DIFF
--- a/media/css/firefox/developer/developer.scss
+++ b/media/css/firefox/developer/developer.scss
@@ -5,6 +5,7 @@
 @import '../../pebbles/includes/lib';
 @import '../../pebbles/components/newsletter';
 @import '../../hubs/sub-nav';
+@import '../../hubs/nav-download-button-remove';
 
 @import 'includes/lib';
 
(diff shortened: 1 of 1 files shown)


media/css/firefox/developer/developer.scss
```diff
@@ -87,7 +87,8 @@
     // Hide it for now pending a better solution.
     #global-nav-download-firefox,
     #sub-nav-download-firefox,
-    #nav-download-firefox {
+    #nav-download-firefox,
```
alexgibson: There's a [handy include here](https://github.com/mozilla/bedrock/blob/master/media/css/hubs/_nav-download-button-remove.scss) which means we don't have to rewrite this for every page.

craigcook: Nice... I suspect that didn't exist when I first built this page. I'll update.

PR #6483, opened 2018-11-16
[fix #XXXX] Add Fx Concert Series & FxA OAuth relier.

jpetto:
## Description

Adds a page for the Firefox Concert Series as well as FxA OAuth relier functionality.


bedrock/mozorg/util.py
```diff
@@ -170,3 +181,49 @@ def get_tweets(account):
...
+    except:
+        email = None
+
+    return email
+
+
+def fxa_concert_rsvp(email):
+    data = {
+        'email': email,
+        'is_firefox': 'Y',  # or 'N'. for the view to figure out.
```
pmac: Need to figure this out via the request to the oauth view from the `user-agent` header and then pass it to this function.

bedrock/mozorg/util.py
```diff
@@ -170,3 +181,49 @@ def get_tweets(account):
...
+        'campaign_id': 'livenation-2018',  # or something. just here in case there's another of these some day.
```
pmac: I'm sure this is fine to leave as is, but to reiterate this is just for future-proofing so that we can differentiate this campaign from any future ones.

bedrock/mozorg/util.py
```diff
@@ -170,3 +181,49 @@ def get_tweets(account):
...
+        'api_key': settings.BASKET_API_KEY,
+    }
+
+    basket.request('post', 'fxa-concerts-rsvp', data=data)
```
pmac: Will need some error handling around this.

bedrock/mozorg/views.py
```diff
@@ -215,3 +220,56 @@ def moss_view(request):
...
+        print('state mismatch error')
+        error = True
+    else:
+        token = get_fxa_oauth_token(code)
+
+        if not token:
+            print('token error')
+            error = True
+        else:
+            email = get_fxa_profile_email(token)
```
pmac: Call to the `fxa_concert_rsvp` function will be in this area right?

jpetto: I've got the basket call commented out below (in the non `error` flow), but yes, basically. 😁

jpetto:
Closed in favor of #6489.

PR #6489, opened 2018-11-19
[fix #6488] Add FxA OAuth relier.

### LINKED ISSUE #6488: Add FxA OAuth relier functionality
Foundational work needed to finish #6476.

jpetto:
## Description

Adds FxA OAuth relier functionality. Also adds very basic templates for testing the flow. (All templates and corresponding URLs are behind a switch to test flow on stage, but wait until templates are finished for prod.)

## Testing

https://www-demo5.allizom.org/firefox/concerts/

The above OAuth flow should work (redirecting you to `/firefox/concerts/` with a cookie that will display a "Verified" message instead of the initial form/link) for all browsers, both signed in and out of FxA.


media/js/firefox/concerts.js
```diff
@@ -0,0 +1,84 @@
...
+        } else {
+            // generate a state token and put it in a cookie
+
+            // the suffix below is superfluous for now, but will help us identify
+            // different OAuth flows in the future
+            state = Math.random() + '_concert_q42018';
+
+            var d = new Date();
+            d.setTime(d.getTime() + (cookieDays * 24 * 60 * 60 * 1000));
+            Mozilla.Cookies.setItem(stateStorageKey, state, d.toUTCString(), '/');
```
jgmize: We should probably be using a [signed cookie](https://docs.djangoproject.com/en/1.8/topics/signing/) here, but it's not a blocker for this use case.

media/js/firefox/concerts.js
```diff
@@ -0,0 +1,84 @@
...
+        // get metrics flow stuff from FxA
+        fetchTokens(metricsFlowEndpoint);
+
+        // get tokens from FxA for analytics purposes
+        // copied from base/mozilla-fxa-form.js
+        function fetchTokens(destURL) {
+            // add required params to the token fetch request
+            destURL += '?utm_campaign=firefox-concert-series-q4-2018&utm_source=mozilla.org';
+
+            fetch(destURL).then(function(resp) {
```
jgmize: Ideally we'd use a circuit breaker here to handle a slow or unresponsive metrics endpoint.

jpetto: As the metrics data here is not critical, I've removed the `fetch` call being a blocker. The user can now submit the form whenever they are ready. If the `fetch` call completed beforehand, they'll have metrics data.

PR #6497, opened 2018-11-21
Remove legacy navigation (Fixes #6377)

### LINKED ISSUE #6377: Remove legacy navigation from bedrock
Once we're happy that the new Protocol navigation is here to stay, we need to remove both the switch and the code for the legacy nagivation from bedrock.

alexgibson:
## Description
- Removes legacy global navigation shown to English locales.

## Issue / Bugzilla link
#6377

## Testing
Demo: https://bedrock-demo-navigation.oregon-b.moz.works/en-US/

Sucessful test run: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/216/pipeline/


alexgibson:
Ok, I think this is ready. This cleans up a bunch of leftovers and also reduces the number of requests for every page, so would be great to get in.

media/js/base/protocol/init-navigation.js
```diff
@@ -3,26 +3,22 @@
...
     if (typeof Mzp === 'undefined' || typeof Mzp.Menu === 'undefined' || typeof Mzp.Navigation === 'undefined') {
         return;
     }
 
-    if (!Mzp.Menu.isSupported()) {
-        return;
-    }
-
-    var _mqWideNav = window.matchMedia('(min-width: 768px)');
+    var hasMediaQueries = typeof window.matchMedia !== 'undefined';
```
stephaniehobson: I was more comfortable with `Mzp.Menu.isSupported`. That way if the menu's requirements change the check in init is still valid.

bedrock/base/templates/includes/global-nav.html
```diff
@@ -1,295 +0,0 @@
...
-          <title>{{_('Menu')}}</title>
-            <rect class="p1" fill="#231f20" x="3" width="5" height="6"></rect>
-            <rect class="p2" fill="#231f20" x="3" y="9.5" width="5" height="6"></rect>
-            <rect class="p3" fill="#231f20" x="3" y="19" width="5" height="6"></rect>
-        </svg>
-      </button>
-
-      <h2 class="nav-logo">
-        <a href="{{ url('mozorg.home') }}">
-          {# Bug 1438302 Avoid duplicate content for en-CA and en-GB pages. #}
```
stephaniehobson: I don't see this in the new nav file. Not a blocker for removing this code but if @rraue thinks it is important from an SEO perspective we should migrate it

media/js/base/protocol/init-navigation.js
alexgibson: I realized that `Menu` calls `isSupported()` during `init()` anyway, so I think this was somewhat redundant.

bedrock/base/templates/includes/global-nav.html
alexgibson: Oh good call, I'll update to make sure we don't lose this 👍

alexgibson:
> Not going to hit the merge button yet in case one of the others tagged for review wants to take a peek. Let's merge and push Wednesday morning if no one else has commented by then.

@stephaniehobson If possible, it would be great to try and push this sooner rather than later. I don't forsee any issues and this should be pretty safe, but getting this pushed out today would give us a whole week to let things settle before all-hands?

bedrock/base/templates/includes/global-nav.html
alexgibson: Updated to make sure we preserve this in the new navigation.

stephaniehobson:
Needs a rebase, then sure, we can get it out today.

alexgibson:
@stephaniehobson rebased, thank you!

Issue #6465, opened 2018-11-12
[11/27] Annual donation fundraising banner


ejregithub:
Description: MoFo donations - fundraising banner on mozilla.org

djst:
Has this merged or will this merge today at 9am PST?

Also, related issue #6518

craigcook:
It merged in 5ca985f03d7312851361ca16c6988a4eaef0760e and is now live in production.

PR #6501, opened 2018-11-21
[#6465] Add 2018 Fundraising banner to home page

craigcook:
## Description
Adds a fundraising banner to the home page (both the new English one and the old localized one). It's behind the switch `fundraising-eoy2018`. 

This is resurrected from last year's banner and uses all the same strings using the l10n tag `home-fundraiser-EOY2017` so any locales that still have that tag in their home.lang should get the localized banner with no additional effort.

## Issue / Bugzilla link
#6465 

## Testing
https://bedrock-demo-craigcook.oregon-b.moz.works

- [x] Dismissing the banner sets a three day cookie so it's hidden on subsequent visits. Delete the `fundraiser-eoy2018` cookie to get the banner back.
- [x] Test for responsiveness as usual. 
- [x] Test RTL (it's translated in Arabic and Farsi).
- [x] Test that the form submits to donate.mozilla.org and that the selected values are passed through to the pre-fill the form on the other side. It should have the correct amount, the once/recurring option, and the currency (we assume currency based on language, not geolocation, but users can change currency on donate.m.o so it's just a suggestion. Also note that not all the locales we have on mozorg are supported by donate.m.o so some may fall back to en-US and US$ when you get there.)


bedrock/mozorg/templates/mozorg/home/includes/fundraiser-eoy2018.html
```diff
@@ -0,0 +1,69 @@
...
+<aside class="c-fundraiser" id="fundraiser">
+  <button type="button" class="c-fundraiser-close" id="fundraiser-close"><span>{{ _('Close') }}</span></button>
+  <div class="content mzp-l-content">
+    <h1 class="c-fundraiser-title">
+      {{ _('We all love the web. Join Mozilla in defending it.') }}
+    </h1>
+
+    <div class="c-fundraiser-content">
+      <p>
+      {% if LANG == 'en-US' %}
```
alexgibson: Could we show this string for all English locales?
```suggestion
      {% if LANG.startswith('en-') %}
```

bedrock/mozorg/views.py
```diff
@@ -194,15 +193,23 @@ class DeveloperView(BlogPostsView):
...
 
     if locale.startswith('en-'):
         template_name = 'mozorg/home/home-en.html'
 
-        ctx['pocket_articles'] = PocketArticle.objects.all()[:4]
     else:
         template_name = 'mozorg/home/home.html'
 
-    return l10n_utils.render(request, template_name, ctx)
+    return l10n_utils.render(request, template_name, {
```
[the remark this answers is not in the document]
alexgibson: I think this should be fine and do no harm 👍

media/css/mozorg/home/fundraiser2018.scss
```diff
@@ -0,0 +1,122 @@
...
+        &:first-child {
+            @include bidi(((margin-right, $spacing-lg, 0),(margin-left, 0, $spacing-lg)));
+        }
+    }
+}
+
+.c-fundraiser-options {
+    label {
+        display: inline-block;
+        margin: 0 $spacing-sm $spacing-md 0;
```
alexgibson: The margin values here could do with a `bidi` mixin?

screen shot 2018-11-22 at 16 07 33

media/css/mozorg/home/fundraiser2018.scss
```diff
@@ -0,0 +1,122 @@
...
+        position: relative;
+
+        &:first-child {
+            margin-left: 0;
```
alexgibson: `bidi`?

media/css/mozorg/home/fundraiser2018.scss
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+
+    @media #{$mq-md} {
+        display: inline;
+        margin: 0 1em 0 0;
```
alexgibson: `bidi`?

media/css/mozorg/home/fundraiser2018.scss
```diff
@@ -0,0 +1,122 @@
...
+    }
+}
+
+// Close button
+.c-fundraiser-close {
+    @include bidi(((right, $spacing-sm, auto), (left, auto, $spacing-sm)));
+    background: transparent url('/media/img/notification/close.svg') center center no-repeat;
+    background-size: 18px 18px;
```
alexgibson: Nit: `background-size` mixin?

media/css/mozorg/home/fundraiser2018.scss
```diff
@@ -0,0 +1,122 @@
...
+    border-radius: 50%;
+    border: none;
+    display: none;
+    height: 30px;
+    padding: 6px;
```
alexgibson: `$spacing-sm`?

media/css/mozorg/home/fundraiser2018.scss
```diff
@@ -0,0 +1,122 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+$font-path: '/media/fonts';
+$image-path: '/media/protocol/img';
+
+@import '../../../protocol/css/includes/lib';
+@import '../../../protocol/css/components/button';
```
alexgibson: The `button` component is now included in the base protocol style sheet for bedrock, so there should be no need to import it again here.

media/css/mozorg/home/fundraiser2018.scss
```diff
@@ -0,0 +1,122 @@
...
+    @media #{$mq-md} {
+        display: inline;
+        margin: 0 1em 0 0;
+    }
+}
+
+// Close button
+.c-fundraiser-close {
+    @include bidi(((right, $spacing-sm, auto), (left, auto, $spacing-sm)));
+    background: transparent url('/media/img/notification/close.svg') center center no-repeat;
```
alexgibson: Not a blocker, but could we use the Protocol close icon here (same one as we use in the modal)?

bedrock/mozorg/templates/mozorg/home/includes/fundraiser-eoy2018.html
```diff
@@ -0,0 +1,69 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+<aside class="c-fundraiser" id="fundraiser">
+  <button type="button" class="c-fundraiser-close" id="fundraiser-close"><span>{{ _('Close') }}</span></button>
+  <div class="content mzp-l-content">
+    <h1 class="c-fundraiser-title">
```
alexgibson: This is introducing a second `h1` on the page, would it be better an `h2`?

media/js/mozorg/home/fundraiser-eoy2018.js
```diff
@@ -0,0 +1,33 @@
+(function($, Mozilla) {
+    'use strict';
+
+    var $fundraiser = $('#fundraiser');
```
alexgibson: Nit: not a blocker, but there's a couple of lines of jQuery here that could pretty easily be vanilla JS with just a couple of small changes :)

alexgibson:
Depends on https://github.com/mozilla/bedrock/pull/6505

media/css/mozorg/home/fundraiser2018.scss
craigcook: I think it's needed for the old Pebbles-based home page. So it's a bit redundant to include here, but probably less redundant than making a whole separate Pebbles version of the style sheet. I do cringe a bit mixing Protocol and Pebbles on the same page but at least this banner will be short-lived.

alexgibson: Ah I see, that makes sense. The only downside to this is it leaves a fair chunk of duplicate CSS in the English homepage bundle. Looking at the code, from what I can see this is only for a single button. Would hard-coding those styles in this (temporary) stylesheet be worth it to avoid the double-import?

media/js/mozorg/home/fundraiser-eoy2018.js
```diff
@@ -0,0 +1,34 @@
...
+            var d;
+
+            fundraiser.parentNode.removeChild(fundraiser);
+
+            if (cookiesOK) {
+                d = new Date();
+                d.setTime(d.getTime() + (cookieDurationDays * 24 * 60 * 60 * 1000)); // 2 day expiration
+                Mozilla.Cookies.setItem(storageKey, true, d.toUTCString(), '/');
+            }
+        });
```
alexgibson: NIt: add `false` as the last param to `addEventListener` for better cross-browser compat.

craigcook: Updated and squashed.

WillatMozFdn:
Requesting copy change to this banner for EN, to keep things fresh and avoid burnout from repeat of the same copy.  New copy below; has been approved through regular process by relevant parties (Abigail - Legal; Marshall - Compliance; Chris - Policy; Lotta - MoFo Comms; Sara - MoFo Advocacy.) 

This is a bit longer than current copy; let me know if this presents problems and I can edit down some.

-----------
The future of the internet is at stake, with new threats to our online privacy and security virtually every day. The not-for-profit Mozilla Foundation fights to save a healthy internet, with projects like our *Privacy Not Included buyer's guide. Your donations make this work possible, **so please chip in today.**
-------------

Issue #6339, opened 2018-10-22
Accounts Benefits Page


bensternthal:
| Estimates |  |
| ------------- | ------------- | 
| TL | 5|
| PM | 12|
| ENG | 60|
| SEO | 0|
| ANA | 8|
| DES | 60|
| COPY | 40|

Description, good enough that reading the card 1 month from now, you still know what it means.

---

#### :yellow_heart: Success Criteria :yellow_heart:

- [x] We've determined which page it replaces/kills
- [x] The other page (non-winner) is redirected here
- [x] A user understands what product services work w. an account
- [x] Launch to en-US
- [x] Page must be localizable at launch
- [x] Visuals (not animated)
- [x] Visuals that explain benefits clearly to English and non-English speaking users
- [ ] Build in Protocol w. new components added
- [ ] Protocol code/design review
- [ ] Conditional CTA based on audience - download or signup/in
- [ ] User understands what an account is
- [ ] Maintainable addition and removal of benefits
- [ ] Design and coded page must be approved by the team and the Approver
- [ ] Analytics QA
- [ ] Legal Approval
- [ ] Increased proportion of traffic to the page
- [ ] a user understands why an account is beneficial
- [ ] Uses Raphael's SEO keyword recommendations
- [x] Links out to each product/services
- [ ] Animation similar to the device sync animation on the old sync page)
- [x] Animations / visuals do not include text

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] What will trigger updates?

#### Tasks
- [ ] UX - @dzingeek 
- [ ] Write copy - @clay-chaffin 
- [ ] Design (create design) - @mrkwvr 
- [ ] Code - @amychurchwell 
- [ ] QA testing of design (the team)
- [ ] QA testing of design (Approver)
- [ ] QA testing of coded page on demo (the team)
- [ ] QA testing by Analytics
- [ ] QA testing of coded page on demo (Approver)

ejregithub:
Blocked by lack of info/data/research in https://github.com/mozilla/bedrock/issues/6344

amychurchwell:
#6345 -- **final URL**: /firefox/benefits

Redirects - https://github.com/mozilla/bedrock/issues/6345#issuecomment-435737625

dzingeek:
Latest designs can be found here. 

[...]

PR #6503, opened 2018-11-22
[#6339] New Firefox Accounts page

craigcook:
## Description
Redesigns the page at /firefox/accounts/ and removes the old Accounts Benefits page, as well as the Sync and Send Tabs feature pages.

## Issue / Bugzilla link
#6339 

## Testing
https://www-demo3.allizom.org/firefox/accounts/

- [x] If on desktop, product cta's will open modal with QR code & app store buttons.
- [x] If on mobile, product cta's will be app store buttons.

features user flow secondary

**NOT DOING CUSTOM DOWNLOAD/THANKS (ignore on chart)**

- [x] if using browser other than firefox, user gets download firefox button.
- [x] if using firefox, user gets sign up for fxa cta. Bottom banner is form, hero is button.
- [x] if using firefox & signed into firefox accounts, user gets firefox mobile app ctas. (Modal with QR code for hero for desktop, app store buttons for mobile & bottom banner.)

features user flow main


craigcook:
@peiying2 Here's a fresh PR with strings in a new template at `firefox/accounts-2018.html` so it can get its own lang file, but slower locales can fall back to the old page without breaking anything.

peiying2:
@craigcook Thanks for creating the new template. Let's remove the l10n label in #6499.

bedrock/privacy/templates/privacy/faq.html
```diff
@@ -206,8 +206,8 @@ <h2>
             {%- endtrans -%}
           </p>
           <p>
-            {%- trans -%}
-              You don't need an account to use Firefox.  Accounts are required to sync data across devices, but we only ask you for an email address.  We don't want to know things like your name, address, birthday and phone number.
+            {%- trans accounts=url('firefox.accounts') -%}
```
peiying2: @alexgibson @craigcook We need a tag to update this string. This page is activated for quite a few locales: https://l10n.mozilla-community.org/langchecker/?locale=all&website=0&file=privacy/faq.lang

flodolo:
On top of what Peiying already (correctly) pointed out, I'm not sure who had the idea of creating a new GitHub label that nobody is monitoring. The L10N labels are used by the l10n team to track pending work, they shouldn't be changed, definitely not without consulting with them.

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,237 @@
...
+  <p>{{ _('Send an open tab on one device to all your others with a single tap. Much easier than texting or emailing yourself those links.') }}</p>
+  {% endcall %}
+
+  {% call feature_card(
+    title=_('Remember your passwords in Lockbox'),
+    image_url='firefox/accounts/fxa-lockbox.png',
+    include_highres_image=True,
+    aspect_ratio='mzp-has-aspect-1-1',
+    class='mzp-l-card-feature-left-half lockbox-card',
+    link_cta='Get the Lockbox App',
```
flodolo: All these strings are hard-coded in English

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,237 @@
...
+            </li>
+            <li>
+              <a href="{{ firefox_ios_url(app_store_campaign) }}" data-link-type="download" data-download-os="iOS" data-download-location="{{ button_location }}" id="{{ ios_id }}">
+                <img src="{{ l10n_img('firefox/ios/btn-app-store.svg') }}" alt="{{ _('Download on the App Store') }}" width="152" height="45">
+              </a>
+            </li>
+          </ul>
+        </div>
+        <div class="show-fxa-supported-signed-out">
+          {{ fxa_email_form(entrypoint='accounts-page', utm_source='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'accounts-page', 'medium': 'referral'}, button_class='button button-blue', button_text='Create an Account') }}
```
flodolo: hard-coded button label.

bedrock/privacy/templates/privacy/faq.html
craigcook: Sorry I forgot this. Added the tag `firefox-accounts-112018`

craigcook:
Updated with a tag for the privacy FAQ and enclosing the strings I missed. Thanks for the review.

I wasn't sure of the difference between the general "L10N" label and "L10N - Ready" and wasn't sure which ones are used by us and which by the l10n team, sorry for the confusion.

flodolo:
It would really help to not force-push once reviewed has started.

alexgibson:
Side note: removing these old pages is likely going to break a few functional tests, so we should push this to the `run-integration-tests` branch before merging.

amychurchwell:
**Testing**
- [ ] If on desktop, product cta's will open modal with QR code & app store buttons.
- [ ] If on mobile, product cta's will be app store buttons.

features user flow secondary

**NOT DOING CUSTOM DOWNLOAD/THANKS (ignore on chart)**

- [ ] if using browser other than firefox, user gets download firefox button.
- [ ] if using firefox, user gets sign up for fxa cta. Bottom banner is form, hero is button.
- [ ] if using firefox & signed into firefox accounts, user gets firefox mobile app ctas. (Modal with QR code for hero for desktop, app store buttons for mobile & bottom banner.)

features user flow main

amychurchwell:
This PR is ready for code review. Still waiting on analytics feedback, design feedback has been implemented and is complete.

flodolo:
It also looks like this page is removing several old pages, they should be marked as obsolete on our side

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+
+{% block content %}
+
+<section class="mzp-c-hero mzp-has-image mzp-t-firefox mzp-t-product-firefox">
+  <div class="mzp-l-content">
+    <div class="mzp-c-hero-body">
+      <h1 class="mzp-c-hero-title">{{ _('One log-in. Power and privacy everywhere.') }}</h1>
+      <div id="hero-cta" class="mzp-c-hero-cta">
+        <div class="show-fxa-supported-signed-out">
+          <a href="https://accounts.firefox.com/signup?service=sync&context=fx_desktop_v3&entrypoint=mozilla.org-button-light&utm_content=%2Ffirefox%2Faccounts%2F&utm_campaign=Create an Account" class="button account-benefits" data-button-name="Create account" data-link-type="button" data-cta-position="primary cta" id="features-hero-account">
```
stephaniehobson: Can we use the `fxa_create_account_button` macro here? It's possible to configure the button label

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+  {% endcall %}
+
+  {% call feature_card(
+    title=_('Put quality content in your Pocket'),
+    image_url='firefox/accounts/fxa-pocket.png',
+    include_highres_image=True,
+    aspect_ratio='mzp-has-aspect-1-1',
+    class='mzp-l-card-feature-left-half pocket-card',
+    link_cta=_('Get the Pocket App'),
+    link_url='#'
```
stephaniehobson: Missing URL

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+              <a href="{{ firefox_ios_url(app_store_campaign) }}" data-link-type="download" data-download-os="iOS" data-download-location="{{ button_location }}" id="{{ ios_id }}">
+                <img src="{{ l10n_img('firefox/ios/btn-app-store.svg') }}" alt="{{ _('Download on the App Store') }}" width="152" height="45">
+              </a>
+            </li>
+          </ul>
+        </div>
+        <div class="show-fxa-supported-signed-out">
+          {{ fxa_email_form(entrypoint='accounts-page', utm_source='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'accounts-page', 'medium': 'referral'}, button_class='button button-blue', button_text=_('Create an Account')) }}
+          <div id="fxa-sign-in">
+            <p>{{_('Already have an account?')}} <a href="#">{{ _('Sign In') }}</a></p>
```
stephaniehobson: Missing URL

bedrock/privacy/templates/privacy/faq.html
```diff
@@ -206,9 +206,15 @@ <h2>
...
           </p>
           <p>
+          {% if l10n_has_tag('firefox-accounts-112018') %}
+            {%- trans accounts=url('firefox.accounts') -%}
+              You don't need an account to use Firefox.  <a href="{{ accounts }}">Accounts</a> are required to sync data across devices, but we only ask you for an email address.  We don't want to know things like your name, address, birthday and phone number.
+            {%- endtrans -%}
+          {% else %}
             {%- trans -%}
-              You don't need an account to use Firefox.  Accounts are required to sync data across devices, but we only ask you for an email address.  We don't want to know things like your name, address, birthday and phone number.
+            You don't need an account to use Firefox. Accounts are required to sync data across devices, but we only ask you for an email address.  We don't want to know things like your name, address, birthday and phone number.
```
stephaniehobson: Looks like some white space was removed here. @peiying2 does that effect the translation?

bedrock/firefox/redirects.py
```diff
@@ -592,4 +592,9 @@ def firefox_channel(*args, **kwargs):
 
     # issue 6186
     redirect(r'^vote/?', '/firefox/election/'),
+
+    # fxa
+    redirect(r'^firefox/accounts/features/?', '/firefox/accounts/'),
+    redirect(r'^firefox/features/sync/?', '/firefox/accounts/'),
+    redirect(r'^firefox/features/send-tabs/?', '/firefox/accounts/'),
```
stephaniehobson: 👍

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+  # License, v. 2.0. If a copy of the MPL was not distributed with this
+  # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% from "macros-protocol.html" import card, call_out, feature_card with context %}
```
stephaniehobson: Looks like we only need to import `feature_card`

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+{% from "macros.html" import fxa_email_form, google_play_button, send_to_device, fxa_create_account_button with context %}
```
stephaniehobson: `fxa_create_account_button` and `send_to_device` aren't used on this page.

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+          <img src="{{ l10n_img('firefox/ios/btn-app-store.svg') }}" alt="{{ _('Download on the App Store') }}" width="152" height="45">
+        </a>
+      </li>
+    </ul>
+  </section>
+
+  <section>
+    <ul class="mobile-download-buttons mobile-lockbox">
+      <li class="ios">
+        <a id="appStoreLink" href="https://app.adjust.com/kq8zhhz" data-link-type="download" data-download-os="iOS">
```
stephaniehobson: The id `appStoreLink` is used here and on line 161. ID's can only be used once on a  page.

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+            </a>
+          </li>
+        </ul>
+      </div>
+  </section>
+
+  <section>
+    <ul class="mobile-download-buttons mobile-pocket">
+      <li class="android">
+        {{ google_play_button(href='https://app.adjust.com/h5iycst', id='playStoreLink') }}
```
stephaniehobson: The id `playStoreLink` is used here an on line 101. IDs can only be used once on a page.

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+      </li>
+      <li class="ios">
+        <a id="appStoreLink" href="https://app.adjust.com/h5iycst" data-link-type="download" data-download-os="iOS">
```
stephaniehobson: The id `appStoreLink` is used here and on line 71  and 161. IDs can only be used once on a  page.

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+    link_url='#'
+  ) %}
+    <p>{{ _('Discover the web’s best content, and absorb it anytime – even offline – on any device. Pocket’s listen feature will even read any article aloud to you. All from the Firefox toolbar.') }}</p>
+
+    <ul class="mobile-download-buttons mobile-pocket">
+      <li class="android">
+        {{ google_play_button(href='https://app.adjust.com/h5iycst', id='playStoreLink') }}
+      </li>
+      <li class="ios">
+        <a id="appStoreLink" href="https://app.adjust.com/h5iycst" data-link-type="download" data-download-os="iOS">
```
stephaniehobson: The id `appStoreLink` is used here and on line 71  and 161. IDs can only be used once on a  page.

stephaniehobson:
Here's a photo of the bug in IE 11:

img_3557

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+        {{ google_play_button(href='https://play.google.com/store/apps/details?id=org.mozilla.testpilot.notes') }}
+      </li>
+    </ul>
+  {% endcall %}
+
+</div>
+
+<aside class="fxa-benefits mzp-u-modal-content">
+  <p class="mobile-title mobile-pocket">{{ _('Scan the QR Code to download the Pocket App on your mobile device.') }}</p>
+  <p class="mobile-title mobile-lockbox">{{ _('Scan the QR Code to download the Lockbox App on your mobile device.') }}</p>
```
stephaniehobson: Maybe update just to say Apple device?

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+  <p class="mobile-title mobile-notes">{{ _('Scan the QR Code to download the Notes App on your mobile device.') }}</p>
```
stephaniehobson: Maybe update just to say Android device?

bedrock/firefox/templates/firefox/accounts-2018.html
stephaniehobson: Hm. Normally we'd use a button for a pure JavaScript trigger like that. I'm not sure the macro will support that.  What do you think of pointing to https://getpocket.com/ for users with JS disabled?

amychurchwell: I wrote some javascript applying the modal click event listener to all card cta links. If I include a link url it will take all users to the address before the modal loads regardless of whether JS is disabled or not. Unsure how to reconcile. 🤔

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+        <div class="show-fxa-supported-signed-out">
+          {{ fxa_email_form(entrypoint='accounts-page', utm_source='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'accounts-page', 'medium': 'referral'}, button_class='button button-blue', button_text=_('Create an Account')) }}
+          <div id="fxa-sign-in">
+            <p>{{_('Already have an account?')}} <a href="#">{{ _('Sign In') }}</a></p>
+          </div>
+        </div>
+      </div>
+      <div class="mobile-ctas">
+        <span class="show-fxa-ios">
+          <a class="button button-hollow button-light" rel="external" href="https://support.mozilla.org/kb/sync-firefox-bookmarks-and-browsing-history-iOS">{{ _('Set up Sync for iOS') }}</a>
```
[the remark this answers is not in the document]
amychurchwell: :( do you think I should call this out specifically with the `html.ios` selector in .scss?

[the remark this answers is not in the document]
amychurchwell: I have `css/base/mozilla-fxa-state.scss` in the static bundle for `firefox_accounts_2018` :(
maybe I should go ahead and hide it myself under the `html.ios` css selector?

bedrock/firefox/templates/firefox/accounts.html
```diff
@@ -61,5 +61,5 @@ <h1>{{_('Sign up, sign in, stay in sync.')}}</h1>
 {% endblock %}
 
 {% block js %}
-  {{ js_bundle('firefox_accounts') }}
+  {{ js_bundle('firefox_accounts_2018') }}
```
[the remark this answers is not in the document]
amychurchwell: I wrote some custom JS for this page and the original bundle will be for the fallback page, unless I am not considering something? 🤔

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+  <div id="bottom-cta">
+    <div id="fxa-sign-up-form">
+      <div class="fxa-cta show-fxa-supported-signed-in show-fxa-supported-signed-out">
+        <div class="show-fxa-supported-signed-in">
+          <ul class="mobile-download-buttons" id="bottom-banner-mobile">
+            <li>
+              {{ google_play_button(id=android_id, extra_data_attributes={'download-location': button_location}) }}
+            </li>
+            <li>
+              <a href="{{ firefox_ios_url(app_store_campaign) }}" data-link-type="download" data-download-os="iOS" data-download-location="{{ button_location }}" id="{{ ios_id }}">
```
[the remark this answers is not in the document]
amychurchwell: is it significant what the unique ID is?

bedrock/firefox/templates/firefox/accounts-2018.html
stephaniehobson: [`e.preventDefault()`](https://developer.mozilla.org/en-US/docs/Web/API/Event/preventDefault)!

bedrock/firefox/templates/firefox/accounts-2018.html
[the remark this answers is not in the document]
stephaniehobson: I don't think so. It might end up in the analytics somewhere? Don't stress about it too much as long as they're different.

bedrock/firefox/templates/firefox/accounts.html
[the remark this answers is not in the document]
stephaniehobson: This is the fallback page, isn't it?

bedrock/firefox/templates/firefox/accounts-2018.html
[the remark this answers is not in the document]
stephaniehobson: I am pretty sure I had to write some custom JS to get the mobile stuff to work on the old benefits page because it's detected separately from the fxa stuff but now I can't *find* it. Don't do anything for now, I'll keep looking. Or let me know if you solve it first.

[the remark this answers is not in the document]
stephaniehobson: Ah! I think this is a display bug. I suspect the button is visible but off screen:

screen shot 2018-11-28 at 8 57 24 pm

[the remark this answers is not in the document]
stephaniehobson: Do you think you could swap this custom code out for a [Protocol compact call out](https://protocol.mozilla.org/patterns/organisms/call-out.html#call-out-compact)?

You can't use the macro because OMG custom code but maybe you can still use the component by hard coding the HTML onto the page.

I am thinking that might solve some of the cross browser layout stuff but if you think that's more work than just tweaking the custom component do what you think is best :)

stephaniehobson:
The bottom component on the page has some display problems at smaller screen sizes:

screen shot 2018-11-28 at 9 11 45 pm

stephaniehobson:
I will not block code review on this because you have followed the spec but I think it is worth going back to the stakeholder and asking if we can fix it (and if they want to delay for us to fix it, or not).

It is weird to me to see and the Play store button on my iPhone and the Apple store on Android. And there is no explanation for the ones that are only available in one store. On desktop it makes sense to show both links but on mobile we know what their mobile device is.

I would like some kind of "coming soon" (if that's accurate) indication for the unsupported platform.

img_3595

media/css/firefox/accounts-2018.scss
```diff
@@ -0,0 +1,216 @@
...
+}
+
+// Cards
+
+.mzp-c-card-feature-content {
+    .mzp-c-cta-link {
+        color: #0a84ff;
+        &:hover,
+        &:focus {
+            color: #0060df;
```
stephaniehobson: $color-link?

media/css/firefox/accounts-2018.scss
```diff
@@ -0,0 +1,216 @@
...
+            border-bottom: 2px solid #0060df;
```
stephaniehobson: $color-link?

media/css/firefox/accounts-2018.scss
```diff
@@ -0,0 +1,216 @@
...
+    h1, h2, h3, h4 {
+        font-weight: normal;
+    }
+}
+
+// Cards
+
+.mzp-c-card-feature-content {
+    .mzp-c-cta-link {
+        color: #0a84ff;
```
stephaniehobson: Vince just had me add these on the fffy page too :( Hopefully we can back-port them to Protocol. I don't know if you got specific direction from a designer for these states but you could also borrow from what Vince and I worked out: https://github.com/mozilla/bedrock/blob/6472526f9bf1b1c23fd49d212adc4a1f0ce022d4/media/css/firefox/fights-for-you.scss#L163-L177

media/css/firefox/accounts-2018.scss
```diff
@@ -0,0 +1,216 @@
...
+.mobile-pocket {
+    display: none;
+}
+
+.mzp-c-modal-inner {
+    @media #{$mq-lg} {
+        max-width: 500px;
+        margin: 0 auto;
+    }
+    @media #{$mq-md} {
```
stephaniehobson: Nit: Usually we put the smallest media queries higher up in the file than the bigger ones.

media/css/firefox/accounts-2018.scss
stephaniehobson: PS you can look up Protocol colour variables here: https://protocol.mozilla.org/fundamentals/tokens.html

media/css/firefox/accounts-2018.scss
```diff
@@ -0,0 +1,216 @@
...
+        padding: .5em;
+        text-align: center;
+
+        .mobile-title {
+            margin: 20px;
+        }
+
+        .mobile-download-buttons {
+            li {
+            display: inline-block;
```
stephaniehobson: Nit: needs an indent

media/css/firefox/accounts-2018.scss
```diff
@@ -0,0 +1,216 @@
...
+            color: #fff;
+        }
+    }
+}
+
+/* -------------------------------------------------------------------------- */
+// Conditional
+
+html.ios,
+html.android {
```
stephaniehobson: 👍

stephaniehobson:
✅ Images compressed ✨

media/css/firefox/accounts-2018.scss
```diff
@@ -0,0 +1,216 @@
...
+
+@import '../../protocol/css/includes/lib';
+@import '../../protocol/css/components/button';
+@import '../../protocol/css/components/billboard';
+@import '../../protocol/css/components/call-out';
+@import '../../protocol/css/components/feature-card';
+@import '../../protocol/css/components/hero';
+@import '../../protocol/css/templates/card-layout';
+@import '../../protocol/css/components/newsletter-form';
+@import '../../protocol/css/components/modal';
```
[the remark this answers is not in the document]
alexgibson: Correct,no need to import `button` again here.

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+      <div id="hero-cta" class="mzp-c-hero-cta">
+        <div class="show-fxa-supported-signed-out">
+          <a href="https://accounts.firefox.com/signup?service=sync&context=fx_desktop_v3&entrypoint=mozilla.org-button-light&utm_content=%2Ffirefox%2Faccounts%2F&utm_campaign=Create an Account" class="button account-benefits" data-button-name="Create account" data-link-type="button" data-cta-position="primary cta" id="features-hero-account">
+            {{ _('Create a Firefox Account') }}
+          </a>
+        </div>
+        <div class="show-fxa-supported-signed-in">
+          <button class="mzp-c-button" type="button" id="fx-modal">{{ _('Download the Firefox App') }}</button>
+        </div>
+        <div id="features-hero-download" class="show-fxa-not-fx">
```
alexgibson: You're using this ID to target the functional test, but it should actually be on the button itself and not the wrapper div.

```
{{ download_firefox(dom_id='features-hero-download', download_location='primary cta') }}
```

bedrock/firefox/urls.py
```diff
@@ -77,8 +72,6 @@
         VariationTemplateView.as_view(template_name='firefox/features/private-browsing.html',
                                       template_context_variations=['a']),
         name='firefox.features.private-browsing'),
-    page('firefox/features/send-tabs', 'firefox/features/send-tabs.html'),
-    page('firefox/features/sync', 'firefox/features/sync.html'),
```
alexgibson: It we're removing these page URLs, please make sure to update all references to pages that may be linking to them. For example, [this page](https://bedrock-integration-tests.oregon-b.moz.works/en-US/firefox/features/) is currently throwing a 500 because I'm guessing it contains a `{{ url('firefox.features.sync') }}` somewhere.

This is the cause of one of the test failures.

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,234 @@
...
+  </section>
+</aside>
+
+<div id="bottom-banner">
+  <div class="mzp-l-content">
+    <div class="banner-title">
+      <p>{{ _('You deserve peace of mind everywhere.') }}</p>
+    </div>
+  </div>
+  <div id="bottom-cta">
```
alexgibson: The content breaks out of the standard page container here because this `bottom-cta` div is not a child of the `mzp-l-conent` wrapper above

screenshot_2018-11-29 get a firefox account keep your data private safe and synced

alexgibson:
To fix the test failure:

- Remove [this line](https://github.com/mozilla/bedrock/blob/master/tests/functional/newsletter/test_newsletter_embed.py#L57), since the page no longer exists.
- Chenge the [URL here](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/features/index.html#L64) to `{{ url('firefox.accounts') }}`.

I can also see 8 other occurences of the old syn page URL in the codebase that will also need updating.

You also need to search and replace for `url('firefox.accounts-features')` and `url('firefox.features.send-tabs')`.

media/static-bundles.json
```diff
@@ -867,6 +853,14 @@
       ],
       "name": "firefox_accounts"
     },
+    {
+      "files": [
+        "css/base/mozilla-fxa-state.scss",
+        "css/base/mozilla-fxa-form.scss",
```
alexgibson: We should be careful when importing Pebbles CSS files into the same bundle as Protocol. This one imports the pebbles `/pebbles/includes/lib/` and `/pebbles/components/buttons/`. Both can have adverse side effects, such as overriding mixins, and in this case the download buttons that should be blue are showing up as green (which is something specific to non-protocol pages).

If you need these styles, I would create a Protcol equivalent and [put it here](https://github.com/mozilla/bedrock/tree/master/media/css/protocol/components)

dzingeek:
> I will not block code review on this because you have followed the spec but I think it is worth going back to the stakeholder and asking if we can fix it (and if they want to delay for us to fix it, or not).
> 
> It is weird to me to see and the Play store button on my iPhone and the Apple store on Android. And there is no explanation for the ones that are only available in one store. On desktop it makes sense to show both links but on mobile we know what their mobile device is.
> 
> I would like some kind of "coming soon" (if that's accurate) indication for the unsupported platform. And to not show the link to the wrong platform for them.
> 
> img_3595

@stephaniehobson @amychurchwell 
I worry about using "coming soon" since we have no idea if or when it might actually happen. I do think we could have it say something like, "Currently available for iOS only."

With that said, I am not sure we have time to get new copy in on this page? If we can't get new copy then I think we should just show the badges even if that means we show the app store badge for android users. It's not ideal but I think not showing anything is worse. 

What are you thoughts on this?

bedrock/firefox/templates/firefox/accounts-2018.html
amychurchwell: Going to wrap `<div id="bottom-cta">` in it's own `mzp-l-content` because I would like them to be side by side without more custom css. Is this bad practice? (@stephaniehobson)

bedrock/firefox/templates/firefox/accounts-2018.html
amychurchwell: do you think I should link to the app store for the others? Not all of them have an 'official website'

bedrock/firefox/templates/firefox/accounts-2018.html
stephaniehobson: I don't think it's bad practice but I'm not sure it will fix the problem :/

Both the banner title and the cta need to be inside an element that has a max-width set on it. `mzp-l-content` does have a max width but if the two are size by side instead of in the same one I think it will end up putting the containers side by side but still too wide.

I don't see a way to solve this bug without adding another wrapper in the HTML and moving the grid info on to that.

```
<div id="bottom-banner">
    <div class="mzp-l-content"><!-- this will keep stuff from getting too wide -->
        <div class="bottom-grid"> <!-- move display:grid and other grid properties here-->
            <div class="banner-title">
               ...
            </div>
            <div id="bottom-cta">
              ...
            </div>
        </div>
    </div>
</div>
```

stephaniehobson: Not the only solution of course ;)

stephaniehobson:
I agree not showing anything is worse than just showing the wrong one.

Another option might be showing both app store badges for all products and a greyed out/disabled badge for the non-available app store.

I didn't realize this page was being localized. I guess it's too late to add the string as a quick fix :(

Anyway, not a blocker.

dzingeek:
I think this is a great thing to bring up and have a conversation about even if we can't actually do anything now. We should keep this in mind for the future for sure. 

I do think that having a greyed out badge might feel weird. I have never seen it before and I worry it might feel broken. However, this could be something to test in the future.

bedrock/firefox/templates/firefox/accounts-2018.html
amychurchwell: Ok. Notes currently doesn't have an adjust link yet so I'll use `https://play.google.com/store/apps/details?id=org.mozilla.testpilot.notes`

amychurchwell: A bug(?)-- on click, mobile buttons appear behind modal. They disappear again when you close the modal. :( 

screen shot 2018-11-29 at 3 59 04 pm

bedrock/firefox/templates/firefox/accounts.html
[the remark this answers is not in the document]
amychurchwell: This is the new page, the previous page would be the fallback right?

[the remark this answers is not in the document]
craigcook: This line is in the file`accounts.html` which is the old fallback page, the new page is `accounts-2018.html`. `accounts.html` should use the old `firefox_accounts` JS bundle. 

Mostly we don't need to touch `accounts.html` at all, it remains frozen in time until we can eventually delete it.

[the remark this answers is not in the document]
craigcook: Alternatively, we could rename the old page to something like `accounts-old.html` and name the new one `accounts.html` but that might have other complications.

bedrock/firefox/templates/firefox/accounts-2018.html
stephaniehobson: That makes sense given what I saw of the javascript.

I wouldn't block launch on it but you can fix it if you like.

bedrock/firefox/templates/firefox/accounts.html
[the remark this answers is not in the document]
amychurchwell: oh I see, I was thinking this was the other file. Not sure why this was touched.

bedrock/firefox/templates/firefox/accounts-2018.html
amychurchwell: I had forgotten about this road bump until I started playing with the macro again-- 
screen shot 2018-11-30 at 11 07 29 am
The macro automatically displays the mobile buttons when logged in & in firefox, but we wanted to do a custom modal with the QR & buttons instead so I opted for this handwritten option.

stephaniehobson: 👍

media/static-bundles.json
craigcook: We've added a Protocol-flavored form style

dzingeek:
A few small bugs:

- On fx desktop > signed into an account: button should be blue, not black
- Footer: Features link is broken (500 error)
- Hero section: Add more space (2x as much) between the title and button
- On fx desktop> not signed in to an account: add space between the Create an Account button and the copy below (Already have an account...)
- On fx mobile > signed into an account: there is a ghost button that should not be there. "Set up Sync for iOS"

screen shot 2018-11-30 at 1 36 05 pm
screen shot 2018-11-30 at 1 46 46 pm
screen shot 2018-11-30 at 1 44 16 pm

stephaniehobson:
@dzingeek In terms of priorities. The only one of these that looks like it's worth 3 of us working Friday at 7:00 for looks like: "On fx mobile > signed into an account: there is a ghost button that should not be there. "Set up Sync for iOS""

Do you agree with that prioritizing?

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,243 @@
...
+  </div>
+
+  <section>
+    <div id="fx-app-buttons" class="show-fxa-supported-signed-in">
+        <ul class="mobile-download-buttons">
+          <li>
+            {{ google_play_button(id=android_id, extra_data_attributes={'download-location': button_location}) }}
+          </li>
+          <li>
+            <a href="{{ firefox_ios_url(app_store_campaign) }}" data-link-type="download" data-download-os="iOS" data-download-location="primary cta" id="{{ ios_id }}">
```
stephaniehobson: Please replace `{{ ios_id }}` with an ID

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -0,0 +1,243 @@
...
+    <div class="banner-cta">
+
+      <div class="fxa-cta show-fxa-supported-signed-in show-fxa-supported-signed-out">
+        <div class="show-fxa-supported-signed-in">
+          <ul class="mobile-download-buttons" id="bottom-banner-mobile">
+            <li>
+              {{ google_play_button(id=android_id, extra_data_attributes={'download-location': 'secondary cta'}) }}
+            </li>
+            <li>
+              <a href="{{ firefox_ios_url(app_store_campaign) }}" data-link-type="download" data-download-os="iOS" data-download-location="secondary cta" id="{{ ios_id }}">
```
stephaniehobson: Please replace {{ ios_id }} with an ID

bedrock/firefox/templates/firefox/accounts.html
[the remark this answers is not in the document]
stephaniehobson: So this needs to be switched back?

stephaniehobson:
Just a couple small things left! So much work done between Wednesday and today!

PR #6505, opened 2018-11-22
Update currency presets for 2018 Fundraising banner

TheoChevalier:
Those changes will be particularly important for the upcoming fundraising banner in https://github.com/mozilla/bedrock/pull/6501

Changes:
- Adding CAD since we now have en-CA support on moz.org
- Adjusting currency presets based on latest exchange rates.
- Fixed CLP values, they were quite off

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -755,8 +755,8 @@ DONATE_PARAMS = {
     'bn-BD': {
         'currency': 'bdt',
         'symbol': u'৳',
-        'presets': '1600,800,400,240',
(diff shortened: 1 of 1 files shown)


TheoChevalier:
For reference, proceeding with some value updates on donate.mozilla.org too in https://github.com/mozilla/donate.mozilla.org/pull/2121

kyoshino:
Haven’t this deployed yet? I’m seeing GB£ instead of CA$ on [/en-CA/](https://www.mozilla.org/en-CA/).

PR #6508, opened 2018-11-23
Document locale functionality of view helper

stephaniehobson:


pmac:
This looks good to me. My only note is that this is a shorter version of the more full documentation of this from the [l10n docs](https://bedrock.readthedocs.io/en/latest/l10n.html#locale-specific-templates). Perhaps a link over to the full docs would be good here as well?

stephaniehobson:
Actually this doesn't work (see https://github.com/mozilla/bedrock/pull/6520). I'll update the procedure and resubmit.

pmac:
Well it does if there if the language has an active lang file for that page, just like every other translated page. But if it doesn't have or need a .lang file then you have to do the thing from #6520.

Issue #6504, opened 2018-11-22
Remove jQuery as a dependency from Protocol base template


alexgibson:
All Protocol components use vanilla JS and have zero dependencies, but jQuery is still bundled in the Protocol base template due to legacy scripts in bedrock.

It would be great to eliminate jQuery from these dependencies, so we can lighten the overall JS payload across the site. We could do this by porting the following files to vanilla JS (or by removing them entirely):

- [x] `js/base/mozilla-utils.js`
- [x] `js/base/mozilla-image-helper.js` (I don't think this is being actively used on any Protocol pages)
- [x] `js/base/nav-main-resp.js` (can be removed once we localize the new navigation)
- [x] `js/base/base-page-init.js`
- [x] `js/base/core-datalayer.js`
- [ ] `js/base/core-datalayer-init.js`
- [x] `js/base/search-params.js`
- [ ] `js/base/stub-attribution.js`

Note: we may still want to keep a version of jQuery around to serve to [legacy IE](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/base-pebbles.html#L178-L180), but that should be fine as it's in a conditional comment.

stephaniehobson:
Could I do this? I've been wanting a chance to really focus on JS.

alexgibson:
Go for it 👍

stephaniehobson:
@agibson What is our minimum viable browser support for these files?

alexgibson:
Most of the dependencies in this issue were resolved, however a couple still remain. Digging deeper, there are still more pieces of legacy code that aren't covered here. I'm going to close this issue and open new issues for the remaining places.

PR #6515, opened 2018-11-26
Remove jQuery dependency from search-params.js

stephaniehobson:
## Description
Remove jQuery dependency from search-params.js

## Issue / Bugzilla link
#6504

## Testing
File has very good unit tests associated with it.

### DIFF
--- a/media/js/base/search-params.js
+++ b/media/js/base/search-params.js
@@ -11,14 +11,14 @@ var _SearchParams = function (search) {
     search = search.match(/^\??(.*)/)[1];
     search = search ? search.split(/[&;]/m) : [];
 
-    search.forEach(function (param) {
-        param = param.split('=');
+    for (var i = 0; i < search.length; i++) {
+        var param = search[i].split('=');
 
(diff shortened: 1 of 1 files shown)


media/js/base/search-params.js
```diff
@@ -11,7 +11,7 @@ var _SearchParams = function (search) {
     search = search.match(/^\??(.*)/)[1];
     search = search ? search.split(/[&;]/m) : [];
 
-    $.each(search, function (index, param) {
+    search.forEach(function (param) {
```
alexgibson: Looking in our bundle definitions, this file is [included in our legacy IE8 bundle](https://github.com/mozilla/bedrock/blob/master/media/static-bundles.json#L1527). According to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/forEach), `Array.forEach` is only supported in IE9, so we may need to support even more ancient machinery here :(

stephaniehobson:
Updated.

Issue #6387, opened 2018-10-31
Firefox Fights For You Campaign Page


ejregithub:
[MasterProjectDoc](https://docs.google.com/document/d/1l4VKr7uEeZLha1MSsF0Rb4yhYhFG6ng8daV0zbESrcs/edit#)

| Estimates |  |
| ------------- | ------------- | 
| TL | 10|
| PM | 12 |
| ENG | 14 |
| SEO | 1 |
| ANA | 4 |
| DES | 40 |
| COPY | 20 |

Description, good enough that reading the card 1 month from now, you still know what it means.
---
#### :yellow_heart: Success Criteria :yellow_heart:
- [x] Analytics in page (GMT - data attributes) to enable proper data reporting post launch/page is tagged appropriately to ensure data is sent and record
- [x] Copy Deck reviewed and approved by MJ
- [x] Copy translated into German - due on 11/21
- [x] Launch to en-US and DE (confirmed w. Janis on 10/31)
- [x] Visuals
- [x] Build in Protocol w. new components added
- [ ] ~Protocol code/design review~
- [x] Design must be approved by the team and the Approver
- [ ] Legal has approved the copy
- [x] The Coded page must be approved by the team and the Approver
- [x] Analytics has QAed and approved tagging in the page
- [x] Page includes a section on Firefox Desktop Browser as an entry point for talking about how FFFY
- [x] Page includes sections on sub-products as proof of Firefox Fighting for You
- [x] Page is set to be indexed (@rraue feels doing so will be of benefit)
- [x] Video component
- [x] We need to understand video content so we can write copy around it
- [x] Survey link included

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [x] Localization as a blocker to launch - ask Michele (en-US and DE required for launch)
- [x] Video component (the video) may not land on this page
- [x] Time line needs to built and shared
- [x] The creative team is out all next week (11/12->)
- [x] Design calls for a new web font we don't have a license to

ejregithub:
Eric pinged Janis in the FFFY-page Slack channel "for the page being produced here - is DE required for launch (is it tied to campaigns/messaging on/after the specified launch date) ?"

ejregithub:
Janis confirmed that the page must be localized to DE for launch. "we will have DE media driving as well.  It will also start 11/27"  So, I'm clearing the flag in Success Criteria and nothing EN and DE as Success Criteria

stephaniehobson:
[...]

PR #6520, opened 2018-11-27
Set active locales for FFY campaign page (#6387)

alexgibson:
## Description
- Sets active locales for `/firefox/fights-for-you/` campaign page ([docs link](https://bedrock.readthedocs.io/en/latest/l10n.html#specifying-active-locales-in-views)).

## Issue / Bugzilla link
#6387

## Testing
- With `Dev=False` in your `.env` file, both English and German should be available in the language switcher in the page footer.


alexgibson:
@stephaniehobson I noticed an issue accessing the German translation of the FFY page when `Dev=False`. This is relatively new functionality in bedrock, so sorry this slipped through review. This PR fixes the issue for me locally.

PR #6526, opened 2018-11-27
[#6465] Fundraising banner, fix close button z-index

craigcook:
## Description
Fixes a bug wherein the close button is often covered by the content div at some screen sizes and becomes unclickable (the button is visible because the div is padded, but it's "behind" the padded area of the div). Giving the button a z-index brings it to the front.


craigcook:
Closing in favor of #6527 with its sensibly lower z-index value. I humbly apologize for my overenthusiasm.

Issue #6444, opened 2018-11-06
LiveNation campaign landing page


ejregithub:
| Estimates |  |
| ------------- | ------------- | 
| TL | |
| PM | |
| ENG | |
| SEO | |
| ANA | |
| DES | |
| COPY | |

Description: webpage on mozilla.org supporting the LiveNation campaign promoting FxAccounts

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [x] Live on 11/30
- [x] Includes FxA form for users to receive notifications about event *and* access to it.

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] Content to enable copywriting

ejregithub:
![image](https://user-images.githubusercontent.com/20547886/48519808-25c58080-e823-11e8-82fe-953e312e05ae.png)

craigcook:
Note: I initially kept this open since I was expecting ongoing updates to the page and wanted to keep it all in one place, but as it turned out the page hasn't needed any changes for a while and we're still not sure if/when the next concert will happen. So closing this now is good cleanup and we'll do any further concert updates as new issues. 👍

PR #6546, opened 2018-11-30
[#6444] Add Firefox Concert Series landing page

craigcook:
## Description
A landing page for the Firefox Concert Series, promoting Firefox Accounts. This uses our new FxA OAuth flow so you can create an account and sign in from any browser. It's not tied to Sync or UITour so you'll see a "create account" form even if you're signed into Firefox.

It's English only so no l10n, and the concert stream will only be available in the US anyway. Uses geolocation to determine if the user in the US and can access the stream; all non-US countries get an apology. If JS fails or geolookup fails we show the concert info by default (the idea is to assume the user has access unless we can determine otherwise; at showtime it's livenation's problem).

## Issue / Bugzilla link
#6444 

## Testing
https://www-demo5.allizom.org/firefox/concerts/

You won't be able to sign in locally without setting a client_id in settings/base.py 

A few states to verify:

- [x] In US - show concert info and FxA section
- [x] Outside US - show a sincere apology and the newsletter form
- [x] Not authed - show the one-field FxA form
- [x] Authed - show a "you're signed in" message


craigcook:
We are apparently semi-blocked by getting approval from the artist (or his representation, anyway) so it's possible we'll need to change the image and copy at the last second, but all the functionality and layout is ready for review.

stephaniehobson:
The geofencing is not keeping this Canadian out.

stephaniehobson:
I just remembered the geo fencing doesn't work on demo. We will have to test it on stage tomorrow.

stephaniehobson:
This text is a bit too low contrast:

screen shot 2018-11-29 at 8 47 02 pm

bedrock/firefox/templates/firefox/concerts.html
```diff
@@ -2,48 +2,189 @@
...
+            <p class="mzp-c-card-tag ended">Ended</p>
+          {% else %}
+            <p class="mzp-c-card-tag countdown" id="countdown-one">Starts in <span class="time days"></span> days, <span class="time hours"></span> hours</p>
+          {% endif %}
+            <p class="date"><time datetime="2018-12-11T21:30:00-0500">December 11 at 9:30pm Eastern</time></p>
+            <h2 class="mzp-c-card-feature-title">Phosphorescent</h2>
+            <p class="mzp-c-card-feature-desc">Live from DC, get a front row seat for one of the last shows on Phosphorescent’s <cite lang="fr">C’est La Vie</cite> tour.</p>
+
+          {% if switch('firefox-concert-one-live') %}
+            <p class="stream-link"><a href="#TBD" class="mzp-c-button mzp-t-primary mzp-t-dark">Watch Now</a></p>
```
stephaniehobson: This button doesn't do anything.

Of course, I wouldn't be seeing it if I wasn't in dev.

bedrock/firefox/templates/firefox/concerts.html
```diff
@@ -2,48 +2,189 @@
...
 
+{% block body_class %}firefox-concerts{% endblock %}
+
 {% block content %}
 <main role="main">
-  <div class="content">
-    <div class="conditional-content verified-content">
-      <p>You have been verified!</p>
+  <div class="concert-content">
+    <header class="mzp-c-hero mzp-t-firefox mzp-t-product-firefox mzp-t-dark">
```
stephaniehobson: FYI we have a macro for hero now.

Don't change this though, let's just get it into production.

media/js/firefox/concerts.js
```diff
@@ -1,6 +1,173 @@
...
+        US_COUNTRY_CODE: 'us',
+        COOKIE_ID: 'firefox-concerts',
+        COOKIE_EXPIRATION_DAYS: 3,
+    };
+
+    var _geoTimeout;
+    var _requestComplete = false;
+
+    // take params from URL and pass through signup button
+    ConcertPage.passThroughParams = function() {
```
stephaniehobson: 👍

media/js/firefox/concerts.js
```diff
@@ -1,6 +1,173 @@
...
+                return match[1].toLowerCase();
+            }
+            return false;
+        }
+        return false;
+    };
+
+    ConcertPage.verifyLocation = function(location) {
+        if (location) {
+            return location !== ConcertPage.US_COUNTRY_CODE;
```
stephaniehobson: Is this supposed to be a negative match?

media/js/firefox/concerts.js
```diff
@@ -1,6 +1,173 @@
...
+
+    ConcertPage.onRequestComplete = function(data) {
+        var country = typeof data === 'string' ? data : 'none';
+
+        clearTimeout(_geoTimeout);
+
+        if (!_requestComplete) {
+            _requestComplete = true;
+
+            // Set a cookie so we don't have to query location on repeated page loads.
```
stephaniehobson: When I set my cookie to be `ca` the page still loads.

I know we can't really test the geofencing until it's on stage though :(

media/js/firefox/concerts.js
```diff
@@ -1,6 +1,173 @@
...
+    };
+
+    ConcertPage.init = function() {
+        ConcertPage.passThroughParams();
+
+        var cookiesEnabled = typeof Mozilla.Cookies !== 'undefined' || Mozilla.Cookies.enabled();
+        var override = ConcertPage.hasGeoOverride();
+
+        // only show excluded content if cookies are enabled (let LiveNation deal with it)
+        if (cookiesEnabled) {
```
stephaniehobson: Not a blocker but we could still use the geofencing if cookies are disabled. This should be updated *after* we get this into production.

media/js/firefox/concerts.js
```diff
@@ -1,6 +1,173 @@
...
+        } else {
+            ConcertPage.showConcertContent();
+        }
+    };
+    window.Mozilla.ConcertPage = ConcertPage;
+
+
     var stateStorageKey = 'fxaOauthState';
     var verifiedStorageKey = 'fxaOauthVerified';
     var cookieDays = 1;
```
stephaniehobson: Why 1? That seems short.

media/js/firefox/concerts.js
```diff
@@ -70,12 +237,76 @@
...
     // apply className determined above to show proper content
-    document.documentElement.className += ' ' + className;
+    document.body.classList.add(className);
+
+    // Phosphorescent - December 11 at 9:30pm Eastern
+    var showtimeOne = 'December 11 2018 21:30:00 GMT-0500';
+    var countdownOne = document.getElementById('countdown-one');
+
+    // Calculate time from now to a future endtime
+    function getTimeRemaining(endtime) {
```
stephaniehobson: Ugh dates in javascript. 😵

media/js/firefox/concerts.js
craigcook: Yes, "if location is not US"

media/css/firefox/concerts.scss
```diff
@@ -1,11 +1,248 @@
...
+//* -------------------------------------------------------------------------- */
+// General page styles
+
+.mzp-l-content.l-content-narrow {
+    max-width: $content-lg;
+}
+
+.firefox-concerts {
+    #outer-wrapper {
+        background: #41085e linear-gradient(to bottom, #000 0%, #06043c 400px, #41085e 60%);
```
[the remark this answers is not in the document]
craigcook: [The mix-in](https://github.com/mozilla/protocol/blob/master/src/assets/sass/protocol/includes/mixins/_gradients.scss#L15) just does `-webkit` but I can double-declare so old browsers get a solid color.

media/js/firefox/concerts.js
craigcook: Good point... this was cribbed from the Yandex page, which I think had a more strict cookie requirement because it's doing more than just geofencing.

media/js/firefox/concerts.js
craigcook: Ask jpetto, he wrote this part ;) But yeah, I think I'll bump that up... maybe a week? Two weeks?

bedrock/firefox/templates/firefox/concerts.html
craigcook: Yeah, we won't get the stream URL from LiveNation until later so we already know there will be at least one update to this page before showtime.

bedrock/firefox/templates/firefox/concerts.html
craigcook: Changed to the macro, though it outputs a `section` and I would prefer a `header` in this instance.

PR #6548, opened 2018-11-30
Fix issues caused by duplicate articles in Pocket feed

### LINKED ISSUE #6547: Duplicate articles in pocket feed causing DB updates to fail
Here's [the exception in the logs](https://gist.githubusercontent.com/metadave/aa83329f5e62102a4d23006d0bc6a1d7/raw/b917c266595cbb33e6676497ae51a6f18649d954/gistfile1.txt) (thanks @metadave).

The issue is that the feed (https://getpocket.com/@MozillaHQ) currently contains the same article twice. This caused said article to be added to the database twice, which worked fine at the time, though is suboptimal for UX reasons. However, when bedrock went to update the articles again it encountered a `MultipleObjectsReturned` exception when trying to run this line:

https://github.com/mozilla/bedrock/blob/28705df22c4efa7775b2106cb900b923d0caf4e6/bedrock/pocketfeed/models.py#L39

So we need to do two things:

1. Ensure that we don't add dupe articles to the DB in future.
2. Deal with the case when dupe articles do make it into the DB.

I'm working on a PR now.

pmac:
Fix #6547

### DIFF
--- a/bedrock/pocketfeed/models.py
+++ b/bedrock/pocketfeed/models.py
@@ -32,6 +32,11 @@ class PocketArticleManager(models.Manager):
         article_ids = []
         articles_to_update = []
         for article in articles:
+            # apparently it's possible for an article to be
+            # in the feed multiple times ¯\_(ツ)_/¯
+            if article['id'] in article_ids:
+                continue
+
             article_ids.append(article['id'])
 
             try:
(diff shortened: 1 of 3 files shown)


pmac:
Thanks!

PR #6549, opened 2018-11-30
[fix bug 1511104] Propagate utm_* params from /new/ to /download/thanks/

jpetto:
## Description

For macOS visitors on `/firefox/new/` only, takes a set of pre-defined `utm_` params in the URL (`campaign`, `medium`, `source`, and `content`), validates their values, removes the `utm_` prefix, and then adds them as query params to all links pointing to `/firefox/download/thanks/`.

The intent of this change is to have attribution on macOS downloads that come from AMO.

This PR also modifies the JS in `firefox/new/variation-scene1.js`. Instead of blindly appending _all_ query parameters in the current URL on to links pointing to `/firefox/download/thanks/`, we now _only_ append the `v` or `xv` parameters (to ensure the correct template is loaded on `/download/thanks/`).

## Issue / Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1511104

## Testing

- [x] Test `/firefox/new/` with the specific  `utm_` params/values (see the tests) and ensure they are present **without the `utm_` prefix** on the primary and modal download buttons.

- [x] Also make sure download button links are unchanged if there are no/incorrect `utm_` params in the URL.

- [x] Test existing `xv=` `/new/` variations with any random params (`utm_` and/or other) and ensure only the `xv` param is appended to the download button.


media/js/firefox/new/scene1.js
```diff
@@ -108,21 +108,69 @@
         });
     };
 
+    var getUtmParamString = function() {
+        var utmParamString = '';
+        var params = new window._SearchParams().utmParams();
```
stephaniehobson: 👍

jpetto:
> I'm a little worried about this clashing with https://github.com/mozilla/bedrock/blob/master/media/js/firefox/new/variation-scene1.js but this doesn't seem to be included in the out put of static bundle for some reason even though it's in the list.

Hm, good catch. This will cause issues if both `scene.js` and `variation-scene1.js` are included on the page. Will update.

jpetto:
@stephaniehobson updated!

This approach is a little heavy-handed, but should work fine for our current use cases.

However, if we start adding any query params to the `/download/thanks/` link pre-JS, this will all need to be re-written.

stephaniehobson:
How urgent is this? 

This works but I'm concerned we're leaving a time bomb for a future developer. I'd rather we combined the functions and included them on all scene1 pages by default. How do you feel about that given urgency/availability considerations?

jpetto:
I think it's _kind of_ urgent, as we currently are getting no attribution data for macOS downloads. 

I definitely hear your concern, and had the same while writing the PR. A full fix here requires answering the following questions:

1. On variations, we currently throw the _entire_ querystring on to the download links. Is this required? Functionally (meaning to serve the correct scene 2 template), all we need is the `xv` parameter. I believe we did all the values because it was easier and not hurting anything.

2. For this specific fix (getting `utm_` params added to the download links), would it hurt to just have all query params?

I can ask #2 in the bug. I'm not sure who would answer #1 - someone in analytics, most likely.

The easiest fix here would be to move the JS from `variation-scene1.js` into `scene1.js` - just appending all query params to all download links all the time. I'm wary to do this without confirmation from both analytics and product, as I don't want to mess up any existing reporting/data processing.

Because we're adding _all_ params in variations, it _might_ be safe to assume we can do this on the canonical `/new` page as well. 

I'll ask in [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1511104), which should satisfy the product side of things.

jpetto:
After comments from hoosteeno and raue, this approach may not work in regards to analytics. Not closing quite yet, but will likely be closed soon.

(May need to look into some sort of stub attribution for macOS.)

jpetto:
We have a path forward:

For macOS users only, we will take all `utm_` params present on `/firefox/new/` and, after removing the `utm_` part, append them to all download links pointing to `/firefox/download/thanks/`.

This PR also updates the JS run on `/firefox/new/` variations. This JS _used_ to blindly append all params in the URL to the download button, including `utm_` params. This could potentially impact analytics, so `utm_` params are now explicitly skipped. *NEED TO VERIFY THIS CHANGE WILL NOT HURT ANYTHING*, hence Do Not Merge.

jpetto:
Verified w/Justin C that the changes here will not negatively impact analytics. Should (again) be good for review.

media/static-bundles.json
```diff
@@ -1635,6 +1640,7 @@
         "js/base/mozilla-client.js",
         "js/base/nav-main-resp.js",
         "js/base/class-list-polyfill.js",
+        "js/base/object-assign-polyfill.js",
```
alexgibson: As we're adding a new polyfill to the base bundle for every page, I would recommend pushing this branch to `run-integration-tests`, just to make sure we're not activating any new code paths that have previously not been tested in older browsers.

media/static-bundles.json
```diff
@@ -1079,7 +1081,8 @@
     {
       "files": [
         "js/base/mozilla-modal.js",
-        "js/firefox/new/scene1.js"
+        "js/firefox/new/scene1.js",
+        "js/firefox/new/scene1-macos-attribution.js"
```
alexgibson: Reading the discussion in the bug I understand that the current ask is for this to only apply to `/firefox/new/`, but my gut says this code should really belong in the [stub attribution bundle](https://github.com/mozilla/bedrock/blob/master/media/static-bundles.json#L1146-L1152) along with the windows code? Less because it's neater code-wise, but more because that bundle is isolated to do one thing, so if something throws an error it will not halt excecution of the page JS or break basic functionality.

I'd be interested in @pmac's thoughts here too.

media/js/firefox/new/scene1-macos-attribution.js
```diff
@@ -0,0 +1,46 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function(Mozilla) {
```
alexgibson: I know there's not much code here, but seeing as it does some pretty important functionality it would be great if it had some unit tests?

media/js/firefox/new/variation-scene1.js
```diff
@@ -8,12 +8,40 @@ For 'standard' /new variations, this file will append all query params
 so the correct template is rendered on /download/thanks/.
 */
 
-(function($) {
+(function() {
```
alexgibson: This code feels like it's also now getting complicated enough that it could do with some tests?

jpetto:
Added a commit refactoring into a separate bundle with unit tests. Waiting for a response in the bug to see if/how we can further limit the values migrated from the querystring into the DOM.

jpetto:
[Integration tests are passing.](https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/233/pipeline/)

jpetto:
Updated the code again to have unit tests as well as a separate bundle enabled by a switch.

Should now be good for final review.

bedrock/firefox/templates/firefox/new/fx/scene1.html
```diff
@@ -80,5 +80,8 @@ <h2>Need to download again?</h2>
 {% endblock %}
 
 {% block js %}
+  {% if switch('stub-attribution-macos') %}
```
alexgibson: This experiment is finished and was shown only to existing Firefox users, is this attribution code something we need to include here? (one less thing to manually test and verify if the answer is no)

bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -160,5 +160,8 @@ <h4 class="heading-linux">{{_('Download Firefox <span>for Linux</span>') }}</h4>
 {% endblock %}
 
 {% block js %}
+  {% if switch('stub-attribution-macos') %}
```
alexgibson: Nit: from a user perspective all the critical UI logic is done in `firefox_new_fx_scene1`. Is there any reason why this `stub-attribution-macos` bundle needs to come before that? Could it be moved to last in the queue?

media/js/firefox/new/variation-scene1.js
```diff
@@ -8,12 +8,35 @@ For 'standard' /new variations, this file will append all query params
...
+    var href;
+    var newQs;
+    var params = new window._SearchParams().params;
+    var prefix;
+
+    // we need to propogate 'v' or 'xv' query params to ensure the correct template is loaded on /download/thanks/
+    // only one of these keys should be present
+
+    // 'xv' denotes a 'variant' page for a specific marketing campaign
+    if (params.hasOwnProperty('xv')) {
```
alexgibson: This is perhaps a little beyond the scope of this PR but may be easy enough to add as we're touching this code. Should we also add a little sanitisation to these param values? e.g. I think we'd just need to allow alpha numeric values and `-` characters (looking at what we use currently).

media/js/firefox/new/variation-scene1.js
```diff
@@ -8,12 +8,35 @@ For 'standard' /new variations, this file will append all query params
...
+        newQs = 'xv=' + params['xv'];
+    // 'v' denotes an experiment variation
+    } else if (params.hasOwnProperty('v')) {
```
alexgibson: Should this be an `if` instead of an `else if`? Currently if there's both an `xv` param and `v` param, only the former gets appended and the latter gets dropped.

media/js/base/stub-attribution-macos.js
```diff
@@ -0,0 +1,134 @@
...
+        if (params.hasOwnProperty('utm_medium') && params['utm_medium'] === 'referral') {
+            finalParams['medium'] = 'referral';
+        }
+
+        if (params.hasOwnProperty('utm_source') && params['utm_source'] === 'addons.mozilla.org') {
+            finalParams['source'] = 'addons.mozilla.org';
+        }
+
+        // utm_content should *always* start with "rta%3A", followed by alphanumeric characters
+        if (params.hasOwnProperty('utm_content') && (/^rta%3A[a-zA-Z0-9_\-]+$/).test(params['utm_content'])) {
```
alexgibson: 🙏

media/js/firefox/new/variation-scene1.js
jpetto: I originally had this as two separate `if` statements, but then thought that there's no way both `v` and `xv` can meaningfully co-exist.

Buuuuut, maybe if an `xv` has an experiment running with different `v` values, they could? I'll switch back to two separate `if` statements. Worst case scenario is one clobbers the other - more of a task for the view in that case.

alexgibson: Experience variations can (and sometimes have) had multiple A/B test variations, so yes I think this is something we'll need to support 👍

PR #6557, opened 2018-12-01
Update tests for new accounts page

craigcook:
## Description
I think I had a version of this working at some point previous to the new page merging but it's possible I never committed it, or never pushed it, or pushed it and it got overwritten in a rebase, or something who knows... the week is a bit of a blur. Anyway, I think this updates the tests for the new accounts page and it's passing locally so maybe I did something right. At the time I'm submitting this PR (late on Friday night, I might add) I've pushed it to the `run-integration-tests` branch but Jenkins is unresponsive and probably needs a kick, but I'm going to submit the PR anyway. If I'm completely off base feel free to just close it but if I'm on the right track I'm happy to tweak until we get there. And I'm babbling and should probably go to bed.

### DIFF
--- a/tests/functional/firefox/test_accounts.py
+++ b/tests/functional/firefox/test_accounts.py
@@ -7,17 +7,17 @@ import pytest
 from pages.firefox.accounts import FirefoxAccountsPage
 
 
-@pytest.mark.skip_if_firefox(reason='Download button is displayed only to non-Firefox users')
+@pytest.mark.skip_if_firefox(reason='Download button is only displayed to non-Firefox users')
 @pytest.mark.nondestructive
 def test_download_button_displayed(base_url, selenium):
     page = FirefoxAccountsPage(selenium, base_url).open()
-    assert not page.is_accounts_form_displayed
+    assert not page.is_create_account_form_displayed
     assert page.download_button.is_displayed
 
 
-@pytest.mark.skip_if_not_firefox(reason='Accounts iFrame is only displayed to Firefox users')
+@pytest.mark.skip_if_not_firefox(reason='Create Account form is only displayed to Firefox users')
 @pytest.mark.nondestructive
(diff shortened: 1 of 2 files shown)


tests/functional/firefox/test_accounts.py
```diff
@@ -7,17 +7,17 @@
 from pages.firefox.accounts import FirefoxAccountsPage
 
 
-@pytest.mark.skip_if_firefox(reason='Download button is displayed only to non-Firefox users')
+@pytest.mark.skip_if_firefox(reason='Download button is only displayed to non-Firefox users')
 @pytest.mark.nondestructive
 def test_download_button_displayed(base_url, selenium):
     page = FirefoxAccountsPage(selenium, base_url).open()
-    assert not page.is_accounts_form_displayed
+    assert not page.create_account_button.is_displayed
```
alexgibson: I don’t believe `create_account_button` has a property called `is_displayed` as this is something our reusable code to test download buttons has, but not regular elements. I think removing the `.is_displayed` part should be enough for these tests to pass again, but then `assert not page create_account_button` isn’t very readable. I would maybe rename the property to `is_create_account_form_displayed` which sounds a bit clearer.

craigcook:
I made some changes but still don't really have any clue what I'm doing.

tests/pages/firefox/accounts.py
```diff
@@ -22,4 +22,4 @@ def download_button(self):
 
     @property
     def create_account_button(self):
```
alexgibson: You need to rename this to `is_create_account_form_displayed`, then it should work.

craigcook: D'oh, of course it was something obvious.

craigcook:
Last build on run-integration-tests is finally green! I still think we should work on adding/improving tests for this page but at least this much is passing and we can get the page live.

PR #6561, opened 2018-12-04
[#6444] Concerts - update start time, add opening act, add stream URL

craigcook:
## Description
Moving the start time up because the opening act wants to get in on the stream. This also adds the URL to the button, though the page at the other end isn't up yet.

### DIFF
--- a/bedrock/firefox/templates/firefox/concerts.html
+++ b/bedrock/firefox/templates/firefox/concerts.html
@@ -48,12 +48,11 @@
           {% else %}
             <p class="mzp-c-card-tag countdown" id="countdown-one">Starts in <span class="time days"></span> days, <span class="time hours"></span> hours</p>
           {% endif %}
-            <p class="date"><time datetime="2018-12-11T21:30:00-0500">December 11 at 9:30pm Eastern</time></p>
-            <h2 class="mzp-c-card-feature-title">Phosphorescent</h2>
-            <p class="mzp-c-card-feature-desc">Live at the 9:30 Club, get a front row seat for one of the last shows on Phosphorescent’s <cite lang="fr">C’est La Vie</cite> tour.</p>
+            <h2 class="mzp-c-card-feature-title"><time datetime="2018-12-11T20:00:00-0500">Tuesday, Dec. 11</time>: Phosphorescent</h2>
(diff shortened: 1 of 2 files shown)


media/js/firefox/concerts.js
```diff
@@ -247,7 +247,7 @@
     document.body.classList.add(className);
 
     // Phosphorescent - December 11 at 9:30pm Eastern
```
alexgibson: Nit: update comment to match the new time?

bedrock/firefox/templates/firefox/concerts.html
```diff
@@ -48,12 +48,11 @@
...
           {% endif %}
-            <p class="date"><time datetime="2018-12-11T21:30:00-0500">December 11 at 9:30pm Eastern</time></p>
-            <h2 class="mzp-c-card-feature-title">Phosphorescent</h2>
-            <p class="mzp-c-card-feature-desc">Live at the 9:30 Club, get a front row seat for one of the last shows on Phosphorescent’s <cite lang="fr">C’est La Vie</cite> tour.</p>
+            <h2 class="mzp-c-card-feature-title"><time datetime="2018-12-11T20:00:00-0500">Tuesday, Dec. 11</time>: Phosphorescent</h2>
+            <p class="mzp-c-card-feature-desc">Livestreamed from DC’s 9:30 Club, get a front row seat for one of the last shows on Phosphorescent’s <cite lang="fr">C’est La Vie</cite> tour. Special guest Liz Cooper and The Stampede kick off the show 8:00pm EST.</p>
 
           {% if switch('firefox-concert-one-live') %}
-            <p class="stream-link"><a href="#TBD" class="mzp-c-button mzp-t-primary mzp-t-dark">Watch Now</a></p>
+            <p class="stream-link"><a href="https://firefoxconcerts.livenation.com" rel="external" class="mzp-c-button mzp-t-primary mzp-t-dark">Watch Now</a></p>
```
alexgibson: I don't think `mzp-t-primary` is an in-use Protocol class name?

craigcook: oh I thought we had primary and secondary... but maybe primary is just the default. So it's doing no harm but also no good. Will remove.

PR #6569, opened 2018-12-06
Convert CSP_DEFAULT_SRC from static to everett config

jgmize:
## Description

This will allow us to add *.allizom.org to the CSP header in dev and staging deployments.

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/6570


jgmize:
@pmac I pushed a fix for the issue we saw in https://circleci.com/gh/mozilla/bedrock/15122 and now all tests are passing.

PR #6575, opened 2018-12-07
Remove jQuery dependency (#6504)

stephaniehobson:
## Description
- remove jQuery dependency
- add test for initLangSwitcher

## Issue / Bugzilla link
#6504

### DIFF
--- a/media/js/base/mozilla-utils.js
+++ b/media/js/base/mozilla-utils.js
@@ -56,7 +56,7 @@ if (typeof Mozilla === 'undefined') {
     // In order to use it, you need a block string_data bit inside your template,
     // then, each key name needs to be preceded by data- as this uses data attributes
     // to work. After this, you can access all strings defined inside the
-    // string_data block in JS using Mozilla.Utils.trans('keyofstring'); Thank @mkelly
+    // string_data block in JS using Mozilla.Utils.trans('key-of-string'); Thank @mkelly
     var _strings = document.getElementById('strings');
     Utils.trans = function(stringId) {
         if (_strings) {
--- a/media/js/base/mozilla-utils.js
+++ b/media/js/base/mozilla-utils.js
@@ -16,7 +16,7 @@ if (typeof Mozilla === 'undefined') {
     Utils.initMobileDownloadLinks = function() {
         if (site.platform === 'android') {
             var playLinks = document.querySelectorAll('a[href^="https://play.google.com/store/apps/"]');
-            for(var i = 0; i < playLinks.length; ++i) {
+            for (var i = 0; i < playLinks.length; ++i) {
                 var playLink = playLinks[i];
                 var oldHref = playLink.getAttribute('href');
                 var newHref = oldHref.replace('https://play.google.com/store/apps/', 'market://');
@@ -39,7 +39,7 @@ if (typeof Mozilla === 'undefined') {
             links[i].setAttribute('href', distributionLink);
         }
         for (var j = 0; j < images.length; j++) {
-            var distributionSrc = images[j].dataset[distribution + 'Link'];
+            var distributionSrc = images[j].getAttribute('data-' + distribution + 'Link');
             images[j].setAttribute('src', distributionSrc);
         }
     };
@@ -60,7 +60,7 @@ if (typeof Mozilla === 'undefined') {
     var _strings = document.getElementById('strings');
     Utils.trans = function(stringId) {
         if (_strings) {
-            return _strings.dataset[stringId];
+            return _strings.getAttribute('data-' + stringId);
         } else {
             return undefined;
         }
--- a/media/js/base/mozilla-utils.js
+++ b/media/js/base/mozilla-utils.js
@@ -15,10 +15,13 @@ if (typeof Mozilla === 'undefined') {
     // Replace Google Play links on Android devices to let them open the marketplace app
     Utils.initMobileDownloadLinks = function() {
         if (site.platform === 'android') {
-            $('a[href^="https://play.google.com/store/apps/"]').each(function() {
-                $(this).attr('href', $(this).attr('href')
-                    .replace('https://play.google.com/store/apps/', 'market://'));
-            });
+            var playLinks = document.querySelectorAll('a[href^="https://play.google.com/store/apps/"]');
+            for(var i = 0; i < playLinks.length; ++i) {
+                var playLink = playLinks[i];
+                var oldHref = playLink.getAttribute('href');
+                var newHref = oldHref.replace('https://play.google.com/store/apps/', 'market://');
+                playLink.setAttribute('href', newHref);
+            }
         }
     };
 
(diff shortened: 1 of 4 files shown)


tests/unit/spec/base/mozilla-utils.js
```diff
@@ -9,6 +9,67 @@ describe('mozilla-utils.js', function() {
...
+            spyOn(Mozilla.Utils, 'doRedirect');
+            var $select = $('#page-language-select');
+            // init
+            Mozilla.Utils.initLangSwitcher();
+            // change the drop down
+            $select.val('de');
+            // fire the event
+            $select.trigger('change');
+            // check redirected to expected page
+            expect(location.href === 'http://localhost:9876/debug.html?lang=de#&step=1');
```
alexgibson: Instead of asserting `location.href` I think it may be enough to just assert that `Utils.doRedirect()` was called? I'm not sure if we can really depend on the localhost address of the test runner to assert against reliably. If it may help, there are some equivalent tests for this feature over in the [protocol repo](https://github.com/mozilla/protocol/blob/master/tests/unit/protocol-lang-switcher.js#L67-L80) if you would like some plain JS examples for reference.

media/js/base/mozilla-utils.js
```diff
@@ -45,10 +52,10 @@ if (typeof Mozilla === 'undefined') {
 
     // language switcher
     Utils.initLangSwitcher = function() {
-        var $language = $('#page-language-select');
-        var previousLanguage = $language.val();
-        $language.on('change', function() {
-            var newLanguage = $language.val();
+        var language = document.getElementById('page-language-select');
```
alexgibson: While jQuery fails gracefully, this now throws an error on Protocol pages where `language` is undefined, so we may need to add a check that it exists before trying to query it.

```
jQuery.Deferred exception: language is null Utils.initLangSwitcher@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:155:13
@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:1690:9
j@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:29997
g/</k<@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:30313
setTimeout handler*g/<@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:30520
i@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:28012
fireWith@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:28783
fire@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:28817
i@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:28012
fireWith@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:28783
ready@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:31932
S@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:3:33
EventListener.handleEvent*@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:3:147
@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:220
@http://127.0.0.1:8000/media/js/BUNDLES/common-protocol.js:2:2
 undefined jquery-3.2.1.min.js:2
TypeError: language is null[Learn More] mozilla-utils.js:56

    initLangSwitcher mozilla-utils.js:56 <anonymous> base-page-init.js:23 j2jQuery
```

tests/unit/spec/base/mozilla-utils.js
stephaniehobson: I'm really trying to check that doRedirect is passed the correct parameters but it's passed something funny and I wasn't able to check it.

stephaniehobson: And you're probably right, we can't rely on the host :/

stephaniehobson: Updated to just check window.location.search. Better?

tests/unit/spec/base/mozilla-utils.js
```diff
@@ -9,6 +9,67 @@ describe('mozilla-utils.js', function() {
...
+            spyOn(Mozilla.Utils, 'doRedirect');
+            var $select = $('#page-language-select');
+            // init
+            Mozilla.Utils.initLangSwitcher();
+            // change the drop down
+            $select.val('de');
+            // fire the event
+            $select.trigger('change');
+            // check redirected to expected page
+            expect(window.location.search === '?lang=de#&step=1');
```
alexgibson: What I find slightly uneasy is we're already depending on the output of a completely unrelated tests here in order for this particular test to pass. From what I can see, `&step=1` is an unintended side effect from an existing test, so if that changes then this test suddenly starts failing. Ideally, I'd like to not have the test runner put through any kind of real redirects at all, as the more things like this we add the more prone it becomes to odd errors that are hard to debug :/

stephaniehobson: Updated just to look for the `lang=de` value.

Though I suppose this will be deleted in #6635

stephaniehobson:
Updated.

alexgibson:
This needs a rebase now that the footer PR merged.

stephaniehobson:
Updated.

media/js/base/mozilla-utils.js
```diff
@@ -50,9 +57,13 @@ if (typeof Mozilla === 'undefined') {
     // then, each key name needs to be preceded by data- as this uses data attributes
     // to work. After this, you can access all strings defined inside the
     // string_data block in JS using Mozilla.Utils.trans('keyofstring'); Thank @mkelly
-    var _$strings = $('#strings');
+    var _strings = document.getElementById('strings');
     Utils.trans = function(stringId) {
-        return _$strings.data(stringId);
+        if (_strings) {
+            return _strings.dataset[stringId];
```
alexgibson: Sorry I missed this originally. Using `dataset` here will likely throw errors in older browsers. I'd suggest using `getAttribute` for better legacy support (which is something jQuery did behind the scenes).

media/js/base/mozilla-utils.js
```diff
@@ -15,10 +15,13 @@ if (typeof Mozilla === 'undefined') {
     // Replace Google Play links on Android devices to let them open the marketplace app
     Utils.initMobileDownloadLinks = function() {
         if (site.platform === 'android') {
-            $('a[href^="https://play.google.com/store/apps/"]').each(function() {
-                $(this).attr('href', $(this).attr('href')
-                    .replace('https://play.google.com/store/apps/', 'market://'));
-            });
+            var playLinks = document.querySelectorAll('a[href^="https://play.google.com/store/apps/"]');
+            for(var i = 0; i < playLinks.length; ++i) {
```
alexgibson: Nit: space before paranthesis

stephaniehobson:
Updated.

media/js/base/mozilla-utils.js
```diff
@@ -50,9 +57,13 @@ if (typeof Mozilla === 'undefined') {
     // then, each key name needs to be preceded by data- as this uses data attributes
     // to work. After this, you can access all strings defined inside the
     // string_data block in JS using Mozilla.Utils.trans('keyofstring'); Thank @mkelly
-    var _$strings = $('#strings');
+    var _strings = document.getElementById('strings');
     Utils.trans = function(stringId) {
-        return _$strings.data(stringId);
+        if (_strings) {
+            return _strings.getAttribute('data-' + stringId);
```
alexgibson: Hmm, this isn't quite going to work as-is.

Whilst some calls to this function [use dashes](https://github.com/mozilla/bedrock/blob/master/media/js/firefox/accounts-2018.js#L25) to separate attribute names, other calls use [camelCase](https://github.com/mozilla/bedrock/blob/bb1a65b048f130a4e1baa70a1a040eb462838c55/media/js/mozorg/contact-spaces.js#L23), which for the jQuery API (and native dataset) is totally fine, but would break here as we're assuming dashes only.

Either this function should handle both cases (more complex), or we should choose one method over the other, and update all calls to it to be the same syntax. Thoughts?

stephaniehobson: Good catch. 🎣 

In a perfect world we handle both cases but in the name of expediency I'm tempted to just go with dashes and update the code base.

stephaniehobson: OTOH there is a certain irony to talking about expediency in a 3 month old PR.

stephaniehobson: Dash is way more common, converting to dashes.

PR #6577, opened 2018-12-10
[#6444] Add promotional video to concert page

craigcook:
## Description
Replaces the artist image with an embedded youtube video (using a different image as a poster). When the show is live (via the `firefox-concert-one-live` switch) the youtube video is replaced by a link to the livestream.


craigcook:
Tweaks done, should be good to go.

media/js/firefox/concerts.js
```diff
@@ -309,4 +320,68 @@
         });
     }, false);
 
+
+    // Video
+    var tag = document.createElement('script');
+    tag.src = 'https://www.youtube.com/iframe_api';
+    var firstScriptTag = document.getElementsByTagName('script')[0];
+    firstScriptTag.parentNode.insertBefore(tag, firstScriptTag);
```
jpetto: We should wrap the above in a conditional to prevent the YouTube JS from loading when it's not needed (e.g. when the concert is actually live).

Easiest way is probably to check for the existence of the `.js-video-play` element prior to executing the code above.

Not critical, but probably good to rename `firstRunOnYouTubeIframeAPIReady` to something more appropriate - like `firefoxConcertsOnYouTubeIframeAPIReady`.

craigcook: Oh good idea, I shall make it so.

jpetto:
LGTM - feel free to merge when tests finish up!

PR #6578, opened 2018-12-11
Double pytest reruns

jgmize:
Per our discussion last week, we're going to increase the number of pytest reruns to reduce the need to rerun the entire test suite on failure.

### DIFF
--- a/bin/run-integration-tests.sh
+++ b/bin/run-integration-tests.sh
@@ -20,9 +20,7 @@ CMD="${CMD} -r a"
 CMD="${CMD} --verbose"
 CMD="${CMD} -n ${PYTEST_PROCESSES}"
 CMD="${CMD} --base-url ${BASE_URL}"
-# rerun a flaky test once
-# DO NOT INCREASE THIS
-CMD="${CMD} --reruns 1"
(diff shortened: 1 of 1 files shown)


alexgibson:
I’m on PTO currently, but thought i’d leave a note that i’m not sure this has been working for some time. We already have reruns set to `1`, but from what i’ve seen it doesn’t currently retry at all.

bin/run-integration-tests.sh
```diff
@@ -20,9 +20,7 @@ CMD="${CMD} -r a"
 CMD="${CMD} --verbose"
 CMD="${CMD} -n ${PYTEST_PROCESSES}"
 CMD="${CMD} --base-url ${BASE_URL}"
-# rerun a flaky test once
-# DO NOT INCREASE THIS
-CMD="${CMD} --reruns 1"
+CMD="${CMD} --reruns 2"
```
pmac: This may explain why it's not working, according to the README for the pytest-rerunfailures package the number represents "the maximum number of times you'd like the test to run". So I think it was essentially off :(

https://github.com/pytest-dev/pytest-rerunfailures#re-run-all-failures

Issue #6583, opened 2018-12-11
Update IRL section on /about


amychurchwell:
Update the IRL section on the About Page with IRL Season imagery and updated copy.

Image is here: https://mozilla.app.box.com/file/342705897087

Copy:

In Mozilla's IRL podcast, host Manoush Zomorodi shares real stories of life online and real talk about the future of the Web.

PR #6590, opened 2018-12-11
Update IRL card on /about (#6583)

amychurchwell:
## Description
IRL card image and copy change on /about

## Issue / Bugzilla link
#6583


bedrock/mozorg/templates/mozorg/about-en.html
```diff
@@ -71,16 +71,29 @@ <h1 class="c-page-title">{{ _('Our Mission in Action') }}</h1>
           aspect_ratio='mzp-has-aspect-3-2',
           link_url='https://blog.mozilla.org/Internetcitizen/2018/06/06/the-today-show-mozilla-firefox-facebook-container/'
       )}}
+      {% if l10n_has_tag('irl_card_about_2018') %}
```
craigcook: Looks like this page hasn't been exposed for l01n yet so we don't need an l10n tag after all, you can just update the card with no fallback.

amychurchwell:
@craigcook Should be good to go. Noticed another `leaders.jpg` was also the same size as hi-res, so I resized that one as well.

Issue #6479, opened 2018-11-15
[On 12/27/2018] 2018 H1 Transparency Report


ejregithub:
Description: Add 2018/H1 Transparency report to mozilla.org with [this content](https://docs.google.com/document/d/1inLs0TpZHt2kfmaU3ODlhRWCybLl4_HhrARel5xv5W4/edit)

Success Criteria:
- [ ] Go live on 12/27/18 - put it behind a switch to facilitate this.

ejregithub:
Today I pinged Mika with the link to the demo in email asking for stakeholder review

stephaniehobson:
Blocked by sign off from legal team.

ejregithub:
@amychurchwell - please note updates from the stakeholder.  Also, note that this needs to go live in the last week of the year and not before.

ejregithub:
Mika has approved the latest updates, this is now stakeholder QA approved.

@amychurchwell @craigcook - Go live is now 12/27 - let's target flipping the switch at 9 am so it's live before noon.  ;)

PR #6599, opened 2018-12-12
Transparency Report: jan-june 2018 update ( #6479)

amychurchwell:
## Description
Updated transparency report with [current information.](https://docs.google.com/document/d/1inLs0TpZHt2kfmaU3ODlhRWCybLl4_HhrARel5xv5W4/edit#heading=h.vvyxsb7n8bru)

**Switch config:** https://github.com/mozmeao/www-config/pull/176

## Issue / Bugzilla link
 #6479 

## Testing
https://bedrock-demo-achurchwell.oregon-b.moz.works/en-US/about/policy/transparency/

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
+++ b/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
@@ -9,7 +9,7 @@
     <nav>
       <h3>{{ _('Previous Reports') }}</h3>
       <ul>
-        {% if switch('transparency_report') %}
+        {% if switch('transparency-report') %}
         <li><a href="{{ url('mozorg.about.policy.transparency.jul-dec-2017') }}">{{ _('July-December 2017') }}</a></li>
         {% endif %}
         <li><a href="{{ url('mozorg.about.policy.transparency.jan-jun-2017') }}">{{ _('January-June 2017') }}</a></li>
--- a/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
+++ b/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
@@ -9,7 +9,7 @@
     <nav>
       <h3>{{ _('Previous Reports') }}</h3>
       <ul>
-        {% if switch('transparency-report') %}
+        {% if switch('transparency_report_dec_2018') %}
         <li><a href="{{ url('mozorg.about.policy.transparency.jul-dec-2017') }}">{{ _('July-December 2017') }}</a></li>
         {% endif %}
         <li><a href="{{ url('mozorg.about.policy.transparency.jan-jun-2017') }}">{{ _('January-June 2017') }}</a></li>
--- a/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
+++ b/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
@@ -9,7 +9,9 @@
     <nav>
       <h3>{{ _('Previous Reports') }}</h3>
       <ul>
+        {% if switch('transparency_report') %}
         <li><a href="{{ url('mozorg.about.policy.transparency.jul-dec-2017') }}">{{ _('July-December 2017') }}</a></li>
+        {% endif %}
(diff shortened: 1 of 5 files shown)


craigcook:
Marking DNM for right now until we get signoff from the legal team.

amychurchwell:
Approved by stakeholder. Looking to be live on 12/27.

bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
```diff
@@ -9,6 +9,9 @@
     <nav>
       <h3>{{ _('Previous Reports') }}</h3>
       <ul>
+        {% if switch('transparency-report') %}
```
craigcook: It might be good to add a date to the switch name, just for context. Though really we should clean up the switch right after it goes live anyway so it probably doesn't matter. Up to you.

And I also think we're now supposed to favor underscores instead of dashes in switch names since that's what www-config uses, but I don't think it's documented anywhere. The config will match dashes and since this should be short lived it's still up to you if you want to change it again, no biggie.

bedrock/mozorg/templates/mozorg/about/policy/transparency/index.html
```diff
@@ -53,13 +53,21 @@ <h3 data-accordion-role="tab">{{ _('What is the scope of the Transparency Report
             <div data-accordion-role="tabpanel">
               <p>
                 {# NOTE: Remember to update these links when publishing a new report. #}
-                  {% set link_userdata=url('mozorg.about.policy.transparency.jul-dec-2017') + '#government-user-data' %}
-                  {% set link_removal=url('mozorg.about.policy.transparency.jul-dec-2017') + '#government-content-removal' %}
-                  {% set link_copytrade=url('mozorg.about.policy.transparency.jul-dec-2017') + '#copyright-trademark' %}
-                  {% set link_supplement=url('mozorg.about.policy.transparency.jul-dec-2017') + '#supplement' %}
+                  {% set link_userdata=url('mozorg.about.policy.transparency.jan-jun-2018') + '#government-user-data' %}
```
craigcook: These updated links should be behind the same switch, with the old links as the `else` fallback.

bedrock/mozorg/templates/mozorg/about/policy/transparency/jan-jun-2018.html
```diff
@@ -0,0 +1,253 @@
...
+                <a href="{{ url('mozorg.about.policy.transparency.index') + '#dfn-natl-security-request' }}">{{ _('National Security Requests') }}</a> <sup><a href="#footnote-1">1</a></sup>
+              </th>
+              <td>0</td>
+              <td>{{ _('N/A') }}</td>
+            </tr>
+          </tbody>
+        </table>
+
+        <aside class="footnotes">
+          <section id="footnote-1">
```
craigcook: Each footnote should be in its own section with its own ID, so the links can link directly to them separately. Look at the Jan-June 2017 report for an example (our only previous report with more than one footnote): https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templates/mozorg/about/policy/transparency/jan-jun-2017.html#L87

bedrock/mozorg/templates/mozorg/about/policy/transparency/jan-jun-2018.html
```diff
@@ -0,0 +1,257 @@
...
+                <a href="{{ url('mozorg.about.policy.transparency.index') + '#dfn-search-warrant' }}">{{ _('Search Warrants') }}</a>
+              </th>
+              <td>0</td>
+              <td>{{ _('N/A') }}</td>
+            </tr>
+            <tr>
+              <th scope="row">
+                <a href="{{ url('mozorg.about.policy.transparency.index') + '#dfn-subpoena' }}">{{ _('Subpoenas') }}</a>
+              </th>
+              <td>1<sup><a href="#footnote-1">2</a></sup></td>
```
craigcook: Need to update these links for the separate footnote IDs, so 1 links to `#footnote-1`, 2 links to `#footnote-2`, etc. 

And should also rearrange them in the right numerical order, so 1 for subpoenas, 2 for court orders, 3 for national security requests.

PR #6601, opened 2018-12-12
Revert "Skip integration tests because of bug 1513622"

stephaniehobson:
This reverts commit 6e558016e64cb109547f7c86dbc9ce8467762335.

### DIFF
--- a/jenkins/branches/master.yml
+++ b/jenkins/branches/master.yml
@@ -7,3 +7,12 @@ regions:
   - iowa-a
 apps:
   - bedrock-dev
+integration_tests:
+  oregon-b:
+    - firefox
+    - headless
+    - download
+  iowa-a:
+    - download
+    - firefox
+    - headless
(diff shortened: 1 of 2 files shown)


metadave:
@pmac is the process to skip integration tests documented anywhere?

pmac:
We should probably not merge this until bug 1513622 is resolved. The 404s persist as of now.

pmac:
> is the process to skip integration tests documented anywhere?

Yup https://bedrock.readthedocs.io/en/latest/pipeline.html#configuration

alexgibson:
Could we just skip the `download` tests here, any reason to keep on skipping every test we have?

pmac:
Alex has a great point. @stephaniehobson can you amend this to exclude the `- download` lines and we can get some tests running on deployment again?

pmac:
I've just been told in the bug that the downloads should be working again, and in my personally testing on the page this is true. I think we can just merge this.

Issue #6124, opened 2018-09-05
Automate updates to homepage editorial (card) content


alexgibson:
Opening this issue to track building automation to facilitate updates to the [homepage](https://www.mozilla.org/en-US/) editorial cards.

Rationale:

- Updating the cards monthly takes a lot of dev time and cross-team coordination.

Notes from meeting with @pmac, @djst and @ejregithub:

- Data could live in a separate GitHub repo in markdown format.
- Bedrock could pull in & parse markdown post-merge.
- Websites team could still code review pull requests initially, if required.
- Could possibly share common image processing tasks that we build for the leadership page in #6123.

Things to consider:

- Even though the editorial content is `en-US` only for now, we could expect l10n to become a requirement in the future for locales such as `de`.
- Need checks to ensure image and quality, size aspect ratio.

alexgibson:
Some specs:

**Card order:**

- Card 1 (large)
- Card 2 (small)
- Card 3 (small)
- Card 4 (small)
- Card 5 (small)
- Card 6 (medium)
- Card 7 (medium)
- Card 8 (medium)
- Card 9 (medium)
- Card 10 (small)
- Card 11 (small)
- Card 12 (small)
- Card 13 (small)
- Card 14 (small)
- Card 15 (small)
- Card 16 (medium)
- Card 17 (medium)

Note: cards are interspersed with other static content, such as billboard components. The page would need a way to insert / reference cards by some unique identifier, based on number/size.

I think we would need to say the order above & size of cards does not change. This is the layout.

**All cards have the following properties:**

- Thumbnail image
  - One of three aspect ratios: 1:1, 3:2, 16:9
  - Large cards
    - 16:9 only
    - 1860px width high-res, 930px width low-res
  - Medium cards
    - 1200px width high-res, 600px width low-res
  - Small cards
    - 860px width high-res, 430px width low-res
- Tag e.g. "Firefox", "Privacy"
- Title
- Description
- Link
- Icon (either video or audio)
- YouTube video ID (if provided, clicking the home page card would open a modal and inject a youtube video)

alexgibson:
@pmac @ejregithub just checking in here - is this something we need to regroup on in terms of planning work / timing? I think Eric mentioned the home page automation would be the most immediately useful thing to start work on, but I guess we need to try size this in terms of LOE and how long we think something like this may take to build.

pmac:
[...]

Issue #6585, opened 2018-12-11
DE version of en-US www homepage


djst:
| Estimates |  |
| ------------- | ------------- |
| TL | |
| PM | |
| ENG | 3 days |
| SEO | |
| ANA | |
| DES | |
| COPY | |

Description: We need the equivalent of the en-US homepage for German, meaning, the same template, same ability to curate "cards" linking to German blog posts.

From a dev point of view, this is likely fairly straightforward since we'll be reusing the en-US templates with some new configurations.

From a design point of view, we're reusing the design as is. Only changes might be the lack of a Pocket feed, but that should be minimal design work.

From a copy point of view, l10n should come from the Marketing team, meaning, Clay won't need to be engaged here. 

From a SEO point of view, there may be work to identify appropriate keywords, tags etc. 

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] This is an example of success criteria. Use a checklist!
- [ ] This is an example of success criteria.

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] This is an example of risk. Use a checklist!
- [ ] This is an example of risk.

#### Tasks
- [ ]

[Master project doc](https://docs.google.com/document/d/10wvOenVYQM7CpZIaVU5y_5TqpIz3stC093D5DM4EfHI/edit#)
- Includes link brief, seven sheets, decision log, meeting notes, etc
------

ejregithub:
Contexting session is booked for Monday Jan 28th

alexgibson:
Blocked on content

alexgibson:
@ejregithub what was the due date for delivering this content? I seem to remember the team thought they could deliver it last Friday (since they wanted it live within a week). Can you check what the status is here? Thanks

ejregithub:
@alexgibson - see here (it’s the 2/6/2019 Edition) in [Airtable](https://airtable.com/shrsePL6DurBLe9mg/tblgNmPLbDjXxeViI/viwYLv28cjNOdZMm0?date=2018-06-16&mode=month)

And here is the sheet with all the content: ([content is in tab 1 + banners are in tab 2](https://docs.google.com/spreadsheets/d/1gLv1pfj7TIuGXZyPnzi3iwdG0heYVpID1HWe4wrLT3c/edit?ts=5c5c56da#gid=1177271466))

alexgibson:
Great, picking this one up today 👍

alexgibson:
[...]

PR #6602, opened 2018-12-13
Use external data for home page content cards

pmac:
This is a WIP for the home page external admin work. @alexgibson let me know what you think of this first pass when you get some time.

Re #6124


pmac:
I still have to deal with image uploading and linking, but this is a decent start I think.

pmac:
It's working for me locally if I do the following:

* `make run`
* open a new terminal also to bedrock repo
* `docker-compose exec app bash`
* `./manage.py migrate`
* `./manage.py update_content_cards`
* then load the dev server in a browser

The image will be broken, but everything else about card 1 should work.

alexgibson:
I ran the steps above and everything worked as expected 👍 

Some thoughts on GA and l10n:

- I'm not sure we need the `ga_title` field in the external repo or in the DB. The field itself is just to store an English version of the card title for GA purposes, however in the world of this type of automation, I think it's pretty safe that this will all be in English anyway, and we can just use `title` to populate the data attribute in the template macro.
- If we do go on to support l10n in this system, I think it wil be unlikely that we would support string translations in the typical sense like we do on bedrock. That kind of thing will be for evergreen content, and not something in www-admin. 
- What I do think we need to support however, is for specific locales to have their own content. For example we do want to build a German version of the home page, which will have its own curated content and links to German blog posts (See https://github.com/mozilla/bedrock/issues/6585).

What we have here so far looks great, let me know if you would like to chat this week about the above. Thanks @pmac!

pmac:
Great! I've basically made it open ended. I think it's just a convenience to have all locales of a specific card inherit the attributes from the en-US one so that it's easier to update them all if that's what you need to do (change the image e.g.), but if that feels over-engineered and potentially error prone I'm happy to remove it. The way it basically works now is that you can provide data for any argument to the `card()` macro and it'll pass through. Some are handled in a special way (e.g. just saying `aspect_ratio: 16-9` and have that become the right class name) but basically that's it.

And yes, I'd love to have a chat this week to further sync about the status and future. Thanks for the review!

pmac:
Changed recently:

* Added tests for importing the card data and the model functions.
* Improved how the template functions work to allow it to get the cards from the context or to be passed in.
* Fixed ability to provide either a full URL or URL name in the data

bedrock/base/templates/macros-protocol.html
```diff
@@ -95,11 +95,15 @@
 
 
 {# Card: https://protocol.mozilla.org/patterns/molecules/card.html #}
-{% macro card(title, ga_title, image_url, link_url, desc=None, meta=None, cta=None, tag_label=None, media_icon=None, class=None, heading_level=2, aspect_ratio='mzp-has-aspect-3-2', include_highres_image=False, youtube_id=None) -%}
+{% macro card(title, ga_title, image_url, link_url, desc=None, meta=None, cta=None, tag_label=None, media_icon=None, class=None, heading_level=2, aspect_ratio='mzp-has-aspect-3-2', include_highres_image=False, highres_image_url=None, youtube_id=None) -%}
 <section class="mzp-c-card {% if class %}{{ class }}{% endif %} {% if aspect_ratio %}{{ aspect_ratio }}{% endif %} {% if media_icon %}{{ media_icon }}{% endif %} {% if youtube_id %}has-video-embed{% endif %}">
   <a class="mzp-c-card-block-link" href="{{ link_url }}" data-link-name="{{ ga_title }}" data-link-type="link" data-link-group="card">
     <div class="mzp-c-card-media-wrapper">
-      {{ lazy_img(image_url=image_url, include_highres_image=include_highres_image, placeholder_url='placeholder.png', optional_attributes={'class': 'mzp-c-card-image'}) }}
+      {% if highres_image_url %}
```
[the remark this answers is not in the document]
pmac: Comment added. LMK if what I wrote is clear and helpful enough. Good suggestion 👍

.env-dist
```diff
@@ -2,3 +2,5 @@ DEBUG=True
 DEV=True
 ALLOWED_HOSTS=*
 SWITCH_NEWSLETTER_MAINTENANCE_MODE=False
+CSP_DEFAULT_SRC='self',*.mozilla.net,*.mozilla.org,*.mozilla.com,*.s3.amazonaws.com
+CONTENT_CARDS_URL=http://bedrock-dev-media.s3.amazonaws.com/media/
```
alexgibson: Do we need to add this variable for demo deployments?

pmac: oooooooooo... good point. I think we should point demos to https://www-dev.allizom.org/media/. Should probably use that both places... I'll do that. Good catch.

.env-dist
```diff
@@ -2,3 +2,5 @@ DEBUG=True
 DEV=True
 ALLOWED_HOSTS=*
 SWITCH_NEWSLETTER_MAINTENANCE_MODE=False
+CSP_DEFAULT_SRC='self',*.mozilla.net,*.mozilla.org,*.mozilla.com,*.allizom.org
```
alexgibson: When I run the site locally (not using Docker), this line seems to prevent all assets from loading? I also see things like `Content Security Policy: Couldn't parse invalid host self'` in the web console.

Things seems to be fine when using Docker, which is odd.

docs/content-cards.rst
```diff
@@ -0,0 +1,69 @@
...
+        ctx = {'page_content_cards': get_page_content_cards('home', locale)}
+        return l10n_utils.render(request, 'sweet-words.html', ctx)
+
+The ``get_page_content_cards`` returns a dict of card data dicts for the given page (``home`` in this case)
+and locale. The dict keys are the names of the cards (e.g. ``card_1``). If the ``page_content_cards`` context
+variable is available in the template, then the ``content_card()`` macro will discover it automatically.
+
+.. note::
+
+    The ``get_page_content_cards`` function is not all that clever as far as l10n is concerned. If you have
```
alexgibson: > you should have cards for every locale for which the page is active or the function will return an empty dict.

I'm not sure if having cards translated for every locale a page is active for is something we can realistically try to adhere to? Locales can be activated at any time, independent of changes to either bedrock or the www-admin repo. The warning you have here is great, but should we also be strongly encouraging developers to be explicit when passing `locale` into `get_page_content_cards`?

pmac: It's a tough question/problem. Are we prepared to say that a template using these should only be locale specific, and then hardcode? That's fine, but a bit limiting. Perhaps we should recommend wrapping the cards section of the template in a language specification at least? 

```jinja
<% if LANG in ['de', 'fr', pt-BR'] %>
{# do cards #}
<% endif %>
```

Or decide in the view and only do cards if the `page_content_cards` variable is truthy?

.env-dist
pmac: @alexgibson I believe I've fixed this. It was an issue with the quotes around `'self'` and getting that out of the environment. Our library to do that strips quotes. So I moved to only neededing to provide the "extra" items to add to the `CSP_DEFAULT_SRC` list. This seems better anyway.

docs/content-cards.rst
pmac: I updated the docs with an attempt. LMK what you think.

pmac:
@alexgibson I believe I've addressed the things you've found. I've also rebased on current master and pushed my demo again. I also just updated the data in the www-admin repo with the new cards on the current home page. I need to squarsh, but other than that we might be good to go here.

pmac:
demo is up https://bedrock-demo-pmac.oregon-b.moz.works/en-US/

.env-dist
```diff
@@ -2,3 +2,5 @@ DEBUG=True
 DEV=True
 ALLOWED_HOSTS=*
 SWITCH_NEWSLETTER_MAINTENANCE_MODE=False
+CSP_DEFAULT_SRC=*.allizom.org
```
alexgibson: Hmm, when running locally (not using Docker) I still see CSP issues e.g.

```
Content Security Policy: The page's settings blocked the loading of a resource at http://localhost:8000/media/js/BUNDLES/site.js ("script-src").
```

docs/content-cards.rst
alexgibson: Works for me 👍

.env-dist
pmac: @alexgibson it's working for me. I honestly have no idea. Does the `CSP_DEFAULT_SRC` setting in your `.env` file look exactly like this line?

alexgibson: I tried this again with a fresh `.env` and it worked both locally and using docker. Not sure what I did wrong the first time, but I can't seem to replicate this error any longer? :)

PR #6611, opened 2018-12-14
[fix #6610] Add query params to FxA sign in links.

jpetto:
## Description

Add query params to FxA sign in links, specifically `service=sync` so the user is actually signed in to Sync.

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/6610

## Testing

Ensure querystring for `accounts-2018.html` is accurate/good enough for now.

Ensure adding this querystring won't negatively impact any user flow (e.g. user's not using Firefox).

PR #6612, opened 2018-12-14
Revert "[fix #6610] Add query params to FxA sign in links."

### LINKED ISSUE #6610: FxA: all https://accounts.firefox.com/signin buttons should sign the user in to Sync
We found some issues where the "Login" button, it is missing crucial query params. The "Sign up" gets the user into Sync. However if they press "Log in" then they would get into accounts, but get no Sync. 

Ref: https://github.com/mozilla/bedrock/blob/1a93087b557dbbc2d10e849b44e52ec538195c54/bedrock/firefox/templates/firefox/accounts-2018.html#L215

Ref: https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/accounts-2018.html#L215

Ref: https://github.com/mozilla/bedrock/pull/6554/files#r241784410

jpetto:
Reverts mozilla/bedrock#6611

Extra `?` in there. :(


craigcook:
Do we really need to revert the whole thing, or just remove the extra `?` ?

Issue #6586, opened 2018-12-11
Win 10 funnelcake test on /new


djst:
Alex has understanding here and we still Daniel Z’s final input on traffic cop allocations

 - an important request from product

djst:
Given that @alexgibson is on PTO, I think this one is blocked at least until @ejregithub is back.

alexgibson:
The only context I have on this request is what’s in the bug, so I don’t have any additional background. If it’s urgent, I wouldn’t call myself a blocker here until I return.

djst:
Thanks @alexgibson. Unassigning you so someone else on the team can pick this up.

craigcook:
I'll see what I can do on this, if I can parse the info in the bug.

craigcook:
(This info is also in the bug, just noting it again here for posterity):

> Context: We're investigating the impact that placing a Firefox shortcut on the Windows 10 taskbar would have on engagement and retention. The hypothesis being that users find it more convenient to open Firefox from the taskbar in a context where pinning shortcuts to the taskbar is not something most users know how to do.

There are two Funnelcake builds; 138 is the control and 139 has the taskbar pinning. This is targeting only users on Windows 10 and only in the en-US locale. Since there are only two builds and one is already set up as a control, it will be a simple 50/50 split in Traffic Cop.

craigcook:
Update: based on expected download rates the sample will be a 9%/9% split between the two funnelcakes and the remaining 82% getting the normal build.

We'll run the test for two weeks so as to avoid needing to turn it off during the holiday break, starting on 20 December (Thursday) and ending on 3 January (two Thursdays hence).

The code merged in https://github.com/mozilla/bedrock/commit/2ecc2a883d0456acd4ea468df6339e0c6520c0d6

PR #6618, opened 2018-12-14
[#6586] Add win10 taskbar pinning funnelcake experiment

craigcook:
## Description
Adds a funnelcake experiment to /new targeting Windows 10 and en-US, with a 50/50 split between funnelcake builds 138 (control) and 139 (the experiment).

## Issue / Bugzilla link
#6586
https://bugzilla.mozilla.org/show_bug.cgi?id=1493595

### DIFF
--- a/media/js/firefox/new/experiment-win10-taskbar-funnelcake.js
+++ b/media/js/firefox/new/experiment-win10-taskbar-funnelcake.js
@@ -1,0 +1,27 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function(Mozilla) {
+    'use strict';
+
+    var ua = navigator.userAgent;
+    var isLikeFirefox = /Iceweasel|IceCat|SeaMonkey|Camino|like\ Firefox/i.test(ua);
+    var isFirefox = /\s(Firefox|FxiOS)/.test(ua) && !isLikeFirefox;
+
+    var isWin10 = (ua.indexOf('Windows NT 10') > -1);
+    var isWin64 = (ua.indexOf('WOW64') > -1 || ua.indexOf('Win64') > -1);
+
+    if (window.site.platform === 'windows' && isWin10 && isWin64 && !isFirefox) {
+        var mclane = new Mozilla.TrafficCop({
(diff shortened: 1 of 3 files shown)


craigcook:
Awaiting confirmation on the percentages so DNM for now.

craigcook:
Update: based on expected download rates the sample will be a 9%/9% split between the two funnelcakes and the remaining 82% getting the normal build. 

We'll run the test for two weeks so as to avoid needing to turn it off during the holiday break, starting on 20 December (Thursday) and ending on 3 January (two Thursdays hence).

media/js/firefox/new/experiment-win10-taskbar-funnelcake.js
```diff
@@ -0,0 +1,28 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function(Mozilla) {
+    'use strict';
+
+    var ua = navigator.userAgent;
+    var isLikeFirefox = /Iceweasel|IceCat|SeaMonkey|Camino|like\ Firefox/i.test(ua);
+    var isFirefox = /\s(Firefox|FxiOS)/.test(ua) && !isLikeFirefox;
+    var isMobile = /^(android|ios|fxos)$/.test(window.site.platform);
```
jpetto: I'm not sure we need this check (and below), as we're already checking that `window.site.platform` is exactly `windows` - which infers *not* mobile...right?

craigcook: good catch, will remove

media/js/firefox/new/experiment-win10-taskbar-funnelcake.js
```diff
@@ -0,0 +1,27 @@
...
+    if (window.site.platform === 'windows' && isWin10 && isWin64 && !isFirefox) {
+        var mclane = new Mozilla.TrafficCop({
+            id: 'experiment_win10_taskbar_funnelcake',
+            variations: {
+                'f=138': 9, // control build
+                'f=139': 9  // taskbar build
+            }
+        });
+
+        mclane.init();
```
jpetto: Best xmas movie. (Yes, better than Gremlins.)

craigcook: Agreed. 🏢 💥 🚁

Issue #6619, opened 2018-12-14
New Firefox Accounts l10n page missed a few strings 


peiying2:
L10n string extraction missed a few strings and string updates between PR #6503 commits. 

Refer to bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1514343

Some of the strings were hard coded in this commit: https://github.com/mozilla/bedrock/pull/6503/commits/00dbad57ace2d0ed0ff0b7742422a218bc28b8dc 

11 days later, they were tagged correctly for l10n and two strings were revised too: https://github.com/mozilla/bedrock/pull/6503/commits/e7f7dbf7a21199a6f31f0ca923dabbca71cf1f1d. 

Somehow, when I extracted the strings (see the conversations and timeline), I got the copy from the old version with the new file name. How can we fix this? 

@craigcook @amychurchwell

craigcook:
Sorry, that whole pull request got a bit messy. I'm not sure how best to clean it up at this point. Can we just re-extract strings from the final template? Would that undo all the previous l10n work?

peiying2:
> Sorry, that whole pull request got a bit messy. I'm not sure how best to clean it up at this point. Can we just re-extract strings from the final template? Would that undo all the previous l10n work?

I extracted the file today to confirm what the bug listed is correct. For the "hard coded" strings, adding the new strings make no difference (they are  exposed in English now already). For the updated strings, if we replace them with new strings (Scan...), which means more English strings on the site until the communities with activated pages catch up again.

craigcook:
For those strings that changed between commits (the "Scan..." ones), we can at least wrap them in a tag with the old form as the fallback. I'll do that.

craigcook:
We added the tag `accounts_2018_qrcodes` in #6620 for the two strings that changed but alas there's nothing we can do for the ones that were hardcoded in the original extraction. Sorry for the headaches.

PR #6620, opened 2018-12-15
6619 accounts strings

craigcook:
## Description
Two strings on this page changed after the initial extraction, so locales have translated the old versions but don't match the final text. This adds the l10n tag `accounts_2018_qrcodes` with the old strings as the fallback until active locales can update.

## Issue / Bugzilla link
#6619 
https://bugzilla.mozilla.org/show_bug.cgi?id=1514343


craigcook:
@peiying2

PR #6627, opened 2018-12-17
Upload static media to s3 for www-dev

pmac:
This adds pushing media to a dev bucket where we were only uploading to a stage and prod bucket during prod pushes.

### DIFF
--- a/bin/upload-staticfiles.sh
+++ b/bin/upload-staticfiles.sh
@@ -4,7 +4,11 @@ set -exo pipefail
 
 source docker/bin/set_git_env_vars.sh
 
-if [[ "$BRANCH_NAME" != "prod" ]]; then
+if [[ "$BRANCH_NAME" == "prod" ]]; then
+    BUCKETS=(stage prod)
+elif [[ "$BRANCH_NAME" == "master" ]]; then
+    BUCKETS=(dev)
+else
     # nothing to do if not prod deployment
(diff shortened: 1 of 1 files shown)


pmac:
The new bucket was created in https://github.com/mozmeao/infra/pull/990

pmac:
This is very similar to the technique I'm using in a [similar file at the www-admin repo](https://github.com/mozmeao/www-admin/blob/master/run.sh) to upload those files to s3.

bin/upload-staticfiles.sh
```diff
@@ -4,7 +4,11 @@ set -exo pipefail
 
 source docker/bin/set_git_env_vars.sh
 
-if [[ "$BRANCH_NAME" != "prod" ]]; then
+if [[ "$BRANCH_NAME" == "prod" ]]; then
+    BUCKETS=(stage prod)
+elif [[ "$BRANCH_NAME" == "master" ]]; then
+    BUCKETS=(dev)
```
jgmize: What would be the consequences of pushing to dev and stage on the master branch and only updating the prod bucket on push to prod?

pmac: We will be doing that in a future PR when we implement the new CI pipeline we discussed at Mozlando. This plus a number of other things about the deployment scripts will need to change.

Issue #6471, opened 2018-11-13
Update "About" card promo in main navigation


alexgibson:
The card under "About" in the main navigation is promoting Mozfest, which happened in October.
<img width="340" alt="screen shot 2018-11-13 at 15 12 18" src="https://user-images.githubusercontent.com/400117/48422451-98b4f580-e756-11e8-81d3-7398bb16601a.png">
We should probably update it with something newer and more evergreen.

djst:
I like the idea of having something evergreen as fallback but think we could also view this as a place for more recent features, e.g. in December we could use it for the end of year fundraising, etc. 

Side-note: this would be a great candidate for automation, i.e. a markdown file we'd update following a basic format (heading, text, image, link).

alexgibson:
Thanks @dzingeek 👍

PR #6642, opened 2018-12-20
Localize Protocol navigation (Fixes #6498)

### LINKED ISSUE #6498: Localize new navigation
Once we remove the legacy global nav from bedrock, we should be ready to start localizing the new navigation so we can roll out to all locales.

alexgibson:
## Description
- [x] Enables new navigation for all locales.
- [x] Removes legacy l10n navigation from bedrock 🎉 💥 
- [x] Adds functional tests for new navigation.

~**Strings are not yet final, will update when ready**~

## Issue / Bugzilla link
#6498

## Testing
Demo server: https://bedrock-demo-navigation.oregon-b.moz.works/


alexgibson:
Final strings are dependent on https://github.com/mozilla/bedrock/issues/6471

alexgibson:
Ok, strings for the navigation are now final and are updated on demo above.

@peiying2 we'd like to please localize the new navigation here, so that all locales can finally use it. I understand we can't use any switches here as this is a global change, so we're fine for this to sit and wait until we have enough coverage.

The new strings will likely need to be added to `main.lang`, and can be found in the template files below:

- https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/includes/protocol/navigation/index.html
- https://github.com/mozilla/bedrock/tree/master/bedrock/base/templates/includes/protocol/navigation/menu

Any questions just let me know, thanks!

peiying2:
@alexgibson Thanks for pointing me to the new templates for the new strings. There are lots of new strings if we put them all in the main.lang file. Even after I remove the duplicates, there are still a lot of strings to localize. Here is the current completion rate: https://l10n.mozilla-community.org/langchecker/?locale=all&website=0&file=main.lang - 50 locales (about 97% of the market). 

If we add these new strings to main.lang, we need to refer to the web dashboard for status to enable the 100% completed locales. This also assumes we don't add new strings unrelated to the navigation update in the next month or two. 

If we create new pages based on the template names, we can activate them as soon as a page is completed. Obviously we want all the pages to be completed before enabling them all as a set, like we did for Firefox Quantum launch, so we certainly have a different kind of challenge to manage the activation. 

I also want to point out that a lot products are listed. Some of the products are only available in certain markets, and some are localized in only a few languages, so the user experience could be different depending on locales.

flodolo:
Named templates and activation won't work, because these files are shared.

I would strongly suggest to create a new shared file (e.g. navigation.lang), and only use it for navigation strings.

As far as I remember, the only change needed in Bedrock [should be here](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L237). On our side, it needs to be added as a normal file, and also as a "non activated" file [here](https://github.com/mozilla-l10n/langchecker/blob/master/app/config/sources.inc.php#L3-L9).

alexgibson:
> I also want to point out that a lot products are listed. Some of the products are only available in certain markets, and some are localized in only a few languages, so the user experience could be different depending on locales.

Yeah, we understand that not all pages linked in the navigation are necessarily translated for all languages, but it's important for mozorg to have a consistent navigation experience across all locales. We have seen a significant lift in downloads since the new navgation was added to English locales, so doing this could have some real impact site wide. If we need to do any locale-specific tailoring of the navigation links then we can do so later on an as-needed basis, but I think that will likely be minimal.

> I would strongly suggest to create a new shared file (e.g. navigation.lang), and only use it for navigation strings.

Thanks that's a good suggestion. I have added `navigation` to `DOTLANG_FILES` in `base.py`.

peiying2:
@alexgibson https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/includes/protocol/navigation/menu/firefox.html#L17 - Mac OS -> macOS

alexgibson:
@peiying2 thank you, updated.

alexgibson:
Sucessful integration test run: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/229/pipeline

alexgibson:
I think this is close enough to be ready for code review. Whilst tests are passing, a good amount of testing and exploration on demo would be appreciated here.

peiying2:
@alexgibson a few localizers reported string fallback logic issues. They noticed that the identical English strings they translated in both navigation.lang and previous shared files, the strings in previous shared files take priority as a fallback string. In other words, the strings they translated (differently) in navigation.lang don't show up on the test server. Instead, they come from main.lang and download_button.lang.

bedrock/settings/base.py
```diff
@@ -234,7 +234,7 @@ def lazy_langs():
 FEED_CACHE = 3900
 DOTLANG_CACHE = 600
 
-DOTLANG_FILES = ['main', 'download_button']
+DOTLANG_FILES = ['main', 'download_button', 'navigation']
```
flodolo: Can we try setting this to `'navigation', 'download_button', 'main'`?

main.lang has a ton of strings, my hope is that we use the ones in navigation first like this

alexgibson:
@peiying2 @flodolo thanks. Made the suggested change and pushed it to demo. Please let me know if there's anything else I can do to help here.

alexgibson:
@peiying2 hi just checking in here - how are we doing in terms of locale coverage?

craigcook:
Heads up: there's been a request to change a bunch of strings in [bug 1522698](https://bugzilla.mozilla.org/show_bug.cgi?id=1522698). Not cause for panic just yet, but we should discuss how/if we can/should do this. 

I also have mild concerns about the UX impact of changing "Projects" to "Labs" since the Projects label was arrived at after a significant amount of research and testing. @dzingeek may have thoughts?

alexgibson:
@craigcook @dzingeek let's not let that bug disrupt this work please. Agree I think we need to think about that request and decide what we should make changes to and when. The current strings have been undergoing translation for several weeks, and I don't want to cause the l10n team any more disruption. This change has taken us several years to get to also, so I don't want to disrupt it now. Let's try and move this discussion to that bug, so as not to sidetrack the work here.

alexgibson:
Rebased after https://github.com/mozilla/bedrock/pull/6695 merged.

PR #6644, opened 2018-12-21
Add concert video to Firefox Concerts page

craigcook:
## Description
Adds a clip from the livestream to the page, moves the "TBA" copy to a separate block, and brings back the accounts illustration so as not to repeat the concert image. Also cleaning up the no-longer-needed switches.


bedrock/firefox/templates/firefox/concerts.html
```diff
@@ -50,40 +36,33 @@
...
+            <p class="mzp-c-card-tag ended">Dec. 11, 2018</p>
+            <h2 class="mzp-c-card-feature-title">Phosphorescent</h2>
+            <p class="mzp-c-card-feature-desc">On December 11 the Firefox Concert Series livestreamed one of the last shows on Phosphorescent’s <cite lang="fr">C’est La Vie</cite> tour from DC’s 9:30 Club. Enjoy this exclusive clip.</p>
+          </div>
+        </div>
+      </section>
+    </div>
+
+    <div class="artist-next mzp-l-content l-content-narrow">
+      <section class="artist-card mzp-c-card-feature mzp-has-aspect-16-9 mzp-l-card-feature-right-half mzp-t-firefox mzp-t-dark">
```
alexgibson: Not a blocker as it wasn't introduced in this PR, but we should try and avoid creating components using the Protocol naming conventions that don't exist imho. These should likely drop the `mzp-` namespace.

alexgibson: Oh, brain malfunction! I'm totally mistaken here and this does exist. For some reason I thought the feature card had a different naming convention. I was confused as this wasn't using the feature card macro, so I assumed it was something custom. My bad!

craigcook: Yeah, sorry, I didn't use the macro because... some reason... there was some piece of it that did need to be custom but now I forget what it was. The countdown maybe. But it's a slightly modified feature card.

PR #6675, opened 2019-01-04
Separate stage from prod deployment

jgmize:
## Description

This will speed up deployment times by allowing deployment and testing against dev and stage deployments in parallel, and allowing immediate deployment to prod branch if needed.

## Issue / Bugzilla link

https://github.com/mozmeao/infra/issues/981

## Testing

Push this to the stage branch, look for notification of job starting in #www-notifications in Mozilla slack with link to Jenkins job

### DIFF
--- a/docs/pipeline.rst
+++ b/docs/pipeline.rst
@@ -44,17 +44,24 @@ and the full suite of headless and UI tests are then run against
 Firefox on Windows 10 using `Sauce Labs`_. This is handled by the pipeline, and is subject
 to change according to the settings in the `master.yml file`_ in the repository.
 
+Push to stage branch
+~~~~~~~~~~~~~~~~~~~~~
+
+Whenever a change is pushed to the stage branch, a production docker image is built, published to
+`Docker Hub`_, and deployed to a `public staging environment`_. Once the new image is deployed, the
+full suite of UI tests is run against it with Firefox, Chrome, and Internet Explorer on
+Windows 10, and the sanity suite is run with IE9.
+
 .. _tagged-commit:
 
 Push to prod branch (tagged)
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-When a tagged commit is pushed to the prod branch, everything that happens during a
-untagged push is still run. In addition, the full suite of UI tests is run against
-Chrome and Internet Explorer on Windows 10, and the sanity suite is run against older
-versions of Internet Explorer (currently IE6 & IE7). If successful, the change is
-pushed to staging, tested, and then to production and the same tests are then run against
-production. As with untagged pushes, this is all handled by the pipeline, and is subject
+When a tagged commit is pushed to the prod branch, a production docker image is built and published
+to `Docker Hub`_ if needed (usually this will have already happened as a result of a push to the stage branch),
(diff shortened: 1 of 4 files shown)


pmac:
What I remember discussing is deploying to dev and stage on a push to `master`. Is there a reason we're not doing that?

jgmize:
> What I remember discussing is deploying to dev and stage on a push to master. Is there a reason we're not doing that?

One of the advantages of this approach is that a dev can push to stage independently of a push to master, and the stage deployment can run in parallel to the master deployment, unlike in #6676.

jpetto:
My preference would be for this PR over #6676, but I do think we should get consensus from the moz.org team before going forward.

jgmize:
@alexgibson @craigcook @amychurchwell any thoughts on staging deployments and tests being triggered by a separate `git push origin master:stage` step which can run in parallel to the dev deployments and tests? Note that the stage push would also become an optional (although highly recommended) step with this change, and other branches could be more easily deployed to stage without going to master or prod if needed.

docs/pipeline.rst
```diff
@@ -44,13 +44,23 @@ and the full suite of headless and UI tests are then run against
 Firefox on Windows 10 using `Sauce Labs`_. This is handled by the pipeline, and is subject
 to change according to the settings in the `master.yml file`_ in the repository.
 
+Push to stage branch
+~~~~~~~~~~~~~~~~~~~~~
+
+Whenever a change is pushed to the stage branch, a production docker image is built and
+published, and the full suite of UI tests is run against Chrome and Internet Explorer on
+Windows 10, and the sanity suite is run against older versions of Internet Explorer (currently IE6 & IE7).
```
alexgibson: Sanity tests are now run against IE8.

docs/pipeline.rst
```diff
@@ -44,13 +44,23 @@ and the full suite of headless and UI tests are then run against
 Firefox on Windows 10 using `Sauce Labs`_. This is handled by the pipeline, and is subject
 to change according to the settings in the `master.yml file`_ in the repository.
 
+Push to stage branch
+~~~~~~~~~~~~~~~~~~~~~
+
+Whenever a change is pushed to the stage branch, a production docker image is built and
+published, and the full suite of UI tests is run against Chrome and Internet Explorer on
```
alexgibson: Also run against Firefox.

jpetto:
It's a small thing, but if we go this route, might be worth adding a confirmation step to the [tag-release.sh](https://github.com/mozilla/bedrock/blob/master/bin/tag-release.sh) script - something like "Did you test on stage first? If not, are you sure you want to skip going to stage?".

docs/pipeline.rst
jgmize: now IE9 as of https://github.com/mozilla/bedrock/pull/6696

jgmize:
Thanks for the early feedback; I've addressed your comments and taken off the WIP label, and pushed this branch to origin for testing. You can see the notifications in #www-notify on slack, and an example pipeline here: https://ci.us-west.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/stage/2/pipeline

PR #6715, opened 2019-01-15
Update whatsnew test from send_to_device to send_yourself

amychurchwell:
## Description
Updates whatsnew test to test the functionality of the `send_yourself` widget.

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -268,7 +268,9 @@
           <p class="mzp-c-sendyourself-field">
             <label for="{{ id }}-input" class="mzp-js-input-label" data-alt="{{ _('Enter your email or phone number') }}">{{ _('Enter your email') }}</label>
             <input id="{{ id }}-input" class="mzp-c-sendyourself-input" name="phone-or-email" type="text" required>
-            <button type="submit" class="mzp-c-button">{{ _('Send') }}</button>
+            <div class="form-submit">
+              <button type="submit" class="mzp-c-button {{ button_theme }}">{{ _('Send') }}</button>
+            </div>
           </p>
           <p class="mzp-c-sendyourself-note mzp-is-sms">
             {{ _('SMS service available in select countries only. SMS &amp; data rates may apply.') }}
--- a/tests/functional/firefox/test_whatsnew.py
+++ b/tests/functional/firefox/test_whatsnew.py
@@ -29,10 +29,10 @@ def test_send_to_device_fails_when_missing_required_fields(base_url, selenium):
 
 @pytest.mark.skip_if_not_firefox(reason='Whatsnew pages are shown to Firefox only.')
 @pytest.mark.nondestructive
-def test_firefox_rocket_qr_code(base_url, selenium):
+def test_firefox_rocket_send_yourself(base_url, selenium):
     page = FirefoxWhatsNewPage(selenium, base_url, locale='id').open()
-    assert not page.send_to_device.is_displayed
-    assert page.is_qr_code_displayed
+    assert not page.is_qr_code_displayed
+    assert page.send_yourself.is_displayed
 
 
 @pytest.mark.skip_if_not_firefox(reason='Whatsnew pages are shown to Firefox only.')
--- a/tests/pages/firefox/whatsnew.py
+++ b/tests/pages/firefox/whatsnew.py
@@ -5,7 +5,7 @@
 from selenium.webdriver.common.by import By
 
 from pages.firefox.base import FirefoxBasePage
-from pages.regions.send_to_device import SendToDevice
+from pages.regions.send_yourself import SendYourself
 
 
 class FirefoxWhatsNewPage(FirefoxBasePage):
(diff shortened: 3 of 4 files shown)


tests/pages/firefox/whatsnew.py
```diff
@@ -5,7 +5,7 @@
 from selenium.webdriver.common.by import By
 
 from pages.firefox.base import FirefoxBasePage
-from pages.regions.send_to_device import SendToDevice
```
alexgibson: When running the whatsnew tests locally, tests are failing with the following error:

```
E       AttributeError: 'FirefoxWhatsNewPage' object has no attribute 'send_to_device'
```

This is becuase [this test](https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/test_whatsnew.py#L13) is still using the old send to device widget, and it expects the `SendToDevice` object to be present for the test to run.

This PR doesn't need to replace `SendToDevice` with `SendYourself` just yet, since we haven't deprecated the old widget. All you should need here is add `SendYourself` and then use that in the [Indonesia whats new test](https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/test_whatsnew.py#L32).

tests/pages/firefox/whatsnew.py
```diff
@@ -16,13 +16,5 @@ class FirefoxWhatsNewPage(FirefoxBasePage):
     _zh_tw_qr_code_locator = (By.CSS_SELECTOR, 'img.qrcode')
 
     @property
-    def send_to_device(self):
```
alexgibson: This is also still needed.

tests/pages/firefox/whatsnew.py
```diff
@@ -16,13 +16,5 @@ class FirefoxWhatsNewPage(FirefoxBasePage):
...
-        return SendToDevice(self)
-
-    @property
-    def is_qr_code_displayed(self):
-        return self.is_element_displayed(*self._qr_code_locator)
-
-    @property
-    def is_zh_tw_qr_code_displayed(self):
```
alexgibson: This selector is needed for the `zh-TW` locale test.

bedrock/base/templates/macros.html
```diff
@@ -268,7 +268,9 @@ <h3 class="mzp-c-sendyourself-title">
           <p class="mzp-c-sendyourself-field">
             <label for="{{ id }}-input" class="mzp-js-input-label" data-alt="{{ _('Enter your email or phone number') }}">{{ _('Enter your email') }}</label>
             <input id="{{ id }}-input" class="mzp-c-sendyourself-input" name="phone-or-email" type="text" required>
-            <button type="submit" class="mzp-c-button">{{ _('Send') }}</button>
+            <div class="form-submit">
```
alexgibson: Is this change to the HTML needed for the test to pass? You might be able to use a selector like `.mzp-c-sendyourself-field button[type="submit"]` without changing the markup perhaps.

bedrock/base/templates/macros.html
```diff
@@ -268,7 +268,9 @@ <h3 class="mzp-c-sendyourself-title">
...
+              <button type="submit" class="mzp-c-button {{ button_theme }}">{{ _('Send') }}</button>
```
alexgibson: Is `{{ button_theme }}`` related to this PR?

tests/functional/firefox/test_whatsnew.py
```diff
@@ -29,10 +29,10 @@ def test_send_to_device_fails_when_missing_required_fields(base_url, selenium):
 
 @pytest.mark.skip_if_not_firefox(reason='Whatsnew pages are shown to Firefox only.')
 @pytest.mark.nondestructive
-def test_firefox_rocket_qr_code(base_url, selenium):
+def test_firefox_rocket_send_yourself(base_url, selenium):
     page = FirefoxWhatsNewPage(selenium, base_url, locale='id').open()
-    assert not page.send_to_device.is_displayed
-    assert page.is_qr_code_displayed
+    assert not page.is_qr_code_displayed
+    assert page.send_yourself.is_displayed
```
alexgibson: As this page no longer features a QR code, I think we should change the test to test the functionality of the form, rather than just assert that it is displayed. You can see an example of this in the [existing whatsnew page test](https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/test_whatsnew.py#L16-L19). You should just need to change `sent_to_device` to `send_yourself`.

tests/functional/firefox/test_whatsnew.py
```diff
@@ -29,10 +29,10 @@ def test_send_to_device_fails_when_missing_required_fields(base_url, selenium):
 
 @pytest.mark.skip_if_not_firefox(reason='Whatsnew pages are shown to Firefox only.')
 @pytest.mark.nondestructive
-def test_firefox_rocket_qr_code(base_url, selenium):
+def test_firefox_rocket_send_yourself(base_url, selenium):
     page = FirefoxWhatsNewPage(selenium, base_url, locale='id').open()
-    assert not page.send_to_device.is_displayed
-    assert page.is_qr_code_displayed
+    assert not page.is_qr_code_displayed
```
alexgibson: This QR code no longer exists in the template, so we can likely remove this assets and it's selector.

amychurchwell:
@alexgibson I agree! I will put this in the page's PR and close this. Just pushed the tests now.

Issue #6718, opened 2019-01-16
Lazy load images on /firefox


stephaniehobson:
To help the /firefox landing page load faster let's lazy load some of the image assets.

alexgibson:
This was fixed

PR #6719, opened 2019-01-16
Lazyload images (#6718)

stephaniehobson:
## Description
Switch from high res helper to lazy loader helper.
Add lazy loader to js bundle and init.

## Issue / Bugzilla link
#6718 

## Testing
- Are the images still there?
- In both high and low res?


alexgibson:
This PR looks great, more performance wins! ⚡️ 

Whilst it's hard to test locally because images load locally from disk and appear quickly, have we thought about content jumping when images lazy load more slowly? Pushing this to demo and loading on a throttled connection might be worth trying, as UX can sometimes be impacted.

For the home page cards that lazy load, we used some CSS to preserve the spacing that images would occupy before they load (using their [aspect ratio to calculate](https://github.com/mozilla/protocol/blob/master/src/assets/sass/protocol/includes/_mixins.scss#L210-L227)), in order to minimize content jumping.

stephaniehobson:
It's on [demo here](https://bedrock-demo-shobson.oregon-b.moz.works/en-US/firefox/).

It's frustrating that the CSS does not respect the height and width declared on the image tag when calculating proportions 😠 I made the spacer gif 16:9 so the placeholder has the correct aspect ratio on desktop and there's no content jumping there.

Forgot to test mobile though and that does have a problem.

alexgibson:
@stephaniehobson I don't see lazy-loading happening on that demo URL. Is there perhaps another branch pushed to that instance?

stephaniehobson:
Ooops, yeah that's the screen shot branch now 😖

stephaniehobson:
Updated to add fade in and lazy load `/media/img/firefox/home/new-tab.png`.

docs/mozilla-lazy-load.rst
```diff
@@ -72,4 +72,6 @@ By default the lazy load plugin will simply swap in images as they load. To add
 
 This works because the plugin removes the ``data-src`` attribute as each image lazy loads.
 
+These styles are also available in `/media/css/base/mozilla-lazy-load.scss` if you want to include that in the static bundle for the page.
```
alexgibson: 👍

PR #6278, opened 2018-10-05
Port privacy notice pages to Protocol (#6198)

alexgibson:
## Description
- Ports privacy notice page to Protocol.
- Removed accordion that hides content.

## Issue / Bugzilla link
#6198

## Testing
Demo:

- https://bedrock-demo-agibson.oregon-b.moz.works/en-US/privacy/firefox/
- https://bedrock-demo-agibson.oregon-b.moz.works/en-US/privacy/firefox-rocket/

PR #6635, opened 2018-12-19
Update footer

stephaniehobson:
## Description
- Updates footer and language switcher markup to match Protocol 4.0.0 for all systems
- Adds [new IA](https://docs.google.com/document/d/1g0N8oy0WSOY218LJjDIslBbVOmNaBY4Agnfo-cvp6wA/edit#heading=h.vvqh5av6uul) to English pages
- Adds new `Supports` and `Utils` files to bundles.

## Issue / Bugzilla link
Fix #6104 

## Testing
- [x] Protocol, Pebbles, and Sandstone pages
  - [x] RTL
  - [x] mobile, tablet, desktop, wide desktop
  - [x] language switcher still works
- [x] Header nav still works? Did I break anything global when we moved stuff into the `Utils` and `Supports` files?

PR #6722, opened 2019-01-16
Revert removal of accordion from /privacy/firefox/ pages (Fixes #6713)

### LINKED ISSUE #6713: Revert to accordion on Fx Privacy Notice page
| Estimates |  |
| PM | .5 |
| ENG | 2 |

Description:
Near the end of last year, the accordions/expanders on the Firefox Privacy Notice were removed. It's important that we put them back.

GDPR created an requirement that Privacy Notices be very short and understandable. But it also allows us to create a "layered" privacy policy. The first layer is the headlines. The second layer (under the accordions) has more explanation. Since legal creates and maintains this page to manage our legal risk, it's important to us to have control over how it looks.

---

- [ ] When a user lands on the page, they see only the headlines, and then can expand them for more information.

- [ ] This is an example of risk. Use a checklist!
- [ ] This is an example of risk.

#### Tasks
- [ ] Revert the page to the previous accordion design (not the content, just the accordion)
- [ ] Restore or otherwise place anchors

alexgibson:
## Description
- Reverts removal of accordion functionality from `privacy/firefox` pages, that was removed in https://github.com/mozilla/bedrock/pull/6278.

Note: whilst restoring this code I realised it was never very accessible. The ARIA support isn't great, and the individual sections are not expandable via keyboard 😩. There seems to be some urgency in restoring this that I don't fully understand (from a legal point of view), but if we are to support this accordion as a feature long term, then we should at least file bugs and make sure our privacy content is accessible to everyone.

## Issue / Bugzilla link
#6713

## Testing
- [ ] Accordion should function as it was previously on all related privacy pages.
- [ ] Accordion should be keyboard accessible.

### DIFF
--- a/bedrock/privacy/templates/privacy/base-protocol.html
+++ b/bedrock/privacy/templates/privacy/base-protocol.html
@@ -73,4 +73,6 @@
 </article>
 {% endblock %}
 
-{% block js %}{% endblock %}
+{% block js %}
+  {{ js_bundle('privacy_protocol') }}
+{% endblock %}
--- a/bedrock/privacy/templates/privacy/notices/firefox-quantum.html
+++ b/bedrock/privacy/templates/privacy/notices/firefox-quantum.html
@@ -7,5 +7,6 @@
 {% block body_class %}{{ super() }} product-firefox{% endblock %}
 
 {% block js %}
+  {{ super() }}
   {{ js_bundle('privacy_firefox') }}
 {% endblock %}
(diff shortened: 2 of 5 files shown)


alexgibson:
Ok, I've spent some time improving the accordion accessibility on these pages. Ideally I had wanted to use the new Protocol [Details molecule](https://protocol.mozilla.org/patterns/molecules/details.html), but we're blocked on using that until the footer is added to bedrock in https://github.com/mozilla/bedrock/pull/6635.

For now, I've just done the minimal amount of extra work needed to make sure that the accordion is keyboard accessible, and that the sections have appropriate ARIA information to convey state.

We should follow up long term to remove all the custom JS on this page and use the standard Protocol component.

This is now ready for review!

PR #6738, opened 2019-01-22
Update external data during docker build

pmac:
* Only pull data that doesn't require auth during build as secrets aren't available.
* Also run migrations and db updates during dev image build.


jpetto:
Code changes LGTM.

alexgibson:
Circle is failing with an error, but I pulled this locally and ran `make build`, and everything worked as expected 👍

pmclanahan:
Same error in Circle for master builds at the moment, and the failure looks unrelated to this change. I'm forcing a merge so that we can fix dev.

PR #6745, opened 2019-01-23
Update clock deploy image to fix mozmeao/infra#1010

jgmize:
## Description

update image in {clock-,}deploy.yaml in bin/deploy.sh

## Issue / Bugzilla link

mozmeao/infra#1010

## Testing
Merge this to master, verify clock-deploy.yaml is updated in the git-sync-operator branch of mozmeao/www-config


jgmize:
Thanks @pmac, good catch! I've applied your suggestions.

Issue #6711, opened 2019-01-15
Launch new SEO page "Free VPN" 


ejregithub:
| Estimates |  |
| ------------- | ------------- |
| PM | 2 |
| ENG | 2 |
| SEO | .5 |

Description: using the template established for SEO pages from the NA Desktop team, create the new page 'Free VPN'

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Page is live before end of January 2019

#### Tasks
- [ ] Assign the URL /firefox/features/free-vpn/

Native Sketch file can be found [here](https://share.goabstract.com/75a5ee3e-51e2-415f-8950-3ae3e70200ef) (in Abstract). Look for the `VPN Landing Page" layer.

Here's the [copy doc](https://docs.google.com/document/d/1eIZr2BPO3uSO_BCfbznOtKWqsC60KCdBn0flTKTYP7I/edit).

ejregithub:
Blocked on meta data and URL

pmac:
@ejregithub Have we passed this copy by our privacy and security folks? It would be pretty embarrassing if we recommended a VPN service that actually hurt user privacy and/or security as many of the free ones do.

amychurchwell:
@ejregithub demo link -- https://bedrock-demo-seo-vpn.oregon-b.moz.works/en-US/firefox/features/free-vpn/

ejregithub:
From Scott (AMO): This content was written last summer, before AMO instituted their review policy. So none of these have been reviewed. Scott has added them to the security review queue for this month and he hopes that they’ll be done in a few weeks.

ejregithub:
Closing this due to issues cited above. We can create a new issue if need be

PR #6751, opened 2019-01-24
Add SEO page, "Free VPN" (#6711)

amychurchwell:
## Description
* Page is live before end of January 2019
* New page is at URL /firefox/features/free-vpn/

## Issue / Bugzilla link
#6711 

## Testing
🍝 copy pasta errors
https://bedrock-demo-seo-vpn.oregon-b.moz.works/en-US/firefox/features/free-vpn/


ejregithub:
Blocked for now, yes.

AMO is supposed to chime in and the security would need to bet any addons we would want to directly recommend- so this will probably die.

PR #6768, opened 2019-01-30
[no bug] Fix /foundation redirects

craigcook:
## Description
Don’t require a trailing slash when redirecting to foundation.m.o

## Testing
Is my regex correct?
Does this break any other redirects?


craigcook:
Yup, looks like I broke something...

Issue #6736, opened 2019-01-22
[2019-03-19] WNP 66


ejregithub:
[Master project doc](https://docs.google.com/document/d/121RNIt_W8RSjubBHugQgCXKH8F8solk2hXP0ziunXsQ/edit)

| Estimates |  |
| ------------- | ------------- | 
| TL | |
| PM | 1/2 day |
| ENG | 4 |
| SEO | |
| ANA | |
| DES | |
| COPY | |

Description: the What's New page for Firefox 66, promoting Firefox Accounts

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Page is live a week before the release date to allow for QA (external)

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] String changes 

#### Tasks:
- [ ] Code the page with creative assets provided by the Accounts team

- Schedule
Deadline for [final copy](https://docs.google.com/document/d/1vuNn8ct1ub0DAwz2YivEmNIDwP2yR4NtAmo8FSR6LnA/edit?usp=sharing) - 2/8
Submit PR w. strings - 2/13
Deadline for final design - 2/19
Coded page on demo for stakeholder QA review - 3/5
RC week - 3/11
WNP 66 go live - on or before - 3/12
Release date - 3/19

hoosteeno:
@amychurchwell can you update here with relevant links & status?

amychurchwell:
Final design a la mham.
<img width="1035" alt="screen shot 2019-02-19 at 1 05 43 pm" src="https://user-images.githubusercontent.com/30483988/53116076-9ce27300-34fc-11e9-82fa-93fcae065cef.png">
https://share.goabstract.com/e11de043-482a-436d-a88b-6223c841bd4e

amychurchwell:
Does Firefox Send have an adjust link? I am not seeing it as an app store product, only finding https://send.firefox.com/
If we are linking directly do we have utm param requirements? Also the design will change from a button to a link for that section as it would not have a QR code.

amychurchwell:
@ejregithub demo link -- https://bedrock-demo-achurchwell.oregon-b.moz.works/en-US/firefox/66.0/whatsnew/
will also post in slack channel.

ejregithub:
Closing, this is live in prod -

PR #6812, opened 2019-02-11
Add WNP 66 l10n copy

amychurchwell:
## Description
Contains [final copy](https://docs.google.com/document/d/1vuNn8ct1ub0DAwz2YivEmNIDwP2yR4NtAmo8FSR6LnA/edit?usp=sharing) for translation.

## Issue / Bugzilla link
#6736

### DIFF
--- a/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx66.html
+++ b/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx66.html
@@ -1,0 +1,45 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+  # License, v. 2.0. If a copy of the MPL was not distributed with this
+  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+ {% add_lang_files "firefox/whatsnew_66" %}
+
+ {% extends "firefox/base-protocol.html" %}
+
+ {% block page_title %}{{ _('Get a Firefox Account – Keep your data private, safe and synced') }}{% endblock %}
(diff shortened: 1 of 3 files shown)


amychurchwell:
@ejregithub final copy for WNP66 ready for l10n. 🙂

ejregithub:
@peiying2 - flagging this one for string extraction, please. This is final copy for the /whatsnew page intended for the Firefox 66 release on 3/19.

peiying2:
Once we have the demo URL (pending design finalization), I will prep the strings for l10n.

peiying2:
Localization targets 3/7 to complete as many locales as possible in order to prepare the locale list for the rel eng group.

peiying2:
@amychurchwell We need to check with the copyrwriter for this one. 

A comment from one of our l10n contributor questions the use of some of the common phrases and they appear different from the past. Has the style changed?

"...and when you’re signed into your Account," should be  "... signed in to ..." instead?
"One log-in. Power and security everywhere", shouldn't it be "One login..." - there are two instances in this file

Of course the strings are leveraged from the firefox/accounts-2018 template. So both files need to be reviewed. 

In checking on how they were used before here: https://transvision.mozfr.org/, the common expressions are "login" and "sign in to...".

We have extracted strings from this one so any revision will require a tag. If we make changes here, are we going firefox/accounts-2018 template too?

amychurchwell:
@peiying2 I checked in with the copywriter, this is not an intentional style change. We are wanting to keep the original style of "login" and "sign in to..."
Seeing as the strings have been extracted, when I submit the final PR with design I will tag copy changes for these instances. Thank you for calling this out. :+1:

alexgibson:
The WNP for FF 66 was merged in https://github.com/mozilla/bedrock/pull/6883, so I think we can close this one now.

Issue #6669, opened 2019-01-03
Stocktake of active experiments


alexgibson:
Below is a list (in no particular order) of all experiments in the bedrock code base. Some have been running for a long time, and we've never made a decision on them or followed up. Others may not have enough data yet to be conclusive, and some are inactive and probably just need clean up.

- [x] 1. Feature page A/B test without main header (Issue #6155). The code exists in bedrock but I can't see the switch enabled in production. Was this experiment ever run?
- [x] 2. Pre-download experiment rerun (Issue #6456). We've finished the rerun of this experiment for English locales, but the experiment is still running for Germany and France.
- [x] 3. Firstrun copy experiment ([bug 1451051](https://bugzilla.mozilla.org/show_bug.cgi?id=1451051)). This is an old page, and the experimet is inactive. Not sure if we ever followed up on it, but the code and templates can likely all be removed.
- [x] 4. SEM browser compare experiment for Chrome (Issue #5952). Experiment is active in production and has been running for a long time.
- [x] 5. SEM browser compare experiment for Edge (Issue #5951). Experiment is active in production and has been running for a long time.
- [x] 6. SEM browser compare experiment for Opera (Issue #6130). Experiment is active in production and has been running for a long time.
- [x] 7. SEM browser compare experiment for Internet Explorer (Issue #6466). Same as above, but has been running for less time. 
- [x] 8. We're also actively serving a Facebook video chat funnelcake to users who visit /firefox/ from a FB referrer (#6206). This has been running for a long time. Given that users cannot easily update their funnelcake build via the regular FF update channel, is serving a long running funnelcake like this benefiting aDAU, or harming it?
- [x] 9. Experiment to add FxA CTA to /new (Issue #6173). Active in production and has been running for a long time.
- [x] 10. Yandex experiment on /ru/firefox/new/ (Issue #6300). This is still running, and nobody knows quite what to do with it, or if it's something we want to continue doing long term.

I think that's it, but I'll add more if / when I come across them.

alexgibson:
10. We're also actively serving a Facebook video chat funnelcake to users who visit `/firefox/` from a FB referrer (https://github.com/mozilla/bedrock/issues/6206). This has been running for a long time. Given that users cannot easily update their a funnelcake via the regular FF update channel, is serving a long running funnelcake like this benefiting aDAU in the long run, or harming it?

hoosteeno:
[...]

PR #6457, opened 2018-11-09
Re-launch pre-download experiment on /firefox/ (Fixes #6456)

alexgibson:
## Description
- Re-launch pre-download experiment on `/firefox/` (Reverts #6224).
- Do not merge until we get updated experiment cohorts & duration.

## Issue / Bugzilla link
#6456

## Testing
TODO

PR #6825, opened 2019-02-14
Remove pre-download experiment code from /firefox/ (#6457)

alexgibson:
## Description
- All pre-download experiments have now finished. This PR removes the experiment code from bedrock.

## Issue / Bugzilla link
#6457
#6669

## Testing
- [ ] Make sure I didn't miss any stray references?


alexgibson:
Do not merge until I get the ok to remove this code.

alexgibson:
@hoosteeno any objection to removing this experiment code?

hoosteeno:
If the experiment were to graduate, would we want this code? If so, please don't remove. If not, go ahead.

alexgibson:
> If the experiment were to graduate, would we want this code? If so, please don't remove. If not, go ahead.

This experiment was written fast and to only ever target a single page. There are also many UX problems it presents, that weren't solved in the experiment. If we want to graduate it, it would need pretty much rewriting from scratch. I think we're OK to remove it 👍

PR #6827, opened 2019-02-14
(Fixes #6826) Fix l10n href bug.

### LINKED ISSUE #6826: Mozilla legal contact update missing variable in localizable string
## Description
PR 6821 update missing variable in localizable string

Describe the issue that you're seeing.
See comment in the https://github.com/mozilla/bedrock/pull/6821/files#r256880657

### Expected result
Need variable in the localizable string to reference back to the URL

@amychurchwell; @craigcook

amychurchwell:
## Description
Replace URL in localizable string with variable.

## Issue / Bugzilla link
#6826

### DIFF
--- a/bedrock/privacy/templates/privacy/index.html
+++ b/bedrock/privacy/templates/privacy/index.html
@@ -53,7 +53,9 @@
         </span>
       </p>
       {% if l10n_has_tag('privacy_email_2019') %}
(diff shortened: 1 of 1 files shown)


craigcook:
I don't think we need this, putting DNM for the moment while we clarify with Peiying.

peiying2:
I think the change is correct. My description of the issue is incorrect.

craigcook:
Confusion resolved! The string is good, no need to change it here.

PR #5748, opened 2018-05-28
Bug 1464779 - Update google_play_button macro to accommodate use of a…

l-hedgehog:
…djust.com links.

## Description
Update `google_play_button` so that:
* `data-mozillaonline-link` is automatically included when using an alternative adjust.com link to download Firefox for Android;
* Include `data-link-type` and `data-download-os` automatically for Firefox Focus;
* Rename `anchor_attributes` to `extra_data_attributes` and make it impossible to define `class`, `id` etc. in it;
* Only include `noopener noreferrer` in `rel` when `target` is used.

## Issue / Bugzilla link
https://bugzil.la/1464779

## Testing
* Markup of most Google Play buttons should stay unchanged. The ones using adjust.com links for Firefox should have an additional `data-mozillaonline-link` if not already added manually.

PR #6835, opened 2019-02-15
Bug 1528163 - Make sure direct FxA links have replacement urls for Mo…

l-hedgehog:
…zillaOnline build

## Description
When viewing with a China repack of desktop Fx, FxA signup/signin links on /firefox/accounts/ and /firefox/64.0/whatsnew/ still points to global FxA services, while they should be replaced with urls to China FxA services.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/1528163

## Testing
* Setup UITour permission for test [1](http://bedrock.readthedocs.io/en/latest/uitour.html#local-development)
* Set "distribution.id" with `Services.prefs.getDefaultBranch("distribution.").setCharPref("id", "MozillaOnline");` in Firefox's browser console.
* Visit [/firefox/accounts/](https://www.mozilla.org/firefox/accounts/), the "Create a Firefox Account" button and the "Sign In" link just above the footer  should point to https://accounts.firefox.com.cn/

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -301,6 +301,28 @@
   </section>
 {%- endmacro %}
 
+{% macro fxa_link_fragment(entrypoint, action='signin', utm_params={}) -%}
+  {% set fxa_path = '' %}
+  {% set fxa_queries = [
+    'service=sync',
+    'context=fx_desktop_v3'
+  ] %}
+
+  {% if action == 'signup' %}
+    {% set fxa_path = 'signup' %}
+  {% else %}
+    {% do fxa_queries.insert(0, 'action=email') %}
+  {% endif %}
+  {% do fxa_queries.append('entrypoint={}'.format(entrypoint)) %}
+  {% for param, val in utm_params.items() %}
+    {% if val and param in ['campaign', 'content', 'medium', 'source'] %}
+      {% do fxa_queries.append('utm_{}={}'.format(param, val.replace('/', '%2F'))) %}
+    {% endif %}
+  {% endfor %}
+
+  href="{{ settings.FXA_ENDPOINT}}{{ fxa_path }}?{{ fxa_queries|join('&') }}" data-mozillaonline-link="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}{{ fxa_path }}?{{ fxa_queries|join('&') }}"
+{%- endmacro %}
+
 {% macro fxa_create_account_button(button_location, id_prefix, campaign, entrypoint, button_text = '', button_class = '', content = '', source = '') %}
   {# - creates the "create account" button which links to accounts.firefox.com
        and fallback states including:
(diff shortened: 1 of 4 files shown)


bedrock/base/templates/macros.html
```diff
@@ -301,6 +301,28 @@ <h3>{{ _('Your download link was sent.') }}</h3>
   </section>
 {%- endmacro %}
 
+{% macro fxa_link_fragment(entrypoint, action='signin', utm_params={}) -%}
```
alexgibson: This is turning into quite a lot of complex logic for a macro to be responsible for. My gut says we should probably convert this into a jinja helper in Python, so that it can have tests. Thoughts, @pmac ?

bedrock/firefox/templates/firefox/whatsnew/fx64/whatsnew-fx64-base.html
```diff
@@ -32,6 +32,12 @@
   {% set sync_desc = _('Send an open tab on one device to all your others with a single tap. Much easier than texting or emailing yourself those links.') %}
 {% endif %}
 
+{% set fxa_utm_params = {
+  'campaign': 'wnp64',
+  'content': '/firefox/WNP/64/',
```
alexgibson: Should this value be url encoded like the original?

l-hedgehog: It is handled in the macro:

`{% do fxa_queries.append('utm_{}={}'.format(param, val.replace('/', '%2F'))) %}`

bedrock/base/templates/macros.html
```diff
@@ -301,6 +301,28 @@ <h3>{{ _('Your download link was sent.') }}</h3>
...
+
+  {% if action == 'signup' %}
+    {% set fxa_path = 'signup' %}
+  {% else %}
+    {% do fxa_queries.insert(0, 'action=email') %}
+  {% endif %}
+  {% do fxa_queries.append('entrypoint={}'.format(entrypoint)) %}
+  {% for param, val in utm_params.items() %}
+    {% if val and param in ['campaign', 'content', 'medium', 'source'] %}
+      {% do fxa_queries.append('utm_{}={}'.format(param, val.replace('/', '%2F'))) %}
```
alexgibson: Would it be worthwhile URL encoding the values here, rather than special-casing for  `/`?

bedrock/firefox/templates/firefox/whatsnew/fx64/whatsnew-fx64-base.html
```diff
@@ -62,8 +68,8 @@ <h1 class="c-page-header-up-to-date">{{ latest }}</h1>
    include_cta=True
   ) %}
   <p class="show-fxa-supported-signed-out show-fxa-default">
-    <a href="https://accounts.firefox.com/signup?service=sync&context=fx_desktop_v3&entrypoint=mozilla.org-fx.wnp64&utm_content=%2Ffirefox%2FWNP%2F64%2F&utm_campaign=wnp64&utm_source=wnp-64" class="mzp-c-button mzp-t-download">{{ button_create }}</a><br>
-    <span class="wn64-signin">{{ login_desc }} <a href="https://accounts.firefox.com/?action=email&service=sync&context=fx_desktop_v3&entrypoint=mozilla.org-fx.wnp64&utm_content=%2Ffirefox%2FWNP%2F64%2F&utm_campaign=wnp64&utm_source=wnp-64">{{ login_link }}</a></span>
+    <a {{ fxa_link_fragment(entrypoint='mozilla.org-fxa.wnp64', action='signup', utm_params=fxa_utm_params) }} class="mzp-c-button mzp-t-download">{{ button_create }}</a><br>
```
alexgibson: The entrypoint here should be `mozilla.org-fx.wnp64` to match the rest (you have an extra `a` between `f` and `x`.

bedrock/base/templates/macros.html
alexgibson: We can address this in a follow up PR.

bedrock/base/templates/macros.html
alexgibson: We can address this in a followup PR.

bedrock/firefox/templates/firefox/whatsnew/fx64/whatsnew-fx64-base.html
l-hedgehog: Fixed for both links.

l-hedgehog:
> Seeing as a lot of the code here is already in a macro, I think it's ok not to block on this requiring a refactor. But I will create a follow up issue so that we can address this, and write better tests for it in the future.
> 
Thanks. I really should take some time to get this and a similar refactor for #5748 done.

PR #6845, opened 2019-02-21
Use HTTPS for links to licenses

fmarier:
The HTTP scheme was deprecated by Mozilla a while back: https://blog.mozilla.org/security/2015/04/30/deprecating-non-secure-http/

### DIFF
--- a/bedrock/mozorg/templates/mozorg/mpl/headers/index.html
+++ b/bedrock/mozorg/templates/mozorg/mpl/headers/index.html
@@ -1,6 +1,6 @@
 {# This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+# file, You can obtain one at https://mozilla.org/MPL/2.0/. -#}
 
 {% extends "mozorg/base-resp.html" %}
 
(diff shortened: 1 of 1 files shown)


fmarier:
Note that the license itself (under "Exhibit A") suggests an HTTPS URL: https://www.mozilla.org/en-US/MPL/2.0/

pmac:
OHAI @fmarier o/

bedrock/mozorg/templates/mozorg/mpl/headers/index.html
```diff
@@ -1,6 +1,6 @@
 {# This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+# file, You can obtain one at https://mozilla.org/MPL/2.0/. -#}
```
pmac: In fact we could update these to the newest full URL: https://www.mozilla.org/en-US/MPL/2.0/

pmac: well... we can leave out the `en-US/` bit as it will redirect, but we could add the www. Thoughts?

fmarier: Maybe it should match the blurb in the license though?

pmac: Good point. Seems fine 👍

PR #6846, opened 2019-02-22
l10n style for Vietnamese (Fixes 6844)

### LINKED ISSUE #6844: The font used in the home page is not full internationalized
## Description

Texts include diacritics looks rather bad in intro section. You can see:

I think this font don't (`Zilla Slab`) support or not include path support for Vietnamese. 

### Steps to reproduce

Right at homepage: https://www.mozilla.org/vi/

### Expected result

The words and punctuation marks are displayed normally.

### Actual result

The texts is broken.

### Environment

What browser and operating system are you using?

Chrome, Windows 10

pierreneter:
## Description

Support font style for Vietnamese, texts include diacritics will be displayed normally.

## Issue / Bugzilla link

Close #6844


pierreneter:
ping @craigcook

craigcook:
Deepest apologies for neglecting this. While these changes do indeed fix the problem, I think we want to give some careful thought to how we might redefine variables for different locales in a way that will scale. We're working on some new font rules in our design system and I want to explore our options. We might actually end up taking this approach after all. 

For the moment I've made a somewhat simpler fix in #7031 that also solves the problem but with hopefully fewer repercussions since it's contained to just the vi `intl.css` file. I'm going to close this pull request in favor of the other one. Thanks for your help!

PR #6900, opened 2019-03-08
[fix #6887] Add FxA button in global navbar

### LINKED ISSUE #6887: Graduate Global FxA CTA
## Description

We previously ran this experiment and need to make it permanent in the locales in which it has been localized.

The original work - #6629 
The next rev with localizations - #6792

### Estimates

| Estimates |  |
| TL | |
| PM | |
| ENG | |
| ANA | ? |

---

- [ ] Global CTA is live in en, DE and FR
- [ ] Content (strings) are set to be localizable to all locales

#### Tasks
- [ ] Submit PR with strings tagged for L10n
- [ ] Reach out to Peiying with DE and FR content in request to extract strings - @ejregithub 
- [ ] Communicate to Accounts team (Alex Davis) that this is happening - @ejregithub 

Localized strings
EN | DE | FR
-- | -- | --
Get a Firefox Account | Firefox-Konto erstellen | Créez un compte Firefox
Check out the Benefits | Das sind deine Vorteile | Découvrez les avantages

craigcook:
## Description
Makes the former experiment permanent with a few small tweaks, and removes the old experiment code. 

For signed-in users I'm changing the button's href to point to the /accounts page. I was also intending to hide the "See the benefits" link but since we can't tell if someone is signed in until the DOM is ready the link was still briefly visible and I didn't like that... so the link is still there, even though it's redundant. However, I think it still adds some context just by clearly stating "See the benefits" so might do more good than harm? Unless we can think of a better treatment to make it appear or hide after the DOM loads without it being super clunky.

## L10n
These are two new strings, but it's a global element so we can't use an l10n tag. We had the strings in German and French as part of the test, so I've added an `if LANG in ['en-US', 'en-CA', 'en-GB', 'en-ZA', 'de', 'fr']` conditional. 

However, the German and French test strings were hardcoded and now we'll need to get them into a lang file. Can we do that manually for those two locales before this merges? (Pinging @peiying2)

OR we can make the conditional English-only for launch and let the regular l10n workflow happen, prioritizing top tier languages and update the conditional as soon as possible. Which approach is best? I'm marking it Do Not Merge until we know which way to go.

## Issue / Bugzilla link
#6887 

## Testing
- [ ] Firefox Account button is shown to Firefox users in English, German, and French.
- [x] Firefox Account button links to accounts page for Firefox users who are signed.
- [x] Firefox download button is shown to non-Firefox browsers.
- [x] Firefox download button is shown to all other locales, including Firefox browsers.

Should we add some automated tests for this? Just a basic test for "Firefox gets account button, non-Firefox gets download button"?


media/js/base/protocol/init-navigation.js
```diff
@@ -41,6 +41,62 @@
...
+
+        var userMajorVersion = Mozilla.Client._getFirefoxMajorVersion();
+        var latestMajorVersion = parseInt(document.documentElement.getAttribute('data-latest-firefox'), 10);
+
+        if (!userMajorVersion || !latestMajorVersion) {
+            return false;
+        }
+
+        // User should be on Firefox Quantum or greater.
+        return userMajorVersion >= 57;
```
alexgibson: I know the original ask for this was to show the button for FF >=  57, but there could be a simpler, more performant way to cut down on all this logic (perhaps)?

1. Default the button `href` to `/firefox/accounts/`.
2. Use the `is-firefox` class that gets added in `site.js` to determine showing the button (no additional JS required, just some CSS).
3. Then use UITour here to upgrade the button `href` to point to `acconts.firefox.com` if not already signed in.

This would mean the button would get shown to all Firefoxes, and not just desktop, but the logic to handle all those cases should be built into the /accounts/ page (I think?). 

Just an idea, but it does change some parameters. And what you have works, so totally feel free to say no 🙂

media/js/base/protocol/init-navigation.js
```diff
@@ -41,6 +41,62 @@
         }
     }
 
+    var nav = document.querySelector('.mzp-c-navigation');
```
alexgibson: Nit: suggest putting all this code into a function that gets called after the navigation initializes? The more stuff we put in front of that code, the longer it takes to run for everyone. Seeing as this is applicable only to Firefox users, it's probably safe for it to come after?

bedrock/base/templates/includes/protocol/navigation/index.html
```diff
@@ -24,22 +24,12 @@
...
-                {{ fxa_button_text }}
+              <a class="c-navigation-fxa-cta mzp-c-button mzp-t-secondary mzp-t-small mzp-t-download" {{ fxa_link_fragment(entrypoint='mozilla.org-globalnav', action='signup', utm_params={'campaign': 'globalnav', 'content': 'get-firefox-account', 'medium': 'referral', 'source': 'www.mozilla.org'}) }} data-button-name="Get a Firefox Account" data-link-type="button" data-cta-position="secondary cta" data-alt-href="{{ url('firefox.accounts') }}">
+                {{ _('Get a Firefox Account') }}
               </a>
-              <p class="c-navigation-fxa-cta-small-link"><a data-link-type="nav" data-link-name="Check out the Benefits" href="{{ url('firefox.accounts') }}">{{ fxa_link_text }}</a></p>
+              <p class="c-navigation-fxa-cta-small-link"><a data-link-type="nav" data-link-name="Check out the Benefits" href="{{ url('firefox.accounts') }}">{{ _('Check out the Benefits') }}</a></p>
```
alexgibson: If I'm signed into FxA I see two links that go to the same page:

Screenshot 2019-03-12 at 13 41 16


Should we only show this small link if signed out?

bedrock/base/templates/includes/protocol/navigation/index.html
```diff
@@ -24,22 +24,12 @@
...
-              {% set fxa_button_text = 'Get a Firefox Account' %}
-              {% set fxa_link_text = 'Check out the Benefits' %}
-            {% endif %}
+          {% if LANG in ['en-US', 'en-CA', 'en-GB', 'en-ZA', 'de', 'fr'] %}
             <div class="c-navigation-fxa-cta-container">
-              <a class="c-navigation-fxa-cta mzp-c-button mzp-t-secondary mzp-t-small mzp-t-download" {{ fxa_link_fragment(entrypoint='mozilla.org-globalnav', action='signup', utm_params={'campaign': 'globalnav', 'content': 'get-firefox-account', 'medium': 'referral', 'source': 'www.mozilla.org'}) }} data-button-name="Get a Firefox Account" data-link-type="button" data-cta-position="secondary cta">
-                {{ fxa_button_text }}
+              <a class="c-navigation-fxa-cta mzp-c-button mzp-t-secondary mzp-t-small mzp-t-download" {{ fxa_link_fragment(entrypoint='mozilla.org-globalnav', action='signup', utm_params={'campaign': 'globalnav', 'content': 'get-firefox-account', 'medium': 'referral', 'source': 'www.mozilla.org'}) }} data-button-name="Get a Firefox Account" data-link-type="button" data-cta-position="secondary cta" data-alt-href="{{ url('firefox.accounts') }}">
```
alexgibson: `mzp-t-download` no longer exists now that bedrock is on Protocol 5.0, should be `mzp-t-product`.

bedrock/base/templates/includes/protocol/navigation/index.html
```diff
@@ -24,22 +24,12 @@
...
-              {% set fxa_button_text = 'Firefox-Konto erstellen' %}
-              {% set fxa_link_text = 'Das sind deine Vorteile' %}
-            {% elif LANG == 'fr' %}
-              {% set fxa_button_text = 'Créez un compte Firefox' %}
-              {% set fxa_link_text = 'Découvrez les avantages' %}
-            {% else %}
-              {% set fxa_button_text = 'Get a Firefox Account' %}
-              {% set fxa_link_text = 'Check out the Benefits' %}
-            {% endif %}
+          {% if LANG in ['en-US', 'en-CA', 'en-GB', 'en-ZA', 'de', 'fr'] %}
```
alexgibson: We need to expose the new strings below for l10n. Once that happens, opening a PR to add the German and French transaltions to the [l10n repo](https://github.com/mozilla-l10n/www.mozilla.org/) should be easy enough. This feels like a better way to handle the existing translations, seeing as the ask is to open this up for all locales?

alexgibson:
> Should we add some automated tests for this? Just a basic test for "Firefox gets account button, non-Firefox gets download button"?

If we can remove the reliance on UITour for showing this button, then yes we certainly should. Currently it gets show behind a UITour call which makes it tricky. I made a suggestion that could simplify some of the logic and would also make testing this possible.

bedrock/base/templates/includes/protocol/navigation/index.html
```diff
@@ -24,22 +24,12 @@
...
             <div class="c-navigation-fxa-cta-container">
-              <a class="c-navigation-fxa-cta mzp-c-button mzp-t-secondary mzp-t-small mzp-t-download" {{ fxa_link_fragment(entrypoint='mozilla.org-globalnav', action='signup', utm_params={'campaign': 'globalnav', 'content': 'get-firefox-account', 'medium': 'referral', 'source': 'www.mozilla.org'}) }} data-button-name="Get a Firefox Account" data-link-type="button" data-cta-position="secondary cta">
-                {{ fxa_button_text }}
+              <a class="c-navigation-fxa-cta mzp-c-button mzp-t-secondary mzp-t-small mzp-t-download" {{ fxa_link_fragment(entrypoint='mozilla.org-globalnav', action='signup', utm_params={'campaign': 'globalnav', 'content': 'get-firefox-account', 'medium': 'referral', 'source': 'www.mozilla.org'}) }} data-button-name="Get a Firefox Account" data-link-type="button" data-cta-position="secondary cta" data-alt-href="{{ url('firefox.accounts') }}">
+                {{ _('Get a Firefox Account') }}
```
alexgibson: Not a blocker, but I wonder if there could be another string to show to people who are already signed in, something that speaks to the features/benefits of their account? Maybe something we could add in a future iteration?

alexgibson: Ooh, we already have the string "Your Firefox Account" in the nav. Maybe that could work?

bedrock/base/templates/includes/protocol/navigation/index.html
craigcook: I tried that but then the small link is visible for a split second until the JS determines the user state and hides it. And since it also has an effect on the navbar height, showing/hiding that link can make the entire page jump by a few pixels. I decided the redundant link was better, but maybe there's another way around it?

bedrock/base/templates/includes/protocol/navigation/index.html
craigcook: Hmm... I would expect that to lead me to a place where I can manage my account, not just a marketing page. What if instead of the benefits page we link to `https://accounts.firefox.com/signin` ? Maybe not right now but that's an idea for the future.

alexgibson: This is the text we already use in the nav, so seems consistent at least? My issue was more with the copy, but maybe let's leave this for now?

bedrock/base/templates/includes/protocol/navigation/index.html
alexgibson: Maybe only show it if not already signed in?

bedrock/base/templates/includes/protocol/navigation/index.html
craigcook: Yeah, and at least it's better than "Get" when you clearly already have one. Maybe we should ask Jennifer for a UX opinion before we make the decision.

craigcook: So I've tinkered a bit and this has the same issue as hiding the benefits link; the "Get" button is visible for a second before we can get the account state and change it to "Your." We could hide everything, show no button at all until we determine state and that's how we decide whether to show "Get" or "Your" ... but then Download will be visible for a split second instead.

Maybe this will be fixed when we fix `getFxaDetails` if we no longer have to wait until the DOM finishes, but for the moment, and as long as there's some delay before we can determine signed-in/signed-out state (either for domready or uitour or something else), I don't see a way to change the button label or hide the benefits link without that little jolt of movement in the corner of your eye right after the page loads. And if this was lower down on the page I wouldn't worry about it, but right there in the navbar it really stands out. At least changing the href is invisible.

bedrock/base/templates/includes/protocol/navigation/index.html
craigcook: That's also possible, but then it's hidden and suddenly pops into view once we determine you're signed out and it's just as bad as it being visible and then vanishing if you're signed in.

Something that might help is if we 1) hide it by default, but reserve space for it to avoid the page jitter using `visibility:hidden` rather than `display:none` and 2) add a slight transition so it fades in rather than becoming visible instantly. Then it still draws attention but might feel more deliberate. Worth a try maybe?

media/js/base/protocol/init-navigation.js
craigcook: This is slightly complicated by the `fxa_link_fragment` macro which includes the firefox.com.cn link for China. We could update that macro, but then it's also in use on other pages and I'm worried about blowing up the scope of this and delaying it even further 😟 

But I think we could definitely just show it to all Firefoxes and not worry about the version and all the other checks. Do we need to still check that it's desktop and not mobile? What happens if you get the button on mobile? I guess I should try that...

bedrock/base/templates/includes/protocol/navigation/index.html
alexgibson: > Maybe this will be fixed when we fix getFxaDetails if we no longer have to wait until the DOM finishes

I've submitted a PR to fix this issue, so hopefully it might help a bit? We do have pages where the entire main CTA is dependent on signed-in state and suffers from the same issue, so I don't know if this is a real show stopper or not. People who are using FF and already signed-in are in a definite minority, so it could arguably be better for only those users to see a flash of content, as opposed to making everyone wait to see the critical content?

That said, if you think it's too distracting to not be worth it, we could totally just stick to the original plan and only update the href. I think this still works OK seeing as the destination is `/accounts/`.

bedrock/base/templates/includes/protocol/navigation/index.html
alexgibson: Good point yeah, the `visibility: hidden` suggestion sounds like a good middle ground. 

But perhaps we're now on the brink of over complicating this just for users who are signed-in, which to be honest are a minority. Maybe showing this link is acceptable enough for either state. If someone has an account, they can still learn more about the benefits they may not know about?

media/js/base/protocol/init-navigation.js
alexgibson: Good point, let's not creep the scope on this further than it needs to be. We definitely shouldn't be pointing mobile users to accounts.firefox.com though. The FxA team we're pretty specific on not doing this when we ran the experiment, as that page is for desktop users only.

bedrock/base/templates/includes/protocol/navigation/index.html
craigcook: I switched to visibility:hidden so it's still briefly visible, but at least once it hides it doens't make the whole page jump. I can live with that.

PR #6906, opened 2019-03-13
Fix race condition in Mozilla.Client.getFxaDetails() (Fixes #6904)

### LINKED ISSUE #6904: Mozilla.Client.getFxaDetails() returns incorrect data if called multiple times in quick succession
## Description

There is a race condition bug in `Mozilla.Client.getFxaDetails()` that returns incorrect data.

### Steps to reproduce

Call the function twice in quick succession and you can often see different results:

### Expected result

The data returned should always be accurate and consistent.

### Actual result

The results are often different:

```
Object { firefox: true, legacy: false, mobile: false, setup: false, desktopDevices: false, mobileDevices: false }
```

```
Object { firefox: true, legacy: false, mobile: false, setup: true, desktopDevices: 1, mobileDevices: 1 }
```

The code short-circuits itself, since it creates a reference to `Mozilla.FxADetails` too early, before the first callback returns:

https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-client.js#L352-L367

### Environment

Bug can be encountered using any Firefox desktop channel.

alexgibson:
## Description

- Fixes a bug where calling `getFxADetails()` twice in quick succession sometimes resulted in the second call returning too early with incorrect data.

## Issue / Bugzilla link
#6904

## Testing

This is a tiny code change but could do with some careful testing:

- [x] When any bedrock page loads we already query FxA for GA [here](https://github.com/mozilla/bedrock/blob/master/media/js/base/core-datalayer-init.js#L34-L38). This should still return the correct data as it does currently.
- [x] Adding an additional call to`getFxADetails()` in a page's JS bundle should now always result in the exact same data being returned. A page should not need to wait until DOMReady to make a second call (this has been a red herring).
- [x] Existing pages like `/firefox/accounts/` should still show the correct state.

### DIFF
--- a/media/js/base/core-datalayer-init.js
+++ b/media/js/base/core-datalayer-init.js
@@ -3,7 +3,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
 // init core dataLayer object and push into dataLayer
-// This needs to happen on DOM ready for UITour (see https://github.com/mozilla/bedrock/issues/6624).
 $(function() {
     var analytics = Mozilla.Analytics;
     var client = Mozilla.Client;
--- a/media/js/base/mozilla-client.js
+++ b/media/js/base/mozilla-client.js
@@ -357,7 +357,7 @@ if (typeof Mozilla === 'undefined') {
             return;
         }
         // set up the object with default values of false
-        var details = Client.FxaDetails = {
+        var details = {
             'firefox': false,
             'legacy': false,
             'mobile': false,
@@ -423,6 +423,7 @@ if (typeof Mozilla === 'undefined') {
 
         function returnFxaDetails() {
             window.clearTimeout(timer);
+            Client.FxaDetails = details;
             callback(details);
         }
 


alexgibson:
@craigcook the best way for you to test this might be to try adding this to your branch(es) to confirm it fixes the issue. Else you could try and recreate the issue locally on this branch and then merge / rebase?

craigcook:
I've been able to reproduce the original issue by firing `getFxaDetails` twice in quick succession with different results, then after adding these changes to `mozilla-client.js` I get the same results from both. Success!

craigcook:
> When any bedrock page loads we already query FxA for GA here. This should still return the correct data as it does currently.

I'm not sure how best to verify this. Do we need to look in GA? Does that mean pushing this to demo first to test, or are we OK proceeding with the assumption that it's working and maybe just verify on stage?

alexgibson:
A console log or breakpoint here on what `details` returns should be enough to verify GA locally: https://github.com/mozilla/bedrock/blob/master/media/js/base/core-datalayer-init.js#L34

craigcook:
> A console log or breakpoint here on what `details` returns should be enough to verify GA locally: https://github.com/mozilla/bedrock/blob/master/media/js/base/core-datalayer-init.js#L34

Well that's simple enough. I thought we needed to verify it was logging in GA. Verified it's returning correct results locally, so that's good enough for me.

Issue #6823, opened 2019-02-14
Update Release Notes template to use Protocol


stephaniehobson:
Update Release Notes template to use Protocol

Blocked on:
- ~Second nav in pages mozilla/protocol#300~

Nice to have:
- ~Label for button of second mobile nav mozilla/protocol#313~

WIP screenshot:
![screenshot_2019-02-08 firefox 65 0 see all new features updates and fixes](https://user-images.githubusercontent.com/854701/52757294-0c1a0d80-2fb9-11e9-9c8a-73fa28b1641f.png)

PR #6912, opened 2019-03-13
Release notes test first pass

stephaniehobson:
## Description

Functional tests for release notes pages

### DIFF
--- a/bedrock/firefox/templates/firefox/releases/aurora-notes.html
+++ b/bedrock/firefox/templates/firefox/releases/aurora-notes.html
@@ -4,7 +4,7 @@
 
 {% extends "firefox/releases/release-notes.html" %}
 
-{% block gtm_page_id %}data-gtm-page-id="/firefox/android/auroranotes/"{% endblock %}
+{% block gtm_page_id %}data-gtm-page-id="/firefox/auroranotes/"{% endblock %}
 
 {% block page_css %}
   {{ css_bundle('firefox_releasenotes_firefox') }}
--- a/tests/functional/firefox/test_notes.py
+++ b/tests/functional/firefox/test_notes.py
@@ -1,0 +1,113 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.firefox.accounts import FirefoxNotesPage
+
+
+@pytest.mark.nondestructive
+def test_firefox_desktop_release_page(base_url, selenium):
+    #firefox/65.0/releasenotes/
+    page = FirefoxNotesPage(selenium, base_url).open()
+    assert page.version is '65.0'
+    assert page.channel is 'Firefox Release'
+    assert page.gtm_id is '/firefox/releasenotes/'
+    assert page.download_button_text is 'Download Firefox'
+    assert page.download_button_link.contains('firefox-stub')
+    assert page.logo is '/media/img/logos/firefox/logo-quantum-wordmark-large.png'
+    assert page.download_all_url.contains('/firefox/all/')
+
+@pytest.mark.nondestructive
+def test_firefox_desktop_esr_page(base_url, selenium):
+    #/firefox/60.5.2/releasenotes/
+    page = FirefoxNotesPage(selenium, base_url).open()
+    assert page.version is '60.5.2'
+    assert page.channel is 'Firefox ESR'
+    assert page.gtm_id is '/firefox/esr/releasenotes/'
+    assert page.download_button_text is 'Download Firefox'
+    assert page.download_button_link.contains('firefox-stub')
+    assert page.logo is '/media/img/logos/firefox/logo-quantum-wordmark-large.png'
+    assert page.download_all_url.contains('/firefox/all/')
+
+@pytest.mark.nondestructive
+def test_firefox_desktop_beta_page(base_url, selenium):
+    #/firefox/65.0beta/releasenotes/
+    page = FirefoxNotesPage(selenium, base_url).open()
+    assert page.version is '65.0beta'
+    assert page.channel is 'Firefox Beta'
+    assert page.gtm_id is '/firefox/beta/releasenotes/'
+    assert page.download_button_text is 'Download Firefox Beta'
+    assert page.download_button_link.contains('firefox-beta-stub')
+    assert page.logo is '/media/img/logos/firefox/logo-quantum-wordmark-large.png'
+    assert page.download_all_url.contains('/firefox/beta/all/')
+
+@pytest.mark.nondestructive
+def test_firefox_desktop_developer_page(base_url, selenium):
+    #/firefox/54.0a2/auroranotes/
+    page = FirefoxNotesPage(selenium, base_url).open()
+    assert page.version is '54.0a2'
+    assert page.channel is 'Firefox Developer Edition'
+    assert page.gtm_id is '/firefox/auroranotes/'
+    assert page.download_button_text is 'Download Firefox Developer Edition'
+    assert page.download_button_link.contains('firefox-devedition-stub')
+    assert page.logo is '/media/img/firefox/developer/title-inverse.png'
+    assert page.download_all_url.contains('/firefox/developer/all/')
+
+@pytest.mark.nondestructive
+def test_firefox_desktop_aurora_page(base_url, selenium):
+    #/firefox/32.0a2/auroranotes/
+    page = FirefoxNotesPage(selenium, base_url).open()
+    assert page.version is '32.0a2'
+    assert page.channel is 'Firefox Aurora'
+    assert page.gtm_id is '/firefox/auroranotes/'
+    assert page.download_button_text is 'Download Firefox Developer Edition'
+    assert page.download_button_link.contains('firefox-devedition-stub')
+    assert page.logo is '/media/img/firefox/template/header-logo.png'
+    assert page.download_all_url.contains('/firefox/developer/all/')
+
+@pytest.mark.nondestructive
+def test_firefox_desktop_nightly_page(base_url, selenium):
+    #/firefox/65.0a1/releasenotes/
+    page = FirefoxNotesPage(selenium, base_url).open()
+    assert page.version is '65.0a1'
+    assert page.channel is 'Firefox Nightly'
+    assert page.gtm_id is '/firefox/releasenotes/'
+    assert page.download_button_text is 'Download Firefox Nightly'
+    assert page.download_button_link.contains('firefox-nightly-stub')
+    assert page.logo is '/media/img/firefox/template/header-logo-nightly.png'
+    assert page.download_all_url.contains('/firefox/nightly/all/')
+
+@pytest.mark.nondestructive
+def test_firefox_android_release_page(base_url, selenium):
+    #/firefox/android/65.0/releasenotes/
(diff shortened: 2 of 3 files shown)


stephaniehobson:
@agibson It's not functional code yet but you should be able to see where I was going with it.

This is in preparation for #6823.

The basic idea is to test that the correct information is being displayed on the correct pages. The pages I want to test on are the comment on the first line of the test. None of the things being tested require the most recent version of release notes so I think hard coded URLs are okay.

I don't know how to feed the URL directly to the test and could use some help with that 🙂  I can't just feed the version number because the URL structure changes for aurora, android, and iOS.

bedrock/firefox/templates/firefox/releases/aurora-notes.html
```diff
@@ -4,7 +4,7 @@
 
 {% extends "firefox/releases/release-notes.html" %}
 
-{% block gtm_page_id %}data-gtm-page-id="/firefox/android/auroranotes/"{% endblock %}
+{% block gtm_page_id %}data-gtm-page-id="/firefox/auroranotes/"{% endblock %}
```
alexgibson: Is changing this intentional? The URL in prod seems to be `/firefox/android/{% version %}/auroranotes/`, as far as i can tell?

tests/functional/firefox/test_notes.py
```diff
@@ -0,0 +1,113 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.firefox.accounts import FirefoxNotesPage
+
+
+@pytest.mark.nondestructive
+def test_firefox_desktop_release_page(base_url, selenium):
+    #firefox/65.0/releasenotes/
```
alexgibson: If you want to pass this into a test to open a specific URL / version, you could try using a parameterized test to pass in a slug to the URL. You can see an existing example of that [here](https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/features/test_features.py#L10-L18) and [here](https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/features/feature.py#L13).

tests/functional/firefox/test_notes.py
```diff
@@ -0,0 +1,113 @@
...
+    page = FirefoxNotesPage(selenium, base_url).open()
+    assert page.version is '65.0'
+    assert page.channel is 'Firefox Release'
+    assert page.gtm_id is '/firefox/releasenotes/'
```
alexgibson: The gtm page ID is something not really critical to a piece of user functionality on the page, so I'm not sure it really belongs in a functional test like this? You could try and test for this without needing to spin up a browser, using a [python unit test](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/tests/test_views.py#L791-L809), or failing that a headless test.

tests/pages/firefox/notes.py
```diff
@@ -0,0 +1,55 @@
...
+        el = self.find_element(*self._channel_name_locator)
+        return el.textContent
+
+    @property
+    def gtm_id(self):
+        el = self.find_element(*self._html_locator)
+        return el.get_attribute('data-gtm-page-id')
+
+    @property
+    def download_button_text(self):
```
alexgibson: Testing for download button text feels perhaps overly specific for a functional test (if someone changes the copy in the future, does the website break?). Alternatively, you could test that the correct dl button CSS class is applied correctly for each channel perhaps?

If you do add properties like this, I'd suggest adding them to the [download button page region](https://github.com/mozilla/bedrock/blob/master/tests/pages/regions/download_button.py), so other pages can reuse it.

tests/functional/firefox/test_notes.py
```diff
@@ -0,0 +1,113 @@
...
+    assert page.download_button_text is 'Download Firefox'
+    assert page.download_button_link.contains('firefox-stub')
```
alexgibson: Something like this will only pass on Windows, so I'm not sure if this will work? 

We already have good unit tests for dl button URLs in Python. As far as a functional test goes, I'd try and concentrate on things that the visitor experiences on the page:

- Is the download button visible?
- Is the right download button channel shown for the type of release notes page?

I think asserting the two things above would probably be enough? ^

tests/functional/firefox/test_notes.py
```diff
@@ -0,0 +1,113 @@
...
+    assert page.logo is '/media/img/logos/firefox/logo-quantum-wordmark-large.png'
```
alexgibson: This may pass locally but will fail in production when the filepath contains a hash. It's also a bit overly specific for a functional test. For example if we move or rename the file then the test could fail even though the right thing is still shown on the page.

I'd suggest asserting that the image is displayed on the page, rather than the filepath used in the HTML.

bedrock/firefox/templates/firefox/releases/aurora-notes.html
stephaniehobson: Wow, that is a can of worms.

Both https://www.mozilla.org/en-US/firefox/android/57.0/auroranotes/ and https://www.mozilla.org/en-US/firefox/android/57.0/releasenotes/ seem to work but if you view source on the auroranotes URL you can see that the gtm-page-id indicates that it is using the release notes template.

There is no aurora version of Firefox for Android AFAIK. And the [Aurora pages](https://www.mozilla.org/en-US/firefox/32.0a2/auroranotes/) are being incorrectly identified as Android ones right now.

I suspect the android aurora notes are the product of a configured URL redirect? But I will follow up with @pmac to try to sort this out.

stephaniehobson: How did you find the Android aurora notes URLs?

tests/functional/firefox/test_notes.py
stephaniehobson: It's something I had to work really hard to get right in #6823. I suppose a unit test would be more appropriate but I don't know how to unit test logic distributed across 3 templates.

tests/pages/firefox/notes.py
stephaniehobson: This is also something I had to work really hard to get right in #6823. I suppose a unit test would be more appropriate but I don't know how to unit test logic distributed across 3 templates.

tests/functional/firefox/test_notes.py
stephaniehobson: By looking at the first child I'm always looking at the windows button :)

> Is the right download button channel shown for the type of release notes page?

This is indeed what I really want to test, how would you suggest I test it?

tests/functional/firefox/test_notes.py
stephaniehobson: I want to test that the correct logos are appearing based on the product, channel, and version. How would you suggest I check that?

tests/functional/firefox/test_notes.py
stephaniehobson: Do we have headless tests currently?

bedrock/firefox/templates/firefox/releases/aurora-notes.html
alexgibson: > How did you find the Android aurora notes URLs?

I'm not sure you can navigate to them directly any more. I think Aurora only exists on android now, since Aurora became Dev Edition on desktop?

tests/functional/firefox/test_notes.py
alexgibson: > Do we have headless tests currently?

Not as far as I'm aware. The unit test I links to is an example of how to request a URL and parse the HTML for something.

tests/pages/firefox/notes.py
alexgibson: I would perhaps favor testing for a CSS class over the button text? I know release notes are only in English (currently), but testing for English strings also makes tests hard to run against other locales.

tests/functional/firefox/test_notes.py
alexgibson: > This is indeed what I really want to test, how would you suggest I test it?

I could look for the right CSS class in the download button.

tests/functional/firefox/test_notes.py
alexgibson: Use a CSS selector to target the image, and then locate that element and assert it is displayed on the page.

bedrock/firefox/templates/firefox/releases/aurora-notes.html
pmac: Yeah. The aurora situation remains weird. Aurora is no longer a thing for any browser, but the old aurora release notes should stick around, so we do need to keep the aurora URLs for the aurora releases. The last Fx Android Aurora I see in release notes is 54.0a2:

https://www.mozilla.org/en-US/firefox/android/54.0a2/auroranotes/
https://github.com/mozilla/release-notes/blob/master/releases/firefox-for-android-54.0a2-aurora.json

tests/pages/firefox/notes.py
stephaniehobson: The string is literally the thing I want to test. I want to make sure the right product name is on the button.

tests/functional/firefox/test_notes.py
stephaniehobson: I don't see a CSS class that corresponds to channel on the div, ul, li, or a 🙁

stephaniehobson: Ah, there isn't one on release but there is one for beta and nightly.

tests/pages/firefox/notes.py
alexgibson: If the string is the absolutely only way you can be sure that the page is showing the right thing then ok. It's usually something to avoid where possible for the reasons above, but if it's the best thing to do here, then go for it :)

tests/functional/firefox/test_notes.py
alexgibson: You can also give each button a unique Id to make targeting easier

bedrock/firefox/templates/firefox/releases/aurora-notes.html
stephaniehobson: So there is no Android Aurora 57 but there is a page for it: https://www.mozilla.org/en-US/firefox/android/57.0/auroranotes/ :(

stephaniehobson: Oh fun, it works with Desktop release too: https://www.mozilla.org/en-US/firefox/58.0/auroranotes/

stephaniehobson: I poked for 15 minutes and now I need to move onto something else.

For Android and Desktop versions that do not have aurora versions the template being used is the release template, So, the URL is deceptive but the content is accurate. 

I will add an Android Aurora test case to my tests and make sure the content is accurate when an aurora version exists, not sure what else to do. @pmac is this worthy of a bug?

stephaniehobson: I filed https://github.com/mozilla/bedrock/issues/6930 to fix the gtm_page_id problem but that doesn't solve the fact that the URLs resolve when they should probably 404 or forward.

pmac: Maybe worthy of a bug. We should probably redirect `/auroranotes/` to `/releasenotes/` when it's not aurora, but I'm hoping that traffic to a non-aurora release at the former URL is extremely low and maybe we can just not worry about it?

stephaniehobson: Yeah, I'm going to vote for not worrying about it. No one should see this unless they're typing the URLs in themselves.

PR #6917, opened 2019-03-14
Update WNP66 'sign in' strings for l10n

amychurchwell:
## Description
Updating fxa CTA strings for an l10n bugfix

## Issue / Bugzilla link
https://github.com/mozilla-l10n/www.mozilla.org/issues/281

### DIFF
--- a/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx66.html
+++ b/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx66.html
@@ -88,9 +88,16 @@
     <a {{ fxa_link_fragment(entrypoint='mozilla.org-whatsnew66', action='signup', utm_params={'campaign': 'wnp66', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'wnp66'}) }} class="mzp-c-button mzp-t-product" data-button-name="Create account" data-link-type="button">
       {{ _('Create a Firefox Account') }}
     </a><br>
-    <span class="wn64-signin">{{ _('Already have an account?') }}
-      <a {{ fxa_link_fragment(entrypoint='mozilla.org-whatsnew66', action='login', utm_params={'campaign': 'wnp66', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'wnp66'}) }}>{{ _('Sign in') }}</a>
+
+    {% if l10n_has_tag('wnp66_log_in') %}
+    <span class="wn64-signin">{{ _('Have an account?') }}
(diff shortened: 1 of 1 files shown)


amychurchwell:
@peiying2 👋

peiying2:
l10n is done with leveraging strings from another file.

flodolo:
I'm starting to think this is not the best approach after all. You need a "Sign In", and it looks weird in English, it might be completely broken in other languages. Has someone looked into how that string is used? We should always make sure that the context is the same when we do this.

Proposal:
* Expose the new string "Have an account? Log in" behind a tag.
* The fall back without the tag would only be a link with just "Already have an account?", and we ignore the "Sign In" part.

I'd also like to figure out how we ended up in this situation, given that the copy was frozen weeks ago. Was the original request unclear? That's part of the larger discussion about WNP, and the mess this process is.

amychurchwell:
@flodolo 

> I'd also like to figure out how we ended up in this situation, given that the copy was frozen weeks ago.

--totally my (the developer's) fault. What happened was two-fold:
1) I had put the copy deck I received in the l10n PR. Once I started development on page I realized there was no copy for a user who was _already_ signed into firefox accounts, so I made a request to add in the strings we used from previous whatsnew pages for that state.
2) Only after the page was in production I noticed that the 'Create a Firefox Account' and following 'Log in' link were not localized because they were not included in that original l10n PR. 🤦‍♀️ Again, that original l10n PR _only_ included strings that were handed off from the copywriter. The copywriter would have not known to include every string from every button .etc. on the page, that was my responsibility that I missed due to inexperience.

I am in the process of creating whatsnew copy documentation/template so these things are not missed in the future.

In my opinion, I think we are really close to this process not being a total mess. If I had not missed these holes in copy in the first place none of this would have happened.

flodolo:
Thanks for the explanation.

> --totally my (the developer's) fault. What happened was two-fold:

I'm not convinced it's on the developer to provide all possible flows, and which strings should be used for each of those. But I understand that we're going way off-topic for this PR, and it's mostly an internal discussion at that point.

amychurchwell:
@flodolo 


>Expose the new string "Have an account? Log in" behind a tag.
>The fall back without the tag would only be a link with just "Already have an account?", and we ignore the "Sign In" part.

Does "Sign In" read weird in English? It doesn't seem strange to me. Can't speak if it will be broken in other languages, not sure how to find out.

I will put "Have an account? Log in" behind a tag and do your recommendation of dropping "Sign In" because we are unsure if it will be understandable in other locales. Can't thank you both enough for your help on this issue.

flodolo:
> Does "Sign In" read weird in English? It doesn't seem strange to me. Can't speak if it will be broken in other languages, not sure how to find out.

It looks weird with Title Case.

PR #6936, opened 2019-03-19
Small release notes test

stephaniehobson:


stephaniehobson:
@alexgibson smaller test case just trying to test the value of the `data-gtm-page-id'`. Is this on the right track?

alexgibson:
@stephaniehobson this looks to be along the right track yes, although the test should probably be in the [release notes app](https://github.com/mozilla/bedrock/blob/master/bedrock/releasenotes/tests/test_base.py), and should use the `release_notes()` view instead of `new()`.

Digging into that `release_notes()` view does look pretty complicated though, maybe @pmac could advise better how to do this? 

An alternate method could be to use a headless test that doesn't require spinning up a browser. You can see an [example here](https://github.com/mozilla/bedrock/blob/master/tests/functional/test_download.py) that uses beautiful soup to parse the HTML response for a set of URLs. Using the marker `@pytest.mark.headless` will exclude it from functional tests.

PR #6880, opened 2019-03-02
Update Release Notes to Protocol

stephaniehobson:
## Description

Changed the release notes template to be based on Protocol.
- Created notes.html to be templates all channels (including release) use
  - notes.html extends base-protocol
  - Changed release-notes to extend notes.html
  - Included defaults to be over-ridden in channel templates
- Added custom typography for release notes
- Attempted to simplify template logic
  - Rely on `release` object values as much as possible instead of
    setting custom text values
  - Moved channel logic to channel templates as much as possible
- Standardized product names (Firefox Beta for Android -> Firefox for
  Android Beta)
- Added pre-quantum logos for all products
- Removed depreciated blocks:
  - `page_image`
  - `site_header_logo
  - `notes_heading_primary` (this is now the `primary_heading` variable)
- Removed unused assets
  - checked system_requirements does not rely on any of them
- Page headings no longer say "what's new" if they're not the latest
  release
  - old pages say "Release Notes"
- Updated release notes index page to use protocol-basic

## Known issues
- CSS needs comments

## Testing
- Removing the icon font was out of scope for this work, but it sure would be nice to have SVGs instead.
- Make sure I did not break system requirements pages.
- Release
  - http://localhost:8000/en-US/firefox/66.0/releasenotes/
  - http://localhost:8000/en-US/firefox/65.0/releasenotes/
  - http://localhost:8000/en-US/firefox/55.0/releasenotes/
  - http://localhost:8000/en-US/firefox/43.0/releasenotes/
  - http://localhost:8000/en-US/firefox/35.0/releasenotes/
- ESR
  - http://localhost:8000/en-US/firefox/52.7.1/releasenotes/
- Beta
  - http://localhost:8000/en-US/firefox/66.0beta/releasenotes/
  - http://localhost:8000/en-US/firefox/55.0beta/releasenotes/
- Aurora 2011 (Fx5) - Dev Ed 2014 - April 2017 (Fx54)
  - http://localhost:8000/en-US/firefox/54.0a2/auroranotes/
  - http://localhost:8000/en-US/firefox/41.0a2/auroranotes/
  - http://localhost:8000/en-US/firefox/32.0a2/auroranotes/
- Nightly
  - http://localhost:8000/en-US/firefox/67.0a1/releasenotes/
  - http://localhost:8000/en-US/firefox/55.0a1/releasenotes/ (includes .bug-id)
- Android
  - http://localhost:8000/en-US/firefox/android/60.0.2/releasenotes/
[...]

PR #6966, opened 2019-03-22
Release notes nav no longer sticky (#6880)

stephaniehobson:
## Description

Removing sticky menu in preparation for port to Protocol.

## Issue / Bugzilla link

As part of #6880 

## Testing
- Did I remove all files and functions?
- Any JS errors?
http://localhost:8000/en-US/firefox/60.0.2/releasenotes/

### DIFF
--- a/media/js/firefox/releasenotes.js
+++ b/media/js/firefox/releasenotes.js
@@ -4,23 +4,11 @@
 
 /* eslint no-unused-vars: [2, { "varsIgnorePattern": "stickyNav" }] */
 
-(function($, Mozilla, Waypoint) {
+(function($, Mozilla) {
     'use strict';
 
     var client = Mozilla.Client;
     var $html = $('html');
-    var $nav = $('#nav');
-
-    var stickyNav = new Waypoint.Sticky({
-        element: $nav,
-        handler: function(direction) {
-            if (direction === 'down'){
-                $nav.addClass('fixedNav');
-            } else{
-                $nav.removeClass('fixedNav');
-            }
-        }
-    });
 
     // navigation drop-down menu
     function initSubMenu() {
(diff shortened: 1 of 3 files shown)


media/js/firefox/releasenotes.js
```diff
@@ -4,23 +4,11 @@
...
 
     var client = Mozilla.Client;
     var $html = $('html');
-    var $nav = $('#nav');
-
-    var stickyNav = new Waypoint.Sticky({
-        element: $nav,
-        handler: function(direction) {
-            if (direction === 'down'){
-                $nav.addClass('fixedNav');
```
amychurchwell: It looks like there are some [css rules](https://github.com/mozilla/bedrock/blob/master/media/css/firefox/releasenotes-firefox.less#L108-L118) that could be deleted with the removal of this. 🙂

media/js/firefox/releasenotes.js
```diff
@@ -76,4 +64,4 @@
 
     initSubMenu();
 
-})(window.jQuery, window.Mozilla, window.Waypoint);
+})(window.jQuery, window.Mozilla);
```
amychurchwell: not a blocker-- but I know in other places we are removing jQuery. Do you think it's worth removing it from this file? 🤷‍♀️ or maybe it doesn't matter because this page is being ported to protocol?

media/js/firefox/releasenotes.js
stephaniehobson: ✅

media/js/firefox/releasenotes.js
stephaniehobson: I think all of this will be replaced by Protocol js in the port :)

stephaniehobson:
Updated to remove CSS.

Issue #6868, opened 2019-02-28
Update eslint to use @mozilla-protocol/eslint-config


alexgibson:
## Description

Now that bedrock has updated to Gulp 4, we should also update the version of ESLint that bedrock uses. This would also be a good time to switch to using the new [Protocol shared config](https://github.com/mozilla/eslint-config-protocol/).

### Estimates

| Estimates |  |
| ------------- | ------------- |
| TL | |
| PM | |
| ENG | 1 day |
| SEO | |
| ANA | |
| DES | |
| COPY | |

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Bedrock is using the latest version of ESLint.
- [ ] Bedrock config inerits from Protocol config.
- [ ] JS is free of linting errors.

#### Tasks
- [x] Update `gulp-eslint` dependency to the laterst NPM version.
- [x] Import new Protocol shared config.
- [x] Update any existing linting rules where necessary.

PR #7009, opened 2019-04-01
Update ESLint to consume @mozilla-protocol/eslint-config

alexgibson:
## Description
- Updates `eslint` and `gelp-eslint` to latest versions.
- Updates bedrock to consume `@mozilla-protocol/eslint-config`.
- Fixes linting errors.

## Issue / Bugzilla link
#6868

## Testing
TODO


jpetto:
With the help of @jgmize, we figured out the issue is with extracting a tarball that tries to replace a file with a directory of the same name (meaning the existing `gulpfile.js` _file_ is not being overwritten with the `gulpfile.js` _directory_). To get around this, we just need to delete the existing Jenkins workspace directory/directories so there's no old file to overwrite. This issue will come up in each stage of release - master, stage, and prod - so we'll need a back-end dev on-deck to manually delete directories in Jenkins. This should be a one-time issue. Once the directory exists in all the places, we should be good going forward.

Just be sure to coordinate with one of the back-end/SRE folks when this is ready to get to prod.

alexgibson:
Thanks @jpetto and @jgmize! This sounds like a royal pain, so considering that restructuring the gulpfile here is only facilitate a different set of linting rules on a single file, I'm going to find another solution that doesn't require all this manual effort.

jgmize:
We could potentially work around this by adding `sh 'rm -f gulpfile.js'` right above each of the places `unstash` is called in the groovy[0][1], which may allow the extraction of workspace.tar.gz.

[0] https://github.com/mozilla/bedrock/blob/master/jenkins/default.groovy#L33
[1] https://github.com/mozilla/bedrock/blob/master/jenkins/utils.groovy#L56

.eslintrc.js
```diff
@@ -1,52 +1,36 @@
...
-            "smart"
-        ],
-        "one-var-declaration-per-line": [
-            2,
-            "always"
-        ],
-        "new-cap": 2
+    extends: '@mozilla-protocol/eslint-config',
+    rules: {
+        'no-useless-escape': 1,
```
alexgibson: Note for reviewer: I didn't go and fix all the regular expressions this rule picked up on, as it would have required a lot of manual testing. I set this to a warning instead, and we can work on eliminating these over time.

alexgibson:
Ok, this is finally ready for code review.

Issue #7015, opened 2019-04-02
Download tests are hanging


jgmize:
## Description

The download tests are repeatedly hanging. 

- https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/master/1587/pipeline
- https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/master/1588/pipeline
- https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/prod/402/pipeline

Running them in isolation, they passed: https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/276/pipeline

alexgibson:
I believe this issue resolved itself, closing,

PR #7016, opened 2019-04-02
Temporarily skip download link tests for stage and prod (#7015)

alexgibson:
#7015


jgmize:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1541227

Issue #7028, opened 2019-04-05
Issues running smoke tests whilst pushing to run-integration-tests branch


alexgibson:
## Description

I'm trying to push the latest bedrock `master` branch to `run-integration-tests`.

### Steps to reproduce

```
git push upstream master:run-integration-tests
```

### Expected result

The test run should be green, like the [last commit to master](https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/master/1592/pipeline).

### Actual result

The `smoke` tests for the /firefox/mobile/ page [all fail](https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/289/tests).

Running the tests locally, they all pass. I also deployed to demo and ran the tests locally against that, and they also still pass.

The last person to [push a job to the testing branch ](https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/287/pipeline/)was making changes on the /firefox/mobile/ page. Is there a chance that there's some issue with the workspace, or there's still different code deployed? I'm, at a loss as to what's going on, as the tests passed fine on the last commit to master.

To make sure I'm not going crazy, I tried [rerunning the last master job](https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/master/1594/pipeline/), but now I get a different error:

```
Error response from daemon: Conflict. The container name "/bedrock-code-master-01acfb05c98bfe64fd433b6b743fd5ea97a565ee" is already in use by container "cffc0b09ffcd5e338a5124ac6c9c0a6e33ce9dc15f8d65f0d22e897b042645ab". You have to remove (or rename) that container to be able to reuse that name.
```

And now I'm even more confused :/

alexgibson:
@jgmize any ideas what might be happening here?

alexgibson:
I'd also like to bring this up as an opportunity to discuss how much value these smoke tests are adding. They all get run agains each deployment anyway, and are super difficult to debug because they are inside a remote docker container?

alexgibson:
Ok, the latest commit to master was fine: https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/master/1595/pipeline/39/

I'm going to try pushing this to `run-integration-tests` and see what happens.

alexgibson:
[...]

PR #7030, opened 2019-04-05
Remove smoke marker from tests

jgmize:
## Description

This PR eliminates the deployment pipeline step of running the smoke tests against a local instance of bedrock and a local selenium grid in docker containers on the jenkins instance. The smoke marker is still used in other situations, including running against the deployed instances.

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/7028

## Testing

I pushed this branch to run-integration-tests branch, triggering https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/293/pipeline


docs/testing.rst
```diff
@@ -196,7 +196,6 @@ you can also read the `pytest markers`_ documentation for more options.
 Smoke tests
```
alexgibson: Should we just remove this section, seeing as the smoke markers don't do anything any longer?

docs/pipeline.rst
```diff
@@ -37,10 +37,8 @@ unit tests, as well as the smoke suite of redirect headless HTTP(s) response che
 Push to master branch
 ~~~~~~~~~~~~~~~~~~~~~
 
-Whenever a change is pushed to the master branch, the smoke suite of
-headless (see :ref:`testing-redirects`) and UI tests (see :ref:`smoke-functional-tests`)
```
alexgibson: Hmm, should we still run the redirect tests at this step? I'm not clear if those will get run against a deployment or not, looking at the current pipeline?

alexgibson: Could we move the redirect tests to run against a deployment instead perhaps?

docs/testing.rst
jgmize: This PR removes the machinery for running the smoke tests via local docker container, but it doesn't remove the smoke markers themselves, so the note in this section about running them from the command line still applies.

docs/pipeline.rst
jgmize: I was under the impression that we were already running the redirect tests against the deployments as part of the headless test suite, but I'm double checking now.

jgmize: @alexgibson we are running the redirect tests against remote, but we don't currently appear to run tests with the smoke marker against remote deployments. I've pushed a new commit that should run them against the deployments in parallel to the other tests, and that's running in https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/296/pipeline/ now

alexgibson: @jgmize at this point I really wonder what use the smoke markers are doing. Why run those only in Firefox in a separate run, and not just run all the tests for each browser?

alexgibson: Also, I do see the smoke tests running against the other browser runs. Happy to chat about this more in person if it helps.

jgmize: Good points @alexgibson, I agree the smoke markers aren't adding any value anymore, and I've pushed a new commit which eliminates them entirely and all references to them.

docs/testing.rst
jgmize: As per my other comment, this PR now removes the smoke markers entirely, along with this section.

jgmize:
https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/297/pipeline/ is a successful run triggered by pushing https://github.com/mozilla/bedrock/commit/be619e61cabe37459073a174e683dd7cdcf04d8d to the run-integration-tests branch, and the CircleCI failure appears to have been a temporary infrastructure issue rather than one with the code; I'm rerunning it now.

alexgibson:
@jgmize this looks great. There's a flake8 error, but aside from that I think this is good to merge 👍

jgmize:
Thanks @alexgibson, I've pushed a new commit to fix the flake8 error.

alexgibson:
Thanks, @jgmize 💯

Issue #7053, opened 2019-04-10
Desktop home page hero image is 4x retina, using 2x saves 330kb load


joeyfb:
## Description

Max width of homepage hero image, `fxa-hero-img.png` across breakpoints is 584px, so 2400px high res image image might be too much. Removing `at2x` and hard coding 1200px wide image would save 330kb initial download on desktop home page.

Only downside I see is super high density displays, above 3x, would have a lower quality.  Apologies if I'm missing that 4x high DPI screens should be targeted, didn't see it in best practices!

Alternative enhancement could be using 1200px as `fxa-hero-img-high-res.png` and having a 800-600px width image as `fxa-hero-img.png`.

- mobile doesn't show the hero image, not affected
- in between widths have ~360px width image so could potentially use lower res image and keep 2x resolution
- desktop image is only 584px wide 1200px width is still 2x retina

### Steps to reproduce

Visit https://www.mozilla.org/en-US/ at desktop width.
View network tab, see `fxa-hero-img-high-res.png` image loaded
Hero image is 584px wide but only needs to 1200px for 2x retina

**Width of desktop image**
<img width="1141" alt="Screen Shot 2019-04-10 at 8 36 25 AM" src="https://user-images.githubusercontent.com/5576685/55892966-2cef8100-5b6c-11e9-9b5b-c282564eaa2a.png">

**468kb load for image**
<img width="1440" alt="Screen Shot 2019-04-10 at 8 36 08 AM" src="https://user-images.githubusercontent.com/5576685/55892983-37117f80-5b6c-11e9-8300-d4dfdcb6b374.png">

### Expected result

2x retina for lighter load

### Actual result

4x retina image with higher load

### Environment

MBP 13, MacOS, firefox,

alexgibson:
Great bug report, thanks!

alexgibson:
Fixed in https://github.com/mozilla/bedrock/pull/7054

PR #7054, opened 2019-04-10
Desktop homepage hero image using normal res #7053

joeyfb:
## Description

Shifts down resolution of `fxa-hero-img` used on homepage and account page

- `fxa-hero-img-high-res.png` goes from 2400px wide to 1200px
- `fxa-hero-img.png` goes from 1200px wide to 800px

Since largest width of image is 584px on homepage and 680px on accounts there shouldn't be a perceived lost of quality.

## Issue / Bugzilla link

#7053

## Testing

- For http://localhost:8000/en-US/ and http://localhost:8000/en-US/firefox/accounts/
  - On widest break point check that 1200px `fxa-hero-img-high-res.png` loaded without perceived lost of quality
  - On narrower breakpoints check that 800px `fxa-hero-img.png` loaded without perceived lost of quality
- Narrowest break point doesn't show hero image so shouldn't load either


alexgibson:
Many thanks for this pull request @joeyfb, and for filing the issue! This is already a good saving, but I quite like your alternative suggestion in the issue:

> Alternative enhancement could be using 1200px as fxa-hero-img-high-res.png and having a 800-600px width image as fxa-hero-img.png.

This would mean visitors on low-res displays have an even smaller download. Worth trying?

joeyfb:
Definitely! I'll look into it tonight and make a new pull request keeping `at2x` code and shifting resolutions down so `fxa-hero-img-high-res.png` becomes new high res.

joeyfb:
New commit shifts down resolutions, pull request description updated to reflect it

Issue #7035, opened 2019-04-05
Add non-Fx content to FxA benefits page


ejregithub:
## Description

60% of email referrals to FxA benefits page are in non-Fx browsers. Many came from e.g. an email they opened in Mobile Safari or Chrome on Android. The email said, "Get an Account". When they get to this page, they see a big "Get Firefox" button. This is a broken experience.

We should add content, shown to non-Fx visitors, that helps them access a smaller set of services that are supported by FxA but not dependent on Firefox. As far as I know, these services include...

* Send
* Monitor
* Notes
* Pocket

The content for these services is already substantially produced on e.g. [the Whatsnew 66 page](https://www.mozilla.org/en-US/firefox/66.0/whatsnew/), though not in exactly the right form factor. 

This is a short-term tactic to maximize FxA signups by serving non-Fx users landing on this page. The long-term solution is to rethink how we market FxA and build FxA funnels on the site, but that is *out of scope* for this issue.

[Plan](https://docs.google.com/document/d/1_nTLCcVZ5UcbT60yxdj4KI4OkFKCGVrAR2dVUo862vc/edit?usp=drivesdk)

### Estimates

| Estimates |  |
| ------------- | ------------- |
| TL | |
| PM | |
| ENG | |
| SEO | |
| ANA | |
| DES | |
| COPY | |

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] A chrome user landing on /firefox/accounts/ can find a way to make an account, if they want to access the non-Firefox-dependent services we offer.

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] We may see a drop in Fx acquisitions as a result

#### Tasks
- [ ] produce UX specifications
- [ ] developer review of UX
- [ ] design for new specifications (unlikely)
- [ ] copy for new specifications (likely)
- [ ] code changes to accomplish UX

alexgibson:
Related issues:

- https://github.com/mozilla/bedrock/issues/6907
- https://github.com/mozilla/bedrock/issues/6789
- https://github.com/mozilla/bedrock/issues/6837

hoosteeno:
Blocked by https://github.com/mozilla/bedrock/issues/7062

hoosteeno:
This is ready for review from 2 perspectives: 
1. Code
2. UX/UI

hoosteeno:
[...]

Issue #6953, opened 2019-03-20
Change log-in to login on Fx Benefits page


hoosteeno:
## Description
The page at https://www.mozilla.org/en-US/firefox/accounts/ has a headline including the string "log-in". This should be changed to "login" the next time we make major string changes on the page (e.g. in  card https://github.com/mozilla/bedrock/issues/6914).

![Screen Shot 2019-03-20 at 11.35.41 AM.png](https://images.zenhubusercontent.com/5c871077d74bd07972c6db24/3116a5a3-285a-4b6a-84eb-f54296db9f7c)

Depends on https://github.com/mozilla/bedrock/issues/6914.

### Estimates

| Estimates |  |
| ------------- | ------------- |
| TL | |
| PM | |
| ENG | |
| SEO | |
| ANA | |
| DES | |
| COPY | |

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] "log-in" is changed to "login" on the FxA benefits page, and that change is ready for localizers.

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] This is an example of a risk. Use a checklist!
- [ ] This is an example of a risk.

#### Tasks
- [ ] This is an example of a task. Use a checklist!
- [ ] This is an example of a task.

Issue #6914, opened 2019-03-14
Add Firefox Send to the /firefox/accounts/ page


alexgibson:
## Description

Firefox Send (https://send.firefox.com/) has now launched and adds a real incentive for signing up and using a Firefox Account (send files up to 2.5gb if you're signed-in). It is also not dependent on using Firefox and works with any browser. 

We should add Send as a feature to the /accounts/ page as a benefit.

### Estimates

| Estimates |  |
| ------------- | ------------- |
| TL | |
| PM | |
| ENG | |
| SEO | |
| ANA | |
| DES | |
| COPY | |

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Design and copy delivered by the end of the sprint (ends 4/12)
- [ ] Is tagged for L10n

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] This is an example of a risk. Use a checklist!
- [ ] This is an example of a risk.

#### Tasks
- [ ] Identify Celeste as A in RASCI, Michele likely as C
- [ ] Clay to write Copy
- [ ] Websites to create Design (Mark)

hoosteeno:
Whoever picks this up, please see https://github.com/mozilla/bedrock/issues/6953.

mrkwvr:
> Do we have design and copy for this yet?

no not yet, Clay and I will be working on those

claychaffin:
Hi All, I'm *blocked* on _accurate_ Send copy until we get a messaging brief. I've asked Celeste/Lauren for it in FxA - benefits channel.

claychaffin:
https://github.com/mozilla/bedrock/issues/6914
Hi All,  I'm *blocked* on _accurate_ Send copy until we get a messaging brief. I've asked Celeste/Lauren for it in FxA - benefits channel.

hoosteeno:
This is ready for design and copy review. Copy and image in attached screenshot comes from WNP66.

![Screen Shot 2019-04-12 at 5 18 04 PM](https://user-images.githubusercontent.com/94519/56070746-f3bf3880-5d46-11e9-8396-91bfba2b3f38.png)

hoosteeno:
@claychaffin @mrkwvr Can you take a look? [Full screenshot for context here](https://user-images.githubusercontent.com/94519/56070849-9ecff200-5d47-11e9-8073-cbadc28f584f.png).

claychaffin:
[...]

Issue #6789, opened 2019-02-04
Update firefox/accounts modal links to buttons


amychurchwell:
Semantically, the links on firefox/accounts that trigger the QR code modal window should be buttons as they do not take the user to a new URL. 

<img width="1287" alt="screen shot 2019-02-04 at 1 08 04 pm" src="https://user-images.githubusercontent.com/30483988/52237300-02d4c680-287e-11e9-8577-9df7d8c6171c.png">

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Demo reviewed by designer

PR #7069, opened 2019-04-12
Add send to FxA Benefits. Handle non-FxA better.

hoosteeno:
## Description
This PR attempts to address the following issues:

#6914
#7035
#6953
#6789

Attached is a screenshot of the non-Fx rendering of this page.
localhost_8000_en-US_firefox_accounts_ (3)


bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -66,14 +77,26 @@ <h2 id="subtitle">{{ _('Own your life on every device with a Firefox Account.')
     class='mzp-l-card-feature-right-half mzp-t-firefox'
   ) %}
   <p>{{ _('Send an open tab on one device to all your others with a single tap. Much easier than texting or emailing yourself those links.') }}</p>
+  <ul class="show-fxa-not-fx">
+      <li>
```
stephaniehobson: It doesn't need to be a list if there's only one item in it ;)

craigcook:
All the new and/or changed strings need to be behind an l10n tag so locales can update in their own time. It should have a descriptive name, perhaps including a date since we can expect this page to get more updates in due time. Maybe `firefox_send_042019`?

See https://bedrock.readthedocs.io/en/latest/l10n.html#staging-copy-changes

craigcook:
Oh also tests are failing but they're minor issues:
```
tests/functional/firefox/test_accounts.py:10:1: E302 expected 2 blank lines, found 1
tests/functional/firefox/test_accounts.py:25:1: W391 blank line at end of file
tests/pages/firefox/accounts.py:11:1: F401 'DownloadButton' imported but unused
tests/pages/firefox/accounts.py:13:1: E302 expected 2 blank lines, found 1
```

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -58,28 +55,35 @@ <h2 id="subtitle">{{ _('Own your life on every device with a Firefox Account.')
...
     image_url='firefox/accounts/fxa-lockbox.png',
     include_highres_image=True,
     aspect_ratio='mzp-has-aspect-1-1',
-    class='mzp-l-card-feature-left-half mzp-t-firefox lockbox-card',
-    link_cta=_('Get the Lockbox App'),
-    link_url='https://app.adjust.com/kq8zhhz',
+    class='mzp-l-card-feature-left-half mzp-t-firefox lockbox-card qr-modal',
     ga_title='Get the Lockbox App'
   ) %}
     <p>{{ _('No more getting locked out of apps and websites. Lockbox secures all the passwords you’ve saved in Firefox and gives you easy access across all your iOS devices.') }}</p>
```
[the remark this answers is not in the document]
hoosteeno: This seems like a good change, but the plumbing doesn't seem to be in place for it. The QR code talks about Apple. I just don't feel confident that we have all the bits in place to accomplish it in this PR, and I don't want to delay this PR while we figure them out. A follow-up issue seems reasonable!

tests/pages/firefox/accounts.py
```diff
@@ -2,24 +2,35 @@
...
-    _download_button_locator = (By.ID, 'features-hero-download')
+    _download_buttons_locator = (By.CLASS_NAME, 'features-download')
     _create_account_form_locator = (By.ID, 'fxa-email-form')
 
     @property
-    def download_button(self):
-        el = self.find_element(*self._download_button_locator)
-        return DownloadButton(self, root=el)
+    def is_create_account_form_displayed(self):
+        wait = Wait(self, timeout=3)
```
alexgibson: We shouldn't need to add any implicit waits like this, as our tests only run once the page has fully loaded and JS has executed.

tests/functional/firefox/test_accounts.py
```diff
@@ -7,17 +7,19 @@
 from pages.firefox.accounts import FirefoxAccountsPage
 
 
-@pytest.mark.skip_if_firefox(reason='Download button is only displayed to non-Firefox users')
+# test Firefox state
+@pytest.mark.skip_if_not_firefox(reason='Non-Firefox user does not see Firefox state')
 @pytest.mark.nondestructive
-def test_download_button_displayed(base_url, selenium):
+def test_firefox_state(base_url, selenium):
```
alexgibson: Try and keep test names descriptive. If this fails in CI, does it give a good indication of what might be wrong? I'd suggest keepoing this something along the lines of `test_download_buttons_displayed`

tests/functional/firefox/test_accounts.py
```diff
@@ -7,17 +7,19 @@
...
-    assert page.download_button.is_displayed
+    assert not page.are_download_buttons_displayed, "Download buttons are displayed for Firefox user"
+    assert page.is_create_account_form_displayed, "Create account form is not displayed for Firefox user"
 
 
-@pytest.mark.skip_if_not_firefox(reason='Create Account form is only displayed to Firefox users')
+# test non-Firefox state
+@pytest.mark.skip_if_firefox(reason='Firefox user does not see non-Firefox state')
 @pytest.mark.nondestructive
-def test_create_account_form_displayed(base_url, selenium):
```
alexgibson: `test_firefox_accounts_buttons_displayed`?

tests/pages/firefox/accounts.py
```diff
@@ -2,24 +2,35 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+from selenium.common.exceptions import TimeoutException
 from selenium.webdriver.common.by import By
+from selenium.webdriver.support import expected_conditions as expected
+from selenium.webdriver.support.ui import WebDriverWait as Wait
 
 from pages.firefox.base import FirefoxBasePage
-from pages.regions.download_button import DownloadButton
```
alexgibson: We should always try and use the download button page region, as this asserts that there's only one download link displayed within the button. Without this, the download button CSS could be busted and the test could still pass.

tests/pages/firefox/accounts.py
```diff
@@ -2,24 +2,35 @@
...
 
 
 class FirefoxAccountsPage(FirefoxBasePage):
 
     URL_TEMPLATE = '/{locale}/firefox/accounts/'
 
-    _download_button_locator = (By.ID, 'features-hero-download')
+    _download_buttons_locator = (By.CLASS_NAME, 'features-download')
```
alexgibson: I believe right now the test is only matching the first instance of the download button it finds in the page. We should probably add an assert for each instance, so if one breaks we can know more easily which one has an issue.

alexgibson:
@hoosteeno can you please rebase your branch against master? I have a [commit here](https://github.com/alexgibson/bedrock/commit/a44e080f02169bd76f47da0edf7eb0f2a0a5d968) to update the functional tests, and am happy to open a PR but your branch is a few commits behind mine.

alexgibson:
I've approved the test changes, but I'll leave it to @craigcook for the final approval here.

bedrock/firefox/templates/firefox/accounts-2018.html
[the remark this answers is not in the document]
craigcook: Well my only misgivings there are that it will mean even more string changes in the very near future, I think it would be better on l10n to bundle them up. Maybe we can resolve this quickly and not delay the PR. Who do we need to talk to?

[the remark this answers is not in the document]
craigcook: The `https://app.adjust.com/kq8zhhz` has already been updated to be cross-platform, so I think all the plumbing is in order, we just need to update copy.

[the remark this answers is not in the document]
craigcook: So let's change this to "across all your Android and iOS devices" (behind the same l10n tag)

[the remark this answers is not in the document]
hoosteeno: Done.

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -58,29 +59,43 @@ <h2 id="subtitle">{{ _('Own your life on every device with a Firefox Account.')
...
+  {% if l10n_has_tag('firefox_send_20190420') %}
+    {% call feature_card(
+      title=_('Take total control of the files you share'),
+      image_url='firefox/accounts/fxa-fxsend.png',
+      include_highres_image=True,
+      aspect_ratio='mzp-has-aspect-1-1',
+      class='mzp-l-card-feature-right-half mzp-t-firefox',
+      ga_title='Try Firefox Send'
+    ) %}
+    <p>{{ _('Feel like a secret agent when you send files online, with a private, encrypted link that ‘self-destructs.’') }}</p>
```
craigcook: I just noticed: should these be double quotes around `self destructs` rather than single quotes?

craigcook: Clay says single quotes, so I guess we'll go with that.

PR #7088, opened 2019-04-18
Bug 1542118 - Unify Bengali as 'bn', remove obsolete locales

flodolo:
Locales removed:
* bn-BD, bn-IN (now 'bn')
* as, en-ZA, mai, or

Also removed references to en-ZA from code.


flodolo:
Need to confirm with @pmac what's the status of legal-docs

pmac:
Legal docs are ready, I can merge this and then immediately update legal docs to the version that only has `bn` and we should be good. Thanks flod!

PR #7091, opened 2019-04-18
Update legal-docs to add bn locale and remove bn-* ones.

pmac:
Re bug 1542118


pmac:
This is the followup to #7088.

pmac:
I've tried it locally and everything seems to be working.

PR #7106, opened 2019-04-23
[fix #7067] Add FxA email-first form to legacy firstrun page

### LINKED ISSUE #7067: Add email-first form to backup/default firstrun page
## Description

The backup/default firstrun experience that we recently changed in #6149 was an important source of new FxA. Now that we've removed the signup form, it is no longer an important source, and as a result FxA acquisitions are noticeably down (https://github.com/mozilla/fxa/issues/818). 

We should add the iframeless form to the backup/default firstrun experience.

### Estimates

| Estimates |  |
| TL | |
| PM | |
| ENG | |
| SEO | |
| ANA | |
| DES | |
| COPY | |

---

- [ ] The backup/default firstrun experience has an iframeless email-first FxA signup form
- [ ] The form's contributions to FxA signup/signin are observable in Amplitude

#### Tasks
- [ ] This is an example of a task. Use a checklist!
- [ ] This is an example of a task.

craigcook:
## Description
Adds the FxA email-first form to the legacy firstrun page, which is seen by older versions of Firefox when they're run for the first time. This also ports the page to Protocol (not really using Protocol components, it's just the base framework).

## Issue / Bugzilla link
#7067 

## Testing
https://bedrock-demo-craigcook.oregon-b.moz.works/en-US/firefox/firstrun/


media/css/firefox/firstrun/firstrun-quantum.scss
```diff
@@ -35,12 +39,9 @@ body {
...
                 linear-gradient(to bottom, #003eaa 40%, #004ec2 60%, #0060df 80%, #0080ff 90%, #00c7ff 100%) top center no-repeat,
                 #003eaa;
-    margin: 0;
-    min-height: 100vh;
-}
-
-* {
     font-family: 'Fira Sans Light', sans-serif;
+    min-height: 100vh;
+    -moz-osx-font-smoothing: unset;
```
craigcook: This bit was needed because the global font smoothing from Protocol really doesn't play well with Fira Sans Light on this background. Compared to the Pebbles-based page the text was extra thin. I wonder if we should remove the font smoothing in Protocol... but then again, it's really only a problem with light fonts so I think this hack will work for now on this one page (which needs a rebranding soon anyway).

bedrock/firefox/templates/firefox/firstrun/firstrun-quantum.html
```diff
@@ -4,23 +4,44 @@
...
+{% block extrahead %}
+  {% if switch('firefox-update-notification-modal', ['en-US']) %}
+    {{ css_bundle('firefox-update-notification-firstrun-whatsnew') }}
+  {% endif %}
+{% endblock %}
+
+{# Bug 1387984 #}
+{% block update_notification %}
+  {% if switch('firefox-update-notification-modal', ['en-US']) %}
+    {% include 'includes/firefox-update-instructions.html' %}
```
alexgibson: Passing note: We've never ported this notification to Protocol, so I'm not sure how well it will play here without some testing?

craigcook: Good point, I'll test it. I think I can try installing an old Firefox, or else just fake it locally.

bedrock/firefox/templates/firefox/firstrun/base-pebbles.html
```diff
@@ -1,42 +0,0 @@
-{# This Source Code Form is subject to the terms of the Mozilla Public
```
alexgibson: 🎉

bedrock/firefox/templates/firefox/firstrun/firstrun-quantum.html
```diff
@@ -4,23 +4,44 @@
...
+    {{ js_bundle('firefox-update-notification-firstrun-whatsnew') }}
+  {% endif %}
+{% endblock %}
+
 {% block global_nav %}{% endblock %}
 {% block site_header %}{% endblock %}
 
 {% block content %}
-  <div id="scene">
+  <div id="scene" class="t-firefox">
```
alexgibson: Is this still needed now all pages default to Firefox branding?

bedrock/firefox/templates/firefox/firstrun/firstrun-quantum.html
craigcook: This was a good catch. The update banner was expecting styles from sandstone base, which were obviously missing from Protocol so it was unstyled. Added the styles to the bundle for this page and it looks good, despite the mix of frameworks. 

At some point we'll need to decide the fate of this banner and either kill it or redesign it for Protocol.

bedrock/firefox/templates/firefox/firstrun/firstrun-quantum.html
```diff
@@ -38,10 +59,13 @@ <h1>
...
         {% endif %}
         </p>
+        <p><a href="{{ url('firefox.accounts') }}">{{_('Learn more about Firefox Accounts')}}</a></p>
       {% endblock %}
       </header>
       <div id="fxaccounts-wrapper">
-        <a class="button" href="{{ url('firefox.accounts') }}">{{_('Learn more about Firefox Accounts')}}</a>
+        {{ fxa_email_form(entrypoint='firstrun-page', utm_source='firstrun-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'firstrun-page', 'medium': 'referral'}, button_class='mzp-c-button mzp-t-product',) }}
+
+        <p class="fxa-signin">{{ _('Already have an account?') }} <a {{ fxa_link_fragment(entrypoint='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': '/firefox/firstrun/', 'source': 'firstrun-page'}) }}>{{ _('Sign In') }}</a></p>
```
alexgibson: Nit: suggest chanign to `'content': 'firefox-firstrun'`? Looking at other pages, we seem to favor dashes over trailing slashes?

media/static-bundles.json
```diff
@@ -1487,6 +1490,9 @@
     },
     {
       "files": [
+        "js/base/mozilla-fxa.js",
```
alexgibson: I don't think we need this file to determine signed in state, is that right?

media/static-bundles.json
```diff
@@ -658,6 +659,8 @@
     },
     {
       "files": [
+        "css/base/mozilla-fxa-state.scss",
```
alexgibson: Same with this file, not sure we need this?

alexgibson:
I tested in Firefox 32 and I see a blank screen with an error:

Screenshot 2019-04-25 at 18 00 14

Would adding a feature tetect in `mozilla-fxa-form.js` fix it?

```javascript
var supportsFetch = 'fetch' in window;

if (!supportsFetch) {
    return;
}
```

media/static-bundles.json
craigcook: Nope, we're doing anything to detect state, or at least I wasn't given any guidance on what to show to signed-in users, and since it's firstrun it's unlikely folks seeing it would already be signed in. So I'll remove the state detection bits.

bedrock/firefox/templates/firefox/firstrun/firstrun-quantum.html
craigcook: Nope, and I ended up not even doing anything with this class anyway, this is an artifact. I'll remove.

craigcook:
Updated and rebased. I've tested it in Firefox 13.0, which is the oldest working installer I have handy (probably have even older ones on backup drives somewhere...). It needed a few more CSS fixes to be presentable in such an outdated Firefox, but the supportsFetch check does at least let it display.

Issue #7047, opened 2019-04-08
Fundraising banner to follow Internet Health Report Release


WillatMozFdn:
## Description

MoFo is planning a late-April fundraising campaign and wants to bring back the home page banner for a brief time, with minor changes from Valentine's Day campaign: a) plain black background, and b) "Variant B" copy from the V-day campaign, as was implemented here: https://github.com/mozilla/bedrock/issues/6693 . It will be English-only.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [x] Banner launched April 26
- [ ] Banner pulled down May 1st

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] This is an example of a risk. Use a checklist!
- [ ] This is an example of a risk.

#### Tasks
- [x] Code it
- [x] Code switch

ejregithub:
This is live. Eric will file a separate issue to turn it off.

PR #7107, opened 2019-04-24
[#7047] Add April 2019 fundraising banner

craigcook:
## Description
Adds a fundraising banner to coincide with publication of the new Internet Health Report. Uses the switch name `home-fundraising-apr2019` and it's English only. I'll file a www-config PR to flip the switch once this makes it to prod. The original request was for the 26th but the report is already out so we could probably do it sooner.

## Issue / Bugzilla link
#7047 

## Testing
https://bedrock-demo-craigcook.oregon-b.moz.works/en-US/

- [ ] Banner displays on the home page in all English locales (en-US, en-GB, en-CA).
- [ ] Banner does **not** display on the German home page.
- [ ] Banner does not display on the home page in other locales (they're still on the old template).
- [ ] Dismissing the banner sets a 1 day cookie named `fundraiser-apr2019`.
- [ ] When the cookie is present, the banner remains hidden.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/home/home-en.html
+++ b/bedrock/mozorg/templates/mozorg/home/home-en.html
@@ -42,6 +42,10 @@
 {% endblock %}
 
 {% block site_header %}
+  {% if LANG.startswith('en') and switch('home-fundraising-apr2019') %}
+    {% include 'mozorg/home/includes/fundraiser-apr2019.html' %}
+  {% endif %}
+
   {% include 'includes/protocol/navigation/index.html' %}
 {% endblock %}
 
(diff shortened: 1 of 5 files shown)


amychurchwell:
Not a blocker -- but I'm curious if we want to dismiss the banner for two days? It's not persistent across the site. I think we should dismiss it per session, not set a cookie. Or at most 1 day?

craigcook:
> Not a blocker -- but I'm curious if we want to dismiss the banner for two days? It's not persistent across the site. I think we should dismiss it per session, not set a cookie. Or at most 1 day?

Good point. This won't be running for long anyway. I've changed the cookie to one day.

Issue #6854, opened 2019-02-26
Pass through utm parameters on FxA referrals


hoosteeno:
## Description

Pass through utm parameters from upstream referrers when referring www.mozilla.org visitors to FxA signup form. *Continue sending an entrypoint generated by mozilla.org.* [Bug with detailed specs](https://bugzilla.mozilla.org/show_bug.cgi?id=1527318#c6).

Here's a complete URL, and I'll break it down.

`https://accounts.firefox.com/signup?service=sync&context=fx_desktop_v3&entrypoint=mozilla.org-globalnav&utm_source=desktop-snippet&utm_content=rel-esr&utm_medium=referral&utm_term=4242&utm_campaign=F100_4242_otherstuff_in_here

Breakdown:

`?service=sync&context=fx_desktop_v3`
Required by FxA server.
`&entrypoint=mozilla.org-globalnav`
This is the entrypoint as defined in FxA client metrics spec[0]. Where did the user enter from? From the global nav on mozilla.org. On mozilla.org, we should use a standard format (see other issue).

`&utm_source=desktop-snippet&utm_content=rel-esr&utm_medium=referral&utm_term=4242&utm_campaign=F100_4242_otherstuff_in_here`
These are all the standard UTM parameters that snippets and emails and other referrers send. We should validate these using the validation logic used by the FxA server, and if valid, we should pass them through to FxA so they end up in Amplitude, where we can use them to understand upstream referrers.

If there are not valid UTM parameters, we should add site-default utm parameters from Mozilla.org in the following format:

`utm_content=[urlsafe content of CTA]&utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=[urlsafe category of CTA]`

Here is an example from our current site:

`utm_content=get-firefox-account&utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=globalnav`

### Estimates (How much time in hours (up to 2), then in 1/2 day increments it will take each role to complete the work required to meet the Success Criteria established for this issue).

| Estimates |  |
| ------------- | ------------- |
| TL = TeamLead | |
| PM = Program Manager | |
| Eng = Engineer | 3 |
| SEO | |
| ANA = Analytics | |
| DES = Design | |
| COPY | |

---

#### :yellow_heart: Success Criteria :yellow_heart:
For every referral point on the website, ensure...
- [x] pre-existing utm parameters are sent on to FxA servers when referrals occur
[...]

PR #7110, opened 2019-04-24
FxA validation of utm parameters (#6854)

amychurchwell:
## Description
Per conversation in #6854
https://github.com/mozilla/bedrock/issues/6854#issuecomment-482443889
This is a rough draft of validating utm_source params.

## Issue
#6854

## Testing

Search url for preexisting source params.
1. if utm_source is present, [validate with regex](https://mozilla.github.io/application-services/docs/accounts/metrics.html#descriptions-of-metrics-related-query-parameters)
2. if utm_source is not present, use value passed to `fxa_link_fragment` macro.

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -247,12 +247,10 @@
 
 {%- endmacro %}
 
-{% macro fxa_cta_link(button_location, id_prefix, campaign, entrypoint, button_text = '', button_class = '', content = '', source = '', action='signup', utm_params={}) %}
-<div class="show-fxa-supported-signed-out">
-  <a {{ fxa_link_fragment(entrypoint="{{ entrypoint }}", action="{{ action }}", utm_params={'campaign': campaign, 'content': (content or request.path_info), 'source': source}) }} class="js-fxa-cta-link {{ button_class }}" data-button-name="Create account" data-link-type="FxA-Sync" data-cta-position="{{ button_location }}" id="{{ account_id }}">
+{% macro fxa_cta_link(entrypoint, campaign, content, source, button_class, button_location, button_text, account_id, action='signin', utm_params={}) %}
+  <a {{ fxa_link_fragment(entrypoint=entrypoint, action="{{ action }}", utm_params={'campaign': campaign, 'content': (content or request.path_info), 'source': source}) }} class="js-fxa-cta-link {{ button_class }}" data-button-name="Create account" data-link-type="FxA-Sync" data-cta-position="{{ button_location }}" id="{{ account_id }}">
     {{ button_text }}
   </a>
-</div>
 {%- endmacro %}
 
 {% macro fxa_create_account_button(button_location, id_prefix, campaign, entrypoint, button_text = '', button_class = '', content = '', source = '') %}
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -243,7 +243,7 @@
     {% endif %}
   {% endfor %}
 
-  href="{{ settings.FXA_ENDPOINT }}{{ fxa_path }}?service={{ service_type }}&{{ fxa_queries|join('&') }}" data-mozillaonline-link="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}{{ fxa_path }}?{{ fxa_queries|join('&') }}"
+  href="{{ settings.FXA_ENDPOINT }}{{ fxa_path }}?service={{ service_type }}&{{ fxa_queries|join('&') }}" data-mozillaonline-link="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}{{ fxa_path }}?service={{ service_type }}&{{ fxa_queries|join('&') }}"
 
 {%- endmacro %}
 
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -225,10 +225,9 @@
   </section>
 {%- endmacro %}
 
-{% macro fxa_link_fragment(entrypoint, action='signin', utm_params={}) -%}
+{% macro fxa_link_fragment(entrypoint, service_type='sync', action='signin', utm_params={}) -%}
   {% set fxa_path = '' %}
   {% set fxa_queries = [
-    'service=sync',
     'context=fx_desktop_v3'
   ] %}
 
@@ -244,7 +243,16 @@
     {% endif %}
   {% endfor %}
 
-  href="{{ settings.FXA_ENDPOINT}}{{ fxa_path }}?{{ fxa_queries|join('&') }}" data-mozillaonline-link="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}{{ fxa_path }}?{{ fxa_queries|join('&') }}"
+  href="{{ settings.FXA_ENDPOINT }}{{ fxa_path }}?service={{ service_type }}&{{ fxa_queries|join('&') }}" data-mozillaonline-link="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}{{ fxa_path }}?{{ fxa_queries|join('&') }}"
+
+{%- endmacro %}
+
+{% macro fxa_cta_link(button_location, id_prefix, campaign, entrypoint, button_text = '', button_class = '', content = '', source = '', action='signup', utm_params={}) %}
+<div class="show-fxa-supported-signed-out">
+  <a {{ fxa_link_fragment(entrypoint="{{ entrypoint }}", action="{{ action }}", utm_params={'campaign': campaign, 'content': (content or request.path_info), 'source': source}) }} class="js-fxa-cta-link {{ button_class }}" data-button-name="Create account" data-link-type="FxA-Sync" data-cta-position="{{ button_location }}" id="{{ account_id }}">
+    {{ button_text }}
+  </a>
+</div>
 {%- endmacro %}
 
 {% macro fxa_create_account_button(button_location, id_prefix, campaign, entrypoint, button_text = '', button_class = '', content = '', source = '') %}
@@ -346,7 +354,7 @@
 
     DO NOT include utm_source in this dictionary! (It's already a required param.)
 #}
-{% macro fxa_email_form(entrypoint, utm_source, class_name='fxa-email-form', utm_params={}, intro_text='', button_text='', button_class='button button-blue') -%}
+{% macro fxa_email_form(entrypoint, utm_source, class_name='fxa-email-form', utm_params={}, intro_text='', button_text='', button_class='js-fxa-cta-link button button-blue') -%}
   <form action="{{ settings.FXA_ENDPOINT }}" data-mozillaonline-action="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}" id="fxa-email-form" class="{{ class_name }}">
     <input type="hidden" name="action" value="email" />
     <input type="hidden" name="context" value="fx_desktop_v3" />
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -247,8 +247,8 @@
 
 {%- endmacro %}
 
-{% macro fxa_cta_link(entrypoint, campaign, content, source, button_class, button_location, button_text, account_id, action='signin', utm_params={}) %}
-  <a {{ fxa_link_fragment(entrypoint=entrypoint, action="{{ action }}", utm_params={'campaign': campaign, 'content': (content or request.path_info), 'source': source}) }} class="js-fxa-cta-link {{ button_class }}" data-button-name="Create account" data-link-type="FxA-Sync" data-cta-position="{{ button_location }}" id="{{ account_id }}">
+{% macro fxa_cta_link(entrypoint, campaign, content, source, button_class, button_location, button_text, account_id, action='signin', medium='referral') %}
+  <a {{ fxa_link_fragment(entrypoint=entrypoint, action="{{ action }}", utm_params={'campaign': campaign, 'content': (content or request.path_info), 'source': source, 'medium': medium }) }} class="js-fxa-cta-link {{ button_class }}" data-button-name="Create account" data-link-type="FxA-Sync" data-cta-position="{{ button_location }}" id="{{ account_id }}">
     {{ button_text }}
   </a>
 {%- endmacro %}
--- a/bedrock/firefox/templates/firefox/accounts-2018.html
+++ b/bedrock/firefox/templates/firefox/accounts-2018.html
@@ -3,7 +3,7 @@
   # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
 
 {% from "macros-protocol.html" import feature_card with context %}
-{% from "macros.html" import fxa_email_form, fxa_link_fragment, google_play_button with context %}
+{% from "macros.html" import fxa_email_form, fxa_link_fragment, fxa_cta_link, google_play_button with context %}
 
 {% extends "base-protocol.html" %}
 
--- a/bedrock/firefox/templates/firefox/accounts-2018.html
+++ b/bedrock/firefox/templates/firefox/accounts-2018.html
@@ -38,10 +38,7 @@
       {% endif %}
       <div id="hero-cta" class="mzp-c-hero-cta">
         <div class="show-fxa-supported-signed-out">
-          <a
-          {{ fxa_link_fragment(entrypoint='mozilla.org-accounts_page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-primary mzp-t-product" data-button-name="Create account" data-link-type="FxA-Sync" data-cta-position="primary cta" id="features-hero-account">
-            {{ _('Create a Firefox Account') }}
-          </a>
+          {{ fxa_cta_link(entrypoint='mozilla.org-accounts_page', account_id='features-hero-account', action='signup', button_text=_('Create a Firefox Account'), button_class='mzp-c-button mzp-t-primary mzp-t-product', campaign='fxa-benefits-page', content='accounts-page-top-cta', medium='referral', source='accounts-page') }}
         </div>
         <div class="show-fxa-supported-signed-in">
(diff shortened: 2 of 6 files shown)


amychurchwell:
cc: @alexgibson 
I wrote a comment in `utm-source-check.js`, having issues with function firing before DOM has loaded. Added a temporary class name to `fxa_link_fragment` which is currently disrupting class-based styles.

amychurchwell:
`entrypoint_variation` & `entrypoint_experiment` discussed [here](https://github.com/mozilla/fxa/issues/693), my feeling is that they should be added to `utm_params={}` when needed as they are experiment specific.

bedrock/base/templates/macros.html
```diff
@@ -319,9 +318,14 @@ <h3>{{ _('Your download link was sent.') }}</h3>
       {% do fxa_queries.append('utm_{}={}'.format(param, val.replace('/', '%2F'))) %}
     {% endif %}
   {% endfor %}
+  {% if not service_type %}
+    {% set service_type = 'sync' %}
+  {% endif %}
 
-  href="{{ settings.FXA_ENDPOINT}}{{ fxa_path }}?{{ fxa_queries|join('&') }}" data-mozillaonline-link="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}{{ fxa_path }}?{{ fxa_queries|join('&') }}"
+  href="{{ settings.FXA_ENDPOINT }}{{ fxa_path }}?service={{ service_type }}&{{ fxa_queries|join('&') }}" data-mozillaonline-link="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}{{ fxa_path }}?{{ fxa_queries|join('&') }}"
+  class="fxa_link_fragment"
```
alexgibson: The buttons that use the `fxa_link_fragment` macro already have a `class` attribute defined in their page CSS, so you can't overwrite that here without browsers ingoring it. This seems to be why the node list returned by your `querySelector` is empty.

What we may need here is to create a new `fxa_cta_link` macro, that consumes `fxa_link_fragment` and returns the full button HTML, instead of just the `href`? We already have `fxa_create_account_button`, but this seems to do way more things than we need. All we require is a single button 😕

If we create an `fxa_cta_link` macro, then you could pass a `button_class` parameter to it so that individual pages could still apply their own classes, whilst still retaining your JS hook, like so:

```
class="js-fxa-cta-link {{ button_class }}"
```

Pages could then call the `fxa_cta_link()` macro directly, and pass in the values they need?

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -38,7 +38,8 @@ <h1 class="mzp-c-hero-title">{{ _('One log-in. Power and privacy everywhere.') }
       {% endif %}
       <div id="hero-cta" class="mzp-c-hero-cta">
         <div class="show-fxa-supported-signed-out">
-          <a {{ fxa_link_fragment(entrypoint='accounts-page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-primary mzp-t-product" data-button-name="Create account" data-link-type="button" data-cta-position="primary cta" id="features-hero-account">
+          <a
+          {{ fxa_link_fragment(entrypoint='mozilla.org-accounts-page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-primary mzp-t-product" data-cta-position="primary cta" id="features-hero-account">
```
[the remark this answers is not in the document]
amychurchwell: This is being moved to PR #7113 to separate the parameter validation (which will be _this_ PR) from standardizing naming and values.

bedrock/base/templates/macros.html
amychurchwell: I think I'm confused how many places this new button would touch? What does `fxa_cta_link` replace? All fxa buttons on bedrock? is it replacing `fxa_link_fragment` in some situations?
`fxa_create_account_button` is complex, but I'm not sure why this wouldn't belong in it?

alexgibson: The fxa_cta_link macro would be used everywhere we show the standard button, instead of creating a custom link everywhere like we do now. Let me know if you want to talk through this in person if it doesn’t make sense.

amychurchwell: @alexgibson maybe we should set up a time to talk because I'm experiencing that any `{{ button_class }}` value passed will overwrite any preexisting classes. So in the case of: `class="js-fxa-cta-link {{ button_class }}"`, `js-fxa-cta-link` will be overwritten.

alexgibson: @amy perhaps this might be easier if you can share your code? If you can push what you have here, i’ll take a look

amychurchwell: @alexgibson I pushed up a rough draft of `fxa_cta_link` macro. I feel a little clumsy on everywhere it touches. Maybe we can organize a time to chat more?

bedrock/base/templates/macros.html
```diff
@@ -320,7 +319,16 @@ <h3>{{ _('Your download link was sent.') }}</h3>
     {% endif %}
   {% endfor %}
 
-  href="{{ settings.FXA_ENDPOINT}}{{ fxa_path }}?{{ fxa_queries|join('&') }}" data-mozillaonline-link="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}{{ fxa_path }}?{{ fxa_queries|join('&') }}"
+  href="{{ settings.FXA_ENDPOINT }}{{ fxa_path }}?service={{ service_type }}&{{ fxa_queries|join('&') }}" data-mozillaonline-link="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}{{ fxa_path }}?{{ fxa_queries|join('&') }}"
+
+{%- endmacro %}
+
+{% macro fxa_cta_link(button_location, id_prefix, campaign, entrypoint, button_text = '', button_class = '', content = '', source = '', action='signup', utm_params={}) %}
+<div class="show-fxa-supported-signed-out">
```
alexgibson: No need for this wrapper div I think? The macro can just render a button.

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -38,7 +38,7 @@ <h1 class="mzp-c-hero-title">{{ _('One log-in. Power and privacy everywhere.') }
       {% endif %}
       <div id="hero-cta" class="mzp-c-hero-cta">
         <div class="show-fxa-supported-signed-out">
-          <a {{ fxa_link_fragment(entrypoint='accounts-page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-primary mzp-t-product" data-button-name="Create account" data-link-type="button" data-cta-position="primary cta" id="features-hero-account">
+          <a {{ fxa_cta_link(entrypoint='accounts-page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-primary mzp-t-product" data-button-name="Create account" data-link-type="button" data-cta-position="primary cta" id="features-hero-account">
```
alexgibson: This is trying to render a link inside a link, which is invalid markup (and could be the cause of your issues)

alexgibson: Instead you just need to do 

```
{{ fxa_cta_link(entrypoint='accounts-page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }}
```

Which will then spit out a link:

```
<a href="foo">some cta</a>
```

media/js/base/fxa-utm-referral.js
```diff
@@ -0,0 +1,57 @@
...
+    };
+
+    UtmUrl.getAttributionData.init = function (){
+        UtmUrl.getAttributionData();
+
+        var ctaLinks = document.getElementsByClassName('js-fxa-cta-link');
+        var url;
+
+        for (var i = 0; i < ctaLinks.length; i++) {
+            console.log(ctaLinks[i].href);
```
amychurchwell: @alexgibson having trouble appending the validated utm_source to each of the `.js-fxa-cta-link` urls. 😓

media/js/base/fxa-utm-referral.js
```diff
@@ -0,0 +1,57 @@
...
+
+    UtmUrl.getAttributionData = function () {
+        var params = new window._SearchParams().utmParams();
+        var finalParams = {};
+
+        // utm_source
+        if (params.hasOwnProperty('utm_source') && (/^[\w/.%-]+$/).test(params['utm_source'])) {
+            // this value is already URI encoded, so we need to decode here as
+            // it will get re-encoded in _SearchParams.objectToQueryString
+            finalParams['source'] = decodeURIComponent(params['utm_source']);
```
alexgibson:
```suggestion
            finalParams['utm_source'] = decodeURIComponent(params['utm_source']);
```

You want to replace the `utm_source` parameter value, not change it to `source`.

media/js/base/fxa-utm-referral.js
```diff
@@ -0,0 +1,57 @@
...
+            console.log(finalParams);
+        }
+        return Object.getOwnPropertyNames(finalParams);
```
alexgibson:
```suggestion
        return Object.getOwnPropertyNames(finalParams).length === 1 ? finalParams : null;
```

Note: you'll need to change the `length` check to equal the total number of utm params you're validating.

media/js/base/fxa-utm-referral.js
```diff
@@ -0,0 +1,57 @@
...
+        if (url.indexOf('?') > 0) {
+            linkParams = window._SearchParams.queryStringToObject(url.split('?')[1]);
+            finalParams = Object.assign(finalParams, linkParams);
+        }
+
+        return url.split('?')[0] + '?' + window._SearchParams.objectToQueryString(finalParams);
+    };
+
+    UtmUrl.getAttributionData.init = function (){
+        UtmUrl.getAttributionData();
```
alexgibson:
```suggestion
        var params = UtmUrl.getAttributionData();
```

This function returns a value, so you'll need to store it in a variable.

media/js/base/fxa-utm-referral.js
alexgibson:
```suggestion
            if (params && link) {
                var url = UtmUrl.appendToDownloadURL(link, params);
                ctaLinks[i].href = url;
            }
```

If utm params exist in the URL, and a link exists that needs updating, update the link.

Note: you may want to update `data-mozillaonline-link` here too.

media/js/base/fxa-utm-referral.js
```diff
@@ -0,0 +1,57 @@
...
+        return Object.getOwnPropertyNames(finalParams);
+    };
+
+    UtmUrl.appendToDownloadURL = function (url, data) {
+        var finalParams = data;
+        var linkParams;
+
+        if (url.indexOf('?') > 0) {
+            linkParams = window._SearchParams.queryStringToObject(url.split('?')[1]);
+            finalParams = Object.assign(finalParams, linkParams);
```
alexgibson: We need to be a little careful when adding code to the base bundle in every page. `Object.assign` doesn't have great support in [older browsers](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Browser_compatibility), so you may need to find an alternative to this function.

You may also want to check `Object.getOwnPropertyNames`.

alexgibson:
@amychurchwell I tried to open a pulle request to your branch, but it looks like you need to rebase against master. I left a commit here that seems to get things working: https://github.com/alexgibson/bedrock/commit/80c48ffd5e68fbbba702a29a8f92b565085ac033

Feel free to pick at what you need 👍

alexgibson:
@amychurchwell looks like you still need to update the buttons to use `fxa_cta_link`. Maybe that was lost in the rebase?

amychurchwell:
Screen Shot 2019-05-03 at 12 40 23 PM

@alexgibson `fxa_cta_link` replaces the fxa button but not the `fxa_link_fragment` 'log in' link. 🤔

bedrock/base/templates/macros.html
```diff
@@ -324,7 +324,7 @@ <h3>{{ _('Your download link was sent.') }}</h3>
 {%- endmacro %}
 
 {% macro fxa_cta_link(entrypoint, campaign, content, source, button_class, button_location, button_text, account_id, action='signin', utm_params={}) %}
-  <a {{ fxa_link_fragment(entrypoint=entrypoint, action="{{ action }}", utm_params={'campaign': campaign, 'content': (content or request.path_info), 'source': source}) }} class="js-fxa-cta-link {{ button_class }}" data-button-name="Create account" data-link-type="FxA-Sync" data-cta-position="{{ button_location }}" id="{{ account_id }}">
```
alexgibson: I'm confused, why is this being removed? Didn't the code rely on it?

bedrock/base/templates/includes/protocol/navigation/index.html
```diff
@@ -25,9 +25,7 @@
         <div class="mzp-c-navigation-download">
           {{ download_firefox(alt_copy=_('Download Firefox'), dom_id='protocol-nav-download-firefox', button_color='mzp-t-secondary mzp-t-small', download_location='nav') }}
           <div class="c-navigation-fxa-cta-container">
-            <a class="c-navigation-fxa-cta mzp-c-button mzp-t-secondary mzp-t-small mzp-t-product" {{ fxa_link_fragment(entrypoint='mozilla.org-globalnav', action='signup', utm_params={'campaign': 'globalnav', 'content': 'get-firefox-account', 'medium': 'referral', 'source': 'www.mozilla.org'}) }} data-button-name="Get a Firefox Account" data-link-type="FxA-Sync" data-cta-position="secondary cta" data-alt-href="{{ url('firefox.accounts') }}">
-              {{ _('Get a Firefox Account') }}
-            </a>
+            {{ fxa_cta_link(entrypoint='mozilla.org-globalnav', action='signup', button_text=_('Get a Firefox Account'), button_class='c-navigation-fxa-cta mzp-c-button mzp-t-secondary mzp-t-small mzp-t-product', button_location='secondary cta', content='get-firefox-account', campaign='globalnav', source='www.mozilla.org') }}<br>
```
alexgibson: data-alt-href="{{ url('firefox.accounts') }}" is missing here, which is breaking this logic https://github.com/mozilla/bedrock/blob/master/media/js/base/protocol/init-navigation.js#L59-L70

It looks like we're also missing an `account_id` here

amychurchwell: Only the account benefits page seemed to have an id defined, which was styles related. Is the way I wrote the macro mean that `account_id` is not optional?

alexgibson:
Seeing as we just merged the eslint PR, it may be worth rebasing this branch just to make sure it passed the new linting rules?

bedrock/base/templates/includes/protocol/navigation/index.html
alexgibson: Looking in the generated HTML, the button currently outputs an empty `id=""` if an Id is not passed. Maybe just add it as an optional parameter?

amychurchwell: Now that I'm thinking of it, maybe this particular one doesn't need to be a `fxa_cta_link` macro. It doesn't make sense to at data-alt-href to the macro because this feels like the only place that uses it that I've come across. It may make more sense to just add the class of `js-fxa-cta-link`?

media/js/base/fxa-utm-referral.js
```diff
@@ -0,0 +1,100 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function () {
+
+    var UtmUrl = {};
+
```
alexgibson: Looks like ESLint is complaining on a missing `'use strict';`here:

```
error  Use the function form of 'use strict'  strict
```

media/js/base/fxa-utm-referral.js
```diff
@@ -0,0 +1,100 @@
...
+        }
+
+        // feature detect support for object modification.
+        if (typeof Object.assign !== 'function' || typeof Object.getOwnPropertyNames !== 'function') {
+            return;
+        }
+
+        for (var i = 0; i < ctaLinks.length; i++) {
+            // get the current FxA links.
+            var oldAccountsLink = ctaLinks[i].href;
```
alexgibson: I think there's a possible race condition here between this code and [this code](https://github.com/mozilla/bedrock/blob/master/media/js/base/protocol/init-navigation.js#L59-L70). If true, we could end up adding UTM params to an internal link to `/firefox/accounts/`, which would confound our GA data by creating new sessions for those that click the link.

To try and avoid this and make our logic here a bit more robust, it may be worth adding a check to make sure that both `oldAccountsLinks` and `newAccountsLink` both start with `https://accounts.firefox.com/` before updating them. This way, we can be more confident that we're only updating an `href` that points to an external domain. Thoughts?

amychurchwell: I think you are right, thanks for catching.
Would this be worth separating out into a separate issue? (FxA validation for global nav?)

alexgibson: hmm, I'm not certain if we can safely merge this without it causing issues?

alexgibson: Or do you mean not applying this logic to the global nav?

amychurchwell: I mean not applying the logic/`fxa_cta_link` to the nav in this PR and moving it to it's own issue. Maybe that is too granular? Wdyt?

alexgibson: I guess it depends how much of the original issue we want to offload for a later task. This is probably a couple of extra lines of JS to add, but we could always skip the navigation button for now if we think that makes sense? My only concern about shipping a partial implementation is that we'll not get time to go back and fix things for a while, or forget to do it.

Issue #7061, opened 2019-04-10
Standardize parameters and values in FxA referrals


ejregithub:
## Description

When we send traffic to FxA signups, we need to send uniform parameters so measurement is straightforward. 

1. For the `entrypoint` url parameter that we generate and send to FxA servers, please use the following standard format: `domain-point_of_CTA`. Examples from our existing site: 

```
entrypoint=mozilla.org-globalnav, 
entrypoint=mozilla.org-fxa_benefits, 
entrypoint=mozilla.org-firefox_new_modal,
```

We may add other entrypoints later.

2. For datalayer values in FxA links, please use the following standard format ([ref](https://bedrock.readthedocs.io/en/latest/analytics.html?highlight=data-link-type)): `FxA-ServiceName`. Examples from our existing site:

```
data-link-type="FxA-Sync"
```

We may add other data link types later.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Parameters are standardized and code is live

#### Tasks
- [ ] Insert the parameters

alexgibson:
@ejregithub I'm confused what this issue is for. Can you add some more information? Thanks

ejregithub:
This the task to complete the work by adding the new parameters. The parameters are named so we can land this code: mozilla/fxa#693 (comment)

hoosteeno:
I added two comments to the closed card mentioned above. Can we discuss here?

1: @amychurchwell , do you think it would be possible to add to this code a uniform event category for FxA links? Right now, our event instrumentation of referrals depends on labels (e.g. "Get a Firefox Account"), which makes them brittle. If we used a category (e.g. "FxA Referral" or similar) we would not be dependent on a CTA string.

2: The parameters are named, so there is no further reason to delay landing this code: mozilla/fxa#693 (comment)

Note that these parameters imply usage. We can use them when we have variations on a CTA on the same entrypoint, as part of an experiment.

In cases where we don't, we can leave them blank.

All of this suggests that what we send in entrypoint should be uniform and verbose. I'd like to suggest the following uniform examples for the entrypoint parameter, and encourage your feedback:

```
entrypoint=site-page-element
entrypoint=mozilla.org-globalnav
entrypoint=mozilla.org-fxa.features-send.tabs
entrypoint=mozilla.org-wnp66-topbar
entrypoint=mozilla.org-firstrun-footer
```

Then when we have experimental variations, we can use the two new parameters

hoosteeno:
[...]

PR #7113, opened 2019-04-25
Standardizing fxa utm parameters and values (#7061)

amychurchwell:
## Description
Standardizing FxA parameters so they are uniform across the site.

## Issue / Bugzilla link
#7061

## Testing
- Format for `entrypoint`: `domain-point_of_CTA`

### DIFF
--- a/bedrock/firefox/templates/firefox/accounts-2018.html
+++ b/bedrock/firefox/templates/firefox/accounts-2018.html
@@ -39,7 +39,7 @@
       <div id="hero-cta" class="mzp-c-hero-cta">
         <div class="show-fxa-supported-signed-out">
           <a
-          {{ fxa_link_fragment(entrypoint='mozilla.org-accounts-page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-primary mzp-t-product" data-cta-position="primary cta" id="features-hero-account">
+          {{ fxa_link_fragment(entrypoint='mozilla.org-accounts_page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-primary mzp-t-product" data-cta-position="primary cta" id="features-hero-account">
             {{ _('Create a Firefox Account') }}
           </a>
         </div>
@@ -126,7 +126,7 @@
       {{ download_firefox(download_location='primary cta', button_color='mzp-c-button mzp-t-secondary mzp-t-product') }}
     </div>
     <div class="show-fxa-supported-signed-out">
-      <a {{ fxa_link_fragment(entrypoint='accounts-page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-secondary mzp-t-product" data-button-name="Create account" data-link-type="button">
+      <a {{ fxa_link_fragment(entrypoint='mozilla.org-accounts_page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-secondary mzp-t-product" data-button-name="Create account" data-link-type="button">
           {{ _('Create a Firefox Account') }}
       </a>
     </div>
@@ -147,7 +147,7 @@
       {{ download_firefox(dom_id='sync-download', download_location='primary cta', button_color='mzp-c-button mzp-t-secondary mzp-t-product') }}
     </div>
     <div class="show-fxa-supported-signed-out">
-      <a {{ fxa_link_fragment(entrypoint='accounts-page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-secondary mzp-t-product" data-button-name="Create account" data-link-type="button">
+      <a {{ fxa_link_fragment(entrypoint='mozilla.org-accounts_page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-secondary mzp-t-product" data-button-name="Create account" data-link-type="button">
           {{ _('Create a Firefox Account') }}
       </a>
     </div>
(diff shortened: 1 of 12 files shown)


bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -289,7 +290,13 @@ <h2>{{ _('You deserve peace of mind everywhere.') }}</h2>
         </div>
 
         <div class="show-fxa-supported-signed-out">
-          {{ fxa_email_form(entrypoint='accounts-page', utm_source='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'accounts-page', 'medium': 'referral'}, button_class='mzp-c-button mzp-t-primary mzp-t-product', button_text=_('Create an Account')) }}
+          {{ fxa_email_form(
+            entrypoint='mozilla.org-fxa_benefits',
+            utm_source='mozilla.org-fxa_benefits',
+            utm_params={'referrer': 'https://www.mozilla.org/en-US/firefox/accounts/'},
```
alexgibson: Is there a need to add `referrer` here? From our convo yesterday, it sounded like this is only necessary if `entrypoint` is missing. If we do require adding this, we should make sure it's adding the correct locale in the URL.

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -38,7 +38,8 @@ <h1 class="mzp-c-hero-title">{{ _('One log-in. Power and privacy everywhere.') }
       {% endif %}
       <div id="hero-cta" class="mzp-c-hero-cta">
         <div class="show-fxa-supported-signed-out">
-          <a {{ fxa_link_fragment(entrypoint='accounts-page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-primary mzp-t-product" data-button-name="Create account" data-link-type="button" data-cta-position="primary cta" id="features-hero-account">
+          <a
+          {{ fxa_link_fragment(entrypoint='mozilla.org-accounts-page', action='signup', utm_params={'campaign': 'fxa-benefits-page', 'content': 'accounts-page-top-cta', 'medium': 'referral', 'source': 'accounts-page'}) }} class="mzp-c-button mzp-t-primary mzp-t-product" data-cta-position="primary cta" id="features-hero-account">
```
alexgibson: Reading the updated issue [description](https://github.com/mozilla/bedrock/issues/7061#issue-431606721), it sounds like this should be `mozilla.org-accounts_page`? (using underscores to separate the point of CTA)

bedrock/firefox/templates/firefox/accounts-2018.html
```diff
@@ -289,7 +290,13 @@ <h2>{{ _('You deserve peace of mind everywhere.') }}</h2>
         </div>
 
         <div class="show-fxa-supported-signed-out">
-          {{ fxa_email_form(entrypoint='accounts-page', utm_source='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'accounts-page', 'medium': 'referral'}, button_class='mzp-c-button mzp-t-primary mzp-t-product', button_text=_('Create an Account')) }}
+          {{ fxa_email_form(
+            entrypoint='mozilla.org-fxa_benefits',
```
alexgibson: Is there a reason that this `entrypoint` is different to the one above, if it's coming from the same page?

bedrock/firefox/templates/firefox/accounts.html
```diff
@@ -38,7 +38,7 @@ <h1>{{_('Sign up, sign in, stay in sync.')}}</h1>
       <div class="fxa-cta show-fxa-supported-signed-in show-fxa-supported-signed-out">
         <p class="show-fxa-supported-signed-in signed-in">{{_('You’re signed in and ready to start using Sync.')}}</p>
         <div class="show-fxa-supported-signed-out">
-          {{ fxa_email_form(entrypoint='accounts-page', utm_source='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'accounts-page', 'medium': 'referral'}) }}
+          {{ fxa_email_form(entrypoint='mozilla.org-accounts', utm_source='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'accounts-page', 'medium': 'referral'}) }}
```
alexgibson: Should this be `mozilla.org-accounts_page` to match the other template, or is it better being something unique?

bedrock/firefox/templates/firefox/new/scene1.html
```diff
@@ -165,7 +165,7 @@ <h4 class="heading-linux">{{_('Download Firefox <span>for Linux</span>') }}</h4>
   <h4 class="fxa-title">{{ _('You’ve already got the Firefox browser. Now get everything else Firefox.') }}</h4>
   <p class="fxa-intro">{{ _('Sync bookmarks, send tabs, save to Pocket and more with a Firefox Account.') }} <a href="{{ url('firefox.accounts') }}">{{ _('Learn more') }}</a></p>
 
-  {{ fxa_email_form(entrypoint='firefox-new', button_text=_('Get a Firefox Account'), utm_source='firefox-new', utm_params={'campaign': 'firefox-new', 'content': 'firefox-new-fxa', 'term': 'existing-users', 'medium': 'referral'}) }}
+  {{ fxa_email_form(entrypoint='mozilla.org-firefox_new', button_text=_('Get a Firefox Account'), utm_source='firefox-new', utm_params={'campaign': 'firefox-new', 'content': 'firefox-new-fxa', 'term': 'existing-users', 'medium': 'referral'}) }}
 
   <p class="fxa-signin">{{ _('Already have an account?') }} <a {{ fxa_link_fragment(entrypoint='accounts-page', utm_params={'campaign': 'fxa-embedded-form', 'content': '/firefox/new/', 'source': 'download-page'}) }}>{{ _('Sign In') }}</a></p>
```
alexgibson: There's an `entrypoint` here with an incorrect value of `accounts-page`, which seems like it needs updating?

alexgibson:
Looks like there's a couple of entry points that still need updating:

https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/whatsnew/fx64/whatsnew-fx64-base.html#L72

https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/firstrun/firstrun-quantum.html#L66-L68 (you may need to rebase on master to get this one that was just added).

bedrock/firefox/templates/firefox/accounts.html
amychurchwell: I will update it to match the other template, I'm not sure if they need to be unique.

bedrock/firefox/templates/firefox/accounts-2018.html
amychurchwell: really good point/catch. Do we have logic to add `referrer` if `entrypoint` is missing? If not I think that would make a good issue.

alexgibson: I don't know of anywhere where we don't pass `entrypoint` already, so there's no logic for referrer afaik.

alexgibson:
@amychurchwell changes look good, but I think there are a few data-attributes that still need updating (the button in the global nav at least).

alexgibson:
Merging as per: https://github.com/mozilla/bedrock/pull/7113#issuecomment-488332980

amychurchwell:
Updating title of this PR to be more specific about what parameters were touched. UTM work continues in #7397

PR #7124, opened 2019-04-30
Experiment: page hero rendering performance (Fixes #7118)

### LINKED ISSUE #7118: Experiment: page hero rendering performance
## Description

[Mana page for experiment](https://mana.mozilla.org/wiki/display/EN/Page+hero+rendering+performance)

In issue https://github.com/mozilla/bedrock/issues/6741 we conducted an analysis of existing page load times in GA and their effect on conversions. Looking at the GA data, it was determined that while there is a direct correlation between page load times and bounce rates in general, there wasn't as strong a relationship between page load and conversions, such as downloads. The _assumption_ here is that if people search for "Download Firefox", then they are more likely to endure a slow loading web page, as they are already comitted to wanting to download the product.

However, there are a couple of caveats with this type of analysis:

1. For people on slow data connections, GA might never load before they give up waiting. We likely don't have GA data on these visitors.
2. Using page load time has long been a common metric to measure website performance, but in some respects it is flawed. For example, a slow loading image further down a page might hold up a page load event by ~10 seconds, despite everything else already having finished rendering at the top of the page. In this scenario, a slow load time is much less likely to be a negative influence on conversions.

With the two above scenarios in mind, I plan to run an alternative type of performance experiment - concentrating on more modern metrics: [speed index](https://sites.google.com/a/webpagetest.org/docs/using-webpagetest/metrics/speed-index), and _last hero painted_. TL:DR - how long the main hero section of a page **actually takes to render**.

Hypothesis:

Variations of a page with a higher speed index score (lower is better), and a longer time to last hero painted, will have a higher bounce rate, and also a lower conversion rate on in-page CTAs.

Null hypothesis:

No matter how long a page variation takes to render, there will be no impact on either bounce rate or conversion rate.

Experiment criteria:

- Target audience : The [about page](https://www.mozilla.org/en-US/about/) (all locales). This is not a main product download page, so there is less risk in adversely effecting downloads. People viewing this page are also less likely commited to an immediate task.
- Experiment configuration: A multi variant test, which incrementally delays the time between start render and last hero painted (resulting in a higher speed index score per variant).

---

- [ ] Experiment is runs for one week and provides enough data to satisfy the hypothesis.

- [ ] Variations that are slowest to load may result in a decrease in conversions.

#### Tasks
- [x] Mana page created.
- [x] Code experiment.
- [ ] Launch experiment.
- [ ] Stop experiment.
- [ ] Provide analysis.

alexgibson:
## Description
- Adds an experiment on /about/ that aims to test the effects of hero rendering performance against bounce rate and conversions.

## Issue / Bugzilla link
#7118

## Testing
Demo: https://bedrock-demo-agibson.oregon-b.moz.works/about/

- [x] Experiment should trigger correctly.
- [x] Variations should increment rendering delay by ~1 second (a begin control, j being the slowest)

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about-2019.html
+++ b/bedrock/mozorg/templates/mozorg/about-2019.html
@@ -16,12 +16,18 @@
 
 {% block body_id %}about{% endblock %}
 
+{% block experiments %}
+  {% if switch('experiment_about_page_performance') %}
+    {{ js_bundle('experiment-about-page-performance') }}
+  {% endif %}
+{% endblock %}
+
 {% block page_css %}
   {{ css_bundle('about-en') }}
 {% endblock %}
 
 {% block content %}
-<section class="mzp-c-hero t-mission">
+<section class="mzp-c-hero t-mission" {% if variant %}data-variant="{{ variant }}"{% endif%}>
   <div class="mzp-l-content">
     <div class="mzp-c-hero-body">
       <h1 class="mzp-c-hero-title">{{ _('Mozilla makes browsers, apps, code and tools that put people before profit.') }}</h1>
@@ -31,7 +37,7 @@
       </div>
 
       <p class="mzp-c-hero-cta">
-        <a id="read-mission-button" class="mzp-c-button" href="{{ url('mozorg.mission') }}">{{ _('Read Our Mission') }}</a>
+        <a id="read-mission-button" class="mzp-c-button" href="{{ url('mozorg.mission') }}" data-button-name="Read Our Mission" data-link-type="button" data-cta-position="primary cta">{{ _('Read Our Mission') }}</a>
       </p>
     </div>
   </div>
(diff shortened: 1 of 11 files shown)


alexgibson:
This is now ready for code review.

alexgibson:
@stephaniehobson this is low priority and not an urgent thing, but you've provided some great thoughts / insight into how to conduct this experiment. A review from yourself would be really welcome? 🙂

alexgibson:
@stephaniehobson thanks, I think reducing the text delay by a factor makes sense. I'll do some measurements and update with what I think simulates a slower connection a bit better 🙂

alexgibson:
@stephaniehobson I updated the variations, thanks for your suggestions. I opted for:

- Making default text visible earlier (at the same time as backgrounds), for a more meaningful first paint.
- Delay loading Zilla Slab incrementally, so it swaps in around the same time as images.

I ran some preliminary benchmarks using WebPageTest, and there still seems to be a good correlation between variations and both Speed Index and Hero Paint times:

image

Note: variation J appears broken in the eyes of WPT, since it takes so long to load it appears to assume everythinghas finished before it actually has. This is probably a good place to stop 😅

stephaniehobson:
R+ 🐢

PR #7137, opened 2019-05-01
Add subheading to elections page

### LINKED ISSUE #7136: Add subheading to EU elections pages
## Description

Update elections page with subheading.

EU_Elections_Hero

stephaniehobson:
## Description

Add subheading to elections page

## Issue / Bugzilla link

Fix #7136.


media/css/firefox/election.scss
```diff
@@ -18,6 +18,13 @@ $image-path: '/media/protocol/img';
     padding-bottom: 200px;
 }
 
+.mzp-c-hero-title {
+    @include at2x('/media/img/firefox/election/firefox-logo.png', 100px);
```
alexgibson: Suggest adding this to `/img/logos/firefox/` as a central place, so we don't end up with multiple copies of the same wordmark logo?

media/css/firefox/election.scss
```diff
@@ -18,6 +18,13 @@ $image-path: '/media/protocol/img';
...
+    background-repeat: no-repeat;
+    background-position: top left;
```
alexgibson: Should the default position for this be `top center`, to match the behavior of the default hero component?

image

stephaniehobson:
Issue for the "logo lockup" https://github.com/mozilla/protocol/issues/373

media/css/firefox/election.scss
stephaniehobson: The same image only twice the size is used on the WN67 page. The size on 67 seems more likely to be the one used centrally so I moved that one. Which means I removed the `a2x` mixin here and just use the low-res version at half its size.

media/css/firefox/election.scss
stephaniehobson: ✅

PR #7148, opened 2019-05-02
Remove lingering nose functions from tests (post django 1.11 update)

jpetto:
## Description

A few nose functions remain after the Django 1.11 update.

### DIFF
--- a/bedrock/firefox/tests/test_base.py
+++ b/bedrock/firefox/tests/test_base.py
@@ -382,10 +382,9 @@ class TestWhatsNew(TestCase):
         req.locale = 'en-US'
         self.view(req, version='67.0')
         template = render_mock.call_args[0][1]
-        eq_(template, ['firefox/whatsnew/whatsnew-fx67.html'])
+        assert template == ['firefox/whatsnew/whatsnew-fx67.html']
         context = render_mock.call_args[0][2]
-        ok_('show_newsletter' in context)
-        eq_(True, context['show_newsletter'])
+        assert context['show_newsletter']
 
     def test_fx_67_0_no_newsletter(self, render_mock):
         """Should not include newsletter for non-translated locales"""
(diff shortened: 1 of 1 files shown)


bedrock/firefox/tests/test_base.py
```diff
@@ -382,21 +382,21 @@ def test_fx_67_0(self, render_mock):
         req.locale = 'en-US'
         self.view(req, version='67.0')
         template = render_mock.call_args[0][1]
-        eq_(template, ['firefox/whatsnew/whatsnew-fx67.html'])
+        assert template == ['firefox/whatsnew/whatsnew-fx67.html']
         context = render_mock.call_args[0][2]
-        ok_('show_newsletter' in context)
-        eq_(True, context['show_newsletter'])
+        assert 'show_newsletter' in context
+        assert True == context['show_newsletter']
```
jgmize: NIT: `True ==` is redundant here
```python
assert context['show_newsletter']
```

jgmize: actually, the previous assertion is redundant too

bedrock/firefox/tests/test_base.py
```diff
@@ -382,21 +382,21 @@ def test_fx_67_0(self, render_mock):
...
         req.locale = 'el'
         self.view(req, version='67.0')
         template = render_mock.call_args[0][1]
-        eq_(template, ['firefox/whatsnew/whatsnew-fx67.html'])
+        assert template == ['firefox/whatsnew/whatsnew-fx67.html']
         context = render_mock.call_args[0][2]
-        ok_('show_newsletter' in context)
-        eq_(False, context['show_newsletter'])
+        assert 'show_newsletter' in context
+        assert False == context['show_newsletter']
```
pmac:
```suggestion
        assert not context['show_newsletter']
```

jgmize:
flake8 cares more about the style nits than I do:
```
bedrock/firefox/tests/test_base.py:388:21: E712 comparison to True should be 'if cond is True:' or 'if cond:'
bedrock/firefox/tests/test_base.py:399:22: E712 comparison to False should be 'if cond is False:' or 'if not cond:'
2       E712 comparison to True should be 'if cond is True:' or 'if cond:'
```

jpetto:
Cleaned it up!

PR #7149, opened 2019-05-02
Import newrelic.agent first in wsgi app

jgmize:
## Description

We are getting the following message in NR: 
> Some modules were uninstrumented during the current time window: meinheld.server, gunicorn.app.base. Make sure newrelic.agent.initialize() is called before any of these modules are imported. Visit the documentation for more information.

This PR re-orders the imports in the wsgi app as per the message above and [the docs](https://docs.newrelic.com/docs/agents/python-agent/installation/python-agent-advanced-integration#manual-integration).

### DIFF
--- a/wsgi/app.py
+++ b/wsgi/app.py
@@ -1,9 +1,6 @@
 # flake8: noqa
-import os
-
-from bedrock.base.config_manager import config
-
-
+# newrelic import & initialization must come first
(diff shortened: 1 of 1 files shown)


glogiotatidis:
https://github.com/mozmeao/infra/issues/1106

PR #7153, opened 2019-05-03
[fix #7144] Remove April 2019 fundraising banner

### LINKED ISSUE #7144: Shut off Fundraising banner to follow Internet Health Report Release on 5/3
## Description

Shut off the Fundraising banner that followed Internet Health Report Release in #7047

---

- [x] The banner is shut off, no longer shown to users.

#### Tasks
- [x] Flip the switch

craigcook:
## Description
The switch was flipped to hide the banner, this just cleans it up in bedrock.


craigcook:
> Needs a rebase, but otherwise this looks good to me.

Rebased!
 
> A side thought for the future: this banner seems to be a semi-annual thing that we re-code and then take down again. Could it be worth making into a more pernament feature of the home page, that we can quickly and easily toggle on / off when it's needed?

I have had similar thoughts... I've resisted making it permanent because I kept believing we would soon rebuild the home page for all locales. Also, it was only ever a year-end thing but now we've been asked to bring it back for short periods at other times so it's becoming even more repetitive to keep adding and removing. 

I wouldn't be opposed to making it permanent behind a switch (though the switch would mean it would always appear in dev mode so that could be annoying). The copy and background also keep changing regularly, but we could have a base version that is evergreen and even get it localized.

Issue #7156, opened 2019-05-06
Incorrect link on /releasenotes


amychurchwell:
## Description

On /firefox/66.0.4/releasenotes/ (any releasenotes #) the 'Desktop' link goes to the Android release notes. Vice versa on Android release notes.

We need to write some logic so that when you are looking at one platform the other is a link & the current one isn't. Also, perhaps the current view is bolded. 

### Bugzilla
[1549219](https://bugzilla.mozilla.org/show_bug.cgi?id=1549219)

PR #7160, opened 2019-05-07
Bug 1549219: don't link to android relnotes under "Desktop" heading a…

jcristau:
…nd vice-versa

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1549219
https://github.com/mozilla/bedrock/issues/7156

### DIFF
--- a/bedrock/firefox/templates/firefox/releases/notes.html
+++ b/bedrock/firefox/templates/firefox/releases/notes.html
@@ -90,13 +90,13 @@
               <div class="mzp-c-navigation-menu">
                 <nav class="mzp-c-menu mzp-is-basic">
                   <ul class="mzp-c-menu-category-list">
-                    {% if release.channel == 'Release' and equivalent_release_url %}
+                    {% if release.channel == 'Release' and release.product == 'Firefox for Android' and equivalent_release_url %}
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/releases/notes.html
```diff
@@ -90,13 +90,13 @@
               <div class="mzp-c-navigation-menu">
                 <nav class="mzp-c-menu mzp-is-basic">
                   <ul class="mzp-c-menu-category-list">
-                    {% if release.channel == 'Release' and equivalent_release_url %}
+                    {% if release.channel == 'Release' and release.product == 'Firefox for Android' and equivalent_release_url %}
                       <li class="mzp-c-menu-category"><a class="mzp-c-menu-title" href="{{ equivalent_release_url }}">{{ _('Desktop') }}</a></li>
                     {% else %}
                       <li class="mzp-c-menu-category"><a class="mzp-c-menu-title" href="{{ url('firefox.notes') }}">{{ _('Desktop') }}</a></li>
```
pmac: We don't really need a link at all if we're already looking at desktop notes right? But I guess this would take you to the latest notes if you were looking at an older one. Hmm... This does seem like an improvement regardless.

PR #7174, opened 2019-05-09
Indicate next steps and give feedback when cancelling

stephaniehobson:
- Script now suggests a letter to press to back out
- Script only requires a single character of input before progressing
- Script outputs "Cancelled." if the tag is cancelled

### DIFF
--- a/bin/tag-release.sh
+++ b/bin/tag-release.sh
@@ -19,10 +19,12 @@ while [[ $# -ge 1 ]]; do
   shift # past argument or value
 done
 
-echo "Did you test on stage first? If not, are you sure you want to skip going to stage?"
-read stage
+# prompt for confirmation, evaluate after first letter typed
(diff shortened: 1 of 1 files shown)


stephaniehobson:
Yeah, I want to make that improvement too :) I've already got the command to find the commit hash for stage head:`git ls-remote git://github.com/mozilla/bedrock.git refs/heads/stage | cut -f 1`

I don't know how to check if all the tests have passed though so I still want to prompt with the "are you sure?"

stephaniehobson:
Also, I think it would be worth double checking that local master HEAD matches the remote master HEAD.

stephaniehobson:
Or rather, we can check tests have passed but I don't know how to check if stage deployment succeeded.

pmac:
yeah. that would be a lot more work. Maybe it's actually already at the point of diminishing returns and we should just keep the prompt for every prod push to make sure people are thinking about it?

Issue #6480, opened 2018-11-15
Automate image compression as part of asset pipeline.


stephaniehobson:
Image compression is a time consuming but easily automatable step in the development process.

Tinypng has a developer API and can do both PNG and JPEGs.
SVGO is available as a npm package.

amychurchwell:
[@alexgibson @craigcook @stephaniehobson]
 
The Tinypng developer API has a price associated per image: https://tinypng.com/developers but there are potential other options if that is a blocker.

[SVGO ](https://github.com/svg/svgo) has quite a few formats, including [gulp](https://github.com/sindresorhus/grunt-svgmin). We could potentially write a gulp task for this & watch img directory for changes. Thoughts?

alexgibson:
I think SVGO is a great tool to handle svg files, and a gulp task makes the most sense 👍 

Tinypng is probably my favorite lossy image compression tool (in terms of results). For non-lossy compression there's also ImageOptim (which also has a [gulp plugin](https://github.com/allbitsnbytes/gulp-imageoptim)).

If we're to invest in automation this year, I think it's likely there will be some costs associated with that investment, so perhaps this is something to talk to Justin about? Personally I don't find ImageOptim on its own to be enough, so it would be nice to have something like TinyPng if we're to really streamline our workflow.

alexgibson:
There's a new `Imagemin` lib out that may be worth investigating here? https://web.dev/fast/use-imagemin-to-compress-images

It looks like it handles, jpg/png compression, svg optimization and also webp. There's a gulp plugin available too.

amychurchwell:
@alexgibson It is looking like imagemin has it's own plugins for png/svg/etc. For both `imagemin-pngquant` and `imagemin-mozjpeg` have a quality field to determine compression level.
(EX: `[pngquant({quality: [0.5, 0.5]})]`, this example forces compression to be at 50%)
Do you know if there a way to best determine what this range should be?

alexgibson:
Done in https://github.com/mozilla/bedrock/pull/7243

PR #7243, opened 2019-05-24
Add script to optimize images #6480

stephaniehobson:
## Description

This bash script works on images in git's staging area. It will:
- optimize PNG, JPG, and SVG files
- check the SVGs have a viewbox (needed for IE support)
- check that any high res images have low res equivalents (this should turn up naming miss matches between high and low res files).

## Issue / Bugzilla link

See #6480

## Testing
Add some images to the git staging area.
Run the script like `./bin/img.sh`
- see if it prompts you for a TinyPNG API key
- see if it saves properly
- create test .png and .svg and .jpg files and see if it compresses them
- create a file with the -high-res sufix and see if it complains there's no low res version
- create a low res version and see if it stops complaining
- does it work as expected? How is the UX? Do you understand the report it generates?
- are my assumptions good?

### DIFF
--- a/bin/img.sh
+++ b/bin/img.sh
@@ -1,0 +1,69 @@
+#!/bin/bash
+
+# are there images to optimize?
+fileschanged=$(git diff --diff-filter=ACM --name-only HEAD | grep ".png\|.jpg\|.svg" | wc -l)
+if [ $fileschanged == 0 ]
+then
+    echo "No images to optimize. Did you remember to stage them?"
+    exit 1
+fi
+
+msgs=()
+
+# ask to optimize with tinypng, we don't want to do this multiple times
+read -p "Tinypng png and jpg images? " -n 1 -r
+echo    # new line
+if [[ $REPLY =~ ^[Yy]$ ]]
+then
+    # is there a tinyPNG API key
+    key="${HOME}/.tinypng"
+    if ! [ -f "$key" ]; then
+        echo "TinyPng API key not found."
+        # let's save one, shall we?
+        echo "What is your API key? (You can sign up for one here: https://tinypng.com/developers)"
+        read apikey
+        printf $apikey > ${HOME}/.tinypng
+        echo "API key saved in $key"
+    fi
+
+    # tinypng png and jpg images
+    git diff --diff-filter=ACM --name-only HEAD | grep ".png\|.jpg" | xargs tinypng
+fi
+
+# svgo svg images
+echo "Optimizing SVGs..."
+git diff --diff-filter=ACM --name-only HEAD | grep ".svg" | xargs svgo --disable=removeViewBox
+
+# check SVGs have viewbox
+echo "Checking for viewboxes..."
+svgs=`git diff --diff-filter=ACM --name-only HEAD | grep ".svg"`
+for svg in $svgs
+do
+    if ! grep -qi viewbox "$svg"; then
+        msgs+=("✘ $svg is missing viewbox")
+    fi
+done
+
+# check -high-res have corresponding low res
+echo "Checking high-res images have a matching low-res..."
+highresimages=`git diff --diff-filter=ACM --name-only HEAD | grep "\-high\-res"`
+for highresimage in $highresimages
+do
+    lowresimage=${highresimage/-high-res/}
+    if ! [ -f "$lowresimage" ]; then
+        msgs+=("✘ $lowresimage not found")
+    fi
+done
+
+if [ ${#msgs[@]} -eq 0 ]; then
+    echo "... done. Everything looks good!"
+    exit 0
+else
+    echo "... hmmm something isn't right."
+    for i in "${msgs[@]}"
+        do
+            echo $i
+    done
+    exit 1
+fi
+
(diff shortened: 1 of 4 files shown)


stephaniehobson:
I'd like a review from a front-end dev and someone who reads bash please :)

pmac:
Is this meant to be run locally or in the docker image? It'd be a good bit of setup for someone using the docker setup to run things to then set all of this up as well. Can these tools be installed in the docker image? If so then I think it'd be good to add Makefile task to run them via docker. Also, from the issue it seems most devs are interested in this being a gulp task. Was there a blocker for gulp integration here that necessitated the bash script?

stephaniehobson:
This script depends on knowing which files are staged to be committed in git so it's intended to be run locally. An `npm install` should install everything needed for that. I had hoped to attach it to a pre-commit hook but it looks like those can only be configured locally.

The thing with the PNG/JPG compression is that it should really only be run on an image once. I didn't know how to get around that with gulp so I went this way. It's not what we wanted but it is a time saver until we can have what we wanted. 

I need to add instructions :)

pmac:
Ah! Makes sense. Thanks. I believe there are projects for managing git commit hooks. I'll look that up.

pmac:
@stephaniehobson I found the projects I was thinking of. [ghooks](https://www.npmjs.com/package/ghooks) is a NodeJS project that manages your git hooks via config in `package.json`. There is also a Python project by a fellow Mozillian called [Therapist](https://therapist.readthedocs.io/en/latest/) that does something similar. Perhaps you and the team could standardize on something like one of those and easily share hooks.

stephaniehobson:
Ah, cool, using ghooks we could totally switch to doing this in gulp.

bin/img.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/bash
+
+# are there images to optimize?
+fileschanged=$(git diff --diff-filter=ACM --name-only HEAD | grep ".png\|.jpg\|.svg" | wc -l)
```
pdehaan: Not sure if this would be any easier w/ something like [**lint-staged**](https://github.com/okonet/lint-staged) and [**imagemin-cli**](https://github.com/imagemin/imagemin-cli). It would arguably be more cross-OS compatible, but probably wouldn't support things like the viewBox and high-res/low-res checks.

package.json
```diff
@@ -42,7 +42,9 @@
     "karma-coverage": "^0.5.5",
     "karma-firefox-launcher": "^0.1.7",
     "karma-jasmine": "^0.3.8",
-    "sinon": "^1.17.3"
+    "sinon": "^1.17.3",
+    "svgo": "^1.2.2",
```
alexgibson: You need to install these using `yarn` to add them to the lockfile.

bin/img.sh
```diff
@@ -0,0 +1,69 @@
...
+        echo "TinyPng API key not found."
+        # let's save one, shall we?
+        echo "What is your API key? (You can sign up for one here: https://tinypng.com/developers)"
+        read apikey
+        printf $apikey > ${HOME}/.tinypng
+        echo "API key saved in $key"
+    fi
+
+    # tinypng png and jpg images
+    git diff --diff-filter=ACM --name-only HEAD | grep ".png\|.jpg" | xargs tinypng
```
alexgibson: Pointing this to `xargs ./node_modules/tinypng-cli/tinypng-cli.js` worked for me.

bin/img.sh
```diff
@@ -0,0 +1,69 @@
...
+fi
+
+# svgo svg images
+echo "Optimizing SVGs..."
+git diff --diff-filter=ACM --name-only HEAD | grep ".svg" | xargs svgo --disable=removeViewBox
```
alexgibson: `xargs ./node_modules/svgo/bin/svgo --disable=removeViewBox`

alexgibson:
Re: the suggestions for using Gulp for this - I agree longer term a Gulp-like workflow like may be preferable, and also offer better cross-platform compatibility. For now though, I think this script looks great, and seeing as our core front-enders are all on macOS this is something we could easily iterate on.

Nice work, @stephaniehobson 💯

stephaniehobson:
Updated to (hopefully) solve the errors and add documentation ✍️

docs/coding.rst
```diff
@@ -33,6 +33,27 @@ template data as keyword arguments:
...
 
+Optimizing images
+----------------
+
+Images can take a long time to load and eat up a lot of bandwidth. Always take care
+to optimize images before uploading them to the site.
+
+The script `img.sh` can be used to optimize images locally on the command line:
+
+# Before you run it for the first time you will need to run `npm install`
```
alexgibson: run `yarn` to install dependencies.

docs/coding.rst
```diff
@@ -33,6 +33,27 @@ template data as keyword arguments:
...
+# Add the image files to git's staging area `git add *`
+# Run the script `./bin/img.sh`
+# The optimized files will not automatically be staged, so be sure to add them before commiting
+
+The script will:
+
+- optimize JPG and PNG files using `tinypng <https://tinypng.com/>`_ (this step is optional
```
alexgibson: Is it worth mentioning that you'll also need to sign up for an API key?

stephaniehobson:
Updated with doc updates.

Also, rebased on the updated dependancies. @agibson could you make sure the `yarn.lock` file looks okay to you?

PR #7247, opened 2019-05-28
Localized /firefox/67.0.1/whatsnew/ URLs redirect to /en-US/ (Fixes #7246)

alexgibson:
Fixes https://github.com/mozilla/bedrock/issues/7246

PR #7250, opened 2019-05-29
67.0.5 (it was a codename) will be 67.0.1

sylvestre:


sylvestre:
@nthomas-mozilla I know this isn't ideal but do you expect any issue?

sylvestre:
https://github.com/mozilla/bedrock/pull/7247

PR #7197, opened 2019-05-16
[fix #7132] New Firefox accounts page

craigcook:
## Description
Redesigns the accounts benefits page at /firefox/accounts/

## Issue / Bugzilla link
#7132 

## Testing
https://bedrock-demo-craigcook.oregon-b.moz.works/firefox/accounts/

PR #7255, opened 2019-05-30
Buttons and links to FxA trigger FxA-Sync event action only (#7252)

### LINKED ISSUE #7252: Use standardized FxA Referrals GA config in fxa_email_form
## Description

In #7061 it was decided that when we send someone to FxA we would like to record it in GA with an Event Action that includes FxA. FxA-ServiceName was decided on with FxA-Sync being the only service currently passed to FxA from moz.org.

With the work on #7197 I realized the button in the fxa_email_form was not updated at that time.

An [example of this form](https://www.mozilla.org/en-US/firefox/accounts/) can be found on the accounts page.

stephaniehobson:
## Description

In #7061 it was decided that when we send someone to FxA we would like to record it in GA with an Event Action that includes FxA. `FxA-ServiceName` was decided on with `FxA-Sync` being the only service currently passed to FxA from moz.org.

With the work on #7197 I realized the button in the fxa_email_form was not updated at that time.

In testing that I realized that the other links we updated in #7113 were sending two events for each click.

Screen Shot 2019-05-29 at 5 25 49 PM

This PR:

- Updates fxa_email_form to trigger event action FxA-Sync click (via the data-link-type attribute)
- Changes data-button-name to data-link-name when link-name is FxA to improve data quality

## Issue / Bugzilla link

Fix #7252

## Testing
- [ ] data-link-type and data-button-name should not be present on the same elements
- [ ] things that send users to FxA should have `data-link-type="FxA-Sync"`

### DIFF
--- a/bedrock/base/templates/includes/protocol/navigation/index.html
+++ b/bedrock/base/templates/includes/protocol/navigation/index.html
@@ -25,7 +25,7 @@
         <div class="mzp-c-navigation-download">
           {{ download_firefox(alt_copy=_('Download Firefox'), dom_id='protocol-nav-download-firefox', button_color='mzp-t-secondary mzp-t-small', download_location='nav') }}
           <div class="c-navigation-fxa-cta-container">
-            <a class="c-navigation-fxa-cta mzp-c-button mzp-t-secondary mzp-t-small mzp-t-product" {{ fxa_link_fragment(entrypoint='mozilla.org-globalnav', action='signup', utm_params={'campaign': 'globalnav', 'content': 'get-firefox-account', 'medium': 'referral', 'source': 'www.mozilla.org'}) }} data-button-name="Get a Firefox Account" data-link-type="FxA-Sync" data-cta-position="secondary cta" data-alt-href="{{ url('firefox.accounts') }}">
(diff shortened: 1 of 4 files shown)


stephaniehobson:
@hoosteeno This will impact our GA data in the following ways:
- fewer `button click` event actions
- links triggering `FxA-Sync click` will now have an event label
- the standard FxA sign in/up form will now trigger `FxA-Sync click` event actions

Are these changes okay? Could they impact our ability to measure the impact of our trailhead changes and should they therefore be left the way they are?

hoosteeno:
Thank you *very much* for the clear explanation and question.

> Are these changes okay?

If I am reading this correctly, then the only kinds of reports we might harm with this change are GA reports counting `button click` events. I have never depended on 'button click' events for reporting anything; it's a useless category, in my eyes. @gkabbz or @zerokarmaleft, do you see any reason we should not replace a few `button click` event actions in www.mozilla.org with `FxA sync click` event actions?

>  Could they impact our ability to measure the impact of our trailhead changes and should they therefore be left the way they are?

AFAIK we use Amplitude for our most important reporting on FxA funnels. If @gkabbz or @zerokarmaleft is not concerned about this change, I am not.

zerokarmaleft:
@hoosteeno @stephaniehobson We don't currently rely on this for reporting in any way, as @hoosteeno indicated, critical reporting is done through Amplitude. However, it may be useful to have this instrumentation (particularly by service) before users enter the registration funnel.

irrationalagent:
From the FxA perspective this doesn't really affect our reporting, so no objections here.

stephaniehobson:
Sounds like this is good for review then :)

stephaniehobson:
Rebased!

PR #7257, opened 2019-05-30
Use standardized adjust links across the site (Fixes #7214)

### LINKED ISSUE #7214: App store badges don't work as expected on desktop
## Description

Adjust links such as the ones on WNP67 (click "Get Lockbox" and then on an app store badge) are designed to redirect mobile user agents to their appropriate app store. On desktop, these links redirect to the Apple App Store, even when clicking on the Google Play Store button. With more interest in our non-browser products and more marketing of them, this problem will continue to confound users (see bug reports linked below).

We should fix it. Ideas:

* ~Eliminate app store badges for desktop users.~
* ~Eliminate Adjust from the mix for desktop users.~
* ~Configure an Adjust fallback link for desktop users https://docs.adjust.com/en/tracker-generation/#redirecting-users-to-landing-pages-or-in-app-locations~
* **Create Adjust links for each badge individually, that redirect to the store that corresponds with the badge selected https://docs.adjust.com/en/tracker-generation/#cross-platform-trackers-for-multiple-app-stores**

##### For Pocket:
Android: https://app.adjust.com/1ek4zlx?redirect=https%3A%2F%2Fplay.google.com%2Fstore%2Fapps%2Fdetails%3Fid%3Dcom.ideashower.readitlater.pro
iOS: https://app.adjust.com/438mmzr?redirect=https%3A%2F%2Fitunes.apple.com%2Fus%2Fapp%2Fpocket-save-read-grow%2Fid309601447

##### For Lockwise:
Android: https://app.adjust.com/gi0ycmr?redirect=https%3A%2F%2Fplay.google.com%2Fstore%2Fapps%2Fdetails%3Fid%3Dmozilla.lockbox
iOS: https://app.adjust.com/qywno6t?redirect=https%3A%2F%2Fitunes.apple.com%2Fapp%2Fid1314000270%3Fmt%3D8

##### For Notes:
No changes needed, it's already a direct link to Play Store.

----

Reported in webcompat: https://github.com/webcompat/web-bugs/issues/31572
Reported on HackerNews: https://news.ycombinator.com/item?id=19969818

----

### Steps to reproduce

Visit https://www.mozilla.org/en-US/firefox/67.0/whatsnew/ on a desktop machine. Click "Get The Lockbox App". Click the Google Play Store badge.

### Expected result

Go to Google Play Store listing.

### Actual result

Go to Apple App Store listing

### Environment

Desktop browser.

alexgibson:
## Description
- Creates standardized adjust links for mobile products that all redirect to the correct app stores if clicked on desktop.

## Issue / Bugzilla link
#7214

## Testing
Demo page: https://www-demo1.allizom.org/en-US/adjust-test/

- [ ] Links should take mobile users to the correct destinations.
- [ ] Links should take desktop users to the correct destinations.

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -54,7 +54,7 @@
 </aside>
 {%- endmacro %}
 
-{% macro google_play_button(class_name='', extra_data_attributes={}, extra_img_attributes={}, href=settings.GOOGLE_PLAY_FIREFOX_LINK, id='', product='Firefox', target='') -%}
+{% macro google_play_button(class_name='', extra_data_attributes={}, extra_img_attributes={}, href=settings.GOOGLE_PLAY_FIREFOX_LINK_UTMS, id='', product='Firefox', target='') -%}
 {% set optional_img_attributes = {'alt': _('Get it on Google Play'), 'width': '152', 'height': '45', 'l10n': True} %}
 {% do optional_img_attributes.update(extra_img_attributes) %}
 <a{% if class_name %} class="{{ class_name }}"{% endif %}{% if id %} id="{{ id }}"{% endif %}{% if target %} target="{{ target }}" rel="external noopener noreferrer"{% else %} rel="external"{% endif %} href="{{ href }}"{% if product in ['Firefox', 'Focus'] %} data-link-type="download" data-download-os="Android"{% endif %}{% if product == 'Firefox' %} data-mozillaonline-link="{{ settings.GOOGLE_PLAY_FIREFOX_LINK_MOZILLAONLINE }}"{% endif %}{% for attr, val in extra_data_attributes.items() %} data-{{ attr }}="{{ val }}"{% endfor %}>
@@ -70,7 +70,7 @@
   {% if ios_link == '' %}
     {% set ios_link = firefox_ios_url('mozorg') %}
   {% endif %}
-  {% set android_link = settings.GOOGLE_PLAY_FIREFOX_LINK %}
+  {% set android_link = settings.GOOGLE_PLAY_FIREFOX_LINK_UTMS %}
   <section id="send-to-device" class="{% if include_logo %}logo {% endif %}{% if include_title %}title{% else %}no-title{% endif %}" data-countries="{{ send_to_device_sms_countries(message_set) }}"{% if spinner_color %} data-spinner-color="{{ spinner_color }}"{% endif %}>
     <div class="form-container">
       {% if include_title %}
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -344,7 +344,7 @@ class FirefoxAndroid(_ProductDetails):
         'release': 'fennec-latest',
     }
 
-    store_url = settings.GOOGLE_PLAY_FIREFOX_LINK
+    store_url = settings.GOOGLE_PLAY_FIREFOX_LINK_UTMS
     # Product IDs defined on Google Play
     # Nightly reuses the Aurora ID to migrate the user base
     store_product_ids = {
--- a/bedrock/firefox/templates/firefox/accounts-2018.html
+++ b/bedrock/firefox/templates/firefox/accounts-2018.html
@@ -93,15 +93,15 @@
     {% endif %}
 
     <div>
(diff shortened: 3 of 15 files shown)


stephaniehobson:
I might be totally off base here but...

Aren't these links generated on a page by page basis and encoded with the origin of the link? That's why they're different on each page even when linking to the same product. I'm concerned this could break some tracking.

It might be better to update the macro to accept the link code as a parameter.

@hoosteeno  Can we confirm this somehow?

alexgibson:
> Aren't these links generated on a page by page basis and encoded with the origin of the link? That's why they're different on each page even when linking to the same product. I'm concerned this could break some tracking.

Hmm, I had assumed adjust can get this information automatically by referrer. But I could be wrong. these are links provided by Justin & Lauren in the issue, but let's find out.

alexgibson:
@hoosteeno @stephaniehobson Looking at some of our existing adjust links, I see things like `campaign=moz.org&adgroup=Focus&creative=android`. Perhaps this is a better solution for tracking origins, rather than pre-baking info into the adjust links themselves (if that is indeed happening). Creating a custom adjust links for each and every use on the website is not something that scales well, imho.

hoosteeno:
@alexgibson you have found the solution that I've been discussing with @lwa-moz. He is setting up a document that will specify the parameters for us to use. Meanwhile, we could get our macros updated to include the possible parameters: `campaign`, `adgroup`, `creative` ([ref](https://docs.adjust.com/en/tracker-generation/#creating-segments-of-users-with-tracker-hierarchies)). 

I'll comment when the values of these params are specified.

hoosteeno:
Loren set up [a helpful spreadsheet to create adjust links](https://docs.google.com/spreadsheets/d/1KXiB4Dp0Bar5cdwIjFjYPhzLXYpMS53s-DPKOPH5hp8/edit#gid=198057346); for our needs, the rules in the spreadsheet are probably most helpful. I'll try to paraphrase the rules here:

1. There is a single adjust shortlink per product that can be used anywhere on www.mozilla.org. See the spreadsheet for links per product.
2. One of the parameters of that shortlink is a redirect to an app store. This redirect ensures that desktop browsers end up where they expect after clicking. For each product, there are two possible values for this parameter, one leading to the app store listing and one to the play store listing. See the spreadsheet for redirect parameter values per product.
3. The shortlink can have other optional parameters to help attribute clicks to marketing. Suggested:

```
campaign=www.mozilla.org #the site it's coming from
adgroup=accounts-page #the page it's coming from
creative=experiment-id-variation-id-etc #totally optional, suggest building for it, but excluding from urls until we have something to put in it
```

alexgibson:
Thanks @hoosteeno. I've updated the best I can based on the spreadsheet.

A couple of observations:

1. Some raw adjust links still redirect to URLs with what seems like old campaign parameters. e.g. `https://app.adjust.com/m54twk` redirects to `https://play.google.com/store/apps/details?id=com.ideashower.readitlater.pro&referrer=adjust_reftag%3Dc471bIa7BKZMU%26utm_source%3DFirefox%2BProperties%2B%2528whatsnew%252C%2Bsnippets%252C%2Bsites%2529`. I can only assume this will confound any custom parameters we attach in bedrock, so we may need to double-check the original adjust links?
2. We're still missing adjust links for Klar.

hoosteeno:
I've commented in the spreadsheet asking @lwa-moz to jump in and address.

alexgibson:
This is now ready for code review.

bedrock/mozorg/urls.py
```diff
@@ -317,4 +317,7 @@
     url(r'^xbl$', views.namespaces, {'namespace': 'xbl'}),
 
     page('locales', 'mozorg/locales.html'),
+
+    # TODO remove before merging!
```
alexgibson: Note to self: remove this test page before merging.

ejregithub:
This will close https://bugzilla.mozilla.org/show_bug.cgi?id=1553497

bedrock/mozorg/urls.py
amychurchwell: 😄 time to remove!

alexgibson:
@amychurchwell thanks! test URL removed

PR #7259, opened 2019-05-30
Skip accounts page test

craigcook:
## Description
Test fails because the page is behind a switch, so we'll just skip it for now. We can unskip once the switch is cleaned up post-launch.

### DIFF
--- a/tests/functional/firefox/test_accounts.py
+++ b/tests/functional/firefox/test_accounts.py
@@ -7,6 +7,7 @@ import pytest
 from pages.firefox.accounts import FirefoxAccountsPage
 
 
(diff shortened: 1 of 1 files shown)


tests/functional/firefox/test_accounts.py
```diff
@@ -7,6 +7,7 @@
 from pages.firefox.accounts import FirefoxAccountsPage
 
 
+@pytest.mark.skipif(reason='Page is temporarily behind the switch "firefox_accounts_trailhead"')
```
pmac: I believe you need to use `skip()` here. `skipif()` takes another argument that is a conditional.

```suggestion
@pytest.mark.skip(reason='Page is temporarily behind the switch "firefox_accounts_trailhead"')
```

Issue #7244, opened 2019-05-24
Trailhead - WNP 67.0.1 - Design Review/Feedback


dzingeek:
@amychurchwell Just a few small tweaks.

- [ ] Button color should be #0060DF 
- [x] It's missing the "Congrats! You're using the latest..." bar in the nav area
- [ ] The text below the box should go from edge to edge of the 'shadow box' above _(See image below for reference (left = live site right = how it should look)_
- [ ] The text does not appear to be Firefox Sharp Sans or Inter (sharp sans for headlines and inter for body copy) 

![Screen Shot 2019-05-24 at 3 37 37 PM](https://user-images.githubusercontent.com/3740746/58359998-127f1780-7e3b-11e9-9626-fa4fd176ae76.png)

stephaniehobson:
The button colour and font face updates are coming in https://github.com/mozilla/bedrock/pull/7227 which is waiting on design and code review.

PR #7262, opened 2019-05-30
Trailhead WNP: add up to date banner JS, design tweaks. (Fixes #7256, #7244)

### LINKED ISSUE #7256: Trailhead /whatsnew page does not display the "Congrats! You’re using the latest version of Firefox." message.
## Description

When loading the page there is a white banner at the top where the up-to-date message should be displayed. The strings are in the template, but we're not showing them.

### Steps to reproduce

Visit: https://www.mozilla.org/en-US/firefox/67.0.1/whatsnew/ (using a Firefox browser)

### Expected result

There should be a message at the top saying you're up-to-date:

image

### Actual result

The bar is empty:

image

### Environment

Tested in both Firefox 68 and Firefox Nightly.

amychurchwell:
## Description
- Tweak design slightly (omitting new fonts and button blue color because that will arrive in a protocol update)
- Add JS to display "Congrats! You’re using the latest version of Firefox." banner.

## Issue / Bugzilla link
#7256
#7244 

## Testing
https://bedrock-demo-achurchwell.oregon-b.moz.works/en-US/firefox/67.0.1/whatsnew/

### DIFF
--- a/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.0.5.html
+++ b/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.0.5.html
@@ -22,8 +22,8 @@
 {% block content %}
 
 <main class="mzp-t-firefox">
-  <header class="mzp-l-content c-page-header">
-    <div class="c-page-header-inner">
+  <header class="c-page-header">
+    <div class="mzp-l-content c-page-header-inner">
       {{ high_res_img('firefox/whatsnew_67/logo.png', {'alt': 'Firefox', 'width': '147', 'height': '55', 'class': 'c-page-header-logo-fx'}) }}
       <h1 class="c-page-header-up-to-date"><span>{{ _('Congrats! You’re using the latest version of Firefox.') }}</span></h1>
     </div>
@@ -44,11 +44,7 @@
             }}
           {% endblock %}
         </div>
-        {% if LANG.startswith('en-') %}
-          <a href="{{ url('firefox.accounts') }}">{{ _('Learn more about joining Firefox') }}</a>
-        {% else %}
-          <a href="{{ url('firefox.accounts') }}">{{ _('Learn More about joining Firefox') }}</a>
-        {% endif %}
+        <a href="{{ url('firefox.accounts') }}">{{ _('Learn more about joining Firefox') }}</a>
       </div>
       <div class="show-fxa-supported-signed-in">
         <h1>{{ _('Your account connects you to all of our products. ') }}</h1>
--- a/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.0.5.html
+++ b/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.0.5.html
@@ -85,9 +85,7 @@
         <img class="glyphs" src="{{ static('img/firefox/whatsnew_67/heart_glyph.svg') }}" alt="">
         <div class="benefit-content">
           <h3>{{ _('Fierce defense of your privacy') }}</h3>
-          <p>{{ _('Everything we make and do honors our Personal Data Promise:') }}
-            <strong>{{ _('Take less. Keep it safe. No secrets.') }}</strong>
-          </p>
+          <p>{{ _('Everything we make and do honors our Personal Data Promise: <strong>Take less. Keep it safe. No secrets.</strong>') }}</p>
         </div>
       </div>
       <div class="show-fxa-supported-signed-in">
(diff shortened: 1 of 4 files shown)


bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.0.5.html
```diff
@@ -70,7 +74,7 @@ <h2>{{ _('Including Firefox Monitor, your watchful eye for data breaches') }}</h
         <div class="benefit-content">
           <h3>{{ _('Products that respect') }}</h3>
           <p>
-            {% trans send='https://send.firefox.com/', lockwise='https://lockbox.firefox.com/', monitor='https://monitor.firefox.com/' %}
+            {% trans send='https://send.firefox.com/', lockwise='https://lockwise.firefox.com/', monitor='https://monitor.firefox.com/' %}
             Firefox products like <a href="{{ send }}">Send</a>, <a href="{{ lockwise }}">Lockwise</a> and <a href="{{ monitor }}">Monitor</a> help you get things done without invading your privacy.
```
craigcook: This string doesn't match the original, which didn't include links, so it's not being translated. I think the only way to fix this is to do it "for real" by updating the lang files. Or else we can remove the links and go back to the original string, but I know marketing will want those links. We should talk to @peiying2 and give her a heads-up.

https://github.com/mozilla-l10n/www.mozilla.org/blob/master/en-US/firefox/whatsnew_67.0.5.lang#L83

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.0.5.html
```diff
@@ -85,7 +89,9 @@ <h3>{{ _('Knowledge that protects') }}</h3>
         <img class="glyphs" src="{{ static('img/firefox/whatsnew_67/heart_glyph.svg') }}" alt="">
         <div class="benefit-content">
           <h3>{{ _('Fierce defense of your privacy') }}</h3>
-          <p>{{ _('Everything we make and do honors our Personal Data Promise: <strong>Take less. Keep it safe. No secrets.</strong>') }}</p>
+          <p>{{ _('Everything we make and do honors our Personal Data Promise:') }}
```
craigcook: This will also break translations. The original string had no `<strong>` tag so it became a mismatch when that was added. Now it's being split, so it's two new strings, neither of which are translated. We need to either remove the `<strong>` to revert back to the original string, or update the lang files. 

Since I know marketing will want the emphasis, and since I think we have to update the lang files for the other string change above already, I think we should do this too.

https://github.com/mozilla-l10n/www.mozilla.org/blob/master/en-US/firefox/whatsnew_67.0.5.lang#L99

craigcook:
A minor nitpick, but the updated message is a little off-center.

image

It's caused by the `margin-left: 0;` on `.c-page-header`. That 0 margin is overriding the margins that are coming from `.mzp-l-content` ... so I think the fix is to remove that margin style from `.c-page-header` and remove the `mzp-l-content` class from that element, and instead add an `mzp-l-content` class to the `c-page-header-inner` element. Then the outer header is full width and the inner container gets the content width and is correctly centered. If that makes sense... sorry, it's hard to explain in words.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.0.5.html
craigcook: OK, With @peiying2's blessing I think we're ready to move ahead with updating the lang files, so this string is good.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.0.5.html
craigcook: And I also think splitting this into two strings is better than one, so this is also good.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.0.5.html
```diff
@@ -44,7 +44,11 @@ <h1>{{ _('You’ve got the latest browser. Now get the rest of Firefox.') }}</h1
             }}
           {% endblock %}
         </div>
-        <a href="{{ url('firefox.accounts') }}">{{ _('Learn more about joining Firefox') }}</a>
+        {% if LANG.startswith('en-') %}
```
craigcook: Since we're updating strings anyway, let's go ahead and remove this conditional.

Issue #7265, opened 2019-05-31
Design: Incorporate MSI in new /all UI


hoosteeno:
## Description

We want to include MSI installers in the new designs for /all. This is an interface design challenge. We need to explore it and mock it up. 

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Installer types, like MSI, are included in the /all interface somehow
- [ ] Mockups are ready for coding

#### Tasks
- [ ] This is an example of a task. Use a checklist!
- [ ] This is an example of a task.

hoosteeno:
I met with Mark, who'll give us some direction on the UI here. Mark will follow up on this comment with final direction. Below are a couple drafts, not quite ready for coding.

The demo is at https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/all/.

We discussed possibly adding installer types to the OS dropdown, as in the attached images. From a design perspective, this would mean...

- [ ] Get final names for the things in the OS dropdown. [There appear to be MSI installers for Release/Beta/Dev/Nightly Windows 64](https://support.mozilla.org/en-US/kb/deploy-firefox-msi-installers?redirectlocale=en-US&redirectslug=firefox-customization-msi-installers). That would mean splitting the Win64 item into`Windows 64-bit [exe or something]` and `Windows 64-bit [msi or something]`
- [ ] Change the label on the dropdown (are these "platforms?" instead of OS? And you are no longer "using" the one you download, necessarily). Tyler Ahn can direct us on this. [Here's an artifact](https://docs.google.com/spreadsheets/d/1-eRoBpWwYUT5nwn7aPdiOEEa-c6pHvs3vlvwewWNJ1A/edit#gid=0) where we track this kind of content.
- [ ] Add an info modal for different installer types: At launch time, this box should include .exe and .msi types. In the near future it will have pkg and dmg types. This, too, requires copy, which should be obtained from Tyler and updated in the crib sheet linked above.

Is that everything?

![Screenshot_2019-06-03 Download the Firefox browser in English (US) and more than 90 other languages](https://user-images.githubusercontent.com/94519/58813916-fe2cde80-85e1-11e9-9c5e-30a7ab58e7cd.png)

<img width="1294" alt="Screen Shot 2019-06-03 at 9 22 54 AM" src="https://user-images.githubusercontent.com/94519/58813923-008f3880-85e2-11e9-9506-996d662e8507.png">

mrkwvr:
[...]

PR #7266, opened 2019-05-31
Add support for win-msi and win64-msi in download URL helpers

pmac:
Also move `firefox-esr-next` to use bouncer aliases setup for that product. Re [bug 1408868](https://bugzilla.mozilla.org/show_bug.cgi?id=1408868).

Followup work to actually add the buttons to pages is still needed.  To use this you simply use `win-msi` or `win64-msi` as the `platform` argument to the download helpers.

Re [bug 1493205](https://bugzilla.mozilla.org/show_bug.cgi?id=1493205)
Re #7265

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -228,13 +228,21 @@ class FirefoxDesktop(_ProductDetails):
         :param locale_in_transition: Include the locale in the transition URL
         :return: string url
         """
-        _version = version
+        # no longer used, but still passed in. leaving here for now
+        # as it will likely be used in future.
+        # _version = version
         _locale = 'ja-JP-mac' if platform == 'osx' and locale == 'ja' else locale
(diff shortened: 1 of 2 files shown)


alexgibson:
@pmac from talking to @hoosteeno earlier it sounded like listing this in platforms may not be optimal, from a UI perspective. Happy to discuss more on Monday, and how this impacts the other PR (as well as the existing /all page).

pmac:
@alexgibson it sounds like the idea is to only use it on the new /all page as the others can't afford to get any wider. It doesn't need to be listed in platforms, but having a different platform string seemed like the easiest way for template developers to specify the MSI versions. I could add another `msi=True` parameter to the functions, but that gets weird if `platform` isn't a Windows.

pmac:
Also it looks like we'll have to be careful for which channels we use this. It doesn't look like ESR has those aliases ready yet e.g.

PR #7271, opened 2019-06-03
Fetch FxA metrics for non-FF browsers on /accounts (Fixes #7269)

### LINKED ISSUE #7269: Metrics are not passed via /accounts form signups using non-Firefox browsers
## Description

When completing the form on /accounts using a non-Firefox browser, `flow_id` and `flow_begin_time` are not passed.

### Steps to reproduce

1. Using Chrome, visit https://www.mozilla.org/en-US/firefox/accounts/.
2. Enter an email address and click "Continue"

### Expected result

The accounts.firefox.com URL should contain a valid `flow_id` and `flow_begin_time` parameter.

### Actual result
The accounts.firefox.com URL contains empty values for both parameters e.g. `&flow_id=&flow_begin_time=&`

### Environment

Latest Chrome browser.

alexgibson:
## Description
- Fetch `flow_id` and `flow_begin_time` for non-Firefox browsers when signing up from /accounts.

## Issue / Bugzilla link
#7269

## Testing
- [ ] Fetch request should occur for non-Firefox browsers.
- [ ] Fetch request should still occur for Firefox browsers.

### DIFF
--- a/media/js/base/mozilla-fxa-form.js
+++ b/media/js/base/mozilla-fxa-form.js
@@ -14,12 +14,7 @@ Mozilla.FxaForm = (function(Mozilla) {
     var fxaSubmitButton = document.getElementById('fxa-email-form-submit');
     var fxaFormContextField = fxaForm.querySelector('[name="context"]');
     var fxaFormServiceField = fxaForm.querySelector('[name="service"]');
-
     var supportsFetch = 'fetch' in window;
-
-    if (!supportsFetch) {
-        return;
-    }
 
     // swap form action for Fx China re-pack
     function init() {
@@ -45,6 +40,10 @@ Mozilla.FxaForm = (function(Mozilla) {
 
     // get tokens from FxA for analytics purposes
     function fetchTokens() {
+        if (!supportsFetch) {
+            return;
+        }
+
         fxaSubmitButton.disabled = false;
 
         var destURL = fxaForm.getAttribute('action') + 'metrics-flow';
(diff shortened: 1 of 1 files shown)


media/js/base/mozilla-fxa-form.js
```diff
@@ -81,6 +80,7 @@ Mozilla.FxaForm = (function(Mozilla) {
             // This allows non-Firefoxes to create accounts.
             fxaFormContextField.remove();
```
craigcook: This is my mistake. `ChildNode.remove()` isn't supported in IE, so these fields aren't being omitted and FxA tries to sign a non-Firefox into Sync, which is impossible so it just spins and spins. https://caniuse.com/#feat=childnode-remove

Alas, removing them still doesn't make the form work in IE, it just throws a 500 error, but that's arguably better than a loading animation that never finishes.

craigcook: As a fix that doesn't on `remove()`, we could just set the fields to `disabled` ?

alexgibson: The changes I've added here don't try to address IE, just fix the metrics being missing from non-FF browsers. Let's try and fix IE in a followup?

alexgibson: Oh wait, nevermind it took me a minute to understand the problem. I'll take a look

alexgibson:
@craigcook fixed the `remove()` error by switching to `removeChild()`.

Issue #7169, opened 2019-05-09
FxA utm validation: tests & remaining touchpoints


amychurchwell:
## Description

Close-out of the work in https://github.com/mozilla/bedrock/pull/7110

---

#### Tasks
- [x] Global navigation fxa cta button
- [x] Sign-in links
- [x] Write tests

stephaniehobson:
@amychurchwell Have you started this or is it okay for me to pick it up?

amychurchwell:
@stephaniehobson I haven't yet, feel free 🙂 thank you!

stephaniehobson:
This is taking longer than I expected because (as is inevitable) writing the tests had turned up bugs.

amychurchwell:
@stephaniehobson yeah, it's a bit of a can of worms. 😅

stephaniehobson:
The Sign into Monitor button on the WN 67.0.5 page should be considered when the email forms are updated.

ejregithub:
@amy - can you please apply an Estimate to this card?

ejregithub:
Noting that these tasks were pulled out and complete in #7417 
- [ ] Email-first form
- [ ] Monitor sign-in button

Closing this issue.

PR #7295, opened 2019-06-07
Pass through utm parameters on more FxA referrals

stephaniehobson:
## Description

This work is part of mozilla/bedrock#6854. The goal is to attribute FxA account signups to their point of origin. If that's mozilla.org that's us. If someone arrived on mozilla.org by following a link with UTM params, we should pass that attribution to FxA.

Technically what happens is:
- Links to FxA are generated with the macro `fxa_link_fragment` which automatically adds UTM params giving attribution to mozilla.org (the format for these is specified in mozilla/bedrock#7061)
- When the page loads `fxa-utm-referral.js` runs and if there are links on the page with class `js-fxa-cta-link` and UTM params in the query string it will append or replace the UTM params on the link.
  - The function first verifies that the links are to accounts.firefox.com to avoid race conditions with [`initFxAButton`](https://github.com/mozilla/bedrock/blob/master/media/js/base/protocol/init-navigation.js#L59-L70) as described in [this comment ](https://github.com/mozilla/bedrock/pull/7110/files#r281961672)on #7110 
  - The function also updates the query string on the `data-mozillaonline-link` attribute, which is present for the China Repack.
- In addition to the UTM attribution tracking we also track FxA referrals in GA for our own use. This is achieved with `data-link-type="FxA-ServiceName"` (as specified in mozilla/bedrock#7061). `data-link-type` must not be used with `data-button-name` but can be used with `data-link-name`.

The foundation of this behaviour was coded in #7110, and #7169 was created to track follow up work.

---
This PR will:

- DOCUMENT THIS MESS
- Modify fxa-utm-referral.js to facilitate testing
- Modify fxa-utm-referral.js to decode URL before doing regex validation
- Add tests for fxa-utm-referral.js
- Add class `js-fxa-cta-link` to links using `fxa_link_fragment` macro
- Make `js-fxa-cta-link` not optional in fxa_email_form macro

## Issue / Bugzilla link

The original issue for this work is #6854 which was closed by #7110.
This PR addresses some of the follow up tasks outlined in #7169.

## Testing
- Do the tests run?
- Do the tests pass?
- Do we need to test anything else?
- Does the regex still protect us from XSS?
- Do links get UTMs by default?
- Do those UTMs get over-ridden if valid ones are in the URL?
- Please check the global nav button specifically. Maybe spot check another place?
- Think about things you should test I have not listed here.


ejregithub:
@stephaniehobson - Please size the work done here in the Estimate menu

docs/firefox-accounts.rst
```diff
@@ -45,3 +45,81 @@ Firefox:
 #. If testing locally, be sure to use ``127.0.0.1`` instead of ``localhost``
 
 .. _instructions: https://github.com/vladikoff/fxa-dev-launcher#basic-usage-example-in-os-x
+
+
+
+===============================
+Linking to accounts.firefox.com
```
alexgibson: ❤️

media/js/base/fxa-utm-referral.js
```diff
@@ -78,24 +75,29 @@
...
-            var newAccountsLink = UtmUrl.appendToDownloadURL(oldAccountsLink, params);
-
-            // set the FxA button to use the new link.
-            ctaLinks[i].href = newAccountsLink;
-
-            // also handle mozilla-online links for FxA China Repack.
-            if (oldMozillaOnlineLink) {
-                var newMozillaOnlineLink = UtmUrl.appendToDownloadURL(oldMozillaOnlineLink, params);
-                ctaLinks[i].setAttribute('data-mozillaonline-link', newMozillaOnlineLink);
+            // verify this a link to accounts.firefox.com or dev server, otherwise we shouldn't touch it
```
alexgibson: Small typo `verify this is a link`?

media/js/base/fxa-utm-referral.js
```diff
@@ -78,24 +75,29 @@
...
+            if(oldAccountsLink.indexOf('https://accounts.firefox.com') === 0 || oldAccountsLink.indexOf('https://latest.dev.lcip.org') === 0) {
```
alexgibson: Thank you for carefully going back through the comments in the old PR and picking out this piece of code that was omitted 🙏

tests/unit/spec/base/fxa-utm-referral.js
```diff
@@ -0,0 +1,251 @@
+/* For reference read the Jasmine and Sinon docs
+ * Jasmine docs: http://pivotal.github.io/jasmine/
+ * Sinon docs: http://sinonjs.org/docs/
+ */
+
+/* global describe, it, expect, Mozilla */
+
+describe('fxa-utm-referral.js', function() {
```
alexgibson: Tests! 💯

bedrock/base/templates/macros.html
```diff
@@ -412,7 +412,7 @@ <h2 class="fxa-email-form-title">{{ form_title }}</h2>
         <input type="email" name="email" id="fxa-email-field" class="fxa-email-field" placeholder="user@example.com" required>
       </p>
 
-      <button type="submit" class="{{ button_class }}" id="fxa-email-form-submit" data-link-type="FxA-{{ service|title }}" data-link-name="Continue">
+      <button type="submit" class="js-fxa-cta-link {{ button_class }}" id="fxa-email-form-submit" data-link-type="FxA-{{ service|title }}" data-link-name="Continue">
```
alexgibson: Placing this class on a `<button>` results in the error `TypeError: oldAccountsLink is undefined`, since the JS code expects an `href` property. (this can be observed when loading /accounts with a valid query param).

To handle attribution on form submissions, I imagine we might need a separate piece of logic to append / update hidden inputs with utm values. Thoughts?

media/js/base/fxa-utm-referral.js
```diff
@@ -78,24 +75,29 @@
         }
 
         for (var i = 0; i < ctaLinks.length; i++) {
-            // get the current FxA links.
+            // get the link off the element
             var oldAccountsLink = ctaLinks[i].href;
```
alexgibson: Would adding a check that the link `href` is not undefined here be worthwhile, to avoid the script throwing an error should someone accidentally add the class to another type of element?

alexgibson:
Do we also need to do this for the "Sign In to Monitor" button on the 67.0.5 whatsnew page?

bedrock/base/templates/macros.html
stephaniehobson: Yeah, I think the forms need to be addressed in a separate PR. I've removed the class here to avoid the error.

stephaniehobson: ✅

media/js/base/fxa-utm-referral.js
stephaniehobson: You mean like I did? 😅 

✅

media/js/base/fxa-utm-referral.js
stephaniehobson: ✅

stephaniehobson:
Updated.

I think we do need it for the Sign In to Monitor button, I will make a note in #7169.

I think it's possible that in the long run we will want this code to run on any link that has UTM params. But one thing at a time.

alexgibson:
This has been reviewed and r+'d, but we're not mergin until after all-hands.

PR #7304, opened 2019-06-11
Add MSI installers to /all page (Fixes #7209)

### LINKED ISSUE #7209: Add MSI installers to redesigned /all page
In the new /all page (#7076) we allow visitors to choose a version of Firefox to download using a set of dropdowns. The following user story applies to this interface:

* As a user, I want to download Firefox MSI installers from the /all page

The following functional requirements support this user story:

1. MSI installers should be included in the list of products available via the dropdowns
2. When MSI installers are selected in the dropdowns, the dynamic elements on the page and in the URL bar change to reflect the MSI installer is selected

The following task would improve the value of this feature:

* The SUMO page linking directly to MSI downloads (https://support.mozilla.org/en-US/kb/deploy-firefox-msi-installers?redirectlocale=en-US&redirectslug=firefox-customization-msi-installers) instead links to the new /all page with a deep link for the MSI installer.

alexgibson:
## Description
- Adds MSI installers to `/firefox/all/` page and `download_firefox_desktop_list()` helper.

**Waiting on final design / copy**

## Issue / Bugzilla link
#7209

## Testing
Demo: https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/all/

- [x] MSI installers for Windows 32bit and 64bit should be present in the dropdown on `/all`.
- [x] MSI installer should not be present or ESR.
- [x] Clicking "Download Now" should give the correct MSI installer.
- [x] MSI installers should be present in the "Advanced Install Options & Other Platforms" modal on `/firefox/new/`.
- [x] No regressions in download buttons.

Successful test run: https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/324/pipeline


alexgibson:
@pmac strings aren't final, but I think in terms of functionality the backend changes here are ready. r?

bedrock/firefox/firefox_details.py
```diff
@@ -143,6 +150,7 @@ def latest_builds(self, locale, channel='release'):
                 # Append 64-bit builds
                 if 'Windows' in _builds:
                     _builds['Windows 64-bit'] = _builds['Windows']
+                    # TODO do we need to do anything here for MSI?
```
alexgibson: @pmac one area of code I wasn't sure about was here. I'm not entirely clear on the need for special casing 64-bit builds. But it made me wonder do we need to do anything for MSI, or is it safe to remove this todo?

pmac: Oh right. I saw that and meant to address it after running locally but forgot. Surprisingly I don't think the `latest_builds` function is used for much of anything anymore. It's called in the `desktop_builds` function in `helpers.py`, but the data isn't used. Even the version returned that is passed to `get_download_url()` isn't used by the function anymore since it no longer produces any bouncer links with version numbers. For now you can safely ignore this, but soon we should be able to clean all of this up a bit more and remove some of this legacy code.

alexgibson:
Ok final copy is in place and this is finally good to review / merge. Upping this to a P1, since MSI installers are long overdue.

claychaffin:
Hi @alexgibson where can I review the copy in the design, please?

alexgibson:
@claychaffin it’s linked on demo above.

claychaffin:
> @claychaffin it’s linked on demo above.

Thank you. Copy changes verified. Nice work, sir — looks great, works great. One comment: I don't understand how the SUMO page helps with the user's installer choice per se. The person is looking to determine what type of machine they have— to download Firefox, likely for the first time —yet the SUMO information is about "How to tell whether you _have_ 32-bit or 64-bit Firefox" — is this the best SUMO help we have for the user's questions in this instance? Maybe I just don't understand?

alexgibson:
Thanks @claychaffin good catch - I did a search on SUMO and found this article, which seems more appropriate https://support.mozilla.org/en-US/kb/choosing-firefox-cpu-architecture-windows-os

PR #7318, opened 2019-06-13
Add signed-in state on /firefox/accounts/ (Fixes #6907)

### LINKED ISSUE #6907: The /firefox/accounts/ page gives no clear indication that I'm signed-in.
## Description

If I'm already signed-in to a Firefox Account, the `/firefox/accounts/` page gives no clear indication of this.

### Steps to reproduce

In Firefox for desktop, make sure you're signed into a Firefox Account and then visit: https://www.mozilla.org/en-US/firefox/accounts/

### Expected result

The page should give some clear indication that I'm already signed into my Firefox Account, and then describe what I can do next to make the most of the benefits.

Perhaps somethign like: "Congrats! You're already signed-in to your Firefox Account. Now go get the Firefox mobile app so you can enjoy all the benefits".

### Actual result

All the the main CTA says is "Download Firefox App" and nothing else, which feels a little too brief to be clear (especially seeing as I'm already using Firefox). A little added context to go along with this button could make things a lot clearer?

### Environment

Any Firefox desktop browser that is signed-in to an account.

alexgibson:
## Description
- Adds signed-in state for Firefox users on `/firefox/accounts`.
- New string added behind the tag `accounts-signed-in-06132019`.

**Do not merge** until we have copy approval.

## Issue / Bugzilla link
#6907

## Testing
Demo: https://www-demo5.allizom.org/en-US/firefox/accounts/

- [ ] Signed-in users should see a Firefox Monitor CTA.
- [ ] All other states should still see the Join Firefox CTA.
- [ ] Locales that don't have the new strings translated should still get the current experience.

### DIFF
--- a/bedrock/firefox/templates/firefox/accounts-2019.html
+++ b/bedrock/firefox/templates/firefox/accounts-2019.html
@@ -3,7 +3,7 @@
   # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
 
 {% from "macros-protocol.html" import feature_card with context %}
-{% from "macros.html" import fxa_email_form, fxa_link_fragment with context %}
+{% from "macros.html" import fxa_email_form, fxa_link_fragment, monitor_button with context %}
 
 {% extends "firefox/base-protocol.html" %}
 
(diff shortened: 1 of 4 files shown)


alexgibson:
Copy is final, this is good for review.

amychurchwell:
Not a blocker or the purpose of this PR, but I think we should consider in the future adding the 'download firefox' cta in the nav for non-firefox browsers on this page.

Screen Shot 2019-06-13 at 10 14 47 AM

alexgibson:
I consider the nav button beyond scope of this bug, but yes I agree.

PR #7322, opened 2019-06-14
Remove fxa_create_account_button macro (#7294)

### LINKED ISSUE #7294: Update or remove fxa_create_account_button macro
The `fxa_create_account_button` macro was designed to create a "Create Firefox Account" Call to Action for Firefox Desktop users, fall back to "Download Firefox" for non Fx users or link to SUMO documentation for Fx Mobile users.

With the trailhead re-focus this macro is no longer used. Can we remove it?

Was there an associated JS file that can also be removed?

stephaniehobson:
## Description

Remove fxa_create_account_button macro

## Issue / Bugzilla link

Fix #7294

## Testing

I checked the [original PR where this macro was added](https://github.com/mozilla/bedrock/pull/5749) and the other JS changes made with it were generic functions added to global files. They are still in use to do things like hide and show content based on the user's FxA Sign in status.


alexgibson:
Looks good but needs a rebase.

Issue #7327, opened 2019-06-24
Improve functional tests for /firefox/accounts/


alexgibson:
## Description

Now that #7318 has merged, we have two distinct states for the /firefox/accounts/ page. One for signed-out users, showing the "Join Firefox" form, and one for signed-in users, showing a Firefox Monitor CTA.

Our current functional tests for this page are pretty basic. All we have is a [single test](https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/test_accounts.py#L13) that asserst that the form is displayed, but not much else.

To better test that this page works as intended it would be good to:

1. Expand the existing test to cover the functionality of the form itself. If we enter information and click "Continue", does it do the right thing? I don't think we need to test the full account signup flow, as that is not within bedrock's remit. However it would be good to assert that the right URL is loaded, together with the expected email parameter e.g. 

    - https://accounts.firefox.com/signup?action=email&entrypoint=mozilla.org-accounts_page&form_type=email&flow_id=d45ef062b25c7428583c2b765dbd370457275bfbf6f29bec07629fbe1bed551c&flow_begin_time=1561365277162&utm_source=accounts_page&utm_content=form-upper&utm_medium=referral&utm_campaign=accounts-trailhead&style=trailhead&email=success%40example.com

2. Add a new functional test to cover the signed-in state. Automating the full sign-up and verification flow is likely to be a large and complicated task. That would be good to cover at some point, but for now we could provide a shortcut just for testing purposes e.g. if `&s=t` is in the URL, show the signed-in state. It's not ideal, but would at least provide a lower effort testing mechanism. This param might also be helpful for stakeholder demo reviews.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [x] New tests are passing in automation.

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] Not improving tests increases risk of accidental breakage.

#### Tasks
- [x] Improve signed-out test.
- [x] Add signed-in test.

stephaniehobson:
Closed by #7329

PR #7329, opened 2019-06-24
Improve functional tests for /firefox/accounts/ page

alexgibson:
## Description
- Updates existing test to cover functionality of the "Join Firefox" form.
- Adds a new test for signed-in state.
- Adds a `signed-in=true` query param override to `mozilla-fxa.js` for easier state testing.

## Issue / Bugzilla link
#7327

## Testing
Successful integration test run: https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/325/pipeline

### DIFF
--- a/media/js/base/mozilla-fxa.js
+++ b/media/js/base/mozilla-fxa.js
@@ -27,17 +27,18 @@ if (typeof window.Mozilla === 'undefined') {
     FxaState.convertFxaDetailsToStateAndDo = function(details, callback) {
         var stateClass = '';
 
-        if(details.firefox) {
-            if(details.legacy) {
+        if (details.firefox) {
+            if (details.legacy) {
                 stateClass = 'state-fxa-unsupported';
-            } else if(details.mobile) {
-                if(details.mobile === 'ios') {
+            } else if (details.mobile) {
+                if (details.mobile === 'ios') {
                     stateClass = 'state-fxa-ios';
                 } else {
                     stateClass = 'state-fxa-android';
                 }
             } else {
-                if(details.setup) {
+                // Query param `signed-in=true` is used for integration testing and demo review only.
+                if (details.setup || window.location.search.indexOf('signed-in=true') !== -1) {
                     stateClass = 'state-fxa-supported-signed-in';
                 } else {
                     stateClass = 'state-fxa-supported-signed-out';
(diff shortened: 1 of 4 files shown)


media/js/base/mozilla-fxa.js
```diff
@@ -27,17 +27,18 @@ if (typeof window.Mozilla === 'undefined') {
...
+            } else if (details.mobile) {
+                if (details.mobile === 'ios') {
                     stateClass = 'state-fxa-ios';
                 } else {
                     stateClass = 'state-fxa-android';
                 }
             } else {
-                if(details.setup) {
+                // Query param `signed-in=true` is used for integration testing and demo review only.
+                if (details.setup || window.location.search.indexOf('signed-in=true') !== -1) {
```
stephaniehobson: 😍

tests/functional/firefox/test_accounts.py
```diff
@@ -3,11 +3,23 @@
...
 
 @pytest.mark.nondestructive
 def test_account_form(base_url, selenium):
-    page = FirefoxAccountsPage(selenium, base_url).open()
-    assert page.is_create_account_form_displayed, "Create account form is displayed for signed out users"
+    page = FirefoxAccountsPage(selenium, base_url, params='').open()
+    page.join_firefox_form.type_email('success@example.com')
+    page.join_firefox_form.click_continue()
+    url = urllib.unquote(selenium.current_url)
+    assert 'email=success@example.com' in url, 'Email address is not in URL'
```
stephaniehobson: I don't know if Selenium is actually clicking with a mouse or with JS. If it's with JS there's a chance the form can be submitted without being visible. Should we also assert the form is displayed?

alexgibson: Good question! Selenium fires real click events based on coordinates, rather than simulating mouse events directly on DOM nodes (this is one of the main reasons I favour it over most JS-based solutions). If something isn't visible on a page and you try to click it, a `element not interactable` error will be thrown. I think this is enough to certify that the form is / isn't visible indirectly 👍

stephaniehobson: Then this is good to go!

PR #7350, opened 2019-06-27
Upgrade protocol to 7.0.1

amychurchwell:
## Description
Update protocol to the latest & greatest. 🙂

### DIFF
--- a/media/css/firefox/features/adblocker.scss
+++ b/media/css/firefox/features/adblocker.scss
@@ -18,12 +18,6 @@ ul li span {
     font-weight: bold;
 }
 
-blockquote {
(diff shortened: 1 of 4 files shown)


stephaniehobson:
If you do want to do remove the custom blockquote styles separately, it's okay just to make an issue for it. I didn't mean to assign you more work if you don't have time for it.

amychurchwell:
@stephaniehobson good point! Thanks for calling that out. I found two pages on protocol with custom blockquote styles that I've removed in this as well. 🙂

stephaniehobson:
Thanks for doing the clean up!

PR #7395, opened 2019-07-08
Compare master and stage hash during deployment

stephaniehobson:
## Description

- Checks `master` and `stage` to make sure a tagged release matches what is on staging
- Fixes the double question that was hurting my head

## Issue / Bugzilla link

No bug this has just been, er, bugging me.

### DIFF
--- a/bin/tag-release.sh
+++ b/bin/tag-release.sh
@@ -22,9 +22,9 @@ done
 # compare master and stage
 echo "Comparing master branch to staging..."
 git fetch "$moz_git_remote"
-stage_hash=`git rev-parse $moz_git_remote/stage`
-master_hash=`git rev-parse master`
-if [ $stage_hash != $master_hash ]; then
+stage_hash=$(git rev-parse "${moz_git_remote}/stage")
+master_hash=$(git rev-parse master)
+if [[ "$stage_hash" != "$master_hash" ]]; then
     read -p "Master branch does NOT match stage branch! Are you sure you want to continue? (Type Override to continue, n to cancel)" no_match
     if [ "$no_match" == "${no_match#[Override]}" ]; then
         # do not continue tagging release
--- a/bin/tag-release.sh
+++ b/bin/tag-release.sh
@@ -19,9 +19,25 @@ while [[ $# -ge 1 ]]; do
   shift # past argument or value
 done
 
+# compare master and stage
+echo "Comparing master branch to staging..."
+git fetch "$moz_git_remote"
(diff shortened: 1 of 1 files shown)


bin/tag-release.sh
```diff
@@ -19,9 +19,25 @@ while [[ $# -ge 1 ]]; do
...
 done
 
+# compare master and stage
+echo "Comparing master branch to staging..."
+git fetch "$moz_git_remote"
+stage_hash=`git rev-parse $moz_git_remote/stage`
+master_hash=`git rev-parse master`
+if [ $stage_hash != $master_hash ]; then
+    read -p "Master branch does NOT match stage branch! Are you sure you want to continue? (Type Override to continue, n to cancel)" no_match
+    if [ "$no_match" == "${no_match#[Override]}" ]; then
```
pmac: I think this should just be:

```suggestion
    if [[ "$no_match" == "Override" ]]; then
```

bin/tag-release.sh
```diff
@@ -19,9 +19,25 @@ while [[ $# -ge 1 ]]; do
   shift # past argument or value
 done
 
+# compare master and stage
+echo "Comparing master branch to staging..."
+git fetch "$moz_git_remote"
+stage_hash=`git rev-parse $moz_git_remote/stage`
```
pmac: It's more reliable and portable to use `$()` instead of the back-ticks. 

```suggestion
stage_hash=$(git rev-parse "${moz_git_remote}/stage")
```

bin/tag-release.sh
```diff
@@ -19,9 +19,25 @@ while [[ $# -ge 1 ]]; do
...
+master_hash=`git rev-parse master`
+if [ $stage_hash != $master_hash ]; then
```
pmac: style NIT: always wrap variables in double quotes:

```suggestion
if [[ "$stage_hash" != "$master_hash" ]]; then
```

This protects against the case when one is blank and would throw a syntax error without the quotes.

bin/tag-release.sh
pmac: Or if you want/need to only check that it starts with "Override" then this would work:

```suggestion
    if [[ "$no_match" == Override* ]]; then
```

stephaniehobson: I tried both suggestions but both continued when I typed no :/

stephaniehobson:
Updated with review fixes.

bin/tag-release.sh
pmac: Huh. Oh well. Sorry about that :/

PR #7396, opened 2019-07-08
Removed Attributes causing duplication of events [Fixes #7392]

### LINKED ISSUE #7392: Remove instances where both data-link-type and data-button-name attributes are used.
## Description

We should not use `data-link-type` and `data-button-name` attributes together or the event will be captured twice by GA.

I have found two places in bedrock this occurs: 
- In the mission CTA in the `about-2019.html` template.
https://github.com/mozilla/bedrock/blob/f6b1b88b778b3096da45bfe2b4322147e9eeb71d/bedrock/mozorg/templates/mozorg/about-2019.html#L40

- The `fxa_banner` macro in `macros.html`
https://github.com/mozilla/bedrock/blob/f6b1b88b778b3096da45bfe2b4322147e9eeb71d/bedrock/mozorg/templates/mozorg/home/includes/macros.html#L62

Errolyn:
## Description
- removed two cases of "data-link-type=button" where "data-link-type" is captured.
- Double checked for other instances where this occurs.
- Ran tests locally - all pass
## Issue / Bugzilla link
#7392
## Testing
We might want to make sure we are getting events as expected?

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about-2019.html
+++ b/bedrock/mozorg/templates/mozorg/about-2019.html
@@ -37,7 +37,7 @@
       </div>
 
       <p class="mzp-c-hero-cta">
-        <a id="read-mission-button" class="mzp-c-button" href="{{ url('mozorg.mission') }}" data-button-name="Read Our Mission" data-link-type="button" data-cta-position="primary cta">{{ _('Read Our Mission') }}</a>
+        <a id="read-mission-button" class="mzp-c-button" href="{{ url('mozorg.mission') }}" data-button-name="Read Our Mission" data-cta-position="primary cta">{{ _('Read Our Mission') }}</a>
       </p>
     </div>
   </div>
--- a/bedrock/mozorg/templates/mozorg/about-2019.html
+++ b/bedrock/mozorg/templates/mozorg/about-2019.html
@@ -37,7 +37,7 @@
       </div>
 
       <p class="mzp-c-hero-cta">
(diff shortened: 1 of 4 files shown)


bedrock/mozorg/templates/mozorg/about-2019.html
```diff
@@ -37,7 +37,7 @@ <h1 class="mzp-c-hero-title">{{ _('Mozilla makes browsers, apps, code and tools
       </div>
 
       <p class="mzp-c-hero-cta">
-        <a id="read-mission-button" class="mzp-c-button" href="{{ url('mozorg.mission') }}" data-button-name="Read Our Mission" data-link-type="button" data-cta-position="primary cta">{{ _('Read Our Mission') }}</a>
+        <a id="read-mission-button" class="mzp-c-button" href="{{ url('mozorg.mission') }}" data-button-name="Read Our Mission" data-cta-position="primary cta">{{ _('Read Our Mission') }}</a>
```
alexgibson: If these are links then should we be using `data-link-type="button"` and `data-link-name="Read Our Mission"` instead?

I'm actually not 100% sure on this, so I asked the question in the issue for clarification:

https://github.com/mozilla/bedrock/issues/7392#issuecomment-509545915

Errolyn:
I went through and searched the rest of the project to make sure there weren't any other instances of using data-button-name inside of an <a> tag, and found a couple more which I switched to data-link-name & data-link-type="button"

PR #7397, opened 2019-07-08
Standardize fxa_email_form UTM values. (Fixes #7387)

### LINKED ISSUE #7387: UTM values for FxA referrals are inconsistent across /whatsnew pages
## Description

Whilst working on https://github.com/mozilla/bedrock/pull/7369 it became clear that our /whatsnew page templates lack a consistent format for passing UTM values on FxA referrals.

### Steps to reproduce

Examples:

- [Firefox 68 trailhead template](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx68-trailhead.html#L13-L15)
- [Firefox 68 standard template](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx68.html#L13-L14)
- [Firefox 67 trailhead template](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.0.5.html#L41-L43)
- [Firefox 67 standard template](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.html#L20-L24)
- Firefox 66 template uses two conflicting sets of values:
  - [These seem unused](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx66.html#L19-L23)
  - [Are these ones correct?](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx66.html#L52)
- [Firefox 65 template](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx65.html#L36)

### Expected result

There should be a standardized set of parameters, and these should be documented somewhere.

### Actual result

I have no real idea which format is correct.

### Environment

N/A

amychurchwell:
## Description
There were some inconsistencies on different /whatsnew templates on format of utm params in the `fxa_email_form` macros, this PR should standardize them.

## Issue / Bugzilla link
#7387

## Testing
- Consistent UTM values across FxA referrals on /whatsnew pages.
- Any templates missed?

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -292,7 +292,7 @@
 
     DO NOT include utm_source in this dictionary! (It's already a required param.)
 #}
-{% macro fxa_email_form(entrypoint, utm_source, style, class_name='fxa-email-form', utm_params={}, form_title='', intro_text='', button_text='', button_class='button button-blue') -%}
+{% macro fxa_email_form(entrypoint, utm_source, style, campaign, content, term, class_name='fxa-email-form', form_title='', intro_text='', button_text='', button_class='button button-blue') -%}
   {% set service  = 'sync' %}
   <form action="{{ settings.FXA_ENDPOINT }}" data-mozillaonline-action="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}" id="fxa-email-form" class="{{ class_name }}">
     <input type="hidden" name="action" value="email" />
@@ -304,14 +304,30 @@
     <input type="hidden" name="flow_id" value="" />
     <input type="hidden" name="flow_begin_time" value="" />
 
-    {# utm_source is required, other utm_ params are optional #}
+    {# utm_source is required #}
     <input type="hidden" name="utm_source" value={{ utm_source }} id="fxa-email-form-utm-source" />
-  {% for attr, val in utm_params.items() %}
-    <input type="hidden" name="utm_{{ attr }}" value="{{ val }}" id="fxa-email-form-utm-{{ attr }}" />
-  {% endfor %}
-  {% if style %}
-    <input type="hidden" name="style" value="{{ style }}" />
-  {% endif %}
+    <input type="hidden" name="utm_medium" value="referral" id="fxa-email-form-utm-medium" />
+
+    {# utm_campaign is used to identify specific marketing campaigns. (For example: trailhead or skyline) If there is one, the campaign name is appended to the default value of `fxa-embedded-form`. #}
+    {% if campaign %}
+      <input type="hidden" name="utm_campaign" value="{{ campaign }}-fxa-embedded-form" id="fxa-email-form-utm-campaign" />
+    {% else %}
+      <input type="hidden" name="utm_campaign" value="fxa-embedded-form" id="fxa-email-form-utm-campaign" />
+    {% endif %}
+
+    {# utm_content is optional. It should only be declared when there is new messaging associated with the form. (For example: `get-the-rest-of-firefox`) #}
+    {% if content %}
+      <input type="hidden" name="utm_content" value="{{ content }}" id="fxa-email-form-utm-content" />
+    {% endif %}
+
+    {# utm_term is optional. It is used for paid search keywords. This should be provided if needed. #}
+    {% if term %}
+      <input type="hidden" name="utm_term" value="{{ term }}" id="fxa-email-form-utm-term" />
+    {% endif %}
+
+    {% if style %}
+      <input type="hidden" name="style" value="{{ style }}" />
+    {% endif %}
 
     {% if form_title %}
       <h2 class="fxa-email-form-title">{{ form_title }}</h2>
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -292,7 +292,7 @@
 
     DO NOT include utm_source in this dictionary! (It's already a required param.)
 #}
-{% macro fxa_email_form(entrypoint, utm_source, style, campaign, content, term, class_name='fxa-email-form', form_title='', intro_text='', button_text='', button_class='button button-blue') -%}
+{% macro fxa_email_form(entrypoint, utm_source, utm_campaign, utm_content, utm_term, style, class_name='fxa-email-form', form_title='', intro_text='', button_text='', button_class='button button-blue') -%}
   {% set service  = 'sync' %}
   <form action="{{ settings.FXA_ENDPOINT }}" data-mozillaonline-action="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}" id="fxa-email-form" class="{{ class_name }}">
     <input type="hidden" name="action" value="email" />
@@ -310,19 +310,19 @@
 
     {# utm_campaign is used to identify specific marketing campaigns. (For example: trailhead or skyline) If there is one, the campaign name is appended to the default value of `fxa-embedded-form`. #}
     {% if campaign %}
-      <input type="hidden" name="utm_campaign" value="{{ campaign }}-fxa-embedded-form" id="fxa-email-form-utm-campaign" />
+      <input type="hidden" name="utm_campaign" value="{{ utm_campaign }}-fxa-embedded-form" id="fxa-email-form-utm-campaign" />
     {% else %}
       <input type="hidden" name="utm_campaign" value="fxa-embedded-form" id="fxa-email-form-utm-campaign" />
     {% endif %}
 
     {# utm_content is optional. It should only be declared when there is new messaging associated with the form. (For example: `get-the-rest-of-firefox`) #}
     {% if content %}
-      <input type="hidden" name="utm_content" value="{{ content }}" id="fxa-email-form-utm-content" />
+      <input type="hidden" name="utm_content" value="{{ utm_content }}" id="fxa-email-form-utm-content" />
     {% endif %}
 
     {# utm_term is optional. It is used for paid search keywords. This should be provided if needed. #}
     {% if term %}
-      <input type="hidden" name="utm_term" value="{{ term }}" id="fxa-email-form-utm-term" />
+      <input type="hidden" name="utm_term" value="{{ utm_term }}" id="fxa-email-form-utm-term" />
     {% endif %}
 
     {% if style %}
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -309,19 +309,19 @@
     <input type="hidden" name="utm_medium" value="referral" id="fxa-email-form-utm-medium" />
 
     {# utm_campaign is used to identify specific marketing campaigns. (For example: trailhead or skyline) If there is one, the campaign name is appended to the default value of `fxa-embedded-form`. #}
-    {% if campaign %}
+    {% if utm_campaign %}
       <input type="hidden" name="utm_campaign" value="{{ utm_campaign }}-fxa-embedded-form" id="fxa-email-form-utm-campaign" />
     {% else %}
       <input type="hidden" name="utm_campaign" value="fxa-embedded-form" id="fxa-email-form-utm-campaign" />
     {% endif %}
 
     {# utm_content is optional. It should only be declared when there is new messaging associated with the form. (For example: `get-the-rest-of-firefox`) #}
-    {% if content %}
+    {% if utm_content %}
       <input type="hidden" name="utm_content" value="{{ utm_content }}" id="fxa-email-form-utm-content" />
     {% endif %}
 
     {# utm_term is optional. It is used for paid search keywords. This should be provided if needed. #}
-    {% if term %}
+    {% if utm_term %}
       <input type="hidden" name="utm_term" value="{{ utm_term }}" id="fxa-email-form-utm-term" />
     {% endif %}
 
--- a/bedrock/firefox/templates/firefox/developer/whatsnew-fx68.html
+++ b/bedrock/firefox/templates/firefox/developer/whatsnew-fx68.html
@@ -10,8 +10,7 @@
 {{ fxa_email_form(
     entrypoint='mozilla.org-whatsnew68a2',
     button_class='mzp-c-button',
-    utm_source='whatsnew',
-    utm_params={'campaign': 'fxa-embedded-form', 'content': 'whatsnew', 'medium': 'referral'})
+    utm_source='whatsnew68'
 }}
 {% endblock %}
 
--- a/bedrock/firefox/templates/firefox/developer/whatsnew-fx68.html
+++ b/bedrock/firefox/templates/firefox/developer/whatsnew-fx68.html
@@ -10,7 +10,7 @@
 {{ fxa_email_form(
     entrypoint='mozilla.org-whatsnew68a2',
     button_class='mzp-c-button',
-    utm_source='whatsnew68'
+    utm_source='whatsnew68')
 }}
 {% endblock %}
 
--- a/bedrock/firefox/templates/firefox/developer/whatsnew-fx68.html
+++ b/bedrock/firefox/templates/firefox/developer/whatsnew-fx68.html
@@ -10,7 +10,7 @@
 {{ fxa_email_form(
     entrypoint='mozilla.org-whatsnew68a2',
     button_class='mzp-c-button',
-    utm_source='whatsnew68')
+    utm_source='whatsnew68a2')
 }}
 {% endblock %}
 
--- a/bedrock/firefox/templates/firefox/firstrun/firstrun-quantum.html
+++ b/bedrock/firefox/templates/firefox/firstrun/firstrun-quantum.html
@@ -65,7 +65,7 @@
       <div id="fxaccounts-wrapper">
         {{ fxa_email_form(
           entrypoint='mozilla.org-firstrun',
-          utm_source='firstrun,
+          utm_source='firstrun',
           button_class='mzp-c-button mzp-t-product')
         }}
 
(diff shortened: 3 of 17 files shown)


amychurchwell:
Added trailhead prefix to the `utm_source` of the trailhead template of wnp68. (`whatsnew68-trailhead`)
I did this because I believe `style='trailhead'` is passed purely to effect the branding of the fxa sign-up and I thought it was important to be able to tell the difference between wnp68 and the trailhead version.

Also, I'm not sure if passing `content` is needed in the context of the `fxa_email_form`?

alexgibson:
I'll also defer to @hoosteeno's verification that the value changes made here are OK.

alexgibson:
Another observation: sometimes we’re passing `medium`, other times we’re not including it.

amychurchwell:
@alexgibson thanks for catching those extra templates, I'll update today. Agree with thoughts about `medium` and `campaign`. I'm going to go ahead and bake those into the macro. That will also ensure that they are always being passed.

I mentioned this briefly before but:
I think we should seek direction on what is useful information to have in `content`. It's usually purpose of the CTA ("urlsafe content of CTA"), but the purpose can often be inferred by what it is. (For example: `fxa_email_form` will always have the messaging of `get-firefox-account`. Is this redundant or helpful?) [docs](https://bedrock.readthedocs.io/en/latest/firefox-accounts.html#utm-parameters)

hoosteeno:
Thanks for working on this!

Before getting in the weeds, I want to point out that, excepting `utm_source`, the utm parameters are rarely used in www FxA referral analyses I'm aware of. This may be an artifact of them not being uniform and useful, so this PR is important work; and even so, we can make a quick & reasonable decision about our parameters' values without achieving perfection.

> Side thought: if some values never change, such as campaign and medium, could they be baked into the macro itself vs having to pass them explicitly every time?

I agree, as long as we can override them. See below.

> Added trailhead prefix to the utm_source of the trailhead template of wnp68. (whatsnew68-trailhead)

In general we should try to follow convention on utm parameters. [Convention suggests](https://ga-dev-tools.appspot.com/campaign-url-builder/) we put "trailhead" somewhere in the campaign parameter. 

Our varied use of UTM parameters means that convention often fails us. In these cases it's generally better to follow patterns we've established elsewhere on the site. [Patterns also suggest](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/accounts-2019.html#L72-L77) we put "trailhead" in the campaign parameter.

Considering these, I wonder if you agree with the below approach?

### Defaults
(context: https://github.com/mozilla/bedrock/issues/7387#issuecomment-509399696)
```
utm_source=whatsnew{{release_number}}
utm_campaign=fxa-embedded-form # only on embedded forms
utm_medium=referral
utm_content= # no default; there is not a useful default value to send here
```

### Example params (including overrides) for WNP68 Trailhead, signed out state
```
utm_source=whatsnew68
utm_campaign=trailhead-fxa-embedded-form 
utm_medium=referral # no override needed
utm_content=get_the_rest_of_firefox # this is a terse summary of the novel content in snake-case, possibly useful in this instance and particularly useful for experiments where we might have another treatment. It could as easily be "v2" or "green_button" if such were helpful for future analysis.
```

### Example params (including overrides) for WNP68 non-trailhead, signed out state
```
utm_source=whatsnew68
utm_campaign=fxa-embedded-form # no override needed
utm_medium=referral # no override needed
utm_content= # there is probably not a useful value to send here, unless the content is really novel
```

Does this seem reasonable? Is this general spec helpful?

bedrock/base/templates/macros.html
```diff
@@ -292,13 +292,15 @@ <h3>{{ _('Your download link was sent.') }}</h3>
...
-{% macro fxa_email_form(entrypoint, utm_source, style, class_name='fxa-email-form', utm_params={}, form_title='', intro_text='', button_text='', button_class='button button-blue') -%}
+{% macro fxa_email_form(entrypoint, utm_source, style, class_name='fxa-email-form', campaign='', utm_params={}, form_title='', intro_text='', button_text='', button_class='button button-blue') -%}
   {% set service  = 'sync' %}
   <form action="{{ settings.FXA_ENDPOINT }}" data-mozillaonline-action="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}" id="fxa-email-form" class="{{ class_name }}">
     <input type="hidden" name="action" value="email" />
     <input type="hidden" name="context" value="fx_desktop_v3" />
     <input type="hidden" name="service" value="{{ service }}" />
     <input type="hidden" name="entrypoint" value="{{ entrypoint }}" id="fxa-email-form-entrypoint" />
+    <input type="hidden" name="utm_campaign" value="{{ campaign }}-fxa-embedded-form" id="fxa-email-form-utm-campaign" />
+    <input type="hidden" name="utm_medium" value="referral" id="fxa-email-form-utm-medium" />
```
alexgibson: If I try and override either of these values it creates a duplicate input instead of overriding the value e.g.

```
utm_params={'campaign': 'accounts-trailhead', 'medium': 'foo'})
```

At this point, I'm not sure if the `for` loop for utm_params still makes sense?

```
{% for attr, val in utm_params.items() %}
    <input type="hidden" name="utm_{{ attr }}" value="{{ val }}" id="fxa-email-form-utm-{{ attr }}" />
{% endfor %}
```

Would it be better to make each utm param explicit in the macro, so we can then make sure values are overridden and not duplicated?

bedrock/base/templates/macros.html
```diff
@@ -292,13 +292,15 @@ <h3>{{ _('Your download link was sent.') }}</h3>
...
 #}
-{% macro fxa_email_form(entrypoint, utm_source, style, class_name='fxa-email-form', utm_params={}, form_title='', intro_text='', button_text='', button_class='button button-blue') -%}
+{% macro fxa_email_form(entrypoint, utm_source, style, class_name='fxa-email-form', campaign='', utm_params={}, form_title='', intro_text='', button_text='', button_class='button button-blue') -%}
   {% set service  = 'sync' %}
   <form action="{{ settings.FXA_ENDPOINT }}" data-mozillaonline-action="{{ settings.FXA_ENDPOINT_MOZILLAONLINE }}" id="fxa-email-form" class="{{ class_name }}">
     <input type="hidden" name="action" value="email" />
     <input type="hidden" name="context" value="fx_desktop_v3" />
     <input type="hidden" name="service" value="{{ service }}" />
     <input type="hidden" name="entrypoint" value="{{ entrypoint }}" id="fxa-email-form-entrypoint" />
+    <input type="hidden" name="utm_campaign" value="{{ campaign }}-fxa-embedded-form" id="fxa-email-form-utm-campaign" />
```
alexgibson: Justin seems to suggest [in the issue](https://github.com/mozilla/bedrock/pull/7397#issuecomment-509692686) that this value should be `fxa-embedded-form` or `trailhead-fxa-embedded-form`, but we seem to be passing things like `accounts-fxa-embedded-form` or `firefox-new-fxa-embedded-form` which is specific to the page itself. Is that what we want? (I'm not clear here, it may be the case but thought I should ask).

Also, if we omit the `campaign` parameter here the value ends up being `-fxa-embedded-form` (extra `-` at the start, which isn't quite what we need).

bedrock/firefox/templates/firefox/whatsnew/beta/whatsnew-fx68.html
```diff
@@ -10,8 +10,7 @@
 {{ fxa_email_form(
     entrypoint='mozilla.org-whatsnew68beta',
     button_class='mzp-c-button',
-    utm_source='whatsnew',
-    utm_params={'campaign': 'fxa-embedded-form', 'content': 'whatsnew', 'medium': 'referral'})
```
alexgibson: `utm_campaign` value on this form ends up being `-fxa-embedded-form` because we're omitting the value.

bedrock/firefox/templates/firefox/whatsnew/fx64/whatsnew-fx64-base.html
```diff
@@ -35,7 +35,7 @@
 {% set fxa_utm_params = {
   'campaign': 'wnp64',
```
alexgibson: Do `campaign` and `content` here need updating?

bedrock/base/templates/macros.html
alexgibson: I'd also suggest ordering these next to `utm_source` which is already hard-coded down on line 310.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.html
```diff
@@ -18,9 +18,8 @@
 {% endblock %}
 
 {% set fxa_utm_params = {
-  'campaign': 'wnp67',
-  'content': '/firefox/WNP/67/',
-  'source': 'wnp-67'
+  'campaign': 'fxa-embedded-form',
```
alexgibson: This is creating a duplicate `utm_campaign` input in the form.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx67.html
```diff
@@ -18,9 +18,8 @@
...
+  'content': 'get-firefox-account'
```
alexgibson: In the issue @hoosteeno suggests making utm_content snake_case. I'm not quite clear why that is. Can we just use dashes like here so it's all consistent? (this makes it much easier for devs to remember).

I'm also not quite clear on when utm_content is required, and what does / doesn't make it useful. We know the purpose of this form, so is `get-firefox-account` useful information?

stephaniehobson: I'm not sure what the right answer is here but either way could it please be[ added to the documentation](https://bedrock.readthedocs.io/en/latest/firefox-accounts.html).

amychurchwell: In regards to content: when I was looking at [this link](https://ga-dev-tools.appspot.com/campaign-url-builder/) shared in [this comment](https://github.com/mozilla/bedrock/pull/7397#issuecomment-509692686) it says that `utm_content` is:
> Used for A/B testing and content-targeted ads. Use utm_content to differentiate ads or links that point to the same URL.
> Examples: logolink or textlink

So perhaps it should only be used in A/B test templates?

stephaniehobson: I take it back, I have an opinion ;) Looking at all our other UTM values I see no underscores so we should stick with dashes unless there is a specific reason for the change here.

hoosteeno: > we should stick with dashes unless there is a specific reason for the change here

Fine with me, I was just seeking uniformity. If we standardize on kebab case, that's great. The standardization is the main thing.

> I'm also not quite clear on when utm_content is required...
> So perhaps it should only be used in A/B test templates?

This is a good shorthand. All 4 utm parameters are *required* for stub attribution; that fact can confuse us in other contexts like this one, where utm_content is *optional*. As an occasional analyst, I think utm_content can also be useful when we are doing something really different, just to help downstream analysts orient themselves. The "get more from Firefox" claim is pretty different, which is why I thought it might provide some modest value downstream, at minimal cost I think?

bedrock/base/templates/macros.html
amychurchwell: I think I was confused because of what was already there. I think the standard should be campaign names if applicable: ex: `fxa-embedded-form` or `trailhead-fxa-embedded-form`, and in the future `skyline-embedded-form`.
I will remove all page-specific values. 👍

bedrock/base/templates/macros.html
amychurchwell: Agree with making the utm params that we've been discussing explicit.

As far as the `far` loop. It may need to remain? In `scene1.html` I'm seeing `'term': 'existing-users'` in the `utm_params` items which I am not seeing in other places. I'm not sure what is happening there, but it feels important and other pages in the future may need to add a unique param for whatever reason.

bedrock/base/templates/macros.html
alexgibson: I believe there are only a finite set of allowed utm params. Couldn’t we add term as a parameter? It only seems to be used in one place, and the for loop can lead to a confusing state.

bedrock/base/templates/macros.html
alexgibson: I believe there are only a finite set of allowed utm params. Couldn’t we add term as a parameter? It only seems to be used in one place, and the for loop can lead to a confusing state.

amychurchwell: [According to Google](https://support.google.com), there are 5 utm params you can add. Including `utm_term` which is used to "identify paid search keywords".
So it makes total sense to just make each param explicit. Tossing the loop.

bedrock/firefox/templates/firefox/whatsnew/fx64/whatsnew-fx64-base.html
amychurchwell: It looks like this template and the templates extending it aren't using the imported `fxa_email_form` macro so I'm going to remove it.
It also seems these values are being used by `fxa_link_fragment` which I'll update. Going to make a separate issue to standardize UTMs in other macros as well.

bedrock/base/templates/macros.html
```diff
@@ -292,7 +292,7 @@ <h3>{{ _('Your download link was sent.') }}</h3>
 
     DO NOT include utm_source in this dictionary! (It's already a required param.)
 #}
-{% macro fxa_email_form(entrypoint, utm_source, style, class_name='fxa-email-form', utm_params={}, form_title='', intro_text='', button_text='', button_class='button button-blue') -%}
+{% macro fxa_email_form(entrypoint, utm_source, style, campaign, content, term, class_name='fxa-email-form', form_title='', intro_text='', button_text='', button_class='button button-blue') -%}
```
alexgibson: Nit: suggest prefixing these new parameters with `utm_*`, so that they are consistent with the `utm_source` parameter?

bedrock/firefox/templates/firefox/firstrun/firstrun-quantum.html
```diff
@@ -63,7 +63,11 @@ <h1>
       {% endblock %}
       </header>
       <div id="fxaccounts-wrapper">
-        {{ fxa_email_form(entrypoint='mozilla.org-firstrun_page', utm_source='firstrun-page', utm_params={'campaign': 'fxa-embedded-form', 'content': 'firstrun-page', 'medium': 'referral'}, button_class='mzp-c-button mzp-t-product',) }}
+        {{ fxa_email_form(
+          entrypoint='mozilla.org-firstrun',
+          utm_source='firstrun,
```
alexgibson: Missing `'` here at the end of `firstrun`

bedrock/base/templates/macros.html
```diff
@@ -304,14 +304,30 @@ <h3>{{ _('Your download link was sent.') }}</h3>
...
-  {% for attr, val in utm_params.items() %}
-    <input type="hidden" name="utm_{{ attr }}" value="{{ val }}" id="fxa-email-form-utm-{{ attr }}" />
-  {% endfor %}
-  {% if style %}
-    <input type="hidden" name="style" value="{{ style }}" />
-  {% endif %}
+    <input type="hidden" name="utm_medium" value="referral" id="fxa-email-form-utm-medium" />
+
+    {# utm_campaign is used to identify specific marketing campaigns. (For example: trailhead or skyline) If there is one, the campaign name is appended to the default value of `fxa-embedded-form`. #}
+    {% if campaign %}
```
alexgibson: This should be `utm_campaign`?

bedrock/base/templates/macros.html
```diff
@@ -304,14 +304,30 @@ <h3>{{ _('Your download link was sent.') }}</h3>
...
+      <input type="hidden" name="utm_campaign" value="{{ utm_campaign }}-fxa-embedded-form" id="fxa-email-form-utm-campaign" />
+    {% else %}
+      <input type="hidden" name="utm_campaign" value="fxa-embedded-form" id="fxa-email-form-utm-campaign" />
+    {% endif %}
+
+    {# utm_content is optional. It should only be declared when there is new messaging associated with the form. (For example: `get-the-rest-of-firefox`) #}
+    {% if content %}
```
alexgibson: `utm_content`

bedrock/base/templates/macros.html
```diff
@@ -304,14 +304,30 @@ <h3>{{ _('Your download link was sent.') }}</h3>
...
+      <input type="hidden" name="utm_content" value="{{ utm_content }}" id="fxa-email-form-utm-content" />
+    {% endif %}
+
+    {# utm_term is optional. It is used for paid search keywords. This should be provided if needed. #}
+    {% if term %}
```
alexgibson: `utm_term`

bedrock/firefox/templates/firefox/developer/whatsnew-fx68.html
```diff
@@ -10,8 +10,7 @@
 {{ fxa_email_form(
     entrypoint='mozilla.org-whatsnew68a2',
     button_class='mzp-c-button',
-    utm_source='whatsnew',
-    utm_params={'campaign': 'fxa-embedded-form', 'content': 'whatsnew', 'medium': 'referral'})
+    utm_source='whatsnew68')
```
alexgibson: This should be `whatsnew68a2`

PR #7402, opened 2019-07-09
Update stylelint (Fixes #6869)

### LINKED ISSUE #6869: Update bedrock to the latest version of Stylelint.
## Description
Now that bedrock has updated to Gulp 4.0, we should also update our version of Stylelint.
---

- [x] Bedrock uses the latest version of Stylelint.
- [x] CSS assets are free of linting errors.

#### Tasks
- [x] Update `gulp-stylelint` to the latest version on NPM.
- [x] Update bedrock's stylelint rules where necessary.
- [x] Fix linting issues where necessary.

alexgibson:
## Description
- Updates `stylelint` and `gulp-stylelint` to the latest version.
- Updates `merge-stream` (dependency for bumping CircleCI image to Node.js 10.0).

## Issue / Bugzilla link
#6869

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -36,5 +36,8 @@
     "no-extra-semicolons": true,
     "no-missing-end-of-source-newline": true,
     "no-eol-whitespace": true
-  }
+  },
+  "ignoreFiles": [
+    "media/css/sandstone/mixins/at2x.less"
+  ]
 }
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -341,7 +341,8 @@ function cssLint() {
             reporters: [{
                 formatter: 'string',
                 console: true
-            }]
+            }],
+            debug: true
         }));
 }
 gulp.task('css:lint', cssLint);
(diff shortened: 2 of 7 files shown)


.stylelintrc
```diff
@@ -36,5 +36,8 @@
     "no-extra-semicolons": true,
     "no-missing-end-of-source-newline": true,
     "no-eol-whitespace": true
-  }
+  },
+  "ignoreFiles": [
+    "media/css/sandstone/mixins/at2x.less"
```
alexgibson: Note for reviewer: I had to move the Less version of the `.at2x()` mixin to a separate file here, so that it can be ignored. The latest Stylelint throws an error when trying to parse the inline JavaScript contained inside the `at2x` function. 

I considered filing an issue over on [postcss-less](https://github.com/shellscape/postcss-less), however I also learned that inline JS [has been deprecated](http://lesscss.org/usage/#command-line-usage-options-specific-to-lessc) in Less. I decided rather than go down that rabbit hole, to make an exception here and instead concentrate on removing Less from bedrock.

PR #7006, opened 2019-03-29
Update firefox/mobile/ english pages with new brand (#6987)

stephaniehobson:
## Description

- Add a temporary English only firefox/mobile page with new branding.
- Protocol style sheet for send to device widget
- Protocol style sheet for firefox subnav

## Issue / Bugzilla link

Fix #6987

## Testing
- [x] shows to all English locales
- [x] send to device widget (opens when clicking blue "Get Firefox" button)
- [x] Focus QR code
- [ ] good fallbacks for old browsers
- [x] Google Play or Apple App store images appear if user is already in Android or iOS

PR #7458, opened 2019-07-16
Protocol based styles for the send to device widget

stephaniehobson:
## Description

Protocol based styles for the send to device widget.

These were tested and okay as part of #7006, but those changes are unlikely to make it into production. This bit is useful so I'm pulling it out so others can use it.

## Issue / Bugzilla link

n/a

## Testing

Was tested as part of #7006 so it's probably okay. You can pull that branch down and test there if you want.

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -120,7 +120,7 @@
               <input id="id-input" name="phone-or-email" type="text" required>
             </div>
             <div class="form-submit">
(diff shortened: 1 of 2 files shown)


alexgibson:
Nice! Can I suggest we update the docs to promote using the new protocol based styles, over the old .less version? https://bedrock.readthedocs.io/en/latest/send-to-device.html

Issue #7358, opened 2019-06-28
Convert /privacy section to Protocol


stephaniehobson:
These pages are on a very out of date template. The brand team would like to update the section to better reflect our commitment to privacy. We've talked to legal and they would also like that.

With #7141 we have a template that can be used for most of the pages in the section. 

Legal has also requested that we update https://www.mozilla.org/en-US/about/legal/terms/mozilla/ as part of this work.

alexgibson:
This was done

PR #7462, opened 2019-07-17
Update privacy section to use Protocol base templates (#7358)

stephaniehobson:
## Description

- all HTML is in base-protocol template now
- notice-headings and notice-paragraphs will process the markdown to fit the common template
  - privacy notices seem to fit one of 2 patterns (except Reality which needs a bit of help)
- add navigation menu to all pages
  - it still only contains the same notices as the homepage though
- use .mzp-c-details to hide/show with the notice-headings template
- create custom hide/show behaviour for the old paragraph style notices to avoid changing any legal wording
- some notices do not have hide/show functionality and have the `format-none` class on them
- Create a class for "side-reference-title" and apply to all headings in side-references
- Move the standstone template used for archived privacy notices into archive folder
- Add side_extra content block to privacy base template
- Update index, faq, and principles pages to use privacy base template

## Issue / Bugzilla link

See #7358.

## Testing

- the privacy notices are cached for 60 minutes after they're built. You will have to re-start your web server if you make changes to the HTML and want to see it take effect.
- there are more privacy notices impacted by this work than can be reached by the nav menu, have a look in the folder to see all the names of the notices
- the contents of the side nav will be revisited in the next few weeks, for now I just re-used what was on the privacy homepage.
- please compare the content to the live site carefully, I am worried I've missed something and don't want to screw up on our legal notices.


stephaniehobson:
Updated the remaining 3 pages in this section of the site to use Protocol and added that work to this PR.

Issue #7314, opened 2019-06-12
Add a link to survey on /contribute/ 


LucyeoH:
## Description

We've created a short, [4 question survey](https://www.surveygizmo.com/s3/5058890/Contribute-Site-Survey) that we'd like to add to the top of the contribute page to help us better understand users current experience of the site. This survey has been reviewed and approved by the survey group. It is currently live but not posted publicly.

**Placement:** As a banner or pop-up on the site (whatever is easiest to implement)

**Survey link:** https://www.surveygizmo.com/s3/5058890/Contribute-Site-Survey

**Banner text:** "We would to love hear from you: [Share Your Opinion]"

Sample : 
![image (1)](https://user-images.githubusercontent.com/8070833/59376415-042b6980-8d1e-11e9-8c6e-0956f1f90b50.png)

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] 10k Respondents

LucyeoH:
Note I'm also working to translate this into the top 11 additional languages for /fr/contribute etc. but don't want that to block this one if it doesn't have to.

ejregithub:
@LucyeoH - do you have a developer prepared to submit a pull request for this work or do you need the Websites team's assistance for doing so?  If the latter, we need to understand the requested go live date, if the survey is already live (do we need to coordinate go live if not?).

LucyeoH:
Hey the survey is already live, we'd love to see this go live by the end of the quarter. If your team has time that would be great. I'm also asking to see if we have availability on our end and will update here.

LucyeoH:
Hey @johngian will submit a PR - July 5th

LucyeoH:
@nukeador to share translated versions with @johngian

nukeador:
EN: Share your opinion
GE: Sag uns deine Meinung
FR: Donnez votre avis
ES: Comparte tu opinion
RU: Поделитесь своим мнением

These will show as link below the English text:

> We would to love hear from you:
> [Share your opinion]
> [Sag uns deine Meinung]
> [Donnez votre avis]
> [Comparte tu opinion]
> [Поделитесь своим мнением]

alexgibson:
This survey was added in the PR referenced above. I'm going to close this - feel free to open a new issue if any other locales need to be added.

PR #4279, opened 2016-08-16
[fix bug 1290996] Deploy survey to existing Firefox users coming back…

alexgibson:
## Description
- Adds a survey link to /firefox/new for up-to-date / returning visitors.

Marking as do-not-merge for now until we get the final survey link / copy.

## Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1290996

## Checklist
- [ ] Requires l10n changes.
- [x] Related functional & integration tests passing.

… to the download page

PR #7466, opened 2019-07-17
Show survey (per locale) in /contribute page

johngian:
## Description
Shows up a `surveygizmo` survey in /contribute page. The survey is going to be different per locale. Currently only focusing in `en-US`. This is going to be extended for the rest of top 5 locales of this page.

## Issue / Bugzilla link
#7314 

## Testing
I've already tested locally that different (dummy) surveys shows up per different locale. For now `en-US` shows up as expected.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/contribute/index.html
+++ b/bedrock/mozorg/templates/mozorg/contribute/index.html
@@ -150,7 +150,27 @@
 
 {% endblock contrib_content %}
 
+{% block cta %}
+
+  {% set localized_surveys = {'en-US': 'GHJYW6D'} %}
+
+  {% if LANG in localized_surveys and switch('contribute-survey') %}
+    <div id="survey-message">
+      <span class="survey-invite">
+        Hello! Would you be willing to take a minute to answer a few questions?
+      </span>
+      <a class="survey-button" href="https://www.surveymonkey.com/r/{{ localized_surveys[LANG] }}">
+        Sure. I’ll help.
+      </a>
+    </div>
+  {% endif %}
+
+{% endblock %}
+
 {% block js %}
   {{ js_bundle('contribute-base') }}
   {{ js_bundle('contribute-landing') }}
+  {% if switch('contribute-survey') %}
+    {{ js_bundle('contribute-survey') }}
+  {% endif %}
 {% endblock %}
--- a/media/static-bundles.json
+++ b/media/static-bundles.json
@@ -916,6 +916,12 @@
     },
     {
       "files": [
+        "js/mozorg/contribute/contribute-survey.js"
+      ],
+      "name": "contribute-survey"
+    },
+    {
+      "files": [
         "js/firefox/new/variation-scene1.js",
         "js/firefox/campaign/berlin-scene1-variation.js"
       ],
--- a/bedrock/mozorg/templates/mozorg/contribute/contribute-base.html
+++ b/bedrock/mozorg/templates/mozorg/contribute/contribute-base.html
@@ -120,6 +120,9 @@
   </section>
   {% endblock contrib_footer %}
 </main>
+
+{% block cta %}{% endblock %}
+
 {% endblock content %}
 
 {% block email_form %}{% endblock %}
--- a/media/css/mozorg/contribute/contribute-base.less
+++ b/media/css/mozorg/contribute/contribute-base.less
@@ -1832,3 +1832,37 @@ textarea:focus {
     }
 
 }
+
+#survey-message {
+    .border-box;
+    .font-size(12px);
+    background: #fff0b3;
+    bottom: 0;
+    display: none;
+    min-height: 30px;
+    overflow: hidden;
+    padding: 10px 20px;
+    position: fixed;
+    text-align: center;
+    width: 100%;
+    z-index: 90;
+    transform: translateY(100%);
+
+    .survey-button {
+        background-color: #00b6e0;
+        border-radius: 2px;
+        color: #fff;
+        height: 20px;
+        margin-left: 10px;
+        padding: 2px 5px;
+    }
+
+    &.show {
+        display: block;
+    }
+
+    &.show.animate {
+        transition: transform .25s ease-in;
+        transform: translateY(0);
+    }
+}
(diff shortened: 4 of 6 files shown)


alexgibson:
Thankf for this PR @johngian :)

We've used Survey Gizmo on mozorg in the past, but we usually just link directly to surveys using plain old HTML. Think we can do that here rather than make another CSP exception for a third-party script? It's not something we've really vetted (and it looks like it loads things like jQuery in the background which shouldn't really be needed). 

It would also be great not to have the link shown in a modal on page load, as this kind of ttreatment can negatively impact UX / SEO. Embedding the link in-page would be our preffered treatment. Thanks

johngian:
@alexgibson Hey! Do you have an example on how you used surveygizmo in the past? I am up to change the PR with just HTML, i just followed the surveygizmo docs.

alexgibson:
We've used various treatments in the past. Here's one that displayed as a small strip at the bottom of the viewport: 

image

At other times we've inserted a CTA in the mian flow of the page near the main header. The implementation is really up to you, but I'm happy to try and help if needed.

alexgibson:
Here's an example of how we did this survey in the past: https://github.com/mozilla/bedrock/pull/4279/files

johngian:
Hey @alexgibson I updated the PR using a CTA for the survey link.  My 2 concerns at this point are:

* I am using the same styling with what you send me (due to lack of an alternative :) )
* Is there a reason why the survey you mentioned was desktop only? My PR also shows up the CTA for mobile as well.

Let me know what you think.

bedrock/mozorg/templates/mozorg/contribute/index.html
```diff
@@ -150,6 +150,23 @@ <h2 class="section-title">{{ _('Not ready to dive in just yet?') }}</h2>
...
 {% endblock contrib_content %}
 
+{% block cta %}
+
+  {% set localized_surveys = {'en-US': 'GHJYW6D'} %}
+
+  {% if LANG in localized_surveys %}
+    <div id="survey-message">
+      <span class="survey-invite">
+        {{ _('Hello! Would you be willing to take a minute to answer a few questions?') }}
```
alexgibson: Note that this string isn't localized, so I'd suggest that we hardcode it as opposed to wrap it in the L10n function call here. When it comes to adding localized surveys, we'll need to provide a translations here. But we can worry about that when the time comes.

bedrock/mozorg/templates/mozorg/contribute/index.html
```diff
@@ -150,6 +150,23 @@ <h2 class="section-title">{{ _('Not ready to dive in just yet?') }}</h2>
...
+      </span>
+      <a class="survey-button" href="https://www.surveymonkey.com/r/{{ localized_surveys[LANG] }}">
+        {{ _('Sure. I’ll help.') }}
```
alexgibson: Same as above, please hardcode the string.

media/static-bundles.json
```diff
@@ -910,7 +910,8 @@
   "js": [
     {
       "files": [
-        "js/mozorg/contribute/contribute-nav.js"
+        "js/mozorg/contribute/contribute-nav.js",
+        "js/mozorg/contribute/contribute-survey.js"
```
alexgibson: Instead of adding this to the page's base bundle, I'd suggest we create a new bundle just for the survey JS. We can then include that in the template, so we can control showing it via a switch. This lets us turn the survey on / off independently of a deployment.

bedrock/mozorg/templates/mozorg/contribute/index.html
```diff
@@ -150,6 +150,23 @@ <h2 class="section-title">{{ _('Not ready to dive in just yet?') }}</h2>
...
+      </a>
+    </div>
+  {% endif %}
+
+{% endblock %}
+
 {% block js %}
   {{ js_bundle('contribute-base') }}
   {{ js_bundle('contribute-landing') }}
```
alexgibson: Suggest adding a bundle just for the survey here, which we can turn on / off via a switch:

```
  {% if switch('contribute-survey') %}
    {{ js_bundle('contribute-survey') }}
  {% endif %}
```

alexgibson:
> Is there a reason why the survey you mentioned was desktop only? My PR also shows up the CTA for mobile as well.

The prior survey was only applicable to desktop users. If you want this to show for everyone, that's fine too :)

alexgibson:
I pushed this branch to our test runner, and it looks like the survey is breaking an integration test as it overlaps a clickable element: https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/350/pipeline/

This is a non-critical page, so I think we're fine to skip the test for the duration that the survey is shown. Please add the following marker to [this test](https://github.com/mozilla/bedrock/blob/master/tests/functional/contribute/test_contribute.py#L10)

```python
@pytest.mark.skip(reason='https://github.com/mozilla/bedrock/issues/7314')
```

bedrock/mozorg/templates/mozorg/contribute/index.html
johngian: Since the feature doesn't rely only on JS to be toggled I added a switch to the `div` as well.

johngian:
Hey @alexgibson thanks for the feedback. I updated the PR after your comments.

alexgibson:
Changes look good, thanks @johngian!

If it's safe to assume that this is going to be a short-lived survey, and you're ok showing it to 100% of users then I think we're ok to merge this as-is. If it's something that may live for several weeks or longer, then I think we should probably make it so the banner can be dismissed.

Thoughts?

johngian:
> Changes look good, thanks @johngian!
> 
> If it's safe to assume that this is going to be a short-lived survey, and you're ok showing it to 100% of users then I think we're ok to merge this as-is. If it's something that may live for several weeks or longer, then I think we should probably make it so the banner can be dismissed.
> 
> Thoughts?

I will defer to @LucyeoH for that

LucyeoH:
@alexgibson @johngian Our goal is ~7k respondents (roughly 7% of monthly visitors). My hope is that we'll be able to get that within a few weeks. My suggestion would be to launch as is and then look at # of responses in 2 weeks, if it looks like we're not on track for that, we add a dismiss button.

alexgibson:
Ok great, let's see how engagement goes. If you get way more survey takers than you need, we can always shut it off quickly.

Thanks both!

PR #7470, opened 2019-07-18
Port unsupported-systems page to Protocol (Fixes #7454)

### LINKED ISSUE #7454: Port /firefox/unsuppported-systems/ page to Protocol
## Description

The [unsupported systems](https://www.mozilla.org/en-US/firefox/unsupported-systems/) page is still using Sandstone & Less. This one should be a relatively quick and easy port to Protocol.

---

- [ ] Page is updated in production using Protocol.

#### Tasks
- [x] Update template.
- [x] Remove legacy .less assets.

alexgibson:
## Description
- Ports `/firefox/unsupported-systems/` page to Protocol base template.
- Fixes unwrapped strings.

## Issue / Bugzilla link
#7454

## Testing
- [x] Legacy assets are all removed.


alexgibson:
Note for reviewer: I decided to omit the newsletter from this updated version of the page, as we don't yet have a suitable drop-in graphic. I figured people aren't very likely to want the newsletter anyway, if we're telling them they can't use our product - it feel like a perhaps inappropriate crosssell? :/

Errolyn:
I'm going to hold off on merging to get your thoughts first :D

alexgibson:
@Errolyn thanks for the review and suggestion! I tried applying a callout to the section, but it doesn't really fit well enough seeing as this is a narrow page, and the callout is designed to run full width. I think the new emphasis box is probably the right solution here and would look really great. I'm not going to block merging this seeing as that's still in development though. It would be great to revisit this page and make that change when we update Protocol.

Going to merge this as is for now, and we can revisit later :)

PR #7471, opened 2019-07-18
Port MPL pages to Protocol (Fixes #6464)

### LINKED ISSUE #6464: Update MPL page to Protocol
https://www.mozilla.org/en-US/MPL/ is still using outdated styling. It could use an update to use Protocol. My guess is that there are still a few of these around, but I saw this one mentioned on Slack today.

alexgibson:
## Description
Ports the following pages to use Protocol templates:

- http://localhost:8000/en-US/MPL/
- http://localhost:8000/en-US/MPL/historical/
- http://localhost:8000/en-US/MPL/license-policy/
- http://localhost:8000/en-US/MPL/headers/
- http://localhost:8000/en-US/MPL/1.1/
- http://localhost:8000/en-US/MPL/1.1/FAQ/
- http://localhost:8000/en-US/MPL/1.1/annotated/
- http://localhost:8000/en-US/MPL/2.0/
- http://localhost:8000/en-US/MPL/2.0/FAQ/
- http://localhost:8000/en-US/MPL/2.0/Revision-FAQ/
- http://localhost:8000/en-US/MPL/2.0/combining-mpl-and-gpl/
- http://localhost:8000/en-US/MPL/2.0/differences/
- http://localhost:8000/en-US/MPL/2.0/permissive-code-into-mpl/

Note: Some pages, such as `/MPL/1.1/` and `/MPL/2.0/` are treated like pre-formatted license pages, and should remain looking exactly like they did prior to this change.

## Issue / Bugzilla link
#6464

## Testing
- [x] Lists should all be styled.
- [x] Headings should be appropriately sized.
- [x] Content should be appropriately spaced out and readable. 
- [x] Check I didn't miss any other MPL pages?


media/css/mozorg/mpl-1-1.scss
```diff
@@ -2,7 +2,10 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-@import '../pebbles/includes/lib';
+$font-path: '/media/fonts';
```
stephaniehobson: This file and the other mpl .scss files don't need to include the library. Not a blocker but a bit tidier.

alexgibson:
@stephaniehobson thanks, updated!

PR #7473, opened 2019-07-18
Add relative link to license file (mozilla#7360)

AConati:
Added link to license file in the README. I haven't contributed to open source before so let me know if I did something wrong. Thanks!

(mozilla#7360)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,10 @@ License
 -------
 
 This software is licensed under the [MPL version 2.0][MPL]. For more
-information, read the file ``LICENSE``.
+information, read this repository's [LICENSE][LICENSE].
 
 [MPL]: https://www.mozilla.org/MPL/
+[LICENSE]: https://github.com/mozilla/bedrock/blob/master/LICENSE
 
 ![](http://i.imgur.com/ElotJSI.jpg)
 


README.md
```diff
@@ -46,7 +46,7 @@ License
 -------
 
 This software is licensed under the [MPL version 2.0][MPL]. For more
-information, read the file ``LICENSE``.
+information, read the file [LICENSE](LICENSE).
```
amychurchwell: In markdown format we can link i[n a couple ways](https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#links). For this doc we can follow the format of the `[MPL]` link by creating a reference link. (Which I see you are halfway there! 😄)

On line 51 you will see: `[MPL]: https://www.mozilla.org/MPL/`
That is the link that `[MPL version 2.0][MPL]` is referencing. We can do the same for `LICENSE` by adding underneath line 51:

`[LICENSE]: https://github.com/mozilla/bedrock/blob/master/LICENSE`

amychurchwell: We will also need to update `[LICENSE](LICENSE)` to use brackets for both because we are using a reference link: 
`[LICENSE][LICENSE]`

Issue #7351, opened 2019-06-27
Update jQuery library version


pdehaan:
## Description

I ran a Lighthouse audit in Google Chrome for https://www.mozilla.org/en-US/firefox/accounts/ and saw that it identified **jquery@3.2.1** as a questionable version.

**Snyk results:**

- https://snyk.io/test/npm/jquery/3.2.1
- https://snyk.io/vuln/npm:jquery?lh=3.2.1&utm_source=lighthouse&utm_medium=ref&utm_campaign=audit

### Steps to reproduce

1. Open Google Chrome
2. Go to https://www.mozilla.org/en-US/firefox/accounts/
3. Open DevTools > Audits
4. Click [Run audits] button.
5. Scroll down to "Best Practices" section in results after scan is done.

![Get_a_Firefox_Account_–_Keep_your_data_private__safe_and_synced](https://user-images.githubusercontent.com/557895/60299338-5bfece80-98e1-11e9-8d37-c2da123630f1.png)

### Expected result

No errors or warnings.

### Actual result

![Get_a_Firefox_Account_–_Keep_your_data_private__safe_and_synced](https://user-images.githubusercontent.com/557895/60299448-98cac580-98e1-11e9-87cf-782aa3ae14a8.png)

### Environment

macOS w/ latest Google Chrome.

alexgibson:
Thanks for filing this, @pdehaan 👍

ejregithub:
@alexgibson - please apply an Estimate (just your work, not Peter's) in this card

alexgibson:
Done, thanks

PR #7495, opened 2019-07-30
Fix fraud report attachment submission (Bug 1567021)

alexgibson:
## Description
- Looks like this bug is a regression from updating jQuery in https://github.com/mozilla/bedrock/issues/7351. Rather than updating the form validation plugin, I've opted for removing the plugin entirely, as browsers today have sufficient enough form validation built-in (we still also have server-side validation and error handling in place).
- Also fixes speaker request form bug (See https://github.com/mozilla/bedrock/issues/7496).

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1567021

## Testing
- [ ] Fraud report form should submit ok when an image is attached.
- [ ] Speaker Request form should submit ok when a PDF is attached.


alexgibson:
Added a second commit that also fixes #7496

alexgibson:
@pmac r? (the python change here seems simple enough, but thought I should tag you as these pages may have some logic I'm not quite clear on).

alexgibson:
Hmm, speaker request tests are now failing with `TypeError: '>' not supported between instances of 'Mock' and 'int'`. Will try and figure out what's up.

alexgibson:
Ok, think I figured out the tests.

PR #7534, opened 2019-08-06
Use releng product-details repo directly

### LINKED ISSUE #6998: Switch to Releng's git repo for product-details data
We are currently keeping our own git repo updated for product-details data that includes some checks for data integrity. Releng is adding our checks to their update process in https://github.com/mozilla/release-services/pull/1987, so we should be able to switch to using their repo for bedrock's data update needs:

https://github.com/mozilla-releng/product-details/tree/production/public/1.0

pmac:
Instead of scraping product-details.mozilla.org we can directly use the source files for that from a git repo managed by release engineering.

Fix #6998

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -91,10 +91,6 @@ TEST_RUNNER = 'django.test.runner.DiscoverRunner'
 # http://www.i18nguy.com/unicode/language-identifiers.html
 LANGUAGE_CODE = 'en-US'
 
-# Use Ship It as the source for product_details
-PROD_DETAILS_URL = config('PROD_DETAILS_URL',
-                          default='https://product-details.mozilla.org/1.0/')
-
 # Tells the product_details module where to find our local JSON files.
 # This ultimately controls how LANGUAGES are constructed.
 PROD_DETAILS_CACHE_NAME = 'product-details'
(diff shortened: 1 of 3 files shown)


bedrock/settings/base.py
```diff
@@ -105,12 +101,10 @@ def path(*args):
...
                                      default=path('product_details_json'))
 PROD_DETAILS_JSON_REPO_URI = config('PROD_DETAILS_JSON_REPO_URI',
-                                    default='https://github.com/mozilla/product-details-json.git')
+                                    default='https://github.com/mozilla-releng/product-details.git')
+PROD_DETAILS_JSON_REPO_BRANCH = config('PROD_DETAILS_JSON_REPO_BRANCH', default='production')
 # path to updated p-d data for testing before loading into DB
-PROD_DETAILS_TEST_DIR = str(Path(PROD_DETAILS_JSON_REPO_PATH).joinpath('product-details'))
-# if the repo is cloned it will be most up-to-date
-if Path(PROD_DETAILS_TEST_DIR).is_dir():
-    PROD_DETAILS_DIR = PROD_DETAILS_TEST_DIR
```
jgmize: There are still some references to `PROD_DETAIL_DIR` in `bedrock/firefox/tests/test_firefox_details.py` and `bedrock/firefox/tests/test_base.py`

pmac: Yeah. It's a default setting in the `product_details` library. Using it in those tests is fine.

pmac:
Tests are passing and it seems to be working for me. Think we're ready to get this landed and try it out on dev?

bedrock/settings/base.py
l-hedgehog: I think with this removal it's no longer possible to use `product_details.storage.PDFileStorage` as `PROD_DETAILS_STORAGE`. The files stored in `PROD_DETAILS_TEST_DIR` could be updated by the management command, but they are never used.

pmac: I think you're right. We used to need that so that things would work without the external DB, but since we moved to the SQLite distributed db we no longer need all of this. I should do some more cleanup to remove more of these settings and just hard-code `PDDatabaseStorage`.

Issue #7532, opened 2019-08-05
Stop double-requesting Flow IDs on WNP69+


hoosteeno:
## Description

On WNP 67.0.1 and 68, we double-request Flow IDs. We do it once for the form shown to signed-out users (https://github.com/mozilla/bedrock/blob/1608cf38e0b530769c50bdfaa6db22dbcccf9829/media/js/base/mozilla-monitor-button.js) and once for the form shown to signed-in users (https://github.com/mozilla/bedrock/blob/432985279224bc492d2a0b591f25c272b2c7f456/media/js/base/mozilla-fxa-form.js).

Here's what that looks like in the network request panel:
![Screen Shot 2019-08-05 at 5 15 49 PM](https://user-images.githubusercontent.com/94519/62500672-da824200-b7a4-11e9-8ab8-9d82f3284a9f.png)

**We should not request Flow IDs when we don't show the UI for the flow they relate to.** Doing so inaccurately increments the number of views of that UI element. Specifically, in this case, when a signed-in user sees the Monitor button, Amplitude still increments the number of visitors to the FxA email-first sign-in form, even though a signed-in user could not interact with that form. 

We should only request one Flow ID per pageload. We should make this change to WNP69+, but not to WNP68* and below.

amychurchwell:
@alexgibson I recommended on the WNP69 issue that we keep `js/base/mozilla-fxa-form.js` and `js/base/mozilla-monitor-button.js` in separate JS blocks separated by some logic (if user is signed in = `mozilla-monitor-button.js`, if signed out = `mozilla-fxa-form.js`) to avoid the double request.

Do you know of a way that we can detect if the fetch request to fxa has been made already that we could include in the JS itself? (or perhaps Python?) 
It would be nice if there was a way to say: if server value `settings.FXA_ENDPOINT` has had a `fetch` request in this session, do not make a second request. Could we test for this?

alexgibson:
Fixed in https://github.com/mozilla/bedrock/pull/7588

Issue #7533, opened 2019-08-05
Retrieve deviceID from metrics-flow requests and add to downstream referrals


hoosteeno:
## Description

We make requests to the FxA servers "metrics-flow" endpoint in several places on Bedrock:

https://github.com/mozilla/bedrock/search?q=metrics-flow&unscoped_q=metrics-flow

The response to these requests should (now) include a deviceId:

https://github.com/mozilla/fxa/blob/master/packages/fxa-content-server/server/lib/routes/get-metrics-flow.js#L122

We need to add this deviceId to any downstream referrals we make:

https://github.com/mozilla/bedrock/search?q=flow_id&unscoped_q=flow_id

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Device ID is retrieved and added to downstream referral links

alexgibson:
Fixed in https://github.com/mozilla/bedrock/pull/7588

Issue #7417, opened 2019-07-11
Standardize UTM parameters on mozilla.org for FxA referrals


amychurchwell:
## Description
**Macros that utilize utm parameters:**
- [x] `fxa_email_form` [#7397]
- [x] `fxa_link_fragment`
- [x] `fxa_cta_link`
- [x] `monitor_button`

**Remember to:**
- [ ] [ document these changes.](https://bedrock.readthedocs.io/en/latest/firefox-accounts.html)
 
| utm parameters| definition | standard format | example in site |
|---|---|---|---|
|`utm_source` & `entrypoint` | Unambiguous identifier of which page of the site linked.  | `mozilla.org-directory-page` | `mozilla.org-firefox-accounts` or `whatsnew#` |
|`utm_medium`| What type of link was used to direct to the page. **this value should be `referral`** except in the case of stub attribution which expects `download_button`. | `referral` | `referral` |
|`utm_campaign`| Used to identify specific marketing campaigns. Should have default value. | **If campaign**: appends campaign name to default value which is descriptive of element. | **with campaign:** `trailhead-fxa-embedded-form` **without campaign:** `fxa-embedded-form` |
|`utm_term`| Optional.  Used for paid search keywords. | Brief keyword, this should be provided. | `existing-users` |
|`utm_content`| Optional. It should only be declared when there is new content pointing to the same place. | Brief description of variation. Or if A/B test, name of experiment. | `get-the-rest-of-firefox` |

_Sources:_
- https://mozilla.github.io/application-services/docs/accounts/metrics.html
- https://ga-dev-tools.appspot.com/campaign-url-builder/
- https://github.com/mozilla/bedrock/pull/7397#issuecomment-509692686

amychurchwell:
There has already been a lot of great analytics work done already that touches several places, but I wanted to distill our standards specifically about UTM parameters and bake-in as much as possible to the macros affected.

PR #7546, opened 2019-08-08
Add monitor_button params and retrieve deviceId from metrics-flow requests. (Fixes #7531, #7533, #7532)

### LINKED ISSUE #7531: WNP and /accounts Monitor button requests metrics-flow without any parameters
## Description

When we create FxA funnels on www.mozilla.org, we ping the FxA servers to get a Flow ID and a Flow Start Time. This allows a user journey to span domains and pages during an FxA registration. 

In order to attribute the first contact properly, we need to include a set of parameters on every request we make to the FxA servers, including the initial ping when we request a Flow ID.

Here's an example of a well-formed request for a Flow ID:

https://github.com/mozilla/bedrock/blob/432985279224bc492d2a0b591f25c272b2c7f456/media/js/base/mozilla-fxa-form.js#L54-L64

Here's the Monitor button's request for a Flow ID (on WNP and /accounts):

https://github.com/mozilla/bedrock/blob/1608cf38e0b530769c50bdfaa6db22dbcccf9829/media/js/base/mozilla-monitor-button.js#L23-L25

We need to add entrypoint, form_type, and utm parameters to the request.

Entrypoint and UTMs should follow [our uniform schema](https://github.com/mozilla/bedrock/issues/7417).

form_type should be derived from the output of https://github.com/mozilla/fxa/issues/963

amychurchwell:
## Description
1. Send additional parameters to FxA servers during metrics-flow request for monitor button.
2. Retrieve new deviceId value from FxA requests.

## Issue / Bugzilla link
#7531
#7533
#7532

## Testing
Adds to monitor button FxA metrics-flow fetch. [`js/base/mozilla-monitor-button.js`]:
- `entrypoint` 
- `form_type`
- utm parameters

_Values follow [our uniform schema](https://github.com/mozilla/bedrock/issues/7417)._

All [metrics-flow requests](https://github.com/mozilla/bedrock/search?q=flow_id&unscoped_q=flow_id) now collect `deviceId` value from FxA.

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -300,9 +300,10 @@
     <input type="hidden" name="service" value="{{ service }}" />
     <input type="hidden" name="entrypoint" value="{{ entrypoint }}" id="fxa-email-form-entrypoint" />
     <input type="hidden" name="form_type" value="email" />
-    {# flow_id and flow_begin_time will be populated via JS on page load #}
+    {# device_id, flow_id and flow_begin_time will be populated via JS on page load #}
     <input type="hidden" name="flow_id" value="" />
     <input type="hidden" name="flow_begin_time" value="" />
+    <input type="hidden" name="device_id" value="" />
 
     {# utm_source is required #}
     <input type="hidden" name="utm_source" value={{ utm_source }} id="fxa-email-form-utm-source" />
@@ -354,12 +355,12 @@
         {% endtrans %}
       </p>
 
-      <button 
-        type="submit" 
-        class="{{ button_class }}" 
-        id="fxa-email-form-submit" 
-        data-cta-type="FxA-Sync" 
-        data-cta-text="Register" 
+      <button
+        type="submit"
+        class="{{ button_class }}"
+        id="fxa-email-form-submit"
+        data-cta-type="FxA-Sync"
+        data-cta-text="Register"
         data-cta-position="Primary"
       >
       {% if button_text %}
--- a/media/js/base/mozilla-monitor-button.js
+++ b/media/js/base/mozilla-monitor-button.js
@@ -26,7 +26,9 @@ Mozilla.MonitorButton = (function () {
     var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
 
     // collect values from monitor button
-    var utmParams = _SearchParams.queryStringToObject(buttonURL);
+    var utmParams = window._SearchParams.queryStringToObject(buttonURL);
+
+    console.log(utmParams);
 
     var utmSource = utmParams.utm_source;
     var utmCampaign = utmParams.utm_campaign;
--- a/media/js/base/mozilla-monitor-button.js
+++ b/media/js/base/mozilla-monitor-button.js
@@ -38,11 +38,11 @@ if (typeof window.Mozilla === 'undefined') {
         var params = window._SearchParams.queryStringToObject(buttonURLParams);
 
         // add required params to the token fetch request
-        destURL += '&entrypoint=' + params.entrypoint;
+        destURL += '?entrypoint=' + params.entrypoint;
         destURL += '&form_type=' + params.form_type;
-        destURL += '&utm_campaign=' + params.campaign;
-        destURL += '&utm_content=' + params.content;
-        destURL += '&utm_source=' + params.source;
+        destURL += '&utm_campaign=' + params.utm_campaign;
+        destURL += '&utm_content=' + params.utm_content;
+        destURL += '&utm_source=' + params.utm_source;
 
         fetch(destURL).then(function(resp) {
             return resp.json();
--- a/media/js/base/mozilla-monitor-button.js
+++ b/media/js/base/mozilla-monitor-button.js
@@ -7,40 +7,56 @@ if (typeof window.Mozilla === 'undefined') {
     window.Mozilla = {};
 }
 
-Mozilla.MonitorButton = (function() {
+(function() {
     'use strict';
 
-    var monitorButton = document.getElementById('fxa-monitor-submit');
+    var MonitorButton = {};
 
-    // fetch request
-    var supportsFetch = 'fetch' in window;
+    MonitorButton.init = function(buttonId){
 
-    if (!supportsFetch || !monitorButton) {
-        return;
-    }
+        // check for custom buttonId, used in case of multiple buttons.
+        if(!buttonId){
+            buttonId = 'fxa-monitor-submit';
+        }
 
-    var buttonURL = monitorButton.getAttribute('href');
-    // strip url to everything after `?`
-    var buttonURLParams = buttonURL.match(/\?(.*)/)[1];
+        var monitorButton = document.getElementById(buttonId);
 
-    var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
+        // fetch request
+        var supportsFetch = 'fetch' in window;
 
-    // collect values from monitor button
-    var params = window._SearchParams.queryStringToObject(buttonURLParams);
+        if (!supportsFetch || !monitorButton) {
+            return;
+        }
 
-    // add required params to the token fetch request
-    destURL += '&' + params;
+        var buttonURL = monitorButton.getAttribute('href');
+        // strip url to everything after `?`
+        var buttonURLParams = buttonURL.match(/\?(.*)/)[1];
 
-    fetch(destURL).then(function(resp) {
-        return resp.json();
-    }).then(function(r) {
-        // add retrieved deviceID, flowBeginTime and flowId values to cta url
-        buttonURL += '&deviceId=' + r.deviceId;
-        buttonURL += '&flowBeginTime=' + r.flowBeginTime;
-        buttonURL += '&flowId=' + r.flowId;
-        monitorButton.setAttribute('href', buttonURL);
-    }).catch(function() {
-        // silently fail: deviceId, flowBeginTime, flowId are not added to url.
-    });
+        var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
+
+        // collect values from monitor button
+        var params = window._SearchParams.queryStringToObject(buttonURLParams);
+
+        // add required params to the token fetch request
+        destURL += '&entrypoint=' + params.entrypoint;
+        destURL += '&form_type=' + params.form_type;
+        destURL += '&utm_campaign=' + params.campaign;
+        destURL += '&utm_content=' + params.content;
+        destURL += '&utm_source=' + params.source;
+
+        fetch(destURL).then(function(resp) {
+            return resp.json();
+        }).then(function(r) {
+            // add retrieved deviceID, flowBeginTime and flowId values to cta url
+            buttonURL += '&deviceId=' + r.deviceId;
+            buttonURL += '&flowBeginTime=' + r.flowBeginTime;
+            buttonURL += '&flowId=' + r.flowId;
+            monitorButton.setAttribute('href', buttonURL);
+        }).catch(function() {
+            // silently fail: deviceId, flowBeginTime, flowId are not added to url.
+        });
+    };
+
+    window.Mozilla.MonitorButton = MonitorButton;
 })();
 
--- a/media/js/base/mozilla-monitor-button.js
+++ b/media/js/base/mozilla-monitor-button.js
@@ -38,10 +38,16 @@ if (typeof window.Mozilla === 'undefined') {
         var params = window._SearchParams.queryStringToObject(buttonURLParams);
 
         // add required params to the token fetch request
+        if (params.utm_content) {
+            destURL += '&utm_content=' + params.utm_content;
+        }
+
+        if (params.utm_campaign) {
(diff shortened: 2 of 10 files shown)


media/js/base/mozilla-monitor-button.js
```diff
@@ -22,15 +22,32 @@ Mozilla.MonitorButton = (function() {
     var buttonURL = monitorButton.getAttribute('href');
     var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
 
+    // collect values from monitor button
+    var utmSource = buttonURL.match(/(utm_source=.*?)&/)[1].toString();
```
alexgibson: We have a [small helper library](https://github.com/mozilla/bedrock/blob/master/media/js/base/search-params.js#L30-L31) in the base bundle that makes this kind of task much easier to do without needing custom code. Might that make this easier to avoid needing to keep the params in a certain order?

bedrock/base/templates/macros.html
```diff
@@ -255,6 +255,11 @@ <h3>{{ _('Your download link was sent.') }}</h3>
   </a>
 {%- endmacro %}
 
+<!--
+  monitor_button:
+  - entrypoint must remain last parameter on url for regex in metrics flow js.
+  - form_type value will change on output of https://github.com/mozilla/fxa/issues/963
```
alexgibson: Is this something we need to call out in the code, or should we file an issue in GitHub to keep track of it. It seems there is nothing to do right now, so could this get lost / forgotten?

bedrock/base/templates/macros.html
```diff
@@ -255,6 +255,11 @@ <h3>{{ _('Your download link was sent.') }}</h3>
   </a>
 {%- endmacro %}
 
+<!--
+  monitor_button:
+  - entrypoint must remain last parameter on url for regex in metrics flow js.
```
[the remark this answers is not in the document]
amychurchwell: **Most definitely**, but I am struggling getting the regex logic right for this. Going to continue working on this today.

bedrock/base/templates/macros.html
amychurchwell: Agreed! Filed issue: https://github.com/mozilla/bedrock/issues/7551

media/js/base/mozilla-monitor-button.js
amychurchwell: Having similar issue with this function because the query is being split by `&`'s

https://github.com/mozilla/bedrock/blob/2ff934c938ca23ba380999655ec8ad2165132b18/media/js/base/search-params.js#L34

so the `utm_source` key in this object is the remainder of the url because it begins with `?`. :( 
Screen Shot 2019-08-09 at 2 05 29 PM

Going to try and tweak this function to account for this.

media/js/base/mozilla-monitor-button.js
```diff
@@ -20,17 +20,41 @@ Mozilla.MonitorButton = (function() {
...
 
     var buttonURL = monitorButton.getAttribute('href');
+    // strip url to everything after `?`
+    buttonURL = buttonURL.match(/\?(.*)/)[1];
+
     var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
 
-    fetch(destURL).then(function(resp) {
+    // collect values from monitor button
+    var utmParams = window._SearchParams.queryStringToObject(buttonURL);
```
alexgibson: Nit: you're getting all types of params here, not just utms, so suggest renaming this to `params`?

media/js/base/mozilla-monitor-button.js
```diff
@@ -20,17 +20,41 @@ Mozilla.MonitorButton = (function() {
...
+
+    console.log(utmParams);
+
+    var utmSource = utmParams.utm_source;
```
alexgibson: Nit: these values only seem to be used once, so are they worth storing as variables? Could this be simplified to:

```javascript
destURL += '&' + params.utm_source;
```

media/js/firefox/concerts.js
```diff
@@ -13,7 +13,7 @@ function onYouTubeIframeAPIReady() {
 }
 
 
-(function(Mozilla) {
+(function (Mozilla) {
```
alexgibson: Are these stylistic changes, or are you having issues with your editor's ESLint config? In terms of bedrock, we don't have any rules set about whitespace changes like this (afaik)?

media/js/base/mozilla-monitor-button.js
```diff
@@ -20,17 +20,41 @@ Mozilla.MonitorButton = (function() {
...
+    // strip url to everything after `?`
+    buttonURL = buttonURL.match(/\?(.*)/)[1];
+
     var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
 
-    fetch(destURL).then(function(resp) {
+    // collect values from monitor button
+    var utmParams = window._SearchParams.queryStringToObject(buttonURL);
+
+    console.log(utmParams);
```
alexgibson: Remove `console.log()`

media/js/base/mozilla-monitor-button.js
```diff
@@ -20,17 +20,41 @@ Mozilla.MonitorButton = (function() {
     }
 
     var buttonURL = monitorButton.getAttribute('href');
+    // strip url to everything after `?`
+    buttonURL = buttonURL.match(/\?(.*)/)[1];
```
alexgibson: Suggest storing the regex match in a variable here as opposed to setting `buttonURL`. Further down on line 55 we have:

```javascript
monitorButton.setAttribute('href', buttonURL);
```

So currently the monitor button is being turned into a relative link, vs sending people to monitor.firefox.com.

media/js/base/mozilla-monitor-button.js
amychurchwell: whoops 😅

media/js/firefox/concerts.js
amychurchwell: These changes were made when I selected `Format Document` in vscode. Seems to be the default formatting.
I added `"javascript.format.insertSpaceBeforeFunctionParenthesis": false` to my `settings.json` file in my editor so hopefully that will prevent the difference in the future.
I also added `"eslint.autoFixOnSave": true` to ensure eslint is being applied on save. :)

media/js/base/mozilla-monitor-button.js
```diff
@@ -20,17 +20,27 @@ Mozilla.MonitorButton = (function() {
...
+    var buttonURLParams = buttonURL.match(/\?(.*)/)[1];
+
     var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
 
-    fetch(destURL).then(function(resp) {
+    // collect values from monitor button
+    var params = window._SearchParams.queryStringToObject(buttonURLParams);
+
+    // add required params to the token fetch request
+    destURL += '&' + params.utm_source;
```
alexgibson: Don't you still need to do this for the other params in the issue as well?

media/js/firefox/concerts.js
alexgibson: I'm not quite sure what kind of rules VSCode uses with the `Format Document` command, but I'd recommend to try installing [this ESLint plugin](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint). Once installed it should use the version of ESLint that's found locally in the `node_modules` folder within the workspace, and will also use the rules defined in our root `.eslintrc`. The tool will also lint as you type, which is pretty cool :)

media/js/base/mozilla-monitor-button.js
amychurchwell: ah jeeze. 🤦‍♀ whoops.

media/js/firefox/concerts.js
amychurchwell: Totally! I have the ESLint plugin, I just made the mistake of using `Format Document` which overrode with it's own formatting opinions. 😅 I couldn't find a way to correct with ESlint because our `.eslintrc` file doesn't appear to have opinions about spacing.

I found that if I added to `settings.json` both: 
`"javascript.format.insertSpaceBeforeFunctionParenthesis": false` and `"javascript.format.insertSpaceAfterFunctionKeywordForAnonymousFunctions": false` it reverted back to our preferred spacing format.

media/js/base/mozilla-monitor-button.js
```diff
@@ -20,17 +20,27 @@ Mozilla.MonitorButton = (function() {
...
+    // strip url to everything after `?`
+    var buttonURLParams = buttonURL.match(/\?(.*)/)[1];
+
     var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
 
+    // collect values from monitor button
+    var params = window._SearchParams.queryStringToObject(buttonURLParams);
+
+    // add required params to the token fetch request
+    destURL += '&' + params;
```
alexgibson: This is passing an object to the metrics flow, so the request comes out like `/metrics-flow&[object%20Object]`, which results in a 404.

I'd also suggest we only pass the specific params in the issue description, vs everything a URL query string may contain?

amychurchwell:
Included logic in `account-conditional-ctas.js` that should resolve issue #7532 as well.
Refactored so that if there are multiple monitor buttons on a page you can pass an optional `buttonId` argument to `Mozilla.MonitorButton.init()` to specify which one you would like to initialize.

media/js/base/mozilla-monitor-button.js
```diff
@@ -7,30 +7,56 @@ if (typeof window.Mozilla === 'undefined') {
...
+        // strip url to everything after `?`
+        var buttonURLParams = buttonURL.match(/\?(.*)/)[1];
+
+        var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
+
+        // collect values from monitor button
+        var params = window._SearchParams.queryStringToObject(buttonURLParams);
+
+        // add required params to the token fetch request
+        destURL += '&entrypoint=' + params.entrypoint;
```
alexgibson: The first parameter in a query string should start with a `?` instead of ` &` (this is causing the metric flow request to throw the 404).

media/js/base/mozilla-monitor-button.js
```diff
@@ -7,30 +7,56 @@ if (typeof window.Mozilla === 'undefined') {
...
+        destURL += '&form_type=' + params.form_type;
+        destURL += '&utm_campaign=' + params.campaign;
```
alexgibson: This value is coming through as `undefined`. Should this be `params.utm_campaign`? Same for the other utm params below.

media/static-bundles.json
```diff
@@ -1398,7 +1400,8 @@
         "js/base/mozilla-fxa-init.js",
         "js/base/mozilla-fxa-form.js",
         "js/base/mozilla-monitor-button.js",
-        "js/firefox/accounts-2019.js"
+        "js/firefox/accounts-2019.js",
+        "js/firefox/whatsnew/account-conditional-ctas.js"
```
alexgibson: This script is fine for /whatsnew, but I think it may be worth creating a separate piece of code to run on /accounts? The reason is that on /accounts we want to call `Mozilla.FxaForm.init();` even on non-Firefox browsers, and the `getFxaDetails();` function doesn't really need to be called in that scenario.

I'd suggest adding some code to `accounts-2019.js` for this:

```javascript
if (Mozilla.Client.isFirefoxDesktop) {
    Mozilla.Client.getFxaDetails(function(details) {
        if (details.setup) {
            Mozilla.MonitorButton.init();
        } else {
            Mozilla.FxaForm.init();
        }
    });
} else {
    Mozilla.FxaForm.init();
}

File: media/js/base/mozilla-monitor-button.js [hunk 14]
```diff
@@ -7,30 +7,56 @@ if (typeof window.Mozilla === 'undefined') {
...
+        var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
+
+        // collect values from monitor button
+        var params = window._SearchParams.queryStringToObject(buttonURLParams);
+
+        // add required params to the token fetch request
+        destURL += '?entrypoint=' + params.entrypoint;
+        destURL += '&form_type=' + params.form_type;
+        destURL += '&utm_campaign=' + params.utm_campaign;
+        destURL += '&utm_content=' + params.utm_content;
```
alexgibson: `utm_content` is coming through as `NaN` here, since it's not present in the monitor button URL. Because of this it surfaces as an empty parameter value in the final URL e.g. `utm_source=mozilla.org-firefox-accounts&utm_campaign=trailhead&utm_content=&form_type=email&`.

Because only `utm_source` is a required parameter, we should probably make both `utm_campaign` and `utm_content` optional here? e.g.

```javascript
if (params.utm_content) {
   destURL += '&utm_content=' + params.utm_content;
}
```

media/static-bundles.json
```diff
@@ -1397,6 +1405,7 @@
         "js/base/mozilla-fxa.js",
         "js/base/mozilla-fxa-init.js",
         "js/base/mozilla-fxa-form.js",
+        "js/base/mozilla-fxa-form-init.js",
```
alexgibson: This is initializing the form twice I think?

alexgibson:
Closing in favor of https://github.com/mozilla/bedrock/pull/7588

PR #7554, opened 2019-08-09
Browser Support Page added

### LINKED ISSUE #7389: Re-evaluate browser support on bedrock.
## Description

The front-end team spend a considerable amount of time supporting older browsers, such as IE9 & IE10. We should evaluate how much traffic we still get from these old browsers, and whether or not the time invested coding for them is still a worthwhile investment. 

Can we consider dropping first-class support for IE9 or IE10, and instead giving them the universal stylesheet we serve to IE8? Doing so may have the following advantages:

- It enables us to ship code faster.
- It enables us to use more advanced browser features.
- It reduces the number of polyfills we need to bundle.

Bedrock's browser support matrix isn't formally documented anywhere (yet), but here's the current snapshot:

- We deliver 1st class CSS & JS to:
  - All evergreen browsers (Firefox, Chrome, Safari, Edge, Opera etc).
  - IE 9, 10 & 11.
- We deliver degraded * support to:
  - IE 8 and below.

* Degraded support consists of no page-specific CSS or JS. Instead, we deliver basic semantic HTML, a universal CSS stylesheet that gets applied to all pages, and a universal JS bundle that only handles downloading Firefox (click a button, get a file), and Google Analytics. The only pages where we make an exception is for `/firefox/new/` and `/firefox/download/thanks/`, which still deliver 1st class CSS to all browsers (but not 1st class JS).

---

- [x] A recommendation is made on which (if any) browsers we can drop 1st class support for.
- [x] The decision is documented.

- [ ] Dropping 1st class support for a browser can potentially impact download conversions.

#### Tasks
- [x] Evaluate IE9 and IE10 traffic. How do they compare to IE8, or IE11 levels of traffic?
- [x] Make a recommendation for which browsers we still need to deliver 1st class support for.
- [x] Add a "Browser support" section to the bedrock docs to document the decision.

hoosteeno:
Fixes #7389

### DIFF
--- a/docs/browser-support.rst
+++ b/docs/browser-support.rst
@@ -1,0 +1,52 @@
+.. This Source Code Form is subject to the terms of the Mozilla Public
+.. License, v. 2.0. If a copy of the MPL was not distributed with this
+.. file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+.. _browser_support:
+
+=================
+Browser Support
+=================
+
+We seek to provide usable experiences of our most important web content to all user agents. But newer browsers are far more capable than older browsers, and the capabilities they provide are valuable to developers and site visitors. We **will** take advantage of modern browser capabilities. Older browsers **will** have a different experience of the website than newer browsers. We will strike this balance by generally adhering to the core principles of `Progressive Enhancement <https://en.wikipedia.org/wiki/Progressive_enhancement>`_::
+
+    * Basic content should be accessible to all web browsers
+    * Basic functionality should be accessible to all web browsers
+    * Sparse, semantic markup contains all content
+    * Enhanced layout is provided by externally linked CSS
+    * Enhanced behavior is provided by unobtrusive, externally linked JavaScript
+    * End-user web browser preferences are respected
+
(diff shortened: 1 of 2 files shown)


docs/browser-support
```diff
@@ -0,0 +1,49 @@
...
+
+Basic support consists of no page-specific CSS or JS. Instead, we deliver basic semantic HTML, a universal CSS stylesheet that gets applied to all pages, and a universal JS bundle that only handles downloading Firefox (click a button, get a file), and Google Analytics. 
+
+Browser Support Matrix (Updated 20190809)
+-------------------
+
+**The following browsers have enhanced support:**
+
+  * All evergreen browsers (Firefox, Chrome, Safari, Edge, Opera, etc.)
+  * IE11.
```
alexgibson: We're going to have to make some incremental changes before we can drop support for IE9 & 10, so until that happens I suggest we reflect the current landscape and change this to `IE9 and above`.

docs/browser-support
```diff
@@ -0,0 +1,49 @@
...
+
+**The following browsers have basic support:**
+
+  * All other user agents.
```
alexgibson: Nit: suggest changing this to `All other IE browsers.`, since that is the only user agent to support conditional comments, which is the mechanism we currently use to deliver basic support styles.

PR #7563, opened 2019-08-12
Update /contribute survey (Fixes #7553)

### LINKED ISSUE #7553: Update language + increase visibility of /contribute survey
## Description

The current survey has a low 0.03% conversion rate. There is a similar CTA on support.mozilla.org where the banner is at the top and the wording was proposed by the research team, and they have a conversion rate of 1.5%. 

Suggestions: 

* Update Wording - "We would love to get your opinion on your experience with our site with a short survey [Take Survey]"
* Move Banner - So it appears at the top of the page. 

Any other ideas? 

cc @johngian

alexgibson:
## Description
- Moves survey to the top of the viewport.
- Updates copy.
- Increase survey message font size
- Use a background color from our brand guidelines.

## Issue / Bugzilla link
#7553

## Testing
- [x] Survey should appear at the top of the viewport.
- [x] Survey should be sticky on wide viewports only.


alexgibson:
@johngian r?

media/js/mozorg/contribute/contribute-survey.js
```diff
@@ -2,15 +2,8 @@
...
-        var $survey = $('#survey-message');
-        $survey.addClass('show');
-        setTimeout(function() {
-            $survey.addClass('animate');
-        }, 500);
-    });
-
-})(window.jQuery);
\ No newline at end of file
+    document.getElementById('survey-message').classList.add('show');
```
stephaniehobson: I'm guessing by looking at the old code that this was only hidden initially so it could be animated in. With the animation gone I think we can just show the survey by default even with JS disabled :)

alexgibson: Yup you're right this is left-over, removing 👍

PR #7595, opened 2019-08-16
Fundraiser banner

Errolyn:


media/js/mozorg/home/fundraiser.js
```diff
@@ -0,0 +1,38 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+(function(Mozilla) {
+    'use strict';
+
+    var fundraiser = document.getElementById('fundraiser');
+    var fundraiserClose = document.getElementById('fundraiser-close');
+    var cookieDurationDays = 1;
+    var cookiesOK = typeof Mozilla.Cookies !== 'undefined' && Mozilla.Cookies.enabled();
+    var storageKey = 'fundraiser-jun2019';
```
stephaniehobson: For the sake of being through I'd probably update this. But the code works just fine as-is and the cookies from the last campaign will have already expired so not a blocker.

Errolyn: updated :)

Errolyn:
This should be ready for another pass :D

media/css/mozorg/home/fundraiser.scss
```diff
@@ -104,7 +103,7 @@ content {
 
 .c-fundraiser-options {
     label {
-        @include bidi(((margin, 0 $spacing-sm 0 0, 0 0 0 $spacing-sm),));
+        @include bidi(((margin, $spacing-sm $spacing-sm 0 0, 0 0 0 $spacing-sm),));
```
stephaniehobson: RTL will need the top margin too.

Errolyn: 🤦‍♀

craigcook:
We usually control these banners with a switch so we can enable/disable it without doing a full production push (https://bedrock.readthedocs.io/en/latest/install.html#feature-flipping-aka-switches). Should this one be the same?

Errolyn:
Yes there should be, I missed adding it back in, I'll add that as well.

Issue #7552, opened 2019-08-09
(11/4) - Display page content specific to users physically located in India


ejregithub:
## Description

The Taipei team wants to leverage the WNP to show content promoting Firefox Lite (which will be rebranded) to users located in India who have updated their Firefox desktop browser to the latest version at 11/4

Issue #6084 has described how this may be accomplished by parsing information out of request headers to then create a URL or convert an existing URL that will look at the country code and redirects the user to a URL based on the country, such as /en-US/firefox/whatsnew/?country=IN , in this particular case where users in India are the target audience.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Country (India) specific content is shown to users when updating their Firefox desktop browser 

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] This is an example of a risk. Use a checklist!
- [ ] This is an example of a risk.

#### Tasks
- [x] Obtain content, copy and design assets, from the Taipei team - MVP is copy and images (QR code)
- [ ] Set up URL for users located in India
- [ ] Configure forwarding via header content identifying country
- [x] Provide assets to Silverorange to code
- [ ] QA (team and stakeholder) of coded page on demo
- [ ] Code review
- [ ] Merge, launch on Nov 4 (India Standard Time Zone)

alexgibson:
It’s worth noting here that the functionality described in #6084 does not yet exist, and is likely either non-trivial or requires a level of engineering effort tbd.

hoosteeno:
Regarding feasibility, [Slack conversation](https://mozilla.slack.com/archives/G7ACJHJCB/p1565371997480400) makes it sound like this is a problem we can solve using a solution like the one described in #6084 without risking site stability. I will explore priority.

alexgibson:
Note: if we're going to do this for FF 70 then we'll need an English version of the Firefox Lite whatsnew page: https://www.mozilla.org/id/firefox/whatsnew/

StellaTseng:
Yes, I will provide the English version of the Firefox Lite whatsnew page for India users.
Like what we did before for Indonesia users (but this time we will use QRcode instead of email): https://www.mozilla.org/id/firefox/whatsnew/ 
We are working on the assets and target to share it in the end of September (around 9/25). 
Is that okay for you to meet the FF 70 launch? Thank you!!

alexgibson:
In order to answer this I need to fully understand the ask here:

[...]

PR #7627, opened 2019-08-27
Add generic Geo redirector view class

### LINKED ISSUE #6084: Add geolocation-based redirector view class and documentation
We typically do geolocation on bedrock on the client side by requesting the user's country code at our [country-code.json](https://www.mozilla.org/country-code.json) endpoint. This works by parsing the information out of the request headers, as our CDN has already used IP-based geolocation to gather that info and has sent it to us.

Another use of that header for us could be to create a URL (or convert an existing URL) that is purely a redirector (i.e. it only returns 302 responses) that looks at the country code and then sends the user to one of potentially several URLs based on the country. This needs to be a simple redirector view because it has to be uncached by the CDN and so all traffic to the URL will hit bedrock, and so we don't want to spend time rendering templates or hitting the database. We simply lookup the proper URL and return a redirect response, and then we let the normal caching happen at the resulting final destination URL.

This might be the desired way to handle geolocation since it keeps the front-end code more simple (no async requests), it makes the UX better and less complex (no content swapping), and potentially makes the page feel faster (time from first paint to last should in theory be shorter).

This issue is to document how this might most easily be done and to add helper classes to make this easy. It should work something like this in the end:

```python
# views.py

class FirefoxGeoView(GeolocationView):
    geolocation = {
        'de': 'firefox.germany',
        'fr': 'firefox.france',
        'default': 'firefox.index',
    }
```

That view would then redirect users with "de" country code to the view with the name `firefox.germany` for example.

pmac:
This is my proposal so far. I've not yet written docs yet because I'd like some feedback on this approach first. The idea is basically that in your `views.py` file you'd import this new class, create a subclass for your redirector, and then add a URL for that. It would look like this:

```python
from bedrock.base.views import GeoRedirectView

class WhatsNewRedirectorView(GeoRedirectView):
    geo_urls = {
        'US': 'firefox.whatsnew-us',
        'IN': 'firefox.whatsnew-in',
    }
    default_url = 'firefox.whatsnew'
```

This would redirect people in the US to the URL named `firefox.whatsnew-us` and in India to `firefox.whatsnew-in` and everyone else to `firefox.whatsnew`. Then you just add this view to the `urls.py` at a URL and the others as well and we're good to go. You can also use full URLs and not just URL names.

This will fix #6084 

What do you think @alexgibson ?

### DIFF
--- a/docs/coding.rst
+++ b/docs/coding.rst
@@ -15,11 +15,15 @@ URL patterns should be as strict as possible. It should begin with a
 `^` and end with `/$` to make sure it only matches what you specifiy.
 It also forces a trailing slash. You should also give the URL a name
 so that other pages can reference it instead of hardcoding the URL.
-Example::
+Example:
+
+.. code-block:: python
 
     url(r'^channel/$', channel, name='mozorg.channel')
 
-Bedrock comes with a handy shortcut to automate all of this::
+Bedrock comes with a handy shortcut to automate all of this:
+
+.. code-block:: python
 
     from bedrock.mozorg.util import page
     page('channel', 'mozorg/channel.html')
@@ -27,6 +31,8 @@ Bedrock comes with a handy shortcut to automate all of this::
 You don't even need to create a view. It will serve up the specified
 template at the given URL (the first parameter). You can also pass
 template data as keyword arguments:
+
+.. code-block:: python
 
     page('channel', 'mozorg/channel.html',
          latest_version=product_details.firefox_versions['LATEST_FIREFOX_VERSION'])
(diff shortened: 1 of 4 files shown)


alexgibson:
@pmac I was able to take this commit and then hack together a quick demo example for how it might handle something like https://github.com/mozilla/bedrock/issues/7552

https://github.com/alexgibson/bedrock/commit/520514ac7c48a1367ddde45e7e6c05c02d602e20

Seems to be working with only minimal changes to our (somewhat overly complex) whatsnew logic. We may want to consider porting all our current whatsnew URLs to whatever we decide the default (non-geo specific) URL should be, but that doesn't need to happen in this PR.

Docs would be super helpful. Nice work 👍

pmac:
Thanks @alexgibson ! Glad it's working! Does it properly maintain the data in the URL for your specific case? i.e. does it maintain the version number in the URL? I thought it should, but that bit is a little overly magic.

alexgibson:
> Does it properly maintain the data in the URL for your specific case? i.e. does it maintain the version number in the URL? 

It seems to yes (which feels like functionality we'd need to maintain as far as /whatsnew goes).

pmac:
Yeah. The generic `RedirectView` class does it for us by passing to the reverse call of the redirect URLs the same kwargs it got from the request. If we need fancier options in future for this we can look at it, but for now this seems to be working perfectly.

pmac:
The relevant Django code for reference:

https://github.com/django/django/blob/stable/2.2.x/django/views/generic/base.py#L178

pmac:
Docs are done and several more things are improved. I think this is ready for r?

lib/l10n_utils/__init__.py
```diff
@@ -175,3 +176,7 @@ def get_context_data(self, **kwargs):
     def render_to_response(self, context, **response_kwargs):
         return render(self.request, self.get_template_names(),
                       context, **response_kwargs)
+
+
+class L10nTemplateView(LangFilesMixin, TemplateView):
```
pmac: I added this as a convenience as it should really already have existed and I wanted the docs to be easier to understand as I was reworking them.

lib/l10n_utils/__init__.py
```diff
@@ -164,7 +165,7 @@ class LangFilesMixin:
     add_active_locales = None
 
     def get_context_data(self, **kwargs):
-        ctx = super(LangFilesMixin, self).get_context_data(**kwargs)
+        ctx = super().get_context_data(**kwargs)
```
pmac: Just a little left over Python 2 -> 3 cleanup. They work the same but the new way is cleaner IMHO.

docs/coding.rst
```diff
@@ -216,6 +264,38 @@ valid variation were given in the URL.
...
+    class CanadaIsSpecialView(GeoRedirectView):
+        geo_urls = {
+            'CA': 'app.canada-is-special',
+        }
+        default_url = 'app.everyone-else'
+
+In this example people in Canada would go to the URL that Django returns using `reverse()`
+(i.e. the name of the URL) and everyone else would go to the `app.everyone-else` URL. You
+may also use full URLs instead of URL names if you want to. It will look for strings that
+start with "http(s)://" and use it as is. The country code must be 2 characters and upper case.
```
alexgibson: Is it worth linking `country code` here to a full list of valid codes?

pmac: Good idea. I linked it to the official ISO 3166-1 alpha-2 list on wikipedia.

Issue #7585, opened 2019-08-15
Pass through utm parameters on FxA referrals


hoosteeno:
## Description

In https://github.com/mozilla/bedrock/issues/6854 we built logic to pass through utm parameters from upstream sources on some referrals to FxA, but we didn't add that capability to the email-first form and monitor button.

We should pass through upstream utm parameters to downstream referrals in the FxA ecosystem. Solving this bug for the general case -- e.g. by including the functionality in the macros that create FxA referrals -- would be ideal.

### Success Criteria
- [x] Visiting https://www.mozilla.org/en-US/firefox/accounts/?utm_source=test&utm_campaign=test&utm_term=test&utm_medium=test&utm_content=test in a signed-in Firefox will add the utm parameters to the URL on the monitor button.
- [ ] Visiting https://www.mozilla.org/en-US/firefox/accounts/?utm_source=test&utm_campaign=test&utm_term=test&utm_medium=test&utm_content=test ini a signed-out Firefox will add the utm parameters to the URL on the form.
- [x] Visits with any number of upstream utm parameters will prevent us from adding all of our own (https://github.com/mozilla/bedrock/issues/7321) -- we should not mix utm parameters from upstream sources with our own.
- [x] The page default entrypoint and all other required FxA parameters will be added in any case -- we should not change the entrypoint due to an upstream factor.

alexgibson:
Depends on https://github.com/mozilla/bedrock/issues/7321

PR #7638, opened 2019-09-03
If passing through UTMs to FxA, do not fall back to our default values (Fixes #7321)

### LINKED ISSUE #7321: If passing through UTMs to FxA, do not fall back to our default values
After the work on #6854 if a user arrives on mozilla.org with UTM parameters we pass those along to FxA if they continue on to there.

The current code works like this:
1) put our own UTMs on
2) if there are UTMs in the URL replace ours with theirs
If the URL has 3/4 possible UTMs we leave the 4th value unchanged.

In https://github.com/mozilla/bedrock/issues/6854#issuecomment-500951475 Justin clarified that we should remove our own when adding others.

> If utm_* are present, let's not add any of our own. We do add missing ones on desktop downloads because all 4 (source/campaign/medium/content) are required by stub attribution, but there is no such requirement in FxA, and adding them just muddies the water.

Tasks
- [x] update the JS to remove our UTMs if other UTMs are present
- [x] update the tests
- [x] update the docs

alexgibson:
## Description
- Updates `fxa-utm-referral.js` so that we no longer muddle together UTM params from different campaigns when passing refferal data.

## Issue / Bugzilla link
#7321

## Testing
- [ ] Visiting http://127.0.0.1:8000/en-US/?utm_source=source-one&utm_campaign=campaign-one should see only the page URL's UTM params passed through to the FxA navigation button link.


alexgibson:
>  should we do something similar for the monitor button?
> it seems like this code should always run before mozilla-fxa-form-init.js by this being in the common bundle does that ensure that?

Both of those things are not yet touched by this code. I suggest we save that for https://github.com/mozilla/bedrock/issues/7585 as opposed to add more scope here.

alexgibson:
@amychurchwell thanks for the review, updated.

PR #7639, opened 2019-09-03
Do not let data update failure stop the Docker build

pmac:
This change should still allow us to get DMS notifications
if the data fails to update during normal operation but will
avoid breaking the docker image build if some data fetch fails.


alexgibson:
Thanks @pmac 👍

PR #7660, opened 2019-09-04
Add structured_data template block (#7653)

### LINKED ISSUE #7653: Add structured data block to templates
To facilitate adding structured data to all pages, we need to have a template block where we can put the includes.

This is a follow up to the successful test in #7338.

stephaniehobson:
## Description

- add structured_data template block
- add `<script>` wrapper for template block that only appears if there is an include
- convert homepage include to `.json` file

## Issue / Bugzilla link

Fix #7653

## Testing

- [x] json-ld is still included on EN homepage
- [x] EN homepage includes `<script>` wrapper around data
- [x] json-ld not on FR homepage
- [x] FR homepage does not include empty `<script>` tags


bedrock/base/templates/base-pebbles.html
```diff
@@ -127,6 +127,14 @@
     <!--[if !lte IE 8]><!-->
     {% block js %}{% endblock %}
 
+    {% if self.structured_data()|trim|length or self.structured_data()|trim|length %}
```
alexgibson: Hmm, what's the need for repeating `self.structured_data()|trim|length` using an `or`?

Also, should this be repeated in each base template for consistency?

stephaniehobson: Oh, heh, that's probably a copy pasta error.

`base-article`, `base-protocol`, and `base-protocol-mozilla` all inherit from this template. But `base-resp` does not and I made a bad assumption that it did, so thank you.

Issue #7654, opened 2019-09-04
Add structured data to Firefox product pages


stephaniehobson:
- [x] add JSON to bedrock
  - [x] update JSON with logo references
- [x] add structured data block to templates (#7653)
- [x] remove micro data from /firefox and /firefox/new and /firefox/mac, /firefox/windows, /firefox/linux (#7658)
- [x] add JSON to individual pages as appropriate

stephaniehobson:
See also https://github.com/mozilla/bedrock/issues/5741

PR #7664, opened 2019-09-04
Add json-ld strucured data

stephaniehobson:
## Description

- adds JSON-LD structured data
- adds JSON linting

## Issue / Bugzilla link

See #7654

### DIFF
--- a/bedrock/base/templates/includes/structured-data/organization/mozilla-corporation-organization.json
+++ b/bedrock/base/templates/includes/structured-data/organization/mozilla-corporation-organization.json
@@ -1,0 +1,61 @@
+{
+    "@context": "http://schema.org",
+    "@type": "Organization",
+    "@id": "https://www.mozilla.org/#organization",
+    "name": "Mozilla",
+    "alternateName": "MoCo",
+    "legalName": "Mozilla Corporation",
+    "description": "Mozilla makes browsers, apps, code and tools that put people before profit. Our mission: Keep the internet open and accessible to all.",
+    "url": "https://www.mozilla.org",
+    "logo": "https://www.mozilla.org/media/img/mozorg/mozilla-256.4720741d4108.jpg",
+    "diversityPolicy": "https://blog.mozilla.org/inclusion/mozilla-workplace-transition-policy-guidelines/",
+    "ethicsPolicy": "https://www.mozilla.org/about/manifesto/",
+    "foundingDate": "2005-08-03",
+    "founder": "Mitchell Baker",
+    "sameAs": [
+      "https://www.wikidata.org/wiki/Q169925",
+      "https://twitter.com/mozilla",
+      "https://www.facebook.com/mozilla",
+      "https://github.com/mozilla",
+      "https://www.instagram.com/mozilla/",
+      "https://www.youtube.com/user/Mozilla",
+      "https://en.wikipedia.org/wiki/Mozilla_Corporation",
+      "https://www.linkedin.com/company/mozilla-corporation"
+    ],
+    "brand": [{
+        "@type": "brand",
+        "@id": "https://www.mozilla.org/#mozilla",
+        "url": "https://www.mozilla.org",
+        "logo": "https://www.mozilla.org/media/img/mozorg/mozilla-256.4720741d4108.jpg",
+        "name": "Mozilla",
+        "sameAs": ["https://en.wikipedia.org/wiki/Mozilla", "https://www.wikidata.org/wiki/Q9661"]
+      },
+      {
+        "@type": "brand",
+        "@id": "https://www.mozilla.org/#firefoxbrand",
+        "url": "https://www.mozilla.org/firefox/",
+        "logo": "https://www.mozilla.org/media/img/logos/firefox/logo-master-wordmark-dark.9969bf7e6192.svg",
+        "name": "Firefox",
+        "sameAs": [
+          "https://twitter.com/firefox",
+          "https://www.instagram.com/firefox/",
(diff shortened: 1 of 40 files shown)


stephaniehobson:
@rraue here is some of my progress on adding the JSON files.

I'm blocked by this appearing in a few files:
```
      "owns": [
      {
      include all products
    	}
```

I think we have a few options:
1. Remove files with "owns" declarations and focus just on products.
2. Remove the "owns" parts of the files that include them (is this valid? I don't know).
3. I figure how how to get all the products listed in there.

3 will be very slow, but 1 or 2 could happen quickly.

rraue:
I think it needs to be option 3 as exactly the connection who own is an important data point itself. To make it faster, we could define two versions of each product in two files, one which can stand for its own (this is how the jsons are designed right now) and one being included as "owns" and "hasPart" in the softwareapllication as well. 

Would this two files policy speed it up? Then i would work on the files so you just need to include them.

bedrock/base/templates/includes/structured-data/organization/mozilla-corporation-organization.json
```diff
@@ -0,0 +1,61 @@
...
+        "url": "https://www.mozilla.org",
+        "logo": "https://www.mozilla.org/media/img/mozorg/mozilla-256.4720741d4108.jpg",
+        "name": "Mozilla",
+        "sameAs": ["https://en.wikipedia.org/wiki/Mozilla", "https://www.wikidata.org/wiki/Q9661"]
+      },
+      {
+        "@type": "brand",
+        "@id": "https://www.mozilla.org/#firefoxbrand",
+        "url": "https://www.mozilla.org/firefox/",
+        "logo": "https://www.mozilla.org/media/img/logos/firefox/logo-master-wordmark-dark.9969bf7e6192.svg",
```
pdehaan: What's the likelihood that these SHAs in the URLs will change/break in the future?

bedrock/base/templates/includes/structured-data/organization/mozilla-foundation-organization.json
```diff
@@ -0,0 +1,31 @@
...
+      "name": "Mozilla Foundation",
+      "url": "https://foundation.mozilla.org",
+      "sameAs": [
+        "https://www.wikidata.org/wiki/Q55672",
+        "https://foundation.mozilla.org",
+        "https://en.wikipedia.org/wiki/Mozilla_Foundation"
+      ],
+      "subOrganization": {
+        "@type": "Organization",
+		"@id": "https://www.mozilla.org/#organization",
```
pdehaan: nit: indenting is a bit off in some of these files re: tabs-vs-spaces.
You could try <kbd>npx prettier ./bedrock/base/templates/includes/structured-data/**/*.json --write</kbd> to auto-format the JSON files, but there are a couple errors in some of the files.

bedrock/base/templates/includes/structured-data/organization/mozilla-foundation-organization.json
```diff
@@ -0,0 +1,31 @@
...
+      	"https://twitter.com/mozilla",
+      	"https://www.facebook.com/mozilla",
+      	"https://github.com/mozilla",
+      	"https://www.instagram.com/mozilla/",
+      	"https://www.youtube.com/user/Mozilla",
+      	"https://en.wikipedia.org/wiki/Mozilla_Corporation",
+      	"https://www.linkedin.com/company/mozilla-corporation"
+    	]
+      }
+  }
```
pdehaan: Build out of pure curiosity, but wondering if having a link checker for all these URLs would be useful.
The one huge caveat is that we'd need to know what schema each JSON file has to determine where to look for URLs (which may be more trouble than it's worth):

```js
const fs = require("fs");
const checkLinks = require("check-links");

// See https://medium.com/javascript-inside/safely-accessing-deeply-nested-values-in-javascript-99bf72a0855a
const get = (p, o) => p.reduce((xs, x) => (xs && xs[x]) ? xs[x] : null, o);

main();

async function main() {
  const json = loadJsonSchema("./bedrock/base/templates/includes/structured-data/organization/mozilla-foundation-organization.json");
  const properties = [
    "url",
    "sameAs",
    "subOrganization.url",
    "subOrganization.logo",
    "subOrganization.sameAs"
  ];
  const links = getProps(properties, json);
  const res = await checkLinks(links);

  const brokenLinks = Object.entries(res)
    .reduce((arr, [url, res]) => arr.push(Object.assign({url}, res)) && arr, [])
    .filter(link => link.status !== "alive");
  console.log(brokenLinks);
}

function getProps(properties, json) {
  // NOTE: `Array#flat()` requires Node 12+.
  const res = properties.map(prop => get(prop.split("."), json)).flat();
  // Use a temporary `Set()` to remove potential duplicate URLs.
  return [...new Set(res)].sort();
}

function loadJsonSchema(p) {
  return require(p);
}
```

bedrock/base/templates/includes/structured-data/organization/mozilla-corporation-organization.json
stephaniehobson: Good. But, these aren't the final links to the images. We can't use .svg for this purpose.

stephaniehobson:
@pdehaan This is a draft and not ready for review. Don't spend too much time reviewing it if you have other tasks :)

stephaniehobson:
@rraue The two file policy would speed it up 👍

rraue:
I updated the organisations json and included all products in a valid way, so we do not have two files, as this makes no sense, we will have to update two files anyways for now. Lets think about a better structure after october.

PR #7665, opened 2019-09-05
[fix #7370] Remove outdated whatsnew templates

### LINKED ISSUE #7370: Decommission old /whatsnew templates.
## Description

Our logic for displaying the `/whatsnew` page is [growing unwieldly](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/views.py#L594-L629). It's time to remove some of the older templates that we're shown to past releases.

---

- [ ] Old templates are removed.
- [ ] URLs fallback to the default /whatsnew page.

- [ ] We should make sure traffic is acceptably low to templates being removed.

#### Tasks
- [ ] Remove templates and associated assets.
- [ ] Update tests.

craigcook:
## Description
Deletes WNP templates prior to 67.0. 

67.0.1 was the trailhead release so the template logic gets a little kooky around that version.

This also removes the 68.0 experiment files.

## Issue / Bugzilla link
#7370 

## Testing

- [x] Firefox 66.0 and older loads the old whatsnew page (quantum design with device widget).
- [x] Firefox 67.0 loads the 67.0 page (pre-trailhead, purple and orange).
- [x] Firefox 67.0.* loads the 67.0.5 Trailhead page for EN/FR/DE locales , the standard 67 page for other locales.
- [x] Firefox 68 loads the 68 Trailhead page for EN/FR/DE (same design as 67.0.1, just different tracking params), the standard 68 page for other locales (same design as 67.0, just different tracking params).
- [x] Firefox 69.0 loads the 69 page.
- [x] Did I forget to delete any obsolete assets?


alexgibson:
There are some context variables here that can be cleaned up:

https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/views.py#L548-L578

craigcook:
> Loading either of these URLs throws an error:
> 
>     * http://127.0.0.1:8000/it/firefox/68.0beta/whatsnew/
> 
>     * http://127.0.0.1:8000/it/firefox/68.0a2/whatsnew/
> 
> 
> It looks like both Beta and Dev edition are extending some other templates that are being removed here. Perhaps we can remove some of these too? (rather then keep them around).

Doh. Seems the non-trailhead 68beta template extends the 66 template, not the 67 template like it does in 68 release. I'll update.

bedrock/firefox/templates/firefox/developer/whatsnew-fx68.html
```diff
@@ -2,32 +2,82 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
-{% add_lang_files "firefox/whatsnew_66" %}
+{% add_lang_files "firefox/whatsnew_67" %}
```
alexgibson: Not entirely sure if this is something to worry about, but if we're switching templates here does that have l10n implications?

I'm guessing locales that have `whatsnew_66.lang` and `whatsnew_67.lang` set to active may not necessarily map 1:1, and afaik release engineering only trigger each whatsnew release using a list of locales that have each page translated. I'm not sure on the numbers this may equatesto for beta and dev edition, but is there a risk some people could see a page in the wrong language here?

alexgibson: Could falling back to the (more widely translated) default WNP be a viable option here, if this is an issue?

craigcook: I'm actually not sure what the best path here is... there was some kind of special rule for Dev Edition for 68 and then for 68-trailhead, and that seems to no longer be very relevant... so I'm thinking we can possibly decom both of those as and have just the single standard WNP for dev edition.

alexgibson: This is a good point - those pages we're meant as a one-off thing for trailhead, and we agreed that the regular Dev Edition pages shold return after. If it simplifies this PR, I'd be in favor of retiring those one-off pages and letting the original defaults take precedence?

craigcook: I shall make it so.

hoosteeno:
This is waiting on an update from @craigcook. Not a p1.

alexgibson:
Looks like this one needs a rebase

alexgibson:
@craigcook this is looking good but I think there's still a couple of tests that need updating (CircleCI failure)

bedrock/firefox/templates/firefox/whatsnew/beta/whatsnew-fx68.html
```diff
@@ -2,32 +2,82 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
-{% add_lang_files "firefox/whatsnew_66" %}
```
alexgibson: Wait, I thought we said we we're going to remove these templates for beta (like you have for dev edition), since changing the lang file could cause problems? Shouldn't beta fallback to the default /whatsnew?

craigcook: You're right, I need to do beta as well. Sorry this has been so messy, gonna try to clean it up.

alexgibson:
Looks like this needs a rebase

craigcook:
Rebased and removed the 68beta templates.

Issue #7536, opened 2019-08-06
Establish standard naming conventions for Fluent ids.


hoosteeno:
Requirements:
* namespaces strings so that a template including strings from multiple files will not cause accidental collisions
* gives some semantic value that helps someone reading the ID to know where to look for it in a template
* anticipates the need to change string IDs to trigger relocalization, even when string namespace and template position do not change. E.g. global_nav_cta_v11 -> global_nav_cta_v12.

alexgibson:
Possible example:

L10n filename: `firefox-new-skyline.ftl`

Fluent ID convention: L10n filename prefix + semantic element name + version modifier suffix.

Example ID: `firefox_new_skyline_main_page_title_v2`.

Open questions / observations:

- The `firefox-new-skyline` filename prefix is explicit, but could make for a long ID.
- Are underscores the desired convention, vs dashes?
- Is `v2` for the modifier suffix descriptive enough? Could we use a date instead e.g. `20191208`?

alexgibson:
I don't know if @pmac has any preference here maybe, but my thinking is currently:

- The filename prefix is probably OK.
- Use underscores (Python convention)
- A version modifier (e.g. `v_2`) is shorter and easier to read than a date format.

hoosteeno:
>     * A version modifier (e.g. `_v2`) is shorter and easier to read than a date format.

Would a campaign-specific modifier be acceptable? E.g. `firefox_new_trailhead`? I can imagine that we might one day wish to make a set of changes across several fluent files as part of one large initiative. In other systems (e.g. instrumentation, template naming) we have used campaign names like this.

I'm not advocating for it, just trying to tease out how rigid we will be with the standard.

alexgibson:
Campaigns usually involve redesigned pages, in which case we'd probably create a new Fluent file for that campaign (e.g. `firefox_new_skyline.ftl`). However for minor content updates you're right, we may want to use a uniform modifier across multiple files. 

Perhaps `_v_*` could be flexible enough, but we should have some rules around it (e.g. keep it short & readable).

PR #7679, opened 2019-09-06
Integrate Fluent L10n system into bedrock

pmac:
@Pike this is a very rough first go at getting this working. I'd like your thoughts as I'm sure I'm doing the whole `FluentBundle`, `FluentResource` dance incorrectly.

This is work toward #6775


lib/l10n_utils/__init__.py
```diff
@@ -52,10 +53,13 @@ def render(request, template, context=None, **kwargs):
     if isinstance(template, list):
         template = template[0]
 
+    if ftl_files:
+        context['ftl_files'] = ftl_files
```
Pike: I'd create something like a `FluentLocalization` here instead. That'll allow you to cache bundles, which, sadly, are quite expensive to create.

I've started to hack on a branch for that on https://github.com/Pike/python-fluent/pull/1.

lib/l10n_utils/fluent.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+def has_message(message_id, bundle):
+    # assume English locales have the message
+    if bundle.locales[0].startswith('en-'):
+        return True
+
+    if not bundle._bundle_cache:
+        # need to warm the cache in the bundle
+        bundle.format_value('warm')
```
Pike: This one is worst-case performance, it will load the full fallback chain always.

I'd be up for taking this API into python-fluent, too.

Big picture question, though:

This is part of the activation strategy? Should that be a runtime thing or a deployment thing?

pmac: I've reworked this a good bit. The use case is that in the template we need to be able to know whether a translation has a message_id or not so that we can show some new part of the page or keep it hidden until the translation of the new string(s) is done.

lib/l10n_utils/fluent.py
```diff
@@ -0,0 +1,80 @@
...
+from django.utils.encoding import force_bytes
+
+from fluent.runtime import FluentLocalization, FluentResourceLoader
+
+
+cache = caches['l10n']
+
+
+class FluentL10n(FluentLocalization):
+    def has_message(self, message_id):
```
pmac: @Pike I moved that logic here. I was going to suggest it as an addition to your fallback PR but I'm not sure the whole `if bundle.locales[0] != 'en'` is appropriate for a wider audence, so it seemed like a bedrock specific need. If you think you can pull it in then great.

lib/l10n_utils/fluent.py
```diff
@@ -5,16 +5,30 @@
...
+        # assume English locales have the message
+        if self.locales[0].startswith('en-'):
+            return True
+
+        for bundle in self._bundles():
+            # bundle.locales[0] is the locale of the loaded FTL files
+            # we want to know if the non-en locale has the message
+            if bundle.locales[0] != 'en' and bundle.has_message(message_id):
+                return True
```
Pike: I was thinking about doing a `break` here, which could be optional.

And instead of checking for `en`, you could check if `bundle.locales[0] == self.locales[0]`, I think that's generic enough to upstream.

lib/l10n_utils/fluent.py
```diff
@@ -5,16 +5,30 @@
...
 
 
 cache = caches['l10n']
 
 
+class FluentL10n(FluentLocalization):
+    def has_message(self, message_id):
+        # assume English locales have the message
+        if self.locales[0].startswith('en-'):
+            return True
```
Pike: Do you want en-GB to pass here? But yes, I think this would be better suited for a subclass.

One question I have is if you want this to return true for English, or for in-development. If just for English is good enough, this could have an option "including_fallback=False", which would skip the `break` and the locale check.

lib/l10n_utils/fluent.py
pmac: Well `return` is another way of saying `break`. That updated check sounds good to me 👍

lib/l10n_utils/fluent.py
pmac: We'll probably need to return True for development as well, that's a good point. I'm not sure yet whether to return `False` for `en-US` for example. I think the majority of FTL files for `en-*` are going to be very short and the `en` fallback for whatever it is will be good enough for the code block to be shown. This is probably something I just need to discuss with the dev team.

lib/l10n_utils/fluent.py
```diff
@@ -0,0 +1,90 @@
...
+from fluent.runtime import FluentLocalization, FluentResourceLoader
+
+
+cache = caches['l10n']
+
+
+class FluentL10n(FluentLocalization):
+    _localized_messages = None
+
+    def _localized_bundles(self):
```
pmac: @Pike perhaps this is a good way to go for the `FluentLocalization` class? Then you can do things like below.

bedrock/mozorg/templates/mozorg/mission_ftl.html
```diff
@@ -0,0 +1,77 @@
...
+
+{% block page_css %}
+  {{ css_bundle('mission') }}
+{% endblock %}
+
+{% block article %}
+<article>
+  <h1 class="mzp-c-article-title">{{ ftl('mission_were_building_a_better_internet') }}</h1>
+  {% if l10n_has_tag('bug_1251061_copy_update') %}
+    <p class="lede">{{ ftl('mission_our_mission_is_to_ensure_the_internet') }}</p>
```
pmac: @alexgibson this template was auto-converted by my new management commands. I'd love to hear your thoughts.

alexgibson: I'll take this for a spin and report back 👍

bedrock/mozorg/templates/mozorg/mission_ftl.html
```diff
@@ -0,0 +1,73 @@
...
+
+  <div class="mzp-l-content">
+    <aside class="mzp-c-newsletter t-love">
+      <div class="mzp-c-newsletter-image">
+        <img src="{{ static('img/newsletter/love-news.svg') }}" alt="" height="172" width="200">
+      </div>
+
+      <div class="newsletter-content">
+        {% if LANG.startswith('en-') %}
+          {{ email_newsletter_form(protocol_component=true, newsletters='mozilla-foundation', title=_('Sign up. Read up.<br> Make a difference.'), subtitle=_('Get the Mozilla newsletter and help us keep the Web free and open.')) }}
```
alexgibson: I noticed that there's still some `_()` calls here in the newsletter module. This may be something that the template conversion didn't pick up?

Running locally, I noticed these strings still appear translated OK, so they must be pulling from the old lang file still? Perhaps this is ok, but I could see it causing some confusion, especially around activation status.

We need to decide what is the most appropriate thing to here. If a template has been converted to Fluent, should it then ignore the .lang file associated with the template? (Excluding widgets & global elements?).

alexgibson: Oh wait, other locales are shown a different set of strings for the newsletter here (which pulls from a global lang file). So what I thought I was seeing isn't strictly correct.

pmac: Right. I saw that too. The reason my template converter didn't get these is that these strings didn't appear in the lang file being converted. So they're in some global lang file. The template converter only replaces calls to `_()` where it has a corresponding new fluent message ID.

pmac:
This is getting to the "too large" stage. I'm calling it ready to review. The more eyes and people trying it out the better.

pmac:
It could use some more tests, but we can also getting some real user testing would be good as far as the documented usage instructions.

NOTE: The activation percent and required strings rules don't yet do anything since those will be enforced in the automation between the l10n team's repo and ours, which isn't yet setup. I wanted to get the documentation done though, and to set up the JSON schema for the metadata files.

bedrock/settings/base.py
```diff
@@ -227,10 +232,24 @@ def lazy_langs():
 LANGUAGES = lazy(lazy_langs, dict)()
 
 FEED_CACHE = 3900
-DOTLANG_CACHE = 600
+DOTLANG_CACHE = config('L10N_CACHE_TIMEOUT', default='600', parser=int)
```
alexgibson: I had to set the value for `L10N_CACHE_TIMEOUT` to `10` locally when converting pages to avoid caching issues. I'm guessing others may also have similar issues, so documenting this may be useful?

docs/l10n.rst
```diff
@@ -9,17 +9,303 @@ Localization
...
+
+Fluent File Activation
+----------------------
+
+Fluent files are activated automatically when processed from the l10n team's repo
+into our own based on a couple of rules.
+
+1. If a fluent file has a group of required strings, all of those strings must be present in
+   the translation in order for it to be activated.
+2. A translation must contain a minimum percent of the string IDs from the English file to be activated.
```
alexgibson: Where is the minimum percentage for activation set? (what is the default value?), and what should a developer do if they are creating a new page and want it to use a different percentage?

bedrock/settings/base.py
```diff
@@ -227,10 +232,24 @@ def lazy_langs():
...
 
 DOTLANG_FILES = ['navigation', 'download_button', 'main']
 
+FLUENT_DEFAULT_FILES = ['navigation', 'footer', 'download_button']
```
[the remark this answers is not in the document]
alexgibson: Side question: how do our default activation rules come into play with regard to global files? Do we have logic in place so these are always active?

bedrock/mozorg/urls.py
```diff
@@ -32,7 +32,7 @@
     # Bug 981063, catch all for old calendar urls.
     # must be here to avoid overriding the above
     redirect(r'^projects/calendar/', 'https://www.thunderbird.net/calendar/', locale_prefix=False),
-    page('mission', 'mozorg/mission.html'),
+    page('mission', 'mozorg/mission_ftl.html', url_name='mozorg.mission', ftl_files=['mozorg/mission']),
```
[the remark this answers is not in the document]
pmac: No reason apart from demonstration or if in future we think there's a good chance of needing to roll back. We can always grab the old version from git though. I'll go ahead and just update the original file.

bedrock/settings/base.py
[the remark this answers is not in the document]
pmac: A comment is probably a good idea, and I can also just leave this list blank for now until we actually have these things converted.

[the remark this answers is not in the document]
pmac: I tried to make the activation thing clear in the docs, so please let me know if I've not accomplished that. The way it works is that all of the .ftl files are evaluated for activation, but the only time that matters is if it's the file being used as the primary for the page. So even if a global file isn't "active" for a locale, if the one that's first in the list for the current page is active, then the page will be available regardless of the state of the other ftl files in the list passed to the view or the global list.

docs/fluent-conversion.rst
```diff
@@ -0,0 +1,97 @@
+.. This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
pmac: I've also noticed some oddities in the TOC. I didn't add it because it's a temporary file that's only useful during our transition to the new system. That said this transition period could last for quite a while... I'm happy to add it if you think it'd be more useful there instead of just a link from the l10n doc..

docs/fluent-conversion.rst
```diff
@@ -0,0 +1,97 @@
...
+or the ``page()`` function in urls.py.
+
+.. code-block:: python
+
+    # urls.py
+    urlpatterns = [
+        page('mission', 'mozorg/mission.html', ftl_files=['mozorg/mission']),
+    ]
+
+.. note::
```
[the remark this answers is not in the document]
pmac: Sure 👍

docs/l10n.rst
```diff
@@ -9,17 +9,303 @@ Localization
...
+directly editing the default language (English in our case) Fluent files and using
+the string IDs created there in their templates and views.
+
+The default files for the Fluent system live in the ``l10n`` directory in the root
+of the bedrock project. This directory houses directories for each locale the developers
+directly implement (mostly simplified English "en", and "en-US"). The simplified English
+files are the default fallback for every string ID on the site and should be strings that
+are plain and easy to understand English, as free from colloquialisms as possible. The
+translators are able to easily understand the meaning of the string, and can then add their
+own local flair to the ideas.
```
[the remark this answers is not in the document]
pmac: I'm not sure about that. I think all of the .ftl files we add in bedrock will be synced over to the l10n team's repo. I'm guessing they'll ignore en-US for the most part, but I think they'll be there. Bedrock will get pretty much all its strings in production from the external repo.

docs/l10n.rst
```diff
@@ -9,17 +9,303 @@ Localization
...
+
+The `Fluent file syntax`_ is well document on the Fluent Project's site. We use "double hash" or
+"group" comments to indicate strings required for activation. A group comment only ends when
+another group comment starts however, so you should either group your required strings at the
+bottom of a file, or also have a "not required" group comment. Here's an example:
+
+.. code-block:: fluent
+
+    ## Required
+    page_title = The Page Title
```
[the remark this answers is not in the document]
pmac: Oh yes. Good catch. We should definitely have a section on string ID creation.

bedrock/settings/base.py
pmac: I can also have a much shorter default when `DEBUG == True`

bedrock/settings/base.py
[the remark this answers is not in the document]
pmac: A comment is probably a good idea, and I can also just leave this list blank for now until we actually have these things converted.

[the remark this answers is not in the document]
alexgibson: Ok, that took a couple of seconds to sink in before I really grokked how activation effects these files, but I think your explaination already in the docs totally makes sense. My brain is still wired to how global files behave in the legacy system, and it may just take a while to override that :)

docs/l10n.rst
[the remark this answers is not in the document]
pmac: That works for me. Thanks for the update. So our automation should copy over changes to `en` files for translation, but not `en-US` or any other locale we directly add to bedrock, right? And then we'll configure the system to look in both locations, preferring the strings in bedrock itself?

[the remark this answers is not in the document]
Pike: Right. Or we disambiguate `en` in the bedrock and the l10n repository (by changing the locale name), giving us the option to fast-path landings there.

lib/l10n_utils/fluent.py
```diff
@@ -0,0 +1,192 @@
...
+        for bundle in self._bundles():
+            messages.update(bundle._messages)
+
+        return messages
+
+    @cached_property
+    def _localized_messages(self):
+        messages = {}
+        for bundle in self._localized_bundles():
+            messages.update(bundle._messages)
```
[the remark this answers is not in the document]
pmac: these are both only used (so far) to get a full de-duped list of string IDs. I kept it returning dicts mostly to be consistent, but can move to just lists of IDs.

lib/l10n_utils/fluent.py
```diff
@@ -0,0 +1,192 @@
...
+
+        :return: list of message IDs
+        """
+        messages = set()
+        for resources in self.resource_loader.resources('en', self.resource_ids):
+            for resource in resources:
+                in_required = False
+                for item in resource.body:
+                    if isinstance(item, GroupComment):
+                        in_required = item.content.lower().startswith('required')
```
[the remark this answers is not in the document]
pmac: I don't follow. It doesn't seem to allow for nested group comments.

pmac:
I believe I've addressed your comments @alexgibson. Thanks for the review!

pmac:
@Pike thanks for the review! Replies inline below.

> For the Fluent files, I have a wishlist:
> 
>     * No hashes in comments. These comments are for localizers to read, and we shouldn't train them to not read them by putting cryptic data inside that.

These will not make it to the l10n team repo. They exist only during the transition and only as an easy and compact way of representing a string ID so that it can be matched in the .lang translations and the templates.

>     * Message IDs should use hypen instead of snakes. https://firefox-source-docs.mozilla.org/intl/l10n/l10n/fluent_review.html#message-identifiers is the docs we have right now, and we advise for that choice across mozilla, not just Firefox

In the absence of guidance we choose underscores in #7536 instead of hyphens. If there's a compelling reason to switch we can, but this is a python convention and so we went with it, I'm guessing similarly to how JS projects seem to prefer camel case. 

>     * URLs should be in a group comment, so that they're visible across all messages in Pontoon. I've asked for more details in https://bugzilla.mozilla.org/show_bug.cgi?id=1374605

Which URLs are you referring to? The URL for which this file is to be used? Shouldn't that be in a whole file comment (a triple hash comment)?

>     * Localizers benefit from context, and having the strings in the Fluent file match page order helps a lot.

👍 

> Other general comments: For linting and error checking, I suggest to tie back to `moz-l10n-lint` in the `compare-locales` package. That does a lot more than just getting parse errors, and there's little value in writing a competing implementation of that.

I need a linter that I can point at all files in a directory, and that project seems to rely on an `l10n.toml` file. We don't yet have such a file, and when we do it will almost certainly not refer to every file in the repo. Perhaps there are some internals in that package that I can leverage? If so I can't find any documentation, or even the code for this project as it's not linked from the [PyPI page](https://pypi.org/project/compare-locales/).
 
> We do have a migration toolchain in https://firefox-source-docs.mozilla.org/intl/l10n/l10n/fluent_migrations.html. That can be extended to support git. I have a branch, but it's not well tested. Also .lang, but that's a tad more work to make the parsers work with c-l. I didn't look tooo deep into the attempts here yet.

What I have so far is working pretty well, and I didn't know about this, and it doesn't seem to support .lang, like you said. So I think leveraging the bedrock code for parsing .lang and then porting is going to be our best bet. Bedrock doesn't do a lot of fancy string work since it's more a static site than a dynamic application, so it should be (and so far is) pretty straight forward to port.

lib/l10n_utils/fluent.py
```diff
@@ -0,0 +1,192 @@
...
+    'translate',
+]
+cache = caches['l10n']
+
+
+class FluentL10n(FluentLocalization):
+    def _localized_bundles(self):
+        for bundle in self._bundles():
+            if bundle.locales[0] == self.locales[0]:
+                yield bundle
```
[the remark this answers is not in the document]
pmac: > This code shouldn't work for `en-US`, with most strings falling back to `en`, right?

This is only used for checking the percent translated and whether a translation has a message or not, so it's really only invoked on a single language instance of this class (the way I use it anyway). The `has_message` method always returns true for `en-*`, so it's working as expected.
 
> Also, it'd be more efficient to `break` on the first bundle that's not what we want, so that we don't need to exhaust the iterator, and thus, parse all locales.

Can you not have another language bundle from a different directory later in the list?

docs/l10n.rst
```diff
@@ -9,17 +9,393 @@ Localization
...
+
+.. code-block:: python
+
+    # views.py
+    from lib.l10n_utils import render
+    from lib.l10n_utils.fluent import ftl
+
+    def about_view(request):
+        hello_string = ftl('about_hello')
+        render(request, 'about.html', {'hello', hello_string}, ftl_files='about')
```
glogiotatidis: You want a `set` here or a `dict`?

docs/l10n.rst
```diff
@@ -9,17 +9,393 @@ Localization
...
+
+    def about_view(request):
+        if has_messages('about_hello_v2', 'about_title_v2'):
+            hello_string = ftl('about_hello_v2')
+            title_string = ftl('about_title_v2')
+        else:
+            hello_string = ftl('about_hello')
+            title_string = ftl('about_title')
+
+        render(request, 'about.html', {'hello', hello_string, 'title': title_string}, ftl_files='about')
```
glogiotatidis: Same here, `set` or `dict`?

bedrock/settings/base.py
```diff
@@ -227,10 +232,25 @@ def lazy_langs():
 LANGUAGES = lazy(lazy_langs, dict)()
 
 FEED_CACHE = 3900
-DOTLANG_CACHE = 600
+DOTLANG_CACHE = config('L10N_CACHE_TIMEOUT', default='2' if DEBUG else '600', parser=int)
```
alexgibson: When pulling the latest from this branch I noticed my macbook's fans we're running like crazy. Increasing the default value here from `2` to `10` seemed to calm things down. Not sure if that's the optimal value, but you may want to test a little here for performance.

lib/l10n_utils/fluent.py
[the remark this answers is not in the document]
Pike: What it does allow is

```
## Project Trailhead Copy
##
## Required
```

I don't know if we'll want that in the end, but right now, we couldn't have a description for the required group of strings.

lib/l10n_utils/fluent.py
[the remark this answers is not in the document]
Pike: > > This code shouldn't work for `en-US`, with most strings falling back to `en`, right?
> 
> This is only used for checking the percent translated and whether a translation has a message or not, so it's really only invoked on a single language instance of this class (the way I use it anyway). The `has_message` method always returns true for `en-*`, so it's working as expected.

Could this end up not wanting to activate en-US?

> > Also, it'd be more efficient to `break` on the first bundle that's not what we want, so that we don't need to exhaust the iterator, and thus, parse all locales.
> 
> Can you not have another language bundle from a different directory later in the list?

That'd be bad, I think. Technically, yes, it's possible, but in practice, we shouldn't show English strings if we have German ones, right? The built-in `_generate_bundles` strictly goes locale by locale.

lib/l10n_utils/fluent.py
```diff
@@ -0,0 +1,192 @@
...
+
+    @cached_property
+    def _messages(self):
+        messages = {}
+        for bundle in self._bundles():
+            messages.update(bundle._messages)
```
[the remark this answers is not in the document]
pmac: Updated to be `_message_ids()` and return a list of string IDs instead of the messages, as that's all I need.

lib/l10n_utils/fluent.py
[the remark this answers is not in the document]
pmac: Updated to be `_localized_message_ids()` and return a list of string IDs instead of the messages, as that's all I need.

lib/l10n_utils/fluent.py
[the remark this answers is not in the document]
pmac: Do contiguous lines of group comments like that result in a single multi-line group comment string? If so I could make that work.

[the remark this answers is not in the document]
Pike: Yes. I've demo-ed some content on https://projectfluent.org/play/?id=6bfc32a0025b9ad6776169d3f28b8126, you can look at the AST tab to see what you'll get exactly.

lib/l10n_utils/utils.py
```diff
@@ -18,3 +41,22 @@ def strip_whitespace(message):
...
+
+    try:
+        i = p.index('templates')
+        p = p[i + 1:]
+    except ValueError:
+        pass
+
+    path = '/'.join(p)
+    base, ext = os.path.splitext(path)
+    return base
```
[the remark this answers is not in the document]
pmac: Great catch! Fixed and added tests.

lib/l10n_utils/management/commands/template_to_ftl.py
```diff
@@ -0,0 +1,117 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import re
+from django.utils.functional import cached_property
```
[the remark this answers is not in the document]
pmac: Blasted auto import adding by PyCharm. It's very nice but I always forget to go clean up its mess. Thanks for catching that.

docs/l10n.rst
pmac: 🤦‍♂

bedrock/settings/base.py
pmac: Works for me. It's a balance for sure. Need long enough to cache multiple calls during a request, but short enough not to span multiple requests. 10 should be fine under normal conditions.

pmac:
> Sorry for not jumping in on that ticket. We shared docs on good IDs in the PRD phase, though they're not as explicit as the reviewer docs that we created for Firefox at a later stage. Fluent is cross-platform, and so are the ID guidelines we have for Fluent files.

If you have a strong preference for hyphens instead of underscores that's fine with me as long as @alexgibson has no objection. It wouldn't be too difficult to change at this point.

alexgibson:
No objections here 👍

lib/l10n_utils/fluent.py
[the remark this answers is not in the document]
pmac: Great! This is now done. All that is required is that one of the lines start with the word "Required" case-insensitive.

lib/l10n_utils/fluent.py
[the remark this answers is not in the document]
pmac: Right, but what if we had a `de/mozorg/mission.ftl` in our bedrock l10n folder, and there was also one in the strings from the l10n team's repo. That is a situation that is possible if the German marketing team wants to override a string but we leave the rest alone.

pmac:
> You lost me here. My understanding is that the `en` files in the l10n repo will be one-to-one copies of the files in bedrock.

Well yes... but during the porting process is the only time these hash comments would be created, then when moving the `en` file over to our repo I filter those comments out. I should probably add another command to remove them from the file in bedrock as well so they don't accidentially move into your repo sometime later. I don't want to automatically remove them in case the dev need to run things again, but they should be easily able to remove them before committing the ported file.

pmac:
> You can call into https://hg.mozilla.org/l10n/compare-locales/file/tip/compare_locales/lint/linter.py#l25, with a path and `None` for the other two args.

@Pike that's cool. I'm just a bit wary of using it since there are no docs and it looks to be built for a much broader scope than just ftl files. I think we should file a sparate issue for improving ftl linting and be okay with what we've got so far for this PR.

Pike:
Made it through the l10n doc today, nice to see docs.

Some high-level comments, though:

I'd avoid any wording that talks about change from A to B. Those stay around for years. It might be good to contrast to gettext still, but only because that's abundant among django developers, not because it's the past of bedrock, too.

I'd also like to clarify where we hope to hook up Fluent. In other projects, we try to be as far down-stream as possible, in the case of bedrock, that'd mean that we try to resolve all things in the template, as much as possible.

The code examples in the doc talk a lot about resolving strings on the python side and then passing them in as params to the template.

Is that just convenience in showing the code? As the mission page does what I'd expect in the template.

Also, is the doc on `render` short on purpose? Reading the docs on the `L10nTemplateView`, it surprised me to find the bulk of the functionality on `render` itself.

Pike:
As this has been mentioned a few times, here's an OK-to-start l10n.toml file:

https://gist.github.com/Pike/191217502f5188807d92e4b35a0a1f44

This is prepared to exclude stuff we want in BWX, but doesn't include one as we don't yet know what's supposed to be in there.

pmac:
> I'd avoid any wording that talks about change from A to B. Those stay around for years. It might be good to contrast to gettext still, but only because that's abundant among django developers, not because it's the past of bedrock, too.

I'm not sure what you're referring to here. Which part of the docs is this?
 
> I'd also like to clarify where we hope to hook up Fluent. In other projects, we try to be as far down-stream as possible, in the case of bedrock, that'd mean that we try to resolve all things in the template, as much as possible.

Most will be in the template, but we do have some situations where we've had to do things in views, and so I document that as well.

> The code examples in the doc talk a lot about resolving strings on the python side and then passing them in as params to the template.

It does happen.

> Is that just convenience in showing the code? As the mission page does what I'd expect in the template.

Only doing things in the templates will cover the vast majority of cases.

> Also, is the doc on `render` short on purpose? Reading the docs on the `L10nTemplateView`, it surprised me to find the bulk of the functionality on `render` itself.

It's pretty much Django's render function with some l10n magic, so I really only cover the difference, which here is pretty much the `ftl_files` parameter.

docs/l10n.rst
```diff
@@ -9,17 +9,421 @@ Localization
...
 depending on your ``DEV`` setting). If the folder is already present, it will
-update the repository to the latest version.
+update the repository to the latest version. It do the same thing for the
+repository for the .ftl translation files in ``git-repos/www-l10n`` directory.
+
+Fluent
+======
+
+Bedrock's Localization (l10n) system is based on `Project Fluent`_. This is a
+departure from the legacy system (see below) that relied on a gettext work flow
```
Pike: This is the part of the docs that talks about changes.

docs/l10n.rst
```diff
@@ -9,17 +9,421 @@ Localization
...
+    from lib.l10n_utils import render
+    from lib.l10n_utils.fluent import ftl, has_messages
+
+    def about_view(request):
+        if has_messages('about-hello-v2', 'about-title-v2'):
+            hello_string = ftl('about-hello-v2')
+            title_string = ftl('about-title-v2')
+        else:
+            hello_string = ftl('about-hello')
+            title_string = ftl('about-title')
```
Pike: Can we turn this example into a feature switch thing? Like, 

```python
    use_skyline_layout = has_messages('about-hello-v2', 'about-title-v2')
    render(request, 'about.html', {'use_skyline_layout': use_skyline_layout}, ftl_files='about')
```

You can probably come up with a better flag name. I hope that that's closer what will happen in practice?

docs/l10n.rst
```diff
@@ -9,17 +9,421 @@ Localization
...
+locale. English locales (e.g. en-US, en-GB) will never use the fallback and will print the simplified
+english version of the primary string if not overridden in the more specific locale.
+
+You can also pass in replacement variables into the ``ftl()`` function for use with `fluent variables`_.
+If you had a variable in your fluent file like this:
+
+.. code-block:: fluent
+
+    welcome = Welcome, { $user }!
+
```
Pike: Should we make it explicit that you can't use `fallback` as a variable name? They're right next to each other, but maybe that's still good to call out?

docs/l10n.rst
```diff
@@ -9,17 +9,421 @@ Localization
...
+    from lib.l10n_utils import render
+    from lib.l10n_utils.fluent import ftl
+
+    def about_view(request):
+        hello_string = ftl('about-hello')
+        render(request, 'about.html', {'hello': hello_string}, ftl_files='about')
+
+If you need to use this string in a view, but define it outside of the view itself, you can use the
+``ftl_lazy`` variant which will delay evaluation until render time. This is mostly useful for defining
+messages shared among several views in constants in a ``views.py`` or ``models.py`` file.
```
Pike: Should we prefer a shared fluent file for these? Or at least mention that that's a perfectly fine option?

Pike:
Theo noticed different comments in https://github.com/mozmeao/www-l10n/blob/master/en/mozorg/mission.ftl and https://github.com/mozmeao/www-l10n/blob/master/fr/mozorg/mission.ftl.

He also raised the question what happens in the migration if we have differently cased strings in the lang files, like `Download Now` and `Download now`.

pmac:
> Theo noticed different comments in https://github.com/mozmeao/www-l10n/blob/master/en/mozorg/mission.ftl and https://github.com/mozmeao/www-l10n/blob/master/fr/mozorg/mission.ftl.

I believe that's because of differences between https://github.com/mozilla-l10n/bedrock-l10n/blob/master/en-US/mozorg/mission.lang and https://github.com/mozilla-l10n/bedrock-l10n/blob/master/de/mozorg/mission.lang. We can only do as good with porting these as we have.
 
> He also raised the question what happens in the migration if we have differently cased strings in the lang files, like `Download Now` and `Download now`.

They are different strings in the legacy system, and also I assume in the new one. We can try to clean that up, but might not catch everything. That's probably going to have to be what PR review is for. Would we not need separate string IDs for capitalization differences if we want it to look like that on the English page?

Pike:
> > Theo noticed different comments in https://github.com/mozmeao/www-l10n/blob/master/en/mozorg/mission.ftl and https://github.com/mozmeao/www-l10n/blob/master/fr/mozorg/mission.ftl.
> 
> I believe that's because of differences between https://github.com/mozilla-l10n/bedrock-l10n/blob/master/en-US/mozorg/mission.lang and https://github.com/mozilla-l10n/bedrock-l10n/blob/master/de/mozorg/mission.lang. We can only do as good with porting these as we have.

I disagree on this. We have full control over what's created as output, and we don't need to grandfather problems of the old setup into the new one.

> > He also raised the question what happens in the migration if we have differently cased strings in the lang files, like `Download Now` and `Download now`.
> 
> They are different strings in the legacy system, and also I assume in the new one. We can try to clean that up, but might not catch everything. That's probably going to have to be what PR review is for. Would we not need separate string IDs for capitalization differences if we want it to look like that on the English page?

Right, and I think his point was that the ID generation code in the migration helpers creates the same ID for both strings.

pmac:
> Right, and I think his point was that the ID generation code in the migration helpers creates the same ID for both strings.

No. It ensures that each ID is unique for the file and will append a suffix if it finds that the ID already exists.

pmac:
> I disagree on this. We have full control over what's created as output, and we don't need to grandfather problems of the old setup into the new one.

We do, but I can't really do anything about it with automation. We'll need to manually fix it, and honestly I don't know what the problem is.

pmac:
I have updated this to the latest `fluent.runtime` release so I think we're about ready to merge. Thoughts @alexgibson ? I definitely wanted to wait until Skyline things calmed down. So let me know when you feel comfortable. I'll probably also do some squashing before we merge.

alexgibson:
@pmac I'm happy to go with whenever you feel ready. What's the scope for this impacting skyline? It feels pretty self-contained, what's the likelyhood of impact?

alexgibson:
Plan is to merge this on Monday 28th

pmac:
> What's the scope for this impacting skyline? It feels pretty self-contained, what's the likelyhood of impact?

I think there is very low risk of any impact at this point, but I do feel like it's a lot of change and I just didn't want to take any risk at this critical time really. My focus with that statement is to avoid introduction of a lot of churn in the git history just in case anything has to be reverted and/or quickly deployed. I think Monday is great. Thanks 👍

PR #7680, opened 2019-09-06
Create Lockwise page [ Fixes  #7570 ]

### LINKED ISSUE #7570: Code the new Lockwise product page at www.mozilla.org/firefox/lockwise
Description: Create the product page for Lockwise

----

***Success Criteria***
- [ ] Page is live on or before October 22nd to Tier One locales

***Risks***

***Tasks***
- [x] Code the page
- [x] Add conditional content for the Add-on install button for dekstop
- [ ] Find all links on www.mozilla.org that point at lockwise.firefox.com and point at the new page. **NOTE** Remove same-site utm parameters. If we previously included utm params on lockwise.firefox.com, we should not include them when we change the link to www.mozilla.org/*.
- [ ] Set up a 301 redirect lockwise.firefox.com to www.mozilla.org/firefox/lockwise
- [x] Source [localized images](https://drive.google.com/drive/folders/19Zrt7jc11e25PmSydNzsRkkJOzo6vLHL), code in the page
- [x] Source Adjust links, code in the page
- [x] Extract strings for L10n w. Tier One as MVP - but goal to make it available to all locales (which implies extracting strings at least two weeks prior to the target launch date)
- [x] Identify Protocol component requirements. Build / update new components that are required.
- [x] Wrap the addon CTA in conditional logic to work only for Fx 69 and below
- [x] Link to the product FAQ that is moving to SUMO

Errolyn:
## Description
Create the Lockwise page. I have a demo of the page up at: https://bedrock-demo-lockwise.oregon-b.moz.works/en-US/firefox/lockwise/

## Issue / Bugzilla link
 #7570

## Testing
firefox/lockwise/ has 4 different possibilities for lockwise cta button, and needs to be tested for all four channels:

- Mobile(all browsers): should not show up (only the app store buttons should be there) 
- Firefox v70+ should open menu and highlight logins and passwords
- Firefox v69- should take you to download the add-on
- In a different browser should take you to install firefox


bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,83 @@
+{% add_lang_files "firefox/products/lockwise" %}
+
+{% from "macros.html" import google_play_button with context %}
+{% from "macros-protocol.html" import hero, picto_card with context %}
+
+{% extends "firefox/base/base-protocol.html" %}
+
+{% block page_title %}{{_('Firefox Lockwise — password manager —  take your passwords everywhere')}}{% endblock %}
+
+{% block page_desc %}{{ _('Firefox Lockwise lets you securely access the passwords you’ve saved in Firefox from anywhere— even outside of the browser. Features 256-bit encryption and Face/Touch ID.') }}{% endblock %}
```
peiying2: @Errolyn you need to add an extra space before the M dash, the changes you made in the file for l10n. Otherwise the string won't show up. It needs to match. 

`...from anywhere— even outside ...` to .`... from anywhere — even outside ...`

Errolyn: thanks for the heads up!

ejregithub:
@peiying2 - please note this can go out to Tier One and 'it' and 'es-ES'

bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,91 @@
...
+          </div>
+        </section>
+        <section class="mzp-c-card-picto">
+          <div class="mzp-c-card-picto-content touch">
+            <p class="picto-content">{{ _('Get to your passwords securely with Face or Touch ID') }}</p>
+          </div>
+        </section>
+        <section class="mzp-c-card-picto">
+          <div class="mzp-c-card-picto-content privacy">
+            <p class="picto-content"><span>{{ _('Your privacy comes first.')}}</span><span>{{ _('We keep your data safe, never sold.') }}</span></p>
```
mnoorenberghe: Drive-by nit: This renders without a space between the two strings: "…first.We…"

bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,105 @@
...
+          </div>
+        </section>
+        <section class="mzp-c-card-picto">
+          <div class="mzp-c-card-picto-content touch">
+            <p class="picto-content">{{ _('Get to your passwords securely with Face or Touch ID') }}</p>
+          </div>
+        </section>
+        <section class="mzp-c-card-picto">
+          <div class="mzp-c-card-picto-content privacy">
+            <p class="picto-content"><span>{{ _('Your privacy comes first. ')}}</span><span>{{ _('We keep your data safe, never sold.') }}</span></p>
```
mnoorenberghe: Shouldn't the space be between the `</span> <span>` otherwise all localizers will need to include the space too?

alexgibson: You're right the space should be between the `<span></span>`. I'd be careful taking the space out of the `Your privacy comes first.` string though, as even a white space changes can cause translations to break.

bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,105 @@
...
+            <div class="mzp-c-button-download-container">
+              <p >{{ _('Only in the Firefox browser') }}</p>
+              <a href="{{ url('firefox.download.thanks') }}" class="mzp-c-button mzp-t-product" data-cta-type="Lockwise-DownloadFX" data-cta-position="Primary">{{ _('Get Firefox') }}</a>
+            </div>
+          </div>
+          <div class="mzp-c-hero-image">
+          {% if LANG.startswith('en-') %}
+            <img src="{{ static('img/firefox/lockwise/EN.svg') }}" alt="">
+          {% elif LANG == 'de' %}
+            <img src="{{ static('img/fmake irefox/lockwise/DE.svg') }}" alt="">
```
alexgibson: Typo here in the file path. I'd also suggest making the filenames lowercase.

bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,105 @@
...
+          <a href='https://support.mozilla.org/products/firefox-lockwise?utm_source=www.mozilla.org-lockwise&amp;utm_campaign=Lockwise&amp;utm_medium=referral' >{{ _('Support') }}</a>
+          <a href='https://lockwise.firefox.com/privacy.html'>{{ _('Your Privacy') }}</a>
+          <a href='https://github.com/mozilla-lockwise'>{{ _('GitHub') }}</a>
+        </section>
+      </section>
+    </section>
+  </main>
+{% endblock %}
+
+{% block js %}{{ js_bundle('lockwise') }}{% endblock %}
```
alexgibson: Nit: indent block

media/css/firefox/lockwise/lockwise.scss
```diff
@@ -0,0 +1,70 @@
+
```
alexgibson: NIt: include license at top of file

media/css/firefox/lockwise/lockwise.scss
```diff
@@ -0,0 +1,70 @@
...
+    margin-bottom: $spacing-lg;
+}
+.mzp-c-hero-image {
+    padding-top: $spacing-lg;
+}
+
+
+@media #{$mq-sm} {
+    .mobile-download-buttons {
+    margin-bottom: $spacing-lg;
```
alexgibson: Nit: indentation

media/img/firefox/lockwise/arrows.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
alexgibson: Optimise SVG (this one and others).

media/js/firefox/lockwise/lockwise.js
```diff
@@ -0,0 +1,26 @@
+(function() {
+    'use strict';
+
+    var client = window.Mozilla.Client;
+
+    if (client._isFirefox() === true && client._getFirefoxMajorVersion() >= '70') {
```
alexgibson: Nit: can simplify the conditional a bit here:
```suggestion
    if (client.isFirefox && client._getFirefoxMajorVersion() >= '70') {
```

media/js/firefox/lockwise/lockwise.js
```diff
@@ -0,0 +1,26 @@
...
+        document.querySelector('.for-firefox-69-and-below').remove();
+        document.querySelector('.for-non-firefox-users').remove();  
+    }
+    else if ( client._isFirefox() === true && client._getFirefoxMajorVersion() < '70' ) {
```
alexgibson: Rather than repeating the call to `client._getFirefoxMajorVersion()` again here, could you store the result in a variable for reference a second time? Maybe the conditionals here could be reworked a little?

media/js/firefox/lockwise/lockwise.js
```diff
@@ -0,0 +1,26 @@
...
+        document.querySelector('.for-firefox-70-and-above').remove();
+        document.querySelector('.for-non-firefox-users').remove();
+    }
+    else if ( client._isFirefox() === false ){
```
alexgibson:
```suggestion
    else if (!client.isFirefox) {
```

media/js/firefox/lockwise/lockwise.js
```diff
@@ -0,0 +1,26 @@
...
+        document.querySelector('.for-firefox-69-and-below').remove();
+        document.querySelector('.for-firefox-70-and-above').remove();
+    } else {
+        document.querySelector('.for-firefox-70-and-above').remove();
+        document.querySelector('.for-firefox-69-and-below').remove();
+        document.querySelector('.for-non-firefox-users').remove();
+    }
+
+    document.querySelector('#lockwise-button').addEventListener('click', function() {
```
alexgibson: We should probably only add this event listener if the visitor is on firefox 70+.

In Chrome I see `Uncaught TypeError: Cannot read property 'addEventListener' of null`

bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,105 @@
...
+          <div class="mzp-c-hero-cta add-on-download for-firefox-69-and-below" >
+            <div class="mzp-c-button-download-container">
+              <p >{{ _('Try Lockwise now') }}</p>
+              <a href="https://github.com/mozilla-lockwise/lockwise-addon/releases/download/2.2.5-alpha/lockwise-signed.xpi" class="mzp-c-button mzp-t-product" data-cta-type="Lockwise-Install" data-cta-position="Primary">{{ _('Install for Firefox') }}</a>
+            </div>
+          </div>
+          <div class="mzp-c-hero-cta add-on-download for-firefox-70-and-above" >
+            <div class="mzp-c-button-download-container">
+              <p >{{ _('Try Lockwise now') }}</p>
+              <a class="mzp-c-button mzp-t-product" id="lockwise-button" data-cta-type="Lockwise-Open" data-cta-position="Primary">{{ _('Open in Firefox') }} </a>
```
alexgibson: This should probably be a `<button>` rather than a link without an `href`.

bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,105 @@
...
+          </div>
+          <div class="mzp-c-hero-cta add-on-download for-firefox-69-and-below" >
+            <div class="mzp-c-button-download-container">
+              <p >{{ _('Try Lockwise now') }}</p>
+              <a href="https://github.com/mozilla-lockwise/lockwise-addon/releases/download/2.2.5-alpha/lockwise-signed.xpi" class="mzp-c-button mzp-t-product" data-cta-type="Lockwise-Install" data-cta-position="Primary">{{ _('Install for Firefox') }}</a>
+            </div>
+          </div>
+          <div class="mzp-c-hero-cta add-on-download for-firefox-70-and-above" >
+            <div class="mzp-c-button-download-container">
+              <p >{{ _('Try Lockwise now') }}</p>
```
alexgibson: This is coming through as untranslated when I view it in French. I don't see this string in the .lang file? https://github.com/mozilla-l10n/www.mozilla.org/blob/master/en-US/firefox/products/lockwise.lang

bedrock/firefox/templates/firefox/lockwise/lockwise.html
alexgibson: I also don't see `Open in Firefox` in the .lang file?

bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,105 @@
...
+              <a class="mzp-c-button mzp-t-product" id="lockwise-button" data-cta-type="Lockwise-Open" data-cta-position="Primary">{{ _('Open in Firefox') }} </a>
+            </div>
+          </div>
+          <div class="mzp-c-hero-cta add-on-download for-non-firefox-users" >
+            <div class="mzp-c-button-download-container">
+              <p >{{ _('Only in the Firefox browser') }}</p>
```
alexgibson: This string is also untranslated

media/js/firefox/lockwise/lockwise.js
```diff
@@ -0,0 +1,26 @@
...
+        document.querySelector('.for-non-firefox-users').remove();  
+    }
+    else if ( client._isFirefox() === true && client._getFirefoxMajorVersion() < '70' ) {
+        document.querySelector('.for-firefox-70-and-above').remove();
+        document.querySelector('.for-non-firefox-users').remove();
+    }
+    else if ( client._isFirefox() === false ){
+        document.querySelector('.for-firefox-69-and-below').remove();
+        document.querySelector('.for-firefox-70-and-above').remove();
+    } else {
```
alexgibson: I'm a bit confused about this last conditional, who gets here?

media/css/firefox/lockwise/lockwise.scss
```diff
@@ -0,0 +1,70 @@
...
+    }
+    .mzp-c-hero-image img {
+        margin: 0 $layout-lg;
+    }
+}
+
+@media #{$mq-lg} {
+    .add-on-download {
```
alexgibson: We probably shouldn't be making assumptions about someone's operating system based on screen size. For example on mobile I see only the app store buttons (which feels right), but on a large screen iPad I see app store buttons as well as a download Firefox button.

I'd suggest relying on the `Mozilla.Client` library for platform detection. I think there are properties like `.isDesktop` and `.isMobile` that may help.

alexgibson:
At smaller viewport sizes the product image isn't quite centered, so it bleeds off the edge of the page a little:

lockwise-bleed

bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,105 @@
...
+
+{% block body_id %}firefox-lockwise{% endblock %}
+{% block body_class %}firefox-lockwise{% endblock %}
+
+{% block content %}
+  <main role="main">
+    <section class="mzp-c-hero mzp-has-image">
+      <div class="mzp-l-content">
+        <div class="mzp-c-hero-body">
+          <img src="{{ static('img/firefox/lockwise/lockwise-horizontal.svg') }}" class="lockwise-logo" />
```
alexgibson: If you rebase this branch against master to pick up Protocol 9.0, it contains all the new product wordmarks

bedrock/firefox/templates/firefox/lockwise/lockwise.html
Errolyn: yes there are a few strings that got added after I did the initial string pulling. I'm working with Eric to get it added, but pages won't go live until they are there.

bedrock/firefox/templates/firefox/lockwise/lockwise.html
alexgibson: @rraue pointed out in Slack that this should contain some text for SEO since it's the product name. I'd suggest trying a heading here, with the `image-replaced` mixin (which should replace `<h1>Firefox Lockwise</h1>` with the wordmark image).

bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,105 @@
...
+{% from "macros.html" import google_play_button with context %}
+{% from "macros-protocol.html" import hero, picto_card with context %}
+
+{% extends "firefox/base/base-protocol.html" %}
+
+{% block page_favicon %}{{ static('img/firefox/lockwise/favicon.ico') }}{% endblock %}
+
+{% block page_title %}{{_('Firefox Lockwise — password manager —  take your passwords everywhere')}}{% endblock %}
+
+{% block page_desc %}{{ _('Firefox Lockwise lets you securely access the passwords you’ve saved in Firefox from anywhere — even outside of the browser. Features 256-bit encryption and Face/Touch ID.') }}{% endblock %}
```
stephaniehobson: You can also declare a Lockwise specific social sharing image for this page like this:

```
{% block page_image %}{{ static('/protocol/img/logos/firefox/lockwise/og.png') }}{% endblock %}
```

bedrock/firefox/templates/firefox/lockwise/lockwise.html
Errolyn: I'm not sure how to make this work with hero macro. Should I unwind and not use the hero macro?

media/img/firefox/lockwise/arrows.svg
```diff
@@ -0,0 +1,15 @@
...
+<svg width="37px" height="37px" viewBox="0 0 37 37" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
```
alexgibson: Optimize SVGs here and below https://jakearchibald.github.io/svgomg/

bedrock/firefox/templates/firefox/lockwise/lockwise.html
alexgibson: I linked to an example for how you can do this below in the comments, but there should be no need to rewind the macro. Happy to discuss in person if it helps?

alexgibson:
At smaller viewports the Lockwise wordmark is not center aligned like the rest of the content: 

image

media/js/firefox/lockwise/lockwise.js
```diff
@@ -0,0 +1,27 @@
+(function() {
+    'use strict';
+
+    var client = window.Mozilla.Client;
+    var version = client._getFirefoxMajorVersion();
+
+    if (client.isMobile) {
+        document.querySelector('.for-firefox-70-and-above').remove();
```
alexgibson: This isn't a blocker, and something we could always fix up later - but we may want to consider the edge case of this JS either not running (i.e. in a legacy browser, or a noscript extension), or failing for some unforseen reason. Until this code runs, all the CTA's are visible on the page:

image

An alternative way to approach this might be to hide all of these JS dependent CTA's by default (using CSS), and then to only show them explicitly here by using a CSS class hook.

bedrock/firefox/templates/firefox/lockwise/lockwise.html
Errolyn: I realize I was thinking to firmly inside the box... I can just not add the Lockwise image header to the hero 🤦‍♀

media/css/firefox/lockwise/lockwise.scss
```diff
@@ -0,0 +1,69 @@
...
+    list-style: none;
+}
+.mzp-c-hero-desc {
+    margin-bottom: $spacing-lg;
+}
+.mzp-c-hero-image {
+    padding-top: $spacing-lg;
+}
+.mzp-c-hero-title {
+    background: url('/media/protocol/img/logos/firefox/lockwise/logo-word-hor-2xl-high-res.png') no-repeat;
```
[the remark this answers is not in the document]
Errolyn: I ended up updating the macro to get this working as described.

media/js/firefox/lockwise/lockwise.js
Errolyn: I have given them all the "hidden" class and remove the class from the button via javascript, if the script doesn't run the buttons are hidden.

Errolyn:
Ok, I think I got everything. Should be ready for another look, I updated the demo branch and also rebased. @alexgibson

media/css/firefox/lockwise/lockwise.scss
```diff
@@ -0,0 +1,95 @@
...
+.mzp-c-card-picto{
+    list-style: none;
+}
+
+.mzp-c-hero-desc {
+    margin-bottom: $spacing-lg;
+}
+
+h1 {
+    @include at2x('/media/protocol/img/logos/firefox/lockwise/logo-word-hor-2xl.png', auto, 100%);
```
alexgibson: This image image is a lot bigger than it needs to be. I think you can just use `logo-word-hor-sm.png` here?

media/css/firefox/lockwise/lockwise.scss
```diff
@@ -0,0 +1,95 @@
...
+
+@media #{$mq-sm} {
+    .mobile-download-buttons {
+        margin-bottom: $spacing-lg;
+    }
+}
+
+@media #{$mq-md} {
+    h1 {
+        @include at2x('/media/protocol/img/logos/firefox/lockwise/logo-word-hor-2xl.png', auto, 48px);
```
alexgibson: No need to redeclare the image again here, `@include background-size(auto 48px);` should do it?

bedrock/base/templates/macros-protocol.html
```diff
@@ -27,11 +27,14 @@
...
 
 {# Hero: https://protocol.mozilla.org/patterns/organisms/hero.html #}
-{% macro hero(title, desc=None, class=None, include_cta=False, heading_level=2, image_url=None, include_highres_image=False, image_alt='') -%}
+{% macro hero(title, subheading, desc=None, class=None, include_cta=False, heading_level=2, subheading_level=3, image_url=None, include_highres_image=False, image_alt='') -%}
 <section class="mzp-c-hero{% if class %} {{ class }}{% endif %}{%if image_url%} mzp-has-image{% endif %}">
   <div class="mzp-l-content">
     <div class="mzp-c-hero-body">
       <h{{ heading_level }} class="mzp-c-hero-title">{{ title }}</h{{ heading_level }}>
+      {% if subheading %}
+      <h{{ subheading_level }} class="mzp-c-hero-subheading">{{ subheading }}</h{{ subheading_level }}>
```
alexgibson: `mzp-c-hero-subheading` isn't a class name that exists in the protocol component, so I'd suggest maybe leaving it out for now? We could file an issue to include a sub-heading in the original component if we want to add this.

Another way to fix this without modifying the macro would be include a custom sub heading within the hero description. You can add any kind of arbitrary HTML underneath the hero heading using the macro. An example of how to do this can be [seen here](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/enterprise/index.html#L45-L53).

That said, what you have here does work - whichever way you want to fix this is fine, but I'd suggest at least removing the non-existant clas name.

alexgibson: Doh, I forgot you're already doing a custom hero description (it's early I need more coffee) :)

So in this case you could do the sub heading by adding the following at line 50:

```
  <h2>{{ _('Take your passwords everywhere') }}</h2>
  <p class="mzp-c-hero-desc">{{ _('Securely access the passwords you’ve saved in Firefox from anywhere — even outside of the browser.') }}</p>
```

bedrock/firefox/templates/firefox/lockwise/lockwise.html
```diff
@@ -0,0 +1,95 @@
...
+    {% endif %}
+    {% call hero(
+      title='Firefox Lockwise',
+      subheading=_('Take your passwords everywhere'),
+      desc=_('Securely access the passwords you’ve saved in Firefox from anywhere — even outside of the browser.'),
+      class='mzp-c-hero mzp-has-image lockwise-hero',
+      image_url= _img,
+      include_cta=True,
+      heading_level=1,
+      subheading_level=4,
```
alexgibson: Nit: This should probably be a level 2 as it follows an h1. You may need to set a custom heading size in your CSS to compensate.

media/css/firefox/lockwise/lockwise.scss
```diff
@@ -0,0 +1,95 @@
...
+.mobile-download-buttons {
+    padding-bottom: $spacing-md;
+}
+
+.encryption .mzp-c-card-picto-content::before {
+    background: url('/media/img/firefox/lockwise/arrows.svg') no-repeat;
+}
+
+.touch .mzp-c-card-picto-content::before {
+    background: url('/media/img/firefox/lockwise/fingerprinters.svg') no-repeat;
```
alexgibson: These icons are a little off center:

image

bedrock/base/templates/macros-protocol.html
alexgibson: We discussed this on Slack, and @Errolyn is going to file a protocol issue to incorporate a sub-heading in the hero component. We're going to keep this as-is, with the aim to bring these changes back into Protocol.

media/css/firefox/lockwise/lockwise.scss
```diff
@@ -0,0 +1,102 @@
...
+h1 {
+    @include at2x('/media/protocol/img/logos/firefox/lockwise/logo-word-hor-sm.png', auto, 100%);
+    background-position: center;
+    @include image-replaced;
+    background-repeat: no-repeat;
+    margin-bottom: $spacing-md;
+}
+
+.hero-subheading {
+    font-size: 36px;
```
alexgibson: This heading should probably use a type size on our standard scale. `@include text-display-md;` is the closest.

media/css/firefox/lockwise/lockwise.scss
```diff
@@ -0,0 +1,102 @@
...
+}
+
+.privacy .mzp-c-card-picto-content::before {
+    background: url('/media/img/firefox/lockwise/mask.svg') no-repeat;
+    background-position: center;
+}
+
+.mzp-c-card-picto .mzp-c-card-picto-title {
+    font-weight: normal;
+    font-size: 18px;
```
alexgibson:
```suggestion
    @include text-body-lg;
```

Issue #7682, opened 2019-09-09
Non-English locales still show old footer content


alexgibson:
## Description

The main page footer on bedrock shows different content for English / non-English locales. They should be the same.

### Steps to reproduce

1. Visit https://www.mozilla.org/en-US/
2. Look at the footer content:

![image](https://user-images.githubusercontent.com/400117/64520667-86600580-d2ee-11e9-861e-04424aa56656.png)

3. Visit https://www.mozilla.org/de/
4. Look at the footer content:

![image](https://user-images.githubusercontent.com/400117/64520717-a0014d00-d2ee-11e9-92d0-431a535d2b68.png)

### Expected result

All locales should have the same footer content.

### Actual result

Only English locales get the latest footer content. It looks like we have a hard-coded conditional here:

https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/includes/site-footer.html

I can't see the strings in the new footer in the L10n repo, so it looks like they were never exposed?

### Environment

Any browser

Shivansh2407:
The site-footer.html Must extend the index.html file so as to continue changing according to the dynamic data otherwise it will feel like the data is hard-coded.

Patil2099:
@alexgibson  Can I work on this issue?

alexgibson:
This is now fixed

PR #7689, opened 2019-09-10
Update site-footer.html

Shivansh2407:
The site-footer.html Must extend the index.html file so as to continue changing according to the dynmic data otherwise it will feel like the data is hard-coded.
Fixing issue #7682


alexgibson:
Hi @Shivansh2407 - thanks for this pull request, however I'm afraid this does not fix the particular issue.

In order to fix issue https://github.com/mozilla/bedrock/issues/7682, we need to expose the strings in the English footer for translation, and then remove the non-English footer once that work is complete. This is not something that contributors can do easily unfortunately, as it requires push rights to another GitHub repository.

If you're looking for good first bugs, there's a few here that may be of interest: https://github.com/mozilla/bedrock/labels/Good%20First%20Bug

Thanks again!

Issue #7684, opened 2019-09-09
Update the banners at the top and bottom of the homepage


stephaniehobson:
## Description

Update the banners at the top and bottom of the homepage in all Teir 1 languages.

Assets will be attached to [this Trello card](https://trello.com/c/dK19hesi/538-web-work-coding-new-headers-on-mozorg-in-en-de-and-fr-3) in the NA Firefox Brand team's sprint board.

[Bug 1569237](https://bugzilla.mozilla.org/show_bug.cgi?id=1569237)

---

#### Tasks
- [x] Update all strings
- [x] Remove old assets
- [x] Add new assets
- [x] Tweak typography
- [x] String change on /accounts to support the "family" language change.

stephaniehobson:
On [demo/6](https://bedrock-demo-6.oregon-b.moz.works/)

PR #7695, opened 2019-09-10
String change to support family of products marketing (#7684)

stephaniehobson:
## Description

String change to support family of products marketing

## Issue / Bugzilla link

See #7684 (Corresponds to the change coming to the homepage banners, trying to give a coherent story to the user)


stephaniehobson:
And I'll need someone to extract the strings for me.

Issue #7508, opened 2019-07-31
Build a sandbox for CRO


hoosteeno:
## Description

Implement the sandbox defined in #7505: https://docs.google.com/document/d/1RXzBVUk3LSbyBeJOesPrKcekvA5bOUBiUIiQy7frHgQ/edit?usp=sharing

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Sandbox is built and a) ready to port content to b) ready to send audience to

alexgibson:
The initial pages in the sandbox will be:

- `/en-US/exp/firefox/new/`
- `/de/exp/firefox/new/`

For now we'll be hardcoding the strings for `de`, rather than exposing the strings through pontoon.

alexgibson:
Blocked by https://github.com/mozilla/bedrock/issues/7700

alexgibson:
@hoosteeno I've implemented convert in the PR referenced above, and pushed it to demo here:

https://www-demo3.allizom.org/en-US/exp/firefox/new/

In the convert dashboard I've set up a simple A/B test (in the staging project) that changes the headline. It seems to be working ok and tracking as expected. Please take a look.

Things to note:

- I've implemented loading their script dynamically (async). This means that if the script is slow to load, or unresponsive, our visitors won't be left looking at a blank screen & loading spinner. This can mean that swapping out varation content may be visible during page load, but as long as their CDN is fast I don't think the experience is too jarring. This is a better (imho) alternative to blocking on a third-party resource.

PR #7754, opened 2019-09-19
Make sure jQuery is included if convert switch is set to off (#7508)

alexgibson:
## Description
- Small followup to #7508, to make sure that jQuery is still loaded if the convert swithc is set to off.

## Issue / Bugzilla link
#7508

## Testing
- [ ] When `Dev=False` a JS error should no longer be thrown in the console.


alexgibson:
@amychurchwell small followup to the other PR, r?

PR #7752, opened 2019-09-18
Fix eslint 'no-useless-escape' errors. (#7747)

Shivansh2407:
## Description
Fix eslint 'no-useless-escape' errors. #7747

PR #7770, opened 2019-09-23
Update concerts.js

Shivansh2407:


amychurchwell:
We don't need to open a separate PR for every file, they can be added to your original PR by pushing a commit to the original branch `patch-3.` - #7752

PR #7793, opened 2019-09-26
Remove SECRET_KEY from all .env files

### LINKED ISSUE #7792: Make it obvious that our SECRET_KEY values in the repo are not used in production
We get occasional bug reports about secret key values found in the git repo. They would normally be correctly concerning, but ours are there for good reason though that is not obvious from the file or context. We should add comments and change the values to make it clear that we've added these values to the repo on purpose and that all is well.

pmac:
It turns out we don't need them because we have a default value in settings/base.py. This covers the cases where we need to run management commands during the build and test phases. Real SECRET_KEY values are set in all of our running instances of the site.

Fix #7792


pmac:
To be clear, these values have always been public and have never been used on a server, but removing them avoids confusion and security concerns.

Issue #7798, opened 2019-09-27
"No parser was explicitly specified" warning in `privacy/views.py`


alexgibson:
## Description

When building bedrock locally I noticed a warning in the console. This is probably harmless, but it sounds like it could be fixed easily enough.

### Steps to reproduce

Run `make build`.

### Expected result

No console warning.

### Actual result

```
/app/bedrock/privacy/views.py:27: UserWarning: No parser was explicitly specified, so I'm using the best available HTML parser for this system ("lxml"). This usually isn't a problem, but if you run this code on another system, or in a different virtual environment, it may use a different parser and behave differently.

The code that caused this warning is on line 27 of the file /app/bedrock/privacy/views.py. To get rid of this warning, pass the additional argument 'features="lxml"' to the BeautifulSoup constructor.

  soup = BeautifulSoup(content)
```

### Environment

Bedrock local Docker image.

imskr:
I would like to work on this issue

imskr:
@alexgibson This is fixed!

PR #7812, opened 2019-09-30
"No parser was explicitly specified" warning Fixed

imskr:
## Description
fixed beautifulSoup Error warning
## Issue / Bugzilla link
#7798

### DIFF
--- a/bedrock/privacy/views.py
+++ b/bedrock/privacy/views.py
@@ -24,7 +24,7 @@ def process_legal_doc(content):
 
     :param content: HTML Content of the legal doc.
     """
-    soup = BeautifulSoup(content, 'html.parser')
+    soup = BeautifulSoup(content, "lxml")
 
     # Convert the site's full URLs to absolute paths
     for link in soup.find_all(href=HREF_PATTERN):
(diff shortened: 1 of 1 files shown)


imskr:
@alexgibson Please review the PR. I have made the changes.

alexgibson:
When I run this change locally and try to view /privacy/firefox/ the pase errors with:

```
jinja2.exceptions.UndefinedError: 'None' has no attribute 'section'
```

alexgibson:
The value you've set here is a different one to that suggested in the warning. Any ideas on this one @pmac? (is this a non-issue, or a bug worth fixing?)

imskr:
@alexgibson It should fix the issue I guess according to here: https://stackoverflow.com/questions/33511544/how-to-get-rid-of-beautifulsoup-user-warning

alexgibson:
@imskr rather than the suggested stackoverflow solution, have you tried using the parser suggested in the warning message that's linked in the issue?

pmac:
I think we should go with `lxml` as the parser since that's what we're using now, it's already installed, and known to be working. We could experiment with `html5lib` which also is installed, but we should do that in a separate issue.

imskr:
@pmac Is it good now?

PR #7827, opened 2019-10-03
Changed url() to re_path() for Django 2.2 updates

thewolfcommander:
## Description
I have changed url() function to re_path() function which is more better for Django 2.2 version and also have changed imported modules from django.conf.urls to django.urls

### DIFF
--- a/bedrock/urls.py
+++ b/bedrock/urls.py
@@ -4,7 +4,7 @@
 
 from django.conf import settings
 from django.conf.urls import handler404
-from django.urls import include, re_path
+from django.urls import include, path
 
 from django.utils.module_loading import import_string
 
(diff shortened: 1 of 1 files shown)


bedrock/urls.py
```diff
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 from django.conf import settings
-from django.conf.urls import handler404, include, url
+from django.urls import handler404, include, re_path
```
amychurchwell: Tests are failing because it appears that `handler404` is not available to be imported from `django.urls`

`ImportError: cannot import name 'handler404' from 'django.urls' (/venv/lib/python3.7/site-packages/django/urls/__init__.py)`

amychurchwell:
@pmac [at a glance](https://docs.djangoproject.com/en/2.2/ref/urls/#url), it seems like `url()` will be deprecated at some point. Do you have thoughts about us moving to `re_path()`? I'm not sure if there are other dev tasks that need to happen. 🤔

pmac:
It is true that `path()` and `re_path()` have replaced `url()` as the preferred functions in Django. I would say that most of the lines in this urls file in particular could just be `path()` calls and don't need any regex at all. Also there are a lot of urls.py files in the project that could use this kind of cleanup I believe. I'm all for it, but let's see if we can make it even more simple.

I'm fine with merging this change for just this file, especially since this is the first contribution from @thewolfcommander (thank you!), but we do need to fix the import issue and I would like to see an attempt to move to mostly `path()` calls. We can file a followup issue for the rest of the urls.py files.

thewolfcommander:
Thank you all. Sorry for the import issue. Let me fix the issue and I will make changes as soon as possible related to the urls which do not need regex.

pmac:
The test failures are flake8 complaining about the indentation on lines 38 and 40. The `name=` needs to start at the same column as the space after `(` on the above line, like it was before. It's now mismatched since the new function name is longer. 

Also we'd like you to try to convert these to `path` calls instead of `re_path` since they're all simple base urls.

bedrock/urls.py
```diff
@@ -17,25 +18,25 @@
...
+    path('etc/', include('bedrock.etc.urls')),
+    path('exp/', include('bedrock.exp.urls')),
+
+    path('healthz/', watchman_views.ping, name="watchman.ping"),
+    path('readiness/', watchman_views.status, name="watchman.status"),
+    path('healthz-cron/', base_views.cron_health_check),
+    path('csp-violation-capture', base_views.csp_violation_capture,
         name='csp-violation-capture'),
-    url(r'^country-code\.json$', base_views.geolocate,
+    path('country-code\.json', base_views.geolocate,
```
pmac: Need to remove the `\` from this one. Only needed for the regex version.

bedrock/urls.py
```diff
@@ -17,25 +18,25 @@
...
+    path('', include('bedrock.mozorg.urls')),
+    path('', include('bedrock.newsletter.urls')),
+    path('etc/', include('bedrock.etc.urls')),
+    path('exp/', include('bedrock.exp.urls')),
+
+    path('healthz/', watchman_views.ping, name="watchman.ping"),
+    path('readiness/', watchman_views.status, name="watchman.status"),
+    path('healthz-cron/', base_views.cron_health_check),
+    path('csp-violation-capture', base_views.csp_violation_capture,
         name='csp-violation-capture'),
```
pmac: Sorry. Flake8 is very picky. The indentation of this line needs to line up the `n` under the `'` character from the line above. Same with line 40 below.

bedrock/urls.py
```diff
@@ -12,39 +15,36 @@
...
     def show404(request):
         return import_string(handler404)(request, '')
 
-    urlpatterns += (
-        url(r'^404/$', show404),
-        url(r'^500/$', import_string(handler500)),
-    )
+
+    urlpatterns += (path('404/', show404), path('500/',
+                    import_string(handler500)))
```
pmac: I'd much rather keep the formatting for these lists like the following:

```suggestion
    urlpatterns += (
        path('404/', show404), 
        path('500/', import_string(handler500)),
    )
```

amychurchwell:
Is this still being worked on?

alexgibson:
Closing this due to inactivity - please feel free to reopen if you have time finish up what's here. Thanks

Issue #7826, opened 2019-10-03
Firefox Account referral docs list incorrect parameters for FxA signup form


alexgibson:
## Description

The table of accepted parameters for the `fxa_email_form` macro are incorrect. It contains rows for `cta_type` and `cta_position`, which belong to the `monitor_button` macro further down the page.

### Steps to reproduce

1. Visit https://bedrock.readthedocs.io/en/latest/firefox-accounts.html#configuration
2. Observe the table rows for `cta_type` and `cta_position`.
3. View the available [macro parameters](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/macros.html#L221).

### Expected result

I believe these two table rows belong to the monitor button macro, so should be in this table: https://bedrock.readthedocs.io/en/latest/firefox-accounts.html#id5

### Actual result

They are in the wrong table.

### Environment

N/A

imskr:
Should I remove `cta_type` and `cta_position` rows from https://bedrock.readthedocs.io/en/latest/firefox-accounts.html#configuration and add it to https://bedrock.readthedocs.io/en/latest/firefox-accounts.html#id5

alexgibson:
That’s correct

imskr:
@alexgibson I've made the PR

PR #7828, opened 2019-10-04
added cta_type  and cta_position in correct table

imskr:
## Description
Added cta_type and cta_position rows in correct table
## Issue / Bugzilla link
#7826


imskr:
Sorry for that @alexgibson I've fixed it

Issue #7742, opened 2019-09-17
WNP 70 Instrumentation Plan - Analytics


ejregithub:
***Tasks***
- [ ] Create instrumentation plan
- [ ] Work with dev for including instrumentation, tags in the coded page, in #7677

sghoseWI:
WNP 70 Analytics Instrumentation:
On ‘Sign Up for Monitor’ button:
data-cta-type=“whatsnew70-Monitor”
data-cta-position="Primary"

On SUMO/Lockwise/Release Notes links
utm_source=www.mozilla.org-whatsnew70
utm_campaign=whatsnew70
utm_medium=referral

sghoseWI:
Conventions
-----------

When choosing URL parameter values, the following conventions help support uniformity in code and predictability in retroactive analysis.

* Use lower case characters in parameter values.
* Separate words in parameter values with hyphens.
* Follow parameter naming patterns established in previous iterations of a page.

hoosteeno:
@sghoseWI can you update with the latest here?

sghoseWI:
@hoosteeno Done (edited previous comment above to avoid confusion) - thanks!

Issue #7677, opened 2019-09-06
WNP 70 code


ejregithub:
## Description

Code the WNP 70 page.
Epic - https://github.com/mozilla/bedrock/issues/7678

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Page is live in Production October 11

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] This is an example of a risk. Use a checklist!
- [ ] This is an example of a risk.

#### Tasks
- [ ] Code it

hoosteeno:
Like the protection report page, this page will have 3 states to determine the CTA:

For Firefox 70+: See what Firefox has blocked for you (uses [a new uiTour function](https://bugzilla.mozilla.org/show_bug.cgi?id=1577031) to open a newtab)
For Firefox <70: [Update your Firefox browser](https://support.mozilla.org/kb/update-firefox-latest-release)
For non-Fx: [Download the Firefox browser](https://www.mozilla.org/firefox/download/thanks/) 

The first two of these should be instrumented with our Generic CTA datalayer attributes. The third is instrumented already. @sghoseWI do you agree about this? Do you have ideas for the actual attribute values?

amychurchwell:
![image](https://user-images.githubusercontent.com/30483988/65799997-442a2700-e12a-11e9-88d6-6d4a6e6686f9.png)
(design for my quick reference.)

hoosteeno:
We need localized strings in fr/de. @ejregithub can you help us keep this moving?

ejregithub:
@hoosteeno - yes, the card in Airtable is for just that  - https://airtable.com/tbl05XxAo8cjqzcvY/viw9kla3Mmk5qMErn/recCbq6Mccece15dL  - I've added you to the Slack channel #marketing-l10n-requests, where you can see the same request to Anika and Malcolm that I posted yesterday (Sunday).

amychurchwell:
instrumentation for reference - https://github.com/mozilla/bedrock/issues/7742

cgonz0:
@amychurchwell  **Design feedback below**
UI is looking awesome - please let me know if you have any questions or need me to clear anything up.
___________________________________

**Conditional content -- Signed in/Signed out of Firefox account:**

- Currently, for both: 
_Have you checked out Firefox Monitor? It’s included with your Firefox account. 
Button: “Sign Up for Monitor”_

- Should be:
[...]

PR #7829, opened 2019-10-04
Add whatsnew page for Firefox 70 (#7677)

amychurchwell:
## Description

Adds a whatsnew page for Firefox 70, with locale-specific templates for English, French, and German.

⚠️ **This page needs to go to production on October 10th.**

## Issue / Bugzilla link
#7677

## Testing
➡️ [https://bedrock-demo-achurchwell.oregon-b.moz.works/firefox/70.0/whatsnew/](https://bedrock-demo-achurchwell.oregon-b.moz.works/firefox/70.0/whatsnew/)

- [x] **English, French, and German** locales receive templates with [hard-coded strings](https://docs.google.com/spreadsheets/d/1P_dICdX2eoxxZI3VK3pbxsLY1r4YneOi4yn3zqM-n_M/edit?ts=5d931779#gid=757298539) & **monitor CTA**.
- [x] All other locales have the **protection report** (`about:protections`) as main CTA.
- [x] [Follows instrumentation format.](#7742)
- [x] All JS-dependent CTAs have `href` fallbacks with correct utm params.

**Protection Report CTA** receives the following states:
- For Firefox 70+: See what Firefox has blocked for you. `Mozilla.UITour.showProtectionReport()`
- For Firefox <70: Update your Firefox browser.
- For non-Fx: Download the Firefox browser.


bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx70-de.html
```diff
@@ -0,0 +1,114 @@
...
+
+          <h3 class="c-picto-block-title">{{ _('Lass über 2000 Arten von Online-Trackern einfach blockieren – standardmäßig.') }}</h3>
+          <div class="c-picto-block-body">
+            <p>{{ _('Alle Firefox Browser für Desktop und Mobile kommen mit standardmäßig eingestelltem verbesserten Tracking-Schutz.') }}</p>
+            <a class="mzp-c-cta-link show-firefox-desktop-70 js-open-about-protections" href="https://support.mozilla.org/kb/enhanced-tracking-protection-firefox-desktop/?utm_source={{ _entrypoint }}&utm_medium=referral&utm_campaign={{ _utm_campaign }}&entrypoint={{ _entrypoint }}">{{ _('Sehen, was Firefox für dich geblockt hat') }}</a>
+
+            <a class="mzp-c-cta-link show-firefox-desktop-old" href="https://support.mozilla.org/kb/update-firefox-latest-release/?utm_source={{ _entrypoint }}&utm_medium=referral&utm_campaign={{ _utm_campaign }}&entrypoint={{ _entrypoint }}">{{ _('Firefox Browser updaten') }}</a>
+
+            <div class="show-default">
+              {{ download_firefox(alt_copy=_('Lad dir den Firefox Browser herunter'), download_location='primary cta', button_color='mzp-c-button mzp-t-primary mzp-t-product') }}
```
alexgibson: Should this be `download_location='secondary cta'`? (same for other locale-specific templates).

media/js/firefox/whatsnew/whatsnew-70.js
```diff
@@ -0,0 +1,59 @@
...
+            document.querySelector('main').classList.add('state-firefox-desktop-70');
+
+            // Intercept link clicks to open about:protections page using UITour.
+            Mozilla.UITour.ping(function() {
+                var protectionReportLinks = document.querySelectorAll('.js-open-about-protections');
+                var lockwiseLinks = document.querySelectorAll('.js-open-lockwise');
+
+                // For UI Tour CTAs, we will treat these more as an action rather than a page navigation.
+                for (var i = 0; i < protectionReportLinks.length; i++) {
+                    protectionReportLinks[i].setAttribute('role', 'button');
```
alexgibson: I flip-flopped on this multiple times when working on the privacy/products page, but in the end I decided to omit `role="button"` here. 

I think I agree with your original thought that this does feel more like a destination, rather than an action. It just happens to be an in-product page, instead of one on the www.

One thing I did do to help solidify this was to add `protectionReportLinks[i].href = 'about:protections';`, which overwrites the existing sumo links, and gives better context to where the link now points to. Worth adding here instead?

media/static-bundles.json
```diff
@@ -1074,6 +1080,17 @@
       ],
       "name": "firefox_whatsnew_69"
     },
+    {
+      "files": [
+        "js/base/uitour-lib.js",
+        "js/base/mozilla-fxa.js",
+        "js/base/mozilla-fxa-init.js",
+        "js/base/mozilla-monitor-button.js",
```
alexgibson: Need to add `mozilla-monitor-button-init.js` here too?

Is it ok for the minotor button Js to be run from all versions of this page, or should it only be from the templates that feature a sign-in button?

amychurchwell:
**String change**

`2000+ types of trackers blocked — automatically`
to `2000+ trackers blocked — automatically`

**German**: `Wir blocken über 2.000 Tracker – automatisch`

media/static-bundles.json
amychurchwell: @alexgibson do you think that there is an excessive performance trade-off for including it on all versions? I didn't originally plan to make separate bundles, but I could.

alexgibson: I wouldn't worry too much about performance, as the files are pretty small and will be minified. I was more wondering if redundant metrics flow requests could somehow skew some analytics somewhere on the FxA side? It may be fine though, I don't really know.

amychurchwell: oooh, excellent point. I am not sure either, so let's separate it just to be safe.

media/js/firefox/whatsnew/whatsnew-70.js
amychurchwell: Sounds good. I think I will keep the current logic for the lockwise links because that does feel like more of an action vs. a destination.

amychurchwell:
Adding blocked label while we wait on the new French string.

amychurchwell:
instrumentation reference - https://github.com/mozilla/bedrock/issues/7742#issuecomment-535716469

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx70-en.html
```diff
@@ -0,0 +1,114 @@
...
+    </div>
+  </header>
+
+  <section class="content-main">
+    <div class="mzp-l-content">
+      <div class="mzp-c-box-emphasis">
+        <img class="c-emphasis-box-logo" src="{{ static('img/logos/firefox/logo-monitor.svg') }}" alt="">
+        <h2 class="c-emphasis-box-title">
+          {{ _('Have you checked out Firefox Monitor?') }}
+          <span class="show-fxa-supported-signed-in">{{ _('It’s included with your Firefox account.') }}</span>
```
amychurchwell: @craigcook the line "It's included with your Firefox account." is being shown to all fx users. Other places I use `show-fxa-supported-signed-in` this is not the case, is there something about this nesting that is voiding that css hook?

image

media/css/firefox/whatsnew/whatsnew-70.scss
```diff
@@ -0,0 +1,212 @@
...
+        display: block;
+        margin: 0 auto $spacing-lg;
+        width: 50px;
+    }
+
+    .c-emphasis-box-title {
+        @include text-display-sm;
+
+        span {
+            display: inline;
```
craigcook: That "It’s included with your Firefox account." bit is showing for all users because of this rule. You might need more specificity on a hiding rule, or maybe this rule isn't needed?

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx70-en.html
craigcook: Added a comment on the specific line. There's a more specific rule setting this to display.

`.mzp-c-box-emphasis .c-emphasis-box-title span {
    display: inline;
}`

media/css/firefox/whatsnew/whatsnew-70.scss
amychurchwell: I included the `inline` property because `show-fxa-supported-signed-in` adds a class of `display: block` which separates the span on a different line. 
writing:
```
body.state-fxa-supported-signed-in {
    span.show-fxa-supported-signed-in {
        display: inline;
    }
}
```
for more specificity.

Issue #7825, opened 2019-10-03
fxa_cta_link macro has empty value for `data-cta-position` attribute


alexgibson:
## Description

The [`fxa_cta_link`](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/macros.html#L185-L188) macro has a redundant / non-used `data-cta-position` attribute. It's not used anywhere (since it is blank everywhere), so we should either remove it, or implement it properly and document it.

### Steps to reproduce

1. https://www.mozilla.org/en-US/firefox/66.0/whatsnew/all/
2. View source on the "Create a Firefox Account" button.

### Expected result

There should be a value like `data-cta-position="primary"` in the markup.

### Actual result

The attribute exists, but is empty:

```
<a href="https://accounts.firefox.com/signup?service=sync&amp;entrypoint=mozilla.org-whatsnew66&amp;context=fx_desktop_v3&amp;utm_source=mozilla.org-whatsnew66&amp;utm_medium=referral&amp;utm_campaign=wnp66&amp;utm_content=accounts-page-top-cta" data-mozillaonline-link="https://accounts.firefox.com.cn/signup?service=sync&amp;entrypoint=mozilla.org-whatsnew66&amp;context=fx_desktop_v3&amp;utm_source=mozilla.org-whatsnew66&amp;utm_medium=referral&amp;utm_campaign=wnp66&amp;utm_content=accounts-page-top-cta" class="js-fxa-cta-link mzp-c-button mzp-t-product" data-cta-text="Create account" data-cta-type="FxA-Sync" data-cta-position="">
  Create a Firefox Account
</a>
```

### Environment

N/A

Shivansh2407:
Can I please work on this issue ? Please let me know the File name to be changed here (if possible)..

alexgibson:
Depends on https://github.com/mozilla/bedrock/issues/8185

alexgibson:
Fixed in https://github.com/mozilla/bedrock/issues/8185

PR #7830, opened 2019-10-05
Update macros.html

Shivansh2407:
## Description
fxa_cta_link macro has empty value for `data-cta-position` attribute #7825


alexgibson:
@Shivansh2407 I'm afraid this does not fix the problem described in the issue. There isn't a clear path to resolution yet, as we stil need to work out what the most appropriate solution should be, so I'm going to close this PR. Thanks for the interest though.

PR #7806, opened 2019-09-27
Add standard sized & named quantum logos, update to use them (#7656)

stephaniehobson:
## Description

In preparation for the upcoming logo update this PR:
- Adds files for the current browser logo & wordmark that are the same size and file name as the files with the new logo
- Replaces updates current release logos and wordmarks to use those files
  - Except favicons, those will be handled separately

## Issue / Bugzilla link

See #7656

## Testing

- Any 404s?
- Did I delete anything still in use?
- Did I miss anything that needs updates?

---

- http://localhost:8000/ar/firefox/accounts/
- http://localhost:8000/de/firefox/campaign/?xv=berlin
- http://localhost:8000/de/firefox/campaign/?xv=privatsphare
- http://localhost:8000/de/firefox/new/?xv=berlin
- http://localhost:8000/en-US/
- http://localhost:8000/en-US/about/legal/firefox/
- http://localhost:8000/en-US/firefox/
- http://localhost:8000/en-US/firefox/40.0/firstrun/
- http://localhost:8000/en-US/firefox/63.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/64.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/65.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/66.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/67.0.5/whatsnew/all/
- http://localhost:8000/en-US/firefox/67.0.5/whatsnew/all/
- http://localhost:8000/en-US/firefox/67.0/whatsnew/
- http://localhost:8000/en-US/firefox/67.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/67.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/68.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/68.0/whatsnew/all/?v=b
- http://localhost:8000/en-US/firefox/68.0/whatsnew/all/?v=c
- http://localhost:8000/en-US/firefox/69.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/accounts/
- http://localhost:8000/en-US/firefox/accounts/
- http://localhost:8000/en-US/firefox/best-browser/
- http://localhost:8000/en-US/firefox/campaign/
- http://localhost:8000/en-US/firefox/campaign/
- http://localhost:8000/en-US/firefox/campaign/?xv=betterbrowser
- http://localhost:8000/en-US/firefox/campaign/?xv=edge
[...]

Issue #7656, opened 2019-09-04
Update Browser logo for Fx70 (2019-10-22) 


stephaniehobson:
## Description

The new browser logo will launch for "release" with Firefox 70 on Oct 22. We should be ready to publish the new logo before that.

---

#### Tasks
- [x] Inventory current logo use
- [x] Determine what sizes and formats are needed for each asset
  - [x] Sizes https://github.com/mozilla/protocol-assets/issues/32
  - [x] Formats https://github.com/mozilla/protocol-assets/issues/27
- [x] Add sizes and formats to protocol-assets https://github.com/mozilla/protocol-assets/pull/36
- [x] Update Protocol to consume new version of protocol-assets https://github.com/mozilla/protocol/pull/491
- [x] Update bedrock to consume the new version of Protocol #7789 
- [x] Update bedrock to remove old logos and use new ones https://github.com/mozilla/bedrock/issues/7956 ~Blocked on #7863~
- [x] Update structured data to reference logos
- [x] generate favicons with new logo #7889 
- [x] update bedrock to use new favicons #7915

stephaniehobson:
Product's logo inventory:
https://docs.google.com/spreadsheets/d/1hbHhihkm8za4diQ_uaTEvNcd2jWeKlSU7X2_pYtnZcM/edit#gid=0

Bedrock specific logo inventory:
https://docs.google.com/spreadsheets/d/1F-__ZuJIvKGkjsp_vOWiSXSJr5z3xltt3gRXKnLoF3A/edit#gid=0

PR #7848, opened 2019-10-08
[#7656] Use backgrounds for accounts page logos

craigcook:
## Description

The accounts page was changed in #7806 to use inline images from Protocol with the `high_res_img()` helper, but that helper rewrites part of the file path so it ends up being a 404, even though it works locally.

Until we can refactor the helper to handle Protocol assets, a quick fix is to use image replacement for these logos (the CSS gets the path right).


stephaniehobson:
R+ 🔼

PR #7849, opened 2019-10-08
Add Heroku support for Review apps

pmac:
With the decommissioning of Deis Workflow our method of creating a demo instance using a named branch will be going away. This is an attempt to use [Heroku Review Apps](https://devcenter.heroku.com/articles/github-integration-review-apps) to replace that. Once this merges I can enable Review Apps for our new `www-demo` Heroku Pipeline and we can try it out. It should create a new deployment for every pull-request and update the PR with a link to it.

Will fix https://github.com/mozmeao/infra/issues/1164

### DIFF
--- a/bin/sync-all.sh
+++ b/bin/sync-all.sh
@@ -6,6 +6,13 @@ if [ ! -e ./manage.py ]; then
     cd $script_parent_dir
 fi
 
+# for heroku demos
+# heroku doesn't pull git submodules
+if [[ ! -d ./vendor-local/src/legal-docs/firefox_privacy_notice ]]; then
+    rm -rf ./vendor-local/src/legal-docs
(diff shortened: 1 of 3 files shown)


metadave:
:+1:

bin/sync-all.sh
```diff
@@ -6,6 +6,12 @@ if [ ! -e ./manage.py ]; then
     cd $script_parent_dir
 fi
 
+# for heroku demos
```
pmac: Note: this is here because Heroku doesn't currently support fetching submodules for automated deployments from github. I think this is fine for demos, and we want to move away from submodules anyway.

pmac:
Thanks y'all. I'm going to go ahead and merge this then and see how it goes.

PR #7865, opened 2019-10-09
Add site wide feature detect snippet (Fixes #7391)

### LINKED ISSUE #7391: Move to a feature-detection system for delivering 1st class browser support
## Description

We usually do a good job of making sure to deliver usable experiences to people both with and without JS enabled, but handling scenarios where JS fails to load is a much more common problem on most websites today. We often resort to using feature detection for this type of problem, but it's nearly always done on a page-by-page basis, or at a component level. 

If we moved toward using a global `cutsTheMustard()` type feature detect (real name TBD), and only ran page-specific JS if the function passes, then we could much more confidently ship code without worrying that we're breaking the site in old browsers.

My initial idea was a function that runs a callback only if the feature detect returns true, but there may be other ways to do this too.

---

- [x] Deliver feature detection on a single page (e.g. /firefox/all)
- [ ] Document the feature detection snippet and incorporate into norms going forward

- [ ] Where should this live - should it be in the `<head>`, so we can leverage it for CSS class hooks, or should it be in the common bundle only? There are likely tradeoffs for each.

#### Tasks
- [ ] Create a feature detect. Ideally it should **return false for IE10, but true for IE11 and above**.
- [x] Decide where to implement it and how to should be used (see risks above).
- [x] Update the page-specific JS on /firefox/all/ to use the new feature detect.
- [ ] Document the feature detect and how it should be used on our [browser support page](https://bedrock.readthedocs.io/en/latest/browser-support.html).

alexgibson:
## Description
- Add site wide feature detection snippet (targeting IE11+).
- Update browser support docs.
- Use said feature detect on `/firefox/all/`.

## Issue / Bugzilla link
#7391

## Testing
Demo: https://bedrock-demo-agibson.oregon-b.moz.works/en-US/firefox/all/

- [x] firefox/all JS should load for IE11+
- [x] firefox/all should degrade to CSS only layout for IE10.

Successful test run: https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/388/pipeline


alexgibson:
Ok, I think this is ready for review.

stephaniehobson:
How do you see this working with `is-js-dependent`? And do you think we can consolidate the approach some how? If not can you please document when we should one vs the other approach.

I'm sorry this review has sat for a while. Can you push the updated branch to your demo please? (I'd like to test the firefox/privacy/products page)

stephaniehobson:
Working as expected in Edge, IE 11, Firefox ESR 24.

alexgibson:
> How do you see this working with is-js-dependent? And do you think we can consolidate the approach some how? If not can you please document when we should use one vs the other approach.

Good question - to make this less confusing, I've renamed `is-js-dependent` to `hide-from-legacy-ie`. I've also added a note in the docs about it.

> I'm sorry this review has sat for a while. Can you push the updated branch to your demo please? (I'd like to test the firefox/privacy/products page)

Done, although as this PR suggests this change is only implemented on /firefox/all/ to try and keep the scope of this PR small.

docs/browser-support.rst
```diff
@@ -17,42 +22,105 @@ We seek to provide usable experiences of our most important web content to all u
...
 **The following browsers have enhanced support:**
 
   * All evergreen browsers (Firefox, Chrome, Safari, Edge, Opera, etc.)
-  * IE10 and above.
+  * IE11 and above.
 
 **The following browsers have basic support:**
 
-  * All other IE browsers.
+  * Outdated evergreen browser versions.
```
stephaniehobson: I think we need to add a 3rd class here. IE10 gets basic JS support but enhanced CSS support.

media/css/firefox/privacy/products.scss
```diff
@@ -239,7 +239,9 @@ $image-path: '/media/protocol/img';
 /* -------------------------------------------------------------------------- */
 // Conditional CTA Hooks
 
-.is-js-dependent {
+.show-firefox-desktop-70,
```
stephaniehobson: Thanks for making this clearer :)

docs/browser-support.rst
alexgibson: I had hoped that definition of what basic support consists of is explained in enough detail below that we don’t necessarily need a 3rd class here. Even though IE10 is still served the same stylesheet, it doesn’t mean it gets enhanced CSS support by default. Using feature detection to deliver a degraded experience where needed is the most suitable method here.

I guess it depends if we think of basic support as a binary thing or not. It could mean “serving a universal stylesheet”, or it could mean “serving a degraded experience”.

I’m open to what others think makes the most sense.

media/js/base/site.js
```diff
@@ -135,6 +135,11 @@
             return 32;
         },
 
+        // Universal feature detect to deliver graded browser support (targets IE 11 and above).
+        cutsTheMustard: function () {
+            return 'classList' in document.createElement('div') && 'MutationObserver' in window;
```
stephaniehobson: There is no mustard emoji. I am heartbroken.

media/js/base/site.js
```diff
@@ -192,6 +197,13 @@
             h.className += ' is-firefox';
         }
 
+        // Add class to reflect browsers that get 1st class JS & CSS support.
+        var isModernBrowser = window.site.isModernBrowser = window.site.cutsTheMustard();
+
+        if (isModernBrowser) {
+            h.className += ' is-modern-browser';
```
stephaniehobson: You can use `classList.add('is-modern-browser');` if it cuts the mustard!

stephaniehobson: (not a blocker)

media/js/firefox/all/all-downloads-unified-init.js
```diff
@@ -5,47 +5,42 @@
...
+
+        // Browser help modal.
+        browserHelpIcon.addEventListener('click', function(e) {
             e.preventDefault();
-            showHelpModal.call(this, installerHelpContent, e.target.textContent, 'Get Installer Help');
+            showHelpModal.call(this, browserHelpContent, browserHelpIcon.textContent, 'Get Browser Help');
         }, false);
+
+        // Installer help modal.
+        for (var i = 0; i < installerHelpIcon.length; i++) {
```
stephaniehobson: Can't wait until we drop support for IE11.

docs/browser-support.rst
stephaniehobson: In my mind we have 3 types of browsers:

- those who get JS & CSS
- those which get CSS but not JS
- those that do not get CSS or JS

I think the most clear thing would be to be explicit about what each browser gets. But maybe I am thinking about this from a different angle?

stephaniehobson: Or maybe we have two categories:

- does not get JS or CSS
- gets JS and CSS but only if they pass the feature tests.

In which case, IE 10 still belongs in the enhanced support catagory.

alexgibson: > In my mind we have 3 types of browsers:

I think what we have is more in terms of:

- those who get 1st class JS & CSS.
- those who get a gracefully degraded experience.
- those who get a universal CSS / JS bundle.

I added a commit to update the docs that hopefully makes this a little clearer. Let me know what you think.

media/js/base/site.js
alexgibson: I'm going to try and resist doing this. Even though it's fine to use `classList` in this one place, unless someone follows the code carefully it could be easy to assume the API is ok to use in the rest of this file.

media/js/firefox/all/all-downloads-unified-init.js
alexgibson: I hear there's a rumor that MS might roll out the new Edge browser to replace IE11, since it features an IE11 mode (whatever that is). If so, dropping IE11 may come sooner than we hope :)

stephaniehobson:
The doc change helps make it clearer in my mind :) Thanks!

PR #7873, opened 2019-10-10
Add ADR: Use Convert and Cloudflare Workers for multi-variant testing

alexgibson:
@hoosteeno r?


docs/architecture/decisions/0003-use-convert-and-cloudflare-workers-for-multi-variant-testing.md
```diff
@@ -0,0 +1,23 @@
...
+
+We would like to make our end-to-end experimentation process faster, with increased capacity, whilst also minimizing the performance impact and volume of code churn related to experiments running on our most important web pages.
+
+## Decision
+
+We will implement a (vetted and approved) third-party experimentation tool called [Convert](https://www.convert.com/) for use on standalone, experimental pages only. We will then use [Cloudflare Workers](https://www.cloudflare.com/en-gb/products/cloudflare-workers/) to redirect a small percentage of traffic to the experimental pages. The worker code will live in the [www-workers](https://github.com/mozmeao/www-workers) repository.
```
jgmize: It might make more sense to put the statement about using Cloudflare Workers in front of the statement about using Convert to match the order of operations for each request. Perhaps something like the following:

> We will use [Cloudflare Workers](https://www.cloudflare.com/en-gb/products/cloudflare-workers/) to redirect a small percentage of traffic to the experimental pages. The worker code will live in the [www-workers](https://github.com/mozmeao/www-workers) repository. We will implement a (vetted and approved) third-party experimentation tool called [Convert](https://www.convert.com/) for use on the pages the Cloudflare Workers redirect to.

It would also be helpful to add some more context to the `(vetted and approved)` by adding a link here or a statement about the vetting process involving our Privacy team in the Context section above.

alexgibson: @hoosteeno any opinions on what is the most suitable document, or statement, that we could include here, which would give some additional context as to the process that was undertaken for the review of Convert?

alexgibson: I've updated this with a link to the (Mozilla confidential) RRA bug for using Convert.

PR #7894, opened 2019-10-14
Decommission /contribute/events

### LINKED ISSUE #6461: Decommission `/contribute/events/` page
The "Events" link under About -> Contribute in the new navigation now points to https://events.mozilla.org and not `/contribute/events/`.

We should decide what we want to do with the [old events page](https://bedrock-demo-agibson.oregon-b.moz.works/en-US/contribute/events/). Is there a good reason to keep it around, or should we retire it and redirect to the new events site?

Patil2099:
Fixes #6461


alexgibson:
As per discussion in the issue, we can’t decom this page until the community portal site is ready. That should be sometime this quarter, but I don’t think it’s live yet.

PR #7896, opened 2019-10-14
Made .scss Equivalent 

### LINKED ISSUE #7449: Make .scss equivalent of mozilla-modal.less
## Description

There are several page bundles in bedrock that are using Pebbles (.scss) for page styles, but are importing`mozilla-modal.less`.

This is a relatively small file. If we make .scss equivalents then it become much easier to remove as a dependency.

---

- [ ] Pages use scss equivalent and no longer have a .less dependency.

#### Tasks
- [ ] Convert file to have .scss equivalent.
- [ ] Update bundles to use .scss versions

Patil2099:
Fixes #7449

### DIFF
--- a/media/css/base/mozilla-accordion.less
+++ b/media/css/base/mozilla-accordion.less
@@ -1,0 +1,84 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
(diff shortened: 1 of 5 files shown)


Patil2099:
@alexgibson Can you explain to me how I have to do it?
Thanks!

alexgibson:
I'm sorry, but I don't currently have enough free time to guide you through the intricacies of our frameworks, which aren't so easy to navigate unfortunately. If you're unsure about how to proceed with this, please close the pull request and take a look for something else. Thanks!

PR #7902, opened 2019-10-15
Fix statsd for bedrock

pmac:
No data is flowing to datadog from bedrock instances. This implements the basket setup of statsd which is working. We'll need to set the statsd client in settings to `django_statsd.clients.normal`.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -5,6 +5,8 @@
 
 import json
 import platform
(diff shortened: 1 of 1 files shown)


pmac:
@jgmize do you know of a reason why we might have not wanted statsd to work for bedrock? I'd like to fix it so that I can collect some newsletter numbers for some people who are asking.

Issue #7904, opened 2019-10-15
Untranslated string in site footer


alexgibson:
## Description

There's an untranslated string in the new footer. It's currently visible in the German locale.

![image](https://user-images.githubusercontent.com/400117/66817490-b163f780-ef33-11e9-86fb-2c6df7c611b8.png)

### Steps to reproduce

1. Visit https://www.mozilla.org/de/firefox/
2. Scroll to the bottom of the page and read the footer.

### Expected result

All strings should be in German.

### Actual result

There's an English string that reads "Visit Mozilla Corporation’s not-for-profit parent, the Mozilla Foundation."

### Environment

N/A

Shivansh2407:
Can I please work on this ???

craigcook:
That string is translated in Pontoon: https://pontoon.mozilla.org/translate/de/mozillaorg/footer.lang/?string=204294

I thought it might be a string mismatch between the [template](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/includes/protocol/footer/firefox-new.html#L84) and the [lang file](https://github.com/mozilla-l10n/www.mozilla.org/blob/master/de/footer.lang#L105) but they look identical to me so I'm not sure what the issue might be, perhaps I'm missing something.

stephaniehobson:
The strings look the same to me too @peiying2 can you please help us diagnose the problem?

peiying2:
Work on demo server: https://bedrock-demo-navigation.oregon-b.moz.works/de/firefox/

craigcook:
So that demo server includes footer.lang in the list of globally active files but that file is NOT globally active in production... I got my head jumbled and for some reason was thinking we could show the new footer to any locales that had completed it, but that's not the case. We also have to make those translations available site-wide.

But footer.lang is already complete in 33 locales with a few more almost done. There are some who will catch up quickly and others who will likely continue to straggle behind because they're less active communities and haven't updated any strings in months anyway.

So after talking to Peiying we've decided we're in a pretty good spot to globally activate footer.lang and remove the conditional logic. I'll also comment on #7913 with more specifics.

PR #7913, opened 2019-10-15
Activate new footer for more translated locales

stephaniehobson:
## Description

Enable the new footer for more locales

## Issue / Bugzilla link

n/a

## Testing
- new mozilla and firefox footers available in more languages
- no errors rendering new mozilla footer, old mozilla footer, or new firefox footer

### DIFF
--- a/bedrock/base/templates/includes/protocol/footer/firefox.html
+++ b/bedrock/base/templates/includes/protocol/footer/firefox.html
@@ -2,8 +2,8 @@
   # License, v. 2.0. If a copy of the MPL was not distributed with this
   # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
- {% if LANG.startswith('en-') or LANG in ['de'] %}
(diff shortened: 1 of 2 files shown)


stephaniehobson:
I'm not sure we should merge this until #7904 is fixed.

craigcook:
So it turns out I was wrong (yes, really!) and this won't work anyway. footer.lang needs to be a global shared file by [adding it to the list in settings/base.py](https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L234).

Then we can remove the conditional logic and start showing the new footer to all locales a bit ahead of when we expected.

craigcook:
Also, if we're removing the conditional logic I suppose that also means we can delete the old Mozilla footer and rename things to lose the -new suffixes and just have footer/firefox.html and footer/mozilla.html ... but that's a bigger change we may want to think through and test.

stephaniehobson:
Closing in favour of https://github.com/mozilla/bedrock/pull/7936

Issue #7915, opened 2019-10-15
Update favicons


stephaniehobson:
## Description

Since #7889 there is now a central source of browser favicons. All favicon references in bedrock should be updated to use those.

---

#### Tasks
- [x] Mozilla
- [x] Firefox master brand
- [x] Firefox Browser
- [x] Firefox Browser Beta
- [x] Firefox Browser Developer
- [x] Firefox Browser Nightly
- [x] Firefox Lockwise
- [x] Pocket

stephaniehobson:
I think this is done.

PR #7916, opened 2019-10-15
Get Mozilla favicons from favicons folder (#7915)

stephaniehobson:
## Description

- move existing favicons into img/favicons/mozilla
- update symlinks in root
- update template references

## Issue / Bugzilla link

#7915 

## Testing

- [x] Do the favicons in the root still load? (I don't know how to test this, it fails on master)
- [x] Are the favicons on the homepage still Mozilla, are the URLs right?
- [x] Are the favicons on other Mozilla pages still Mozilla, are the URLs right? (you don't need to test them all, just a random sample).


stephaniehobson:
The root ones are working on demo 4 :)

PR #7935, opened 2019-10-16
Update image helpers to always use the full path to the image

### LINKED ISSUE #7863: High-res image helper doesn't work with images from Protocol NPM package
## Description

See https://github.com/mozilla/bedrock/pull/7848 for context, where we had to revert a change because the high_res_img helper doesn't play quite right with images that are in `/protocol/`, rather than `/media/img/`.

To make it more fiddly, the helper seems to work as expected when running locally, but not when deployed to demo.

### Steps to reproduce

I'm actually not quite sure how to reproduce this locally (what are the best steps for emulating the same environment?), but adding the following to a template and pushing to demo should reproduce the bug:

```
{{ high_res_img('/protocol/img/logos/firefox/monitor/logo-word-hor-stack-md.png') }}
```

### Expected result

The page should give a 200 response.

### Actual result

The page throws a 500 error.

### Environment

N/A

pmac:
This is less magic, and allows images to be used with the helpers from outside of the `media/img` directory. Unfortunately it means adding `img/` to all of the uses of the helpers.

Fix #7863

### DIFF
--- a/bedrock/base/templates/macros-protocol.html
+++ b/bedrock/base/templates/macros-protocol.html
@@ -7,11 +7,10 @@
 {% macro billboard(title, ga_title, desc, link_cta, link_url, image_url, include_highres_image=False, reverse=False, heading_level=2) -%}
 <section class="mzp-c-billboard {% if reverse %}mzp-l-billboard-left{% else %}mzp-l-billboard-right{% endif %}">
   <div class="mzp-c-billboard-image-container">
-    {% set image = image_url if include_highres_image else 'img/' + image_url %}
     {% if include_highres_image %}
-      {{ high_res_img(image, {'alt': '', 'class': 'mzp-c-billboard-image', 'width': '346', 'height': '346'}) }}
+      {{ high_res_img(image_url, {'alt': '', 'class': 'mzp-c-billboard-image', 'width': '346', 'height': '346'}) }}
     {% else %}
-      <img src="{{ static(image) }}" class="mzp-c-billboard-image" width="346" height="346" alt="">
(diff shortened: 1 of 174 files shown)


stephaniehobson:
Random clicking around found a 500 on the lockwise page. https://bedrock-demos-pr-7935.herokuapp.com/en-US/firefox/lockwise/

stephaniehobson:
If you haven't already pushed to the `run-integration-tests` branch that would be a good idea.

pmac:
Yeah... I'm wondering if this isn't just too much. I was going to and still could just special-case image URLs that start with `protocol/`. That would be a MUCH smaller change. I think this is the Right Way™️ but might be overkill too, and as you said very quick to get out of date...

pmac:
Hmmm... I see that error in Sentry: `ValueError: Missing staticfiles manifest entry for '/protocol/img/logos/firefox/lockwise/og.png'` Strange. Is that working elsewhere?

stephaniehobson:
I agree this is The Right Way ™️ and would like to find a way to make it happen. 

The lockwise og image should be in /media (https://www.mozilla.org/media/protocol/img/logos/firefox/lockwise/og.png)

stephaniehobson:
Oh you have found the other thing I would like to get fixed. I can never remember if I should start with a leading slash or not. I think removing the leading slash fixes the lockwise problem.

pmac:
I believe I've fixed tests and I've rebased on master. I also filed #7949 to address the slash prefix issue during dev.

pmac:
I also pushed the branch to `run-integration-tests` and it all passed.

PR #7937, opened 2019-10-16
Port Transparency Report pages to Protocol (Fixes #7530)

### LINKED ISSUE #7530: Port Transparency Report pages to Protocol
## Description

The Transparency Report pages are using Pebbles.
https://www.mozilla.org/about/policy/transparency/

---

- [ ] Transparency Report pages are updated in production using Protocol.

#### Tasks
- [ ] Update templates.

amychurchwell:
## Description
Updated Transparency Report templates to use Protocol.

## Issue / Bugzilla link
#7530 

## Testing
http://localhost:8000/about/policy/transparency/
All templates under /about/policy/transparency/ are no longer using Pebbles.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
+++ b/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/past-reports.html
@@ -2,21 +2,17 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
-
 {# Add links to previous reports as new reports are published. #}
-<div class="nav-previous">
-  <div class="content">
-    <nav>
-      <h3>{{ _('Previous Reports') }}</h3>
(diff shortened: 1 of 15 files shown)


alexgibson:
At smallest screen sizes, the title overflows the edge of the viewport ever so slightly:

image

alexgibson:
Would it be more closely aligned with the current design if we made the whole page use the narrow column layout, instead of just the top section?

image

alexgibson:
Visually, it feels like the three nav buttons could sit inside the narrow column margins, rather than float outside in their own space?

image

amychurchwell:
> The page title also still runs wider than the screen.

Struggling to reproduce this bug, what width is it? I can say that padding that is probably causing that is coming from `mzp-l-content`

alexgibson:
> Struggling to reproduce this bug, what width is it?

Using Firefox, resize the browser window to it's smallest width (320px, which is the smallest phones go). I don't thik this needs to be a blocker though

Issue #7941, opened 2019-10-16
Double counting events on lockwise page


hoosteeno:
## Description

In https://github.com/mozilla/bedrock/blob/d09bd3bdce9a778c951fce74fe9d58866629fa81/bedrock/firefox/templates/firefox/lockwise/lockwise.html#L55, we see the following:

```
 <a href="{{ lockwise_adjust_url('ios', 'lockwise-page') }}" data-link-type="download" data-download-os="iOS" class="ios-button"  data-cta-text="apple-app-store" data-cta-type="lockwise-app-store" data-cta-position="primary">
```

I believe `data-link-type` causes an event to fire, for an iOS (browser) download. And `data-cta-type` causes an event to fire. We should not fire both events.

We should use the spec here for the data attributes on links on this page:

https://github.com/mozilla/bedrock/issues/7583#issuecomment-536762745

Specifically:

* remove data-link-type and data-download-os from the lockwise app store links

Shivansh2407:
I have removed data-link-type and data-download-os from the lockwise app store links. Please check the PR #7944 .

alexgibson:
@hoosteeno it looks like fixing this here on one page is easy, but the Google play store badge uses a macro that is shared on many pages across the site. Removing `data-link-type="download" data-download-os="Android"` from the macro will therefore effect download counts on many more pages.

The suggested values `data-cta-type=“lockwise-app-store” data-cta-text=“google-play-store”` only work assuming the macro is used for Lockwise. If we we're to make the macro have a default value (say data-cta-type=“firefox-app-store”`) that we could then override, would that work?

hoosteeno:
I'm not sure I understand: Are we using the same macro to build app store buttons for every kind of app in the app store? And that means we have to count all clicks on app store buttons as a download click on a mobile browser's download button?

alexgibson:
> Are we using the same macro to build app store buttons for every kind of app in the app store? And that means we have to count all clicks on app store buttons as a download click on a mobile browser's download button?

Yes. 99% of these are just for Firefox Android, but a few are for Focus. I don't know the history for why it's configured this way, but removing the default dtat-attributes will havce knock on offects on buttons for other pages.

hoosteeno:
Since Lockwise is not a browser download, we shouldn't count it as one. It sounds like the solution is more complicated than removing a parameter; in fact, we need to replace the macro on non-browser download buttons. It could be a different macro; or a more flexible macro; or bespoke code on this one page. 

[...]

Issue #7583, opened 2019-08-15
Lockwise - Build analytics plan and implement analytics


ejregithub:
## Description

We need an analytics instrumentation plan for Lockwise when it lands at www.mozilla.org. We should evaluate the current state of Lockwise analytics on lockwise.firefox.com, and design an ideal analytics setup, and specify the instrumentation so we can build it.

Notes
* If there are existing analytics we should prioritize continuity.
* There are 3 CTAs: Download on play store, download on app store, and install extension. The extension CTA will disappear later.
* We might be able to use our new [Generic CTA](https://github.com/mozilla/bedrock/issues/7491#issuecomment-520981154) tooling for this, in which case we simply need to specify the values of the CTA datalayer attributes.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Analytics plan is specified adequately for a developer to implement

ejregithub:
@sghoseWI - please review tasks here and consult w. @Errolyn who is coding this page.

sghoseWI:
UPDATES: 

Tags
On App Store download buttons:
data-cta-type=“lockwise-app-store”
data-cta-text=“apple-app-store|google-play-store” // there are 2 of these buttons; use only one text option depending on the platform
data-cta-position=“primary”

On ‘Install for Firefox’ button:
data-cta-type=“lockwise-install”
data-cta-position=“primary”

On ‘Get Firefox’ button:
data-cta-type=“lockwise-download-fx”
data-cta-position=“primary”

On ‘Open in Firefox’ button:
data-cta-type=“lockwise-open-in-fx”
data-cta-position=“primary”

On Offsite links only (not on any onsite links):
utm_source=www.mozilla.org-lockwise
utm_campaign=lockwise
utm_medium=referral
entrypoint = “mozilla.org-lockwise”

sghoseWI:
One caveat (from Justin): All of our download buttons are very finely tuned. If we change the data params on them, we break download goal counting. We should use a default download button for "Get Firefox".  @Errolyn

PR #7944, opened 2019-10-17
Update lockwise.html

Shivansh2407:
## Description

Double counting events on lockwise page #7941

### DIFF
--- a/bedrock/firefox/templates/firefox/lockwise/lockwise.html
+++ b/bedrock/firefox/templates/firefox/lockwise/lockwise.html
@@ -55,7 +55,7 @@
         <a href="{{ lockwise_adjust_url('ios', 'lockwise-page') }}" class="ios-button"  data-cta-text="apple-app-store" data-cta-type="lockwise-app-store" data-cta-position="primary">
           <img src="{{ l10n_img('firefox/ios/btn-app-store.svg') }}" alt="{{ _('Download on the App Store') }}" width="152" height="45">
         </a>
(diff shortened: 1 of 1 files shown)


Shivansh2407:
But from android part what should be removed @alexgibson 

See here 
        {{ google_play_button(href=lockwise_adjust_url('android', 'lockwise-page'), extra_data_attributes={'download-location': button_location, 'cta-type': 'lockwise-app-store', 'cta-text': 'google-play-store', 'cta-position': 'primary' } ) }}

alexgibson:
The spec is linked in the issue: https://github.com/mozilla/bedrock/issues/7583#issuecomment-536762745

I'll paste it again here:

```
On App Store download buttons:
data-cta-type=“lockwise-app-store”
data-cta-text=“apple-app-store|google-play-store” // there are 2 of these buttons; use only one text option depending on the platform
data-cta-position=“primary”
```

alexgibson:
It looks like the macro you may need to update is here:

https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/macros.html#L57-L63

On further inspection, I'm not sure this is so straight forward, as removing the default attributes is going to effect analytics on other pages. I'm going to mark this as do-not-merge until I can suggest a path forward. Thanks again

Shivansh2407:
Made some changes in the lockwise.html file

alexgibson:
> Made some changes in the lockwise.html file

Please hold while I can suggest a correct set of changes. Thanks

PR #7869, opened 2019-10-09
[fix #7748] Add lifecycle/welcome page 2

craigcook:
## Description
Adds a new lifecycle page promoting Pocket.

## Issue / Bugzilla link
#7748 

## Testing
https://bedrock-demo-craigcook.oregon-b.moz.works/firefox/welcome/2/

PR #7954, opened 2019-10-17
Welcome CTA is at top and bottom

hoosteeno:
## Description
This PR puts the welcome CTAs at top and bottom, and modifies some vertical spacing to make the page flow well. I collaborated on it with @cgonz0 so it has been design-reviewed. 

With 2 CTAs we need to modify the FxA flow begin scripts so they can handle an array of buttons but make only a single request for flow. 

Hopefully this does not break existing macro users and does not require any changes for them.

image

### DIFF
--- a/bedrock/firefox/templates/firefox/welcome/base.html
+++ b/bedrock/firefox/templates/firefox/welcome/base.html
@@ -35,6 +35,10 @@
       <div class="body-secondary">
         {% block content_secondary %}{% endblock %}
       </div>
+
+      <div class="secondary-cta">
+        {% block secondary_cta %}{% endblock %}
+      </div>
     </div>
   </section>
 
--- a/bedrock/firefox/templates/firefox/welcome/page1.html
+++ b/bedrock/firefox/templates/firefox/welcome/page1.html
@@ -24,12 +24,24 @@
     title=_('You’re on track to stay protected'),
     desc=_('You’ve got the web browser that protects your privacy — now it’s time to get a lookout for hackers.'),
     class='mzp-t-firefox mzp-t-dark',
-    include_cta=False,
+    include_cta=True,
     heading_level=1,
     image_url=None,
     include_highres_image=False,
     image_alt=''
   ) %}
+
+  <p class="primary-cta">
+      {{ monitor_button(
+        entrypoint='mozilla.org-firefox-welcome-1',
+        utm_campaign='welcome-1-monitor',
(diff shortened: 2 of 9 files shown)


alexgibson:
If there are now 2 buttons on this page we might need to make sure they don't have duplicate IDs, which is used by the integration test here: https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/welcome/page1.py#L14

craigcook:
> If there are now 2 buttons on this page we might need to make sure they don't have duplicate IDs, which is used by the integration test here: https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/welcome/page1.py#L14

I think it will also mean refactoring mozilla-monitor-button.js to apply the FxA params to multiple buttons. Right now it's assuming a single button with a unique ID, I think we can change that to either a class or a maybe an array of IDs, just something we can loop through and set the same params on each button.

amychurchwell:
This branch may need a rebase now that https://github.com/mozilla/bedrock/pull/7869 has been merged! Also, should we make this change for welcome 2 also?

alexgibson:
I think we already have logic in place to handle multiple buttons and ID’s https://bedrock.readthedocs.io/en/latest/firefox-accounts.html#id5

craigcook:
> I think we already have logic in place to handle multiple buttons and ID’s https://bedrock.readthedocs.io/en/latest/firefox-accounts.html#id5

We can pass in an ID, but I think that would require firing two different button inits, which would perform the fetch twice, right? Wouldn't it be better to fetch once and apply the same parameters to all the buttons on a page? Or am I misreading mozilla-monitor-button.js?

WNP69 has two Monitor buttons but only one gets initialized. I don't think we've ever done two buttons at once.

hoosteeno:
We have logic for multiple IDs, but unless I'm mistaken, we don't have logic to cache the fetched params. If we init the button twice, we'll cause two different flow begins; get two different flow IDs; and make too many requests. 

I am near to a fix on this.

craigcook:
image

The top button is picking up link styles from the `.mzp-c-hero.mzp-t-dark` rule. But really we should just remove those styles and replace them with `@include light-links;` (and link colors in dark themes is something that needs to change upstream in Protocol; I jumped the gun on this page).

amychurchwell:
Picking up this PR.

amychurchwell:
This work is continued in #7972, closing this PR. 😄

PR #7962, opened 2019-10-18
Updated images on /firefox

### LINKED ISSUE #7964: Images on /firefox could use some further optimization
## Description

The new [/firefox](https://www.mozilla.org/en-US/firefox/) page weighs in at over 3MB high-res screens, which is more than twice as resource heavy as the previous design.

There may be further optimizations we could do, but looking at the waterfall chart there are a couple of images that jump out; both of which could likely be optimized further.

- https://www.mozilla.org/media/img/firefox/home/master/respect-high-res.png (1.4MB)
- https://www.mozilla.org/media/img/firefox/home/master/pocket-high-res.png (615KB)

stephaniehobson:
## Description

Updated images on /firefox
- converted some to SVG
- optimized others

## Issue / Bugzilla link

Fix #7964 

## Testing

http://localhost:8000/en-US/firefox/


alexgibson:
I noticed the respect image could use optimizing as well https://github.com/mozilla/bedrock/issues/7964

stephaniehobson:
I converted the respect image to JPG and did some slight of hand with the gradient instead to avoid a transparent PNG. The JPG can be compressed further but then it doesn't exactly match the background colour provided by the CSS. I tried a compressed PNG and it keeps the background colour but the file size is actually larger than the uncompressed JPG. The JPG was exported at 70% quality so it's a reasonable file size even without extra compression.

PR #7969, opened 2019-10-21
Update MOSS india page with l10n fallback strings

alexgibson:
I merged a change in https://github.com/mozilla/bedrock/commit/a14dabea7f02b336313fb9e7337c3c83ccfce125 without realising that one of the pages was translated for hi-IN. This is a PR to fix that.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/moss/mission-partners-india.html
+++ b/bedrock/mozorg/templates/mozorg/moss/mission-partners-india.html
@@ -21,15 +21,28 @@
   <h2 class="moss-section-heading">{{ _('About') }}</h2>
 
   <p class="moss-copy">
-  {% trans url1=url('mozorg.moss.mission-partners') %}
-    <strong>The Global Mission Partners: India track is now closed.</strong> All applications which would previously have been submitted to the Global Mission Partners: India track will now be considered under the Mission Partners track. Please see the <a href="{{ url1 }}">Mission Partners page</a> for more information.
-  {% endtrans %}
+  {% if l10n_has_tag('moss_updates_20191021') %}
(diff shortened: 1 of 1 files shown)


alexgibson:
@mehan I'm going to need to revert the changes from your PR temporarily. My appologies, but I thought the pages we're en-US only, but it turns out they are translated in `hi-IN` also.

I'll work on a new PR that updates the pages with your changes, but does it in a safe way that won't break translations.

alexgibson:
Ok, I worked out a way to make this a much smaller change, without needing to revert. It turns out that the main MOSS pages aren't translated in hi-IN, but [only this page](https://github.com/mozilla-l10n/www.mozilla.org/blob/master/hi-IN/mozorg/moss/mission-partners-india.lang).

I've updated above to use an l10n tag with fallback strings, which means we can keep the other changes.

mehan:
@alexgibson great, many thanks for your help with this! just FYI, that MOSS mission partners india page is now deprecated (we will no longer link to it from the main MOSS site once these changes get merged), so hopefully this shouldn't be an issue in the future.

alexgibson:
@mehan if the India page is deprecated, is there any reason why we couldn't remove the page and redirect to https://www.mozilla.org/en-US/moss/mission-partners/ ?

From a maintainability point of view, this may be easier than keeping this content around longer than it's needed.

mehan:
@alexgibson That works for me! I wasn't sure what the usual procedures are around this, so I didn't want to delete the page. Redirecting to the Mission Partners page would be perfect. Thanks!

alexgibson:
Ok great - that simplifies what I need to do here :)

PR #7972, opened 2019-10-21
Add hero ctas on firefox/welcome/1 and /2

amychurchwell:
## Description
Continuation of PR #7954 

Modifies monitor button and pocket button FxA flow begin scripts:
- ensures only one fetch request is made no matter how many ctas.
- every button receives the same `deviceId`, `flowBeginTime`, and `flowId`
- does not require unique ID's for each button. (uses class hook `js-monitor-button`)

_note_: 
- removing `cta_position` from docs because this is not a value that is passed to monitor. We pass `utm_content` if there is a distinction to be made. `data-cta-type` records on our end whether the `primary` or `secondary` cta was clicked.
- removing `button_id` from docs because this PR makes it so button ID is no longer the way we differentiate different buttons.

## Testing

- does this change break previous implementations of the monitor button macro?
- ctas reflect the same `deviceId`, `flowBeginTime`, and `flowId` values?


amychurchwell:
accidental branch on mozilla/bedrock 😅

PR #7834, opened 2019-10-07
Add lifecycle page 1 to CRO Sandbox (Fixes #7802)

alexgibson:
## Description

- Adds an en-US only version of /firefox/welcome/1/ at /exp/firefox/welcome/1/
- Adds headless tests for /exp/ pages.

## Issue / Bugzilla link
#7802

## Testing
- [x] Experiment page should look & function the same as regular page.
- [x] Experiment page should load Convert JS (when DNT is disabled).
- [x] Experimental page be noindex (to match the original)

Successful test run: https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/373/pipeline

PR #7973, opened 2019-10-21
Additional ctas on welcome/1 and /2 (Fixes #7961)

### LINKED ISSUE #7961: Put CTA at top and bottom of welcome pages
## Description

We should move the CTA from the middle of the welcome pages to the top and the bottom. This will require some adjustment to vertical spacing, and some javascript refactoring. 

---

- [ ] Welcome page CTAs are at top and bottom.
- [ ] reviewed by design
- [ ] no string changes

amychurchwell:
## Description
Continuation of PR #7954 
⚠️ **Blocking #7834**

Modifies monitor button and pocket button FxA flow begin scripts:
- ensures only one fetch request is made no matter how many ctas.
- every button receives the same `deviceId`, `flowBeginTime`, and `flowId`
- does not require unique ID's for each button. (uses class hook `js-monitor-button`)

_note_: 
- removing `button_id` from docs because this PR makes it so button ID is no longer the way we differentiate different buttons.

## Issue
#7961

## Testing

- does this change break previous implementations of the monitor button macro?
- ctas reflect the same `deviceId`, `flowBeginTime`, and `flowId` values?

### DIFF
--- a/bedrock/firefox/templates/firefox/welcome/page1.html
+++ b/bedrock/firefox/templates/firefox/welcome/page1.html
@@ -18,18 +18,30 @@
 
 {% set _entrypoint = 'mozilla.org-firefox-welcome-1' %}
 {% set _utm_campaign = 'welcome-1-monitor' %}
+{% set _cta_type = 'lifecycle-monitor' %}
 
 {% block content_intro %}
   {% call hero(
     title=_('You’re on track to stay protected'),
     desc=_('You’ve got the web browser that protects your privacy — now it’s time to get a lookout for hackers.'),
     class='mzp-t-firefox mzp-t-dark',
-    include_cta=False,
+    include_cta=True,
     heading_level=1,
     image_url=None,
     include_highres_image=False,
     image_alt=''
   ) %}
+
+  <p class="primary-cta">
+    {{ monitor_button(
+      entrypoint=_entrypoint,
+      utm_campaign=_utm_campaign,
+      button_text=_('Check Your Breach Report'),
+      cta_type=_cta_type,
+      cta_position='primary'
+    ) }}
+  </p>
+
   {% endcall %}
 {% endblock %}
 
@@ -39,15 +51,6 @@
   <div class="body-primary-body">
     {# L10n: "Firefox Monitor" is a product name and shouldn't be translated. #}
     <p>{{ _('Firefox Monitor shows you if your information has been leaked in a known data breach, and alerts you in case it happens in the future.') }}</p>
-
-    <p class="primary-cta">
-      {{ monitor_button(
-        entrypoint=_entrypoint,
-        utm_campaign=_utm_campaign,
-        button_text=_('Check Your Breach Report'),
-        cta_type="Lifecycle-Monitor",
-      ) }}
-    </p>
   </div>
 {% endblock %}
 
@@ -84,6 +87,18 @@
     </div>
   </div>
   {% endif %}
+{% endblock %}
+
+{% block secondary_cta %}
+  <p>
+    {{ monitor_button(
+      entrypoint=_entrypoint,
+      utm_campaign=_utm_campaign,
+      button_text=_('Check Your Breach Report'),
+      cta_type=_cta_type,
+      cta_position='secondary'
+    ) }}
+  </p>
 {% endblock %}
 
 {% block content_utility %}
--- a/media/js/base/mozilla-monitor-button.js
+++ b/media/js/base/mozilla-monitor-button.js
@@ -15,17 +15,17 @@ if (typeof window.Mozilla === 'undefined') {
     MonitorButton.init = function() {
         var buttons;
 
+        // Exit if no fetch support
+        var supportsFetch = 'fetch' in window;
+        if (!supportsFetch) {
+            return;
+        }
+
         // Collect all monitor buttons
         buttons = document.getElementsByClassName('js-monitor-button');
 
         // Exit if no valid button in DOM
         if (buttons.length === 0) {
-            return;
-        }
-
-        // Exit if no fetch support
-        var supportsFetch = 'fetch' in window;
-        if (!supportsFetch) {
             return;
         }
 
--- a/media/js/base/mozilla-monitor-button.js
+++ b/media/js/base/mozilla-monitor-button.js
@@ -12,27 +12,28 @@ if (typeof window.Mozilla === 'undefined') {
 
     var MonitorButton = {};
 
-    MonitorButton.init = function(buttonId){
+    MonitorButton.init = function() {
+        var buttons;
 
-        // check for custom buttonId, used in case of multiple buttons.
-        if(!buttonId){
-            buttonId = 'fxa-monitor-submit';
-        }
+        // Collect all monitor buttons
+        buttons = document.getElementsByClassName('js-monitor-button');
 
-        var monitorButton = document.getElementById(buttonId);
-
-        // fetch request
-        var supportsFetch = 'fetch' in window;
-
-        if (!supportsFetch || !monitorButton) {
+        // Exit if no valid button in DOM
+        if (buttons.length === 0) {
             return;
         }
 
-        var buttonURL = monitorButton.getAttribute('href');
+        // Exit if no fetch support
+        var supportsFetch = 'fetch' in window;
+        if (!supportsFetch) {
+            return;
+        }
+
+        var buttonURL = buttons[0].getAttribute('href');
         // strip url to everything after `?`
         var buttonURLParams = buttonURL.match(/\?(.*)/)[1];
 
-        var destURL = monitorButton.getAttribute('data-action') + 'metrics-flow';
+        var destURL = buttons[0].getAttribute('data-action') + 'metrics-flow';
 
         // collect values from monitor button
         var params = window._SearchParams.queryStringToObject(buttonURLParams);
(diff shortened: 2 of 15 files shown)


media/js/base/mozilla-monitor-button.js
```diff
@@ -12,27 +12,28 @@ if (typeof window.Mozilla === 'undefined') {
...
-        // check for custom buttonId, used in case of multiple buttons.
-        if(!buttonId){
-            buttonId = 'fxa-monitor-submit';
-        }
+        // Collect all monitor buttons
+        buttons = document.getElementsByClassName('js-monitor-button');
 
-        var monitorButton = document.getElementById(buttonId);
+        // Exit if no valid button in DOM
+        if (buttons[0] === undefined) {
```
alexgibson: Use `length` property?
```suggestion
        if (buttons.length === 0) {
```

media/js/base/mozilla-monitor-button.js
```diff
@@ -12,27 +12,28 @@ if (typeof window.Mozilla === 'undefined') {
...
+            return;
+        }
 
-        // fetch request
+        // Exit if no fetch support
         var supportsFetch = 'fetch' in window;
-
-        if (!supportsFetch || !monitorButton) {
+        if (!supportsFetch) {
```
alexgibson: Minor performance suggestion: move this further up before the check for `buttons.length`, to save redundant DOM queries in non-supporting browsers?

media/js/base/mozilla-monitor-button.js
```diff
@@ -62,7 +63,10 @@ if (typeof window.Mozilla === 'undefined') {
             buttonURL += '&deviceId=' + r.deviceId;
             buttonURL += '&flowBeginTime=' + r.flowBeginTime;
             buttonURL += '&flowId=' + r.flowId;
-            monitorButton.setAttribute('href', buttonURL);
+            // applies url to all buttons and adds cta position
+            for (var i=0;i<buttons.length;i++) {
```
alexgibson: Nit spaces after `;` separators
```suggestion
            for (var i=0; i<buttons.length; i++) {
```

bedrock/firefox/templates/firefox/welcome/page1.html
```diff
@@ -86,6 +89,18 @@ <h3 class="c-picto-block-title">{{ _('Stay in the know') }}</h3>
...
 {% endblock %}
 
+{% block secondary_cta %}
+  <p>
+      {{ monitor_button(
+        entrypoint=_entrypoint,
+        utm_campaign=_utm_campaign,
+        button_text=_('Check Your Breach Report'),
+        cta_type=_cta_type,
+        cta_position='secondary'
```
alexgibson: Should this also have `utm_content='secondary'` to match the button metrics above?

media/js/base/mozilla-monitor-button.js
```diff
@@ -12,27 +12,28 @@ if (typeof window.Mozilla === 'undefined') {
...
+        // Exit if no fetch support
         var supportsFetch = 'fetch' in window;
-
-        if (!supportsFetch || !monitorButton) {
+        if (!supportsFetch) {
             return;
         }
 
-        var buttonURL = monitorButton.getAttribute('href');
+        var buttonURL = buttons[0].getAttribute('href');
```
alexgibson: Hmm, so this always gets the first match of a button in the page, which is then used to overwrite the `href` for every button in the page. Would this wipe out button specific params such as `utm_content=primary|secondary`?

If so, it feels like we should be preserving the original parameters, rather than overwriting. Thoughts?

alexgibson: Also: it would be great to add a unit test for this new functionality, to make sure it's doing the right things?

media/js/firefox/welcome-page2.js
```diff
@@ -13,23 +13,29 @@ if (typeof window.Mozilla === 'undefined') {
     var PocketButton = {};
 
     PocketButton.init = function(){
+        var buttons;
 
-        var pocketButton = document.getElementById('pocket-cta');
+        // Collect all monitor buttons
+        buttons = document.getElementsByClassName('js-pocket-button');
```
alexgibson: Not a blocker for this PR and I'm not saying you need to refactor it here - but this seems almost identical to the monitor button logic. We should probably be making this more DRY. Let's save that for a future issue though?

tests/pages/firefox/accounts.py
```diff
@@ -12,7 +12,7 @@ class FirefoxAccountsPage(FirefoxBasePage):
 
     URL_TEMPLATE = '/{locale}/firefox/accounts/{params}'
 
-    _firefox_monitor_button_locator = (By.ID, 'fxa-monitor-submit')
+    _firefox_monitor_button_locator = (By.CLASS_NAME, 'js-monitor-button')
```
alexgibson: Like `querySelector` in JS, this only matches the first element found with the class name in the page, so it won't cover two buttons. 

If desirable, I'd suggest making two different selectors (`_firefox_monitor_primary_button_locator`, `_firefox_monitor_secondary_button_locator`), and then having the test assert that both are visible more explicitly.

tests/pages/firefox/welcome/page1.py
```diff
@@ -11,7 +11,7 @@ class FirefoxWelcomePage1(FirefoxBasePage):
 
     URL_TEMPLATE = '/{locale}/firefox/welcome/1/'
 
-    _monitor_button_locator = (By.ID, 'fxa-monitor-submit')
+    _monitor_button_locator = (By.CLASS_NAME, 'js-monitor-button')
```
alexgibson: Same as above, this will only match one button, but not the other.

alexgibson:
Oh, so I didn't realize we don't have any unit tests for this monitor button code. I don't suggest delaying this PR to write some - but let's make sure we file an isse as follow up, as this code is used on enough pages that it really should have some coverage?

bedrock/firefox/templates/firefox/welcome/page1.html
amychurchwell: nice catch! some leftover code. 
I was in a discussion about whether `utm_content` should be passed to monitor to share info about which button was clicked. However, I don't think we need to pass this info because we are already collecting whether "primary" or "secondary" was clicked from `data-cta-position` on our end.

amychurchwell: I am going to change both to `cta_position`

alexgibson: If we don't need to pass those metrics over in the flow request, then we should remove the code that does it? https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-monitor-button.js#L45-L56

I guess my question is, what's required and what's not?

media/js/firefox/welcome-page2.js
amychurchwell: +100 agree. I was thinking that perhaps we can refactor the monitor JS to be usable by all fxa-product ctas?
We had originally separated it out because we were unsure if the pocket button should be made into it's own macro.

media/js/base/mozilla-monitor-button.js
alexgibson: If these values aren't important to pass to the fetch request, then can we remove the code that includes them to begin with? https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-monitor-button.js#L45-L56

bedrock/firefox/templates/firefox/welcome/page1.html
amychurchwell: We should keep that code because `utm_content` is optional.
> `utm_content`: Declared when more than one piece of content (on a page or at a URL) links to the same place, to distinguish between them.

the parameters that are required are:
- `entrypoint`
- `utm_campaign`

according to https://bedrock.readthedocs.io/en/latest/firefox-accounts.html#linking-to-monitorfirefoxcom

(though, I'm pretty sure `form_type='button'` and `'utm_medium=referral'` is also required.)

I guess the question here is that is `utm_content=primary/secondary` helpful information for monitor to have? I will confirm.

alexgibson: Ok I'm confused - the code I'm linking to above only passes over those values in the flow request. Removing that code will still keep those values in the final href. Unless I'm not following what you mean?

amychurchwell: oh! I see what you are saying. "Why are we passing these values specifically to the flow request if they are not required?" a good question! I think this is copy pasta from `mozilla-fxa-form.js`?

amychurchwell:
Opened issue for writing monitor button unit tests in #7981

tests/pages/firefox/accounts.py
amychurchwell: Thank you! I updated `page1.py` and `page2.py` as firefox/accounts only has one. 😄

media/js/firefox/welcome-page2.js
amychurchwell: Filed issue for this: #7985

media/js/base/mozilla-monitor-button.js
```diff
@@ -12,27 +12,28 @@ if (typeof window.Mozilla === 'undefined') {
...
+        }
 
-        // fetch request
+        // Exit if no fetch support
         var supportsFetch = 'fetch' in window;
-
-        if (!supportsFetch || !monitorButton) {
+        if (!supportsFetch) {
             return;
         }
```
alexgibson: Suggest moving this feature detect to the top of the function?

media/js/firefox/welcome-page2.js
```diff
@@ -12,24 +12,30 @@ if (typeof window.Mozilla === 'undefined') {
...
+        if (buttons.length === 0) {
+            return;
+        }
 
-        if (!supportsFetch || !pocketButton) {
+        // Exit if no fetch support
+        var supportsFetch = 'fetch' in window;
+        if (!supportsFetch) {
             return;
         }
```
alexgibson: Suggest moving this feature detect to the top of the function?

tests/pages/firefox/welcome/page1.py
```diff
@@ -11,8 +11,13 @@ class FirefoxWelcomePage1(FirefoxBasePage):
 
     URL_TEMPLATE = '/{locale}/firefox/welcome/1/'
 
-    _monitor_button_locator = (By.ID, 'fxa-monitor-submit')
+    _monitor_primary_button_locator = (By.CSS_SELECTOR, '.primary-cta .js-monitor-button')
```
alexgibson: I think you still need to update the actual test files that use these selector changes. Please push this branch to `run-integration-tests` to make sure things are passing.

amychurchwell: ah whoops! thank you. Looks like tests are passing now :+1:

PR #7986, opened 2019-10-22
Create What's New Page for India (Fixes #7552)

### LINKED ISSUE #7552: (11/4) - Display page content specific to users physically located in India

wen-2018:
Create What's New Page for India (Fixes #7552)

Note that this page uses the Geo Redirect View to show an English-language page to all visitors in India

## Description
The changes in bedrock/firefox/views.py is made referencing https://github.com/silverorange/bedrock/pull/5/files#r335481028.

## Issue / Bugzilla link
Fixes #7552

## Testing
This needs to be tested to make sure the Geo Redirect View works for visitors in India.


bedrock/firefox/views.py
```diff
@@ -547,6 +549,48 @@ def get_redirect_url(self, *args, **kwargs):
...
+
+        # add version to context for use in templates
+        version = self.kwargs.get('version') or ''
+        match = re.match(r'\d{1,2}', version)
+        ctx['version'] = version
+        ctx['num_version'] = int(match.group(0)) if match else ''
+
+        return ctx
+
+    def get_template_names(self):
```
alexgibson: @pmac is there a neater way we could implement this view, without needing to duplicate all the fallback locales from the regular whatsnew view, as below?

I think the only real difference here is:

```
if locale.startswith('en-'):
    template = 'firefox/whatsnew/index-lite.in.html'
```

alexgibson: I did some hacking on this and came up with:

```python
class WhatsNewIndiaView(WhatsnewView):
    def get_template_names(self):
        template = super(WhatsNewIndiaView, self).get_template_names()
        locale = l10n_utils.get_locale(self.request)
        version = self.kwargs.get('version') or ''
        channel = detect_channel(version)

        if locale.startswith('en-') and channel not in ['nightly', 'alpha', 'beta']:
            # return a list to conform with original intention
            return ['firefox/whatsnew/index-lite.in.html']

        return template
```

It seems to work and do the right things, although it would be good if @pmac could give it a sanity check.

Note: since the class extends `WhatsnewView`, I had to move it after the definition in the file for it to work.

alexgibson: It would alse be good to add a unit test for this new view / template

sgarrity: > It would alse be good to add a unit test for this new view / template

@alexgibson  Good idea - Do you know if there's a unit test for a similar view / template that we could use as an example?

alexgibson: @sgarrity tests for the existing whatsnew view can be found here: https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/tests/test_base.py#L228

Following the same pattern for the india-specific view should work.

bedrock/firefox/views.py
```diff
@@ -666,6 +668,20 @@ def get_template_names(self):
         return [template]
 
 
+class WhatsNewIndiaView(WhatsnewView):
```
pmac: @alexgibson I do like this better! Was going to suggest this yesterday and failed to get around to it. I do have a couple of NITs, but nothing major.

bedrock/firefox/views.py
```diff
@@ -666,6 +668,20 @@ def get_template_names(self):
...
+    def get_template_names(self):
+        template = super(WhatsNewIndiaView, self).get_template_names()
```
pmac: No need to pass anything to `super()` anymore now that we're on Python 3.7.

```suggestion
        template = super().get_template_names()
```

There is also no need to do this at the beginning of this method. You can call the above on line 682 so that it only does that work if it needs to.

bedrock/firefox/views.py
```diff
@@ -679,6 +678,8 @@ def get_template_names(self):
             # return a list to conform with original intention
             return ['firefox/whatsnew/index-lite.html']
 
+        template = super().get_template_names()
```
pmac: NIT: better to just `return super().get_template_names()`. No need for the extra variable.

pmac: Unless you'd like to have a single return statement, which is nice stylistically. e.g.:

```python
        if locale.startswith('en-') and channel not in ['nightly', 'alpha', 'beta']:
            # return a list to conform with original intention
            templates = ['firefox/whatsnew/index-lite.html']
        else:
            templates = super().get_template_names()

        return templates

**Comment by alexgibson** [comment]
Feedback from stakeholder:

1. Icons remove : please remove the two icons on the upper-right hand side of mobile.
The original source file was already updated before.
The reward program and the account feature will be only available in Indonesia on the first launch(11/4), NOT India.

image

2. Layout Cropping : The current layout from my desktop with 100% zoom level is a bit too far/small.
(The screenshot below was from my desktop.)
a. QR code needs to be fully displayed along with the copy (Scan the QR code to get started.) so users can directly take their mobile to scan QR code without scrolling down the screen. 

image

b. The expected layout is like this :
Zoom in the whole screen
image.png

image

3. Logo and Description :
a. The logo and the sentence is a bit small now, please refer to the logo and the sentence on the Indonesia WNP to make it bigger.

image

b. The description : shall we make these wording gray color as well?
Then the visual on the whole page would be more consistent. (Same as Indonesia WNP)

image

**Comment by alexgibson** [comment]
Please note the URL for the Firefox Lite logo is changing in https://github.com/mozilla/bedrock/pull/8017

**Comment by sgarrity** [comment]
>     1. please remove the two icons on the upper-right hand side of mobile.
>        The original source file was already updated before.

@alexgibson We can probably just get rid of these icons with some photoshop magic, but can you clarify what you mean by the original source file already being updated?

**Comment by alexgibson** [comment]
@sgarrity link to the source files here, thanks! https://drive.google.com/drive/folders/1GLEIkGnaPEog4kkCYUnoMkFHNyT8c1Wk?usp=sharing

**Comment by sgarrity** [comment]
@alexgibson We've addressed the size/color suggestions you made here. We had to take some liberties with the scale and layout to adapt the fixed original design into a responsive page, but we may have taken a bit too much creative license :-)   Should be closer to the original design now.

The remaining issues I'm aware of are:
1. Removing the icons in the mobile screenshot - I see you've provided the link to the source file - we'll update that now.
2. Getting a unit test in place for this view - I haven't written any unit tests for bedrock before (shame!), but we'll take a crack at this today
3. When wrapping the main headline for potential localization (probably not a blocker, as this page is English-only), I'm trying to figure out the best way to format the string for translation, as it contains some markup. So far, I'm thinking this:
```
{{ _('The new %s')|format('<span><strong>Firefox</strong> Lite</span>')|e }}
```
However, the `|e` isn't preventing the escaping of the markup as I expected. Do you know the appropriate syntax for this situation? I was looking for similar examples, but couldn't find any examples where the replacement string shouldn't be escaped.

alexgibson:
> When wrapping the main headline for potential localization (probably not a blocker, as this page is English-only), I'm trying to figure out the best way to format the string for translation, as it contains some markup.

@sgarrity I think including the markup in the translated string is probably just fine. I'd recommend adding an L10n comment directly above the line, to indicate that the tags are for visual formatting, e.g. https://github.com/mozilla/bedrock/blob/c9f6a8fc4cd4bcb5530801f03cd83fb472a3ab81/bedrock/mozorg/templates/mozorg/home/home.html#L56

sgarrity:
@alexgibson I think we're all caught up here. We've got the mobile screenshot updated, a unit test added, and the l10n string wrapping updated.

alexgibson:
@sgarrity please note https://github.com/mozilla/bedrock/pull/8017 merged, so you many need to update the path to the logo here in this PR.

wen-2018:
> @sgarrity please note #8017 merged, so you many need to update the path to the logo here in this PR.

Hi @alexgibson ! The logo path has been updated. Thanks!

alexgibson:
@sgarrity I think this may need a rebase against master again as there are commits here that don't belong to the PR.

wen-2018:
Hi @alexgibson ! This has been rebased again. Thanks!

alexgibson:
It's launch day, merging! 🎉

Issue #8051, opened 2019-10-29
OSError: [Errno 16] Device or resource busy: '/app/locale' when running `make test` on Docker for Mac


jgmize:
## Description

When running `make test` on Docker for Mac, multiple threads are attempting to mutate the `/app/locale` directory in parallel, resulting in test failures with `OSError: [Errno 16] Device or resource busy: '/app/locale'`. This is likely due to the delegated volume mounts for `/app/locale` in https://github.com/mozilla/bedrock/blob/f57cd17bfe25622cdba18202b502cea26bc328c5/docker-compose.yml#L45 and https://github.com/mozilla/bedrock/blob/f57cd17bfe25622cdba18202b502cea26bc328c5/docker-compose.yml#L65

### Steps to reproduce

Run `make test` on Docker for Mac. @amychurchwell reported this originally, and I was able to replicate.

### Expected result

Tests should pass.

### Actual result

Multiple test failures

### Environment

Docker for Mac

jgmize:
Fixed by #8052

PR #8052, opened 2019-10-29
Disable locale volume mount in test container

jgmize:
## Description

Temporarily disable locale volume mount

## Issue / Bugzilla link

#8051

## Testing

`make test` on Docker for Mac passes with these changes

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -42,7 +42,8 @@ services:
       - ./scripts/:/app/scripts:delegated
       - ./vendor-local/:/app/vendor-local:delegated
       - ./wsgi/:/app/wsgi:delegated
-      - ./locale/:/app/locale:delegated
+      # temporarily disabling mount due to mozilla/bedrock#8051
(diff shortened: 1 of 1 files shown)


jgmize:
I restored the volume mount on the app container in 700a444  and the tests are still passing locally.

jgmize:
I also tried replacing `delegated` with `consistent` and `cached` [configurations](https://docs.docker.com/docker-for-mac/osxfs-caching/) but got the same errors.

jgmize:
The [line triggering the errors](https://github.com/mozilla/bedrock/blob/master/bedrock/base/tests/test_accepted_locales.py#L45) is attempting to move the whole directory as part of the setup for each test. It may make more sense to remove this mount permanently given what the test setup is doing.

pmac:
Oh wow. Those tests are a bit crazy. I think maybe the right answer in this case is to just remove that test file as we'll be phasing out the old l10n system soon which this is testing.

PR #8055, opened 2019-10-30
Remove mozilla-accordion.js (Fixes #8019)

### LINKED ISSUE #8019: Remove legacy mozilla-accordion.js jQuery script
## Description

As soon as #7530 is fixed, I think we're in a position to remove the legacy jQuery `mozilla-accordion.js` script from the repository.

Looking through the repo, the [accordion bundle](https://github.com/mozilla/bedrock/blob/master/media/static-bundles.json#L1481-L1487) is still in use on several pages, but from what I can see that code is redundant, since all the pages it's on are now ported to Protocol, which uses the `details` JS component instead.

---

- [ ] JS library is removed from the repository.

- [ ] Need to double check code is not still being used by any pages.

#### Tasks
- [ ] Remove library.
- [ ] Remove bundles.
- [ ] Remove unit tests.
- [ ] Remove documentation.

alexgibson:
## Description
All pages on mozorg that have an accordion now use Protcol's Details component, so we can finally remove the legacy jQuery accordion lib 🎉

## Issue / Bugzilla link
#8019

## Testing
- [ ] Effected templates should still have functioning accordions where they appear.
- [ ] No other stray assets left.


alexgibson:
Oh nice catch, let me take a look at what's up there. I don't think I've ever seen that `privacy.js` file before :)

PR #8061, opened 2019-10-30
[fix #8060] Update welcome 1 page title

### LINKED ISSUE #8060: Update the hero section headline copy on the welcome 1 page
## Description

https://www.mozilla.org/firefox/welcome/1/

We need to update the headline copy on the welcome 1 page with something that we believe will perform better than the current copy based on testing in another location (mobile).

The new string for the hero headline replacing the current "You’re on track to stay protected" is:

"Stay ahead of hackers. Check for data breaches with Firefox Monitor."

---

- [ ] New string is live in en-US

#### Tasks
- [ ] Update headline copy in en-US and submit PR

craigcook:
## Description
Updates the title behind the tag `monitor_title_30102019`
Also adjusts the font-size since the new headline is much longer and wrapped awkwardly otherwise.

## Issue / Bugzilla link
#8060

### DIFF
--- a/bedrock/firefox/templates/firefox/welcome/page1.html
+++ b/bedrock/firefox/templates/firefox/welcome/page1.html
@@ -18,14 +18,15 @@
 
 {% set _entrypoint = 'mozilla.org-firefox-welcome-1' %}
 {% set _utm_campaign = 'welcome-1-monitor' %}
-{% set _utm_content = 'stay-ahead-hackers' %}
 {% set _cta_type = 'lifecycle-monitor' %}
 
 {% if l10n_has_tag('monitor_title_30102019') %}
   {# L10n: "Firefox Monitor" is a product name and shouldn't be translated. #}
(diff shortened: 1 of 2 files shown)


bedrock/firefox/templates/firefox/welcome/page1.html
```diff
@@ -18,11 +18,19 @@
 
 {% set _entrypoint = 'mozilla.org-firefox-welcome-1' %}
 {% set _utm_campaign = 'welcome-1-monitor' %}
+{% set _utm_content = 'stay-ahead-hackers' %}
```
amychurchwell: I think we should also put this value behind an l10n tag so that the analytics only change when the string is available?

craigcook: Good idea, though I expect this will be available today anyway. But it couldn't hurt.

PR #8068, opened 2019-11-01
Updated bug bounty sites

psiinon:
## Description

Add new sites to the web bug bounty program.

## Issue / Bugzilla link

n/a

## Testing

n/a

## Reviewers

- @april 
- @moz-jvehent 
- @claudijd

### DIFF
--- a/bedrock/security/templates/security/bug-bounty/web-eligible-sites.html
+++ b/bedrock/security/templates/security/bug-bounty/web-eligible-sites.html
@@ -99,6 +99,23 @@
     <li>aus5.mozilla.org</li>
   </ul>
 
+  <h3>Lando</h3>
+  <ul class="mzp-u-list-styled">
+    <li>api.lando.services.mozilla.com</li>
+    <li>api-private.lando.services.mozilla.com</li>
+    <li>lando.services.mozilla.com</li>
+  </ul>
+
+  <h3>Location Services</h3>
+  <ul class="mzp-u-list-styled">
+    <li>location.services.mozilla.com</li>
+  </ul>
+
+  <h3>Normandy</h3>
+  <ul class="mzp-u-list-styled">
+    <li>self-repair.mozilla.org</li>
+  </ul>
+
   <h3>Phabricator</h3>
   <ul class="mzp-u-list-styled">
     <li>phabricator.services.mozilla.com</li>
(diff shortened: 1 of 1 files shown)


bedrock/security/templates/security/bug-bounty/web-eligible-sites.html
```diff
@@ -99,6 +99,23 @@ <h3>Firefox Updates (AUS/Balrog)</h3>
...
+  </ul>
+
+  <h3>Location Services</h3>
+  <ul class="mzp-u-list-styled">
+    <li>location.services.mozilla.com</li>
+  </ul>
+
+  <h3>Normandy</h3>
+  <ul class="mzp-u-list-styled">
+    <li>self-repair.mozilla.org</li>
```
jvehent: I think normandy is 100% served via kinto nowadays, so the URL would be `https://firefox.settings.services.mozilla.com/v1/buckets/main/collections/normandy-recipes/records` but @mythmon can confirm

april:
This all looks good to me, but they're not my sites.  :)

claudijd:
All good here.

bedrock/security/templates/security/bug-bounty/web-eligible-sites.html
mythmon: Some very old clients pull from self-repair.mozilla.org. An intermediate age of browser pulls from normandy.cdn.mozilla.net. New versions of Firefox pull from Remote Settings. All three methods are actively used today, since we have a significant population of old users.

jwhitlock:
I’d like to keep location.services off the list until we deploy the new infrastructure, planned before end of year. @willkg, do you agree?

jvehent:
> I’d like to keep location.services off the list until we deploy the new infrastructure, planned before end of year. @willkg, do you agree?

@jwhitlock Effectively, location has been on this list since forever, because its a core firefox service. We just forgot to list it explicitly. I think it's safe to add it now, it won't drastically change the amount of submissions we're getting, and if a reporter finds something legitimate, we'd want to know anyway.

bedrock/security/templates/security/bug-bounty/web-eligible-sites.html
psiinon: firefox.settings.services.mozilla.com is already on the list, so that doesnt need to be added.

psiinon:
In that case I think this can be merged - @april or @claudijd can one of you do this honours? (I dont have perms;)

Issue #8045, opened 2019-10-29
Improve the quality of migrated Fluent files


alexgibson:
## Description

before we start porting legacy .lang files over to .ftl, we should make sure we try and adhere to a defined level of quality for those files. This could be achieved by a combination of improved automation and / or documentation.

alexgibson:
- Order of strings in fluent file should follow order in template.
- Avoid superfluous comments from .lang files that are no longer needed.
- Quality of auto-generated IDs?
- URLs should point to a resolving page.

alexgibson:
@Pike do you feel like the work in https://github.com/mozilla/bedrock/pull/8087 was enough to cover the main points above? Can this issue be closed?

alexgibson:
I'm going to close this based on the guidelines that were added to https://bedrock.readthedocs.io/en/latest/fluent-conversion.html

Feel free to re-open if there's still something missing in our process / documentation.

Issue #7877, opened 2019-10-10
Determine approach for migrating existing localizations


hoosteeno:
## Description

When migrating existing legacy content (mission.html and mission.lang become mission.html and mission.ftl), we want to port the localizations as well (bringing over existing localizations from .lang into .ftl files). 

* We need to make sure it's done in a way that does not trigger a ton of new work for Pontoon localizers
* We don't want to carry over all the cruft that has accrued in .lang files

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] In a meeting between pmac and Pike, we have figured out how to code this so that...
- [ ] The fluent files for reference English are well structured
- [ ] No existing localizations are lost
- [ ] The fluent files for all the localizations follow the improved structure from reference English

alexgibson:
The definition of a process in which these migrations will happen is still somewhat in flux, and seems to have become a bit of a blocker currently. @pmac has written some scripts in bedrock for migrating .lang files to .ftl, migrating jinja templates, and also migrating existing localizations, so that they are ready to submit as a PR to the L10n repo.

That work seems to fit most of what's defined above, but @pike called out some ways in which we need to improve the quality of migrated .ftl files, which are captured in https://github.com/mozilla/bedrock/issues/8045.

There also still seems to be some consensus that we need to change some of what we've already implemented, to better suit the needs of Pontoon; but I'm not the best person to define what those needs are. I've asked others to try and help articulate here.

pmac:
Here's what I have so far for a flow chart. See if this gels with what we've been saying @alexgibson @Pike 

![Basic Flowchart](https://user-images.githubusercontent.com/19176817/68242369-62fbc300-ffde-11e9-879e-d508d246c7f7.png)

alexgibson:
Thanks @pmac! In terms of how data gets from bedrock to each l10n repository, I think I follow along ok here, although I'm still a little confused about how data should flow between www-l10n and the l10n-team repo. If I'm interpreting this right, there would be a PR opened against the l10n-team repo, but the www-l10n repo would already contain the FTL files and metadata?

Pike:
The chart looks good to me, too.

Yes, when we have a migration recipe, we generate two exports, one into `www-l10n` and one into `l10n-team`.

Once the PR on `l10n-team` is merged, the next automation kicks in that exports from `l10n-team` into `www-l10n`. Given the "once" can be days, we need to do both.

[...]

PR #8087, opened 2019-11-06
Use recipes for Fluent migration

Pike:
## Description

This uses `fluent.migrate` behind the scenes to convert templates to Fluent. The rough overview of the new flow would be

```bash
./manage.py fluent recipe path/to/template.html
./manage.py fluent ftl path/to/template.html
./manage.py fluent template path/to/template.html
```

with manual edits after the `recipe` step to create good IDs, and after the `ftl` step to create good comments.

## Issue / Bugzilla link

#8045 is probably the core of this, but also #7877 

## Testing

This requires a few branches on the l10n libraries side, but I managed to migrate the `bedrock/firefox/templates/firefox/new/scene1.html` template, with `en` and localized Fluent files.

The docs are not yet where I'd like them to be, but I wanted to put this out there already.

## Todo
* [x] Finalize `.lang` patch in `compare-locales`. Has a review already, just needs follow-ups.
* [x] trans blocks
* [x] Limit change scope in `fluent.migrate`, get that reviewed and published.
* [x] Probably create an extra requirements file to use when creating and running migrations.

## Follow-ups
* Activation logic
* Cleanup

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -255,7 +255,8 @@ FLUENT_PATHS = [
     # remote FTL files from l10n team
     FLUENT_REPO_PATH,
 ]
-FLUENT_MIGRATIONS_PATH = ROOT_PATH / 'l10n' / 'bedrock_migrations'
+FLUENT_MIGRATIONS = 'lib.fluent_migrations'
+FLUENT_MIGRATIONS_PATH = ROOT_PATH / 'lib' / 'fluent_migrations'
 
 # Paths that don't require a locale code in the URL.
 # matches the first url component (e.g. mozilla.org/gameon/)
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -255,6 +255,7 @@ FLUENT_PATHS = [
     # remote FTL files from l10n team
     FLUENT_REPO_PATH,
 ]
+FLUENT_MIGRATIONS_PATH = ROOT_PATH / 'l10n' / 'bedrock_migrations'
 
 # Paths that don't require a locale code in the URL.
 # matches the first url component (e.g. mozilla.org/gameon/)
--- a/docs/fluent-conversion.rst
+++ b/docs/fluent-conversion.rst
@@ -14,39 +14,74 @@ from .lang to .ftl files. This document will cover the usage of these.
 If you've got a translated page you'd like to convert as-is to the Fluent system
 then you can follow this procedure to get a big head start.
 
-Convert a .lang file to English .ftl
-====================================
+The key concept in converting a page to Fluent is a recipe. The recipe is a
+piece of python code that can programmatically generate a Fluent file. It can
+use existing Fluent files as templates, and .lang localizations as data sources.
 
-The first step is to convert an existing english .lang file into an equivalent
-.ftl file in the ``en`` folder of the ``l10n`` directory in bedrock. Let's say
-for example you have a file called ``mozorg/mission.lang``; you would run:
+All the functionality is provided by the ``fluent`` management command via
+subcommands for each phase.
+
+Create a recipe from a template
+===============================
+
+The first step is to create the recipe. Let's say you want to convert
+``bedrock/mozorg/templates/mozorg/mission.html``, then you'll run
+
 
 .. code-block:: bash
 
     $ make run-shell
-    $ ./manage.py lang_to_ftl mozorg/mission
+    $ ./manage.py fluent recipe bedrock/mozorg/templates/mozorg/mission.html
 
-This should create ``l10n/en/mozorg/mission.ftl`` which has all of the strings,
-generated string IDs, and a comment above each one containing the md5 hash
-of the original string ID so that we can later match the new string IDs to the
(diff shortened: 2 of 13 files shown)


bedrock/settings/base.py
```diff
@@ -252,6 +252,8 @@ def lazy_langs():
     # remote FTL files from l10n team
     FLUENT_REPO_PATH,
 ]
+FLUENT_MIGRATIONS = 'lib.fluent_migrations'
+FLUENT_MIGRATIONS_PATH = ROOT_PATH / 'lib' / 'fluent_migrations'
```
pmac: can these live in the `FLUENT_LOCAL_PATH`, or should they stay with the code? It seems like we'll want/need to delete them once the migration makes it to production.

docs/fluent-conversion.rst
```diff
@@ -14,39 +14,74 @@ from .lang to .ftl files. This document will cover the usage of these.
...
+The first step is to create the recipe. Let's say you want to convert
+``bedrock/mozorg/templates/mozorg/mission.html``, then you'll run
+
+
+.. code-block:: bash
+
+    $ make run-shell
+    $ ./manage.py fluent recipe bedrock/mozorg/templates/mozorg/mission.html
+
+This will parse all of the calls to ``_()`` and the ``trans`` blocks, process the strings in the
```
pmac: One of the reasons I got the strings from the .lang files instead of templates is that templates often include strings that are translated in other files, e.g. a global file like newsletter.lang. Extracting everything in the template will result in string IDs with no translations in the .lang files you're porting.

lib/l10n_utils/management/commands/fluent.py
```diff
@@ -0,0 +1,83 @@
...
+            return self.handle_help(**kwargs)
+        if subcommand == 'recipe':
+            return self.create_recipe(**kwargs)
+        if subcommand == 'ftl':
+            return self.create_ftl(**kwargs)
+        if subcommand == 'template':
+            return self.create_template(**kwargs)
+
+    def handle_help(self, **kwargs):
+        self.stdout.write('''\
```
pmac: I find it best to use `textwrap.dedent` in these situations so that you can maintain the file indentation.

lib/l10n_utils/management/commands/fluent.py
```diff
@@ -0,0 +1,83 @@
...
+
+        template_parser = subparsers.add_parser(
+            'template',
+            description='Create template_ftl.html file with existing recipe'
+        )
+        template_parser.add_argument('template', type=Path)
+
+    def handle(self, subcommand, **kwargs):
+        if subcommand in (None, 'help'):
+            return self.handle_help(**kwargs)
```
pmac: Move this to the end with no `if` statement to also catch an invalid sub-command.

bedrock/settings/base.py
Pike: One thing that the migration library does is that it loads them as python modules.

I didn't feel great about that being in a data path, so I moved them to `lib`. Yes, I started with them being in `l10n` ;-) .

In Firefox, they're actually just part of the default virtualenv path, too.

And yes, they should stay with the code. I see how that's a tad tricky to ensure that we're doing things right, but there's also not a whole lot of risk. One thing the migration library does is that if a string is already in the target, it won't overwrite it.

I wouldn't recommend to overwrite all strings with all recipes all the time, but it's kinda safe and won't destroy any updates that localizers did, for example.

And yes, we clean those recipes up on the Firefox side once we don't need them around anymore. On bedrock, this could probably be an issue to do once the migrated content made the roundtrip through the repos to prod. Or as we do on fx, just a kinda-monthly chore.

docs/fluent-conversion.rst
Pike: https://github.com/mozilla/bedrock/blob/712b92ec732c8e9708bd63443b1c73fc9dba067e/lib/l10n_utils/management/commands/_fluent_recipe.py#L118-L129 keeps track of which string is coming from which lang file.

One nice thing about the migration being the first thing is that it'll create a migration for each file found. If that touches more .lang files than you want right now, you just remove the `add_transforms` call for that ftl file from the recipe, and the next steps will only bother about the content you left in the recipe.

Well, it doesn't do the default ones yet, not sure how to tackle those yet.

Pike:
Laid the foundations for an activation subcommand. I've talked about what that thing is supposed to do with pmac, who's going to dig in. I wrote the scaffolding now, which deduces the template name from a recipe name, and then finds the lang files and gets the activated languages for each. pmac will tell us what to do with that data ;-) (Sung to an old tune by Marillion)

pmac:
[This is the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=931046) I mentioned that should shed light on how it currently works and how it's desired to work.

Pike:
I lobby to take the activation and cleanup parts out of this PR. In particular the activation piece is gonna be easier to reason about if we do it in a separate PR.

Pike:
Nothing new in the patch, just rebased. Luckily there were no conflicts on the documentation changes.

Issue #7526, opened 2019-08-05
Allow a developer to specify the localization approach


hoosteeno:
* Vendor/community/hybrid
* per page
* drives (and maybe lives in?) the system that l10n team manages

PR #8092, opened 2019-11-07
Add l10n project configuration for Pontoon and vendor (Fixes #7875)

### LINKED ISSUE #7875: l10n project configuration for pontoon & vendor
## Description

Build project configuration files and commit to Bedrock.

---

- [ ] Ensure a developer can specify a localization approach for a particular fluent file
- [ ] Establish a default localization approach (Community)

Pike:
## Description

Add project configuration for Pontoon and vendor/staff localization.

@pmac @alexgibson I'm mostly interested in the comments here.

Also, adding docs to https://bedrock.readthedocs.io/en/latest/l10n.html ?

@peiying2, this also needs you review on the locales, in particular for vendors.

As for pontoon locales, I just started with everything we have in vcs now. Is that right?

## Issue / Bugzilla link

#7875 

## Testing

I tested this against the `www-l10n` data, and mission goes to vendor/staff for de, fr, as requested by Jennifer.


l10n/configs/vendor.toml
```diff
@@ -0,0 +1,27 @@
...
+#     reference = "en/firefox/*.ftl"
+#     l10n = "{locale}/firefox/*.ftl"
+#     locales = [
+#         "de",
+#         "fr",
+#         "ru",
+#     ]
+
+[[paths]]
+    reference = "en/mozorg/mission.ftl"
```
pmac: I don't think the mission page is going to be handled by the vendor.

pmac: Nevermind. I missed your comment in the description.

l10n/configs/special-templates.toml
```diff
@@ -0,0 +1,16 @@
+# For locales with dedicated templates that don't go through
```
alexgibson: I'm curious on the use case here - if content doesn't go through an L10n process, then we typically use hard-coded, locale-specific templates that bedrock already has functionality to support. Is this something that we'd need to make exceptions for here?

l10n/configs/vendor.toml
alexgibson: If `de` and `fr` are to be handled by Mozilla staff as opposed to community, then would we need to list out every `.ftl` file here explicitly, or could we just say "all .ftl files that map to either de or fr locales"?

Also if both, vendor and staff are covered in this config file, do we need a way to specify who's responsible for what?

l10n/configs/pontoon.toml
```diff
@@ -0,0 +1,121 @@
...
+    "bn",
+    "br",
+    "bs",
+    "ca",
+    "cak",
+    "crh",
+    "cs",
+    "cy",
+    "da",
+    "de",
```
alexgibson: Should `de` and `fr` be here if they are in `vendor.toml`?

l10n/configs/vendor.toml
alexgibson: From our discussion in Slack the other day, it sounds like Marketing are only responsible for a sub-set of pages (which are still tbd) for both German and French, so we may not be able to create such general rules as I queried about above.

l10n/configs/special-templates.toml
Pike: Yes.

l10n/configs/pontoon.toml
Pike: AFAICT, marketing doesn't intend to own *all* of mozilla.org, so yes.

Pike:
On a general note, it's important that `de` and `fr` go into our vendor system. That's where they can tap into paid translators on demand, w/out having to wait for a bunch of other people to do config changes for them.

How to best implement that is a different question, and really depends on the answer if the marketing coverage for tier 1 is effectively that of tier 2, or if they're different things.

There's no programmatic change, we'd just put `de` and `fr` into a different leaf config than the other vendor languages, and include both for the vendor, and exclude both for Pontoon.

But that's all opportunity and theory at this point, as we don't know what we're supposed to do.

alexgibson:
Whichever config we go for, it needs to be able to handle the localization approaches we agreed on in https://github.com/mozilla/bedrock/issues/7526

Hopefully we'll have a clearer idea of how this breaks down between different pages / sections of the site soon.

alexgibson:
I'm pasting this list of vendor locales from the email thread confirming them:

ru, pt-BR, pl, id, zh-TW, zh-CN, it, nl, ja, tr, vi, hu, cz, es-ES, es-MX, es-AR, hi-IN

Note: I've added in Hindi at the end, which was a follow up ask.

alexgibson:
One open question on the above list I notice: we have es-ES, es-MX, es-AR in there, but the original email (which Jeff replied to with this list) says "we've decided that localizing in a single Spanish locale is adequate". We may need some clarification here.

l10n/configs/vendor.toml
```diff
@@ -0,0 +1,47 @@
+# All content handled by the vendor or Mozilla staff should
+# be covered by the [[paths]] in this file, with the relevant locales.
+
+basepath = ".."
+
+locales = [
+    "cs",
+    "de",
+    "en-CA",
+    "en-GB",
```
Pike: We'll need to confirm that our English variants are actually staff-powered.

l10n/configs/vendor.toml
```diff
@@ -0,0 +1,47 @@
...
+    "es-AR",
+    "es-CL",
```
Pike: `es-CL` wasn't on the lists of Spanish, @gueroJeff, was that a mistake or intentional?

l10n/configs/vendor.toml
```diff
@@ -0,0 +1,47 @@
+# All content handled by the vendor or Mozilla staff should
+# be covered by the [[paths]] in this file, with the relevant locales.
+
+basepath = ".."
+
+locales = [
+    "cs",
```
Pike: The locale code for Czech is `cs`, `CZ` is the region code. This deviates from the spec, so calling that out.

Pike:
I'd like to get this reviewed by @hoosteeno and @gueroJeff for the locale choices, and @pmac and/or @alexgibson for technical accuracy.

I short story on the paths:

Looking at the spreadsheet, we're somewhat interwoven in terms of file paths as to what goes to vendor or staff or community.

We can start with explicit paths for now, and go for more general patterns once we know what the exceptions are, which is the approach I've take now.

Alternatively, we could put the ftl files into directories inside `l10n`, say, `l10n/en/s/` for staff, `l10n/en/v/`for vendors, and `l10n/en/c/` for community-only. I expect re-categorizations of content to be extremely rare, in particular rare w/out actual work on a page. In that case, we're probably fine to do some repo magic?

And last but not least, I've not introduced `l10n/configs/staff.toml`. Depending on our choices above, and how different the content story between "staff and vendors" vs "just staff and maybe community", it might make sense to do so. We can also do that when we find out it's good to have at a later point.

gueroJeff:
> I'd like to get this reviewed by @hoosteeno and @gueroJeff for the locale choices, and @pmac and/or @alexgibson for technical accuracy.

This is the list in the vendor bucket, as discussed with @hoosteeno:
    hi-IN (Hindi)
    ru (Russian)
    pt-BR (Portuguese, Brazil)
    pl (Polish)
    id (Bahasa Indonesian)
    ms (Bahasa Malay)
    zh-TW (Traditional Chinese)
    zh-CN (Simplified Chinese)
    it (Italian)
    nl (Dutch)
    ja (Japanese)
    tr (Turkish)
    vi (Vietnamese)
    hu (Hungarian)
    cz (Czech)
    es-ES (Spain Spanish)
    es-MX (Mexican Spanish)
    es-AR (Argentine Spanish)

Pike:
I've updated the PR here with the latest input.

There are some differences to the vendor-list that Jeff gave:
* addition of `de` and `fr`, as they're staff and go into the vendor system
* `cz` -> `cs`

I also talked about `en-*` with Jeff explicitly, and we're OK with them being in the community bucket. The teams keep up with their work on Pontoon well, and don't seem to have staff involved right now.

Added Malay. Dropped Chilean Spanish.

I can't set reviewers on PRs here, I'd like to have an r+ from @hoosteeno , could somebody set that?

pmac:
> I can't set reviewers on PRs here, I'd like to have an r+ from @hoosteeno , could somebody set that?

done

Issue #8059, opened 2019-10-30
Put Pocket button on firefox/pocket page.


ejregithub:
## Description

We built a new Pocket button with the CTA to 'Activate Pocket' in the [Welcome 2](https://www.mozilla.org/en-US/firefox/welcome/2/) browser welcome page and should put that button, with proper instrumentation, on firefox/pocket page.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Activate Pocket button is on the /firefox/pocket/ product page
- [ ] Is live prior to end of S2/Q4

#### Tasks
- [x] Add Pocket button
- [x] Instrument for attribution
- [x] Set it up as a Python helper
- [x] Add unit tests

alexgibson:
As part of this I think we should also tackle https://github.com/mozilla/bedrock/issues/7985

alexgibson:
Blocked by https://github.com/mozilla/bedrock/issues/7985

alexgibson:
Instrumentation for this buttons on this page is as follows:

Primary button
---------------

- `entrypoint`: `mozilla.org-firefox-pocket`
-  `s`: `ffpocket`
- `data-cta-text`: `Try Pocket Now`
- `data-cta-type` : `activate pocket`
- `data-cta-position`: `primary`

Secondary Button
------------------

- `entrypoint`: `mozilla.org-firefox-pocket`
-  `s`: `ffpocket`
- `data-cta-text`: `Try Pocket Now`
- `data-cta-type` : `activate pocket`
- `data-cta-position`: `secondary`

PR #8115, opened 2019-11-11
Refactor mozilla-monitor-button.js to be reusable for other services (Fixes #7985)

### LINKED ISSUE #7985: Refactor `mozilla-monitor-button.js` to be reusable for other fxa product CTA's.
## Description

We utilized the same JS for the pocket button in #7973 which is essentially a dupe of `mozilla-monitor-button.js`, we should refactor this code to be able to be utilized by any fxa product CTA that needs to make a fetch for `deviceId`, `flowBeginTime`, and `flowId`

alexgibson:
## Description
- Renames `mozilla-monitor-button.js` to `mozilla-fxa-product-button.js`.
- Renames `Mozilla.MonitorButton` to `Mozilla.FxaProductButton`.
- Removes duplicate `Mozilla.PocketButton` code.

## Issue / Bugzilla link
#7985

## Testing
- [ ] Pages that make FxA metrics flow requests should continue to do so, with no regressions.
- [ ] For Pocket, test `/firefox/welcome/2/`. All other affected pages should use a Monitor CTA.


alexgibson:
Marking this as a P2 since it blocks https://github.com/mozilla/bedrock/issues/8059

tests/unit/spec/base/mozilla-monitor-button.js
```diff
@@ -40,14 +40,14 @@ describe('mozilla-monitor-button.js', function() {
...
             expect(window.fetch).toHaveBeenCalledTimes(1);
             expect(window.fetch).toHaveBeenCalledWith('https://accounts.firefox.com/metrics-flow?entrypoint=mozilla.org-firefox-accounts&form_type=button&utm_source=mozilla.org-firefox-accounts&utm_campaign=trailhead');
         });
     });
 
     it('should attach flow parameters to button hrefs in the metrics response', function() {
-        return Mozilla.MonitorButton.init().then(function() {
+        return Mozilla.FxaProductButton.init().then(function() {
             var buttons = document.querySelectorAll('.js-monitor-button');
             expect(buttons[0].href).toEqual('https://monitor.firefox.com/oauth/init?form_type=button&entrypoint=mozilla.org-firefox-accounts&utm_source=mozilla.org-firefox-accounts&utm_campaign=trailhead&utm_medium=referral&deviceId=848377ff6e3e4fc982307a316f4ca3d6&flowBeginTime=1573052386673&flowId=75f9a48a0f66c2f5919a0989605d5fa5dd04625ea5a2ee59b2d5d54637c566d1');
```
amychurchwell: Since we are moving to the more generic `FxaProductButton`, should these tests be flexible enough to cover other product hrefs such as `https://getpocket.com/...`?

bedrock/base/templates/macros.html
```diff
@@ -229,7 +229,7 @@ <h2 class="thank-you hidden">{{ _('Your download link was sent.') }}</h2>
     {% do fxa_queries.append('entrypoint_variation=' ~ entrypoint_variation) %}
   {% endif %}
```
amychurchwell: perhaps we should consider renaming the `monitor_button` macro to something more generic and passing a `product='Monitor'` or `product='Pocket'` value to change the href? It could also add a selector to the macro we could target in tests to make sure the correct href for the product is being shown.

docs/firefox-accounts.rst
```diff
@@ -275,7 +275,7 @@ Invoking the macro will automatically include a set of default UTM parameters as
 Tracking Sign-ups / Sign-ins
 ----------------------------
 
-For both Firefox Accounts form submissions and Firefox Monitor referrals, we also pass ``device_id``, ``flow_id`` and ``flow_begin_time`` parameters to track top-of-funnel metrics. These are values fetched from a metrics flow API endpoint, and are instered back into the form / link along with the other standard referral parameters. This functionality is handled by ``mozilla-fxa-form.js`` and ``mozilla-monitor-button.js`` respectively.
+For both Firefox Accounts form submissions and Firefox Monitor referrals, we also pass ``device_id``, ``flow_id`` and ``flow_begin_time`` parameters to track top-of-funnel metrics. These are values fetched from a metrics flow API endpoint, and are instered back into the form / link along with the other standard referral parameters. This functionality is handled by ``mozilla-fxa-form.js`` and ``mozilla-fxa-product-button.js`` respectively.
```
amychurchwell: 🔥 👏

bedrock/base/templates/macros.html
alexgibson: This will come in the next PR (I'm just keeping the scope of this PR to the original isssue for now). But honestly, I don't think we should continue to make our macros more and more complex like this, as it already feels brittle. I think something that doesn this type of product link generation makes sense, but it should be a Python helper so that we can have unit tests.

alexgibson:
Thanks for the careful review @amychurchwell 💯 

I pushed a commit with some fixes, and also pushed this to the run-integration-tests branch: https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/454/pipeline

bedrock/base/templates/macros.html
amychurchwell: That sounds good, I created an issue for that work here: #8143

Issue #8118, opened 2019-11-11
Remove duplicate logo on /firefox/accounts


amychurchwell:
## Description

There is a logo duplicate on /firefox/accounts, we should remove the one above the headline.
![image](https://user-images.githubusercontent.com/30483988/68611777-c5861080-046f-11ea-9e25-69686e3ee369.png)

We should remove the logo background property on `.c-accounts-hero-title` in [`firefox/accounts-2019.scss`](https://github.com/mozilla/bedrock/blob/master/media/css/firefox/accounts-2019.scss#L28-L29)

imskr:
I would like to take this issue!

amychurchwell:
@imskr okay, I will assign you this bug.

PR #8122, opened 2019-11-11
Pass utm params from upstream referrals on fxa_email_form (Fixes #8001)

### LINKED ISSUE #8001: Not passing through utm parameters from upstream referrals on /firefox/accounts/
## Description

The page at www.mozilla.org/firefox/accounts/ is not passing through parameters coming from upstream referrers. It should be.

We have worked on this several times in 2019, most recently at https://github.com/mozilla/bedrock/issues/7585. We should review that work to understand if this bug relates to that in any way.

### Steps to reproduce

1. Visit about:welcome in a *Firefox Profile that is not signed in to Sync*.
2. Click the "Learn More about the Benefits" link at the bottom of the firstrun experience. 
3. Note the parameters on the URL: `https://www.mozilla.org/en-US/firefox/accounts/?utm_campaign=firstrun&utm_medium=referral&utm_source=activity-stream&utm_term=trailhead-join`
4. Fill in the Sync sign-up form (perhaps with a fake email address) and submit.
5. Note the parameters on the URL: `https://accounts.firefox.com/signup?action=email&context=fx_desktop_v3&service=sync&entrypoint=mozilla.org-firefox-accounts&form_type=email&flow_id=7c45f2ee51998ba8ae782391e1fbf4813174e8253668742d916bc0afd77fa440&flow_begin_time=1571857270270&device_id=78ad24a987b64c338da60447a29a7c70&utm_source=mozilla.org-firefox-accounts&utm_medium=referral&utm_campaign=trailhead-fxa-embedded-form&style=trailhead&email=flibbijy%40mailinator.com`\

### Expected result

The second URL should have the utm_source, utm_campaign, utm_medium, and utm_term that the first URL has. 

### Actual result

The second URL has the default utm parameters from this page.

amychurchwell:
## Description
Refactors `mozilla-fxa-form.js` to check for utm parameters on the url before applying it's own.

## Issue / Bugzilla link
#8001

## Testing

- [x] Should apply url utm parameter values to corresponding hidden input field values if present.
- [x] If there are no utm parameters on the url, it's own values are passed.
- [x] utm parameters are validated. 

- http://localhost:8000/en-US/firefox/accounts/?utm_campaign=firstrun&utm_medium=referral&utm_source=activity-stream&utm_term=trailhead-join

Screen Shot 2019-11-11 at 3 17 25 PM

- http://localhost:8000/en-US/firefox/accounts/

Screen Shot 2019-11-11 at 3 20 11 PM

### DIFF
--- a/media/js/base/mozilla-fxa-form.js
+++ b/media/js/base/mozilla-fxa-form.js
@@ -12,7 +12,7 @@ if (typeof window.Mozilla === 'undefined') {
 
     var FxaForm = {};
 
-    FxaForm.init= function(){
+    FxaForm.init = function() {
         var fxaForm = document.getElementById('fxa-email-form');
         var fxaSubmitButton = document.getElementById('fxa-email-form-submit');
         var fxaFormContextField = fxaForm.querySelector('[name="context"]');
@@ -33,7 +33,6 @@ if (typeof window.Mozilla === 'undefined') {
                     if (data.accurate && data.distribution && data.distribution.toLowerCase() === 'mozillaonline') {
                         fxaForm.action = mozillaonlineAction;
                     }
-
                     fetchTokens();
                 });
             } else {
@@ -49,14 +48,37 @@ if (typeof window.Mozilla === 'undefined') {
 
             fxaSubmitButton.disabled = false;
 
-            var destURL = fxaForm.getAttribute('action') + 'metrics-flow';
+            // track external UTM referrals
+            var urlParams = new window._SearchParams().utmParams();
+            var attributionData = Mozilla.UtmUrl.getAttributionData(urlParams);
+
+            if (attributionData) {
+                var utms = ['utm_source', 'utm_campaign', 'utm_content', 'utm_term', 'utm_medium'];
+                for (var i = 0; i < utms.length; i++) {
+                    if (Object.prototype.hasOwnProperty.call(attributionData, utms[i])) {
+                        // check if input is available
+                        if (fxaForm.querySelector('[name="' + utms[i] + '"]')) {
+                            fxaForm.querySelector('[name="' + utms[i] + '"]').value = attributionData[utms[i]];
+                        } else {
+                            // create input if one is not present
+                            var input = document.createElement('input');
+                            input.type = 'hidden';
+                            input.name = utms[i];
+                            input.value = attributionData[utms[i]];
+                            fxaForm.appendChild(input);
+                        }
+                    } else {
+                        // do not update form params.
+                    }
+                }
+            }
+
             var entrypoint = document.getElementById('fxa-email-form-entrypoint');
             var entrypointExp = document.getElementById('fxa-email-form-entrypoint-experiment');
             var entrypointVar = document.getElementById('fxa-email-form-entrypoint-variation');
             var utmSource = document.getElementById('fxa-email-form-utm-source');
-            var utmCampaign = document.getElementById('fxa-email-form-utm-campaign');
-            var utmContent = document.getElementById('fxa-email-form-utm-content');
-            var utmTerm = document.getElementById('fxa-email-form-utm-term');
+
+            var destURL = fxaForm.getAttribute('action') + 'metrics-flow';
 
             // add required params to the token fetch request
             destURL += '?form_type=email';
@@ -64,12 +86,17 @@ if (typeof window.Mozilla === 'undefined') {
             destURL += '&utm_source=' + utmSource.value;
 
             // add optional utm params to the token fetch request
-            if (utmCampaign) {
-                destURL += '&utm_campaign=' + utmCampaign.value;
-            }
+
+            var utmTerm = document.getElementById('fxa-email-form-utm-term');
+            var utmCampaign = document.getElementById('fxa-email-form-utm-campaign');
+            var utmContent = document.getElementById('fxa-email-form-utm-content');
 
             if (utmContent) {
                 destURL += '&utm_content=' + utmContent.value;
+            }
+
+            if (utmCampaign) {
+                destURL += '&utm_campaign=' + utmCampaign.value;
             }
 
             if (utmTerm) {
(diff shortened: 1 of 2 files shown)


media/js/base/mozilla-fxa-form.js
```diff
@@ -49,21 +48,39 @@ if (typeof window.Mozilla === 'undefined') {
...
-            var destURL = fxaForm.getAttribute('action') + 'metrics-flow';
+            // track external UTM referrals
+            var urlParams = new window._SearchParams().utmParamsFxA();
+            var attributionData = Mozilla.UtmUrl.getAttributionData(urlParams);
+
+            if (attributionData) {
+                var utms = ['utm_source', 'utm_campaign', 'utm_content', 'utm_term', 'utm_medium'];
+                for (var i = 0; i < utms.length; i++) {
+                    if (attributionData.hasOwnProperty(utms[i])) {
+                        fxaForm.querySelector(`[name="${utms[i]}"]`).value = `${attributionData[utms[i]]}`;
```
amychurchwell: tests are failing due to the notation on this line: 
```
(`[name="${utms[i]}"]`).value = `${attributionData[utms[i]]}`
```
but I'm not quite sure how to achieve this without the back-ticks?

alexgibson: String template literals don't yet have broad enough browser support to use natively: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals

You'll probably need to use regular string concatenation instead:

```javascript
'[name="' + utms[i] + '"]'.value = attributionData[utms[i]]
```

media/css/firefox/accounts-2019.scss
```diff
@@ -25,8 +25,6 @@
 
     .c-accounts-hero-title {
         @include text-display-lg;
-        background: transparent url('/media/protocol/img/logos/firefox/logo-word-hor.svg') center top no-repeat;
```
alexgibson: This seems unrelated to this issue, but if this is for https://github.com/mozilla/bedrock/issues/8118 then let's remove the logo from exp/firefox/accounts/ to?

amychurchwell: oh whoops, this was accidentally left in while I was filing that issue. 🤦‍♀

bedrock/base/templates/macros.html
```diff
@@ -264,24 +264,18 @@ <h2 class="thank-you hidden">{{ _('Your download link was sent.') }}</h2>
...
     {% else %}
-      <input type="hidden" name="utm_campaign" value="fxa-embedded-form" id="fxa-email-form-utm-campaign" />
+      <input type="hidden" name="utm_campaign" value="fxa-embedded-form" />
     {% endif %}
 
     {# utm_content is optional. It should only be declared when there is new messaging associated with the form. (For example: `get-the-rest-of-firefox`) #}
-    {% if utm_content %}
-      <input type="hidden" name="utm_content" value="{{ utm_content }}" id="fxa-email-form-utm-content" />
-    {% endif %}
+    <input type="hidden" name="utm_content" value="{{ utm_content }}" id="fxa-email-form-utm-content"/>
```
alexgibson: If these are no longer included conditionally, should they have default values?

bedrock/base/templates/macros.html
```diff
@@ -264,24 +264,18 @@ <h2 class="thank-you hidden">{{ _('Your download link was sent.') }}</h2>
 
     {# utm_campaign is used to identify specific marketing campaigns. (For example: trailhead or skyline) If there is one, the campaign name is prepended to the default value of `fxa-embedded-form`. #}
     {% if utm_campaign %}
-      <input type="hidden" name="utm_campaign" value="{{ utm_campaign }}-fxa-embedded-form" id="fxa-email-form-utm-campaign" />
+      <input type="hidden" name="utm_campaign" value="{{ utm_campaign }}-fxa-embedded-form" />
     {% else %}
-      <input type="hidden" name="utm_campaign" value="fxa-embedded-form" id="fxa-email-form-utm-campaign" />
```
alexgibson: I think you can still use an `id` to target the input here?

bedrock/base/templates/macros.html
amychurchwell: I don't think they should have default values. if `value=""` it does not get passed and it is not a required value.

alexgibson: Ok as long as it doesn't get passed that should keep things nice and simple 👍

alexgibson: It looks like loading the page without any utm params in the URL results in a 400 response from the metrics request:

```
GET https://latest.dev.lcip.org/metrics-flow?form_type=email&entrypoint=mozilla.org-firefox-accounts&utm_source=mozilla.org-firefox-accounts&utm_campaign=trailhead-fxa-embedded-form&utm_content=&utm_term= 400
```

The value for `utm_content` also seems to be defaulting

media/js/base/mozilla-fxa-form.js
```diff
@@ -49,21 +48,39 @@ if (typeof window.Mozilla === 'undefined') {
 
             fxaSubmitButton.disabled = false;
 
-            var destURL = fxaForm.getAttribute('action') + 'metrics-flow';
+            // track external UTM referrals
+            var urlParams = new window._SearchParams().utmParamsFxA();
```
alexgibson: The `utmParamsFxA()` function here seems to have been written some time ago, and has some strange behavior which I'm not sure we're expecting here. For example, if `utm_content` is missing from the URL then it returns the pathname (e.g. `/firefox/accounts/`) instead.

I'm not entirely convinced we should be using this function here, as it seems like an old piece of code and is no longer used anywhere on the site. I'd suggest using `utmParams()` instead, since this just returns the values without any hidden logic?

bedrock/base/templates/macros.html
amychurchwell: Oh shoot, you are right. Reaching out to Saptarshi for a recommendation on what a default value should be.

amychurchwell: oh shoot, you are right. I spoke with Saptarshi and Justin about a potential default `utm_content` value and Justin asked why we don't insert the hidden `utm_content` input if the value is present on the URL. I thought that could potentially be a good solution-

if `utm_content` exists on URL: check that input field exists > If exists: overwrite // If does not exist: insert one.

Issue #6794, opened 2019-02-05
Move from Jenkins to GitLab for CI/CD


pmac:
- [x] Setup CI for Bedrock in GitLab
- [x] Hook GitLab CI up to Github for build reporting
- [x] Configure build and deployment pipeline

See https://github.com/mozilla/kitsune/pull/3455 for a similar effort in Kitsune and [our infra docs](https://github.com/mozmeao/infra/blob/master/docs/gitlab_ci.md) for pipeline help.

jgmize:
This was completed by https://github.com/mozilla/bedrock/pull/8222

PR #8141, opened 2019-11-13
Add .gitlab-ci.yml and bin/update_config.sh 

jgmize:
Build and deploy to iowa-a/bedrock-dev on gitlab-ci-dev & master for #6794


jgmize:
https://gitlab.com/mozmeao/bedrock/pipelines/95744275 is linked above as a success, which in turn triggered https://gitlab.com/mozmeao/bedrock/pipelines/95744275 which together demonstrate and end-to-end build, deployment, and test pipeline via gitlab. There are some improvements yet to be made, such as notifying slack at the beginning of the pipeline, but as discussed with @pmac we can follow up on that after merging to master.

Issue #7542, opened 2019-08-08
Implement use of ES2015 in bedrock via transpiling build step


alexgibson:
## Description

As part of bedrock's technical roadmap we'd like to enable front-end developers to write JavaScript using ES2015 (ES6), with an aim to making code cleaner, more pleasing to write and easier to maintain. To enable this we'll need to add a transpiling build step to our existing [Gulp script](https://github.com/mozilla/bedrock/blob/master/gulpfile.js), using a tool like [Babel](https://babeljs.io/).

Some key considerations:

- Transpiled code will need to work in IE9 if we're to use it broadly (until we can drop 1st class support for that browser at least).
- We'll likely not want to transpile everything. If existing JavaScript code is already written in ES5, there's no need to transpile it. Some pages & bundles will also likely need to support legacy browsers, so may need excluding. Perhaps we could specify a particular file format to identify code that needs transpiling? (e.g. my-script.es6.js or similar).
- We'd need to consider how transpiling works in relation to running unit tests (via `gulp js:test`).
- We'd need to update our ESLint config to enable ES2015 in our source.
- We may want to compare bundle sizes of before / after transpilation with regard to performance.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Transpiling is implemented in build system.
- [ ] One page (TBD) in production is using transpiled code.

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] Need to ensure performance impact is acceptable.
- [ ] Need to ensure that supported browsers can still run transpiled code.

#### Tasks
- [ ] Implement build step.
- [ ] Document the changes.
- [ ] Port a single page (TBD) to use transpiled code.

Errolyn:
From what I cant tell IE9 will be work fine with the ES5 polyfills, but I'm going to venture a guess that there will be some caveat we'll run into.

hoosteeno:
@Errolyn is this still in progress?

Errolyn:
Yes is it part of the webpack work  that is still on going, though this portion in the webpack work is mostly done. I don't want to close this until all the webpack stuff is done and merged.

alexgibson:
Closing this until we circle back on webpack evaluation.

PR #8169, opened 2019-11-16
Add javaScript transpiling to Bedrock [Fixes: #7542]

Errolyn:
## Description
Adds support for JavaScript transpiling to Bedrock
## Issue / Bugzilla link
#7542


Errolyn:
@alexgibson I wanted to get this up since this is the first "working" version of transpiling javascript and I thought that it would be good to get some input from you about file structure, and where things should live and be called before I go to far with what I'm doing.

gulpfile.js
```diff
@@ -471,7 +485,7 @@ const defaultTask = gulp.series(
     clean,
     assetsCopy,
     gulp.parallel(sassCompileAllFiles, lessCompileAllFiles),
-    gulp.parallel(jsCompileBundles, cssCompileBundles),
+    gulp.parallel(jsCompileBundles, cssCompileBundles, transpileJs),
```
alexgibson: I think we'll likely need to transpile JS prior to bundling, so this should likely happen in parallel on the above line, when sass and less are also compiled?

gulpfile.js
```diff
@@ -259,6 +263,15 @@ function cssWatch(cb) {
     cb();
 }
 
+/**
+ * Transpile ES6 JavaScript files.
+ */
+function transpileJs() {
+    return gulp.src('media/js/components/*.es6.js')
+        .pipe(babel({presets: ['@babel/preset-env']}))
+        .pipe(gulp.dest('media/js/compiled'))
```
alexgibson: Instead of checking-in compiled JS into GitHub, we may want follow how Sass & Less assets are compiled and bundled as convention:

1. Transpile JS files into `static_build`
2. The content of `static_build` is then bundled and minified into `static_final` ready for consumption.

gulpfile.js
```diff
@@ -259,6 +263,15 @@ function cssWatch(cb) {
     cb();
 }
 
+/**
+ * Transpile ES6 JavaScript files.
+ */
+function transpileJs() {
+    return gulp.src('media/js/components/*.es6.js')
```
alexgibson: We probably want to avoid copying these `*.es6.js` files over to `static_final` without being compiled first: https://github.com/mozilla/bedrock/blob/master/gulpfile.js#L375

PR #8173, opened 2019-11-18
/whatsnew pages should extend from a noindex template (Fixes #8172)

### LINKED ISSUE #8172: /whatsnew pages should extend from a noindex template
## Description

By default, /whatsnew pages used to extend from a template that included a `noindex` meta tag. Looking at the pages we're maintaining currently, it looks like none of them contain such a tag, and all are extending from the default Firefox base template.

alexgibson:
## Description
- Creates a new (Protocol-based) base template for all existing /whatsnew pages to extend.

## Issue / Bugzilla link
#8172

## Testing
- [x] All current /whatsnew pages should contain a noindex meta tag.


bedrock/firefox/templates/firefox/whatsnew/base.html
```diff
@@ -2,23 +2,9 @@
...
+{% extends "firefox/base/base-protocol.html" %}
 
 {# "noindex" pages should not have the canonical or hreflang tags: bug 1442331 #}
 {% block canonical_urls %}<meta name="robots" content="noindex,follow">{% endblock %}
 
 {% block gtm_page_id %}data-gtm-page-id="/firefox/whatsnew/"{% endblock %}
-
-{% block extrahead %}
-  {% if switch('firefox-update-notification-modal', ['en-US']) %}
-    {{ css_bundle('firefox-update-notification-firstrun-whatsnew') }}
```
alexgibson: Note for reviewer: I'm not carrying this old notification over, since it's not being shown currently ans has never been ported to Protocol.

alexgibson:
Rebased after #8207 merged.

bedrock/firefox/templates/firefox/whatsnew/base.html
stephaniehobson: 👍

alexgibson:
> What about the WN 71 page in the exp folder? Should that extend the same template as the other WN pages?

The CRO page is already noindex, so I think we're good here.

PR #8175, opened 2019-11-18
Add link to download in another language on /thanks (Fixes #8116)

### LINKED ISSUE #8116: Add a link to choose a different language on /firefox/download/thanks/
## Description

There used to be a "Download in another language" link on /firefox/download/thanks/, but it looks like it was removed in a recent redesign. We should reinstate it to make selecting a different language easier.

### Steps to reproduce

1. Visit https://www.mozilla.org/firefox/ and click download.
2. Once https://www.mozilla.org/firefox/download/thanks/ loads try and select a different language to the one that was auto-downloaded.

### Expected result

There should be a "Download in another language" link on the page that links to /firefox/all/.

### Actual result

There is no link. The only way to get another language is to change the dropdown in the page footer, which is not easy to find.

### Environment

N/A

alexgibson:
## Description
Adds "Download in another language" link to the bottom of `/firefox/download/thanks/` page.

## Issue / Bugzilla link
#8116

## Testing
- [x] String should already be localized.

### DIFF
--- a/media/css/firefox/new/trailhead/thanks.scss
+++ b/media/css/firefox/new/trailhead/thanks.scss
@@ -195,10 +195,6 @@ main {
 
 .download-another-language-link {
     @include text-body-sm;
-    padding: $layout-md 0;
+    padding: $layout-sm 0;
     text-align: center;
-
-    @media #{$mq-lg} {
-        padding-bottom: $layout-lg 0;
-    }
 }
--- a/media/css/firefox/new/trailhead/thanks.scss
+++ b/media/css/firefox/new/trailhead/thanks.scss
@@ -192,3 +192,13 @@ main {
         }
     }
 }
+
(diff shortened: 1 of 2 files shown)


media/css/firefox/new/trailhead/thanks.scss
```diff
@@ -192,3 +192,13 @@ main {
         }
     }
 }
+
+.download-another-language-link {
+    @include text-body-sm;
+    padding: $layout-md 0;
```
amychurchwell: nit: in longer languages, the link is out of view. Perhaps we can decrease top padding?

Screenshot_2019-11-21 Загрузите самый быстрый в истории Firefox

alexgibson: Changed default padding to `$layout-sm` 👍

PR #8199, opened 2019-11-20
[fix #7887] Show WNP70 content to 72beta

### LINKED ISSUE #7887: Show beta users the release WNP at 72.0beta
## Description

Beta users will be offered the 72.0 release a week prior to the rapid release schedule

We have the opportunity to show content (WNP used in the rapid release channel or otherwise) to these beta users.

Reference https://github.com/mozilla/bedrock/pull/7073 for when this was done for 68.0, at that major bump timing (Monday of RC week)

[RELENG BUG HERE](https://bugzilla.mozilla.org/show_bug.cgi?id=1597873)

---

- [ ] Page is live at assigned URL anytime before November 21st, 2019

- [ ] Timing

#### Tasks
- [ ] Assign WNP 70.0 content to the beta URL - https://www.mozilla.org/firefox/72.0beta/whatsnew/

craigcook:
## Description
Shows the WNP70 content to 72 Beta. Other beta versions get the default WNP.

I ended up duplicating the template just for the sake of setting different tracking params, so I'm not happy with the repetition of code but I'm not sure there's any other way. It's also possible we don't even need to do this and are happy to have signups from beta counted as coming from the regular page.

**This needs to be in production on 21 November (Thursday).**

## Issue / Bugzilla link
#7887 
https://bugzilla.mozilla.org/show_bug.cgi?id=1597873

## Testing
http://localhost:8000/firefox/72.0beta/whatsnew/all/

- [x] 72.0beta shows the 72beta template (same as content as WNP70, but verify the entrypoint param).
- [x] 71.0beta shows the default template.


craigcook:
Bleh, I've just realized that the copy that is hard-coded in the WNP70 template is not what appears in the whatsnew_70.lang file so the page isn't being properly localized. I think we'll need to duplicate the three hard-coded templates for EN/FR/DE and let all other locales get the default.

bedrock/firefox/views.py
```diff
@@ -541,7 +541,17 @@ def get_template_names(self):
             else:
                 template = 'firefox/whatsnew/index.html'
         elif channel == 'beta':
-            template = 'firefox/whatsnew/index.html'
+            if version.startswith('72.'):
```
alexgibson: Not a blocker: you could probably have one less `else` conditional here by saying

```suggestion
            if version.startswith('72.') and locale in ['en-US', 'en-CA', 'en-GB']:
```

alexgibson: Doh, nevermind I read this wrong!

craigcook:
> LGTM - it may have been cleaner to extend the original templates and override the analytics parameters, rather then duplicate the templates. But we're in a crunch and this works.

That was the first thing I tried and it didn't seem to work, but it's quite likely I did something wrong. Does it entail doing something other than just `{% set _entrypoint = 'mozilla.org-whatsnew72beta' %}`? Because the final buttons seem to still render the value set in the base template, unless I also repeat the buttons in the extended template. I think the setting and output need to be in the same template, unless there's another method.

> The only thing I think would be nice to add here still is a functional test for the new URL, as this kind of reuse makes dependencies complicated, and has bitten us in the past. Not a blocker if you would rather add a test in a followup PR though.

Good call, I'll add a test.

PR #8202, opened 2019-11-21
Remove Firefox notification banner (Fixes #6726)

### LINKED ISSUE #6726: Improve or deprecate the Firefox out-of-date notification banner
In 2017 we implemented a site wide notification banner that is shown to out of date Firefox users (2 major verions out of date or more). Clicking the "Update Firefox" button takes people to `/download/thanks/`. You can see a screenshot of the banner below:

screenshot_2019-01-18 the new fast browser for mac pc and linux firefox

This was implemented in a relatively short sprint cycle, and was to meet a Q3 goal to increase the volume of users updating old versions of Firefox from 500k weekly to 750k. The notification met that goal (in terms of the number of clicks the button got), but the banner has a number of flaws that we never addressed. I think we should evaluate how effective this banner is, and either fix those flaws, or else consider removing it from the site.

Issues:

- If a user has an update waiting, all they really need to do is restart their browser. They have already downloaded the update, and to make them re-download the entire product again is sub-optimal.
- If someone downloads Firefox using Firefox and then attempts to run the installer, they are presented with an error because Firefox is already running (a frustrating UX).
- We have no idea if people who clicked on the "Update Firefox" button actually updated their browser or not, all we know is if they click the button.
- The banner is also shown on outdated `/firstrun` and `/whatsnew` pages. Telling someone their browser is out of date when they literally just updated it, is a also frustrating UX.
- Visually the banner is unlike anything on any page, it's a snowflake in terms of design, and it is perhaps overly obtrusive. If someone really doesn't want to update their browser, it is likely quite annoying (it's possible to dismiss, but the cookie only lasts a week or so).
- The banner is quite JS heavy, and increases the bundle size of all our pages even though it's only applicable to a single browser, and only on outdated versions.
- The banner has never been ported to Protocol, and so only exists on legacy Pebbles and Sandstone pages.
- The banner was never localized, and only exists for en-US audiences.

I can think of ways to try and improve some of these issues, but first I think we should take a look at the existig data for how well it is performing (do we know how many clicks turn into re-installs, for instance?).

Relevant bugs/issues:
https://bugzilla.mozilla.org/show_bug.cgi?id=1381776
https://github.com/mozilla/bedrock/pull/5001

alexgibson:
## Description
- Removes legacy Firefox notification banner 🎉 

## Issue / Bugzilla link
#6726

## Testing
Integration test run: https://ci.vpn1.moz.works/blue/organizations/jenkins/bedrock_multibranch_pipeline/detail/run-integration-tests/472/pipeline


mkaply:
Did this go live?

PR #8213, opened 2019-11-25
Add default /whatsnew template promoting Firefox Accounts (Fixes #8196)

### LINKED ISSUE #8196: Build the WNP fallback using Fluent
## Description

The fallback WNP needs to be built using Fluent for localized content

---

- [ ] Page is live and fully localized using Fluent for Tier 1 locales.
- [ ] Non Tier 1 locales see the existing default /whatsnew page (mobile promo).

- [ ] Full Fluent L10n workflow is not yet in a finished state.

#### Tasks
- [x] Create Fluent file with final strings.
- [x] Create template to use said strings.
- [ ] Configure the file to be localized for Tier 1 locales.

alexgibson:
## Description
- Adds new default /whatsnew page for Firefox 60 +.
- Page is localized using Fluent.

http://localhost:8000/en-US/firefox/60.0/whatsnew/all/

## Issue / Bugzilla link
#8196

## Testing
- [ ] Page should be shown for all fallback /whatsnew URLs on Firefox 60+.
- [ ] Fluent integration should work as expected.
- [ ] Fluent file should be formatted correctly.

Successful test run: https://gitlab.com/mozmeao/www-config/pipelines/102252665


alexgibson:
This is ready, but is blocked on Fluent integration with Pontoon.

alexgibson:
@Pike r on the .ftl formatting?

@pmac r on the python bits? (particularly on integrating the .ftl file in the view correctly)

bedrock/firefox/views.py
```diff
@@ -577,11 +587,24 @@ def get_template_names(self):
...
 
         # return a list to conform with original intention
         return [template]
 
+    def render_to_response(self, context, **response_kwargs):
+        templates = self.get_template_names()
+        ftl_dict = {
+            'firefox/whatsnew/index-sync.html': 'firefox/whatsnew/whatsnew-sync'
+        }
+        ftl_files = ftl_files = ftl_dict.get(templates[0])
```
pmac: Duplicated `ftl_files =`

bedrock/firefox/views.py
```diff
@@ -577,11 +587,24 @@ def get_template_names(self):
...
+                template = 'firefox/whatsnew/index-sync.html'
+            else:
+                template = 'firefox/whatsnew/index.html'
 
         # return a list to conform with original intention
         return [template]
 
+    def render_to_response(self, context, **response_kwargs):
+        templates = self.get_template_names()
+        ftl_dict = {
```
pmac: NIT: Since this is a static dict it would be (slightly) better to defined it as a class variable instead of in this function since it'll be created once that way vs. every time in the method body. Super picky micro-optimization, but it's something.

bedrock/firefox/views.py
```diff
@@ -577,11 +587,24 @@ def get_template_names(self):
...
-            template = 'firefox/whatsnew/index.html'
+            if show_default_sync_whatsnew(version) and ftl_file_is_active('firefox/whatsnew/whatsnew-sync'):
+                template = 'firefox/whatsnew/index-sync.html'
+            else:
+                template = 'firefox/whatsnew/index.html'
 
         # return a list to conform with original intention
         return [template]
 
+    def render_to_response(self, context, **response_kwargs):
```
pmac: This is going to be a much more common pattern I think. I'm going to add (I should have done this before) a `get_ftl_files()` method that you can override to do this kind of thing instead of doing all of this every time.

alexgibson:
Thanks @pmac updated

bedrock/firefox/views.py
alexgibson: Agree I think this could be a common pattern 👍

alexgibson: Updated to use the class dictionary helper

l10n/en/firefox/whatsnew/whatsnew-sync.ftl
```diff
@@ -0,0 +1,25 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### URL: https://www-dev.allizom.org/firefox/whatsnew/
+
+-brand-name-firefox = Firefox
+-brand-name-firefox-browser = Firefox Browser
```
Pike: We should put these into a shared file, I'd suggest `l10n/en/brands.ftl`, and add that to the list of ftl files, appending at the end.

Pike: And add a file-level comment there pointing to https://mozilla-l10n.github.io/styleguides/mozilla_general/index.html#brands-copyright-and-trademark, please? With a "Please read the trademarks localization guide" or so

alexgibson:
@Pike thanks for the suggestions. I added a global brands.ftl file.

Re: the mix between "Browser" and "browser" this is intentional. I double checked with our copy writer and it should be capitalized for the product name, but fine to use lowercase when it's a regular noun.

This PR will target tier 1 locales, so I believe at least en-US and German will be covered via staff.

gueroJeff:
@alexgibson en-US, de, and fr should be covered by staff (falling back to vendors, where needed). en-GB, and en-CA are also part of that bucket, but I'm trying to get confirmation from the copy team on whether/how they will resource for those. Also, let's avoid using the "tier" nomenclature and replace it with "staff translation bucket" and "vendor translation bucket" please.

alexgibson:
Ok, we're unblocked on merging this to master - so this is now good for final review.

alexgibson:
I've made an update here by splitting common (reusable) strings that typically appear in every /whatsnew page insto a separate Fluent file. If I've got the activation order right, then only locales that have `whatsnew-accounts.ftl` active should still see the new page.

pmac:
> If I've got the activation order right, then only locales that have whatsnew-accounts.ftl active should still see the new page.

It's the first in the list, so it looks right to me 👍

PR #8218, opened 2019-11-25
update TEST_IMAGE in www-config .gitlab-ci.yml

jgmize:
## Description

Now that https://github.com/mozmeao/www-config/pull/245 has landed with support for the `TEST_IMAGE` variable, we can automatically update it for pushes to master, stage & prod to ensure that we are always testing with an image that matches the commit we are deploying.


pmac:
I'll let you merge when ready.

jgmize:
This change was included in https://github.com/mozilla/bedrock/pull/8222

PR #8222, opened 2019-11-26
Gitlab stage & prod

jgmize:
This switches the stage and prod branches from Jenkins to Gitlab for CI/CD. 
It depends on https://github.com/mozmeao/www-config/pull/247 which in turn depends on https://github.com/mozmeao/www-config/pull/246.


docs/pipeline.rst
```diff
@@ -173,23 +173,30 @@ default can be set and then can be overridden in the individual job configuratio
 Adding test runs
 ~~~~~~~~~~~~~~~~
 
-Test runs can be added by creating a new properties section in the
-`integration tests script <https://github.com/mozilla/bedrock/blob/master/docker/bin/run_integration_tests.sh>`_
-with the parameters of the new test run. This is simply a bash script and you can duplicate a clause of the case staement.
+Test runs can be added by creating a new job in the `.gitlab-ci.yml file in the www-config repository`_
```
jgmize: This section arguably belongs under https://www-config.readthedocs.io/#testing but for this PR I'm leaving them here to keep the changes in context, with the intention of moving them in a future pair of PRs to www-config and bedrock.

bin/update_config.sh
```diff
@@ -18,5 +18,9 @@ for CLUSTER in ${CLUSTERS:=iowa-a}; do
         fi
     done
 done
+
+TEST_IMAGE=mozorg/bedrock_test:${GIT_COMMIT}
+sed -i -e "s|TEST_IMAGE: .*|TEST_IMAGE: ${TEST_IMAGE}|;s|image: mozorg/bedrock_test.*|image: ${TEST_IMAGE}|" .gitlab-ci.yml
+
```
pmac: This is from #8218 right?

jgmize: right, I based this on #8218 and will rebase if that one merges first, or we can close it out in favor of this one if we want to just merge this.

PR #8239, opened 2019-12-03
Port /firefox/channel/ page to Protocol. (Fixes #7452)

### LINKED ISSUE #7452: Update /firefox/channel/ page
## Description

The [Firefox channel](https://www.mozilla.org/en-US/firefox/channel/desktop/) page is still using Sandstone / Less and needs updating to use the new branding. I'm not sure if the NA brand team have any immediate plans to redesign this page, so we need to figure out a low effort way to port it to Protocol.

---

- [ ] Page is updated in production using Protocol,

#### Tasks
- [ ] Firgure out the appropriate design treatment.
- [ ] Update the template.
- [ ] Remove old assets.

amychurchwell:
## Description
Ports the Firefox channel page(s) to Protocol and removes unused assets.

## Issue / Bugzilla link
#7452

## Testing
- [ ] http://localhost:8000/firefox/channel/desktop/
- [ ] http://localhost:8000/firefox/channel/android/
- [ ] http://localhost:8000/firefox/channel/ios/


amychurchwell:
Adding 'do not merge' label to allow for a small design review. This page will ultimately require a larger redesign, but I have been in communication with Cristina on what a good mvp might look like for the time being.

media/css/firefox/channel.scss
```diff
@@ -0,0 +1,68 @@
...
+    background: transparent;
+}
+
+.mzp-l-notes {
+    margin: $layout-lg $layout-xs;
+}
+
+@media #{$mq-lg} {
+    .mzp-t-beta {
+        @include at2x('/media/protocol/img/logos/firefox/browser/beta/logo-md.png');
```
alexgibson: You're no doubt already aware, but the logo sizing needs some work on larger viewports:

image

bedrock/firefox/templates/firefox/channel/android.html
```diff
@@ -17,9 +18,11 @@
 
 {% block page_desc %}
 {% if l10n_has_tag('firefox_title_meta_bug1357373') %}
-  {{_('Experience cutting-edge features in a pre-release browser for Android: Firefox Beta and Firefox Nightly. Install now!')}}
+  {{_('Experience cutting-edge features in a pre-release browser for Android: Firefox Beta and Firefox Nightly. Install
```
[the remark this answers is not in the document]
amychurchwell: bit of `format document` fallout, will have to tweak the defaults there.

bedrock/firefox/templates/firefox/channel/base.html
```diff
@@ -1,71 +1,86 @@
...
-          <li><a rel="external" class="more" href="https://developer.mozilla.org/docs/Mozilla/QA/Bug_writing_guidelines">{{ _('Tips for filing a bug') }}</a>
-        </ul>
-      </section>
     </div>
+  </nav>
+
+  {% call hero(
+    title=_('Take a browse on the wild side.'),
+    desc=_('Be among the first to explore future releases of Firefox for desktop, Android and iOS.'),
+    class='mzp-t-firefox mzp-t-product-firefox',
```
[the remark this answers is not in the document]
amychurchwell: Agreed! When I spoke to Cristina they mentioned that design is "trying to open up the pages a bit more and not use these dark headers unless we are working within templates and don't have much choice (moments pages) or focusing on privacy", for some context.

bedrock/firefox/templates/firefox/channel/android.html
```diff
@@ -31,75 +32,97 @@
...
-  <div class="container">
-    <header>
-      <h2>{{_('Beta')}}</h2>
-      <p>{{_('Try the latest Android features, before they get released to the rest of the world.')}}</p>
-    </header>
-    {{ download_firefox('beta', platform='android', alt_copy=_('Download'), dom_id="android-beta-download") }}
-    <p class="notice">
+<section id="beta">
+  <section class="mzp-c-call-out-compact mzp-t-firefox">
+    <img class="logo" width="120" height="120" alt="" src="{{ static('protocol/img/logos/firefox/browser/beta/logo.svg') }}" />
```
alexgibson: Sorry I may not be understanding the context here. It feels like there's a fair amount of custom code that's trying to fight against the default call-out component, or trying to re-implement things that the component may already be able to handle? If it's becoming tricky to make it do the thing you want, then perhaps all these overrides aren't worth it, and this could be a custom one off thing instead?

Some thoughts:

- I see you're using a custom container width here via `mzp-l-narrow`, but the call out component already has a narrower width container when used in the regular way, which seems similar-ish already to what you want? You can see it in the demo page here: https://protocol.mozilla.org/demos/call-out-compact.html
- The default component also has built in icons for the different products, using `mzp-t-product-firefox`, `mzp-t-product-beta` classes etc. Using these would mean you don't have to hide the logo at small screen sizes as you are in the workaround here. The default logos are smaller than the current icons, but that should be possible to change via some CSS if you wanted.
- You could also consider putting the release notes links below the notes section, similar to how they appear in terms of order in the current page, as opposed to try and fit them into the call out description?

Last thing: the way things are stacked currently, there are several layers of layout containers in play:

```html
<div class="mzp-l-content mzp-l-narrow">
  <section class="mzp-c-call-out-compact">
    <div class="mzp-l-content"></div>
  </section>
</div>
```

Ideally, there should be only one `mzp-l-content`, as nesting children creates extra padding around content that's visible at smaller screen widths.

Perhaps it would help to talk through this in person to understand the requirements. What you have looks ok visually, but I'm wondering if there may be a simpler solution to what you have? :)

bedrock/firefox/templates/firefox/channel/android.html
```diff
@@ -31,75 +32,97 @@
...
+      <div class="mzp-c-call-out-cta">
+        <div class="mzp-c-call-out-cta-container">
+          {{ download_firefox('nightly', platform='android', alt_copy=_('Download'), dom_id="android-nightly-download") }}
+        </div>
+      </div>
     </div>
-  </footer>
+  </section>
+
+  <div class="mzp-l-notes">
```
alexgibson: Nit: avoid adding mzp namespace to things that aren't implemented in Protocol?
```suggestion
  <div class="l-notes">
```

bedrock/firefox/templates/firefox/channel/base.html
```diff
@@ -1,71 +1,82 @@
...
+  <ul>
+    <li><a class="mzp-c-cta-link" href="https://bugzilla.mozilla.org/enter_bug.cgi">{{ _('File a bug now') }}</a></li>
+    <li><a class="mzp-c-cta-link" href="https://developer.mozilla.org/docs/Mozilla/QA/Bug_writing_guidelines">{{ _('Tips for filing a bug') }}</a></li>
+  </ul>
+  {% endcall %}
+
+  <aside class="mzp-l-content">
+    <div class="mzp-c-newsletter">
+      <div class="mzp-c-newsletter-image">
+        {{ high_res_img('img/home/2018/newsletter-graphic.png', {'alt': ''}) }}
```
alexgibson: I think the comment about the Mozilla branded signup is related to this graphic. A Firefox branded image might work better?

amychurchwell: I spoke with Cristina & we are going to reuse the styles from the newsletter on  firefox/developer for the time-being. 

Screen Shot 2019-12-05 at 10 02 22 AM

bedrock/firefox/templates/firefox/channel/android.html
amychurchwell: This makes a lot of sense, I think I may have over-complicated things. I've simplified this PR to be using more defaults. :) thanks for the guidance.

media/css/firefox/channel.scss
```diff
@@ -0,0 +1,180 @@
...
+
+    @media #{$mq-md} {
+        &.mzp-t-product-firefox,
+        &.mzp-t-product-beta,
+        &.mzp-t-product-developer,
+        &.mzp-t-product-nightly,
+        &.mzp-t-product-focus {
+            .mzp-c-call-out-content {
+                background-size: $layout-xl $layout-xl;
+                padding: 0 0 0 $layout-2xl;
```
alexgibson: This might need a bidi mixin to apply the same padding on TRL screens

```suggestion
                @include bidi(((padding, 0 0 0 $layout-2xl, 0 $layout-2xl 0 0),));
```

bedrock/firefox/templates/firefox/channel/ios.html
```diff
@@ -23,27 +22,20 @@
...
-{% block body_id %}platform-ios{% endblock %}
-
 {% block channels %}
-<section id="testflight" class="channel">
-  <div class="container">
-    <header>
-      <h2>{{_('TestFlight')}}</h2>
-      <p>{{_('Test beta versions of Firefox for iOS via Apple’s TestFlight program.')}}</p>
-    </header>
-    <a href="{{ url('firefox.ios.testflight') }}" class="button testflight-cta">{{_('Sign up now')}}</a>
```
alexgibson: Looks like this `testflight-cta` class name is being relied upon by a functional test: https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/channel/ios.py#L14

media/css/firefox/channel.scss
```diff
@@ -0,0 +1,180 @@
...
+    background: transparent;
+
+    @media #{$mq-md} {
+        &.mzp-t-product-firefox,
+        &.mzp-t-product-beta,
+        &.mzp-t-product-developer,
+        &.mzp-t-product-nightly,
+        &.mzp-t-product-focus {
+            .mzp-c-call-out-content {
+                background-size: $layout-xl $layout-xl;
```
alexgibson: Nit: use `background-size()` mixin?
```suggestion
               @include background-size($layout-xl $layout-xl);
```

PR #8244, opened 2019-12-04
Port /firefox/ios/testflight/ to Protocol. (Fixes #7439) 

### LINKED ISSUE #7439: Port /firefox/ios/testflight/ to Protocol
## Description

The [TestFlight for Firefox iOS](https://www.mozilla.org/en-US/firefox/ios/testflight/) page is built upon Sandstone / Less, but can probably be ported to Protocol pretty easily (swapping out the blue hero background for purple).

---

- [ ] Page is updated in production using Protocol.

#### Tasks
- [ ] This is an example of a task. Use a checklist!
- [ ] Update tests where needed.

amychurchwell:
## Description
Ports firefox/ios/testflight to Protocol and removes unused assets.

## Issue / Bugzilla link
#7439

## Testing
http://localhost:8000/firefox/ios/testflight/


bedrock/firefox/templates/firefox/testflight.html
```diff
@@ -1,131 +1,138 @@
...
-  {{ _('Firefox for iOS Beta') }} —
+{{ _('Firefox for iOS Beta') }} —
 {% endblock %}
 
 {% block page_title %}{{ _('TestFlight') }}{% endblock %}
 
 {% block page_desc %}
-  {{ _('Sign up to test pre-release beta versions of Firefox for iOS via Apple’s TestFlight program and help make our mobile browser for iPhone, iPad and iPod touch even better.') }}
+{{ _('Sign up to test pre-release beta versions of Firefox for iOS via Apple’s TestFlight program and help make our
+mobile browser for iPhone, iPad and iPod touch even better.') }}
```
alexgibson: Remove line break?

bedrock/firefox/templates/firefox/testflight.html
```diff
@@ -1,131 +1,138 @@
...
-        TestFlight, Apple's App Store program that manages pre-release
-        software, to participate in the program.
-      {% endtrans %}
-
-        <strong>{{ _('Please accept it.') }}</strong>
-      </p>
-
-      <p>{{ _('Be aware that this email may be in your spam folder.') }}</p>
+          <p>
+            <label for="privacy" class="mzp-u-inline">
```
alexgibson: The `for` attribute should map to the id of the input here?
```suggestion
            <label for="id_terms" class="mzp-u-inline">
```

media/static-bundles.json
```diff
@@ -1588,6 +1588,12 @@
         "js/firefox/enterprise/sla.js"
       ],
       "name": "firefox-enterprise-sla"
+    },
+    {
+      "files": [
+        "protocol/js/protocol-newsletter.js"
```
alexgibson: There's a JS error when the page loads:

```javascript
TypeError: Argument 1 of Window.getComputedStyle is not an object.
```

As a result, when I submit the form, it doesn't display the custom success message that's in the page. I think this may be because there's no `newsletter-details` container to expand?

bedrock/firefox/templates/firefox/testflight.html
```diff
@@ -1,131 +1,138 @@
...
+  desc=_('Sign up to test pre-release versions to make Firefox even better.'),
+  class='mzp-t-firefox mzp-t-dark mzp-t-product-firefox',
+  include_cta=False,
+  heading_level=1) %}
+  {% endcall %}
+
+  <aside class="mzp-l-content">
+    <div class="mzp-c-newsletter">
+      <div class="mzp-c-newsletter-image">
+        {{ high_res_img('img/home/2018/newsletter-graphic.png', {'alt': ''}) }}
```
alexgibson: This is a mozilla themes graphic - think we could either drop it, or find a more Firefox centric image?

bedrock/firefox/templates/firefox/testflight.html
```diff
@@ -1,131 +1,138 @@
...
+          When Mozilla receives data from you or TestFlight, we handle it in
+          accordance with our <a href="{{ url }}">Privacy Policy</a>.
+          {% endtrans %}
+        </p>
       </li>
       <li>
-        {{ _('You may receive emails from TestFlight or Mozilla in connection with your participation in this Program.') }}
+        <p>
+          {{ _('You may receive emails from TestFlight or Mozilla in connection with your participation in this
+          Program.') }}
```
alexgibson: Remove line break?

bedrock/firefox/templates/firefox/testflight.html
```diff
@@ -1,131 +1,138 @@
...
+        <p>{{ _('Be aware that this email may be in your spam folder.') }}</p>
+      </div>
     </div>
-  </div>{#-- /.container --#}
-</section>{#-- /#intro --#}
+  </aside>
 
-<section id="terms-conditions">
-  <div class="container">
+  <section class="mzp-l-content" id="terms-conditions">
```
alexgibson: Could this content feels a little wide for readability. Perhaps it could benefit from a slightly narrower container?

media/static-bundles.json
amychurchwell: I'm not sure what content to put in `newsletter-details` for this form. 🤔

alexgibson: If you take a look at other newsletter forms, such as on the [home page]([url](https://www.mozilla.org/en-US/)), then inspect the dom you can see what's inside there.

amychurchwell: I nested information under `newsletter-details` but I'm still struggling to trigger the custom success message. The current success message is from `newsletter/one_newsletter_signup.html`

alexgibson: The current success message is already in the page. If the JS is working, it should show the message depending on the class names used in the newsletter JS.

amychurchwell: Right! I feel a little lost because I'm not receiving JS console errors for some reason.

alexgibson: Looks like you need to include `"js/newsletter/form-protocol.js"` as well

alexgibson: When testing it out there may be some input ID's you need to update in the HTML to match what's expected in the JS. I think it also expects a value for country, which is something this particular form doesn't have - so you may need to make that an optional thing that gets passed.

amychurchwell: ah ok! I think I have provided a fix for this & things seem to be working as expected.

```
<input type="hidden" name="lang" id="id_lang"  value="en" aria-required="false">
<input type="hidden" name="country" id="id_country"  aria-required="false">
```

bedrock/firefox/templates/firefox/testflight.html
```diff
@@ -1,131 +1,135 @@
...
+      <fieldset class="mzp-c-newsletter-content">
+        <div class="mzp-c-form-errors" id="newsletter-errors"></div>
+
+        <p>
+          <label for="id_email">{{ _('Email') }}</label>
+          <input type="email" class="mzp-js-email-field" id="id_email" name="email" required aria-required="true">
+        </p>
+
+        <div id="newsletter-details" class="mzp-c-newsletter-details">
+          <input type="hidden" name="lang" id="id_lang"  value="en" aria-required="false">
```
alexgibson: It looks like there's already a duplicate `name="lang"` input above on line 37, so perhaps all you need is to add the correct `id` there?

bedrock/firefox/templates/firefox/testflight.html
```diff
@@ -1,131 +1,135 @@
...
+          <input type="hidden" name="country" id="id_country"  aria-required="false">
```
alexgibson: This seems to workaround the form not having a country selector ok, but I don't know if including a field without a value here is a problem for basket or not. Maybe @pmac can advise here?

pmac: Country is not a required field, so I don't think you need to include this at all.

alexgibson: I think this is a workaround because the JS expects a country field. We'd just like toverify this isn't going to cause any weirdness.

amychurchwell: Thanks ya'll! :+1:

pmac: Ah! Cool. I don't think it should be weird.

Out of curiosity, why are we avoiding country for this? In my discussions with the email team they'd really like to have Country info as often as they can get it.

alexgibson: I’m not too sure there - this PR is just updating a legacy page and it’s the way it was. If we wanted to add it we could maybe file a follow up issue for later?

tests/pages/firefox/ios_testflight.py
```diff
@@ -13,14 +13,14 @@ class iOSTestFlightPage(FirefoxBasePage):
...
     _email_locator = (By.ID, 'id_email')
-    _html_format_locator = (By.ID, 'id_fmt_0')
+    _html_format_locator = (By.ID, 'format-html')
     _privacy_policy_checkbox_locator = (By.ID, 'id_privacy')
-    _privacy_policy_link_locator = (By.CSS_SELECTOR, 'label[for="id_privacy"] span a')
-    _submit_button_locator = (By.ID, 'footer_email_submit')
+    _privacy_policy_link_locator = (By.CSS_SELECTOR, 'label[for="id_privacy"] a')
+    _submit_button_locator = (By.ID, 'newsletter-submit')
     _terms_checkbox_locator = (By.ID, 'id_terms')
     _terms_link_locator = (By.CSS_SELECTOR, 'label[for="id_terms"] span a')
```
alexgibson:
```suggestion
    _terms_link_locator = (By.CSS_SELECTOR, 'label[for="id_terms"] a')
```

tests/pages/firefox/ios_testflight.py
```diff
@@ -13,14 +13,14 @@ class iOSTestFlightPage(FirefoxBasePage):
...
-    _text_format_locator = (By.ID, 'id_fmt_1')
-    _thank_you_locator = (By.CSS_SELECTOR, '#newsletter-form-thankyou h2')
+    _text_format_locator = (By.ID, 'format-text')
+    _thank_you_locator = (By.ID, 'newsletter-thanks')
```
alexgibson: Try adding:

```python
_form_details_locator = (By.ID, 'newsletter-details')

def expand_form(self):
    # scroll newsletter into view before expanding the form
    self.scroll_element_into_view(*self._email_locator)
    assert not self.is_form_detail_displayed, 'Form detail is already displayed'
    # Click the submit button to expand the form when not visible.
    # Ideally here we would focus on the email field, but Selenium has issues
    # dealing with focus events when the window is not active.
    # See bug https://bugzilla.mozilla.org/show_bug.cgi?id=704583
    self.find_element(*self._submit_button_locator).click()
    self.wait.until(expected.visibility_of_element_located(
        self._privacy_policy_checkbox_locator))

@property
def is_form_detail_displayed(self):
    return self.is_element_displayed(*self._form_details_locator)
```

alexgibson: At the start of each test you'll then also need to add `page.expand_form()`

PR #8247, opened 2019-12-04
Whatsnew experiment template should extend /exp/ base template (Fixes #8246)

### LINKED ISSUE #8246: Whatsnew experiment template should extend /exp/ base template
## Description

The /whatsnew experiment template is not configured to load Convert correctly.

### Steps to reproduce

1. https://www.mozilla.org/en-US/exp/firefox/71.0/whatsnew/all/

### Expected result

It should load the Convert bundle

### Actual result

The Convert bundle is not present.

### Environment

N/A

alexgibson:
## Description

http://localhost:8000/en-US/exp/firefox/71.0/whatsnew/all/

## Issue / Bugzilla link
#8246

## Testing
- [x] Convert bundle should be loaded
- [x] Page should be no index

### DIFF
--- a/bedrock/exp/templates/exp/firefox/whatsnew/whatsnew-fx71.html
+++ b/bedrock/exp/templates/exp/firefox/whatsnew/whatsnew-fx71.html
@@ -4,9 +4,10 @@
 
 {% from "macros.html" import monitor_button, fxa_cta_link with context %}
 
-{% add_lang_files "firefox/whatsnew_71" %}
+{% extends "exp/base/base-firefox.html" %}
 
-{% extends "firefox/base/base-protocol.html" %}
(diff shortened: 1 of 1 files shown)


bedrock/exp/templates/exp/firefox/whatsnew/whatsnew-fx71.html
```diff
@@ -4,9 +4,10 @@
 
 {% from "macros.html" import monitor_button, fxa_cta_link with context %}
 
-{% add_lang_files "firefox/whatsnew_71" %}
```
stephaniehobson: I'm assuming the lang file isn't needed because the experiment is EN only?

alexgibson: Correct yeah

PR #8255, opened 2019-12-04
[fix #8223] Update newsletter country label

### LINKED ISSUE #8223: Update copy for country selector in newsletter sign up form
## Description

Per direction in https://wiki.mozilla.org/Lists_of_Countries_and_Regions, we need to update the copy on mozilla.org "Country or Region" instead of only 'Country' where on its own or coupled with 'select' where used in conjunction with the [product details region selection list ](https://product-details.mozilla.org/1.0/regions/en-US.json )

Example:  The newsletter sign up form is used across multiple pages. In the drop down, visible to non- en locale visitors (e.g. - https://www.mozilla.org/de/newsletter/), is the string 'Select country' in its various localized forms.

***Tasks***
To cover this, two strings need to be updated in the /newsletters.lang file:

- [ ] "Select country" needs to be changed to "Select Country or Region"
- [ ] "Country" needs to be changed to "Country or Region"
- [ ] Extract the new strings, make available to localizers

craigcook:
## Description
Updates newsletter "Country" labels to "Country or region", to encompass places that are distinct regions but not separate countries.

Strings were already added in https://github.com/mozilla-l10n/www.mozilla.org/pull/357

## Issue / Bugzilla link
#8223

### DIFF
--- a/bedrock/newsletter/forms.py
+++ b/bedrock/newsletter/forms.py
@@ -263,7 +263,10 @@ class NewsletterFooterForm(forms.Form):
             lang, country = lang.split('-', 1)
         else:
             country = ''
-            regions.insert(0, ('', _lazy('Select country')))
+            if locale.startswith('en'):
+                regions.insert(0, ('', _lazy('Select country or region')))
+            else:
+                regions.insert(0, ('', _lazy('Select country')))
         lang_choices = get_lang_choices(newsletters)
         languages = [x[0] for x in lang_choices]
         if lang not in languages:
--- a/bedrock/newsletter/forms.py
+++ b/bedrock/newsletter/forms.py
@@ -9,7 +9,7 @@ from django import forms
 from django.core.exceptions import ValidationError
 from django.forms import widgets
 from django.utils.safestring import mark_safe
-from lib.l10n_utils.dotlang import _, _lazy
+from lib.l10n_utils.dotlang import _, _lazy, lang_file_has_tag
 from product_details import product_details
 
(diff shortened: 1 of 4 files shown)


bedrock/newsletter/forms.py
```diff
@@ -263,7 +263,10 @@ def __init__(self, newsletters, locale, data=None, *args, **kwargs):
             lang, country = lang.split('-', 1)
         else:
             country = ''
-            regions.insert(0, ('', _lazy('Select country')))
+            if locale.startswith('en'):
```
pmac: Hmm... perhaps I was wrong... it looks like this won't ever be shown to any of our `en` locales since they all include the country, which pre-selects a country and doesn't show this text. So maybe we should just include a call to the new string here so it can be extracted and translated and then switch to it later?

craigcook: oh, it seemed to work locally... but yeah, I think I follow your logic. The string is already out to localizers so we can just remove this and we'll update later once we've got the l10n coverage.

pmac: Okay, after looking into it I think we can use the same tag from the template in here:

```python
if lang_file_has_tag('newsletter/existing', locale, 'country_region_122019'):
```

Assuming `newsletter/existing` is the lang file that would get this tag.

pmac: Oh, and you'd have to include `lang_file_has_tag` in the import from `lib.l10n_utils.dotlang`.

Issue #8186, opened 2019-11-18
Code - newsletter subscription preferences page UI updates


ejregithub:
## Description

In #7620 the new design was established for the newsletter subscription preferences page. This is the issue to code and implement those changes

[brief](https://docs.google.com/document/d/1ahtsCMOlf3tBNFnW11qehTLPK1ER-4XycKY40HQ3B_Y/edit). 

Consult @cgonz0 for more detail on the updates
---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Updates are live
- [ ] the page is ported to Protocol

#### Tasks
- [ ] Code updates
- [ ] String update - change “Country” to “Country or Region”, above the corresponding drop down menu with listing. 
- [ ] Maintain the localization as 'opt-in'/allow for mixed locales as in the current version of the page.

ejregithub:
![emal_preferences_approved](https://user-images.githubusercontent.com/20547886/69093407-3a72c080-0a03-11ea-91fc-a8cb59d77d26.png)

Errolyn:
After a conversation in the durable team room, we are going to see about using checkboxes instead of the switch, they will be more accessible, and easier to understand. I'm marking this blocked until those details are confirmed by Cristina.

ejregithub:
@Errolyn - please note the additional tasks - 

- String update - change “Country” to “Country or Region”, above the corresponding drop down menu with listing.
- Maintain the localization as 'opt-in'/allow for mixed locales as in the current version of the page.

Errolyn:
Having trouble with getting a demo working correctly but here is a screenshot for now
![Screenshot_2019-12-02 Newsletter Subscriptions](https://user-images.githubusercontent.com/1214589/70002524-1fc93d00-1515-11ea-8a74-85094a20c095.png)

cgonz0:
> Having trouble with getting a demo working correctly but here is a screenshot for now


@Errolyn looking good. can we change all the link/button/checkbox/radio etc colors to our firefox blue (#0060df)?

Errolyn:
Sure I will give it a go :)

Errolyn:
@cgonz0 so after looking at it, it looks like to change the color of the checkbox I basically have to hide the checkbox then create a new one from scratch. It looks like Stephanie is already working on a component to do that. So I think it might be best to leave the checkbox as/is for and create a separate issue to port this page over to protocol, and use the new checkbox component. 

[...]

PR #8258, opened 2019-12-04
Email subscription page updates [Fixes: #8186]

Errolyn:
## Description
Changing UI for the subscriptions page to have checkboxes, also changing the color of blue used

## Issue / Bugzilla link
#8186
## Testing
Checkboxes should be aligned cross browsers, clicking and unclicking boxes should update subscriptions.

### DIFF
--- a/media/css/newsletter/newsletter.less
+++ b/media/css/newsletter/newsletter.less
@@ -179,10 +179,8 @@ body {
 
             .subscription td {
                 min-width: 75px;
-                label {//this should not be displayed BUT if some one uses an addon or accessibility feature to display labels this will make sure they don't look weird on the page.
-                    font-size: 12px;
-                    color: #666;
-                    display: none;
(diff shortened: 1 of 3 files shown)


media/css/newsletter/newsletter.less
```diff
@@ -170,6 +179,11 @@ body {
 
             .subscription td {
                 min-width: 75px;
+                label {//this should not be displayed BUT if some one uses an addon or accessibility feature to display labels this will make sure they don't look weird on the page.
+                    font-size: 12px;
+                    color: #666;
+                    display: none;
```
alexgibson: If an element is `display: none` then it usually means that screen readers won't acknowledge it. Might using the `.visually-hidden()` mixin be a good solution here?

Errolyn: Done

PR #8277, opened 2019-12-06
Use newrelic-admin in run-prod.sh

### LINKED ISSUE #7372: Uninstrumented modules in New Relic agent initialization
## Description

Getting the following error on https://rpm.newrelic.com/accounts/1299394/applications/57236602 since the latest prod deploy:

> Some modules were uninstrumented during the current time window: meinheld.server, gunicorn.app.base. Make sure newrelic.agent.initialize() is called before any of these modules are imported. Visit the [documentation](https://docs.newrelic.com/docs/agents/python-agent/installation-configuration/python-agent-integration#manual-integration) for more information.

This appears to be a regression since the Django-2.2/Python3 upgrade.

pmac:
Instead of importing newrelic in the wsgi file.

Fix #7372

### DIFF
--- a/bin/run-prod.sh
+++ b/bin/run-prod.sh
@@ -1,4 +1,5 @@
 #!/bin/bash -xe
+NEW_RELIC_CONFIG_FILE=newrelic.ini newrelic-admin run-program \
 gunicorn wsgi.app:application -b 0.0.0.0:${PORT:-8000} \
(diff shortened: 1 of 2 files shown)


jgmize:
Since merging this, bedrock-dev on https://rpm.newrelic.com/accounts/1299394/applications/57024613 no longer has the warning about uninstrumented modules for newer time windows. Older time windows will still show the warning.

Issue #8236, opened 2019-12-02
Publish Mozilla CA Policy 2.7


pmac:
https://bugzilla.mozilla.org/show_bug.cgi?id=1600707

PR #8281, opened 2019-12-09
Fix incorrect formatting for v2.7 of the security certs policy

pmac:
## Issue / Bugzilla link

Re bug 1600707
See #8236

## Testing

Look at /about/governance/policies/security-group/certs/policy/ and see if things are formatted properly compared to the version on Github.


pmac:
Wayne has reviewed the changes in the demo and approved of them in the bug.

PR #8286, opened 2019-12-09
Add build-test job to .gitlab-ci.yml

jgmize:
This job enables building custom test images by pushing to the `build-test` branch. See https://gitlab.com/mozmeao/bedrock/pipelines/101797359 (also linked from the ci/gitlab/build-test details below) for example of this job.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -24,6 +24,16 @@ build:
     - FROM_DOCKER_REPOSITORY=mozorg/bedrock docker/bin/push2dockerhub.sh
     - bin/upload-staticfiles.sh
 
+build-test:
+  stage: build
+  only:
+    - build-test
+  tags:
+    - oregon-b-shell
+  script:
+    - make clean build-ci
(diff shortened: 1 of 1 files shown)


jgmize:
I've also updated the docs on https://www-config.readthedocs.io/#testing

pmac:
Can we do something like this but do the full thing and get `run-integration-tests` app back?

pmac:
We could check with the team but I think that might be preferable to doing this and then manually doing the pipeline in github.

jgmize:
> Can we do something like this but do the full thing and get run-integration-tests app back?

The original `run-integration-tests` app was deployed on Deis Workflow, but I can set up a dedicated deployment for running integration tests against.

pmac:
Yeah. That's what I'm wondering if it wouldn't be best to just do that. The team was used to pushing to that branch if they needed integration tests for in-progress code.

We can do this in the interim, but that seems like a better UX for the team, though it does mean one more deployment to setup :/

jgmize:
I agree, the previous UX was much better, so I've re-implemented it (mozilla/bedrock#8288) and I'm closing this one. Thanks @pmac.

PR #8291, opened 2019-12-10
Add /firefox/mobile/get-app/ page (Fixes #8266)

### LINKED ISSUE #8266: Build a landing page for SMS referrals
## Description

We should build a bare-bones SMS interface, identical to the SMS modal, that we can direct users to. It would resemble the (sync-signed-in) page at https://accounts.firefox.com/sms.

---

- [x] Interface is up at a URL such as www.mozilla.org/firefox/mobile/get-app/
- [x] Interface is localized in en-US only
- [x] Interface works exactly as the modal works, except with no opening and no closing
- [ ] **Interface is launched, online, tested by 2019-12-13**

alexgibson:
## Description
- Adds `/firefox/mobile/get-app/` page with send to device widget.

## Issue / Bugzilla link
#8266

## Testing
- [x] Widget should function as expected.
- [x] No regressions from moving the /mobile page assets to a new directory.

Successful test run: https://gitlab.com/mozmeao/www-config/pipelines/102045227


stephaniehobson:
I'm prompted to give a phone number but then it doesn't work. In prod I'm not given the phone number option. Is this just a demo difference?  

Also, en-GB gets to email but en-CA gets a QR Code. *pouts*

bedrock/firefox/templates/firefox/mobile/get-app.html
```diff
@@ -0,0 +1,62 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. #}
+
+{% from "macros.html" import google_play_button, send_to_device with context %}
+
+{% add_lang_files "firefox/sendto" %}
```
stephaniehobson: We're losing the translations that have already been made for the mobile page :( Any way around that?

ejregithub: I'm fairly certain the issue with using the widget are due to it being on demo, iirc from past testing whilst on demo.

@stephaniehobson - @hoosteeno indicated in the original issue that "We can copy the widget from /mobile entirely, and put in the center of the page on a white background"

If the strings in the page are intended to be verbatim to that used in /mobile, can simply point to that /firefox/mobile-2019.lang file?

alexgibson: @stephaniehobson @ejregithub the ask for this page is en-US only right now, no translations required from another page. The SMS works via geo location, not page locale. On demo it defaults to assuming you’re in the US. I think we’re fine to move forward here.

PR #8303, opened 2019-12-11
Remove tabzilla. (Fixes #8301)

### LINKED ISSUE #8301: Remove tabzilla from bedrock.
## Description
image

Tabzilla was a universal tab for Mozilla websites. 

[The repo is archived](https://github.com/mozilla/tabzilla/) and it is no longer used with our current branding. We should remove it and related assets from Bedrock.

It also utilizes `.less` which is being removed as a dependency.

amychurchwell:
## Description
Removes everything tabzilla.

## Issue / Bugzilla link
#8301


tests/redirects/map_globalconf.py
```diff
@@ -1050,10 +1050,6 @@
 
     # Bug 1255785
     url_test('/styleguide/identity/mozilla/logo-prototype/', '/styleguide/'),
-    # Bug 1268847
-    url_test('/styleguide/websites/sandstone/buttons/', '/styleguide/'),
```
alexgibson: I don't think there's any reason to remove these redirect tests, as they are for tabzilla content that's already been removed?

tests/redirects/map_globalconf.py
```diff
@@ -1224,9 +1220,6 @@
     url_test('/collusion/', 'https://addons.mozilla.org/firefox/addon/lightbeam/'),
     url_test('/lightbeam/{,about/}?', 'https://addons.mozilla.org/firefox/addon/lightbeam/'),
 
-    # Bug 1428150
-    url_test('/tabzilla/media/css/tabzilla.css', 'https://mozorg.cdn.mozilla.net/media/css/tabzilla-min.css'),
```
alexgibson: Same as above

media/css/tabzilla/tabzilla-static.less
```diff
@@ -1,33 +0,0 @@
...
-    position: relative;
-    float: right;
-
-    a {
-        position: relative;
-        display: block;
-        width: 147px;
-        height: 37px;
-        .image-replaced();
-        .at2x('/media/img/tabzilla/tabzilla-static.png', 147px, 37px);
```
alexgibson: Looks like there are some images here that can be removed.

PR #8324, opened 2019-12-16
Add ADR for Fluent L10n system

### LINKED ISSUE #7786: Write an ADR for the new l10n system
We record major architectural decisions in [ADRs](https://bedrock.readthedocs.io/en/latest/architecture/decisions/0001-record-architecture-decisions.html). We should do so for this change.

pmac:
Fix #7786


pmac:
cc @metadave

PR #8330, opened 2019-12-18
create whatsnew-72 [Fixes: #8331]

Errolyn:
## Description 
Create whatsnew-72 page

Link to page on demo: https://www-demo-whats-new-72-yeyfvjkm.herokuapp.com/en-US/firefox/72.0/whatsnew/all/
## Issue / Bugzilla link
#8331

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -553,8 +553,8 @@ class WhatsnewView(l10n_utils.LangFilesMixin, TemplateView):
                 template = 'firefox/whatsnew/index.html'
         elif locale == 'id':
             template = 'firefox/whatsnew/index-lite.id.html'
-        elif version.startswith('72.') and lang_file_is_active('firefox/whatsnew_72', locale):
-            template = 'firefox/whatsnew/whatsnew-fx71.html'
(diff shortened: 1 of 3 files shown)


bedrock/firefox/views.py
```diff
@@ -553,6 +553,8 @@ def get_template_names(self):
                 template = 'firefox/whatsnew/index.html'
         elif locale == 'id':
             template = 'firefox/whatsnew/index-lite.id.html'
+        elif version.startswith('72.') and lang_file_is_active('firefox/whatsnew_72', locale):
```
alexgibson:
```suggestion
        elif version.startswith('72.') and lang_file_is_active('firefox/whatsnew_71', locale):
```

bedrock/firefox/views.py
```diff
@@ -553,6 +553,8 @@ def get_template_names(self):
...
+            template = 'firefox/whatsnew/whatsnew-fx72.html'
```
alexgibson: Adding a [test](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/tests/test_base.py#L159) to make sure that the right template is shown here would be good too?

PR #8332, opened 2019-12-19
Define a localization config for Fluent /whatsnew page (Fixes #8326)

### LINKED ISSUE #8326: Define a localization config for Fluent /whatsnew page
## Description

In https://github.com/mozilla/bedrock/issues/8196 we built a /whatsnew page using Fluent for L10n. Right now, it's being shown for en-US only, until we get the L10n pipeline up and running.

What's still needed for this page in the interim, is a well defined L10n configuration. The original issue cited the page being needed for English. German, and French only. Based on discussion in last weeks meeting, it sounds like that's still tbd.

We need to implement a configuration for the page, using one of the three buckets being defined in https://github.com/mozilla/bedrock/pull/8092.

This basically equates to a choice of:

- Staff
- Staff + vendor
- Staff + vendor + community

Note: we can no longer say "staff + community for everything else". Choosing staff + community would leave a vendor sized hole.

---

- [x] A configuration exists for `firefox/whatsnew/whatsnew-account.ftl` and `firefox/whatsnew/whatsnew.ftl`.

- [x] `firefox/whatsnew/whatsnew.ftl` contains generic strings that may be shared and reused among different /whatsnew releases. We may want to consider a broad strategy for this particular file.

#### Tasks
- [x] Implement a config for both fluent files.

alexgibson:
## Description
- Adds a default configuration rule for Fluent /whatsnew pages, for broad staff + vendor + community locale coverage.

Note: community is already [configured here](https://github.com/mozilla/bedrock/blob/master/l10n/configs/pontoon.toml#L118-L120) with a wildcard rule.

## Issue / Bugzilla link
#8326

## Testing
- [ ] Did i get this right? (check for typos)

### DIFF
--- a/l10n/configs/pontoon.toml
+++ b/l10n/configs/pontoon.toml
@@ -116,5 +116,11 @@ locales = [
 ]
 
 [[paths]]
-    reference = "en/**/*.ftl"
-    l10n = "{locale}/**/*.ftl"
+    reference = "en/mozorg/mission.ftl"
+    l10n = "{locale}/mozorg/mission.ftl"
+[[paths]]
+    reference = "en/firefox/whatsnew/whatsnew-account.ftl"
+    l10n = "{locale}/firefox/whatsnew/whatsnew-account.ftl"
+[[paths]]
+    reference = "en/firefox/whatsnew/whatsnew.ftl"
+    l10n = "{locale}/firefox/whatsnew/whatsnew.ftl"
--- a/l10n/configs/vendor.toml
+++ b/l10n/configs/vendor.toml
@@ -43,3 +43,6 @@ locales = [
 [[paths]]
     reference = "en/mozorg/mission.ftl"
     l10n = "{locale}/mozorg/mission.ftl"
+[[paths]]
+    reference = "en/firefox/whatsnew/*.ftl"
+    l10n = "{locale}/firefox/whatsnew/*.ftl"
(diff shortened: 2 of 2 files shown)


alexgibson:
@hoosteeno r? I'm proposing we set a default configuration rule for all /whatsnew pages here, but we can switch to a more explicit rule if we think this may not be the general case we look to work toward.

@Pike r? Just to sanity check I'm doing the right thing here.

alexgibson:
Another option here is: we switch vendor to use a default wildcard rule for all Fluent files, like we already do for community. This may make sense, if marketing’s approach to vendor l10n is a simple one and this fits within the expected budget.

Pike:
IIRC, we want *most* whatsnew content to be not exposed to community?

And then just a bit of fallback page, that's long lived, gets exposed to community as well. Is that this page? I recall we talked staff and maybe vendor?

And yes, we'll need to be more restrictive on the wildcards for the Pontoon config to support that.

alexgibson:
> IIRC, we want most whatsnew content to be not exposed to community?

I'll defer to @hoosteeno, but if so then we need a simple way to exclude the directory from community (barring the well translated fallback page).

> And then just a bit of fallback page, that's long lived, gets exposed to community as well. Is that this page? I recall we talked staff and maybe vendor?

Right, that's what this page is - so we're going for staff + vendor + community.

hoosteeno:
We want this fallback to be as widely localized as it possibly can, so it should get staff+vendor+community (i.e. all locales).

In general, WNP will only get staff+vendor localization. And we should anticipate, if possible, that the list of vendor-localized locales will shift a bit as we exercise this new system.

alexgibson:
I've updated the pontoon configuration to more explicit. I'm not clear if there's a simpler way, such as being able to exclude a path. Happy to take further suggestions

PR #8342, opened 2019-12-20
Fix /whatsnew page 72 logos (Fixes #8341)

### LINKED ISSUE #8341: Logo paths on /whatsnew page 72 are broken
It looks like https://github.com/mozilla/bedrock/pull/8276 removed some logos that https://github.com/mozilla/bedrock/pull/8330 was using, so the page is throwing a 500.

https://www.allizom.org/en-US/firefox/72.0/whatsnew/all/

alexgibson:
## Description
- Fixes logo image paths for 72 WNP.
- Adds functional test for the page.

http://localhost:8000/en-US/firefox/72.0/whatsnew/all/

## Issue / Bugzilla link
#8341

## Testing
Test run: https://gitlab.com/mozmeao/www-config/pipelines/104577579

### DIFF
--- a/bedrock/firefox/tests/test_base.py
+++ b/bedrock/firefox/tests/test_base.py
@@ -398,7 +398,7 @@ class TestWhatsNew(TestCase):
         req.locale = 'es-ES'
         self.view(req, version='72.0')
         template = render_mock.call_args[0][1]
-        assert template == ['firefox/whatsnew/index.html']
+        assert template == ['firefox/whatsnew/index-account.html']
 
     # end 72.0 whatsnew tests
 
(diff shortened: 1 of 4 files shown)


bedrock/firefox/tests/test_base.py
```diff
@@ -398,7 +398,7 @@ def test_fx_72_0_0_fallback(self, render_mock):
         req.locale = 'es-ES'
         self.view(req, version='72.0')
         template = render_mock.call_args[0][1]
-        assert template == ['firefox/whatsnew/index.html']
+        assert template == ['firefox/whatsnew/index-account.html']
```
pmac: I don't see a change that would cause this test to change. Is it currently failing?

alexgibson: Yeah this is currently failing on master. The branch that merged wasn't up to date, so the failure didn't get picked up by Circle.

pmac: That is an issue with our current process. Maybe something we can fix next year.

PR #8343, opened 2019-12-20
Document Fluent file configuration (Fixes #8335)

### LINKED ISSUE #8335: Update docs on configuring Fluent files for extraction (pontoon / vendor)
## Description

During the process of creating new Fluent files for translation, we also need to configure those files in bedrock so that extraction happens into the correct buckets (i.e. pontoon / vendor). Currently our docs don't mention this part of the process. We should detail how we have our buckets configured, how to make changes to them, and what the implications are for making choices.

The syntax for the config files is documented here: https://moz-l10n-config.readthedocs.io/

alexgibson:
@pmac @Pike r?

### DIFF
--- a/docs/l10n.rst
+++ b/docs/l10n.rst
@@ -329,6 +329,52 @@ the template with the context. You can use the ``ftl_files`` argument with this 
     def about_view(request):
         render(request, 'about.html', {'name': 'Duder'}, ftl_files='mozorg/about')
 
+Fluent File Configuration
+-------------------------
+
+In order for a Fluent file to be extracted through automation and sent out for localization,
+it must first be configured to go through one or more distinct pipelines. This is controlled
+via a set of configuration files:
+
+- `Vendor`_, locales translated by an agency, and paid for by Marketing (locales covered by staff are also included in this group).
+- `Pontoon`_, locales translated by Mozilla contributors.
+- `Special templates`_, for locales with dedicated templates that don't go through the localization process (not currently used).
+
+Each configuration file consists of a pre-defined set of locales for which each group is
+responsible for translating. The locales defined in each file should not be changed without
+first consulting the with L10n team, and such changes should not be a regular occurence.
+
+To establish a localization strategy for a Fluent file, it needs to be included as a path
+in one or more configuration files. For example:
+
+.. code-block:: text
+
+    [[paths]]
+        reference = "en/mozorg/mission.ftl"
+        l10n = "{locale}/mozorg/mission.ftl"
+
+You can read more about configuration files in the `L10n Project Configuration`_ docs.
+
+Using a combination of vendor and pontoon configuration offers a flexible but specific set of
+options to choose from when it comes to defining an l10n strategy for a page. The available
+choices are:
+
+#. Staff locales.
+#. Staff + select vendor locales.
+#. Staff + all vendor locales.
+#. Staff + vendor + pontoon.
+#. All pontoon locales (for non-marketing content only).
+
+When choosing an option, it's important to consider that vendor locales have a cost associated
+with them, and pontoon leans on the goodwill of our volunteer community. Typically, only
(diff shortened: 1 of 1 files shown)


docs/l10n.rst
```diff
@@ -329,6 +329,39 @@ the template with the context. You can use the ``ftl_files`` argument with this
...
+Fluent File Configuration
+-------------------------
+
+In order for a Fluent file to be extracted through automation and sent out for localization,
+it must first be configured to go through one or more distinct pipelines. This is controlled
+via a set of configuration files:
+
+- `Vendor`_, locales translated by an agency, and paid for by Marketing (locales covered by staff are also included in this group).
+- `Pontoon`_, locales translated by Mozilla contributors.
+- `Special templates`_, for locales with dedicated templates that don't go through the localization process.
```
pmac: I'm still confused about this. Why do we need configuration for this if there won't be a fluent file for these templates? Is it just for the case that we have a fallback template and would want to exclude locales that are already hard coded to avoid superfluous work? Maybe we can add some more information about these at some point.

alexgibson: Yeah, I asked @Pike about this in https://github.com/mozilla/bedrock/pull/8092#pullrequestreview-313368378

I'm still not quite sure as to the real world use case. I guess we could perhaps use this for a pseudo-localization locale. Something that used Fluent files (e.g for testing purposes), but doesn't leave bedrock?

gueroJeff: I would encourage the use of different terminology here. "Vendor" will include staff locales too, meaning that for some locales it won't be translated by an agency. Maybe "external" is the better designation?

Pike: Yes, this is about avoiding un-used work. @gueroJeff told me that staff can do that inside BWX, though. They can decline a particular item of work.

Probably need to gather some experience to see if that's working well with our flow.

docs/l10n.rst
```diff
@@ -329,6 +329,39 @@ the template with the context. You can use the ``ftl_files`` argument with this
...
+.. code-block:: text
+
+    [[paths]]
+        reference = "en/mozorg/mission.ftl"
+        l10n = "{locale}/mozorg/mission.ftl"
+
+It's possible for any given Fluent file to be included in both vendor and pontoon configurations,
+in order for a page to have the broadest possible coverage. When making choices, it's important
+to consider that vendor has a cost associated with it, and pontoon leans on the good will of
+our volunteer community. Choose wisely.
```
gueroJeff: We should clarify this statement for marketing content (which has been very specifically defined for mozorg), as there's are only very specific options available to choose from. The config options  for this content boil down to:
1) Staff locales (2)
2) Staff locales (2) + select vendor locales (n)
3) Staff (2) + all vendor locales (18)
4) Staff + vendor + community

For anything non-marketing, it's assumed that it will always go to Pontoon.

docs/l10n.rst
alexgibson: @gueroJeff I’m using vendor here as a term because that’s what the configuration file itself is called. I’m happy to rename it, but note that the docs above do call out that vendor includes staff locales also.

docs/l10n.rst
alexgibson: I’ll see if I can include expand on this, thanks 👍

docs/l10n.rst
gueroJeff: Gotcha, thanks for clarifying. If other places make that clarifying reference, I think this should be fine.

docs/l10n.rst
gueroJeff: Thank you!

alexgibson:
@gueroJeff I've updated with a more specific list of options as per your recommendation. Let me know if this sounds ok? Thanks!

alexgibson:
Thanks @gueroJeff!

@pmac if this looks ok to you, shall we merge?

Issue #7574, opened 2019-08-14
Browser Comparison SEO pages 


ejregithub:
## Description

The idea is to provide a one-stop content resource that compares Firefox against our top five competitors in the market. This will consist of one main overview page that provides an intro and overview graphic, and that then links to the five sub-pages (repurposing SEM content) comparing Firefox side-by-side against competitors.

[Brief](https://docs.google.com/document/d/1BkZL_29Qx9rhf1_foCYMD8kju4bHdaGhpv2PcXvPImo/edit)

Structure:
Overview page: Compare Web Browsers
Chrome page: Compare Firefox vs Google Chrome
Safari page: Compare Firefox vs Safari
Edge page: Compare Firefox vs Microsoft Edge
Internet Explorer page: Compare Firefox vs Internet Explorer
Opera page: Compare Firefox vs Opera

We are doing this to boost our organic non-brand visibility and to drive incremental downloads from SEO content. 

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] This is an example of a success criteria. Use a checklist!
- [ ] This is an example of a success criteria.

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] This is an example of a risk. Use a checklist!
- [ ] This is an example of a risk.

#### Tasks (in this order for go live priority)
- [x] Launch Overview page: Compare Web Browsers
- [x] Launch Chrome page: Compare Firefox vs Google Chrome
- [ ] Launch Safari page: Compare Firefox vs Safari
- [ ] Launch Edge page: Compare Firefox vs Microsoft Edge
- [ ] Launch Internet Explorer page: Compare Firefox vs Internet Explorer
- [ ] Launch Opera page: Compare Firefox vs Opera

ejregithub:
Design for Overview page: Compare Web Browsers

![image](https://user-images.githubusercontent.com/20547886/63061057-d0e59200-bea8-11e9-8a89-9cb6908cf264.png)

guyfitzpatrick:
Main page English back-translation from German: https://docs.google.com/document/d/1W3CgeHU0-IOY-XZlBIhxPY88Oszx_li8aApHmhJB4AA/edit#heading=h.zgjwb84g9sby

rraue:
The urls will be:
Main page: /firefox/browsers/compare/
Chrome: /firefox/browsers/compare/chrome/
Safari: /firefox/browsers/compare/safari/
[...]

PR #8347, opened 2019-12-22
Combined PR with browser-comparison pages for Edge, IE, Safari, Opera, and image updates for the hub page

sgarrity:
## Description
Adds comparison pages for IE, Edge, Safari, and Opera, and makes some updates to the existing page for Chrome as well as the landing page.

## Issue
#7574 is the epic tracking all of this work

## Testing

- [ ] [Landing page](https://www-demo-browser-compar-nvjrww.herokuapp.com/firefox/browsers/compare/)
- [ ] [Chrome](https://www-demo-browser-compar-nvjrww.herokuapp.com/firefox/browsers/compare/chrome/)
- [ ] [Edge](https://www-demo-browser-compar-nvjrww.herokuapp.com/firefox/browsers/compare/edge/)
- [ ] [IE](https://www-demo-browser-compar-nvjrww.herokuapp.com/firefox/browsers/compare/ie/)
- [ ] [Opera](https://www-demo-browser-compar-nvjrww.herokuapp.com/firefox/browsers/compare/opera/)
- [ ] [Safari](https://www-demo-browser-compar-nvjrww.herokuapp.com/firefox/browsers/compare/safari/)


cgonz0:
@sgarrity @craigcook 
i believe the media query issue with browser logo sizes has already been fixed.
**for /edge**
can we move this block of copy above the existing svg? “At Firefox our privacy policy is transparent….” (security & privacy section)
**for /opera**
the svg (security & privacy) is not behaving properly - when i "view image" in its own browser tab, it looks as it should. i may have to look into this on my end...
**for /ie**
i know some copy changes are still forthcoming, but just noting format for “peril” text (first p / security and privacy section).

craigcook:
> **for /edge**
> can we move this block of copy above the existing svg? “At Firefox our privacy policy is transparent….” (security & privacy section)

Done.

> **for /opera**
> the svg (security & privacy) is not behaving properly - when i "view image" in its own browser tab, it looks as it should. i may have to look into this on my end...

It seems to be working for me, but then I guess I don't know what it's supposed to look like. I do note there's a fairly long pause between animation cycles, so depending on when it scrolls into view it can appear to be frozen for a while until it starts again. Should shorten the delay a tad?

> **for /ie**
> i know some copy changes are still forthcoming, but just noting format for “peril” text (first p / security and privacy section).

Done.

cgonz0:
> It seems to be working for me, but then I guess I don't know what it's supposed to look like. I do note there's a fairly long pause between animation cycles, so depending on when it scrolls into view it can appear to be frozen for a while until it starts again. Should we shorten the delay a tad?
> 
pause is fine so as not to distract from reading text, but for some reason there's an element that disappears in this view and not when i view it outside of the page

craigcook:
> pause is fine so as not to distract from reading text, but for some reason there's an element that disappears in this view and not when i view it outside of the page

Oh I see now, the eyelashes vanish... I wonder if there's some other CSS on the page messing with them. I can look into it (no pun) 👁

cgonz0:
> Oh I see now, the eyelashes vanish... I wonder if there's some other CSS on the page messing with them. I can look into it (no pun) 👁

yea, something going on - would appreciate that, thanks! 👁 👁

sgarrity:
I did a little digging into the eyelashes SVG animation. I don't have it solved, but I will shared what I've learned so far:

First, there's some rogue vendor-prefixed CSS that I think can be removed:

```
#eyelashes {
  -webkit-transform-origin: -25px 150px;
  -moz-transform-origin: 150px 150px;
}
```

I think this can be safely replaced with:
```
#eyelashes {
  transform-origin: 150px 150px;
}
```

Next, it appears that the issue of the animation not working when the SVG is included in the page via the `` tag seems to be related to the `perspective` property.

If I remove all of the `perspective(400px)` from the eyelashes animation, it seems to work. For example, replacing:
`transform:perspective(400px) rotate3d(1,0,0,-5deg)`
with
`transform:rotate3d(1,0,0,-5deg)`

sgarrity:
After doing a bit more experimenting, it seems the original animation (with the `perspective` attributes) works if the SVG is included inline in the HTML file, rather than via the `` tag.

craigcook:
@sgarrity Excellent sleuthing! I hadn't managed to trace it to perspective transform yet, but I did find that embedding the SVG inline worked when the img tag didn't so I wonder if this is down to some quirk of the rendering engine. But I think removing the perspective can fix it.

cgonz0:
> @sgarrity Excellent sleuthing! I hadn't managed to trace it to perspective transform yet, but I did find that embedding the SVG inline worked when the img tag didn't so I wonder if this is down to some quirk of the rendering engine. But I think removing the perspective can fix it.

although subtle, the perspective adds to the animation, so i'd prefer find another fix - if possible.

craigcook:
Temporarily blocked by getting FINAL final approved copy for IE, Edge, and Safari pages.

craigcook:
With final approved copy in place this is ready for a last round of code review. I had reviewed it previously but made enough updates myself along the way that I feel like someone else should give it a quick once-over to make sure I haven't missed something and/or broken something anew.

bedrock/firefox/templates/firefox/compare/chrome.html
```diff
@@ -282,38 +287,55 @@ <h2>{{ _('<strong>Portability</strong>') }}</h2>
...
       </p>
 
+      <p class="compare-version-disclaimer">
+        {% trans %}
+        The comparisons made here were done so across browser release versions as follows:<br> Firefox (70) | Chrome (74)
+        {% endtrans %}
+      </p>
+
+      <aside class="bottom-download-cta">
+        {{ download_firefox(dom_id='download-bottom', alt_copy=_('Download Firefox'), download_location='bottom cta') }}
```
alexgibson: I think our convention here is `secondary cta`?
```suggestion
        {{ download_firefox(dom_id='download-bottom', alt_copy=_('Download Firefox'), download_location='secondary cta') }}
```

bedrock/firefox/templates/firefox/compare/chrome.html
```diff
@@ -201,42 +203,45 @@ <h2>{{ _('<strong>Utility</strong>') }}</h2>
...
       </p>
 
-      <p class="compare-content">
+      <p>
         {% trans %}
         As for customization, our fans will tell you one of the things they love most about our browser is its ability to allow you to move and arrange a majority of the UI elements to best suit your needs. Chrome allows you to hide certain UI elements but there’s not much allowance, if any, for moving things around based on your preferences. However, it should be noted that both Chrome and Firefox make it pretty easy to change your browser’s appearance and theme.
         {% endtrans %}
       </p>
 
+      {% include '/firefox/compare/includes/download-menu-list.html' %}
```
alexgibson: Only the first occurence of this menu in a page seems to work? The others don't seem to open when clicked.

bedrock/firefox/templates/firefox/compare/chrome.html
alexgibson: This is the same for other pages.

bedrock/firefox/templates/firefox/compare/chrome.html
craigcook: I see the problem. I didn't realize the details JS uses an ID to know which menu to open and the include has duplicate IDs (which is a problem anyway, but now it's also breaking the JS). So when you click the second menu it's actually opening the _previous_ one further up the page.

I started to make the ID a variable to pass to the include, but realized that because Stephanie is brilliant and already thought of this, the JS will automatically assign a unique ID itself if ID. So just removing the ID from the include also fixes it.

craigcook:
Updated, now with subnav!

PR #8391, opened 2020-01-06
Webpack work

Errolyn:
## Description
This will Switch our build to use Webpack instead of Gulp


bedrock/base/templates/includes/protocol/navigation/menu-firefox/join.html
```diff
@@ -14,7 +14,7 @@
           <li>
             <section class="mzp-c-menu-item mzp-has-icon">
               <a class="mzp-c-menu-item-link" href="{{ url('firefox.accounts') }}" data-link-name="Join Firefox" data-link-type="nav" data-link-position="topnav" data-link-group="join">
-                <img class="mzp-c-menu-item-icon" src="{{ static('protocol/img/logos/firefox/logo.svg') }}" class="mzp-c-menu-item-icon" alt="" width="24" height="24">
+                <img class="mzp-c-menu-item-icon" src="{{ static('node_modules/@mozilla-protocol/core/protocol/img/logos/firefox/logo.svg') }}" class="mzp-c-menu-item-icon" alt="" width="24" height="24">
```
alexgibson: @Errolyn I haven't pulled this PR locally, but one thing to consider here is I don't believe we ship the content of `node_modules` in the production image for bedrock. So even though this may work locally, once deployed it may have issues still? Static assets like this should probably be kept in `/media/`, since that root directory is served via the CDN and cached more heavily if the filename is hashed.

It may be worth considering keeping the current behavior of copying the protocol assets over to /media. You may be able to achieve this using something like the [copy-webpack-plugin](https://www.npmjs.com/package/copy-webpack-plugin) perhaps?

media/css/exp/firefox/accounts-2019.scss
```diff
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-@import '../../../protocol/css/includes/lib';
+@import '../../~@mozilla-protocol/core/protocol/css/includes/lib';
```
alexgibson: Curious why this needs to be a relative path still?

If we're delving into `node_modules` for the source, could things like this just be `@import '~@mozilla-protocol/core/protocol/css/includes/lib';`?

media/css/exp/firefox/accounts-2019.scss
```diff
@@ -48,12 +48,12 @@
     }
 
     @media #{$mq-md} {
-        background: transparent url('/media/img/firefox/accounts/trailhead/accounts-hero-logged-out.svg') right -720px top no-repeat;
+        background: transparent url('~media/img/firefox/accounts/trailhead/accounts-hero-logged-out.svg') right -720px top no-repeat;
```
alexgibson: Can you help explain why this is needed everywhere?

webpack.config.js
```diff
@@ -0,0 +1,915 @@
...
+  ]),
+  "site": resolveBundles([
+    "js/base/core-datalayer-page-id.js",
+    "js/base/dnt-helper.js",
+    "js/base/mozilla-cookie-helper.js",
+    "js/base/site.js"
+  ]),
+  "stub-attribution": resolveBundles([
+    "js/base/stub-attribution-init.js",
+    "js/base/stub-attribution.js"
```
alexgibson: The JS errors are occuring because of the order the files in each bundle. If we we're using JS modules for importing dependencies this wouldn't be an issue, but bedrock currently handles bundling via concatenation. To fix the errors, I think you're gonna have to avoid the temptation to reorder things?

Here `stub-attribution-init.js` needs to come after `stub-attribution.js`. This is the same for the rest of the JS bundles.

babel.config.js
```diff
@@ -0,0 +1,19 @@
+const presets = [
+  [
+    "@babel/env",
+    {
+      targets: {
```
alexgibson: Is it worth adding babel at a later stage to this work? I understand adding it to the config is easy, but we have to consider how & when we'll apply transpiling (see considerations in https://github.com/mozilla/bedrock/issues/7542).

bedrock/exp/templates/exp/firefox/index.de.html
```diff
@@ -51,25 +51,25 @@ <h2 class="c-showcase-title">Das sind die Produkte der Firefox-Familie</h2>
         <ul class="c-showcase-list">
           <li>
             <a class="mzp-c-cta-link" href="{{ url('firefox.new') }}" data-cta-type="link" data-cta-text="Browsers">
-              <img alt="" src="{{ static('protocol/img/logos/firefox/browser/logo-md.png') }}" height="40" width="40"><br>
+              <img alt="" src="{{ static('node_modules/@mozilla-protocol/core/protocol/img/logos/firefox/browser/logo-md.png') }}" height="40" width="40"><br>
```
alexgibson: If we copied over the Protocol images to `/protocol/` like we do currently, would this mean we don't have to touch so many files? Or is this just generally tidier to do long term?

gulpfile.js
```diff
@@ -25,6 +25,8 @@ const argv = require('yargs').argv;
 const browserSync = require('browser-sync').create();
 const merge = require('merge-stream');
 const staticBundles = require('./media/static-bundles.json');
+const babel = require('gulp-babel');
```
alexgibson: Can this file and its npm dependencies be removed?

alexgibson:
If I’m correct in assuming that we’re still going to be using `collectstatic` in Python to manage hashing asset urls, like we do currently, then unless I’m mistaken, all webpack should have to do is ensure the image files are all copied to the right directory.

webpack.config.js
```diff
@@ -0,0 +1,915 @@
...
+  return fileList.map((f) => {
+    if (f.match(/^protocol\//)) {
+      return `@mozilla-protocol/core/${f}`;
+    }
+    return path.resolve(__dirname, "media", f);
+  });
+}
+
+const entry = {
+  "about": resolveBundles([
```
alexgibson: I wonder if instead of creating a new list that needs to be updated manually, we could import `static-bundles.json` and then traverse that here? Either that, or write a script convert one format to the other? Either way, separating this from the webpack config itself may be worthwhile.

Errolyn:
I'm going to close this for now, we can always look back at it to get info on what I did here.

PR #8400, opened 2020-01-07
Increase the width of the level span by one pixel

tomrittervg:
On my computer at least, "MFA 2020-02" is width enough that it overflows to two lines. Increasing the width by one pixel puts it back on one line.


media/css/security/components/level.scss
```diff
@@ -25,7 +25,7 @@ $image-path: '/media/protocol/img';
     font-weight: bold;
     line-height: 1;
     padding: 4px 2px 4px 4px;
```
amychurchwell:
```suggestion
    padding: 4px;
```
I think this would also be a good addition. 🙂

PR #8404, opened 2020-01-08
[no bug] remove social-widgets.less and send-to-device.less

amychurchwell:
## Description
Removes `social-widgets.less` & `send-to-device.less`


amychurchwell:
`social_widgets` is a legacy helper from when there was a twitter feed on the homepage. It is also currently updating the DB from [accounts in settings](https://github.com/mozilla/bedrock/blob/ee1ef78bb247b4d057ab5c7296b4f97b20bb7c8c/bedrock/settings/base.py#L630).

from @pmac :
> it looks like if we remove the credentials it'll stop updating, or I can just remove twitter from the update commands

Question - @hoosteeno do we have plans to support twitter feeds in the future? Or should we remove this code? We could always retrieve from git history if we need to.

amychurchwell:
cc: @alexgibson do you have an opinion about keeping the twitter feed infra?

alexgibson:
@amychurchwell this type of feed hasn't been used on bedrock in a number of years, so I'm happy to remove it. Things can always be resurrected from git history if needed. I don't think it needs to stay tied to the old .less file either way.

Is it ok to remove the backend code here as-is, or do we need to do anything with the DB first? /cc @pmac

hoosteeno:
> Question - @hoosteeno do we have plans to support twitter feeds in the future? Or should we remove this code? We could always retrieve from git history if we need to.

I have no reason to think we'll embed social posts on our website soon. Agree with removing.

pmac:
> Is it ok to remove the backend code here as-is, or do we need to do anything with the DB first?

I think we should probably add a migration to remove the DB table, and then follow up with a PR to remove the whole backend. I'm happy to handle the backend removal bits of this if it would help.

alexgibson:
> I think we should probably add a migration to remove the DB table, and then follow up with a PR to remove the whole backend. I'm happy to handle the backend removal bits of this if it would help.

Ok, let's remove the CSS file here in this PR, and follow up with the backend removal. Does thyat sound ok to you @amychurchwell?

amychurchwell:
@alexgibson that sounds good to me. Is what I have in this PR too much removal or okay as-is?

alexgibson:
Just removing the .less file is good for now I think 👍

PR #8462, opened 2020-01-21
Redesigned Firefox What's New page for India (Fixes #8300)

sgarrity:
## Description

This PR implements the updated content and design for the Firefox What's New page for India as defined in issue #8300.

## Issue / Bugzilla link

Addresses issue #8300

## Testing

This page is targeted (and designed for) desktop-sized viewports, though it has been basically adapted to be responsive.

PR #8464, opened 2020-01-21
Update design/content for What's New page for Indonesia (Fixes #8302)

### LINKED ISSUE #8302: (Code review) Update design in WNP template for Indonesia
## Description

Update the content in the Indonesian WNP for 72, featuring Firefox Lite

New content is [here](https://drive.google.com/drive/folders/1zJbfgledNRHT5pGCfAP9TvWfVouWUeDE?usp=sharing) - use the ' indonesia final' .psd file 

---

- [ ] Is live on 12/27

#### Tasks
- [ ] Update the content

sgarrity:
## Description

This PR implements the updated content and design for the Firefox What's New page for Indonesia as defined in issue #8302.

## Issue / Bugzilla link

Addresses issue #8302

## Testing

This page is targeted (and designed for) desktop-sized viewports, though it has been basically adapted to be responsive.


sgarrity:
For reference, here's the desktop design that was provided. I've taken some discretion in adapting the design to respond to smaller viewports:

image

media/css/firefox/whatsnew/whatsnew-lite.scss
```diff
@@ -63,158 +63,161 @@
...
+    padding-left: $spacing-lg;
+    padding-right: $spacing-lg;
+    background-repeat: no-repeat;
+    background-position: top left;
+    background-size: 42%;
     @media #{$mq-md} {
-        min-height: 180px;
+        display: flex;
+        justify-content: flex-end;
+        background-image: url('/media/img/firefox/whatsnew/lite/photo.webp');
```
alexgibson: webp! 🔥

alexgibson: @pmac I think this is the first .webp image format we've added to bedrock. Just checking, is there anything we need to do to make sure this is served correctly via the CDN? I assume this should be fine, but thought I should double-check.

bedrock/firefox/templates/firefox/whatsnew/index-lite.id.html
```diff
@@ -20,49 +20,35 @@
...
             <img src="{{ static('img/logos/firefox/logo-lite.png') }}"  alt="" height="146" width="155">
-            <h1><strong>Firefox</strong> Lite Baru</h1>
+            {# L10n: the <strong> below is for visual formatting only - bolding the word 'Firefox' #}
+            <h1>{{ _('<strong>Firefox</strong> Lite') }}</h1>
+          </div>
+          {# Rough translation: A lightning fast browsing experience with a mini APK size for Android. #}
+          <p class="lite-text">Pengalaman menjelajah cepat kilat dengan ukuran APK mini untuk Android.</p>
+          <div class="lite-qrcode-container">
+            <img src="{{ static('img/firefox/whatsnew/lite/qrcode-in.png') }}" alt="QR code">
+            <p>{{ _('Scan the QR code to get started.') }}</p>
```
alexgibson: Should this string be in indonesian?

bedrock/firefox/templates/firefox/whatsnew/index-lite.id.html
```diff
@@ -20,49 +20,35 @@
...
-        <div class="lite-container">
+    </header>
+    <div class="lite-body">
+      <div class="lite-content mzp-l-content">
+        <div class="lite-copy">
           <div class="lite-heading">
             <img src="{{ static('img/logos/firefox/logo-lite.png') }}"  alt="" height="146" width="155">
-            <h1><strong>Firefox</strong> Lite Baru</h1>
+            {# L10n: the <strong> below is for visual formatting only - bolding the word 'Firefox' #}
+            <h1>{{ _('<strong>Firefox</strong> Lite') }}</h1>
```
alexgibson: This should probably be an `h2` since "Your Firefox is up to date." is already an `h1` above.

media/css/firefox/whatsnew/whatsnew-lite.scss
```diff
@@ -63,158 +63,161 @@
...
+    margin-bottom: $spacing-xl;
 
     h1 {
-        @include font-firefox;
-
-        @include text-display-md;
-
         font-weight: normal;
-        text-align: start;
+        font-size: 2rem;
```
alexgibson: Please try and avoid setting arbitrary font sizes, and choose from one of the predefined options in Protocol: https://protocol.mozilla.org/fundamentals/typography.html

media/css/firefox/whatsnew/whatsnew-lite.scss
pmac: Hmmm... Only issue I could foresee would be that it would be served with the wrong content-type header. I'll check to see if our stuff should understand webp.

sgarrity: A quick note on the rationale for using WebP here: 
* This page is aimed at recent versions of  Firefox only, so support isn't an issue
* The design called for an unusual combination of a circularly-masked photographic element (usually best compressed via JPEG) and simpler solid color elements.

Originally I was going to either embed a JPEG (for the photographic element) in an SVG (for the masking and simpler shapes), or put a JPEG as a background element on a div with border-radius to handle the masking.

WebP was much simpler than both options. I used the cleverly-named [WebPShop](https://developers.google.com/speed/webp/docs/webpshop) plugin for Photoshop to export the WebP image

bedrock/firefox/templates/firefox/whatsnew/index-lite.id.html
sgarrity: Yes! Fixed.

media/css/firefox/whatsnew/whatsnew-lite.scss
pmac: It does seem that our mime-type guessing probably won't get the right answer for the webp images. It will probably do `binary/octet-stream`. I think the image will still work on the page, but not if viewed directly. Thoughts?

sgarrity:
@alexgibson I've fixed that untranslated string (good catch). I've also made the same h1->h2 and Protocol font size changes referenced over in the India page PR: https://github.com/mozilla/bedrock/pull/8462#issuecomment-577230934

media/css/firefox/whatsnew/whatsnew-lite.scss
alexgibson: If we push this to demo, we should be able to test that it works ok, right?

pmac: I think demos serve the static media themselves, where dev, stage, and prod sync the media to an s3 bucket and serve from there. I don't think we'll be sure until we get it on dev unfortunately :/

alexgibson: Ok, let's see how this goes once it merges. Is it worth filing a separate issue so that we can fix the mime-type?

pmac: I think so. We should really send the proper mime type. Let's verify that it is in fact the wrong type, then we can file an issue to fix it.

pmac: I'm going to go ahead and file this issue. I'm pretty confident it will be wrong, but I've found a pretty easy way to fix it, so I'll document that in the issue. Good catch!

sgarrity:
This branch is now up on demo2 for preview: https://www-demo2.allizom.org/id/firefox/whatsnew/all/

bedrock/firefox/templates/firefox/whatsnew/index-lite.id.html
```diff
@@ -20,49 +20,35 @@
...
             <img src="{{ static('img/logos/firefox/logo-lite.png') }}"  alt="" height="146" width="155">
-            <h1><strong>Firefox</strong> Lite Baru</h1>
+            {# L10n: the <strong> below is for visual formatting only - bolding the word 'Firefox' #}
+            <h2>{{ _('<strong>Firefox</strong> Lite') }}</h2>
+          </div>
+          {# Rough translation: A lightning fast browsing experience with a mini APK size for Android. #}
+          <p class="lite-text">Pengalaman menjelajah cepat kilat dengan ukuran APK mini untuk Android.</p>
+          <div class="lite-qrcode-container">
+            <img src="{{ static('img/firefox/whatsnew/lite/qrcode-in.png') }}" alt="{{ _('kode QR') }}">
+            <p>{{ _('Pindai kode QR untuk memulai.') }}</p>
```
alexgibson: Nit: no need to wrap this string for L10n as it's hard-coded.

tests/functional/firefox/whatsnew/test_whatsnew.py
```diff
@@ -29,10 +29,7 @@ def test_send_to_device_fails_when_missing_required_fields(base_url, selenium):
...
-def test_firefox_rocket_send_yourself(base_url, selenium):
+def test_firefox_lite_qr_code(base_url, selenium):
     page = FirefoxWhatsNewPage(selenium, base_url, locale='id').open()
-    assert page.send_to_device.is_displayed
-    send_to_device = page.send_to_device
-    send_to_device.type_email('success@example.com')
-    send_to_device.click_send()
-    assert send_to_device.send_successful
+    assert not page.send_to_device.is_displayed
+    assert page.is_qr_code_displayed
```
alexgibson: This test is failing because it's for a QR code defined on a different page to the one being tested here.

I'd suggest using a new property here, like `assert page.is_firefox_lite_qr_code_displayed`.

You can then define that property in [whatsnew.py](https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/whatsnew/whatsnew.py) using:

```python
_firefox_lite_qr_code_locator = (By.CSS_SELECTOR, '.lite-qrcode-container img')

@property
def is_firefox_lite_qr_code_displayed(self):
    return self.is_element_displayed(*self._firefox_lite_qr_code_locator)

alexgibson:
@sgarrity In an email @ejregithub relayed some feedback from Stella, that the QRCodes on these pages are "below the fold". They look ok to me, as per the approved designs, but I guess they could be an issue on smaller screens. Perhaps there's a solution here that could be achieved using a vertical media query?

/cc @cgonz0

sgarrity:
> @sgarrity In an email @ejregithub relayed some feedback from Stella, that the QRCodes on these pages are "below the fold". They look ok to me, as per the approved designs, but I guess they could be an issue on smaller screens. Perhaps there's a solution here that could be achieved using a vertical media query?

While the vertical position varies across the whole breadth of viewport widths, I'm inclined to see if we can just compress the vertical arrangement of that main content column at all widths. I'll try this out.

pmac:
fold

sgarrity:
I've made some vertical layout adjustments to make it more likely that the QR code will be entirely visible on a broader set of viewport sizes.

Here's a quick preview of the before/after:

responsive-preview

cgonz0:
> @sgarrity In an email @ejregithub relayed some feedback from Stella, that the QRCodes on these pages are "below the fold". They look ok to me, as per the approved designs, but I guess they could be an issue on smaller screens. Perhaps there's a solution here that could be achieved using a vertical media query?
> 
> /cc @cgonz0

i'm good with @sgarrity's solution below

Issue #8471, opened 2020-01-22
Remove ContributorActivity model and contributor-data/ URL


pmac:
Along with the `update_tableau_data` management command. These things are no longer used. They updated data for a JSON endpoint like https://www.mozilla.org/contributor-data/github.json, but all of those URLs return empty JSON objects now.

@alexgibson do you agree, or is there some plan to revive this?

alexgibson:
Huh, I had no idea this was a thing. Was this used for the old mozilla logo prototype that was never used? If that's the case, then yeah let's remove it 👍

pmac:
It was that plus some numbers on an old home page and the main old contributor page IIRC. Sounds good. Adding it to the PR of removal :) 💥

alexgibson:
We still have contributor number here, but I'm not sure where that comes from: https://www.mozilla.org/en-US/contribute/

Issue #8417, opened 2020-01-10
Remove twitter feed backend machinery


pmac:
Following up on #8404 we should remove the Twitter feed updating and storage code. This will need to be done in two steps:

- [ ] Remove most of the code and add a DB migration to remove the DB table
- [ ] Remove the rest of everything including all DB migrations

Issue #8402, opened 2020-01-07
Remove events data from the DB


alexgibson:
## Description

Once https://github.com/mozilla/bedrock/issues/6461 is complete, we can then remove event data from the DB and clean up the backend code: https://github.com/mozilla/bedrock/tree/master/bedrock/events

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Event data is removed from the DB

#### Tasks
- [ ] Add table removal migration.
- [ ] Once that's in prod, remove the remaining backend code.

alexgibson:
Depends on https://github.com/mozilla/bedrock/issues/6461

alexgibson:
@pmac fyi - the events page has been removed, so I think we should be fine to remove event data from the DB.

PR #8473, opened 2020-01-22
Remove models and code for unused data

pmac:
For Twitter, Events, and Contributor info

Re #8471, #8417, #8402


pmac:
After this I'll do a quick follow up to completely remove the events app. This needs to get all the way to prod first though, so I'll just file a new issue.

pmac:
huh... thanks. I'll check it out.

PR #8476, opened 2020-01-22
Use dockerized aws-cli for s3 sync of static files

### LINKED ISSUE #8470: Upload .webp files with the proper mime type
The AWS cli that we use to upload our static media to the S3 buckets uses the Python standard library [`mimetypes` package](https://docs.python.org/3/library/mimetypes.html) to "guess" the types for each file. In #8464 we've added our first .webp file to bedrock, and it does not properly guess the type. This is because the type isn't quite in the standards yet, and the Python lib only adds the standards. However, the library does read [some other files](https://github.com/python/cpython/blob/570724667d37fcb160cdef2b34190dc36e28155c/Lib/mimetypes.py#L42-L52) that we can create or modify. Currently the `aws s3 sync` command is run directly on the GitLab runner box, so we have a couple of options:

1. Add the `/usr/local/etc/mime.types` file to the runner machines
2. Move the s3 sync operation into docker and add the file there

I'm leaning toward option 2 since it means that our runners don't need to be any more special, and we keep all the code necessary to make the site work in the repo. The format for the file would be:

```
image/webp       webp
```

Related to #6766

pmac:
This will use the [docker image from this repo](https://github.com/mozmeao/aws-cli-docker) which includes a mime type definition for webp (`image/webp`). It'll be challenging to test this before merging to master, so we should discuss before doing so.

Fix #8470

### DIFF
--- a/bin/aws-docker
+++ b/bin/aws-docker
@@ -1,0 +1,6 @@
+#!/usr/bin/env bash
+set -eo pipefail
+docker run --rm \
+           -v "${HOME}/.aws:/home/webdev/.aws:ro" \
+           -v "$(pwd):/app" \
+           mozmeao/aws-cli:latest "$@"
--- a/bin/upload-staticfiles.sh
+++ b/bin/upload-staticfiles.sh
@@ -32,7 +32,7 @@ bin/move_hashed_staticfiles.py "${TMP_DIR}" "${TMP_DIR_HASHED}"
 
 for BUCKET in "${BUCKETS[@]}"; do
     # hashed filenames
-    aws s3 sync \
+    bin/aws-docker s3 sync \
         --acl public-read \
         --cache-control "max-age=315360000, public, immutable" \
         --profile bedrock-media \
@@ -40,7 +40,7 @@ for BUCKET in "${BUCKETS[@]}"; do
     # non-hashed filenames
     # may not need to include non-hashed files
     # TODO look into this if this is slow or makes the bucket too large
-    aws s3 sync \
+    bin/aws-docker s3 sync \
         --acl public-read \
         --cache-control "max-age=21600, public" \
         --profile bedrock-media \


pmac:
@duallain @glogiotatidis I'd love your thoughts on this. If you look at the linked repo I'm basically creating a container for the aws-cli that includes a custom mime.types file so that we can get `aws s3 sync` to add the proper mime type to `.webp` image files. It will allow us to do more fun stuff in future as well should we need to obviously.

glogiotatidis:
The solution itself is well implemented. Myself I'm undecided on whether we should create custom images (like in this case) and bundle tools in the repo (what I did in lumbergh) or make those changes in the workers.

pmac:
Right. The other option is to take this custom `mime.types` and install it on all the runners via ansible. It's another fine choice. My thinking was that in this way we can keep customization on the runners lower, and keep these necessary tweaks in public repos. But really I'm happy either way.

@metadave do you have an opinion on this?

metadave:
I think this is good. Some time in the near future, I'd like to get away from using creds, but that doesn't have to be part of this work.

Do we need to do any extra work to keep the image up to date?

pmac:
> Do we need to do any extra work to keep the image up to date?

Yes. I didn't (or haven't yet) hook the repo up to docker hub for auto builds. I did setup dependabot so we will be notified of dependency upgrades. In that regard the aws cli will likely be more up-to-date than one installed on the runner. I looked for an official aws cli image but didn't see one.

glogiotatidis:
As a  side note: If this syncing is taking "too long", maybe we can consider `rclone` in the future, as we did with careers. Rclone supports setting mime.types as does aws-cli https://rclone.org/overview/#features

pmac:
I'm not sure exactly how long these are taking. Perhaps I should add some timing output in the build logs. Right now the build jobs are taking between 6 and 15 min or so... so not THAT bad IMO. Could be better probably. Maybe rclone already supports more mime types? Does it do good type guessing like aws-cli?

pmac:
Looks like rclone guesses mime types using the Go `mime` library, which uses similar files to the python equivalent https://golang.org/pkg/mime/#TypeByExtension

pmac:
I've gone ahead and updated the ansible for the gitlab runners to add `/etc/apache2/mime.types`, so this should no longer be necessary.

PR #8483, opened 2020-02-04
Add coarse UA detection to stub attribution (Fixes #8405)

### LINKED ISSUE #8405: Set up an experiment for UA attribution with coarse user agent detection
## Description

Probably using #8363 to complete [Bug 1406005 - Run onboarding experiment passing UA string variation to attribution service](https://bugzilla.mozilla.org/show_bug.cgi?id=1406005)

The high level intent is to correctly attribute the top 5 desktop browsers and pass that data to Firefox via stubattribution. Other browsers (or modified UAs) that have strings that look like one of the top default user agent strings will just be treated as false positives. Also noting that we don't need to capture the browser version.

Here's some example code from https://bugzilla.mozilla.org/show_bug.cgi?id=1406005#c47:
```js
function deriveVariation(ua) {
  if (ua.includes("Trident")) return "ie";
  if (ua.includes("Edg"))     return "edge";
  if (ua.includes("Version")) return "safari";
  if (ua.includes("Firefox")) return "firefox";
  if (ua.includes("Chrome"))  return "chrome";
  return "other";
}
```

alexgibson:
## Description
- Adds a new `ua` parameter to stub attribution.

## Issue / Bugzilla link
#8405

## Testing
Demo: https://www-demo5.allizom.org/en-US/

### DIFF
--- a/media/js/base/stub-attribution.js
+++ b/media/js/base/stub-attribution.js
@@ -156,6 +156,38 @@ if (typeof window.Mozilla === 'undefined') {
     };
 
     /**
+     * Returns a browser name based on coarse UA string detection for only major browsers.
+     * Other browsers (or modified UAs) that have strings that look like one of the top default user agent strings are treated as false positives.
+     * @param {String} ua - Optional user agent string to facilitate testing.
+     * @returns {String} - Browser name.
+     */
+    StubAttribution.getUserAgent = function(ua) {
+        ua = typeof ua !== 'undefined' ? ua : navigator.userAgent;
+
+        if (/MSIE|Trident/i.test(ua)) {
+            return 'ie';
+        }
+
+        if (/Edg|Edge/i.test(ua)) {
+            return 'edge';
+        }
+
+        if (/Firefox/.test(ua)) {
+            return 'firefox';
+        }
+
+        if (/Opera/.test(ua)) {
+            return 'opera';
+        }
+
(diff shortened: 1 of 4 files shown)


media/js/base/stub-attribution.js
```diff
@@ -155,6 +155,38 @@ if (typeof window.Mozilla === 'undefined') {
...
+        if (/Edg|Edge/i.test(ua)) {
+            return 'edge';
+        }
+
+        if (/Firefox/.test(ua)) {
+            return 'firefox';
+        }
+
+        if (/Opera/.test(ua)) {
+            return 'opera';
```
Mardak: No need to fix this, as we didn't list Opera in the bug and this is acceptable as a false positive. FYI, here's the user agent from Opera on mac:

Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_1) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3945.130 Safari/537.36 OPR/66.0.3515.72

and windows:
Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.87 Safari/537.36 OPR/66.0.3515.44

alexgibson: If Opera isn't a concern I'll go ahead and remove this altogether 👍

Issue #8407, opened 2020-01-09
Code: new 404 and 500 pages


ejregithub:
## Description

The card is for the tasks required to code the newly design 404 and 500 error pages on www.mozilla.org, using our current coding standards. 

The new designs for this page are here: https://github.com/mozilla/bedrock/issues/8329#issuecomment-572775691

### Success Criteria 
- [ ] The page is built with Protocol components
- [ ] Strings are extracted, made available in Pontoon to be localized in Fluent by the Community

### Reviews:
- [ ] Brand (copy and design)
- [ ] Team QA
- [ ] Team Lead

------

## Original Report

I ended up on https://www.mozilla.org/en-US/firefox/xr/ ([from clicking a link like this](https://bugzilla.mozilla.org/show_bug.cgi?id=1604853))

<img width="1248" alt="Screen Shot 2019-12-18 at 12 48 33 PM" src="https://user-images.githubusercontent.com/26739/71110150-bf80ff00-2194-11ea-8c32-bd0735c717c5.png">

My impression was that this was some sort of marketing campaign. But it was cryptic. What's a "Whoops!"? Also, the attempt to be funny with "Did you make a left at that last URL instead of a right?" is funny but it's not working. It could work if I've immediately understood that it's a page-not-found page. 

Also the bullet point list is unfocussed in that I have no idea what those bullet points are trying to communicate (I know it's about getting you away from the not found page to the right page, but that's not clear and takes time to understand)

### Steps to reproduce

Go to https://www.mozilla.org/en-US/firefox/xr/ and imagine you don't know what a www.mozilla.org's page-not-found pages look like. 

### Expected result

A title like "Page not found". Or a sub-title like "The URL could not be found". 
Perhaps some graphic that gets the impression that something went wrong. 

### Actual result

See description.

Errolyn:
@ejregithub I'm not seeing a source for the copy? I'm just using what I see in the design but to go any further I need access to the copy doc. Thanks :D 

@cgonz0 what famicon should we use for the 404 & 500 pages? And to double check there should not be a footer on this page?

cgonz0:
@Errolyn 
correct, no nav or footer. let me get back to you on the favicon:)
the (m)ozilla and desktop icons are used in the nav - although not exactly sure where they live in the directory.
i don't think the firefox logomark icon exists on any webpage just yet, so will send that to you.

> @cgonz0 what famicon should we use for the 404 & 500 pages? And to double check there should not be a footer on this page?
[...]

PR #8518, opened 2020-02-12
Error page updates [Fixes: #8407]

Errolyn:
## Description

issue: #8407

Update 404/500 page
Approved by design

## Testing
- http://localhost:8000/en-US/500/
- http://localhost:8000/en-US/404/


alexgibson:
@Errolyn looks like this branch might need a rebase as there are commits from other issues in here?

bedrock/base/templates/404.html
```diff
@@ -2,47 +2,63 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
-{% set_lang_files "mozorg/404" %}
-
-{% extends "base-pebbles.html" %}
+{% set_lang_files "mozorg/404-update-2020" %}
```
alexgibson: Is the plan here to lose translations for the existing pages? Usually when making updates to existing pages we'll use an [L10n conditional tag](https://bedrock.readthedocs.io/en/latest/l10n.html#staging-copy-changes), so locales that don't yet have the new content translated can fall back to the old strings. Technically, we don't _have_ to do that, but it means the error pages would be displayed in English until communities catch up again.

bedrock/base/templates/404.html
```diff
@@ -2,47 +2,63 @@
...
-    <header>
-      <h1><span>{{ _('Whoops!') }}</span></h1>
-      <p>{{ _('Did you make a left at that last URL instead of a right? No problem. Here are some tips to get you back on your way:') }}</p>
-    </header>
-    <div class="error-content">
-      <ul class="prose">
-        <li>{{ _('If you typed in the address, check your spelling. Could just be a typo.') }}</li>
+  <main role="main">
+    <section class="mzp-l-content message-section">
+      <h2>{{ _("Sorry, we can't find that page") }}</h2>
```
alexgibson: Nit: single quotes for wrapping L10n strings, and use curly apostrophes inside copy.
```suggestion
      <h2>{{ _('Sorry, we can’t find that page') }}</h2>
```

bedrock/base/templates/404.html
```diff
@@ -2,47 +2,63 @@
...
+      <p>{{_("We're all about a healthy internet but sometimes broken URLs happen.") }}</p>
```
alexgibson:
```suggestion
      <p>{{_('We’re all about a healthy internet but sometimes broken URLs happen.') }}</p>
```

bedrock/base/templates/404.html
```diff
@@ -2,47 +2,63 @@
...
+      <div>
+        <a class="back-button" href="javascript:history.back()">{{ _("Go Back") }}</a>
```
alexgibson: We should try and avoid using inline JS, and instead move this to an external script. Also, what happens if I'm on Twitter and hit a link which 404's? If it opens in a new tab, there's nothing to go back to?

bedrock/base/templates/404.html
```diff
@@ -2,47 +2,63 @@
...
+      <ul>
         <li>
-        {% trans bugzilla='https://bugzilla.mozilla.org/enter_bug.cgi?product=www.mozilla.org' %}
-          If you’ve found an issue with one of our websites, we’d appreciate it if you could report the problem in Bugzilla, our <a href="{{ bugzilla }}">bug tracker</a>. One of our developers will take a look at it as soon as possible.
-        {% endtrans %}
+          <section class="mzp-has-icon">
+            <img class="list-icon" alt="" src="{{ static('protocol/img/icons/mozilla.svg') }}" width="30" height="30">
+              <p class="list-content">
+                {% trans about=url('mozorg.about') %}
+                  <a class="no-highlight" href="{{ about }}">Learn</a> about Mozilla, the non-profit behind Firefox.
```
alexgibson: It might be worth trying to keep things like `class` names out of string translations, as it requires localizers to type everything correctly in order not to break styling. Instead you could try something like `.list-content a` in the CSS?

bedrock/base/templates/404.html
```diff
@@ -2,47 +2,63 @@
...
+        <a class="back-button" href="javascript:history.back()">{{ _("Go Back") }}</a>
+      </div>
+    </section>
+    <section class="mzp-l-content">
+      <ul>
         <li>
-        {% trans bugzilla='https://bugzilla.mozilla.org/enter_bug.cgi?product=www.mozilla.org' %}
-          If you’ve found an issue with one of our websites, we’d appreciate it if you could report the problem in Bugzilla, our <a href="{{ bugzilla }}">bug tracker</a>. One of our developers will take a look at it as soon as possible.
-        {% endtrans %}
+          <section class="mzp-has-icon">
```
alexgibson: This isn't a protocol component as such, so suggest dropping the `mzp-` prefix?

bedrock/base/templates/404.html
```diff
@@ -2,47 +2,63 @@
...
+              <a class="no-highlight" href={{ explore }}>Explore</a> the entire family for Firefox products designed to respect your privacy.
+            {% endtrans %}
+            </p>
+          </section>
         </li>
+        <li>
+          <section class="mzp-has-icon">
+            <img class="list-icon" alt="" src="{{ static('protocol/img/icons/desktop.svg') }}" width="30" height="30" color="black">
+            <p class="list-content">
+            {% trans  download='https://firefox.com/new'%}
```
alexgibson: Currently this link redirects to /new, so until firefox.com exists as a real project I'd suggest linking directly?
```suggestion
            {% trans download=url('firefox.new') %}
```

bedrock/base/templates/500.html
```diff
@@ -15,22 +15,13 @@
...
-      {% endtrans %}
-      </p>
-    </div>
-    <aside class="download-firefox">
-      {{ download_firefox() }}
-    </aside>
-  </div>
+<main role="main">
+  <section class="mzp-l-content 500-content">
+    <h2 class="500-header">{{ _("Something went wrong") }}</h2>
```
alexgibson: Nit: single quotes

bedrock/base/templates/500.html
```diff
@@ -15,22 +15,13 @@
...
-    </div>
-    <aside class="download-firefox">
-      {{ download_firefox() }}
-    </aside>
-  </div>
+<main role="main">
+  <section class="mzp-l-content 500-content">
+    <h2 class="500-header">{{ _("Something went wrong") }}</h2>  
+    <p class="500-description">{{ _('It’s probably just a server error and we’re working to fix it.') }}<br>
+    {% trans firefox="https://firefox.com", mozilla="https://mozilla.org" %}
```
alexgibson:
```suggestion
    {% trans firefox="https://firefox.com", mozilla=url('mozorg.home') %}
```

media/css/base/page-not-found.scss
```diff
@@ -2,79 +2,68 @@
...
+.no-highlight:link, .no-highlight:visited, .no-highlight:hover, .no-highlight:active {
+    color: black;
 }
 
 main {
-    min-height: 450px;
-
-    header {
-        margin-bottom: 40px;
+    font-size: 18px;
```
alexgibson: Nit: try and avoid setting aritrary font sizes in pixels, and stick to the predefined type mixins defined in protocol? https://protocol.mozilla.org/fundamentals/typography.html#styles

Same applies below

media/css/base/page-not-found.scss
```diff
@@ -2,79 +2,68 @@
...
+}
 
-        h1 {
-            @include font-size-huge;
-            color: #fff;
-            line-height: 1.3;
+h2 {
```
alexgibson: This is a pretty broad selector and will apply to anything on the page. It may be ok as is, but should we add any more content in the future it could have unintended consequences. Suggest using something more specific to the page, like `.500-header`?

media/css/base/page-not-found.scss
```diff
@@ -2,79 +2,68 @@
...
+    font-size: 40px;
+}
 
-            span {
-                @include box-decoration-break(clone);
-                background-color: #000;
-            }
-        }
+svg, img {
```
alexgibson: Same for these images, suggest making the selectors less broad.

media/css/base/page-not-found.scss
```diff
@@ -2,79 +2,68 @@
...
+    vertical-align: top;
+    padding-right: $spacing-md;
+}
 
-        p {
-            max-width: 35em;
-        }
-    }
+li {
```
alexgibson: Same here, this will apply to any list item on the page.

bedrock/base/templates/404.html
alexgibson: Should this also be an `h1`?

bedrock/base/templates/500.html
alexgibson: Should this also be an `h1`?

media/css/base/page-not-found.scss
```diff
@@ -2,79 +2,68 @@
...
+}
 
-            span {
-                @include box-decoration-break(clone);
-                background-color: #000;
-            }
-        }
+svg, img {
+    vertical-align: top;
+    padding-right: $spacing-md;
```
alexgibson: Might also want to use a `bidi()` mixin here so the padding appears on the left for RTL locales?

alexgibson:
There's a functional test here for the 404 page which may need updating:

- https://github.com/mozilla/bedrock/blob/master/tests/pages/not_found.py
- https://github.com/mozilla/bedrock/blob/master/tests/functional/test_not_found.py

media/css/base/page-not-found.scss
Errolyn: I originally had used the mix-ins but was asked by design not to and set the font sizes manually with specific break points for this page. (https://github.com/mozilla/bedrock/issues/8407#issuecomment-587989081) I reached out to Cristina and asked. The mixins we have defined don't match the sizing designed for the page.

bedrock/base/templates/404.html
Errolyn: It looks like we are wanting to use these pages for testing moving over to fluent. I'm going to get on a bit earlier to talk about this with you and make sure I understand what needs to be done.

media/css/base/page-not-found.scss
alexgibson: We spent a long time trying to standardise on type-scale across mozorg pages, which at one time lacked any real kind of consistency. I can only encourage you to encourage our designers to try and use what’s in the design system where possible. If we really can’t avoid setting arbitrary type sizes, then I’d suggest using the `font-size()` mixin, so modern browsers will get rem units instead of fixed pixel sizes.

bedrock/base/templates/404.html
alexgibson: Happy to chat about it, but don’t worry about the Fluent part. We can tackle that after this PR has merged 🙂

bedrock/base/templates/404.html
```diff
@@ -2,47 +2,66 @@
...
-    </header>
-    <div class="error-content">
-      <ul class="prose">
-        <li>{{ _('If you typed in the address, check your spelling. Could just be a typo.') }}</li>
+  <main role="main">
+    <section class="mzp-l-content message-section">
+      <h1 class="error_header">{{ _('Sorry, we can’t find that page') }}</h2>
+      <p>{{_('We’re all about a healthy internet but sometimes broken URLs happen.') }}</p>
+      <div class="hide-back" id="go-back">
+        <a class="back-button" onclick="goBack()">{{ _("Go Back") }}</a>
```
alexgibson: Generally we should try and avoid creating global functions to handle click events. Suggest moving this `onClick` handler to the script you created and using `addEventListener()` to handle the event?

The link should probably also have an `href='#'` value.

alexgibson:
```suggestion
        <a class="back-button" onclick="goBack()">{{ _('Go Back') }}</a>
```

media/css/base/page-not-found.scss
```diff
@@ -2,79 +2,70 @@
...
+    color: $color-link;
+    cursor: pointer;
+}
 
-        &:link,
-        &:visited {
-            color: $color-link;
-        }
+.hide-back {
+    display: none;
```
alexgibson: Instead of hiding this link conditionally if there's no history, might it be worth hiding it by default (since it relies on JS to work), and then only showing it if there's something to go back to?

media/js/base/error-pages.js
```diff
@@ -0,0 +1,17 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ 
+// eslint-disable-next-line
```
alexgibson: Please try to avoid creating global functions (see comment above)

bedrock/base/templates/base-error.html
```diff
@@ -0,0 +1,19 @@
...
+
+{% block page_title_prefix %}{% endblock %}
+{% block page_title_suffix %}{% endblock %}
+
+{% block body_class %}{% endblock %}
+
+{% block site_header %}
+  <header class="mzp-l-content header-spacing">
+    <div class="header-image">
+    {{ high_res_img('img/co-brand/firefox_moz_co-brand_sm.png') }}
```
alexgibson: Should this image has some alt text perhaps?

alexgibson:
This may need a rebase after https://github.com/mozilla/bedrock/pull/8609 merged.

Errolyn:
For some reason getting setting a conditional for a tag is not working for the 404 page. I'm going to try and look into it more tomorrow, but if @alexgibson you have any ideas I would love your input :D

alexgibson:
@Errolyn I think this may be an issue with the way Django handles the view for the 404 page by default. I managed to fix it by treating it the same way as we do the 500 page: https://github.com/alexgibson/bedrock/commit/ae25b67b51a37253eda1774da55923bc4a8b0b52

@pmac does this look ok - anything we're missing here?

pmac:
Oh wow. I guess I don't understand then how the current 404 page is translated... The change you linked to looks fine to me @alexgibson .

bedrock/base/templates/500.html
```diff
@@ -2,9 +2,9 @@
  # License, v. 2.0. If a copy of the MPL was not distributed with this
  # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
 
-{% set_lang_files "mozorg/500" %}
+{% set_lang_files "mozorg/404-update-2020" %}
```
alexgibson: Note we still need to keep existing translations around for the 500 page too, like we did for the 404 page.

bedrock/base/templates/404.html
```diff
@@ -3,46 +3,98 @@
...
+        </aside>
+      </div>
+    </main>
+    
+  {% endif %}
 {% endblock %}
+
+{% block js%}
+  {{ js_bundle('page_not_found') }}
+{% endblock %}
```
alexgibson: Nit: we should generally have a blank line at end of files. I can recommend [setting up editorconfig](https://bedrock.readthedocs.io/en/latest/coding.html#configuring-your-code-editor) to work with your code editor as it takes care of a lot of things like this automatically, as well as indentation and whitespace.

bedrock/base/views.py
```diff
@@ -171,6 +171,11 @@ def server_error_view(request, template_name='500.html'):
     return l10n_utils.render(request, template_name, status=500)
 
 
+def page_not_found_view(request, template_name='404.html'):
```
alexgibson: I think Paul commented on this in Slack, ybut you may need `exception` in here as an argument.
```suggestion
def page_not_found_view(request, exception, template_name='404.html'):
```

tests/functional/test_not_found.py
```diff
@@ -8,6 +8,13 @@
 
 
 @pytest.mark.nondestructive
-def test_is_download_button_displayed(base_url, selenium):
+def test_go_back_button_is_not_displayed_without_history(base_url, selenium):
```
alexgibson: Nice job on the tests! These are passing for me locally, but I would recommend you push this to `run-integration-tests` before we merge 👍

media/js/base/error-pages.js
```diff
@@ -0,0 +1,17 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+ 
+// eslint-disable-next-line 
+
+(function(){
+    'use strict';
+    // Hides back button if there is no previous page to go back to.
+    if (window.history.length > 1) {
+        var div = document.getElementById('go-back');
+        div.classList.remove('hide-back');}
```
alexgibson: Nit: closing bracket on line below `}`

bedrock/base/templates/base-error.html
```diff
@@ -0,0 +1,19 @@
...
+
+{% block page_title_prefix %}{% endblock %}
+{% block page_title_suffix %}{% endblock %}
+
+{% block body_class %}{% endblock %}
+
+{% block site_header %}
+  <header class="mzp-l-content header-spacing">
+    <div class="header-image">
+    {{ high_res_img('img/co-brand/firefox_moz_co-brand_sm.png', {'alt': 'Firefox | Mozilla'}) }}
```
alexgibson: Might this sound better to a screen reader?

```suggestion
    {{ high_res_img('img/co-brand/firefox_moz_co-brand_sm.png', {'alt': 'Firefox, Mozilla'}) }}
```

bedrock/base/views.py
```diff
@@ -171,6 +171,11 @@ def server_error_view(request, template_name='500.html'):
     return l10n_utils.render(request, template_name, status=500)
 
 
+def page_not_found_view(request, exception, template_name='404.html'):
```
alexgibson: Try:

```suggestion
def page_not_found_view(request, exception=None, template_name='404.html'):
```

Issue #7705, opened 2019-09-11
Create automation for checking and processing FTL files


pmac:
This automation will do a number of things:

- [x] pull changes in from the l10n team's fluent file repository
- [x] check those changed files for errors
- [x] copy the files without errors to [mozmeao/www-l10n](https://github.com/mozmeao/www-l10n/) repo
- [x] check the changed files that aren't yet active to decide if they should be activated
- [x] write the activitation state of each file to a JSON file of the same name (e.g. for a FTL file `de/mozorg/home.ftl` write a `metadata/mozorg/home.json` file to contain metadata)
- [ ] push the updates to the [mozmeao/www-l10n](https://github.com/mozmeao/www-l10n/) repo

This will make it as easy as possible for bedrock to know the things it needs to for page activation.

This automation will live in bedrock and should get a GitLab CI job that will be triggered by a webhook on commit in the l10n team's repo.

TODO:

- [ ] hook up the automation to do all of the above on commits to the l10n team's repo

hoosteeno:
This issue can be almost completely resolved; but closing it out will require the l10n team's repository to be built. Meanwhile we can build a temporary version of that somewhere.

alexgibson:
A script that pulls in updates and processes activation metadata was added in https://github.com/mozilla/bedrock/pull/8523

The remaining steps involve running this script via automation.

pmac:
Also need to improve said script from #8523 to include updating the configs in addition to the .ftl files.

PR #8557, opened 2020-02-20
Include the TOML configs when updating the Fluent repo

pmac:
Re #7705

### DIFF
--- a/lib/l10n_utils/management/commands/process_ftl.py
+++ b/lib/l10n_utils/management/commands/process_ftl.py
@@ -47,6 +47,7 @@ class Command(BaseCommand):
         self.update_l10n_team_files()
         no_errors = self.copy_ftl_files()
         self.set_activation()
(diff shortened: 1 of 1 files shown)


pmac:
@Pike for this command, you also have a better method of linting the .ftl files right?

Pike:
> @Pike for this command, you also have a better method of linting the .ftl files right?

Not ad-hoc. We have an algorithm inside compare-locales which does create a copy w/out errors, and it's a lot more powerful. Sadly, it requires some artifacts that are Firefox-specific right now, so that's not usable out of the box.

It's probably a good idea to think about this some more, as it's a thing that bothers me in general.

FWIW, stripping parse errors is the least of your possible problems, as they're absolutely harmless at runtime as well. c-l does strip them, but it also strips messages that fatally fail linter checks. Right now, there's none of them affecting bedrock, though.

For now, this is good enough.

PR #8583, opened 2020-02-24
Scale Structured data to locales (Fixes #8522)

### LINKED ISSUE #8522: Scale Structured data to tier 1 locales
## Description

We want to scale our structured product data to all locales (currently only on en-US). To integrate this on our locales, we need to automate some changes to the json files the data is based on, the urls are locales specific. To be precise some data is only language specific, not locale specific and some are locale specific:

I will try to catch all changes here and we need to discuss what's in scope so we can ship V1. IDs, URLs and sameAs: wikipedia MUST be localized, all other references should be but the system will (partly) work incomplete as well.

## IDs
All ids need to be specific to the locale. 

For example ` "@id": "https://www.mozilla.org/#firefoxlockwise",` stays like it is on en-US, but en-CA will get ` "@id": "https://www.mozilla.org/#firefoxlockwise-en-ca",`

For de it would be `"@id": "https://www.mozilla.org/#firefoxlockwise-de",`

For pt-PT it will be `"@id": "https://www.mozilla.org/#firefoxlockwise-pt-pt",` and for pt-BR it will be `"@id": "https://www.mozilla.org/#firefoxlockwise-pt-br",`. As pt only is part of our hreflang but does not resolve in a status code 200 itself it will not have any id. 

## URLs

URLs point always to the locale specific url

For example ` "url": "https://www.mozilla.org/en-US/firefox/lockwise/",` stays like it is on en-US, but en-CA will get ` "url": "https://www.mozilla.org/en-CA/firefox/lockwise/",`

For de it would be `"url": "https://www.mozilla.org/de/firefox/lockwise/",`

## sameAs

Not all urls in the sameAs need to be altered. Only wikipedia, amazon and playstore/appstore links need to be language (not locale) specific):

For wikipedia we just alter the subdomain to fit the language part of our locales:

for en-US, en-CA and en-GB it will be all `"sameAs": [
      "https://en.wikipedia.org/wiki/Firefox_Lockwise",`

for de it will be `"sameAs": [
      "https://de.wikipedia.org/wiki/Firefox_Lockwise",`

For pt-PT and pt-BR it will be`"sameAs": [
      "https://pt.wikipedia.org/wiki/Firefox_Lockwise",`

I know that some of these links will not work, as the wikipedia pages do not exist, but right now I cant think of any workaround which would complicate everything by far and I am sure that not resolving links will be just ignored.

For the playstore links we have to just add &hl=language at the url:

for en-US, en-CA and en-GB it will be all `"sameAs": [
      "https://play.google.com/store/apps/details?id=mozilla.lockbox&hl=en",`

for de it will be `"sameAs": [
      "https://play.google.com/store/apps/details?id=mozilla.lockbox&hl=de",`

For pt-PT and pt-BR it will be`"sameAs": [
      "https://play.google.com/store/apps/details?id=mozilla.lockbox&hl=pt",`

Appstore: TBD as their hreflang setup is a bit weird and I need to map it to ours.
Amazon: TBD as their hreflang setup is a bit weird and I need to map it to ours.

## Pages we need to localize structured data on:

- Homepage
- /new
- /mobile
- /mac
- /linux
- /windows
- /developer
- /channel/desktop
- /enterprise
- /lockwise

But the "system" we build should work on all pages, whenever we include structured data in a new /en-US/ (planned is /firefox and any "new" product pages on bedrock)

## Example include on /moz.org/de/

[moz_org_de_structured_data.txt](https://github.com/mozilla/bedrock/files/4199498/moz_org_de_structured_data.txt)

---

- [ ] This is an example of a success criteria. Use a checklist!
- [ ] This is an example of a success criteria.

#### Tasks
- [ ] Check all existing structured data and make sure all urls are locale based (currently some are for testing purpose pointing to the redirect without a locale)

- [ ] Define how we will technically scale and automate this so we do not need json files for every single locale if possible

alexgibson:
## Description
- Exposed structured data to more locales.
  - Adds helpers to generate structured data IDs, app store links, and shortened locale codes.
  - Moves app store links into `appstores.py`.
  - Use `url()` helper for generating links.

## Issue / Bugzilla link
#8522

## Testing
Demo: https://www-demo2.allizom.org/en-US/

Successful test run: https://gitlab.com/mozmeao/www-config/pipelines/120549500

### DIFF
--- a/bedrock/mozorg/templatetags/misc.py
+++ b/bedrock/mozorg/templatetags/misc.py
@@ -599,6 +599,51 @@ def ifeq(a, b, text):
     return jinja2.Markup(text if a == b else '')
 
 
+@library.global_function
+@jinja2.contextfunction
+def app_store_url(ctx, base_url):
+    """Returns a localised app store URL for a given product"""
+    locale = getattr(ctx['request'], 'locale', 'en-US')
+    countries = settings.APPLE_APPSTORE_COUNTRY_MAP
+    if locale in countries:
+        return base_url.format(country=countries[locale])
+    else:
+        return base_url.replace('/{country}/', '/')
+
+
+@library.global_function
+@jinja2.contextfunction
+def play_store_url(ctx, base_url):
+    """Returns a localised play store URL for a given product"""
+    locale = getattr(ctx['request'], 'locale', 'en-US')
+    return base_url + '&hl=' + locale.split('-')[0]
+
+
+@library.global_function
+@jinja2.contextfunction
+def structured_data_id(ctx, id, domain=None):
+    """
+    Returns an identifier for a structured data object based on
+    a supplied id e.g. https://www.mozilla.org/#firefoxbrowser
+    """
+    canonical = settings.CANONICAL_URL if not domain else domain
+    locale = getattr(ctx['request'], 'locale', 'en-US')
+    suffix = ''
+
+    if locale != 'en-US':
+        suffix = '-' + locale.lower()
+
+    return canonical + '/#' + id + suffix
+
+
+@library.global_function
+@jinja2.contextfunction
+def lang_short(ctx):
+    """Returns a shortened locale code e.g. en."""
+    locale = getattr(ctx['request'], 'locale', 'en-US')
+    return locale.split('-')[0]
+
+
 def _get_adjust_link(adjust_url, app_store_url, google_play_url, redirect, locale, adgroup, creative=None):
     link = adjust_url
     params = 'campaign=www.mozilla.org&adgroup=' + adgroup
--- a/bedrock/mozorg/templatetags/misc.py
+++ b/bedrock/mozorg/templatetags/misc.py
@@ -601,8 +601,9 @@ def ifeq(a, b, text):
 
 @library.global_function
 @jinja2.contextfunction
-def app_store_url(ctx, base_url):
+def app_store_url(ctx, product):
     """Returns a localised app store URL for a given product"""
+    base_url = getattr(settings, f'APPLE_APPSTORE_{product.upper()}_LINK')
     locale = getattr(ctx['request'], 'locale', 'en-US')
     countries = settings.APPLE_APPSTORE_COUNTRY_MAP
     if locale in countries:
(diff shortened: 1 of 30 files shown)


alexgibson:
@pmac if you have some spare time, r on the python bits here? I'd be interested if there are any easier / simpler ways to do some of the helpers.

bedrock/mozorg/templatetags/misc.py
```diff
@@ -599,6 +599,51 @@ def ifeq(a, b, text):
...
+    else:
+        return base_url.replace('/{country}/', '/')
+
+
+@library.global_function
+@jinja2.contextfunction
+def play_store_url(ctx, base_url):
+    """Returns a localised play store URL for a given product"""
+    locale = getattr(ctx['request'], 'locale', 'en-US')
+    return base_url + '&hl=' + locale.split('-')[0]
```
pmac: Could be simplified by using your helper function below, removing the line above, and an F string.
```suggestion
    return f'{base_url}&hl={lang_short(ctx)}'
```

bedrock/mozorg/templatetags/misc.py
```diff
@@ -599,6 +599,51 @@ def ifeq(a, b, text):
     return jinja2.Markup(text if a == b else '')
 
 
+@library.global_function
+@jinja2.contextfunction
+def app_store_url(ctx, base_url):
```
pmac: I think it would be nice to simplify how these are called. I think it might be a pain to look up the setting name every time you want to use this and the `play_store_url` helpers. I can see a couple of good options.

You could use these functions as just local python helpers and expose a function per product to the templates. So you'd have `app_store_url_lockwise()`e.g.. The other good option IMO is to use the fact that you've got a good variable naming convention going and use a simple string passed into the function to get the right variable. For example:

```python
def app_store_url(ctx, product):
    base_url = getattr(settings, f'APPLE_APPSTORE_{product.upper()}_LINK')
```

Certainly not required, but should help when you want to use these in future.

alexgibson:
Thanks @pmac - i updated with your suggestions and pushed another commit.

I also re-ran integration tests for good measure: https://gitlab.com/mozmeao/www-config/pipelines/121235818

If all looks ok, I'll merge 👍

pmac:
Looks great! :shipit:

alexgibson:
@rraue is happy with the data so I'm going to merge this. If we get any errors in the google search console we can follow up with fixes, but these should be invisible to users.

PR #8590, opened 2020-02-25
Firefox.com mode

pmac:
This is a very quick demonstration of how this could work. There is A LOT more work to do to make bedrock able to serve both appropriately, but this gives you the idea.

Set `FIREFOX_MODE=True` in your environment and you'll see firefox content as main.

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -6,6 +6,7 @@ from operator import itemgetter
 from urllib.parse import urlencode
 
 from django.conf import settings
+from django.urls import reverse_lazy
 
 from everett.manager import ListOf
 from product_details import ProductDetails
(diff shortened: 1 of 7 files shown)


alexgibson:
I did some preliminary testing and found some things worth noting:

- Stub attribution would need a [small updat](https://github.com/mozilla/bedrock/blob/master/media/js/base/stub-attribution.js#L229-L232)e to work.
- Vanify redirects such as /accounts/ -> /firefox/accoounts/ are going in an endless loop.
- The page at /firefox/privacy/ [clobbers](https://www-demo-firefox-com-mo-ybg0u5.herokuapp.com/en-US/privacy/) the page at /privacy/. Although the privacy notices themselves are [still intact](https://www-demo-firefox-com-mo-ybg0u5.herokuapp.com/en-US/privacy/firefox/).
- The [contact pages](https://www-demo-firefox-com-mo-ybg0u5.herokuapp.com/en-US/contact/) are throwing errors with `NoReverseMatch: Reverse for 'press.press-inquiry' not found.`

Will continue to add if I encounter anything else. Overall though this looks great, nice work!

pmac:
The big thing to figure out IMO will be how to make it output URLs for www.m.o for pages still there when in Firefox mode, and the reverse for www.f.c when in Mozilla mode. Might be able to deal with it in the `url()` helper, but we'll see.

alexgibson:
> The big thing to figure out IMO will be how to make it output URLs for www.m.o for pages still there when in Firefox mode, and the reverse for www.f.c when in Mozilla mode. Might be able to deal with it in the url() helper, but we'll see.

Yeah, seeing as the URLs are all relative, I'm not sure how we be able to tell which app the path is defined in. Could switch to using absolute URLs, prepending the right domain maybe?

alexgibson:
Canonical and hreflang URLs are all still pointing to www.mozilla.org when in firefox.com mode.

alexgibson:
Sitemap reflects mozorg: https://www-demo-firefox-com-mo-ybg0u5.herokuapp.com/en-US/sitemap.xml

pmac:
Closing this for now. Leaving the branch intact.

Issue #8595, opened 2020-02-26
Port firefox/switch page to Fluent


alexgibson:
## Description

The L10n team would like to run a second test for the vendor pipeline, so we've settled on porting the [firefox/switch](https://www.mozilla.org/en-US/firefox/switch/) page.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Page is updated in production to use Fluent for translation.

#### Tasks
- [ ] Migrate .lang files to .ftl
- [ ] Migrate template.
- [ ] Expose strings.

Pike:
First step, create recipe ;-)

alexgibson:
note to self: https://bedrock.readthedocs.io/en/latest/fluent-conversion.html

alexgibson:
@peiying2 has asked if we can re-run this migration using placeholders for brand names. Re-opening.

PR #8599, opened 2020-02-27
Migration and switch

Pike:
## Description

Migrate /firefox/switch/ to Fluent

## Issue / Bugzilla link

#8595

## Testing

Tested the `en-US` page as well as `de` and `bn`.

Migration to mozmeao/www-l10n is going to be a separate PR onto that repo.


l10n/en/firefox/switch-en.ftl
```diff
@@ -0,0 +1,8 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### TODO: Merge the content here with switch.ftl to expose new string.
```
alexgibson: 👍

Issue #8600, opened 2020-02-27
Experiment: sticky navigation on /firefox


amychurchwell:
## Description
Hypothesis: If we make the global navigation a sticky navigation, CTA clicks in global nav CTA will increase.

## Test Plan
https://docs.google.com/document/d/1eja21ZlEJQfi-9jjWUhxx1zD6Gnm5EeaGMR8BZBIOP8/edit?usp=sharing

#### Tasks
- [x] Put proposed nav on demo for design review.
- [x] Code experiment for final review.

#### Demo
https://www-demo-demo-nav-sdavucfmzlyu.herokuapp.com/en-US/exp/firefox/

amychurchwell:
@stephaniehobson or @alexgibson would either of you be able to review the draft of this experiment on convert? (ID 10034124)

alexgibson:
Link to experiment summary for posterity: https://app.convert.com/#/projects/1003343/summary/10034124

alexgibson:
Some initial review feedback:

- Is there a test plan for this experiment, I can't seem to find one linked here?
- I think I managed to fix the preview links by changing the editor URL to point to stage (I had to manually copy over the CSS code as I initially clicked the wrong thing (sorry), hope I did that ok?)
- Traffic distribution for the experiment is set to 24%. Should this be 100%? I'm guessing we need to redirect 24% of traffic from /firefox/ to /exp/firefox/?
- Should we add bounce rate as a secondary goal perhaps? (I'm guessing more people might end up navigating to a different page?)
- Would it be worth adding a vertical media query, to stop the nav taking over on smaller desktop screens?
- The goal for "Click: Download button in navigation on /firefox" is set as "Site Wide". Should it be specific to /firefox?
- The same goal is set to track clicks on `#protocol-nav-download-firefox`, but the actual download link is a child of that element. Should it be something more specific? e.g. `DIV#protocol-nav-download-firefox .download-list > li > a.download-link`?

alexgibson:
Another thing that may be worth exploring:

- Do we need to consider the effect this may have on the page's primary goals (which aren't neessarily to download Firefox or navigate)? Do we need to measure if an increase in clicks in the nav leads to a decrease in click throughs for other products?

amychurchwell:
[...]

PR #8601, opened 2020-02-27
Sticky navigation demo (#8600)

amychurchwell:
## Description
Rough draft of sticky navigation.

## Issue / Bugzilla link
#8600

## Testing
https://www-demo-demo-nav-sdavucfmzlyu.herokuapp.com/en-US/exp/firefox/

### DIFF
--- a/media/css/exp/firefox/index.scss
+++ b/media/css/exp/firefox/index.scss
@@ -19,6 +19,22 @@ $image-path: '/media/protocol/img';
 }
 
 // ------------------------------------------------------------------------------
+// Sticky nav test.
+.mzp-c-navigation {
+    @media #{$mq-lg} {
+        position: fixed;
+        z-index: 999; // sorry
+        top: 0;
+        width: 100%;
+        border-bottom: 1px solid rgba(0, 0, 0, 0.2);
+    }
+}
+
+.c-lead {
+    margin-top: 100px !important; // also sorry
+}
+
+// ------------------------------------------------------------------------------
 // component over rides (content menu list, feature card)
 
 .firefox-home {


alexgibson:
I was excited to tes this out - hope you don't mind some feedback on a couple of UX issues I noteced:

1. On mobile I couldn't scroll the nav content or the screen when open:

nav-mobile

2. On desktop, if the user's screen is small then the nav can (sometimes accidentally) cover the entire screen, preventing scroll and making the nav hard to close. If the nav content is also longer than the screen height, that content is also inaccessible.

Screenshot 2020-02-28 at 10 23 20

amychurchwell:
@alexgibson -- thanks for the feedback, Alex! This isn't quite ready for review yet. Current state is a super quick rough draft to make sure I was in the right direction.

amychurchwell:
On UX issues:

> 1. On mobile I couldn't scroll the nav content or the screen when open:

I think for this test we should make the sticky nav desktop only. There is no CTA in the nav on mobile, so it wouldn't help out the experiment imo.

> 2. On desktop, if the user's screen is small then the nav can (sometimes accidentally) cover the entire screen, preventing scroll and making the nav hard to close. If the nav content is also taller than the screen height, that content is then inaccessible.

When the nav covers the entire screen, it seems like it doesn't prevent scroll but you can't see where you are scrolling & a user might not realize. Perhaps we should prevent scrolling if the nav covers the entire view? 

If the nav covering the entire screen makes it trickier to close- maybe we should move the 'x' to the left instead? Or maybe make the nav narrower so you can see the page to the left and right that would allow a user to click off the nav to close?

Also, I am going to tinker with making the nav container vertically responsive so it does not break out of screen height.

justinkropp:
@amychurchwell This LGTM. 

I agree with the following:

> I think for this test we should make the sticky nav desktop only. There is no CTA in the nav on mobile, so it wouldn't help out the experiment imo.

The default behavior on mobile/small viewports in the demo feels right. Users on mobile are more than willing to scroll, so letting them scroll through the expanded nav _and_ the page is ideal. I would also keep the "X" as is. The majority of users are right-handed and moving the "X" to the left could present some possible usability issues (harder for them to reach with their thumbs if operating the one-handed).

Hope that helps :)

amychurchwell:
Thank you for the feedback everyone. Closing this pull request and moving conversation to the issue. 👍

Issue #8546, opened 2020-02-19
(3-2-2020) - WNP 74 - Code


ejregithub:
## Description

Code the WNP for the 74 product release on 3/10/20

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] Page is live on 2/28

#### Approvals
- [ ] Brand
- [ ] PMM (TBD)
- [ ] Team Lead
- [ ] Accountable stakeholder
- [ ] Legal

craigcook:
Demo: https://www-demo3.allizom.org/firefox/74.0/whatsnew/all/

hoosteeno:
We need...

* icons (@cgonz0 ?)
* final copy in en (@rantley ?)
* final copy de/fr (@ejregithub ?)
* an analytics review (@hoosteeno )

PR #8603, opened 2020-02-28
[#8546] Add WNP 74

craigcook:
## Description
Adds whatsnew page for Firefox 74.0

There are some locale-specific, hard-coded blurbs behind conditionals for English, French, and German.

## Issue / Bugzilla link
#8546 

## Testing
https://www-demo3.allizom.org/firefox/74.0/whatsnew/all/


craigcook:
Note for reviewer: because there's so little time for l10n before this release, we're using the list of active locales for WNP73 as a proxy for WNP74, then the view logic sets the 73 page as a fallback for any locales that don't have the 74 page active yet. This is a bit of a kludge to buy localizers more time. But please double-check the view to make sure it's sensible and doesn't break anything else.

PR #8628, opened 2020-03-05
Convert FxA macro buttons to python helpers (Fixes #8185)

### LINKED ISSUE #8185: Convert Sync buttons to new FxA product button Python helper.
## Description

The work done in #8154 added a dedicated [Python helper for creating FxA product buttons.](https://github.com/mozilla/bedrock/blob/master/bedrock/mozorg/templatetags/misc.py#L723-L742) 

We should convert the Sync buttons to use the new helper.

Same task for Monitor buttons: #8167

---
#### Tasks
- [ ] Convert Sync buttons (`fxa_cta_link`) to the new Python helper.
- [ ] Write tests.

alexgibson:
## Description
- Removes `fxa_cta_link` and `fxa_link_fragment` macros.
- Replaces with `fxa_button` and `fxa_link_fragment` python helpers.
- Adds unit tests 🎉
- Updates documentation.

URLs:
- http://localhost:8000/en-US/firefox/ (navigation and footer links, navigation FxA button, FxA button at bottom of page)
- http://localhost:8000/en-US/firefox/accounts/
- http://localhost:8000/en-US/exp/firefox/accounts/
- http://localhost:8000/de/exp/firefox/
- http://localhost:8000/en-US/exp/firefox/
- http://localhost:8000/en-US/exp/firefox/71.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/browsers/
- http://localhost:8000/en-US/firefox/firstrun/
- http://localhost:8000/en-US/firefox/new/ (legacy template)
- http://localhost:8000/en-US/firefox/privacy/products/
- http://localhost:8000/en-US/firefox/products/
- http://localhost:8000/en-US/firefox/welcome/3/
- http://localhost:8000/en-US/firefox/60.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/71.0/whatsnew/all/

## Issue / Bugzilla link
#8185

## Testing
- [ ] I made a couple of small consistency improvements to tracking parameters, but largely the buttons and links should all contain the same paths and parameters?

Successful test run: https://gitlab.com/mozmeao/www-config/pipelines/123942605

### DIFF
--- a/bedrock/mozorg/templatetags/misc.py
+++ b/bedrock/mozorg/templatetags/misc.py
@@ -858,7 +858,7 @@ def fxa_link_fragment(ctx, entrypoint, action='signup', optional_parameters=None
 
         {% set signin = fxa_link_fragment(entrypoint='mozilla.org-firefox-accounts') %}
         {% set class_name = 'js-fxa-cta-link js-fxa-product-button' %}
-        <p>Already have an account? <a {{ sign_in }} {{ class_name }}>Sign In</a> to start syncing.</p>
+        <p>Already have an account? <a {{ sign_in }} class="{{ class_name }}">Sign In</a> to start syncing.</p>
     """
 
     if action == 'email':
@@ -898,10 +898,8 @@ def fxa_button(ctx, entrypoint, button_text, action='signup', class_name=None, i
         'data-mozillaonline-link': _fxa_product_url(mozillaonline_product_url, entrypoint, optional_parameters)
     }
 
-    if optional_attributes:
-        optional_attributes.update(mozillaonline_attribute)
-    else:
-        optional_attributes = mozillaonline_attribute
+    optional_attributes = optional_attributes or {}
+    optional_attributes.update(mozillaonline_attribute)
 
     return _fxa_product_button(product_url, entrypoint, button_text, class_name, is_button_class, include_metrics,
                                optional_parameters, optional_attributes)
--- a/bedrock/mozorg/templatetags/misc.py
+++ b/bedrock/mozorg/templatetags/misc.py
@@ -766,18 +766,28 @@ def lockwise_adjust_url(ctx, redirect, adgroup, creative=None):
     return _get_adjust_link(adjust_url, app_store_url, play_store_url, redirect, locale, adgroup, creative)
 
 
-def _fxa_product_button(product_url, entrypoint, button_text, class_name=None, is_button_class=True,
-                        optional_parameters=None, optional_attributes=None):
-    href = f'{product_url}?entrypoint={entrypoint}&form_type=button&utm_source={entrypoint}&utm_medium=refferal'
-    css_class = 'js-fxa-cta-link js-fxa-product-button'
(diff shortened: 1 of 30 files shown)


craigcook:
Found an interesting bug on the /firefox and /firefox/pocket pages when logged in... the button in the navbar gets its URL changed to point to /firefox/accounts (the URL swap happens in `init-navigation.js`) but because it still has the `js-fxa-product-button` class it gets targeted by `mozilla-fxa-product-button.js` and the flow params get appended. The button then leads to a 404 because it's a malformed query string (missing a `?` at the start).

My first idea for a fix is to update `init-navigation.js` to remove the `js-fxa-product-button` class when it does the URL swap so it's no longer treated as an FxA button. There may be other ways too. We could possibly add a check in `mozilla-fxa-product-button.js` to make sure it really is a product button before it fetches and appends the flow params, maybe just checking for `?entrypoint` in the href?

alexgibson:
Thanks @craigcook nice catch!

I added two commits that do the following:

- Only add metrics params to whitelisted domains.
- Exclude the nav button from metrics params (_technically_ we could add them here, but currently we don't apply this to the nav site wide, only to in-page primary CTAs). Rather than start doing it only on some pages, let's reconsider this as a site-wide strategy later, if we need to?

bedrock/mozorg/templatetags/misc.py
```diff
@@ -825,4 +838,70 @@ def monitor_fxa_button(ctx, entrypoint, button_text, class_name=None, is_button_
...
+
+    Examples
+    ========
+
+    In Template
+    -----------
+
+        {% set signin = fxa_link_fragment(entrypoint='mozilla.org-firefox-accounts') %}
+        {% set class_name = 'js-fxa-cta-link js-fxa-product-button' %}
+        <p>Already have an account? <a {{ sign_in }} {{ class_name }}>Sign In</a> to start syncing.</p>
```
pmac: shouldn't that be?

```suggestion
        <p>Already have an account? <a {{ sign_in }} class="{{ class_name }}">Sign In</a> to start syncing.</p>
```

bedrock/mozorg/templatetags/misc.py
```diff
@@ -825,4 +838,70 @@ def monitor_fxa_button(ctx, entrypoint, button_text, class_name=None, is_button_
...
+        action = '?action=email'
+
+    product_url = f'{settings.FXA_ENDPOINT}{action}'
+    mozillaonline_product_url = f'{settings.FXA_ENDPOINT_MOZILLAONLINE}{action}'
+
+    mozillaonline_attribute = {
+        'data-mozillaonline-link': _fxa_product_url(mozillaonline_product_url, entrypoint, optional_parameters)
+    }
+
+    if optional_attributes:
```
pmac: Style suggestion. Take it or leave it. Definitely not a required change:

```python
optional_attributes = optional_attributes or {}
optional_attributes.update(mozillaonline_attribute)
```

PR #8638, opened 2020-03-09
Only show nightly or dev edition CTA when appropriate channel (Fixes #8579)

### LINKED ISSUE #8579: Only show nightly down if ?channel=nightly on firefox/installer-help/
## Description

Users arriving on firefox/installer-help/ from the nightly channel should be shown the Firefox nightly download button instead of all of the download buttons. Like we do for the other channels.

[beta](
https://www.mozilla.org/en-US/firefox/installer-help/?channel=beta&installer_lang=en-US)
Screen Shot 2020-02-21 at 2 47 47 PM

[nightly](
https://www.mozilla.org/en-US/firefox/installer-help/?channel=nightly&installer_lang=en-US)
Screen Shot 2020-02-21 at 2 47 36 PM

amychurchwell:
## Description
- Only show nightly download if `?channel=nightly`
- Only show dev edition download if `?channel=alpha` or `aurora`

## Issue / Bugzilla link
#8579

## Testing
- [ ] tests needing updating?

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -43,7 +43,8 @@ INSTALLER_CHANNElS = [
     'release',
     'beta',
     'alpha',
-    # 'nightly',  # soon
+    'nightly',
+    'aurora',  # deprecated name for dev edition
 ]
 SEND_TO_DEVICE_MESSAGE_SETS = settings.SEND_TO_DEVICE_MESSAGE_SETS
 
@@ -69,7 +70,10 @@ def installer_help(request):
         context['installer_lang'] = installer_lang
 
     if installer_channel and installer_channel in INSTALLER_CHANNElS:
-        context['installer_channel'] = installer_channel
+        if installer_channel == 'aurora':
+            context['installer_channel'] = 'alpha'
+        else:
+            context['installer_channel'] = installer_channel
 
     return l10n_utils.render(request, 'firefox/installer-help.html', context)
 


amychurchwell:
(note: Not sure the history behind the win64 code being commented out & unsure if there are other places needing updating?)

media/css/pebbles/components/_buttons-download.scss
```diff
@@ -107,8 +107,7 @@ ul.download-list {
 .download-button .os_win64-msi,
 .download-button .os_linux,
 .download-button .os_linux64,
-// Uncomment this to enable the win64 download buttons:
-// .win7up.x86.x64 .download-button .os_win,
+.win7up.x86.x64 .download-button .os_win,
```
alexgibson: @amychurchwell this represents a pretty large and untested change that effects the whole of the website, so we may need to be a little careful here.

As far as I know, 64bit download buttons have never been enabled by default for Windows, because we understand that relying on user agent detection is not entirely accurate when it comes to system architecture (I think the product teams may have some data here). There may also be memory requirements to consider too. So instead, we've been relying on the stub installer delivering the correct binary, which it can do more reliably than we can here on the website.

As part of bedrock's technical roadmap, I'd like to see us try and reduce our reliance on user agent strings as much as is technically possible. We are seeing clear signals from other browsers that they intend to deprecate and freeze such information over time.

There is a suggestion in the original bug you linked to that may be worth exploring:

```
Maybe stub installer can pass along its detection result and make this happen?
```

If the architecture could be passed along via a query parameter to the stub installer page, then we may be able to work out a way to show the right build. Thoughts?

If/when we do enable 64bit builds by default for Windows, then we should probably do that as it's own distinct piece of work, and figure out the best way to implement and monitor it.

bedrock/firefox/views.py
```diff
@@ -43,7 +43,7 @@
     'release',
     'beta',
     'alpha',
-    # 'nightly',  # soon
+    'nightly',
```
alexgibson: Do we need to add `aurora` here too? I still see traffic with that parameter in GA

amychurchwell: When I append `?channel=aurora` to the URL it shows a download for `?product=firefox-aurora-latest-ssl` which 404's, so I don't think we should but it's possible I am misunderstanding.

amychurchwell: *The 404 behavior happens when `aurora` is included in this array. 

If `aurora` is what dev edition used to be, we might need to redirect those users to `alpha` or ask that that query parameter is replaced with `alpha`?

alexgibson: Perhaps we chould be treat `alpha` and `aurora` as the same? I believe we should be able to do that without changing or rewriting the query parameter itself, if we handled it in the view logic.

media/css/pebbles/components/_buttons-download.scss
amychurchwell: Thank you for the info, Alex. Definitely do not want to enable by default.
I am going to separate these two issues because this feels larger scope & I would like to get a bugfix in for the dev edition & nightly CTA issue (#8579)

bedrock/firefox/views.py
```diff
@@ -68,6 +69,9 @@ def installer_help(request):
     if installer_lang and installer_lang in firefox_desktop.languages:
         context['installer_lang'] = installer_lang
 
+    if installer_channel and installer_channel == 'aurora':
+        context['installer_channel'] = 'alpha'
+
     if installer_channel and installer_channel in INSTALLER_CHANNElS:
```
alexgibson: The conditional you added above doesn't seem to be working, because it is overwritten by this conditional here below. Something like this should work?

```suggestion
    if installer_channel and installer_channel in INSTALLER_CHANNElS:
        if installer_channel == 'aurora':
            context['installer_channel'] = 'alpha'
        else:
            context['installer_channel'] = installer_channel
```

amychurchwell: ah whoops, thank you Alex.

bedrock/firefox/views.py
```diff
@@ -68,7 +69,14 @@ def installer_help(request):
     if installer_lang and installer_lang in firefox_desktop.languages:
         context['installer_lang'] = installer_lang
 
+    if installer_channel and installer_channel == 'aurora':
```
alexgibson: I think you can remove this as it's now redundant?

bedrock/firefox/views.py
```diff
@@ -68,7 +69,14 @@ def installer_help(request):
...
+        context['installer_channel'] = 'alpha'
+
     if installer_channel and installer_channel in INSTALLER_CHANNElS:
+        if installer_channel == 'aurora':
+            context['installer_channel'] = 'alpha'
+        else:
+            context['installer_channel'] = installer_channel
         context['installer_channel'] = installer_channel
```
alexgibson: This is overriding the conditional you've added above. I think if you remove this line then things should work ok.

PR #8642, opened 2020-03-10
Decommission /technology landing page (Fixes #6824)

### LINKED ISSUE #6824: Determine the fate of the /technology page
The [technology page](https://www.mozilla.org/en-US/technology/) hasn't been updated in a long time. We no longer link to it from the global navigation, and we're now replacing links to it from the home page pointing to https://labs.mozilla.org/ instead.

Unless we plan on updating this content, I think we need to consider what to do with this page long term, especially as we are now using https://labs.mozilla.org as the hub to talk about these types of projects.

Options are:

1. We keep the page as-is and let it grow stale. There may be some SEO benefits here, but I'm not sure if it really does Mozilla any huge favors if the content is outdated and no longer reflects how we are focusing this type of work. The page is also not using Protocol, so keeping it around is not debt-free.
2. We decommission the page and redirect it to https://labs.mozilla.org/.
3. We re-purpose the page as an index for all the recently added SEO pages (porting it to Protocol):
    - https://www.mozilla.org/en-US/technology/what-is-a-browser/
    - https://www.mozilla.org/en-US/technology/update-your-browser/
    - https://www.mozilla.org/en-US/technology/incognito-browser/
    - https://www.mozilla.org/en-US/technology/browser-history/
    - (and soon to be more)
4. We keep the current purpose of the page, redesign it and update the content (porting it to Protocol).

My preference would probably be 3, as creating all these orphaned pages probably isn't great for SEO anyways.

Interested in @rraue's thoughts here.

alexgibson:
## Description
- Removes the /technology landing page and redirects to https://labs.mozilla.org/

## Issue / Bugzilla link
#6824

## Testing
- [ ] Did I miss any stray assets or links?

Successful test run: https://gitlab.com/mozmeao/www-config/pipelines/124911867


tests/redirects/map_globalconf.py
```diff
@@ -1315,4 +1315,7 @@
 
     # Issue 8418
     url_test('/styleguide/', 'https://mozilla.design/'),
+
+    # Issue 6824
```
amychurchwell: `redirects.py` & `map_globalconf.py` are having merge conflicts 🥀

alexgibson:
@amychurchwell thanks, rebased 👍

PR #8671, opened 2020-03-17
welcome page 7

Errolyn:
## Description

Add welcome page based on whatsnew 74

## Issue / Bugzilla link
#8574

## Testing
There is one coded block for FR, DE, en- and everything else. They should all work.
http://localhost:8000/firefox/welcome/7/


bedrock/firefox/templates/firefox/welcome/page7.html
```diff
@@ -0,0 +1,118 @@
...
+{% call hero(
+    title=_('It’s okay to like Facebook'),
+    desc=_('If you still kinda like Facebook but don’t trust them, then try the Facebook Container extension by Firefox and make it harder for them to track you around the web.'),
+    class='mzp-t-firefox mzp-t-dark',
+    include_cta=True,
+    heading_level=1
+  ) %}
+
+  <p class="primary-cta">
+    <a href="https://addons.mozilla.org/firefox/addon/facebook-container/?src=external-www.mozilla.org-whatsnew74&utm_source=www.mozilla.org-whatsnew74&utm_medium=referral&utm_campaign=feb2020" class="mzp-c-button mzp-t-product" id="facebook-container-button" rel="external">{{ _('Get Facebook Container') }}</a>
```
craigcook: These URL params still mention whatsnew74 and need to be updated for this page, here and on the other links. @hoosteeno or @sghoseWI might have input on what the preferred values should be.

media/css/firefox/welcome.scss
```diff
@@ -71,14 +71,15 @@ $image-path: '/media/protocol/img';
     }
 }
 
-.mzp-c-hero-title {
+.mzp-c-hero-title, .hero-subheading {
```
craigcook: I don't think the `.hero-subheading` class is used. Is this is a leftover?

craigcook:
Also don't forget to update that string on welcome 6!

media/css/firefox/welcome.scss
Errolyn: yep you are totally right, I was trying something and missed taking that bit out!

PR #8681, opened 2020-03-20
Fix Fluent error on supported nonlocale pages (Fixes #8680)

alexgibson:
Fixes https://github.com/mozilla/bedrock/issues/8680

PR #8682, opened 2020-03-20
Fix Fluent error on supported nonlocale pages

### LINKED ISSUE #8680: Fluent error on pages that don't have a locale
## Description

This is a regression from https://github.com/mozilla/bedrock/pull/8665

Pages that don't have a locale, such as https://bedrock-dev.gcp.moz.works/locales/ are not loading correctly.

In Sentry I see:

```
KeyError: 'fluent_l10n'
(19 additional frame(s) were not displayed)
...
  File "bedrock/mozorg/util.py", line 71, in _view
    return django_render(request, tmpl, kwargs)
  File "django_jinja/backend.py", line 63, in render
    return mark_safe(self._process_template(self.template.render, context, request))
  File "django_jinja/backend.py", line 109, in _process_template
    return handler(context)
  File "lib/l10n_utils/templatetags/fluent.py", line 26, in ftl
    return jinja2.Markup(fluent.translate(ctx['fluent_l10n'], message_id, fallback, **kwargs))
```

pmac:
This change results in passing every page render through the l10n system, even ones in the SUPPORTED_NONLOCALE list.

Fix #8680

### DIFF
--- a/bedrock/mozorg/tests/test_util.py
+++ b/bedrock/mozorg/tests/test_util.py
@@ -7,8 +7,6 @@
 import os
 
 from django.test import RequestFactory
-from django.test.utils import override_settings
-
 from mock import ANY, patch
 
 from bedrock.mozorg.tests import TestCase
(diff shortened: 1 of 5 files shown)


pmac:
@alexgibson I think this approach might be more thorough and better long-term than #8681. Let me know what you think.

PR #8700, opened 2020-03-25
[#8564] Update WNP75 state logic

craigcook:
## Description
We should only show the Monitor pitch to users signed into FxA with at least one mobile device. All others get the Sync pitch.

## Testing
https://www-demo3.allizom.org/firefox/75.0/whatsnew/all/

- [ ] Signed out users get Sync
- [ ] Signed in users with no devices get Sync
- [ ] Signed in users with one or more devices get Monitor

### DIFF
--- a/media/js/firefox/whatsnew/whatsnew-75.js
+++ b/media/js/firefox/whatsnew/whatsnew-75.js
@@ -19,6 +19,15 @@ if (typeof window.Mozilla === 'undefined') {
                 document.querySelector('.c-page-header').classList.add('show-up-to-date-message');
             }
         });
+
+        client.getFxaDetails(function(details) {
+            if (details.setup && details.browserServices.sync.mobileDevices > 0) {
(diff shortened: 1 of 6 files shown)


media/js/firefox/whatsnew/whatsnew-75.js
```diff
@@ -19,6 +19,15 @@ if (typeof window.Mozilla === 'undefined') {
                 document.querySelector('.c-page-header').classList.add('show-up-to-date-message');
             }
         });
+
+        client.getFxaDetails(function(details) {
+            if (details.setup && details.browserServices.sync.mobileDevices > 0) {
```
craigcook: Detecting synced devices isn't currently part of our standardized mozilla-fxa.js lib, but I wonder if we should add it there?

alexgibson: Could be worth adding yeah, file a followup issue perhaps?

craigcook:
Updated! Also assigned body to a variable.

PR #8715, opened 2020-03-27
Remove events app

### LINKED ISSUE #8474: Remove events app
Follow up to #8473 (#8402). Those changes need to go to prod so that the database cleanup will happen. After that we can fully remove the events app and other migrations that have deleted models.

pmac:
Fix #8474


pmac:
Doh! Failed to remove it from `INSTALLED_APPS`. I'll update.

PR #8743, opened 2020-04-03
Experiment: Add social proof block to release notes.

amychurchwell:
## Description
Adds a social proof section on release notes with download button.

> Hypothesis: People visiting a release notes page are looking for a reason to try the browser, something compelling. If we add a block of social proof content to the page, near the top, with a CTA, we will see a bigger conversion rate on the page. 

**Test Plan:** https://docs.google.com/document/d/1x2FsYMqU_CFZpPDA2ktxsj9NhbbtuUnBCUX1Cv13atE/edit?usp=sharing

## Issue / Bugzilla link
**CRO Team Card:** https://airtable.com/tblO5zD4uNk7ISyG0/viwIUJsHKDrOFl90B/recfUUVL7JULESFZp

## Testing

- [ ] **Audience is en-US, desktop, non-firefox users.** Total distribution is **100%**, **50%** per group.


media/js/firefox/social-proof-experiment.js
```diff
@@ -0,0 +1,42 @@
...
+                    'v=a': 50,
+                    'v=b': 50
+                }
+            });
+
+            cop.init();
+        }
+    };
+
+    if (client.isDesktop && !client.isFirefoxDesktop) {
```
alexgibson, line 38: This is throwing an error `TypeError: can't access property "isDesktop", client is undefined` because Traffic Cop is running in the `<head>` of the page (before `Mozilla.Client` is included). Rather than bundle the whole of `Mozilla.Client` here also, it might be worth just taking the small bits of code you need to do the logic and apply them directly here instead?

amychurchwell: Thanks, Alex! I moved the logic into the file & things should work as expected now. 👍

hoosteeno:
@sghoseWI the experiment plan calls for 2 variations. @amychurchwell , if I'm reading this right, there is only 1:

https://github.com/mozilla/bedrock/pull/8743/files#diff-0b513c98d40c021a905d72c0c95a6b10R31-R32

I think we should try 2. I initially hoped we would try several, but at this point with a plan already written, I'd be happy to try 2. 

Amy, if I give you new copy can we add one more?

amychurchwell:
@hoosteeno I believe the current test plan only calls for one variation:
> **Traffic Distribution % per Group:** 50%
> **Total Distribution % (among 2 groups, including control):** 100%

Which I read as control gets 50% and variation gets 50%. Will confirm with Saptarshi in the AM(PST), but I'd be more than happy to run a multi-variation experiment either now or if this initial test proves promising.

hoosteeno:
Came in here late just to say, "yep let's merge". :shipit:

PR #8751, opened 2020-04-06
Update Fluent python view utils with better dev UX

pmac:
This updates the `ftl`, `ftl_lazy`, and `has_messages` functions that are for use in Python code to not require fluent files to be defined if using the default files, to allow passing a single fluent file as a string instead of a list, and to require fluent files to be passed in via the `ftl_files` keyword argument.


pmac:
@alexgibson okay. docs are updated and tests are pretty good. r?

PR #8754, opened 2020-04-07
Small debugging improvement for release notes update

duallain:
## Description
We've had a few instances in dev of bedrock not being updated with release notes.
Adds a success message in the run-upload-db.py file, and adding newlines to both skips and successes.
Non-quiet release notes jobs will output the current hashes, and new hash, which should help debugging.

## Testing
* clock container logs should be a little easier to read
* update_release_notes_job should have an additional log line (container script may not show this, as it has the quiet flag passed).

### DIFF
--- a/bin/run-db-upload.py
+++ b/bin/run-db-upload.py
@@ -85,10 +85,10 @@ def main(args):
     prev_data = get_prev_db_data()
     new_data = get_db_data()
     if not force and prev_data and prev_data['checksum'] == new_data['checksum']:
-        print('No update necessary\n')
+        print('No update necessary')
         return 0
 
-    print('Attempting a db update\n')
+    print('Attempting a db update')
 
     set_db_data(new_data)
     if '--no-upload' in args:
--- a/bin/run-db-upload.py
+++ b/bin/run-db-upload.py
@@ -85,8 +85,10 @@ def main(args):
     prev_data = get_prev_db_data()
     new_data = get_db_data()
     if not force and prev_data and prev_data['checksum'] == new_data['checksum']:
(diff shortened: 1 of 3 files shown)


bin/run-db-upload.py
```diff
@@ -85,9 +85,11 @@ def main(args):
     prev_data = get_prev_db_data()
     new_data = get_db_data()
     if not force and prev_data and prev_data['checksum'] == new_data['checksum']:
-        print('No update necessary')
+        print('No update necessary\n')
```
glogiotatidis, line 88: Do you need the extra newline? Print adds one by itself.

pmac: Yeah. Don't need the trailing newline.

duallain: I was thinking the extra new line would help make it a bit easier to parse:

```+ python manage.py update_pocketfeed --quiet
No update necessary
[2020-04-08 17:28:25.748025] Clock job update_upload_database@bedrock-dev-clock-5f947c7699-z4xlt: finished successfully
[2020-04-08 17:32:44.212303] Clock job update_upload_database@bedrock-dev-clock-5f947c7699-z4xlt: starting```

I think having an empty newline in that log would be helpful, between the start/stop messages in the long. Easier to scan through the logs and see the start/stop of each job.  Though it looks like I put it in the wrong place, so will update to be on the finished successfully message.

pmac: I don't think that's a thing. Each line is a line in papertrail and they're all jumbled anyway. I don't see how an empty line is going to help in a log.

Issue #8773, opened 2020-04-13
Experiment: re-run /nav on /new


stephaniehobson:
## Description

The Brand team would very much like to include the global navigation on the download page to:
- keep users arriving on this page who don't want to download on our site
- enforce the master brand narrative that Firefox is more than a browser

We believe it is worth running again because:
- installs were not measured last time
- load times were not compared last time
- changes to /firefox mean that traffic arriving here from search engines may include different segments than before those changes were picked up by search engines

What are we wanting to do? What are we hoping to affect?
- decrease bounce rate
- improve download:install ratio
- no decrease in installs

---

#### Test Plan

https://docs.google.com/document/d/13lxBKBI5pv8WMUnaHWGSweXYJfLIABRyztEZYSYqHqo/edit

#### Tasks
- [ ] Update /new template to show nav with a variable
- [ ] Add a variation which includes the nav HTML and JS but does not show it

stephaniehobson:
This experiment is ready for review: https://app.convert.com/#/projects/1003343/experiments_report/10035273

stephaniehobson:
@alexgibson can you review please ^?

stephaniehobson:
@alexgibson Oh it was the stub attribution Amy and I wanted to ask about. Can you please make sure that's working as intended as part of the review 🙏

alexgibson:
I tested this on Windows and everything seems ok with stub attribution. I see the convert experiment Id and variation passing along automatically 👍 

I do wonder if the targeting URL is incorrect however. It's currently set to `https://www.allizom.org/en-US/exp/firefox/new/?v=a`, so how is traffic going to get to `?v=a`? Seems like this query parameter might be redundant?

Once I do hit the target URL, I get redirected to a variation correctly. Clicking the download button registers as a goal as expected.

One other thing I did notice was a new CSP error in the console:

```
convert.eb14ea70d301.js:1 Refused to connect to 'https://1003343.metrics.convertexperiments.com/log-rep/' because it violates the following Content Security Policy directive: "connect-src 'self' *.mozilla.net *.mozilla.org *.mozilla.com *.allizom.org www.googletagmanager.com www.google-analytics.com logs.convertexperiments.com https://accounts.stage.mozaws.net/ https://accounts.firefox.com.cn/".
```

This doesn't seem to impede the experiment as far as I can tell, but I'll open a pull request to add an exception for it nonetheless 👍

stephaniehobson:
[...]

PR #8774, opened 2020-04-14
Add view to exp.firefox.new to enable variations #8773

stephaniehobson:
## Description

- changes nav to display conditionally in default /exp/firefox/new template
- adds view to pass variants and experiences to /exp/firefox/new template
- add test for new view

## Issue / Bugzilla link

#8773 

## Testing

No nav:
http://localhost:8000/en-US/exp/firefox/new/
http://localhost:8000/en-US/exp/firefox/new/?xv=a

Nav:
http://localhost:8000/en-US/exp/firefox/new/?xv=nv
→ I gave this an obscure name to avoid having what's in the URL bar influence the experiment results


bedrock/exp/urls.py
```diff
@@ -15,8 +15,7 @@
 urlpatterns = (
     page('opt-out', 'exp/opt-out.html'),
     page('firefox', 'exp/firefox/index.html', active_locales=['en-US', 'en-GB', 'en-CA', 'de']),
-    page('firefox/new', 'exp/firefox/new/download.html', active_locales=['en-US', 'en-GB', 'en-CA', 'de']),
-    page('firefox/new/nav', 'exp/firefox/new/download-nav.html'),
```
alexgibson, line 19: Should we redirect this old URL instead of let it 404?

stephaniehobson: Feels like unnecessary tech debt to me. It's not indexed and I don't know why anyone would have it bookmarked. No one has visited the page since the experiment ended.

bedrock/exp/tests/test_views.py
```diff
@@ -0,0 +1,24 @@
...
+from bedrock.exp import views
+from bedrock.mozorg.tests import TestCase
+
+
+@override_settings(DEV=False)
+@patch('bedrock.exp.views.l10n_utils.render')
+class TestExpFirefoxNew(TestCase):
+    def test_download_template(self, render_mock):
+        req = RequestFactory().get('/exp/firefox/new/')
+        req.locale = 'en-US'
```
stephaniehobson, line 20: @pmac Can you help me out with the test failure here please? flak8 is complaining this variable is defined but not used. But if I'm reading it right we're adding a property to a variable that *is* used on the next line. Am I reading that right?

stephaniehobson: Oh I'm looking at the wrong file nevermind >_<

bedrock/exp/urls.py
alexgibson: I see a small amount of daily traffic to this URL in GA, but it's pretty small. Your call 👍

stephaniehobson: Oh I had my search parameters wrong when I did my search.

But this is a real problem. We need to figure out how people are getting here.

stephaniehobson: I'll redirect these people to the real /new page.

But I also want to fix the root of the problem.

stephaniehobson:
Tests fixed, redirect added.

bedrock/exp/redirects.py
```diff
@@ -0,0 +1,7 @@
+from bedrock.redirects.util import redirect
+
+redirectpatterns = (
+
+    redirect(r'^exp/firefox/new/nav?$', 'firefox.new'),
```
alexgibson, line 5: The ending `v` on `nav` is optional here, should this be a trailing slahs instead? Having a test for this redirect would be good too, but not a blocker.

```suggestion
    redirect(r'^exp/firefox/new/nav/?$', 'firefox.new'),
```

stephaniehobson:
Committed the suggestion, hopefully GTG.

Issue #8636, opened 2020-03-09
Error when viewing pages locally with dev=True, debug=False


alexgibson:
## Description

When running bedrock locally with `dev=True` and `debug=False`, pages throw an error.

### Steps to reproduce

1. In your `.env`, set `dev=True`, `debug=False` and start bedrock running.
2. Try loading any page e.g. http://localhost:8000/en-US/

### Expected result

The page should load without error.

### Actual result

The page displays the error message:

```
A server error occurred.  Please contact the administrator.
```

In the terminal session I see:

```
[0] During handling of the above exception, another exception occurred:
[0] 
[0] Traceback (most recent call last):
[0]   File "/usr/local/Cellar/python/3.7.4_1/Frameworks/Python.framework/Versions/3.7/lib/python3.7/wsgiref/handlers.py", line 137, in run
[0]     self.result = application(self.environ, self.start_response)
[0]   File "/Users/agibson/Documents/GitHub/bedrock/venv/lib/python3.7/site-packages/django/core/handlers/wsgi.py", line 141, in __call__
[0]     response = self.get_response(request)
[0]   File "/Users/agibson/Documents/GitHub/bedrock/venv/lib/python3.7/site-packages/django/core/handlers/base.py", line 75, in get_response
[0]     response = self._middleware_chain(request)
[0]   File "/Users/agibson/Documents/GitHub/bedrock/venv/lib/python3.7/site-packages/django/core/handlers/exception.py", line 36, in inner
[0]     response = response_for_exception(request, exc)
[0]   File "/Users/agibson/Documents/GitHub/bedrock/venv/lib/python3.7/site-packages/django/core/handlers/exception.py", line 90, in response_for_exception
[0]     response = handle_uncaught_exception(request, get_resolver(get_urlconf()), sys.exc_info())
[0]   File "/Users/agibson/Documents/GitHub/bedrock/venv/lib/python3.7/site-packages/django/core/handlers/exception.py", line 129, in handle_uncaught_exception
[0]     return callback(request, **param_dict)
[0]   File "/Users/agibson/Documents/GitHub/bedrock/bedrock/base/views.py", line 171, in server_error_view
[0]     return l10n_utils.render(request, template_name, status=500)
[0]   File "/Users/agibson/Documents/GitHub/bedrock/lib/l10n_utils/__init__.py", line 120, in render
[0]     return django_render(request, template, context, **kwargs)
[...]

PR #8777, opened 2020-04-14
Make running locally in prod mode easier

pmac:
This will give us some commands to more easily build and run the prod Docker image locally. It will run using your local version of the Django app, but not the media, which must be built separately.

## Using

If you want to run things locally as if they're prod and see changes in templates or Python files you can simply:

```
make pull run-prod
```

That will pull the latest prod image and run it in prod mode locally. If you need to build new static assets (images, CSS, JS, etc.) then you can do:

```
make build-prod run-prod
```

Let me know what you think.

Re #8636

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,9 @@ build: .docker-build-pull
 	${DC} build --pull app assets
 	touch .docker-build
 
+build-prod: .docker-build-pull
+	${DC} build --pull release
+
 pull: .env submodules
 	-GIT_COMMIT= ${DC} pull release app assets builder app-base
 	touch .docker-build-pull
@@ -48,6 +51,9 @@ rebuild: clean build
 
 run: .docker-build-pull
 	${DC} up assets app
+
+run-prod: .docker-build-pull
+	${DC} up release
 
 stop:
 	${DC} stop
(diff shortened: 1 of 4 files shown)


pmac:
This obviously still needs docs, but I wanted you to try it first before we document it and merge this. It may well not be that useful since you have to build the image every time you want to view it with changed CSS...

alexgibson:
Thanks for this @pmac!

I ran both of the above commands and they worked OK 👍 

One thing I didn't expect was still needing to set `Dev=False` and `Debug=False` in my `.env`. If we still need to do this manually then I think that's fine, but let's document it and explain what the commands do / don't handle.

I think this will be very useful for thing like checking locale activation (template fallbacks), as well as debugging errors that don't happen when `Dev=True`.

pmac:
Good point. It does still use your `.env` file. I could have it use a hard coded `prod.env` file instead but I'm guessing that setting env vars will be a thing you will want to do for this at some point. I'm honestly not sure which way is best to go for this. I'm leaning toward leaving it as-is and documenting that you have to update your `.env` file with what you want. Thoughts?

alexgibson:
> Thoughts?

I can picture someone wanting to set `Dev=False` but also `Debug=True` for some error tracing, so I think leaving things as-is and documenting it may be fine?

pmac:
Added docs. Ready for review.

PR #8779, opened 2020-04-15
Migrate common UI strings from main.lang to Fluent (Fixes #8657)

### LINKED ISSUE #8657: Migrate main.lang to Fluent
## Description
- https://github.com/mozilla-l10n/www.mozilla.org/blob/master/en-US/main.lang

Note: main.lang was a catch all for common strings for many years. In retrospect, allowing it to build up this was probably wasn't the ideal situation. When we migrate this to Fluent, it may be worth exploring:

- Culling any old strings that are no longer in use globally.
- Splitting out any other common groups of strings into separate, more distinct .ftl files (such as fundraising banner, firefox accounts form etc)

---

- [ ] Existing .lang strings are migrated into a suitable .ftl alternative.

alexgibson:
## Description
- Migrates common UI string from `main.lang` to `ui.ftl`.
- Updates templates to use new Fluent IDs.

Note to reviewer: I've been somewhat selective here in what we migrate, by trying to stick to generic UI strings that are common and/or heavily used. There are _a lot_ of old strings in `main.lang`, and we want to try and move away from a creating files of global, unrelated or ungrouped strings in Fluent. Pontoon will now suggest similar / matching strings to localizers from other files, so global shared strings like aren't needed as they used to be.

This PR is the last of what I propose we migrate from `main.lang`. Templates that still rely on strings from there can continue to use them, until we come to port those pages to Fluent. At that stage they can be migrated to other files.

That said, I'm open to adding any further suggestions that may still be migrated [from here](https://github.com/mozilla-l10n/www.mozilla.org/blob/master/en-US/main.lang).

## Issue / Bugzilla link
#8657

## Testing
- [ ] Are there any other common UI related strings worth importing?
- [ ] Did I miss any references to the strings migrated here that still need updating?

Successful test run: https://gitlab.com/mozmeao/www-config/pipelines/136185630

### DIFF
--- a/bedrock/base/templates/base-pebbles.html
+++ b/bedrock/base/templates/base-pebbles.html
@@ -81,12 +81,9 @@
 
   <body {% if self.body_id() %}id="{% block body_id %}{% endblock %}" {% endif %}class="html-{{ DIR }} {% block body_class %}{% endblock %}" {% block body_attrs %}{% endblock %}>
     <div id="strings"
-      data-global-close="{{ _('Close') }}"
-      data-global-next="{{ _('Next') }}"
-      data-global-previous="{{ _('Previous') }}"
-      data-global-fx-out-of-date-banner-heading="{{ _('Your Firefox is out-of-date.') }}"
-      data-global-fx-out-of-date-banner-message="{{ _('Get the most recent version to keep browsing securely.') }}"
-      data-global-fx-out-of-date-banner-confirm="{{ _('Update Firefox') }}"
+      data-global-close="{{ ftl('ui-close') }}"
+      data-global-next="{{ ftl('ui-next') }}"
+      data-global-previous="{{ ftl('ui-previous') }}"
       {% block string_data %}{% endblock %}></div>
 
     {% block page_banner %}{% endblock %}
--- a/bedrock/mozorg/templates/mozorg/manifesto-base.html
+++ b/bedrock/mozorg/templates/mozorg/manifesto-base.html
@@ -28,13 +28,13 @@
   <div class="side-reference">
     <h4 class="side-reference-title">{{ _('Our Products') }}</h4>
     <p>{{ _('Software and other innovations designed to advance our mission.') }}</p>
-    <a class="more" href="{{ url('firefox') }}">{{ _('Learn More') }}</a>
+    <a class="more" href="{{ url('firefox') }}">{{ ftl('ui-learn-more') }}</a>
   </div>
 
   <div class="side-reference">
     <h4 class="side-reference-title">{{ _('Get Involved') }}</h4>
     <p>{{ _('Become a volunteer contributor in a number of different areas.') }}</p>
-    <a class="more" href="{{ url('mozorg.contribute.index') }}">{{ _('Learn More') }}</a>
+    <a class="more" href="{{ url('mozorg.contribute.index') }}">{{ ftl('ui-learn-more') }}</a>
   </div>
 
   {% block sidebar_extra %}{% endblock %}
--- a/bedrock/mozorg/templatetags/misc.py
+++ b/bedrock/mozorg/templatetags/misc.py
@@ -318,7 +318,8 @@ def video(ctx, *args, **kwargs):
         'h': 360,
         'autoplay': False,
         'preload': False,
-        'id': 'htmlPlayer'
+        'id': 'htmlPlayer',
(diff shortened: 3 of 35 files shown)


bedrock/base/templates/base-pebbles.html
```diff
@@ -81,12 +81,9 @@
 
   <body {% if self.body_id() %}id="{% block body_id %}{% endblock %}" {% endif %}class="html-{{ DIR }} {% block body_class %}{% endblock %}" {% block body_attrs %}{% endblock %}>
     <div id="strings"
-      data-global-close="{{ _('Close') }}"
-      data-global-next="{{ _('Next') }}"
-      data-global-previous="{{ _('Previous') }}"
-      data-global-fx-out-of-date-banner-heading="{{ _('Your Firefox is out-of-date.') }}"
-      data-global-fx-out-of-date-banner-message="{{ _('Get the most recent version to keep browsing securely.') }}"
-      data-global-fx-out-of-date-banner-confirm="{{ _('Update Firefox') }}"
```
alexgibson, line 89: Note: these strings are no longer used anywhere, so I'm removing them.

alexgibson:
@pmac r? I think this is ready for review, but let's hold off merging until the L10n PR has also been reviewed.

bedrock/mozorg/templates/mozorg/manifesto-base.html
```diff
@@ -28,13 +28,13 @@
   <div class="side-reference">
     <h4 class="side-reference-title">{{ _('Our Products') }}</h4>
     <p>{{ _('Software and other innovations designed to advance our mission.') }}</p>
-    <a class="more" href="{{ url('firefox') }}">{{ _('Learn More') }}</a>
```
pmac, line 31: I noticed that this is technically a different string than some other ones since "More" is capitalized. Hopefully it's fine to change that, but I did want to note the change.

bedrock/mozorg/templatetags/misc.py
```diff
@@ -318,7 +318,8 @@ def video(ctx, *args, **kwargs):
         'h': 360,
         'autoplay': False,
         'preload': False,
-        'id': 'htmlPlayer'
+        'id': 'htmlPlayer',
+        'fluent_l10n': ctx['fluent_l10n']
```
pmac, line 322: NIT: always best to add a trailing comma.

bedrock/mozorg/tests/test_helper_misc.py
```diff
@@ -12,6 +12,7 @@
 from jinja2 import Markup
 from mock import patch
 from pyquery import PyQuery as pq
+from lib.l10n_utils.fluent import fluent_l10n
```
pmac, line 15: NIT: import ordering rules would put this down in the group with the other bedrock imports since it's in the project.

bedrock/mozorg/templates/mozorg/manifesto-base.html
alexgibson: Yeah - we currently have two different strings in `main.lang`, one with `more` and another with `More`. Rather than keeping two, I decided to opt for the most prevalent occurence which is the former. I think this likely makes more sense than migrating two different occurences, as many locales seem to ignore the capitalization. Maybe this is something we could have a preference with the the `en` file only, which we can always change later on?

pmac: could probably change it with CSS for en-* locales too. But it does bring up the question of whether there are locales that translated the two differently, and which they'd prefer in the migration. Maybe we need some L10n team input here?

peiying2: Localized strings from the main.lang, and only a handful are different. 
- [Learn more](https://transvision.mozfr.org/string/?entity=mozilla_org/main.lang:d59048f21fd887ad520398ce677be586&repo=mozilla_org).
- [Learn More](https://transvision.mozfr.org/string/?entity=mozilla_org/main.lang:f5395c9793af8a11b406ca7c1ac70da9&repo=mozilla_org).

Let's keep the one that agrees with the current style.

alexgibson: Peiying says there's not much difference between the two string in terms of translations, so I think we're ok to keep one and not the other. Good call asking to verify 👍

Issue #7723, opened 2019-09-12
Automate l10n activities


hoosteeno:
### from bedrock

- [x] Lint fluent files for validity post-commit on Bedrock: fluent is valid, and fluent files included in templates actually exist (if not, CI fails)
- [x] Copy new fluent files in bedrock over to l10n team's fluent repositories in a branch
- [x] Config files -- which are authoritative in Bedrock's l10n repo -- are copied to l10n team's repository at this time
- [x] Open a PR with these changes against the https://github.com/mozilla-l10n/www-l10n repo

---

### to bedrock

- [x] Retrieve fluent files from l10n team's fluent repositories on some schedule or as a result of some event
- [x] Lint fluent files when retrieved from fluent repositiories: fluent is valid (if not, don't import)
- [x] Evaluate every fluent file imported for activation status & manage activation lists
- [x] Ensure a developer can override the site-wide default activation % for a particular fluent file
- [x] Push into [mozmeao/www-l10n](https://github.com/mozmeao/www-l10n) repository

pmac:
@Pike for the "from bedrock" part, we want to push a branch to the mozilla-l10n repo and open a PR during that automation right?

pmac:
Much of the "to bedrock" part was done in #8523. #7705 indicates what's left to do.

Pike:
> @Pike for the "from bedrock" part, we want to push a branch to the mozilla-l10n repo and open a PR during that automation right?

Yep.

pmac:
>  and fluent files included in templates actually exist

I'm not sure this part is necessary. It would be pretty obvious if the ftl file wasn't right during development and review since no strings would show up on the page and would only have the string IDs in the HTML. We could look at including this in tests, but it seems like overkill to me, at least for now.

pmac:
I think we're done here! Please reopen if we discover we need something more.

PR #8786, opened 2020-04-17
Add automation to submit a PR for Fluent file changes

pmac:
## Description

This adds the `manage.py open_ftl_pr` command. It requires authentication tokens for Github to fully work. It includes changes to make the automation run for every merge to master that includes changes to files in the `l10n` directory of the repo.

## Issue / Bugzilla link

Re #7723

## Testing

You can run this locally and it will fetch updates from the l10n team repo, copy any changed files from `l10n/` into it, create a new branch and commit those changes to it. It will then fail since it has no credentials for Github. I tested locally by forking the l10n team's repo and using a personal token for creating PRs in my fork.

e.g. https://github.com/pmac/www-l10n/pull/2


pmac:
@Pike for this to work once it merges the @MozmarRobot account will need write access to the https://github.com/mozilla-l10n/www-l10n repo.

Pike:
Invite sent.

Pike:
CCing @flodolo so that he knows where this is coming from.

I contemplated adding MozmarRobot to the tool-bots, but figured better safe than sorry, and only added it to the one repo.

pmac:
Invitation accepted. Thanks!

pmac:
> For new content when we commit to master, the command will be run and any changes to the `/l10n/` directory will manifest as a pull request against the L10n team repo. This would satisfy the creation of new content, and updates to existing content (is that correct?)

Correct

> For the migration of legacy content, what would the workflow now be? I'm guessing we'd need to open an accompanying PR for just the localized strings, that would then need to be reviewed alongside the automated PR for en?

So far yes, that is right. This automation doesn't do anything for doing the migration and opening a PR for that. We should discuss how we might accomplish that.

lib/l10n_utils/management/commands/open_ftl_pr.py
```diff
@@ -0,0 +1,105 @@
...
+from subprocess import CalledProcessError
+
+from django.core.management.base import CommandError
+from django.conf import settings
+from django.template.defaultfilters import slugify
+from django.utils.functional import cached_property
+
+from bedrock.utils import github
+
+from dirsync import sync
```
glogiotatidis, line 16: :+1:

bin/open-ftl-pr.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+set -exo pipefail
+
+DOCKER_IMAGE=mozorg/bedrock:latest
```
glogiotatidis, line 4: nit: Use `CI_COMMIT_SHORT_SHA` to ensure that you use the image that you just built?

lib/l10n_utils/management/commands/open_ftl_pr.py
```diff
@@ -0,0 +1,105 @@
...
+            return False
+
+        self.stdout.write('Committed changes to local repo')
+        return True
+
+    def push_changes(self):
+        try:
+            self.l10n_repo.git('push', self.git_push_url, 'HEAD')
+        except CalledProcessError:
+            raise CommandError(f'There was a problem pushing to {self.l10n_repo.remote_url}')
```
glogiotatidis, line 80: If we reach this point, next run should not be affected b/c it will be a new branch, correct?

bin/open-ftl-pr.sh
pmac: I thought about that, but to keep things fast I just went with this. It's not guaranteed that this will run on the same runner that built this image. Also it's not really important that this be run by the same exact version of the site. Maybe it would help with some issues though? I'm not sure. Hmmm...

pmac: Yeah. But even if we rerun this was run only inside the docker container which was deleted and will start fresh again.

pmac: If I did do that then I wouldn't have to pass in `CI_COMMIT_SHORT_SHA` in the `docker run` command either since the image would have it built in via the `GIT_SHA` var.

pmac: Okay @glogiotatidis. Great call. I've updated things. I sourced the `set_git_env_vars.sh` script and used the `$DEPLOYMENT_DOCKER_IMAGE` variable that will be the image name of the newly pushed image which includes the branch name.

pmac:
@alexgibson I think this is good to go. As soon as we're ready for these PRs to start being automatically created we can merge this. We should talk more about the content migration workflow, but for now this should help I think.

pmac:
e.g. should we document a process for finding the PR that this automation opens, and pushing the migrated fluent files to that branch so that it can all be considered together, or should that go in a separate PR?

alexgibson:
I’m not sure I entirely follow your line of thinking. But here are my thoughts:

- I don’t see why we couldn’t merge this PR to master now, if it handles automation for new fluent content being added to the repo.
-  What I don’t yet understand is how this automated workflow would work along side PRs that also migrate existing content. I think it would be good to add documentation here for how these workflows should happen once this PR merges.

alexgibson:
Oh wait, I think I understand what you’re envisaging, and that could work yes. Hmmm

alexgibson:
Ok, I think this option for skipping the automated L10n process is a good idea, and something we can certainly iterate on if we need. I think this is good to merge now, and we can test out the process and document it in a followup PR. Thanks @pmac! 💯

PR #8790, opened 2020-04-17
Add b and c variants of exp/firefox/new page (#8773)

stephaniehobson:
## Description

Add b and c variants of exp/firefox/new page. Both have nav menu but b does not display it. A is control.

## Issue / Bugzilla link

#8773

## Testing

-http://localhost:8000/en-US/exp/firefox/new/?v=a
-http://localhost:8000/en-US/exp/firefox/new/?v=b
-http://localhost:8000/en-US/exp/firefox/new/?v=c

### DIFF
--- a/bedrock/exp/templates/exp/firefox/new/download.html
+++ b/bedrock/exp/templates/exp/firefox/new/download.html
@@ -27,9 +27,18 @@
 {#- This will appear as <meta property="og:description"> which can be used for social share -#}
 {% block page_og_desc %}Faster page loading, less memory usage and packed with features, the new Firefox is here.{% endblock %}
 
+{% block page_css %}
+  {% if v == 'b' %}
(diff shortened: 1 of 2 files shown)


bedrock/exp/templates/exp/firefox/new/download.html
```diff
@@ -27,9 +27,18 @@
 {#- This will appear as <meta property="og:description"> which can be used for social share -#}
 {% block page_og_desc %}Faster page loading, less memory usage and packed with features, the new Firefox is here.{% endblock %}
 
+{% block page_css %}
+  {% if v == 'b' %}
+  <style>
```
alexgibson, line 32: We should really avoid inline CSS because CSP, but I don't think we're blocking it and this is a temporary experiment, so 👍

PR #8794, opened 2020-04-18
Add variant of /thanks for chrome post-download experiment.

amychurchwell:
## Description
Adds a variant of /thanks for an experiment. Traffic cop not needed because the Convert tool will be used to split traffic. 

Will target en-US, desktop, nonFX (Chrome)

## Issue / Bugzilla link
[CRO team airtable card](https://airtable.com/tblO5zD4uNk7ISyG0/viwIUJsHKDrOFl90B/recUZVMgB8JzM3Zml)


bedrock/firefox/templates/firefox/new/trailhead/thanks-b.html
```diff
@@ -0,0 +1,107 @@
...
+
+  <section class="mzp-l-content">
+    <ol class="l-columns-two">
+      <li class="c-picto-block">
+        <h3 class="c-picto-block-title">{{ _('Click on the Firefox installer.') }}</h3>
+        <div class="c-picto-block-image">
+          <img src="{{ static('img/firefox/new/trailhead/browser-status-bar.png') }}" alt="">
+        </div>
+        <div class="c-picto-block-body">
+          <p><strong>{{ _('Look down, look left.') }}</strong> {{ _('Click on the Firefox installer in the lower bottom
```
alexgibson, line 70: Not something we need worry about for the experiment, but these instructions might not be accurate for RTL locales.

bedrock/firefox/templates/firefox/new/trailhead/thanks-b.html
```diff
@@ -0,0 +1,107 @@
...
+
+      <li class="c-picto-block">
+        <h3 class="c-picto-block-title">{{ _('Select ‘Yes.’') }}</h3>
+        <div class="c-picto-block-image">
+          <img src="{{ static('img/firefox/new/trailhead/browser-modal.png') }}" alt="">
+        </div>
+        <div class="c-picto-block-body">
+          <p>
+            {% trans %}
+            Click YES for Firefox. <strong>Select YES</strong> on the pop-up window to finish the install.
```
alexgibson, line 83: This is true for Windows 10, but do we need to worry about other Windows versions in this experiment?

I also wonder if this is going to pose a challenge for L10n, since translators may not have access to a Windows version in their language to test?

bedrock/firefox/templates/firefox/new/trailhead/thanks-b.html
```diff
@@ -0,0 +1,107 @@
...
+        <div class="c-picto-block-body">
+          <p><strong>{{ _('Look down, look left.') }}</strong> {{ _('Click on the Firefox installer in the lower bottom
+            left-hand corner.') }}</p>
+        </div>
+      </li>
+
+      <li class="c-picto-block">
+        <h3 class="c-picto-block-title">{{ _('Select ‘Yes.’') }}</h3>
+        <div class="c-picto-block-image">
+          <img src="{{ static('img/firefox/new/trailhead/browser-modal.png') }}" alt="">
```
alexgibson, line 78: English copy in the image here. Maybe something we could solve with CSS later on, if this experiment is successful?

bedrock/firefox/templates/firefox/new/trailhead/thanks-b.html
```diff
@@ -0,0 +1,107 @@
...
+            {% endtrans %}
+          </p>
+        </div>
+      </li>
+    </ol>
+
+    <aside class="sumo-section">
+      <h3>{{ _('Complete your install and get the automatically private and wildly fast Firefox.') }}</h3>
+      <p>
+        {% trans sumo='' %}
```
alexgibson, line 93: Missing URL here?

media/css/firefox/new/trailhead/thanks.scss
```diff
@@ -11,6 +11,76 @@ $image-path: '/media/protocol/img';
 @import '../../../../protocol/css/components/picto-card';
 @import '../../../../protocol/css/templates/card-layout';
 
+// Chrome post-download instructions experiment.
```
alexgibson, line 14: Should we split this CSS out into its own file? The selectors added here are quite broad, but whilst they don't interfere with the regular page they are not needed for the non-experimental version?

bedrock/firefox/templates/firefox/new/trailhead/thanks-b.html
amychurchwell: whoops. thank you. :)

bedrock/firefox/templates/firefox/new/trailhead/thanks-b.html
amychurchwell: I think that is a good idea. We could potentially absolutely position a string on the image.

media/css/firefox/new/trailhead/thanks.scss
amychurchwell: While the non-experimental version doesn't need the styles, the new variant does. I separated the styles into it's own file and added both to it's bundle. :+1:

bedrock/firefox/templates/firefox/new/trailhead/thanks-b.html
amychurchwell: > This is true for Windows 10, but do we need to worry about other Windows versions in this experiment?

We are going to target Windows 10 specifically. I am going to work on creating the audience in convert based on user agent. :+1:

> I also wonder if this is going to pose a challenge for L10n, since translators may not have access to a Windows version in their language to test?

I am not sure what the ultimate localization plan for this one would end up being, but you are right it would need a larger conversation.

bedrock/firefox/templates/firefox/new/trailhead/thanks-b.html
```diff
@@ -0,0 +1,107 @@
...
+            {% endtrans %}
+          </p>
+        </div>
+      </li>
+    </ol>
+
+    <aside class="sumo-section">
+      <h3>{{ _('Complete your install and get the automatically private and wildly fast Firefox.') }}</h3>
+      <p>
+        {% trans sumo='https://support.mozilla.org/bm/kb/how-download-and-install-firefox-windows' %}
```
alexgibson, line 93: Should this URL remove `/bm/`?
```suggestion
        {% trans sumo='https://support.mozilla.org/kb/how-download-and-install-firefox-windows' %}
```

bedrock/firefox/templates/firefox/new/trailhead/thanks-b.html
```diff
@@ -0,0 +1,107 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros-protocol.html" import call_out, picto_card with context %}
```
alexgibson, line 5: I don't think `picto_card` is needed here for this version of the page.

Issue #8675, opened 2020-03-18
Sticky Navigation: Code


hoosteeno:
The epic at https://github.com/mozilla/bedrock/issues/8676 talks about implementing sticky navigation to replace global navigation bars on all pages that have global navigation. This card is about the coding.

Success Criteria:
- [ ] Sticky navigation code is live in production

Tasks:
- [x] Build it in Protocol

alexgibson:
Sticky nav is now ready for code review over in Protocol: https://github.com/mozilla/protocol/pull/561

alexgibson:
Adding a blocker to this until it's in Protocol.

alexgibson:
https://github.com/mozilla/protocol/pull/561 has been merged, so the next step is now to publish a new NPM release and then update bedrock to use it.

hoosteeno:
@craigcook is publishing a new NPM release something you can do?

craigcook:
> @craigcook is publishing a new NPM release something you can do?

Yup, I'm on it.

hoosteeno:
This is dependent on https://github.com/mozilla/bedrock/pull/8796

PR #8791, opened 2020-04-17
Graduate installer-help redesign experiment into prod.

amychurchwell:
## Description
Graduates the installer-help redesign experiment! 🎉  ([Original PR](https://github.com/mozilla/bedrock/pull/8731))

**Experiment Analysis Report:** https://docs.google.com/document/d/1RlsO0zjuZT2r717TOdfv8PpOGBfvlVHu3SO5-cpk5F4/edit# 

## Issue
[Airtable card](https://airtable.com/tblO5zD4uNk7ISyG0/viwIUJsHKDrOFl90B/reclbibhH3nIMzdyp)

PR #8796, opened 2020-04-20
Update Protocol to 11.0.2

craigcook:
## Description
Updates Protocol to [v11.0.2](https://github.com/mozilla/protocol/blob/master/CHANGELOG.md#1102)

This update includes the sticky nav but I haven't actually changed the nav in bedrock, we're only gaining the ability. Unblocks #8675 

## Testing
This ended up touching almost every css file because it renames the `text-display-*` mixins to `text-title-*` and I should have caught them all, but the old `text-display-*` ones are aliased so nothing should break even if I missed some. But it's worth giving it a fairly careful look to make sure I didn't get overzealous in my find/replace.

It also updates base form label styling and now declares a text color, so be on the lookout for forms on dark backgrounds where the labels might no longer inherit their color. I only found a few but I could have missed some.

There are minor updates to notification bars and the emphasis box but they should be superficial.

### DIFF
--- a/media/css/base/banners/fundraiser.scss
+++ b/media/css/base/banners/fundraiser.scss
@@ -51,6 +51,7 @@ $image-path: '/media/protocol/img';
     margin-bottom: $spacing-lg;
 
     label {
+        color: $color-white;
         display: inline-block;
 
         &:first-child {
--- a/media/css/base/banners/includes/_base.scss
+++ b/media/css/base/banners/includes/_base.scss
@@ -39,7 +39,7 @@ $image-path: '/media/protocol/img';
 
 .c-banner-title {
     @include font-mozilla;
-    @include text-display-lg;
+    @include text-title-lg;
 
     @media #{$mq-lg} {
         @include bidi((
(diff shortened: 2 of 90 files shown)


alexgibson:
Form label color issue on the default /whatsnew page (http://localhost:8000/en-US/firefox/whatsnew/all/)

image

alexgibson:
Not a blocker, but the new form label sizes create some minor alignment issues with the (?) buttons on /firefox/all/. I'm not aware of any other page on the site where this might also be effected:

image

alexgibson:
Notification bar on /firefox/download/thanks/ is misaligned:

image

alexgibson:
Can this override be removed now? https://github.com/mozilla/bedrock/blob/master/media/css/protocol/protocol.scss#L16

alexgibson:
Looks like this needs a rebase, and I can still see some misaligned notifications on some pages. Would it be owrth adding a global override to fix this in bedrock until https://github.com/mozilla/protocol/pull/570 is published?

http://localhost:8000/en-US/newsletter/updated/

image

alexgibson:
Oh, also these two PRs merged that add new pages / CSS. You may want to check if either of these need updating:

- https://github.com/mozilla/bedrock/pull/8794
- https://github.com/mozilla/bedrock/pull/8791

PR #8816, opened 2020-04-21
Add pages for Privacy by Default messaging experiment.

ericawright:
## Description
@craigcook, please review :) 

We will be adding 3 pages for rapid privacy messaging experimentation, 2 privacy messages, and one unrelated message to be used as a control. We will use a heartbeat message to encourage users to visit these pages. 
This is to give you a better idea of our intentions. Please let me know if I should move the files elsewhere. It needs to be hosted on mozilla.org because we are linking directly to the protections report, using the UI tour. 

Note: Images are placeholders.


ericawright:
How would I indicate that these pages should not be translated? This will be used for a brief experiment in en-US only.

bedrock/firefox/templates/firefox/messaging-experiment/privacy_by_default.html
```diff
@@ -0,0 +1,69 @@
...
+      title=_('Firefox: Private By Default'),
+      desc=_("Privacy is more than a mission. It's built right into our products. Our default privacy and security settings block malware, stop trackers, and keep companies from following you around the web."),
+      class='mzp-t-firefox mzp-t-dark',
+      include_cta=True,
+      heading_level=1
+    ) %}
+  
+    <p class="primary-cta">
+      <div>
+        <a class="mzp-c-button mzp-t-product js-open-about-protections" data-cta-text="Go to your Protections Report" data-cta-type="button">{{ _('Check out your protections') }}</a>
```
craigcook, line 25: This should probably have an href even though it triggers a JS event. Elsewhere we've linked to SUMO, just in case JS fails or is disabled. [Here's an example from an older whatsnew page.](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx70-en.html#L74)

media/static-bundles.json
```diff
@@ -824,6 +824,13 @@
       ],
       "name": "firefox_welcome_page5"
     },
+    {
+      "files": [
+        "js/base/uitour-lib.js",
+        "js/firefox/privacy/messaging-experiment.js"
+      ],
+      "name": "firefox_welcome_page8"
```
craigcook, line 832: These pages are probably temporary but just in case they live for a while, let's rename this bundle to make it distinct from the welcome pages (since we're likely to have a welcome8 in the near future).

hoosteeno:
I can't jump into the code at this time, but I want to be sure these pages...
* include a noindex metatag
* ARE NOT listed in a robots.txt file (listing them there makes it far more likely that they will be indexed, because when listed there crawlers can see their URLs, but can't load the page to access the noindex tag)
* can gracefully handle the possibility that someone without the right version of Firefox clicks the CTAs
* have www.mozilla.org GA set up so we have analytical record of their pageloads
* have experiment parameters set up in datalayer so we can isolate them in GA

@craigcook can you help ensure these are true?

craigcook:
Turning this into a checklist:

- [ ] include a noindex metatag, and DO NOT include hreflang and canonical tags
- [ ] ARE NOT listed in a robots.txt file (listing them there makes it far more likely that they will be indexed, because when listed there crawlers can see their URLs, but can't load the page to access the noindex tag)
- [ ] can gracefully handle the possibility that someone without the right version of Firefox clicks the CTAs
- [ ] have [www.mozilla.org](http://www.mozilla.org) GA set up so we have analytical record of their pageloads
- [ ] have experiment parameters set up in datalayer so we can isolate them in GA

craigcook:
Pages on demo:

https://www-demo3.allizom.org/en-US/firefox/privacy-by-default/
https://www-demo3.allizom.org/en-US/firefox/privacy-promise/
https://www-demo3.allizom.org/en-US/firefox/mobile-promo/ (this one re-uses /welcome/4/)

media/img/icons/cryptomining-pink.svg
```diff
@@ -0,0 +1 @@
+<svg width="60" height="50" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink"><defs><path d="M46.005 7h-24a6.009 6.009 0 00-4.8 2.4l-3.6 4.8c-1.482 1.98-.072 4.8 2.4 4.8h30V7z" id="a"/></defs><g fill="none" fill-rule="evenodd"><path d="M34 7V5.5A1.5 1.5 0 0032.5 4h-3A1.5 1.5 0 0028 5.5v39a1.5 1.5 0 001.5 1.5h3a1.5 1.5 0 001.5-1.5V7z" fill="#FF8AC5"/><use fill="#E31587" xlink:href="#a"/></g></svg>
```
craigcook, line 1: We usually include both width/height and viewBox in SVGs. I can't remember which browser needs which... I know old IEs didn't support viewBox, and maybe they ALL support width/height so this could be fine as-is. But it still seems like it can't hurt to have both. So here we can add `viewBox="0 0 60 50"` (and update the others as well).

ericawright: Could I just turn around users who are on a different browser? Send them to the homepage or something.

media/js/firefox/privacy/messaging-experiment.js
```diff
@@ -0,0 +1,28 @@
...
+(function() {
+    'use strict';
+
+    var client = Mozilla.Client;
+
+    function handleOpenProtectionReport(e) {
+        e.preventDefault();
+        Mozilla.UITour.showProtectionReport();
+
+        window.dataLayer.push({
```
ericawright, line 14: Also, what does this do?

craigcook: It sends an event to Google Analytics when the button is clicked. If it were some other web page we could follow the traffic and know people came from this page, but since this opens an in-product page we need a special event to count those clicks.

ericawright: okay, I should change the label in that case

media/img/icons/cryptomining-pink.svg
craigcook: Adding viewBox is certainly not a blocker, just a nice-to-have (and I'm pretty sure as long as width and height are there viewBox is optional anyway... it might be different if it was only viewBox and no width/height). I don't think we need to forcibly prevent non-Firefoxes from loading these pages, they're very unlikely to get any odd traffic.

bedrock/firefox/templates/firefox/messaging-experiment/privacy_promise.html
```diff
@@ -0,0 +1,66 @@
...
+      title=_('With Firefox, you’re protected every time you get online'),
+      desc=_('Firefox automatically protects your privacy without sacrificing performance. So go ahead and browse with confidence, knowing Firefox is protecting what you do online.'),
+      class='mzp-t-firefox mzp-t-dark',
+      include_cta=True,
+      heading_level=1
+    ) %}
+  
+    <p class="primary-cta">
+      <div>
+        <a href="{{ url('firefox.privacy.index') }}" class="mzp-c-button mzp-t-product" data-cta-text="Read our privacy promise" data-cta-type="button">{{ _('Read our privacy promise') }}</a>
```
craigcook, line 25: Another place for title case: "Read Our Privacy Promise"

bedrock/firefox/templates/firefox/messaging-experiment/privacy_by_default.html
```diff
@@ -0,0 +1,74 @@
...
+        <h3 class="c-picto-block-title">Cryptominers stopped in their tracks</h3>
+        <div class="c-picto-block-body">
+          <p>{{ _('Cryptominers are a type of malware that use your system’s computing power to mine digital money. Cryptominers are blocked by default with Firefox’s Standard settings.') }}</p>
+        </div>
+      </div>
+    </div>
+  {% endblock %}
+  
+  {% block js %}
+    {{ js_bundle('default_privacy_messaging') }}
```
craigcook, line 72: Since we're no longer linking to about:protections we should be able to delete this bundle.

ericawright:
need to revert that last commit - turns out we do still want to link to about:protections. (and some legal translations snuck in???)

PR #8819, opened 2020-04-22
Remove duplicate Fluent ID from navigation.ftl

alexgibson:


alexgibson:
The L10n migration to accompany this PR has been reviewed and merged. The commit in bedrock here is a single line code change (removing a duplicate line from the migraton file), so I'm going to merge.

PR #8847, opened 2020-04-27
Lint .ftl files using moz-l10n-lint (Fixes #8846)

### LINKED ISSUE #8846: Lint .ftl files in pull requests
## Description

We have a couple of commands to help lint .ftl files, which will pick out things such as duplicate IDs or other errors.

```
moz-l10n-lint l10n/l10n-pontoon.toml
moz-l10n-lint l10n/l10n-vendor.toml
```

So far I've been running this manually when doing migrations, but it's easy to forget. We should have this running in CI against pull requests, and have checks fail if anything is caught.

alexgibson:
Fixes https://github.com/mozilla/bedrock/issues/8846

### DIFF
--- a/bin/run-tests.sh
+++ b/bin/run-tests.sh
@@ -1,6 +1,8 @@
 #!/bin/bash -xe
 
 flake8
+moz-l10n-lint l10n/l10n-pontoon.toml
+moz-l10n-lint l10n/l10n-vendor.toml
 python manage.py lint_ftl -q
 python manage.py runscript check_calendars
 python manage.py version


bin/run-tests.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash -xe
 
 flake8
+moz-l10n-lint l10n/l10n-pontoon.toml
+moz-l10n-lint l10n/l10n-vendor.toml
 python manage.py lint_ftl -q
```
pmac, line 6: do we still need this command?

alexgibson: I don't know, does `lint_ftl` run against the files pulled from the external repo? If that's the case, I'd say let's keep them? This PR adds linting mainly for the `en` files that live in bedrock.

PR #8855, opened 2020-04-27
Add exp versions of the homepage and mobile (fixes #8795)

### LINKED ISSUE #8795: Clean up and add exp templates for experiments.
## Description

We should remove the exp templates that are not used for running experiments and add the ones we need.

---

#### exp templates needed.
_en-US only for now._
- [x] /firefox
- [x] /firefox/new
- [ ] /firefox/mobile
- [ ] homepage

amychurchwell:
## Description
Adds firefox/mobile and the homepage to the exp app for convert tests.

## Issue / Bugzilla link
#8795

## Testing
[Creating exp pages docs.](https://bedrock.readthedocs.io/en/latest/abtest.html#creating-experimental-pages)

### DIFF
--- a/bedrock/exp/templates/exp/firefox/mobile.html
+++ b/bedrock/exp/templates/exp/firefox/mobile.html
@@ -1,0 +1,217 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import google_play_button, send_to_device with context %}
+{% from "macros-protocol.html" import feature_card, hero, call_out_compact with context %}
+
+{% extends "exp/base/base.html" %}
+
+{% block page_css %}
+  {{ css_bundle('exp-firefox-mobile') }}
+{% endblock %}
+
+{% block canonical_urls %}
+  {# Experimental pages should set canonical to the original page URL. #}
+  <link rel="canonical" href="{{ settings.CANONICAL_URL }}/{{ LANG }}/firefox/mobile/">
+{% endblock %}
+
+{% block page_title %}{{ _('Download the Firefox Browser on your Mobile for iOS and Android') }} | {{ _('Firefox') }}{% endblock %}
+{% block page_desc %}{{ _('Firefox Browser for Mobile blocks over 2000 trackers by default, giving you the privacy you deserve and the speed you need in a private mobile browser.') }}{% endblock %}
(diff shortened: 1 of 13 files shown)


bedrock/exp/templates/exp/firefox/mobile.html
```diff
@@ -0,0 +1,217 @@
...
+
+{% extends "exp/base/base.html" %}
+
+{% block page_css %}
+  {{ css_bundle('exp-firefox-mobile') }}
+{% endblock %}
+
+{% block canonical_urls %}
+  {# Experimental pages should set canonical to the original page URL. #}
+  <link rel="canonical" href="{{ settings.CANONICAL_URL }}/{{ LANG }}/firefox/mobile">
```
alexgibson, line 16: Nit add trailing slash?
```suggestion
  <link rel="canonical" href="{{ settings.CANONICAL_URL }}/{{ LANG }}/firefox/mobile/">
```

bedrock/exp/templates/exp/firefox/mobile.html
```diff
@@ -0,0 +1,217 @@
...
+{% endblock %}
+
+{% block page_title %}{{ _('Download the Firefox Browser on your Mobile for iOS and Android') }} | {{ _('Firefox') }}{% endblock %}
```
alexgibson, line 19: I don't suggest changing this now as you have everything in place already, but we're on the verge of switching to a new L10n system, and these pages are never going to be translated. It doesn't hurt leaving the `_()` calls in the templates, but they are going to become redundant soon and eventually removed later down the road. We should probably try and resist adding them where they're not needed.

bedrock/exp/views.py
```diff
@@ -24,3 +28,29 @@ def new(request):
...
+    locale = l10n_utils.get_locale(request)
+    donate_params = settings.DONATE_PARAMS.get(
+        locale, settings.DONATE_PARAMS['en-US'])
+
+    # presets are stored as a string but, for the home banner
+    # we need it as a list.
+    donate_params['preset_list'] = donate_params['presets'].split(',')
+    ctx = {
+        'donate_params': donate_params,
+        'pocket_articles': PocketArticle.objects.all()[:4]
```
alexgibson, line 43: I think you need to set `active_locales` in the context here seeing as these templates don't rely on a .lang file. See this view as an example: https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/views.py#L791

alexgibson:
Oh, adding these URLs to this test would be good too?

https://github.com/mozilla/bedrock/blob/master/tests/functional/test_experiment_pages.py

PR #8879, opened 2020-05-04
Fix navigation visibility in IE when home banner is off

alexgibson:
Fixes https://gitlab.com/mozmeao/www-config/-/jobs/537668386


alexgibson:
To test this PR:

- Open the home page in IE11.
- Either turn the fundraising banner off, or close it with the (x).
- The navigation should remain visible (vs obscured like on master branch)

PR #8900, opened 2020-05-08
Use a single docker image for all deployments

pmac:
We are setting up a process (k8s DaemonSet) that will update the database file and the l10n files (basically run `bin/run-file-clock.sh`) in a volume. Bedrock will then be configured to mount that volume at `/app/data` and the files will be kept updated that way. Therefore at launch the app will have all the right data immediately.

This PR updates bedrock to be ready for that by moving the database and l10n file repos in to the new `data` directory. It also updates deployment things to no longer build different images depending on the branch.

Fix https://github.com/mozmeao/infra/issues/1306


.gitlab-ci.yml
```diff
@@ -35,7 +35,7 @@ build-test:
     - mozmeao
     - aws
   script:
-    - make clean build-ci
+    - bin/build-release-image.sh
```
glogiotatidis, line 38: I can't find this file in the repo. Maybe you forgot to `git add`?

pmac: That is so strange. I remember creating it, but it's just gone now. I can't find a trace of it. Good catch.

pmac: Okay. I've recreated it.

pmac:
Had a good run on integration tests https://gitlab.com/mozmeao/www-config/pipelines/151053178

pmac:
I'm struggling to figure out how to properly deploy this. I think I need to split this into 2 PRs. The first will move the data into the data directory in preparation for using the volume from the daemonset. The second would modify the deployment to remove the branch name from the image tags and whatnot. 

The reason is that if we merge this then dev will briefly have prod data. Maybe that's okay. But it also needs to be deployed so that we can have an image we can use in the daemonset that will populate the data directory and thus the volume. So we have a kind-of 🐔 🥚 thing happening. What would be nice is if we had another deployment we could use to try all of this out. It should be dev, but that's the master branch and the place where people review translations and release notes and that makes me nervous that something could go wrong. We could always roll-back though.

Any thoughts @glogiotatidis @metadave @duallain ?

pmac:
@duallain we still want to do this?

duallain:
@pmac Yeah, I think so. Trying to think of the best way to do it:

Do the split into two prs you mentioned above
clear the runway with everyone (frontend devs etc.)
Push the changes through to dev/stage, wait a day/half day, push to iowa prod (leave frankfurt as a backup plan for a minute)?

pmac:
I'll look and think over it some more. It should be fine to push through to prod because it will require a setting and deployment change to use the database on the k8s node instead of inside the container.

pmac:
Things to do to enable this:

- [ ] Merge this
- [ ] Update dev deployment YAML with the volume for the data
- [ ] Update the deployment to run `bin/run-prod.sh` instead of `bin/run.sh`
- [ ] Delete the `RUN_SUPERVISOR` setting from the deployment

Issue #8501, opened 2020-02-11
Dot on channel page only in Chrome


stephaniehobson:
## Description

There's a weird lonely dot hanging out on the channel page.

### Steps to reproduce

Open: https://www.mozilla.org/en-US/firefox/channel/desktop/ in Chrome

Scroll down:
<img width="581" alt="Screen Shot 2020-02-10 at 15 58 43" src="https://user-images.githubusercontent.com/854701/74201336-4fd04300-4c1e-11ea-8cd9-01a1e01429bb.png">

### Expected result

No dot.

<img width="618" alt="Screen Shot 2020-02-10 at 15 59 58" src="https://user-images.githubusercontent.com/854701/74201383-6e363e80-4c1e-11ea-8ef5-20a1e6b1fadb.png">

### Actual result

Dot.

### Environment

Mac OS and Chrome.

craigcook:
It's a stray list marker. The picto card macro defaults to `li` as the base element but in this case it doesn't have a `ul` parent so it's not getting the global `list-style: none;` styling. Chrome evidently adds bullets to `li` elements when it doesn't know how else to treat them (Firefox seems to treat them as nothing; `li` isn't valid HTML without a parent `ul` or `ol`). 

Switching the macro call to a `div` or `section` should fix the bug and also be valid and more semantically correct.

ishantd:
Hi sorry for making a pull request before mentioning it here.
If it's okay, please review it. Thanks.

PR #8903, opened 2020-05-09
Random dot on chrome #8501

ishantd:
## Description
Random dot due to the macro making _**li**_ elements. Instead changed the _li_ to _div_ as mentioned by @craigcook for better semantics.

## Issue / Bugzilla link
#8501

### DIFF
--- a/bedrock/base/templates/macros-protocol.html
+++ b/bedrock/base/templates/macros-protocol.html
@@ -197,7 +197,7 @@
 
 
 {# Picto Card: https://protocol.mozilla.org/patterns/molecules/picto-card.html #}
-{% macro picto_card(title=None, desc=None, class=None, heading_level=2, base_el='li', custom_desc=False) -%}
+{% macro picto_card(title=None, desc=None, class=None, heading_level=2, base_el='div', custom_desc=False) -%}
 <{{ base_el }} class="mzp-c-card-picto {% if class %}{{ class }}{% endif %}">
   <div class="mzp-c-card-picto-content">
     {% if title %}<h{{ heading_level }} class="mzp-c-card-picto-title">{{ title }}</h{{ heading_level }}>{% endif %}
--- a/bedrock/exp/templates/exp/firefox/mobile.html
+++ b/bedrock/exp/templates/exp/firefox/mobile.html
@@ -50,7 +50,7 @@
               </button>
             </div>
             <div class="mobile-download-buttons-wrapper">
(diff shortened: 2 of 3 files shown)


bedrock/base/templates/macros-protocol.html
```diff
@@ -197,7 +197,7 @@
 
 
 {# Picto Card: https://protocol.mozilla.org/patterns/molecules/picto-card.html #}
-{% macro picto_card(title=None, desc=None, class=None, heading_level=2, base_el='li', custom_desc=False) -%}
+{% macro picto_card(title=None, desc=None, class=None, heading_level=2, base_el='div', custom_desc=False) -%}
```
alexgibson, line 200: Changing this does fix the issue on the /firefox/channel/desktop/ page, however in the majority of other pages this now creates invalid markup, because there the macro is commonly used as a list item, and a `<div>` shouldn't be a direct child of a `<ul>`.

So, to fix this bug fully you could either:

- Update every other instance of the macro  use `li` only where needed, and leave the default here set to `div`.
- Or alternatively set the macro on the /firefox/channel/desktop/ page to use `base_el='div'`, and leave the default value as `li`.

The first solution is probably the right thing to do, but it does require more work and testing than the second solution. Seeing as this is your own free time, I'll leave it up to you which option you want to go for here.

bedrock/exp/templates/exp/firefox/mobile.html
```diff
@@ -50,7 +50,7 @@ <h3>{{ _('Get automatic privacy on mobile') }}</h3>
               </button>
             </div>
             <div class="mobile-download-buttons-wrapper">
-              <ul class="mobile-download-buttons firefox" id="mobile-download-buttons-firefox">
```
alexgibson, line 53: I don't think this change is related to this PR?

bedrock/firefox/templates/firefox/mobile/index.html
```diff
@@ -47,7 +47,7 @@ <h3>{{ _('Get automatic privacy on mobile') }}</h3>
               </button>
             </div>
             <div class="mobile-download-buttons-wrapper">
-              <ul class="mobile-download-buttons firefox" id="mobile-download-buttons-firefox">
```
alexgibson, line 50: I don't think this change is related to this PR?

alexgibson:
Thanks for this, but another contributor fixed this bug in https://github.com/mozilla/bedrock/pull/9077. Closing

PR #8914, opened 2020-05-13
Use the UITour for the FxA flow

vladikoff:
Ref: https://bugzilla.mozilla.org/show_bug.cgi?id=1635219

@rfk @alexgibson this is a draft of the changes we could make to route all the FxA links and forms using the UITour. Alex, see the chart below, we are working on moving the login to OAuth and to be able to generate OAuth urls we would need to "Create the FxA url" in Desktop.

TODO

* [x] Figure out the /metrics flow params. Those might need changes in Desktop :(
* [x] Should we convert the links into `<buttons>` instead? Otherwise the "non-sync" link can be copy pasted and opened

Bedrock UITour

### DIFF
--- a/docs/firefox-accounts.rst
+++ b/docs/firefox-accounts.rst
@@ -37,6 +37,20 @@ predictability in retroactive analysis.
     <https://mozilla.github.io/application-services/docs/accounts/metrics.html#descriptions-of-metrics-related-query-parameters>`_
     rules applied by the Firefox Accounts server.
 
+UITour Flow
+-----------
+
+Since Firefox 80 the FxA links and email forms use the UITour to show the Firefox Accounts page and log the browser
+into Sync or an Account. For non-Firefox browsers or if the UITour is not available the flow uses normal links that
+allow users to log into FxA as a website only without connecting the Firefox Desktop client.
+This UITour flow allows the Firefox browser to determine the correct FxA server and authentication flow.
+This transition was introduced to later migrate Firefox Desktop to an OAuth based client authentication flow.
+
+The current code automatically detects if you are in the supported browser for this flow and updates links and forms
+to drive them through the UITour API. The UITour ``showFirefoxAccounts`` action supports flow id parameters,
+UTM parameters and the email data field.
+
+We hope to remove the legacy non-UITour login logic after 1 or 2 ESRs.
 
 Signup Form
 -----------
--- a/docs/uitour.rst
+++ b/docs/uitour.rst
@@ -48,6 +48,13 @@ Note that ``browser.uitour.testingOrigins`` can be a comma separated list of dom
     must also exclude the domain protocol e.g. ``https://``. A browser restart is also required
     after adding an allowed domain.
 
+    If you are working on Firefox Accounts integration, you can use the ``identity.fxaccounts.autoconfig.uri``
+    config property to change the Accounts server. For example, to change it to stage environment use this value:
+    ``https://accounts.stage.mozaws.net/``. Restart the browser and make sure the configuration updated.
+    ``identity.fxaccounts.remote.root`` preference should now point to ``https://accounts.stage.mozaws.net``.
+    If it has not changed for some reason, update it manually.
+    Ref: https://mozilla-services.readthedocs.io/en/latest/howtos/run-fxa.html
+
 JavaScript API
 --------------
 
--- a/media/js/base/fxa-utm-referral.js
+++ b/media/js/base/fxa-utm-referral.js
@@ -14,9 +14,12 @@ if (typeof window.Mozilla === 'undefined') {
 
     var allowedList = [
         'https://accounts.firefox.com/',
+        'https://accounts.firefox.com.cn/',
+        'https://accounts.stage.mozaws.net/',
         'https://monitor.firefox.com/',
         'https://getpocket.com/',
-        'https://latest.dev.lcip.org/'
+        'https://latest.dev.lcip.org/',
+        'https://stable.dev.lcip.org/'
     ];
 
     /**
--- a/media/js/base/mozilla-fxa-form.js
+++ b/media/js/base/mozilla-fxa-form.js
@@ -30,26 +30,12 @@ if (typeof window.Mozilla === 'undefined') {
     /**
      * Pass through utm_params from URL if present,
      * to attribute external marketing campaigns.
+     *
+     * @returns {Object} of utm parameters
      */
     FxaForm.getUTMParams = function() {
         var urlParams = new window._SearchParams().utmParams();
-        var attributionData = Mozilla.UtmUrl.getAttributionData(urlParams);
-
-        if (attributionData) {
-            var utms = ['utm_source', 'utm_campaign', 'utm_content', 'utm_term', 'utm_medium'];
-            for (var i = 0; i < utms.length; i++) {
-                if (Object.prototype.hasOwnProperty.call(attributionData, utms[i])) {
-                    // check if input is available
-                    if (formElem.querySelector('[name="' + utms[i] + '"]')) {
-                        formElem.querySelector('[name="' + utms[i] + '"]').value = attributionData[utms[i]];
-                    } else {
-                        // create input if one is not present
-                        var input = FxaForm.createInput(utms[i], attributionData[utms[i]]);
-                        formElem.appendChild(input);
-                    }
-                }
-            }
-        }
+        return Mozilla.UtmUrl.getAttributionData(urlParams) || {};
     };
 
     /**
@@ -102,6 +88,55 @@ if (typeof window.Mozilla === 'undefined') {
         });
     };
 
+
+    /**
+     * Intercept event handler for FxA forms, lets the browser drive the FxA Flow using
+     * the `showFirefoxAccounts` UITour API. Attaches several UTM parameters from the current page
+     * that will be forwarded to the browser and later on to FxA services.
+     * @param event {Event}
+     * @private
+     */
+    FxaForm.interceptFxANavigation = function(event) {
+        event.preventDefault();
+        var extraURLParams = FxaForm.getUTMParams();
+
+        var entrypointInput = document.getElementById('fxa-email-form-entrypoint');
+        var entrypointExp = document.getElementById('fxa-email-form-entrypoint-experiment');
+        var entrypointVar = document.getElementById('fxa-email-form-entrypoint-variation');
+        var entrypoint = null;
+        if (entrypointInput && entrypointInput.value) {
+            entrypoint = entrypointInput.value;
+        }
+        if (entrypointExp && entrypointExp.value) {
+            extraURLParams['entrypoint_experiment'] = entrypointExp.value;
+        }
+        if (entrypointVar && entrypointVar.value) {
+            extraURLParams['entrypoint_variation'] = entrypointVar.value;
+        }
+
+        var email = document.getElementById('fxa-email-field');
+        if (email) {
+            email = email.value;
+        }
+
+        var formElem = document.getElementById('fxa-email-form');
+        if (formElem) {
+            var deviceId = formElem.querySelector('[name="device_id"]');
+            var flowId = formElem.querySelector('[name="flow_id"]');
+            var flowBeginTime = formElem.querySelector('[name="flow_begin_time"]');
+            if (deviceId && deviceId.value) {
+                extraURLParams['device_id'] = deviceId.value;
+            }
+            if (flowId && flowId.value) {
+                extraURLParams['flow_id'] = flowId.value;
+            }
+            if (flowBeginTime && flowBeginTime.value) {
+                extraURLParams['flow_begin_time'] = parseInt(flowBeginTime.value, 10);
+            }
+        }
+        return Mozilla.UITour.showFirefoxAccounts(extraURLParams, entrypoint, email);
+    };
+
     /**
      * Configures Sync for Firefox browsers.
      * Only Firefox < 71 requires `service=sync`.
@@ -110,7 +145,16 @@ if (typeof window.Mozilla === 'undefined') {
         var contextField = formElem.querySelector('[name="context"]');
         var serviceField = formElem.querySelector('[name="service"]');
         var userVer = parseFloat(Mozilla.Client._getFirefoxVersion());
+        var self = this;
+        var useUITourForFxA = userVer >= 80 && typeof Mozilla.UITour !== 'undefined';
 
+        if (useUITourForFxA) {
+            Mozilla.UITour.ping(function() {
+                // intercept the flow and submit the form using the UITour API instead.
+                // In the future we should fully migrate to this API for Firefox Desktop login.
+                formElem.addEventListener('submit', self.interceptFxANavigation);
+            });
+        }
         /**
          * Context is required for all Firefox desktop clients.
          */
@@ -180,7 +224,17 @@ if (typeof window.Mozilla === 'undefined') {
         return new window.Promise(function(resolve, reject) {
             if (formElem) {
                 // Pass through UTM params from the URL to the form.
-                FxaForm.getUTMParams();
+                var utms = FxaForm.getUTMParams();
+                Object.keys(utms).forEach(function (i) {
+                    // check if input is available
+                    if (formElem.querySelector('[name="' + i + '"]')) {
+                        formElem.querySelector('[name="' + i + '"]').value = utms[i];
+                    } else {
+                        // create input if one is not present
+                        var input = FxaForm.createInput(i, utms[i]);
+                        formElem.appendChild(input);
+                    }
+                });
 
                 // Configure Sync for Firefox desktop browsers.
                 if (Mozilla.Client._isFirefoxDesktop()) {
--- a/media/js/base/mozilla-fxa-link.js
+++ b/media/js/base/mozilla-fxa-link.js
@@ -12,6 +12,13 @@ if (typeof window.Mozilla === 'undefined') {
 
     var FxaLink = {};
     var client = Mozilla.Client;
+    var allowedList = [
+        'https://accounts.firefox.com/',
+        'https://accounts.stage.mozaws.net/',
+        'https://latest.dev.lcip.org/',
+        'https://stable.dev.lcip.org/',
+        'https://accounts.firefox.com.cn/'
+    ];
 
     /**
      * Configures Sync for Firefox browsers.
@@ -35,14 +42,62 @@ if (typeof window.Mozilla === 'undefined') {
     };
 
     /**
+     * Returns the hostname for a given URL.
+     * @param {String} url.
+     * @returns {String} hostname.
+     */
+    FxaLink.getHostName = function(url) {
+        var matches = url.match(/^https?:\/\/(?:[^/?#]+)(?:[/?#]|$)/i);
+        return matches && matches[0];
+    };
+
+    /**
+     * Intercept event handler for FxA links and forms, lets the browser drive the FxA Flow using
+     * the `showFirefoxAccounts` UITour API. Attaches several UTM parameters from the current page
+     * that will be forwarded to the browser and later on to FxA services.
+     * @param event {Event}
+     * @private
+     */
+    FxaLink.interceptFxANavigation = function(event) {
+        event.preventDefault();
+        var search = new URL(event.target.href).search;
+        var urlParams = new window._SearchParams(search).params;
+        var extraURLParams = {};
+        var allowedChars = /^[\w/.%-]+$/;
+        var knownParams = ['flow_id', 'flow_begin_time', 'device_id',
+            'entrypoint', 'entrypoint_experiment', 'entrypoint_variation',
+            'utm_source', 'utm_campaign', 'utm_content', 'utm_term', 'utm_medium'];
+
+        for (var i = 0; i < knownParams.length; i++) {
+            var known = knownParams[i];
+            if (Object.prototype.hasOwnProperty.call(urlParams, known)) {
+                var param = decodeURIComponent(urlParams[known]);
+                if ((allowedChars).test(param)) {
+                    extraURLParams[known] = param;
+                }
+            }
+        }
+        var entrypoint = extraURLParams.entrypoint;
+        // 'entrypoint' is sent separately and should not be in the extras
+        delete extraURLParams.entrypoint;
+        return Mozilla.UITour.showFirefoxAccounts(extraURLParams, entrypoint);
+    };
+
+    /**
      * Updates FxA links with Sync params.
      * Only applicable for Firefox desktop user agents.
      */
-    FxaLink.init = function() {
+    FxaLink.init = function(callback) {
+        callback = callback || function noop() {
+            // do nothing
+        };
         if (!client._isFirefoxDesktop()) {
             return;
         }
 
+        var userVer = parseFloat(client._getFirefoxVersion());
+        var useUITourForFxA = userVer >= 80 && typeof Mozilla.UITour !== 'undefined';
+        var self = this;
         var fxaSigninLink = document.querySelectorAll('.js-fxa-cta-link');
 
         for (var i = 0; i < fxaSigninLink.length; i++) {
@@ -57,9 +112,32 @@ if (typeof window.Mozilla === 'undefined') {
                 link.setAttribute('data-mozillaonline-link', FxaLink.updateURL(mozillaOnlineLink));
             }
         }
+
+        if (useUITourForFxA) {
+            Mozilla.UITour.ping(function() {
+                for (var i = 0; i < fxaSigninLink.length; i++) {
+                    var link = fxaSigninLink[i];
+                    var hostName = FxaLink.getHostName(link.href);
+                    // check if link is in the FxA referral allowedList domains.
+                    if (hostName && allowedList.indexOf(hostName) === -1) {
+                        continue;
+                    }
+                    link.setAttribute('role', 'button');
+                    link.oncontextmenu = function(e) {
+                        e.preventDefault();
+                    };
+                    // intercept the flow and submit the form using the UITour API instead.
+                    // In the future we should fully migrate to this API for Firefox Desktop login.
+                    link.addEventListener('auxclick', self.interceptFxANavigation);
+                    link.addEventListener('click', self.interceptFxANavigation);
+                }
+                callback();
+            });
+        } else {
+            callback();
+        }
     };
 
     window.Mozilla.FxaLink = FxaLink;
 
 })();
-
--- a/media/static-bundles.json
+++ b/media/static-bundles.json
@@ -810,7 +810,6 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/libs/spin.min.js",
         "protocol/js/protocol-modal.js",
         "js/base/send-to-device.js",
@@ -820,21 +819,18 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/welcome/welcome8.js"
       ],
       "name": "firefox_welcome_page8"
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/privacy/messaging-experiment.js"
       ],
       "name": "default_privacy_messaging"
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/base/mozilla-fxa.js",
         "js/base/mozilla-fxa-init.js",
         "js/base/mozilla-fxa-product-button.js",
@@ -846,7 +842,6 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/whatsnew/up-to-date.js",
         "js/base/mozilla-fxa-product-button.js",
         "js/base/mozilla-fxa-product-button-init.js"
@@ -855,7 +850,6 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/whatsnew/up-to-date.js",
         "js/libs/spin.min.js",
         "protocol/js/protocol-modal.js",
@@ -866,7 +860,6 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/whatsnew/up-to-date.js",
         "js/firefox/whatsnew/whatsnew-74.js"
       ],
@@ -874,7 +867,6 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/whatsnew/up-to-date.js",
         "js/base/mozilla-fxa-product-button.js",
         "js/base/mozilla-fxa-product-button-init.js",
@@ -884,7 +876,6 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/whatsnew/up-to-date.js",
         "js/firefox/whatsnew/whatsnew-76.js"
       ],
@@ -892,7 +883,6 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/whatsnew/up-to-date.js",
         "js/firefox/whatsnew/whatsnew-77.js"
       ],
@@ -900,7 +890,6 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/whatsnew/up-to-date.js",
         "js/base/mozilla-video-poster.js",
         "js/firefox/whatsnew/whatsnew-77-zhCN.js"
@@ -915,7 +904,6 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/lockwise/lockwise.js"
       ],
       "name": "lockwise"
@@ -935,7 +923,6 @@
     },
     {
       "files": [
-        "js/base/uitour-lib.js",
         "js/firefox/new/desktop/download.js"
       ],
       "name": "firefox_desktop_download"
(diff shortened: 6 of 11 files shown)


rfk:
>  Should we convert the links into instead?

I think you might be missing a word here, because I can't parse this question as-is.

media/js/base/mozilla-fxa-form.js
```diff
@@ -111,6 +111,28 @@ if (typeof window.Mozilla === 'undefined') {
         var serviceField = formElem.querySelector('[name="service"]');
         var userVer = parseFloat(Mozilla.Client._getFirefoxVersion());
 
+
+        var useUITourToSync = userVer > 77 && Mozilla.UITour;
```
rfk, line 115: Would it work to feature-detect the presence of `Mozilla.UITour.showFirefoxAccounts` rather than do a version comparison here?

alexgibson: I don't think a feature detect will work here, as `Mozilla.UITour.showFirefoxAccounts` is just a JS lib as opposed to a browser feature.

alexgibson:
@vladikoff thanks for taking this on. My main question here is how this will interplay with the /metrics flow logic that lives in [mozilla-fxa-product-button.js](https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-fxa-product-button.js). Right now, I think your new UITour call will clobber this data?

vladikoff:
> > Should we convert the links into instead?
> 
> I think you might be missing a word here, because I can't parse this question as-is.

hah sorry github converted my HTML tag 😄 , it's:

> Should we convert the links into `<buttons>` instead?

alexgibson:
> Should we convert the links into <buttons> instead?

If this is still performing a page navigation, I would side on it making slightly more sense to remain a link.

vladikoff:
> My main question here is how this will interplay with the /metrics flow logic that lives in mozilla-fxa-product-button.js. 

@alexgibson Here's a patch to support flow params in Firefox UITour https://phabricator.services.mozilla.com/D77542

vladikoff:
@alexgibson could you please take a look at these changes? 

There is a bit of a dependency of `form` on `link` introduced in this. Is there some other way we can deal with that?

media/js/base/mozilla-fxa-link.js
```diff
@@ -43,12 +86,21 @@ if (typeof window.Mozilla === 'undefined') {
...
         for (var i = 0; i < fxaSigninLink.length; i++) {
             var link = fxaSigninLink[i];
             var mozillaOnlineLink = link.getAttribute('data-mozillaonline-link');
 
+            if (useUITourForFxA) {
+                // intercept the flow and submit the form using the UITour API instead.
+                // In the future we should fully migrate to this API for Firefox Desktop login.
+                link.addEventListener('auxclick', this.interceptFxANavigation);
+                link.addEventListener('click', this.interceptFxANavigation);
+                return;
```
alexgibson, line 102: The `return` statement here is preventing the UITour even from being bound to all but the first link in the page.

media/js/base/mozilla-fxa-link.js
```diff
@@ -43,12 +86,21 @@ if (typeof window.Mozilla === 'undefined') {
...
         var fxaSigninLink = document.querySelectorAll('.js-fxa-cta-link');
 
         for (var i = 0; i < fxaSigninLink.length; i++) {
             var link = fxaSigninLink[i];
             var mozillaOnlineLink = link.getAttribute('data-mozillaonline-link');
 
+            if (useUITourForFxA) {
+                // intercept the flow and submit the form using the UITour API instead.
+                // In the future we should fully migrate to this API for Firefox Desktop login.
+                link.addEventListener('auxclick', this.interceptFxANavigation);
```
alexgibson, line 100: When I try to right click one of these links it immediately opens a new tab, as opposed to letting me use the context menu (for example to inspect the element or the link URL). This feels a little unexpected/disorienting - is there another way to deal with someone trying to trigger one of these links in a new tab?

alexgibson:
@vladikoff before I dive into a code level review, here are some thoughts on the approach:

- You can no doubt help explain here, but I'm still a little in the dark as to why this new user flow needs to use UITour? We have a growing number of conditionals in bedrock to support different user flows for various Firefox versions, and it's starting to egde on being complicated. In the long term, it would be great if we could perhaps work with the FxA team to reduce some of the legacy logic, which would make it easier for downstream referrers like bedrock to maintain our signup flows. In the short term, it would be great if you could add some documentation for this change to [our docs](https://bedrock.readthedocs.io/en/latest/firefox-accounts.html)?
- We should try and consider UITour like a form of _progressive enhancement_. Currently if UITour is disabled (for example the user could have a profile issue, which does happen sometimes), then clicking on the fxa links / submit button doesn't do anything right now. It's a dead-end. I'd expect a valid user navigation to still occur here in all cases. Ideally, people should still be able to get to accounts.firefox.com and sign up using an existing user flow? I'd suggest wrapping the event bindings behind a call to `UITour.ping()` to accomodate this.
- I agree there's too much of a dependency between `mozilla-fxa-form.js` and `mozilla-fxa-link.js` as it stands. I'm not opposed to a small amount of code duplication if that means we can more easily keep concerns separated, as this I feel this would help make the code easier to reason about. I'd expect all the code relevant to the FxA form to live in `mozilla-fxa-form.js`, for example.
- We're going to need to be careful that this intercept doesn't overwrite links to other services, such as the buttons to monitor.firefox.com. It may be worth adding a whitelist similar to the one we use in [`mozilla-product-button.js`](https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-fxa-product-button.js#L15-L21)?
- With this new flow, the action always opens accounts.firefox.com. I'm assuming there's a way to change this via a config to a dev / staging server when we need?

I'll likely have some other questions / comment in the review, but I'll leave this here for now so you can have a chance to reply. Thanks!

vladikoff:
Thanks for the questions!

> @vladikoff before I dive into a code level review, here are some thoughts on the approach:
> 
> * You can no doubt help explain here, but I'm still a little in the dark as to why this new user flow needs to use UITour? 

We are migrating the fxa auth flow to OAuth (in effort to consolidate how we authenticate Sync clients). We use this new flow in iOS and Fenix already with the help of our shared components. To start "an OAuth flow" the clients need to keep track of a state, know the client id, redirect params and generate public key bundles for Sync keys. This is a lot of work to put on the Bedrock's side. The UITour seemed like the candidate that can help begin and hand over the flow to the Browser so it can do all the hard work.

Another related bug deals with security of the flow: https://bugzilla.mozilla.org/show_bug.cgi?id=1449520

> We have a growing number of conditionals in bedrock to support different user flows for various Firefox versions, and it's starting to egde on being complicated. In the long term, it would be great if we could perhaps work with the FxA team to reduce some of the legacy logic, which would make it easier for downstream referrers like bedrock to maintain our signup flows. In the short term, it would be great if you could add some documentation for this change to [our docs](https://bedrock.readthedocs.io/en/latest/firefox-accounts.html)?

I am hoping by handing over the flow to the UITour we would no longer need to keep the conditions in bedrock. I'm also optimistic that we can remove the conditionals in the future and just use the UITour to open FxA flows. For example I am hoping the `context` logic will go away after we ship 1 ESR after these changes land.

> * We should try and consider UITour like a form of _progressive enhancement_. Currently if UITour is disabled (for example the user could have a profile issue, which does happen sometimes), then clicking on the fxa links / submit button doesn't do anything right now. It's a dead-end. I'd expect a valid user navigation to still occur here in all cases. Ideally, people should still be able to get to accounts.firefox.com and sign up using an existing user flow? I'd suggest wrapping the event bindings behind a call to `UITour.ping()` to accomodate this.

The UITour ping idea could help ! Good suggestion!

> * I agree there's too much of a dependency between `mozilla-fxa-form.js` and `mozilla-fxa-link.js` as it stands. I'm not opposed to a small amount of code duplication if that means we can more easily keep concerns separated, as this I feel this would help make the code easier to reason about. I'd expect all the code relevant to the FxA form to live in `mozilla-fxa-form.js`, for example.

Sounds good! 

> * We're going to need to be careful that this intercept doesn't overwrite links to other services, such as the buttons to monitor.firefox.com. It may be worth adding a whitelist similar to the one we use in [`mozilla-product-button.js`](https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-fxa-product-button.js#L15-L21)?

Will check! thanks!

> * With this new flow, the action always opens accounts.firefox.com. I'm assuming there's a way to change this via a config to a dev / staging server when we need?

This would involve changing the FxA Configuration in Firefox Desktop. Like so: 
image


> 
> I'll likely have some other questions / comment in the review, but I'll leave this here for now so you can have a chance to reply. Thanks!

stephaniehobson:
I'm sorry I don't have time to parse the code, this may already have been taken into account 😅 

The flows in the diagrams in the description show only the Firefox flow but the form on /accounts is presented to and can be filled in by users of any browser not just Firefox.

alexgibson:
I believe this update is for the Sync flow, which is Firefox only - but @vladikoff can confirm here. Other browsers should still enter the regular sign up flow.

vladikoff:
> I believe this update is for the Sync flow, which is Firefox only - but @vladikoff can confirm here. Other browsers should still enter the regular sign up flow.

Correct. The non-Firefox flow remains unchanged.

media/js/base/mozilla-fxa-link.js
vladikoff: Hm I think the only way would be to make it a `<button>`?

alexgibson: It might be easier to add `role=button` to the links for now, and then disable the context menu?

```javascript
link.setAttribute('role', 'button');
link.oncontextmenu = function(e) {
    e.preventDefault();
};
```

This then avoids the odd menu behaviour, and should at least provide user agents with a hint as to the role of the element.

vladikoff: Awesome, thank you for the tip! Yesterday I tried replacing the DOM node with the button. It worked well for the most part except the styling. I think your suggestion should work better

docs/firefox-accounts.rst
```diff
@@ -37,6 +37,18 @@ predictability in retroactive analysis.
     <https://mozilla.github.io/application-services/docs/accounts/metrics.html#descriptions-of-metrics-related-query-parameters>`_
     rules applied by the Firefox Accounts server.
 
+UITour Flow
+-----------
+
+Since Firefox 79 the FxA links and email forms use the UITour to show the Firefox Accounts page and log the browser
```
vladikoff, line 43: @rfk anything else to mention here?

media/js/base/mozilla-fxa-form.js
```diff
@@ -102,6 +102,50 @@ if (typeof window.Mozilla === 'undefined') {
...
+     * Intercept event handler for FxA links and forms, lets the browser drive the FxA Flow using
+     * the `showFirefoxAccounts` UITour API. Attaches several UTM parameters from the current page
+     * that will be forwarded to the browser and later on to FxA services.
+     * @param event {Event}
+     * @private
+     */
+    FxaForm.interceptFxANavigation = function(event) {
+        event.preventDefault();
+        var urlParams = new window._SearchParams().utmParams();
+        var attributionData = Mozilla.UtmUrl.getAttributionData(urlParams);
```
alexgibson, line 116: Looks like we already have a `FxaForm.getUTMParams()` function above on line 24. Rather than repeat the logic here, might it be worth altering that function to return only the params, and then the calling code can reuse it for each context? That way we'd only have to query the params once, rather than do it all again here.

media/js/base/mozilla-fxa-link.js
```diff
@@ -34,6 +34,49 @@ if (typeof window.Mozilla === 'undefined') {
...
+            if (deviceId && deviceId.value) {
+                extraURLParams['device_id'] = deviceId.value;
+            }
+            if (flowId && flowId.value) {
+                extraURLParams['flow_id'] = flowId.value;
+            }
+            if (flowBeginTime && flowBeginTime.value) {
+                extraURLParams['flow_begin_time'] = parseInt(flowBeginTime.value, 10);
+            }
+        }
```
alexgibson, line 76: I don't think we need this logic that deals with the form, as we split this out into `mozilla-fxa-form.js`, right?

media/static-bundles.json
```diff
@@ -1296,6 +1296,7 @@
         "js/base/core-datalayer.js",
         "js/base/core-datalayer-init.js",
         "js/base/search-params.js",
+        "js/base/uitour-lib.js",
```
alexgibson, line 1299: If we're adding this to the base bundle, then please add it to `common`, `common-protocol`, and `common-protocol-no-jquery`. It would then also need removing from any page-specific bundles to avoid including it twice.

tests/unit/spec/base/mozilla-fxa-form.js
```diff
@@ -146,6 +148,62 @@ describe('mozilla-fxa-form.js', function() {
...
+        it('should pass through utm parameters from the URL to the form into the UITour', function() {
+            spyOn(window.Mozilla.Client, '_isFirefoxDesktop').and.returnValue(true);
+            spyOn(window.Mozilla.Client, '_getFirefoxVersion').and.returnValue('79.0');
+            spyOn(Mozilla.Client, 'getFirefoxDetails').and.callFake(function(callback) {
+                callback({
+                    'accurate': true,
+                    'distribution': undefined,
+                });
+            });
+            window.Mozilla.UITour = {
```
alexgibson, line 161: Rather than overwriting these function I'd suggest using [sinon](https://sinonjs.org/releases/v9.0.2/stubs/) to stub things out. You can see an example of that here:

https://github.com/mozilla/bedrock/blob/master/tests/unit/spec/base/core-datalayer-page-id.js#L12-L16

media/js/base/mozilla-fxa-link.js
```diff
@@ -34,6 +34,49 @@ if (typeof window.Mozilla === 'undefined') {
...
+    /**
+     * Intercept event handler for FxA links and forms, lets the browser drive the FxA Flow using
+     * the `showFirefoxAccounts` UITour API. Attaches several UTM parameters from the current page
+     * that will be forwarded to the browser and later on to FxA services.
+     * @param event {Event}
+     * @private
+     */
+    FxaLink.interceptFxANavigation = function(event) {
+        event.preventDefault();
+        var urlParams = new window._SearchParams().utmParams();
```
alexgibson, line 46: We already have some logic for adding utm parameters automatically to Fxa links in [fxa-utm-referrer.js](https://github.com/mozilla/bedrock/blob/master/tests/unit/spec/base/fxa-utm-referral.js). Rather than reconstruct things again from the URL, might it be easier to get the existing query string directly from the link?

This way you would also pick up flow_id params that might also exist there, which seem to get missed with the current logic here? (unles I'm misreading something).

docs/firefox-accounts.rst
```diff
@@ -37,6 +37,18 @@ predictability in retroactive analysis.
...
+into Sync or an Account. For non-Firefox browsers or if the UITour is not available the flow uses normal links that
```
alexgibson, line 44: > For non-Firefox browsers or if the UITour is not available the flow uses normal links that
allow users to log into FxA as a website only without connecting the Firefox Desktop client.

I'm curious as to what the behaviour should be here, as right now if UITour isn't available we still have all the legacy logic in place to support signing in to Sync for older Firefox versions. I'm assuming we want to keep this, so we should probably make that clearer here?

alexgibson:
When I click one of the UITour generated links, the URL I'm redirected to is:

https://accounts.firefox.com/signin?context=fx_desktop_v3&entrypoint=uitour&action=email&service=sync

One point worth considering here is `entrypoint` is overwritten with the value of `uitour`. This means we're likely going to drop page-specific information such as `entrypoint=mozilla.org-whatsnew60`?

hoosteeno:
> One point worth considering here is entrypoint is overwritten with the value of uitour. This means we're likely going to drop page-specific information such as entrypoint=mozilla.org-whatsnew60?

We agreed at the start that existing attribution would be unchanged with this integration and that still seems to be an important requirement. Please reach out if we need to discuss more.

vladikoff:
> We agreed at the start that existing attribution would be unchanged with this integration and that still seems to be an important requirement. 

Sounds good, I will set it up to forward the entrypoints, it is an important one

docs/firefox-accounts.rst
vladikoff: > if UITour isn't available we still have all the legacy logic in place to support signing in to Sync for older Firefox versions.

I don't know what the deprecation strategy is in place right now. My thoughts on this were: we can support the legacy logic for 1 or 2 ESRs + keep an eye on FxA metrics and see where the logins are coming from. Once it is down to almost zero we can remove the legacy flows. I will update the docs above

hoosteeno:
> > We agreed at the start that existing attribution would be unchanged with this integration and that still seems to be an important requirement.
> 
> Sounds good, I will set it up to forward the entrypoints, it is an important one

(This isn't a comment on present code because I'm not reading the code, but it is a requirement we should ensure present code handles, relevant to code comments just above.)

We should pass through and auto-populate the same params the present FxA authN flow passes through, as well as handle the entire set of potential parameters. @alexgibson , please jump in if I'm missing something. This is documented fairly well at https://bedrock.readthedocs.io/en/latest/firefox-accounts.html#configuration. 

1. Pass through any entrypoint or utm-* params present on initial request URLs
2. If not present, auto-populate entrypoint and utm-* params with www.mozilla.org default values and send to FxA servers
3. Allow Bedrock to set any additional params mentioned in docs (e.g. experiment variation) to be sent to FxA servers

Does that make sense?

alexgibson:
I think it we take the approach of using what's already in the FxA link query strings, vs trying to rebuild params over again via page URL, then we should be able to preserve the data that's already there.

vladikoff:
I have filed a bug to add support for missing parameters:   'entrypoint', 'entrypoint_experiment', 'entrypoint_variation'

The rest should already be supported

https://bugzilla.mozilla.org/show_bug.cgi?id=1648180

media/js/base/mozilla-fxa-link.js
vladikoff: > which seem to get missed with the current logic here?

The `fxa-link` doesn't have `flow_id`, only the form has the flow data.

vladikoff:
The full list of params that I gathered for the FxA login flow supports (not `/metrics-flow`) is the following:
```
flow_id
flow_begin_time
device_id
utm_source
utm_content
utm_medium
utm_term
utm_campaign
entrypoint_experiment
entrypoint_variation
```

Let me know if I am missing something

hoosteeno:
> Let me know if I am missing something

It was discussed earlier, so the code probably includes it. But your list of all params excludes `entrypoint`.

docs/firefox-accounts.rst
rfk: nope, nothing in particular to add, thanks!

docs/uitour.rst
```diff
@@ -48,6 +48,9 @@ Note that ``browser.uitour.testingOrigins`` can be a comma separated list of dom
     must also exclude the domain protocol e.g. ``https://``. A browser restart is also required
     after adding an allowed domain.
 
+    If you are working on Firefox Accounts integration, you can use the ``identity.fxaccounts.autoconfig.uri``
+    config property.
```
rfk, line 52: Trying to read this doc with my beginner's hat on: what exactly would I use this property to *do*, and *why*? Ideally I think this would say this would say something like "use the blah-bah config property to use a locally-running FxA server" and link to FxA docs with more details about doing so.

media/js/base/mozilla-fxa-link.js
alexgibson: We add flow params to primary call-t-action links to FxA. This is handled by [`mozilla-fxa-product-button.js`](https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-fxa-product-button.js), so something we still need to accomodate for here.

media/js/base/mozilla-fxa-link.js
```diff
@@ -57,6 +93,25 @@ if (typeof window.Mozilla === 'undefined') {
                 link.setAttribute('data-mozillaonline-link', FxaLink.updateURL(mozillaOnlineLink));
             }
         }
+
+        if (useUITourForFxA) {
+            Mozilla.UITour.ping(function() {
+                for (var i = 0; i < fxaSigninLink.length; i++) {
```
alexgibson, line 99: We need to make sure that this only intercepts links to accounts.firefox.com, and not links to other fxa related flows, such as monitor.firefox.com. I'd suggest adding an allowed list of domains, similar to how we do here: https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-fxa-product-button.js#L15-L21

alexgibson: Also, do we need to add in extra logic to this change to handle the China repack for accounts.firefox.com.cn?

vladikoff: For the China repack, the UITour should intercept it and then hand it off to the client and that would use the FxA configuration that is on the client. I think if the user has the client configured for China servers then the client would send them to those servers.

media/js/base/fxa-utm-referral.js
```diff
@@ -14,6 +14,7 @@ if (typeof window.Mozilla === 'undefined') {
 
     var allowedList = [
         'https://accounts.firefox.com/',
+        'https://accounts.stage.mozaws.net/',
```
alexgibson, line 17: Let's add `https://stable.dev.lcip.org/` and `https://accounts.firefox.com.cn/` here for parity with `mozilla-fxa-link.js`?

media/js/base/mozilla-fxa-form.js
```diff
@@ -34,22 +34,18 @@ if (typeof window.Mozilla === 'undefined') {
...
-                        // create input if one is not present
-                        var input = FxaForm.createInput(utms[i], attributionData[utms[i]]);
-                        formElem.appendChild(input);
-                    }
+                    utmParams[i] = attributionData[utms[i]];
                 }
             }
         }
+
+        return utmParams;
```
alexgibson, line 48: Nit: add  comment at the top of this function to describe what's now being returned e.g. `* @returns {Array} of utm parameters`

media/js/base/mozilla-fxa-form.js
```diff
@@ -102,6 +98,55 @@ if (typeof window.Mozilla === 'undefined') {
         });
     };
 
+
+    /**
+     * Intercept event handler for FxA links and forms, lets the browser drive the FxA Flow using
```
alexgibson, line 103:
```suggestion
     * Intercept event handler for FxA forms, lets the browser drive the FxA Flow using
```

media/js/base/mozilla-fxa-form.js
```diff
@@ -110,7 +155,16 @@ if (typeof window.Mozilla === 'undefined') {
         var contextField = formElem.querySelector('[name="context"]');
         var serviceField = formElem.querySelector('[name="service"]');
         var userVer = parseFloat(Mozilla.Client._getFirefoxVersion());
+        var self = this;
+        var useUITourForFxA = userVer > 79 && Mozilla.UITour;
```
alexgibson, line 159: Should this now be 80? Also suggest making the check for `UITour` a little more well formed?

```suggestion
        var useUITourForFxA = userVer > 80 && typeof Mozilla.UITour !== 'undefined';
```

media/js/base/mozilla-fxa-link.js
```diff
@@ -34,21 +41,68 @@ if (typeof window.Mozilla === 'undefined') {
...
+     * @param event {Event}
+     * @private
+     */
+    FxaLink.interceptFxANavigation = function(event) {
+        event.preventDefault();
+        var search = new URL(event.target.href).search;
+        var urlParams = new window._SearchParams(search).params;
+        var extraURLParams = {};
+        var allowedChars = /^[\w/.%-]+$/;
+        // There are no `flowId` params below because only the `fxa-form` starts a new flow.
```
alexgibson, line 67: We still need to handle `flowId` params on links, as they can be added to main call-to-action links via `mozilla-fxa-product-button.js`.

media/js/base/mozilla-fxa-link.js
```diff
@@ -34,21 +41,68 @@ if (typeof window.Mozilla === 'undefined') {
...
+    FxaLink.init = function(callback) {
+        callback = callback || function noop() {
+            // do nothing
+        };
         if (!client._isFirefoxDesktop()) {
             return;
         }
 
+        var userVer = parseFloat(client._getFirefoxVersion());
+        var useUITourForFxA = userVer > 79 && Mozilla.UITour;
```
alexgibson, line 99:
```suggestion
        var useUITourForFxA = userVer > 80 && typeof Mozilla.UITour !== 'undefined';
```

media/js/base/search-params.js
```diff
@@ -80,6 +80,21 @@
         return utms;
     };
 
+    _SearchParams.prototype.entrypointParams = function() {
```
alexgibson, line 83: Is this used anywhere? If not it can likely be deleted.

media/js/base/mozilla-fxa-form.js
vladikoff: Sorry, this should be >= 80

vladikoff:
Updated!

media/js/base/mozilla-fxa-link.js
```diff
@@ -34,21 +41,68 @@ if (typeof window.Mozilla === 'undefined') {
...
+     * @param event {Event}
+     * @private
+     */
+    FxaLink.interceptFxANavigation = function(event) {
+        event.preventDefault();
+        var search = new URL(event.target.href).search;
+        var urlParams = new window._SearchParams(search).params;
+        var extraURLParams = {};
+        var allowedChars = /^[\w/.%-]+$/;
+        var knownParams = ['flow_id', 'flow_begin_time', 'device_id',
```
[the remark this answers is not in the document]
alexgibson: Hmm - but when I change the above locally, the call to `showFirefoxAccounts()` fails silently. I wonder, is this because the required format has changed?

If that's the case would we need to change this convention in our helpers site-wide? How would that effect older clients?

media/js/base/mozilla-fxa-form.js
```diff
@@ -30,26 +30,24 @@ if (typeof window.Mozilla === 'undefined') {
...
      * to attribute external marketing campaigns.
+     *
+     * @returns {Array} of utm parameters
      */
     FxaForm.getUTMParams = function() {
         var urlParams = new window._SearchParams().utmParams();
         var attributionData = Mozilla.UtmUrl.getAttributionData(urlParams);
+        var utmParams = {};
 
         if (attributionData) {
```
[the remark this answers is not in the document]
vladikoff: Good catch 👍

media/js/base/mozilla-fxa-link.js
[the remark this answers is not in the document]
vladikoff: Yeah to match the `utm_*` parameters, this followed with snake_case. hmmm

media/js/base/mozilla-fxa-form.js
```diff
@@ -102,6 +100,55 @@ if (typeof window.Mozilla === 'undefined') {
...
+                extraURLParams['device_id'] = deviceId.value;
+            }
+            if (flowId && flowId.value) {
+                extraURLParams['flow_id'] = flowId.value;
+            }
+            if (flowBeginTime && flowBeginTime.value) {
+                extraURLParams['flow_begin_time'] = parseInt(flowBeginTime.value, 10);
+            }
+        }
+        return Mozilla.UITour.showFirefoxAccounts(extraURLParams, entrypoint, email);
```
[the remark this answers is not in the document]
vladikoff: >  Similarly, do we need a sane default for entrypoint which isn't null? What would happen if the form doesn't contain an email field? etc

For `entrypoint` the browser checks for it. It will default to `uitour` if nothing is reported.
For `email` the browser checks for that and has a fork in the logic: https://searchfox.org/mozilla-central/source/browser/components/uitour/UITour.jsm#531

media/js/base/mozilla-fxa-link.js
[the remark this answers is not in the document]
vladikoff: I have adjusted the function to convert the values for the proper case. Was there anything more you had in mind for utm_ value validation? The `getAttributionData` function does a bit of validation already: 

https://github.com/mozilla/bedrock/blob/9e6ed7eaa873bb797dc9f3185a1cae86083e0ff8/media/js/base/fxa-utm-referral.js#L55-L56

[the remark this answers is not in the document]
alexgibson: Yeah, I think letting `getAttributionData` do its thing should be enough for utms 👍

docs/uitour.rst
```diff
@@ -48,6 +48,10 @@ Note that ``browser.uitour.testingOrigins`` can be a comma separated list of dom
     must also exclude the domain protocol e.g. ``https://``. A browser restart is also required
     after adding an allowed domain.
 
+    If you are working on Firefox Accounts integration, you can use the ``identity.fxaccounts.autoconfig.uri``
```
alexgibson, line 51: I've tried setting this to both `https://accounts.firefox.com.cn/` and `https://accounts.stage.mozaws.net/`, but I can't seem to get UITour to notice the change. Is there anything else I need to do, other than restart Firefox?

alexgibson: I've also tried setting `Services.prefs.getDefaultBranch("distribution.").setCharPref("id", "MozillaOnline");` in the browser console, to make Firefox act like the China Repack, but clicking the links I'm still seeing UITour open `https://accounts.firefox.com/`

media/js/base/mozilla-fxa-link.js
```diff
@@ -34,15 +41,66 @@ if (typeof window.Mozilla === 'undefined') {
...
+        var extraURLParams = {};
+        var allowedChars = /^[\w/.%-]+$/;
+        var knownParams = ['flowId', 'flowBeginTime', 'deviceId',
+            'entrypoint', 'entrypoint_experiment', 'entrypoint_variation',
+            'utm_source', 'utm_campaign', 'utm_content', 'utm_term', 'utm_medium'];
+
+        for (var i = 0; i < knownParams.length; i++) {
+            var known = knownParams[i];
+            // convert the param to snake_case for the UITour API
+            var knownFormat = known.replace(/([A-Z])/g, ' $1');
```
alexgibson, line 74: This works fine, but I notice when clicking the links that the final URL query string also contains snake_case params: `&flow_id=76df5964e3e15eae43a6ce13c751414d2fe3b7f96d55c4965bd61e296bbb045a&flow_begin_time=1594374070072&device_id=e801638be46f47d49076faca22eeac28`

Assuming that this format is accepted by the FxA server, I'd perhaps favor changing our default params to snake_case everywhere, as opposed to having to alter them here for this specific flow? You should be able to do that pretty easily here: https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-fxa-product-button.js#L80-L82

media/static-bundles.json
```diff
@@ -1253,14 +1236,12 @@
       "files": [
         "js/base/send-to-device.js",
         "js/libs/spin.min.js",
-        "js/base/uitour-lib.js",
```
alexgibson, line 1256: You may need to rebase this branch, but I still see an stray uitour-lib.js on line 294:

```
    {
      "files": [
        "js/base/uitour-lib.js",
        "js/firefox/new/desktop/download.js"
      ],
      "name": "firefox_desktop_download"
    },
```

alexgibson:
Test run was good https://gitlab.com/mozmeao/www-config/-/pipelines/165178826

media/js/base/mozilla-fxa-link.js
vladikoff: FxA accepts both cases: https://github.com/mozilla/fxa/blob/8a8138b20baf069d31e0a3b1968e4529faaaca0c/packages/fxa-content-server/app/scripts/models/flow.js#L51-L52

So that should be fine

docs/uitour.rst
vladikoff: image

@alexgibson could you check if your `identity.fxaccounts.remote.root` updated when you changed the autoconfig uri? 
I tried it locally in Nightly and it was not set, but after I updated and restarted it started working.

It is set in https://searchfox.org/mozilla-central/source/services/fxaccounts/FxAccountsConfig.jsm#282

media/js/base/mozilla-fxa-link.js
alexgibson: That's good to know - if we can change our default above to snake_case to avoid the extra conversion again here then that would help to keep things a bit simpler?

docs/uitour.rst
alexgibson: Yeah this seems to be the problem: my `identity.fxaccounts.remote.root` is still set to `https://accounts.firefox.com/`.

I can't seem to get it to change at all though. I've tried updating Nightly and restarting the browser, but no luck?

alexgibson: Ok, I was able to get this to work when trying our a new profile in Nightly, so I can confirm that the China Repack support still works as expected. No idea what's wrong with my regular profile still, but at least it seems to be related to my install.

vladikoff: Yea I was wondering if I should add that to the doc, it could be an issue in Firefox setting multiple config prefs in a row?? 😕 . I will add a note to the doc just so it is explained. I am happy you got it to work!

media/js/base/mozilla-fxa-link.js
vladikoff: I have updated the `flowParams` logic for that, it seems like that was the only spot that had to change.

alexgibson:
Deployed to demo: https://www-demo2.allizom.org/en-US/ (just a note for myself in the morning)

PR #8924, opened 2020-05-14
[fix #8836] Add whatsnew 77

### LINKED ISSUE #8836: Code - WNP77

craigcook:
## Description
WNP for Firefox 77.0 promoting Picture-in-Picture with a video of cuddly wuddly red pandas 🐼 

It's not possible to measure PiP use on a given web page (it doesn't send any kind of event we can detect) but there are aggregate stats on overall use of the feature in Firefox telemetry. So we're going to track video interactions on this page and try to correlate that with a hopeful uptick in feature usage. I'm using the Youtube API instead of a standard iframe embed so we can do event tracking.

I haven't added any functional tests at this point because really the only thing to test for is the video, and I'm not sure how to write a test for that or if it's even possible.

## Issue / Bugzilla link
#8836

## Testing
http://localhost:8000/firefox/77.0/whatsnew/all/
https://www-demo3.allizom.org/firefox/77.0/whatsnew/all/


alexgibson:
> I'm using the Youtube API instead of a standard iframe embed so we can do event tracking.

Do we need the YouTube API here to track play/pause events? We should be able to use navive HTML5 video here iiuc. It might be a little more work, but it could to avoid the possibility of YouTube registering as a tracker on page load (depending on your Firefox settings). On things like the home page, we defer loading the YouTube API until someone registers interest in wathcing a video - but that's not going to be possible here since we want the PiP icon to show.

I guess it also depends on if we can get the video source and upload it to the CDN?

craigcook:
True, I don't think there's any reason this has to be on youtube. Let's see if we can get the video hosted on our own CDN. Should we be worried about the bandwidth of hosting video on such a high-traffic page?

> On things like the home page, we defer loading the YouTube API until someone registers interest in wathcing a video - but that's not going to be possible here since we want the PiP icon to show.

If we end up sticking with youtube I think we could do the same here and load the API on click.

craigcook:
With some concerns about the cost of serving a lengthy video on a high-traffic page we opted to stick with youtube, but now the API doesn't load until a user clicks.

media/static-bundles.json
```diff
@@ -864,6 +870,14 @@
       ],
       "name": "firefox_whatsnew_76"
     },
+    {
+      "files": [
+        "js/base/uitour-lib.js",
```
alexgibson, line 875: Nit: do we need this?

alexgibson: Ah, I see why we need this (for the call in `whatsnew/up-to-date.js` 👍

PR #8936, opened 2020-05-20
Use mozmeao dockerhub repo.

glogiotatidis:
See https://github.com/mozmeao/infra/issues/1284


pmac:
Also these new repos in `mozmeao` don't yet exist. We should do that first right? Or can they be moved?

glogiotatidis:
> The changes look good. But I found more mentions of `mozorg/bedrock` in `.gitlab-ci.yml`.

TIL `rg` requires `--hidden` to search hidden files.

glogiotatidis:
> Also these new repos in `mozmeao` don't yet exist. We should do that first right? Or can they be moved?

`docker push` will create them, no prior action needed.

glogiotatidis:
> > Also these new repos in `mozmeao` don't yet exist. We should do that first right? Or can they be moved?
> 
> `docker push` will create them, no prior action needed.

We can move existing tags from `mozorg` to `mozmeao` if that's valuable?

pmac:
> `docker push` will create them, no prior action needed.

I had no idea it would do that. Neat.

> We can move existing tags from `mozorg` to `mozmeao` if that's valuable?

Nah. No need that I can think of.

pmac:
> TIL `rg` requires `--hidden` to search hidden files.

I used this tip I found in the ripgrep help:

```
Tip: to disable all smart filtering and make ripgrep behave a bit more like classical grep, use 'rg -uuu'.
```

glogiotatidis:
@pmac updated.

PR #8942, opened 2020-05-26
Move front-end linting and tests to NPM scripts (Fixes #8479)

### LINKED ISSUE #8479: Run CSS/JS linting and tests via npm scripts
## Description

Given that most code editors lint CSS/JS on-the-fly today, without requiring any project specific build scripts, it may be worth migrating Gulp tasks for ESLint, Stylelint, and Karma to run as simple npm scripts instead.

Doing this may reduce the number of npm dependencies we currently have, and also simplify/lighten our gulpfile.

---

- [x] Lint / test gulp tasks are migrated to run as npm scripts in CI.

alexgibson:
## Description
Begins to reduce our dependency on Gulp by moving JS tests and JS/CSS/JSON linting from Gulp to NPM scripts.

- Adds NPM scripts for linting and testing front-end assets.
- Removes `gulp-eslint`, `gulp-stylelint`, and `gulp-jsonlint` plugins.
- Updates `eslint` to version 7.
- Uses ESLint plugin to lint JSON files.
- Updates docker images to use Node v12.

## Issue / Bugzilla link
#8479

## Testing
Make sure to update dependencies by running `yarn` and `make build` first.

- Running `npm run test` should lint front end assets before running JS unit tests.
- Running `make lint` should lint both front-end and back-end code.

Successful test run: https://gitlab.com/mozmeao/www-config/pipelines/149677839

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,7 @@
 ########
 # assets builder and dev server
 #
-FROM node:8-slim AS assets
+FROM node:12-slim AS assets
 
 ENV PATH=/app/node_modules/.bin:$PATH
 WORKDIR /app
@@ -11,11 +11,12 @@ COPY package.json yarn.lock ./
 
 # install dependencies
 RUN yarn install --pure-lockfile
-RUN yarn global add gulp-cli@2.0.1
+RUN yarn global add gulp-cli@2.2.1
 
 # copy supporting files and media
-COPY .eslintrc.js .stylelintrc gulpfile.js ./
+COPY .eslintrc.js .eslintignore .stylelintrc .stylelintignore gulpfile.js ./
 COPY ./media ./media
+COPY ./tests/unit ./tests/unit
 
 RUN gulp build --production
 
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,7 @@ help:
 	@echo "  clean         - remove all build, test, coverage and Python artifacts"
 	@echo "  rebuild       - force a rebuild of all of the docker images"
 	@echo "  submodules    - resync and fetch the latest git submodules"
-	@echo "  lint          - check style with flake8, jshint, and stylelint"
+	@echo "  lint          - check style with flake8, eslint, and stylelint"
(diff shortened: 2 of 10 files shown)


Dockerfile
```diff
@@ -1,7 +1,7 @@
 ########
 # assets builder and dev server
 #
-FROM node:8-slim AS assets
+FROM node:12-slim AS assets
```
alexgibson, line 4: @pmac note I had to bump our node docker image here, as it was still running v8. Node 12 is the current LTS version, according to https://nodejs.org/en/.

I pushed this for a test run, and all seems to be ok? https://gitlab.com/mozmeao/www-config/pipelines/149677839

pmac: Sounds good 👍

stephaniehobson:
I'm getting this error when I run `yarn` in the virtual environment `error eslint@7.0.0: The engine "node" is incompatible with this module. Expected version "^10.12.0 || >=12.0.0". Got "8.16.0"`

We might need to list it as a requirement explicitly in `pip install -r requirements/dev.txt`?

stephaniehobson:
I broke my local dev so I'm not sure the comment I left is valid 😬

alexgibson:
@stephaniehobson looks like you’re running an outdated version of node, which is installed globally not part of the python dependencies. I’d suggest trying to download the latest long term support (LTS) version, which right now is v12: https://nodejs.org/en/

Makefile
```diff
@@ -95,7 +95,7 @@ clean:
 
 lint: .docker-build-pull
 	${DC} run test flake8
-	${DC} run assets gulp js:lint css:lint json:lint
+	${DC} run assets npm run lint
 
 test: .docker-build-pull
 	${DC} run test
```
pmac, line 101: We could also add `${DC} run assets npm run test` here

alexgibson: I don’t think we can do that as the tests require browsers, which aren’t in the container?

pmac: AH! I forgot about that. We can figure that out later.

PR #8953, opened 2020-05-28
Remove autofill from /switch page (Fixes #8651)

### LINKED ISSUE #8651: Remove "autofill" from switch page.
## Description

We should cut "autofill" from the switch page copy as it is not data we import through the switching process. Here's the [supporting documentation](https://wiki.mozilla.org/QA/Firefox_migrators#Supported_data_types).

---

- [ ] New page is launched with improved copy.

alexgibson:
## Description
Removes mention of "autofill" from /firefox/switch/ page as it's something we no longer import.

## Issue / Bugzilla link
#8651

## Testing
- [ ] Both page description and page title should no longer mention "autofill"

### DIFF
--- a/l10n/en/firefox/switch.ftl
+++ b/l10n/en/firefox/switch.ftl
@@ -5,7 +5,9 @@
 ### URL: https://www-dev.allizom.org/firefox/switch/
 
 switch-switch-from-chrome = Switch from { -brand-name-chrome } to { -brand-name-firefox } in just a few minutes
+switch-switching-to-firefox-is-fast-updated = Switching to { -brand-name-firefox } is fast, easy and risk-free, because { -brand-name-firefox } imports your bookmarks, passwords and preferences from { -brand-name-chrome }.
 switch-switching-to-firefox-is-fast = Switching to { -brand-name-firefox } is fast, easy and risk-free, because { -brand-name-firefox } imports your bookmarks, autofills, passwords and preferences from { -brand-name-chrome }.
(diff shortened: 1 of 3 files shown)


l10n/en/firefox/switch.ftl
```diff
@@ -5,7 +5,9 @@
 ### URL: https://www-dev.allizom.org/firefox/switch/
 
 switch-switch-from-chrome = Switch from { -brand-name-chrome } to { -brand-name-firefox } in just a few minutes
+switch-switching-to-firefox-is-fast-updated = Switching to { -brand-name-firefox } is fast, easy and risk-free, because { -brand-name-firefox } imports your bookmarks, passwords and preferences from { -brand-name-chrome }.
```
alexgibson, line 8: @peiying2 what's the recommended process for adding a new string and deprecating an older one?

Here I'm removing the word "autofill" from an existing string. Should I add a comment to convey that the old string is obsolete?

Also note I'm updating a string below to use the brand name for Facebook like you mentioned earlier.

peiying2: If we don't have a tagging process in place, ideally we have a new ID with the changed string. We just remove it and add a new ID/string. I am not sure it is good practice to use the same ID to update the string. Let me check what has been done with other fluent projects.

peiying2: @alexgibson New ID for revised string.

PR #8955, opened 2020-05-28
[no bug] WNP77, use self-hosted video for zh-CN

craigcook:
## Description
YouTube is blocked in China, so we'll show a self-hosted video to zh-CN. Other locales should still get the YouTube embed.

This is behind the switch `firefox-whatsnew77-video-zhCN` so in case we approach exceeding our bandwidth we can quickly turn it off and zh-CN should fall back to the WNP76 template.

The 77 release is **2020-06-02** so we need to get this into production before then. Marking as P1.

## Testing
http://localhost:8000/zh-CN/firefox/77.0/whatsnew/all/
https://www-demo3.allizom.org/zh-CN/firefox/77.0/whatsnew/all/

- [x] zh-CN gets the self-hosted video when the switch is on
- [x] zh-CN gets the WNP76 template when the switch is off

### DIFF
--- a/media/js/firefox/whatsnew/whatsnew-77-zhCN.js
+++ b/media/js/firefox/whatsnew/whatsnew-77-zhCN.js
@@ -1,0 +1,37 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    function trackVideoInteraction(title, state) {
+        window.dataLayer.push({
+            'event': 'video-interaction',
+            'videoTitle': title,
+            'interaction': state
+        });
+    }
+
+    function initVideoInteractionTracking() {
+        var video = document.getElementById('pip-video');
+
+        video.addEventListener('play', function() {
+            trackVideoInteraction(this.getAttribute('data-video-title'), 'video play');
+        }, false);
+
+        video.addEventListener('pause', function() {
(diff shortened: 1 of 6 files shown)


media/js/firefox/whatsnew/whatsnew-77-zhCN.js
```diff
@@ -0,0 +1,37 @@
...
+            'videoTitle': title,
+            'interaction': state
+        });
+    }
+
+    function initVideoInteractionTracking() {
+        var video = document.getElementById('pip-video');
+
+        video.addEventListener('play', function() {
+            trackVideoInteraction(this.getAttribute('data-video-title'), 'video play');
```
alexgibson, line 20: The title is coming through as `undefined` in the dataLayer event. Should this be `data-ga-label`?
```suggestion
            trackVideoInteraction(this.getAttribute('data-ga-label'), 'video play');
```

media/js/firefox/whatsnew/whatsnew-77-zhCN.js
```diff
@@ -0,0 +1,37 @@
...
+        }, false);
+
+        video.addEventListener('pause', function() {
+            var action = this.currentTime === this.duration ? 'video complete' : 'video paused';
+            trackVideoInteraction(this.getAttribute('data-video-title'), action);
```
alexgibson, line 25:
```suggestion
            trackVideoInteraction(this.getAttribute('data-ga-label'), action);
```

media/js/firefox/whatsnew/whatsnew-77-zhCN.js
craigcook: Oops, actually the attribute should be changed in the HTML. Copypasta error 🍝

Issue #8811, opened 2020-04-21
Update and relaunch Browser Comparison Pages


ejregithub:
## Description

We need to update the browser comparison pages to the current release of each competitor and reflect the latest features in them.

We should update the copy in each page as specified in copy docs. 

Here is the latest [comparison table](https://docs.google.com/spreadsheets/d/1dyxx-M27jpit4RKStXxR54kkBN1sa9Tbca2a8MionKc/edit#gid=1512926430) vetted by PMM (Tyler Ahn)

Here is a[ listing of links](https://docs.google.com/spreadsheets/d/1UrSn0qR24sI6cdhNVF2fri3aAQr5YdNC2l6DvvSUeIQ/edit#gid=0) to all the things germane to this work

We should go in this order:

* Update all the browser-specific pages
* Update the main page 
* Relaunch the content
* (Migrate to Fluent)
* Build additional pages (e.g. Brave)

Link to [decisions document](https://docs.google.com/document/d/1bOQV23eZDI7AFaRISv9w-fEH3ZPG8KDKk10dLN-QTiI/edit)

---

#### Success Criteria
- [ ] Content is updated & browser versions are prominent

#### Approvals
- [ ] (C) Product: Arthur Edelstein for PrivSec sections
- [ ] (C) Product: Rachel Tublitz & Romain Testard for utility & portability sections
- [ ] (C) - Brand Design: Clio
- [ ] (C) - Brand Copy: Natalie
- [ ] (C) - PMM: TBD
- [ ] (R) - Websites Team Lead: Hoosteeno
- [ ] (A) - TBD

hoosteeno:
@ejregithub Can you add a link to the doc where we captured our copy suggestions?

@rantley Can you update with status on this? How long until we have copy to code? We need to understand timing because our Fluent l10n system is coming online in 2 weeks.

ejregithub:
@hoosteeno - here's that doc (added in the description to this card as well)
https://docs.google.com/document/d/1bOQV23eZDI7AFaRISv9w-fEH3ZPG8KDKk10dLN-QTiI/edit

craigcook:
The pages are live as of 26 June 2020

PR #8958, opened 2020-05-29
Update and relaunch browser comparison pages

craigcook:
## Description
Restoring the previously removed/redirected pages, migrating them all to Fluent, and updating copy and info. Also adding a new page for Brave.

My first time doing Fluent migrations so be gentle.

## Issue / Bugzilla link
#8811 

## Testing
http://localhost:8000/firefox/browsers/compare/
https://www-demo1.allizom.org/firefox/browsers/compare/


craigcook:
This is awaiting stakeholder review to double-check for accuracy so it's still a do-not-merge for the moment, but we can start code review.

alexgibson:
Oh, also don't forget to add `[skip l10n]` to the commit message here as the migration will be handled manually.

l10n/en/brands.ftl
```diff
@@ -17,6 +17,7 @@
 -brand-name-mozilla-corporation = Mozilla Corporation
 -brand-name-mozilla-foundation = Mozilla Foundation
 -brand-name-twitter = Twitter
+-brand-name-github = GitHub
```
alexgibson, line 20: Should these be in alphabetical order, to match what was added to https://github.com/mozilla-l10n/www-l10n/blob/master/en/brands.ftl ?

alexgibson:
Approved and merge at will - although it may be a good idea to wait until the migration has been reviewed by Peiying, in the event that she suggests changes and you need to re-run it.

PR #8989, opened 2020-06-05
Add middleware to compress responses with gzip, brotli, or zstandard

### LINKED ISSUE #8983: HTML is not being served compressed
## Description

HTML should be compressed with gzip for perf reasons.

### Steps to reproduce

1) Open devtools  network tab
2) Visit www.mozilla.org
3) Click on the HTML request
4) Switch to the Headers tab
5) Look for "content-encoding" in the response.

### Expected result

Should be gzipped.

### Actual result

Not compressed

cc: @MEAOCloud

pmac:
Fix #8983


pmac:
Demo is up and seems to be working well. I'm going to also push to integration tests.

pmac:
Integration tests run https://gitlab.com/mozmeao/www-config/-/pipelines/153382489

Issue #8971, opened 2020-06-01
WNP 78 Code


ejregithub:
Code and launch the Whatsnew page for the Firefox 78 release.

Content: Mobile

[Copy - DE, EN and FR](https://docs.google.com/document/d/1XGljKf8hl88z56zHaqcerEy3gQlNH1C-PB-uhKqj08o/edit) for use in the test

[Releng bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1644222)

---

####  Success Criteria 
- [ ] Is fully approved and live in production on or before June 26th

#### Approvals
- [x] Accountable - LNiolet
- [x] Mobile PMM - LAustin
- [x] Brand (Design) - Clio
- [x] Brand (Copy) - Natalie
- [x] Team Lead - Hoosteeno

Errolyn:
@ejregithub Can you add the due date for this so I have it here?

ejregithub:
@Errolyn - done - the very last day (we should strive to have the content, all templates with copy test variations, live in prod sooner) for go live is June 26th.

PR #9033, opened 2020-06-22
Whatsnew 78 [fixes: 8971]

Errolyn:
## Description

Add what new page, the default is our current default page and there will be 4 variations on that design we are trying.

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/8971


bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx78-b-de.html
```diff
@@ -0,0 +1,69 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros.html" import send_to_device with context %}
+
+{% extends "firefox/whatsnew/base.html" %}
```
alexgibson, line 7: There's quite a bit of code duplication in the templates here for each variation. As far as I can tell, only the title is changing - so would it be less code to write / maintain if we extended the control template and then update the title using a block?

```jinja
{# This Source Code Form is subject to the terms of the Mozilla Public

{% from "macros.html" import send_to_device with context %}

{% extends "firefox/whatsnew/whatsnew-fx78-a-de.html" %}

{% block page_heading %}Privat unterwegs? Hol dir Firefox für Mobile.{% endblock %}
```

bedrock/firefox/views.py
```diff
@@ -591,6 +591,33 @@ def get_template_names(self):
                 template = 'firefox/whatsnew/index.html'
         elif locale == 'id':
             template = 'firefox/whatsnew/index-lite.id.html'
+        elif version.startswith('78.'):
+            if variation == '2' and locale in ['en-US', 'en-CA', 'en-GB']:
```
alexgibson, line 595: There's a lot of logic here that's quite long winded. It also feels a little confusing for me to work out which template I'm seeing, when we use `?v=2` in the URL but `a` in the template name as an identifier? 

I'd suggest renaming the templates to be consistent with the variation parameter. Then we could also cut down on a lot of this logic using something like:

```python
variations = ['2', '3', '4', '5']
locales = ['en-US', 'en-CA', 'en-GB', 'de', 'fr']

if variation in variations and locale in locales:
    locale = locale.split('-')[0]
    template = 'firefox/whatsnew/whatsnew-fx78-{0}-{1}.html'.format(variation, locale)
else:
```

Adding some unit tests for the above would be nice too.

alexgibson:
At smaller screen sizes, I see an empty white line that appears underneath the main content that stretches the width of the viewport:

image

media/js/firefox/whatsnew/whatsnew-exp-78.js
```diff
@@ -0,0 +1,62 @@
...
+                    'v=5': 20,
+
+                }
+            });
+
+            cop.init();
+        }
+    };
+
+    if (!isMobile) {
```
alexgibson, line 58: Do we need this check here, seeing as it's a /whatsnew page?

media/js/firefox/whatsnew/whatsnew-exp-78.js
```diff
@@ -0,0 +1,62 @@
...
+
+    var initTrafficCop = function () {
+        if (href.indexOf('v=') !== -1) {
+            if (href.indexOf('v=1') !== -1) {
+                window.dataLayer.push({
+                    'data-ex-variant': 'whatsnew-78-control',
+                    'data-ex-name': 'whatsnew-78-Experiment'
+                });
+            } else if (href.indexOf('v=2') !== -1) {
+                window.dataLayer.push({
```
alexgibson, line 21: This script needs to be included in all the variation templates as well. currently it's only added to index.html, which means none of these dataLayer events get fired.

bedrock/firefox/views.py
```diff
@@ -591,6 +591,33 @@ def get_template_names(self):
...
+                template = 'firefox/whatsnew/whatsnew-fx78-a-en.html'
+            elif variation == '3' and locale in ['en-US', 'en-CA', 'en-GB']:
+                template = 'firefox/whatsnew/whatsnew-fx78-b-en.html'
+            elif variation == '4' and locale in ['en-US', 'en-CA', 'en-GB']:
+                template = 'firefox/whatsnew/whatsnew-fx78-c-en.html'
+            elif variation == '5' and locale in ['en-US', 'en-CA', 'en-GB']:
+                template = 'firefox/whatsnew/whatsnew-fx78-d-en.html'
+            elif variation == '2' and locale == 'de':
```
alexgibson, line 603: When I set `Dev=False` in my `.env` and try to load this URL I get redirected to en-US because the locale isn't activated:

http://localhost:8000/de/firefox/78.0/whatsnew/all/?v=2

Adding the following to the templates seems like the simplest fix for the experiment variations:

```jinja
{% add_lang_files "firefox/whatsnew" %}
```

bedrock/firefox/templates/firefox/whatsnew/index.html
```diff
@@ -77,3 +77,9 @@ <h1 class="main-title">{{ _('Download Firefox for your smartphone and tablet.')
 {% block js %}
   {{ js_bundle('firefox_whatsnew') }}
 {% endblock %}
+
+{% block experiments %}
+  {% if switch('whatsnew-78-exp') %}
```
alexgibson, line 82: We should also add the supported locales to the switch here, to avoid non-experiment locales from being redirected.

alexgibson:
This may need a rebase after https://github.com/mozilla/bedrock/pull/9041 merged

media/js/firefox/whatsnew/whatsnew-exp-78.js
```diff
@@ -0,0 +1,57 @@
...
+                    'data-ex-variant': 'whatsnew-78-v1',
```
alexgibson, line 20: Should this be `v2`? Same applies to the res of the variants below, as they seem to be out by 1.

PR #9047, opened 2020-06-24
Only show browser comparison pages in DEV mode

craigcook:
**Note:** we should only merge this **if** we need to deploy to production **before** we get a go-ahead on the comparison pages. We can also just wait to push prod.

## Description
The browser comparison pages are finished, but it was only after they merged that I learned we're still waiting for a final go/no-go from leadership before publishing them, and we won't get word until tomorrow (25 June) at the earliest. 

Since the pages already merged we're effectively blocked from doing a production push until we have permission to relaunch the comparison pages. However, l10n is already in progress so we need to make sure the pages are still available on dev/demo, so I don't want to revert the commit or redirect the URLs. I considered putting the pages behind a switch but I think that would entail routing them through a view (don't think we can do switches in urls.py) and this dev mode rule was simpler.

tl;dr: this is a quick-and-dirty hack just to unblock us from pushing production. We might not even need it.

## Testing
http://localhost:8000/firefox/browsers/compare/ (and sub-pages)

- [ ] Pages should load normally when `DEV=True`
- [ ] Pages should redirect to /firefox when `DEV=False`

### DIFF
--- a/bedrock/firefox/urls.py
+++ b/bedrock/firefox/urls.py
@@ -6,6 +6,8 @@ from django.conf.urls import url
 import bedrock.releasenotes.views
 from bedrock.mozorg.util import page
 from bedrock.releasenotes import version_re
+from django.conf import settings
+from bedrock.redirects.util import redirect
 
 from bedrock.firefox import views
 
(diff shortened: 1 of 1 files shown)


craigcook:
Looks like we're gonna need this after all!

Issue #9010, opened 2020-06-10
Make adding new brand names to brands.ftl easier


alexgibson:
## Description

Currently adding a new brand name to brands.ftl involves a set of very laborious steps:

1. Add a new brand name to en/brands.ftl in bedrock.
2. Take that brand name and add it to every locale in the l10n-team repo.
3. Once that is merged in the l10n-team repo, pull it back into bedrock via l10n_update.

This is required because brand name placables do not fall back to English like a regular string if they do not exist for a locale.

The above takes a front end developer several hours of laborious copy / paste for every new brand name added.

We should make this easier to automate, or alternatively change the default behaviour to fall back to English.

pmac:
With #9048 merged I'm going to mark this as closed. I don't think we need this automation if english fallbacks work on the site. Please reopen if you disagree.

PR #9048, opened 2020-06-24
Add a fallback to the en version of a brand term for Fluent strings

pmac:
In Fluent each term you use in a string has to be availalbe in that locale bundle. This is a bit of a hack around that. This change looks for the presence of a term ID in the translated string, and if it sees one it will add the string as a new ID to a new translation context with the en brands.ftl loaded so that it will attempt to use the default english terms.

I'll be adding a new term to a page to test and make sure it doesn't show in the translations and deploy a demo.

### DIFF
--- a/lib/l10n_utils/fluent.py
+++ b/lib/l10n_utils/fluent.py
@@ -8,7 +8,7 @@ from django.core.cache import caches
 from django.utils.encoding import force_bytes
 from django.utils.functional import lazy, cached_property
 
-from fluent.runtime import FluentLocalization, FluentResourceLoader
+from fluent.runtime import FluentLocalization, FluentResource
 from fluent.syntax.ast import GroupComment, Message
 
 from lib.l10n_utils import translation
@@ -25,6 +25,7 @@ __all__ = [
 ]
 cache = caches['fluent']
 REQUIRED_RE = re.compile(r'^required\b', re.MULTILINE | re.IGNORECASE)
+TERM_RE = re.compile(r'\{\s*-[a-z-]+\s*\}')
 
 
 class FluentL10n(FluentLocalization):
@@ -92,6 +93,28 @@ class FluentL10n(FluentLocalization):
             return True
 
         return message_id in self._localized_message_ids
+
+
+class FluentResourceLoader:
+    """A resource loader that will add english brand terms to every bundle"""
+    @staticmethod
+    def resources(locale, resource_ids):
+        for root in settings.FLUENT_PATHS:
+            resources = []
+            for resource_id in resource_ids:
+                path = root.joinpath(locale, resource_id)
+                if not path.is_file():
+                    continue
+
+                with path.open(encoding='utf-8') as ftl_file:
+                    resources.append(FluentResource(ftl_file.read()))
+            if resources:
+                if locale != 'en':
+                    path = settings.FLUENT_LOCAL_PATH.joinpath('en', 'brands.ftl')
+                    with path.open(encoding='utf-8') as ftl_file:
+                        resources.append(FluentResource(ftl_file.read()))
+
+                yield resources
 
 
 def _cache_key(*args, **kwargs):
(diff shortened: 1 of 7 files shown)


pmac:
@alexgibson this is potentially a fix for #9010 that doesn't involve manually updating all of the locale files.

pmac:
Actually, deploying a demo for this will be tough since it will require adding a page to use a new string and a non-en FTL file to test a translation. What I did to test locally was to update locales, then in `git-repos/www-l10n/de/brands.ftl` I deleted lines for `-brand-name-firefox` and others. English worked fine, but the term IDs showed up in the German translation until I implemented this fix.

pmac:
I've got some tests to add/update, but it should be good for someone to pull down and manually test.

pmac:
@Pike you may want to have a look at this too. It's definitely a hack, but it works pretty well for our case, and I think is significantly better than having to write automation to handle all of the files in all of the locales.

pmac:
Okay. Tests added/updated.

pmac:
That's how all of the other strings work right? It's just an FTL file so my assumption was that it would work like all the others.

Pike:
This is quite a hack, and so much that it took @stasm 5 minutes to figure out what it actually does.

Somewhere probably only in aural conversations, I have a proposal that language fallback on Term references would be working across languages, but only w/out logic.

Under the assumption that the `en` brands.ftl is w/out logic, we could just add it as last resource to reach bundle that wants branding.

This does effectively the same thing, but in a way that's less surprising to us Fluent folks. It would also not be tied to the exact visual presentation we use for missing terms. Like, it's more coincidence than design that this works as it does.

pmac:
> Under the assumption that the en brands.ftl is w/out logic, we could just add it as last resource to reach bundle that wants branding.

Interesting. So if I just add the english `brands.ftl` to the end of every bundle it will use those strings only if it's not in the localized `brands.ftl` earlier in the bundle?

Pike:
> Interesting. So if I just add the english `brands.ftl` to the end of every bundle it will use those strings only if it's not in the localized `brands.ftl` earlier in the bundle?

Yeah, that happens here: https://github.com/projectfluent/python-fluent/blob/aa7d65eeeae767bbfef0e961dffdddb90fc99abd/fluent.runtime/fluent/runtime/__init__.py#L67-L68

pmac:
Ah right! `allow_override` is `False` so it keeps the first value for the string ID it gets 👍

pmac:
I'll update this and it will be WAY less hacktastic 😄

pmac:
Update is done. I like this a lot better. It simply adds the `en/brands.ftl` to every non-en bundle via the resource loader.

Pike:
Just to make sure about the edge cases, do we have localizations that don't include `brand`?

But yeah, this is a lot more fluent-y. Thanks for the update.

pmac:
> Just to make sure about the edge cases, do we have localizations that don't include `brand`?

Not following. Is the question whether there are locales that don't have a `brands.ftl` file? I don't think there are, and I don't think it matters even if they don't. Before this change it would matter because the translations wouldn't be able to find the term, but now they can even if there are no terms from `brands.ftl` in the locale.

Pike:
The question is if we add `brands.ftl` to every page, or if some pages opt out of that.

pmac:
Oh! It's on the list of default fluent files that are loaded for every page along with ones for e.g. the nav and footer. 

https://github.com/mozilla/bedrock/blob/318b49c4b4df40cde4fc58938e642b52560f3cd4/bedrock/settings/base.py#L248-L257

Pike:
'k, then this looks good.

Saying that, I didn't actually review the tests.

pmac:
Thanks for taking a look and for the suggestion for the much better method for fixing this.

pmac:
> One question - Is it worth adding an additional test to make sure that if a locale _does_ have a [translated brand](https://github.com/mozilla-l10n/www-l10n/blob/master/es-MX/brands.ftl#L26) term, then that translation is used instead of the english term?

Yes. Good call. I'll add that.

pmac:
Test added. I think we can merge after the tests run.

alexgibson:
Thanks @pmac !

PR #9051, opened 2020-06-24
Add firefox/accounts & fr /firefox exp templates. [no bug]

amychurchwell:
## Description
Adds firefox/accounts and the french /firefox page to exp templates to enable convert testing.
While doing this I noticed some exp templates that were missing open graph share urls.

## Issue / Bugzilla link
N/A

## Testing
[Creating exp page docs](https://bedrock.readthedocs.io/en/latest/abtest.html#creating-experimental-pages)


bedrock/exp/templates/exp/firefox/accounts.html
```diff
@@ -0,0 +1,200 @@
...
+{% from "macros.html" import fxa_email_form with context %}
+
+{% extends "firefox/base/base-protocol.html" %}
+
+{% set _entrypoint = 'mozilla.org-firefox-accounts' %}
+{% set _utm_source = 'mozilla.org-firefox-accounts' %}
+{% set _utm_campaign = 'trailhead' %}
+
+{%- block page_title -%}
+  {{ ftl('firefox-accounts-there-is-a-way-to', fallback='firefox-accounts-get-a-firefox-account') }}
```
amychurchwell, line 15: note for reviewer: I know that our exp templates should not be a part of our localization system-- is there a way to retrieve the strings from these fluent ID's? (have not had a chance to interact with the new fluent system very much)

alexgibson: @amychurchwell the strings for the associated page can be found here: https://github.com/mozilla/bedrock/blob/master/l10n/en/firefox/accounts.ftl

It's up to you if you want to go as far as replacing them really. The docs are just meant as a guide to reinforce that experimental pages can be standalone and lower maintenance.

amychurchwell: Ah, okay. I'll leave this as-is then. Thank you.

PR #9057, opened 2020-06-25
Use new string ID for facebook share string (Fixes #9056)

### LINKED ISSUE #9056: Fluent string update requires a new ID
## Description

Describe the issue that you're seeing.
Fixing a bug should involve adding a new string ID, not making changes with existing ID. By reusing the same ID, there is no warning when the placeable or variable is different between source and target language.

### Steps to reproduce
(I missed this one during the review): firefox/switch.ftl
https://bugzilla.mozilla.org/show_bug.cgi?id=1648377
- switch-share-to-facebook = Share to Facebook
+ switch-share-to-facebook = Share to { -brand-name-facebook }

### Expected result

Add a new string and remove the old one. Or add an updated string. 

@alexgibson

alexgibson:
## Description
- Reverts a string change in `switch.ftl` and adds a new string ID instead.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/9056

## Testing
- [ ] Page should still display the same text, but using a brand name placeable for "Facebook", falling back to the original string.


peiying2:
Thanks for the fix. LGTM.

PR #9065, opened 2020-06-25
Add brand term detection and replacement to fluent recipe migration script

### LINKED ISSUE #8612: Enhance Fluent migration script to interpolate brand names
## Description

The current Fluent migration script automatically handles things like variables when creating migration recipies, but it does not yet incorporate things like placeables for [brand names](https://github.com/mozilla/bedrock/blob/master/l10n/en/brands.ftl). These have to be done manually (see https://github.com/mozilla/bedrock/pull/8611)

We should update [fluent_recipe.py](https://github.com/mozilla/bedrock/blob/master/lib/l10n_utils/management/commands/_fluent_recipe.py#L126) to handle replacing common brand names with placeables automatically when creating recipies.

pmac:
Fix #8612

## Testing

I ran the following command to test:

```bash
./manage.py fluent recipe bedrock/firefox/templates/firefox/nightly_whatsnew.html
```

### DIFF
--- a/docs/fluent-conversion.rst
+++ b/docs/fluent-conversion.rst
@@ -49,9 +49,10 @@ The existing recipe will already have the template name as prefix, though.
 You can also choose to remove strings from the conversion, if you just
 want to convert a subset of the strings.
 
-If you want to handle things such as updating existing translations to use
-brand terms as placeholders, you can update the recipe to replace existing
-string content by using the ``REPLACE()`` helper:
+The migrations will automatically add ``REPLACE()`` helpers when it detects
+that any of the strings in ``brands.ftl`` are in the text. Make sure these
+are correct, and remove any of them that shouldn't be used. If you need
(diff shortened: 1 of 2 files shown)


pmac:
Ah yes. Great catch.... Perhaps we need to look for the strings in length order and remove the match from the string before checking for others. I'll try to make that happen. Thanks for the thorough explanation!

alexgibson:
> Perhaps we need to look for the strings in length order and remove the match from the string before checking for others.

I think that might work yeah, great! 👍

pmac:
@alexgibson I've done those updates and it seems to be working much better. Let me know what you think.

pmac:
Doh! Should be fixed now.

pmac:
Oh! You mentioned the docs. Forgot about that. I'll update them quickly before we merge.

pmac:
Docs updated.

Issue #8515, opened 2020-02-12
Default homepage picto card responsive styles bug.


amychurchwell:
## Description

![Screen Shot 2020-02-12 at 9 01 57 AM](https://user-images.githubusercontent.com/30483988/74358589-8f11a780-4d76-11ea-9d19-9766516ad77a.png)

Picto card list items break formation at certain screen widths.
example: https://www.mozilla.org/hi-IN/

alexgibson:
On closer inspection this seems to be a float clearing issue with the content above, rather than an issue with the picto card styles themselves. Giving the section directly above a class name like `impact-innovations` and then clearing the floated content in that section should force the picto cards to wrap onto the next line:

```css
.impact-innovations {
    @include clearfix;
}
```

mitchell-frost:
Hey! I am new to mozilla. Can I work on this?

codebreakeragr8:
@amychurchwell @alexgibson Hey! If this issue is still open then i would like to work on it.

alexgibson:
@codebreakeragr8 @mitchell-frost thanks for showing interest in this bug. Pull requests are always welcome, and contributing docs are here: https://github.com/mozilla/bedrock/blob/master/.github/CONTRIBUTING.md

Rudresh22:
Hello sir, I want to work on this issue. It would be great if you could guide me through this.

craigcook:
Sorry, this is already being worked on (I hadn't seen this issue when I opened the PR). It'll be fixed with https://github.com/mozilla/bedrock/pull/9089

alexgibson:
Fixed in https://github.com/mozilla/bedrock/pull/9089

PR #9089, opened 2020-06-29
[fix bug #1640569] home page, clear floats in the two column section

craigcook:
## Description
The two column "Impact and Innovations" section can collide with floats in the picto card section below when they're not the same height. This adds a clearfix to the column container to prevent that.

The bug was filed by a localizer working on the Taiwanese version but it likely happens in other locales as well.

image

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1640569

## Testing
http://localhost:8000/zh-TW/

The row of picto cards should line up correctly because the two columns above them are now contained.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/home/home.html
+++ b/bedrock/mozorg/templates/mozorg/home/home.html
@@ -74,7 +74,7 @@ accessible to all.') }}
     image_url='img/home/dino.svg'
     )}}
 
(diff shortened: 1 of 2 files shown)


craigcook:
Looks like this was also reported in https://github.com/mozilla/bedrock/issues/8515

PR #8741, opened 2020-04-02
Header logos on 404 and 500 pages should be links (fix bug 8716)

imrishabh18:
## Description
Added the link to mozilla homepage for 404 and 500 pages

## Issue / Bugzilla link
#8716

PR #9092, opened 2020-06-30
Manoj71 patch 1

manoj71:

### DIFF
--- a/bedrock/base/templates/base-error.html
+++ b/bedrock/base/templates/base-error.html
@@ -12,10 +12,10 @@
 {% block site_header %}
   <header class="mzp-l-content header-spacing">
     <div class="header-image">
-    <a href=""https://www.firefox.com">{{ high_res_img('img/logos/firefox/logo-reality.png', {'alt': 'Firefox'}) }}</a>
+    <a href="https://www.firefox.com">{{ high_res_img('img/logos/firefox/logo-reality.png', {'alt': 'Firefox'}) }}</a>
     </div>
     <div class="header-image">
-      <a href=""https://www.mozilla.org"> {{ high_res_img('img/logos/mozilla/black.svg', {'alt': ' Mozilla'}) }}</a>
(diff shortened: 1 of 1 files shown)


bedrock/base/templates/base-error.html
```diff
@@ -12,8 +12,10 @@
 {% block site_header %}
   <header class="mzp-l-content header-spacing">
     <div class="header-image">
-    {{ high_res_img('img/co-brand/firefox_moz_co-brand_sm.png', {'alt': 'Firefox, Mozilla'}) }}
+    <a href="https://www.firefox.com">{{ high_res_img('img/logos/firefox/logo-reality.png', {'alt': 'Firefox'}) }}</a>
```
alexgibson, line 15: This is the logo for Firefox Reality browser, which is incorrect as it's not the regular Firefox logo. Also, a high-res version of this logo doesn't exist, so it's currently 404 in the page.

I'd suggest using `protocol/img/logos/firefox/logo-word-hor-xs.png`

The `href` should use our URL helper:

```
href="{{ url('firefox') }}"
```

bedrock/base/templates/base-error.html
```diff
@@ -12,8 +12,10 @@
...
     </div>
+    <div class="header-image">
```
alexgibson, line 17: This is missing a closing `</div>`

bedrock/base/templates/base-error.html
```diff
@@ -12,8 +12,10 @@
...
+      <a href="https://www.mozilla.org"> {{ high_res_img('img/logos/mozilla/black.svg', {'alt': ' Mozilla'}) }}</a>
```
alexgibson, line 18: There's no need to use `high_res_img()` helper for an svg asset.

I'd suggest using `protocol/img/logos/mozilla/black.svg`

The `href` should also use the URL helper `href="{{ url('mozorg.home') }}"`

alexgibson:
@manoj71 it looks like someone is already working on this bug and is a little further along with their changes: https://github.com/mozilla/bedrock/pull/8741

I'm going to close this pull request and we'll see if we can take the other one to the finish line. Thanks again for trying!

PR #9097, opened 2020-07-01
Simplify site.js by removing ancient platforms (Fixes #8454)

### LINKED ISSUE #8454: Simplify site.js platform detection logic
## Description

We have a global script called [site.js](https://github.com/mozilla/bedrock/blob/master/media/js/base/site.js) that sits in the `<head>` of every web page. This script is used for platform detection, and acts as a hook for displaying the correct link in our download buttons.

Much of the logic in here is from old, legacy code. Some stretching back to the old PHP site. We're likely in a position whereby we can now remove and simplify some of what's here. For example:

- Can we remove the detection for [very old versions of Windows](https://github.com/mozilla/bedrock/blob/master/media/js/base/site.js#L14-L36)? Does the TLS certificate for www.mozilla.org today even support these old clients? how much traffic do we see from them?
- Can we remove [Firefox OS detection](https://github.com/mozilla/bedrock/blob/master/media/js/base/site.js#L68-L70)? This isn't used as a hook anywhere on the site any more, so the answer is likely yes.
- Can we remove detection for [old MacOS versions](https://github.com/mozilla/bedrock/blob/master/media/js/base/site.js#L47-L52)? Also [here](https://github.com/mozilla/bedrock/blob/master/media/js/base/site.js#L62-L67)?.
- Do we still need [64bit detection for Windows](https://github.com/mozilla/bedrock/blob/master/media/js/base/site.js#L164-L167)? This code has never been enabled, since we know it's not always acurate, and the stub installer handles serving 32/64bit binaries much more accurately.

Tidying up some of this old code may help us move toward migrating to [Client Hints](https://github.com/mozilla/bedrock/issues/8440) more easily. Making this script smaller also has performance benefits since it is a blocking request.

---

- [ ] Legacy code is removed from site.js

- [ ] Results in download buttons being shown to really old clients, even if they may not be able to install / run Firefox.

alexgibson:
## Description

Demo: https://www-demo1.allizom.org/en-US/

- Removes "unsupported" messaging from download buttons:
- Defaults unknown systems to "undetermined" messaging
- Removes unused detection for Firefox OS.
- Removes unused logic for Windows 64bit download links. (the stub installer handles detection)
- Removes detection for Android architecture (download buttons all link to the Play Store, so this is now redundant?)

## Issue / Bugzilla link
#8454

## Testing
Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/162469319

### DIFF
--- a/media/css/pebbles/components/_buttons-download.scss
+++ b/media/css/pebbles/components/_buttons-download.scss
@@ -95,8 +95,6 @@ ul.download-list {
     .ios-download,
     .linux-arm-download,
     .unrecognized-download,
-    .unsupported-download,
-    .unsupported-download-osx,
     .nojs-download {
         display: none;
     }
@@ -108,8 +106,6 @@ ul.download-list {
 .download-button .os_win64-aarch64,
 .download-button .os_linux,
 .download-button .os_linux64,
-// Uncomment this to enable the win64 download buttons:
-// .win7up.x86.x64 .download-button .os_win,
 .android .download-button-desktop,
 .windows.arm .download-button .os_win,
 .linux.arm .download-button .os_linux,
@@ -124,8 +120,6 @@ ul.download-list {
     display: none !important;
 }
 
-// Uncomment this to enable the win64 download buttons:
-// .win7up.x86.x64 .download-button .os_win64,
 .linux .download-button .os_linux,
 .linux.x86.x64 .download-button .os_linux64,
 .windows .download-button .os_win,
@@ -134,27 +128,21 @@ ul.download-list {
 .android .download-button .os_android,
 .download-button-android .os_android,
 .android .download-button-desktop .download-list,
-.android .download-button-desktop small.os_win,
 .download-button-ios .os_ios,
 .ios .download-button .os_ios,
 .ios .download-button .ios-download,
 .ios .download-button-desktop .download-list,
-.other .download-button-android .download-list,
-.other .download-button small.os_win {
+.other .download-button-android .download-list {
     display: block !important;
 }
 
-.linux.arm .download-button .linux-arm-download,
-.oldwin .download-button .unsupported-download,
-.oldmac .download-button .unsupported-download {
+.linux.arm .download-button .linux-arm-download {
     display: block;
     max-width: 250px;
 }
 
 // Hide the privacy link if platform is unsupported.
-.linux.arm .download-button .fx-privacy-link,
-.oldwin .download-button .fx-privacy-link,
-.oldmac .download-button .fx-privacy-link {
+.linux.arm .download-button .fx-privacy-link {
     display: none;
 }
 
@@ -170,22 +158,6 @@ ul.download-list {
     .unrecognized-download {
         display: block;
     }
-}
-
-// Android architecture detection
-.download-button .download-list .os_android.x86,
-.download-button .download-other.os_android .arm,
-.android.x86 .download-button .download-list .os_android.armv7up,
-.android.x86 .download-button .download-other.os_android .x86 {
-    display: none !important;
-}
-
-.android.x86 .download-button .download-list .os_android.x86 {
-    display: block !important;
-}
-
-.android.x86 .download-button .download-other.os_android .armv7up {
-    display: inline !important;
 }
 
 /* stylelint-enable */
--- a/media/js/base/site.js
+++ b/media/js/base/site.js
@@ -6,36 +6,11 @@
     'use strict';
     window.site = {
         getPlatform: function (ua, pf) {
-            // Firefox OS navigator.platform is an empty string, which equates to a falsey value in JS
-            // Ths means we must use an ugly ternary statement here to make testing easier.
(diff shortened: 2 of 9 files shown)


media/css/pebbles/components/_buttons-download.scss
```diff
@@ -123,38 +119,32 @@ ul.download-list {
...
-// Uncomment this to enable the win64 download buttons:
-// .win7up.x86.x64 .download-button .os_win64,
 .linux .download-button .os_linux,
 .linux.x86.x64 .download-button .os_linux64,
 .windows .download-button .os_win,
 .osx .download-button .os_osx,
 .android .download-button .os_android,
 .download-button-android .os_android,
 .android .download-button-desktop .download-list,
-.android .download-button-desktop small.os_win,
```
alexgibson, line 135: Note for reviewer: these selectors with `small.os_win` in them are from back when there was a list of links underneath the download buttons. These are now redundant

media/css/pebbles/components/_buttons-download.scss
```diff
@@ -172,20 +162,4 @@ ul.download-list {
     }
 }
 
-// Android architecture detection
```
alexgibson, line 175: As far as I can tell all Android download buttons now point to the Play Store, so these selectors aren't used anywhere.

media/js/base/site.js
```diff
@@ -110,12 +70,6 @@
                 return 'armv8';
             }
 
-            // PowerPC
```
alexgibson, line 113: We don't use this check for PowerPC processors anywhere, so removing it.

media/js/base/site.js
```diff
@@ -155,16 +109,11 @@
...
         if (platform === 'windows') {
-            // Add class to allow Windows XP/Vista users to download Firefox 52 ESR, though these
-            // legacy platforms are now officially unsupported
+            // Add class for Windows XP/Vista users to display
+            // unsupported messaging on /download/thanks/ page.
             if (_version >= 5.1 && _version <= 6) {
                 h.className += ' xpvista';
             }
-
-            // Add class to support downloading Firefox for Windows 64-bit on Windows 7 and later
```
alexgibson, line 164: 64bit downloads for Windows have never been enabled since the stub installer can handle these checks more accurately than we can do using UA, so removing the logic.

media/js/base/site.js
```diff
@@ -174,18 +123,16 @@
...
         var isARM = archType.match(/armv(\d+)/);
 
+        // Used for Windows and Linux ARM processor detection.
         if (archType !== 'x86') {
             h.className = h.className.replace('x86', archType);
 
             if (isARM) {
                 h.className += ' arm';
-
-                // Add class to support downloading Firefox for Android on ARMv7 and later
```
alexgibson, line 183: This also seems to be unused now since we link to the Play Store everywhere,

media/js/base/site.js
```diff
@@ -6,36 +6,11 @@
...
-            // Ths means we must use an ugly ternary statement here to make testing easier.
-            pf = (pf === '') ? '' : pf || navigator.platform;
+            pf = pf || navigator.platform;
             ua = ua || navigator.userAgent;
 
-            if (/Win(16|9[x58]|NT( [1234]| 5\.0| [^0-9]|[^ -]|$))/.test(ua) ||
-                    /Windows ([MC]E|9[x58]|3\.1|4\.10|NT( [1234]\D| 5\.0| [^0-9]|[^ ]|$))/.test(ua) ||
-                    /Windows_95/.test(ua)) {
-                /**
-                 * Officially unsupported platforms are Windows 95, 98, ME, NT 4.x, 2000
```
alexgibson, line 18: These operating systems are now so old it's difficult to make a case for keeping this logic. I can't find any VMs services that still support these systems, which makes testing pretty much impossible. All this is doing imho is increasing the chances of false positives.

alexgibson: I took a look in GA and less than 0.03% of Windows visitors over the past year have been on these systems combined. As such I think we're ok to drop this conditional.

media/js/base/site.js
```diff
@@ -45,29 +20,17 @@
...
-                    pf.indexOf('iPad') !== -1 ||
-                    pf.indexOf('iPod') !== -1 ||
-                    pf.indexOf('MacIntel') !== -1 && 'standalone' in navigator) {
+                pf.indexOf('iPad') !== -1 ||
+                pf.indexOf('iPod') !== -1 ||
+                pf.indexOf('MacIntel') !== -1 && 'standalone' in navigator) { //iPadOS
                 return 'ios';
             }
-            if (ua.indexOf('Mac OS X') !== -1) {
+            if (ua.indexOf('Mac OS X') !== -1 && !/Mac OS X 10.[0-8]\D/.test(ua)) {
```
alexgibson, line 31: OS X 10.0-8 made up ~2% of Mac visits over the past year. With this in mind, I think it's still worth serving them some conditional messaging.

alexgibson:
I'm going to fold this up into https://github.com/mozilla/bedrock/issues/9102. Closing

Issue #9103, opened 2020-07-02
Add compare link to /features/


rraue:
## Description

We hat do [remove the link](https://github.com/mozilla/bedrock/commit/8241a7f69aee9faed5f55c5a2c1792e0d6a58d30) to the compare pages on /features/ but now they are up again, we can add it again. I know the new feature page is nearly ready and migrated to protocol but as the /compare pages have issues getting re indexed, it would help even for a couple of days.

PR #9104, opened 2020-07-02
Add compare link

rraue:
The compare pages are back online and we can add the link again, so removing the comment: https://github.com/mozilla/bedrock/issues/9103

### DIFF
--- a/bedrock/firefox/templates/firefox/features/index.html
+++ b/bedrock/firefox/templates/firefox/features/index.html
@@ -93,15 +93,13 @@
           <h3>{{ _('Browse faster') }}</h3>
         </a>
       </li>
-    {# Temporarily remove comparison pages
-     # {% if l10n_has_tag('firefox_features_compare_browsers') %}
-     # <li class="features-list-item compare">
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/features/index.html
```diff
@@ -93,15 +93,13 @@ <h3>{{ _('Balanced memory usage') }}</h3>
...
-     # <li class="features-list-item compare">
-     #   <a href="{{ url('firefox.compare.index') }}" data-link-type="link" data-link-name="Compare Browsers">
-     #     <h3>{{ _('Compare Browsers') }}</h3>
-     #   </a>
-     # </li>
-     # {% endif %}
-     #}
+     {% if l10n_has_tag('firefox_features_compare_browsers') %}
+     <li class="features-list-item compare">
+        <a href="{{ url('firefox.compare.index') }}" data-link-type="link" data-link-name="Compare Browsers">
```
alexgibson, line 98: Ah, looks like this now needs to be:

```suggestion
        <a href="{{ url('firefox.browsers.compare.index') }}" data-link-type="link" data-link-name="Compare Browsers">
```

PR #9108, opened 2020-07-03
Exclude id locale from /whatsnew pages in Fluent config (Fixes #9100)

### LINKED ISSUE #9100: Remove `id` from the monthly WNP# template
## Description

WNP # page (not the default) that ties to a particular Fx Desktop release is not available in `id` like the other locales on production. The `id` template is hard coded with content from the Emerging Market's marketing team. Let's remove it from the page configuration so it doesn't show up in Pontoon for the community..

### Expected result

When WNP # page is pushed to Pontoon, `id` is excluded.

alexgibson:
Fixes: #9100

### DIFF
--- a/l10n/configs/special-templates.toml
+++ b/l10n/configs/special-templates.toml
@@ -14,3 +14,10 @@ basepath = ".."
 #         "de",
 #         "fr",
 #     ]
+
+[[paths]]
+    reference = "en/firefox/whatsnew/**/*.ftl"
+    l10n = "{locale}/firefox/whatsnew/**/*.ftl"
+    locales = [
+        "id",
+    ]


alexgibson:
@Pike r?

mathjazz:
I don't think so:
https://bugzilla.mozilla.org/show_bug.cgi?id=1602896

alexgibson:
Given the above, let me know if there are any problems to prevent merging this, or if there are any steps you may need to coordinate?

Pike:
Thanks, there's no problem other than that we need to think if we follow up with actions on the l10n side.

The files are currently translated, so there's no request for work that would go to waste, and this config change makes sure that if we take in new work, it won't bother the Indonesians.

That's a good state, I think.

PR #9114, opened 2020-07-06
Add Windows ARM64/AArch64 builds to /firefox/all/ (Fixes #8929)

### LINKED ISSUE #8929: mozilla.org does not provide a download option for Windows ARM64/AArch64 build
## Description

See: https://bugzilla.mozilla.org/show_bug.cgi?id=1636582

Tracking this to add as a download option to /firefox/all/ initially.

---

- [ ] Windows ARM64/AArch64 builds are available on /firefox/all/

alexgibson:
## Description
- Add Windows ARM64/AArch64 builds to /firefox/all/ and (/firefox/new/ modal).

http://127.0.0.1:8000/en-US/firefox/all/
http://127.0.0.1:8000/en-US/firefox/new/

## Issue / Bugzilla link
#8929

## Testing
Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/163494753

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -30,6 +30,7 @@ class FirefoxDesktop(_ProductDetails):
     platform_labels = OrderedDict([
         ('win64', 'Windows 64-bit'),
         ('win64-msi', 'Windows 64-bit MSI'),
+        ('win64-aarch64', 'Windows ARM64/AArch64'),
         ('win', 'Windows 32-bit'),
         ('win-msi', 'Windows 32-bit MSI'),
         ('osx', 'macOS'),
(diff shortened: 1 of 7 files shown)


alexgibson:
@pmac this seems to be working and tests passed. Not sure if there is anything else that I missed here?

alexgibson:
> I don't see anything in here about auto detection of the ARM platform for the default download button. Is that not requested or not possible?

This is a good question. Right now, I think Windows ARM users get a message saying "Your system doesn't meet the requirements to run Firefox." (see https://github.com/mozilla/bedrock/blob/master/media/css/protocol/components/_download-button.scss#L115).

Let me dig in and see if we can display the right download link instead.

The tricky part will be finding someone to help test that things work as expected on a real device.

pmac:
> Let me dig in and see if we can display the right download link instead.

That sounds like it could be a separate issue. It's fine with me if you want to go ahead and merge this and file a separate issue for the platform detection.

alexgibson:
> That sounds like it could be a separate issue. 

Ok, let's get this merged and I'll file a followup 👍

Issue #9112, opened 2020-07-06
CSP image error on /firefox/browsers/compare/ in Chrome


alexgibson:
## Description

When visiting the browser comparison index page in Chrome, there's a CSP error reported in the console. Firefox does not seem to report the error.

### Steps to reproduce

1.) In Chrome visit https://www.mozilla.org/en-US/firefox/browsers/compare/
2.) Open the web console.

### Expected result

There should be no CSP error.

### Actual result

```
Refused to frame 'https://www.mozilla.org/media/img/firefox/compare/compare-hero.7a35f38ce6db.svg' because it violates the following Content Security Policy directive: "frame-src www.googletagmanager.com www.google-analytics.com www.youtube-nocookie.com trackertest.org www.surveygizmo.com accounts.firefox.com accounts.firefox.com.cn www.youtube.com".
```

The root of the error might be caused by trying to load an [image referenced as an `<object>`](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/browsers/compare/index.html#L50-L52), which I'm guessing Chrome considers to be a frame.

Thankfully Chrome still renders the nested fallback image ok.

### Environment

Chrome, macOS.

craigcook:
As the comment in the template says, that image ended up being an `object` because of a bug in Safari 13.1 (since fixed) but Chrome wouldn't render the `object`, and clearly I never spotted the CSP violation when trying to debug the issue. The nested image worked so I `¯\_(ツ)_/¯`ed and moved on. But now maybe we no longer need the nested fallback image. I'll have to test it some more when I get the chance.

Thanks for fixing!

PR #9115, opened 2020-07-06
Update CSP_FRAME_SRC with the domain pointing to object.data URL.

musale:
## Description
Chrome throws the CSP error for failing to have the domain in the `frame-src` directive.

## Issue / Bugzilla link
#9112 

## Testing
Adding `localhost:8080` into the `CSP_FRAME_SRC` list locally fixes the error.


musale:
@alexgibson that's a much neater approach. I have updated it.

pmac:
Not that I know of. This looks good to me 💯

PR #9123, opened 2020-07-07
Add an ADR for using a single docker image for all deployments

pmac:
Re https://github.com/mozmeao/infra/issues/1306


pmac:
We have already implemented this. Merging.

PR #9124, opened 2020-07-08
EN homepage redesign

stephaniehobson:
## Description

Update homepage layout:

Screen Shot 2020-07-10 at 16 40 01 Screen Shot 2020-07-10 at 16 40 27

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/9135

## Testing

New cards merged to www-admin in https://github.com/mozmeao/www-admin/pull/76.

I could not see them locally so you may need to test on the heroku deployment.


stephaniehobson:
Oops, broke FR and DE. Will resubmit in a bit.

PR #9129, opened 2020-07-09
Enable download buttons for Windows ARM64 builds. (Fixes #9117)

### LINKED ISSUE #9117: Enable download buttons for Windows ARM64 builds.
## Description

Firefox builds for ARM64 / AArch64 were added to /firefox/all/ in https://github.com/mozilla/bedrock/issues/8929.

When someone tries to download Firefox directly via /new however, they still get the following message:

image

This issue is to investigate enabling the download button for those users.

One thing I'm not sure about is whether we need to worry about 32bit (seeing as the ARM builds are 64bit only). I don't think our UA detection is fool proof here. That said, anything may be better than the current experience.

---

#### Success Criteria
- [ ] ARM64 builds are served in production via the regular download button.

#### Approvals
- [ ] The Accountable
- [ ] Brand - design?
- [ ] Brand - copy?
- [ ] Team Lead
- [ ] Others?

alexgibson:
## Description
- Enabled download buttons for Windows ARM46 / AArch64 devices.

## Issue / Bugzilla link
#9117

## Testing
- [x] You can mock the behaviour locally by adding the class names `windows arm` (instead of `osx`) to the `<html>` element. Direct download button links should then point to `&os=win64-aarch64`.

### DIFF
--- a/media/css/pebbles/components/_buttons-download.scss
+++ b/media/css/pebbles/components/_buttons-download.scss
@@ -129,6 +129,7 @@ ul.download-list {
 .linux .download-button .os_linux,
 .linux.x86.x64 .download-button .os_linux64,
 .windows .download-button .os_win,
+.windows.arm .download-button .os_win64-aarch64,
 .osx .download-button .os_osx,
 .android .download-button .os_android,
 .download-button-android .os_android,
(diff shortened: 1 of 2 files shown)


alexgibson:
This has been verified as working as expected by the Firefox QA team, on ARM64 Windows devices (see: https://github.com/mozilla/bedrock/issues/9117#issuecomment-657750820). Marking as ready for review.

PR #9130, opened 2020-07-09
Add ability to run selenium tests locally

pmac:

### DIFF
--- a/docker/bin/run_integration_tests.sh
+++ b/docker/bin/run_integration_tests.sh
@@ -16,6 +16,10 @@ case $1 in
     BROWSER_VERSION=latest
     PLATFORM="Windows 10"
     ;;
+  firefox-local)
+    BROWSER_NAME=firefox
+    DRIVER="Remote"
+    ;;
   ie)
     BROWSER_NAME="internet explorer"
     PLATFORM="Windows 10"
@@ -55,6 +59,36 @@ if [ -z "${BASE_URL}" ]; then
 
   DOCKER_LINKS=(--link bedrock-code-${BRANCH_AND_COMMIT}:bedrock)
   BASE_URL="http://bedrock:8000"
+fi
+
+if [ "${DRIVER}" = "Remote" ]; then
+  # Start Selenium hub and NUMBER_OF_NODES (default 5) firefox nodes.
+  # Waits until all nodes are ready and then runs tests
+
+  SELENIUM_VERSION=${DOCKER_SELENIUM_VERSION:-"3.141.59-20200525"}
+
+  docker pull selenium/hub:${SELENIUM_VERSION}
+  docker pull selenium/node-firefox:${SELENIUM_VERSION}
+
+  # start selenium grid hub
+  docker run -d --rm -p 4444:4444 \
+    --name bedrock-selenium-hub-${BRANCH_AND_COMMIT} \
+    selenium/hub:${SELENIUM_VERSION}
+  DOCKER_LINKS=(${DOCKER_LINKS[@]} --link bedrock-selenium-hub-${BRANCH_AND_COMMIT}:hub)
+  SELENIUM_HOST="hub"
+
+  # start selenium grid nodes
+  for NODE_NUMBER in `seq ${NUMBER_OF_NODES:-5}`; do
+    docker run -d --rm --shm-size 2g \
+      --name bedrock-selenium-node-${NODE_NUMBER}-${BRANCH_AND_COMMIT} \
+      ${DOCKER_LINKS[@]} \
+      selenium/node-firefox:${SELENIUM_VERSION}
+    while ! ${SELENIUM_READY}; do
+      IP=`docker inspect --format '{{ .NetworkSettings.IPAddress }}' bedrock-selenium-node-${NODE_NUMBER}-${BRANCH_AND_COMMIT}`
+      CMD="docker run --rm --link bedrock-selenium-hub-${BRANCH_AND_COMMIT}:hub tutum/curl curl http://hub:4444/grid/api/proxy/?id=http://${IP}:5555 | grep 'proxy found'"
+      if eval ${CMD}; then SELENIUM_READY=true; fi
+    done
+  done
 fi
 
 # make sure results dir exists or docker will create it


bin/run-integration-tests.sh
```diff
@@ -36,6 +36,7 @@ fi
 
 # Sauce Labs arguments
 if [ "${DRIVER}" = "SauceLabs" ]; then
+  exit 0
```
alexgibson, line 39: Would this still allow us to run download and headless tests in the pipeline? Worth pushing for a test run to make sure?

alexgibson: Thinking about it - might it be worth adding our most high traffic URLs to the headless tests just to assert a 200 response? This might be enough to catch at least some errors?

alexgibson: I think en-US would probably be fine as nearly all pages use a common template.

We already have /firefox/new/ and /firefox/all/ covered by headless (I think), so maybe add:

- `/en-US/`
- `/firefox/`
- `/firefox/mobile/`
- `/firefox/browsers/`
- `/firefox/proucts/`
- `/firefox/accounts/`
- `/firefox/%version%/whatsnew/`
- `/firefox/installer-help/`
- `/firefox/releasenotes/`
- `/firefox/channel/desktop/`
- `/firefox/developer/`

Anything else product related?

pmac: Looks like most of these actually are already covered by our headless tests. I think those may only be looking for the redirects though. Maybe we can make them test the final destination as well?

https://github.com/mozilla/bedrock/blob/f9f5df9248e88f64e8cf2f7783b40b98e542fd49/tests/redirects/test_urls.py#L27-L60

alexgibson: Some of those urls are long gone and 404. If we can update the list to be more recent like above then great.

alexgibson: You could also add to something like these tests if you didn’t want to change redirect tests https://github.com/mozilla/bedrock/blob/master/tests/functional/test_generated_pages.py

pmac: I've added some tests and am running integration tests again.

pmac: If we later wanted something comprehensive I could have it download the sitemap and crawl the whole thing. Seems a bit much for this but possibly useful if we have to scale back the other.

tests/redirects/test_urls.py
```diff
@@ -58,3 +58,32 @@ def test_404_url(base_url):
...
+    '/en-US/firefox/mobile/',
+    '/en-US/firefox/browsers/',
+    '/en-US/firefox/products/',
+    '/en-US/firefox/accounts/',
+    '/en-US/firefox/channel/desktop/',
+    '/en-US/firefox/lockwise/',
+    '/en-US/firefox/developer/',
+    '/en-US/firefox/installer-help/',
+    '/en-US/firefox/notes/',
+    '/en-US/firefox/whatsnew/all/',
```
alexgibson, line 80: We host many different /whatsnew pages, which are some of our highest traffic URLs. I don't think it's worth adding them all here, but testing the latest 2 major releases would be nice:

- https://www.mozilla.org/en-US/firefox/78.0/whatsnew/all/
- https://www.mozilla.org/en-US/firefox/77.0/whatsnew/all/

tests/redirects/test_urls.py
```diff
@@ -58,3 +58,32 @@ def test_404_url(base_url):
...
+    '/en-US/security/advisories/',
+    '/en-US/privacy/websites/',
+    '/en-US/privacy/firefox/',
+    '/credits/',
+])
+def test_real_url(url, base_url, follow_redirects=True):
+    """Ensure that major URLs work and return a 200"""
```
alexgibson, line 87: Not a blocker, but this isn't really testing redirects any longer - would it make sense to move it under `/functional`?

pmac: True. It was just similar code-wise. I can move it if you'd rather.

tests/redirects/test_urls.py
pmac: Yeah. I just figured that would become outdated very quickly. Not sure how best to make that work well. How do you test those pages normally?

tests/redirects/test_urls.py
alexgibson: Up to you as it works either way - just thought it might have caused someone confusion in the future if this ever merges :)

tests/redirects/test_urls.py
alexgibson: We have functional tests for each major release. When we remove templates, we remove the accompanying tests.

pmac: Looks like none of those pages have any tests in functional at the moment. 76 is the latest test I can find.

pmac:
I think what would really be nice is if we could fall back to spinning up our own Selenium cluster with just Firefox and run the tests there. It wouldn't be as good as Sauce but it would be better than not running functional tests at all. I think we used to do that in fact.

tests/redirects/test_urls.py
alexgibson: Sigh, they should have tests :(

the tests for Dev edition and Nightly do exist, they just don't change version to version:

https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/whatsnew/test_whatsnew_developer_70.py
https://github.com/mozilla/bedrock/blob/master/tests/functional/firefox/whatsnew/test_whatsnew_nightly_70.py

Can we please just add these URLs so they have some level of coverage?

alexgibson:
> I think what would really be nice is if we could fall back to spinning up our own Selenium cluster with just Firefox and run the tests there. It wouldn't be as good as Sauce but it would be better than not running functional tests at all. I think we used to do that in fact.

We did yes - if that's something you can spin up again, then it would be much better than the current situation we're facing imho.

pmac:
Looks like we got rid of the local selenium tests in #7030. Perhaps it's time to try to bring that back.

tests/redirects/test_urls.py
pmac: Yeah. I was just looking to see if maybe there was some automated way to discover the tests to run. I could automatically grab all of the firefox URLs and test those I think... Though I may look into running the selenium tests locally instead.

alexgibson:
> Perhaps it's time to try to bring that back.

Yeah, if it's resurrectable without a ton of work it might be a better option than this. Ideally it would be great to run tests in both Firefox and at least one non-Firefox browser (e.g. Chrome). The main reason we removed the local selinium docker container was because tests are really hard to debug there (plus we can't specify the platform). I'd still favor that over dropping tests entirely, especially if it doesn't cost us anything.

pmac:
Okay. I've changed this up entirely. Now this adds the `firefox-local` target to the `run_integration_tests.sh` script so that we can run our selenium tests in Firefox in a local grid when we can't run with a 3rd party service and more browsers and platforms. I'll follow up with a PR to www-config to switch to running this task instead of the others if we need it. But I think we could go ahead and merge this so that we're ready if need be. It will also be useful if our 3rd party test runner service has an outage.

pmac:
Turns out I need to do all of this in the [www-config](https://github.com/mozmeao/www-config) repo. Closing this.

PR #9131, opened 2020-07-10
Cache-bust doubleclick pixel on /download/thanks (Fixes #9128)

### LINKED ISSUE #9128: Cache-bust doubleclick pixel on /download/thanks
## Description

After a call with iProspect, one of Mozilla's marketing agency partners, they flagged an error with the doubleclick pixel that gets loaded on the `/download/thanks/` page. The bug stems from the fact that the pixel URL never changes, and so browsers will cache subsequent requests, and this also effects the overall accuracy of reporting.

They recommend enumeration of the `num` parameter with a random value of alphanumeric characters, between 8 and 15 digits long, which we can do using JS before making the request.

### Steps to reproduce

1.) In a browser with DNT disabled, visit https://www.mozilla.org/firefox/new/ and click the download button.
2.) On the /download/thanks/ page, observe the network request for doubleclick.

### Expected result

In the request the `num` parameter should have a random value.

### Actual result

The value is always `1` e.g:

```
https://ad.doubleclick.net/ddm/activity/src=6417015;dc_pre=CMiR6eDfv-oCFdfb1QodXJwKMQ;type=deskt0;cat=mozil0;dc_lat=;dc_rdid=;tag_for_child_directed_treatment=;ord=1;num=1?&_dc_ck=try
```

iProspect provided some (hairy looking) pseudo-code, which we would need to refactor (because `document.write()` is bad practice and we also respect things like DNT). It is enough to provide context for what needs doing at the least.

```html
<!--
Start of Floodlight Tag: Please do not remove
Activity name of this tag: Mozilla Desktop Download
URL of the webpage where the tag is expected to be placed: http://www.mozilla.org
This tag must be placed between the <body> and </body> tags, as close as possible to the
opening tag.
Creation Date: 06/17/2020
-->
<script type="text/javascript">
var axel = Math.random() + "";
var a = axel * 10000000000000;
document.write('<img
src="https://ad.doubleclick.net/ddm/activity/src=6417015;type=deskt0;cat=mozil0;dc_lat
=;dc_rdid=;tag_for_child_directed_treatment=;tfua=;npa=;ord=1;num=' + a + '?" width="1"
height="1" alt=""/>');
</script>
<noscript>
<img
src="https://ad.doubleclick.net/ddm/activity/src=6417015;type=deskt0;cat=mozil0;dc_lat
=;dc_rdid=;tag_for_child_directed_treatment=;tfua=;npa=;ord=1;num=1?" width="1"
height="1" alt=""/>
</noscript>
<!-- End of Floodlight Tag: Please do not remove -->
```

### Environment

Chrome, macOS

alexgibson:
## Description
- Adds random number param to cache bust doubleclick request on /download/thanks/.
- Replaces jQuery code with vanilla JS.
- Adds test for cache bust param.

## Issue / Bugzilla link
#9128

## Testing

Test using a browser with DNT disabled:

- [ ] Pixel request should fire on /thanks/ as expected.
- [ ] Pixel request should contain a `num` parameter with a random value.

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -1510,7 +1510,7 @@ CSP_FRAME_SRC = CSP_CHILD_SRC
 # Bug 1331069 - Double Click tracking pixel for download page.
 AVAILABLE_TRACKING_PIXELS = {
     'doubleclick': ('https://ad.doubleclick.net/ddm/activity/src=6417015;type=deskt0;cat=mozil0;dc_lat=;dc_rdid=;'
-                    'tag_for_child_directed_treatment=;ord=1;num=1?&_dc_ck=try'),
+                    'tag_for_child_directed_treatment=;tfua=;npa=;ord=1'),
 }
 ENABLED_PIXELS = config('ENABLED_PIXELS', default='doubleclick', parser=ListOf(str))
 TRACKING_PIXELS = [AVAILABLE_TRACKING_PIXELS[x] for x in ENABLED_PIXELS if x in AVAILABLE_TRACKING_PIXELS]
--- a/media/js/base/mozilla-pixel.js
+++ b/media/js/base/mozilla-pixel.js
@@ -21,13 +21,13 @@ if (typeof window.Mozilla === 'undefined') {
     var Pixel = {};
 
     Pixel.getPixelData = function() {
-        return $('#strings').data('pixels');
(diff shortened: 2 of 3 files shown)


alexgibson:
Ok, iProspect have confirmed that the demo instance above it firing as expected. I think this is ready for review.

PR #9133, opened 2020-07-10
Switch from chrome /thanks variation. [no bug]

amychurchwell:
## Description
Custom designed /thanks page to be linked to in a convert experiment.

## Issue / Bugzilla link
[Airtable card](https://airtable.com/tblO5zD4uNk7ISyG0/viwIUJsHKDrOFl90B/rec7dWsJPvKgsF1QQ)

## Testing
**Demo:** https://www-demo5.allizom.org/en-US/firefox/download/thanks/?v=a

image

### DIFF
--- a/bedrock/firefox/templates/firefox/new/trailhead/thanks-a.html
+++ b/bedrock/firefox/templates/firefox/new/trailhead/thanks-a.html
@@ -11,10 +11,6 @@
   {{ css_bundle('firefox_exp_thanks') }}
 {% endblock %}
 
-{% block site_header %}
-  {% include 'includes/protocol/navigation/menu-firefox/index.html' %}
-{% endblock %}
-
 {% block content %}
 <main role="main">
   <div class="mzp-c-notification-bar mzp-t-warning unsupported">
--- a/bedrock/firefox/templates/firefox/new/trailhead/thanks-a.html
+++ b/bedrock/firefox/templates/firefox/new/trailhead/thanks-a.html
@@ -1,0 +1,83 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros-protocol.html" import hero, picto_card with context %}
+
+{% extends "firefox/new/trailhead/thanks.html" %}
+
+{% block extrahead %}
+  {{ css_bundle('protocol-core') }}
+  {{ css_bundle('firefox_exp_thanks') }}
+{% endblock %}
+
+{% block site_header %}
+  {% include 'includes/protocol/navigation/menu-firefox/index.html' %}
+{% endblock %}
+
+{% block content %}
+<main role="main">
+  <div class="mzp-c-notification-bar mzp-t-warning unsupported">
+    <p>
+      {{ ftl('firefox-new-youre-using-an-insecure-outdated', url='https://support.mozilla.org/kb/end-support-windows-xp-and-vista') }}
+    </p>
+  </div>
+  <div class="mzp-c-notification-bar mzp-t-success auto-download">
+    {# fallback_url is replaced by the platform download link via JS, but if
+       something fails the user should still get a link to a working download path. #}
+    <p>
+      {{ ftl('firefox-new-your-firefox-new-should-begin', id='direct-download-link', fallback_url=url('firefox.all'))}}
+    </p>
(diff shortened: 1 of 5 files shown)


alexgibson:
In IE the notification prompt is over to the side of the screen:

image

bedrock/firefox/views.py
```diff
@@ -662,11 +662,11 @@ def get_template_names(self):
 class DownloadThanksView(L10nTemplateView):
     ftl_files_map = {
         'firefox/new/trailhead/thanks.html': ['firefox/new/download'],
-        'firefox/new/trailhead/thanks-b.html': ['firefox/new/download'],
```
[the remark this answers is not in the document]
amychurchwell: It seemed safe to remove because the template does not exist, but happy to remove this change if it causes confusion.

bedrock/firefox/views.py
```diff
@@ -687,7 +687,9 @@ def get_template_names(self):
         if variant not in self.variations:
             variant = None
 
-        if ftl_file_is_active('firefox/new/download'):
+        if variant == 'a':
```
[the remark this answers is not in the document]
amychurchwell: not necessarily, because a convert experiment on /new will link to this page & the cloudflare worker will handle locale. Going to add in anyway just to be explicit.

bedrock/firefox/templates/firefox/new/trailhead/thanks-a.html
```diff
@@ -0,0 +1,83 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% from "macros-protocol.html" import hero, picto_card with context %}
+
+{% extends "firefox/new/trailhead/base.html" %}
```
alexgibson, line 7: The download isn't working because there's no JS bundle included in the template here. If you extend the regular `thanks.html` template then it should work again, without having to redefine things.
```suggestion
{% extends "firefox/new/trailhead/thanks.html" %}
```

bedrock/firefox/templates/firefox/new/trailhead/thanks-a.html
```diff
@@ -0,0 +1,83 @@
...
+
+{% block extrahead %}
+  {{ super() }}
```
alexgibson, line 10: If you extend from `thanks.html`, then also make this change:

```suggestion
  {{ css_bundle('protocol-core') }}
```

bedrock/firefox/templates/firefox/new/trailhead/thanks-a.html
```diff
@@ -0,0 +1,83 @@
...
+  {{ css_bundle('firefox_exp_thanks') }}
+{% endblock %}
+
+{% block site_header %}
```
alexgibson, line 14: If you extend from `thanks.html` then you can also remove this block.

bedrock/firefox/templates/firefox/new/trailhead/thanks-a.html
```diff
@@ -0,0 +1,83 @@
...
+      <li class="c-steps-item">Follow <a href="https://support.mozilla.org/kb/import-bookmarks-google-chrome">this tutorial</a> to import your passwords, bookmarks, browsing history, and other data from Chrome.</li>
+      <li class="c-steps-item">Enjoy exploring the internet with more speed and privacy.</li>
+    </ol>
+  </section>
+
+  <section class="c-share">
+    <div class="mzp-l-content">
+      <h2 class="c-share-title">Spread the word about Firefox and help your favorite people say goodbye to Chrome.</h2>
+
+      {% set share_url = settings.CANONICAL_URL + canonical_path %}
```
alexgibson, line 59: The share buttons are stil pointing to /firefox/download/thanks/ instead of /firefox/new/?

bedrock/firefox/templates/firefox/new/trailhead/thanks-a.html
amychurchwell: I added it in because the design included the nav:
image

alexgibson: What I mean is, the current /thanks page already has the nav - so if you extend that template then there's no need to re-include it again.

amychurchwell: ohh whoops, removing now.

Issue #9135, opened 2020-07-10
Update English Homepage Layout


stephaniehobson:
MJ has requested an update to our homepage layout. It makes some card bigger, moves some of the section separators around, and changes the banner to be privacy focused.

<img width="251" alt="Screen Shot 2020-07-10 at 16 40 01" src="https://user-images.githubusercontent.com/854701/87211022-8e9d1e80-c2cc-11ea-9864-561ab5476909.png"> <img width="247" alt="Screen Shot 2020-07-10 at 16 40 27" src="https://user-images.githubusercontent.com/854701/87211026-91980f00-c2cc-11ea-9f09-d270885ad477.png">

alexgibson:
This looks like it was done, closing

Issue #5824, opened 2018-06-19
Homepage needs a localization plan


alexgibson:
The [redesigned home page](https://www.mozilla.org/en-US/) is being shown for all English locales, however all other non-English locales, *except* DE, are still being served the [old home page design](https://www.mozilla.org/fr/), because the editorial content is aimed for English audiences and our blog content is largely not translated. The card content is also likely to change every 2 weeks, which is pretty pointless to try and translate.

We need to develop a strategy for l10n and the new home page, as currently 60% of our global audience is not seeing it. The download CTA sections and billboards are evergreen and are well suited for translation, but we need some other content to replace the card sections. We can easily serve non-English locales a slightly different variation on the content hierarchy, but the overall template design should remain consistent and use Protocol components.

michaelmccombie:
For a temporary fix we could convert the old homepage content we have to homepage cards. It would keep the designs consistent and we wouldn't have to go through the process of translating and testing new content.

Ideally we would come back and create something brand new, but it's probably best not to go down that route until we are sure the current homepage design is performing well.

![image](https://user-images.githubusercontent.com/5924575/41606674-13065a72-7399-11e8-9fe9-33e0cbe91ecf.png)

michaelmccombie:
We could also take the landing page templates (which we haven't started building yet) and adapt that work to create a non /en-US home page. 

![image](https://user-images.githubusercontent.com/5924575/41690798-191c3f9e-74ac-11e8-98c4-219075a49d04.png)

michaelmccombie:
Personally, I'd lean towards the landing page look because it makes more sense for our current content (assuming we don't want to turn this into a bigger redesign project) and it would be nice to have a project in mind already when we start putting together those components.

ejregithub:
Flagging the new to get the brand teams involved for determine evergreeen content on the page for non-Tier one locales.

ejregithub:
I emailed Eddie Ibarra (TL for NA Brand) and then sent a meeting invite, adding Robin (TL for EU Brand) to that for June 24th

ejregithub:
From the meeting. Robin self-assigned the task to talk to Janis about this including determining who the Accountable person should be here.

pmac:
@alexgibson with all of the changes since this was last filed, do we still need to have this one open?

alexgibson:
Closing in favor of https://github.com/mozilla/bedrock/issues/12282

PR #9136, opened 2020-07-11
EN homepage redesign

stephaniehobson:
## Description

Update homepage layout:

Screen Shot 2020-07-10 at 16 40 01 Screen Shot 2020-07-10 at 16 40 27

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/9135

## Testing

New cards merged to www-admin in https://github.com/mozmeao/www-admin/pull/76. They are in a different directory so the live page will work until this update is published and then the old folder can be removed.

I could not see them locally so you may need to test on the heroku deployment.

Make sure I didn't break DE or FR.


stephaniehobson:
This version of the homepage is English only because the content is primarily links to blog posts and the blog posts are not translated to very many languages. There's off and on work tracked in https://github.com/mozilla/bedrock/issues/5824 to update the localized version of the homepage.

Issue #9061, opened 2020-06-25
[July 17] WNP 79 - Code


ejregithub:
## Description

Code the page supporting the marketing theme selected for the WNP shown to users upgrading to Firefox 79 (release on 7/28)

[Principal project doc](https://docs.google.com/document/d/1DW2-pohjlRxmtSf7l4vhsVUCkhLQuXh6VdkOyopWPDk/edit)

---

#### Success Criteria
- [ ] Page is live in prod on or before July 17

#### Approvals
- [ ] The Accountable - MWarther
- [ ] Brand - Clio
- [ ] Brand - Natalie
- [ ] Team Lead - Hoosteeno
- [ ] Others?

craigcook:
Merged in https://github.com/mozilla/bedrock/commit/f771dc08768835577e7873a89e193e4158f2bcfb

Issue #9060, opened 2020-06-25
WNP 79 - Design


ejregithub:

cgonz0:
**assets** 
note: please set hero img width to 230px.
[https://app.zenhub.com/files/1616665/34626163-1b45-47ac-9871-10f5aa7f67a0/download](https://app.zenhub.com/files/1616665/34626163-1b45-47ac-9871-10f5aa7f67a0/download)

**mocks**
![wnp79.png](https://images.zenhubusercontent.com/5d7c23b1d8206c00011723a8/f9ab23af-9ae3-44c7-8a9a-4f73d9a06596)
![wnp79 - zhn.png](https://images.zenhubusercontent.com/5d7c23b1d8206c00011723a8/cc31b4bd-b2e8-4f22-b24a-ac7cd1fea14d)
![wnp79 - already Set to Default.png](https://images.zenhubusercontent.com/5d7c23b1d8206c00011723a8/a468ed6a-a85e-4126-aa07-74f72e863e07)
![wnp79 Non EN.png](https://images.zenhubusercontent.com/5d7c23b1d8206c00011723a8/f4da7966-06e3-4a1f-b113-7a3b301aed69)

PR #9139, opened 2020-07-13
Add WNP79

craigcook:
## Description
Adds what's new page for Firefox 79.0.

## Issue / Bugzilla link
#9061 
Design issue (with screenshots for reference): #9060 

## Testing
http://localhost:8000/firefox/79.0/whatsnew/


bedrock/firefox/views.py
```diff
@@ -529,7 +529,8 @@ def get_redirect_url(self, *args, **kwargs):
 class WhatsnewView(L10nTemplateView):
 
     ftl_files_map = {
-        'firefox/whatsnew/index-account.html': ['firefox/whatsnew/whatsnew-account', 'firefox/whatsnew/whatsnew']
+        'firefox/whatsnew/index-account.html': ['firefox/whatsnew/whatsnew-account', 'firefox/whatsnew/whatsnew'],
+        'firefox/whatsnew/whatsnew-fx79.html': ['send_to_device', 'firefox/whatsnew/whatsnew', 'firefox/whatsnew/whatsnew-fx79']
```
alexgibson, line 533: The .ftl file `send_to_device` is now a global file, so no need to include it here. 

Much the same as .lang, the order of these files can also govern locale activation. The file `firefox/whatsnew/whatsnew-fx79` should be first, before `firefox/whatsnew/whatsnew`.

bedrock/firefox/views.py
```diff
@@ -549,7 +550,10 @@ def get_context_data(self, **kwargs):
 
         analytics_version = str(num_version) + channel
         entrypoint = 'mozilla.org-whatsnew' + analytics_version
-        campaign = 'whatsnew' + analytics_version
+        if version.startswith('79.'):
+            campaign = 'mozilla-vpn-v1-0-announcement'
```
alexgibson, line 554: do we need to add a `channel` check in here to make sure only this WNP gets the custom param?

media/static-bundles.json
```diff
@@ -907,6 +914,17 @@
       ],
       "name": "firefox_whatsnew_77_zhCN"
     },
+    {
+      "files": [
+        "js/base/uitour-lib.js",
```
alexgibson, line 919: As of https://github.com/mozilla/bedrock/pull/8914 this lib is now included in our common bundles. so you should be able to remove it here.

alexgibson:
Oh, seeing as you've already done the L10n PR manually for this page you may want to add `[skip l10n]` to the commit message here.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx79.html
```diff
@@ -0,0 +1,155 @@
...
+        </div>
+
+        <div class="state-is-default">
+          <button class="mzp-c-button mzp-t-product mzp-t-small js-modal-link">{{ ftl('whatsnew79-fallback-cta') }}</button>
+        </div>
+      </div>
+    </div>
+
+    <aside id="modal" class="mzp-u-modal-content mzp-l-content">
+      <h2 class="modal-logo">{{ ftl('brand-name-firefox') }}</h2>
```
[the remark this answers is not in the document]
craigcook: Even better, since this is just a decorative image it doesn't need to be a separate heading at all. I've reworked it to use a pseudo-element.

craigcook:
Updated and rebased (and added [skip l10n])

Issue #9110, opened 2020-07-03
RTL styling issues on /firefox/browsers/ page


alexgibson:
## Description

The Firefox Accounts form on the /firefox/browsers/ page is inaccessible on RTL locales due to styling issues.

### Steps to reproduce

https://www-dev.allizom.org/ar/firefox/browsers/

### Expected result

Form should be accessible.

### Actual result

![image](https://user-images.githubusercontent.com/400117/86476111-cab90800-bd3d-11ea-9336-f503a5d74bea.png)

### Environment

Firefox.

Abhirup-99:
is this issue still up? Would like to contribute.

alexgibson:
@Abhirup-99 please see our contributing guide to get started, thanks https://github.com/mozilla/bedrock#contributing

berliners1:
Hello, I just made a pull request with a potential fix for this issue.

PR #9168, opened 2020-07-17
 RTL styling issues on /firefox/browsers/ page

berliners1:
## Description
I made a small change that will fix this issue, by utilizing z-index to make sure the image is underneath the form container at all times, so the image cannot ever make the form unusable by being in front of it.

## Issue / Bugzilla link
#9110 

## Testing
http://localhost:8000/ar/firefox/browsers/

### DIFF
--- a/media/css/firefox/browsers-products.scss
+++ b/media/css/firefox/browsers-products.scss
@@ -549,6 +549,7 @@ $url-download-link-hover: svg-url($download-link-hover);
             right: $layout-md;
             bottom: $layout-md;
             left: calc(560px + #{$layout-md * 2});
+            z-index:0;
         }
     }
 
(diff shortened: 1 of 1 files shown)


berliners1:
Okay, I replaced my z-index-based solution with this one, and to me it looks good now. How does it look for you?

media/css/firefox/browsers-products.scss
```diff
@@ -549,6 +549,10 @@ $url-download-link-hover: svg-url($download-link-hover);
             right: $layout-md;
```
alexgibson, line 549: You can remove the existing properties for both `left` and `right` now that the mixin is defining them.

Nit: try and keep things in alphabetical order, so the `@include bidi()` can come first at the top of the list.

Thanks!

berliners1:
Ok, made these changes too. But... I didn't realize @ isn't allowed in commit messages? Oops. Do you know offhand if I can change that?

berliners1:
Ok, my best solution to fixing that was to amend the commit to remove the @, following this documentation: https://docs.github.com/en/github/committing-changes-to-your-project/changing-a-commit-message

Let me know if everything looks good with that.

Issue #9075, opened 2020-06-26
Update to Protocol 12.0.0


stephaniehobson:
Update to Protocol 12.0.0.

PR #9169, opened 2020-07-17
Remove pebbles base templates and assets (Fixes #8868)

### LINKED ISSUE #8868: Remove pebbles base templates and assets
## Description

Once the following issues are resolved, we should be unblocked on removing the Pebbles framework dependencies and related assets:

- https://github.com/mozilla/bedrock/issues/8375
- https://github.com/mozilla/bedrock/issues/8686

---

- [ ] All Pebbles related code is removed from bedrock.

#### Tasks
- [ ] Remove Pebbles scss library and related files.
- [ ] Remove / rename Pebbles base template(s).

alexgibson:
## Description
- Remove pebbles related .scss files and bundles.
- Remove pebbles base HTML templates.
- Renames `common-protocol.js` bundle to `common.js`.
- Renames `common-protocol-no-jquery.js` to `common-no-jquery.js`.
- Renames `protocol-core.css` bundle to `common.css`.
- Renames `protocol-oldIE.css` bundle to `common-old-ie.css`.
- Remove unused CSS/image assets related to hubs and pebbles sub navigation (both now gone).

Note: I resisted changing `base-protocol.html` to `base.html` as it would touch so many files, but happy to go the extra mile and do this if we think it's worthwhile?

## Issue / Bugzilla link
#8868

## Testing
- [ ] Are there any stray pebbles assets I missed?
- [ ] Are there any pages still using any of these resources?

Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/167777451


stephaniehobson:
Can we remove `button button-green` from the download button macro 🥺

stephaniehobson:
> Can we remove `button button-green` from the download button macro 🥺

I did it in #9075.

media/js/base/mozilla-video-poster.js
```diff
@@ -4,7 +4,7 @@
 
 /* A simple HTML5 Video poster image helper.
  * https://bugzilla.mozilla.org/show_bug.cgi?id=1040213
- * CSS Styles: 'media/css/pebbles/_video-poster.scss' */
+ */
```
stephaniehobson, line 7: Looks like there's no CSS for this now? I think that means we don't need to add/remove the `supports-video` class further down.

media/css/pebbles/base/elements/_lists.scss
```diff
@@ -1,75 +0,0 @@
...
-    }
-
-    ol {
-        list-style: decimal;
-        margin-left: 20px;
-        margin-bottom: 0;
-    }
-}
-
-ol.prose {
```
stephaniehobson, line 32: There are a few instances of `prose` lingering in the HTML.

alexgibson:
> Also, can/did you check if we can delete any of the fonts since Protocol brings its own with it?

I didn't - let's save that for https://github.com/mozilla/bedrock/issues/9170

media/js/base/mozilla-video-poster.js
alexgibson: This class still might be useful for pages that use the helper, so I think keeping it around won't do any harm

alexgibson:
Thanks for the review @stephaniehobson

PR #9180, opened 2020-07-20
Add Sentry client SDK (Fixes #8444)

### LINKED ISSUE #8444: Implement client side error reporting
## Description

Bedrock doesn't currently have any form of client side error reporting, so we have no idea when or where JavaScript errors are occuring, unless they are lucky enough to be caught by functional tests.

We should do an evaluation of client side error reporting tools and look at what might be applicable to bedrock. For example, we already use [Sentry](https://sentry.io/welcome/) on the backend, but don't have the client side reporting tool enabled. If third-party tools aren't an option (e.g. due to our websites privacy policy), then we could also look to implement a (crude) first-party reporting mechanism.

---

- [ ] Choose and implement an appropriate client side reporting tool.

- [ ] Chosen will need to comply with our websites privacy policy.

#### Tasks
- [ ] Evaluate the options available.
- [ ] Implement a solution.

alexgibson:
## Description
WIP demo: https://www-demo2.allizom.org/en-US/

## Issue / Bugzilla link
#8444

## Testing
- [ ] Sentry JS should only initialize if DNT is disabled.
- [ ] Sentry JS should only load when switch is enabled.


bedrock/settings/base.py
```diff
@@ -1529,3 +1529,6 @@ def get_default_gateway_linux():
 
 # Issue 7508 - Convert.com experiment sandbox
 CONVERT_PROJECT_ID = ('10039-1003350' if DEV else '10039-1003343')
+
+# Issue 8444 - Sentry Client side SDK
+SENTRY_DATA_SOURCE_NAME = config('SENTRY_DATA_SOURCE_NAME', default='')
```
alexgibson, line 1534: @pmac I took a pop at adding this to https://www-demo1.allizom.org/en-US/

Do we already have the Sentry DSN's for each deployment set up as environment variables?

pmac: We do, but I think the one for the client-side is different. The server side one is secret.

https://github.com/mozilla/bedrock/blob/1cd982199cd19dfdca7fd8fd5cdd78309f94b4d7/bedrock/settings/base.py#L1434

pmac:
I just opened #9183 to upgrade our Sentry python library from Raven to the new Sentry-SDK, and have made sure that there is now a `SENTRY_DSN` setting you can use for this. Once that is merged I'll update all of the configs for the platforms to use the safe DSN version and we should be good for an even better test.

alexgibson:
This is now working on demo: https://sentry.prod.mozaws.net/operations/bedrock-demos/issues/9190127/?referrer=alert_email

alexgibson:
Note to self: look into filtering https://docs.sentry.io/platforms/javascript/configuration/filtering/

pmac:
The Sentry-SDK update for the backend has just merged (again) and so you should be good to base this work on that so that they share DSN settings etc.

alexgibson:
@pmac I've rebased this and set it to a 10% sample rate.

Seems to be working on https://www-demo2.allizom.org/en-US/

https://sentry.prod.mozaws.net/operations/bedrock-demos/issues/9190127/?query=is%3Aunresolved

pmac:
Nice! We about ready to try it for real?

alexgibson:
I think so. We’re still waiting back on the RRA result, but I don my expect there to be anything too unexpected. We could have this running on dev / stage until then.

Thoughts?

pmac:
If we put a switch around the sentry JS bundle I think that'd be great. Just adding a switch will get us dev enabled and then we can enable stage and prod when we're ready.

alexgibson:
Added a switch in the base template for the JS bundle 👍

media/js/mozorg/home/home.js
```diff
@@ -8,4 +8,7 @@
     // Lazyload images
     Mozilla.LazyLoad.init();
 
+    // TODO remove this is a test
+    wmyUndefinedFunction();
```
alexgibson, line 12: Whoope I need to remove this still

alexgibson:
Ok, I think this is now ready

pmac (approved):
LGTM 🚢

PR #9183, opened 2020-07-20
Move from Raven to Sentry-SDK for error reporting

### LINKED ISSUE #6378: Upgrade to new Sentry SDK
Sentry dropped raven package in favor of a new unified SDK.

-   New Package: https://github.com/getsentry/sentry-python
-   Docs: https://docs.sentry.io/quickstart/?platform=python

While I couldn't find a deprecation schedule, please plan to upgrade to the new SDK in the near future to avoid incompatibility issues with the server.

pmac:
Fix #6378


bedrock/settings/base.py
```diff
@@ -1430,11 +1429,13 @@ def get_default_gateway_linux():
 
 DEAD_MANS_SNITCH_URL = config('DEAD_MANS_SNITCH_URL', default='')
 
-RAVEN_CONFIG = {
-    'dsn': config('SENTRY_DSN', default=''),
-    'site': '.'.join(x for x in [APP_NAME, CLUSTER_NAME] if x),
-    'release': config('GIT_SHA', default=''),
-}
+SENTRY_DSN = config("SENTRY_DSN", default='')
```
glogiotatidis, line 1432: nit: `s/"/'/` (also below)

and when we convert to using Black we can changed everything back :rofl:

bedrock/settings/base.py
```diff
@@ -1498,7 +1499,7 @@ def get_default_gateway_linux():
 CSP_REPORT_ONLY = config('CSP_REPORT_ONLY', default='false', parser=bool)
 CSP_REPORT_ENABLE = config('CSP_REPORT_ENABLE', default='false', parser=bool)
 if CSP_REPORT_ENABLE:
-    CSP_REPORT_URI = config('CSP_REPORT_URI', default='/csp-violation-capture')
+    CSP_REPORT_URI = config('CSP_REPORT_URI', default='')
```
glogiotatidis, line 1502: I'd remove default value here, to prevent it from starting if no URI has been defined by `CSP_REPORT_ENABLE` is `True`

pmac:
> I found that `django-csp` accepts [`CSP_REPORT_PERCENTAGE`](https://django-csp.readthedocs.io/en/latest/reports.html#throttling-the-number-of-reports) config variable which looks interesting, especially in high traffic sites.

Good find! That would possibly enable us to use the report in prod. We don't currently turn on CSP failure reporting in prod, only in stage and dev I think. I'll file another issue to discuss whether we want to do that with the team.

bedrock/settings/base.py
pmac: I decided to remove the `CSP_REPORT_ENABLE` setting entirely. It's no longer needed since we don't have a default URL for the reports. I tested having `None` as the default for the setting and it is disabled as expected.

PR #9185, opened 2020-07-20
[no bug] Remove old whatsnew pages

craigcook:
## Description
This removes all the custom whatsnew pages for Firefox versions prior to 76 (they’ll get the evergreen base page) as well as related assets and tests.

## Testing
Make sure I didn't miss anything that should be deleted
Make sure I didn't delete anything that should be kept


bedrock/firefox/views.py
```diff
@@ -606,18 +606,8 @@ def get_template_names(self):
             else:
                 template = 'firefox/whatsnew/index.html'
         elif channel == 'beta':
```
alexgibson, line 608: I think can can perhaps remove this conditional, and let beta users fallback to whatever the default wnp is?

bedrock/firefox/views.py
```diff
@@ -606,18 +606,8 @@ def get_template_names(self):
...
-                    template = 'firefox/whatsnew/whatsnew-fx70-de.html'
-                elif locale == 'fr':
-                    template = 'firefox/whatsnew/whatsnew-fx70-fr.html'
-                else:
-                    template = 'firefox/whatsnew/index.html'
-            else:
-                template = 'firefox/whatsnew/index.html'
-        elif locale == 'id' and show_firefox_lite_whatsnew(version):
+            template = 'firefox/whatsnew/index.html'
+        elif locale == 'id':
```
alexgibson, line 610: Looks like you may have lost something in the rebase here?

```suggestion
        elif locale == 'id' and show_firefox_lite_whatsnew(version):
```

PR #9195, opened 2020-07-21
Add note to convert QA docs

amychurchwell:
## Description

Adds a note to the convert QA docs. [Reference](https://support.convert.com/hc/en-us/articles/115000009891-Defining-Audiences-for-your-Experience).

### DIFF
--- a/docs/abtest.rst
+++ b/docs/abtest.rst
@@ -64,6 +64,11 @@ During review, the engineer will compare the following to the experiment plan:
 #. The distribution percentages are configured.
 #. The target URLs are configured.
 
+.. Note::
+
+    If you select more than one audience in convert it is joined in an OR fashion, not AND.
+
+
 Once the engineer is satisfied, the engineer (or someone else with write privileges) will:
 
 #. Add ``https://www.mozilla.org/*`` to the list of urls the experiment can run on.


docs/abtest.rst
```diff
@@ -64,6 +64,11 @@ During review, the engineer will compare the following to the experiment plan:
 #. The distribution percentages are configured.
 #. The target URLs are configured.
 
+.. Note::
+
+    If you select more than one audience in convert it is joined in an OR fashion, not AND.
```
alexgibson, line 69: This may make more sense to someone who uses the tool, but to an outsider or QA reviewer - what does this imply?

amychurchwell: I wanted to leave a note that setting more than one audience in convert means that convert treats it as audience X **OR** audience Y and does not combine them, because it's a common misunderstanding. I thought the QA docs would be a helpful place for this note because someone who does not use convert regularly would not know this.

Is there a way we can make it more clear? Or is there a better place for this to live?

alexgibson: I think this is probably the right place for this to live - but I was wondering how to interpret it. Say you're on PTO and I'm asked (theoretically) to review an experiment. If it has multiple audiences set, I can read this and understand it means a logical OR. But then what do I do? Should I suggest changes, ignore it, or something else? 

I guess I'm looking for an action or outcome?

amychurchwell: I guess it depends on the experiment. Here is what I'd like to avoid--
Say, an experiment targets desktop users using chrome. It would be wrong to set the audience with both `Browser Is Chrome` and `Device is Desktop` because convert would say, "okay **ANY** user on desktop." It seems correct on the surface, but all we need is `Browser Is Chrome`.

amychurchwell: Maybe one audience per experiment would solve for this?

alexgibson: Maybe something worded from your last two replies above would be good context to help explain this in the docs? 👍

PR #9198, opened 2020-07-21
Only initialize Sentry when configured

pmac:


pmac:
Things have been reverted and now there's another PR with this change included. Closing.

PR #9210, opened 2020-07-22
Parametrize the download test functions

pmac:
## Description

This will mean that every URL tested will show as a single test in PyTest thus allowing it to parallelize the tests over a number of processes and speeing up tests. This is necessary as there are now 4671 URLs being tested by the download tests.

## Testing

I ran the following command to test:

```
py.test -r a -n 4 --verbose --base-url https://bedrock-dev.gcp.moz.works --reruns 2 -m "download" tests
```


pmac:
@alexgibson I started working on this because I wanted to see why this was taking so long. I saw a "download" job recently that [took over 43 minutes](https://gitlab.com/mozmeao/www-config/-/jobs/651024816)! Making each URL a test means that it should be much faster. We also might want to consider running these tests less often since it's getting to be quite a large number of URLs to test.

duallain:
I like this change a ton.  Makes understanding what's going on inside the pipeline job quite a bit easier.

pmac:
> The question is where - stage perhaps? Then at least it can block a deployment should something be broken.

I was thinking that too. Maybe running it on all dev deployments is a bit much, but stage and prod still seem good. I'm even thinking that maybe we just need it to happen hourly on prod since these links can also be broken by a change in release management. But changing it to no longer run for dev does seems good. Really feels like overkill to run them this much. I'll file a PR over in the www-config repo for that and we can discuss more there.

pmac:
> This looks good to me! I'm not sure if you've tried running this in the pipeline yet, so I'll leave it to you to merge.

That's a good point. This should really go through the pipeline tests. I'll run that now.

alexgibson:
Looks like this passed in just over 12 minutes - seems like a huge improvement? https://gitlab.com/mozmeao/www-config/-/jobs/653379591 👍

pmac:
Hopefully so. It at least looks better and hopefully will be reliably faster. Shall we merge?

alexgibson:
Sounds good to me

Issue #5818, opened 2018-06-14
Move legal-docs content to the DB


pmac:
Following on from #5721 we're looking to further automate the deployment of changes to legal-docs. We want to:

* Move from a git submodule to simply automatically pulling in changes to their repo a la security advisories
* Use a new `production` branch on the legal-docs repo as the source for the content on the website.

We do need the linting from #5721 before we can put this in prod, but we can get this working going simultaneously.

pmac:
Fixed in #9213

PR #9213, opened 2020-07-23
Load and use Legal Docs from the database

### LINKED ISSUE #8385: Automate legal-docs updates
## Description

It would be nice if legal-docs could be updated automatically, without the need for a manual git submodule update.

---

- [ ] Privacy / legal page content can be updated independently to bedrock.

pmac:
Also remove the git submodule \o/

This will need to be deployed to prod in 2 steps. We'll deploy the 1st commit to add the model and fill up the DB. After that is running successfully we can deploy the 2nd commit wich switches the site to use the docs from the DB.

Fix #8385 #5818


pmac:
This doesn't do any validity checking beyond parsing the Markdown into HTML. I'll add a Dead Man's Switch for this job so that it will complete and not cause other data to not load if there are errors, but will also notify us if there are parse errors. The www-sitemap-generator job should tell us if any of the pages start to return 500 errors.

Another thing we should consider is to have the legal docs team use a separate `prod` branch that we can set bedrock prod to use. That way they can merge things and look at them on www-dev before pushing them to prod.

pmac:
Great. I'm waiting to discuss with the legal docs people so they can weigh-in on the process change.

pmac:
I've gotten buy-in from Peiying for the `prod` branch in the legal-docs repo. I've updated this PR to automatically use `master` branch when `DEV=True` and otherwise to use `prod`. I think this might be good to go.

alexgibson:
@pmac shall I merge this, or is there anything else we're waiting on?

pmac:
I think we're good. Yeah. I say we give it a shot 👍 You're free to do the honors.

Issue #9177, opened 2020-07-20
/fraud-report/ - Success notification appears at the bottom of the page


alexgibson:
## Description

Upon sucessfully submitting a fraud-report, the success notification appears right at the bottom of the page, providing no clear indication that the report was sent.

### Steps to reproduce

1. Visit https://www.mozilla.org/en-US/about/legal/fraud-report/
2. Submit the required information.

(you can also visit https://www.mozilla.org/en-US/about/legal/fraud-report/?submitted=True directly)

### Expected result

Upon submission, there should be a notification visible in the viewport confirming that the report was sent.

### Actual result

The notification is shown right at the bottom of the page, which is not visible for most screen sizes. It should probably appear at the top of the page.

![image](https://user-images.githubusercontent.com/400117/87915855-9433fa80-ca6a-11ea-86be-f759b6c5d604.png)

### Environment

N/A

berliners1:
I just submitted a pull request for a potential fix to this issue, also accounting for the red Error notification that can occur there.

PR #9226, opened 2020-07-26
Adjust /fraud-report/ notification location

berliners1:
## Description
I moved both the Success and Error notifications to render at the top instead of bottom, and made the h2 'Fraud Report' no longer render upon successful form submission since it would then have nothing under it.

## Issue / Bugzilla link
#9177

## Testing
For Success notification, visit: http://localhost:8000/en-US/about/legal/fraud-report/?submitted=True

For Error notification, visit: http://localhost:8000/en-US/about/legal/fraud-report/
And enter in an invalid url before submitting

### DIFF
--- a/bedrock/legal/templates/legal/fraud-report.html
+++ b/bedrock/legal/templates/legal/fraud-report.html
@@ -27,6 +27,20 @@
 {% endblock %}
 
 {% block article %}
+
+{% if form_submitted %}
+<aside class="mzp-c-notification-bar mzp-t-success">
+  <p><strong>{{ _('Your report has been sent.') }}</strong></p>
+  <p>{{ _('Thank you for your help. It’s our community that makes us great.') }}</p>
+</aside>
+{% endif %}
+
+{% if form_error %}
+<aside class="mzp-c-notification-bar mzp-t-error">
+  <p><strong>{{ _('An error has occurred with your submission. Please review your information and try again.') }}</strong></p>
+</aside>
+{% endif %}
+
 {# L10n: The line break below is for visual formatting only #}
 <h1 class="mzp-c-article-title">{{_('Protect the Fox<br> (and More!)')}}</h1>
 <h2>{{ _('Help us safeguard Mozilla’s trademarks by reporting misuse.') }}</h2>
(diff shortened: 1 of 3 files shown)


bedrock/legal/templates/legal/fraud-report.html
```diff
@@ -27,6 +27,20 @@ <h3 class="mzp-c-sidemenu-label">{{ self.page_title() }}</h3>
 {% endblock %}
 
 {% block article %}
+
+{% if form_submitted %}
+<aside class="mzp-c-notification-bar mzp-t-success">
```
alexgibson, line 32: Thanks! It would be nice to add some bottom margin below the notification here:

image

Using `margin-bottom: $spacing-xl;` should do it:

image

bedrock/legal/templates/legal/fraud-report.html
```diff
@@ -27,6 +27,20 @@ <h3 class="mzp-c-sidemenu-label">{{ self.page_title() }}</h3>
...
+  <p><strong>{{ _('Your report has been sent.') }}</strong></p>
+  <p>{{ _('Thank you for your help. It’s our community that makes us great.') }}</p>
+</aside>
+{% endif %}
+
+{% if form_error %}
```
alexgibson, line 38: Same applies for the error notification:

image

berliners1:
Ok, I added the margin-bottom to the element, let me know how it looks. Thanks!

media/css/protocol/protocol.scss
```diff
@@ -18,7 +18,7 @@ $image-path: '/media/protocol/img';
 // Already fixed upstream in https://github.com/mozilla/protocol/commit/548e7ce3070372e7fe92f268fb3847e911d5038e
 @media #{$mq-sm} {
     #outer-wrapper .mzp-c-notification-bar {
-        margin: $layout-xs auto 0;
+        margin: $layout-xs auto $spacing-xl;
```
alexgibson, line 21: Whilst this works it also applies the same bottom margin to every notification on the website, which is something we probably don't want at this time. My suggestion would be to update this selector:

https://github.com/mozilla/bedrock/blob/master/media/css/legal/legal.scss#L70-L72

Note: you may need to increase the specificity to get the changes to apply.

berliners1:
Ok, I added the same id of outer-wrapper to legal.scss's class that protocol.scss has and that did the trick. Let me know how that looks. Thanks!

PR #9232, opened 2020-07-28
Make GA event fire before opening the protection report

ericawright:
r? @craigcook 

It seems we are not getting an expected amoung of "link click" events, and the event may not be firing because the page navigates away to the protection report before the event is being added to the data layer. This makes the event fire before the navigation. 

note, this is a bug in all places where protection report is being opened, and will need to be updated in those places in the future.


craigcook:
Marking Do Not Merge for the moment since we may also need to adjust the traffic cop to compensate for the shorter experiment run.

PR #9241, opened 2020-07-29
add data-cta-text params

ericawright:
links below the fold and the main CTA (when JS is not enabled) are lacking data-cta-text params - this fixes that.
Luckily we did included UTM params and so have a decent record of click throughs on below-the-fold links.
r? @craigcook 

~This may not need to be merged, but adding it awaiting on the teams decision.~ 
Team has decided this is good to go. please take note of the timestamp when this is pushed to production

### DIFF
--- a/bedrock/firefox/templates/firefox/welcome/page8.html
+++ b/bedrock/firefox/templates/firefox/welcome/page8.html
@@ -64,7 +64,7 @@
         heading_level=1
       ) %}
       <p class="primary-cta protection-report">
-        <a href="https://support.mozilla.org/kb/enhanced-tracking-protection-firefox-desktop/?utm_source={{ _utm_source }}&utm_medium=referral&utm_campaign={{ _utm_campaign }}&utm_content={{_utm_content}}" class="mzp-c-button mzp-t-product" id="button" rel="external">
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/welcome/page8.html
```diff
@@ -64,7 +64,7 @@
         heading_level=1
       ) %}
       <p class="primary-cta protection-report">
-        <a href="https://support.mozilla.org/kb/enhanced-tracking-protection-firefox-desktop/?utm_source={{ _utm_source }}&utm_medium=referral&utm_campaign={{ _utm_campaign }}&utm_content={{_utm_content}}" class="mzp-c-button mzp-t-product" id="button" rel="external">
+        <a href="https://support.mozilla.org/kb/enhanced-tracking-protection-firefox-desktop/?utm_source={{ _utm_source }}&utm_medium=referral&utm_campaign={{ _utm_campaign }}&utm_content={{_utm_content}}" class="mzp-c-button mzp-t-product" id="button" rel="external" data-cta-text="View your report - no JS" data-cta-type="button">
```
craigcook, line 67: Since these attributes are evidently required even when there's an explicit event, I don't think we should say "no JS" in the name.

craigcook: We should also add `data-cta-position="Primary"` to help further distinguish the big button from the secondary link below.

PR #9250, opened 2020-07-31
Removed jQuery from mozilla-lazy-load.js & mozilla-smoothscroll.js (Fixes #8381)

### LINKED ISSUE #8381: Remove jQuery code from mozilla.smoothscroll.js and mozilla.lazy-load.js libs
## Description

Both `mozilla.smoothscroll.js` and `mozilla.lazy-load.js` libs have portions of jQuery code that can either be removed or converted to native functions.

---

- [ ] Both libraries are updated to use native code only.

vinay-deshmukh:
## Description

For mozilla-lazy-load.js:
1. Only one occurrence of $, which was replaced with querySelectorAll

For mozilla-smoothscroll.js:
1. Removed a unit test spec which expected jQuery functions to be called
2. Modified a unit test spec to not refer to window.jQuery

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/8381

## Testing

I've run the units tests via `karma start karma.conf.js`

### DIFF
--- a/media/js/base/mozilla-lazy-load.js
+++ b/media/js/base/mozilla-lazy-load.js
@@ -73,16 +73,18 @@ if (typeof window.Mozilla === 'undefined') {
      * @param _selector (String) - CSS selector for target images e.g. '.list-item img'.
      */
     LazyLoad.loadAllFallback = function(_selector) {
-        $(_selector).each(function() {
-            var srcset = this.getAttribute('data-srcset');
+        var nodeList = document.querySelectorAll(_selector);
+        for(var ni = 0; ni < nodeList.length; ni++) {
+            var self = nodeList[ni];
+            var srcset = self.getAttribute('data-srcset');
 
             if (srcset) {
-                this.srcset = srcset;
+                self.srcset = srcset;
             }
 
-            this.src = this.getAttribute('data-src');
-            this.onload = LazyLoad.onImageLoad;
-        });
+            self.src = self.getAttribute('data-src');
+            self.onload = LazyLoad.onImageLoad;
+        }
     };
 
     /**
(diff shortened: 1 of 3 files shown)


media/js/base/mozilla-lazy-load.js
```diff
@@ -73,16 +73,18 @@ if (typeof window.Mozilla === 'undefined') {
      * @param _selector (String) - CSS selector for target images e.g. '.list-item img'.
      */
     LazyLoad.loadAllFallback = function(_selector) {
-        $(_selector).each(function() {
-            var srcset = this.getAttribute('data-srcset');
+        document.querySelectorAll(_selector)
+            .forEach(function(currentValue) {
```
alexgibson, line 77: `forEach` on a NodeList isn't supported on Internet Explorer, so we may need to use a traditional `for` loop here.

vinay-deshmukh: Replaced this `forEach` with a normal `for` loop

vinay-deshmukh:
I had one question about making sure the code runs on older browsers like Internet Explorer. What is the repository's policy/opinion on polyfills?

It would be great to use polyfills for new features like `NodeList.forEach` and `window.fetch`, but I believe these polyfills would have to be created in a centralized `.js` file (as opposed to defining them in each file), which would then be imported in files where it's needed.

Are there any plans for such a "polyfill header file"?

alexgibson:
> Are there any plans for such a "polyfill header file"?

We do bundle some polyfills in our common JS bundle, although we currently have to consider the impact of shipping polyfills to all browsers (not just the ones that need them). I think adding small polyfills that are likely commonly used, such as `NodeList.forEach` should be fine, although I don't think we use `fetch` commonly enough to bundle it everywhere. We could do that on a per-bundle basis though.

alexgibson (approved):
Thanks!

PR #9255, opened 2020-08-01
Fix #8293 truncated logo on \mobile for RTL languages

### LINKED ISSUE #8293: Firefox logo truncated on /mobile for some RTL languages
## Description

The Firefox logo on /firefox/mobile/ is truncated at medium screen widths for some RTL locales. This may be more prevalent for locales such as Arabic, which have larger line-heights by default.

### Steps to reproduce

1. Visit https://www-dev.allizom.org/ar/firefox/mobile/
2. Resize the browser window to around 960px.

### Expected result

The logo should not be truncated.

### Actual result

The logo is truncated. The wordmark is also blown up larger than the source file dimensions, so it appears a pixelated. This may be contributing to the truncation.

image

### Environment

Any browser

mdieske:
## Description
Fixed the bug by adding backgound-size attribute for c-wordmark class in the mobile.scss.

## Issue / Bugzilla link
#8293 

## Testing
I have tested the change in local envirioment on the following pages:
localhost:8000/ar/firefox/mobile/
localhost:8000/he/firefox/mobile/
localhost:8000/fa/firefox/mobile/
(Following to check if the change doesn't have negative effects on non RTL)
localhost:8000/de/firefox/mobile/
localhost:8000/en-us/firefox/mobile/

### DIFF
--- a/media/css/firefox/mobile/mobile.scss
+++ b/media/css/firefox/mobile/mobile.scss
@@ -158,7 +158,7 @@ main .mzp-l-content {
 //  Wordmark
 
 .c-wordmark {
-    @include at2x('/media/protocol/img/logos/firefox/browser/logo-word-hor-sm.png', auto, 100%);
+    @include at2x('/media/protocol/img/logos/firefox/browser/logo-word-hor-sm.png', contain);
     @include bidi(((padding-right, $spacing-sm, padding-left, 0),));
     @include image-replaced;
     background-position: top center;
--- a/media/css/firefox/mobile/mobile.scss
+++ b/media/css/firefox/mobile/mobile.scss
@@ -163,7 +163,6 @@ main .mzp-l-content {
     @include image-replaced;
     background-position: top center;
     background-repeat: no-repeat;
-    background-size: contain;
     content: '';
     display: block;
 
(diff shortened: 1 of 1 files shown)


media/css/firefox/mobile/mobile.scss
```diff
@@ -163,6 +163,7 @@ main .mzp-l-content {
     @include image-replaced;
     background-position: top center;
     background-repeat: no-repeat;
+    background-size: contain;
```
alexgibson, line 166: The fix works on low-res screens (nice!), but the `at2x()` mixin is creating another `background-size` property for high-res screens that takes precedence. You may want to try and pass `contain` to the mixin above instead of `auto, 100%`?

mdieske:
Thanks for your review and feedback @alexgibson.
I added your requested change. `contain` is now passed to the `at2x()` mixin. I tested on my local enviroment simulating a high-res screen and it seems to work well.

media/css/firefox/mobile/mobile.scss
alexgibson: There's no need for this duplicate property now that you've specified `contain` in the mixin. Please remove this line.

mdieske: You are right. I thought that this was needed for the two different cases (low-res, high-res) but as the image itself is included by the `at2x()` mixin it's not necessary.

PR #9262, opened 2020-08-03
Allow specification of multiple FTL files for locale activation

### LINKED ISSUE #9251: Older versions of translated pages not linked to newer ones
The /new page was recently redesigned. That means there are now three versions of this page. V1 has been migrated to fluent and has wide translation support. V2 remains in lang files. V3 is our shiny new one. 

The language selector in the footer of the English page only includes the locales which have translated V3 and it should be possible to access all translations from the drop down.

pmac:
## Issue / Bugzilla link

Fix #9251

## Testing

I tested locally by setting `DEV=False` and `SWITCH_NEW_REDESIGN=on`. With the current code you would see the new page but not the lang switcher, and with these changes you should see the full list.

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -662,6 +662,7 @@ class DownloadThanksView(L10nTemplateView):
         'firefox/new/trailhead/exp-thanks.html': ['firefox/new/download'],
         'firefox/new/desktop/thanks.html': ['firefox/new/desktop'],
     }
+    ftl_activations = ['firefox/new/download', 'firefox/new/desktop']
 
     # place expected ?v= values in this list
     variations = ['a', 'b', 'c']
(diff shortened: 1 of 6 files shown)


bedrock/firefox/views.py
```diff
@@ -704,6 +704,7 @@ class NewView(L10nTemplateView):
         'firefox/new/trailhead/download-yandex.html': ['firefox/new/download', 'banners/firefox-mobile'],
         'firefox/new/desktop/download.html': ['firefox/new/desktop'],
     }
+    ftl_activations = ['firefox/new/download', 'firefox/new/desktop']
```
alexgibson, line 707: Should we also add this to [DownloadThanksView](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/views.py#L660), since that URL shares the same localizations file(s) as /new?

pmac: Good catch. Done.

Issue #9253, opened 2020-07-31
Update /new exp template with new redesign.


amychurchwell:
## Description

Now that the firefox/new redesign landed in #9202 we should update the corresponding exp template to match.

amychurchwell:
Fixed in #9263

PR #9263, opened 2020-08-03
Update /new exp template with new redesign. (#9253)

amychurchwell:
## Description
Updates the exp /new template and corresponding JS/CSS to match the new design of /new

## Issue / Bugzilla link
#9253

## Testing
- The redesign seems to have a lot of functionality, while exp templates are more simplified versions we should make sure that the expected behavior is the same.
- I decided that both templates should share the same fluent file so future copy changes don't require a manual dev task.


bedrock/exp/templates/exp/firefox/new/download.html
```diff
@@ -2,50 +2,500 @@
...
+  </aside>
+  {% endif %}
+
 </main>
 {% endblock %}
+
+{% block js %}
+  {{ js_bundle('firefox_desktop_download') }}
+
+  {% if show_firefox_join_modal %}
```
alexgibson, line 494: This variable is not set anywhere, so the modal script isn't loading?

alexgibson:
Loading http://localhost:8000/de/exp/firefox/new/ shows the old page design (sans any CSS). Since you're now using localization files you can remove `download.de.html`?

bedrock/exp/templates/exp/firefox/new/download.html
```diff
@@ -2,50 +2,500 @@
...
+
+    {{ download_firefox(dom_id='download-join-firefox-modal', alt_copy=ftl('firefox-desktop-download-just-download-the-browser'), button_color='mzp-t-secondary', locale_in_transition=True, download_location='other') }}
+  </aside>
+  {% endif %}
+
 </main>
 {% endblock %}
+
+{% block js %}
+  {{ js_bundle('firefox_desktop_download') }}
```
alexgibson, line 492: Do you still need to create bundles for the CSS and JS files you've duplicated? The /exp/ page seems to be loading the originals.

bedrock/exp/templates/exp/firefox/new/download.html
amychurchwell: I noticed this on the original page and did not want to remove it because I don't have the context for what it's for. Maybe @stephaniehobson knows more?

alexgibson: The original page has the variable defined here: https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/new/desktop/download.html#L8

bedrock/exp/templates/exp/firefox/new/download.html
amychurchwell: ah whoops, I missed the JS. Thank you.

bedrock/exp/templates/exp/firefox/new/download.html
amychurchwell: Ah, okay. I will add that back in right now. I must've not realized what I was looking at.

media/js/exp/firefox/new/desktop/join-modal.js
```diff
@@ -0,0 +1,49 @@
...
+        e.preventDefault();
+
+        // Open the modal
+        Mzp.Modal.createModal(this, joinFirefoxContent, {
+            title: e.target.textContent,
+            className: 'join-firefox-modal'
+        });
+
+        // Count the click in GA
+        window.dataLayer.push({
```
amychurchwell, line 21: @alexgibson Should I add an experiment specific value to this dataLayer push?

alexgibson: THe event should be tied to the URL in GA already, so I don't think there should be anything extra required (although I'm no GA expert)

amychurchwell: That sounds right. Thanks! I think I've addressed all the feedback now.

bedrock/exp/templates/exp/firefox/new/download.html
```diff
@@ -2,50 +2,518 @@
...
+
 {% extends "exp/firefox/new/base.html" %}
 
-{% block page_title_prefix %}Download Firefox — {% endblock %}
+{% block extrahead %}
+  {{ super() }}
 
-{% block page_title %}Free Web Browser{% endblock %}
+  <!--[if IE 9]>
+    {{ css_bundle('firefox_desktop_download_ie9') }}
```
alexgibson, line 20: Are you OK with using the source originals for these assets, vs making /exp/ ones?

amychurchwell: yes, I think that is fine for IE styles.

PR #9267, opened 2020-08-04
Allow views to specify multiple ftl or template files for lang activation list

pmac:
## Description

Follup to #9262. This changes the `ftl_activations` parameter to `activation_files` and allows devs to include template filenames in the list so that .lang activations will be included.

## Issue / Bugzilla link

Re #9251

## Testing

I tested by commenting out the template file name in the `activation_files` parameter on the `NewView` class and looking at the list of locales in the source of the `/firefox/new/` page. Then I uncommented the line and looked at the list and saw that it was significantly longer.

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -663,7 +663,11 @@ class DownloadThanksView(L10nTemplateView):
         'firefox/new/trailhead/exp-thanks.html': ['firefox/new/download'],
         'firefox/new/desktop/thanks.html': ['firefox/new/desktop'],
     }
-    ftl_activations = ['firefox/new/download', 'firefox/new/desktop']
+    activation_files = [
+        'firefox/new/download',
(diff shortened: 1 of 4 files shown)


bedrock/firefox/views.py
```diff
@@ -663,7 +663,7 @@ class DownloadThanksView(L10nTemplateView):
         'firefox/new/trailhead/exp-thanks.html': ['firefox/new/download'],
         'firefox/new/desktop/thanks.html': ['firefox/new/desktop'],
     }
-    ftl_activations = ['firefox/new/download', 'firefox/new/desktop']
+    activation_files = ['firefox/new/download', 'firefox/new/desktop']
```
alexgibson, line 666: Add `firefox/new/protocol/thanks.html` here?

pmac: Done.

Issue #8860, opened 2020-04-29
[Epic] Make a simplified version of the download button


alexgibson:
## Description

The current download button helper is versatile in the way that it supports multiple products (release, beta, nightly etc) and can also have the option to download directly (vs link to `/download/thanks/`). A large proporation of use-cases on the website exist to just serve the main Firefox release verson however, and all link directly to the `/download/thanks/` page.

For those use-cases, the HTML could probably be simplified to just the following:

```html
<div class="mzp-c-button-download-container">
  <a href="/%LOCALE%/firefox/download/thanks/" class="mzp-c-button mzp-t-product" data-direct-link="https://download.mozilla.org/?product=firefox-stub&amp;os=win&amp;lang=%LOCALE%">
    Download Firefox
  </a>
  <small class="mzp-c-button-download-privacy-link">
    <a href="/%LOCALE%/privacy/firefox/">Firefox Privacy Notice</a>
  </small>
</div>
```

Note that the `data-direct-link` attribute would still need to exist, but only to serve Windows builds to legacy IE users (IE9 and below).

The other thing that's omitted here is all the conditional markup for [unsupported messaging]( https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/includes/download-button.html#L42-L59), such as:

```html
<p class="unsupported-download">
  Your system doesn't meet the <a href="/firefox/system-requirements/">requirements</a> to run Firefox.
</p>
 ```

Having all this content in every download button, often 2 or 3 times per page, adds a lot of hidden content that's often not needed to the user. This can add to page weight, make our content harder to understand for crawlers, and can also surface as confusing messages in search results (see https://github.com/mozilla/bedrock/issues/8739).

Instead of having this content in every page of the website, we could move it to `/download/thanks/`. This page could then serve as the main place for such conditional messaging, allowing us to use a simplified download button helper for most of the site.

pmac:
Love it. Concentrating the complexity on the `/thanks/` page makes a lot of things easier to reason about, and potentially offer a better UX.

mitchell-frost:
Can I work on this?

alexgibson:
Depends on https://github.com/mozilla/bedrock/issues/9102

alexgibson:
Depends on #8454

alexgibson:
Depends on #9137

alexgibson:
[...]

Issue #7166, opened 2019-05-08
Remove jQuery as a dependency on /firefox/download/thanks/


alexgibson:
## Description

Right now, the `/firefox/download/thanks/` page relies on jQuery's non-standard `:visible` selector in order to [determine which download URL to trigger](https://github.com/mozilla/bedrock/blob/master/media/js/firefox/new/scene2.js#L10). All this is hinged upon a [crafty piece of CSS](https://github.com/mozilla/bedrock/blob/master/media/css/firefox/new/scene2.scss#L13-L25) that hides the button visually on the page, but still permits jQuery to consider it as visible for the auto-download to work. This feels like a brittle dependency between our (complex) download button CSS, jQuery's non-standard selector engine, and some page-specific CSS trickery.

Given that we have hopes of deprecating jQuery on bedrock, and coupled with the news that future versions of jQuery will [remove the Sizzle selector engine](https://blog.jquery.com/2019/04/10/jquery-3-4-0-released/), it's about time we pay off this piece of technical debt and refactor how we handle auto-downloading.

I'd like to suggest that we switch to a vanilla JS solution that can be more easily unit tested.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [ ] jQuery is removed as a dependency in `scene2.js`.
- [ ] jQuery is removed as a dependency from `mozilla-pixel.js`.
- [ ] Vanilla JS solution now handles auto-downloading.

#### :heavy_exclamation_mark: Risks :heavy_exclamation_mark:
- [ ] There's a risk of error that could impact downloads.

#### Tasks
- [ ] Implement vanilla JS solution.
- [ ] Write unit tests for the new code.

alexgibson:
Note there is also another file on this page [mozilla-pixel.js](https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-pixel.js) that also uses jQuery.

alexgibson:
I'm going to tackle this as part of https://github.com/mozilla/bedrock/issues/9102

PR #9274, opened 2020-08-05
Move download specific messaging to the /thanks/ page (Fixes #9102) 

### LINKED ISSUE #9102: Move download specific messaging to the /thanks/ page
## Description

As a dependency to making a simplified version of the download button in https://github.com/mozilla/bedrock/issues/8860, we need to first find an alternate solution for the various pieces of [conditional messaging](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/includes/download-button.html#L46) contained in the full download button helper.

After some discussion we have agreed that the best place for this messaging to live is in the /download/thanks/ page.

This issue to to track migration of that messaging.

---

#### Success Criteria
- [x] Conditional messaging contained in the download button is shown on /thanks, instead of on the page where the download click originates from.

#### Approvals
- [ ] The Accountable
- [ ] Brand - design?
- [ ] Brand - copy?
- [ ] Team Lead
- [ ] Others?

alexgibson:
## Description
- Moves download specific platform messages to the `/thanks` page.
- Removes _"Your system does not meet the requirements to run Firefox"_ from download buttons and defaults to _"Your system may not meet the requirements for Firefox, but you can try one of these versions:"_.
- Removes unused / ancient platform detection from `site.js`.
- Removes jQuery dependency from `thanks.js`.
- Adds tests for `thanks.js` download logic.

Note: this PR does not yet introduce the simplified download button helper, that will happen in https://github.com/mozilla/bedrock/issues/8860

Demo: https://www-demo3.allizom.org/en-US/firefox/new/

## Issue / Bugzilla link
#9102
#8454
#7166

## Testing
- The most reliable way to test this is to use BrowserStack on the demo link above using real devices. 
- Failing that, you can also try a [UA switcher](https://addons.mozilla.org/en-US/firefox/addon/uaswitcher/) which may work for some cases.
- Finally, it's also possible to edit `base/site.js`locally to return the platform name you want to test. 

Please test the following conditions when viewing /thanks:

- [x] Platforms that identify as `windows`, `osx`, `linux`, `ios`, and `android` should all still auto-download / redirect to the app store as expected.
- [x] Platforms that identify as `windows xpvista` should see the message `Mozilla no longer provides security updates on Windows XP or Vista, but you can still use Firefox.` underneath the page heading.
- [x] Platforms that identify as `other` should see the message `Your system may not meet the requirements for Firefox, but you can try one of these versions.` (linking to the /firefox/all/ page). Browsers with JS disabled should also see this same message.
- [x] Users with JS disabled should see the same message as above.
- [x] Platforms that identify as `linux arm` should see the message `Please follow these instructions to install Firefox.`
- [x] Verify all of the above is true for both the [current](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/new/desktop/thanks.html) and [legacy](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/new/trailhead/thanks.html) /thanks templates.

Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/174604268


alexgibson:
Ok, I think this is ready for review. There are some pretty big changes here, so could use some careful review / testing.

I'd be happy to walk/talk through this with whomever does review.

alexgibson:
> Browserstack doesn't have Linux images (as far as I know) so I'm not sure how to verify it, but the other OSes are correctly detected so I'll wager Linux is too. I can at least verify that the correct message is visible when I change the class locally.

Yeah, I did get a couple of different people at Mozilla to help verify that Linux was serving the correct download a week or so ago. So that seems to be work as as it was before, as far as I can see.

> So this is a tentative r+ from me me but I'm always nervous about anything that could seriously impact downloads so I don't know how much more thoroughly we should test this. And since it took me so long to review I'm also nervous about pushing this to production right before a long weekend. Maybe we should wait until next week to merge.

This is a great call out. Let's certainly not merge this before Tuesday next week. I'll also loop in Justin and see if he has any recommendations here as to further testing we could do.

Issue #9215, opened 2020-07-24
Fenix Launch Support: Make /mobile more accurate BY AUG 18


hoosteeno:
## Description

On www.mozilla.org/firefox/mobile, in the extensions section, remove the hyperlink and all text after the word "extensions". See attached mockup.

![Screen Shot 2020-08-05 at 11 50 21 AM](https://user-images.githubusercontent.com/94519/89446577-ee42ea00-d711-11ea-9436-8d33750019b1.png)

---

#### Success Criteria
- [x] Designs are complete and attached.
- [ ] Code is merged and live before Aug. 18

#### Approvals
- [ ] Dev
- [x] PMM

PR #9275, opened 2020-08-05
Fix 9215: Remove hyperlink and copy from /mobile.

hoosteeno:
## Issue / Bugzilla link

#9215

### DIFF
--- a/bedrock/firefox/templates/firefox/mobile/index.html
+++ b/bedrock/firefox/templates/firefox/mobile/index.html
@@ -94,17 +94,19 @@
     </section>
 
     <section>
-      <div class="mzp-l-content t-extend">
-        {% set extend_title = '<span>'|safe + ftl('firefox-mobile-android-only') + '</span><br>'|safe + ftl('firefox-mobile-make-android-your-own') %}
-        {% call feature_card(
-          title=extend_title,
-          image_url='img/firefox/mobile/protocol/screen-frame.svg',
-          aspect_ratio='mzp-has-aspect-1-1',
-          class='mzp-l-card-feature-right-half mzp-t-firefox',
-        ) %}
-        <p>{{ ftl('firefox-mobile-customize-your-firefox-mobile', url='https://addons.mozilla.org/firefox/extensions/') }}</p>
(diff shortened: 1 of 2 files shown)


bedrock/firefox/templates/firefox/mobile/index.html
```diff
@@ -102,7 +102,7 @@ <h3>{{ ftl('firefox-mobile-stop-companies-from-following') }}</h3>
           aspect_ratio='mzp-has-aspect-1-1',
           class='mzp-l-card-feature-right-half mzp-t-firefox',
         ) %}
-        <p>{{ ftl('firefox-mobile-customize-your-firefox-mobile', url='https://addons.mozilla.org/firefox/extensions/') }}</p>
+        <p>{{ ftl('firefox-mobile-customize-your-firefox-mobile') }}</p>
```
stephaniehobson, line 105: I'm sorry this is complicated but I think removing the variable will cause errors in other locales.

Options are:
1) change the token name at the same time you change the string contents
2) keep the token unchanged as a [fallback](https://bedrock.readthedocs.io/en/latest/l10n.html#the-ftl-helper-function) and add a new token for the shorter string

stephaniehobson: Possible option 3, update the string in English to remove the text & link but continue passing the URL for locales which haven't caught up.

pmac: My understanding is that unless we use a new string ID no one will be notified to update their translation. So I think #2 is the way to go: create a new string ID for the new string, and use the old one as the fallback, keeping the parameter in the call for the ones that haven't yet translated the new string.

hoosteeno: The point of this change is to avoid making false promises about a product. We shouldn't wait for l10n completeness in order to avoid making those promises. So I'll take option 4: wrap the entire "extensions" card in a conditional that depends on the presence of the new string. In that case I think (based on code search) that we can remove the old ftl token. PR forthcoming.

PR #9277, opened 2020-08-05
Adjust Legal Docs Cache Timeout

pmac:
It was set to an hour for dev and prod. In the new system 10min on prod and 1min on dev is good and helps get changes out faster for the legal team.


pmac:
I'm going to go ahead and merge this. It's a tiny change, but one I need in dev so that the legal-team can more easily see the process.

PR #9287, opened 2020-08-06
Explore options for featuring previous Nightly builds on www.mozilla.org 

### LINKED ISSUE #9271: Explore options for featuring previous Nightly builds on www.mozilla.org
## Description

Originally reported in - https://bugzilla.mozilla.org/show_bug.cgi?id=1614106

A link to get previous Nightly releases on https://www.mozilla.org/en-US/firefox/channel/desktop/#nightly would be nice. The use case for this is for people who are trying to do a rough bisection for a regression and need old nightlies, but don't know that they need to go dig around in ftp.mozilla.org to get these. Or for people like myself who don't want to click all the way through ftp.mozilla.org's directory hierarchy.

The recommendation is to add a link on the Nightly download page to https://ftp.mozilla.org/pub/firefox/nightly/

---

#### Success Criteria
- [ ] Tasks should be represented here as something like "the new design is done and approved, live on the page"

#### Approvals
- [ ] The Accountable
- [ ] Brand - design?
- [ ] Brand - copy?
- [ ] Team Lead
- [ ] Others?

khushi1234945:
https://bugzilla.mozilla.org/show_bug.cgi?id=1614106
Closes #9271

### DIFF
--- a/bedrock/firefox/templates/firefox/channel/desktop.html
+++ b/bedrock/firefox/templates/firefox/channel/desktop.html
@@ -96,6 +96,7 @@
           {{ ftl('firefox-channel-all-languages-and-platforms', fallback='firefox-channel-all-languages-and-builds') }}
         </a>
       </li>
+      <li><a href="https://ftp.mozilla.org/pub/firefox/nightly/">{{ ftl('firefox-channel-previous-nightly-builds') }}</a></li>
     </ul>
     <p>
       {% if ftl_has_messages('firefox-channel-nightly-is-an-unstable-testing') %}
--- a/l10n/en/firefox/channel.ftl
+++ b/l10n/en/firefox/channel.ftl
@@ -45,6 +45,7 @@ firefox-channel-download-and-test-the-latest-desktop = Download and test the lat
 firefox-channel-beta = { -brand-name-beta }
 firefox-channel-test-about-to-be-released = Test about-to-be-released features in the most stable pre-release build.
 firefox-channel-release-notes = Release Notes
+firefox-channel-previous-nightly-builds = Previous Nightly Builds
 firefox-channel-developer-edition = { -brand-name-developer-edition }
 firefox-channel-build-test-scale-and-more = Build, test, scale and more with the only browser built just for developers.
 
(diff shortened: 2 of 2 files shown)


bedrock/firefox/templates/firefox/channel/desktop.html
```diff
@@ -96,6 +96,7 @@
           {{ ftl('firefox-channel-all-languages-and-platforms', fallback='firefox-channel-all-languages-and-builds') }}
         </a>
       </li>
+      <li><a href="https://ftp.mozilla.org/pub/firefox/nightly/">{{ ftl('firefox-channel-previous-nightly-builds') }}</a></li>
```
craigcook, line 99: This string doesn't exist in the Fluent file at `/firefox/channel.ftl`. It needs to be added there so it can be rendered on the page and localized in other languages. See https://bedrock.readthedocs.io/en/latest/l10n.html

Without a text string defined the template will display the string ID:
image

khushi1234945: String is added in the Fluent file at /firefox/channel.ftl

khushi1234945:
String is added in the Fluent file at /firefox/channel.ftl

khushi1234945:
Hi @craigcook ! Thanks for the review!
I added the string in the fluent file at /firefox/channel.ftl

l10n/en/firefox/channel.ftl
```diff
@@ -45,6 +45,7 @@ firefox-channel-download-and-test-the-latest-desktop = Download and test the lat
 firefox-channel-beta = { -brand-name-beta }
 firefox-channel-test-about-to-be-released = Test about-to-be-released features in the most stable pre-release build.
 firefox-channel-release-notes = Release Notes
+firefox-channel-previous-nightly-builds = Previous Nightly Builds
```
craigcook, line 48: Sorry I didn't reply sooner, it's been a crazy week.

Our brand names get special treatment for translation, and those names are kept in a separate file that we reference as variables in strings. We should format this string like:

```suggestion
firefox-channel-previous-nightly-builds = Previous { -brand-name-nightly } Builds
```

khushi1234945:
Hi @craigcook 
I commit the suggestion which was provded by you.

khushi1234945:
Thank you too @craigcook

PR #9294, opened 2020-08-10
Add Firefox Relay terms of service (Fixes #9240)

### LINKED ISSUE #9240: Push Private Relay ToS to https://www.mozilla.org/about/legal/terms/firefox-relay/
cc: @mafeld

Hi @pmac and @alexgibson, 

I'm writing to see if you could push the Firefox Relay Terms of Service to the following url: **https://www.mozilla.org/about/legal/terms/firefox-relay/**

The ToS are here: **https://github.com/mozilla/legal-docs/blob/master/firefox_relay_ToS/en-US.md**

As soon as possible would be good, but August 24th at the latest. @groovecoder let me know if I've got this date wrong.

Many thanks!
Emily

alexgibson:
## Description
http://localhost:8000/en-US/about/legal/terms/firefox-relay/

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/9240

### DIFF
--- a/bedrock/legal/templates/legal/terms/firefox-relay.html
+++ b/bedrock/legal/templates/legal/terms/firefox-relay.html
@@ -1,0 +1,7 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "legal/docs-base.html" %}
+
+{% block page_title %}Firefox Relay Terms of Service{% endblock %}
--- a/bedrock/legal/urls.py
+++ b/bedrock/legal/urls.py
@@ -46,6 +46,9 @@ urlpatterns = (
     url(r'^terms/firefox-private-network/$', LegalDocView.as_view(template_name='legal/terms/firefox-private-network.html',
         legal_doc_name='Firefox_Private_Network_ToS'), name='legal.terms.firefox-private-network'),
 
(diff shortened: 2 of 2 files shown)


bedrock/legal/templates/legal/terms/firefox-relay.html
```diff
@@ -0,0 +1,7 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "legal/docs-base.html" %}
+
+{% block page_title %}{{ _('Firefox Relay: About Your Rights') }}{% endblock %}
```
pmac, line 7: This is a new string right? Should we use Fluent instead?

alexgibson: These legal pages all appear to use this type of title string that has never been localised. We should figure out what to do about that, and it should definitely use Fluent if a string is needed. For now though, I think this is probably OK as it sticks with the existing convention?

alexgibson: Updated to a hard-coded English string for now.

pmac: Yeah. Good point. I'm wondering if we couldn't just yank the title from the document itself for use in the title, but we should save that for a larger rework of all of these. Hardcoded or just leaving it like the others is fine with me 👍

alexgibson: Yeah let’s tackle this when we port the legal pages to Fluent. Thanks @pmac !

PR #9301, opened 2020-08-10
Open cookies portion via cookies link (Fixes #8354)

### LINKED ISSUE #8354: Cookies link in footer goes to closed accordion
## Description

https://www.mozilla.org/en-US/privacy/websites/#cookies which is linked to from the site footer does not take the user to the cookie section of the privacy policy.

### Steps to reproduce

Go to mozilla.org and click on "cookies".

### Expected result

Link to the cookie portion of the websites privacy notice.

### Actual result

View of the footer on the privacy notice.

mdieske:
## Description
Open the cookies portion of the website by javascript if it is visited via the cookies link. 

## Issue / Bugzilla link
#8354 

## Testing
Tested on local enviroment in different languages by clicking on "cookies" in the footer.

### DIFF
--- a/media/js/privacy/privacy-protocol.js
+++ b/media/js/privacy/privacy-protocol.js
@@ -21,13 +21,14 @@
 })();
 
 if (window.location.href.includes('cookies')) {
-	window.onload = function() {
-		
-		var cookieButton = document.querySelector('#user-choices button');
-		var cookieText = document.getElementById('expand-formatparagraphssummary-2');
-		
-		cookieButton.setAttribute('aria-expanded', 'true');
-		cookieText.classList.remove('is-closed');
-		cookieText.setAttribute('aria-hidden', 'false');
-		}
-	};
+    window.onload = function() {
+        'use strict';
+        
+        var cookieButton = document.querySelector('#user-choices button');
+        var cookieText = document.getElementById('expand-formatparagraphssummary-2');
+        
+        cookieButton.setAttribute('aria-expanded', 'true');
+        cookieText.classList.remove('is-closed');
+        cookieText.setAttribute('aria-hidden', 'false');
+        };
+    }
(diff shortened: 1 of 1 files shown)


media/js/privacy/privacy-protocol.js
```diff
@@ -19,3 +19,16 @@
         buttons[i].setAttribute('data-close', closeText);
     }
 })();
+
+if (window.location.href.includes('cookies')) {
```
alexgibson, line 23: Suggest using `window.location.hash` here to be a little more explicit:
```suggestion
if (window.location.hash === '#cookies') {
```

media/js/privacy/privacy-protocol.js
```diff
@@ -19,3 +19,16 @@
...
+    window.onload = function() {
```
alexgibson, line 24: There should be no need to override the `onload` event handler here? Suggest moving this logic inside the closure above, together with the existing code for this page.

media/js/privacy/privacy-protocol.js
```diff
@@ -19,3 +19,16 @@
...
+        'use strict';
+        
+        var cookieButton = document.querySelector('#user-choices button');
+        var cookieText = document.getElementById('expand-formatparagraphssummary-2');
+        
+        cookieButton.setAttribute('aria-expanded', 'true');
```
alexgibson, line 30: These lines will throw errors should the selectors above not be found, and since the HTMl on this page is generated via markdown it could be a little brittle.

To avoid throwing errors should something go wrong, I'd suggest adding a check that both selectors have found a match:

```javascript
if (cookieButton && cookieText) {
    cookieButton.setAttribute('aria-expanded', 'true');
    cookieText.classList.remove('is-closed');
    cookieText.setAttribute('aria-hidden', 'false');
}
```

media/js/privacy/privacy-protocol.js
mdieske: Good suggestion, thank you!

media/js/privacy/privacy-protocol.js
mdieske: You are right. Thought it was necessary but defining the variables after the existing code works also.

mdieske:
Thank you for your feedback and suggestions @alexgibson. As I am a beginner this is very valuable for me. I'll commit your requested changes.

PR #9320, opened 2020-08-21
WNP80, hide Android promo with switch for en-US/CA [fix #9319]

### LINKED ISSUE #9319: WNP80, hide Android promo blurb from en-US until 27 August
## Description

https://www.mozilla.org/en-US/firefox/80.0/whatsnew/all/

The new Firefox for Android is releasing to most locales on the 25th, the same date as Firefox 80 for desktop. But it's being delayed in en-US until the 27th. We shouldn't promote the new Android browser in en-US until it's actually available in that locale.

We'll put the blurb behind a conditional switch so it will be hidden for en-US users until we activate it on the 27th.

---

#### Success Criteria
- [x] The Android promo is hidden in en-US by default
- [ ] A switch will reveal the Android promo in en-US when needed

#### Approvals
- [x] Lauren
- [x] Mei
- [ ] Hoosteeno

craigcook:
## Description

The new Firefox for Android is releasing to most locales on the 25th (the same date as Firefox 80 for desktop) but it's being delayed in en-US and en-CA until the 27th. WNP80 promotes the new Android browser, but we shouldn't show that promo to en-US and en-CA until it's actually available in those locales.

This puts the Android blurb behind the switch `firefox-whatsnew80-daylight-en-us` only for en-US and en-CA; all other locales will be able to see it by default.

This also fixes a typo, but avoids altering the original string by treating en-US as a locale (I've already submitted the same via Pontoon for en-CA and en-GB).

We need to get this to production on **Monday the 24th** ahead of the Fx80 release.

## Issue / Bugzilla link
#9319 

## Testing
http://localhost:8000/en-US/firefox/80.0/whatsnew/

- [x] In Prod mode (switch off), the Android promo is hidden in en-US and en-CA
- [x] In Prod mode (switch off), the Android promo is visible in other locales
- [x] In Dev mode (switch on), the Android promo is visible in en-US and en-CA (as well as all others)
- [ ] English locales show "lightning" instead of "lightening"

### DIFF
--- a/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx80.html
+++ b/bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx80.html
@@ -66,7 +66,7 @@ data-whatsnew80-redirect-url="{{ settings.FXA_ENDPOINT + 'sms/?' + utm_params }}
 
   <section class="content-extra">
     <div class="mzp-l-content">
-      <div class="l-columns-three">
+      <div class="{% if switch('firefox-whatsnew80-daylight-en-us', ['en-US', 'en-CA']) %}l-columns-three{% elif LANG in ['en-US', 'en-CA'] %}l-columns-two{% else %}l-columns-three{% endif %}">
 
         <div class="c-picto-block">
           <div class="c-picto-block-image">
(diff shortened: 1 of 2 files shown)


alexgibson (approved):
I still see `lightening-fast` for en-GB but this may just be because the new string hasn't caught up in Pontoon just yet. Codewise this LGTM 👍

alexgibson:
After merging this it occured to me that you may not technically have needed a new string to change the typo for the en-US file, since Fluent uses unique IDs vs the original copy. Still, what you have should work ok.

craigcook:
> After merging this it occured to me that you may not technically have needed a new string to change the typo for the en-US file, since Fluent uses unique IDs vs the original copy. Still, what you have should work ok.

🤦 D'oh, you're right... I'm still thinking in .lang and didn't want to change the string, but it's just the ID we need to worry about now. Silly me.

PR #9326, opened 2020-08-25
Hide Firefox Send [fix #9316]

### LINKED ISSUE #9316: Remove Send from www.mozilla.org
## Description

Firefox Send is currently suspended and we're uncertain when it will be active again. So meanwhile we're promoting and linking to a service that nobody can actually use. We should remove promotional links to Send until it's back in service.

If we assume this is temporary we can just comment them out, or perhaps put them behind a switch so we can quickly restore them.

It also doesn't need to be an exhaustive purge of every mention, just prominent spots like in the nav and on the main Firefox pages.

---

#### Success Criteria
- [ ] Prominent promotions of Firefox Send are removed/hidden.
- [ ] In such a way that they can be restored when/if Send is available again.

craigcook:
## Description
Removes promotional references/links to Firefox Send, hiding them temporarily behind the switch `firefox-send-active`. Send is hidden by default when the switch is OFF, so we can quickly switch it ON and unhide Send when/if Send becomes available again (and we'd follow that up with some cleanup to remove the switch and make it permanent again).

## Issue / Bugzilla link
#9316 

## Testing
Note that the switch hides Send when it's OFF and switches are treated as ON in dev mode. You'll need to test locally in prod mode to verify it's hidden.

Send should be removed from:

Main nav
Footer
http://localhost:8000/exp/firefox/
http://localhost:8000/firefox/accounts/
http://localhost:8000/exp/firefox/accounts/
http://localhost:8000/firefox/privacy/products/
http://localhost:8000/firefox/welcome/3/

### DIFF
--- a/bedrock/firefox/templates/firefox/accounts.html
+++ b/bedrock/firefox/templates/firefox/accounts.html
@@ -100,7 +100,7 @@
       {% endif %}
     </h2>
 
-    <ul class="c-product-list">
+    <ul class="c-product-list {% if switch('firefox-send-active') %}l-columns-two{% else %}l-columns-three{% endif %}">
       <li class="c-product-list-item t-product-firefox">
         <a href="{{ url('firefox.new') }}">
           <h3 class="c-product-list-title">{{ ftl('firefox-accounts-firefox-browser') }}</h3>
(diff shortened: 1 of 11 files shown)


bedrock/firefox/templates/firefox/accounts.html
```diff
@@ -100,7 +100,7 @@ <h2 class="c-section-title">
       {% endif %}
     </h2>
 
-    <ul class="c-product-list">
+    <ul class="c-product-list {% if switch('firefox-send-active') %}l-columns-two{% else %}l-columns-three{% endif %}">
```
alexgibson, line 103: Should we also apply this to the /exp/firefox/accounts/ page? (the CSS changes might need updating there too).

alexgibson (changes requested):
One visual difference spotted between /firefox/accounts/ and /exp/firefox/accounts/, but otherwise this lgtm 👍

craigcook:
Updated!

alexgibson (approved):

PR #9345, opened 2020-08-28
Make simplified download button (Fixes #8860)

### LINKED ISSUE #8860: [Epic] Make a simplified version of the download button

alexgibson:
## Description
- Adds a new simplified download button helper.
- Implements the simple button on the following pages / areas:
  - Download button in the global navigation.
  - Download buttons on http://localhost:8000/en-US/firefox/new/
  - Download buttons on home pages:
    - http://localhost:8000/en-US/
    - http://localhost:8000/de/
    - http://localhost:8000/fr/
    - http://localhost:8000/es-ES/

## Issue / Bugzilla link
#8860

## Testing
Demo: https://www-demo2.allizom.org/en-US/firefox/new/

- [x] Clicking download on a desktop browser navigates to /thanks and initiates a download as expected.
- [x] Clicking download in an old IE browser initiates a download via a popup, before navigating to /thanks as expected.
- [x] Clicking download on Android / iOS navigated to /thanks before redirecting to the app/play store.

Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/188897807

### DIFF
--- a/bedrock/firefox/templates/firefox/includes/download-button-thanks.html
+++ b/bedrock/firefox/templates/firefox/includes/download-button-thanks.html
@@ -1,0 +1,19 @@
+<div class="mzp-c-button-download-container c-button-download-thanks">
+  <a href="{{ transition_url }}"
+     id="{{ id }}"
+     class="mzp-c-button mzp-t-product {% if button_class %}{{ button_class }}{% else %}mzp-t-xl{% endif %}"
+     data-direct-link="{{ download_link_direct }}"
+     data-link-type="download"
+     {% if download_location %}data-download-location="{{ download_location }}"{% endif %}>
+    {% if alt_copy %}
+      {{ alt_copy }}
+    {% else %}
+      {{ ftl('download-button-download-firefox') }}
+    {% endif %}
+  </a>
+  <small class="mzp-c-button-download-privacy-link">
+    <a href="{{ url('privacy.notices.firefox') }}">
+      {{ ftl('download-button-firefox-privacy-notice') }}
+    </a>
+  </small>
+</div>
--- a/bedrock/firefox/templates/firefox/includes/download-button-thanks.html
+++ b/bedrock/firefox/templates/firefox/includes/download-button-thanks.html
@@ -1,7 +1,6 @@
-<div class="mzp-c-button-download-container c-button-download-thanks">
+<div id="{{ id }}" class="mzp-c-button-download-container c-button-download-thanks">
   <a href="{{ transition_url }}"
-     id="{{ id }}"
-     class="mzp-c-button mzp-t-product {% if button_class %}{{ button_class }}{% else %}mzp-t-xl{% endif %}"
+     class="download-link mzp-c-button mzp-t-product {% if button_class %}{{ button_class }}{% else %}mzp-t-xl{% endif %}"
      data-direct-link="{{ download_link_direct }}"
      data-link-type="download"
      {% if download_location %}data-download-location="{{ download_location }}"{% endif %}>
--- a/bedrock/firefox/templatetags/helpers.py
+++ b/bedrock/firefox/templatetags/helpers.py
@@ -175,6 +175,59 @@ def download_firefox(ctx, channel='release', platform='all',
 
 @library.global_function
 @jinja2.contextfunction
+def download_firefox_thanks(ctx, dom_id=None, locale=None, alt_copy=None, button_class=None,
+                            locale_in_transition=False, download_location=None):
+    """ Output a simple "download firefox" button that only points to /download/thanks/
+
+    :param ctx: context from calling template.
+    :param dom_id: Use this string as the id attr on the element.
+    :param locale: The locale of the download. Default to locale of request.
+    :param alt_copy: Specifies alternate copy to use for download buttons.
+    :param button_class: Classes to add to the download button, contains size mzp-t-xl by default
+    :param locale_in_transition: Include the page locale in transitional download link.
+    :param download_location: Specify the location of download button for
+            GA reporting: 'primary cta', 'nav', 'sub nav', or 'other'.
+    """
+
+    channel = 'release'
+    locale = locale or get_locale(ctx['request'])
+    funnelcake_id = ctx.get('funnelcake_id', False)
+    dom_id = dom_id or 'download-button-thanks'
+    transition_url = '/firefox/download/thanks/'
+    version = firefox_desktop.latest_version(channel)
+
+    if funnelcake_id:
+        # include funnelcake in /download/thanks/ URL
+        transition_url += '?f=%s' % funnelcake_id
+
+    if locale_in_transition:
+        transition_url = '/%s%s' % (locale, transition_url)
+
+    download_link_direct = firefox_desktop.get_download_url(
+        channel, version, 'win', locale,
+        force_direct=True,
+        force_full_installer=False,
+        force_funnelcake=False,
+        funnelcake_id=funnelcake_id,
+    )
+
+    data = {
+        'id': dom_id,
+        'transition_url': transition_url,
+        'download_link_direct': download_link_direct,
+        'alt_copy': alt_copy,
+        'button_class': button_class,
+        'download_location': download_location,
+        'fluent_l10n': ctx['fluent_l10n']
+    }
+
+    html = render_to_string('firefox/includes/download-button-thanks.html', data,
+                            request=ctx['request'])
+    return jinja2.Markup(html)
+
+
+@library.global_function
+@jinja2.contextfunction
 def download_firefox_desktop_list(ctx, channel='release', dom_id=None, locale=None,
                                   force_full_installer=False):
     """
--- a/bedrock/firefox/templatetags/helpers.py
+++ b/bedrock/firefox/templatetags/helpers.py
@@ -196,8 +196,9 @@ def download_firefox_thanks(ctx, dom_id=None, locale=None, alt_copy=None, button
     transition_url = '/firefox/download/thanks/'
     version = firefox_desktop.latest_version(channel)
 
+    # if there's a funnelcake param in the page URL e.g. ?f=123
     if funnelcake_id:
-        # include funnelcake in /download/thanks/ URL
+        # include param in transitional URL e.g. /firefox/download/thanks/?f=123
         transition_url += '?f=%s' % funnelcake_id
 
     if locale_in_transition:
--- a/bedrock/firefox/tests/test_helpers.py
+++ b/bedrock/firefox/tests/test_helpers.py
@@ -310,6 +310,57 @@ class TestDownloadButtons(TestCase):
         assert pq(list[0]).attr('class') == 'os_ios'
 
 
+class TestDownloadThanksButton(TestCase):
+
+    def get_l10n(self, locale):
+        return fluent_l10n([locale, 'en'], settings.FLUENT_DEFAULT_FILES)
+
+    def test_download_firefox_thanks_button(self):
+        """
+        Download link should point to /firefox/download/thanks/
+        """
+        rf = RequestFactory()
+        get_request = rf.get('/fake')
+        get_request.locale = 'en-US'
+        doc = pq(render("{{ download_firefox_thanks() }}",
+                        {'request': get_request, 'fluent_l10n': self.get_l10n(get_request.locale)}))
+
+        links = doc('.c-button-download-thanks > a')
+        assert links.length == 1
+
+        link = pq(links)
+        href = link.attr('href')
+
+        assert href == ('/firefox/download/thanks/')
+        assert link.attr('id') == 'download-button-thanks'
+        assert link.attr('data-link-type') == 'download'
+
+        # Direct attribute for legacy IE browsers should always be win 32bit
+        assert link.attr('data-direct-link') == 'https://download.mozilla.org/?product=firefox-stub&os=win&lang=en-US'
+
+    def test_download_firefox_thanks_attributes(self):
+        """
+        Download link should support custom attributes
+        """
+        rf = RequestFactory()
+        get_request = rf.get('/fake')
+        get_request.locale = 'en-US'
+        doc = pq(render("{{ download_firefox_thanks(dom_id='test-download', button_class='test-css-class', "
+                        "download_location='primary cta', locale_in_transition=True) }}",
+                        {'request': get_request, 'fluent_l10n': self.get_l10n(get_request.locale)}))
+
+        links = doc('.c-button-download-thanks > a')
+        assert links.length == 1
+
+        link = pq(links)
+        href = link.attr('href')
+
+        assert href == ('/en-US/firefox/download/thanks/')
+        assert link.attr('id') == 'test-download'
+        assert link.attr('data-download-location') == 'primary cta'
+        assert 'test-css-class' in link.attr('class')
+
+
 class TestDownloadList(TestCase):
 
     def latest_version(self):
--- a/bedrock/firefox/tests/test_helpers.py
+++ b/bedrock/firefox/tests/test_helpers.py
@@ -332,7 +332,7 @@ class TestDownloadThanksButton(TestCase):
         link = pq(links)
         href = link.attr('href')
 
-        assert href == ('/firefox/download/thanks/')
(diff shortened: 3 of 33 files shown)


alexgibson:
This needs to be reviewed and merged after https://github.com/mozilla/bedrock/pull/9274

alexgibson:
This is ready for code review, but do-not-merge pending GA testing / verification.

bedrock/firefox/templatetags/helpers.py
```diff
@@ -173,6 +173,59 @@ def download_firefox(ctx, channel='release', platform='all',
...
+    :param alt_copy: Specifies alternate copy to use for download buttons.
+    :param button_class: Classes to add to the download button, contains size mzp-t-xl by default
+    :param locale_in_transition: Include the page locale in transitional download link.
+    :param download_location: Specify the location of download button for
+            GA reporting: 'primary cta', 'nav', 'sub nav', or 'other'.
+    """
+
+    channel = 'release'
+    locale = locale or get_locale(ctx['request'])
+    funnelcake_id = ctx.get('funnelcake_id', False)
```
stephaniehobson, line 194: How does this work? If a `funnelcake_id` is defined somewhere on the page which calls the button this picks that up?

bedrock/firefox/templatetags/helpers.py
```diff
@@ -173,6 +173,59 @@ def download_firefox(ctx, channel='release', platform='all',
...
+        transition_url += '?f=%s' % funnelcake_id
+
+    if locale_in_transition:
+        transition_url = '/%s%s' % (locale, transition_url)
+
+    download_link_direct = firefox_desktop.get_download_url(
+        channel, version, 'win', locale,
+        force_direct=True,
+        force_full_installer=False,
+        force_funnelcake=False,
```
stephaniehobson, line 210: The original `download_firefox` has some documentation about what these mean. If this function will eventually replace that one it would be nice to preserve the documentation.

stephaniehobson (commented):
I haven't started looking at the JS yet but here are a few comments from my initial pass.

media/js/base/mozilla-utils.js
```diff
@@ -21,6 +21,63 @@ if (typeof window.Mozilla === 'undefined') {
...
+    Utils.trackDownloadThanksButton = function() {
+        var downloadButton = document.querySelectorAll('.c-button-download-thanks > a');
+        var data = Utils.getDownloadAttributionValues(window.site);
+
+        for (var i = 0; i < downloadButton.length; ++i) {
+
+            if (data && data.os && data.name && data.version) {
+                downloadButton[i].setAttribute('data-display-os', data.os);
+                downloadButton[i].setAttribute('data-display-name', data.name);
+                downloadButton[i].setAttribute('data-download-version', data.version);
```
stephaniehobson, line 76: There isn't a 1:1 mapping for the version defined by the old button to the version defined in the JS here. (For example win64, win-msi, and standard will be reported as something else) Is that okay?

stephaniehobson:
I can't wait to see the impact implementing this in the nav bar has on performance! 🔥

bedrock/firefox/templatetags/helpers.py
alexgibson: If there's a funnelcake param in the url (e.g. ?f=123) then that will get passed to the /thanks page (e.g. /download/thanks/?f=123). I'll add a comment to document what it does.

bedrock/firefox/templatetags/helpers.py
alexgibson: I don't think this will replace the full download button entirely as it's needed to download pre-release products still, but good call to add some more docs 👍

media/js/base/mozilla-utils.js
alexgibson: Win64 downloads have never been enabled via the dl button (since the stub installer handles detection), and I don't think we'd ever serve msi installers via /thanks. So I think this is OK.

bedrock/firefox/templates/firefox/includes/download-button-thanks.html
```diff
@@ -0,0 +1,19 @@
+<div class="mzp-c-button-download-container c-button-download-thanks">
```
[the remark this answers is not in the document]
alexgibson: You can already add a custom button class and ID to the link itself. I don't think it makes sense to have another set for the wrapper, so if it's more useful I'll move them both to the top HTML element?

bedrock/firefox/templates/firefox/includes/download-button-thanks.html
```diff
@@ -0,0 +1,19 @@
...
+  <a href="{{ transition_url }}"
+     id="{{ id }}"
+     class="mzp-c-button mzp-t-product {% if button_class %}{{ button_class }}{% else %}mzp-t-xl{% endif %}"
```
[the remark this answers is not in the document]
alexgibson: It's not strictly needed, but I think having a singular identifiable class for the link (rather than needing `.c-button-download-thanks > a` makes sense. For compatibility purposes `download-link` wfm 👍

bedrock/firefox/templates/firefox/includes/download-button-thanks.html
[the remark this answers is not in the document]
alexgibson: Ok, we need to have a custom button class on the link so we can do things like `mzp-t-secondary`. It might be potentially confusing to have a different param and class for the wrapper, so what I propose is to make the ID belong to the top level element. Does that work?

bedrock/firefox/templates/firefox/includes/download-button-thanks.html
```diff
@@ -0,0 +1,19 @@
...
+     data-direct-link="{{ download_link_direct }}"
+     data-link-type="download"
+     {% if download_location %}data-download-location="{{ download_location }}"{% endif %}>
+    {% if alt_copy %}
+      {{ alt_copy }}
+    {% else %}
+      {{ ftl('download-button-download-firefox') }}
+    {% endif %}
+  </a>
+  <small class="mzp-c-button-download-privacy-link">
```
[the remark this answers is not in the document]
alexgibson: It looks like the new button was used on 2/5 of those .scss files. I've updated the ones that were effected. nice catch 👍

alexgibson:
Thanks @stephaniehobson - I pushed a commit that addresses some feedback. I also added some docs for the download button to try and help people understand which one to use and when.

alexgibson:
@pmac r on the python bits here? I'm mostly wondering if I've overlooked anything that should still be included in the newer helper.

bedrock/firefox/templatetags/helpers.py
```diff
@@ -173,6 +173,60 @@ def download_firefox(ctx, channel='release', platform='all',
...
+    :param download_location: Specify the location of download button for
+            GA reporting: 'primary cta', 'nav', 'sub nav', or 'other'.
+    """
+
+    channel = 'release'
+    locale = locale or get_locale(ctx['request'])
+    funnelcake_id = ctx.get('funnelcake_id', False)
+    dom_id = dom_id or 'download-button-thanks'
+    transition_url = '/firefox/download/thanks/'
+    version = firefox_desktop.latest_version(channel)
```
pmac, line 197: This is not something we should fix in this PR, but the `version` parameter to the `get_download_url` isn't even used in the function anymore. It's a vestige of the old way of generating direct download links before we only used the `*-latest` bouncer aliases. This doesn't hurt anything, but you could also just pass in `None` for `version` and it would work the same.

bedrock/firefox/tests/test_helpers.py
```diff
@@ -310,6 +310,59 @@ def test_ios(self):
...
+                        {'request': get_request, 'fluent_l10n': self.get_l10n(get_request.locale)}))
+
+        button = doc('.c-button-download-thanks')
+        links = doc('.c-button-download-thanks > .download-link')
+        assert links.length == 1
+
+        link = pq(links)
+        href = link.attr('href')
+
+        assert href == ('/en-US/firefox/download/thanks/')
```
pmac, line 360: style nit: superfluous parens

bedrock/firefox/tests/test_helpers.py
```diff
@@ -310,6 +310,59 @@ def test_ios(self):
...
+                        {'request': get_request, 'fluent_l10n': self.get_l10n(get_request.locale)}))
+
+        button = doc('.c-button-download-thanks')
+        links = doc('.c-button-download-thanks > .download-link')
+        assert links.length == 1
+
+        link = pq(links)
+        href = link.attr('href')
+
+        assert href == ('/firefox/download/thanks/')
```
pmac, line 335: style nit: superfluous parens

pmac (approved):
Looks good! Just a couple of small notes, but 👍 from me.

bedrock/firefox/templates/firefox/includes/download-button-thanks.html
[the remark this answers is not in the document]
stephaniehobson: ID on top-level +1.

I think that's best until we can have parent selectors in CSS.

bedrock/firefox/templatetags/helpers.py
stephaniehobson: So we'd still have to use the other button for a campaign page that has a funnel cake? (That's okay, I just want to understand).

alexgibson: No, funnel cakes should work for both button helpers 🙂

media/js/base/mozilla-utils.js
stephaniehobson: There's not a lot of them but win64, win64-msi are currently being [reported in GA](https://analytics.google.com/analytics/web/#/report/content-event-events/a36116321w64087799p65789850/explorer-table.secSegmentId=analytics.customDimension27&explorer-table.plotKeys=%5B%5D&explorer-table.rowCount=250&explorer-table.filter=win). I don't see standard though, even though it is documented in the read the docs.

alexgibson: There’s a modal on /windows that enables downloading specific builds. On /browsers/windows-64-bit/ my guess is they are following the link to the /all page.

I don’t think any of these events are coming from download buttons since those links have never been enabled to appear.

stephaniehobson: If you[ filter to /firefox/new](https://analytics.google.com/analytics/web/#/report/content-event-events/a36116321w64087799p65789850/explorer-table.plotKeys=%5B%5D&explorer-table.secSegmentId=analytics.customDimension27&_r.drilldown=analytics.eventCategory:~2Ffirefox~2Fnew~2F%20Interactions,analytics.eventAction:Firefox%20Download/) you can see they are being triggered on /new 🤷‍♀️

stephaniehobson: Is that the modal there too?

stephaniehobson: Hm, also seeing them on [/firefox](https://analytics.google.com/analytics/web/#/report/content-event-events/a36116321w64087799p65789850/_r.drilldown=analytics.eventCategory:~2Ffirefox~2F%20Interactions&explorer-table.secSegmentId=analytics.customDimension27&explorer-table.plotKeys=%5B%5D/). No modal there.

alexgibson: The older version of /new has the modal yes. Are these session based events? That may explain things. Still think it this is likely a red herring.

bedrock/firefox/templates/firefox/includes/download-button-thanks.html
```diff
@@ -0,0 +1,18 @@
+<div id="{{ id }}" class="mzp-c-button-download-container c-button-download-thanks">
+  <a href="{{ transition_url }}"
+     class="download-link mzp-c-button mzp-t-product {% if button_class %}{{ button_class }}{% else %}mzp-t-xl{% endif %}"
+     data-direct-link="{{ download_link_direct }}"
+     data-link-type="download"
```
stephaniehobson, line 5: There's a change in the GA reporting. This is showing up with the Event Label "Download for unavailable". I think it is because `data-download-os` is defined on the old button but not this one.

stephaniehobson: Otherwise GA reporting looks to be the same in Chrome/MacOS:

| Prod     | Demo |
| --- | --- |
| Screen Shot 2020-09-15 at 07 22 12 | Screen Shot 2020-09-15 at 07 24 19 |

alexgibson: Strange, because I'm adding `data-download-os` via JS (since the value is dynamic)

alexgibson: I can see `data-download-os` on the link when viewing source on demo:

image

alexgibson: Oh wait - it's `data-display-os` (face palm). At least that's a sinple fix :)

docs/download-buttons.rst
```diff
@@ -0,0 +1,41 @@
+.. This Source Code Form is subject to the terms of the Mozilla Public
+.. License, v. 2.0. If a copy of the MPL was not distributed with this
+.. file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+.. _download-buttons:
+
+========================
+Firefox Download Buttons
```
stephaniehobson, line 8: 📄 💯

bedrock/firefox/templates/firefox/includes/download-button-thanks.html
stephaniehobson: Oh, I was viewing source. Then I don't know what's causing the difference. Are you able to look into it?

alexgibson: Yeah I think I've got it - typo (see comment above ^) :)

stephaniehobson: LOL, I made the same mistake in my comment 🤦‍♀️

alexgibson: Updated data attribute to the correct name

alexgibson:
Leaving do-not-merge in place until adswerve have taken a look in GA

alexgibson:
Ok, adswerve have done some QA and confirmed things look good in GA. I'm going to remove the DNM label from the PR, but we should probably wait until Monday to merge and push to prod (pending final review approval).

stephaniehobson (approved):
R+ 🦋

Issue #9334, opened 2020-08-26
September Campaign Landing Page


stephaniehobson:
Landing page which matches September Campaign look and feel.

In English and German only. No need for localization.

PR #9358, opened 2020-09-01
Add september campaign page and placeholders

stephaniehobson:
## Description

- page available on dev only right now, using placeholder GIFs
- page, CSS, JS, and url for landing page
- layout for hero image, checklist, and further reading

## Issue / Bugzilla link

See #9334 

## Testing

http://localhost:8000/en-US/firefox/campaign/unfck/


Zenix27 (approved):

bedrock/firefox/urls.py
```diff
@@ -170,3 +172,12 @@
     # Issue 8536
     page('firefox/retention/thank-you', 'firefox/retention/thank-you.html'),
 )
+
+if settings.DEV:
```
alexgibson, line 176: 🔥

media/css/firefox/unfck.scss
```diff
@@ -0,0 +1,524 @@
...
+        padding-bottom: $v-grid-lg;
+    }
+
+    ul.download-list,
+    .mzp-c-button-download-container {
+        margin-bottom: 0;
+    }
+}
+
+@supports ( width: 75vw ) {
```
alexgibson, line 158: This is a nice progressive enhancement 👍

alexgibson (approved):
This looks like a great base starting point, thanks! 

It'll be an interesting experiment to see how this goes. I'll likely start working on the EU content on Monday, but I'll try and leave the base HTML/CSS untouched. If there are any specific bits of functionality you could use some help with (such as the form for example), then I'd be happy to help out there as well.

Issue #9344, opened 2020-08-27
WNP81: Merge code by EOD 9/2


hoosteeno:
## Description

Code the WNP81 in our standard WNP template with copy from #9343 as part of epic #9342.

---

#### Success Criteria
- [ ] Coded mockups are approved by the council
- [ ] Code is merged and launched by EOD 9/2 so l10n can commence

PR #9363, opened 2020-09-02
String fixes for WNP81 [#9344]

craigcook:
## Description
More neutral English for easier l10n, and we'll treat the quirky colloquialisms as an American translation.

Also fixes a failing test.

### DIFF
--- a/l10n/en-US/firefox/whatsnew/whatsnew-fx81.ftl
+++ b/l10n/en-US/firefox/whatsnew/whatsnew-fx81.ftl
@@ -2,5 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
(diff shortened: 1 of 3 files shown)


l10n/en-US/firefox/whatsnew/whatsnew-fx81.ftl
```diff
@@ -2,5 +2,6 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+whatsnew81-our-latest-for-android-has = Our latest for { -brand-name-android } has lots of fun, smart and even weirdo features – like a moveable search bar. Because we’re an independent browser. We care about your thumbs.
```
stephaniehobson, line 5: The weirdo comment was lost in the copy-paste.

craigcook: I don't think we need it for en-US, that was for localizers.

amychurchwell (approved):
r+ 🐱

PR #9372, opened 2020-09-07
Add form functionality to sept landing page (#9334)

alexgibson:
## Description
- Adds form view & tests for the september landing page (heavily borrowed from the /press/ section).
- Moves logic for making page appear only on Dev from `urls.py` to a switch in `views.py` (needed for the unit tests).

## Issue / Bugzilla link
#9334

### DIFF
--- a/bedrock/firefox/forms.py
+++ b/bedrock/firefox/forms.py
@@ -6,6 +6,8 @@ import re
 
 from django import forms
 
+from bedrock.mozorg.forms import (HoneyPotWidget)
+
 
 class SendToDeviceWidgetForm(forms.Form):
     number = forms.CharField(max_length=20, min_length=4, required=False)
@@ -22,3 +24,21 @@ class SendToDeviceWidgetForm(forms.Form):
             number = re.sub(r'\D+', '', number)
 
         return number
+
+
+class UnfckForm(forms.Form):
+    unfck_field = forms.CharField(
+        max_length=500,
+        required=True,
+        widget=forms.Textarea(),
+    )
+
+    # honeypot
+    office_fax = forms.CharField(widget=HoneyPotWidget, required=False)
+
+    def clean_office_fax(self):
+        cleaned_data = super(UnfckForm, self).clean()
+        honeypot = cleaned_data.pop('office_fax', None)
+
+        if honeypot:
+            raise forms.ValidationError('Your submission could not be processed')
--- a/bedrock/firefox/forms.py
+++ b/bedrock/firefox/forms.py
@@ -6,7 +6,7 @@ import re
 
 from django import forms
 
-from bedrock.mozorg.forms import (HoneyPotWidget)
+from bedrock.mozorg.forms import HoneyPotWidget
 
 
 class SendToDeviceWidgetForm(forms.Form):
@@ -28,7 +28,7 @@ class SendToDeviceWidgetForm(forms.Form):
 
 class UnfckForm(forms.Form):
     unfck_field = forms.CharField(
-        max_length=500,
+        max_length=2000,
         required=True,
         widget=forms.Textarea(),
     )
--- a/bedrock/firefox/forms.py
+++ b/bedrock/firefox/forms.py
@@ -28,7 +28,7 @@ class SendToDeviceWidgetForm(forms.Form):
 
 class UnfckForm(forms.Form):
     unfck_field = forms.CharField(
-        max_length=2000,
+        max_length=280,
         required=True,
         widget=forms.Textarea(),
     )
--- a/bedrock/firefox/templates/firefox/campaign/unfck/emails/unfck.txt
+++ b/bedrock/firefox/templates/firefox/campaign/unfck/emails/unfck.txt
@@ -1,0 +1,5 @@
+Tell the world what you wanna unfck
+
+------------------------------------
+
+{{ unfck_field }}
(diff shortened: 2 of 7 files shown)


alexgibson:
@stephaniehobson @pmac r?

This doesn't add any fancy ajax form handling just yet, because I'm not 100% sure what the plan for this form is in terms of UX. Maybe that's something we could build on here if needed. For now, this basically just adds the logic to generate an email on submit.

bedrock/firefox/templates/firefox/campaign/unfck/emails/unfck.txt
```diff
@@ -0,0 +1,5 @@
+Tell the world what you wanna unfck
```
alexgibson, line 1: What should the copy be here?

bedrock/firefox/views.py
```diff
@@ -59,6 +67,10 @@
 ]
 STUB_VALUE_RE = re.compile(r'^[a-z0-9-.%():_]+$', flags=re.IGNORECASE)
 
+UNFCK_EMAIL_FROM = 'Mozilla.com <noreply@mozilla.com>'
+UNFCK_EMAIL_SUBJECT = 'Tell the world what you wanna unfck'
+UNFCK_EMAIL_TO = ['TODO@mozilla.com']
```
alexgibson, line 72: Who will be the poor soul to wade into all these emails?

alexgibson:
I'm not quite sure how to test this fully. When submitting the form locally (using my own email), everything works but no email is received. I'm assuming that's because whatever generates those isn't there?

When testing on demo, submitting the form [throws an error](https://sentry.prod.mozaws.net/operations/bedrock-demos/issues/9542999/?referrer=alert_email). The same error is also thrown for other forms such as `/press/press-inquiry/` however, and seems to happen on all demos.

pmac:
Locally it's set to use a dummy email setup that should output to console. In demos it's probably set to use a real smtp server but they don't have credentials.

bedrock/firefox/forms.py
```diff
@@ -6,6 +6,8 @@
 
 from django import forms
 
+from bedrock.mozorg.forms import (HoneyPotWidget)
```
pmac, line 9: superfluous parentheses

bedrock/firefox/forms.py
```diff
@@ -22,3 +24,21 @@ def clean_number(self):
             number = re.sub(r'\D+', '', number)
 
         return number
+
+
+class UnfckForm(forms.Form):
+    unfck_field = forms.CharField(
+        max_length=500,
```
pmac, line 31: Only 500 characters?

bedrock/firefox/templates/firefox/campaign/unfck/emails/unfck.txt
```diff
@@ -0,0 +1,5 @@
...
+
+------------------------------------
+
+{{ unfck_field }}
```
pmac, line 5: I think to be safe we should use the `bleach_tags` filter here so that all HTML is stripped.

```suggestion
{{ unfck_field|bleach_tags }}
```

That filter is defined here https://github.com/mozilla/bedrock/blob/bd043fa6515d8e8a3051d1673bb9e4487147b0f8/bedrock/mozorg/templatetags/misc.py#L556-L558

bedrock/firefox/tests/test_views.py
```diff
@@ -758,3 +761,127 @@ def test_firefox_welcome_page1(self, render_mock):
...
+        Form should be valid.
+        """
+        form = UnfckForm(self.data)
+
+        # make sure form is valid
+        assert form.is_valid()
+
+    def test_form_missing_data(self):
+        """
+        With incorrect data (missing email), form should not be valid
```
pmac, line 842: not "missing email" but blank text right?

pmac (changes requested):
It's looking good. Just a couple of suggestions and questions.

bedrock/firefox/forms.py
alexgibson: Hmm good suggestion, let's increase it to 2000 (does that sound good?)

alexgibson:
Thanks @pmac updated

bedrock/firefox/forms.py
stephaniehobson: I think it needs to be tweet length.

alexgibson: Perhaps we can update this later once the tweet functionality is in place? Right now there's no indication what this form actually does (sending an email or sharing on twitter). I think the UI needs to be made clearer, which can happen in a separate PR.

alexgibson: Ok, I've changed this to 280 characters as suggested. I think the UI will really need to make clear the limit to avoid frustration though.

pmac: Definitely. Probably Twitter-style character countdown.

pmac (approved):
:octocat:

bedrock/firefox/templates/firefox/campaign/unfck/emails/unfck.txt
stephaniehobson: This is good for now.

bedrock/firefox/views.py
stephaniehobson: Still TBD.

stephaniehobson (approved):
R+ 🐍 

Pmac had a few outstanding comments. I'll do a follow up commit for them.

stephaniehobson:
Oh, pmac approved. I'll merge :)

Issue #9370, opened 2020-09-07
Local install fails when installing Pillow


alexgibson:
## Description

I updated my local install of Yarn using `brew upgrade yarn`. However when homebrew updated, it looks like it also updated the version of Python installed on my machine. As a result, my local install of bedrock broke, and I can't install a fresh virtual env using the docs. It fails with an error.

### Steps to reproduce

Follow the local install steps here: https://bedrock.readthedocs.io/en/latest/install.html#local-installation

### Expected result

Bedrock should install successfully.

### Actual result

An error is thrown when installing Pillow:

```
   Running setup.py install for Pillow ... error
    ERROR: Command errored out with exit status 1:
     command: /Users/agibson/Documents/GitHub/bedrock/venv/bin/python3 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/private/var/folders/sy/tzj820zd5vb4nv6wfgxcl2rr0000gn/T/pip-install-yasp_tsp/pillow/setup.py'"'"'; __file__='"'"'/private/var/folders/sy/tzj820zd5vb4nv6wfgxcl2rr0000gn/T/pip-install-yasp_tsp/pillow/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /private/var/folders/sy/tzj820zd5vb4nv6wfgxcl2rr0000gn/T/pip-record-v18v5jll/install-record.txt --single-version-externally-managed --compile --install-headers /Users/agibson/Documents/GitHub/bedrock/venv/include/site/python3.8/Pillow
         cwd: /private/var/folders/sy/tzj820zd5vb4nv6wfgxcl2rr0000gn/T/pip-install-yasp_tsp/pillow/
    Complete output (174 lines):
    running install
    running build
    running build_py
    creating build
    creating build/lib.macosx-10.15-x86_64-3.8
    creating build/lib.macosx-10.15-x86_64-3.8/PIL
    copying src/PIL/MpoImagePlugin.py -> build/lib.macosx-10.15-x86_64-3.8/PIL
    copying src/PIL/ImageMode.py -> build/lib.macosx-10.15-x86_64-3.8/PIL
    copying src/PIL/PngImagePlugin.py -> build/lib.macosx-10.15-x86_64-3.8/PIL
[...]

PR #9382, opened 2020-09-10
Upgrade Pillow to 7.2.0

pmac:
We needed to upgrade Pillow anyway. I'm also hoping that it helps with #9370


pmac:
@alexgibson let me know if this helps with installing on Python 3.7 or 3.8.

alexgibson (approved):
I tried this with 3.7.8 and bedrock installs and runs fine.

I tried again with 3.8 and install works fine, but there are other unrelated errors when running the site. That's to be expected for now I guess, so this lgtm thanks! r+

PR #9383, opened 2020-09-10
Updated grammar in commit.html

vringar:
## Description
Changed the english version of `mozorg.about.governance.policies.commit.access-policy`, replacing a clunky "his or her" with a "their", to be both more inclusive as well as easier to read.

## Testing
There should be no testing required.


pmac (approved):
Thanks ❤️

PR #9387, opened 2020-09-10
Show standard WNP template to Indonesian [fix #9365]

### LINKED ISSUE #9365: Display the standard whatsnew page for Indonesian
## Description

Indonesian (`id`) gets a custom hard-coded whatsnew page promoting Firefox Lite, but as of https://github.com/mozilla/bedrock/pull/9323 the regular whatsnew pages are localizable in `id`. We'll need to update the view logic so `id` can see the regular localized page.

Question: do we want to preserve the custom page for older versions of Firefox and only start showing the regular page in Indonesian for 81+? Or do we want to decommission the old custom template entirely?

Also related: We show a custom page promoting Lite in India and several countries in Africa. Are we decommissioning that as well?

---

#### Success Criteria
- [ ] Tasks should be represented here as something like "Approved designs attached to this issue"

craigcook:
## Description
Removes the hard-coded custom template and the view logic so `id` can receive the standard whatsnew page.

I've left the assets and the English template for now because I haven't heard yet if we also need to stop showing it in India.

## Issue / Bugzilla link
#9365 

## Testing
http://localhost:8000/id/firefox/81.0/whatsnew/all/

WNP81 has been exposed for l10n in `id` but the last several versions were not, so the whatsnew pages for 79 and 80 won't have active `id` Fluent files and should fall back to the evergreen page (only in production mode).


alexgibson (approved):
🇮🇩

PR #9393, opened 2020-09-11
Add a test attribute to thanks page [fix #9389]

### LINKED ISSUE #9389: Add an attribute to the /download/thanks/ page to support FxA tests
## Description

The Accounts team has some functional tests for the "download and sign up" user flow that look for a known element on the /download/thanks/ page to verify the correct page has loaded. We removed an element ID they were previously relying on in https://github.com/mozilla/bedrock/pull/9274 so those tests started failing.

To prevent this from happening again next time we change the template, I think it's better to add something explicitly for testing purposes that we'll know not to change or remove (rather than building tests around page elements subject to change whenever we redesign). A data attribute on the root or body element should work, and we'll be sure to add a comment about what it is.

Once the template is updated the FxA team can update their tests to look for the new attribute.

---

#### Success Criteria
- [ ] Thanks page template has a data attribute for reliable testing
- [ ] FxA tests are updated to detect that new attribute

craigcook:
## Description
This adds the attribute `data-test-fxa-template="firefox-download-thanks"` to the /download/thanks/ template root element, which FxA tests can use to verify the page loads.

There's likely another way to verify a page in a test, but they've been using a CSS selector to look for an element already and this may be easier than completely rewriting their tests. Once this lands they can update their tests to look for the selector `html[data-test-fxa-template="firefox-download-thanks"]`

## Issue / Bugzilla link
#9389

### DIFF
--- a/bedrock/firefox/templates/firefox/new/protocol/base.html
+++ b/bedrock/firefox/templates/firefox/new/protocol/base.html
@@ -8,9 +8,6 @@
 
 {% extends "firefox/base/base-protocol.html" %}
 
-{# FxA automated tests use this attribute to verify the page #}
-{% block html_attrs %}data-test-fxa-template="firefox-download-thanks"{% endblock %}
-
 {% block page_title_prefix %}{{_('Download Firefox')}} — {% endblock %}
 
 {%- block page_title -%}{{_('Free Web Browser')}}{%- endblock -%}
--- a/bedrock/firefox/templates/firefox/new/protocol/base.html
+++ b/bedrock/firefox/templates/firefox/new/protocol/base.html
@@ -8,6 +8,9 @@
 
 {% extends "firefox/base/base-protocol.html" %}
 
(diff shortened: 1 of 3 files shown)


bedrock/firefox/templates/firefox/new/protocol/base.html
```diff
@@ -8,6 +8,9 @@
 
 {% extends "firefox/base/base-protocol.html" %}
 
+{# FxA automated tests use this attribute to verify the page #}
+{% block html_attrs %}data-test-fxa-template="firefox-download-thanks"{% endblock %}
```
alexgibson, line 12: I'm not sure we need this one as this base template is used now only for /firefox/windows/, /firefox/mac/ etc?

alexgibson (changes requested):
Looks good, just one question / suggestion. Thanks for helping out to fix this 👍

craigcook:
Updated!

alexgibson (approved):

PR #9397, opened 2020-09-14
Make stub attribution work for legacy IE browsers (Fixes #9350)

### LINKED ISSUE #9350: Stub attribution is not being applied in old IE browsers
## Description

The stub attribution script is running in older Ie browsers, however the signed attribution code is not being applied to the download URL, due to the way old IE browsers have a different triggering mechanism for the download:

https://github.com/mozilla/bedrock/blob/master/media/js/ie/mozilla-utils-ie.js#L27-L37

### Steps to reproduce

1.) On IE 9 or below, visit https://www.mozilla.org/firefox/new/
2.) Click download Firefox

### Expected result

Stub attribution params should be attached to the file download URL

### Actual result

No params are attached.

### Environment

Windows 7, IE9

alexgibson:
## Description
- Add stub attribution data to `data-direct-link` attributes for legacy IE browsers.
- Update tests.

## Issue / Bugzilla link
#9350

## Testing
Demo: https://www-demo3.allizom.org/en-US/

1. Using IE 8/9 on Windows, navigate to the demo URL above.
2. Open IE developer tools, try to inspect the DOM of the download button in the nav.
3. Find a download link and inspect the attributes.
4. Verify `attribution_code` and `attribution_sig` params have been appended to the URL (see screenshot below).
5. Click download, and verify you still get the expected FF installer.

image

### DIFF
--- a/media/js/base/stub-attribution.js
+++ b/media/js/base/stub-attribution.js
@@ -103,14 +103,23 @@ if (typeof window.Mozilla === 'undefined') {
 
         // target download buttons and other-platforms modal links.
         $('.download-list .download-link, .download-platform-list .download-link').each(function() {
+            var link = this;
             var version;
-            // If this is a transitional download link do nothing.
-            if (this.href && this.href.indexOf('/firefox/download/thanks/') === -1) {
+            var directLink;
+            // Append stub attribution data to direct download links.
+            if (link.href && link.href.indexOf('https://download.mozilla.org') !== -1) {
 
-                version = $(this).data('downloadVersion');
+                version = link.getAttribute('data-download-version');
                 // Append attribution params to Windows 32bit, 64bit, and MSI installer links.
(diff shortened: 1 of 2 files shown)


amychurchwell (approved):
lgtm! r+wc:

I noticed during testing no-js download links don't currently have `data-direct-link` and I am curious if this is by design? Should we implement this for the dumb-download links? did not want to block on this question.

alexgibson:
> I noticed during testing no-js download links don't currently have data-direct-link and I am curious if this is by design? Should we implement this for the dumb-download links? did not want to block on this question.

I'm not sure there's any benefit in adding the attribute to the buttons that get shown with no-js, as without JS the functionality is moot / not needed. Thanks!

amychurchwell:
"why don't we add this js code to the non-js buttons?" lol, yeesh. Thanks Alex!

PR #9402, opened 2020-09-14
Update comparison table on /new (Fixes #9359)

### LINKED ISSUE #9359: Comparison table on /new includes inaccurate info
## Description

The comparison table on our download page says Chrome blocks 3rd party cookies:

Screen Shot 2020-09-01 at 3 33 35 PM

This is only true under certain circumstances -- i.e. in incognito mode. We should be clearer about this comparison. We should add the words "by default" to the end of the comparison, and change the checkmark under Chrome to the "false" symbol:

Screen Shot 2020-09-01 at 3 35 58 PM

amychurchwell:
## Description
image

Adds "by default" to "Blocks third-party tracking cookies" table row and removes Chrome's check.

## Issue / Bugzilla link
#9359

### DIFF
--- a/l10n/en/firefox/new/desktop.ftl
+++ b/l10n/en/firefox/new/desktop.ftl
@@ -30,6 +30,7 @@ firefox-desktop-download-compare-safari = { -brand-name-safari }
 firefox-desktop-download-yes = Yes
 firefox-desktop-download-no = No
 firefox-desktop-download-blocks-third-party = Blocks third-party tracking cookies
+firefox-desktop-download-blocks-third-party-default = Blocks third-party tracking cookies by default
 firefox-desktop-download-autoplay-blocking = Autoplay blocking
 firefox-desktop-download-blocks-social-trackers = Blocks social trackers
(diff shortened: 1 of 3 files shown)


l10n/en/firefox/new/desktop.ftl
```diff
@@ -29,7 +29,7 @@ firefox-desktop-download-compare-edge = { -brand-name-edge }
 firefox-desktop-download-compare-safari = { -brand-name-safari }
 firefox-desktop-download-yes = Yes
 firefox-desktop-download-no = No
-firefox-desktop-download-blocks-third-party = Blocks third-party tracking cookies
+firefox-desktop-download-blocks-third-party = Blocks third-party tracking cookies by default
```
craigcook, line 32: I think this needs to be a new string with a new ID. That also means we can set the old one as the fallback.

craigcook (changes requested):
We should make this same update to the comparison pages but that can be a different PR.

craigcook (approved):

PR #9416, opened 2020-09-17
Remove Firefox Send 

craigcook:
## Description
We hid prominent links to Send behind a switch in #9326 but [now we've announced the permanent closure](https://blog.mozilla.org/blog/2020/09/17/update-on-firefox-send-and-firefox-notes/) so we can make that removal permanent. This deletes all those places where it was switch-hidden and also removes mentions of Send in a few other spots, along with related structured data.

## Issue / Bugzilla link
#9316 

## Testing
Make sure I didn't miss anything.

Send is still mentioned/linked in the fallback string on the Safari comparison page, but that will fall off over time as l01n updates.


l10n/en/firefox/browsers/compare/safari.ftl
```diff
@@ -42,14 +42,12 @@ compare-safari-like-safari-firefox-encourages = Like { -brand-name-safari }, { -
...
-compare-safari-also-when-you-sign-up-for-fallback = Also, when you sign up for a { -brand-name-firefox } account you get access to some unique services like Screenshots, <a { $pocket }>{ -brand-name-pocket }</a> and <a { $send }>{ -brand-name-send }</a> that integrate directly into the browser. Screenshots is a feature built right into the { -brand-name-firefox } browser, allowing you to copy or download any or all part of a web page. When you save the screenshot, you can also choose which folder you want to find it in, instead of cluttering your desktop. The { -brand-name-pocket } for { -brand-name-firefox } button lets you save web pages and videos to { -brand-name-pocket } in just one click, so you can read a clean, distraction-free version whenever and wherever you want — even offline. With { -brand-name-send }, you can share large files with end-to-end encryption and a variety of security controls, such as the ability to set an expiration time for a file link to expire, the number of downloads, and whether to add an optional password for an extra layer of security.
```
alexgibson, line 52: There's a string being removed here from the canonical ftl file - is that expected? Typically we add a comment to mark them as outdated. I'm not entirely use what would happen if we delete the string altogether - maybe something to ask @peiying2?

bedrock/firefox/templates/firefox/products/index.html
```diff
@@ -157,8 +149,6 @@ <h3 class="mzp-c-menu-list-title">{{ ftl('firefox-products-get-pocket') }}</h3>
       {% include 'includes/structured-data/software/firefox-monitor-software.json' %}
       ,
       {% include 'includes/structured-data/software/firefox-lockwise-software.json' %}
-      ,
-      {% include 'includes/structured-data/software/firefox-send-software.json' %}
```
alexgibson, line 161: I see a couple more templates using this file still: https://github.com/mozilla/bedrock/search?q=%7B%25+include+%27includes%2Fstructured-data%2Fsoftware%2Ffirefox-send-software.json%27+%25%7D&unscoped_q=%7B%25+include+%27includes%2Fstructured-data%2Fsoftware%2Ffirefox-send-software.json%27+%25%7D

alexgibson (changes requested):
A couple of pages are throwing an error becuase they still reference the json file, but otherwise this LGTM. I'm not quite sure how best to approach the .ftl string change, so maybe something we should ask Peiying about.

l10n/en/firefox/browsers/compare/safari.ftl
craigcook: So this string has now been updated twice, but I don't think we can do multiple fallbacks. So if we want to preserve the old string I think it'll have to be `if ftl_has_messages()` in the template.

alexgibson: I don't think we need to keep three levels of fallback in the template, I was just meaning I wasn't sure what happens if we delete a string from the english source file.

craigcook: Oh nevermind, I see I did that already, I just messed this up.

alexgibson (approved):

pmac:
In case you're not following the fluent repo, Peiying has requested changes https://github.com/mozilla-l10n/www-l10n/pull/108#pullrequestreview-491579379

PR #9421, opened 2020-09-18
Fix functional tests on 76 WNP (#9316)

alexgibson:
Fixes: https://gitlab.com/mozmeao/www-config/-/pipelines/191738102/failures


craigcook (approved):
Note the 76 page is being removed in https://github.com/mozilla/bedrock/pull/9404 but this will fix those tests until that can be reviewed and merged.

PR #9429, opened 2020-09-22
Add image to bedrock for releasenotes embed

alexgibson:
## Description
Adds an image to /media/ for releng to embed in a release notes publication.

## Issue / Bugzilla link
N/A


alexgibson:
@pmac releng pinged me to ask if they could include an image in a release notes publication, which I've included here. I wasn't aware of any solution they have for this in Nucleus, so I figured we could push an image to prod for them to reference.

Does this sound ok, or is there a more appropriate way to do this?

pmac:
This looks good. We don't have anything better yet unfortunately.

pmac (approved):
🖼

PR #9430, opened 2020-09-22
Exclude GA on /privacy/thunderbird/ (Fixes #9426)

### LINKED ISSUE #9426: Exclude GA from /privacy/thunderbird/
## Description

https://www.mozilla.org/en-US/privacy/thunderbird/

It looks like this page is being opened in-product by Thunderbird, which creates a lot of noise in GA, and also decreases overall conversion rate for the site as a whole. We already exclude GA from /privacy/firefox/, so perhaps we should consider the same here as well?

---

#### Success Criteria
- [ ] GA is excluded from /privacy/thunderbird/

alexgibson:
## Description
Removes Google Analytics from Thunderbird privacy notice.

http://localhost:8000/en-US/privacy/thunderbird/

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/9426

## Testing
- [ ] GA JS bundle should not be present on the page


pmac (approved):
🔒

PR #9432, opened 2020-09-22
Update firefox/all Android tests

alexgibson:
Fixes: https://gitlab.com/mozmeao/www-config/-/jobs/751098290

Looks like this info changed in product details

### DIFF
--- a/bedrock/firefox/firefox_details.py
+++ b/bedrock/firefox/firefox_details.py
@@ -375,8 +375,7 @@ class FirefoxAndroid(_ProductDetails):
         platforms = OrderedDict()
 
         # Supported Android version has been changed with Firefox 56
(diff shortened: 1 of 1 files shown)


pmac:
I think we should fix the details instead actually. It looks like the issue is that the version has been removed from `mobile_details` and we're basing the string on the version:

https://github.com/mozilla/bedrock/blob/5e35c6434b85dc0c2767c30608f557d51b738272/bedrock/firefox/firefox_details.py#L379

So I think we should change that to just always say `4.1` since we're definitely greater 56 but it's not reported.

alexgibson:
Huh, ok nice find!

alexgibson:
@pmac updated

pmac (approved):
✅

PR #9440, opened 2020-09-24
Updates to EN September Campaign page

stephaniehobson:
## Description

* add/update alt text
* standardize styles for CTAs
* change banner to `img` with srcset to move it up in load order

## Issue / Bugzilla link

#9334 

## Testing

http://localhost:8000/en-US/firefox/unfck/


stephaniehobson:
@alexgibson  I'm conflicted on the alt text.

Normally I'd call these decorative and not add alt text but I think that people who have trouble reading the moving text might try to rely on it to help them understand what's going on.

I wanted to spell fuck right so the screen readers can handle it but then I was worried about tripping the censors that you mentioned. I'm not actually sure what a screen reader will do with `fck` (I'll try it out in the AM). Also, the alt text shows up breifly while the lazy loaded images lazy load. Not sure if that's a problem (I suppose we could centre it?)

bedrock/firefox/templates/firefox/campaign/unfck/index.html
```diff
@@ -28,7 +28,7 @@ <h3 class="c-item-title">{{ title }}</h3>
           {% if src_still %}
           <source srcset="{{ src_still }}" media="(prefers-reduced-motion: reduce)" loading="lazy" width="280" height="280"></source>
           {% endif %}
-          <img src="{{ src }}" loading="lazy" width="280" height="280">
+          <img src="{{ src }}" loading="lazy" width="280" height="280" alt="{{ alt }}">
```
alexgibson, line 31: I'm not sure on the right answer here, but is `alt` text on the `img` fallback enough to satisfy screen readers that encounter the `<picture>` element?

alexgibson (approved):
I think in the context of this page adding alt text does make sense. It's a good call, and I'll have to see if i can get some German/French copy here also (I may not be able to do so for everything but I'll try).

Maybe the visible flash of alt text is ok (?), but if we really wanted to hide it we could always add `color: transparent;` to the `img`? It's not ideal but at least the text would be there for screen readers still.

alexgibson:
feel free to merge as / when you're ready. I'll leave it to you

alexgibson:
Oh, also for the spelling of f*ck - at this stage I think we've at least voiced it as a possible concern. That's about as much as we can do realistically given the campaign content. If screen readers manage ok with fck then that's let's use that, but if not then I'm ok with the proper spelling.

bedrock/firefox/templates/firefox/campaign/unfck/index.html
stephaniehobson: My research and testing suggest this is the best way. Chrome didn't idenify it as an image but it did read the alt text. It ended up sounding like a subheading because the heading is first in the source order.

stephaniehobson:
The screen reader on MacOS did well with unfck so I've swapped the correct spelling for that.

PR #9441, opened 2020-09-24
Add sticky promo component to /firefox pages

### LINKED ISSUE #9406: Add sticky promo to /firefox
## Description

We should implement a winning experiment design on /firefox using the new sticky promo component.

"Based on mozilla.org traffic volume between 4/30/20 and 5/27/20, we estimate that implementing the winning shorter copy and logo treatment into production on /firefox will lead to **~1.25M incremental non-FX desktop downloads per year.**"

image

maxxcrawford:
## Description

Approvals: 
- [x] Design approval
- [x] Code approval

To-do: 
- [x] Add Fx-User Product Links with UTMs/attribution
- [x] Write test to confirm the promo shows on load and then hides on dismiss click 
- [x] Confirm layout matches design specs 
- [x] Confirm mobile behavior
- [x] Browser test 
- [x] Add Download Link `download_firefox_thanks`
- [x] Add Fx-User Product Logos
- [x] Promo is limited to the following three pages: `firefox`, `firefox/new` and `firefox/browsers`
- [x] Remove promo bundle file and add as partial for import
- [x] Modal should remember you closed it for the session 

## Issue / Bugzilla link

#9406

## Testing

⚠️ _Be sure to load each link in both a Firefox browser and a non-Firefox browser, as they show different messages!_ 

Testing URLs: 
- https://www-demo-pr-9441.herokuapp.com/en-US/firefox/
- https://www-demo-pr-9441.herokuapp.com/en-US/firefox/new/ 
- https://www-demo-pr-9441.herokuapp.com/en-US/firefox/browsers/

### Screenshots

Firefox User: 

image

Non-Firefox User: 

image


maxxcrawford (commented):
Self-review/notes for discussion with the team

bedrock/base/templates/includes/sticky-promo.html
```diff
@@ -0,0 +1,18 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% block page_css %}
+  {{ css_bundle('firefox_sticky_promo') }}
+{% endblock %}
```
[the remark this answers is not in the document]
craigcook: I think this would go on each page that includes the promo, I don't think it can go in the promo include itself.
It also might be an option to just make it a Sass partial and `@import` it on the pages that need it, rather than making a whole separate bundle.

l10n/en/firefox/sticky-promo.ftl
```diff
@@ -0,0 +1,10 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### URL: https://www-dev.allizom.org/firefox/*
+### Global to all /firefox pages
+
+# Meet our family of products
+firefox-sticky-promo-meet-our-family-of = Meet our family of products
```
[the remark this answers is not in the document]
craigcook: hmm... well if that string is in a global file you can reference the string ID in the template, rather than doing it like this (which I expect might cause other problems, if it works at all). And if it's not in a global file you'd need to reference that file from every page that wants to use it, which can make a mess with file activation.

So even though this string exists elsewhere I think it's probably still best to treat it as new.

media/static-bundles.json
```diff
@@ -740,6 +740,12 @@
         "css/firefox/unfck.de.scss"
       ],
       "name": "firefox-unfck-de"
+    },
```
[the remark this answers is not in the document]
craigcook: Not that I know of. Though for this one I don't know if this should get its own bundle. A better approach may be to either make it a Sass partial to import into pages that need it, OR add it as another file in the individual CSS bundle for each page that uses it. The JS for the promo can also be added to page bundles.

On a third hand, if we want to control the promo with a waffle switch it might need a dedicated bundle after all.

bedrock/base/templates/includes/sticky-promo.html
```diff
@@ -0,0 +1,38 @@
...
+  <button class="mzp-c-sticky-promo-close" type="button">{{ ftl('ui-close') }}</button>
+  <div class="promo-firefox">
+    <h3 class="mzp-c-sticky-promo-title">{{ ftl('firefox-sticky-promo-get-the-latest-firefox') }}</h3>
+    {{ download_firefox_thanks(dom_id='sticky-promo', download_location='other', button_class='mzp-c-button mzp-t-product mzp-t-small') }}
+  </div>
+  <div class="promo-products">
+    <h3 class="mzp-c-sticky-promo-title">{{ ftl('firefox-sticky-promo-meet-our-family-of') }}</h3>
+    <ul class="promo-products-list">
+      <li>
+        <a data-link-name="Browsers" data-link-type="link" class="promo-products-link" href="{{ url('firefox.browsers.index') }}">
```
maxxcrawford, line 15: For internal links – are there more tags needed? I saw other `data-link` attrs like: `data-link-position` and `data-link-group`.

bedrock/base/templates/includes/sticky-promo.html
```diff
@@ -0,0 +1,38 @@
...
+          <img src="{{ static('protocol/img/logos/firefox/browser/logo.svg') }}" alt="">
+          Browsers
+        </a>
+        <a data-link-name="Monitor" data-link-type="link" class="promo-products-link" href="https://monitor.firefox.com/{{ referral|safe }}" rel="external noopener">
```
maxxcrawford, line 19: Same question for external links: Are there more tags needed? Do we need to label all of these links with a `sticky-promo` sort of GTM ID?

media/css/firefox/browsers-products.scss
```diff
@@ -11,6 +11,7 @@ $image-path: '/media/protocol/img';
 @import '../../protocol/css/components/zap';
 @import '../protocol/components/fxa-form';
 @import '../protocol/components/custom-menu-list';
+@import 'includes/sticky-promo';
```
maxxcrawford, line 14: Revised SCSS from a CSS bundle to SCSS import config.

media/img/icons/connection.svg
```diff
@@ -0,0 +1,2 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 38.41 41.33" width="38" height="41">
```
maxxcrawford, line 1: White version of Connection icon. I filed an issue on https://github.com/mozilla/protocol-assets/issues/49 to add VPN assets there for eventual Protocol consumption.

media/js/base/sticky-promo.js
```diff
@@ -0,0 +1,16 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    document.addEventListener('DOMContentLoaded', function() {
```
maxxcrawford, line 8: The function was firing before the page was ready. If there's another more suitable document-ready function, lmk!

media/js/base/sticky-promo.js
```diff
@@ -0,0 +1,16 @@
...
+        var closeButton = document.querySelector('.mzp-c-sticky-promo-close');
+
+        closeButton.addEventListener('click', function(e) {
+            e.preventDefault();
+            e.currentTarget.parentNode.classList.add('mzp-a-fade-out');
+        }, false);
```
maxxcrawford, line 14: This code snippet was taken from https://github.com/mozilla/protocol/blob/master/src/assets/js/protocol/protocol-sticky-promo.js

bedrock/base/templates/includes/sticky-promo.html
craigcook: Maybe `data-link-position="sticky promo"`?

media/js/base/sticky-promo.js
alexgibson: Scripts should typically go at the end of the document body. The one here is being requested before the page content, making it render blocking and hence the need for the event listener here. I’d suggest moving the JS either into the page’s JS block, or alternatively into the existing page bundle.

justinkropp:
@maxxcrawford   LGTM 👍

media/js/base/sticky-promo.js
maxxcrawford: I moved the script below and it loaded without the need to wait. I'll move the entire block into the existing page bundle JS instead. Great callout!

media/js/base/sticky-promo.js
```diff
@@ -0,0 +1,15 @@
...
+    'use strict';
+
+    var closeButton = document.querySelector('.mzp-c-sticky-promo-close');
+
+    closeButton.addEventListener('click', function(e) {
+        e.preventDefault();
+        e.currentTarget.parentNode.classList.add('mzp-a-fade-out');
+    }, false);
+
+})();
```
amychurchwell, line 15: This code is available to you via `protocol/js/protocol-sticky-promo.js` so we can add that to the JS bundles instead :D

amychurchwell (changes requested):
Did a first pass and it's looking great so far!

amychurchwell:
x-post from slack for posterity:
- it may be a good idea to push this branch to `run-integration-tests` since it touches several places.
- we should write some tests for the sticky promo functionality.

Hope you have a fun week off! (ﾉ◕ヮ◕)ﾉ*:･ﾟ✧

bedrock/base/templates/includes/sticky-promo.html
maxxcrawford: The issue here is that this file is included into pages that already have `referrals` defined. Is it okay to overwrite them here? 

Example: `bedrock/firefox/templates/firefox/home/index-master.html`
L48 sets the `referrals`, L60 includes the `sticky-promo.html` file.

amychurchwell (changes requested):
LGTM! One tweak + tests & this should be good to go, imo. 👍

maxxcrawford:
Update to this PR/issue. We need to change the default behavior of this so that we can respect the user if they choose to dismiss this promo. Filed an issue in Protocol to add this behavior. This feature will also affect the tests, so we're holding on them until this is landed. 

https://github.com/mozilla/protocol/issues/638

media/js/firefox/sticky-promo.js
```diff
@@ -0,0 +1,40 @@
...
+
+(function() {
+    'use strict';
+
+    var StickyPromo = {};
+    var stickyPromoCookieID = 'firefoxStickyPromo';
+
+    StickyPromo.setCookie = function() {
+        var date = new Date();
+        var cookieDuration = 1 * 24 * 60 * 60 * 1000; // 1 day expiration
```
maxxcrawford, line 13: Is a day the right time length?

media/js/firefox/sticky-promo.js
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    var StickyPromo = {};
+    var stickyPromoCookieID = 'firefoxStickyPromo';
```
maxxcrawford, line 9: I'm setting the cookie ID one place, rather than passing args through each function. Should I put this variable in ALL CAPS?

maxxcrawford (commented):
Note that one "feature" of this implimentation would be if they dismiss this promo across any of the three pages it's on, it won't show on the others.

media/js/firefox/sticky-promo.js
craigcook: A day seems about right to me 👍

media/js/firefox/sticky-promo.js
craigcook: I'm not sure we have an established convention for this. I guess we can look for other examples and see what we've done before.

craigcook: Looks like elsewhere we've done lowercase with dashes, like `firefox-sticky-promo`

bedrock/firefox/templates/firefox/browsers/index.html
```diff
@@ -28,6 +28,8 @@
 {% set referrals = '?utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=firefox-browsers' %}
 {% set _entrypoint = 'mozilla.org-firefox-browsers' %}
 
+{% include '/includes/sticky-promo.html' %}
```
alexgibson, line 31: Inserting the promo outside of `{% block content %}` here results in the `<aside>` coming before the `<head>` in the DOM. The effect is a visible flash of unstyled HTML before the CSS loads in.

We should also consider SEO here. This promo probably isn't the most important content in the page, so can probably come last at the bottom of the page?

bedrock/base/templates/includes/sticky-promo.html
```diff
@@ -0,0 +1,40 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% set promo_referrals = '?utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=sticky-promo' %}
+
+<aside class="mzp-c-sticky-promo mzp-t-product-firefox mzp-t-dark">
```
alexgibson, line 7: Seeing as this relies on JS, suggest adding a class of `hide-from-legacy-ie` to hide the content from older browsers that don't get first-class CSS/JS?

media/css/firefox/includes/_sticky-promo.scss
```diff
@@ -0,0 +1,65 @@
...
+    display: flex;
+    text-decoration: none !important; /* stylelint-disable-line declaration-no-important */
+    margin-bottom: 1rem;
+    align-items: center;
+    font-weight: bold;
+    img {
+        width: 1.75rem;
+        height: 100%;
+        display: block;
+        margin-right: 1rem;
```
alexgibson, line 63: Product logos could use a little RTL love here. Suggest also using protocol's [spacing tokens ](https://protocol.mozilla.org/fundamentals/tokens.html)for margin / padding as opposed to `rem`?

image

media/js/firefox/sticky-promo.js
```diff
@@ -0,0 +1,40 @@
...
+        date.setTime(date.getTime() + cookieDuration); // 1 day expiration
+        Mozilla.Cookies.setItem(STICKY_PROMO_COOKIE_ID, true, date.toUTCString(), '/');
+    };
+
+    StickyPromo.hasCookie = function() {
+        return Mozilla.Cookies.hasItem(STICKY_PROMO_COOKIE_ID);
+    };
+
+    StickyPromo.show = function (){
+        document.addEventListener('DOMContentLoaded', Mzp.StickyPromo.open);
```
alexgibson, line 23: Not sure you need this event listener here as the content should already exist in the DOM by the time this script runs.

alexgibson (changes requested):
I took a look at the test and made a [branch here](https://github.com/alexgibson/bedrock/commit/722ec0f416fb8a7d1e25ca4d03345111e296103c) with a commit that should work OK with the animation (I think/hope). Feel free to take what you need - I'd suggest pushing this branch to `run-integration-tests` to make sure things pass OK 👍 

Also left a few other small comments / suggestions. The main visual thing I noticed when pulling the branch locally, was a flash of unstyled content because of where the promo is inserted in the DOM. Should be an easy fix I think though.

Also: the makes a fairly substantial change to /firefox/new/. We have an [A/B test](https://github.com/mozilla/bedrock/issues/9637) running on that page right now, so we may want to think about timing for when this rolls out?

media/js/firefox/sticky-promo.js
maxxcrawford: Revised.

media/css/firefox/includes/_sticky-promo.scss
maxxcrawford: Revised with `margin-inline-end` and all the tokens. 

image

bedrock/base/templates/includes/sticky-promo.html
maxxcrawford: How does the `hide-from-legacy-ie` class work? By moving the `aside` below the `{% block content %}`, does that resolve the FOUC?

bedrock/firefox/templates/firefox/browsers/index.html
maxxcrawford: Moved.

media/js/firefox/sticky-promo.js
maxxcrawford: Removed it without issue.

bedrock/base/templates/includes/sticky-promo.html
alexgibson: We have an IE specific stylesheet that will hide elements using this class name: https://github.com/mozilla/bedrock/blob/master/media/css/protocol/common-old-ie.scss#L48-L50

For placement you may want to consider making a block in the base template (e.g. `{% sticky_promo_block %}` from which individual pages can populate: https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/base-protocol.html#L93-L102

maxxcrawford: Moved and added the `hide-from-legacy-ie` for good measure. I did not move it into a block for now.

bedrock/base/templates/includes/sticky-promo.html
```diff
@@ -0,0 +1,38 @@
...
+  <button class="mzp-c-sticky-promo-close" type="button">{{ ftl('ui-close') }}</button>
+  <div class="promo-firefox">
+    <h3 class="mzp-c-sticky-promo-title">{{ ftl('firefox-sticky-promo-get-the-latest-firefox') }}</h3>
+    {{ download_firefox_thanks(dom_id='sticky-promo', download_location='other', button_class='mzp-c-button mzp-t-product mzp-t-small') }}
+  </div>
+  <div class="promo-products">
+    <h3 class="mzp-c-sticky-promo-title">{{ ftl('firefox-sticky-promo-meet-our-family-of') }}</h3>
+    <ul class="promo-products-list">
+      <li>
+        <a data-link-name="Browsers" data-link-type="link" data-link-position="sticky-promo" class="promo-products-link" href="{{ url('firefox.browsers.index') }}{{ referrals }}">
```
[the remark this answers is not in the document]
maxxcrawford: I set the new var via `promo_referrals`!

bedrock/firefox/templates/firefox/new/desktop/download.html
```diff
@@ -497,6 +497,8 @@ <h4 class="join-firefox-title">{{ ftl('firefox-desktop-download-youve-already-go
 
 {% endblock %}
 
+{% include '/includes/sticky-promo.html' %}
```
alexgibson, line 500: The sticky promo is still being inserted into the head before the CSS. It also seems to be causing some kind of markup invalidation:

image

I think this is what's causing IE9 to fail on the test run, as it renders this:

image

alexgibson:
We shoould avoid showing this stick promo on small viewports. On my phone it obscures the page, making content inaccessible:

image

bedrock/base/templates/includes/sticky-promo.html
```diff
@@ -0,0 +1,40 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% set promo_referrals = '?utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=sticky-promo' %}
+
+<aside class="mzp-c-sticky-promo mzp-t-product-firefox mzp-t-dark hide-from-legacy-ie">
```
alexgibson, line 7: Seeing as this adds new strings in a global file, should we only display the sticky promo it if the locale has the main headlines translated? (using `if (ftl_has_messages(): `? Right now it seems like this would show up in english everywhere right away.

l10n/en/firefox/sticky-promo.ftl
```diff
@@ -0,0 +1,17 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### URL: https://www-dev.allizom.org/firefox/*
```
alexgibson, line 5: I think this field pops up in Pontoon for translators to click and preview their work. Suggest just putting one URL in here where the promo appears?
```suggestion
### URL: https://www-dev.allizom.org/firefox/
```

tests/pages/regions/sticky_promo.py
```diff
@@ -0,0 +1,25 @@
...
+
+class StickyPromo(BaseRegion):
+
+    _root_locator = (By.CLASS_NAME, 'mzp-c-sticky-promo')
+    _close_locator = (By.CLASS_NAME, 'mzp-c-sticky-promo-close')
+
+    def close(self):
+        promo = self.selenium.find_element(*self._root_locator)
+        self.find_element(*self._close_locator).click()
+        self.wait.until(lambda s: 'mzp-a-fade-out' in promo.get_attribute('class'))
```
alexgibson, line 18: This should fix the [test failure](https://gitlab.com/mozmeao/www-config/-/jobs/841874461) in Chrome:

```suggestion
        self.wait.until(lambda s: not self.is_displayed)
```

alexgibson (changes requested):
Left a few more comments above.

I've been thinking about the [test failure](https://gitlab.com/mozmeao/www-config/-/jobs/841874457) for the language selector, and it doesn't feel quite right to me to move the test to another page, when it's really highlighting a genuine usability problem.

What do you think about making the stick promo hide underneath the footer once you reach the end of a page? This way the information would still be accessible:

```css
.mzp-c-footer {
    position: relative;
    z-index: 11;
}
```

alexgibson:
I managed to gety a [passing test run](https://gitlab.com/mozmeao/www-config/-/pipelines/214636918) with the following changes:

https://github.com/alexgibson/bedrock/commit/3ca1d077bff4d5a7dc9b6439dd10ca18b008db72

l10n/en/firefox/sticky-promo.ftl
maxxcrawford: Revised the URL and ammended the line below about the scope of this text.

maxxcrawford:
> Left a few more comments above.
> 
> I've been thinking about the [test failure](https://gitlab.com/mozmeao/www-config/-/jobs/841874457) for the language selector, and it doesn't feel quite right to me to move the test to another page, when it's really highlighting a genuine usability problem.
> 
> What do you think about making the sticky promo hide underneath the footer once you reach the end of a page? This way the information would still be accessible:
> 
> ```css
> .mzp-c-footer {
>     position: relative;
>     z-index: 11;
> }
> ```

Would it make sense to add a class to the footer when the sticky promo becomes visible? Rather than modifying the footer everywhere? That way it is only present when needed.

alexgibson:
> Would it make sense to add a class to the footer when the sticky promo becomes visible? Rather than modifying the footer everywhere? That way it is only present when needed.

If you wanna go the extra mile there then it sounds good to me 👍

bedrock/base/templates/includes/sticky-promo.html
```diff
@@ -0,0 +1,41 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% set promo_referrals = '?utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=sticky-promo' %}
+{% if ftl_has_messages('firefox-sticky-promo-get-the-latest-firefox', 'firefox-sticky-promo-meet-our-family-of') %}
```
maxxcrawford, line 6: @alexgibson Added the `ftl_has_messages` IF statement. I only tested for the two new strings, and the other strings are all brand/property names that should already be localized.

media/css/protocol/components/_footer.scss
```diff
@@ -4,6 +4,11 @@
 
 @import '../../../protocol/css/includes/lib';
 
+.mzp-c-footer.sticky-overlay {
```
maxxcrawford, line 7: Is `.sticky-overlay` an okay class name? Not sure if there's a better naming convention I can utilize.

media/js/firefox/sticky-promo.js
```diff
@@ -0,0 +1,58 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    // Check if promo exists on the page
+    if (!document.querySelector('.mzp-c-sticky-promo')) {
+        return;
+    }
```
maxxcrawford, line 11: This is basic "check if it's there logic".

media/js/firefox/sticky-promo.js
```diff
@@ -5,6 +5,13 @@
 (function() {
     'use strict';
 
+    // Check if promo exists on the page or if smaller than tablet.
```
alexgibson, line 8: Seeing as this script is self-contained, it might make sense to treat it as a form of progressive enhancement. Suggest moving all of the logic here inside a `Mozilla.run()` callback?

https://bedrock.readthedocs.io/en/latest/browser-support.html#feature-detection-using-javascript

media/css/protocol/components/_footer.scss
alexgibson: This is a little unusual since class names usually relate to the element itself and not some other element in the page. Maybe `intersect-sticky-overlay` is a little more descriptive?

Also suggest adding a comment in the CSS to explain why this exists.

media/js/firefox/sticky-promo.js
```diff
@@ -30,6 +37,10 @@
         // Open promo
         Mzp.StickyPromo.open();
 
+        // Add modifier class to the footer to make sure the language selection drop-down is not obscured by the sticky promo
+        var footer = document.querySelector('.mzp-c-footer');
+        footer.classList.add('sticky-overlay');
```
alexgibson, line 42: Might be worth checking that a page has a footer before attempting to modify the class?

```javascript
if (footer) {
    footer.classList.add('intersect-sticky-overlay');
}

File: tests/pages/firefox/new/sticky_promo.py [hunk 27]
```diff
@@ -0,0 +1,25 @@
...
+
+class StickyPromoPage(BasePage):
+
+    URL_TEMPLATE = '/{locale}/firefox/new/'
+
+    _sticky_promo_modal_content_locator = (By.CSS_SELECTOR, '.mzp-c-sticky-promo.mzp-a-slide-in')
+
+    def wait_for_page_to_load(self):
+        self.wait.until(lambda s: self.seed_url in s.current_url)
+        promo = self.find_element(*self._sticky_promo_modal_content_locator)
```
alexgibson, line 19: Not sure when this was removed, but I'd suggest still waiting for the `loaded` class on the HTML element before running the tests (in addition to what you already have here): https://github.com/mozilla/bedrock/blob/master/tests/pages/base.py#L32-L33

media/js/firefox/sticky-promo.js
```diff
@@ -0,0 +1,60 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+(function() {
+    'use strict';
+
+    // Check if promo exists on the page or if smaller than tablet.
+    var matchMediaDesktop = window.matchMedia('(min-width: 768px)').matches;
+
+    if (!document.querySelector('.mzp-c-sticky-promo') || !matchMediaDesktop) {
```
alexgibson, line 11: I see two calls to `document.querySelector('.mzp-c-sticky-promo')` in the script (the other in `bindEvents`), so suggest using a variable for reference to avoid needing to query the DOM twice? e.g.

```javascript
var promo = document.querySelector('.mzp-c-sticky-promo');

if (!promo || !matchMediaDesktop) {

}
````

**Review by alexgibson — CHANGES_REQUESTED** [review]
A couple of nits, but otherwise I think this is good and ready to go. r+wc nice work! 🐊

File: media/css/protocol/components/_footer.scss (hunk 23 above)
Comment (maxxcrawford) [reply media/css/protocol/components/_footer.scss:7]: Comment added and class name updated.

File: media/js/firefox/sticky-promo.js (hunk 26 above)
Comment (maxxcrawford) [reply media/js/firefox/sticky-promo.js:42]: Added! 

Side question - because we know which pages this promo is being added to, isn't it assumed a footer is there?

Comment (alexgibson) [reply media/js/firefox/sticky-promo.js:42]: The footer is there for all pages currently, but this seems like a small enough change to cater for unanticipated futures.

**Comment by alexgibson** [comment]
Looks like this needs a rebase. I also don't see the changes made following the last round of review, so maybe the commits need pushing from your local branch still?

File: media/css/protocol/components/_footer.scss [hunk 29]
```diff
@@ -4,8 +4,15 @@
 
 @import '../../../protocol/css/includes/lib';
 
-//* -------------------------------------------------------------------------- */
-// Footer section styles
+// This class is added to address an issue with the sticky promo component on specific pages overlapping the language selection box in the footer.
+.mzp-c-footer.is-intersecting-sticky-overlay {
```
alexgibson, line 8: I think this needs to be `.c-footer.` now that the nav has changed (same below)

alexgibson (changes requested):

media/css/protocol/components/_footer.scss
```diff
@@ -4,8 +4,15 @@
...
 
-//* -------------------------------------------------------------------------- */
-// Footer section styles
+// This class is added to address an issue with the sticky promo component on specific pages overlapping the language selection box in the footer.
+.c-footer.is-intersecting-sticky-overlay {
+    position: relative;
+    z-index: 11;
+}
+
+.c-footer.mzp-t-firefox {
```
alexgibson, line 13: Actually, I don't think this selector is related to this PR, maybe copy pasta from a rebase?

media/js/firefox/sticky-promo.js
```diff
@@ -0,0 +1,66 @@
...
+        };
+
+        StickyPromo.show = function (){
+        // Open promo
+            Mzp.StickyPromo.open();
+
+            // Add modifier class to the footer to make sure the language selection drop-down is not obscured by the sticky promo
+            var footer = document.querySelector('.c-footer');
+            if (footer) {
+                footer.classList.add('sticky-overlay');
```
alexgibson, line 44:
```suggestion
                footer.classList.add('is-intersecting-sticky-overlay');
```

alexgibson (changes requested):

media/js/firefox/sticky-promo.js
maxxcrawford: This is now fixed. Sorry. I was so focused on fixing the `.mzp-c-footer` issue. 🙈

media/css/protocol/components/_footer.scss
maxxcrawford: Yeah – That got dropped in from the rebase. Should I drop it?

maxxcrawford: Removed.

alexgibson (approved):
I think this is ready, nice work! 💯

peiying2:
@maxxcrawford your PR on l10n side is now combined with Amy's PR which is pending for fixes: https://github.com/mozilla-l10n/www-l10n/pull/127. Could you separate it out? Otherwise your page would not get into Pontoon anytime soon. Thank you!

amychurchwell:
The l10n PR for this copy now lives here: https://github.com/mozilla-l10n/www-l10n/pull/140

PR #9450, opened 2020-09-26
 Updates to EN September Campaign page (#9440) 

stephaniehobson:
## Description

- allow multiple links per checklist item
- add Facebook link
- add reading list link
- update copy
- add section header
- finalize GIFs as much as possible
- add production URLs and redirects
- comment out unfinished links & JS

## Issue / Bugzilla link

#9440

## Testing

http://localhost:8000/en-US/firefox/unfck/


stephaniehobson:
@alexgibson Some of the data science folks would like something to be live so they can check that UTMs are being captured. This PR includes adding production URLs. If you have any concerns with that let's talk on Monday. I didn't make any promises ;)

bedrock/firefox/redirects.py
```diff
@@ -653,4 +653,8 @@ def firefox_channel(*args, **kwargs):
     # issue 8641
     redirect(r'^/firefox/windows-64-bit/?$', 'firefox.browsers.windows-64-bit'),
     redirect(r'^/firefox/best-browser/?$', 'firefox.browsers.best-browser'),
+
+    #unfck campaign
+    redirect(r'^firefox/unfuck(/.*)?', '/firefox/unfck'),
```
alexgibson, line 658: I think the regex can be simplified a bit here, and we can also use the page name for the destination?
```suggestion
    redirect(r'^firefox/unfuck/?$', 'firefox.campaign.unfck.index'),
```

Not a blocker, but having a couple of redirect tests would be good here too for an extra measure.

alexgibson (changes requested):
One suggestion with the redirects, and there are a couple of flake8 errors Circle picked up on. Other than that, I'm totally fine with this going to prod for testing (good idea).

I'm going to prepare another PR for the German page today to tidy up a few remaining things. If we could get that reviewed and merged also - do you think we can get the data science team to test both versions of the page?

bedrock/firefox/redirects.py
alexgibson: I added one additional redirect (plus tests) in https://github.com/mozilla/bedrock/pull/9454 if it makes things simpler (I'm also happy to rebase if needed, whatever works).

alexgibson (approved):

PR #9464, opened 2020-09-29
Exclude Sentry JS from IE < 10 (#8444)

alexgibson:
## Description
As per https://docs.sentry.io/platforms/javascript/troubleshooting/supported-browsers/

## Issue / Bugzilla link
#8444

## Testing
- [ ] Sentry JS snippet should not run in old IE browsers

### DIFF
--- a/bedrock/base/templates/base-protocol.html
+++ b/bedrock/base/templates/base-protocol.html
@@ -104,7 +104,9 @@
     {# Issue 8444 #}
     {% block sentry_client %}
       {% if settings.SENTRY_DSN and switch('sentry-js') %}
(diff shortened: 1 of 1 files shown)


alexgibson:
@pmac r? I noticed the Sentry JS bundle throws an error in old IE versions, so let's not load the script there.

I considered seeing if we could polyfill, but they say they don't strictly support those old browsers. I think having coverage for IE 10 and upward is fine, given https://bedrock.readthedocs.io/en/latest/browser-support.html

pmac (approved):
Sounds good :shipit:

PR #9476, opened 2020-09-29
Update daylight promo banner [fix #9412]

### LINKED ISSUE #9412: Update banner on mozilla.org website
## Description
With Firefox for iOS daylight coming out on 9/22/20, we want to update the Daylight banner to include iOS in copy and imagery.

I'll deliver copy and assets. The button functionality is fine as it supports mobile users on both platforms.

Banner to replace: image

---

#### Success Criteria
- [ ] New copy is shipped
- [ ] New visuals are shipped

craigcook:
## Description
Adds a new minimal banner promoting the new Firefox for Android and iOS. This will replace the previous banner but I ended up building it as a separate banner so we can get this into production ahead of time and control them both with separate switches. We don't intend to make the switch until Oct 8 (pending release of Fx 29 for iOS). Doing this weird double-banner/separate-switch treatment gives us more control on the timing. We'll turn the old one off at the same time as we turn the new one on, then we can delete the old one in a follow-up PR.

## Issue / Bugzilla link
#9412 

## Testing
http://localhost:8000/
http://localhost:8000/firefox/mobile/

The banner is hidden if the strings aren't translated so you won't see it in non-English locales while testing.

You'll see both banners stacked in dev mode (because both switches will default to ON; we won't let that happen in prod). You can close the old banner to see what the new one should look like by itself.

### DIFF
--- a/bedrock/mozorg/templates/mozorg/home/home-fr.html
+++ b/bedrock/mozorg/templates/mozorg/home/home-fr.html
@@ -20,7 +20,7 @@ Le saviez-vous ? Mozilla, le concepteur de Firefox, se bat pour qu’Internet, u
   {% endif %}
 
   {% if switch('firefox-daylight-promo-banner') %}
-    {% include 'includes/banners/daylight/daylight-promo.html' %}
+    {{ css_bundle('daylight-promo-banner') }}
   {% endif %}
 {% endblock %}
 
--- a/bedrock/mozorg/templates/mozorg/home/home-fr.html
+++ b/bedrock/mozorg/templates/mozorg/home/home-fr.html
@@ -18,6 +18,10 @@ Le saviez-vous ? Mozilla, le concepteur de Firefox, se bat pour qu’Internet, u
   {% if switch('firefox-daylight-launch-banner') %}
     {{ css_bundle('daylight-launch-banner') }}
   {% endif %}
+
+  {% if switch('firefox-daylight-promo-banner') %}
+    {% include 'includes/banners/daylight/daylight-promo.html' %}
+  {% endif %}
 {% endblock %}
 
 
--- a/media/css/base/banners/daylight-promo.scss
+++ b/media/css/base/banners/daylight-promo.scss
@@ -45,8 +45,9 @@ $image-path: '/media/protocol/img';
         }
 
         .c-cta {
-            padding: 0 $spacing-xl;
             line-height: 0;
+            min-width: 180px;
+            padding: 0 $spacing-xl;
         }
     }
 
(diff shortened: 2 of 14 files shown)


bedrock/mozorg/templates/mozorg/home/home-fr.html
```diff
@@ -18,6 +18,10 @@
   {% if switch('firefox-daylight-launch-banner') %}
     {{ css_bundle('daylight-launch-banner') }}
   {% endif %}
+
+  {% if switch('firefox-daylight-promo-banner') %}
```
alexgibson, line 22: Should this be a CSS bundle instead?

image

media/css/base/banners/daylight-promo.scss
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+
+    // Conditional content
+    .hidden-on-desktop {
+        display: none;
+    }
+
+    .android,
+    .ios {
```
alexgibson, line 60: It looks like the conditional classes aren't quite working here:
```suggestion
    .android &,
    .ios & {
```

media/css/base/banners/daylight-promo.scss
```diff
@@ -0,0 +1,130 @@
...
+        .hidden-on-desktop {
+            display: inline;
+        }
+
+        .hidden-on-mobile {
+            display: none;
+        }
+    }
+
+    .android .mobile-download-buttons {
```
alexgibson, line 70:
```suggestion
    .android & .mobile-download-buttons {
```

media/css/base/banners/daylight-promo.scss
```diff
@@ -0,0 +1,130 @@
...
+        .android {
+            display: inline-block;
+        }
+
+        .ios {
+            display: none;
+        }
+    }
+
+    .ios .mobile-download-buttons {
```
alexgibson, line 80:
```suggestion
    .ios & .mobile-download-buttons {
```

alexgibson:
Minor responsive styling issue on the button at small-mid screen sizes:

image

alexgibson (changes requested):
Spotted one bug on the French home page, and it looks like the conditional buttons aren't quite working due to the CSS selectors. Otherwise this looks good and I welcome the newer more minimal banner styling 🎉

bedrock/mozorg/templates/mozorg/home/home-fr.html
craigcook: Oops, botched copypasta.

craigcook:
Updated! I think those selectors broke when I nested the whole thing in an ID, which I did at the last minute only so the two banners could briefly coexist but neglected to re-test the conditional bits.

alexgibson (commented):
r+ 🍡

alexgibson (approved):

PR #9482, opened 2020-09-30
Add blog post URLs

stephaniehobson:
## Description

Add blog post URLs

## Issue / Bugzilla link

#9334

### DIFF
--- a/bedrock/firefox/templates/firefox/campaign/unfck/index.html
+++ b/bedrock/firefox/templates/firefox/campaign/unfck/index.html
@@ -142,7 +142,9 @@
       info='The more people who use independent tech like Firefox, the greater the effect. Get yourself some unfcked messaging, content, email and more.',
       alt='Un-fck. Un-friend. Un-hate. Un-real. Un-follow.',
       link_text='Meet our Friends',
-      link_url='#',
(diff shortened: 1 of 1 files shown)


bedrock/firefox/templates/firefox/campaign/unfck/index.html
```diff
@@ -181,7 +181,7 @@ <h2 class="c-section-title u-visually-hidden">What we’re doing</h2>
         title='Here’s how Firefox unfcks the internet',
         ga_title='card-unfck',
         image_url=None,
-        link_url='',
+        link_url='https://blog.mozilla.org/firefox/how-to-unfck/',
```
craigcook, line 184: I'm not seeing UTM params added to this one, but they're on the other one... not sure what's up with that, maybe because it's a card?

craigcook (changes requested):
The second link doesn't seem to be getting UTM params, do they need to be hardcoded? 

Neither post is published yet but I verified the slugs are correct.

bedrock/firefox/templates/firefox/campaign/unfck/index.html
craigcook: Oh is it missing `data-cta-text`? That's the only difference I see.

craigcook (approved):

PR #9493, opened 2020-10-01
Update browser comparison version disclaimer [fix #9481]

### LINKED ISSUE #9481: Update disclaimer on browser comparison pages
## Description

By request from Tyler:

`The comparisons made here were done so across browser release versions as follows:`

should be updated to:

`The comparisons made here were done so with default settings and across browser release versions as follows:`

craigcook:
## Description
Updates the disclaimer intro to clarify that comparisons were made with default settings.

## Issue / Bugzilla link
#9481 

## Testing
http://localhost:8000/firefox/browsers/compare/ (and subpages)

### DIFF
--- a/bedrock/firefox/templates/firefox/browsers/compare/brave.html
+++ b/bedrock/firefox/templates/firefox/browsers/compare/brave.html
@@ -243,7 +243,7 @@
 
       <div class="compare-content compare-version-disclaimer">
         <p>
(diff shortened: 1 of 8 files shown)


maxxcrawford (approved):
LGTM! Confirmed fall back works for Spanish (ES), and each sub page has the correct string in EN.

PR #9508, opened 2020-10-02
Add tests for CDN infra

pmac:
Add tests for:

* /media/* urls
* locale redirects and vary headers
* geolocation

Fix https://github.com/mozmeao/infra/issues/1348


pmac:
@glogiotatidis @duallain this is still WIP but this is the start I've gotten if you want to have a look.

pmac:
you run them thusly:

```
$ pytest --base-url https://www.mozilla.org test_infra
```

pmac:
I've added some TLS tests to the suite. You grab the test data via [ssllabs-scan](https://github.com/ssllabs/ssllabs-scan) (can be installed on Mac with `brew install ssllabs-scan`) by running `make TEST_DOMAIN=www.mozilla.org tls-test-data`. The default `TEST_DOMAIN` is `www.mozilla.org`, so you can leave that part off if you want data for that domain.

pmac:
I decided not to check in the `tls.json` file because:

1. it is huge and
2. it's only really useful to run manually and when you've just downloaded the data

duallain:
@pmac I'm sure I'm just doing something wrong, but this didn't end up working for me:

> you run them thusly:
> 
> ```
> $ pytest --base-url https://www.mozilla.org test_infra
> ```

I get ```$ pytest --base-url https://www.mozilla.org test_infra
ERROR: usage: pytest [options] [file_or_dir] [file_or_dir] [...]
pytest: error: unrecognized arguments: --base-url
  inifile: /Users/aalexander/sandbox/bedrock/setup.cfg
  rootdir: /Users/aalexander/sandbox/bedrock```

I did setup a venv, and then `pip install -r requirements/dev.txt` so I think I should have a reasonable version of pytest.  It seems like when I add the command to run-tests.sh it runs ok, so it must be some setup difference.

(this is the patch that ended up working)
```$ git diff
diff --git a/bin/run-tests.sh b/bin/run-tests.sh
index 9639d46d4..873c6c1b5 100755
--- a/bin/run-tests.sh
+++ b/bin/run-tests.sh
@@ -9,3 +9,4 @@ python manage.py version
 python manage.py migrate --noinput
 py.test lib bedrock
 py.test -r a tests/redirects
+pytest --base-url https://www.mozilla.org test_infra
diff --git a/docker-compose.yml b/docker-compose.yml
index c954b90e0..4c604675c 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -58,6 +58,7 @@ services:
       - ./root_files/:/app/root_files:delegated
       - ./scripts/:/app/scripts:delegated
       - ./tests/:/app/tests:delegated
+      - ./test_infra/:/app/test_infra:delegated
       - ./wsgi/:/app/wsgi:delegated
       - ./l10n/:/app/l10n:delegated
```

We should also add the run command somewhere.  I'm guessing there's not too many people who will run this, so it may not make sense to add to Makefile or the readme or what have you.  What do you think of just putting the command into a comment in the test file itself? Or making a terse README.md in that directory?  Just something so we don't have to go spelunking for it next time.

pmac:
@duallain okay. I've added a README to the test_infra directory, switched to using docker to run the ssllabs-scan command, and added a make command to run the cdn tests. It will also now automatically download the TLS report if you don't already have the file locally. Give it a whirl and let me know what you think.

duallain (approved):

duallain:
This new setup ends up working really well, I find it really easy to work with.  Thanks for all of it @pmac

PR #9536, opened 2020-10-12
Improve GA download reporting for legacy / unknown browsers (Fixes #9535)

### LINKED ISSUE #9535: Improve GA download reporting for legacy / unknown browsers
## Description

Since we rolled out the simplified download button in (https://github.com/mozilla/bedrock/issues/8860) we've been seeing a small number of download events in GA come through as `Firefox for unavailable`, where we haven't been able to determine their operating system.

image

The highest proportion of these events appears to be users on IE9. Here I think we're not setting the event as "Firefox for Desktop" explicitly. The rest seem to be from old versions of other operating systems and weird edge case browsers.

Suggested solution:

- Make sure we're registering as `Firefox for Desktop` on old IE versions.
- Set as `Firefox for Unsupported` as a catch all for other instances.

alexgibson:
## Description
- Fixes GA bug where old IE browsers are not setting download event label correctly.
- Add 'Unsupported' fallback label for other platforms.

## Issue / Bugzilla link
#9535

## Testing
Demo: https://www-demo2.allizom.org/en-US/

- [x] Correct data attributes should be viewable on download links is old IE browsers:

image

### DIFF
--- a/media/js/base/mozilla-utils.js
+++ b/media/js/base/mozilla-utils.js
@@ -56,6 +56,10 @@ if (typeof window.Mozilla === 'undefined') {
             data.name = 'Android';
             data.version = 'android';
             break;
+        default:
+            data.os = 'Unsupported';
+            data.name = 'Unsupported';
+            data.version = 'unsupported';
         }
 
         return data;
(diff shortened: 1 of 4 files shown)


stephaniehobson (approved):
R+ 📒

Issue #9327, opened 2020-08-25
Sign in/sign up links in footer have malformed styling on some pages


craigcook:
## Description
The "sign in" and "sign up" links in the footer are getting some added styling from the FxA form on any page where that form appears. This seems due to the reocurring class name `fxa-signin` in the footer, but the same name styles the "already have an account?" link in the form to set the small font size. The /firefox/products/ page also adds a top margin.

![image](https://user-images.githubusercontent.com/205591/91232634-5454d880-e6e4-11ea-8858-c8100f1ec543.png)

The simplest fix is probably to change the class in either the FxA form or in the footer to something more unique.

stephaniehobson:
Careful that some JS is not relying on those classes too.

soulhunter10737:
How it should look like?

PR #9538, opened 2020-10-12
remove fxa-signin class from footer list items

slightlyoffbeat:
The class is not needed for this element. This should fix bug 9327

## Description
This fixes an issue where some links in the footer receive additional styles when a FxA form is on the page

This class appears to be unnecessary for the footer (the only scss files that referenced this class were for the FxA signup). If this is true, a new unique class is not needed, and removing from the footer should suffice. 

## Issue / Bugzilla link
fixes bug #9327

### DIFF
--- a/bedrock/base/templates/includes/protocol/footer/firefox.html
+++ b/bedrock/base/templates/includes/protocol/footer/firefox.html
@@ -44,7 +44,7 @@
             {{ ftl('footer-join') }}
           </h5>
           <ul class="mzp-c-footer-list">
-            <li class="fxa-signin">
+            <li>
               {{ fxa_button(
                 entrypoint='mozilla.org-firefoxfooter',
                 button_text=ftl('footer-sign-up'),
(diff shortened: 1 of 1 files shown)


maxxcrawford (approved):
This patch fixes it (including the margin issue on the [/products](https://www.mozilla.org/en-US/firefox/products/) page too).

I am seeing one issue, but this may warrant a separate issue: The language drop-down (also in the footer) is different widths between the two environments: 

image

_Note - This may be just for me!_

@stephaniehobson I didn't see any references in any JavaScript files to the removed class. `.fxa-signin`. I don't think this will be an issue. The class that DOES need to exist is the `.js-fxa-cta-link`. That class is the hook to add all the Amplitude-specific data attributes on production.

craigcook:
> I am seeing one issue, but this may warrant a separate issue: The language drop-down (also in the footer) is different widths between the two environments:

It doesn't look like that select has an explicit width set, just max-width, so the actual rendered width will depend on the longest option in the menu. That can vary from page to page depending on available translations so I think this is fine, unless we want to explicitly make it `width: 100%;`

alexgibson:
@slightlyoffbeat congrats on your first PR to bedrock! 🌮

PR #9545, opened 2020-10-14
[WIP] Add scroll margin to release note items

vhfmag:
## Description

Adds scroll-margin-top to release notes' list items to avoid sticky header from overlaying an item when using fragments (e.g. https://www.mozilla.org/en-US/firefox/83.0a1/releasenotes/#note-788467). This probably shouldn't have a hard-coded value, but I didn't manage to understand which SASS variables to use.


stephaniehobson (approved):
R+ 📜 

Thanks for the patch!

PR #9546, opened 2020-10-14
Copy updates for US election (#9334)

stephaniehobson:
## Description

- split actions into 2 sets
- add election focused actions
- add and update gifs
- increase space around second section-title

## Issue / Bugzilla link

#9334

## Testing

- http://localhost:8000/en-US/firefox/unfck/
- high-res, low-res, and still images appear for new and edited GIFs
- the two tweet buttons work but pre-populate different content


craigcook (approved):
Looks good, just one URL we might want change but not a showstopper. r+wc 🤬

Issue #9547, opened 2020-10-14
October 2020 Homepage Content Updates


maxxcrawford:
## Description

- Hide Cards 6-9
- Update Pocket Main Card copy/design 

---

#### Success Criteria
- [ ] Use new content from https://github.com/mozmeao/www-admin/issues/83
- [ ] Tasks should be represented here as something like "Approved designs attached to this issue"

PR #9283, opened 2020-08-05
Fix #9278

stephaniehobson:
## Description

Stop text from overlapping image in mobile. Updates to both the default and exp template.

## Issue / Bugzilla link

Fix #9278

## Testing

http://localhost:8000/en-US/firefox/new/

In ios or android, or you can spoof it by changing the body class from `osx` to `android` or `ios`

PR #9548, opened 2020-10-15
October 2020 Homepage (EN) Content Updates

maxxcrawford:
#9283  Description

Adding new content to the EN homepage. This updates removes some cards (Cards 6-9, four-across promoting products and Cards 10-11, two-across) and adds a new Pocket hero section. 

Note that this PR is blocked by https://github.com/mozmeao/www-admin/pull/84

TODO: 

- [x] Optimize Pocket image
- [x] Remove duplicate (alt) Pocket images 
- [x] Add UTMs to Pocket CTA
- [ ] Browser test

## Issue / Bugzilla link

#9547 

## Testing

Review link: https://www-demo-pr-9548.herokuapp.com/en-US/


media/css/mozorg/home/home-en.scss
```diff
@@ -72,6 +72,22 @@ $image-path: '/media/protocol/img';
     }
 }
 
+//* -------------------------------------------------------------------------- */
+// Pocket Hero
+
+.pocket-sunny-hero {
+    &.mzp-t-product-firefox .mzp-c-hero-title {
```
maxxcrawford, line 79: Commandeering this class.  I didn't see one for Pocket, so I overwrote this one. LMK if there's a better way to manage this.

bedrock/mozorg/templates/mozorg/home/home-en.html
```diff
@@ -105,10 +98,17 @@ <h1>{{ self.page_title() }}</h1>
         include_highres_image=True
       )}}
 
-      <div class="mzp-l-card-half">
-        {{ content_card('card_10') }}
-        {{ content_card('card_11') }}
-      </div>
+      {% call hero(
```
stephaniehobson, line 101: The hero is a little narrow because it is being nested inside a `mzp-l-content` div. (this is more obvious on the text side than the image side). Let me know if you want to chat about a fix but I don't think there's a "right" way to fix it so I'm happy to leave the how up to you.

bedrock/mozorg/views.py
```diff
@@ -117,7 +117,7 @@ def home_view(request):
 
     if locale.startswith('en-'):
         template_name = 'mozorg/home/home-en.html'
-        ctx['page_content_cards'] = get_page_content_cards('home-en', 'en-US')
+        ctx['page_content_cards'] = get_page_content_cards('home-2020-oct', 'en-US')
```
stephaniehobson, line 120: I'd drop the `oct` from the folder name since there's a reasonable chance we'll be updating the cards in November, but not changing the layout. (need to change in www-admin too). Not a blocker though.

media/css/mozorg/home/home-en.scss
stephaniehobson: 👍

bedrock/mozorg/templates/mozorg/home/home-en.html
maxxcrawford: I pulled it out of the `.mzp-l-content` wrapper div and put it in its own. I also modified the image to have a high-res variation, match the 3:2 ratio. This overall looks much better: 

 Before: 
image

After: 
image

bedrock/mozorg/views.py
maxxcrawford: Revised!

stephaniehobson (approved):
R+ 🏡

Issue #9453, opened 2020-09-28
Implement local Selenium Grid for testing


glogiotatidis:
Success Criteria
 - [x] A successful test run of our selenium tests in multiple browsers on local Selenium Grid
 - [x] Easy access to failed test logs
 - [x] Recordings for tests - Opted for having the ability to easily run tests on Saucelabs to get video recordings. For screenshots are available through Gitlab's UI as artifacts.

glogiotatidis:
Note to self:

@pmac had a good idea to send failed tests to Saucelabs to get a second try and video recordings.

glogiotatidis:
My work here is done. We will upgrade the runners for speed on Monday, see https://github.com/mozmeao/infra/issues/1359.

@alexgibson good to close? do you need anything else?

alexgibson:
Good to close, thanks @glogiotatidis 💯

PR #9565, opened 2020-10-19
Issue #9453 - Replace pytest-xdist with pytest-parallel.

glogiotatidis:
----

## Description

This replaces pytest-xdist with pytest-parallel for parallel run of the integration tests. xdist has been causing memory leaks while working on #9453, pytest-parallel on the other hand seems to be working just fine. According to its documenation it's ideal for use with Selenium tests.

## Issue / Bugzilla link

Part of #9453.


glogiotatidis:
I remember we always had issues with `xdist` so I guess it's fine to replace it? Thoughts?

pmac:
> I remember we always had issues with `xdist` so I guess it's fine to replace it? Thoughts?

We have indeed. It caused a ton of problems when we moved from raven to sentry-sdk for some still unknown reason. It also causes problems with certain kinds of tests and test discovery. I'm glad there's an alternative.

alexgibson:
This looks good to me and I pushed the branch for a test run, but it seems to have stalled when building: https://gitlab.com/mozmeao/www-config/-/pipelines/204600453

I don't think that's related to the change here, but not quite sure what I can do to get it unstuck?

pmac:
I think we can remove some more dependencies. I installed both of these into fresh venvs and these were only installed by xdist and not parallel: `{'pytest-forked', 'apipkg', 'execnet'}`. Perhaps we can remove these as well. I know excnet was the one I discovered was causing problems with sentry-sdk.

alexgibson (approved):
I'll defer to @pmac on the other dependencies, but looking at https://gitlab.com/mozmeao/www-config/-/jobs/799041238 this looks good to me.

The output is a little confusing in terms of which tests have `PASSED` next to them (some seem to have `PASSEDPASSED`, and others are missing anything). I'm guessing this is because the output is happening in parallel which may effect the ordering a little differently to pytest-xdist.

Still, I think the most important thing here is that we capture failures in a readable way, which will still be output at the end in the test result. So if this gives us extra stability, I'm all for it. r+

glogiotatidis:
> The output is a little confusing in terms of which tests have `PASSED` next to them (some seem to have `PASSEDPASSED`, and others are missing anything). I'm guessing this is because the output is happening in parallel which may effect the ordering a little differently to pytest-xdist.

> Still, I think the most important thing here is that we capture failures in a readable way, which will still be output at the end in the test result. So if this gives us extra stability, I'm all for it. r+

I agree, output is not great. I also tried `pytest-progress` to display the current progress of tests but it wasn't really that good output to read. xdist has the best in progress output, but as you say, the important part is stability and the output at the end.

glogiotatidis:
Removed the extra packages @pmac said

PR #9566, opened 2020-10-19
Add BrowserStack as a functional test driver option

alexgibson:
Adds Browser Stack as a driver option when setting desired capabilities


pmac (approved):
:shipit:

PR #9590, opened 2020-10-22
Add link to explain Pocket [no bug]

stephaniehobson:
## Description

Add a second link in the Pocket section of EN UTI page.

## Issue / Bugzilla link

No bug.


craigcook (approved):
👖

PR #9593, opened 2020-10-23
Only set browserstack capabilities for IE 11

alexgibson:
@glogiotatidis r?

This is just to help me test something in browserstack. It should not effect any tests that we're running in the default pipeline.


glogiotatidis (approved):
:cake:

Issue #9523, opened 2020-10-05
Little Book of privacy Coding


rraue:
## Description

In https://github.com/mozilla/bedrock/issues/9520 we build a Little book of Privacy page. This issue tracks coding.

- Url: /firefox/privacy/book/
- Open for translation

---

#### Success Criteria
- [x] Page is coded
- [x] Page is in pontoon

rraue:
The "Download pdf" button needs to be behind a switch first, as we need to do changes in the pdf itself to reflect the changes from legal, so we cant have the download at first, but that should resolve in the next weeks, so we should code the download button already.

rraue:
We should change the og:image on this page, to fit better the content, as the EU team plans some social/pr campaign with it. Instead of the standard one lets take this one: https://drive.google.com/file/d/14p1PLi5zuZKWvnbWnleS6qKy9yjjlAnt/view?usp=sharing

Issue #9522, opened 2020-10-05
Little Book of privacy Design


rraue:
## Description

in https://github.com/mozilla/bedrock/issues/9520 we build a little book of privacy page, this issue is the track the design.

---

#### Success Criteria
- [x] Design is attached
- [x] @justinkropp approved
- [x] Chad approved
- [ ] Dev knows what to do

rraue:
Design Mockup:
![little book design idea](https://user-images.githubusercontent.com/36700618/95090601-bf52ef80-0725-11eb-9b6d-ca349a437b69.png)

- should be included in the /firefox/privacy/ subnav
- Has 2 cta - download firefox & download pdf in en & de - in all other locales only download as cta
- easy right - left design
- illustrations we take from the pdf or from the .ai file in the gdrive: https://drive.google.com/drive/folders/1EPL0Z_AlRpK6Teq5SQQk5kgEIpD0nFbU and in caronlyns folder: https://drive.google.com/drive/folders/1xvGsE_4YNPsKi5eNONwc4Ldy-B54chqo

markteater:
@rraue here's my initial pass at a basic conversion of The Little Book of Privacy:
![Screenshot_2020-10-15 Little Book of Privacy - Protocol Design System](https://user-images.githubusercontent.com/1934450/96271684-2cbf1580-0f9b-11eb-9046-b43b74eed5af.jpg)

rraue:
One of the things missing in this Design (sorry I didnt catch this earlier) is the subnav of https://www.mozilla.org/en-US/firefox/privacy/ that should include the little book of privacy and should be included on top of this page here as well.

Issue #9520, opened 2020-10-05
Epic: Little Book of Privacy Page


rraue:
## Description

We want to build a page for the little book of privacy
url: /firefox/privacy/book/

Pitch: https://docs.google.com/document/d/17aL69fIIUhftGhFnx87OcfYcWoecs-xno1a1rT6_Nc4
Brief: https://docs.google.com/document/d/1GxaA7BXQhuV4BseXXGOKFUxCtmKyUIrrCR-ILZ5UiaU/edit?usp=sharing
Gdrive: https://drive.google.com/drive/folders/1EPL0Z_AlRpK6Teq5SQQk5kgEIpD0nFbU?usp=sharing

---

#### Issues
- [x] Content https://github.com/mozilla/bedrock/issues/9521
- [x] Design  https://github.com/mozilla/bedrock/issues/9522
- [x] Coding https://github.com/mozilla/bedrock/issues/9523

Issue #9490, opened 2020-10-01
Epic: Evergreen Content for SEO 


rraue:
## Description

What is the overall outcome we aim to achieve? 
We will create 20 evergreen pages for SEO growth. This issue track progress.

Principle:
existing designs, most preferred protocol components only

Link to pitch, brief, or background documents.

---

#### Issues
- [ ] Translate Page https://github.com/mozilla/bedrock/issues/9759
- [ ] Fingerprinting page https://github.com/mozilla/bedrock/issues/9761
- [ ] Password Page https://github.com/mozilla/bedrock/issues/9760
- [x] Pillar Page https://github.com/mozilla/bedrock/issues/9503
- [x] Quantum page https://github.com/mozilla/bedrock/issues/9499
- [x] FAQ page #9491
- [x] Chromebook Page https://github.com/mozilla/bedrock/issues/9512
- [x] Sync Page https://github.com/mozilla/bedrock/issues/9516
- [x] Little Book of Privacy Page https://github.com/mozilla/bedrock/issues/9520

PR #9596, opened 2020-10-24
Add firefox/privacy/book (#9490, #9520, #9523)

amychurchwell:
## Description
Adds firefox/privacy/book and updates cards on firefox/more

## Issue / Bugzilla link
#9490
#9520
#9523

## Testing
[Design](https://github.com/mozilla/bedrock/issues/9522#issuecomment-710085508) (Design of this branch has been approved.)
[Copy](https://docs.google.com/document/d/16yJDqxFUqLKAgWTQgjHL6ZE6DaqEwisqLl3ZvQryF_c/edit?usp=sharing)
[Demo](https://www-demo-pr-9596.herokuapp.com/en-US/firefox/privacy/book/)


amychurchwell:
@rraue in the [design issue](https://github.com/mozilla/bedrock/issues/9522) you mentioned:

> Has 2 cta - download firefox & download pdf in en & de - in all other locales only download as cta

Where are the pfs? We'll have to host them to have the conditional ctas.

rraue:
@amychurchwell The pdf will not be ready when we push the page, so we need to hide these buttons and activate/add them once the pdfs are ready. I just wanted them to be baked in the design so the direction is clear.

l10n/en/firefox/privacy/book.ftl
```diff
@@ -0,0 +1,130 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### URL: https://www-dev.allizom.org/firefox/privacy/book
+
+# HTML page title
+little-book-of = Little Book of Privacy
```
alexgibson, line 8: Please prefix Fluent IDs with a name space.

amychurchwell:
@rraue once the pdfs are ready we can file an issue to add them to this page. 👍

l10n/en/firefox/privacy/book.ftl
amychurchwell: Is this an ID? It's only used once in the doc. Perhaps I don't understand.

alexgibson: All Fluent string IDs in this file should be namespaced with a unique prefix (as per the docs and in other files), and just like you have done in #9586

For example: 

```
privacy-book-little-book-of = Little Book of Privacy
privacy-book-how-to-stay = How to stay relaxed and be secure online.
```

Right now there's no namespacing, which means it's hard to work out what file a string belongs to in the template, and also increases the likelihood of creating duplicate IDs over time.

amychurchwell:
Adding "do not merge" until demo approval. Could still benefit from a pair of eyes because it is _quite_ a lot of content.

l10n/en/firefox/all.ftl
```diff
@@ -8,6 +8,7 @@ firefox-all-check-the-system-requirements = Check the system requirements
 firefox-all-release-notes = Release notes
 firefox-all-source-code = Source code
 firefox-all-need-help = Need help?
+firefox-all-learn-more-about-firefox = Learn more about Firefox
```
rraue, line 11: Same here, this shouldn't be updated, we decided against the link on /all

bedrock/firefox/templates/firefox/all-unified.html
```diff
@@ -90,8 +90,10 @@ <h1 class="c-intro-heading">{{ ftl('firefox-all-choose-which-firefox') }}</h1>
         </div>
         <div class="c-support-links">
           <ul>
+            <li><a href="{{ url('firefox.more') }}">{{ ftl('firefox-all-learn-more-about-firefox') }}</a>
```
rraue, line 93: I think this was part of my branch but shouldn't be. We decided against the link there in https://github.com/mozilla/bedrock/issues/9610. It think I just accidentally added this to the /more branch.

bedrock/firefox/templates/firefox/more.html
```diff
@@ -16,18 +16,51 @@
 {% block content %}
 <main>
   {% call hero(
-    title=ftl('firefox-products-are'),
+    title=ftl('learn-more-about-firefox'),
```
stephaniehobson, line 19: You can keep the old string as a fallback for l10n if you want. Whether or not that makes sense depends on how much work has been done to localize the page.

l10n/en/firefox/more.ftl
```diff
@@ -22,6 +22,14 @@ get-the-privacy = Get the privacy you deserve. Enhanced Tracking Protection make
 take-your-passwords = Take your passwords everywhere
 securely-access-the = Securely access the passwords you’ve saved in { -brand-name-firefox } from anywhere — even outside of the browser.
 choose-which-firefox = Choose which { -brand-name-firefox-browser } to download in your language
+firefox-chromebook = { -brand-name-firefox-browser } for Chromebook
```
stephaniehobson, line 25: I know it's too late to add the namespace to the existing strings but we could start now with the new ones.

stephaniehobson (changes requested):
This page looks great, especially on XL screens - it switches nicely to mobile too 💯 

Only blocking feedback is an update to the heaving levels. Two other optional things I noted if time/energy allows.

stephaniehobson (approved):

stephaniehobson:
R+ 🍂 

Thanks for adding the fallback.

Issue #9610, opened 2020-10-27
Add link to pillar page on /releasenotes


rraue:
## Description

In https://github.com/mozilla/bedrock/issues/9503 we have build a pillar page for our evergreen content and in order to reduce orphaned pages (we will add more card-links to the pillar page) we need to link the pillar page. Currently our nav system is still evolving and in order to get the page and therefor the linked pages crawled and indexed, we need to set a link to /firefox/more/.

We will set a link on all /releasenotes pages in the footer

alexgibson:
Suggest we use some link text that’s slightly more helpful than “More”?

rraue:
Haha, yes! Any suggestion? Its a wild mix of pages we havent figured out the right information architecture for. I need it indexed and can't find a better link text. Maybe "More about Firefox"? Not super precise but at the end the page provides a lot of links to knowledge around browsers and all builds of firefox.

alexgibson:
Honestly I'm a little confused why we'd add this to /all from a user perspective. I understand we want it indexed, but we should favor users over robots.

Would a page like https://www.mozilla.org/en-US/firefox/more/ not be a good source of additional reading to people landing at https://www.mozilla.org/en-US/firefox/ ?

rraue:
I agree completely. We shouldn't have links for robots only (/locale/ might be the only exception) and I don't think the pages is without value (like a category page in a blog isn't without value and still not heavily used). And /firefox/ was my first page to go to as well, but I don't see currently any Text where it makes sense to link to /more (/faq should be linked from /firefox as well and I cant find a place right now). I thought as wlel about the footer, but isnt our footer global for the whole site?

The reason I choose /all is that its sort of the last resort and since the last update with a super clear UX. So the links in the right are the eternal-last resort if you really cant find on the page what you are looking for. /more will eventually guide you then.

But I am fine with any useful link to the page.

alexgibson:
Ultimately I'm ok with putting this on /all if there's really nowhere else. But it feels like we're just adding links for crawlers now in places where users may not find it useful. Imho the focus of the /all page should be helping people get the build they want / need. Pages like /firefox seem to be more for helping people to make up their mind, where evergreen articles like these may be more helpful. This is just my 2 pence though so feel free to ignore 😅

rraue:
[...]

PR #9613, opened 2020-10-27
Add link in footer to /more

rraue:
Fixes https://github.com/mozilla/bedrock/issues/9610


craigcook:
This is a new string but the template is still in .lang. I think we may need to migrate the whole thing to Fluent before we can change or add any strings. 

One quick hack to get this live ASAP to at least a large portion of users (and robots) would be to make it English-only, wrapped in an `(% if LANG.startswith('en-') %}` condition.

rraue:
@craigcook I thought releasenotes pages are only available in english anyways?!

craigcook:
> @craigcook I thought releasenotes pages are only available in english anyways?!

Ah! Indeed they are. In that case this has no l10n impact after all so no worries.

craigcook (approved):
🎓

PR #9622, opened 2020-10-28
Set up experiment redirects for /firefox/new/

### LINKED ISSUE #9576: Implement an A/B testing redirect solution for www pages
## Description

When we migrate to the new CDN, we'll no longer have the ability to run [edge workers](https://github.com/mozmeao/www-workers) for A/B testing redirects.

We're going to need to come up with an alternate solution. Some possibilities include:

1.) A server side redirect mechanism for A/B testing in bedrock.
2.) A lambda solution using Cloudfront.
3.) A client side solution using JS in bedrock.

I think our preference would likely be 1 or 2, since a client side solution would be least performant.

pmac:
Fix #9576

### DIFF
--- a/bedrock/firefox/views.py
+++ b/bedrock/firefox/views.py
@@ -7,6 +7,7 @@ import hashlib
 import hmac
 import re
 from collections import OrderedDict
+from random import random
 from urllib.parse import urlparse
 
 import basket
@@ -14,6 +15,7 @@ import querystringsafe_base64
 from django.conf import settings
 from django.http import (
     HttpResponsePermanentRedirect,
+    HttpResponseRedirect,
     JsonResponse,
 )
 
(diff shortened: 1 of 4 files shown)


bedrock/firefox/views.py
```diff
@@ -722,6 +725,30 @@ def get(self, *args, **kwargs):
...
+        skip_exp = 'automation' in self.request.GET
+        # only engage the experiment infra if some experiments are set
+        if redirect_percents and not skip_exp:
+            locale = l10n_utils.get_locale(self.request)
+            percent = redirect_percents.get(locale, 0)
+            if percent:
+                percent = percent / 100
+                print(percent)
+                if random() <= percent:
+                    response = HttpResponseRedirect(reverse('exp.firefox.new'))
```
alexgibson, line 743: We probably need to make sure that URL params are preserved in the redirect here (sorry, this didn't cross my mind in our discussion yesterday). When I test locally params don't currently seem to be preserved.

alexgibson (changes requested):
This looks great to me. I noticed one issue with URL params being clobbered in the redirect, but otherwise everything seems to work as expected when I test locally 💯

bedrock/firefox/views.py
pmac: Good catch! I'll update that.

pmac: Updated and added a test for params.

alexgibson (approved):
LGTM, thanks @pmac 💯

PR #9623, opened 2020-10-28
update ftl IDs for firefox/quantum [no bug]

amychurchwell:
## Description

Updates firefox/quantum fluent strings to have a standardized prefix per https://bedrock.readthedocs.io/en/latest/l10n.html#ftl-string-ids


stephaniehobson (approved):
R+ 🌓

PR #9625, opened 2020-10-29
Filter Sentry JS errors to /media/ scripts (#8444) 

alexgibson:
## Description
This should hopefully help cut down on a lot of the noise created by external scripts / extensions.

## Issue / Bugzilla link
#8444

## Testing

I pushed an undefined function error to the home page on demo: https://www-demo1.allizom.org/en-US/

Error in Sentry: https://sentry.prod.mozaws.net/operations/bedrock-demos/issues/9190127/

### DIFF
--- a/media/js/base/sentry-init.js
+++ b/media/js/base/sentry-init.js
@@ -16,12 +16,11 @@
             window.Sentry.init({
                 dsn: sentryDsn ,
                 sampleRate: 0.10,
-                denyUrls: [
-                    // Chrome extensions
-                    /extensions\//i,
-                    /^chrome:\/\//i,
-                    // Firefox extensions
-                    /^resource:\/\//i
+                ignoreErrors: [
+                    'https://plugin.ucads.ucweb.com/api/flow/'
+                ],
+                allowUrls: [
+                    '/media/js/'
                 ]
             });
         }


media/js/base/sentry-init.js
```diff
@@ -16,12 +16,11 @@
...
-                    // Chrome extensions
-                    /extensions\//i,
-                    /^chrome:\/\//i,
-                    // Firefox extensions
-                    /^resource:\/\//i
+                ignoreErrors: [
+                    'https://plugin.ucads.ucweb.com/api/flow/'
+                ],
+                allowUrls: [
+                    '/media/js/'
```
alexgibson, line 23: @pmac I think by only allowing errors from JS on specific URLs it should give us the best option for filtering out unwanted noise (e.g. from extensions and external scripts). The `allowedUrls` array here accepts multiple values and regex. By default it checks that a URL contains the value, so this should match for `https://www.mozilla.org/media/js/`, but also `https://foo.com/media/js/`. We could make it more strict, but maybe this is ok?

Thoughts?

pmac: Oh interesting. So all errors are reported against the script URL and not the page one? In that case this looks great!

pmac (approved):
👻

PR #9638, opened 2020-11-02
Add ?automation=true param to functional tests (Fixes #8917)

### LINKED ISSUE #8917: [cro] Implement a way for functional tests to be excluded from experiments
## Description

I'm not sure what a solution looks like right now, but filing this issue as it's been in the back of my head for too long. When we implement traffic cop experiments, such as in https://github.com/mozilla/bedrock/pull/8901, we often have to skip functional tests for a page because the experiment alters the expected functionality. This is more work for the developer, and also loses us test coverage whilst experiments are running.

We should develop a mechanism by which functional tests are excluded from TC experiments. One possible solution could be for tests to use an `automation=true` URL parameter which traffic cop can look for and exclude. There may well be other options worth exploring too.

---

- [x] Functional tests are excluded from experiments.
- [x] Functionality is documented in A/B testing docs.

alexgibson:
## Description
- Adds `automation=true` param to functional tests so we can exclude those sessions from experimentation (see https://github.com/mozilla/bedrock/pull/9622)

## Issue / Bugzilla link
#8917

## Testing
Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/211599010

### DIFF
--- a/tests/pages/firefox/new/download.py
+++ b/tests/pages/firefox/new/download.py
@@ -10,7 +10,7 @@ from pages.regions.modal import ModalProtocol
 
 class DownloadPage(BasePage):
 
-    URL_TEMPLATE = '/{locale}/firefox/new/{params}'
+    _URL_TEMPLATE = '/{locale}/firefox/new/'
 
     _download_button_locator = (By.CSS_SELECTOR, '#download-button-thanks > .download-link')
     _platforms_modal_link_locator = (By.CLASS_NAME, 'js-platform-modal-button')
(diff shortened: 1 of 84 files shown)


alexgibson:
@pmac updated as per your suggestion yesterday, thanks

tests/pages/regions/join_firefox_form.py
```diff
@@ -26,5 +30,6 @@ def type_email(self, value):
         self.find_element(*self._email_locator).send_keys(value)
 
     def click_continue(self):
+        self.wait.until(lambda s: self.is_enabled)
```
alexgibson, line 33: This fixes a flakey test should a page be slow to load.

tests/pages/firefox/new/download.py
```diff
@@ -22,7 +22,10 @@ def is_download_button_displayed(self):
         return self.is_element_displayed(*self._download_button_locator)
 
     def download_firefox(self):
-        self.find_element(*self._download_button_locator).click()
+        link = self.find_element(*self._download_button_locator)
+        href = link.get_attribute('href')
+        self.set_attribute(link, att_name='href', att_value=href + '?automation=true')
```
alexgibson, line 27: It's not ideal having tests modify an in-page link, but for the purpose what we're doing here I think this is likely OK.

pmac: Yeah. Agreed. Can't think of a better way to go about it though.

pmac (commented):
This looks great. Nice work ⭐

pmac (approved):

PR #9641, opened 2020-11-02
Fix run-integration-tests branch test runs

### LINKED ISSUE #9598: run-integration-tests branch is not working
## Description

Looks like this is broken because it relied on the now defunked oregon-b cluster. We should move it to GCP.

pmac:
Requires https://github.com/mozmeao/www-config/pull/324
Fix #9598

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -78,7 +78,7 @@ test:
   only:
     - run-integration-tests
   variables:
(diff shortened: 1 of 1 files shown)


pmac:
Required changes in www-config have merged. Merging this now too. It will fail for a bit until we get the DNS sorted, but after that we should be good to go.

PR #9647, opened 2020-11-02
Fix #9636 - Localze WNP 83 for EN/DE only

### LINKED ISSUE #9636: WNP 83: Remove FTL file and create second de-* template
## Description

Remove FTL file and create German l10n HTML file. Add routes to filter by locale. 

Add a comment to the line, _"[space to save the magic](https://github.com/mozilla/bedrock/blob/master/l10n/en/firefox/whatsnew/whatsnew-fx83.ftl#L13)"_ with `save the good things you find`

maxxcrawford:
## Description

Remove `.ftl` file for WNP 83. Moving to hard-coded EN and DE versions only. All other locales should be redirected to evergreen, localized WNP. 

### TODO

- [x] Add DE String
- [x] Add DE Template Test
- [x] Remove "EN" from EN template

## Issue / Bugzilla link

#9636

## Testing

- Go to http://localhost:8000/firefox/83.0/whatsnew/all/
  - _Note there is no locale in the URL_
-  If English locale set, should see EN version of WNP 83
-  If German (de-*) locale set, should see DE version of WNP 83
-  If neither English or German locale set, should see evergreen WNP


bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx83-en.html
```diff
@@ -25,11 +25,11 @@
   <section class="wnp-content-main">
     <div class="mzp-l-content mzp-t-content-md wnp83-hero-logo">
       {% include 'firefox/whatsnew/includes/fx83/pocket.svg' %}
-      <h1 class="wnp-main-title">{{ ftl('whatsnew83-main-title') }}</h1>
-      <p class="wnp-main-tagline">{{ ftl('whatsnew83-main-intro') }}</p>
+      <h1 class="wnp-main-title">EN Meet Pocket – Our inspired source for news & stories</h1>
```
maxxcrawford, line 28: Remove "EN" text

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx83-de.html
```diff
@@ -0,0 +1,85 @@
...
+{% block site_header %}{% endblock %}
+
+{% block content %}
+<main class="content-wrapper mzp-t-firefox">
+  {% include 'firefox/whatsnew/includes/header.html' %}
+
+  <section class="wnp-content-main">
+    <div class="mzp-l-content mzp-t-content-md wnp83-hero-logo">
+      {% include 'firefox/whatsnew/includes/fx83/pocket.svg' %}
+      <h1 class="wnp-main-title">{DE} Meet Pocket – Our inspired source for news & stories</h1>
```
maxxcrawford, line 28: Add DE strings

bedrock/firefox/tests/test_base.py
```diff
@@ -384,7 +384,7 @@ def test_fx_83_0_0(self, render_mock):
         req.locale = 'en-US'
         self.view(req, version='83.0')
         template = render_mock.call_args[0][1]
-        assert template == ['firefox/whatsnew/whatsnew-fx83.html']
+        assert template == ['firefox/whatsnew/whatsnew-fx83-en.html']
```
maxxcrawford, line 387: Add test confirming DE redirect works as intended

bedrock/firefox/views.py
```diff
@@ -661,6 +663,15 @@ def get_template_names(self):
         return template
 
 
+class WhatsNewGermanView(WhatsnewView):
```
maxxcrawford, line 666: Is this necessary? I used the `WhatsNewChinaView` as a template.

bedrock/firefox/views.py
```diff
@@ -616,8 +616,10 @@ def get_template_names(self):
                 template = 'firefox/developer/whatsnew.html'
             else:
                 template = 'firefox/whatsnew/index.html'
-        elif version.startswith('83.') and ftl_file_is_active('firefox/whatsnew/whatsnew-fx83'):
-            template = 'firefox/whatsnew/whatsnew-fx83.html'
+        elif version.startswith('83.') and locale.startswith('de'):
```
craigcook, line 619: I think instead of `startswith` this should be an equal, since we only have one German: `locale == 'de'`

craigcook (changes requested):
Just a few minor bits of cleanup. The separate view isn't needed (and isn't being used here, China was a special case). And if we're not using Fluent for this we should also delete the file `l10n/en/firefox/whatsnew/whatsnew-fx83.ftl` and remove the reference from ftl_files_map at https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/views.py#L563

bedrock/firefox/views.py
```diff
@@ -560,7 +560,6 @@ class WhatsnewView(L10nTemplateView):
         'firefox/whatsnew/whatsnew-fx80.html': ['firefox/whatsnew/whatsnew-fx80', 'firefox/whatsnew/whatsnew'],
         'firefox/whatsnew/whatsnew-fx81.html': ['firefox/whatsnew/whatsnew-fx81', 'firefox/whatsnew/whatsnew'],
         'firefox/whatsnew/whatsnew-fx82.html': ['firefox/whatsnew/whatsnew-fx80', 'firefox/whatsnew/whatsnew'],
-        'firefox/whatsnew/whatsnew-fx83.html': ['firefox/whatsnew/whatsnew-fx83', 'firefox/whatsnew/whatsnew'],
```
craigcook, line 563: Doh, looks like we do need `firefox/whatsnew/whatsnew` for the up-to-date notification, and will need it for both the English and German templates.

image

craigcook (changes requested):
Sorry, one last thing!

bedrock/firefox/views.py
maxxcrawford: Ahhhh! Great catch.

alexgibson: Hmm... won't this mess with locale activation? (first file in the array is used to activate locales). Perhaps worth testing using `Dev=False` in your `.env`.

maxxcrawford: I added back the `firefox/whatsnew/whatsnew` FTL file for the template.

Also, I need to set them for both `whatsnew-fx83-en.html` and `whatsnew-fx83-de.html`.

craigcook: Working for me in local prod mode:

image

alexgibson: Right, the copy shows ok - but the page says it's available in many more languages (dictated by `firefox/whatsnew/whatsnew.ftl`):

image

alexgibson: As @craigcook pointed out in Slack, looks like this is OK due to the `en` conditional in the view :)

craigcook (approved):
🌳

Issue #9644, opened 2020-11-02
Type error on /firefox/new/ in older Edge browsers


alexgibson:
## Description

Some older versions of Edge (e.g. v15) seem to hit an error on /firefox/new/

```
Object doesn't support property or method 'forEach'
```

https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/9874825/events/

https://github.com/mozilla/bedrock/blob/master/media/js/firefox/new/desktop/download.js#L160

Suprisingly it seems to support `IntersectionObserver`, but not `forEach` on Node Lists.

dougmellon:
Do you mind if I take a shot at this?

dougmellon:
Currently working on this.

PR #9648, opened 2020-11-02
Convert NodeList to array for compatibility

dougmellon:
## Description
The forEach() method of the NodeList interface calls the callback given in parameter once for each value pair in the list, in insertion order. While more modern browsers support forEach on a NodeList, older ones only allow it on arrays. To solve this, I converted the original NodeList to an array than iterated through that new array.

## Issue / Bugzilla link
[Type error on /firefox/new/ in older Edge browsers #9644](https://github.com/mozilla/bedrock/issues/9644)

## Testing
fix bug #9644

### DIFF
--- a/media/js/firefox/new/desktop/download.js
+++ b/media/js/firefox/new/desktop/download.js
@@ -133,12 +133,12 @@
 
     var supportsInsersectionObserver = function() {
         return 'IntersectionObserver' in window &&
-               'IntersectionObserverEntry' in window &&
-               'intersectionRatio' in window.IntersectionObserverEntry.prototype;
+            'IntersectionObserverEntry' in window &&
+            'intersectionRatio' in window.IntersectionObserverEntry.prototype;
     }();
 
     // check for support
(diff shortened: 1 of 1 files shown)


dougmellon:
Second ever pull request so if there are any issues, please feel free to let me know and I will correct them.

alexgibson (changes requested):
Thanks for the pull request!

The code here looks good, but testing in Edge it doesn't look like the animations actually work there. I think the best course of action is probably to opt older Edge browsers out of this code path altogether on line 141, instead of the changes you've made here:

```javascript
// check for support
if(supportsInsersectionObserver && window.NodeList && NodeList.prototype.forEach) {

}
```

dougmellon:
@alexgibson I believe I made the correct changes. Please let me know if there are any further issues with it and I will gladly adjust them.

alexgibson:
Merged in https://github.com/mozilla/bedrock/commit/5a7793bfe8f6dab828acef1daa5832179015534b, thanks @dougmellon! 💯

Issue #9302, opened 2020-08-10
[8/27] Banner promo for Fenix release


ejregithub:
Description: The new release, Fenix, for Firefox mobile on Android is currently planned for global availability on the 27th. Promoting that on www.mozilla.org can be achieved by using the banner component on the site and placing that on high trafficked pages.

[Principal Project Doc](https://docs.google.com/document/d/1cmqftlLoUphUD7eGblUyAiapksP6fpy-PFGU-fORAyM/edit)

#### Success Criteria
- [ ] Banner is live on 2020-8-27 on the homepage, /firefox/, and /firefox/mobile/

#### Tasks
- [ ] Copy - Rob
- [ ] Design - Cristina
- [ ] Code - Craig

#### Approvals
- [ ] Accountable - Mei
- [ ] TL - Hoosteeno
- [ ] Brand copy - Natalie
- [ ] Brand design - Clio
- [ ] NA Brand TL - Eddie

hoosteeno:
@justinkropp can you attach designs to this when done?

justinkropp:
This is the most recent version of the Fenix banner as it'll show up on desktop.

![Fenix Desktop Banner 1312](https://user-images.githubusercontent.com/1461798/91354759-30b09180-e7bb-11ea-8c64-116fa1820cab.png)

I'm going to check with Mei to double-check that the copy is final and approved.

hoosteeno:
@justinkropp Looks nice to me. Can you specify the interaction on this, too? Is it dismissable (and for how long)? What should happen on click?

So far we've planned to land it on the homepage, the /firefox/ page, and the /firefox/mobile/ page. Do you think we should consider anywhere else?

justinkropp:
@hoosteeno Yes. 

- This should be dismissable.
- If the user is on desktop they should see the CTA above ("Get it on your phone"), which should trigger the current send to device modal
- If the user is on mobile, I'd like to show the respective app store buttons  (App Store and Google Play), both of which would open the appropriate app store experience
- Can we also land this on /firefox/products/ ?

> Is it dismissable (and for how long)?
Do you mean how long until the user sees the banner again after dismissing it?

----

By the way, copy is approved.

hoosteeno:
> Do you mean how long until the user sees the banner again after dismissing it?

We set a cookie that can expire. Should it expire? 

> If the user is on desktop they should see the CTA above ("Get it on your phone"), which should trigger the current send to device modal

[...]

PR #9649, opened 2020-11-03
Remove Daylight launch banner [#9302]

craigcook:
## Description
Cleans up the banner that was added in https://github.com/mozilla/bedrock/pull/9336 and is no longer needed.

## Issue / Bugzilla link
#9302 

## Testing
http://localhost:8000/
http://localhost:8000/firefox/mobile/


maxxcrawford (changes requested):
Should the [`l10n/en/banners/firefox-daylight-launch.ftl`](https://github.com/craigcook/bedrock/blob/9302-remove-banner/l10n/en/banners/firefox-daylight-launch.ftl) file be removed?

craigcook:
So I think when a Fluent file is removed from bedrock the automation will open a PR to remove all the translations from www-l10n. I guess we'll find out.

maxxcrawford (approved):

Issue #9286, opened 2020-08-06
Design singular navigation


hoosteeno:
We will collapse our two navigation systems into a single global navigation and footer under the Mozilla brand. This card is about the design work required to specify that navigation.

Here are some principles that I hope we can deliver in a new navigation design:

1. It helps visitors (users & search robots) know where they are, discover what to do next, and move easily to the most important content on our site:
* Browsers and supplementary information about browsers
* Other products, in order of marketing priority (VPN, Pocket, FB Container, Monitor, etc.)
* Information about Mozilla, most importantly information that effectively communicates our excellence
* Developer and innovation content, mostly links to external sites
2. It anticipates the certainty that we will add new products and content types, and remove products and content types, without requiring a full redesign.
3. It is consistent across the entire website.

We should produce a few options with a strong recommendation and attach to this issue.

## Success ##
- [ ] **Approved** designs attached to this issue

## Approvals ##
- [ ] Marketing executive 
- [ ] Websites team lead
- [ ] Websites dev
- [ ] Websites design

alexgibson:
We should consider the footer here also as the content is complimentary to the navigation.

hoosteeno:
Assigning to @justinkropp

justinkropp:
@hoosteeno @rraue Here's a [working draft](https://docs.google.com/document/d/16F9BZjWmnsNSR8xpmZ5tKlLrN3rc87lVUywB_fCBRhU/edit?usp=sharing) of the navigation proposal (both primary and footer navigation systems). Attaching here for reference.

justinkropp:
Following up on the working draft above, attached is a visual skeleton of the proposed navigation.
----
![fx+moz nav](https://user-images.githubusercontent.com/1461798/91904704-5cc48a80-ec73-11ea-860c-3d44d9bf57de.png)

hoosteeno:
Thanks @justinkropp ! Some questions here for you: 

* Are the top-level items clickable? I think @rraue wants them to be.
* The spec should probably note a couple things:
** When content doesn't exist yet (e.g. product promise)
** When the text of the thing in the nav is still TBD (e.g. Firefox Blog)

And some questions for @kiiirby:

* Can you confirm the contents & order of items in Browsers/Products section, sometime in the next few days? Don't over rotate -- we can change this. Getting it live with a pretty good set is the goal. And remember: A spot in nav is *free traffic* to the thing, so if you're on the fence, leaving it in is probably fine.

[...]

PR #9650, opened 2020-11-03
Redesign global nav and footer, add subnavs [fix #9466] [skip l10n]

### LINKED ISSUE #9466: Code singular navigation
Let's code up the singular navigation that was designed in https://github.com/mozilla/bedrock/issues/9286. 

---
#### Success
- [x] Code is launched on demo and approved by UX
- [ ] Nav is approved by marketing leads
- [ ] Merged, pushed, in l10n

craigcook:
## Description
Overhauls and consolidates the main navigation and footer (no more separate ones for Mozilla and Firefox). This is also a redesign, though keeps much of the original markup structure so the changes are mainly visual.

This also adds subnavs to a bunch of pages, and updates the subnav styling and behavior. Subnav now collapses on small screens, so it requires a title to present a clickable element to toggle the menu.

There are oodles of new strings but I'll submit those changes to www-l10n directly for review by Peiying.

## Issue / Bugzilla link
#9466 (code)
#9286 (design)

## Testing
https://www-demo1.allizom.org/

This is a pretty big PR since it touches lots of individual pages to add subnavs, but moreover it impacts the entire site because it's a global element on every page. It should be carefully reviewed and thoroughly tested.


alexgibson:
Unless this is a migration from .lang, you probably want to remove the `[skip l10n]` part of the commit message here, else the string changes won't reach Pontoon when this merges.

craigcook:
> Unless this is a migration from .lang, you probably want to remove the `[skip l10n]` part of the commit message here, else the string changes won't reach Pontoon when this merges.

I'm planning to submit strings directly to www-l10n once they're final, and hopefully even get a headstart on l10n before this gets merged to bedrock (working in it today).

craigcook:
Some stakeholders will be looking this over in the next few days so it may be subject to change, marking do-not-merge for the moment but overall it's ready for code review. Strings have been sent to Pontoon in https://github.com/mozilla-l10n/www-l10n/pull/128

alexgibson (changes requested):
I took a first pass using modern browsers and things look great, mostly just nits and tidy ups.

I'll take a second pass on some older browsers next.

alexgibson:
We currently have some fallback CS & JS for older IE browsers that's specific to the nav, but it looks like that may need updating to work with the new nav:

image

media/js/base/protocol/protocol-menu.js
```diff
@@ -0,0 +1,341 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// Copied from Protocol, to be backported along with nav updates.
```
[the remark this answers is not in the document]
craigcook: Yeah I'm not happy with the duplication. The only significant changes are the names, since I removed the `mzp-` prefix to avoid clashing with Protocol. The plan is that once I can backport the redesign into Protocol these duplicates can be removed, hence the comment.

BUT we could also keep the `mzp-` prefixes in the updated nav and the Protocol JS should all still work without duplicating it. But then we'll need to be careful that there aren't other name collisions or that we don't make a bigger mess. I also thought this might simplify a future update when we get everything into Protocol. It seemed like duplicating was the lesser crime but I can be convinced otherwise.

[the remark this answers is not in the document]
alexgibson: I think as long as we're not comitted to maintaining this code in bedrock long term (vs in protocol) then that's fine. Mostyly I just wasn't sure how much of this needs to be re-reviewed, but if it's just changin selectors then I think that's fine 👍

alexgibson:
From a design point of view, the only real worry I have is menu overflow for locales that may have longer strings.

image

Might it be worth waiting a little for some locales to complete their translations, so we can test some more real-world examples?

craigcook:
I believe I've fixed the IE style and behavior. It's ugly, but usable. 

/firefox/new/ gets some special treatment since that page needs to still look decent even in oldIE, but with the way it was set up there were all kinds of weird problems with the nav, like funky cascade issues that don't happen with the global IE nav styles. In the end the simpler fix for /new seemed to be removing some of the special treatment and letting it get the global styling instead. But I want to make sure I haven't done anything that seriously breaks that page, so it should get some thorough QA (seems to work OK for me in IE 7/8/9)

alexgibson:
> /firefox/new/ gets some special treatment since that page needs to still look decent even in oldIE, but with the way it was set up there were all kinds of weird problems with the nav, like funky cascade issues that don't happen with the global IE nav styles. In the end the simpler fix for /new seemed to be removing some of the special treatment and letting it get the global styling instead. But I want to make sure I haven't done anything that seriously breaks that page, so it should get some thorough QA (seems to work OK for me in IE 7/8/9)

I gave this some testing on older IEs and things generally seem to work OK, although I didn't get the hamburger menu in IE 6 & 7. However, allowing both the /new and /thanks pages get a mixture of default global IE styles aas well as page specific CSS/ JS does make me a little nervous - it feels like there's a risk of introducing some other subtle bugs / issues here maybe? It would certainly require some more careful testing.

Might another option be to make it so old IE's simply don't get the fancy nav on these pages, or instead an alternative simple nav that's just a handful of top level links? This might achieve the main aim of navigating to at least our download pages, and also be simpler to maintain long term. Thoughts?

alexgibson:
Noticed a small bug in German where a link title overflows the container:

image

alexgibson:
Greek seems to suffer from the panel overflow :(

image

craigcook:
We can change the rule to reverse the menu offset on the third item, but then there's a chance it will overflow to the left. That even seems more likely for more locales that have short strings.

image

For the overflowing titles, I'm not sure what's the best fix there apart from hiding the overflow. Is truncated text better than overflowing text?

image

Another, more drastic option would be to unset the grid layout at this awkward, narrow-but-not-yet-mobile breakpoint so the dropdown items stack. And in that case we could even make the panel a bit narrower and probably avoid problem of the panel overflowing the left side of the screen, since that's most likely to happen at this same awkward breakpoint.

craigcook:
OK, I'm stumped. The big problem in IE8 seems to be the inline SVG messing with the rest of the DOM. This is only on /new, though there are lesser rendering glitches on other pages.

image

So I can put the SVGs inside conditional comments and that fixes IE8, but then for some reason they're also hidden from IE9/10/11, even when the comment is explicitly for IE8 or lower (`<!--[if lt IE 9]>`). Trying to hide them with CSS doesn't do anything, the issue is the SVG messing up the DOM. But this doesn't seem to happen with the old nav. I feel like I'm missing some trick, or there's some other obscure glitch happening.

I'm running out of ideas for a fix other than ditching the inline SVG and embedding them all as ``.

alexgibson:
How about my suggestion earlier, where we give old IE’s a simpler, alternative nav that’s slimmed down, and other browsers get the full navigation?

alexgibson:
> I'm running out of ideas for a fix other than ditching the inline SVG and embedding them all as .

This is also another option, we could add `loading=lazy` to avoid making the image requests on page load.

craigcook:
To solve for the dropdown overflowing the viewport it's now a single column on mid-size screens and doesn't switch to the grid until we hit `mq-lg`.

image

However, that made a new problem: When the window is tall enough to get the sticky nav (> 600px) but still pretty short, the long menu can run off the bottom and you can't scroll to it because of the stickiness. So to compensate for that I'm hiding descriptions to shorten the menu in that in-between range when the window is tall enough for a sticky nav but not tall enough to fit the whole menu with descriptions. It's fiddly but I think it works.

alexgibson:
> However, that made a new problem: When the window is tall enough to get the sticky nav (> 600px) but still pretty short, the long menu can run off the bottom and you can't scroll to it because of the stickiness. So to compensate for that I'm hiding descriptions to shorten the menu in that in-between range when the window is tall enough for a sticky nav but not tall enough to fit the whole menu with descriptions. It's fiddly but I think it works.

I think this is a great solution 👍

media/css/protocol/components/_menu-item.scss
```diff
@@ -99,58 +108,69 @@ $icon-size-md: 32px;
...
+        .c-menu-item-desc {
+            display: block;
+            margin-top: $spacing-sm;
+        }
+    }
+}
+
+// Hide descriptions to shorten the menu when nav is sticky
+// Otherwise long menus can't be scrolled
+@media #{$mq-md} and #{$mq-tall} {
```
alexgibson, line 151: `mq-tall` equates to `min-heght: 600px`, so if I have a shorter screen than this value, I still get the overflowing descriptions. Would it be better to use a `max-hight` value here instead?

bedrock/firefox/templates/firefox/new/desktop/download.html
```diff
@@ -21,6 +21,7 @@
   <![endif]-->
 
   <!--[if lt IE 9]>
+    {{ css_bundle('common-old-ie') }}
```
alexgibson, line 24: Is this needed still?

alexgibson (changes requested):
One nit / questions, but otherwise I think this is looking pretty much ready? r+wc

media/css/protocol/components/_menu-item.scss
craigcook: At >600px the nav isn't sticky so you can scroll even if it overflows. Kropp and I felt the descriptions are valuable enough content that we should keep them visible whenever possible. It's really the combination of narrow window + tall (hence sticky nav) that makes trouble.

But thinking about it more now, even if it's scrollable that doesn't mean it's easy to scroll while the menu is open (I'm a habitual scroll-wheel users but not everyone is). Once you move your mouse the menu closes and you can't get it back without scrolling back to the top. So for some small number of people who scroll in odd ways and have viewports just the right size, they may never be able to reach the bottom of the dropdown... so yeah, maybe this should be `$mq-short`.

bedrock/firefox/templates/firefox/new/desktop/download.html
craigcook: This page doesn't get the common IE bundle so without this the nav breaks in IE8.

alexgibson: OK, so just to make sure I'm getting this right: We serve IE8 1st class page CSS, but the nav styling doesn't work well there any more, so we need to serve it the common-old-ie styles as well to compensate?

alexgibson (approved):
r+ to merge whenever you're ready, great work! 💯

Issue #8938, opened 2020-05-20
Add Firefox for iOS and Android to the /firefox/all/ page


alexgibson:
## Description

Firefox for iOS is noticably absent from the /firefox/all/ page. The technical reason for this is because Apple only allows users to download from the App Store, but from a user perspective I'd still expect to find Firefox for iOS mentioned somewhere on this page as an option:

https://www.mozilla.org/en-US/firefox/all/

We should find a simple way to add iOS as a download option, even if it just links to the App Store.

For Android, now that [Fenix](https://github.com/mozilla-mobile/fenix) is Firefox for Android, we don't have a good way to link to the .apk files ([see bug 1663735](https://bugzilla.mozilla.org/show_bug.cgi?id=1663735)): the releases section of the GitHub repo isn't automatically updated and the builds aren't being uploaded so that bouncer can redirect to them.

The proposal is that we remove Android from the download selection form on /firefox/all/ and we add the App Store and Play Store buttons to the bottom of the left-hand column on the page so that people still have options for downloading our mobile apps.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [x] Firefox for iOS is presented as an option for download on /all
- [x] Firefox for Android (Fenix) is presented as an option for download on /all

alexgibson:
> The proposal is that we remove Android from the download selection form on /firefox/all/ and we add the App Store and Play Store buttons to the bottom of the left-hand column on the page so that people still have options for downloading our mobile apps.

I've been thinking about the UX here and even though it may be easier to remove Android from the product menu technically, I'm not sure that makes for the right user experience. Keeping all products in one menu feels clearer to me (including adding iOS ion there). We could point the download button to the relevant store and keep the UI consistent, or alternatively we could replace the button with a store badge.

hoosteeno:
Needs some UX work before it's ready for code -- @justinkropp this is one to pick up when you have bandwidth.

justinkropp:
I've though about both approaches as outline above, and I'm more in alignment with Alex's suggestion that we keep Android in the product menu, add in iOS, and point the download button to the appropriate location. 

As an alternative addition, could we also include the App Store and Google Play buttons on the page (but separate from the product menu) where appropriate?

hoosteeno:
> As an alternative addition, could we also include the App Store and Google Play buttons on the page (but separate from the product menu) where appropriate?

I think we will have to rework the page a bit anyway to accommodate these new products:

[...]

PR #9652, opened 2020-11-03
Display updates to /all (#8938)

stephaniehobson:
## Description

- change download info to a table
- right align button

## Issue / Bugzilla link

See #8938

## Testing

[Design specs](https://github.com/mozilla/bedrock/issues/8938#issuecomment-700214409)
[Link to localhost](http://localhost:8000/en-US/firefox/all/)

### DIFF
--- a/media/css/firefox/all/all-unified.scss
+++ b/media/css/firefox/all/all-unified.scss
@@ -6,6 +6,7 @@ $font-path: '/media/fonts';
 $image-path: '/media/protocol/img';
 
 @import '../../../protocol/css/includes/lib';
+@import '../../../protocol/css/includes/forms/lib';
 @import '../../../protocol/css/templates/main-with-sidebar';
 @import '../../../protocol/css/components/modal';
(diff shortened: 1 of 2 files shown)


media/css/firefox/all/all-unified.scss
```diff
@@ -178,7 +183,31 @@ $image-path: '/media/protocol/img';
...
+    width: 1%;
+    white-space: nowrap;
+}
+
+.c-download-info-content {
+    font-weight: bold;
+}
+
 .c-download {
+    text-align: right;
```
alexgibson, line 209: Might need a bidi mixin for RTL locales here?

alexgibson (changes requested):
One suggestion for better RTL support, but otherwise r+wc

media/css/firefox/all/all-unified.scss
stephaniehobson: Updated.

alexgibson (approved):

PR #9656, opened 2020-11-04
Fix docker image tagging

pmac:
Raphael discovered that the sitemap was behind. I found that this is because the `prod-latest` and git-tag-based docker tags weren't being pushed to Docker Hub. This is because we moved to promoting images instead of building fresh for stage and prod. Because of this the docker hub push was being skipped for stage and prod. This change ensures that the tagging and pushing happens regardless.


pmac:
cc @rraue

pmac:
For now I've manually tagged the latest prod image on docker hub so that the next run of sitemap generator will do the right thing. This PR should fix it moving forward.

duallain (approved):

PR #9663, opened 2020-11-05
Updating the site with our new PGP key

dveditz:
Update the /security pages to reflect our new PGP, replacing the expired one.

## Description
The PGP key used by the security team to receive confidential exploit reports expired in October 2020 and we have generated a new one. The site needs to be updated so it's showing the new public key block and not the expired 2017 one.

## Issue / Bugzilla link
None, just updating data.

## Testing
- import the public key and see it's for security@mozilla.org
- verify that transition.txt is correctly signed by the old and new key as claimed in the text


dveditz:
@pmac Can you please review and merge this update of our PGP key? If I'm doing it wrong let me know.

pmac (approved):
Looks perfect. I verified the key and signatures. Thanks!

PR #9668, opened 2020-11-05
move sitemaps data to db

pmac:
- Add models and views for sitemaps
- Convert sitemaps URLs to use views and data from db

This is done in 2 commits. Once this merges we must deploy the first commit, and then the 2nd one separately and only after the sitemaps data has had a chance to make it into the database.


bedrock/sitemaps/models.py
```diff
@@ -0,0 +1,68 @@
...
+
+    return sitemap, etags
+
+
+def get_sitemap_objs():
+    objs = []
+    sitemap, etags = load_sitemaps_data()
+    for url, locales in sitemap.items():
+        if not locales:
+            locales = [""]
```
glogiotatidis, line 24: nit: style consistency `s/"/'/`until you install `black` in bedrock :D

bedrock/sitemaps/views.py
```diff
@@ -1 +1,33 @@
...
+
+from bedrock.sitemaps.models import SitemapURL
+
+
+class SitemapView(TemplateView):
+    content_type = 'text/xml'
+
+    def _get_locale(self):
+        if self.kwargs['is_none']:
+            locale = ''
```
glogiotatidis, line 11: Hmm shouldn't this be `locale = None` to work with line 28?

If so, I wonder why yours tests pass?

glogiotatidis (commented):
this looks good, except a question about your locale check.

bedrock/sitemaps/views.py
pmac: the naming is confusing here. In the DB I store the empty string for URLs that don't have a locale. The `is_none` kwarg comes from the URL because the file that is for URLs without a locale is `/sitemap_none.xml`. I set it to `None` when the view is for the sitemap index, which is when it's not for the none file, and it's not for one of the files for a locale.

bedrock/sitemaps/models.py
pmac: lol... true. I copied a lot of this code from the www-sitemap-generator repo, which is using black :)

bedrock/sitemaps/views.py
pmac: I've added some more comments to hopefully make things more clear

glogiotatidis: Got it! Still confusing that locale can be empty string but I get it. Thanks for the comments

glogiotatidis (approved):

bedrock/sitemaps/views.py
pmac: Do you think it'd be better if it was some other specific value? I had it be "none" for a bit but thought that was potentially even more confusing. I could have it be `null` in the DB. Just not sure how best to handle the URLs that don't have a locale. The ones I mean are here: https://www.mozilla.org/sitemap_none.xml e.g. https://www.mozilla.org/keymaster/gatekeeper/there.is.only.xul

pmac: I've updated things to use a special string and constant to indicate a URL has no locale. I think this makes the code clearer.

glogiotatidis:
> I've updated things to use a special string and constant to indicate a URL has no locale. I think this makes the code clearer.

I had a review comment that read
 > I guess it's fine the way it is, as long as one knows that some DB entries have empty locale. That's what got me confused.

but I didn't click `submit review`. 

Either way your changes makes things clearer. You rock :metal:

PR #9674, opened 2020-11-06
[no bug] rename new ftl IDs on firefox/more

amychurchwell:
## Description
New fluent ID's in firefox/more are named [following our convention](https://bedrock.readthedocs.io/en/latest/l10n.html#ftl-string-ids)

## Issue / Bugzilla link
https://github.com/mozilla-l10n/www-l10n/pull/127


craigcook:
I'm not sure we can change IDs since this file is already in Pontoon and translated in a bunch of languages. It'll effectively treat these as new strings and break current translations, unless we can run some kind of migration process. I would defer to Alex or pmac on how that might work.

I think we may need to add the new IDs as new strings but keep the old ones with the old IDs, and mark them as outdated. Then set the old IDs as fallbacks in the template.

craigcook:
Oh nevermind, I think these are already new and not yet in Pontoon, my mistake!

craigcook (approved):
🧶

Issue #9620, opened 2020-10-28
Stub Attribution error for Firefox users on Windows


alexgibson:
## Description

Another frequent error we're seeing in Sentry is in the stub attribution script:

```
TypeError: window._SearchParams is not a constructor
  at u.getAttributionData(/media/js/BUNDLES/stub-attribution.d27a6ffd686b.js:1:2303)
  at u.init(/media/js/BUNDLES/stub-attribution.d27a6ffd686b.js:1:3083)
  at ? (/media/js/BUNDLES/stub-attribution.d27a6ffd686b.js:1:3246)
  at ? (/media/js/BUNDLES/stub-attribution.d27a6ffd686b.js:1:3253)
```

https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/9874302/

### Steps to reproduce

Unclear right now. It seems to happen only to Firefox users. Requires more research.

### Expected result

Either the attribution script should function as expected, or it should fail gracefully.

### Actual result

An error is thrown, preventing attribution data from being collected.

### Environment

Latest Firefox on Windows.

alexgibson:
See also: https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/9874328/events/

alexgibson:
I've been trying to figure out the cause of this error and the one thing that seems consistent is this happens in nearly all cases to Firefox users when viewing in-product URLs such as /whatsnew, /welcome, and /privacy:

https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/9874302/events/

Geography seems to suggest this could be an error caused by scripts timing out (possibly on slow connections), as opposed to something like a specific syntax error.

We could add a check to the stub attribution script to make sure that `window._SearchParams` is defined before trying to query data, but this would only really fix the error, and not the root cause. It may be better if we just ship less JS to users under these circumstances.

@hoosteeno do you think there's merit in continuing to include stub attribution JS for existing Firefox users when viewing in-product pages? It seems like offering them a Firefox download is probably an unlikely thing we would do there.

hoosteeno:
If sessions starting on the whatsnew page leave the whatsnew page and, in the same session, go to a page that's part of a download funnel, will they get stub attribution code? In other words, would the impact of removing this code from whatsnew be that we never attribute to whatsnew, but only for pages that have a download button on them?

That seems OK.

alexgibson:
[...]

PR #9679, opened 2020-11-10
Remove Stub Attribution script from in-product Firefox pages (#9620)

alexgibson:
## Description
- Removes Stub Attribution script from in-product pages such as /whatsnew, /welcome, and /privacy/firefox/ (those users are already on Firefox, and if they should go on to download attribution should still happen on the download page)
- Adds a check for `window._SearchParams` in stub attribution script to fail more gracefully when network errors happen.

## Issue / Bugzilla link
#9620

## Testing
- [ ] Stub attribution script should not be loaded for in-product page templates.


stephaniehobson (approved):
R+ 👣

PR #9685, opened 2020-11-12
Enable multiprocess mode for prometheus metrics gathering

pmac:
@duallain I think this works. My mistake before in 89a6b9c8bd9df8272371a5dd5114da7129c8434c was that I added the env var in the Docker image which meant it tried to use it all the time and some of the management commands and the dev server didn't play nice with it. I tested locally by running `make run-prod` and looking at `/prometheus/metrics/` as well as getting a shell on the running container `docker-compose exec release-local bash` and looking at the `/tmp/prometheus_metrics` dir.

### DIFF
--- a/bin/run-prod.sh
+++ b/bin/run-prod.sh
@@ -1,10 +1,15 @@
 #!/bin/bash -xe
 
+export prometheus_multiproc_dir=/tmp/prometheus_metrics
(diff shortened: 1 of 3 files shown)


glogiotatidis (approved):
Looks good. Interested to see who that works. :100:

duallain (approved):

Issue #9321, opened 2020-08-24
Break up common.js bundle


alexgibson:
## Description

Now that we've removed the legacy Sandstone and Pebbles frameworks, we're finally down to a single `common.js` bundle that gets shared among all pages:

This bundle encompasses several different things currently, and dependencies can start to become unclear as it begins to grow in size. A single error can prevent the rest of the script from executing, so for example an analytics JS error in `core-datalayer.js` can halt everything. This seems brittle to me, especially if user facing things come afterward? Now that we have HTTP/2, we can also lean into separating things out more semantically without as much performance impact.

We could consider splitting the common bundle up, into something like the following:

- `lib.js`
  - `polyfills.js`
  - `libraries.js`
  - `utils.js`
- `data.js` (analytics, metrics)
- `ui.js` (page / UI related scripts)

alexgibson:
Looking at our most common JS errors (https://sentry.prod.mozaws.net/operations/bedrock-prod/), I think there's likely some data here that could back up making some changes to the way we bundle our common JS files.

For example:

- User is on an older browser -> browser downloads and parses `common.js` -> An exception occurs somehwere during parsing.
- User then downloads page specific JS bundle -> page specific JS tries to use a common library, such as `Mozilla.UITour`.
- Page then throws an error because `Mozilla.UITour is undefined`.

PR #9687, opened 2020-11-12
Fix JS error when join-firefox-content not found (Fix #9688)

### LINKED ISSUE #9688: JS error on /new in Firefox < 6.0
```
o is undefined
```

https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/9874754/

stephaniehobson:
## Description

- This content is only displayed to Firefox browses
- The error is only appearing in Firefox < 6.0

## Issue / Bugzilla link

Fix #9687

## Testing

- The error can be replicated by deleting the HTML element from the page

### DIFF
--- a/media/js/firefox/new/desktop/join-modal.js
+++ b/media/js/firefox/new/desktop/join-modal.js
@@ -42,8 +42,8 @@
     };
 
     // Only Firefox desktop 57+ (post-quantum) should get the FxA modal
-    if (client.isFirefoxDesktop && client._getFirefoxMajorVersion() >= '57') {
-        initFxAccountModal();
+    if (joinFirefoxContent && client.isFirefoxDesktop && client._getFirefoxMajorVersion() >= '57') {
+        window.Mozilla.run(initFxAccountModal);
     }
 
 })();


alexgibson:
Rather than try to add conditionals for specific lines of code, could we maybe use `Mozilla.run()` here so this script only applies to modern browsers? https://bedrock.readthedocs.io/en/latest/browser-support.html#feature-detection-using-javascript

stephaniehobson:
Added Mozilla.run(). Kept the check that the content element exists on the page since the error could arise in any modern browser if the content is not on the page.

alexgibson:
Ok, I actually think the modal here might be a red herring, since there's already logic that says this code should only run in FF 57 and above in the conditional?

When I run the page locally in Firefox 5 via BrowserStack, I see the following errors:

```
[05:02:50.405] client is undefined @ http://localhost:8000/media/js/BUNDLES/firefox_new_desktop_join_modal.js:205
[05:02:50.401] uncaught exception: [Exception... "Not enough arguments"  nsresult: "0x80570001 (NS_ERROR_XPC_NOT_ENOUGH_ARGS)"  location: "JS frame :: http://localhost:8000/media/js/BUNDLES/common.js :: <TOP_LEVEL> :: line 2"  data: no]
[05:02:50.402] window._SearchParams is not a constructor @ http://localhost:8000/media/js/BUNDLES/stub-attribution.js:207
[05:02:50.403] uncaught exception: [Exception... "Not enough arguments"  nsresult: "0x80570001 (NS_ERROR_XPC_NOT_ENOUGH_ARGS)"  location: "JS frame :: http://localhost:8000/media/js/BUNDLES/firefox_desktop_download.js :: <TOP_LEVEL> :: line 100"  data: no]
```

What I think is happening is there's a JS error happening in both `common.js` and `firefox_desktop_download.js`. The error `NS_ERROR_XPC_NOT_ENOUGH_ARGS` seems to indicate that there are some calls to `addeventListener` missing `false` at the end of the argument. E.g: https://github.com/mozilla/bedrock/blob/master/media/js/firefox/new/desktop/download.js#L100

I'm not sure where the error is in common.js

alexgibson:
This is another reason why I think we should consider https://github.com/mozilla/bedrock/issues/9321, since a JS error anywhere in `common.js` can result in several other unrelated things breaking.

PR #9689, opened 2020-11-13
Migrate firefox/nightly/firstrun to Fluent (Fixes #9684) [skip l10n]

### LINKED ISSUE #9684: Migrate nightly/firstrun page to Fluent
## Description

https://www.mozilla.org/en-US/firefox/nightly/firstrun/

---

#### Success Criteria
- [ ] Page uses Fluent in production along with existing migrations

alexgibson:
## Description
http://localhost:8000/en-US/firefox/nightly/firstrun/

## Issue / Bugzilla link
#9684

## Testing
```
./manage.py l10n_update
```

### DIFF
--- a/bedrock/firefox/redirects.py
+++ b/bedrock/firefox/redirects.py
@@ -37,7 +37,7 @@ redirectpatterns = (
              '/firefox/nightly/firstrun{p}'),
 
     # bug 1275483
-    redirect(r'^firefox/nightly/whatsnew/?', 'firefox.nightly_firstrun'),
+    redirect(r'^firefox/nightly/whatsnew/?', 'firefox.nightly.firstrun'),
 
     # bug 840814
     redirect(r'^projects/firefox'
--- a/bedrock/firefox/templates/firefox/nightly/firstrun.cs.html
+++ b/bedrock/firefox/templates/firefox/nightly/firstrun.cs.html
@@ -1,0 +1,21 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "/firefox/nightly/firstrun.html" %}
+
+{% block locale_promo %}
+  <div class="mzp-l-content mzp-t-content-md locale-specific">
+    <h3>Zjistěte více o místní komunitě Mozilla</h3>
+    <p>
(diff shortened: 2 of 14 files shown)


bedrock/firefox/templates/firefox/nightly-firstrun.html
```diff
@@ -15,85 +15,65 @@
...
       </ul>
     </div>
   </section>
-
-  {% if not request.locale.startswith('en') %}
-    {% if l10n_has_tag('promo_locale') %}
-    {# The following block is entirely defined by locales, and the placeholder in en-US here will never be exposed to production #}
-    <div class="mzp-l-content">
-      <div class="locale-specific">
-        {% l10n promo_locale 20150101 %}
```
alexgibson, line 84: @pmac I don't think we have an equivalent system for this in Fluent, right?

pmac: these were really just localized templates managed in the l10n repo. So we kindof do with l10n templates that we could manage in bedrock if we want/need to keep any of this, it just wouldn't be managed by the l10n team.

alexgibson: @peiying2 is this content you would like to keep? We'd need to manage the content in bedrock when this page transitions to Fluent, which is something I can take care of here, but it would probably be less visible / accessible to the L10n community.

It looks like ~6 locales have custom content: https://github.com/mozilla-l10n/www.mozilla.org/search?p=1&q=promo_locale

alexgibson: OK, I'm going to move this content into bedrock for the time being, just so it's not lost.

pmac:
I looked at the lang files and found that there are indeed several locales that take advantage of the translatable URL for the "start localizing" button. For [bug 1677938](https://bugzilla.mozilla.org/show_bug.cgi?id=1677938) I'd suggest just updating the value in the FTL file.

alexgibson:
> I looked at the lang files and found that there are indeed several locales that take advantage of the translatable URL for the "start localizing" button. For bug 1677938 I'd suggest just updating the value in the FTL file.

Yep, I'll update this, thanks @pmac 👍

alexgibson:
@pmac updated thanks

alexgibson:
I opened a pull request to fix this in the l10n repo as well: https://github.com/mozilla-l10n/www-l10n/pull/135

craigcook (approved):
🌙

PR #9696, opened 2020-11-16
Add image from firefox.com to set proper HSTS settings

### LINKED ISSUE #9695: Add an img tag pointing to firefox.com in mozilla.org template
## Description
A recent HSTS change on firefox.com caused issues for some of our users (https://bugzilla.mozilla.org/show_bug.cgi?id=1666072). The server configuration was fixed and a client-side fix is in nightly to avoid future breakage  but it is too late and risky to uplift it in a release build a few hours before shipping 83.
We need to reset the HSTS value for firefox.com in our users cache and the easiest way is to force a http hit to firefox.com by including an image tag in mozilla.org template so as that Firefox users get their HSTS value for firefox.com fixed on loading the whatsnew page when they upgrade from a previous version.

No real image is needed, we only need to force an http connection to firefox.som so ``  which would return a 200 HTTP code would work.

pmac:
Will apply to firefox.com and all subdmains.

Re bug 1666072
Fix #9695

### DIFF
--- a/bedrock/base/templates/base-protocol.html
+++ b/bedrock/base/templates/base-protocol.html
@@ -142,5 +142,7 @@
       <script src="https://pontoon.mozilla.org/pontoon.js"></script>
     {% endif %}
     <!--<![endif]-->
+    {# Sets the HSTS header for firefox.com. See issue #9695 #}
+    <img src="https://firefox.com/set_hsts.gif" class="hidden">
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


alexgibson (approved):
r+ to merge once https://github.com/mozmeao/www.firefox.com/pull/34/ has merged

pmac:
The www.firefox.com change is merged and deployed to prod and working as expected. Merging this. Thanks Alex!

PR #9697, opened 2020-11-16
Update /firefox/releases/ to include all ESR releases

### LINKED ISSUE #9318: ESR 78 releases not showing up on https://www.mozilla.org/en-US/firefox/releases/
On the page https://www.mozilla.org/en-US/firefox/releases/

If you look at 68, it lists all the ESRs

If you look at 78, it's missing 78.1.

pmac:
The way it used to work assumed that ESR releases followed a standard cadence. This has not always been true and can change. This changes it to include them no matter the cadence.

### Testing

I updated the tests and test data to ensure that ESR 78 is included properly. I also tested by pulling up the current production page, updating my local DB, and running a local server and comparing the pages to ensure that the only updates were to add ESR releases that weren't previously included.

```bash
$ bin/run-db-download.py --force
$ python manage.py runserver 0.0.0.0:8000
```

Fix #9318


bedrock/releasenotes/views.py
```diff
@@ -173,7 +167,8 @@ def releases_index(request, product):
         # number has not been used since Firefox 4, then reintroduced with
         # Firefox ESR 24 (Bug 870540). On this index page, 24.1.x should be
         # fallen under 24.0. This pattern is a tricky part.
-        converter = '%g' if int(major_version) in esr_major_versions else '%s'
+        # the outlier is 33.1 which is in major_releases for some reason.
+        converter = '%g' if int(major_version) >= 24 and release != '33.0' else '%s'
```
glogiotatidis, line 171: I spend a few minutes trying to make this more pythonic but then `33.1` is really annoying. I came up with

```python
converter = f'{major_version:.0f}' if int(major_version) >= 24 and release not in  ['33.0', '33.1'] else f'{major_version:.1f}'
major_pattern = r'^' + re.escape(converter)
```

which is not great either. Anyway what you have works fine, i's just worried that `round` can bite us.

and also TIL that you can do `rf'{major_version}'` and produce a raw string with variable interpolation.

glogiotatidis (approved):
:cake:

bedrock/releasenotes/views.py
alexgibson: I remember [33.1 ](https://www.mozilla.org/en-US/firefox/33.1/releasenotes/)was some kind of special release that was tied to a marketing campaign (we did a fancy /whatsnew page promoting the "Forget button"). Still it's odd to put it in major releases, but I'm guessing this is probably the reason.

alexgibson (approved):
I tested the change locally and it seems to do the correct things. I'll defer to yourself and @glogiotatidis on the Python discussion. r+!

bedrock/releasenotes/views.py
pmac: Yeah. The call to `round` was confusing for me too. I may play with this some more. It does feel convoluted but I'm guessing there were good reasons for it.

pmac: I've updated it. I didn't use the `re.escape` call. What problem was that solving? It seems to be working okay without it.

pmac: Added `re.escape` back in. Of course we need that, the versions have `.` in them 🤦 Good catch.

pmac: I also threw in a fix for `14.0.1` being duplicated.

pmac:
Everything seems good. Merging. Thanks all!

bedrock/releasenotes/views.py
glogiotatidis: lgtm!  :)

Issue #9666, opened 2020-11-05
Migrate CPG pages to Fluent (by end of Nov.)


hoosteeno:
## Description

In #9614 we are changing strings on CPG pages. We need those pages to be migrated to Fluent. We need it all done in November.

---

#### Success Criteria
- [ ] Pages touched by #9614 are migrated to Fluent and those changes are merged/live.

amychurchwell:
Blocked until limited locale config issue is resolved in Pontoon. Peiying is working with Matjaz.
Apparently a fix was put out, but still investigating.  https://bugzilla.mozilla.org/show_bug.cgi?id=1602896

PR #9708, opened 2020-11-18
Migrate /participation/reporting/ to fluent. (#9666) [skip l10n]

amychurchwell:
## Description
Migrates /about/governance/policies/participation/reporting/ to fluent.

## Issue / Bugzilla link
#9666

## Testing
http://localhost:8000/en-US/about/governance/policies/participation/reporting/


amychurchwell:
migration for the existing translations in progress. en is in review.

l10n/configs/pontoon.toml
```diff
@@ -164,6 +164,25 @@ locales = [
 [[paths]]
     reference = "en/mozorg/mission.ftl"
     l10n = "{locale}/mozorg/mission.ftl"
+[[paths]]
```
amychurchwell, line 167: @pmac does this addition look okay to you? The work Alex did in #9735 ended up working & I want to make sure I don't mess it up with the CPG pages. :P

pmac: I think it's not going to work because line 159 above this includes this path, so it will override this IIRC.

pmac: So we'll have to specify a config for each subdomain of `mozorg/about` or something like that :(

amychurchwell: It seems like the only subdirectory in `mozorg/about` for fluent is `governance/policies/` could I replace `en/mozorg/about/**/*.ftl` with what I have?

pmac: Hmm... I just fear that if/when we add more it will not behave properly unless the dev at that point remembers to update the config. Maybe we leave out the more broad config for now to be safe?

amychurchwell: I pushed a commit to delete the broader config. Is that what you meant?

pmac: I think so. I guess it's safer to not include those potential new files in pontoon at all than too broadly. The other option I guess is to just have the subset of locales apply to everything in `mozorg/about` like you suggested above. I'm not sure which is best TBH. What do you think?

amychurchwell: I think this works for now. We can make this broader as we continue porting things over. It is hard to start broad in the beginning, imo.

amychurchwell:
the www-l10n PR for this PR has merged: https://github.com/mozilla-l10n/www-l10n/pull/133 so this is now ready for review.

alexgibson (approved):

PR #9710, opened 2020-11-18
Update the l10n link on the Firefox Nightly first run page

pmac:
Fix [bug 1677938](https://bugzilla.mozilla.org/show_bug.cgi?id=1677938)


alexgibson:
It might make more sense to fix this in https://github.com/mozilla/bedrock/pull/9689 ?

pmac:
Oh! Right. Totally forgot about that one. Do you think it's worth keeping the translatable URL? Are there alternative URLs that exist already?

pmac:
Closing in favor of #9689

PR #9720, opened 2020-11-20
[fix #9719] Set up whatsnew page for Firefox 84.0

### LINKED ISSUE #9719: WNP84: Use WNP82 for WNP84 users. 84 lands 2020-12-15.
## Description

We want to reuse WNP82 for WNP84 users. **84 lands 2020-12-15.**

This is a straightforward effort:

#### Tasks
- [ ] Activate WNP82 template for WNP84 requests
- [ ] Coordinate with l10n team (no l10n should be needed)
- [ ] Coordinate with QA team to validate...
* Target market experience
* Out-of-market experience
* Chinese IP experience

82 Bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=1668938
82 QA issue here: https://jira.mozilla.com/browse/PI-821

---

#### Success Criteria
- [ ] Page is live and available for QA/l10n
- [ ] l10n knows what they need to do (nothing??)
- [ ] QA knows what they need to do

craigcook:
## Description
Updates the whatsnew view to reuse the WNP82 template for Firefox 84.0. There are no other changes, we're just rerunning the exact same page.

However, if we really want to keep each WNP separate and self-contained we could duplicate and rename the 82 template and assets. Reusing them in place means we wouldn't be able to delete them until 84 also cycles out. It incurs a small technical debt and hence the comment to remind future devs (also tests would break), but it also feels kinda silly making exact duplicates just so we can more easily delete some later. If that's the better way I can be persuaded.

## Issue / Bugzilla link
#9719 

## Testing
http://localhost:8000/en-US/firefox/84.0/whatsnew/

It's English only; other locales should fall back to the evergreen page.


craigcook:
We need different UTM params for 84 (but of course, I hadn't thought of that) so while there may be other options I think I'm just going to duplicate the template after all.

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx84.html
```diff
@@ -0,0 +1,74 @@
...
+  {% include 'firefox/whatsnew/includes/header.html' %}
+
+  <section class="wnp-content-main">
+    <div class="mzp-l-content mzp-t-content-md">
+      <img class="wnp-main-image" src="{{ static('img/firefox/whatsnew/whatsnew84/vpn-logo.svg') }}" width="120" alt="">
+      <h1 class="wnp-main-title">Firefox protects your browsing. <strong>Mozilla VPN</strong> protects you everywhere.</h1>
+      <p class="wnp-main-tagline">Mozilla VPN (Virtual Private Network) protects your entire internet connection on your computer, your tablet, and even your phone. For even more security, it comes with a 30-day money back guarantee.</p>
+
+      <div class="wnp-main-cta">
+        <a href="https://vpn.mozilla.org/?utm_source=mozilla.org-whatsnew82&utm_medium=referral&utm_campaign=whatsnew84&entrypoint=mozilla.org-whatsnew84" class="mzp-c-button mzp-t-product mzp-t-lg" rel="external" data-cta-text="Get Mozilla VPN Now" data-cta-type="button">Get Mozilla VPN Now</a>
```
alexgibson, line 32:
```suggestion
        <a href="https://vpn.mozilla.org/?utm_source=mozilla.org-whatsnew84&utm_medium=referral&utm_campaign=whatsnew84&entrypoint=mozilla.org-whatsnew84" class="mzp-c-button mzp-t-product mzp-t-lg" rel="external" data-cta-text="Get Mozilla VPN Now" data-cta-type="button">Get Mozilla VPN Now</a>
```

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx84.html
```diff
@@ -0,0 +1,74 @@
...
+  <aside class="mzp-l-content c-utilities">
+    {% set releasenotes_url = url('firefox.releases.index') if not version else '/firefox/%s/releasenotes/'|format(version) %}
+    <p>{{ ftl('whatsnew-release-notes', url=releasenotes_url) }}</p>
+  </aside>
+
+</main>
+{% endblock %}
+
+{% block js %}
+  {{ js_bundle('firefox_whatsnew_84') }}
```
alexgibson, line 73: This bundle doesn't seem to exist, does it need duplicating?

alexgibson (changes requested):
Spotted a utm param that looks like it still needs updating, and also a missing JS bundle. But otherwise this all looks as expected 👍 

I'm a little lost as to why we're duplicating the templates here in order to provide different utm params. We added a way to standardize utms params on /whatsnew pages, so they are automatically available in the template via a [context variable](https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/views.py#L624-L625). The hope was that this would avoid the need for template duplication like this.

Is there something that's not working there, or that we could improve to help need doing this? (not a blocker for this PR).

craigcook:
D'oh, I'd forgotten about the variable. Though there may still be some tiny advantage to duping the template so it's independent of the 82 template... but the repetition might be more debt than the reuse?

alexgibson (approved):

PR #9722, opened 2020-11-20
Configure firefox/privacy/book.ftl for de, fr, and en only. [no bug]

amychurchwell:
## Description

Configure firefox/privacy/book to be exposed for DE, FR, and EN localization only.

[Context from Peiying:](https://github.com/mozilla-l10n/www-l10n/pull/127#pullrequestreview-529626438)
> Given the length of this page, we are discussing with Justin on how to keep a content heavy page like this one to the target markets and offer an alternative, a shorter version to smaller communities. This page is best left for the l10n vendor to do.

The EU team already has translations available for DE and FR once this is available in Pontoon.

**www-l10n ref:** https://github.com/mozilla-l10n/www-l10n/pull/139

### DIFF
--- a/l10n/configs/pontoon.toml
+++ b/l10n/configs/pontoon.toml
@@ -146,5 +146,13 @@ locales = [
         "ru"
     ]
 [[paths]]
+    reference = "en/firefox/privacy/book.ftl"
+    l10n = "{locale}/firefox/privacy/book.ftl"
+    locales = [
+        "de",
+        "fr"
+    ]
+[[paths]]
     reference = "en/**/*.ftl"
     l10n = "{locale}/**/*.ftl"
+


amychurchwell:
@pmac I am attempting to configure this ftl file for `de`, `fr`, `en-*` exposure in pontoon only. Does this look correct to you? (I am new to this system 😅)

l10n/configs/pontoon.toml
```diff
@@ -148,3 +148,11 @@ locales = [
 [[paths]]
     reference = "en/**/*.ftl"
     l10n = "{locale}/**/*.ftl"
+
+[[paths]]
```
pmac, line 152: I think this new definition should be above the one above it which appears to be a catch-all definition. I'm honestly not that familiar with these config files either, but to me it looks like order is important.

pmac: Ah ha! Found the docs for this file format https://moz-l10n-config.readthedocs.io/en/latest/fileformat.html#paths

amychurchwell: that sounds like a good idea. Updating.

pmac: Cool. Looks good to me then 👍

pmac (approved):
🍂

PR #9723, opened 2020-11-20
l10n copy changes for book.ftl [no bug]

amychurchwell:
## Description

Updates bedrock's copy of `firefox/privacy/book.ftl` to reflect the copy changes that were implemented in the www-l10n repo.


craigcook (approved):
📖

PR #9735, opened 2020-12-01
Add more explicit path definitions to pontoon Fluent config

alexgibson:
See discussion in https://github.com/mozilla-l10n/www-l10n/pull/146

@pmac r? It seems like these config files don't quite work the way we originally thought, and we can't use broad wild cards like the way we have been if we need to expose certain files only to a subset of locales. 

This fix is a little verbose, but (from what I can tell) it's the only real option we have that doesn't involve re-architecting all the L10n files into different directories (which feels like a somewhat large change to make suddenly).

Thoughts?

### DIFF
--- a/l10n/configs/pontoon.toml
+++ b/l10n/configs/pontoon.toml
@@ -159,11 +159,11 @@ locales = [
     reference = "en/mozorg/about/**/*.ftl"
     l10n = "{locale}/mozorg/about/**/*.ftl"
 [[paths]]
-    reference = "en/mozorg/about/contribute.ftl"
-    l10n = "{locale}/mozorg/about/contribute.ftl"
+    reference = "en/mozorg/contribute.ftl"
+    l10n = "{locale}/mozorg/contribute.ftl"
 [[paths]]
-    reference = "en/mozorg/about/mission.ftl"
-    l10n = "{locale}/mozorg/about/mission.ftl"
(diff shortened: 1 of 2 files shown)


l10n/configs/pontoon.toml
```diff
@@ -115,6 +115,55 @@ locales = [
...
+    reference = "en/firefox/set-as-default/**/*.ftl"
+    l10n = "{locale}/firefox/set-as-default/**/*.ftl"
+[[paths]]
+    reference = "en/firefox/welcome/**/*.ftl"
+    l10n = "{locale}/firefox/welcome/**/*.ftl"
+[[paths]]
+    reference = "en/firefox/whatsnew/**/*.ftl"
+    l10n = "{locale}/firefox/whatsnew/**/*.ftl"
+[[paths]]
+    reference = "en/mozorg/about/**/*.ftl"
```
pmac, line 159: Doesn't this one include the next two, or is it really only those within subfolders of `about`?

pmac (approved):
Well this is unfortunate, but it is what it is. Looks good minus my one question below. 🥦

l10n/configs/pontoon.toml
alexgibson: Oh, you're right - shows how brittle this can be :/

alexgibson:
@pmac pushed a fix, thanks

PR #9744, opened 2020-12-02
Update to Protocol 13.0.1 [fix #8876]

### LINKED ISSUE #8876: Update Protocol to implement brand themes
## Description
Theming in Protocol should finally be arriving soonish (https://github.com/mozilla/protocol/issues/447). Once that feature has landed upstream in Protocol there will be some custom follow-up work to implement themes in Bedrock.

Mozilla.org is an atypical (possibly unique) consumer of Protocol in that it's one website with a common codebase that displays pages in both brand themes; some pages are Mozilla branded and other pages are Firefox branded.

Theming on the Protocol side is handled by variables and mixins that output different styles for different brands, which requires compiling the SCSS. In order to display two brands in Bedrock we'll need to compile Protocol twice and make two separate CSS bundles. I believe we can do this with a custom Gulp task. Then we'll have two base templates to choose from, each with its own brand theme, and any given page will inherit its theme from one of those bases.

---

- [ ] Both brand themes are compiled separately.
- [ ] All pages are using the appropriate brand theme.

#### Risks
- [ ] I'm assuming this is possible with Gulp but I haven't actually tried it yet, so it may prove impossible and we'll have to do something else.

#### Tasks
- [ ] Update bedrock to use the new version of Protocol (once it's published).
- [ ] Write a custom Gulp task to copy and compile both Protocol themes.
- [ ] Establish two CSS bundles, one for each theme.
- [ ] Establish two base templates, one for each theme.
- [ ] Update every page to use the appropriate template.

craigcook:
## Description
This is a big update that implements brand themes. It ended up touching a lot of pages to make minor style tweaks, mainly adjusting some font sizes because the type scale has been changed and declaring explicit colors for headings where needed, plus some general cleanup here and there (code formatting, switching to tokens, etc).

[See the release notes migration tips for reference.](https://github.com/mozilla/protocol/releases/tag/v13.0.0)

## Issue / Bugzilla link
#8876 

## Testing
https://www-demo-pr-9744.herokuapp.com/

### DIFF
--- a/bedrock/mozorg/templates/mozorg/home/home-de.html
+++ b/bedrock/mozorg/templates/mozorg/home/home-de.html
@@ -49,7 +49,7 @@
       </div>
 
       {{ billboard(
-        title='Für Dich & das Internet. <br>Schon seit 1998.',
+        title='Für Dich & das Internet. <br>Schon seit 1998.'|safe,
         ga_title='Für Dich & das Internet. Schon seit 1998.',
         desc='Mozilla setzt sich für ein gesundes Internet ein – und damit für die Menschen, die es nutzen. Damals. Heute. Morgen.',
         link_cta='Erfahre mehr über Mozilla',
--- a/media/css/firefox/browsers-products.scss
+++ b/media/css/firefox/browsers-products.scss
@@ -526,7 +526,7 @@ $url-download-link-hover: svg-url($download-link-hover);
 
 // box around form
 .mzp-c-emphasis-box.t-account {
-    @include light-links;
+    @include white-links;
     background: $color-ink-80 linear-gradient($color-violet-50 -100%, $color-ink-80);
     color: $color-white;
     margin: 0 auto $v-grid-xs auto;
@@ -567,6 +567,10 @@ $url-download-link-hover: svg-url($download-link-hover);
 .fxa-form-cta {
     max-width: 560px;
     margin: 0 auto;
+
+    .mzp-c-form-title {
+        @include font-firefox;
+    }
 
     @media #{$mq-md} {
         margin: 0;
--- a/bedrock/base/templates/500.html
+++ b/bedrock/base/templates/500.html
@@ -18,7 +18,7 @@
   {% if l10n_has_tag('404_update_2020') %}
     <main role="main">
       <section class="mzp-l-content">
-        <h1 class="error_header">{{ _('Something went wrong') }}</h1>
+        <h1 class="error-header">{{ _('Something went wrong') }}</h1>
         <p>{{ _('It’s probably just a server error and we’re working to fix it.') }}<br>
         {% trans firefox="https://firefox.com", mozilla=url('mozorg.home') %}
           You can also try refreshing this page or go to <a href={{ firefox }}>firefox.com</a> or <a href={{ mozilla }}>mozilla.org</a>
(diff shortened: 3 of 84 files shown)


bedrock/mozorg/templates/mozorg/home/home-de.html
```diff
@@ -49,7 +49,7 @@ <h1>{{ self.page_title() }}</h1>
       </div>
 
       {{ billboard(
-        title='Für Dich & das Internet. <br>Schon seit 1998.',
+        title='Für Dich & das Internet. <br>Schon seit 1998.'|safe,
```
craigcook, line 52: This is just a fix for a bug I spotted, unrelated to this PR (the `<br>` was rendered as text, not sure how long this bug has been in production).

media/css/firefox/browsers-products.scss
```diff
@@ -526,12 +528,22 @@ $url-download-link-hover: svg-url($download-link-hover);
 
 // box around form
 .mzp-c-emphasis-box.t-account {
-    @include light-links;
     background: $color-ink-80 linear-gradient($color-violet-50 -100%, $color-ink-80);
     color: $color-white;
     margin: 0 auto $v-grid-xs auto;
     position: relative;
 
+    a:link,
```
craigcook, line 536: The `light-links` mixin changed so inverted links now have color, but here and in a few other spots they need to be explicitly white. I've filed https://github.com/mozilla/protocol/issues/648

craigcook:
I think this is ready for review. Although the bit that updates Protocol is itself very small, this turned into a visual check of pretty much every page I could find and making a few adjustments where they seemed necessary.

alexgibson:
I spotted some pages where minor issues with light links occur (or at least link colors have changed compared to prod)

http://localhost:8000/en-US/firefox/windows/

image

image

http://localhost:8000/en-US/firefox/79.0/whatsnew/all/

image

(The release notes link color is also on other whatsnew pages)

alexgibson:
I pushed this branch for a test run and all seems well: https://gitlab.com/mozmeao/www-config/-/pipelines/228406086

alexgibson (changes requested):
I went through the migration tips and things look good too: https://github.com/mozilla/protocol/releases/tag/v13.0.0

r+wc on the small changes. I'm sure there are likely some other areas where we'll have light link issues on dark backgrounds but those seem minor and can easily be fixed.

craigcook:
The dark theme link colors were specified in https://github.com/mozilla/protocol/issues/428 though the values there strayed from the tokens and I ended up picking the token that was closest. But now seeing them in practice I still think the link colors are all too dark and/or saturated on dark backgrounds. That blue on an ink background is a real shock to the retinas.

So I'd like to revise the inverse link colors in the theme and publish a 13.0.1, which can also include the new `white-links` mixin for https://github.com/mozilla/protocol/issues/648 (I've already got it ready, just haven't submitted a PR yet). That will mean letting this PR sit just a little longer, but that's probably better than doing a bunch of overrides in bedrock that we'd just have to remove later.

alexgibson (approved):
Updates look good to me. From a user perspective, I still find things like this a bit confusing:

image

Looking a little deeper, I think it's because the top link is actually a `<button>`, but styled to look like a link. This is really another UX / design problem altogether imho, so not a blocker. But if you want to update the button color to match the new link styles then I'll leave that up to you. r+! 🕶️

craigcook:
Maybe we should just go with white links on those pages anyway... I'll do that.

PR #9768, opened 2020-12-08
Remove ARM64 UA detection for Windows downloads (Fixes #9753)

### LINKED ISSUE #9753: Remove ARM processor UA detection for Windows
## Description

I managed to get hold of an ARM64 Windows device for testing downloads and found the following:

- Our ARM64 UA detection on mozorg does not work, at least on my device. The user agent string does not provide any kind of hint to reveal that it has an ARM64 processor. As a result, the regular Windows stub installer is served when visitng /thanks, instead of the ARM64 build.
- The stub installer already has (much more reliable) built in logic for figuring out the currect build, and downloads the ARM64 version of Firefox correctly.

Given the above, I think we can simply remove our existing (bad) logic for ARM64 detection via UA strings.

---

#### Success Criteria
- [ ] ARM UA detection for Windows is removed from the website.

alexgibson:
## Description
Removes ARM64 / AArch64 UA detection for Windows because:
1. It does not work reliably (returns `false` on my ARM64 Windows laptop).
2. The stub installer already handles detection for the main FF release.
3. For pre-release builds such as Nightly / Beta, ARM64 builds can be download from /firefox/all/ (linked from the /channel page).

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/9753

## Testing
Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/226852818


media/css/protocol/components/_download-button.scss
```diff
@@ -91,7 +90,6 @@ ul.download-list {
 .linux .download-button .os_linux,
 .linux.x86.x64 .download-button .os_linux64,
 .windows .download-button .os_win,
-.windows.arm .download-button .os_win64-aarch64,
```
amychurchwell, line 94: I am confused about everything we'd like to remove in this PR. Are we removing `os_win64-aarch64` detection as well? Or only instances of `windows.arm` AND `os_win64-aarch64`?
I ask because I noted that there are instances of `assert pq(list[2]).attr('class') == 'os_win64-aarch64'` in our download button tests in `test_helpers.py`

media/js/base/mozilla-utils.js
```diff
@@ -33,8 +33,8 @@ if (typeof window.Mozilla === 'undefined') {
         switch(platform) {
         case 'windows':
             data.os = 'Desktop';
-            data.name = site.isARM ? 'Windows ARM64/AArch64' : 'Windows 32-bit';
-            data.version = site.isARM ? 'win64-aarch64': 'win';
```
amychurchwell, line 37: Are we no longer detecting `isARM` for windows? If so, do we need to remove-- https://github.com/mozilla/bedrock/blob/74f53d34a9a2513111cf9fe9432817f079d290a6/tests/unit/spec/base/mozilla-utils.js#L90-L94

amychurchwell (changes requested):
> Removes ARM64 / AArch64 UA detection for Windows

Feeling a bit confused about all we want to remove. Do we need to remove instances of `win64-aarch64` as well? Do we need to remove all instances of `isARM` detection for Windows? Or are we only targeting a specific combination of the two?

Does this affect firefox/all? I see a string in `all.ftl`:
`firefox-all-arm64-installers = ARM64/AArch64 installers` and
`firefox-all-arm64-installers-optimized = ARM64/AArch64 installers optimized for Snapdragon-powered { -brand-name-windows } PCs.`

media/css/protocol/components/_download-button.scss
alexgibson: We're removing UA detection for Arm64 on Windows. We're not removing Arm64 Windows builds, which we still produce and are available via firefox/all/

amychurchwell (approved):
I believe this covers everything, r+ 🚲

PR #9772, opened 2020-12-08
Add EOY2020 fundraising banner [fix #9765]

### LINKED ISSUE #9765: Add EOY 2020 fundraising banner
## Description

We should run the usual MoFo fundraising banner for the last week of the year, **from December 21 to January 4**. 

We should be able to reuse the same banner from last year, though possibly with some copy tweaks. Last year's banner was translated in-house in French and German, if we revise the copy we may need to revise the translations as well, or let the old text remain for those locales. We could also open this up to community l10n (with a conditional in the template so the banner is only visible to locales that have translated it).

This might interfere with the current banner promoting Firefox for Android, but maybe it's time to remove that banner anyway. Or we can see what happens with two banners.

We'll want to show the fundraising banner on at least the home page but possibly on a few other pages as well. We should come up with a list of pages.

---

#### Success Criteria
- [ ] Tasks should be represented here as something like "Approved designs attached to this issue"

craigcook:
## Description
Adds a MoFo fundraising banner to the home page (maybe more), to run from 21 December to 4 January.

## Issue / Bugzilla link
#9765 

## Testing
https://www-demo-pr-9772.herokuapp.com

### DIFF
--- a/l10n/en/banners/fundraising.ftl
+++ b/l10n/en/banners/fundraising.ftl
@@ -1,0 +1,16 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### URL: https://www-dev.allizom.org/
+
+banner-fundraising-title = The future is being decided online. Help { -brand-name-mozilla } fight for a web for all of us.
+banner-fundraising-body = Our personal data is collected every time we access the web, and it's not easy to find out what happens with it. Misinformation is making it harder for us to find the truth. Web-connected devices are pushed to market without meeting minimum security standards. The non-profit { -brand-name-mozilla-foundation } fights to make the internet safe, open and accessible to all. <strong>Will you donate today?</strong>
+banner-fundraising-monthly = Monthly
+banner-fundraising-one-time = One-time
+banner-fundraising-donate = Donate
+
+# Variables:
+#   $amount (number) - Inserts a number in US dollars, e.g. '$50'. Adapt the string in your translation for your locale conventions if needed, ex: { $amount } US$
+# You can also localize the currency (e.g. to use € instead of $), but only if your currency is available on https://donate.mozilla.org (check the currency dropdown)
+banner-fundraising-amount = ${ $amount }
--- a/bedrock/base/templates/includes/banners/fundraiser.html
+++ b/bedrock/base/templates/includes/banners/fundraiser.html
@@ -7,27 +7,20 @@
 {% block banner_id %}fundraising-banner{% endblock %}
 
 {% block banner_title %}
-  <span>{{ _('We all love the web. Join Mozilla in defending it.') }}</span>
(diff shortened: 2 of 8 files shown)


l10n/en/banners/fundraising.ftl
```diff
@@ -0,0 +1,14 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+banner-fundraising-title = The future is being decided online. Help { -brand-name-mozilla } fight for a web for all of us.
```
alexgibson, line 5: Nit insert the URL of the home page here so localizers have a link to preview their work?

l10n/en/banners/fundraising.ftl
```diff
@@ -0,0 +1,14 @@
...
+banner-fundraising-body = Our personal data is collected every time we access the web, and it's not easy to find out what happens with it. Misinformation is making it harder for us to find the truth. Web-connected devices are pushed to market without meeting minimum security standards. The non-profit { -brand-name-mozilla-foundation } fights to make the internet safe, open and accessible to all. <strong>Will you donate today?</strong>
+banner-fundraising-monthly = Monthly
+banner-fundraising-one-time = One-time
+banner-fundraising-donate = Donate
+
+# Variables:
+#   $amount (number) - Inserts a number in US dollars, e.g. '$50'. Adapt the string in your translation for your locale conventions if needed, ex: { $amount } US$
+# You can also localize the currency (e.g. to use € instead of $), but only if your currency is available on https://donate.mozilla.org (check the currency dropdown)
```
alexgibson, line 13: Nit: Suggest keeping this comment all on one line since it's related to the variable?

alexgibson (changes requested):
A couple of minor suggestions in the FTL file, but otherwise this looks good to me. r+wc 💸

l10n/en/banners/fundraising.ftl
alexgibson: Actually, technically you're correct it's not related to the variable but to the surrounding string instead. So maybe this does make the most sense already 👍

l10n/en/banners/fundraising.ftl
```diff
@@ -0,0 +1,16 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### URL: https://www-dev.allizom.org/firefox/accounts/
```
alexgibson, line 5: Copy pasta on the URL?

```suggestion
### URL: https://www-dev.allizom.org/
```

alexgibson (changes requested):

l10n/en/banners/fundraising.ftl
craigcook: 🤦 pasted from another file and didn't actually fix it. I'm an idiot.

alexgibson (approved):

Issue #9774, opened 2020-12-09
Standardize GA data-attributes for download buttons


alexgibson:
## Description

From some [discussion](https://github.com/mozilla/bedrock/pull/9754#discussion_r538671894) in previous PRs it seems like our app store download badges are missing some of the common GA data attributes that our regular download buttons contain.

Example:

Play Store badges have: `data-link-type="download" data-download-os="Android"`.
App Store badges have: `data-link-type="download" data-download-os="iOS"`.
Download buttons have: `data-link-type="download" data-download-os="Desktop" data-display-name="macOS" data-download-version="osx"` (macOS example)

Should both desktop and mobile buttons be the same, or are some attributes optional? If some are optional, then we should document the reasons why it appears in some places and not in others. 

It's not easy to tell, based on [our docs](https://bedrock.readthedocs.io/en/latest/analytics.html) that seem to suggest something entirely different:

Data Attribute | Expected Value
-- | --
data-link-type | ‘Desktop’, ‘Android’, or ‘iOS’
data-download-os | name or text of the link
data-download-version | ‘standard’, ‘developer’, ‘beta’

I'm guessing our docs are incorrect? What should the rules be?

#### Success Criteria
- [ ] Data attributes are standardized for both mobile and desktop.
- [ ] Documentation is up to date.

amychurchwell:
My guess is that the docs are incorrect.

I don't know enough about how these attributes are used in GA to make a call. Maybe it is worth huddling up with data science to get some clarity on what requirements should be?

I will say, it seems like we can safely say that `data-download-os` should be the os and not "name or text of the link"

(An issue that is tangentially relevant (https://github.com/mozilla/bedrock/issues/7392) about the GA conflict in using both `data-link-type` and `data-button-name` that.)

hoosteeno:
Here is how the data attributes from desktop download buttons presently** pass into GA through GTM:

```
event category = pageId + "Interactions"
event action = "Firefox Download"
event label = "Firefox for " + data-download-os
custom dimension 27 = data-download-version
custom dimension 28 = data-download-language
custom dimension 15 = data-download-location
```

The event values are used far more than the custom dimensions, because they end up powering goals. Looks like only data-download-os is required for them. The custom dimensions come in handy now and then, and I think there is no harm in including them on HTML elements if doing so helps to standardize our helpers somehow -- but probably not worth special effort to include on mobile links now.

[...]

PR #9777, opened 2020-12-10
Update analytics docs with correct download attributes (#9774)

alexgibson:
## Description
- Update analytics docs with more accurate download attribute examples.
- Highlight required attributes.

## Issue / Bugzilla link
#9774

### DIFF
--- a/docs/analytics.rst
+++ b/docs/analytics.rst
@@ -68,17 +68,21 @@ For all generic CTA links and <button> elements, add these data attributes (* in
 |    data-cta-position     | Location of CTA on the page (e.g. 'primary', 'secondary', 'header') |
 +--------------------------+---------------------------------------------------------------------+
 
-For all download buttons, add these data attributes:
+For all download buttons, add these data attributes (* indicates a required attribute):
 
-+--------------------------+--------------------------------+
-|    Data Attribute        |        Expected Value          |
-+==========================+================================+
-|    data-link-type        | 'Desktop', 'Android', or 'iOS' |
-+--------------------------+--------------------------------+
-|    data-download-os      | name or text of the link       |
-+--------------------------+--------------------------------+
-|    data-download-version |'standard', 'developer', 'beta' |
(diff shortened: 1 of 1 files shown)


docs/analytics.rst
```diff
@@ -68,17 +68,21 @@ For all generic CTA links and <button> elements, add these data attributes (* in
...
+|    Data Attribute         |        Expected Value                                                                         |
++===========================+===============================================================================================+
+|    data-link-type *       | 'download'                                                                                    |
++---------------------------+-----------------------------------------------------------------------------------------------+
+|    data-download-os *     | 'Desktop', 'Android', 'iOS'                                                                   |
++---------------------------+-----------------------------------------------------------------------------------------------+
+|    data-download-name     | 'Windows 32-bit', 'Windows 64-bit', 'macOS', 'Linux 32-bit', 'Linux 64-bit', 'iOS', 'Android' |
++---------------------------+-----------------------------------------------------------------------------------------------+
+|    data-download-version  | 'win', 'win64', 'osx', 'linux', 'linux64', 'ios', 'android'                                   |
```
amychurchwell, line 82: > (* indicates a required attribute)

Would the download button still be functional without `data-download-version`? Or should we reword "required attributes for analytics"?

amychurchwell (approved):
r+wc 🧄

docs/analytics.rst
alexgibson: As per Justin's comment here, only `data-link-type` and `data-download-os` are required attributes. Attributes such as `data-download-version` are not required for th goal (there's also only one bersion for iOS and Android, which kinda makes it redundant). So I think we're good here :)

alexgibson: After thinking about this a bit more, I agree it's worth adding a note for clarity - which I did here: https://gitlab.com/mozmeao/bedrock/-/commit/853f7bee77e7f993b5a7a2120d78e52d0bcad414 👍

Issue #9767, opened 2020-12-07
Add sticky promo box for downloads


stephaniehobson:
From this [pitch](https://docs.google.com/presentation/d/1zvJ-02NoXdmqCgw77TnP8GVdONYyUWITmyWfAd-3UP8/edit#slide=id.ga2fdac9f2d_0_0) and tracked in this [Trello card](https://trello.com/c/N6FbWQln/168-add-floating-promo-box-for-downloads-to-20-pages) (both Mozilla only links, sorry.)

Add a sticky promo box which:
- displays only to desktop users
- display only to non-Firefox users
- promotes the desktop download

To the following pages:

- [ ] /firefox/releasenotes
- [ ] /firefox/features/adblocker/  
- [ ] /firefox/features/bookmarks
- [ ] /firefox/features/fast
- [ ] /firefox/features/independant
- [ ] /firefox/features/memory
- [ ] /firefox/features/password manager
- [ ] /firefox/features/private browsing
- [ ] /firefox/features/safe browser
- [ ] /firefox/browsers/best-browser
- [ ] /firefox/browsers/browser-history
- [ ] /firefox/browsers/incognito-browser
- [ ] /firefox/browsers/what is a browser
- [ ] /firefox/browsers/win 64
- [ ] /firefox/browsers/Brave
- [ ] /firefox/browsers/Opera
- [ ] /firefox/browsers/Chrome
- [ ] /firefox/browsers/IE
- [ ] /firefox/browsers/Edge
- [ ] /firefox/browsers/Safari

To facilitate this work these pages can be switched from the Mozilla theme to the Firefox theme (got the okay for this from @rraue who commissioned the pages)

- [ ] /firefox/browsers/what-is-a-browser/
- [ ] /firefox/browsers/incognito-browser/
- [ ] /firefox/browsers/browser-history/
- [ ] /firefox/browsers/update-your-browser/ (not getting a promo but should be updated as well)

maxxcrawford:
@stephaniehobson Thanks for whipping this card into shape. I'll get started directly.

hoosteeno:
@stephaniehobson to review today, @maxxcrawford to complete this week.

hoosteeno:
@maxxcrawford to complete this week -- 1/26 complete.

PR #9780, opened 2020-12-10
Add "Download" Sticky Promo to various pages

maxxcrawford:
## Description

This PR adds the "Download" [Sticky Promo element](https://protocol.mozilla.org/patterns/molecules/sticky-promo.html) to mulitple pages. 

### Pages Checklist: 
- [x] /firefox/releasenotes
- [x] /firefox/features/adblocker/ 
- [x] /firefox/features/bookmarks
- [x] /firefox/features/fast
- [x] /firefox/features/independant
- [x] /firefox/features/memory
- [x] /firefox/features/password-manager
- [x] /firefox/features/private-browsing
- [x] /firefox/features/safe-browser
- [x] /firefox/browsers/best-browser
- [x] /firefox/browsers/browser-history
- [x] /firefox/browsers/incognito-browser
- [x] /firefox/browsers/what-is-a-browser
- [x] /firefox/browsers/windows-64-bit/
- [x] /firefox/browsers/compare/brave
- [x] /firefox/browsers/compare/opera
- [x] /firefox/browsers/compare/chrome
- [x] /firefox/browsers/compare/ie
- [x] /firefox/browsers/compare/edge
- [x] /firefox/browsers/compare/safari

New pages: (Not listed in #9767)
- [x] /firefox/features/
- [x] /firefox/browsers/quantum/
- [ ] ~/firefox/browsers/chromebook/~

Updated Theme from Moz to Firefox: 
- [x] /firefox/browsers/what-is-a-browser/
- [x] /firefox/browsers/incognito-browser/
- [x] /firefox/browsers/browser-history/
- [x] /firefox/browsers/update-your-browser/ (not getting a promo but should be updated as well)

## Issue / Bugzilla link

#9767

## Testing

- The promo should be visible only if in a Non-Firefox browser AND you have not dismissed the promo by clicking the "X" to dismiss. 

### Localhost URLs to check

- [localhost:8000/en-US/firefox/releasenotes/](localhost:8000/en-US/firefox/releasenotes/)

### Screenshot

image

### DIFF
--- a/media/js/firefox/sticky-promo.js
+++ b/media/js/firefox/sticky-promo.js
@@ -14,6 +14,14 @@
             return;
         }
 
+        // If the user is on Firefox and is NOT supposed to see sticky promo.
+        var fxUser = document.documentElement.classList.contains('is-firefox');
+        var hideFromFxUser = promo.classList.contains('hide-from-fx-user');
+
+        if ( fxUser && hideFromFxUser) {
+            return;
+        }
+
         var StickyPromo = {};
         var STICKY_PROMO_COOKIE_ID = 'firefox-sticky-promo';
 
--- a/media/static-bundles.json
+++ b/media/static-bundles.json
@@ -542,6 +542,7 @@
     },
     {
       "files": [
+        "css/firefox/sticky-promo.scss",
         "css/firefox/releasenotes.scss"
       ],
       "name": "firefox_releasenotes"
--- a/media/static-bundles.json
+++ b/media/static-bundles.json
@@ -190,6 +190,7 @@
     },
     {
       "files": [
+        "css/firefox/sticky-promo.scss",
         "css/firefox/features/detail.scss",
         "css/firefox/features/features-cards.scss"
       ],
@@ -479,6 +480,7 @@
     },
     {
       "files": [
+        "css/firefox/sticky-promo.scss",
         "css/firefox/features/index.scss",
         "css/firefox/features/features-cards-index.scss"
       ],
@@ -667,6 +669,7 @@
     },
     {
       "files": [
+        "css/firefox/sticky-promo.scss",
         "css/firefox/features/safebrowser.scss"
       ],
       "name": "safebrowser"
--- a/media/static-bundles.json
+++ b/media/static-bundles.json
@@ -639,6 +639,7 @@
     },
     {
       "files": [
+        "css/firefox/sticky-promo.scss",
         "css/firefox/browsers/what-is-a-browser.scss"
       ],
       "name": "what-is-a-browser"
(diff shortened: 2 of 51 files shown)


media/static-bundles.json
```diff
@@ -102,6 +102,12 @@
       ],
       "name": "grants"
     },
+    {
```
maxxcrawford, line 105: Instead of folding the JS and CSS into each page's  static bundle, I made individual JS and CSS bundles to be tacked on a template level. 

This execution was chosen because some of the pages do not have bundles defined. 

❗ An alternate approach is to use the stand-alone bundles only when there are no other bundles available for a specific template.

media/js/firefox/sticky-promo.js
```diff
@@ -14,6 +14,14 @@
             return;
         }
 
+        // If the user is on Firefox and is NOT supposed to see sticky promo.
```
maxxcrawford, line 17: This logic was written in such a way as to not require any modifications to the other pages who are already using the sticky promo. This means **the default behavior is to be visible to Firefox users** (and show them a list of other Firefox products).

media/static-bundles.json
stephaniehobson: I think this is good for the JS - with HTTP2 this might even be preferable. It'd be nice if we could get a `defer` on that script tag... that can be a V2 thing.

With the CSS include going in the header though it becomes render-blocking with all the potential problems of an extra request. I think it would be better for the CSS to use stand-alone bundles only when there are no other bundles available.

media/js/firefox/sticky-promo.js
stephaniehobson: 👍 Nice backwards compatability.

media/static-bundles.json
maxxcrawford: Commenting here to say this point is resolved, but wanna leave it up for posterity. I updated the commit to bundle the sticky promo SCSS with the release notes CSS bundle, as it already existed. 👍

maxxcrawford:
@stephaniehobson Should we also include this on the parent features page: https://www.mozilla.org/en-US/firefox/features/ ? It already exists on https://www.mozilla.org/en-US/firefox/browsers/

tests/pages/firefox/releasenotes.py
```diff
@@ -32,6 +33,7 @@ class FirefoxReleaseNotesPage(BasePage):
     _secondary_play_store_button_locator = (By.ID, 'download-android-secondary')
     _primary_app_store_button_locator = (By.ID, 'download-ios-primary')
     _secondary_app_store_button_locator = (By.ID, 'download-ios-secondary')
+    _sticky_promo_modal_content_locator = (By.CSS_SELECTOR, '.mzp-c-sticky-promo.mzp-a-slide-in')
```
stephaniehobson, line 36: Why do the release notes test need the content_locator defined in the file? I don't see it defined for the other pages.

stephaniehobson: Ooooh, I see the test failure. Poke me on Slack if you need to know how to run isolated tests locally.

maxxcrawford: I think this was copy paste from the [original test](https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/new/sticky_promo.py#L15).

stephaniehobson (commented):
We should not include the promo on the Chromebook page because they can't download and install from /new.

Other than that it looks great and the tests are passing 💯

maxxcrawford:
> We should not include the promo on the Chromebook page because they can't download and install from /new.

This has been reverted/removed! :D Ready to 🚀

stephaniehobson (approved):
R+

PR #9790, opened 2020-12-16
Add experiment URLs for analytics blocking stats

pmac:


bedrock/mozorg/urls.py
```diff
@@ -121,6 +121,11 @@
 
     url(r'^robots\.txt$', views.Robots.as_view(), name='robots.txt'),
 
+    # analytics experiment
+    url(r'^a7s-exp-2020/page-load.gif', views.tiny_gif),
```
pmac, line 125: @alexgibson the idea is that these URLs would be used in various ways on the site. When they load successfully they increment a counter. This one should be in an `` tag, the `ga-page-load.json` should be loaded when GA successfully fires, in  a callback if possible, and `js-page-load.json` should be loaded when the page loads, regardless of GA. In this way we can see the ratio of users blocking GA and JS and the total.

pmac: I can take a stab at the JS bits if you agree that this should work.

stephaniehobson: Ghostery and such actually mock the GA callback so last time I did this I had to opt for a more complicated solution. Documented here for your copy and paste pleasure: https://stephaniehobson.ca/posts/google-analytics-privacy-and-event-tracking/

stephaniehobson: Oh, actually, there's a lot of other stuff in there. Relevant code: `if(window.ga && ga.loaded)`

alexgibson: I think GTM might be a little different to regular GA (I also see things such as `window.google_tag_manager` in the global object?), but some of the same approaches might work here. However I'm not sure if trying to work out if GA has loaded via JS is necessarily going to be as robust as we might hope, given that some extensions shim analytics code.

Might it be a better measurement if we send a genuine non-interaction GA event, and then compare those impressions in GA itself?

```
window.dataLayer.push({
    'eAction': 'ga-loaded',
    'eLabel': 'a7s-exp-2020',
    'event': 'non-interaction'
});
```

pmac: Part of the point is that these should be blocked if GA or GTM is blocked. So I think it should be fine to do some simple stuff here. We're looking for rough numbers we can compare, not necessarily real numbers of views.

stephaniehobson: I think the dataLayer push is the way to go here.

Aayushap1 (approved):

PR #9791, opened 2020-12-16
Fix #9786 - Add CSS qualification to visible sticky promo box

### LINKED ISSUE #9786: Dismissing the sticky promo only makes it invisible
## Description

If you dismiss the sticky Firefox promo that's visible on certain pages, it's still there on next visit to the page but it's invisible (with `opacity: 0`). The result is an area of content that you can't click / interact with.

### Steps to reproduce

1. Visit https://www.mozilla.org/en-US/firefox/
2. Click the close button to dismiss the sticky promo.
3. Refresh the page.
4. Scroll down to the language selector in the bottom footer.
5. Click the language selector.

### Expected result

I should be able to select a language.

### Actual result

I'm redirected to getpocket.com

### Environment

Firefox Nightly

maxxcrawford:
## Description

Sticky Promo box was just being visually hidden, rather than not displayed, once dismissed.

## Issue / Bugzilla link

#9786 

## Testing

Go to [http://localhost:8000/en-US/firefox/](http://localhost:8000/en-US/firefox/) and dismiss the box. It should no longer visible and should not intefer with selecting the language from the footer.

### DIFF
--- a/media/css/firefox/includes/_sticky-promo.scss
+++ b/media/css/firefox/includes/_sticky-promo.scss
@@ -65,7 +65,7 @@ $image-path: '/media/protocol/img';
 }
 
 @media #{$mq-md} {
-    .mzp-c-sticky-promo {
+    .mzp-c-sticky-promo.mzp-a-slide-in {
         display: block;
     }
 }


stephaniehobson (approved):
R+

Issue #9783, opened 2020-12-11
Nightly /whatsnew page should get dark themed footer


alexgibson:
## Description

The Nightly /whatsnew page has a dark theme color but currently ets the light themed footer:

![image](https://user-images.githubusercontent.com/400117/101894087-8d369180-3b9d-11eb-92b0-049a235ccf95.png)

### Steps to reproduce

Visit: https://www.mozilla.org/en-US/firefox/84.0a1/whatsnew/all/

### Expected result

Footer should use dark theme.

### Actual result

Footer uses the light theme.

### Environment

N/A

Ayushsunny:
I wanna work on this issue

diabitbol:
you may have an other add-on activated that prevent you to change the theme

SnehVachhani:
i am working on it.

PR #9801, opened 2020-12-18
Change footer to dark theme

joneskj55:
## Description
This PR adds the dark theme to the nightly/ whatsnew page.
## Issue / Bugzilla link
Issue #9783
## Testing
Tested by running in browser on local machine.

### DIFF
--- a/bedrock/base/templates/includes/protocol/footer/footer.html
+++ b/bedrock/base/templates/includes/protocol/footer/footer.html
@@ -4,7 +4,7 @@
 
 {% set utm_params = 'utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=footer' %}
 
-<footer class="c-footer mzp-t-dark {% if theme_class %}{{ theme_class }}{% endif %}" id="colophon">
+<footer class="c-footer {% if theme_class %}{{ theme_class }}{% endif %}" id="colophon">
   <div class="mzp-l-content">
     <nav class="c-footer-primary">
       <div class="c-footer-sections">
--- a/bedrock/base/templates/includes/protocol/footer/footer.html
+++ b/bedrock/base/templates/includes/protocol/footer/footer.html
@@ -4,7 +4,7 @@
 
 {% set utm_params = 'utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=footer' %}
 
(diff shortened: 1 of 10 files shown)


craigcook (changes requested):
Thanks for the PR! Unfortunately this sets the dark theme for the footer across the entire site. To do it on specific pages we need to pass a theme class to the include on just that page's template, like this:

```
{% block site_footer %}
  {% with theme_class='mzp-t-ink' %}
    {% include 'includes/protocol/footer/footer.html' %}
  {% endwith %}
{% endblock %}
```

Add that at the bottom of all the Nightly template files in https://github.com/mozilla/bedrock/tree/master/bedrock/firefox/templates/firefox/nightly and you should get the ink-colored footer only on those pages without impacting the rest of the site.

joneskj55:
@craigcook Thanks! I'll knock that out and commit here. Quick question, is `mzp-t-ink` the class you want or `mzp-t-dark`? I noticed that ink is more purple and matches with that page but just want to make sure I use the correct class.

craigcook:
> @craigcook Thanks! I'll knock that out and commit here. Quick question, is `mzp-t-ink` the class you want or `mzp-t-dark`? I noticed that ink is more purple and matches with that page but just want to make sure I use the correct class.

I think ink probably works better with the Nightly colors so let's go with that. Thanks!

craigcook:
Oops, I misled you... looks like the language-specific templates all extend other templates, so the only files we really need to change are `firstrun.html` and `whatsnew.html` (the others will get the ink footer from those parent templates).

And I'm not sure why CircleCI failed, doesn't seem related to any of the code changes so maybe just a network issue. Maybe it'll fix itself on the next commit.

joneskj55:
I can go ahead and update those two files and revert the language-specific ones and see if that works.

joneskj55:
That change worked on my local machine. Also, the check passed this time. Hope this works! Fingers crossed 🤞🏾

craigcook:
🤦 I once again misled you. Looks like the firstrun page already has the ink footer, we only need to update `whatsnew.html`. Sorry for all the back-and-forth, thanks for your patience!

joneskj55:
Hahaha no problem, I'll change it real quick!

craigcook (approved):
There we go, sorry about all that. Looks great, thanks for your contribution! 🌃

PR #9813, opened 2020-12-21
Decommission /enterprise/signup/ page (Fixes #9788)

### LINKED ISSUE #9788: Decom or Keep https://www.mozilla.org/en-US/firefox/enterprise/signup/
## Description
As part of researching all things SalesForce I came across ye olde enterpise signup form. This submits to:

https://webto.salesforce.com/servlet/servlet.WebToLead?encoding=UTF-8"

I think Lauren needs to make the call if we still need this form. If so, we will need a new endpoint come February. If not we can decom this page.

---

#### Success Criteria
- [x] A decision on keep or decom is made. 
- [ ] Based on the above we keep and change the form endpoint or we decom.

alexgibson:
## Description
Removes and redirects the following two pages:

- https://www.mozilla.org/en-US/firefox/enterprise/signup/
- https://www.mozilla.org/en-US/firefox/enterprise/signup/thanks/

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/9788

## Testing
- [x] No stray assets left behind.


stephaniehobson (approved):
R+ 📝

PR #9822, opened 2020-12-22
Fix sub navigation on iOS devices (Fixes #9821)

### LINKED ISSUE #9821: Sub navigation does not minimize on iPhone browsers
## Description

On iPhones the sub navigation does not minimize by default on page load.

### Steps to reproduce

1.) On an iPhone, visit a page with sub navigation such as : https://www.mozilla.org/en-US/firefox/84.0/releasenotes/

### Expected result

image

### Actual result

image

### Environment

Happens both in Safari and Firefox on iOS 14.3

alexgibson:
## Description
Fixes sub navigation collapse / expand on iOS devices

## Issue / Bugzilla link
#9821

## Testing
- [ ] STR in the issue should no longer occur.


stephaniehobson (approved):
R+ )

PR #9843, opened 2021-01-14
Add Reimagine Open page [fix #9828]

### LINKED ISSUE #9828: Publish "Reimagine Open" whitepaper
## Description

Mozilla is publishing a paper on the current state and future of the open web. It's currently in a PDF format but we've been asked to put together a canonical web page version as well. 

Visual design should evoke some of the feeling of the PDF but can have a simplified layout. I think we can base it on the Little Book of Privacy page as a starting point and change up the graphics. I'll work with Michael Ham for final assets.

I don't think there's any current desire to localize it but it might be worthwhile to format it in Fluent anyway, just in case.

---

#### Success Criteria
- [ ] Page is on demo for review before 12 January 2021
- [ ] Page is live on or before 13 January 2021

craigcook:
## Description
Publishes a new MoFo white paper about the future of the open internet. No l10n so strings are just hard-coded.

If possible we want to get this to prod today (14 January, 2021) so we can publish a blog post about it.

## Issue / Bugzilla link
#9828 

## Testing
http://localhost:8000/foundation/reimagine-open/
https://www-demo4.allizom.org/foundation/reimagine-open/


maxxcrawford (approved):
LGTM! 🚀 

I found one issue with the background overlapping content at midsized screens. It could be possible to style this breakpoint first, and then applying the layout to extra large (~1920px wide) screens. 

_(Taken at 1280px wide)_

Screen Shot 2021-01-14 at 12 27 29

craigcook:
That overlap is largely unavoidable unless without adjusting the positioning at multiple breakpoints, or do something else entirely. So rather than avoid the overlap I made it an intentional part of the design, and the transparent background on the content container keeps the text readable. I'm still not 100% happy with it and explored some other ideas, but I got Michael Ham's blessing on this treatment so we went with it. If anything it looks worse on super-wide screens because the image ends up way over to the right with a bunch of dead whitespace, but that's more of an edge case.

PR #9846, opened 2021-01-18
Removed lang file l10n commands and docs

### LINKED ISSUE #8962: Remove lang file l10n commands and docs
We are now in read-only mode for lang files. We will not be extracting any more strings using the old commands. We should therefore remove these commands and the docs for them from the project. 

The commands are:

* `l10n_check`
* `l10n_extract`
* `l10n_merge`

Patil2099:
Fixes #8962


alexgibson:
This PR was only partially completed and is now stale, so I'm going to close this PR. Thanks nonetheless @Patil2099!

PR #9847, opened 2021-01-18
Pass GA visit ID to stub attribution (Fixes #9830)

### LINKED ISSUE #9830: pass ga_visit_id into bouncer for windows stub installer downloads
## Description

When a moz.org visitor clicks on the 'download firefox' link to receive a download of the Windows stub installer, bedrock should pass the GA visit id into the bouncer API call so bouncer can emit that (along w a generated download token) into stackdriver for ultimate landing in a BigQuery data set.

---

#### Success Criteria
- [ ] Bouncer successfully receives GA visit id along with requests to initiate a Windows stub installer download

alexgibson:
## Description
- Passes Google Analytics visit ID to stub attribution for Windows visitors.

## Issue / Bugzilla link
#9830

## Testing
Demo: https://www-demo3.allizom.org/

- [ ] Is my method for checking GA loading sane / non-error prone?

### DIFF
--- a/media/js/base/stub-attribution.js
+++ b/media/js/base/stub-attribution.js
@@ -199,6 +199,48 @@ if (typeof window.Mozilla === 'undefined') {
     };
 
     /**
+     * Gets the client ID from the GA object.
+     * @returns {String} client ID.
+     */
+    StubAttribution.getGAVisitID = function() {
+        try {
+            return window.ga.getAll()[0].get('clientId');
+        } catch (e) {
+            return null;
+        }
+    };
+
+    /**
+     * A crude check to see if Google Analytics has loaded. Periodically
+     * attempts to retrieve the client ID from the global `ga` object.
+     * @param {Function} callback
+     */
+    StubAttribution.waitForGoogleAnalytics = function(callback) {
+        var timeout;
+        var pollRetry = 0;
+        var interval = 100;
+        var limit = 20; // (100 x 20) / 1000 = 2 seconds
+
+        function _checkGA() {
+            clearTimeout(timeout);
+            var clientID = StubAttribution.getGAVisitID();
+
+            if (clientID && typeof clientID === 'string') {
+                callback(true);
+            } else {
+                if (pollRetry <= limit) {
+                    pollRetry += 1;
+                    timeout = window.setTimeout(_checkGA, interval);
+                } else {
+                    callback(false);
+                }
+            }
+        }
+
+        _checkGA();
+    };
+
+    /**
      * Gets utm parameters and referrer information from the web page if they exist.
      * @param {String} ref - Optional referrer to facilitate testing.
      * @return {Object} - Stub attribution data object.
@@ -210,6 +252,7 @@ if (typeof window.Mozilla === 'undefined') {
         var variation = params.get('variation') || StubAttribution.experimentVariation;
         var referrer = typeof ref !== 'undefined' ? ref : document.referrer;
         var ua = StubAttribution.getUserAgent();
+        var visitID = StubAttribution.getGAVisitID();
 
         /* eslint-disable camelcase */
         var data = {
@@ -220,7 +263,8 @@ if (typeof window.Mozilla === 'undefined') {
             referrer: referrer,
             ua : ua,
             experiment: experiment,
-            variation: variation
+            variation: variation,
+            visit_id: visitID
         };
         /* eslint-enable camelcase */
 
@@ -290,11 +334,14 @@ if (typeof window.Mozilla === 'undefined') {
         // make the XHR request to the stub authentication service.
         } else if (!StubAttribution.isFirefoxNewScene2()) {
 
-            data = StubAttribution.getAttributionData();
+            // Wait for GA to load so that we can pass along visit ID.
+            StubAttribution.waitForGoogleAnalytics(function() {
+                data = StubAttribution.getAttributionData();
 
-            if (data && StubAttribution.withinAttributionRate()) {
-                StubAttribution.requestAuthentication(data);
-            }
+                if (data && StubAttribution.withinAttributionRate()) {
+                    StubAttribution.requestAuthentication(data);
+                }
+            });
         }
     };
 
--- a/bedrock/firefox/tests/test_views.py
+++ b/bedrock/firefox/tests/test_views.py
@@ -51,6 +51,7 @@ class TestStubAttributionCode(TestCase):
             'experiment': '(not set)',
             'variation': '(not set)',
             'ua': '(not set)',
+            'visit_id': '(not set)',
         }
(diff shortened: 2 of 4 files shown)


alexgibson:
@pmac @hoosteeno this is a pretty primitive method, but given all the ways that third party scripts can either be blocked or just plain slow to load, I'm not sure if there are any more accurate methods. Right now, this will check for existence the GA visit ID at an interval of every 200ms, and then eventually timeout after 4 seconds if GA hasn't created it. I'm happy to tweak this as we see fit. Thoughts?

As it stands, I think this could likely miss out data from people on pretty slow internet connections. However, the tricky thing to balance here is that we also don't want to delay kicking things off too long, because then we also risk missing out on stub attribution data altogether for some visits (e.g. someone has fast internet but uses an adblocker to block GA).

media/js/base/stub-attribution.js
```diff
@@ -290,11 +334,14 @@ if (typeof window.Mozilla === 'undefined') {
         // make the XHR request to the stub authentication service.
         } else if (!StubAttribution.isFirefoxNewScene2()) {
 
-            data = StubAttribution.getAttributionData();
+            // Wait for GA to load so that we can pass along visit ID.
+            StubAttribution.waitForGoogleAnalytics(function() {
```
pmac, line 338: Does this mean that we'll only get Attribution if the user sticks around for over 4 seconds if they don't get GA? If so is it possible to collect the other attribution data more quickly and then tack on the visit_id whenever it becomes available?

alexgibson: What do you mean by collecting other attribution data more quickly? Would that mean making 2 calls to the auth end point? 

If so, what should we do in that instance, throw away the previous data? Or is is possible to make a second call to encode only the visit_id and then add that to what we already have? It feels like this is a point where complexity might start to creep in…

pmac: We would have to throw away the old data... hmm... I actually thought that the signing of the collected data happened later on the download page itself. I forgot that it actually happens immediately. If you think this won't result in much loss of attribution then I'm cool with it like it is. But if we do see that a lot of people don't have GA and we miss a lot of attribution as a result we should consider getting the data in two steps perhaps. It's also possible that the rest of this data is superfluous once we get the GA ID stuff working, so hopefully we can simplify everything after it's proved to be working.

alexgibson: A more robust way to implement this imho would be for bedrock to generate the unique ID, which we then pass both into GA and stub attribution independently. But I’m not 100% clear on prior discussion for this feature or the other work to support it though, so if this is the way it needs to work the I’ll see if I can tweak things a little more.

pmac: the whole idea though is that we want this to respect privacy, and the only way to do that is for this download ID to stay only in our private database. If it's in GA then it's more data that we're giving GA. All that we need to do is get the GA ID to bouncer if one exists. It then generates the download ID and stores them.

alexgibson: It feels a bit like the same level of tracking is going on here regardless to where an ID gets generated. We’re still piecing together the same level of information about visitors. If this is the way it’s been designed though then let’s try and make it work 👍

pmac: The point is to restrict access to the info required to track very tightly.

alexgibson: I think we could probably still do that depending on how it was implemented, but this is probably another convo. I have another approach I might try here tomorrow to get us what we need.

alexgibson:
Ok updated. I had hoped we could use the window load event to capture the GA script loading in a more predictable manner, but I forgot that we load it async (since it's third party), so these types of events ignore those scripts entirely. I think our best bet here is probably to tweak the timing of the existing method to what we think is sensible.

I've gone ahead and changed the max timeout to 2 seconds, so the logic is basically as follows:

1. The stub attribution script runs as soon as it's downloaded and checks for the GA client ID.
2. If the GA client ID exists, the do the data auth dance.
3. If there's no ID, then keep checking every 100ms or so until it's found.
4. If found, do the auth.
5. If still not found after 2 seconds of continually checking, then stop and do stub attribution as normal.

With the above in place I think we can probably assume:

- This should do a reasonable job at capturing client IDs, providing visitors are on a decent / stable internet connection.
- The number of client IDs captured likely cannot be used to reason about things such as % of people who block GA (slower connections will muddle this data).

If you have any other suggestions, I'm happy to try them.

pmac:
I agree. This sounds great. Thanks for all your work on this!

pmac (approved):
This looks great. I think we can do the testing they want to do with demo3, so let's not merge this yet, but I think it's ready.

pmac:
I've updated the config for demo3 to point to [staging bouncer](https://bouncer-bouncer.stage.mozaws.net/) and with the correct signing key for that server.

hoosteeno:
@pmac @alexgibson Will the visit ID get passed to a nightly download? The Firefox change that consumes this visit ID is only in nightly at the moment.

pmac:
I don't see a stub installer download on the site for nightly. The stub installer for release should get the strings regardless though, so I think we should be able to do an end-to-end test anyway.

bedrock/firefox/firefox_details.py
```diff
@@ -293,6 +293,10 @@ def get_download_url(self, channel, version, platform, locale,
 
         product = '%s-%s' % (prod_name, suffix)
 
+        # Temporary hack to return beta stub for testing
```
alexgibson, line 296: TODO: remove before merging

media/js/base/stub-attribution.js
```diff
@@ -107,7 +107,7 @@ if (typeof window.Mozilla === 'undefined') {
             var version;
             var directLink;
             // Append stub attribution data to direct download links.
-            if (link.href && link.href.indexOf('https://download.mozilla.org') !== -1) {
+            if (link.href && (link.href.indexOf('https://download.mozilla.org') !== -1 || link.href.indexOf('https://bouncer-bouncer.stage.mozaws.net') !== -1)) {
```
alexgibson, line 110: TODO: remove before merging

alexgibson:
@pmac just checking in, has there been any update on this from the FF team?

pmac:
I've asked in the slack channel. We can't deploy this until the changes to the stub attribution service are in production. I'll update here when I have a timeline.

alexgibson:
@pmac Ok I've removed the temporary hacks to get this working on demo. If the Firefox team are happy with this, then I think we're ready? We'll have to see if there's a drop in overall attribution data because of the way this is set up, but hopefully it should be small (if any).

pmac:
Sounds good. I'm going to merge and get it going to prod so they can test. Thank you!

PR #9849, opened 2021-01-18
Update donate metadata

### LINKED ISSUE #9848: Update donate link metadata
Some of the metadata for the Donate website is outdated:

- We dropped support of several currencies
- We dropped a couple locales
- Some unsupported locales were added to `DONATE_PARAMS`

E.g https://www.mozilla.org/fi/ is linking to https://donate.mozilla.org/fi/?presets=50,30,20,10&amount=30&utm_source=mozilla.org&utm_medium=referral&utm_content=company&currency=eur&utm_campaign=footer which is a 404, because Finnish is not supported on the donate website.

When a locale is not supported, we should rather link to https://donate.mozilla.org/?utm_source=mozilla.org&utm_medium=referral&utm_content=company&utm_campaign=footer without any custom presets

TheoChevalier:
Fixes #9848

### DIFF
--- a/bedrock/mozorg/templatetags/misc.py
+++ b/bedrock/mozorg/templatetags/misc.py
@@ -380,15 +380,15 @@ def donate_url(ctx, source=''):
 
     For en-US this would output:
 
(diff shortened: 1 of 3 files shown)


alexgibson (approved):
Thanks @TheoChevalier 💯

PR #9850, opened 2021-01-19
Migrate error pages to Fluent (Fixes #9807) [skip l10n]

### LINKED ISSUE #9807: Migrate 404 and 500 error pages to Fluent
## Description

https://www.mozilla.org/en-US/404

---

#### Success Criteria
- [ ] Page is using Fluent in production along with existing migrations

alexgibson:
## Description
- http://localhost:8000/en-US/404/
- http://localhost:8000/en-US/500/

## Issue / Bugzilla link
#9807

## Testing
```
./manage.py l10n_update
```

### DIFF
--- a/l10n/en/404.ftl
+++ b/l10n/en/404.ftl
@@ -1,0 +1,22 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+### URL: https://www-dev.allizom.org/404/
+
+not-found-page-not-found-page-page-not-found = 404: Page Not Found
+not-found-page-sorry-we-cant-find-that-page = Sorry, we can’t find that page
+not-found-page-were-all-about-a-healthy-internet = We’re all about a healthy internet but sometimes broken URLs happen.
+not-found-page-go-back = Go Back
+
+# Variables:
+#   $about (url) - link to https://www.mozilla.org/about/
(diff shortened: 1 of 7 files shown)


l10n/en/404.ftl
```diff
@@ -0,0 +1,22 @@
...
+not-found-page-were-all-about-a-healthy-internet = We’re all about a healthy internet but sometimes broken URLs happen.
+not-found-page-go-back = Go Back
+
+# Variables:
+#   $about (url) - link to https://www.mozilla.org/about/
+not-found-page-learn-about-mozilla-the-non = <a href="{ $about }">Learn</a> about { -brand-name-mozilla }, the not-for-profit behind { -brand-name-firefox }.
+
+# Variables:
+#   $explore (url) - link to https://www.mozilla.org/firefox/
+not-found-page-explore-the-entire-family-for = <a href={ $explore }>Explore</a> the entire family for { -brand-name-firefox } products designed to respect your privacy.
```
craigcook, line 18: Nit: I think this should be "of" instead of "for," so it reads "Explore the entire family **of** Firefox products." 

Though now with VPN I wonder if this should be "Mozilla products" or even just "the entire family of products"? But maybe this isn't the time for a rewrite.

craigcook (approved):
Just a grammar nitpick on one line, r+wc

l10n/en/404.ftl
alexgibson: Makes sense yeah. For now, I think it's probably best to leave this until when we next update this page (strings have already been reviewed on the l10n side).

PR #9859, opened 2021-01-21
Mention and exclude ngrok from domain takeover bounties

ajvb:
Specifically exclude *.ngrok.io and "development-only" domains from the Domain Takeover bounty.

### DIFF
--- a/bedrock/security/templates/security/web-bug-bounty.html
+++ b/bedrock/security/templates/security/web-bug-bounty.html
@@ -66,7 +66,7 @@
         <th scope="row">Domain Takeovers</th>
         <td>$5000</td>
         <td>$2000</td>
-        <td>$500/$200<sup>5</sup></td>
+        <td>$500<sup>5</sup>/$200<sup>6</sup></td>
       </tr>
       <tr>
         <th scope="row">XSS (minor)</th>
@@ -81,7 +81,7 @@
         <td>--</td>
       </tr>
       <tr>
-        <th scope="row">Clickjacking<sup>6</sup></th>
+        <th scope="row">Clickjacking<sup>7</sup></th>
         <td>$1000</td>
         <td>$500</td>
         <td>--</td>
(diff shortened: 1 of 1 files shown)


bedrock/security/templates/security/web-bug-bounty.html
```diff
@@ -100,7 +100,7 @@ <h2 id="payouts-section">Payouts</h2>
     <li>Includes IDORs that bypass authentication or authorization for significant actions</li>
     <li>Significant actions only, such as changing email/passwords, deleting accounts, etc.</li>
     <li>Must be able to conduct significant action (i.e., not defacement, phishing, cookie injection, etc.)</li>
-    <li>For domains falling outside *.mozilla.org, *.mozilla.com, *.mozilla.net, and *.firefox.com</li>
+    <li>For domains falling outside *.mozilla.org, *.mozilla.com, *.mozilla.net, and *.firefox.com. *.ngrok.io and similar development-only domains are excluded.</li>
```
claudijd, line 103: I think we may be overloading on the exclusion here...

I think intent is:

$500 for *.mozilla.org, *.mozilla.com, *.mozilla.net, and *.firefox.com takeovers
$200 for anything else excluding *.ngrok.io and similar development-only domains takeovers
 
What if we put in two exceptions, each with their own number and put the number next to 500 and 200 respectively?   Would that make our intent more clear?

ajvb: Yeah good point. Will make that change.

claudijd (approved):

foxliz (approved):

g-k (approved):

ajvb:
@alexgibson Pinging you for r? based off of past web bounty page changes (i.e. https://github.com/mozilla/bedrock/pull/8966).

alexgibson (approved):

Issue #9864, opened 2021-01-22
Redesign/rebuild S2D widget


hoosteeno:
## Description

The Send-to-device (S2D) widget on www.mozilla.org includes SMS capabilities. As we transition to a new email and contacts provider, we'll EOL SMS as a delivery mechanism for these messages. This epic includes short term and long term work to make that true. We can build task cards for individual activities below as needed.

---

#### Issues
- [x] SHORT TERM: Remove SMS from widget -- eliminate language advertising SMS, and remove features providing it.
LONG TERM:
- [ ] Design a new widget -- it should include email, QR code, and app-store badges
- [ ] Automate QR codes, so Bedrock serves them rather than manually creating
- [ ] Implement & launch the new widget
- [ ] Replace all widgets (in every locale) with new widget, since geography is no longer a constraint

maxxcrawford:
Note - I removed myself from this epic issue and assigned myself to the task #9875 as part of the epic.

alexgibson:
Closing old / stale issues. Feel free to re-open if still relevant

PR #9868, opened 2021-01-23
Remove SMS feature for send to device (#9864)

stephaniehobson:
## Description

Remove SMS feature for send to device 
- rename form field from  `phone-or-email` to `s2d-email` (unique name makes it easier to trace the path through the code)
- remove `legal_note_sms` and `input_label_alt` from macro
- remove sms error and thank you messages and form labeling
- remove geolocation check from helpers and javascript
- remove sms from message sets (keep message sets because not all countries get email option)
- remove sms tests
  - convert one test from sms to email
- update docs to remove sms references
- remove sms strings from ftl files
- update JS and CSS to remove sms behaviour and styles
- update JS tests

## Issue / Bugzilla link

See #9864 

## Testing

Testing values: success@example.com, failure@example.com, invalid@email.
Should still fallback to QR codes for languages without translated emails (eg. AR)

http://localhost:8000/en-US/firefox/mobile/
http://localhost:8000/en-US/firefox/mobile/get-app/
http://localhost:8000/en-US/firefox/whatsnew/all/
http://localhost:8000/en-US/firefox/79.0/whatsnew/all/
http://localhost:8000/en-US/firefox/80.0/whatsnew/all/
http://localhost:8000/en-US/firefox/81.0/whatsnew/all/
http://localhost:8000/en-US/firefox/welcome/4/
http://localhost:8000/en-US/firefox/welcome/5/
http://localhost:8000/en-US/firefox/welcome/6/
http://localhost:8000/en-US/firefox/welcome/7/

### DIFF
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -54,7 +54,7 @@
 
 {# Docs: https://bedrock.readthedocs.io/en/latest/send-to-device.html #}
 {% macro send_to_device(platform='all', product='firefox', message_set='default', id='send-to-device', class='horizontal', include_title=True, title_text=None, input_label=None, legal_note_email=None, spinner_color='#000', button_class=None) %}
-  <section id="{{ id }}" class="send-to-device {{ class }}">
+  <section id="{{ id }}" class="send-to-device {{ class }}" data-spinner-color="{{ spinner_color }}">
     <div class="form-container">
       {% if include_title %}
         <h2 class="form-heading">
--- a/bedrock/base/templates/macros.html
+++ b/bedrock/base/templates/macros.html
@@ -53,8 +53,8 @@
 
 
 {# Docs: https://bedrock.readthedocs.io/en/latest/send-to-device.html #}
-{% macro send_to_device(platform='all', product='firefox', message_set='default', id='send-to-device', class='horizontal', include_title=True, title_text=None, input_label=None, legal_note_email=None, legal_note_sms=None, spinner_color='#000', input_label_alt=None, button_class=None) %}
-  <section id="{{ id }}" class="send-to-device {{ class }}" data-countries="{{ send_to_device_sms_countries(message_set) }}" data-spinner-color="{{ spinner_color }}">
+{% macro send_to_device(platform='all', product='firefox', message_set='default', id='send-to-device', class='horizontal', include_title=True, title_text=None, input_label=None, legal_note_email=None, spinner_color='#000', button_class=None) %}
+  <section id="{{ id }}" class="send-to-device {{ class }}">
     <div class="form-container">
       {% if include_title %}
         <h2 class="form-heading">
@@ -68,9 +68,6 @@
       <h2 class="mzp-u-title-xs thank-you hidden">{{ ftl('send-to-device-your-download-link') }}</h2>
       <form class="send-to-device-form" action="{{ url('firefox.send-to-device-post') }}" method="post">
         <ul class="mzp-c-form-errors hidden">
-          <li class="sms">
-            {{ ftl('send-to-device-sorry-we-cant-send', fallback='send-to-device-sorry-this-number') }}
-          </li>
           <li class="email">{{ ftl('send-to-device-please-enter-an-email') }}</li>
           <li class="system">{{ ftl('send-to-device-an-error-occured') }}</li>
         </ul>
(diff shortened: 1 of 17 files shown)


stephaniehobson:
@pmac I would like to change the name of the input from `['phone-or-email']` to something else to reflect that phone is no longer acceptable. How do I do that without screwing up basket?

stephaniehobson:
@pmac I think I found it in views.py :) Heavy python changes in this one, could use your R.

bedrock/base/templates/macros.html
```diff
@@ -53,8 +53,8 @@
 
 
 {# Docs: https://bedrock.readthedocs.io/en/latest/send-to-device.html #}
-{% macro send_to_device(platform='all', product='firefox', message_set='default', id='send-to-device', class='horizontal', include_title=True, title_text=None, input_label=None, legal_note_email=None, legal_note_sms=None, spinner_color='#000', input_label_alt=None, button_class=None) %}
-  <section id="{{ id }}" class="send-to-device {{ class }}" data-countries="{{ send_to_device_sms_countries(message_set) }}" data-spinner-color="{{ spinner_color }}">
```
alexgibson, line 57: Is removing `data-spinner-color="{{ spinner_color }}` intentional here? Looks like it's still used by the JS? https://github.com/mozilla/bedrock/blob/master/media/js/base/send-to-device.js#L33

bedrock/base/templates/macros.html
```diff
@@ -82,27 +79,18 @@ <h2 class="mzp-u-title-xs thank-you hidden">{{ ftl('send-to-device-your-download
...
-            <label class="mzp-c-field-label" for="{{ id }}-input" data-alt="{% if input_label_alt %}{{ input_label_alt }}{% else %}{{ ftl('send-to-device-enter-your-email-or-phone', fallback='send-to-device-enter-your-email-or-phone-10-digit') }}{% endif %}">
+            <label class="mzp-c-field-label" for="{{ id }}-input">
               {% if input_label %}
                 {{ input_label }}
               {% else %}
                 {{ ftl('send-to-device-enter-your-email') }}
               {% endif %}
             </label>
-            <input id="{{ id }}-input" class="mzp-c-field-control send-to-device-input" name="phone-or-email" type="text" required>
+            <input id="{{ id }}-input" class="mzp-c-field-control send-to-device-input" name="s2d-email" type="text" required>
```
alexgibson, line 89: Might renaming this break something on the receiving end?

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx80.html
```diff
@@ -6,10 +6,6 @@
 
 {% extends "firefox/whatsnew/base.html" %}
 
-{% block html_attrs %}
-data-whatsnew80-redirect-url="{{ settings.FXA_ENDPOINT + 'sms/?' + utm_params }}"
```
alexgibson, line 10: Not entirely clear what this was used for, is it unrelated to this change?

tests/redirects/map_globalconf.py
```diff
@@ -356,7 +356,6 @@
     url_test('/firefox/{performance,happy,speed,memory}/', '/firefox/features/fast/'),
     url_test('/firefox/security/', '/firefox/features/independent/'),
     url_test('/firefox/technology/', 'https://developer.mozilla.org/docs/Tools'),
-    url_test('/firefox/sms/{,sent}', '/firefox/'),
```
alexgibson, line 359: Why is this test being removed?

alexgibson (changes requested):
A few nits / questions, but otherwise the front-end changes lgtm. I'll defer to @pmac on the Python bits.

docs/send-to-device.rst
```diff
@@ -72,23 +72,6 @@ The Jinja macro supports parameters as follows (* indicates a required parameter
...
 |    legal_note_email  | Provides a custom legal note for email use.                            | Localizable String.  | _('The intended recipient of the email must have consented.')      |
 +----------------------+------------------------------------------------------------------------+----------------------+--------------------------------------------------------------------+
-|    legal_note_sms    | Provides a custom legal note for SMS or email.                         | Localizable string   | _('SMS service available in select countries only.')               |
-+----------------------+------------------------------------------------------------------------+----------------------+--------------------------------------------------------------------+
 |    spinner_color     | Hex color for the form spinner. Defaults to '#000'.                    | String               | '#fff'                                                             |
 +----------------------+------------------------------------------------------------------------+----------------------+--------------------------------------------------------------------+
 
-
-Geolocation Callback
```
pmac, line 81: Are we sure we won't need the Geo Callback anymore? We do still sometimes restrict mobile products to certain countries.

pmac (changes requested):
Just a couple of tweaks and questions about things that may not need to be removed. Nice work overall. 🍰

bedrock/firefox/templates/firefox/whatsnew/whatsnew-fx80.html
```diff
@@ -6,10 +6,6 @@
 
 {% extends "firefox/whatsnew/base.html" %}
 
-{% block html_attrs %}
```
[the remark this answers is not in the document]
alexgibson: Perhaps this was left over from an experiment? (that would be my guess anyway)

bedrock/base/templates/macros.html
stephaniehobson: I think the change in `bedrock/firefox/views.py` is what was needed to avoid problems. Unless @pmac says different.

docs/send-to-device.rst
stephaniehobson: Right now the only restrictions we have are to only show the email form to locales which have translated emails and that's handled in the python. Since this is dead code now I'd err on the side of removing it but if there's a specific use case we know is coming I can save it.

stephaniehobson:
Updated and pushed to integration tests again.

bedrock/firefox/views.py
```diff
@@ -186,29 +186,29 @@ def sign_attribution_codes(codes):
...
-    # determine if email or phone number was submitted
-    data_type = 'email' if '@' in phone_or_email else 'number'
+    data = {
+        'platform': request.POST.get('platform'),
+        'email': email,
+    }
 
     # populate data type in form data dict
-    data[data_type] = phone_or_email
+    data[data_type] = email
```
pmac, line 205: This line can be removed.

bedrock/firefox/views.py
```diff
@@ -186,29 +186,29 @@ def sign_attribution_codes(codes):
...
 
     # instantiate the form with processed POST data
     form = SendToDeviceWidgetForm(data)
```
pmac, line 208: This form can also be cleaned up to remove the `number` field and the `clean_number` method.

https://github.com/mozilla/bedrock/blob/bbd6f42a9bf022097d83a1758cadec7f0c2fbf0a/bedrock/firefox/forms.py#L11

stephaniehobson:
Updated, integration tests passing.

bedrock/base/templates/macros.html
alexgibson: Seems ok to me, I tested the form end-to-end and got the email ok 👍

docs/send-to-device.rst
alexgibson: Agree, we can still make an XHR request to the geo URL for other reasons. I don't think it's worth keeping this in the S2D widget if it's functionality no longer related to the form 👍

alexgibson (approved):
Looks good to me. I'll hold off merging just in case @pmac wanted a last look at the most recent updates.

pmac (approved):
🍰

PR #9871, opened 2021-01-26
Fix send-to-device success header color

alexgibson:
## Description

Using dark theme colors, the "Your download link was sent" is not currently readable.

- http://localhost:8000/de/firefox/85.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/60.0/whatsnew/all/
- http://localhost:8000/en-US/firefox/whatsnew/all/

## Issue / Bugzilla link
None

## Testing
- [ ] "Your download link was sent" header should be visible on both light and dark color schemes upon entering an email address.


craigcook (approved):
🕶️

Issue #9717, opened 2020-11-20
Refine sticky promo behaviour


alexgibson:
## Description

In https://github.com/mozilla/bedrock/issues/9406 we added a sticky promo to some of the /firefox pages. One minor user issue that came to light during code review was that the promo creates a blind spot once you scroll to the bottom of a page, forcing you to close the promo in order to see what's there. To get around this we made the promo hide beneath the global page footer on scroll, however this interaction could likely be improved.

One suggestion would be to slide the sticky promo back out of view again as the footer comes into view. Scrolling back up would show the promo again once the footer leaves the viewport.

Similarly, on /new, the fact that promo appears immediately on launch really starts to make it look user hostile (because there are 3 download buttons on screen all of a sudden). Should we have the promo appear on scroll, not on load?

---

#### Success Criteria
- [ ] Promo does not obscure content at the bottom of the page without being closed by the user.
- [ ] Promo does not appear until scroll

PR #9874, opened 2021-01-26
Revise UX/Behavior of Sticky Download Promo 

maxxcrawford:
## Description

This PR adds two revisions to the UX of the [sticky download promo](https://protocol.mozilla.org/patterns/molecules/sticky-promo.html):

1. When the footer is in the viewport, the promo box is hidden. 
2. The promo box is not displayed until a user scrolls on the page. 

TODO: 

- [x] Revise/remove tests that are no longer needed 

## Issue / Bugzilla link

#9717

## Testing

*Be sure to delete any `localhost` cookeis you may have*

1. Go to http://localhost:8000/en-US/firefox/new/
2. Scroll down the page some. 
3. Expected: Promo should appear
4. Scroll down to where the footer is in view. 
5. Expected: Promo should dissapear 
6. Scrioll back up to where the footer is no longer in view
7. Expected: Promo should reappear 

Notes: 
- I couldn't find a page that had the promo that was so short that the footer was already in view. It is assumed that the promo would not be displayed. 
- If you dismiss the promo, and continue to scroll (including into and out of the footer), it remains hidden. On refresh/reload, the promo is not loaded (per the current logic)

### DIFF
--- a/media/css/firefox/sticky-promo.scss
+++ b/media/css/firefox/sticky-promo.scss
@@ -73,7 +73,7 @@ $image-path: '/media/protocol/img';
 }
 
 @media #{$mq-md} {
-    .mzp-c-sticky-promo.mzp-a-slide-in {
+    .mzp-c-sticky-promo.mzp-a-slide-in, .mzp-c-sticky-promo.is-displayed {
         display: block;
     }
 }
--- a/media/css/firefox/sticky-promo.scss
+++ b/media/css/firefox/sticky-promo.scss
@@ -73,7 +73,8 @@ $image-path: '/media/protocol/img';
 }
 
 @media #{$mq-md} {
-    .mzp-c-sticky-promo.mzp-a-slide-in, .mzp-c-sticky-promo.is-displayed {
+    .mzp-c-sticky-promo.mzp-a-slide-in,
+    .mzp-c-sticky-promo.is-displayed {
         display: block;
     }
 }
--- a/media/js/firefox/sticky-promo.js
+++ b/media/js/firefox/sticky-promo.js
@@ -5,6 +5,16 @@
 (function() {
     'use strict';
 
+    function isInViewport(element) {
+        // Calculate on each scroll if the footer is in the view port.
+        var elementTop = element.offsetTop;
+        var elementBottom = elementTop + element.offsetHeight;
+        var viewportTop = window.scrollY || window.pageYOffset;
+        var viewportBottom = viewportTop + window.screen.height;
+
+        return elementBottom > viewportTop && elementTop < viewportBottom;
+    }
+
     function onLoad(){
         // Check if promo exists on the page or if smaller than tablet.
         var matchMediaDesktop = window.matchMedia('(min-width: 768px)').matches;
@@ -50,6 +60,19 @@
             var footer = document.querySelector('.c-footer');
             if (footer) {
                 footer.classList.add('is-intersecting-sticky-overlay');
+
+                document.addEventListener('scroll', function () {
+                    // If the footer is in the viewport, fade out the promo.
+                    // Animate it back in when the footer leaves the viewport
+                    // if the user did not dismiss it.
+                    if (isInViewport(footer)) {
+                        promo.classList.replace('mzp-a-slide-in', 'mzp-a-fade-out');
+                    } else if (!promo.classList.contains('user-dismiss')) {
+                        promo.classList.replace('mzp-a-fade-out', 'mzp-a-slide-in');
+                    }
+                }, {
+                    passive: true
+                });
             }
 
             // Set Close Button event
@@ -57,7 +80,11 @@
 
             stickyBtnClose.addEventListener('click', function(){
                 StickyPromo.setCookie(STICKY_PROMO_COOKIE_ID);
+                promo.classList.add('user-dismiss');
             });
+
+
+
         };
 
         // Check on page load
--- a/media/js/firefox/sticky-promo.js
+++ b/media/js/firefox/sticky-promo.js
@@ -53,8 +53,14 @@
         };
 
         StickyPromo.show = function (){
-        // Open promo
-            Mzp.StickyPromo.open();
+            var stickyPromoShowInit = false;
+            document.addEventListener('scroll', function () {
+                if (!stickyPromoShowInit) {
+                    // Open promo
+                    Mzp.StickyPromo.open();
+                    stickyPromoShowInit = true;
+                }
+            });
 
             // Add modifier class to the footer to make sure the language selection drop-down is not obscured by the sticky promo
             var footer = document.querySelector('.c-footer');
@@ -82,9 +88,6 @@
                 StickyPromo.setCookie(STICKY_PROMO_COOKIE_ID);
                 promo.classList.add('user-dismiss');
             });
-
-
-
         };
 
         // Check on page load
--- a/media/js/firefox/sticky-promo.js
+++ b/media/js/firefox/sticky-promo.js
@@ -93,7 +93,7 @@
             var footer = document.querySelector('.c-footer');
             if (footer) {
                 footer.classList.add('is-intersecting-sticky-overlay');
-                document.addEventListener('scroll', checkForFooterOnScroll, true);
+                document.addEventListener('scroll', checkForFooterOnScroll, false);
             }
 
             // Set Close Button event
--- a/media/js/firefox/sticky-promo.js
+++ b/media/js/firefox/sticky-promo.js
@@ -83,16 +83,17 @@
                 // Animate it back in when the footer leaves the viewport
                 // if the user did not dismiss it.
                 if (isInViewport(footer)) {
-                    promo.classList.replace('mzp-a-slide-in', 'mzp-a-fade-out');
+                    promo.classList.remove('mzp-a-slide-in');
+                    promo.classList.add('mzp-a-fade-out');
                 } else if (!promo.classList.contains('user-dismiss')) {
-                    promo.classList.replace('mzp-a-fade-out', 'mzp-a-slide-in');
+                    promo.classList.remove('mzp-a-fade-out');
+                    promo.classList.add('mzp-a-slide-in');
                 }
             }, 100);
 
             // Add modifier class to the footer to make sure the language selection drop-down is not obscured by the sticky promo
             var footer = document.querySelector('.c-footer');
             if (footer) {
-                footer.classList.add('is-intersecting-sticky-overlay');
                 document.addEventListener('scroll', checkForFooterOnScroll, false);
             }
 
--- a/media/js/firefox/sticky-promo.js
+++ b/media/js/firefox/sticky-promo.js
@@ -1,9 +1,25 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
 (function() {
     'use strict';
+
+    // https://davidwalsh.name/javascript-debounce-function
+    function debounce(func, wait, immediate) {
+        var timeout;
+        return function() {
+            var context = this,
+                args = arguments;
+            var later = function() {
+                timeout = null;
+                if (!immediate) {func.apply(context, args);}
+            };
+            var callNow = immediate && !timeout;
+            clearTimeout(timeout);
+            timeout = setTimeout(later, wait);
+            if (callNow) {func.apply(context, args);}
+        };
+    }
 
     function isInViewport(element) {
         // Calculate on each scroll if the footer is in the view port.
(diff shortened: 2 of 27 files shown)


media/css/firefox/sticky-promo.scss
```diff
@@ -73,7 +73,7 @@ $image-path: '/media/protocol/img';
 }
 
 @media #{$mq-md} {
-    .mzp-c-sticky-promo.mzp-a-slide-in {
+    .mzp-c-sticky-promo.mzp-a-slide-in, .mzp-c-sticky-promo.is-displayed {
```
alexgibson, line 76: nit: selectors on new lines

media/js/firefox/sticky-promo.js
```diff
@@ -43,20 +53,40 @@
         };
 
         StickyPromo.show = function (){
-        // Open promo
-            Mzp.StickyPromo.open();
+            var stickyPromoShowInit = false;
+            document.addEventListener('scroll', function () {
+                if (!stickyPromoShowInit) {
```
alexgibson, line 58: Seeing as this code only runs once, might it make sense to remove the scroll event handler once the promo is shown?

media/js/firefox/sticky-promo.js
```diff
@@ -43,20 +53,40 @@
...
+                    stickyPromoShowInit = true;
+                }
+            });
 
             // Add modifier class to the footer to make sure the language selection drop-down is not obscured by the sticky promo
             var footer = document.querySelector('.c-footer');
             if (footer) {
                 footer.classList.add('is-intersecting-sticky-overlay');
+
+                document.addEventListener('scroll', function () {
```
alexgibson, line 70: This event fires a lot to check for the footer. Might it be possible to throttle / debounce the handler so it only needs to check every 100ms or so? Alternatively, it might be worth considering [Intersection Observer](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) perhaps if performance is a concern?

alexgibson (changes requested):
In terms of interaction I think this looks / feels really great as a website visitor, nice work! 💯 

I left a couple of comments / suggestions related to performance, but up to you to do what you think is best here.

As for the tests, you'll probably need to remove the custom [wait_for_page_to_load](https://github.com/mozilla/bedrock/blob/master/tests/pages/firefox/new/sticky_promo.py#L17-L22) logic, as the promo no longer appears on page load. Instead it might make sense for the tests to:

1.) Load a page as normal.
2.) Scroll down the page to some element.
3.) Wait for the promo to appear and assert.
4.) Click close.
5.) Wait for the promo to disappear and assert.

A couple of tests on key / high traffic pages is likely enough for this promo imho (e.g. /firefox/new, /firefox), as it's largely shared code. I'd say there's no need to repeat the same test on every page where it is featured.

If I can help out with code/advice for the tests above, just let me know!

media/js/firefox/sticky-promo.js
maxxcrawford: @alexgibson Great call out. Looking at the docs for  Intersection Observer, it looks like it is lacking any IE support. Should that be avoided?

alexgibson: @maxxcrawford I'll defer to you on what you think is ultimately the most appropriate thing to use here. But as long as we use feature detection to avoid errors on older browsers, then we should be ok. A couple of options might be:

a) Make Intersection Observer a requirement for user agents that see the sticky promo (IE users wouldn't get it).
b) Fallback to a scroll event handler for browsers that don't support the newer method (wider feature coverage, but more complex).

Personally I think a) might be ok here imho. All modern browsers support Intersection Observer today, so as long as we consider the promo like a form of progressive enhancement, then IE visitors just get the regular page and be none the wiser.

media/js/firefox/sticky-promo.js
```diff
@@ -53,14 +53,14 @@
...
-            var stickyPromoShowInit = false;
-            document.addEventListener('scroll', function () {
-                if (!stickyPromoShowInit) {
-                    // Open promo
-                    Mzp.StickyPromo.open();
-                    stickyPromoShowInit = true;
-                }
-            });
+
+            function openOnScroll(){
```
maxxcrawford, line 57: @alexgibson Revised the logic to turn itself off after firing once.

media/js/firefox/sticky-promo.js
maxxcrawford: Debounce added! I used [the function](https://davidwalsh.name/javascript-debounce-function) from @darkwing. 🙌

tests/functional/firefox/test_home.py
```diff
@@ -16,7 +16,9 @@ def test_download_menu_list_displays(base_url, selenium):
 
 
 @pytest.mark.nondestructive
-@pytest.mark.skip_if_not_firefox(reason='FB Container link shown only to Firefox users.')
-def test_fb_container_fx(base_url, selenium):
+def test_sticky_promo(base_url, selenium):
```
maxxcrawford, line 19: There was not a sticky promo test on /firefox, so I added it here.

tests/pages/firefox/home.py
```diff
@@ -26,3 +34,18 @@ def fb_container_is_displayed(self):
     def browser_menu_list(self):
         el = self.find_element(*self._browser_menu_list_locator)
         return MenuList(self, root=el)
+
+    def init_promo(self):
+        assert not self.is_promo_displayed, 'Promo detail is not displayed'
+        # scroll down page to trigger promo to display
+        self.scroll_element_into_view(*self._page_promise_content_locator)
+        self.wait.until(expected.visibility_of_element_located(
+            self._sticky_promo_modal_content_locator))
```
maxxcrawford, line 43: @alexgibson This function checks that it isn't visible to start and then scrolls down the page to an element (but not so far as in the footer) and then waits for the promo to become visible.

tests/functional/firefox/new/test_sticky_promo.py
```diff
@@ -10,6 +10,7 @@
 @pytest.mark.nondestructive
 def test_sticky_promo(base_url, selenium):
     page = StickyPromoPage(selenium, base_url).open()
+    page.init_promo()
     assert page.promo.is_displayed
     page.promo.close()
     assert not page.promo.is_displayed
```
maxxcrawford, line 16: This matches your original plan: 
- Confirm it isn't there
- Scroll down the page
- See it, comfirm
- Dismiss it, confirm 

🚀

tests/functional/firefox/test_home.py
maxxcrawford: I pushed this as it but received an error when running the test locally: 

```
selenium.common.exceptions.ElementClickInterceptedException: 
Message: Element <button class="mzp-c-sticky-promo-close" type="button"> is not clickable at point (1266,407) because another element <html class="osx x86 js is-firefox is-modern-browser mzp-has-sticky-navigation loaded"> obscures it
```

Looking at the screenshot from the generated report, it doesn't look obscured. Could you please help me debug? 

image

maxxcrawford (commented):
@alexgibson Self-reviewed the new test. Please help me solve the obscured element issue below when you can.

tests/pages/firefox/home.py
alexgibson: I think the problem here is the test is waiting for *any* portion of the sticky promo to be visible in the viewport. At that point, it then tries to click the close button (which can still be off screen because the promo is animating, hence the error message).

Using the existing method of checking for the `is_displayed` class is probably going to be more consistent:

```
promo = self.find_element(*self._sticky_promo_modal_content_locator)
self.wait.until(lambda s: 'is-displayed' in promo.get_attribute('class'))
```

tests/pages/firefox/home.py
```diff
@@ -18,6 +20,12 @@ class FirefoxHomePage(BasePage):
     # facebook container extention link visiability
     _facebook_container_link_locator = (By.ID, 'test-fbc')
 
+    # floating/sticky box - download link
+    _sticky_promo_modal_content_locator = (By.CSS_SELECTOR, '.mzp-c-sticky-promo.mzp-a-slide-in')
```
alexgibson, line 24:
```suggestion
    _sticky_promo_modal_content_locator = (By.CSS_SELECTOR, '.mzp-c-sticky-promo')
```

media/js/firefox/sticky-promo.js
```diff
@@ -43,20 +69,39 @@
...
+                } else if (!promo.classList.contains('user-dismiss')) {
+                    promo.classList.replace('mzp-a-fade-out', 'mzp-a-slide-in');
+                }
+            }, 100);
 
             // Add modifier class to the footer to make sure the language selection drop-down is not obscured by the sticky promo
             var footer = document.querySelector('.c-footer');
             if (footer) {
                 footer.classList.add('is-intersecting-sticky-overlay');
+                document.addEventListener('scroll', checkForFooterOnScroll, true);
```
alexgibson, line 96: Any specific reason why `useCapture` is `true` here?

tests/pages/firefox/home.py
maxxcrawford: Awesome. Revised all the tests to leverage this.

tests/pages/firefox/home.py
maxxcrawford: Removed across all the tests!

maxxcrawford:
@alexgibson I made all the nit revisions and pushed to `run-integration-tests`. See [results](https://gitlab.com/mozmeao/www-config/-/pipelines/253327580)

media/js/firefox/sticky-promo.js
```diff
@@ -43,20 +69,39 @@
...
+            }
+
+            document.addEventListener('scroll', openOnScroll, false);
+
+            var checkForFooterOnScroll = debounce(function() {
+                // If the footer is in the viewport, fade out the promo.
+                // Animate it back in when the footer leaves the viewport
+                // if the user did not dismiss it.
+                if (isInViewport(footer)) {
+                    promo.classList.replace('mzp-a-slide-in', 'mzp-a-fade-out');
```
alexgibson, line 86: It looks like Internet explorer doesn't support `replace()` on `classList`, so you might have to fallback to using `remove()` and `add()` as separate calls.

media/js/firefox/sticky-promo.js
```diff
@@ -43,20 +69,39 @@
...
+                } else if (!promo.classList.contains('user-dismiss')) {
+                    promo.classList.replace('mzp-a-fade-out', 'mzp-a-slide-in');
```
alexgibson, line 88: Same as above

media/js/firefox/sticky-promo.js
```diff
@@ -43,20 +69,39 @@
...
+                }
+            }, 100);
 
             // Add modifier class to the footer to make sure the language selection drop-down is not obscured by the sticky promo
             var footer = document.querySelector('.c-footer');
             if (footer) {
                 footer.classList.add('is-intersecting-sticky-overlay');
```
alexgibson, line 95: Do we need this on the footer still if we're hiding the promo on scroll? Could we remove the line of CSS related to this also? https://github.com/mozilla/bedrock/blob/master/media/css/protocol/components/_footer.scss#L7-L11

tests/pages/firefox/browsers/compare.py
```diff
@@ -44,7 +31,3 @@ def secondary_download_button(self):
     def browser_menu_list(self):
         el = self.find_element(*self._browser_menu_list_locator)
         return MenuList(self, root=el)
-
-    @property
-    def promo(self):
```
alexgibson, line 49: It looks like the [test run](https://gitlab.com/mozmeao/www-config/-/jobs/1016570576) failed becuase you still need to remove the test related to this page object

```
E       AttributeError: 'BrowserComparisonPage' object has no attribute 'promo'
```

tests/pages/firefox/features/feature.py
```diff
@@ -14,18 +14,24 @@ class FeaturePage(BasePage):
     _URL_TEMPLATE = '/{locale}/firefox/features/{slug}/'
 
     _download_button_locator = (By.ID, 'features-header-download')
-    _sticky_promo_modal_content_locator = (By.CSS_SELECTOR, '.mzp-c-sticky-promo.mzp-a-slide-in')
 
-    def wait_for_page_to_load(self):
-        el = self.find_element(By.TAG_NAME, 'html')
-        self.wait.until(lambda s: 'loaded' in el.get_attribute('class'))
+    # Used as a scroll target to move down the page, to trigger the sticky promo element
```
alexgibson, line 18: Up to you - but this is a relatively low impact page. If you wanted to remove the test from here also I think that would probably be fine. Testing the promo on /firefox/ and /firefox/new/ should be enough coverage I think.

alexgibson (changes requested):
Noticed an error in IE, and a suggested fix for the tests. But otherwise this is all looking good!

media/js/firefox/sticky-promo.js
maxxcrawford: Great call. I'll remove this hotfix too.

tests/pages/firefox/features/feature.py
maxxcrawford: Removing.

tests/pages/firefox/browsers/compare.py
maxxcrawford: Double checked all files for the instance of this, and this was the only extra one. 🧹

alexgibson:
Tests look good! 🌮 https://gitlab.com/mozmeao/www-config/-/pipelines/253799385

alexgibson (approved):

Issue #9875, opened 2021-01-27
Redesign Send-to-Device Widget


maxxcrawford:
## Description

Epic #9864

Task: Design a new widget -- it should include email, QR code, and app-store badges

Notes: 
- Include error/success states
- Consider L10N length/copy
- Consult @pmac about how to improve QR code generation
- Confirm we can measure performance between QR Code and Email 
- Reduce size of QR code

---

#### Success Criteria
- [ ] An element contributing to the definition of done for the card
- [ ] Tasks should be represented here as something like "Approved designs attached to this issue"

## Screenshots

QR Code Widget in Hero: (https://www.mozilla.org/it/firefox/mobile/)
![image](https://user-images.githubusercontent.com/2692333/106055927-aa72e100-60b3-11eb-8d9c-4cd326aac4ac.png)

QR Code Widget (above footer) (https://www.mozilla.org/it/firefox/mobile/) 
<img width="466" alt="image" src="https://user-images.githubusercontent.com/2692333/106055860-94fdb700-60b3-11eb-9f00-a2710b9057d2.png">

SMS/Email Widget in Hero: (https://www.mozilla.org/en-US/firefox/mobile/)
![image](https://user-images.githubusercontent.com/2692333/106056017-ca0a0980-60b3-11eb-82ff-319061e5d909.png)

SMS/Email Widget in Hero: (https://www.mozilla.org/en-US/firefox/mobile/)
<img width="582" alt="image" src="https://user-images.githubusercontent.com/2692333/106056045-d42c0800-60b3-11eb-8cb6-f92aea3c7ece.png">

pmac:
I should be able to provide a template helper method that can take some data (a URL most likely) and replace it with SVG data that will be embeded into the page using the [qrcode](https://pypi.org/project/qrcode/) library.

stephaniehobson:
List of pages currently using the S2D widget: 

http://localhost:8000/en-US/firefox/mobile/
http://localhost:8000/en-US/firefox/mobile/get-app/
http://localhost:8000/en-US/firefox/whatsnew/all/
http://localhost:8000/en-US/firefox/79.0/whatsnew/all/
http://localhost:8000/en-US/firefox/80.0/whatsnew/all/
http://localhost:8000/en-US/firefox/81.0/whatsnew/all/
http://localhost:8000/en-US/firefox/welcome/4/
http://localhost:8000/en-US/firefox/welcome/5/
[...]

PR #9877, opened 2021-01-28
Add QR Code generator template helper function

pmac:
Re #9875

## Testing

You can try it locally by simply installing the new requirements and adding a call to a template:

1. `pip install -r requirements/base.txt`
2. Edit a template and add a call to the thing: `{{ qrcode('https://bedrock.rtfd.org') }}`
3. Run the local server and see the QR Code on the page

If you're wanting to use the Docker setup then you'll need to build new images `make build` to get the new deps.


alexgibson:
I might have a use-case for this, so picking up review :)

alexgibson (approved):
I tested this locally and it works great 💯 

I wasn't quite sure if the `mm` units would work well in older browsers, but the QR code I generated locally worked well in both IE9/10 which is probably more than adequate. Nice work! r+

PR #9886, opened 2021-01-30
Move Yandex geolocation check to server (Fix #9883)

### LINKED ISSUE #9883: RU users see outdated /new page
## Description

At the moment all users with RU as their language see an out dated version of the /new page. The old template is used because it has the logic for showing the Yandex page build into it when the geolocation indicates it's appropriate. RU is an active language for the long form /new page which has a higher conversion rate, it would be better to display that page to both sets of users, at the very least it should be possible to display it to RU users outside of that geolocation.

---

#### Success Criteria
- [x] Yandex specific image and copy are preserved and served to RU language users in RU geolocation
- [x] Download button for Yandex content continues to go to Yandex partnership page
- [x] Long form /new page is displayed to RU language users outside of RU geolocation
- [x] Update tests

stephaniehobson:
## Description

- moved Yandex content to its own page, using protocol templates
- removed js display logic and moved it to the view
- updated analytics to work in new structure

## Issue / Bugzilla link

Fix #9883

## Testing

http://localhost:8000/ru/firefox/new/?geo=ru

Check `window.dataLayer` for the analytics push.


bedrock/firefox/views.py
```diff
@@ -789,13 +789,14 @@ def get_context_data(self, **kwargs):
...
         variant = self.request.GET.get('v', None)
+        geo = self.request.GET.get('geo', None)
 
         # ensure variant matches pre-defined value
         if variant not in self.variations:
             variant = None
 
-        if locale == 'ru' and switch('firefox-yandex'):
-            template = 'firefox/new/trailhead/download-yandex.html'
+        if locale == 'ru' and geo == 'ru' and switch('firefox-yandex'):
```
stephaniehobson, line 798: @pmac I could use some direction here for how to do the geolocation server side. We don't need to redirect, just use a different template so I wasn't sure how to use the GeoRedirectView or even if it's possible to use it together with a L10nTemplateView.

pmac: The issue is that for geo to work on the server side we have to turn off the CDN cache because all requests have to make it to the server since we don't know their country of origin until it reaches the server. That's why we've restricted server side geo to redirects so that we don't have to turn off the cache for an HTML page so that it stays fast. What we'd normally do is to use the redirect view and have two other URLs that would load the template you want. Could even just be the same URL but with a query param.

stephaniehobson: I don't think that's a good idea for the download page. I guess it's back to client side detection.

PR #9896, opened 2021-02-02
Update Yandex to current /new template (Fix #9883)

### LINKED ISSUE #9883: RU users see outdated /new page

stephaniehobson:
## Description

- make Yandex content an include
- add include to desktop/download template with conditional for RU lang
  - hidden by default
  - revealed by the same JS script as the old page
  - small CSS adjustments for the new content, also included conditionally
- remove old Yandex page which relied on the trailhead/download template
- remove tests for platform modal (no longer on page)
- update tests with new locators
- don't want to show promo on yandex page, hid it for RU because that's easiest.

## Issue / Bugzilla link

Fix #9883 

## Testing

http://localhost:8000/ru/firefox/new/?geo=ru


stephaniehobson:
Ready for review but waiting for okay from stakeholders to launch.

bedrock/firefox/tests/test_views.py
```diff
@@ -508,19 +508,25 @@ def test_yandex_scene_1(self, render_mock):
         req = RequestFactory().get('/firefox/new/')
         req.locale = 'ru'
         view = views.NewView.as_view()
-        view(req)
+        response = view(req)
         template = render_mock.call_args[0][1]
-        assert template == ['firefox/new/trailhead/download-yandex.html']
+        assert template == ['firefox/new/desktop/download.html']
+        doc = pq(response.content)
+        robots = doc('main[class*="t-yandex"]')
```
stephaniehobson, line 515: @pmac The switch used to control the template but now it controls an include on the page. I'd like to update the tests to look at the response but this test is failing with a `lxml.etree.ParserError: Document is empty` Can you spot my mistake?

pmac: I think the issue is that you're mocking the render function, which means it will not be returning a rendered page, but rather a mock object.

bedrock/firefox/tests/test_views.py
```diff
@@ -555,6 +545,41 @@ def test_experiment_redirect_automation_param(self, render_mock):
         assert 'cache-control' not in resp
 
 
+class TestFirefoxYandex(TestCase):
+    @patch.dict(os.environ, SWITCH_FIREFOX_YANDEX='True')
+    @patch.object(views, 'ftl_file_is_active', lambda *x: True)
+    def test_yandex_show_to_ru(self):
```
stephaniehobson, line 551: @pmac I've progressed to a failure because my request is returning the wrong content. Progress! 

The content is only on the page if the local is `ru`. If I look at the doc being the `lang` attribute on the `<html>` element is `en`... but the copy is in Russian. Not sure if that's a symptom of a different oddity.

pmac: I think it's because you're switching things in the template based on the `LANG` variable, which is different from the `request.lang` property. You should be able to activate a language with:

```python
with self.activate('ru'):
    response = view(req)
```

Replace line 555 with that and it may work. Your other option is to not make these conditionals in the template, and instead make a new yandex template that inherits from this one and that overrides the bits that you want. It'd be a bit easier to test I'm guessing.

stephaniehobson:
Okay to launch.

bedrock/firefox/templates/firefox/new/desktop/download.html
```diff
@@ -63,6 +73,11 @@ <h2 class="c-sub-navigation-title">{{ ftl('sub-navigation-firefox-for-desktop',
 {% endblock %}
 
 {% block content %}
+
+  {% if show_yandex %}
+    {% include 'firefox/new/desktop/includes/yandex.html' %}
+  {% endif %}
```
alexgibson, line 79: Would changing this `endif` to an `else` help make the markup a bit cleaner? (avoid two `<main>` elements, `<h1>` etc in the source code?

bedrock/firefox/views.py
```diff
@@ -787,16 +787,13 @@ def get_context_data(self, **kwargs):
...
     def get_template_names(self):
-        locale = l10n_utils.get_locale(self.request)
         variant = self.request.GET.get('v', None)
 
         # ensure variant matches pre-defined value
         if variant not in self.variations:
             variant = None
 
-        if locale == 'ru' and switch('firefox-yandex'):
-            template = 'firefox/new/trailhead/download-yandex.html'
```
alexgibson, line 798: Can we remove this template now?

bedrock/firefox/templates/firefox/new/desktop/includes/yandex.html
```diff
@@ -0,0 +1,70 @@
...
+  </section>
+
+  {% call call_out_compact(
+    title='Загрузите последнюю версию браузера Firefox.',
+    class='mzp-t-product-firefox mzp-t-firefox')
+  %}
+    {{ download_firefox(alt_copy='Скачать Firefox без яндекса', dom_id='download-secondary', locale_in_transition=True, download_location='secondary cta') }}
+  {% endcall %}
+
+  <!-- TODO: hide promo -->
```
alexgibson, line 68: Is this still todo?

alexgibson (changes requested):
A couple of nits / questions but otherwise this seems to work as indended 👍 

I found that mixing the content into a single template (vs extending & overriding blocks) makes the conditional logic a little harder to follow / reason about, but if this makes overall maintenance easier then I'm not against it. r+wc

bedrock/firefox/templates/firefox/new/desktop/download.html
alexgibson: Oh right, we can’t do this as it’s all client side. Never mind :/

bedrock/firefox/templates/firefox/new/desktop/includes/yandex.html
stephaniehobson: It's do done.

bedrock/firefox/views.py
stephaniehobson: ✅

bedrock/firefox/templates/firefox/new/desktop/download.html
stephaniehobson: Because the geoLocation swap needs to be done client side, both of these chuncks need to go on the page.

I've re-jigged it not to use two `<main>` elements. Only one `<h1>` will be visible at a time. I don't _think_ it's worth removing the hidden one from the DOM since the search engines will either pick up on it before the page renders or treat it as hidden after it renders.

stephaniehobson:
Re-jigged to override the template instead of combining them.

stephaniehobson:
Integration test run: https://gitlab.com/mozmeao/www-config/-/pipelines/258438894

alexgibson (approved):
💯 r+

PR #9903, opened 2021-02-04
Adding dockerhub creds to circleci builds

duallain:
## Description

Adding our dockerhub creds to pull images, so we can avoid getting rate limited.

## Testing
The build failed when I put bad creds into the env variables, and passed when good creds were added.
Checked that the username/password wasn't echo'd in the logs, and that seems to be true for both jobs.


pmac (approved):
🍍

PR #9917, opened 2021-02-08
Fix common strings on privacy policies

craigcook:
## Description
Adds reference to `privacy/index` Fluent file to the legal doc view so all privacy pages can get those strings in the sidebar.

## Testing
http://localhost:8000/privacy/firefox/


pmac (approved):
☃️

PR #9919, opened 2021-02-09
Migrate Mozilla home page to Fluent (Fixes #9911)

### LINKED ISSUE #9911: Migrate legacy home page to Fluent
## Description

https://www.mozilla.org/es-ES/

I had put off filing this issue in the hope that this page would be redesigned, but it looks like that's not happening [at least in the short term](https://github.com/mozilla/bedrock/issues/9880). Holding off migrating these strings is now preventing removing the entire legacy .lang system for bedrock, and it's been three years since filing https://github.com/mozilla/bedrock/issues/5824.

Let's migrate these strings even though they are old, so we can move forward.

---

#### Success Criteria
- [ ] Page is live in production using Fluent along with existing migrations.

alexgibson:
## Description
http://localhost:8000/es-ES/

## Issue / Bugzilla link
#9911

## Testing
```
./manage.py l10n_update
```

### DIFF
--- a/bedrock/mozorg/templates/mozorg/home/home.html
+++ b/bedrock/mozorg/templates/mozorg/home/home.html
@@ -6,18 +6,10 @@
 
 {% extends "base-protocol-mozilla.html" %}
 
-{% add_lang_files "mozorg/home/index-quantum" %}
-
 {% block gtm_page_id %}data-gtm-page-id="Homepage"{% endblock %}
 
-{% block page_title %}
-{{ _('Internet for people, not profit') }}
-{% endblock %}
-
-{% block page_desc %}
-{{ _('Did you know? Mozilla — the maker of Firefox — fights to keep the Internet a global public resource, open and
-accessible to all.') }}
-{% endblock %}
+{% block page_title %}{{ ftl('home-internet-for-people-not-profit') }}{% endblock %}
+{% block page_desc %}{{ ftl('home-did-you-know-mozilla-the-maker') }}{% endblock %}
 
 {% block body_class %}lang-{{ LANG }}{% endblock %}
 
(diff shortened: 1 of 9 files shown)


bedrock/mozorg/templates/mozorg/moss/base.html
```diff
@@ -4,8 +4,6 @@
 
 {% extends "base-protocol-mozilla.html" %}
 
-{% add_lang_files "mozorg/moss/index" %}
```
alexgibson, line 7: Not related to this PR, but removing this line as no longer needed

bedrock/mozorg/templates/mozorg/home/home.html
```diff
@@ -59,85 +51,66 @@
...
     image_url='img/home/dino.svg'
     )}}
 
     <section class="c-column-container">
       <div class="c-column">
         <div class="c-column-content">
-          <h2>{{ _('Our impact') }}</h2>
+          <h2>{{ ftl('home-our-impact') }}</h2>
           <p>
             {% if l10n_has_tag('home-update-20171116') %}
```
craigcook, line 75: Does `l10n_has_tag` still work with Fluent? I think this should be `ftl_has_messages`. 

BUT given how old these strings are maybe we don't need any conditional here and just go with the latest string and delete the other two.

craigcook (changes requested):

bedrock/mozorg/templates/mozorg/home/home.html
alexgibson: Good catch, will update 👍

alexgibson:
@craigcook thanks for the review, removed the old conditional

craigcook (approved):

Issue #7673, opened 2019-09-06
Remove jQuery dependency for send-to-device.js


alexgibson:
## Description

The send-to-device widget is still being used in modern templates, but relies on jQuery. We should replace those function calls with native equivalents.

---

#### :yellow_heart: Success Criteria :yellow_heart:
- [x] send-to-device.js no longer relies on jQuery.

#### Tasks
- [x] Remove dependency.
- [x] Update tests where applicable.

Patil2099:
@alexgibson Can I work on this issue?

vinay-deshmukh:
@alexgibson I have submitted a PR for this issue.

PR #9928, opened 2021-02-16
Remove jQuery from JS unit tests

alexgibson:
## Description
Remove jQuery syntax from JS unit tests

## Issue / Bugzilla link
N/A

## Testing
- [ ] Tests are still passing in CI

### DIFF
--- a/tests/unit/spec/base/core-datalayer.js
+++ b/tests/unit/spec/base/core-datalayer.js
@@ -10,15 +10,14 @@ describe('core-datalayer.js', function() {
 
     describe('pageHasDownload', function() {
 
-        afterEach(function() {
-            $('.download').remove();
-        });
-
         it('will return "true" when download button is present on page.', function() {
             var downloadMarkup = '<a class="download" href="#" data-link-type="download" data-download-os="Desktop">';
 
-            $(downloadMarkup).appendTo('body');
+            document.body.insertAdjacentHTML('beforeend', downloadMarkup);
             expect(Mozilla.Analytics.pageHasDownload()).toBe('true');
+
+            var content = document.querySelector('.download');
+            content.parentNode.removeChild(content);
         });
 
         it('will return "false" when download button is not present on page.', function() {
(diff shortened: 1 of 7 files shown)


tests/unit/spec/base/core-datalayer.js
```diff
@@ -10,15 +10,14 @@ describe('core-datalayer.js', function() {
 
     describe('pageHasDownload', function() {
 
-        afterEach(function() {
-            $('.download').remove();
-        });
-
```
maxxcrawford, line 16: Should the `$('.download').remove();` be replaced with some sort of `...parentNode.removeChild(content)` call?

tests/unit/spec/base/core-datalayer.js
```diff
@@ -28,23 +27,24 @@ describe('core-datalayer.js', function() {
 
     describe('pageHasVideo', function() {
 
-        afterEach(function() {
-            $('#htmlPlayer').remove();
-            $('#iframePlayer').remove();
-        });
-
```
maxxcrawford, line 35: Should both of the `.remove()` calls be replaced?

tests/unit/spec/base/mozilla-fxa.js
```diff
@@ -9,21 +9,21 @@ describe('mozilla-fxa.js', function() {
     describe('FxaState.applyStateToBody', function() {
 
         beforeEach(function() {
-            $('body').attr('class', 'js state-fxa-default');
+            document.body.setAttribute('class', 'js state-fxa-default');
```
maxxcrawford, line 12: Would swapping `setAttribute` for `classList.add` make more sense here?

tests/unit/spec/base/send-to-device.js
```diff
@@ -85,8 +86,8 @@ describe('send-to-device.js', function() {
 
         it('should handle success', function() {
 
-            spyOn($, 'post').and.callFake(function () {
-                var d = $.Deferred();
+            spyOn(window.$, 'post').and.callFake(function () {
+                var d = window.$.Deferred();
```
maxxcrawford, line 90: Is it okay for the `$` to remain here?

maxxcrawford (changes requested):
Blocking with some questions. This may require no changes.  LMK!

tests/unit/spec/base/core-datalayer.js
alexgibson: Only the first test in this block modifies the DOM, so I moved the cleanup code to the test in question, else this will throw an error when the second test tries to remove an element that's not there.

tests/unit/spec/base/core-datalayer.js
alexgibson: Same as above, the two tests don't share any common cleanup code, so I moved it to each respective test.

tests/unit/spec/base/mozilla-fxa.js
alexgibson: Good suggestion 👍

tests/unit/spec/base/send-to-device.js
alexgibson: For now yeah, until we fix https://github.com/mozilla/bedrock/issues/7673

tests/unit/spec/base/core-datalayer.js
alexgibson: Same as above, the third test doesn't modify the DOM, so I moved the code to the relevant tests where cleanup is needed.

alexgibson:
@maxxcrawford thanks for the review, updated

maxxcrawford (approved):

PR #9932, opened 2021-02-17
WNP 86 for /de/ and /fr/ (Fixes #9926)

### LINKED ISSUE #9926: WNP 86 for DE / FR
## Description

- /de/ Firefox Mobile promo via QRCode (adjust URL forthcoming)
- /fr/ Show the content from https://www.mozilla.org/fr/firefox/81.0/whatsnew/all/

---

#### Success Criteria
- [ ] Content is live before 23 Feb.

alexgibson:
## Description
- http://localhost:8000/de/firefox/86.0/whatsnew/all/ (Firefox Mobile promo with QRCode)
- http://localhost:8000/fr/firefox/86.0/whatsnew/all/ (Template from WNP81 with Send to Device)

## Issue / Bugzilla link
#9926

## Testing
- [x] Scanning QRCode should redirect to app store / play store.

Demo
- https://www-demo2.allizom.org/de/firefox/86.0/whatsnew/all/
- https://www-demo2.allizom.org/fr/firefox/86.0/whatsnew/all/

### DIFF
--- a/bedrock/firefox/templates/firefox/whatsnew/whatsnew-mobile-qrcode-de.html
+++ b/bedrock/firefox/templates/firefox/whatsnew/whatsnew-mobile-qrcode-de.html
@@ -1,0 +1,75 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/whatsnew/base.html" %}
+
+{% block page_title %}{{ ftl('whatsnew-page-title') }}{% endblock %}
+{% block page_desc %}{{ ftl('whatsnew-page-description') }}{% endblock %}
+
+{#- This will appear as <meta property="og:description"> which can be used for social share -#}
+{% block page_og_desc %}{{ ftl('whatsnew-page-description') }}{% endblock %}
+
+{#- Override <meta property="og:url"> for social share -#}
+{% block page_og_url %}{{ url('firefox.mobile.index') }}{% endblock %}
+
+{% block body_id %}firefox-whatsnew-mobile-qr-code-de{% endblock %}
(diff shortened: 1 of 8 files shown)


craigcook (approved):
r+ but with some additional thoughts...

I'm not sure how I feel about showing the 81 template to FR, mostly because it means we won't be able to delete it for a while longer (we haven't done a great job cleaning up old WNPs, that's on me). 

The other option would be making a new self-contained copy of the template and maybe hard-coding the FR strings. That's a bit more effort but might incur less technical debt.

The 81 template is also the older design and if we were to go the route of making a new one for FR we could use the current design and recycle the old copy. However that might mean another round of stakeholder review and I know we're racing the clock already.

I leave it to you if you want to put more time into the FR template and I can review any updates, but if the juice isn't worth the squeeze we can merge this as-is and call it done. 🍇

craigcook:
I've shamelessly stolen your functional tests for the EN page in https://github.com/mozilla/bedrock/pull/9936 so whichever one of these PRs merges first the other will need to rebase, hopefully without too many conflicts.

alexgibson:
> I'm not sure how I feel about showing the 81 template to FR, mostly because it means we won't be able to delete it for a while longer (we haven't done a great job cleaning up old WNPs, that's on me).

Yeah I've given this some thought also. After talking with the EU team on how they would like to handle WNPs this year, it sounds like they want to move to a system of more "evergreen" German and French pages which they can rotate, rather than create everything from scratch every 4 weeks. I think this makes good sense now we're no longer localizing WNPs widely, which is why I've been resisting the urge to label everything with a version number (which has been the typical fashion up until now).

As for the WNP 81 template, it's unlikely this will become an evergreen template. Right now, it's just to avoid showing the french locale the same thing again as last release (until we can build up more content). I don't think it's worth duplicating everything according to version number for the reasons above, so If it means we can't delete it for a while longer, then I'm OK with that.

Issue #9399, opened 2020-09-14
Build a new evergreen download page


hoosteeno:
## Description

In #9224 we are designing an evergreen download page that can serve the majority of our locales where we don't have on-demand localization. That issue has extensive conversation about the design. This issue is for coding the new experience.

---

#### Success Criteria
- [ ] Code is merged and ready for localization

PR #9934, opened 2021-02-18
Evergreen download page (#9399)

stephaniehobson:
## Description

The goal of this PR is to: 
1) remove the number of templates we have to maintain for the download page
2) activate as many locales as we can for the download page

Changes made:
- combine platform and trailhead pages into "basic" page template
  - activated by either new/download.ftl or new/platform.ftl
  - can be viewed with `xv` query string param for translators
- simplify download page
    - remove trailhead product spotlights from download and thanks pages
    - remove join and platform modals
    - remove unused strings and assets
- add views for platform pages
    - update associated links in content

## Issue / Bugzilla link

See #9399

## Testing

You will need to set `DEV=False` in your .env file to see the language file activation (you may also want to update your translations `./manage.py l10n_update`).

You can see which locales are active for which file in the [l10n repo](https://github.com/mozmeao/www-l10n/blob/master/metadata/firefox/new/platform.json).

---

- [x] /new page with an active desktop.ftl (http://localhost:8000/en-US/firefox/new/)
  - [x]  display long desktop download page
- [x] /new page without an active desktop.ftl but with an active download.ftl (http://localhost:8000/fi/firefox/new/)
  - [x] display basic page
- [x] platform page with an active platform.ftl but without an download.ftl (http://localhost:8000/et/firefox/mac/)
  - [x] display platform page
  - [x] no untranslated strings
  - [x] display features
  - [x] page title and hero title contain the platform name
- [x] platform page with an active download.ftl but without an active platform.ftl (http://localhost:8000/fi/firefox/mac/)
   - [x] display platform page
   - [x] no untranslated strings
   - [x] does not display features
   - [x] page title and hero title contain the platform name
- [x] platform page with both download.ftl and platform.ftl active (http://localhost:8000/en-US/firefox/mac/)
   - [x] display platform page
   - [x] no untranslated strings
   - [x] display features
   - [x] page title and hero title contain the platform name
- [x] /new page with an active firefox-mobile.ftl file (http://localhost:8000/fi/firefox/new/)
  - you will need to set `SWITCH_FIREFOX_MOBILE_BANNER=True` in your .env file or switch back into dev mode
  - [x] in an iOS or Android device the banner displays
- [x] /new with the basic experience param (http://localhost:8000/en-US/firefox/new/?xv=basic)
  - [x] display basic page
- [x] /new in RU in Russian geo 
  - [x] display Yandex page
- [x] /new in RU outside of Russian geo 
  - [x] display long desktop download page


stephaniehobson:
Blocked by https://github.com/mozilla/bedrock/pull/9896

stephaniehobson:
Integration tests passing: https://gitlab.com/mozmeao/www-config/-/pipelines/259004203

bedrock/firefox/templates/firefox/new/basic/base_platform.html
```diff
@@ -0,0 +1,36 @@
...
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/new/basic/base_download.html" %}
+
+{% block experiments %}{% endblock %}
+
+{# This template can render with EITHER 'firefox/new/download.ftl' OR 'firefox/new/platform.ftl' active
+ # download.ftl contains strings for the hero and small-links for the platform pages
+ # platform.ftl contains strings for the hero, small-links, and features sections for the platform pages
+ # the view for the platform pages will activate the locale if either language file is active  #}
```
alexgibson, line 12: If I'm following along correctly, I think the aim here is to activate as many locales as possible for the download pages, correct? I've tested things as thoroughly as I could using your checklist (really helpful!) and things do seem to be in order ✅

That said, understanding this all felt pretty complicated, and I'm still wondering a little if there's a hole in the logic somewhere. There's certainly a level of complexity we're taking on here long term if we decide to go this route. Looking at the number of active locales for each file right now I see:

- platform.ftl = 53 active locales
- download.ftl = 56 active locales
- combined active locales = 61

This is definitely a small win, but I have to ask the question: is this tech debt acknowledgement worth the number of extra locales we gain? What does long term maintenance look like this these pages?

alexgibson (changes requested):
A couple of minor nits / suggestions at the code level, but functionality wise everything seems to be in working order ✅

I did find the mixing of the L10n files in the platform download pages to be a bit hard to follow, and I wonder if we're making maintenance in the future more difficult by going this route? If we stood to gain a large number of locales then I'd say it's probably worth it. But I have to wonder if there's maybe a better alternative?

- Could we push for locales to catch up on platform.ftl?
- Or lean on the vendor L10n system once that's in place?
- Could we add the platform.ftl strings to download.ftl as new strings (with fallbacks), so the mixed files could eventually be removed?

bedrock/firefox/templates/firefox/new/basic/base_platform.html
stephaniehobson: There is some number of additional locales we will activate with the deleted strings but it is time consuming to estimate how many. For example: ach, ar, and bn should re-activate 🤞 I found these by looking at the translations in Pontoon. BUT we can get this win without sharing strings between the two ftl files.

There are 8 locales where there are active general download pages but not active platform pages (an, ca, fi, hy-AM, ja, lij, mr, ur). Those are the ones that would benifit from mixing the two language files. (because it turns out we have the strings `Download { -brand-name-firefox } <br>for { -brand-name-windows }` in download.ftl 🤷‍♀️) I have asked Raphael to give me an estimate of the value of activating the platform pages in addition to the general download page.

If the value of having both is low I could separate the HTML back into two templates like we had before but then we are back to having two HTML page to maintain and I feel like the HTML changes more than the strings. 

> What does long term maintenance look like for these pages? How do we plan to simplify this again?

We should never touch platform.ftl or download.ftl again. There is no migrating these languages to a new ftl file or a single ftl file, they have been abandoned. My goal was to make the HTML maintainable and never touch the .ftl files again. 

> Which .ftl file should be used going forward? (fallback strings go both ways currently)

I updated the PR to always use the platform ones on the platform page if they are available. I *think* dev will show the English strings from platform.ftl on the platform page since all files are marked as active. 

> How should localizers understand where to find strings / test their changes?

The basic /new has a URL to trigger it so translators can see those strings. 

The platform ones can be viewed on the platform page. With the update I just made so that the platform ones are always preferred changes to platform.ftl should always be reflected there. Again, I *think* the English will display if there's a missing string and not the fallback. I'll see if I can find a way to check.

stephaniehobson:
The answer to most of your questions is, "No new translations are likely to happen. Strings that are added or moved are unlikely to be translated by the locales that are still displaying the old download page." 

If we would rather maintain one HTML template per language file I can separate them again. There are [8 locales](https://editor.mergely.com/WydwF2Eu/) that would activate the platform download changes if merge these changes. However, those 8 locales already have a translated generic /new page. I've asked Raphael to estimate the value of activating platform pages in locales that already have a generic /new page activated.

I think further reviewing can wait until I hear back from Raphael :)

stephaniehobson:
I was wrong, if there is an untranslated string with a fallback the fallback will display in dev mode. Which does complicate the editing experience.

alexgibson:
> We should never touch platform.ftl or download.ftl again. There is no migrating these languages to a new ftl file or a single ftl file, they have been abandoned. My goal was to make the HTML maintainable and never touch the .ftl files again.

In practical terms, I don't know if the assumption that we will never need to never touch these strings ever again feels realistic to me (especially as download.ftl is also tied to a /thanks page template). My understanding of an evergreen page is one that we'll always keep around and maintain as a fallback, as opposed to abandon and never update (allowing us to other legacy templates more easily in the future?)

stephaniehobson:
Investigating combining these files.

pmac:
Just weighing in here: I think the debt we're taking on here is acceptable. I think it is unlikely that we'll touch this page again in the short and medium term. If we do updated it we'll probably be making a new one due to some more global site design direction. If this gets us a new set of supported locales for now then I'm all for it. If we do ever need to touch it again I guess we'll have more to deal with, but that seems like an acceptable level of risk to me.

alexgibson (approved):
@stephaniehobson looks like this hasn't had a rebase in a good while. I took a run through the testing list again and things appear to be correct still, but I wasn't sure if there might be anything that has changed in Protocol since this code was written?

Is it worth pushing this for a test run again just to be safe?

If not, then this looks good to me.

alexgibson:
Also not clear if there was anything else you still wanted to address in the review comments? (but I don't think any are blockers).

alexgibson:
I rebased this branch and pushed it for a test run and things look good: https://gitlab.com/mozmeao/www-config/-/pipelines/331781889

If anything crops up I guess we can always fix it post-merge.

PR #9944, opened 2021-02-19
Move QR Code generator cache from file system to memory

pmac:
Some deployments (e.g. Demos) don't have writable file systems.
Also it seems to be frowned upon to write to the FS in Docker in
general. This should work as well or better.

### DIFF
--- a/bedrock/settings/__init__.py
+++ b/bedrock/settings/__init__.py
@@ -79,6 +79,17 @@ CACHES['externalfiles'] = {
     }
 }
 
+# cache for generated QR codes
+CACHES['qrcode'] = {
+    'BACKEND': 'bedrock.base.cache.SimpleDictCache',
+    'LOCATION': 'qrcode',
+    'TIMEOUT': None,
+    'OPTIONS': {
+        'MAX_ENTRIES': 20,
+        'CULL_FREQUENCY': 4,  # 1/4 entries deleted if max reached
+    }
+}
+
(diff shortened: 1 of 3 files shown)


bedrock/settings/__init__.py
```diff
@@ -79,6 +79,17 @@
...
 }
 
+# cache for generated QR codes
+CACHES['qrcode'] = {
+    'BACKEND': 'bedrock.base.cache.SimpleDictCache',
+    'LOCATION': 'qrcode',
+    'TIMEOUT': None,
+    'OPTIONS': {
+        'MAX_ENTRIES': 20,
+        'CULL_FREQUENCY': 4,  # 1/4 entries deleted if max reached
```
alexgibson, line 89: Just so I understand, if we have more than 20 generated QRCodes on the site at one time and a full cache, what would be the effect on performance?

alexgibson (approved):
Looks good to me. I tested this locally by viewing the QRCode at http://localhost:8000/de/firefox/86.0/whatsnew/all/ and all seems well.

I'll leave merging this in case @duallain has any thoughts.

bedrock/settings/__init__.py
pmac: it would mean a few more times generating the svg. The idea is that if it grows too large it will delete a quarter of the entries in the cache, which will then be generated again if they're requested. We can always grow the size of the cache if we see issues and the size isn't negatively impacting resources.

duallain (approved):

PR #9962, opened 2021-02-25
Use gender neutral term in privacy/passwords.ftl (Fixes #9909)

### LINKED ISSUE #9909: Should use gender neutral term in en/firefox/privacy/safe-passwords
## Description

In https://pontoon.mozilla.org/en-CA/mozillaorg/en/firefox/privacy/passwords.ftl/?string=218027, `... If an attacker can get your password, he can access your account and do anything you could do with that account. ...`

### Expected result

Instead of `he`, suggest to use `they`, so the sentence needs to be rewritten: If *attackers* can get your password, *they* can...

alexgibson:
Fixes #9909


craigcook (approved):
🧞‍♀️

PR #9965, opened 2021-02-25
[no bug] Add doge logo to FxNightly firstrun and whatsnew pages

craigcook:
## Description
Just for funsies. Uses the switch `nightly-doge`.

## Testing
http://localhost:8000/firefox/nightly/firstrun/
http://localhost:8000/firefox/88.0a1/whatsnew/all/


media/css/firefox/whatsnew/whatsnew-nightly.scss
```diff
@@ -23,6 +23,10 @@ $image-path: '/media/protocol/img';
     color: $color-white;
 }
 
+.c-page-header-logo-fx {
+    background-image: none;
```
craigcook, line 27: This was because the original background was visible behind the Nightly one, we evidently never noticed because they just lined up. But the doge's ears are positioned differently and an orange ear was poking out.

stephaniehobson (approved):
R+ 🐕

PR #9998, opened 2021-03-11
fix: remove trailing slash from vpn urls to save a redirect

rlr:
## Description

I noticed the extra redirect when inspecting network traffic.


rlr:
cc: @alexgibson

alexgibson (approved):
Thanks @rlr 💯

Issue #10004, opened 2021-03-12
Update /firefox/central redirect to point to /firefox


maxxcrawford:
## Description

Currently, http://mozilla.org/firefox/central points to https://support.mozilla.org/kb/get-started-firefox-overview-main-features. It needs to point to the root /firefox instead. 

---

#### Success Criteria
- [ ] Visiting the /firefox/central redirects correctly.

maxxcrawford:
Prevous Bugzilla Bug, originally putting this in place: https://bugzilla.mozilla.org/show_bug.cgi?id=979527

maxxcrawford:
@mhoye We should confirm that we have the same level of l10n coverage for this page, as the SUMO page.

Ayushsunny:
I would like to work on this issue

PR #10005, opened 2021-03-12
Fix #1004 - Update /firefox/central redirect

maxxcrawford:
## Description

Update http://mozilla.org/firefox/central redirect to /firefox, instead of SUMO page. 

## Issue / Bugzilla link

#10004

## Testing

1. Go to http://localhost:8000/en-US/firefox/central
2. Should redirect to http://localhost:8000/en-US/firefox/


craigcook (approved):
r+ but it's still marked "Do Not Merge" so I don't want to click the button if there's still some reason we need to wait. If it's good to go, then merge at will 🌵

Issue #9999, opened 2021-03-11
WNP 87 for English (Picture-in-Picture)


craigcook:
## Description

Promoting Picture-in-Picture with some emphasis on the new multiple-window feature. US English specifically might reference the March basketball tournament that shall not be named while en-GB and en-CA might be more general.

DE and FR get their own custom pages, other locales will get the evergreen fallback.

Copy and assets to come.

PR #10016, opened 2021-03-17
Add en-US WNP for Firefox 87, promoting PiP

craigcook:
## Description
Adds a whatsnew page for Firefox 87 promoting Picture-in-Picture. This is only for en-US; en-GB and en-CA fall back to the evergreen page.

## Issue / Bugzilla link
#9999 

## Testing
http://localhost:8000/en-US/firefox/87.0/whatsnew/all/
https://www-demo4.allizom.org/en-US/firefox/87.0/whatsnew/all/


maxxcrawford (approved):
Everything works great - including the non-EN redirect, and the video playing when the "Try it out" button is clicked.

I found one issue, but it's a non-blocking edge case. After the main video has played, clicking the "Try it out" doesn't do anything. I would expect it to reply the video? Or maybe open up the SUMO page on how to use? It's a strange context. Not sure where best to go. 

⏯️

craigcook:
> I found one issue, but it's a non-blocking edge case. After the main video has played, clicking the "Try it out" doesn't do anything. I would expect it to reply the video? Or maybe open up the SUMO page on how to use? It's a strange context. Not sure where best to go.

I brought it up with stakeholders and they like the clear call to action of a clickable button, and didn't feel like it was too much of a problem if it only works once. So we can live with it, and some percentage of people will click it a second time and find it odd that it doesn't do anything, then probably just move on with their lives.

craigcook:
Update: we're going to set the button to a disabled state after it's been clicked.

craigcook:
I've changed the button from a link to a `button`, setting `disabled` on click, and hiding it if JS isn't available.

maxxcrawford (approved):
🚀 🚀 🚀

PR #10018, opened 2021-03-17
Contentful integration

stephaniehobson:
## Description

Support for 3 Contentful page model types and the Hero, Callout, and Card components.


gulpfile.js
```diff
@@ -146,9 +146,9 @@ function sassWatch(cb) {
     sassWatcher.on('change', function(path) {
         return gulp.src(path, {base: buildDir})
             // filter out internal imports (files starting with "_" )
-            .pipe(filter(file => {
-                return !file.relative.startsWith('_');
-            }))
+            // .pipe(filter(file => {
```
alexgibson, line 149: Looks like this is why the JS linting is throwing an error in CircleCI:   

`9:7  error  'filter' is assigned a value but never used  no-unused-vars`.

Not sure why this change is needed here though, without digging in more.

pmac: ah! thanks! I'll dig in. Appreciate the pointer. Not sure why this change is here either.

alexgibson: My hunch is that commenting out this code is not needed (it might have been something Stephanie was testing out and got added accidentally). I can't see how it's related to other changes here.

pmac (approved):
:shipit:

PR #10020, opened 2021-03-18
Switching to allowing gevent workers with gunicorn

duallain:
## Description

Borrowing settings from https://github.com/mdn/developer-portal/blob/master/configs/gunicorn.py
Adding gevent as an optional worker
Adding a dynamic worker count based on the math that gunicorn recommends, but if multiprocess fails failling back


wsgi/config.py
```diff
@@ -1,21 +1,34 @@
...
+errorlog = "-"
+loglevel = getenv("LOGLEVEL", "info")
 
 # Larger keep-alive values maybe needed when directly talking to ELBs
 # See https://github.com/benoitc/gunicorn/issues/1194
-keepalive = getenv('WSGI_KEEP_ALIVE', 2)
-worker_class = getenv('GUNICORN_WORKER_CLASS', 'meinheld.gmeinheld.MeinheldWorker')
-worker_tmp_dir = '/dev/shm'
+keepalive = getenv("WSGI_KEEP_ALIVE", 118)
+worker_class = getenv("GUNICORN_WORKER_CLASS", "meinheld.gmeinheld.MeinheldWorker")
```
pmac, line 24: Should we change this default to "gevent" now?

pmac:
This works for me locally. I'm ready to try it out on dev or stage if you are.

pmac (approved):
⛏️

wsgi/config.py
duallain: Yeah, we should 'someday' but I wanted to be able to do it incrementally so I thought it made sense to stay as meinheld, do the full release cycle then come back and update here.

Issue #9950, opened 2021-02-23
VPN Landing page: updates & localization


alexgibson:
## Description

- Update the VPN landing page design based on of qualitative testing.
- Open up the page for L10n.

---

#### Success Criteria
- [ ] Updated VPN landing page is exposed in Pontoon for L10n.

alexgibson:
Copy doc: https://docs.google.com/document/d/1iSA4iAsI5npi8M1ATbqE6RLjLPzCwNKDyVwtlUm7wmY/edit?usp=sharing

alexgibson:
A/B test plan for headlines: https://docs.google.com/document/d/1clUSwKb3hkX9Fp-rNAV4PgdXwfI7l1aOZTSBaIIJbuY/edit?usp=sharing

PR #10028, opened 2021-03-22
VPN landing page copy tweaks (#9950)

alexgibson:
## Description
http://localhost:8000/en-US/products/vpn/

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/9950

## Testing
- [ ] Landing page copy for number of servers should reflect the variable change.
- [ ] FAQ list of compatible devices should use bullets instead of numbers.


alexgibson:
Added a couple of last minute copy tweaks, so bumping this to a P1 as it blocks getting things into Pontoon.

pmac (approved):
🍰

alexgibson:
Circle Python tests seem to be stuck in a pending state here, but clicking the details link it looks like everything passed. There's a message saying "GitHub returned status 500 adding commit status of success", so I think we're fine to merge here.

bedrock/products/templates/products/vpn/landing.html
```diff
@@ -192,7 +192,7 @@ <h3>{{ ftl('vpn-landing-faq-compatibility-question-heading') }}</h3>
       </summary>
       <p class="vpn-faq-item-detail">{{ ftl('vpn-landing-faq-compatibility-question-desc') }}</p>
 
-      <ol class="vpn-faq-item-detail mzp-u-list-styled">
+      <ul class="vpn-faq-item-detail mzp-u-list-styled">
```
craigcook, line 195: Need to change the closing tag to `</ul>` as well.

pmac: AH! Reviewer fail. Thanks Craig.

craigcook: D'oh, I commented just as it was merged. This shouldn't cause any visible issues, most browsers will be able to handle it. Just a minor fix we can do for the next time the page changes.

craigcook: Fixed in #10029

PR #10052, opened 2021-03-25
Migrate `ms` from Pontoon to Smartling

peiying2:
- Switch Malay (ms) to vendor supported locale
- Remove it from Pontoon
- Add to vendor supported list.

Ref: https://github.com/mozilla-l10n/www-l10n/pull/176

**DO NOT** merge once approved. Needs to be coordinated with  Smartling/L10n team

### DIFF
--- a/l10n/configs/pontoon.toml
+++ b/l10n/configs/pontoon.toml
@@ -71,7 +71,6 @@ locales = [
     "mk",
     "ml",
     "mr",
-    "ms",
     "my",
     "nb-NO",
     "ne-NP",
--- a/l10n/configs/vendor.toml
+++ b/l10n/configs/vendor.toml
@@ -4,26 +4,7 @@
 basepath = ".."
 
 locales = [
-    "cs",
-    "de",
-    "es-AR",
-    "es-ES",
-    "es-MX",
-    "fr",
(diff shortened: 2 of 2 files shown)


alexgibson (approved):
Let me know when you would like to merge r+

peiying2:
@pmac we are going to move Malay (ms) to Smartling soon. Please review the PR and see if this is how you would like it done. Do NOT merge until the day of transition, which we will comment here when it takes place. 

We will follow the same process for any locales we decide to transition to the Smartling platform. Thanks!

pmac:
It looks fine to me, though I really don't know what affect that `vendor.toml` file has. If this works out it looks like it'll be very easy to move languages, so that's nice.

peiying2:
@pmac if this looks good now, you can merge it. It  won't take an effect until Jeff enables Smartling  credential first thing tomorrow morning.

peiying2:
> It looks fine to me, though I really don't know what affect that `vendor.toml` file has. If this works out it looks like it'll be very easy to move languages, so that's nice.

I'm not sure either, but it does make it easier to see what locales are managed through the vendor platform, vs. what are missing  in the pontoon.toml.

PR #10055, opened 2021-03-29
Add error message for VPN sign-ins without a subscription (Fixes #10002)

### LINKED ISSUE #10002: Add message for users who try and sign-in to VPN but are not yet subscribers.
## Description

If someone clicks "Sign-in" on the VPN page and they are not yet a subscriber, but DO already have a Firefox Account, then they are kicked back to the landing page without a message.

### Steps to reproduce

1.) Visit https://www.mozilla.org/en-US/products/vpn/
2.) Click "Sign-In" in the top right navigation.
3.) Sign-in with your Firefox account (must not already be a VPN subscriber).

### Expected result

When redirected back to the landing page, a message should be displayed informing the visitor that they are not yet a subscriber and to choose an option below.

### Actual result

They are redirected back to the landing page without any context.

It may also be worth considering renaming this link to something more specific, such as "Already a subscriber?"

alexgibson:
## Description
http://localhost:8000/en-US/products/vpn/?vpn-sub-not-found=true

## Issue / Bugzilla link
#10002

## Testing
- [ ] Error notification should be displayed when `?vpn-sub-not-found=true` param is present.

### DIFF
--- a/bedrock/products/templates/products/vpn/landing.html
+++ b/bedrock/products/templates/products/vpn/landing.html
@@ -43,6 +43,13 @@
 
 {% block content %}
 <main>
+  {# error message for visitors who try to sign-in without a subscription (issue 10002) #}
+  {% if sub_not_found %}
+    <div class="mzp-c-notification-bar mzp-t-error">
+      <p>{{ ftl('vpn-landing-sub-not-found') }}</p>
+    </div>
+  {% endif %}
+
   {% block hero %}
     {% call vpn_hero(
       heading=ftl('vpn-shared-product-name'),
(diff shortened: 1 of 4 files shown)


l10n/en/products/vpn/shared.ftl
```diff
@@ -75,3 +75,6 @@ vpn-shared-pricing-save-percent = Save { $percent }%
 vpn-shared-pricing-total = { $amount } total
 
 ##
+
+# message shown to visitors who try to sign-in without an active subscription.
+vpn-shared-sub-not-found = Oops, we can’t find your subscription.
```
alexgibson, line 80: This is just placeholder copy for now, will update once we have the final string.

alexgibson:
Ready for review.

PR #10072, opened 2021-04-01
Hide price for VPN if not available (Fixes #10070)

### LINKED ISSUE #10070: Hide price in countries where VPN is not yet available
## Description

Right now if VPN is not available people see the price in the features block. There's a request to hide this if VPN is not yet available in a country, as once it launches there's no guarantee that price/plan will be the same.

---

#### Success Criteria
- [ ] Price should not be displayed in countries where VPN is not yet available.

alexgibson:
## Description
http://localhost:8000/en-US/products/vpn/?geo=de

image

## Issue / Bugzilla link
#10070

## Testing
- [ ] Price should not be visible on the page if VPN is not yet available.


craigcook (approved):
💸

PR #10084, opened 2021-04-02
Fix #10047 - Update strings to correct spelling of 'Add-ons'

### LINKED ISSUE #10047: Change "Addons" to "Add-ons" on mozilla.org
In an effort to be more consistent, we should change "Addons" to "Add-ons" so that we are aligned with AMO.

One example where this will need to be changed is the subnav:

https://www.mozilla.org/en-US/firefox/new/

maxxcrawford:
## Description

Update all instances of "Addons" across the site with "Add-ons" 

## Issue / Bugzilla link

#10047

## Testing

- http://localhost:8000/en-US/firefox/new/
- http://localhost:8000/en-US/firefox/mobile/

_Note - Both instances are in the subnavigation._

### DIFF
--- a/bedrock/firefox/templates/firefox/new/desktop/download.html
+++ b/bedrock/firefox/templates/firefox/new/desktop/download.html
@@ -52,7 +52,7 @@
           <a href="https://support.mozilla.org/products/firefox/{{ referrals }}" data-link-type="nav" data-link-position="subnav" data-link-name="Support">{{ ftl('sub-navigation-support', fallback='navigation-support') }}</a>
         </li>
         <li class="c-sub-navigation-item">
-          <a href="https://addons.mozilla.org/firefox/{{ referrals }}" data-link-type="nav" data-link-position="subnav" data-link-name="Addons">{{ ftl('sub-navigation-addons', fallback='navigation-add-ons') }}</a>
+          <a href="https://addons.mozilla.org/firefox/{{ referrals }}" data-link-type="nav" data-link-position="subnav" data-link-name="Addons">{{ ftl('sub-navigation-add-ons', fallback='sub-navigation-addons') }}</a>
         </li>
         <li class="c-sub-navigation-item">
           <a href="{{ url('firefox.all') }}" data-link-type="nav" data-link-position="subnav" data-link-name="All Languages">{{ ftl('sub-navigation-all-languages', fallback='download-button-systems-languages') }}</a>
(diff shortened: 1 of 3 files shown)


bedrock/firefox/templates/firefox/new/desktop/download.html
```diff
@@ -52,7 +52,7 @@ <h2 class="c-sub-navigation-title">{{ ftl('sub-navigation-firefox-for-desktop',
           <a href="https://support.mozilla.org/products/firefox/{{ referrals }}" data-link-type="nav" data-link-position="subnav" data-link-name="Support">{{ ftl('sub-navigation-support', fallback='navigation-support') }}</a>
         </li>
         <li class="c-sub-navigation-item">
-          <a href="https://addons.mozilla.org/firefox/{{ referrals }}" data-link-type="nav" data-link-position="subnav" data-link-name="Addons">{{ ftl('sub-navigation-addons', fallback='navigation-add-ons') }}</a>
+          <a href="https://addons.mozilla.org/firefox/{{ referrals }}" data-link-type="nav" data-link-position="subnav" data-link-name="Addons">{{ ftl('sub-navigation-add-ons', fallback='sub-navigation-addons') }}</a>
```
maxxcrawford, line 55: Note that I'm leaving `data-link-name="Addons"` alone to not break analytics.

l10n/en/sub_navigation.ftl
```diff
@@ -24,12 +24,12 @@ sub-navigation-desktop-beta-and-developer = Desktop { -brand-name-beta } & { -br
 sub-navigation-desktop-nightly = Desktop { -brand-name-nightly }
 sub-navigation-features = Features
 sub-navigation-support = Support
-sub-navigation-addons = Addons
```
maxxcrawford, line 27: BLOCKING Question: I was removing this string as its no longer used, but I have it set as the fallback. Do I need to leave it for now?

maxxcrawford (commented):
Self-reviewed!

l10n/en/sub_navigation.ftl
craigcook: Yeah, I think it needs to still be in the file to work as a fallback, and you can add a comment above it marking it as obsolete like:
`# Obsolete string`

maxxcrawford: @craigcook Revised!

craigcook (approved):
🩹

PR #10091, opened 2021-04-06
Updated forums page

cshields:
Last weekend we decommissioned our legacy forums (mailman+usenet) and so this page needs to be updated.  Text was reviewed by mhoye as well, not sure of the process for approving this change any further.

(likewise, I've tested this in a local docker environment, but I am not a bedrock dev so please let me know if this is incomplete or breaks something elsewhere)


craigcook (approved):
🪑

PR #10098, opened 2021-04-07
Update root store policy [fix bug #1703063]

craigcook:
## Description
Updates the Mozilla Root Store Policy to v2.7.1. This also adds a table of contents at the top with skip-links to each section, though it's not a FULL list with all the subsections because that was getting kinda long. 

We should get the security folks to review this on demo and give their blessing. Also the policy date is "effective May 1, 2021" so does that mean we should wait until then to publish it? I'll ask for clarification in the bug.

## Issue / Bugzilla link
https://bugzilla.mozilla.org/show_bug.cgi?id=1703063

## Testing
http://localhost:8000/en-US/about/governance/policies/security-group/certs/policy/
https://www-demo-pr-10098.herokuapp.com/en-US/about/governance/policies/security-group/certs/policy/


craigcook:
[Comment from bwilson](https://bugzilla.mozilla.org/show_bug.cgi?id=1703063#c2): "even though the effective date is May 1, 2021, we need to publish it soon so that we can discuss it with the community ahead of it going effective." So this is ready for review and we should try to get live this week, if possible.

pmac (approved):
LGTM 💯 Let's get this published for them.

PR #10131, opened 2021-04-15
L10n fixes for VPN landing page on mobile (Fixes #10130)

### LINKED ISSUE #10130: [l10n] UI issues for “german” language in VPN homepage
#### Affected versions:
 - All Firefox;

#### Affected FxA versions:
- 1.203.7;

#### Affected Platforms:
- Android 11;

#### Prerequisites:
- Have the browser localized in “German”;

#### Steps to reproduce:
1. Open homepage: https://www-dev.allizom.org/de/products/vpn/
2. Scroll down to #pricing section;

#### Expected result:
- There are no display or layout issues.

#### Actual result:
- The text from the section is not wrapped and appears cut. There is an horizontal scroll bar.

#### Notes:
- This also happens at the beginning of the page with the first title.
- Also happens while in a region where VPN is available;
- This does not happen on iOS.

ui

alexgibson:
## Description
Some CSS fixes to better cater for longer strings (such as German) on small viewports. This is mostly just making copy a little smaller on mobile screen sizes. Large viewports should be uneffected.

- Reduce hero font size on small screens.
- Reduce block component padding & font-size on small screens.
- Reduce feature list font size / spacing on small screens.

I've tried to ensure things are still readable (i.e. not truncated or cut off screen) down to 320px in German.

- http://localhost:8000/en-US/products/vpn/
- http://localhost:8000/de/products/vpn/
- http://localhost:8000/fr/products/vpn/

## Issue / Bugzilla link
#10130

## Testing
- [x] Content is readbale and not truncated in German even at small screen sizes.

### DIFF
--- a/media/css/products/vpn/common.scss
+++ b/media/css/products/vpn/common.scss
@@ -166,30 +166,38 @@ main .mzp-c-button.mzp-t-xl {
 
 .vpn-feature-list.mzp-u-list-styled {
     @include bidi(((text-align, left, right),));
-    @include text-body-xl;
+    @include text-body-lg;
     display: inline-block;
     list-style: none;
-    margin-bottom: $spacing-lg;
+    margin: 0 0 $spacing-lg 0;
 
     li {
         @include bidi((
-            (background-position, top 7px left, top 7px right),
-            (padding-left, (18px + $spacing-md), padding-right, 0),
+            (background-position, top 3px left, top 3px right),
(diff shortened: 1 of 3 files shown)


craigcook (approved):
Tested at 320px and everything was readable, no truncation or horizontal scrolling, though a few long words are a very tight fit ("Geschwindigkeit"). If we're really concerned we could drop the font size one more notch at the smallest breakpoint, but I think that's a bit of an edge case. r+ 🍋

PR #10146, opened 2021-04-16
Revert wsgi keep alive as well

duallain:
it's one of the last remaining items different between our known good state and current wild cpu usage


sciurus (approved):
I see this working great by hand in the cluster.

pmac (approved):
💯

PR #10158, opened 2021-04-21
Only try and swap out images for China Repack builds (Fixes #10157)

### LINKED ISSUE #10157: SyntaxError: 'img[data-[RANDOM-IDENTIFIER-link]' is not a valid selector
## Description
One of the most perplexing and frequent JS errors we see in Sentry as along the lines of the following:

`SyntaxError: 'img[data-[RANDOM-IDENTIFIER-link]' is not a valid selector`

All the errors use different values for [RANDOM IDENTIFIER], and are often formatted in a similar way to email addresses. You can see many examples here: https://sentry.prod.mozaws.net/operations/bedrock-prod/?query=is%3Aunresolved+SyntaxError%3A+%27img%5Bdata

For a long time I've chalked this up to some popular extension, but after digging in further it appears the error comes from the following line of code: https://github.com/mozilla/bedrock/blob/master/media/js/base/mozilla-utils.js#L105

This means that the `[RANDOM IDENTIFIER]` values aren't from extensions at all, but are actually (I can only assume?), custom Firefox builds, or at least builds that have non standard distribution IDs.

This is certainly an oddity itself that I don't quite understand fully, but the good news is I think we now understand enough to prevent this error from occuring.

alexgibson:
## Description
Changes the `Mozilla.Utils.maybeSwitchToChinaRepackImages()` function so it will only look for matching selectors if the distribution ID is specifically `mozillaonline`.

Should fix errors such as: https://sentry.prod.mozaws.net/operations/bedrock-prod/?query=is%3Aunresolved+SyntaxError%3A+%27img%5Bdata&sort=freq

## Issue / Bugzilla link
#10157

## Testing
- [ ] Running `Mozilla.Utils.maybeSwitchToChinaRepackImages({distribution: 'regata os'});` should no longer throw an error.

### DIFF
--- a/media/js/base/mozilla-utils.js
+++ b/media/js/base/mozilla-utils.js
@@ -97,11 +97,17 @@ if (typeof window.Mozilla === 'undefined') {
 
     // Bug 1264843: link to China build of Fx4A, for display within Fx China repack
     Utils.maybeSwitchToChinaRepackImages = function(client) {
-        if (!client.distribution || client.distribution === 'default') {
+        if (!client.distribution) {
             return;
         }
 
         var distribution = client.distribution.toLowerCase();
+
+        // only swap out images for China Repack builds (issue 10157)
+        if (distribution !== 'mozillaonline') {
+            return;
+        }
+
         var images = document.querySelectorAll('img[data-' + distribution + '-link]');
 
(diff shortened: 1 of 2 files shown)


pmac (approved):
📦

Issue #10075, opened 2021-04-01
update protocol on bedrock


slightlyoffbeat:

craigcook:
Published [Protocol 14.0.0](https://github.com/mozilla/protocol/releases/tag/v14.0.0) yesterday so this is unblocked.

PR #10163, opened 2021-04-21
Update to Protocol 14.0.3

craigcook:
## Description

Updates from Protocol 13.0.1 to 14.0.3 (see [Protocol's changelog](https://github.com/mozilla/protocol/blob/main/CHANGELOG.md) for a list of changes). Significant things in this update are some new components: Split, Picto, Logo, and Section Heading, as well as a multi-column template. Another major change is increased vertical spacing on content containers (`mzp-l-content`) so most pages will have a bit more whitespace. It also adds some new logo assets for Focus and VPN, and renames some other logo assets (Mozilla and Pocket) to match our current naming convention.

## Issue / Bugzilla link
#10075 

## Testing
This is a global update to our CSS framework and impacts nearly every page in some way so it could use some pretty thorough review to make sure nothing is seriously broken. It adds new macros for Split and Picto and uses them on a few pages, so those could use careful review as well.

https://www-demo-pr-10163.herokuapp.com/


craigcook:
An asset name needs to be fixed upstream in https://github.com/mozilla/protocol/pull/683 and this will update to 14.0.3

craigcook:
Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/290263039

stephaniehobson:
Crawlbars on /new at 1311px :(

stephaniehobson:
Problem with the Pocket logo on http://localhost:8000/en-US/firefox/privacy/products/ Screen Shot 2021-05-03 at 14 45 20

You should probably check there's a size defined for the logo anywhere it's used.

stephaniehobson:
IMHO there's too much whitespace between the "you're up to date" message and the start of the content on the VPN WNPs (84, 85, 88). But it's not a blocker.

media/css/firefox/new/desktop/download.scss
```diff
@@ -8,6 +8,7 @@ $image-path: '/media/protocol/img';
 
 @import '../../../../protocol/css/components/emphasis-box';
 @import '../../../../protocol/css/components/modal';
+@import '../../../../protocol/css/components/section-heading';
```
stephaniehobson, line 11: This page also has a prototype block component that should be able to be swapped out for split (not a blocker for merging though).

media/css/firefox/privacy/products.scss
```diff
@@ -225,7 +225,7 @@ $image-path: '/media/protocol/img';
 }
 
 .privacy-products-pocket .mzp-c-card-feature-title {
-    background-image: url('/media/protocol/img/logos/pocket/logo-word-horz.svg');
+    background-image: url('/media/protocol/img/logos/pocket/logo-word-hor.svg');
```
stephaniehobson, line 228: Needs an explicit background size declaration to go with it (see screen shot in other comment).

stephaniehobson (changes requested):
Lots of excellent code clean up 🧽  Lots of places where customizations have been completely replaced by standard components ✨ A couple tiny nits but overall this massive PR is in excellent shape.

bedrock/firefox/templates/firefox/browsers/chromebook.html
```diff
@@ -36,21 +36,21 @@ <h2 class="c-sub-navigation-title">{{ ftl('sub-navigation-firefox-for-desktop',
...
+  <h1 class="mzp-u-title-xl">{{ ftl('browsers-chromebook-get-firefox-browser') }}</h1>
   <p>{{ ftl('browsers-chromebook-so-youve-got') }}</p>
 
   <div class="mzp-c-menu-list mzp-t-cta mzp-t-download" id="chromebook-download-list">
     <h4 class="mzp-c-menu-list-title">{{ ftl('browsers-chromebook-dropdown-copy') }}</h4>
     <ul class="mzp-c-menu-list-list download-platform-list">
       <li class="mzp-c-menu-list-item">
         <a href="{{ firefox_adjust_url('android', 'firefox-browsers-faq') }}" rel="external noopener" data-link-type="download" data-display-name="Android" data-download-version="android" data-download-os="Android">
-          {{ ftl('browsers-chromebook-get-firefox-for') }}
+          {{ ftl('browsers-chromebook-browsers-chromebook-get-firefox-for') }}
```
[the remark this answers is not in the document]
craigcook: This is actually a bug fix. The canonical string ID has the repeated bit https://github.com/mozilla/bedrock/blob/master/l10n/en/firefox/browsers/chromebook.ftl#L13

But there also seems to be a bug in the current template because the tagline and dropdown don't appear at all on the current page (see prod https://www.mozilla.org/firefox/browsers/chromebook/), I expect due to some problem with how the hero macro is called but I didn't investigate deeply. When I replaced the hero I fixed that glitch and the menu appeared and I discovered the mismatched string ID. We can fix the ID as well but that makes it a new string and this has already been widely translated.

media/css/firefox/new/desktop/download.scss
craigcook: hmm this is a little tricky with the inline SVG. I don't think it can use the macro, it would have to just be coded into the page. Which we can totally do. But I also think that hero image is going to change in the next few weeks for MR1 so maybe we can leave it for now and change to the component when we're doing the other updates.

bedrock/firefox/templates/firefox/browsers/chromebook.html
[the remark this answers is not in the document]
stephaniehobson: Bug fix is fine as is 👍

media/css/firefox/new/desktop/download.scss
stephaniehobson: And the weird browser compat table thing could be as annoying as the SVG. Okay, this can go as-is. 👍

media/css/firefox/new/desktop/download.scss
```diff
@@ -902,7 +902,7 @@ button.mzp-c-cta-link {
     padding-top: 0;
 
     .mzp-c-emphasis-box {
-        margin: 0 $layout-2xs * -1;
```
stephaniehobson, line 937: 😂

media/css/firefox/new/desktop/download.scss
```diff
@@ -902,7 +902,7 @@ button.mzp-c-cta-link {
...
+        margin: 0 -$layout-2xs;
```
stephaniehobson, line 905: 😂

stephaniehobson (approved):
R+ 🍌

PR #10165, opened 2021-04-22
Fix typo on WNP88 :facepalm: [no bug]

craigcook:
## Description
The old "using now" <-> "now using" switcheroo.

## Testing
http://localhost:8000/en-US/firefox/88.0/whatsnew/all/


maxxcrawford (approved):
🙈

PR #10170, opened 2021-04-26
Fallback to variable pricing only if country is unknown and switch is enabled (Fixes #10169)

### LINKED ISSUE #10169: VPN: fallback to variable pricing only if country is unknown and switch is enabled
## Description

This isn't a real user facing issue thankfully, but currently there's a bug where if country code is unknown and page language is 'fr' or 'de', then variable pricing can end up being shown irrespective of switch state.

---

#### Success Criteria
- [ ] Fallback logic should respect switch state.

alexgibson:
## Description
Fixes an edge case bug where variable pricing can be shown irrespective of switch state.

## Issue / Bugzilla link
#10169

## Testing
1.) Set `SWITCH_VPN_LAUNCH_GERMANY_FRANCE=off` in your `.env`.
2.) Set `Dev=False` in your `.env` (this will make `country_code.json` 404 when running localhost, which is useful for triggering this bug).
3.) Open http://localhost:8000/de/products/vpn/#pricing
4.) Verify that the page falls back to fixed pricing (US$4.99/Month).

### DIFF
--- a/media/js/products/vpn/geo.js
+++ b/media/js/products/vpn/geo.js
@@ -123,7 +123,7 @@
         return 'not-available';
     };
 
-    VPN.setAvailability = function(availability, language) {
+    VPN.setAvailability = function(availability, language, variablePricing) {
         if (availability === 'fixed-price') {
             // If VPN is available in countries that support fixed pricing,
             // show the fixed pricing subscribe links.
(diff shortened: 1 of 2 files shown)


alexgibson:
It's launch day, so this is likely no longer required. closing

PR #10203, opened 2021-05-17
Differentiate headings for installations stuff [ fixes #10063]

### LINKED ISSUE #10063: Dev Edition: differentiate headline of upgrades from new installations
See: https://bugzilla.mozilla.org/show_bug.cgi?id=1700518

Abhra303:
## Description
It fixes issue #10063. It works same as the bugzilla issue https://bugzilla.mozilla.org/show_bug.cgi?id=1700518 suggests. 

## Issue / Bugzilla link
#10063
## Testing
go to http://127.0.0.1:8000/en-US/firefox/88.0a2/whatsnew/all/?oldversion=87.0 . Remove the oldversion attribute from the link and compare the change in the heading.

### DIFF
--- a/bedrock/firefox/templates/firefox/developer/whatsnew.html
+++ b/bedrock/firefox/templates/firefox/developer/whatsnew.html
@@ -25,7 +25,7 @@
       {{ high_res_img('protocol/img/logos/firefox/browser/developer/logo-word-hor-md.png', {'alt': '', 'width': '368', 'height': '68'}) }}
 
       {% if ftl_has_messages('firefox-developer-congrats-you-now-have', 'firefox-developer-welcome-to-your-new-favorite') %}
-        {% if oldversion and ftl_has_messages('firefox-developer-congrats-you-now-have-latest') %}
+        {% if ftl_has_messages('firefox-developer-congrats-you-now-have-latest') %}
           <h1 class="intro-title">{{ ftl('firefox-developer-congrats-you-now-have-latest') }}</h1>
         {% else %}
           <h1 class="intro-title">{{ ftl('firefox-developer-congrats-you-now-have') }}</h1>
--- a/bedrock/firefox/templates/firefox/developer/whatsnew.html
+++ b/bedrock/firefox/templates/firefox/developer/whatsnew.html
@@ -25,11 +25,7 @@
       {{ high_res_img('protocol/img/logos/firefox/browser/developer/logo-word-hor-md.png', {'alt': '', 'width': '368', 'height': '68'}) }}
 
       {% if ftl_has_messages('firefox-developer-congrats-you-now-have', 'firefox-developer-welcome-to-your-new-favorite') %}
-        {% if ftl_has_messages('firefox-developer-congrats-you-now-have-latest') %}
-          <h1 class="intro-title">{{ ftl('firefox-developer-congrats-you-now-have-latest') }}</h1>
-        {% else %}
-          <h1 class="intro-title">{{ ftl('firefox-developer-congrats-you-now-have') }}</h1>
-        {% endif %}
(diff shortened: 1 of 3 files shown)


bedrock/firefox/templates/firefox/developer/whatsnew.html
```diff
@@ -25,7 +25,11 @@
       {{ high_res_img('protocol/img/logos/firefox/browser/developer/logo-word-hor-md.png', {'alt': '', 'width': '368', 'height': '68'}) }}
 
       {% if ftl_has_messages('firefox-developer-congrats-you-now-have', 'firefox-developer-welcome-to-your-new-favorite') %}
-        <h1 class="intro-title">{{ ftl('firefox-developer-congrats-you-now-have') }}</h1>
+        {% if oldversion and ftl_has_messages('firefox-developer-congrats-you-now-have-latest') %}
```
alexgibson, line 28: I think we can actually simplify this a bit and remove the `oldversion` check.

This /whatsnew page is only ever shown on a browser update. New installations see a different [/firstrun](https://www.mozilla.org/en-US/firefox/88.0a2/firstrun/) page when a profile is first created. So it's safe to assume that the new string you've added here is always applicable.

Does that make sense?

alexgibson (changes requested):
Thanks for the PR @Abhra303!

The new string you added looks good. I think we can simplify the logic a bit by removing the `oldversion` check. Let me know if you have any questions? Thanks

bedrock/firefox/templates/firefox/developer/whatsnew.html
Abhra303: Actually, I was not sure whether the /whatsnew page is only for update or for both update and installation. So I added the `old_version` check.  
As you clarified now, I am going to remove this check. But I have a question though.
@alexgibson is there any reason for keeping the `old_version` in the get_template_name method of WhatsnewView? Because old_version is used nowhere after the [declaration](https://github.com/mozilla/bedrock/blob/daa9b16a786fe7aaca9d543c08ea71a4aed583a0/bedrock/firefox/views.py#L630).

alexgibson:
> @alexgibson is there any reason for keeping the old_version in the get_template_name method of WhatsnewView? Because old_version is used nowhere after the declaration.

Please keep this declaration as it is sometimes used for targeting specific templates. Thanks

bedrock/firefox/templates/firefox/developer/whatsnew.html
```diff
@@ -25,7 +25,11 @@
       {{ high_res_img('protocol/img/logos/firefox/browser/developer/logo-word-hor-md.png', {'alt': '', 'width': '368', 'height': '68'}) }}
 
       {% if ftl_has_messages('firefox-developer-congrats-you-now-have', 'firefox-developer-welcome-to-your-new-favorite') %}
-        <h1 class="intro-title">{{ ftl('firefox-developer-congrats-you-now-have') }}</h1>
+        {% if ftl_has_messages('firefox-developer-congrats-you-now-have-latest') %}
+          <h1 class="intro-title">{{ ftl('firefox-developer-congrats-you-now-have-latest') }}</h1>
+        {% else %}
+          <h1 class="intro-title">{{ ftl('firefox-developer-congrats-you-now-have') }}</h1>
+        {% endif %}
```
alexgibson, line 32: You can now simplify this conditional to:

```
<h1 class="intro-title">{{ ftl('firefox-developer-congrats-you-now-have-latest', fallback='firefox-developer-congrats-you-now-have') }}</h1>
```

alexgibson (changes requested):

bedrock/firefox/templates/firefox/developer/whatsnew.html
Abhra303: oh sure, thanks :)

alexgibson (approved):
Thanks @Abhra303 r+ 🍡

PR #10204, opened 2021-05-18
Remove .gitmodules

jwhitlock:
The file ``.gitmodules`` was left empty after PR #9213, and ``git clone --recursive`` is no longer needed.


alexgibson (approved):
Thanks @jwhitlock

Issue #10205, opened 2021-05-18
Preference Center sends unneeded update to Basket when unsubscribing from all newsletters


jwhitlock:
## Description

When unsubscribing from all newsletters, Bedrock is sending two requests to Basket:

1. A ``POST`` to ``/news/user/{token}``, with a form body with keys ``api_key`` and ``lang``. 
2. A ``POST`` to ``/news/unsubscribe/{token}``, with a form body with an ``email`` and ``optout=Y``

The first ``POST`` is not needed.

### Steps to reproduce

1. Load the preference center at https://www.mozilla.org/en-US/newsletter/existing/{token}, replacing with a valid token from a new or existing newsletter sign-up
2. Select "Remove me from all the subscriptions on this page:"
3. Select "Save Preferences"

### Expected result

Only the second ``POST`` to ``/news/unsubscribe/{token}`` and ``optout=Y`` should be sent, since no other updates (language or email format) were made.

### Actual result

The two ``POST``s are made, and basket makes an unnecessary call to update the email record in the backing storage before unsubscribing the user from all newsletters.

### Environment

This was observed in the ``basket-dev`` environment, through the extra calls to the backing storage. This is due to the ``api_key`` being set to the setting ``BASKET_API_KEY``. I believe this setting is also populated in production.

There is an existing test that the update ``POST`` is not sent. The test passes due to an empty ``BASKET_API_KEY``:

https://github.com/mozilla/bedrock/blob/489b92654fb4b260f90e8fab35a1bacb57ec092a/bedrock/newsletter/tests/test_views.py#L306-L307

alexgibson:
I believe the behavior here is no longer true today, since we migrated the preference center logic to the front-end in https://github.com/mozilla/bedrock/issues/11946. Feel free to re-open if there's still something to improve here.

PR #10206, opened 2021-05-18
Do not send basket update when nothing changed on newsletter preference page

jwhitlock:
## Description

Apply ``BASKET_API_KEY`` to the basket update ``kwargs`` after checking if any of the settings from the newsletter preference page changed. This prevents detecting that the page always has changes, and avoids sending a needless update to Basket when unsubscribing from all newsletters.

## Issue / Bugzilla link

https://github.com/mozilla/bedrock/issues/10205

## Testing

Unit testing only. I added ``BASKET_API_KEY`` to all ``TestExistingNewsletterView``, which required some test changes and caused ``test_remove_all`` to fail, replicating the issue from basket-dev.  The second commit changes the view logic, and the test now passes again.

### DIFF
--- a/bedrock/newsletter/tests/test_views.py
+++ b/bedrock/newsletter/tests/test_views.py
@@ -3,9 +3,11 @@
 # file, You can obtain one at https://mozilla.org/MPL/2.0/.
 import json
 import uuid
+from unittest import expectedFailure
 
 from django.http import HttpResponse
 from django.test.client import RequestFactory
+from django.test.utils import override_settings
 
 import basket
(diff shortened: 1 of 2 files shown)


jwhitlock:
Sorry, I was busy as well!

I believe so. I'll rebase and see if the code still makes sense.

stevejalim (commented):
This makes sense to me - both the reasoning and the tests. r+

stevejalim (approved):

stevejalim:
Will leave it to @jwhitlock or @pmac to merge after their own sense-check, given the vintage of this fix 👍

Issue #10207, opened 2021-05-19
AMO download attribution experiment


alexgibson:
## Description

The AMO team would like to run an experiment whereby Firefox download links on AMO point directly to /thanks, as opposed to /new (which is the current behaviour). To support this experiment they need to measure the effect it has on installs, but the /thanks page does not (currently) do the stub-attribution dance before triggering a Firefox download (it assumes you have done a mozorg page navigation before deciding to download).

We met to discuss the work/risks involved in adding stub attribution to /thanks and settled on the following:

- To satisfy the experiment we can build a custom /thanks page that gets triggered by a specific query param (TBD). Making this available via a custom page / parameter lowers the risk of this breaking the download funnel, and confines the new logic only to those who enter into the experiment.
- The custom page can look exactly the same as the current /thanks page, but can have some bespoke JS that will attempt to make a stub attribution call before triggering a download.

If the experiment proves successful, then we can work toward a plan to add this logic to the regular / thanks page.

---

#### Success Criteria
- [x] A custom /thanks page is available via a specific query param (TBD)
- [x] Custom page will make a stub attribution call before triggering a Firefox download (if visitor is on Windows)

bobsilverberg:
Thanks for filing, @alexgibson! When you say the custom /thanks page will be available via a specific query param, does that mean that the URL will remain /thanks, and that we'll be passing the custom query param to that URL, or will it be an entirely different URL (e.g., /amo-thanks)?

We would intend to pass all of the other query params we are currently using for the download link to this new URL, and would expect them to continue to work for GA and telemetry. Specifically, those are:
- experiment
- variation
- utm_campaign
- utm_content
- utm_medium
- utm_source
- utm_term

alexgibson:
@bobsilverberg I propose we use a query param on the regular /thanks URL, something similar to:

https://www.mozilla.org/firefox/download/thanks/?source=amo (open to other suggestions if something else might be better)

You can still pass along any other params you need here to attribute your experiment.

bobsilverberg:
Thanks @alexgibson. I have a question regarding IE<=9, which you have indicated should be excluded from using this new URL. The number of users we see on IE<=9 is extremely small (e.g., 40 over a 2 week period), plus we do not support IE at all on AMO and know that the site has rendering issues for anyone on IE. For these reasons we intend to simply ignore this issue, rather than add code to address it.

[...]

PR #10216, opened 2021-05-24
Remove jQuery as a dependency from stub-attribution.js (Fixes #7524)

### LINKED ISSUE #7524: Remove jQuery as a dependency from stub-attribution.js
## Description

There are a few areas in [stub-attribution.js](https://github.com/mozilla/bedrock/blob/master/media/js/base/stub-attribution.js) that rely on jQuery. We should replace these function calls with vanilla JS equivalents.

It's worth noting that stub-attribution is currently served to [even old IE browsers](https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/base-pebbles.html#L121-L125). Before making this change, we need to consider if this level of support is still necessary, or if we can move this bundle to 1st class supported browsers only.

---

- [ ] Code is running in production using vanilla JS.

- [ ] Dropping support for a legacy browser may see a drop in attribution data.

#### Tasks
- [ ] Refactor code.

alexgibson:
## Description
- Removes old jQuery calls from stub-attribution script.
- Prevent the script from running on IE6/7 via conditional comment.
- Add custom success / timeout callbacks that might be useful for https://github.com/mozilla/bedrock/issues/10207
- Update tests.

## Issue / Bugzilla link
#7524

## Testing
To test this use **Chrome on a Windows VM**:

1. Open https://www-demo2.allizom.org/en-US/?utm_source=test&utm_campaign=test
2. Click "Download Firefox" in the navigation on the top right.
3. Hold down the shift key and click the "Try downloading again" link to open the download URL in a new tab.
4. Copy the download URL using command + C
5. Open https://www.base64decode.org/
6. Paste the download link in the text field, and then remove everything except the value of the `attribution_code` param (this is the only bit we need to test).
7. Click "Decode".
8. Verify the following:
  - [x] Both `source` and `campaign` have the value `test`.
  - [x] The `ua` value should be `chrome` (assuming you tested in Chrome).
  - [x] The `visit_id` value should be a random number (your GA visit ID).
  - [x] Everything else should be `(not set)`.

### DIFF
--- a/bedrock/base/templates/base-protocol.html
+++ b/bedrock/base/templates/base-protocol.html
@@ -124,12 +124,14 @@
       <![endif]-->
     {% endblock %}
 
+    <!--[if !lte IE 7]><!-->
     {# Bug 1279291 #}
     {% block stub_attribution %}
       {% if settings.STUB_ATTRIBUTION_RATE %}
         {{ js_bundle('stub-attribution') }}
       {% endif %}
     {% endblock %}
+    <!--<![endif]-->
 
     <!--[if !IE]><!-->
     {% block js %}{% endblock %}
--- a/media/js/base/stub-attribution.js
+++ b/media/js/base/stub-attribution.js
@@ -31,6 +31,14 @@ if (typeof window.Mozilla === 'undefined') {
     StubAttribution.experimentVariation;
 
     /**
+     * Custom event handler callback globals. These can be defined as functions when
+     * calling StubAttribution.init();
+     */
+    StubAttribution.successCallback;
+    StubAttribution.timeoutCallback;
+    StubAttribution.requestComplete = false;
+
+    /**
      * Determines if session falls within the predefined stub attribution sample rate.
      * @return {Boolean}.
      */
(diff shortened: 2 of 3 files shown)


alexgibson:
I've tested this on demo as far back as IE8, and all seems to be working. Think this is ready for review.

stephaniehobson:
My local VMs are not working. I got the GTM JS error and the link didn't match when I tried it in BrowserStack.

Going to see if my Surface has IE8 emulation.

stephaniehobson:
(Oh, maybe the testing steps / expectations are different in IE?)

alexgibson:
> (Oh, maybe the testing steps / expectations are different in IE?)

Oh, I tried spare you the pain of needing to use IE devevloper tools by suggesting Chrome to verify the changes :)

If you really want to test this in old IE browsers (IE 9 and below) then you'll need to look for the stub attribution params attached to the `data-direct-link` attribute in the DOM (since those browsers trigger the download before hitting /thanks): https://github.com/mozilla/bedrock/blob/master/bedrock/firefox/templates/firefox/includes/download-button-thanks.html#L4

IE 10/11 should use the regular method though, if you want to test that flow.

stephaniehobson:
It's not working for me in old Edge (44). Which was an accidental discovery since I didn't know it wasn't up to date. Do we care about that?

stephaniehobson (approved):
R+ if we don't care about EdgeHTML.

alexgibson:
> It's not working for me in old Edge (44). Which was an accidental discovery since I didn't know it wasn't up to date. Do we care about that? (the source and campaign appear as unset)

I don't have Edge 44 to hand, but I tested in BrowserStack on Edge 15 (oldest) and Edge 91 (latest) and both worked as expected. Perhaps a VM error?

I think we should hold off merging this until after the workweek, but unless the error is reproducible I think this is probably ready.

PR #10228, opened 2021-05-27
Leadership ordering, add Surman to MoFo board [no bug]

craigcook:
## Description
We were asked to change the ordering to strictly alphabetical (previously we listed steering committee members first, then everyone else). Also adding Mark Surman to the MoFo board.

## Testing
http://localhost:8000/about/leadership/


stephaniehobson (approved):
R+ ⛓️

PR #10235, opened 2021-05-28
Add macos native context menus image to 89 release notes

pascalchevrel:
## Description
Illustrative image for our release notes.


craigcook (changes requested):
This file could use some optimization to make it smaller, without losing quality. I was able to reduce it by 52% running through tinypng.com

pascalchevrel:
I received more files from marketing, they are all heavy. Let's cancel this PR, I'll make another one with all the files

Issue #10209, opened 2021-05-19
VPN download-first experiment


alexgibson:
## Description

The VPN team would like to experiment with a download-first subscription experience. Proposed flow details can be seen here: https://www.figma.com/file/Rvu8RsVhTksMs48b5pRlup/Mozilla-VPN---Download-Experiment-Flowchart?node-id=0%3A1

The plan is to run this experiment this before https://github.com/mozilla/bedrock/issues/10199, so it's a P1.

Target audience:

- Mac / Windows platforms only.
- All locales / countries.

Experiment params will be:

- `entrypoint_experiment=vpn-landing-page-download-first`
- `entrypoint_variation=a|b`

---

#### Success Criteria
- [ ] A/B test is live in production for date TBD

alexgibson:
Some aditional requirements:

- If page URL contains both `entrypoint_experiment`, `entrypoint_variation`, and FxA flow params, then pass through the flow params to the VPN subscription links. This will help reconnect which variation someone is in when they come back from the product download to subscribe.

alexgibson:
Experiment design doc (Mozilla access only): https://docs.google.com/document/d/1igeYfvV_t1vZRsIw1P2AKS4CpO-RvM4Dc5_0ppx12-0/edit?usp=sharing

PR #10236, opened 2021-05-28
Update VPN download-first exp cookie ID

alexgibson:
## Description
This shouldn't effect anything, but was pointed out by Sarah on the VPN team. I could not resist updating it because I'm OCD.

## Issue / Bugzilla link
#10209


craigcook (approved):
🍪

Issue #10167, opened 2021-04-22
Update homepage for MR1


slightlyoffbeat:
Placeholder for work related to updating the homepage for MR1

craigcook:
Prototyping on demo: https://www-demo4.allizom.org/en-US/

slightlyoffbeat:
@craigcook will we be able to compare the CTR across the 4 variations?

craigcook:
> @craigcook will we be able to compare the CTR across the 4 variations?

Yeah, we can tag them to differentiate. I'd be surprised to see much difference between them, but who knows.

PR #10238, opened 2021-05-28
Fix/update home page tests [#10167]

craigcook:
## Description
I wrote tests assuming the MR1 promo would be everywhere, then neglected to update them when it was removed from DE and FR.

## Testing
Run `tests/functional/test_home.py`


stephaniehobson (approved):
R+ ✂️

Issue #10184, opened 2021-05-03
Update illustrations and screenshots on www.mozilla.org for MR1 


slightlyoffbeat:
## Description

MR1 will introduce an updated UI for Firefox including a new style for tabs. This means we will need to update various screenshots and illustrations across mozilla.org

---

#### Tasks
- [x]  Identify pages that will be affected
- [x]  Identify and catalog which assets on those pages need to be updated.
- [x]  Submit list of assets needed to creative team
- [x]  Receive updated assets from creative team
- [x]  create PR to update Bedrock
- [x]  changes go live on day of MR1

slightlyoffbeat:
I'll take on the ownership of all tasks before handing off to a dev

slightlyoffbeat:
We will have P1 assets from creative by Friday, May 21. 

Here is the document I created that audits assets and assigns priorities:

https://docs.google.com/spreadsheets/d/1Fsx_b-Vs0GOyLi_MoTBK_ZUdoNWsN2L76EUjX2l1aa8/edit?usp=sharing

PR #10245, opened 2021-06-01
Remove outdated assets after MR1 launch

craigcook:
## Description
Removes the switch for the MR1 launch to make those bits permanent, and deletes most of the redundant assets and CSS. 

The DE and FR home pages still have the old promo so that stuff remains, and the legacy home page keeps the older promo as fallback content for locales that haven't translated the MR1 promos.

## Issue / Bugzilla link
#10184 

## Testing
http://localhost:8000/
http://localhost:8000/firefox/new/

Make sure I didn't delete any files we still need.


stephaniehobson (approved):
R+ 📅

PR #10249, opened 2021-06-03
Add locale `sco` per community request

peiying2:
`sco` is for Scots, which is localized for Firefox and ready for Beta. There are links in the product that take users to mozilla.org.

### DIFF
--- a/l10n/configs/pontoon.toml
+++ b/l10n/configs/pontoon.toml
@@ -84,6 +84,7 @@ locales = [
     "rm",
     "ro",
     "ru",
+    "sco",
     "si",
     "sk",
     "sl",


peiying2:
@alexgibson or @pmac please review.

alexgibson:
Hmm, I tried this locally but http://localhost:8000/sco/ is still a 404 for me when `Dev=True`.

Looking in settings, it seems like it looks for the locale folder in `./data/locales`: https://github.com/mozilla/bedrock/blob/master/bedrock/settings/base.py#L138, which is for the old `.lang system.

@pmac should we need to update this to point to the `./data/www-l10n/` instead? I'm not entirely sure what repercussions that might have, if any?

pmac:
Yes we should, but for now that won't work either as `sco` doesn't exist in [that repo](https://github.com/mozilla-l10n/www-l10n) either.

pmac:
And we'll need to add it to the `PROD_LANGUAGES` setting before it'll work in prod.

alexgibson:
> Yes we should, but for now that won't work either as sco doesn't exist in that repo either.

@peiying2 can you add the new locale as a folder in the repo?

peiying2:
@pmac don't know how to create a folder under www-l10n. I created a PR, it doesn't look right: https://github.com/mozilla-l10n/www-l10n/pulls. This is quite different from the markdown folder type that I did. 

Also should the content from `en` be copied over to `sco` or it magically happens after pontoon.toml is merged?

pmac:
I don't know how that part works either. Might have to find out from whoever knows Pontoon best. Maybe Pontoon creates the folder? Have we added a new locale since moving to Fluent?

peiying2:
> I don't know how that part works either. Might have to find out from whoever knows Pontoon best. Maybe Pontoon creates the folder? Have we added a new locale since moving to Fluent?

No, I have not. This is a first. I will find out from our side. For creating the folder, is what I did correct?

pmac:
> For creating the folder, is what I did correct?

I'm not sure. It's hard to tell on GH what you did exactly. Git is strange in that it doesn't really have a concept of a folder, only paths for changes, so it doesn't track them if there's nothing in them. So we need to add some file to that folder to make it exist. I sometimes just add an empty `.gitignore` file to a folder I want to make sure exists when I clone the repo. But I guess we can also just wait until there are some real .ftl files to add to the folder. Not much point in supporting the locale until there are some translations I guess.

alexgibson:
Could we maybe copy/add the `brands.ftl` file to create some content?

pmac:
> Could we maybe copy/add the `brands.ftl` file to create some content?

That's a great idea since that file is not translated.

peiying2:
> > Could we maybe copy/add the `brands.ftl` file to create some content?
> 
> That's a great idea since that file is not translated.

OK, I am over thinking of this... Once the pontoon.toml file is merged, I will add the locale in Pontoon, the files will be available for localization - excluding the pages not available for all locales. No need to add the files before hand.

alexgibson:
> OK, I am over thinking of this... Once the pontoon.toml file is merged, I will add the locale in Pontoon, the files will be available for localization - excluding the pages not available for all locales. No need to add the files before hand.

Ok but just so you're clear, the `/sco/` pages will not be visible on www-dev for translators to preview until some kind of content exists that we can pull in. Is that OK?

peiying2:
> > OK, I am over thinking of this... Once the pontoon.toml file is merged, I will add the locale in Pontoon, the files will be available for localization - excluding the pages not available for all locales. No need to add the files before hand.
> 
> Ok but just so you're clear, the `/sco/` pages will not be visible on www-dev for translators to preview until some kind of content exists that we can pull in. Is that OK?

Understood. We should find out soon.

alexgibson (approved):

PR #10262, opened 2021-06-10
Update VPN staging plan IDs (Fixes #10260)

### LINKED ISSUE #10260: VPN staging IDs for variable pricing in US$ are incorrect
## Description

We're currently using the same VPN staging IDs for US$ variable pricing as we do for Euros. This creates a currency mismatch error when QA are trying to test the new flow. See: https://jira.mozilla.com/browse/FXA-3620

We need to get a new set of staging plan IDs from the subplat team (which are forthcoming).

alexgibson:
Fixes #10260

### DIFF
--- a/bedrock/settings/base.py
+++ b/bedrock/settings/base.py
@@ -1331,19 +1331,19 @@ VPN_VARIABLE_PRICING = {
     },
     'us': {
         '12-month': {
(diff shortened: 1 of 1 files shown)


alexgibson:
This change only effects www-dev and demo servers, and helps unblock QA who are based in Europe - so I'm going to go ahead and merge this and it's a very low risk change.

Issue #10257, opened 2021-06-08
Update moments page for VPN


slightlyoffbeat:
## Description

PMM team has requested the following updates to https://www.mozilla.org/en-US/firefox/welcome/10/

1. Update headline. Copy to be provided by creative team
2. Update CTA to link to pricing section of landing page: https://www.mozilla.org/en-US/products/vpn/?source=welcome10#pricing

UPDATE:
This will run as an a/b test with traffic cop.

### Both Variations:

**Logo:** VPN lockup logo (not just icon)

**CTA text:** Get Mozilla VPN

### Variation 1 (50% traffic)

**Headline:** Ending soon: $4.99/month for a secure connection on every device

**Paragraph:** Protect your entire internet connection — on your computer, your tablet, and even your phone—with this limited-time offer.

### Variation 2 (50% traffic)

**Headline:** Blur your location online for next-level privacy

**Paragraph:** Get Mozilla VPN for $4.99/month — this offer ends soon. Start now.

---

#### Success Criteria
- [ ] Able to provide signup attribution for users who landed on this moments page
- [ ] Able to track and compare CTR of each variation.
- [ ] Able to only send users from EN countries that have access to VPN and can buy in USD

slightlyoffbeat:
@craigcook This copy is in legal review, but is close to final and can be used as a placeholder:

*****

**Logo:** VPN lockup logo (not just icon)

**Headline:** Ending soon: $4.99/month for a secure connection on every device

**Paragraph text:** Protect your entire internet connection—on your computer, your tablet, and even your phone—with this limited-time offer.

**CTA text:** Get Mozilla VPN

slightlyoffbeat:
Craig and I had a conversation about various challenges with availability and pricing by geo.

According to possible targeting criteria, we can target by country. This should remove the need to do any geo-checking for price, or having anything built to handle users viewing this who cannot purchase VPN.

alexgibson:
> According to possible targeting criteria, we can target by country. This should remove the need to do any geo-checking for price, or having anything built to handle users viewing this who cannot purchase VPN.

We should probably double-check this when it launches to make sure the targeting is accurate. Here's the list of countries who saw the German VPN moments page and it doesn't seem geo-targeted, although I'm not 100% sure on what criteria Caro asked for (we left the price off the page because we were in doubt how it really worked).

![image](https://user-images.githubusercontent.com/400117/121860822-24b5a200-ccf1-11eb-8e19-2cd93d9afbd3.png)

PR #10278, opened 2021-06-14
Fix functional tests for welcome 10 [#10257]

craigcook:
## Description
Fixing the test I broke (just need to update the class in the button locator). Also excludes automation from traffic cop.

## Testing
`py.test --base-url http://localhost:8000 --driver Firefox --html tests/functional/results.html tests/functional/firefox/welcome/test_welcome_page10.py`

### DIFF
--- a/media/js/firefox/welcome/welcome10-experiment.js
+++ b/media/js/firefox/welcome/welcome10-experiment.js
@@ -33,7 +33,7 @@
         }
     };
 
-    if (href.indexOf('automation=true') {
+    if (href.indexOf('automation=true') === -1) {
         initTrafficCop();
     }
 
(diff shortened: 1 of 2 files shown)


alexgibson (approved):

media/js/firefox/welcome/welcome10-experiment.js
```diff
@@ -32,6 +32,9 @@
             cop.init();
         }
     };
-    initTrafficCop();
+
+    if (href.indexOf('automation=true') {
```
alexgibson, line 36: Oh wait, this should be:
```suggestion
    if (href.indexOf('automation=true') === -1) {
```

alexgibson (changes requested):

media/js/firefox/welcome/welcome10-experiment.js
craigcook: Doh, I was just thinking that. We only want to initTrafficCop when NOT `automation=true`

alexgibson (approved):

PR #10279, opened 2021-06-15
AMO attribution experiment template (Fixes #10207)

### LINKED ISSUE #10207: AMO download attribution experiment

alexgibson:
## Description
Adds an experiment template for the AMO team at the following URL:

http://localhost:8000/en-US/firefox/download/thanks/?xv=amo

This will help the AMO team understand the impact on installs if they link directly to `/download/thanks/` instead of `/firefox/new/` from AMO pages. The template contains some custom stub attribution logic that will delay the Firefox download (!) if someone lands on the page directly and does not already have a stub attribution cookie.

Obviously this is something we've been cautious about doing be default, but in the confines of an experiment we met and agreed the risk was low enough to test.

~**Note:** this PR relies on some logic added in https://github.com/mozilla/bedrock/pull/10216, so **please review that PR first**.~

## Issue / Bugzilla link
#10207

## Testing
This is up on demo2 for testing:

1.) On a Windows device/VM, open a **new Ingognito window** in Chrome.
2.) Open https://www-demo2.allizom.org/en-US/firefox/download/thanks/?xv=amo&utm_source=test&utm_campaign=test
3.) Wait for the Firefox download to being.
4.) Right click on the "Try downloading again" link and select "copy link address".
5.) Paste the link somewhere so you can inspect it.

The link should contain the following params:
- [x] `attribution_code`
- [x] `attribution_sig`

### DIFF
--- a/bedrock/firefox/templates/firefox/new/desktop/thanks-amo-exp.html
+++ b/bedrock/firefox/templates/firefox/new/desktop/thanks-amo-exp.html
@@ -1,0 +1,12 @@
+{# This Source Code Form is subject to the terms of the Mozilla Public
+ # License, v. 2.0. If a copy of the MPL was not distributed with this
+ # file, You can obtain one at http://mozilla.org/MPL/2.0/. -#}
+
+{% extends "firefox/new/desktop/thanks.html" %}
+
+{% block js %}
+  {% if switch('tracking-pixel') %}
+    {{ js_bundle('firefox_new_pixel') }}
+  {% endif %}
+  {{ js_bundle('firefox_new_thanks_amo_exp') }}
+{% endblock %}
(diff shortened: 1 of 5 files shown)


craigcook (approved):
LGTM and works as intended. r+ 🔌

PR #10303, opened 2021-06-25
Fall back to locale-based country if user data country is invalid

pmac:
For the newsletter/existing/<token>/ form (email preferences) it would
select the first country in the list if the country from the user data
was an invalid selection in the list. e.g. if the data said "Spain"
instead of "ES". This change makes the form default to the best country
based on the locale of the page.


alexgibson (approved):
Tested locally and works as advertised r+

PR #10305, opened 2021-06-25
Update number of VPN servers (Fixes #10304)

### LINKED ISSUE #10304: [VPN] Number of servers is incorrect
## Description

The [VPN landing page](https://www.mozilla.org/en-US/products/vpn/) states "Connect to more than 750 servers in over 30 countries", which is not technically correct.

The correct number should in fact be: 400+

alexgibson:
## Description
http://localhost:8000/en-US/products/vpn/

## Issue / Bugzilla link
#10304


pmac (approved):
#️⃣

Issue #10280, opened 2021-06-15
Add support for new EU countries to VPN landing page


alexgibson:
## Description

Country list and exact launch date is still TBD. 

Note: In some of these new countries there may not be a single national language, and instead of mix of two (or more) different languages that are commonly spoken. In these cases we'll need to add some new functionality to the landing page, whereby certain countries can use alternate subscription plan IDs based on the locale of the page. We still need to finalize exactly how this will work with the sub-plat team.

---

#### Success Criteria
- [x] VPN landing page supports additional countries for variable pricing plans.
- [x] New countries are behind a feature switch.

alexgibson:
To avoid needing to create dozens of plan IDs for every country we add in the future, we're going to change to a matrix of currency/language. This means we'll only need a smaller number of plans long term, but our current code will need a refactor to work in the new format.

alexgibson:
Pricing plans (Mozilla access only): https://docs.google.com/spreadsheets/d/1AEnBXYd_8P5Ft9OiXZli5BBkKmcmAIMB2cTO1kW7epU/edit?usp=sharing

alexgibson:
Some initial user stories:

- a user shows up in a not-supported country: go to waitlist (easy)
- a user shows up in a supported country and they're viewing the page in a matching locale: go to the plan for matching language and currency
- a user shows up in a supported country and they're viewing the page in a non-matching locale: what plan should they be shown? (this will either be determined by a preference from the product team, or alternatively be website visits).

PR #10311, opened 2021-07-05
Avoid using extend for VPN country lists (#10280)

alexgibson:
@pmac r?

I noticed that with `extend()` the country lists we're growing exponentially on each page refresh. I don't quite understand why exactly (my guess is something to do with the setting file acting differently?). Good thing is this is still behind a feature switch.

This PR seems to fix the issue?


pmac:
Bah 🤦‍♂️ should have noticed that in the initial review. The settings lists are kept in memory and `.extend()` updates them in place, so it just keeps adding to the same list. Using the `+` operator like you did creates a new list. This should do it. 💯

pmac (approved):
🧨

PR #10312, opened 2021-07-05
Update Edge information for the browser comparison page

ElijahPepe:
Per the new Edge, autoplay blocking, an in-browser screenshot, and support for macOS and Linux have all been introduced. Sources:

https://www.microsoftedgeinsider.com/en-us/download?platform=linux-deb
https://techcommunity.microsoft.com/t5/articles/introducing-web-capture-for-microsoft-edge/td-p/1721318
https://www.microsoft.com/en-us/edge
https://www.onmsft.com/how-to/how-to-block-auto-playing-media-in-the-new-microsoft-edge
https://answers.microsoft.com/en-us/microsoftedge/forum/all/how-to-prevent-video-autoplay-in-chromium-edge/7b0bf188-1915-4b61-b186-103fbc6ba556

### DIFF
--- a/bedrock/firefox/templates/firefox/browsers/compare/edge.html
+++ b/bedrock/firefox/templates/firefox/browsers/compare/edge.html
@@ -112,7 +112,7 @@
             <tr>
               <th scope="row">{{ ftl('compare-shared-autoplay-blocking') }}</th>
               <td><img alt="{{ ftl('compare-shared-yes') }}" src="{{ static('img/firefox/compare/icon-check.svg') }}" height="24" width="24"></td>
-              <td><img alt="{{ ftl('compare-shared-no') }}" src="{{ static('img/firefox/compare/icon-dash.svg') }}" height="24" width="24"></td>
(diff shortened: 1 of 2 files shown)


alexgibson (changes requested):
Thanks for the PR to update this! 

Looks like there is also a chart that needs the same updates at http://localhost:8000/en-US/firefox/browsers/compare/edge/

If you have time to update this as well, it would be very much appreciated? 🙂

ElijahPepe:
Added those changes there.

alexgibson (approved):

alexgibson:
Thanks!

yoasif:
Does beta really count as a Linux release for Edge?

ElijahPepe:
Really, what the OS availability row appears to count is if it's even available officially for all OSes. What "Beta" seems to be on Edge is just a quicker update time with the downside of containing more bugs that normally would be caught before an official release. A majority of users using Linux probably wouldn't encounter significant issues while using a beta version. The best way to solve this is a notice stating that the Linux version of Edge is in beta, but I don't see much of the need and rush to clarify that over a serious bug or outdated information.

PR #10318, opened 2021-07-08
Add GA custom dimension for AMO experiments (Fixes #10175)

### LINKED ISSUE #10175: Custom dimension for AMO experiments
## Description

We have determined that we cannot currently use `utm_campaign` for recording an experiment variation when we are linking to the Download Firefox page from within AMO. We would like to be able to record the experiment variation in order to use Google Analytics to analyze the conversion rate of these referrals during an experiment. We would like to use a custom GA dimension for this, either a new one that could be created for this purpose, or we could use an existing one if there is one available that makes sense.

---

#### Success Criteria
- [x] AMO is able to create a link to Download Firefox which includes a custom dimension which will be stored in GA for mozilla.org and can be used for reporting conversions.

ccing @alexgibson and @slightlyoffbeat with both of whom I have discussed this on Slack.

alexgibson:
## Description
For URLs that contain `experiment` and `variation` parameters, both of which match validation specific to AMO experiments, add a `dataLayer.push()` event to register then experiment as a custom dimension in GA.

http://localhost:8000/en-US/?experiment=20210708_amo_experiment_name&variation=variation_1_name

## Issue / Bugzilla link
#10175

## Testing
1.) Open the above URL.
2.) Open web console
3.) In the console enter `window.dataLayer` and hit enter.
4.) Verify that the array contains an abject with `{ "data-ex-name": "20210708_amo_experiment_name", "data-ex-variant": "variation_1_name" }`

### DIFF
--- a/media/js/base/core-datalayer-init.js
+++ b/media/js/base/core-datalayer-init.js
@@ -58,6 +58,19 @@
     if (typeof Mozilla.Utils !== 'undefined') {
         Mozilla.Utils.onDocumentReady(function() {
             initCoreDataLayer();
+
+            // Add GA custom dimension for AMO experiments (Issue 10175).
+            if (typeof window._SearchParams !== 'undefined') {
+                var params = new window._SearchParams().params;
+                var validParams = analytics.getAMOExperiment(params);
+
+                if (validParams) {
+                    dataLayer.push({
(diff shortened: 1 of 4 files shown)


craigcook (approved):
🚅

PR #10320, opened 2021-07-09
Stop serving JS to legacy IE browsers (Fixes #10192)

### LINKED ISSUE #10192: Stop bundling JS for legacy IE versions?
## Description

Still thinking through this a bit more, but hear me out...

- We currently serve a legacy `common-ie.js` bundle for old IE browsers (IE 6-9).
- The main role of this bundle is to serve Firefox downloads, although it also does a few other bits and bobs such as load GA, do Stub Attribution etc.
- Maintaining this code has several costs associated with it:
  - The existing IE bundle relies upon an old and vulnerable version of jQuery in order to function. Upgrading this lib is not an option, unfortunately. Later jQuery versions do not support old IE.
   - These old browsers have little to no test coverage, and most versions (IE 6-8) are now officially unsupported by SauceLabs. This means changes have to be manually QA'd by a dev with access to the right VM. The risk of JS breakage slipping through undetected is also higher because of this.
  - Many modern JS bundlers don't easily support such old IE versions, even with polyfills and transpiling. This makes it harder for bedrock to use more modern front-end tooling.

Proposal:

- I think (?) we might be able to get rid of most of this JS code:
  - We already know these visitors are on Windows and need the 32 bit Firefox installer. Can we give them the right file just using CSS and conditional comments?
  - There might be a couple of other scripts we need to load still, such as GA, HTML5 Shiv etc, but these can also be served via conditional comments and don't need to be bundled as such (we could load the scripts directly).

alexgibson:
## Description
- Removes `common-ie.js` bundle.
- Move IE specific logic to use conditional comments.
- Download buttons will simply trigger a file download using good-ol' plain links.
- 2 IE scripts remain which are linked directly via `<script>` tags and are not bundled:
  - `media/js/ie/libs/html5shiv.js` (to enable use of HTML5 sectioning elements in IE).
  - `media/js/ie/libs/gtm-snippet.js` (for basic GA / event support).
- 1 IE bundle remains, for stub attribution.

## Issue / Bugzilla link
#10192

## Testing
Testing on IE8/IE9 on a Windows VM is probably the easiest way to verify these changes:

- [x] Clicking download button on https://www-demo2.allizom.org/en-US/firefox/new/ should directly serve Firefox stub installer for Windows.
- [x] Clicking download buttons on https://www-demo2.allizom.org/en-US/firefox/channel/desktop/ should directly serve each respective pre-release build for Windows.
- [x] Clicking download button on https://www-demo2.allizom.org/en-US/firefox/developer/ should directly serve Dev Edition installer for Windows.

Most other changes are cosmetic, for which I've left comments and screenshots below.

Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/347694419


bedrock/firefox/templates/firefox/new/basic/base_download.html
```diff
@@ -68,6 +68,11 @@ <h2 class="c-sub-navigation-title">{{ ftl('sub-navigation-firefox-for-desktop',
 {% block content %}
 <main role="main" class="main-download" {% if v %}data-variant="{{ v }}"{% endif %}>
 
+  {# Show unsupported notification to IE 6/7 visitors on Windows XP/Vista #}
```
alexgibson, line 71: Moved this notification from the /thanks page to /new

image

media/css/protocol/components/_navigation-ie.scss
```diff
@@ -93,33 +75,28 @@ $image-path: '/media/protocol/img';
...
+        display: inline-block;
+        *display: inline;
+        zoom: 1;
+        margin-bottom: $spacing-2xl;
     }
 
-    // hide panel close buttons
-    .c-menu-button-close {
-        display: none;
+    // only display primary navigation links to IE visitors.
```
alexgibson, line 85: I made a call to only display the top-level navigation links for old IE browsers:

image

I figured this is probably enough, as all we can really hope for is for them to download a new browser (they are probably not going to get much out of the site itself).

Happy to try something else here if we think it's worth it.

alexgibson:
Ok I think this one is ready for review

alexgibson:
Note for reviewer: I verified that the GTM snippet is working ok on demo using IE9. I'll keep watch for events when this goes to production to make sure everything looks ok.

craigcook (approved):
Sorry for neglecting this PR.

In IE 8/9 on Windows 7 I get a download for 92.0b5 for Beta, which is correct, but just "Firefox Installer.exe" for both Dev Edition and Nightly. That's the stub installer but I'm testing this via Browserstack and can't actually install to verify it gets the correct build. But maybe this is correct? Do we not do Dev or Nightly builds for Win7?

In IE 6/7/8 on Windows XP I get a download for 52.0.9esr for all channels, which I believe is the very last build we ever released for XP so that seems like the best we can offer (there's no Beta/Dev/Nightly for XP).

So assuming the Win7 behavior is correct, that's an r+ from me. If we're in doubt, I suppose we can ask someone in release management.

alexgibson:
> In IE 8/9 on Windows 7 I get a download for 92.0b5 for Beta, which is correct, but just "Firefox Installer.exe" for both Dev Edition and Nightly. That's the stub installer but I'm testing this via Browserstack and can't actually install to verify it gets the correct build. But maybe this is correct? Do we not do Dev or Nightly builds for Win7?

I took a look to try and double check this (good call out!) The download links on the channel page point to:

- Beta: https://download.mozilla.org/?product=firefox-beta-latest-ssl&os=win&lang=en-US
- Dev Edition: https://download.mozilla.org/?product=firefox-devedition-stub&os=win&lang=en-US
- Nightly: https://download.mozilla.org/?product=firefox-nightly-stub&os=win&lang=en-US

If I `curl -v` each of these URLs I get:

- Beta: https://download-installer.cdn.mozilla.net/pub/firefox/releases/92.0b5/win32/en-US/Firefox%20Setup%2092.0b5.exe
- Dev Edition: https://download-installer.cdn.mozilla.net/pub/devedition/releases/92.0b5/win32/en-US/Firefox%20Installer.exe
- Nightly: https://download-installer.cdn.mozilla.net/pub/firefox/nightly/latest-mozilla-central-l10n/Firefox%20Installer.en-US.exe

These seem to be the same URLs as I see in prod right now, so they look OK to me.

alexgibson:
> In IE 6/7/8 on Windows XP I get a download for 52.0.9esr for all channels, which I believe is the very last build we ever released for XP so that seems like the best we can offer (there's no Beta/Dev/Nightly for XP).

This logic is all handled in Bouncer, so this looks correct to me yeah 👍

alexgibson:
@craigcook thanks for the review, I left some comments above.

I did spot on other thing I still needed to update with stub attribution, so I'm going to add a do-not-merge label on this until I've verified things are working OK.

craigcook:
> These seem to be the same URLs as I see in prod right now, so they look OK to me.

Ah that makes sense. I was just thrown by the filenames and didn't think to check that they lived at different paths.

alexgibson:
@craigcook Ok I've updated this so that the stub attribution bundle for IE works properly. You can verify this by visiting the demo links in the description above, and you should be able to see the `attribution_sig` and `attribution_code` params attached to the download links URLs. If all looks good to you still, then I think this is ready to merge. Thanks!

craigcook:
Verified, links are getting `attribution_sig` and `attribution_code` params.

PR #10325, opened 2021-07-12
Remove VPN pricing experiment code from welcome 10 [#10257]

craigcook:
## Description

The experiment has concluded and the $4.99 price is ending. This cleans up the experiment code and removes the price (just in case someone links directly to a variant URL, even with traffic cop disabled).

## Issue / Bugzilla link
#10257

## Testing
http://localhost:8000/firefox/welcome/10/


alexgibson (approved):
r+ 💸

PR #10326, opened 2021-07-12
Remove Firefox Lite whatsnew page [fix #10292]

### LINKED ISSUE #10292: Remove the Firefox Lite whatsnew page [30 June]
## Description

India and parts of Africa are still being shown a whatsnew page promoting Firefox Lite, which is being formally deprecated on **30 June 2021**. We should remove that page and disable the geo-based redirects.

---

#### Success Criteria
- [ ] Geo redirects removed; people in India and Africa should see the normal WNP (still depending on locale).
- [ ] Templates and assets deleted.
- [ ] Tests removed.

craigcook:
## Description
Firefox Lite is officially ended. This deletes the custom WNP we've been showing in India and parts of Africa, as well as related tests and assets. <s>I'm not sure if we should redirect the `/whatsnew/india/` and `/whatsnew/africa/` URLs, I assume not since they shouldn't have been indexed and likely don't have many direct links in the wild, so probably safe to let them 404.</s>

## Issue / Bugzilla link
#10292 

## Testing
http://localhost:8000/en-US/firefox/89.0/whatsnew/india/ (should redirect to /89.0/whatsnew/all/)
http://localhost:8000/en-US/firefox/89.0/whatsnew/africa/ (should redirect to /89.0/whatsnew/all/)


alexgibson:
@craigcook you're right that these pages shouldn't show up in search, but I do wonder if there's a lot of people out there who may have shared / bookmarked these URLs, seeing as the /whatsnew page has such a broad reach. Is it worth adding the redirect for this possibility perhaps?

alexgibson (approved):
Looks good locally, I tested the redirects and all seems well.

It would be nice to add a couple of tests for the redifrect, just to make sure things are working as they should. But otherwise r+

PR #10328, opened 2021-07-13
Remove unused macos attribution script (Fixes #9346)

### LINKED ISSUE #9346: Remove macOS download attribution script?
## Description

We have a macOS download attribution script that was added some 2 years ago here: https://github.com/mozilla/bedrock/blob/master/media/js/base/stub-attribution-macos.js

As far as i know, attribution for macOS downloads has never worked fully - and I don't know of any plans to get that working. I don't think it serves us any favors to keep on maintaining and shipping dead code. If this is not being used, there's always Git history to recover things later.

---

#### Success Criteria
- [ ] macOS attribution script is removed from the repo.

alexgibson:
## Description
This script has lived in the repo for several years but has never been used. We've got the OK in the issue to remove it. If we ever need it in the future, Git history is always available.

## Issue / Bugzilla link
#9346

## Testing
- [x] Did I miss any stray references to the file?


maureenlholland (approved):
R + 🧹

Issue #10335, opened 2021-07-14
Add vanity URL for VPN product


pmac:
## Description

The VPN page lives at mozilla.org/products/vpn, but I find myself typing mozilla.org/vpn and hoping that that will take me to where I want to be. I propose we add a redirect from `/vpn` to `/products/vpn/`.

pmac:
@reemhamz this should be a lot like https://github.com/mozilla/bedrock/pull/9572

reemhamz:
Created a redirect that enables a user to add `/vpn` at the end of the mozorg URL to redirect them directly to `/products/vpn`

PR #9572, opened 2020-10-20
Fix #9571: Add more redirects for UFI campaign

pmac:

PR #10338, opened 2021-07-19
Issue #10335: Add simplified VPN redirect

reemhamz:
## Description
Created a redirect that enables a user to add `/vpn` at the end of the mozorg URL to redirect them directly to `/products/vpn`

## Issue / Bugzilla link
Issue #10335  

## Testing
Ran `py.test -r a tests/redirects` and all tests have passed.

### DIFF
--- a/bedrock/mozorg/redirects.py
+++ b/bedrock/mozorg/redirects.py
@@ -743,7 +743,7 @@ redirectpatterns = (
 
     # Issue 9984
     redirect(r'^/about/legal/fraud-report/?$', '/about/legal/defend-mozilla-trademarks/'),
-    
+
     # Issue 10335
     redirect(r'^vpn/?$', 'products.vpn.landing'),
 )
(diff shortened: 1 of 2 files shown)


bedrock/mozorg/redirects.py
```diff
@@ -743,4 +743,7 @@ def decider(request, **kwargs):
 
     # Issue 9984
     redirect(r'^/about/legal/fraud-report/?$', '/about/legal/defend-mozilla-trademarks/'),
+
```
pmac, line 746: The CI tests failed due to some white space in this blank line. We use the `flake8` tool to verify some style things, and this is one of them. I have my editor set to auto remove trailing whitespace to avoid this kind of thing.

pmac (changes requested):
It works!! 🎉 

We should add a test though, to ensure that no future change breaks this redirect. If you look at #9572 you'll see the additional changed file adds a test. It should be something like

```python
url_test('/vpn/', '/products/vpn/'),
```

pmac (approved):
Looks perfect! 🍰

Issue #10341, opened 2021-07-20
Remove/replace deprecated Protocol components


craigcook:
## Description

[Protocol 14.0.0](https://github.com/mozilla/protocol/releases/tag/v14.0.0) introduced some new components that should replace some previously existing components. Namely the new Split can replace the old Hero and Feature Card components, and the new Picto replaces the old Picto Card. However, those older components are still in Protocol for backwards compatibility, giving us time to gradually phase them out.

Bedrock updated to Protocol 14 a while ago, now we should work on replacing deprecated components when and where we can.

Here's a rough audit of templates that use these components, mostly using macros but some have the markup hardcoded: https://docs.google.com/spreadsheets/d/1VLijrBEPevlrZ2RbYt6GCDxkXXHA-GiHqhsTzT55xnY/

craigcook:
Looking through some of these pages I remembered that the Hero component defaults to a centered layout and the image is optional. Several pages use that basic centered hero, so Split isn't a replacement for those. I think we talked about making another component for this use case (a centered page header type thing) but I don't remember what we concluded. It could be reproduced with other elements/components (namely the Section Heading) but but it's also maybe enough of a design pattern to justify a component.

PR #10351, opened 2021-07-28
Replace deprecated components [#10341]

craigcook:
## Description
Replacing picto cards and feature cards on feature pages and heroes on comparison pages.

Most of the comparison pages weren't using the hero macro and already had a semi-customized hero, so I just made it a fully custom element only used on these pages. I think we may be creating a new simple hero component soon and could update these pages again, but for now I think just de-coupling them from the old hero is a good step.

This also replaces most of the feature icons with the ones we added to Protocol a long time ago, finally making good on a promise.

## Issue / Bugzilla link
#10341 

## Testing
http://localhost:8000/firefox/browsers/compare/ et al
http://localhost:8000/firefox/features/ et al


bedrock/firefox/templates/firefox/browsers/compare/index.html
```diff
@@ -19,26 +21,21 @@
...
-        <p>{{ ftl('compare-index-looking-for-a-better-browser', fallback='compare-index-looking-for-a-better-fallback') }}</p>
-      </div>
-
-      {{ download_firefox(dom_id='compare-hero-download', download_location='primary cta') }}
-
+{% call split(
+    image_url='img/firefox/compare/compare-browser-windows.png',
+    include_highres_image=True,
+    block_class='',
+    theme_class='',
```
alexgibson, line 28: Do these need to be defined as empty strings? (the defaults seem to be `None` in the macro)

bedrock/firefox/templates/firefox/features/includes/features-content.html
```diff
@@ -1,82 +1,121 @@
...
-      </div>
-    </section>
+  <ul class="mzp-l-columns mzp-t-columns-three">
+    {% call picto(
+      base_el='li',
+      image_url='protocol/img/icons/brand/orange/speedometer.svg',
+      image_width='32',
+      body=True,
+    ) %}
+      <p><a href="{{ url('firefox.features.fast') }}" rel="external">{{ ftl('features-shared-browse-faster') }}</a></p>
```
alexgibson, line 13: Remove `rel="external"` on links here that are same site?

alexgibson:
The hero images on the following three pages look a bit too large to me?

- http://localhost:8000/en-US/firefox/features/block-fingerprinting/
- http://localhost:8000/en-US/firefox/features/translate/
- http://localhost:8000/en-US/firefox/privacy/safe-passwords/

image

alexgibson (changes requested):
There's a lot of changes here, but I pulled each page locally and everything looks respectable (bar the large hero images on a few pages). I took a pass through the code changes and only noticed a couple of nits.

Not a blocker for this PR, but it would be great if the new split macro supported optional lazy-loading like the feature-card macro does. I don't think it effects any images on these pages in question though. r+wc 🖼️

bedrock/firefox/templates/firefox/browsers/compare/index.html
craigcook: Oops, that was a leftover from pasting that I neglected to remove.

craigcook:
The huge images are because I hadn't included the `mzp-l-split-media-constrain-height` class, and never noticed because of my ancient low-res monitor. Fixed now!

alexgibson (approved):
Looking good on my monitor! r+

Issue #10352, opened 2021-07-28
Link to user interviews from MR1 home promos


craigcook:
## Description

We're posting profiles on the users featured in the home page promos and would link to link to them. Two are up already, the other two coming soon:

https://blog.mozilla.org/en/products/firefox/sidebar-lalo-luevano-restauranteur-bodega/
https://blog.mozilla.org/en/products/firefox/gary-linden-big-wave-surfer/

slightlyoffbeat:
@slightlyoffbeat ping MJ on this

PR #10353, opened 2021-07-28
Link to blog interviews from home promos [#10352]

craigcook:
## Description
Adds links for Gary and Lalo. The other two are forthcoming in a the next few weeks.

## Issue / Bugzilla link
#10352 

## Testing
http://localhost:8000/

You'll probably have to reload the page several times to get both promos (it's random)

### DIFF
--- a/media/css/mozorg/home/home-mr1-promo.scss
+++ b/media/css/mozorg/home/home-mr1-promo.scss
@@ -86,11 +86,6 @@ $image-path: '/media/protocol/img';
     a:link,
     a:visited {
         color: get-theme('body-text-color');
-
-        &:hover,
-        &:active {
-            color: get-theme('link-color-hover');
-        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


media/css/mozorg/home/home-mr1-promo.scss
```diff
@@ -82,6 +82,16 @@ $image-path: '/media/protocol/img';
...
     position: relative;
     z-index: 11;
+
+    a:link,
+    a:visited {
+        color: get-theme('body-text-color');
+
+        &:hover,
+        &:active {
+            color: get-theme('link-color-hover');
```
alexgibson, line 92: From a consistency point of view, this is the only link on the page that seems to change color on hover/focus. Maybe that's ok? I'll leave the call to you

alexgibson (approved):
One question on the link hover state, but not a blocker. r+

media/css/mozorg/home/home-mr1-promo.scss
craigcook: Oh, you're right. Maybe the underline is enough. I'll fix.

alexgibson (approved):

PR #10363, opened 2021-08-02
Update classList polyfill (Fixes #6221)

### LINKED ISSUE #6221: Update bedrock's classList polyfill
I just uncovered a bug in IE 9 when using bedrock's [classList polyfill](https://github.com/mozilla/bedrock/blob/master/media/js/base/class-list-polyfill.js):

Given the following HTML:

```html
<div class="foo bar buzz"></div>
```

Running the following JS:

```javascript
document.querySelector('.foo').classList.remove('bar')
```

Will result in:

```html
<div class="foobuzz">
```

Instead of:

```html
<div class="foo buzz">
```

Our polyfill was [taken from MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#Polyfill) around 2 years ago, but it looks like it's been updated since. We should investigate replacing the polyfill we use in bedrock with the new version.

alexgibson:
## Description
Updtes our `classList` polyfill to the latest one recommended on MDN https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#Polyfill

## Issue / Bugzilla link
#6221

## Testing
Test run: https://gitlab.com/mozmeao/www-config/-/pipelines/346423583


maureenlholland (approved):
r+ ✅

PR #10366, opened 2021-08-03
2021 H1 Transparency report (Fixes #10365)

### LINKED ISSUE #10365: 2021 H1 Transparency report
## Description

Doc (Mozilla access only): https://docs.google.com/document/d/1ckFLilcdQG9otFvGyWazQ-NP38DzcJ8lRBBcA83djZg/edit?usp=sharing

---

#### Success Criteria
- [ ] 2021 H1 report is live in prod

alexgibson:
## Description
Jan-Jun 2021 TR report

## Issue / Bugzilla link
#10365

## Testing

http://localhost:8000/en-US/about/policy/transparency/jan-jun-2021/

### DIFF
--- a/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/reports-nav.html
+++ b/bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/reports-nav.html
@@ -8,7 +8,6 @@
         href="{{ url('mozorg.about.policy.transparency.index') + '#definitions' }}">Definitions</a>
     </li>
     {# This link should point to the latest report. Past reports are added to 'includes/past-reports.html' #}
-    {# There are also links in the faq that need to be updated to point to the latest report. #}
     <li>
       <a class="mzp-c-button" href="{{ url('mozorg.about.policy.transparency.jan-jun-2021') }}">January-June 2021 Report</a>
     </li>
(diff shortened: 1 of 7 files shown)


alexgibson:
This is ready for review, but should not merge until Thursady 5 August.

bedrock/mozorg/templates/mozorg/about/policy/transparency/includes/reports-nav.html
```diff
@@ -10,7 +10,7 @@
     {# This link should point to the latest report. Past reports are added to 'includes/past-reports.html' #}
     {# There are also links in the faq that need to be updated to point to the latest report. #}
```
alexgibson, line 11: Note for reviewer: I could not find any FAQ link that needed updating, but a double-check would be good just in case I missed something.

craigcook: I think there were some links originally that must have been removed in a later edit. We should delete this comment to avoid future confusion.

craigcook (approved):
Looks good. r+wc

Not specifically related to this PR but I noticed these pages were getting the body text color from the Firefox theme and it took me a minute to track down why: both `index.html` and `report-base.html` should be extending `base-protocol-mozilla.html`. If we do that we could also remove some redundant CSS (these pages are explicitly setting Zilla Slab on a bunch of elements that should inherit from the theme)... but we don't need to turn this PR into a bunch of theming cleanup, happy to merge as-is and do that stuff separately. Up to you.

alexgibson:
> Not specifically related to this PR but I noticed these pages were getting the body text color from the Firefox theme and it took me a minute to track down why: both index.html and report-base.html should be extending base-protocol-mozilla.html. If we do that we could also remove some redundant CSS (these pages are explicitly setting Zilla Slab on a bunch of elements that should inherit from the theme)... but we don't need to turn this PR into a bunch of theming cleanup, happy to merge as-is and do that stuff separately. Up to you.

Nice catch! Seemed like a small enough change so I fixed up the base templates and CSS font overrides 👍

PR #10368, opened 2021-08-04
Add media_after param to split macro (fixes #10367)

### LINKED ISSUE #10367: Add stack order param to Split macro
While working on [updating Protocol components](https://github.com/mozilla/bedrock/issues/10341) with Split, we realized the bedrock Split macro always stacks text first on small screen sizes. This is not the current behaviour of most components we want to update to Split.

We need to add a param to the bedrock Split macro to stack either image or text first.

maureenlholland:
## Description

Most of the deprecated components we want to upgrade to use Split stack media before text. In the bedrock split macro, we always set media second, but the Protocol component can handle either source order. This PR updates the bedrock macro to add a media_after param and sets the stack order as media first by default.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/10367

## Testing

The following should stack media after text:
http://localhost:8000/en-US/
http://localhost:8000/en-US/firefox/browsers/quantum/
http://localhost:8000/en-US/firefox/browsers/chromebook/
http://localhost:8000/en-US/about/policy/lean-data/
http://localhost:8000/en-US/about/policy/lean-data/stay-lean/
http://localhost:8000/en-US/about/policy/lean-data/build-security/
http://localhost:8000/en-US/about/policy/lean-data/engage-users/
http://localhost:8000/en-US/firefox/developer/ (bottom dark theme section: Download the Firefox browser made for developers)

The following should stack media before text (sample urls, complete list is quite long now):
http://localhost:8000/en-US/firefox/developer/ (near top light theme section: The browser made for developers)
http://localhost:8000/en-US/firefox/features/bookmarks/

### DIFF
--- a/bedrock/base/templates/macros-protocol.html
+++ b/bedrock/base/templates/macros-protocol.html
@@ -66,17 +66,30 @@
 
 
 {# Split: https://protocol.mozilla.org/patterns/organisms/split.html #}
-{% macro split(id=None, block_class=None, theme_class=None, body_class=None, image_url=None, media_class=None, include_highres_image=False, l10n_image=False, image_alt='') -%}
+{% macro split(id=None, block_class=None, theme_class=None, body_class=None, image_url=None, media_class=None, include_highres_image=False, l10n_image=False, image_alt='', media_after=False) -%}
 
 <section {% if id %}id="{{ id }}"{% endif %} class="mzp-c-split{% if block_class %} {{ block_class }}{% endif %}">
   {% if theme_class %}
   <div class="mzp-c-split-bg {{ theme_class }}">
   {% endif %}
   <div class="mzp-c-split-container">
+    {% if image_url and not media_after %}
+    <div class="mzp-c-split-media{% if media_class %} {{ media_class }}{% endif %}">
+      {% if include_highres_image %}
+        {{ high_res_img(image_url, {'alt': image_alt, 'l10n': l10n_image, 'class': 'mzp-c-split-media-asset'}) }}
+      {% else %}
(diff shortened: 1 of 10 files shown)


alexgibson (approved):
Tested the URLs provided and everything works as described, nice! r+

My only thought was would it make sense to default to media _after_ the copy, instead of before? I'm not sure what the most common usecase is in bedrock, but at a guess if we're replacing hero components with split then using media-after might make sense? I had a look on the Protocol site and most of the examples seem to default to media after as well.

Thoughts?

maureenlholland:
> My only thought was would it make sense to default to media _after_ the copy, instead of before? I'm not sure what the most common usecase is in bedrock, but at a guess if we're replacing hero components with split then using media-after might make sense? I had a look on the Protocol site and most of the examples seem to default to media after as well.
> 
> Thoughts?

It was a 50/50 toss up in my head. Hero replacements generally need media after, Feature Card replacements generally need media before. I went with media_after because it was already a param in Feature Card and features tend to be used multiple times on a page whereas hero is once per page. But it's a close call. Maybe best to double check with @craigcook when he is available?

alexgibson:
> Maybe best to double check with @craigcook when he is available?

Yeah I'll defer to both of you here. Whichever way works for me ultimately :)

craigcook:
I have no strong feelings one way or the other. A case can be made for either being the default, but it does seem like we currently have more heroes to replace, than feature cards so maybe media after wins.

alexgibson:
I think we can always change this later if we wanted, what we have here works! merging :)

PR #10369, opened 2021-08-05
Break up common.js bundle (Fixes #9321)

### LINKED ISSUE #9321: Break up common.js bundle

alexgibson:
## Description
Splits up `common.js` into three separate bundles with more distinct responsibilities:

- `lib.js` (libraries, utilities, polyfills).
- `ui.js` (global components etc).
- `data.js` (analytics and referrals).

Happy to bikeshed the naming if others have ideas or suggestions.

## Issue / Bugzilla link
#9321

## Testing
Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/350097824

### DIFF
--- a/bedrock/base/templates/base-protocol.html
+++ b/bedrock/base/templates/base-protocol.html
@@ -116,7 +116,9 @@
 
     {% block site_js %}
       <!--[if !IE]><!-->
-        {{ js_bundle('common') }}
+        {{ js_bundle('lib') }}
+        {{ js_bundle('ui') }}
+        {{ js_bundle('data') }}
       <!--<![endif]-->
 
       <!--[if IE]>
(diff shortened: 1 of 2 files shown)


craigcook (approved):
🐰

PR #10383, opened 2021-08-11
Mozilla logo scaled

AtharvaParkale:
## Testing
I have property scaled the 'Mozilla ' logo with the 'firefox' logo.

### DIFF
--- a/media/css/base/page-not-found.scss
+++ b/media/css/base/page-not-found.scss
@@ -70,7 +70,9 @@ main {
     .logo-mozilla {
         display: inline-block;
         height: 50px;
-        transform: translateY(-6px);
+        width: 141px;
+        padding-left: 30px;
+        transform: translateY(-2px);
 
         &:after {
             @include bidi(((float, right, left),));
@@ -78,7 +80,7 @@ main {
             content: '';
             height: 50px;
             margin: -10px $spacing-lg 0;
-            width: 2px;
+            width: 0px;
         }
     }
 


media/css/base/page-not-found.scss
```diff
@@ -70,15 +70,17 @@ main {
     .logo-mozilla {
         display: inline-block;
         height: 50px;
-        transform: translateY(-6px);
+        width: 141px;
```
craigcook, line 73: I think a better way to do this is to add a `width="140"` attribute to the logo image in `bedrock/base/templates/base/base-error.html` rather than setting the width in CSS.

media/css/base/page-not-found.scss
```diff
@@ -70,15 +70,17 @@ main {
...
+        padding-left: 30px;
```
craigcook, line 74: This padding doesn't seem necessary and is causing the logo to misalign with the content.

media/css/base/page-not-found.scss
```diff
@@ -70,15 +70,17 @@ main {
...
+        transform: translateY(-2px);
 
         &:after {
             @include bidi(((float, right, left),));
             background: $color-marketing-gray-30;
             content: '';
             height: 50px;
             margin: -10px $spacing-lg 0;
-            width: 2px;
+            width: 0px;
```
craigcook, line 83: This should still be `2px` to make the dividing line visible.

media/css/base/page-not-found.scss
```diff
@@ -70,15 +70,17 @@ main {
...
-        transform: translateY(-6px);
+        width: 141px;
+        padding-left: 30px;
+        transform: translateY(-2px);
 
         &:after {
             @include bidi(((float, right, left),));
             background: $color-marketing-gray-30;
             content: '';
             height: 50px;
```
craigcook, line 81: I think if we increase this to `60px` it'll look better (the dividing line will be slightly taller than the two logos).

craigcook (changes requested):

media/css/base/page-not-found.scss
AtharvaParkale: Will look into it

craigcook:
Closing in favor of #10502. Sorry, but thanks for your help!

PR #10385, opened 2021-08-13
Use Split on Firefox Products page, fix layout [#10341]

craigcook:
## Description
When we updated the Browsers page in https://github.com/mozilla/bedrock/commit/346aace714ff65b0410d256ec875e896b47f6b83 I failed to notice that the Products page uses the same style sheet, so some styling was broken. This updates the Products page and makes a few other minor tweaks.

This is currently broken in production (pushed only yesterday) so we should try to fix this ASAP.

## Testing


reemhamz (approved):
Looks good to me!

Issue #10166, opened 2021-04-22
Replace Lockwise with VPN on various pages


slightlyoffbeat:
The lockwise app is being phased out. A full communication plan and strategy around the future of lockwise is coming. In the meantime, we should replace lockwise with VPN in various discovery pages/paths. 

Remove lockwise from:
- [ ] the floating promo box on /new (and anywhere else it shows up)

Replace Lockwise with VPN on:

- [ ] https://www.mozilla.org/en-US/firefox/
- [ ] https://www.mozilla.org/en-US/firefox/products/
- [ ] https://www.mozilla.org/en-US/firefox/privacy/products/
- [ ] https://www.mozilla.org/en-US/firefox/accounts/

We will not touch the lockwise landing page for now. Work on that page will be covered by a separate issue at a later time.

Success Criteria:
- [x] Thumbs up on visuals and copy from creative studio
- [x] Thumbs up on approach by UX
- [x] coded on demo
- [x] Creative has reviewed visuals + copy on demo and given thumbs up.
- [ ] Links have appropriate parameters
- [x] Don't show content to non-EN if it isn't localized yet.
- [ ] Launched

slightlyoffbeat:
There is overlap with other cards that reference removing lockwise, but also try and cover additional changes. We should consider closing or reworking those cards after this work is completed:

https://github.com/mozilla/bedrock/issues/9976

https://github.com/mozilla/bedrock/issues/9905

slightlyoffbeat:
Affected pages include:

/firefox/products
/firefox/privacy/products
/firefox/new (the floating box with products)
/firefox/lockwise
/firefox/accounts

alexgibson:
@slightlyoffbeat where should we redirect /firefox/lockwise to? Is there some SUMO article or blog post that explains to people what is happening with the product?

slightlyoffbeat:
@alexgibson either just /firefox or a SUMO article. I reached out to SUMO to see if an article exists (if not, I'll request one gets made). I also reached out to Mei to see if she has an opinion on this. I'll update here when I have an answer.

slightlyoffbeat:
As of writing this comment, this is still blocked, but things are moving fast and this may be prioritized for next week.

slightlyoffbeat:
No longer blocked.

slightlyoffbeat:
@craigcook added a bullet to success criteria that we shouldn't show to non-EN unless it is translated.

slightlyoffbeat:
[...]

PR #10386, opened 2021-08-16
Remove more references to Lockwise [#10166]

craigcook:
## Description
This just removes some more references to Lockwise but doesn't specifically replace them with VPN because that doesn't seem appropriate for these places.

A few other pages still link to the Lockwise page in copy but I didn't change those. We'll need to decide the fate of that page eventually and that's when we can update any remaining links.

## Issue / Bugzilla link
#10166 

## Testing
http://localhost:8000/firefox/ (removed structured data)
http://localhost:8000/firefox/products/ (removed structured data)
http://localhost:8000/firefox/accounts/ (subnav)
http://localhost:8000/firefox/sync/ (subnav)
http://localhost:8000/firefox/features/safebrowser/
http://localhost:8000/firefox/more/
http://localhost:8000/firefox/welcome/3/


alexgibson (approved):
r+

Is it worth removing some of the strings from the pages touched here if they are no longer needed?

Issue #10387, opened 2021-08-16
New page for Firefox Tips


slightlyoffbeat:
## Description

There is a social campaign that talks about Firefox tips, tricks, and useful features. This is a simple page that will display various videos that a user can click on to view Mozilla employees talking about their favorite tips.

This was designed using Protocol components.

---

Notes:

Content: N/A
Media: N/A
Design: Please consult with Elise before starting. She can walk through the protocol components and expected behavior.

#### Success Criteria
- [x] URL set up at: firefox/features/tips (URL to be confirmed with M.J.)
- [ ] Event tracking that allows us to see what videos a user is clicking on to open.
- [x] EN only (Social campaign and videos are EN only and will not be translated for now)
- [x] added to /firefox/features/
- [x] All variations of this page have a canonical reference to /firefox/features/tips
- [x] When a card is clicked, the video opens in a modal and auto plays.
- [x] Staged on Demo
- [x] approval from editorial
- [x] approval from comms

slightlyoffbeat:
There's been a bit of discussion on this today. A few notes:

- There will be a page a tips/tricks page that does not use any secret menu language at /firefox/features/tips
- there will be a page variation that swaps out some copy for copy that uses the secret menu language
- we should make sure that any variations point to /firefox/features/tips as the canonical source
- there was some design feedback so dev is on hold until designs are updated.

PR #10399, opened 2021-08-25
Firefox tips 10387

reemhamz:
## Description
There is a social campaign that talks about Firefox tips, tricks, and useful features. This is a simple page that will display various videos that a user can click on to view Mozilla employees talking about their favorite tips.

## Issue / Bugzilla link
#10387 

## Testing
http://localhost:8000/en-US/firefox/features/tips/


alexgibson (dismissed):
I took a pass and overall the page is looking really great, nice work!

I've suggested a fix below for the variation view to make the split videos work. Also if this page isn't going to be localized (which it sounds like is a no), then I'd suggest keeping these strings out of an .ftl file and instead hard-coding them in the template.

craigcook (approved):
Nice work! 🌵

PR #10384, opened 2021-08-12
Update Hero components with Split

reemhamz:
## Description
Replacing the soon-to-be deprecated [Hero](https://protocol.mozilla.org/patterns/organisms/hero.html) components in some pages with the new [Split](https://protocol.mozilla.org/patterns/organisms/split.html) component.

## Issue / Bugzilla link
#10341 

## Testing
http://localhost:8000/firefox/lockwise/
http://localhost:8000/firefox/switch/
http://localhost:8000/en-US/

PR #10407, opened 2021-08-31
Fix home page promo switch logic

alexgibson:
## Description
Fixes test failure in https://gitlab.com/mozmeao/www-config/-/jobs/1549619134

See https://github.com/mozilla/bedrock/pull/10384/files#r699390512 for context.

## Issue / Bugzilla link
N/A

## Testing
MR1 promo should default to being visible on http://localhost:8000/en-US/


reemhamz:
Ah thanks for catching that before prod. r+

reemhamz (approved):
Ah thanks for catching this before prod. r+ 💫

PR #10412, opened 2021-09-01
Fix split call on dev edition pages [no bug]

craigcook:
## Description
Restores (and renames) `id` param in the split macro that was lost in a different merge.
Adds a `base_el` param while I'm in here anyway, which defaults to `section` so doesn't actually change anycurrent uses of this macro but gives us future opportunities.

## Testing
The id param was only used in two places.

http://localhost:8000/en-US/firefox/developer/
http://localhost:8000/en-US/firefox/91.0a2/whatsnew/all/


pmac (approved):
🚁

PR #10413, opened 2021-09-01
FIx home page lazyloading [no bug]

craigcook:
## Description
Card images on the old (non-Contentful) home page were failing to load because the lazyloading script was moved/renamed in https://github.com/mozilla/bedrock/commit/8beef7541e079c73446ab71e56b49e1215b1eed2. This restores the JS for the old home page but shouldn't interfere with Contentful.

## Testing
http://localhost:8000/


pmac (approved):
🖼️

PR #10414, opened 2021-09-01
Update id param in Split macro calls [no bug]

craigcook:
## Description
I got cocky and renamed a parameter in a macro without thoroughly searching for every place that macro was called with the old parameter. This, I think, catches them all.

## Testing
http://localhost:8000/firefox/pocket/
http://localhost:8000/firefox/privacy/book/
http://localhost:8000/foundation/annualreport/2019/


pmac (approved):
🆔

PR #10417, opened 2021-09-02
Turn shared VPN string in multiple contexts into separate strings

alexgibson:
Fixes: https://github.com/mozilla-l10n/www-l10n/issues/225


craigcook (approved):
🧶

PR #10424, opened 2021-09-07
Add platform to images in docker-compose

pmac:
These changes will allow the build to succeed on machines with non-intel processor architectures. This is so far mostly an issue for Mac computers with the M1 chip (ARM64 arch).

See https://docs.docker.com/desktop/mac/apple-silicon/#known-issues

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -10,6 +10,7 @@ services:
       context: .
       target: assets
     image: mozmeao/bedrock_assets:${GIT_COMMIT:-latest}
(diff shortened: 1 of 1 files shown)


pmac:
This is the result of some debugging of problematic docker builds with @maureenlholland 

@alexgibson r? mainly just to make sure that a build still works fine on another Intel Mac (I've already tested).

alexgibson (approved):
This worked for me locally too, r+ 🐿️

Issue #10408, opened 2021-09-01
Add linkable targets to the principles on the Manifesto page


dveditz:
## Description

Mozillians often illustrate their motivations or guide rails for certain projects through references to one or another principal from our Manifesto, but like most of our site there are almost no interior link targets available. There are currently #sec-principles and #principles to nearly the same spot but that's about it.

Please add ID attributes to the Manifesto page so #principle-01, #principle-02, etc (or any reasonable variation)work as link targets.

Of the two links that _do_ exist, https://www.mozilla.org/en-US/about/manifesto/#principles links to where we'd want #principle-01 to go but it's not a very satisfactory experience. If you're already reading the page it can be alright, but if you follow that link from a different page (this one, for instance) the principle will be half-covered by a floating header/advertisement bar. This bad header-covering-link-target behavior is site-wide (example: https://www.mozilla.org/en-US/about/manifesto/details/#advancing) so maybe that aspect should be dealt with more generically in the site's code or styles.

---

#### Success Criteria

 - [ ] Principles from the Manifesto can be individually referenced
 - [ ] The targeted principle is readable when linked from a separate document

PR #10426, opened 2021-09-07
Add linkable targets to the principles on the Manifesto page

nathan-barrett:
## Description
Add linkable targets to the principles on the Manifesto page - added an id to each individual principle and move the id #principles to the header for the principles section so it is viewable when navigating from a different page.
## Issue / Bugzilla link
#10408 
## Testing
localhost:8000en-US/about/manifesto/#principles


craigcook (approved):
⚓

Issue #10429, opened 2021-09-08
Add the user's country to the CDN cache key


pmac:
We currently only use the country in views for redirects since we have to disable the CDN cache for them. But the CDN will include the user's country in the cache key which would allow us to use the country in any view we needed without need for any added complexity.

pmac:
I tested the above with Mozilla VPN. 

```shell
$ # VPN: Off
$ curl -v https://www-dev.allizom.org/en-US/cdn-geo-test/ 2>&1 | egrep "cf-cache-status|country_code"
< cf-cache-status: HIT
{"country_code": "US"}

$ # VPN: Germany
$ curl -v https://www-dev.allizom.org/en-US/cdn-geo-test/ 2>&1 | egrep "cf-cache-status|country_code"
< cf-cache-status: MISS
{"country_code": "DE"}
$ curl -v https://www-dev.allizom.org/en-US/cdn-geo-test/ 2>&1 | egrep "cf-cache-status|country_code"
< cf-cache-status: HIT
{"country_code": "DE"}
$ curl -v https://www-dev.allizom.org/en-US/cdn-geo-test/ 2>&1 | egrep "cf-cache-status|country_code"
< cf-cache-status: HIT
{"country_code": "DE"}

$ # VPN: Off
$ curl -v https://www-dev.allizom.org/en-US/cdn-geo-test/ 2>&1 | egrep "cf-cache-status|country_code"
< cf-cache-status: HIT
{"country_code": "US"}
$ curl -v https://www-dev.allizom.org/en-US/cdn-geo-test/ 2>&1 | egrep "cf-cache-status|country_code"
< cf-cache-status: HIT
{"country_code": "US"}

$ # VPN: Toronto, Canada
$ curl -v https://www-dev.allizom.org/en-US/cdn-geo-test/ 2>&1 | egrep "cf-cache-status|country_code"
< cf-cache-status: MISS
{"country_code": "CA"}
$ curl -v https://www-dev.allizom.org/en-US/cdn-geo-test/ 2>&1 | egrep "cf-cache-status|country_code"
< cf-cache-status: HIT
{"country_code": "CA"}
$ curl -v https://www-dev.allizom.org/en-US/cdn-geo-test/ 2>&1 | egrep "cf-cache-status|country_code"
< cf-cache-status: HIT
{"country_code": "CA"}

$ # VPN: Gothenburg, Sweden
[...]

PR #10431, opened 2021-09-08
cdn cache geo redirects

pmac:
- Remove the CDN geo test page
- Remove the no-cache headers from views that use Geo


alexgibson (approved):
r+ on the code change, although I'm not sure how to consider the likely impact, if any. I'm assuming this will effect the cache on requests for both the WNP redirector and the VPN geo-request. Is this relatively safe to just roll out, or does it need some further testing?

pmac:
I believe it's relatively safe to roll out. We should manually test it on dev/stage though. I did some decent testing on the idea of this and [documented here](https://github.com/mozilla/bedrock/issues/10429#issuecomment-915497296), but we should do some more on these important pages when this lands as well.

PR #10435, opened 2021-09-09
Fix #10434: Add country_code to the context for all pages

### LINKED ISSUE #10434: Add country code to the context for templates
Now that we've completed #10429 we can have the country available in every request, much like we do with `request.LANG`. Perhaps we should do it exactly like that and make it `request.COUNTRY`.

pmac:
Allows for a query param to override the country (e.g. `?geo=de`).

Also removes the old Maxmind stuff since we're unlikely to use that again, and add a GeoTemplateView class to make it easy to switch template based on request country. Adds docs for all of this as well.

### DIFF
--- a/bedrock/base/context_processors.py
+++ b/bedrock/base/context_processors.py
@@ -1,15 +1,18 @@
-# from bedrock.base
-
 from django.conf import settings
 
+from bedrock.base.geo import get_country_from_request
 from lib.l10n_utils import translation
+
+
+def geo(request):
+    return {'country_code': get_country_from_request(request)}
 
 
 def i18n(request):
     return {
         'LANGUAGES': settings.LANGUAGES,
-        'LANG': (settings.LANGUAGE_URL_MAP.get(translation.get_language()) or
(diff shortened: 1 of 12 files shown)


bedrock/base/context_processors.py
```diff
@@ -2,14 +2,25 @@
...
+
+    return {'country_code': country.upper() if country else None}
+
+
 def i18n(request):
     return {
         'LANGUAGES': settings.LANGUAGES,
-        'LANG': (settings.LANGUAGE_URL_MAP.get(translation.get_language()) or
-                 translation.get_language()),
+        'LANG': (settings.LANGUAGE_URL_MAP.get(
```
pmac, line 22: This was a change that flake8 wanted 🤷‍♂️

pmac:
@alexgibson this grew a bit, but it's done I believe. Thanks for the assist!

alexgibson (approved):
I tested out all the new toys locally and everything worked just as described in the docs. Nice work! 🌏

PR #10437, opened 2021-09-09
Add sphinx-autobuild for easier docs development

pmac:
@reemhamz @maureenlholland give this a try. It adds a make command. Do this:

```shell
$ pip install -r requirements/docs.txt
$ make livedocs
```

That will build the docs, start a local server at http://localhost:8000, open a browser tab for it, and watch for changes in the docs directory. If it detects changes it will rebuild the docs and reload your browser.

PR #10438, opened 2021-09-10
More docs improvements

pmac:
Followup from #10437:

- Fix docs formating errors and warnings from Sphinx
- Change the `make docs` command to start an auto-build-reload server using Docker:
  * `make docs` does this using Docker
  * `make livedocs` does this locally
  * `make build-docs` just builds the docs using Docker

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,8 +22,9 @@ help:
 	@echo "  test          - run tests against local files"
 	@echo "  test-image    - run tests against files in docker image"
 	@echo "  test-cdn      - run CDN tests against TEST_DOMAIN"
-	@echo "  docs          - generate Sphinx HTML documentation"
+	@echo "  docs          - generate Sphinx HTML documentation with server and live reload using Docker"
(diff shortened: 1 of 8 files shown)


maureenlholland (approved):
r+ 📝 ! Docker and local working great - thanks!

If we want to be squeaky clean, there is one last warning that could be addressed, but it's not a block:

`/app/docs/install.rst:236: WARNING: Explicit markup ends without a blank line; unexpected unindent.`

pmac:
> If we want to be squeaky clean, there is one last warning that could be addressed, but it's not a block:

I know. I can't seem to find the magical incantation to make that go away. I'll poke at it a bit more. It is very annoying to leave that one warning.

maureenlholland:
@pmac I think it just wants a blank line before the Localization heading

pmac:
Yup. You're right. Sleep helped me. I was banging my head against this on Friday. All of the other warnings I was working on were in the contentful file, and I failed to notice that this one was in install.rst 🤦‍♂️ All fixed up now.

PR #10440, opened 2021-09-10
Output warning on missing image not issue 500 error

pmac:
We saw [this error in Sentry](https://sentry.prod.mozaws.net/operations/bedrock-dev/issues/13026476/?query=is%3Aunresolved) when Contentful failed to upload an image. This will cause a warning to be output to the logs, but not break the whole page when an image can't be found by the static assets system.

This is more than just a contentful issue, and we should ensure that outputting a log warning is enough, but I believe we can all agree that a broken image on a page is superior to a broken page.

### DIFF
--- a/bedrock/base/templatetags/helpers.py
+++ b/bedrock/base/templatetags/helpers.py
@@ -1,4 +1,5 @@
 import datetime
+import logging
 import urllib.parse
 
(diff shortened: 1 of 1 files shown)


alexgibson (approved):
Looks good! r+

PR #10449, opened 2021-09-13
Add support for VPN CTA for Contentful Hero components

### LINKED ISSUE #10447: Get Mozilla VPN link on Contentful home page goes to getpocket.com
## Description

The "Get Mozilla VPN" link on Contentful home page goes to `getpocket.com` instead of `/products/vpn/`.

### Steps to reproduce

1. https://www-dev.allizom.org/en-US/
2. Scroll down the the "Protection for your whole device, on every device" section.
3. Click "Get Mozilla VPN".

Screenshot: 

image

### Expected result

The page should navigate to https://www-demo1.allizom.org/en-US/products/vpn/

### Actual result

The link goes to https://getpocket.com/?utm_source=www.mozilla.org-home&utm_medium=referral&utm_campaign=home

### Environment

N/A

pmac:
Fix #10447

### DIFF
--- a/bedrock/contentful/templates/includes/contentful/cta.html
+++ b/bedrock/contentful/templates/includes/contentful/cta.html
@@ -22,7 +22,13 @@
   ) }}
 
 {% elif action == 'Get Pocket' %}
+
     {% set referral = '?utm_source=' + utm_source + '&utm_medium=referral&utm_campaign=' + utm_campaign %}
     <a href="https://getpocket.com/{{ referral|safe }}" class="mzp-c-button {{ button_class }}" rel="external noopener" data-cta-type="link" data-cta-text="{{ cta_text }}">{{ label }}</a>
+
+{% elif action == 'Get Mozilla VPN' %}
+
+  <a href="{{ url('products.vpn.landing') }}" class="mzp-c-button {{ button_class }}" data-cta-text="{{ cta_text }}" data-cta-type="link" data-cta-position="{{ location }}">{{ label }}</a>
+
 {% endif %}
 
--- a/bedrock/contentful/api.py
+++ b/bedrock/contentful/api.py
@@ -173,15 +173,17 @@ def _make_wordmark(entry):
 
 def _make_cta_button(entry):
     fields = entry.fields()
+    action = fields.get('action')
 
     button_class = [
-        'mzp-t-product',  # TODO, only add on Firefox themed pages
+        # TODO, only add on Firefox themed pages
+        'mzp-t-product' if action != 'Get Mozilla VPN' else '',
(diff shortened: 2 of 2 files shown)


pmac:
This should, after we add the "Get Mozilla VPN" action as a choice in the admin, fix this issue. We should probably think harder about how we'd like to enable this kind of thing more robustly in future though.

bedrock/contentful/templates/includes/contentful/cta.html
```diff
@@ -22,7 +22,13 @@
   ) }}
 
 {% elif action == 'Get Pocket' %}
+
     {% set referral = '?utm_source=' + utm_source + '&utm_medium=referral&utm_campaign=' + utm_campaign %}
     <a href="https://getpocket.com/{{ referral|safe }}" class="mzp-c-button {{ button_class }}" rel="external noopener" data-cta-type="link" data-cta-text="{{ cta_text }}">{{ label }}</a>
+
+{% elif action == 'Get Mozilla VPN' %}
+
+  <a href="{{ url('products.vpn.landing') }}" class="mzp-c-button {{ button_class }}">{{ label }}</a>
```
alexgibson, line 31: Suggest adding some data attributes so we can see the numbner of clicks this gets in GA?
```suggestion
  <a href="{{ url('products.vpn.landing') }}" class="mzp-c-button {{ button_class }}" data-cta-text="{{ cta_text }}" data-cta-type="button" data-cta-position="{{ location }}">{{ label }}</a>
```

alexgibson (changes requested):
Looks good - one suggested change to make sure these buttons get events in GA.

bedrock/contentful/templates/includes/contentful/cta.html
pmac: Thanks! I literally just copied the code from the current VPN CTA on the home page.

pmac: I do see above this one that the `data-cta-type` for Pocket is `link` and not `button`. They're both buttons though. They should be the same type right?

alexgibson: Good question. We seem to use these pretty interchangably and there's no real definition for `link` as a value in our docs: https://bedrock.readthedocs.io/en/latest/analytics.html#gtm-listeners-data-attributes

To err on the side of consistency I'd use `link` as well here. Although throughout the site as a whole I see a good mixture of both.

pmac: Cool. Done.

alexgibson (approved):

PR #10454, opened 2021-09-14
Implement attribution flow for VPN subscription referrals (Fixes #10409)

### LINKED ISSUE #10409: Implement attribution flow for VPN subscription referrals
## Description

Now that we're starting to promote Mozilla VPN more widely on mozorg pages, we need to implement a system that will enable us to track attribution more easily as people click through to the product page. Up until now we've used a `?source=` query param on the page URL (we can't use utm params for same-site links). This was OK enough for in-product pages such as /whatsnew, but if we want to do this on more public surfaces (example see: https://github.com/mozilla/bedrock/issues/10403), then we should probably make a better solution since these URLs can easily be crawled and shared.

The solution is up for discussion, but it could be something like setting a cookie on CTA click, which we then read when someone lands on the product page before updating the subscription link parameters.

---

#### Success Criteria
- [x] When people click on same-site CTA links to the VPN landing page, we can attribute those conversions at the subscription level.

alexgibson:
## Description
- Adds a new VPN product referral link helper.
- Implements helper in navigation CTA button for VPN.

## Issue / Bugzilla link
#10409

## Testing
To test use a Firefox browser with DNT disabled:

1.) Open http://localhost:8000/en-US/
2.) Click on the "Get Mozilla VPN" CTA button in the main navigation.
3.) Once on http://localhost:8000/en-US/products/vpn/#pricing, verify that each subscription link contains:
  - `utm_source=www.mozilla.org`
  - `utm_medium=referral`
  - `utm_campaign=navigation`

Demo link: https://www-demo1.allizom.org/en-US/

### DIFF
--- a/bedrock/products/templatetags/misc.py
+++ b/bedrock/products/templatetags/misc.py
@@ -7,6 +7,7 @@ import jinja2
 from django.conf import settings
 from django_jinja import library
 
+from bedrock.base.urlresolvers import reverse
 from lib.l10n_utils.fluent import ftl
 
 FTL_FILES = ['products/vpn/shared']
@@ -223,5 +224,35 @@ def vpn_saving(ctx, plan='12-month', lang=None):
     markup = (f'<span class="js-vpn-saving-display" {attrs}>'
               f'{default_text}'
               f'</span>')
+
+    return jinja2.Markup(markup)
+
+
+@library.global_function
+@jinja2.contextfunction
+def vpn_product_referral_link(ctx, referral_id='', page_anchor='', link_text=None, class_name=None, optional_attributes=None):
+    """
+    Render link to the /products/vpn/ landing page with referral attribution markup
+
+    Examples
+    ========
+
+    In Template
+    -----------
+
+        {{ vpn_product_referral_link(referral_id='navigation', link_text='Get Mozilla VPN') }}
+    """
+
+    href = reverse('products.vpn.landing')
+    css_class = 'mzp-c-button mzp-t-product js-fxa-product-referral-link'
+    attrs = f'data-referral-id="{referral_id}" '
+
+    if optional_attributes:
+        attrs += ' '.join('%s="%s"' % (attr, val) for attr, val in optional_attributes.items())
+
+    if class_name:
+        css_class += f' {class_name}'
+
+    markup = (f'<a href="{href}{page_anchor}" class="{css_class}" {attrs}>{link_text}</a>')
 
     return jinja2.Markup(markup)
--- a/bedrock/products/tests/test_helper_misc.py
+++ b/bedrock/products/tests/test_helper_misc.py
@@ -601,3 +601,27 @@ class TestVPNSaving(TestCase):
             u'<span class="js-vpn-saving-display" data-price-usd="Save 50%" '
             u'data-price-euro="Save 50%" data-price-chf="Save 45%">Save 50%</span>')
         self.assertEqual(markup, expected)
+
+
+class TestVPNProductReferralLink(TestCase):
+    rf = RequestFactory()
+
+    def _render(self, referral_id, page_anchor, link_text, class_name, optional_attributes):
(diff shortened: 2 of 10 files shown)


alexgibson:
@maureenlholland r? I think this is ready for code review.

I've added a do-not-merge label just until we can get a thumbs up from Dan on the functionality.

maureenlholland (commented):
This is all working as expected for me. Once test is fixed and Dan confirms behaviour, r+ official

bedrock/products/tests/test_helper_misc.py
```diff
@@ -601,3 +601,27 @@ def test_vpn_12_month_saving(self):
...
+                      {'request': req})
+
+    def test_vpn_product_referral_link(self):
+        """Should return expected markup"""
+        markup = self._render(referral_id='navigation', page_anchor='#pricing',
+                              link_text='Get Mozilla VPN', class_name='mzp-t-secondary mzp-t-md',
+                              optional_attributes={'data-cta-text': 'Get Mozilla VPN',
+                                                   'data-cta-type': 'button'})
+        expected = (
+            u'<a href="/products/vpn/#pricing" class="mzp-c-button mzp-t-product '
```
alexgibson, line 624: @pmac not urgent - any idea why this test passes locally, but not in CI? 

The problem seems to be because in the helper I'm using `reverse('products.vpn.landing')` to generate the link URL, and whilst this returns  `/products/vpn/` for me locally, when CircleCI runs it returns with the locale `/en-US/products/vpn/`. How can I make this consistent?

media/css/protocol/components/_navigation.scss
```diff
@@ -249,23 +249,11 @@ html.is-firefox .c-navigation {
...
     }
 
-    .c-navigation-fxa-cta-container {
+    .c-navigation-vpn-cta-container {
         display: block;
     }
 }
 
-// ... but not on mobile platforms
-html.is-firefox.ios .c-navigation,
```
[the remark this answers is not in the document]
alexgibson: Yep! We didn't want to promote the FxA button on mobileas it had some desktop-specific paremeters in the link. But now that we're promoting VPN this should be fine to show to mobile-Firefox users.

bedrock/products/tests/test_helper_misc.py
pmac: It's probably because the test environment has an active locale at that point. You should be able to ensure this is the case using the following for the _render method:

```python
    def _render(self, referral_id, page_anchor, link_text, class_name, optional_attributes):
        with self.activate('en-US'):
            req = self.rf.get('/')
            req.locale = 'en-US'
            return render("{{{{ vpn_product_referral_link('{0}', '{1}', '{2}', '{3}', {4}) }}}}".format(
                      referral_id, page_anchor, link_text, class_name, optional_attributes),
                      {'request': req})
```

pmac: if I'm right then it should always have the `/en-US` prefix.

alexgibson: This worked, thank you!

bedrock/products/templatetags/misc.py
```diff
@@ -225,3 +226,33 @@ def vpn_saving(ctx, plan='12-month', lang=None):
...
+    Examples
+    ========
+
+    In Template
+    -----------
+
+        {{ vpn_product_referral_link(referral_id='navigation', link_text='Get Mozilla VPN') }}
+    """
+
+    href = reverse('products.vpn.landing')
```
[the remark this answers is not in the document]
alexgibson: This will generate the right URL for the landing page, including the locale

alexgibson:
Ok, updated and tests are passing. 

@maureenlholland looks like you need to approve the PR still for me to merge? Thanks

maureenlholland (approved):

PR #10471, opened 2021-09-16
Fix Sentry JS

alexgibson:
## Description
The Sentry JS we use needs to be updated now we've switched to Webpack.

I pushed a branch to demo with an intentional JS error to verify this method works: https://sentry.prod.mozaws.net/operations/bedrock-demos/issues/13155171/?environment=production&referrer=alert_email

## Issue / Bugzilla link
N/A

## Testing
https://www-demo2.allizom.org/en-US/ should trigger a JS error that you can see above in Sentry


pmac (approved):
💯

PR #10491, opened 2021-09-22
Contentful: Card components font size is too big (Fixes #10444)

### LINKED ISSUE #10444: Contentful: Card components font size is too big
## Description

Contentful cards (small/medium) font size seems to be too big. We should investigate and make sure they are correct, or are using the correct version of protocol.

### Steps to reproduce

Compare this contentful homepage:

https://www-dev.allizom.org/en-US/contentful-preview/69xQvSpz9F6fmsyaSxLrjW/

To the Mozilla homepage:

https://www.mozilla.org/en-US/

The font size on the row of 4 cards is different.

### Expected result

We should make sure that the default font size for cards is correct and consistent and is up-to-date.

alexgibson:
## Description
Assume that 4-card layout should always use the `mzp-c-card-extra-small` component class.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/10444

## Testing
http://localhost:8000/en-US/

### DIFF
--- a/bedrock/contentful/templates/includes/contentful/all.html
+++ b/bedrock/contentful/templates/includes/contentful/all.html
@@ -67,6 +67,8 @@
           {% for card_data in entry.cards -%}
             {% if card_data.component == 'large_card' %}
               {% set size_class = 'mzp-c-card-large' %}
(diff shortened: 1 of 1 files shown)


maureenlholland (approved):
r+ 🗃️

PR #10518, opened 2021-09-28
Ignore Sentry JS errors without a source URL (Fixes #10515)

### LINKED ISSUE #10515: Ignore Sentry JS errors without a source URL
## Description

We see quite a lot of JS errors in Sentry that don't have a source URL and instead register as `<anonymous>`. We don't have any inline scripts in bedrock (everything is served from `/media/js/`), so I think it's likely safe to assume these errors are from extensions or browsers injecting inline scripts.

We should be able to ignore these errors using:

```javascript
Sentry.init({
  beforeSend(event) {
    try {
      if (event.exception.values[0].stacktrace.frames[0].filename === `<anonymous>`) {
        return null;
      }
    } catch (e) {}

    return event;
  },
});
```

---

#### Success Criteria
- [ ] JS errors without a valid source URL are ignored.

alexgibson:
## Description
This PR attempts to reduce some of the noise we see in Snetry by filtering out JS errors with `<anonymous>` for the filename/URL (example: https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/13160153/).

## Issue / Bugzilla link
#10515

## Testing
This isn't so easy to test locally, so a general code-level review should be enough I think. I pushed this branch to [demo1](https://www-demo1.allizom.org/en-US/) and regular error logging seems fine. We'll only be able to see the real-world impact of this change on logging once it gets to production.


media/js/base/sentry.es6.js
```diff
@@ -0,0 +1,42 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+'use strict';
+
+const Sentry = require('@sentry/browser');
```
nathan-barrett, line 7: Just a general question here, since this is using es6 - is there a big difference between using the es6 import syntax? `import Sentry from '@sentry/browser` vs the require?

media/js/base/sentry.es6.js
```diff
@@ -0,0 +1,42 @@
...
+            ],
+            beforeSend(event) {
+                try {
+                    // https://github.com/getsentry/sentry-javascript/issues/3147
+                    if (event.exception.values[0].stacktrace.frames[0].filename === `<anonymous>`) {
+                        return null;
+                    }
+                } catch (e) {}
+
+                return event;
```
nathan-barrett, line 38: should this be wrapped in a `finally` block? Or is that not the case since there is a return within the try block? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/try...catch

nathan-barrett (approved):
r+wc

media/js/base/sentry.es6.js
alexgibson: I mostly kept this using `require()` for consistency, since we support commonJS in all scripts, and not just `es6.js` scripts. Happy to switch it if you think it makes more sense though, thoughts?

media/js/base/sentry.es6.js
alexgibson: Yeah I don't think we need a `finally` here as we already return something specific in the `try`, as you say.

media/js/base/sentry.es6.js
alexgibson: That's possible using `require()` as well I think. But we can just go ahead and use the newer syntax anyway I think. I'll make it so 👍

nathan-barrett:
Seems good to merge to me, feel free to go ahead once you make any quality of life changes

Issue #10490, opened 2021-09-21
Setup Brand Indicators for Message Indentification (BIMI) for Mozilla Emails


krattliff11:
## Description

Google has announced general support of Brand Indicators for Message Identification (BIMI) in Gmail
BIMI provides a secure, global framework in which inboxes display sender logos for authenticated messages.

Why is this important?
BIMI along with DMARC, strengthens the trustworthiness of our emails. This is critical because recipients are more likely to open and engage with emails coming from a brand they know and trust.

We are looking for help with 
1) publishing a BIMI record for our domain in the Domain Name System (DNS) which points to the 'M' logo in an SVG Portable/Secure format to be used
2) Obtain a Verified Mark Certificate (VMC). We will need someone to go through the vetting process with a Certified Authority (CA) which proves our right to use the Mozilla logo. It is required that the logo be a registered trademark, which will be validated as part of the vetting process. Today, Entrust and DigiCert support BIMI as CAs and the list is expected to expand in the future. Some mailbox providers support BIMI without a VMC, so if you choose to pass on this final step, there is still value in publishing a “self-asserted” BIMI record.

Details:
https://www.validity.com/blog/google-announces-general-availability-of-bimi-in-gmail/

---

#### Success Criteria
- there is a SVG logo hosted with a DNS record
- Emails send from mozilla@email.mozilla.org and firefox@email.mozilla.org display the 'M' Mozilla logo

pmac:
More info on the image spec here https://bimigroup.org/implementation-guide/

krattliff11:
I cannot verify the svg logo. I'm using this tool: https://bimigroup.org/bimi-generator/

pmac:
I guess we'll have to get it setup and then maybe it'll tell us. Unfortunately we can't do the rest of the setup with the DNS, that's an IT thing, but I can help file that ticket if you'd like. In the mean time we can merge this so that we can move on to the DNS bits.

krattliff11:
I'd like your help filing the ticket with IT for the 2nd half of this work.

Thanks

pmac:
@duallain Do you know the proper place to file a request for DNS changes like this? The SVG image has been deployed and is here: 

https://www.mozilla.org/media/img/logos/mozilla/logo-bimi.svg

duallain:
> @duallain Do you know the proper place to file a request for DNS changes like this? The SVG image has been deployed and is here:
> 
[...]

PR #10519, opened 2021-09-28
Add BIMI-compatible Mozilla logo

pmac:
Re: #10490


pmac:
@krattliff11 is there a way to test and see if this SVG will work?

craigcook (approved):
🚴

PR #10526, opened 2021-09-29
Check for double-encoded RTAMO attribution data (Fixes #10524)

### LINKED ISSUE #10524: RTAMO limitation not always enforced
## Description

In https://github.com/mozilla/bedrock/issues/10337, a check has been added to limit RTAMO attribution to AMO referrals. This limitation is only enforced when the UTM `content` parameter [starts with `rta:`](https://github.com/mozilla/bedrock/blob/ab01bd216b451116cfa4eb59fbb172131dd809a7/media/js/base/stub-attribution.js#L340) but that might be skipped when the value is encoded more than once.

### Steps to reproduce

Download and install FF on Windows using the URL below:

```
https://www.mozilla.org/fr/firefox/new/?utm_campaign=amo-fx-cta-607454&utm_content=rta%25253AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ&utm_medium=referral&utm_source=addons.mozilla.org
```

### Actual result

FF will show the "RTAMO page" after install.

### Expected result

The RTAMO flow shouldn't be triggered.

Bedrock should probably decode the value of `content` until there is no percent sign anymore or the decoding step does not change the result, which is [what Firefox does](https://searchfox.org/mozilla-central/source/browser/components/newtab/aboutwelcome/lib/AboutWelcomeDefaults.jsm#238-249). After that, verifying that the value starts with `rta:` should always return the expected outcome.

### Other info

The URL below (which is what AMO usually generates) should not trigger the RTAMO flow (this is already working but could be useful to compare with the URL in the STR above):

```
https://www.mozilla.org/fr/firefox/new/?utm_campaign=amo-fx-cta-607454&utm_content=rta:dUJsb2NrMEByYXltb25kaGlsbC5uZXQ&utm_medium=referral&utm_source=addons.mozilla.org
```

alexgibson:
## Description
Handles double-encoded attribution data when validating RTAMO referrals.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/10524

### DIFF
--- a/media/js/base/stub-attribution.js
+++ b/media/js/base/stub-attribution.js
@@ -336,8 +336,29 @@ if (typeof window.Mozilla === 'undefined') {
 
     StubAttribution.hasValidData = function(data) {
         if (typeof data.utm_content === 'string' && typeof data.referrer === 'string') {
-            // If RTAMO data does not originate from AMO, drop attribution (Issue 10337).
-            if ((/^rta:/).test(decodeURIComponent(data.utm_content)) && data.referrer.indexOf('https://addons.mozilla.org') === -1) {
+            var content = data.utm_content;
+            var charLimit = 150;
+
+            // If utm_content is unusually long, return false early.
+            if (content.length > charLimit) {
+                return false;
+            }
+
+            // Attribution data can be double encoded
+            while (content.includes('%')) {
+                try {
+                    var result = decodeURIComponent(content);
+                    if (result === content) {
+                        break;
+                    }
+                    content = result;
+                } catch (e) {
+                    break;
+                }
+            }
+
+            // If RTAMO data does not originate from AMO, drop attribution (Issues 10337, 10524).
+            if ((/^rta:/).test(content) && data.referrer.indexOf('https://addons.mozilla.org') === -1) {
                 return false;
             }
         }
--- a/tests/unit/spec/base/stub-attribution.js
+++ b/tests/unit/spec/base/stub-attribution.js
@@ -181,6 +181,48 @@ describe('stub-attribution.js', function() {
             /* eslint-enable camelcase */
 
             expect(Mozilla.StubAttribution.hasValidData(data)).toBeTruthy();
+
+            /* eslint-disable camelcase */
+            const data2 = {
+                utm_source: 'addons.mozilla.org',
+                utm_medium: 'referral',
+                utm_campaign: 'amo-fx-cta-607454',
+                utm_content: 'rta:dUJsb2NrMEByYXltb25kaGlsbC5uZXQ',
+                referrer: 'https://addons.mozilla.org/',
+                ua: 'chrome',
+                visit_id: GA_VISIT_ID
+            };
+            /* eslint-enable camelcase */
+
+            expect(Mozilla.StubAttribution.hasValidData(data2)).toBeTruthy();
+
+            /* eslint-disable camelcase */
+            const data3 = {
+                utm_source: 'addons.mozilla.org',
+                utm_medium: 'referral',
+                utm_campaign: 'non-fx-button',
+                utm_content: 'rta%25253AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ',
+                referrer: 'https://addons.mozilla.org/',
+                ua: 'chrome',
+                visit_id: GA_VISIT_ID
+            };
+            /* eslint-enable camelcase */
+
+            expect(Mozilla.StubAttribution.hasValidData(data3)).toBeTruthy();
+
+            /* eslint-disable camelcase */
+            const data4 = {
+                utm_source: 'addons.mozilla.org',
+                utm_medium: 'referral',
+                utm_campaign: 'non-fx-button',
+                utm_content: '%72%74%61%3AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ',
+                referrer: 'https://addons.mozilla.org/',
+                ua: 'chrome',
+                visit_id: GA_VISIT_ID
+            };
+            /* eslint-enable camelcase */
+
+            expect(Mozilla.StubAttribution.hasValidData(data4)).toBeTruthy();
         });
 
         it('should return false for RTAMO data that does not have AMO as the referrer', function() {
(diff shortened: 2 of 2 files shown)


media/js/base/stub-attribution.js
```diff
@@ -336,8 +336,23 @@ if (typeof window.Mozilla === 'undefined') {
 
     StubAttribution.hasValidData = function(data) {
         if (typeof data.utm_content === 'string' && typeof data.referrer === 'string') {
+            var content = data.utm_content;
```
willdurand, line 339: I am wondering if we should restrict the length of the `utm_content` value given that we introduced a loop below given that we can add a lot of `25` after `rta%` in a string like `rta%25253AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ`

alexgibson: @willdurand what would you consider a sensible length for RTAMO referrals?

alexgibson: Actually, we don't really need to be concerned with preserving the actual value here, since we're only validating. We can just check the first 10 or so characters

willdurand: > @willdurand what would you consider a sensible length for valid RTAMO referrals?

I'd say `120` for the length of the `content` value.

tests/unit/spec/base/stub-attribution.js
```diff
@@ -214,6 +214,38 @@ describe('stub-attribution.js', function() {
...
         });
+
+        it('should validate double-encoded attribution data correctly', function() {
+            /* eslint-disable camelcase */
+            const data = {
+                utm_source: 'addons.mozilla.org',
+                utm_medium: 'referral',
+                utm_campaign: 'amo-fx-cta-607454',
+                utm_content: 'rta%25253AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ',
+                referrer: '',
```
willdurand, line 225: Shouldn't this parameter be set to `'https://addons.mozilla.org/'` and the matcher used below be `toBeTruthy()` given that the `utm_content` value looks valid?

alexgibson: Ok updated

willdurand: Ha, I see now. It shouldn't be possible because AMO wouldn't do that but if that happens for some reasons, the `referer` value should still be checked to avoid the by-pass.

willdurand: Sorry, I just saw the update. I think we need both test cases.

alexgibson: I think both makes sense 👍

willdurand: To be clear, the following test case (copied from a previous iteration) is useful for the bypass (sorry, I got confused when I read it the first time):

```js
        it('should validate double-encoded attribution data correctly', function() {
            /* eslint-disable camelcase */
            const data = {
                utm_source: 'addons.mozilla.org',
                utm_medium: 'referral',
                utm_campaign: 'amo-fx-cta-607454',
                utm_content: 'rta%25253AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ',
                referrer: '',
                ua: 'chrome',
                visit_id: GA_VISIT_ID
            };
            /* eslint-enable camelcase */

            expect(Mozilla.StubAttribution.hasValidData(data)).toBeFalsy();
        });
```

In addition, I think we'd need a test case like this one. We should early return when `utm_content` is longer than `120` characters I believe, instead of trying to decode up to 120 chars.

```diff
diff --git a/tests/unit/spec/base/stub-attribution.js b/tests/unit/spec/base/stub-attribution.js
index 156f5a13d..029eb1aa7 100644
--- a/tests/unit/spec/base/stub-attribution.js
+++ b/tests/unit/spec/base/stub-attribution.js
@@ -246,6 +246,22 @@ describe('stub-attribution.js', function() {
 
             expect(Mozilla.StubAttribution.hasValidData(data)).toBeTruthy();
         });
+
+        it('should limit the length of the UTM content', function() {
+            /* eslint-disable camelcase */
+            const data = {
+                utm_source: 'addons.mozilla.org',
+                utm_medium: 'referral',
+                utm_campaign: 'amo-fx-cta-607454',
+                utm_content: 'rta%' + '25'.repeat(58) + '3AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ',
+                referrer: '',
+                ua: 'chrome',
+                visit_id: GA_VISIT_ID
+            };
+            /* eslint-enable camelcase */
+
+            expect(Mozilla.StubAttribution.hasValidData(data)).toBeFalsy();
+        });
     });
 
     describe('isFirefoxNewScene2', function() {
```

alexgibson: Ok, maybe we can return early if `utm_content` starts with `rta` and length > 120 chars. We'd still be iterating over any length of string less than that value though, which is the same as we're doing now.

Thoughts?

willdurand: > @willdurand On the question of returning early if `utm_content` > 120 chars. This would imply we drop all stub attribution data (not just RTAMO attribution) if `utm_content` > 120 chars. I'm not quite clear on the impact of doing that, which makes me hesitant to accept the suggestion.

I see. I think the only option here would be to use a more conservative value then (`200` maybe?).

> Ok, maybe we can return early if `utm_content` starts with `rta` and length > 120 chars. We'd still be iterating over any length of string less than that value though, which is the same as we're doing now.

Could we start by checking the `referer` first and, if it is AMO, then we restrict the length of `utm_content`?

alexgibson: > I see. I think the only option here would be to use a more conservative value then (200 maybe?).

I'm going to see if I can find out the maximum length of value we support for attribution, perhaps that might make reasoning about this easier.

> Could we start by checking the referer first and, if it is AMO, then we restrict the length of utm_content?

It seems like the usecase we're more concerned about is if the referrer is not AMO, which represents the large majority of attribution data.

tests/unit/spec/base/stub-attribution.js
```diff
@@ -214,6 +214,84 @@ describe('stub-attribution.js', function() {
...
+        });
+
+        it('should limit the length of the UTM content for RTAMO data', function() {
+            /* eslint-disable camelcase */
+            const data = {
+                utm_source: 'addons.mozilla.org',
+                utm_medium: 'referral',
+                utm_campaign: 'amo-fx-cta-607454',
+                utm_content: `rta%${'25'.repeat(58)}3AdUJsb2NrMEByYXltb25kaGlsbC5uZXQ`,
+                referrer: 'https://addons.mozilla.org/',
```
willdurand, line 271: This could be set to an empty string, right?

alexgibson: Why? Would the test then be passing as falsy because of the length, or because of the referrer? Harder to tell imho, unless I'm not understanding why you suggest this?

willdurand: Without the length check, the function would return `true` with _and_ without the `referer` set, which is a problem. Because the added length check is before, it does not matter right now but if there is some refactoring in the future, we might miss this edge case.

alexgibson: Changing this value won't make reasonsing about future changes any clearer though, unless I'm not following your trail of thought here? 

If we don't have the length check it will pass validation yes, but we have to set a sensible limit on how many times we're going to try decoding these values without it. I'm not sure why a future refactor would try and move the length check later on in the code, seeing as it would be redundant if not used as an early return. Are we perhaps leaning a little too far into the edgecases here? :)

alexgibson: I guess I could add another referrer check along with the length check, so we only return early for non-AMO referrals? Thoughts?

willdurand: Maybe you could add both test cases again, with and without the `referer` set?

alexgibson:
@wagnerand I think we're at risk of getting a little too far into the weeds here with the decoding complexity. I'm wondering if we can simplify much of this by taking a simpler approach. Curious to what you think of the following pseudo logic:

```
if `utm_content` starts with `rta:` or `rta%` and `referrer` does not equal `addons.mozilla.org`:
    return false
else
    return true
```

This would keep things much simpler imho, and we don't need to be concerned with decoding, or double-decoding, or while loops/value lengths.

willdurand:
That works for me.

Can we still enforce a length limit?

alexgibson:
> Can we still enforce a length limit?

There's already a limit on the total length of all attribution data I believe, just not individual values. What would setting a limit here achieve for AMO?

willdurand:
> There's already a limit on the total length of all attribution data I believe, just not individual values. What would setting a limit here achieve for AMO?

I edited my reply in the meantime. Let's forget about length ;-)

alexgibson:
Ok, refactored and simplified this a bunch 👍

media/js/base/stub-attribution.js
```diff
@@ -336,8 +336,9 @@ if (typeof window.Mozilla === 'undefined') {
 
     StubAttribution.hasValidData = function(data) {
         if (typeof data.utm_content === 'string' && typeof data.referrer === 'string') {
-            // If RTAMO data does not originate from AMO, drop attribution (Issue 10337).
-            if ((/^rta:/).test(decodeURIComponent(data.utm_content)) && data.referrer.indexOf('https://addons.mozilla.org') === -1) {
+            // If RTAMO data does not originate from AMO, drop attribution (Issues 10337, 10524).
+            if (((/^rta:/).test(data.utm_content) || (/^rta%/).test(data.utm_content)) &&
```
willdurand, line 340: I am sorry to come back to that again but, having talked to @diox about it, we think we must decode the `utm_content` value in a loop (like FF does) and probably limit the length of the value to avoid decoding that value for a very long time (although that is more a perf optimization than an actual issue).

The reason is: `rta:` can also be encoded like this `%72%74%61%3A` and FF will decode it fully and accept it (because the decoded value will start with `rta:`).

Attribution "code" in FF is [limited to 1010](https://searchfox.org/mozilla-central/source/browser/components/attribution/AttributionCode.jsm#143) and AFAIK that includes all the parameters/values. For RTAMO specifically, the length of the `content` value should not be more than `120` characters. It could be more if we don't want to possibly break other stuff (non-AMO stuff). I suppose a limit of 200 chars would already be good because that would already avoid decoding the value for too long but, then again, that is up to you.

alexgibson: How about we simply drop the fully encoded rta: value like in your example here?

I’m not sure I agree it’s worth the effort of doing all the decoding logic tbh.

alexgibson: Ok, I took a look in GA and the longest `utm_content` value I can see (in the last 30 days) was ~100 chars and was from AMO. With that in mind, I don't think I'm quite as concerned about potentially dropping data.

Let me see if i can rework this again to accomodate decoding.

alexgibson: > Fully encoded rta: was just an example, there could be any number of chars encoded, and each could be encoded more than once.

I understand that, my question was more around if the juice is really worth the squeeze here. I did try and assess our current GA data for `utm_content` in the comment above. With that in mind, I think I'm ok with potentially dropping longer values.

alexgibson:
@willdurand Ok, reverted back to the original approach. Thanks for your feedback.

media/js/base/stub-attribution.js
diox: FWIW the maximum length AMO should send is 111 (Add-on ID is 80 chars max, everything ascii, base64 encoded without the padding that should be 107, add 4 chars for `rta:`). Will suggested 120 above to be safe.

alexgibson: I've settled on 150 just to give some extra wiggle room 👍

media/js/base/stub-attribution.js
```diff
@@ -336,8 +336,31 @@ if (typeof window.Mozilla === 'undefined') {
...
-            if ((/^rta:/).test(decodeURIComponent(data.utm_content)) && data.referrer.indexOf('https://addons.mozilla.org') === -1) {
+            var charLimit = 150;
+
+            // If utm_content is unusually long, return false early.
+            if (data.utm_content.length > charLimit) {
+                return false;
+            }
+
+            // Limit the number of chars to iterate through.
+            var content = data.utm_content.substr(0, charLimit);
```
willdurand, line 347: Is this still needed given that there is the early return above?

alexgibson:
Updated

willdurand (approved):
Looks good to me, thanks!

pmac:
We also limit the total attribution data in the server-side view that signs the data to 600 characters after encoding, and the stub attribution service also has this limit, so we'll never store enough data to be much of an issue IMO.

pmac (approved):
LGTM 🍰

PR #10528, opened 2021-09-29
Update ESLint rules for bedrock (Fixes #10527)

### LINKED ISSUE #10527: Update ESLint rules for bedrock
## Description

Now that we're on Webpack and can support new methods for authoring JS, it's a good time to update our ESLint config rules to reflect some newer standards.

---

#### Success Criteria
- [x] ESLint rules are updated and linting checks passing.

alexgibson:
## Description
Updates bedrock ESLint rules for both ES5 and ES2017+ JS files.

Feel free to make other suggestions? https://eslint.org/docs/rules/

## Issue / Bugzilla link
#10527

## Testing
- `npm run lint`

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -6,6 +6,69 @@ module.exports = {
     extends: [
         'plugin:json/recommended'
     ],
+    rules: {
+        // Require strict mode directive in top level functions
+        // https://eslint.org/docs/rules/strict
+        'strict': ['error', 'function'],
+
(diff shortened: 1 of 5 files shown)


alexgibson:
> The only other rule I think I've found a useful guardrail is no-unreachable https://eslint.org/docs/rules/no-unreachable

@maureenlholland good call! I think this rule is in the recommended (default) rules already looking at the link? I think we should be good here! 👍

Issue #10500, opened 2021-09-23
Remove deprecated Protocol components


alexgibson:
## Description

Dump from running `npm start`:

[1] WARNING: Hero component is now deprecated. Please migrate to use Split: https://protocol.mozilla.org/patterns/organisms/split.html
[1]          on line 7 of node_modules/@mozilla-protocol/core/protocol/css/components/_hero.scss
[1]          from line 11 of media/css/mozorg/about.scss

[1] WARNING: Hero component is now deprecated. Please migrate to use Split: https://protocol.mozilla.org/patterns/organisms/split.html
[1]          on line 7 of node_modules/@mozilla-protocol/core/protocol/css/components/_hero.scss
[1]          from line 10 of media/css/mozorg/lean-data.scss

[1] WARNING: Hero component is now deprecated. Please migrate to use Split: https://protocol.mozilla.org/patterns/organisms/split.html
[1]          on line 7 of node_modules/@mozilla-protocol/core/protocol/css/components/_hero.scss
[1]          from line 9 of media/css/firefox/installer-help.scss

[1] WARNING: Hero component is now deprecated. Please migrate to use Split: https://protocol.mozilla.org/patterns/organisms/split.html
[1]          on line 7 of node_modules/@mozilla-protocol/core/protocol/css/components/_hero.scss
[1]          from line 10 of media/css/firefox/testflight.scss

[1] WARNING: Hero component is now deprecated. Please migrate to use Split: https://protocol.mozilla.org/patterns/organisms/split.html
[1]          on line 7 of node_modules/@mozilla-protocol/core/protocol/css/components/_hero.scss
[1]          from line 9 of media/css/firefox/features/picture-in-picture.scss

[1] WARNING: Hero component is now deprecated. Please migrate to use Split: https://protocol.mozilla.org/patterns/organisms/split.html
[1]          on line 7 of node_modules/@mozilla-protocol/core/protocol/css/components/_hero.scss
[1]          from line 13 of media/css/newsletter/newsletter.scss

[1] WARNING: Hero component is now deprecated. Please migrate to use Split: https://protocol.mozilla.org/patterns/organisms/split.html
[1]          on line 7 of node_modules/@mozilla-protocol/core/protocol/css/components/_hero.scss
[1]          from line 10 of media/css/mozorg/home/home.scss

[1] WARNING: Picto Card component is now deprecated. Please migrate to use Picto: https://protocol.mozilla.org/patterns/molecules/picto.html
[...]

PR #10541, opened 2021-10-01
Remove deprecated picto cards [#10500]

maureenlholland:
## Description

We are still deciding the best component to use to replace centered hero components, but we have a picto_card replacement in [picto](https://protocol.mozilla.org/patterns/molecules/picto.html) so we can clean up those remaining components. There should be no more picto_card references in bedrock and we can remove the picto_card macro.

## Issue / Bugzilla link
https://github.com/mozilla/bedrock/issues/10500

## Testing
TBA

### DIFF
--- a/bedrock/base/templates/macros-protocol.html
+++ b/bedrock/base/templates/macros-protocol.html
@@ -291,7 +291,7 @@
 
 
 {# Picto: https://protocol.mozilla.org/patterns/molecules/picto.html #}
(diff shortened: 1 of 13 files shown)


bedrock/base/templates/macros-protocol.html
```diff
@@ -291,7 +291,7 @@
 
 
 {# Picto: https://protocol.mozilla.org/patterns/molecules/picto.html #}
-{% macro picto(title=None, class=None, heading_level=3, base_el='div', body=False, image_url=None, image_width=64, include_highres_image=False, l10n_image=False, lazy_loading=False) -%}
```
maureenlholland, line 294: there's a "startsWith" check on the image_url that will error if default None is used, an empty string default allows this check

PR #10542, opened 2021-10-01
Remove unused img assets (Fixes #9170)

### LINKED ISSUE #9170: Prune unused assets from bedrock
## Description

Images are one of the largest asset types and contribute significantly to the overall footprint of bedrock. Looking through the `/media/img/` directory, I think there is likely scope for a good cleanup of unused images that have been orphaned over the years. Pruning unused images will help reduce the number of bytes we push / pull around the internet, and may also help with memory consumption when running bedrock locally and watching files.

It may also be worth looking in `/media/fonts/` as these are also heavy weight files.

---

#### Success Criteria
- [ ] Unused images are pruned from the repository.

alexgibson:
## Description
Cleans up hopefully a lot of leftover image cruft.

## Issue / Bugzilla link
#9170

## Testing
- [x] None of the images below should be referenced from either jinja templates or scss files.

Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/381885884


maureenlholland (approved):
r+ 🧼

PR #10543, opened 2021-10-01
String fixes for VPN and Focus pages

craigcook:
## Description
Some fixes for recent updates. These are new strings that haven't landed in Pontoon yet, so we don't need to update IDs.

## Issue / Bugzilla link
https://github.com/mozilla-l10n/www-l10n/pull/234#issuecomment-932331831


nathan-barrett (approved):
😎

PR #10545, opened 2021-10-04
Use indexOf over includes for better older browser compat (#10524)

alexgibson:
Fixes https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/13516420/?environment=production&referrer=alert_email


reemhamz (approved):
Looks good to me, internet explorer would appreciate this. r+!

PR #10573, opened 2021-10-11
10570: Add isort support to project

### LINKED ISSUE #10570: Add isort support to Bedrock
## Description

As part of improving our auto-formatting and linting checks, it makes sense to add `isort` support to complement `black` and `prettier`

This work will involve:
* adding isort as a dev dependency
* adding appropriate/sensible config
* applying it to the entire codebase

---

#### Success Criteria
- [ ] `isort` is available to run independently, ideally via a makefile command
- [ ] `isort` is automatically run as part of `pre-commit` checks
- [ ] The codebase passes CI once any auto-formatting by isort is complete

stevejalim:
## Description

This changeset:
* adds `isort` as a dev dependency
* adds a lightweight config for isort -- this results in some possibly surprising results, but I am fine with them as it will at least be stable and automatically policed for us by pre-commit and CI checks. We won't have to think about it much (or at all, if we install isort support in our code editors)
* adds `isort --check` as a code-quality test, alongside existing `black` and `flake8` checks 
* adds a pre-commit config which applies the same release isort as one of its checks
* applies isort to the entire codebase

This changeset also:
* fixes an issue with dependency clashes (soft-clashes, it seems, not fatal) resulting from the dependabot bump of urllib3 without requests being manually updated too 🤦 - see https://github.com/mozilla/bedrock/pull/10573/commits/98a52b903d82044da8e62ae1e5e9c98e58a05ad0 for that specific change

## Issue / Bugzilla link

Resolves #10570 

## Testing

* have pre-commit running/installed
* check out this branch
* rebuild your local containers fully (so that the isort is installed) and run the tests 
    `$ make pull clean build test`
* make a deliberately wrong import change, stage it and try to commit it - pre-commit will catch it, fix it (but leaving it  unstaged for git) and stop the commit, as expected
* confirm CI is happy

### DIFF
--- a/bin/run-tests.sh
+++ b/bin/run-tests.sh
@@ -2,7 +2,7 @@
 
 flake8
 black --check .
-isort . --check
+isort --check .
 moz-l10n-lint l10n/l10n-pontoon.toml
 moz-l10n-lint l10n/l10n-vendor.toml
 python manage.py lint_ftl -q


stevejalim:
@pmac 

> Can you try to update it so that Django has its own section?

Oh yeah, sure. Rob and I discussed it and I deliberately went with a lighter version (a la https://github.com/mozilla/glam/blob/main/setup.cfg), but am happy to break out Django, sure thing

Issue #10580, opened 2021-10-12
Optimise images on /firefox/privacy/book/ page


alexgibson:
## Description

The following page weighs in at a surprising 5.58MB: https://www.mozilla.org/en-US/firefox/privacy/book/

Requests are almost entirely images: 

<img width="792" alt="image" src="https://user-images.githubusercontent.com/400117/136964868-22ac94f7-6a6d-4c31-b412-465fd44b47fd.png">

Some assets, such as `privacy-book-arrow-on-face.png` weigh in at over 1MB each.

There is likely a bit we could do to make this page perform better.

- Lazy load images with `loading="lazy"`
- Reduce the overall dimensions of some images? (is it just me, or are some of them way bigger than they need to be visually?)
- Can images be compressed better?

joneskj55:
Hi @alexgibson I'd like to take this on. I have a few ideas on how to optimize the images and enhance the performance of this page.

alexgibson:
This was fixed - the page is now ~600kb on first page load 🔥

PR #10596, opened 2021-10-14
optimize images on /firefox/privacy/book/

joneskj55:
## Description
This PR attempts to optimize the images on /firefox/privacy/book. Steps taken are as follows.

1. Converted jpg & png files to webp format
2. `` tags were converted to `<amp-img>` for lazy loading of images. https://amp.dev/documentation/components/amp-img/)

## Issue / Bugzilla link
#10580

## Testing
Testing was done using Lighthouse Performance. Each step taken showed a marked improvement in the performance of the page. However, I understand that this test is not perfect and can vary depending on a lot of factors. That said, after converting the images to webp there was an 18 point performance increase. After converting the `` tags to `<amp-img>` there was a 7 point performance increase.

## Opportunities
1. Serve responsive images using `srcset`
2. Reduce the overall dimensions as mentioned in the corresponding issue #10580

### DIFF
--- a/bedrock/base/templates/macros-protocol.html
+++ b/bedrock/base/templates/macros-protocol.html
@@ -81,9 +81,9 @@
         {% if image_url.startswith('https')  %}
           <img class="mzp-c-split-media-asset" src="{{ image_url }}" alt="{{ image_alt }}" loading="{{ loading }}">
         {% elif l10n_image %}
-          <img class="mzp-c-split-media-asset" src="{{ l10n_img(image_url) }}" alt="{{ image_alt }}", loading="{{ loading }}">
+          <img class="mzp-c-split-media-asset" src="{{ l10n_img(image_url) }}" alt="{{ image_alt }}" loading="{{ loading }}">
         {% else %}
-          <img class="mzp-c-split-media-asset" src="{{ static(image_url) }}" alt="{{ image_alt }}", loading="{{ loading }}">
+          <img class="mzp-c-split-media-asset" src="{{ static(image_url) }}" alt="{{ image_alt }}" loading="{{ loading }}">
         {% endif %}
       {% endif %}
     </div>
@@ -103,11 +103,11 @@
         {{ high_res_img(image_url, {'alt': image_alt, 'l10n': l10n_image, 'class': 'mzp-c-split-media-asset', 'loading': loading}) }}
       {% else %}
         {% if image_url.startswith('https')  %}
-          <img class="mzp-c-split-media-asset" src="{{ image_url }}" alt="{{ image_alt }}", loading="{{ loading }}">
+          <img class="mzp-c-split-media-asset" src="{{ image_url }}" alt="{{ image_alt }}" loading="{{ loading }}">
         {% elif l10n_image %}
-          <img class="mzp-c-split-media-asset" src="{{ l10n_img(image_url) }}" alt="{{ image_alt }}", loading="{{ loading }}">
+          <img class="mzp-c-split-media-asset" src="{{ l10n_img(image_url) }}" alt="{{ image_alt }}" loading="{{ loading }}">
         {% else %}
-          <img class="mzp-c-split-media-asset"  src="{{ static(image_url) }}" alt="{{ image_alt }}", loading="{{ loading }}">
+          <img class="mzp-c-split-media-asset"  src="{{ static(image_url) }}" alt="{{ image_alt }}" loading="{{ loading }}">
         {% endif %}
       {% endif %}
     </div>


joneskj55:
Hi @alexgibson I'm looking at doing some resizing of these images. I was thinking to scale down the larger images to more closely align with their corresponding text. For example:

## Current
Screen Shot 2021-11-03 at 10 56 46 AM

## Proposed
Screen Shot 2021-11-03 at 10 56 54 AM


Please let me know if this lines up with what you are looking for.

alexgibson:
@joneskj55 rather than start to alter the look of the page, I think some careful optimisation for a few of the larger images might be enough? 

For example, given the nature of some of these images trying JPG might work better. Taking [this image](https://www.mozilla.org/media/img/firefox/privacy/book/privacy-healthy-internet.f4ecea641644.png), which is 429kb and converting it to a medium quality JPG, and the running it through https://tinypng.com/, I can get it down to 188kb. I think it probably still looks OK in terms of quality comparison

test2

joneskj55:
Ahh ok, thanks for the tip! I'll give this a shot instead.

joneskj55:
I took the 9 largest images and converted them to JPG and then compressed them using TinyPNG. As you stated, the change in quality is barely noticeable and still looks good.

## Results
Screen Shot 2021-11-03 at 1 46 25 PM

Please let me know if you would like me to convert & compress the other images as well.

bedrock/firefox/templates/firefox/privacy/book.html
```diff
@@ -42,7 +42,7 @@ <h2 class="c-sub-navigation-title">{{ ftl('sub-navigation-privacy') }}</h2>
 <section class="mzp-c-hero">
   <div class="mzp-l-content">
     <div class="mzp-c-hero-body">
-      <img src="{{ static('img/firefox/privacy/book/privacy-eyes.png') }}" width="200" height="200" alt="">
+      <img src="{{ static('img/firefox/privacy/book/privacy-eyes.png') }}" width="200" height="200" alt="" loading="lazy">
```
alexgibson, line 45: No need to add `loading="lazy"` to images that appear above the fold.

bedrock/firefox/templates/firefox/privacy/book.html
```diff
@@ -53,7 +53,7 @@ <h1 class="mzp-c-hero-title">{{ ftl('privacy-book-little-book-of') }}</h1>
 </section>
 
 {% call split(
-  image_url='img/firefox/privacy/book/privacy-healthy-internet.png',
+  image_url='img/firefox/privacy/book/privacy-healthy-internet.jpg',
```
alexgibson, line 56: Whilst there's definitely an improvement in these image sizes, the ones used by the `split` macro would still benefit from `loading="lazy"`. To do this you would need to add a new `loading` parameter to the `split` macro here: https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/macros-protocol.html#L75

I'd suggest making the default value for the parameter `eager`, and then individual pages can specify `lazy` when required, which can be applied to the `` elements in the macro. Let me know if you have any questions on how to do this.

alexgibson (changes requested):
Looks good! I think we still need to lazy load the images in the `split` macro, but otherwise this is already a great improvement (saving over 1MB 👍)

bedrock/firefox/templates/firefox/privacy/book.html
joneskj55: Indeed, I will remove it.

bedrock/firefox/templates/firefox/privacy/book.html
joneskj55: Ok, thanks! I'll look into this and follow up if I have any questions.

joneskj55: Got the remainder of the images loading lazily. Thanks for the tips!

Screen Shot 2021-11-04 at 10 33 55 AM

bedrock/base/templates/macros-protocol.html
```diff
@@ -79,11 +79,11 @@
         {{ high_res_img(image_url, {'alt': image_alt, 'l10n': l10n_image, 'class': 'mzp-c-split-media-asset'}) }}
```
alexgibson, line 79:
```suggestion
        {{ high_res_img(image_url, {'alt': image_alt, 'l10n': l10n_image, 'class': 'mzp-c-split-media-asset', 'loading': loading}) }}
```

Let's add this to the `high_res_img` helper too. Thanks!

alexgibson (changes requested):

bedrock/base/templates/macros-protocol.html
joneskj55: Will do.

bedrock/base/templates/macros-protocol.html
```diff
@@ -76,14 +76,14 @@
...
       {% if include_highres_image %}
-        {{ high_res_img(image_url, {'alt': image_alt, 'l10n': l10n_image, 'class': 'mzp-c-split-media-asset'}) }}
+        {{ high_res_img(image_url, {'alt': image_alt, 'l10n': l10n_image, 'class': 'mzp-c-split-media-asset', 'loading': loading}) }}
       {% else %}
         {% if image_url.startswith('https')  %}
-          <img class="mzp-c-split-media-asset" src="{{ image_url }}" alt="{{ image_alt }}">
+          <img class="mzp-c-split-media-asset" src="{{ image_url }}" alt="{{ image_alt }}" loading="{{ loading }}">
         {% elif l10n_image %}
-          <img class="mzp-c-split-media-asset" src="{{ l10n_img(image_url) }}" alt="{{ image_alt }}">
+          <img class="mzp-c-split-media-asset" src="{{ l10n_img(image_url) }}" alt="{{ image_alt }}", loading="{{ loading }}">
```
alexgibson, line 84: Please remove the commas in the `` elements, as they are not needed to separate attributes and are creating invalid markup issues.

alexgibson (changes requested):

bedrock/base/templates/macros-protocol.html
joneskj55: Oh wow, my mistake! Can't believe I missed that.

alexgibson (approved):
This looks great, thanks for all your work! 💯

PR #10608, opened 2021-10-18
Use server-side geo-location for VPN landing page (Fixes #10433)

### LINKED ISSUE #10433: Refactor VPN product page to use country code from CDN cache key
## Description

Now that we have #10429  we should be able to greatly simplify some of the geo-location login on the [VPN landing page](https://www.mozilla.org/en-US/products/vpn/).

We should be able to:

- Determine if VPN is available in the user's country directly in the view.
- Pass the country code from the view to the page template as a context variable.
- We could then use this new context variable to:
  - Set the right pricing / currency to display in the pricing section.
  - Set the right plan ID in each subscription link URL.

One factor that might be worth some consideration here when making changes is SEO. For example, we may still want to always render the pricing section in the HTML response just for crawlers, but hide / show it using a CSS class name (much in the same way we do today, but using JS). Just food for thought.

Doing this work will greatly simplify displaying pricing buttons elsewhere, like on the /whatsnew page or on other VPN pages. It should also help improve page performance as the logic will no longer need to be handled by JS on the client.

~When it comes to testing and rollout, we may want to consider making this an entirely new template behind a switch, which we can then launch once we're satisfied everything functions as before.~ (spoke to Paul here and we're fairly confident to just make this change without a switch)

---

#### Success Criteria
- [ ] VPN landing page geo-locsation logic is moved from client side to server side.

alexgibson:
## Description
- Updates VPN landing page to use server-side geo-location
- Removes FF 92 /whatsnew templates for VPN that are no longer used.
- Removes outdated experiment files.

## Issue / Bugzilla link
#10433

## Testing
- [x] VPN available (pricing USD) http://localhost:8000/en-US/products/vpn/?geo=us
- [x] VPN available (pricing EURO) http://localhost:8000/en-US/products/vpn/?geo=de
- [x] VPN available (pricing CHF) http://localhost:8000/en-US/products/vpn/?geo=ch
- [x] Not available (wait list) http://localhost:8000/en-US/products/vpn/?geo=pl

Successful test run: https://gitlab.com/mozmeao/www-config/-/pipelines/399445005


maureenlholland (approved):
r+ 🧼

alexgibson:
Thanks @maureenlholland!

@stevejalim or @pmac either of you care to take a look at the Python changes here? A second pair of eyes would always be good.

bedrock/products/templatetags/misc.py
```diff
@@ -85,36 +84,24 @@ def vpn_subscribe_link(ctx, entrypoint, link_text, plan="12-month", class_name=N
...
-            plan_id = attrs["default"][plan]["id"]
-
-        plan_attributes.update({f"data-plan-{country}": plan_id})
-
-    if plan_attributes:
-        optional_attributes = optional_attributes or {}
-        optional_attributes.update(plan_attributes)
-
-    product_url = f"{settings.VPN_SUBSCRIPTION_URL}subscriptions/products/{settings.VPN_PRODUCT_ID}?plan={plan_default}"
+    product_url = f"{settings.VPN_SUBSCRIPTION_URL}subscriptions/products/{settings.VPN_PRODUCT_ID}?plan={plan_id}"
```
stevejalim, line 97: So in theory `plan_id` could be `None` at this point (if config is absent). Is that OK?

bedrock/products/templatetags/misc.py
```diff
@@ -13,6 +13,16 @@
 FTL_FILES = ["products/vpn/shared"]
 
 
+def get_available_plans(country_code, lang):
```
pmac, line 16: Suggestion: Include "vpn" in the function name since it's specifically about plans for VPN.

bedrock/products/tests/test_helper_misc.py
```diff
@@ -119,301 +142,529 @@ def render(s, context=None):
...
+        optional_parameters=None,
+        optional_attributes=None,
+    ):
         req = self.rf.get("/")
         req.locale = "en-US"
         return render(
-            "{{{{ vpn_subscribe_link('{0}', '{1}', '{2}', '{3}', '{4}', {5}, {6}) }}}}".format(
-                entrypoint, link_text, plan, class_name, lang, optional_parameters, optional_attributes
+            "{{{{ vpn_subscribe_link('{0}', '{1}', '{2}', '{3}', '{4}', '{5}', {6}, {7}) }}}}".format(
+                entrypoint, link_text, plan, class_name, country_code, lang, optional_parameters, optional_attributes
```
stevejalim, line 160: In the future, it might be nice to use f-strings to make these easier to scan:

```
f"{{{{ vpn_subscribe_link('{entrypoint}', '{link_text}', '{plan}', '{class_name}', '{country_code}', '{lang}', {optional_parameters}, {optional_attributes}) }}}}"
```

pmac (approved):
That's a lot of good work! 🎸

bedrock/products/templatetags/misc.py
alexgibson: I guess theoreticallty that's possible yes, b ut pretty unlikely. I guess we could default to an empty string, which might be a better non-functional parameter value.

bedrock/products/tests/test_helper_misc.py
alexgibson: Ah, yeah that would be much nicer 👍

bedrock/products/templatetags/misc.py
alexgibson: After giving this some consideration, I think I'd rather have the page throw a 500 error if the `plan` is not a valid option, rather than default silently to something not so obvious value. This is easier to catch, and would almost certainly be picked up as a failure in a test run.

Issue #10544, opened 2021-10-04
Use an SQS queue for Contentful change notification


pmac:
Bedrock currently polls Contentful for data changes, but this results in much more API use and data transfer than we need. We should be updating data only when notified by a webhook or the like. 

Contentful has a "webhook" type that will instead place the messages on a queue in SQS:

https://www.contentful.com/marketplace/webhook/aws-sqs/

All we need is an SQS queue with which to work. I believe we'll need 3 queues actually, one each for dev, stage, and prod. 

This issue can cover the work of setting up the queue and switching bedrock to using it. 

Success criteria:

- [ ] DB update code (`update_contentful`) only pulls from Contentful if there's a message in the SQS queue
- [ ] Local/developer users can always get new Contentful content, regardless of SQS state via a `--force` option

pmac:
@duallain how should we go about getting the SQS work done here? Can we just make the via the AWS console, or would SRE prefer that we get this done via some automation of yours?

duallain:
Pulled this in with https://mozilla-hub.atlassian.net/browse/SE-2694. Plan would be to pull it in this week or next.  Not 100% where our commitments are, but SQS queues are pretty easy and that doc has a nice outline of what we need to give to contentful to get everything working.  

We could also use this to teach a backend dev the tf bits if you'd like, it's simple enough this would be a great training ticket.

pmac:
> We could also use this to teach a backend dev the tf bits if you'd like, it's simple enough this would be a great training ticket.

Excellent suggestion. We should involve @robhudson, @stevejalim and me.

PR #10610, opened 2021-10-18
Contentful: switch to using SQS-based signalling rather than polling API every 5 mins

stevejalim:
Switch from polling Contentful directly to checking a message queue for a sentinel

## Description

Prior to this changeset, the update_contentful management command would immediately pull all content from Contentful and then decide whether the local DB needed updating or not.
This change was run every few minutes as part of run-db-update.sh, called by a clock process, and resulted in lots of API calls to Contentful, even though content may not have changed.

This changeset now first checks an SQS message queue for a notification from Contentful (put there via a separately configured webhook at the Contentful side each time content state changes). The messages are checked and, if the indicate that a content update is worthwhile, only _then_ is the previous set of API calls triggered.

The 'go' actions vary depending on whether this is in use on Dev/Stage or Prod, to allow Dev to pull in updated draft/preview content, too. Note that Dev needs a preview-compatible API key to work in DEV=True mode

Also note that the `--force` option now does two things:
 * [ADDED] poll contentful regardless of queue state
 * [RETAINED] update stored data even if the data hash matches

Finally, the idea of going with granular updates, rather than the existing 'replace-all-the-things' approach was considered, but the exisitng approach clean-sweep gives us more reliable behaviour.

Includes tests for all the new code. Will backfill a test for the existing Contenful sync as a follow-up - want to get the changes here into real review and test-driving well ahead of Friday.

## Issue / Bugzilla link

Contributes to #10544 

## Testing

* uodate your .env with dev credentials for Contenful - these are stored with our team secrets
  * `CONTENTFUL_NOTIFICATION_QUEUE_URL`
  * `CONTENTFUL_NOTIFICATION_QUEUE_REGION`
  * `CONTENTFUL_NOTIFICATION_QUEUE_ACCESS_KEY_ID`
  * `CONTENTFUL_NOTIFICATION_QUEUE_SECRET_ACCESS_KEY`

  From Contentful
  * `CONTENTFUL_SPACE_ID`
  * `CONTENTFUL_SPACE_KEY` - use the preview key because we're running in dev mode
 
* pull and run the app container
* `make fresh-data` to get the latest data in – if you don't do this you may get unexpected behaviour (adding not updating pages)
* in contentful, make an edit to the EN US homepage - eg switch the order of a couple of cards - then publish
* shell into the container  with `docker-compose exec app bash` and run `./manage.py update_contentful` - it should **update** two page records, not add any.
* once complete, verify that your local build reflects the changes made in Contentful
* Un-make the changes in contentful
* re-run `update_contentful` and confirm the reversion is shown
* re-run `update_contentful` and confirm there are no changes to reconcile
* try `./manage.py update_contentful --force` and confirm it **updates** two pages

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -192,16 +192,6 @@ the container which you can then use to run any commands you'd like and inspect 
 
     $ docker-compose run test bash
 
-.. note::
-
-    Gotcha wanrning: the command above runs a bash shell in a fresh ``test``
-    container, which doesn't have your local .env file available to it.
-
-    If you want to run a shell on the actual running app container, to be able to
-    use your .env, specify the ``app`` service and ``exec`` not ``run``:
-
-        $ docker-compose exec app bash
-
 Local
 -----
 


stevejalim:
Just added the `Do not merge` label because we'll need SQS queues on Staging and Prod before we can let this go downstream (or we change `run-db-update.sh` to use `--force` in the meantime)

PR #10651, opened 2021-10-29
Add release notes image for Firefox 94

rvandermeulen:


craigcook:
Though the file size is pretty hefty I think it's about as small as it can be under the circumstances. r+! 🎥

craigcook (approved):

PR #10660, opened 2021-11-01
Remove old l10n machinery

robhudson:
## Description

Bedrock has moved to using Fluent for its l10n. This PR removes the unused code associated with the old l10n system.


lib/l10n_utils/__init__.py
```diff
@@ -94,28 +93,23 @@ def render(request, template, context=None, ftl_files=None, activation_files=Non
...
     context["template"] = template
-    context["langfile"] = get_l10n_path(template)
     context["template_source_url"] = template_source_url(template)
 
     # if `locales` is given use it as the full list of active translations
     if "active_locales" in context:
         translations = context["active_locales"]
         del context["active_locales"]
     else:
+        translations = [settings.LANGUAGE_CODE]
```
robhudson, line 103: I needed a sensible default since `translations_for_template(template)` got removed below. Let me know if this makes sense.

bedrock/settings/base.py
```diff
@@ -221,7 +221,6 @@ def data_path(*args):
     "zu",
 )
 
-LOCALES_PATH = DATA_PATH / "locale"
```
pmac, line 224: The next 3 lines can go as well I think

pmac: Along with the `DOTLANG_FILES` constant below

pmac:
I found a mention of the `add_lang_files` template tag in the [send-to-device docs](https://github.com/mozilla/bedrock/blob/master/docs/send-to-device.rst). @alexgibson do you know if we can just remove that step from the docs, or do we need to replace it with something Fluent-related?

alexgibson:
@pmac it looks like `send_to_device.ftl` is now a global file, so yes I think we can just remove that step. Thanks

bedrock/settings/base.py
robhudson: Done!

pmac (approved):
I think this looks ready. It's still marked as a draft, but I think we're probably ready to merge.

🍕

pmac:
You ran this one through integration tests already, right?

robhudson:
Yep, it passed integration testing.

Issue #10620, opened 2021-10-21
Homepage update for MR2


slightlyoffbeat:
## Description

A part of this Epic: #10563

Featured mobile section will be for dev.

All other changes will happen through contentful.

Copy & Design:

https://drive.google.com/open?id=17ttn6j7Y73AKZm8H19lRx8X8Mup-ZuUy&authuser=mham%40mozilla.com&usp=drive_fs

And Figma file here (includes copy):

https://www.figma.com/file/fyK3Ptys7HdQX7eDBBA5vb/MR2-Visual-Guidelines-Copy?node-id=654%3A3457

---

#### Success Criteria
- [ ] TBD

PR #10671, opened 2021-11-02
Make MR2 home promo styles more specific [#10620]

craigcook:
## Description
While looking at the Contentful preview I realized some of my styling for the MR2 promo was applying some padding to other instances of the Split component. This updates the promo styles to be more specific to just the promo so they won't mess with any other Splits on the same page.

## Testing
http://localhost:8000/en-US/
http://localhost:8000/fr/
http://localhost:8000/de/
http://localhost:8000/es-ES/

### DIFF
--- a/bedrock/mozorg/templates/mozorg/home/home.html
+++ b/bedrock/mozorg/templates/mozorg/home/home.html
@@ -18,6 +18,7 @@
 {% block body_id %}home{% endblock %}
 
 {% set referrals="?utm_source=www.mozilla.org&utm_medium=referral&utm_campaign=homepage" %}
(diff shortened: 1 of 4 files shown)


alexgibson (approved):
The changes here look fine to me. Although I did notice something looks very broken on the rest of world home page?

https://www-dev.allizom.org/es-ES/
http://localhost:8000/es-ES/

Should we fix that here, or file a separate issue?

craigcook:
Fixed rest-of-world home page. Looks like Dutch has translated the promo so you can test http://localhost:8000/nl/

alexgibson (approved):

PR #10672, opened 2021-11-02
Move de, fr to vendor supported locale

peiying2:
- Remove de, fr from Pontoon locale list
- Add de, fr to Vendor supported locale list


peiying2:
Background: we have the agreement from Anika and Anastasia to move de, fr to Smartling as vendor supported locales. They have right to make edits on the Smartling platform.

pmac (approved):
Looks good to me 💯

PR #10674, opened 2021-11-02
Update to Protocol 15.1.0

craigcook:
## Description
- Updates @mozilla-protocol/core to v15.1.0.
- Replaces Firefox Relay logo images with Protocol's assets.
- Deletes Relay logos in bedrock.

This is a really minor release of Protocol, only adding Relay logos and not really changing anything else. 

## Testing
Run `npm install` locally to get the latest package. The Relay logo currently appears in the main nav under Products and in the sticky promo on multiple pages (e.g. /firefox/browsers/). Double-check that nothing else is still referring to the files I deleted.


craigcook:
Incidentally, I updated Node/npm not long ago and now any time I run `npm install` it completely reformats package-lock.json (with a warning about an old lockfile). That may be something we need to sort out.

maureenlholland (approved):
r+ 🎩

PR #10684, opened 2021-11-03
Ignore noisy non-actionable JS errors (Fixes #10683)

### LINKED ISSUE #10683: Ignore frequent noisy JS errors in Sentry
## Description

There are a couple of JS errors we see frequently that are non-actionable and just create a lot of noise. I think we can ignore these:

- `SecurityError: Failed to execute 'open' on 'XMLHttpRequest'` https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/14432596/?query=is%3Aunresolved
- `NetworkError when attempting to fetch resource` https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/14443019/?query=is%3Aunresolved

---

#### Success Criteria
- [ ] Update Sentry ignore list to include these errors

alexgibson:
Fixes https://github.com/mozilla/bedrock/issues/10683

### DIFF
--- a/media/js/base/sentry.es6.js
+++ b/media/js/base/sentry.es6.js
@@ -22,7 +22,8 @@ if (
             dsn: sentryDsn,
             sampleRate: 0.1,
             ignoreErrors: [
-                'https://plugin.ucads.ucweb.com/api/flow/',
+                "SecurityError: Failed to execute 'open' on 'XMLHttpRequest'", // issue 10683
+                'NetworkError when attempting to fetch resource', // issue 10683
                 'Non-Error promise rejection captured with value' // issue 10380
             ],
             allowUrls: ['/media/js/', 'https://cdn-3.convertexperiments.com/'],


media/js/base/sentry.es6.js
```diff
@@ -22,7 +22,8 @@ if (
             dsn: sentryDsn,
             sampleRate: 0.1,
             ignoreErrors: [
-                'https://plugin.ucads.ucweb.com/api/flow/',
+                "SecurityError: Failed to execute 'open' on 'XMLHttpRequest'/", // issue 10683
```
maureenlholland, line 25: Do we need the slash inside quotes?
```suggestion
                "SecurityError: Failed to execute 'open' on 'XMLHttpRequest'", // issue 10683
```

maureenlholland (commented):
One syntax question on SecurityError.
For NetworkError, is it valuable to have knowledge of when mozilla requests fail? ie https://sentry.prod.mozaws.net/operations/bedrock-prod/issues/9885478/?query=is%3Aunresolved%20networkerror%20when%20attempting%20to%20fetch%20resource
If so, we could reduce the check to look for 'fluffyadsforeveryday.com'.

maureenlholland:
> If so, we could reduce the check to look for 'fluffyadsforeveryday.com' in ignoreUrls, maybe?

I don't think Sentry has a way to access this value to filter the error.
The moz-fetch error above has: TypeError https://www.mozilla.org/ja/firefox/
The fluffyads-fetch error has: TypeError dt/</</<(/media/js/sentry.28b4b6e9b0b0.js)

Is there any way to filter based on those urls?

alexgibson:
@maureenlholland yeah, I don't think we can filter based on the breadcrumb events, only the actual error. Even still, the source of these errors does differentiate - but they are almost always either noise or something unactionable (e.g. network error)

maureenlholland (approved):
r+ 🙊

